Issue #2818, opened 2019-07-10
When two deployments are pushed, the second doesn't have a corresponding DISTRIBUTED event


MiguelPires:
Related to the test proposed in https://github.com/zeebe-io/zeebe/pull/2817

PR #19637, opened 2024-06-21
feat: add list, create, edit, delete roles to Identity UI

maryarm:
## Description
Identity UI support list/create/edit/delete on Role.

## Related issues

Part of identity issue ([#2818](https://github.com/camunda-cloud/identity/issues/2818))


identity/client/src/pages/roles/modals/DeleteModal.tsx
```diff
@@ -0,0 +1,59 @@
+/*
+ * @license Identity
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under one or more contributor license
+ * agreements. Licensed under a proprietary license. See the License.txt file for more information. You may not use this
+ * file except in compliance with the proprietary license.
+ */
+
```
dlavrenuek, line 7: The license header is from old Identity. We can delete if for now or include the new license header.

identity/client/src/pages/roles/modals/DeleteModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+  };
+
+  return (
+    <Modal
+      open={open}
+      headline={t('Are you sure you want to delete the role "{{ name }}"?', {
+        name: role?.name,
+      })}
```
dlavrenuek, line 50: Descructuring `role` to `{ id, name }` can make the code look a bit better

```suggestion
  entity: { id, name },
  open,
  onClose,
  onSuccess,
}) => {
  const { t } = useTranslate();
  const { enqueueNotification } = useNotifications();

  const [callDeleteApi, { loading }] = useApiCall(deleteRole);

  const handleSubmit = async () => {
    if (role) {
      const { success } = await callDeleteApi({ id });

      if (success) {
        enqueueNotification({
          kind: "success",
          title: t("Role has been deleted."),
        });
        onSuccess();
      }
    }
  };

  return (
    <Modal
      open={open}
      headline={t('Are you sure you want to delete the role "{{ name }}"?', {
        name,
      })}
```

identity/client/src/pages/roles/modals/EditModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+import TextField from "src/components/form/TextField";
+import { useApiCall } from "src/utility/api";
+import useTranslate from "src/utility/localization";
+import { FormModal, UseEntityModalProps } from "src/components/modal";
+import { Role, updateRole } from "src/utility/api/roles";
+import EntityList from "src/components/entityList";
+import usePermissions from "src/pages/roles/modals/usePermissions";
+
+const EditModal: FC<UseEntityModalProps<Role>> = ({
+  entity: role,
```
dlavrenuek, line 18: FYI: I did not use desctucturing here because the variables `name`, `description` and `permissions` already exist, so you would need to name it like `currentName` etc... which in the end is longer than using role.name.

dlavrenuek (changes requested):
One code style suggestion from my side, looks good otherwise?

identity/client/src/pages/roles/modals/DeleteModal.tsx
```diff
@@ -27,26 +20,24 @@ const DeleteModal: FC<UseEntityModalProps<Role>> = ({
   const [callDeleteApi, { loading }] = useApiCall(deleteRole);
 
   const handleSubmit = async () => {
-    if (role) {
-      const { success } = await callDeleteApi({
-        id: role.id,
-      });
+    const { success } = await callDeleteApi({
+      id: id,
```
dlavrenuek, line 24:
```suggestion
      id,
```

dlavrenuek (approved):
One last suggestion from my side to make it perfect, but can also be merged as it is 🏅

PR #19645, opened 2024-06-21
refactor: apply auth profile only if other auth profiles are not enabled

Ben-Sheppard:
## Description
I would like to start making progress on integrating the webapps (starting with Operate) into the new central auth layer. This PR is a small refactor to the logic around when the default auth profile is applied.

Essentially I wanted to centralise that logic and remove the dependency on the Operate specific classes.

### DIFF
--- a/dist/src/main/java/io/camunda/application/initializers/WebappsConfigurationInitializer.java
+++ b/dist/src/main/java/io/camunda/application/initializers/WebappsConfigurationInitializer.java
@@ -7,6 +7,7 @@
  */
 package io.camunda.application.initializers;
 
+import static io.camunda.application.Profile.AUTH_BASIC;
 import static io.camunda.application.Profile.IDENTITY;
 import static io.camunda.application.Profile.IDENTITY_AUTH;
 import static io.camunda.application.Profile.OPERATE;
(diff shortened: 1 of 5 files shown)


dist/src/main/java/io/camunda/application/initializers/WebappsConfigurationInitializer.java
```diff
@@ -34,7 +35,7 @@ public class WebappsConfigurationInitializer
   private static final Set<String> WEBAPPS_PROFILES =
       Set.of(OPERATE.getId(), TASKLIST.getId(), IDENTITY.getId());
   private static final Set<String> LOGIN_DELEGATED_PROFILES =
-      Set.of(IDENTITY_AUTH.getId(), SSO_AUTH.getId());
+      Set.of(IDENTITY_AUTH.getId(), SSO_AUTH.getId(), AUTH_BASIC.getId());
```
houssain-barouni, line 38: @Ben-Sheppard I am trying to understand the difference between `AUTH` , `AUTH_BASIC` and `IDENTITY_AUTH`
is `AUTH_BASIC` considered as authentication handled by (delegated to) identity with some basic functionalities? 
How `Identity` webapp behaves when `AUTH` profile is used?

houssain-barouni (commented):
I had just one question about the difference between different auth profiles

dist/src/main/java/io/camunda/application/initializers/WebappsConfigurationInitializer.java
Ben-Sheppard: Ah good point, so in this case its I guess not delegated in that sense to Identity, the `AUTH_BASIC` profile relies currently on the Spring basic authentication login (the browser popup) so actually in that sense I don't think its considered delegated i.e. to the Identity service.

I will remove it from this set

houssain-barouni (approved):

PR #19657, opened 2024-06-24
feat: display detail of role in Identity UI 

maryarm:
## Description
Identity UI support retrieve on Role.

## Related issues
Part of identity issue ([#2818](https://github.com/camunda-cloud/identity/issues/2818))


identity/client/src/pages/roles/detail/RolePermissions.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const RolePermissions: FC<RoleDetailsProps> = ({ role, loading }) => {
+  const navigate = useNavigate();
+  const { t } = useTranslate();
+
+  const { availableItems } = usePermissions();
+
+  const rolePermissions = availableItems.filter((itemX) =>
+    role?.permissions?.includes(itemX.permission),
+  );
```
dlavrenuek, line 24: I was thinking why we can't just use `role?.permissions` and then understood that it is because `rolePermissions` is used in the entity table and contains the translated description. My idea of the `usePermissions` hook was to separate the reusable code used in both edit and create modals, so the usage here is not 100% intuitive. It would be cleaner to separate the functionality needed here (convert the plain list of string permissions to localized user facing data) to f.e. a separate hook that is also used inside `usePermissions` [here](https://github.com/camunda/camunda/blob/8cb11b8fc5b7d31c0f668483c508e8cbeec49ba6/identity/client/src/pages/roles/modals/usePermissions.tsx#L21). May be something like `useLocalizedPermissions`?

identity/client/src/pages/roles/modals/EditModal.tsx
```diff
@@ -25,7 +25,7 @@ const EditModal: FC<UseEntityModalProps<Role>> = ({
   const [name, setName] = useState(role.name);
   const [description, setDescription] = useState(role.description);
   const { permissions, setPermissions, onSelect, onUnselect, availableItems } =
-    usePermissions();
+    usePermissions(role?.permissions);
```
dlavrenuek, line 28: good catch

dlavrenuek (changes requested):

identity/client/src/pages/roles/modals/useAllPermissionsTranslated.tsx
```diff
@@ -5,25 +5,20 @@
...
+) => {
   const [permissions, setPermissions] = useState<string[]>(initialPermissions);
-  const availableItems: RoleListItem[] = allPermissions.map(
-    (permission: string) => ({
-      permission,
-      description: t(`${permission}.description`),
-    }),
-  );
+  const availableItems: TranslatedPermissionItem[] =
+    usePermissionsTranslated(allPermissions);
```
dlavrenuek, line 21: There is no need to specify the type TranslatedPermissionItem because it is inferred from `usePermissionsTranslated`. Generally I would never specify a type if it is inferred and if it technically not needed to specify it. Typescript will still check all types and eslint ensures that we have no `any`s

identity/client/src/pages/roles/modals/usePermissionsTranslated.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const translatedItems: TranslatedPermissionItem[] = initialPermissions.map(
+    (permission: string) => ({
+      permission,
+      description: t(`${permission}.description`),
+    }),
+  );
+  return translatedItems;
+};
+
+export type { TranslatedPermissionItem };
```
dlavrenuek, line 27: You can also just do `type TranslatedPermissionItem = {...`. But I don't think that there is a need to export the type here.

dlavrenuek (approved):
One suggestion for usage of types, but it is not mandatory and can also be merged as it is.

identity/client/src/pages/roles/modals/usePermissionsTranslated.tsx
maryarm: I'll fix them in the next PR. :)

Issue #19076, opened 2024-06-04
Provide consilidated architecture to enable and to build the v2 REST endpoints


romansmirnov:
## Description

Following the [architecture](https://docs.google.com/presentation/d/1dnnXgx7UENeyf1wYGiYPq7-YVnlTTp8814Q6veZj9P0/edit#slide=id.p), the general archicture must be implemented. That way, the teams are enabled to build and migrate the v2 REST endpoints in the Unified Camunda API.

```[tasklist]
### Search Client Layer
- [ ] https://github.com/camunda/camunda/pull/19040
- [x] https://github.com/camunda/camunda/pull/19043 - Provide an Elasticsearch-specific data storage implementation that transforms a search query in an Elasticsearch-specific search request
- [x] https://github.com/camunda/camunda/pull/19048 - Provide an Opensearch-specific data storage implementation that transforms a search query in an Elasticsearch-specific search request
```

```[tasklist]
### Camunda Service Layer
- [x] https://github.com/camunda/camunda/pull/19049 - Provide basic infrastructure to construct search queries
- [x] https://github.com/camunda/camunda/pull/19049 - Implement a process instance query (a first increment)
```

```[tasklist]
### Application Layer
- [ ] https://github.com/camunda/camunda/pull/19123
- [ ] https://github.com/camunda/camunda/pull/19136
- [ ] https://github.com/camunda/camunda/pull/19170
- [ ] https://github.com/camunda/camunda/pull/19177
- [ ] https://github.com/camunda/camunda/pull/19193
- [ ] https://github.com/camunda/camunda/pull/19419
- [ ] https://github.com/camunda/camunda/pull/19437
```

```[tasklist]
### REST Layer
- [x] Implement a process instance controller to execute a process instance search query
```

```[tasklist]
### Rollout to SaaS and SM
- [ ] https://github.com/camunda-cloud/camunda-operator/pull/2963
- [ ] https://github.com/camunda-cloud/camunda-operator/pull/3009
- [ ] https://github.com/camunda/camunda-platform-helm/pull/1995
```

```[tasklist]
### Testing
- [x] Run newly added tests in CI https://github.com/camunda/camunda/pull/19436
- [ ] Add new tests for search client
- [ ] https://github.com/camunda/camunda/pull/19682
- [ ] https://github.com/camunda/camunda/pull/19663
- [ ] https://github.com/camunda/camunda/pull/19678
```

---

related to https://github.com/camunda/product-hub/issues/2274

[...]

PR #19663, opened 2024-06-24
Add new unit tests for SearchRange queries (ES)

ChrisKujawa:
## Description

I wrote some new tests for the search range queries (ES implementation). I thought about how to do it with less effort and having a good coverage, and thought might make sense to use parameterized tests here.

So I added a parameterized test for SearchRange query (ES) that allows us to easily test and extend tests for different range queries. 

I kind of like it as it checks the actual query which will be sent to ES (it is kind of decoupled from the actual client API/usage). I would like to discuss this approach first with you @romansmirnov  before I continue here. 

## Related issues


romansmirnov (approved):
Nice!

PR #19666, opened 2024-06-24
build: make optimize releasable from the monorepo

misiekhardcore:
## Description

This PR contains changes required to make Optimize releasable from the monorepo
Branch related to this PR is used as a "main" branch in the release configuration, so the changes can be tested without any interference with the repo main branch,


optimize/pom.xml
```diff
@@ -545,8 +545,9 @@
         <configuration>
           <gitFlowConfig>
             <productionBranch>latest</productionBranch>
-            <developmentBranch>master</developmentBranch>
+            <developmentBranch>test-optimize-release</developmentBranch>
```
misiekhardcore, line 548: should I set it to main already?

PR #19667, opened 2024-06-24
ci: fix java client pom

nicpuppa:
## Description

Due to this error in the CI: https://github.com/camunda/camunda/actions/runs/9642437390/job/26591161674

Maven deploy plugin wasn't able to find the repository to deploy the artifact

### DIFF
--- a/zeebe/clients/java/pom.xml
+++ b/zeebe/clients/java/pom.xml
@@ -1,6 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <project>
   <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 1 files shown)


tmetzke (approved):
Ah damn, overlooked this 🙈 👍

zeebe/clients/java/pom.xml
```diff
@@ -1,6 +1,13 @@
 <project>
```
tmetzke, line 1: 🙃 We should add the XML opening tag here, I think.

tmetzke: `<?xml version="1.0" encoding="UTF-8"?>`

PR #19678, opened 2024-06-24
Add new tests for the SearchTermQuery (ES)

ChrisKujawa:
## Description

This PR does the following:

 * Removes some indirection (reducing complexity)
 * Fix an issue with the SearchTermQuery as the caseInsensitive value was not provided to the ES client
 * Add new parameterized test for the SearchTermQuery
 * Improve a bit on the null-handling and a separate test for it

## Related issues

### DIFF
--- a/search/search-client/src/main/java/io/camunda/search/clients/query/SearchTermQuery.java
+++ b/search/search-client/src/main/java/io/camunda/search/clients/query/SearchTermQuery.java
@@ -10,14 +10,9 @@ package io.camunda.search.clients.query;
 import io.camunda.search.clients.types.TypedValue;
 import io.camunda.util.ObjectBuilder;
 import java.util.Objects;
-import java.util.function.Function;
 
-public final record SearchTermQuery(String field, TypedValue value, Boolean caseInsensitive)
+public record SearchTermQuery(String field, TypedValue value, Boolean caseInsensitive)
     implements SearchQueryOption {
-
-  static SearchTermQuery of(final Function<Builder, ObjectBuilder<SearchTermQuery>> fn) {
-    return SearchQueryBuilders.term(fn);
-  }
 
   public static final class Builder implements ObjectBuilder<SearchTermQuery> {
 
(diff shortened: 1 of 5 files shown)


search/search-client/src/main/java/io/camunda/search/clients/query/SearchTermQuery.java
```diff
@@ -62,8 +57,13 @@ public Builder caseInsensitive(final Boolean value) {
...
     @Override
     public SearchTermQuery build() {
-      return new SearchTermQuery(
-          Objects.requireNonNull(field), Objects.requireNonNull(value), caseInsensitive);
+      Objects.requireNonNull(field);
+
+      if (value == null || value == TypedValue.NULL || value.isNull() || value.value() == null) {
+        throw new IllegalArgumentException(
+            String.format("Expected non-null value for term query, for field '%s'.", field));
+      }
```
romansmirnov, line 65: For my understanding, was there anything failing with a NPE? Because the transformer transforms the null value into `FieldValue.NULL` that the ES client can handle in a term query.

romansmirnov (commented):
@Zelldon, thanks for providing the test cases! I just have a question regarding the introduced null check.

search/search-client/src/main/java/io/camunda/search/clients/query/SearchTermQuery.java
Zelldon: Thanks @romansmirnov I think it was multiple things, at the start. For example, the value was never null, as we always create a TypedValue. I later fixed also that null strings are converted to the Null TypedValue, which we didn't before and then it failed on some unexpected places. 

I think we can remove this and I add a new parameterized test case:

romansmirnov: Yes, that makes sense to me! Basically, keep the null-check for null Strings and add a new parameterized test case.

Zelldon: I just found out that require non null https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T- has an overloaded method that supports messages, I would like to use this one over the other. So I can make sure we add some more context.

Zelldon: @romansmirnov I updated the PR

romansmirnov (approved):

PR #19679, opened 2024-06-24
ci: increase timeouts for acquiring locks during build

### LINKED ISSUE #19619: Concurrent maven builds causing failures in CI process.
**Description**

A maven build issue maybe related to [concurrent maven builds](https://issues.apache.org/jira/browse/MNG-7868), is blocking a both normal PRs and back porting https://github.com/camunda/camunda/actions/runs/9602339244/job/26482978516?pr=19599. Sometimes the fix is to delete the cache that the built process restores from, however this doesn't always fix it.

deepthidevaki:
## Description

This is an attempt to workaround the frequent failures caused by 'Could not acquire read lock for ..' 
This change helped to merge #19613. I'm not sure if it really improves the situation, though. But makes sense to apply this changes to main and other stable branches.

## Related issues

closes #19619

### DIFF
--- a/.github/actions/setup-zeebe/action.yml
+++ b/.github/actions/setup-zeebe/action.yml
@@ -137,6 +137,7 @@ runs:
         -D aether.enhancedLocalRepository.splitRemote=true
         -D aether.syncContext.named.nameMapper=file-gav
         -D aether.syncContext.named.factory=file-lock
+        -D aether.syncContext.named.time=120
         -D maven.artifact.threads=32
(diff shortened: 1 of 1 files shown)


lenaschoenburg:
Should be backported to 8.5 too unless I'm missing something

EuroLew (approved):
LGTM.

deepthidevaki:
stable/8.5 already has this fix. No need to backport.

PR #19684, opened 2024-06-24
fix: add name to User's UI

maryarm:
## Description
As identity backend now support Name attribute for the User, UI should also handle it.

## Related issues
Part of identity issue ([#2818](https://github.com/camunda-cloud/identity/issues/2971))


dlavrenuek (approved):
Perfect 💯

PR #19686, opened 2024-06-24
build: rollback java version to 17

### LINKED ISSUE #19658: Spring SDK 8.5.3 requires JDK 21+
**Describe the bug**

The Spring SDK's [version compatibility](https://docs.camunda.io/docs/next/apis-tools/spring-zeebe-sdk/getting-started/#version-compatibility) is documented to require JDK 17+. However, since 8.5.2/8.5.3, it requires JDK 21+.

This may be a documentation issue if not a defect.

**To Reproduce**

- use spring SDK in a project
- build the project with JDK 17
- note the following error:

>has been compiled by a more recent version of the Java Runtime (class file version 65.0), this version of the Java Runtime only recognizes class file versions up to 61.0

**Expected behavior**

Spring SDK 8.5.x should require JDK 17+ only. Or the documentation should be adjusted.

nicpuppa:
## Description

Spring Zeebe SDK should be builded with Java 17, according to our [docs](https://docs.camunda.io/docs/next/apis-tools/spring-zeebe-sdk/getting-started/#version-compatibility)

## Related issues

closes #19658

### DIFF
--- a/.github/workflows/spring-boot-starter-camunda-sdk-ci.yml
+++ b/.github/workflows/spring-boot-starter-camunda-sdk-ci.yml
@@ -42,14 +42,21 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v4
         with:
-          java-version: '21'
-          distribution: 'adopt'
+          java-version: 21
+          distribution: temurin
       - name: Create build output log file
         run: echo "BUILD_OUTPUT_FILE_PATH=$(mktemp)" >> $GITHUB_ENV
       - name: Build Java Client
         uses: ./.github/actions/build-zeebe
         with:
           maven-extra-args: -T1C -am
+
+      - name: Downgrade Java environment
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+
       - name: Maven Test Build
         # we use the verify goal here as flaky test extraction is bound to the post-integration-test
         # phase of Maven https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#default-lifecycle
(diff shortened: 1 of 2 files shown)


.github/workflows/spring-boot-starter-camunda-sdk-ci.yml
```diff
@@ -42,14 +42,21 @@ jobs:
...
+          java-version: 21
+          distribution: temurin
       - name: Create build output log file
         run: echo "BUILD_OUTPUT_FILE_PATH=$(mktemp)" >> $GITHUB_ENV
       - name: Build Java Client
         uses: ./.github/actions/build-zeebe
         with:
           maven-extra-args: -T1C -am
+
+      - name: Downgrade Java environment
```
nicpuppa, line 54: 💭 This step might be useless

nicpuppa:
Hey @korthout, do we have a test set-up to verify/check the builds with different java versions ? I was looking into something related to the `java-client` but I didn't found anything

korthout:
> I was looking into something related to the java-client but I didn't found anything

Wait. I'm confused about this. I can't find it either 😕

.github/workflows/spring-boot-starter-camunda-sdk-ci.yml
korthout: 👍 Let's remove it

korthout (approved):
Thanks @nicpuppa 

🔧 Let's remove the `Downgrade Java environment` step.

LGTM 👍

nicpuppa:
@korthout should we forward port this ? Right now all the 8.6 alphas are affected by this

korthout:
@nicpuppa Is there any reason for 8.6 to require 21+? If not, let's go for the lowest possible requirement ourselves. That would also help us to avoid having to change the docs for 8.6.

korthout:
@nicpuppa There's no need to backport it, I think:

https://github.com/camunda/camunda/pull/19692#issuecomment-2186927151

PR #19625, opened 2024-06-21
feat: implement interfaces and dtos user task search 

nathansandi:
## Description

This is the first of 2 PRs of the implementation of Search User Task API. The goal of this PR is prepare the interfaces and implementation, plus the controller for User Task Search 

The continuation of the implementation:

2) https://github.com/camunda/camunda/pull/19691 - Integrate the Variables on the User Task Search Service and Controller 

The PRs were divided in order to make it easier the review process.

Test example:
```
{
    "filter": {
        "taskState":"CREATED",
        "processInstanceKey": 2251799813685379,
    },
    "sort":[
        {
            "field":"creationTime",
            "order":"desc"
        }
    ],
    "page": {
        "limit": 20
    }
}
```

## Related issues

closes #https://github.com/camunda/camunda/issues/19211

PR #19691, opened 2024-06-24
feat: Integrate variable filter over user task services [wip]

nathansandi:
## Description

This is the third of three PRs for the implementation of the Search User Task API. The goal of this PR is to integrate the Variables on the User Task Search Service and Controller. This completes the setup necessary for comprehensive end-to-end tests of the User Task Search functionality.

Note: Elastic Search does not support join between 2 different indices

https://discuss.elastic.co/t/joining-two-indexes-with-common-field-values/332861

So I am filtering on the follow way:
1) Filter the variables by name/value and collecting the ScopeKeys and processKey that the variable belongs
2) Use this as an input to User Task Search to filter by flowInstanceID and processInstanceKey
From this, I can bring only tasks that contains the variables (either process/task variable)

Note: It is not possible Create a Nested object to Tasklist Task on the current state, because it would require data migration, and this is no longer available on Tasklist. In case we change the solution and go over Nested object, only new tasks would be supported. This is also the reason I have chosen go over this solution

If in case I have added Process/Task Id filter, this will be merged with the returned result from the variable results, and the keys that does not belongs on the search will be filtered out.

The overall implementation is divided into three PRs to facilitate the review process:

1. [PR 1](https://github.com/camunda/camunda/pull/19625) - Prepare the interfaces and implementation, and set up the preparation for the API Mapper and Controller for the User Task Context.
3. **PR 2 (this PR)** - Integrate the Variables on the User Task Search Service and Controller.

From this PR, it is already possible to test the User Task Search solely via the API.

For example:
```json
{
    "filter": {
        "taskState":"CREATED",
        "processInstanceKey": 2251799813685379,
        "variables":[
            {
                "name":"anyTest2",
                "eq":"\"test2\""
            }
        ]
    },
    "sort":[
        {
            "field":"creationTime",
            "order":"desc"
        }
    ],
    "page": {
        "limit": 20
    }
}
```

This also works on User Task Static Builder 

closes: #https://github.com/camunda/tasklist/issues/4967


service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+        originalFilter.created(),
+        originalFilter.completed(),
+        originalFilter.canceled(),
+        originalFilter.creationDateFilter(),
+        originalFilter.completionDateFilter(),
+        originalFilter.dueDateFilter(),
+        originalFilter.followUpDateFilter(),
+        originalFilter.variableFilters(),
+        originalFilter.tenantIds());
+  }
```
romansmirnov, line 128: :x: This will cause issues and must always be adjusted whenever the filter gets extended or whatever. Instead, let's try to move it to the filter and introduce a first concept of allowing to "extend/adjust" a filter, for example:

```java
public final record UserTaskFilter(
  ...) implements FilterBase {

  public static Builder extending(final UserTaskFilter filterToExtend) {
    return new Builder(filterToExtend);
  }

  public static final class Builder implements ObjectBuilder<UserTaskFilter> {


    public Builder() {}
    private Builder(final UserTaskFilter filter) {
      this.userTaskKeys = filter.userTaskKeys();
      ...;
    }
  }
}
```

That way, at this place, it could do the following

```java
final var newFilter = UserTaskFilter.extending(query.filter()).processInstanceKey(allProcessInstanceIds).build();
return UserTaskQuery.of(b -> b.filter(newFilter).sort(query.sort()).page(query.page()));
```

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+      final UserTaskFilter mergedFilter =
+          getUserTaskFilter(query, allProcessInstanceIds, allScopeKeys);
+
+      // Create a new UserTaskQuery with the merged filter
+      final UserTaskQuery modifiedQuery =
+          UserTaskQuery.of(b -> b.filter(mergedFilter).sort(query.sort()).page(query.page()));
+
+      return executor.search(modifiedQuery, UserTaskEntity.class);
+    }
     return executor.search(query, UserTaskEntity.class);
```
romansmirnov, line 95: :x: Please adjust the code in a way that the statement `executor.search(...)` is only there once.

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+        originalFilter.userTaskDefinitionIds(),
+        originalFilter.processNames(),
+        originalFilter.assignees(),
+        originalFilter.userTaskState(),
+        originalFilter.processInstanceKeys() != null
+                && originalFilter.processInstanceKeys().isEmpty()
+            ? allProcessInstanceIds
+            : originalFilter.processInstanceKeys(),
+        originalFilter.processDefinitionKeys(),
+        allScopeKeys,
```
romansmirnov, line 116: :x: What if the user provided `scopeKeys`?

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+  private void processVariableFilter(
+      final VariableValueFilter variableValueFilter,
+      final VariableServices variableServices,
+      final Map<Long, Set<Long>> scopeKeyMapByProcessInstanceKey,
+      final boolean firstIteration) {
+    final VariableFilter.Builder variableFilterBuilder = new VariableFilter.Builder();
+    variableFilterBuilder.variable(variableValueFilter);
+
+    final SearchQueryPage searchQueryPage =
+        new SearchQueryPage.Builder().size(MAX_ELASTICSEARCH_PAGE_SIZE).build();
```
romansmirnov, line 144:
```suggestion
    final var searchQueryPage = SearchQueryPageBuilders.page((p) -> p.size(MAX_ELASTICSEARCH_PAGE_SIZE));
```

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
     return search(SearchQueryBuilders.userTaskSearchQuery(fn));
   }
+
+  private void processVariableFilter(
+      final VariableValueFilter variableValueFilter,
+      final VariableServices variableServices,
+      final Map<Long, Set<Long>> scopeKeyMapByProcessInstanceKey,
+      final boolean firstIteration) {
+    final VariableFilter.Builder variableFilterBuilder = new VariableFilter.Builder();
+    variableFilterBuilder.variable(variableValueFilter);
```
romansmirnov, line 141:
```suggestion
    final var variableFilterBuilder = FilterBuilders.variable((v) -> v.variable(variableValueFilter));
    variableFilterBuilder.variable(variableValueFilter);
```

romansmirnov (changes requested):
:x: The PR requires adjustments and improvements, because it is not easy to follow and to understand what is actually happening. I am not done though, but still wanted to leave the first bunch of feedback.

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+      updateScopeKeyMap(scopeKeyMapByProcessInstanceKey, currentScopeKeyMapByProcessInstanceKey);
+    }
+  }
+
+  private void initializeScopeKeyMap(
+      final Map<Long, Set<Long>> scopeKeyMapByProcessInstanceKey,
+      final Map<Long, Set<Long>> currentScopeKeyMapByProcessInstanceKey) {
+    for (final Map.Entry<Long, Set<Long>> entry :
+        currentScopeKeyMapByProcessInstanceKey.entrySet()) {
+      final Long processInstanceId = entry.getKey();
```
romansmirnov, line 181: :x: Please change to `processInstanceKey`, we should use the same terms for the same things all over the place to avoid confusions.

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+                      : new HashSet<>();
+              final Set<Long> currentScopeKeys =
+                  currentScopeKeyMapByProcessInstanceKey.get(processInstanceId);
+
+              if (currentScopeKeys != null) {
+                validScopeKeys.retainAll(currentScopeKeys);
+              }
+
+              if (entry.getValue() == null || validScopeKeys.isEmpty()) {
+                return currentScopeKeys == null || currentScopeKeys.isEmpty();
```
romansmirnov, line 216: ❓ To my understanding, `currentScopeKeys` will always contain at least one key which would be the `processInstanceKey`. Basically, this will never return `true`. Why is this check then necessary?

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+
+  private void initializeScopeKeyMap(
+      final Map<Long, Set<Long>> scopeKeyMapByProcessInstanceKey,
+      final Map<Long, Set<Long>> currentScopeKeyMapByProcessInstanceKey) {
+    for (final Map.Entry<Long, Set<Long>> entry :
+        currentScopeKeyMapByProcessInstanceKey.entrySet()) {
+      final Long processInstanceId = entry.getKey();
+      final Set<Long> validScopeKeys =
+          entry.getValue().stream()
+              .filter(scopeKey -> !scopeKey.equals(processInstanceId))
```
romansmirnov, line 184: ❓ This filter is never applied to the `currentScopeKeyMapByProcessInstanceKey`. Why?

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+    scopeKeyMapByProcessInstanceKey
+        .entrySet()
+        .removeIf(
+            entry -> {
+              final Long processInstanceId = entry.getKey();
+              final Set<Long> validScopeKeys =
+                  entry.getValue() != null
+                      ? entry.getValue().stream()
+                          .filter(Objects::nonNull)
+                          .collect(Collectors.toSet())
```
romansmirnov, line 206: ❓ The class `CollectionUtil` already contains helpers like `withoutNullValue()`, could this be reused? Probably it would need to be extended accordingly.

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+      final Map<Long, Set<Long>> currentScopeKeyMapByProcessInstanceKey) {
+    scopeKeyMapByProcessInstanceKey
+        .entrySet()
+        .removeIf(
+            entry -> {
+              final Long processInstanceId = entry.getKey();
+              final Set<Long> validScopeKeys =
+                  entry.getValue() != null
+                      ? entry.getValue().stream()
+                          .filter(Objects::nonNull)
```
romansmirnov, line 205: ❓ When can this be `null`?

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+        firstIteration = false;
+      }
+
+      // Collect all unique processInstanceIds and scopeKeys
+      final List<Long> allProcessInstanceIds =
+          new ArrayList<>(scopeKeyMapByProcessInstanceKey.keySet());
+      final List<Long> allScopeKeys =
+          scopeKeyMapByProcessInstanceKey.values().stream()
+              .flatMap(set -> Optional.ofNullable(set).orElse(Collections.emptySet()).stream())
+              .filter(Objects::nonNull)
```
romansmirnov, line 79: :question: What happens here? Why is `filter(Objects::nonNull)` not sufficient?

romansmirnov: Okay, I think I got it. The `flatMap` ensures that a set is returned, and the `filter` filters values inside the set.

service/src/main/java/io/camunda/service/UserTaskServices.java
romansmirnov: What I mean is, once `firstIteration` is set to `false`, `initializeScopeKeyMap()` is not called anymore, hence, the filter is not applied anymore.

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+
+              if (currentScopeKeys != null) {
+                validScopeKeys.retainAll(currentScopeKeys);
+              }
+
+              if (entry.getValue() == null || validScopeKeys.isEmpty()) {
+                return currentScopeKeys == null || currentScopeKeys.isEmpty();
+              }
+
+              return validScopeKeys.isEmpty()
```
romansmirnov, line 219: ❓ Why is this checked again?

service/src/main/java/io/camunda/service/UserTaskServices.java
```diff
@@ -40,11 +58,176 @@ public UserTaskServices withAuthentication(final Authentication authentication)
...
+
+    currentScopeKeyMapByProcessInstanceKey.forEach(
+        (processInstanceId, currentScopeKeys) ->
+            scopeKeyMapByProcessInstanceKey.merge(
+                processInstanceId,
+                currentScopeKeys,
+                (oldSet, newSet) -> {
+                  oldSet.addAll(newSet);
+                  return oldSet;
+                }));
```
romansmirnov, line 231: ❓ Why must it be merged at all? Shouldn't it be enough to remove `processInstanceKeys` and `scopeKeys` from `scopeKeyMapByProcessInstanceKey`?

romansmirnov: Having the following case

```
pi(1) ---> fi(2) --> var: foo==123
    |
   +------> fi(3) --> var: bar == 456
```

Now doing the following user task query:
```json
{"filter": { "variable": [ {"name": foo", "eq": 123 }, {"name": bar", "eq": 456 }] } }
```

According the logic, with the `firstIteration` it will execute a variable query with `{"name": foo", "eq": 123 }`, which will return `pi(1) ---> fi(2) --> var: foo==123`. Basically, the variable `foo`:
```json
{
  "processInstanceKey": 1,
  "scopeKey": 2,
  "name": "foo",
  "value": 123
}
```

The `scopeKeyMapByProcessInstanceKey` looks like this: `{ 1: [2] }`

In the next iteration, it will combine the variable value filter `{"name": bar", "eq": 456 }` with the process instance key `1`
```json
{"filter": { "variable": [ {"name": bar", "eq": 456 }], "processInstanceKeys": 1 } }
```

It will return
```json
pi(1) ---> fi(3) --> var: bar ==456
```

The `currentScopeKeyMapByProcessInstanceKey` looks like this: `{ 1: [3] }`

Based on this, it `removeIf()`, the `scopeKeyMapByProcessInstanceKey` will be changed to `{ 1: [] }` (because `[2]` and `[3]` do not intersect. And executing the `merge()`, it will merge the `scopeKeyMapByProcessInstanceKey` will be merged to `{ 1: [3] }`

service/src/main/java/io/camunda/service/UserTaskServices.java
romansmirnov: Okay, it is not necessarily the `processInstanceKey`, but IMO it cannot be empty or `null`

service/src/main/java/io/camunda/service/UserTaskServices.java
romansmirnov: As a result, it will return the user task (that belongs to flownode instance `3`) which would not be expected.

romansmirnov:
I think, I am done for now with the review. Trying to summarize:

* Extend the search-client accordingly to avoid leaking ES/OS-specificas in the service layer
* Code-improvements for easier understanding and maintainability 
* Missing test cases
* The intersection logic requires clarification

In general, I don't think this is a viable solution because it is insufficient, and IMO does not scale with a growing data set (including process instances, variables, and tasks). Also, it may become tricky when we start to support `or` queries because it requires a "union" logic instead of the intersection logic.

PR #2946, opened 2019-08-07
update go client and zbctl with long polling timeout

deepthidevaki:
## Description

make long polling timeout configurable in go client and zbctl

## Related issues

closes #2931

## Pull Request Checklist

- [x] All commit messages match our [commit message guidelines](https://github.com/zeebe-io/zeebe/blob/develop/CONTRIBUTING.md#commit-message-guidelines)
- [x] The submitting code follows our [code style](https://github.com/zeebe-io/zeebe/wiki/Code-Style)
- [x] If submitting code, please run `mvn clean install -DskipTests` locally before committing

PR #19693, opened 2024-06-24
feat: add role management for a user to identity's UI

maryarm:
## Description
We need to be able to view/assign/unassign roles to a User in UI.

## Related issues
Part of identity issue ([#2946](https://github.com/camunda-cloud/identity/issues/2946))


dlavrenuek (approved):
Thank you for the implementation, looks good 🚀

PR #19694, opened 2024-06-24
ci: fix actions/paths-filter condition expressions from bool to string

cmur2:
## Description

Address the problem reported [via Slack](https://camunda.slack.com/archives/C071KP5BTHB/p1719235453069919) where `java-unit-tests` did not get executed although Java files where touched.

Works now with only Java code changes and without GHA changes: https://github.com/camunda/camunda/actions/runs/9649777741?pr=19694


lenaschoenburg (approved):
Thanks!

PR #19721, opened 2024-06-25
ci: disable integration tests from QA in java-unit-tests job

cmur2:
## Description

As discussed apparently some ITs get executed in https://github.com/camunda/camunda/actions/runs/9658623190/job/26640217174 which is unexpected given the job name.


ChrisKujawa (approved):
👍🏼

PR #19723, opened 2024-06-25
Merge optimize in  artifact not found

buccarel:
## Description

This fixes the following workflow:
* `Operate E2E Tests / test (pull_request)`
* `Tasklist E2E Tests / test (pull_request)`

There are 2 problems:
1) for some reason, the variable spring-boot.version is not being resolved in the CI environment
2) the optimize build is not compatible with the operate e2e environment

Problem (1) is resolved by explicitly stating the unresolved variable with the correct version.
Problem (2) is resolved by excluding optimize from the e2e build of operate

Even if optimize is being excluded, the fix for (1) is still part of the PR because when eventually the 2 modules will be compatible, problem (1) could happen again.


.github/workflows/operate-e2e-tests.yml
```diff
@@ -17,6 +17,8 @@ on:
       - "parent/*"
       - "pom.xml"
   pull_request:
+    branches:
+      - "merge-optimize-in"
```
buccarel, line 21: I'll remove this before merging. I needed it to activate the workflon on pull request against the branch

.github/workflows/operate-e2e-tests.yml
```diff
@@ -115,7 +117,9 @@ jobs:
...
+        # Currently, the e2e environment of operate conflicts with the optimize build. For the moment,
+        # we're excluding optimize from the build, not to impact this operate's workflow.
+        run: mvn clean install -T1C -DskipChecks -P -docker,skipFrontendBuild -DskipTests=true -B -DskipRemoteStaging=true -Dmaven.deploy.skip=true -pl '!optimize,!optimize-distro,!optimize/client,!optimize/upgrade,!optimize/backend,!optimize/util,!optimize/util/dependency-doc-creation,!optimize/util/optimize-reimport-preparation,!optimize/util/optimize-commons,!optimize/plugins,!optimize/plugins/plugin,!optimize/plugins/optimize-test-plugins/optimize-main-test-plugin,!optimize/plugins/optimize-test-plugins/optimize-pluginloading-test-plugin,!optimize/qa,!optimize/qa/engine-it-plugin,!optimize/qa/import-performance-tests,!optimize/qa/data-generation,!optimize/qa/upgrade-tests,!optimize/qa/connect-to-secured-es-tests,!optimize/qa/cleanup-performance-tests,!optimize/qa/clustering-tests,!optimize/qa/optimize-data-generator,!optimize/qa/import-mediator-permutation-test'
```
buccarel, line 122: I tried to find a way to exclude optimize recursively, but I couldn't find any, so I exluded all of the submodules.

I also tried to put this more clean on multiple lines, but I couldn't make it work because the -pl argument takes a single-quoted string that is not working well with the yaml multiline parsing

.github/workflows/operate-e2e-tests.yml
PHWaechtler: Do we have any mechanism to make sure we dont forget?

.github/workflows/operate-e2e-tests.yml
PHWaechtler: Does this mean operate currently cannot be build (outside of CI) because Optimize interferes with it? If so we should fix that instead of excluding I think

.github/workflows/operate-e2e-tests.yml
buccarel: Yes, I am the mechanism: I'll do it right after the  approval of this :P

.github/workflows/operate-e2e-tests.yml
buccarel: No, they can build. The problem is that, in this workflow, they are not only building, they are also instantiating containers. These containers clash with OUR build of optimize-upgrade, which has the instantiation of containers in the build goal.

There are multiple ways to fix this, that require more or less cooperation with the other teams.

I think that for the moment the priority is to make the PR green, then any other cooperation can be done afterwards. I think this because this is their test, doesn't have to work for an extra module like optimize.

Unless there's some other reason I'm not aware of, this is what I understood

HeleneW-dot (approved):

PR #19730, opened 2024-06-25
Add ids query test

ChrisKujawa:
## Description

* Simplify implementation - remove indirection; use different method to construct list
* Make SearchIdsQuery more null tolerant, default to empty lists/values
* Add paremeterized tests covering several ways for nulls and adding values to the query

## Related issues

### DIFF
--- a/search/search-client-elasticsearch/src/test/java/io/camunda/search/es/transformers/query/IdsQueryTransformerTest.java
+++ b/search/search-client-elasticsearch/src/test/java/io/camunda/search/es/transformers/query/IdsQueryTransformerTest.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.search.es.transformers.query;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import co.elastic.clients.elasticsearch._types.query_dsl.Query;
+import io.camunda.search.clients.query.SearchQuery;
+import io.camunda.search.clients.query.SearchQueryBuilders;
+import io.camunda.search.es.transformers.ElasticsearchTransformers;
+import io.camunda.search.transformers.SearchTransfomer;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class IdsQueryTransformerTest {
+
+  private final ElasticsearchTransformers transformers = new ElasticsearchTransformers();
+  private SearchTransfomer<SearchQuery, Query> transformer;
+
+  @BeforeEach
+  public void before() {
+    transformer = transformers.getTransformer(SearchQuery.class);
+  }
+
+  private static Stream<Arguments> provideQueries() {
+    return Stream.of(
+        Arguments.arguments(SearchQueryBuilders.ids(List.of()), "Query: {'ids':{'values':[]}}"),
+        Arguments.arguments(
+            SearchQueryBuilders.ids().values(List.of()).build().toSearchQuery(),
+            "Query: {'ids':{'values':[]}}"),
+        Arguments.arguments(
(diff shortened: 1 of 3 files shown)


search/search-client-elasticsearch/src/test/java/io/camunda/search/es/transformers/query/IdsQueryTransformerTest.java
```diff
@@ -0,0 +1,97 @@
...
+        Arguments.arguments(
+            SearchQueryBuilders.ids().values((String[]) null).build().toSearchQuery(),
+            "Query: {'ids':{'values':[]}}"),
+        Arguments.arguments(
+            SearchQueryBuilders.ids(List.of("value")), "Query: {'ids':{'values':['value']}}"),
+        Arguments.arguments(
+            SearchQueryBuilders.ids("value1", "value2"),
+            "Query: {'ids':{'values':['value1','value2']}}"),
+        Arguments.arguments(
+            SearchQueryBuilders.ids((Collection<String>) List.of("value1", "value2")),
```
houssain-barouni, line 61: I would have tested with a `Set` to avoid casting, but why not, it is just a test :)

search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
```diff
@@ -7,37 +7,30 @@
...
+public record SearchIdsQuery(List<String> values) implements SearchQueryOption {
 
   public static final class Builder implements ObjectBuilder<SearchIdsQuery> {
 
-    private List<String> ids;
+    private final List<String> ids = new ArrayList<>();
 
     public Builder values(final List<String> values) {
-      ids = addValuesToList(ids, values);
+      ids.addAll(Objects.requireNonNullElse(values, List.of()));
```
houssain-barouni, line 23: null collections are evil, we should forbid that everywhere :)

search/search-client-elasticsearch/src/test/java/io/camunda/search/es/transformers/query/IdsQueryTransformerTest.java
Zelldon: Jep I did this before :) but the problem is that the ordering is not deterministic :)

houssain-barouni: `LinkedHashSet` but it is not fun to initiate

search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
Zelldon: I guess we could with extra checkers maybe? 🤔 I think intellij has some

Zelldon: Or we just use kotlin 😯

search/search-client-elasticsearch/src/test/java/io/camunda/search/es/transformers/query/IdsQueryTransformerTest.java
Zelldon: Yeah I felt not worth it :D

Zelldon: Wait I could do `Sets.newLinkedHashSet(1, 2)`

search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
houssain-barouni, line 23: @Zelldon there will be a change of behaviour here
**before** 
`new Builder().values(List.of("a")).values(List.of("b"))` gives `ids` with `["b"]`
**now**
`new Builder().values(List.of("a")).values(List.of("b"))` gives `ids` with `["a", "b"]`

I think it is not expected to invoke many time `Builder.values` but I am wondering if this is a desired result 🤔 ?

Zelldon: Hm 🤔 are you sure? I thought that was intended to have a combined list, this is how I understood the code of `addValuesToList` (and the name :D )

houssain-barouni (approved):
LGTM!

search/search-client-elasticsearch/src/test/java/io/camunda/search/es/transformers/query/IdsQueryTransformerTest.java
Zelldon: Interesting. Was from assertj but also exists in google common. But I guess we can also just leave it as it is.

PR #19736, opened 2024-06-25
Add constant query test (ES)

ChrisKujawa:
## Description

- Simplify SearchHasChildQuery, remove some unused/complexity
- Add descriptive error messages for null checks
- Add parameterized tests for has child query and transformer

## Related issues

### DIFF
--- a/search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
+++ b/search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
@@ -9,9 +9,7 @@ package io.camunda.search.clients.query;
 
 import io.camunda.util.CollectionUtil;
 import io.camunda.util.ObjectBuilder;
-import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 
 public record SearchIdsQuery(List<String> values) implements SearchQueryOption {
 
@@ -25,7 +23,7 @@ public record SearchIdsQuery(List<String> values) implements SearchQueryOption {
     }
 
     public Builder values(final String... values) {
-      return values(Arrays.stream(Objects.requireNonNullElse(values, new String[0])).toList());
+      return values(CollectionUtil.collectValuesAsList(values));
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


houssain-barouni (approved):
LGTM!

search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
```diff
@@ -7,37 +7,30 @@
...
+public record SearchIdsQuery(List<String> values) implements SearchQueryOption {
 
   public static final class Builder implements ObjectBuilder<SearchIdsQuery> {
 
-    private List<String> ids;
+    private final List<String> ids = new ArrayList<>();
 
     public Builder values(final List<String> values) {
-      ids = addValuesToList(ids, values);
+      ids.addAll(Objects.requireNonNullElse(values, List.of()));
```
romansmirnov, line 23: I know this has already been merged, but I still want to leave my feedback here.

IMO, this handling should be "centralized" in a static method or similar, as there are multiple cases where a list of values can be provided. That way, we would have one way of handling these cases and would (to some degree) prevent others from implementing it differently.

For example, we could adjust the `#addValueToList()` to the following:

```
  public static <T> List<T> addValuesToList(final List<T> list, final List<T> values) {
    final List<T> result = Objects.requireNonNullElse(list, List.of());
    result.addAll(values);
    return result;
  }
```

Which could be used that way here:

```
public Builder values(final List<String> values) {
  ids = addValuesToList(ids, values);
}
```

search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
```diff
@@ -7,37 +7,30 @@
...
       return this;
     }
 
-    public Builder values(final String value, final String... values) {
-      return values(collectValues(value, values));
+    public Builder values(final String... values) {
+      return values(Arrays.stream(Objects.requireNonNullElse(values, new String[0])).toList());
```
romansmirnov, line 28: The same as above, I would rather suggest extracting `Arrays.stream(...)` into a centralized "helper" method so that it can be used in different places as well.

search/search-client/src/main/java/io/camunda/search/clients/query/SearchIdsQuery.java
Zelldon: Agree. I think when i first wrote it i actually wanted to hear your thoughts first about it and also regarding the other method

Zelldon: @romansmirnov something like this https://github.com/camunda/camunda/pull/19736/commits/1cc6052c2bc4b219c9d24fce31283bf59a31f9a4 ? 🤔

PR #19739, opened 2024-06-25
Do not evaluate target boundary event correlation key expression on migration

### LINKED ISSUE #19527: Do not evaluate target boundary event correlation key expression on migration
**Description**

While migrating message subscriptions (or even when they are not mapped), we still require a variable for evaluating target element's correlation key even if it won't change after migration (not yet supported). If a variable is not provided for target element boundary event's correlation key, the migration fails with a failed evaluation expression. See: https://github.com/camunda/camunda/blob/7d6a4cc17334272a20801c65143ac5867029e2a9/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java#L175

It is not a good UX since user's should be forced to provide a variable for the correlation key that will not be migrated at all. As a solution, we should prevent evaluating expressions on target boundary events.

berkaycanbc:
## Description
A filter added to `CatchEventBehavior.subscribeToEvents` method to filter out events of an element instance before executing any expression evaluations. This is used to prevent evaluating expressions on events that are migrated. Because, we are currently not supporting expression evaluation on migration. This will become optional (e.g. allowing correlation key or message name re-evaluation) in further migration improvements.

## Related issues
closes #19527

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java
@@ -148,7 +148,7 @@ public final class CatchEventBehavior {
    */
   public Either<Failure, Void> subscribeToEvents(
       final BpmnElementContext context, final ExecutableCatchEventSupplier supplier) {
-    return subscribeToEvents(context, supplier, catchEvent -> true);
+    return subscribeToEvents(context, supplier, executableCatchEvent -> true, catchEvent -> true);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java
```diff
@@ -161,19 +161,26 @@ public Either<Failure, Void> subscribeToEvents(
    * @param context the context of the element instance that subscribes to events
    * @param supplier the supplier of catch events to subscribe to, typically the element of the
    *     element instance that subscribes to events
-   * @param filter the filter for catch events to subscribe to, only events that match the filter
-   *     are subscribed to.
+   * @param filterBeforeEvaluation the filter apply for catch events before applying any
+   *     evaluations. This is especially useful for filtering catch events that doesn't require an
+   *     expression evaluation.
+   * @param filterAfterEvaluation the filter for catch events to subscribe to, only events that
+   *     match the filter are subscribed to.
```
korthout, line 168: 🔧 Let's align the descriptions of the parameters

```suggestion
   * @param filterBeforeEvaluation the filter for catch events to subscribe to. Only events that
   *     match the filter are subscribed to. This filter is applied before evaluating the catch
   *     event's expressions. This is especially useful for filtering catch events that doesn't
   *     require an expression evaluation.
   * @param filterAfterEvaluation the filter for catch events to subscribe to. Only events that
   *     match the filter are subscribed to. This filter is applied after evaluating the catch
   *     event's expressions.
```

korthout (approved):
Nice solution @berkaycanbc 🚀 

💭 To highlight that the changes to these tests verify the bug, you could first commit the test changes to show that they fail without the production code changes. This makes it easy to verify the PR by hand. I worked around this by reverting the first commit locally to see that the tests fail without the fix.

LGTM 👍

PR #19755, opened 2024-06-25
deps: downgrade Netty due to incompatibility with grpc-java

npepinpe:
## Description

Netty 4.1.111.Final is incompatible with grpc-java < 1.65.0 and >= 1.51.0. It can introduce buffer corruption at any point, which if you're lucky is detected, but may also go unnoticed and result in garbage and waste of time.

Wait until grpc-java releases a compatible version before updating to 4.1.111.Final. Note that this also blocks Spring Boot 3.3.1 which depends on Netty 4.1.111.Final.

See:

- https://github.com/grpc/grpc-java/issues/11284#issuecomment-2168264887
- https://github.com/grpc/grpc-java/releases/tag/v1.64.0

## Related issues

blocks #19581
blocks #19265
blocks #19264 
blocks #19259
blocks #19258

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -71,6 +71,7 @@
       "allowedVersions": "!/0.8.9/"
     },
     {
+      "description": "Do not update Netty as it is incompatible with grpc-java < 1.65; this can be removed after that",
       "matchManagers": ["maven"],
       "matchPackagePrefixes": ["io.netty"],
       "allowedVersions": "<=4.1.110.Final"
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -71,6 +71,11 @@
       "allowedVersions": "!/0.8.9/"
     },
     {
+      "matchManagers": ["maven"],
+      "matchPackagePrefixes": ["io.netty"],
+      "allowedVersions": "<=4.1.110.Final"
+    },
+    {
       "description" : "Exclude SNAPSHOT versions, renovate may suggest them for pre-release values.",
       "matchManagers": ["maven"],
       "matchPackagePatterns": [".*"],
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -72,7 +72,7 @@
     <version.mockito>5.12.0</version.mockito>
     <version.model>7.7.0</version.model>
     <version.msgpack>0.9.8</version.msgpack>
-    <version.netty>4.1.111.Final</version.netty>
+    <version.netty>4.1.110.Final</version.netty>
     <version.objenesis>3.4</version.objenesis>
     <version.opensearch>2.8.0</version.opensearch>
     <version.opensearch.testcontainers>2.0.1</version.opensearch.testcontainers>


npepinpe:
I'm now wondering if this may not have caused more issues than we thought, considering the Netty issue implies corruption can go unnoticed until it hits your application, and this was merged on the 11th of June :thinking:

Zelldon:
@npepinpe do we need to configure renovate here ? 🤔

npepinpe:
Maybe? Do you know how to pause for a specific dep? I'm no renovate expert :s

Zelldon:
I think it is similar to something like https://github.com/camunda/camunda/blob/main/.github/renovate.json#L50 

But do you then keep an eye on the updates so we don't miss it ? 🤔

Zelldon:
Regarding renovate we can also ask our expert (summon @cmur2 👻 )

Zelldon:
What you where writing regarding the corruption and 

> DO NOT USE NETTY 4.1.111 WITH GRPC. It appears there is corruption. There's no guarantees the corruption would be detected.

Looks really scary 😅

ChrisKujawa (approved):

.github/renovate.json
```diff
@@ -70,6 +70,11 @@
       "matchPackagePrefixes": ["org.jacoco"],
       "allowedVersions": "!/0.8.9/"
     },
+    {
+      "matchManagers": ["maven"],
```
cmur2, line 74: Please add a description (text + e.g. ticket link) like many of the rules have :)

.github/renovate.json
```diff
@@ -70,6 +70,11 @@
...
+      "matchPackagePrefixes": ["io.netty"],
+      "allowedVersions": "<=4.1.110.Final"
```
cmur2, line 76: Should work https://docs.renovatebot.com/configuration-options/#allowedversions

You'd also have to backport this to `stable/**` branches for full effect since PRs are open there as well: https://github.com/camunda/camunda/pull/19258

Alternatively you can always manually close such a PR to ignore that specific version, instead of adding a Renovate rule

npepinpe: The PRs are blocked on stable branches explicitly, so I think it's fine not to backport for now.

.github/renovate.json
npepinpe: I thought about adding a comment but remembered JSON has no comments, I didn't know about the "description" field (but should've since it's used elsewhere :see_no_evil:)

juliojgd:
@npepinpe You've banned Netty 4.1.111 version but, if you read the full conversation at grpc-java, you will realize that there is nothing wrong with Netty 4.1.111 but the problem is with grpc-java. You should upgrade to Netty 4.1.111 when grpc-java releases the fixed versions (and you adopted these new grpc-java fixed versions).

Zelldon:
> @npepinpe You've banned Netty 4.1.111 version but, if you read the full conversation at grpc-java, you will realize that there is nothing wrong with Netty 4.1.111 but the problem is with grpc-java. You should upgrade to Netty 4.1.111 when grpc-java releases the fixed versions (and you adopted these new grpc-java fixed versions).

I think that is the plan. :) but thanks for raising 👍🏼

npepinpe:
As @Zelldon mentioned, that's exactly our plan :) We're just waiting for the 1.65 release of grpc-java in order to update.

PR #19764, opened 2024-06-26
Add bool query test

ChrisKujawa:
## Description

* simplify bool query, remove some complexity and indirections
* Add parameterized tests for bool query

## Related issues

### DIFF
--- a/search/search-client/src/main/java/io/camunda/search/clients/query/SearchBoolQuery.java
+++ b/search/search-client/src/main/java/io/camunda/search/clients/query/SearchBoolQuery.java
@@ -14,18 +14,13 @@ import io.camunda.util.ObjectBuilder;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.function.Function;
 
-public final record SearchBoolQuery(
+public record SearchBoolQuery(
     List<SearchQuery> filter,
     List<SearchQuery> must,
     List<SearchQuery> mustNot,
     List<SearchQuery> should)
     implements SearchQueryOption {
-
-  static SearchBoolQuery of(final Function<Builder, ObjectBuilder<SearchBoolQuery>> fn) {
-    return SearchQueryBuilders.bool(fn);
-  }
 
   public static final class Builder implements ObjectBuilder<SearchBoolQuery> {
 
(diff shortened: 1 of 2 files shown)


search/search-client/src/main/java/io/camunda/search/clients/query/SearchBoolQuery.java
```diff
@@ -39,53 +33,21 @@ public Builder filter(final List<SearchQuery> queries) {
       return this;
     }
 
-    public Builder filter(final SearchQuery query, final SearchQuery... queries) {
```
houssain-barouni, line 42: this method can be useful when you have a single query to use. Do you think it is overkill ?

Zelldon: Based on my experience with designing APIs is important to keep it simple (KISS) and start small. Most of the things you aren't gonna need (YAGNI).

It is always easier to add more stuff, versus later to remove things again from an API (especially public apis). 

For example, here it could also just be used with `List.of(item)`. But if you say it is definitely a use case we want to severe and support then of course we can add it. 

I am not saying that we don't should support such, I just want to keep the complexity to a minimum. When we add such we also need to have more tests, etc. 

Let me know what you think :)

houssain-barouni (approved):
LGTM!

search/search-client/src/main/java/io/camunda/search/clients/query/SearchBoolQuery.java
Zelldon: I was also eager to remove them, and I think in some I did. For consistency we could also remove the others. Any thoughts?

houssain-barouni: we can ask @romansmirnov maybe he had a different vision when he created these methods

romansmirnov: The vision was just to get it working 😄 just joking. In the first iteration, I implemented the API close to the ES/OS clients and used, for example, mostly the fluent builders when testing the components with underlying storage locally.

Nevertheless, I already expected (and made it transparent to others) that we would adjust the API as necessary in later iterations (including deleting unused code). I am okay with deleting stuff, less code, less to maintain, and if it turns out we need something, then we will for sure know better what we need... So all good, let's merge it 🚀

romansmirnov (approved):

PR #19766, opened 2024-06-26
feat: add new binding type attribute to BPMN model API

### LINKED ISSUE #19461: [`deployment` version binding] Add new `bindingType` attribute to BPMN Model API
### Description
Add a new (optional) `bindingType` attribute to the `ZeebeCalledElement`, `ZeebeCalledDecision`, and `ZeebeFormDefinition` extension elements in the BPMN Model API.

### Acceptance Criteria
- [ ] A new attribute `bindingType` is added to `ZeebeCalledElement`, `ZeebeCalledDecision`, and `ZeebeFormDefinition`.
  - [ ] The attribute is optional.
  - [ ] Allowed values are `deployment` and `latest`.
  - [ ] `latest` is used as the default if the attribute is not set.
- [ ] The attribute can be set via the respective builder classes of the parent elements:
  - [ ] `AbstractBusinessRuleTaskBuilder`
  - [ ] `AbstractCallActivityBuilder`
  - [ ] `AbstractStartEventBuilder`
  - [ ] `AbstractUserTaskBuilder`

jfriedenstab:
## Description
Adds support for a new `bindingType` attribute (that can be `latest` or `deployment`) to the `ZeebeCalledElement`, `ZeebeCalledDecision`, and `ZeebeFormDefinition` extension elements in the BPMN Model API.
This should allow Zeebe to determine the binding type for such elements when a deployed BPMN diagram is processed.

## Related issues

closes #19461

### DIFF
--- a/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractUserTaskBuilder.java
+++ b/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractUserTaskBuilder.java
@@ -22,6 +22,7 @@ import static io.camunda.zeebe.model.bpmn.impl.ZeebeConstants.USER_TASK_FORM_KEY
 import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
 import io.camunda.zeebe.model.bpmn.instance.UserTask;
 import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeAssignmentDefinition;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeBindingType;
 import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeFormDefinition;
 import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeHeader;
 import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeTaskHeaders;
@@ -173,6 +174,14 @@ public abstract class AbstractUserTaskBuilder<B extends AbstractUserTaskBuilder<
   @Override
   public B zeebeExternalFormReferenceExpression(final String expression) {
     return zeebeExternalFormReference(asZeebeExpression(expression));
+  }
+
+  @Override
+  public B zeebeFormBindingType(final ZeebeBindingType bindingType) {
+    final ZeebeFormDefinition formDefinition =
+        getCreateSingleExtensionElement(ZeebeFormDefinition.class);
+    formDefinition.setBindingType(bindingType);
+    return myself;
   }
 
   public B zeebeTaskHeader(final String key, final String value) {
--- a/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/impl/instance/zeebe/ZeebeFormDefinitionImpl.java
+++ b/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/impl/instance/zeebe/ZeebeFormDefinitionImpl.java
@@ -18,6 +18,7 @@ package io.camunda.zeebe.model.bpmn.impl.instance.zeebe;
 import io.camunda.zeebe.model.bpmn.impl.BpmnModelConstants;
 import io.camunda.zeebe.model.bpmn.impl.ZeebeConstants;
 import io.camunda.zeebe.model.bpmn.impl.instance.BpmnModelElementInstanceImpl;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeBindingType;
 import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeFormDefinition;
 import org.camunda.bpm.model.xml.ModelBuilder;
 import org.camunda.bpm.model.xml.impl.instance.ModelTypeInstanceContext;
@@ -30,6 +31,7 @@ public class ZeebeFormDefinitionImpl extends BpmnModelElementInstanceImpl
   protected static Attribute<String> formKeyAttribute;
   protected static Attribute<String> formIdAttribute;
   protected static Attribute<String> externalReferenceAttribute;
+  private static Attribute<ZeebeBindingType> bindingTypeAttribute;
 
   public ZeebeFormDefinitionImpl(final ModelTypeInstanceContext instanceContext) {
     super(instanceContext);
@@ -65,6 +67,16 @@ public class ZeebeFormDefinitionImpl extends BpmnModelElementInstanceImpl
     externalReferenceAttribute.setValue(this, externalReference);
   }
 
+  @Override
+  public ZeebeBindingType getBindingType() {
+    return bindingTypeAttribute.getValue(this);
+  }
+
+  @Override
+  public void setBindingType(final ZeebeBindingType bindingType) {
+    bindingTypeAttribute.setValue(this, bindingType);
+  }
+
   public static void registerType(final ModelBuilder modelBuilder) {
     final ModelElementTypeBuilder typeBuilder =
         modelBuilder
@@ -88,6 +100,13 @@ public class ZeebeFormDefinitionImpl extends BpmnModelElementInstanceImpl
         typeBuilder
             .stringAttribute(ZeebeConstants.ATTRIBUTE_EXTERNAL_REFERENCE)
             .namespace(BpmnModelConstants.ZEEBE_NS)
+            .build();
+
+    bindingTypeAttribute =
+        typeBuilder
+            .enumAttribute(ZeebeConstants.ATTRIBUTE_BINDING_TYPE, ZeebeBindingType.class)
+            .namespace(BpmnModelConstants.ZEEBE_NS)
+            .defaultValue(ZeebeBindingType.latest)
             .build();
 
     typeBuilder.build();
--- a/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeBindingType.java
+++ b/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeBindingType.java
@@ -15,23 +15,7 @@
  */
 package io.camunda.zeebe.model.bpmn.instance.zeebe;
 
-import io.camunda.zeebe.model.bpmn.instance.BpmnModelElementInstance;
-
-public interface ZeebeFormDefinition extends BpmnModelElementInstance {
-
-  String getFormKey();
-
-  void setFormKey(String formKey);
-
-  String getFormId();
-
-  void setFormId(String formId);
-
-  String getExternalReference();
-
-  void setExternalReference(String externalReference);
-
-  ZeebeBindingType getBindingType();
-
-  void setBindingType(ZeebeBindingType bindingType);
+public enum ZeebeBindingType {
+  deployment,
+  latest
 }
--- a/zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/builder/CallActivityBuilderTest.java
+++ b/zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/builder/CallActivityBuilderTest.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright © 2017 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.zeebe.model.bpmn.builder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
+import io.camunda.zeebe.model.bpmn.instance.ExtensionElements;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeBindingType;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeCalledElement;
+import org.camunda.bpm.model.xml.instance.ModelElementInstance;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class CallActivityBuilderTest {
+
+  @Test
+  void shouldSetProcessId() {
+    // when
+    final BpmnModelInstance instance =
+        Bpmn.createExecutableProcess("process")
+            .startEvent()
+            .callActivity("callActivity", c -> c.zeebeProcessId("process-id-1"))
+            .done();
+
+    // then
+    final ModelElementInstance callActivity = instance.getModelElementById("callActivity");
+    final ExtensionElements extensionElements =
+        (ExtensionElements) callActivity.getUniqueChildElementByType(ExtensionElements.class);
+    assertThat(extensionElements.getChildElementsByType(ZeebeCalledElement.class))
+        .hasSize(1)
+        .extracting(ZeebeCalledElement::getProcessId)
+        .containsExactly("process-id-1");
+  }
+
+  @Test
+  void shouldSetProcessIdExpression() {
+    // when
+    final BpmnModelInstance instance =
+        Bpmn.createExecutableProcess("process")
+            .startEvent()
+            .callActivity("callActivity", c -> c.zeebeProcessIdExpression("processIdExpr"))
+            .done();
+
+    // then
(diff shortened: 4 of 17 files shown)


zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractUserTaskBuilder.java
```diff
@@ -175,6 +176,14 @@ public B zeebeExternalFormReferenceExpression(final String expression) {
     return zeebeExternalFormReference(asZeebeExpression(expression));
   }
 
+  @Override
+  public B zeebeFormBindingType(final ZeebeBindingType bindingType) {
```
jfriedenstab, line 180: FYI: Given that you can set a couple of other attributes that are not related to the form, I named this method `zeebeFormBindingType` instead of just `zeebeBindingType`.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/impl/instance/zeebe/ZeebeFormDefinitionImpl.java
```diff
@@ -90,6 +102,12 @@ public static void registerType(final ModelBuilder modelBuilder) {
             .namespace(BpmnModelConstants.ZEEBE_NS)
             .build();
 
+    bindingTypeAttribute =
+        typeBuilder
+            .enumAttribute(ZeebeConstants.ATTRIBUTE_BINDING_TYPE, ZeebeBindingType.class)
+            .namespace(BpmnModelConstants.ZEEBE_NS)
+            .build();
```
jfriedenstab, line 109: I decided not to set a default value here, because the binding type is not applicable when using `formKey` or `externalReference`.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/validation/zeebe/ZeebeFormDefinitionValidator.java
```diff
@@ -56,6 +60,14 @@ public void validate(
...
                 ""));
       }
+      if (!isBlank(formKey) && bindingType != null) {
+        validationResultCollector.addError(
+            0,
+            String.format(
+                ERROR_MESSAGE_INVALID_ATTRIBUTE_COMBINATION,
+                ZeebeConstants.ATTRIBUTE_FORM_KEY,
+                ZeebeConstants.ATTRIBUTE_BINDING_TYPE));
+      }
```
jfriedenstab, line 70: I'm not sure whether such a validation is actually needed (the `bindingType` could also just be ignored if it was used with a `formKey` or `externalReference`). But I think it doesn't hurt to be strict here and return an error.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractUserTaskBuilder.java
korthout: 💭 It would be good to [document such intentions in the commit message](https://github.com/camunda/camunda/wiki/Pull-Requests-and-Code-Reviews#ive-written-a-clear-changelist-description) as well. This makes it easier to re-discover such decisions later, as the git commit message are available locally close to the code in most editors compared to GitHub PR comments which exist on the web. [Splitting up your commit into multiple smaller ones](https://github.com/camunda/camunda/wiki/Pull-Requests-and-Code-Reviews#ive-narrowly-scoped-my-changes) can help highlight changes for which a reason should be documented.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeBindingType.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.zeebe.model.bpmn.instance.zeebe;
+
+public enum ZeebeBindingType {
+  deployment,
+  latest
```
korthout, line 20: 💭 I was surprised to see non-capitalized enum values, but they appear to be needed:
https://github.com/camunda/camunda/pull/16519#discussion_r1508047718 TIL

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/impl/instance/zeebe/ZeebeFormDefinitionImpl.java
korthout: Not setting a default means that existing models with a FormDefinition using `formId` (those without `bindingType`) are implicitly using `latest` even though there is no `bindingType` specified and no default either. I'm not sure whether it's a big problem, but perhaps we see a better solution. 

❓ Do you see any way to overcome this?

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/validation/zeebe/ZeebeFormDefinitionValidator.java
korthout: 🔧 With my previous comment in mind, it may be better to allow `bindingType` and `formKey` at the same time, to allow a default `bindingType` value.

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/builder/CallActivityBuilderTest.java
```diff
@@ -0,0 +1,133 @@
...
+    final ExtensionElements extensionElements =
+        (ExtensionElements) callActivity.getUniqueChildElementByType(ExtensionElements.class);
+    assertThat(extensionElements.getChildElementsByType(ZeebeCalledElement.class))
+        .hasSize(1)
+        .extracting(ZeebeCalledElement::getProcessId)
+        .containsExactly("process-id-1");
+  }
+
+  @Test
+  void shouldSetProcessIdExpression() {
```
korthout, line 53: 👍 Thanks for adding additional coverage 🙇

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/validation/ZeebeTaskValidatorFormTest.java
```diff
@@ -349,6 +350,19 @@ public static Object[][] parameters() {
...
+            .startEvent()
+            .userTask("task")
+            .zeebeFormKey("form-key")
+            .zeebeFormBindingType(ZeebeBindingType.deployment)
+            .endEvent()
+            .done(),
+        singletonList(
+            expect(
+                ZeebeFormDefinition.class,
+                "Attribute 'formKey' must not be used in combination with attribute 'bindingType'"))
```
korthout, line 364: 👍 I do like that we're helping users by clear rejection messages

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/validation/ZeebeTaskValidatorFormTest.java
```diff
@@ -625,7 +639,21 @@ public static Object[][] parameters() {
...
+            .userTask("task")
+            .zeebeUserTask()
+            .zeebeExternalFormReference("reference")
+            .zeebeFormBindingType(ZeebeBindingType.deployment)
+            .endEvent()
+            .done(),
+        singletonList(
+            expect(
+                ZeebeFormDefinition.class,
+                "Attribute 'externalReference' must not be used in combination with attribute 'bindingType'"))
```
korthout, line 655: 🔧 We could consider adding a test that shows what happens when a string is used that is not an enum value of `ZeebeBindingType` as `bindingType`

korthout (commented):
Nice work @jfriedenstab 👏 

👍 I like the changes and choices you've made

❌ There's one open question that we must discuss before I can approve the changes. Please have a look.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractUserTaskBuilder.java
jfriedenstab: Thanks, good points! 👍🏻 We're not really using the commit message body in my team, so I still need to get used to providing more information there ;) I've also realized that I could have split up my commit into multiple smaller ones. From this repo's history I see that the PRs often contain relatively small individual commits.

I will keep the points in mind for the next PR. Or was your comment meant as a suggestion to rewrite the commit history?

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeBindingType.java
jfriedenstab: I actually took `ZeebeExecutionListenerEventType` as an inspiration 🙂. But, yes, it looks a bit unusual indeed.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/impl/instance/zeebe/ZeebeFormDefinitionImpl.java
jfriedenstab: Thanks for raising this 👍🏻.

> Not setting a default means that existing models with a FormDefinition using `formId` (those without `bindingType`) are implicitly using `latest` even though there is no `bindingType` specified and no default either.

Hm, I think the problem is rather that for these models the `bindingType` will be `null` and not `latest`, right? However, `latest` would make sense as this is already the current behavior in Zeebe if a `formId` is used.

I would basically see three options to deal with this:
1. Always use `latest` as the default if the `bindingType` is not explicitly set in the model, even if `formKey` or `externalReference` is used. This would be the simplest solution (and also resolve your [other comment](https://github.com/camunda/camunda/pull/19766#discussion_r1658848845)).

2. Use `latest` as the default if the `bindingType` is not explicitly set in the model, but _only_ if the `formId` is set (probably by enhancing the `getBindingType` method). This would add a bit of extra complexity, but reflect my [original intention](https://github.com/camunda/camunda/pull/19766#discussion_r1655076741).

3. Keep the current implementation. This would require an extra `null` check in Zeebe later when the binding type is actually used to resolve the form by `formId` (`null` would be equivalent to `latest` then).

I would prefer 1 or 2, also to be in line with `ZeebeCalledElementImpl` and `ZeebeCalledDecisionImpl` which actually define a default value.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/validation/zeebe/ZeebeFormDefinitionValidator.java
jfriedenstab: Yes, if we decide to always set a default value for `bindingType`, let's just remove this extra validation.

One use case I had in mind when adding it was that users might have specified the wrong attribute by mistake (e.g. they actually wanted to use `formId` with `bindingType=deployment`, but used `formKey` instead). But that would probably be quite an edge case.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractUserTaskBuilder.java
korthout: @jfriedenstab My intention was simply to point it out as something for future changes. However, if you think it's easy to change then please take it along.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/impl/instance/zeebe/ZeebeFormDefinitionImpl.java
korthout: It seems to me that semantically it's not incorrect to use `latest` with `formKey`. There simply is only one version of it. 🤷 

I'm not a fan of unnecessary complexity. It doesn't appear to me as something that significantly helps our users. If it does, we could always add such rules later as improvements.

Given that option 1 is the simplest, it has my preference until we see a user need.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractUserTaskBuilder.java
jfriedenstab: It was relatively easy, so I split the commit into two.

zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/impl/instance/zeebe/ZeebeFormDefinitionImpl.java
jfriedenstab: Let's go with option 1 then, thanks! :)

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/validation/ZeebeTaskValidatorFormTest.java
jfriedenstab: Good suggestion 👍🏻. I added such tests with https://github.com/camunda/camunda/pull/19766/commits/9efc3ee74a44974b1ce19e615a2dc08b881c087c.

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeCalledDecisionTest.java
```diff
@@ -37,6 +44,43 @@ public Collection<ChildElementAssumption> getChildElementAssumptions() {
...
+
+    // when
+    final ZeebeCalledDecision calledDecision =
+        businessRuleTask.getSingleExtensionElement(ZeebeCalledDecision.class);
+
+    // then
+    assertThatThrownBy(calledDecision::getBindingType)
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage(
+            "No enum constant io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeBindingType.foo");
```
jfriedenstab, line 84: 💭 Maybe it would make sense to add a validation rule to return a more user-friendly error if an invalid binding type was used (something like `Attribute 'bindingType' must be one of: deployment, latest`). But I would add this in a follow-up PR.

jfriedenstab:
@korthout I addressed all the comments. Could you have another look again?

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeCalledDecisionTest.java
```diff
@@ -41,4 +48,39 @@ public Collection<AttributeAssumption> getAttributesAssumptions() {
         new AttributeAssumption(
             BpmnModelConstants.ZEEBE_NS, "bindingType", false, false, ZeebeBindingType.latest));
   }
+
+  @Test
+  public void shouldReadValidBindingType() {
+    // given
+    final BpmnModelInstance modelInstance =
+        Bpmn.readModelFromStream(getClass().getResourceAsStream("ZeebeCalledDecisionTest.bpmn"));
```
korthout, line 56: 🔧 From the name of the test method and from the name of the bpmn resource it's unclear that this verifies that the binding type of the called decision is `deployment`. My suggestion is to adjust the bpmn filename.

🔧 I'm not a big fan of bpmn files in tests, as they make the tests harder to understand (the information is split over multiple places), like above. We could use the bpmn-model api to create the bpmn model, produce the xml, and then read it again all in the same test case to avoid this.

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeCalledDecisionTest.java
korthout: 👍 I like this idea, and would suggest we take this along while this is still fresh. Fine to do so in a separate PR though.

korthout (approved):
Looking great @jfriedenstab 🚀 

🔧 I have a few last suggestions, but nothing blocking this from being merged when you see fit. Feel free to apply the suggestions or just merge.

👍 LGTM

💭 Sorry for the delayed review

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeCalledDecisionTest.java
jfriedenstab: Thanks, good points! I replaced the BPMN files with models created by the API for both cases (valid and invalid binding type).

zeebe/bpmn-model/src/test/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeCalledDecisionTest.java
jfriedenstab: I created a [follow-up issue](https://github.com/camunda/camunda/issues/20054) for this and will tackle it soon.

PR #19773, opened 2024-06-26
feat: add role management for a group to identity's UI

maryarm:
## Description
We need to be able to view/assign/unassign roles to a Group in UI.

## Related issues
Part of identity issue ([#2946](https://github.com/camunda-cloud/identity/issues/2947))


dlavrenuek (changes requested):
Looks good already. I would do some refactoring (as part of copying current implementation) so we have a clean and structured code to work with in the future

npepinpe:
@maryarm - I guess we should close this? It's probably very outdated, no?

cmur2:
@maryarm can this be closed?

Issue #17721, opened 2024-04-24
[EPIC] Only run relevant CI for each change


cmur2:
## Introduction

Currently, the monorepo includes the CI pipelines of Zeebe, Operate and Tasklist side-by-side. They are not integrated and as noted in #17231 the path filters are lengthy, redundantly defined and not applicable to the merge queue. In the end we run CI pipelines of Zeebe, Operate and Tasklist more often than we need and also on unrelated changes, partly because there is no single entry point so using GitHub required status checks makes this duplication necessary. This has a few downsides:

* higher cost (large parts of the CI run on infrastructure that Camunda pays by use)
* less stability due to inherent flakiness (running unrelated flaky tests may fail the whole pipeline)
* confusing UI with 20+ very detailed and partially redundant status checks
* cumbersome configuration of GitHub required status checks

We can get some inspiration from the Console monorepo workflows where Clément recently established a [more efficient and modular CI pipeline structure](https://github.com/camunda-cloud/camunda-cloud-management-apps/blob/main/.github/workflows/ci.yaml) using change detection and conditional execution.

The scope of this Epic is to provide a foundation for unifying the existing CI pipelines running on code changes to have a single entrypoint and only running pipelines relevant for a change. [More details in the planning doc](https://docs.google.com/document/d/1to2jKWSnFvMrL6eMLd0NfcK0Yzp4v8N7R03Fi3LU_qc/edit#heading=h.fzqhk2y369b2).

## Task Breakdown

Zelldon:
I created a follow-up https://github.com/camunda/camunda/issues/19539 to further investigate and improve related to further improve when to run which tests etc.

cmur2:
https://github.com/camunda/product-hub/issues/2629 will also contribute to this goal.

cmur2:
Closing in favor of https://github.com/camunda/camunda/issues/46709

PR #19778, opened 2024-06-26
Add build backend and frontend to unified ci

ChrisKujawa:
## Description

> [!NOTE] 
> This is just an proposal

We havew previously added merge CIs for Operate and Tasklist, which were mostly only building the apps.

I think this can be merged into the Unified CI, and we can get rid of these extra merge workflows. With that, we can also have one place where we build the frontends together.

Additionally, here we can do the java checks (checking for google styleguide and spotbugs) which was previously done in Zeebe CI. This inherently also builds the backend, so we need no extra step for this.

## Related issues

Related to https://github.com/camunda/camunda/issues/17721

### DIFF
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -45,7 +45,8 @@ outputs:
       ${{
         github.event_name == 'push' ||
         steps.filter-common.outputs.github-actions-change == 'true' ||
-        steps.filter-identity.outputs.identity-frontend-change == 'true'
+        steps.filter-common.outputs.frontend-change == 'true' ||
+        steps.filter-identity.outputs.frontend-identity == 'true'
       }}
 
 
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -39,6 +39,14 @@ outputs:
         steps.filter-common.outputs.dockerfile-change == 'true' ||
         steps.filter-common.outputs.frontend-change == 'true'
       }}
+  frontend-change:
+    description: Output whether any frontend has changed
+    value: >-
+      ${{
+        github.event_name == 'push' ||
+        steps.filter-common.outputs.github-actions-change == 'true' ||
+        steps.filter-identity.outputs.identity-frontend-change == 'true'
+      }}
 
 
 runs:
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -39,7 +39,7 @@ outputs:
         steps.filter-common.outputs.dockerfile-change == 'true' ||
         steps.filter-common.outputs.frontend-change == 'true'
       }}
-  frontend-change:
+  frontend-changes:
     description: Output whether any frontend has changed
     value: >-
       ${{
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,7 @@ jobs:
         with:
           directory: ./tasklist/client
       - uses: ./.github/actions/build-frontend
-        id: build-tasklist-fe
+        id: build-identity-fe
         with:
           directory: ./identity/client
       - name: Observe build status
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -423,6 +423,8 @@ jobs:
       - docker-checks
       - identity-frontend-tests
       - integration-tests
+      - build-platform-backend
+      - build-platform-frontend
     steps:
       - run: exit ${{ ((contains(needs.*.result, 'cancelled') || contains(needs.*.result, 'failure')) && 1) || 0 }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,7 @@ jobs:
     if: needs.detect-changes.outputs.java-unit-tests == 'true'
     needs: [detect-changes]
     name: Java checks
+    timeout-minutes: 15
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,25 +65,21 @@ jobs:
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
 
-  build-platform-backend:
+  java-checks:
     if: needs.detect-changes.outputs.java-unit-tests == 'true'
     needs: [ detect-changes ]
+    name: Java checks
     runs-on: ubuntu-latest
-    timeout-minutes: 15
-    outputs:
-      flakyTests: ${{ steps.analyze-test-run.outputs.flakyTests }}
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-zeebe
         with:
-          go: true
+          go: false
+          maven-cache-key-modifier: java-checks
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
-      - uses: ./.github/actions/build-zeebe
-        with:
-          go: false
-          maven-extra-args: -T1C -PskipFrontendBuild
+      - run: ./mvnw -T1C -B -D skipTests -P !autoFormat,checkFormat,spotbugs,skipFrontendBuild verify
       - name: Observe build status
         if: always()
         continue-on-error: true
@@ -93,7 +89,6 @@ jobs:
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
-
 
   build-platform-frontend:
     if: needs.detect-changes.outputs.frontend-changes == 'true'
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -64,6 +65,73 @@ jobs:
...
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          build_status: ${{ job.status }}
+          secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+
+
+  build-platform-frontend:
```
houssain-barouni, line 98: if we use two different jobs, the frontend packages won't be part of the final distribution, because they are built in two different workspaces

Zelldon: For me the only intentention was to add jobs to build the respective components either be or fe. Just to verify whether they can be built. We can of course further improve and reuse stuff via artifacts, but this was not the goal right now.

.github/workflows/ci.yml
```diff
@@ -64,6 +65,73 @@ jobs:
...
+          maven-extra-args: -T1C -PskipFrontendBuild
+      - name: Observe build status
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          build_status: ${{ job.status }}
+          secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
```
Zelldon, line 95: Actually, I think we can completely skip this, as this is already part of the java-unit-tests. So we could remove the Operate merge CI already. WDYT @houssain-barouni @cmur2 do you prefer a separate step for just building 🤔

.github/workflows/tasklist-merge-ci.yml
```diff
@@ -10,14 +10,6 @@ concurrency:
   group: "${{ github.workflow }}-${{ github.ref }}"
 
 jobs:
-  run-build:
-    name: run-build
-    uses: ./.github/workflows/tasklist-ci-build-reusable.yml
```
houssain-barouni, line 15: this also builds tasklist docker image, which is not the case in the unified ci

.github/actions/paths-filter/action.yml
```diff
@@ -40,6 +40,14 @@ outputs:
         steps.filter-common.outputs.frontend-change == 'true' ||
         steps.filter-identity.outputs.identity-frontend-change == 'true'
       }}
+  frontend-change:
+    description: Output whether any frontend has changed
+    value: >-
+      ${{
+        github.event_name == 'push' ||
+        steps.filter-common.outputs.github-actions-change == 'true' ||
+        steps.filter-identity.outputs.identity-frontend-change == 'true'
```
houssain-barouni, line 49: we need filters on operate and tasklist client folders

cmur2:
@houssain-barouni @Zelldon I'm aware of this PR and appreciate this effort towards the unified CI! Due to hiring I can only give qualified feedback tomorrow 🙏🏼

.github/workflows/operate-merge-ci.yaml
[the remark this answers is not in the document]
Zelldon: Which we want to remove :)

.github/workflows/tasklist-merge-ci.yml
Zelldon: For me it is still unclear whether we want or need this as this can also be done via camunda and the right profile ahould be enough for development

.github/workflows/ci.yml
Zelldon: I replaced it with the java checks we have in our Zeebe ci and that are mandatory for the complete project

.github/actions/paths-filter/action.yml
Zelldon: hups

.github/workflows/tasklist-merge-ci.yml
Zelldon: Do you think we or the team can leverage the Camunda docker image instead of the Tasklist image?

On release, we would still create the tasklist image btw. But I think this would reduce a lot of complexity of our CIs if we reduce the docker image builds.

Wdyt @cmur2 ? 🤔 


But I guess might be an issue for SaaS and snapshot usage their 🤔 we might need to adjust then also something 🤔 Hm no sure anymore :D

.github/actions/paths-filter/action.yml
Zelldon: Sure, but why we have now two ways of doing it ?

Like https://github.com/camunda/camunda/blob/main/.github/actions/paths-filter/action.yml#L62-L69 

VS https://github.com/camunda/camunda/blob/main/.github/actions/paths-filter/action.yml#L55-L60

cmur2:
Finally got time to look at this now 👀

.github/workflows/ci.yml
```diff
@@ -64,6 +65,68 @@ jobs:
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
 
+  java-checks:
+    if: needs.detect-changes.outputs.java-unit-tests == 'true'
+    needs: [ detect-changes ]
+    name: Java checks
+    runs-on: ubuntu-latest
+    steps:
```
cmur2, line 73: Need a `timeout-minutes` here

.github/workflows/ci.yml
```diff
@@ -64,6 +65,68 @@ jobs:
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
 
+  java-checks:
+    if: needs.detect-changes.outputs.java-unit-tests == 'true'
```
cmur2, line 69: Probably should consider renaming this variable eventually (not necessarily in this PR) to not confuse readers, it is more generically used now

.github/workflows/ci.yml
```diff
@@ -64,6 +65,68 @@ jobs:
...
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+
+  build-platform-frontend:
+    if: needs.detect-changes.outputs.frontend-changes == 'true'
+    needs: [ detect-changes ]
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    outputs:
+      flakyTests: ${{ steps.analyze-test-run.outputs.flakyTests }}
```
cmur2, line 99: There is no `analyze-test-run` step here so this should go

.github/workflows/ci.yml
Zelldon: I aggree. I did this in the other PR.

.github/actions/paths-filter/action.yml
cmur2: > Sure, but why we have now two ways of doing it ?

Ah, they are functionally equivalent but I thought it might be nice to scope team specific filters (as long as needed) out into separate steps - for a logical partitioning.

.github/workflows/tasklist-merge-ci.yml
cmur2: The `operate-merge-ci` also builds the Operate Docker image, so same question for that. If we want to keep mental load low, we can migrate them as well to the unified CI in their current form.

.github/workflows/ci.yml
cmur2: > the other PR
Oh, which did I miss? 🙈

.github/workflows/tasklist-merge-ci.yml
cmur2: It shouldn't be much effort to migrate those Docker build jobs in their current form (no snapshot build/push needed), can also help with that.

Zelldon: Ok lets do that then.

.github/workflows/ci.yml
Zelldon: :) https://github.com/camunda/camunda/pull/19785/files#diff-1951b8794b638a238e5b4db8432cc73261ef7a353ff0a17ca3efc5501cb1dadfL14

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       java-unit-tests: ${{ steps.filter.outputs.java-unit-tests }}
       docker-ci: ${{ steps.filter.outputs.docker-ci }}
       identity-frontend-tests: ${{ steps.filter.outputs.identity-frontend-tests }}
+      frontend-changes: ${{ steps.filter.output.frontend-changes }}
```
cmur2, line 34: Typo, needs to be: `steps.filter.outputs.frontend-changes` (plural "outputs")

.github/workflows/ci.yml
```diff
@@ -255,6 +318,8 @@ jobs:
       - java-unit-tests
       - docker-checks
       - identity-frontend-tests
+      - build-platform-backend
```
cmur2, line 321: This job does not exist

cmur2 (commented):
Two things which I'll push a fix commit for

.github/workflows/tasklist-merge-ci.yml
cmur2: Actually, the `{operate|tasklist}-ci-buid-reusable` flows do not just build the Docker image, they also push that and the snapshot Maven artifacts to the registries so replacing is a bit more effort.To wrap this PR up I think it makes sense to not delete this nor the operate-merge-ci.yml yet and keep migrating in another PR - in the meantime we have some duplication but that is acceptable temporarily.

Zelldon: Ok I will bring it back then.

Zelldon:
I think I applied all comments @cmur2 @houssain-barouni. If @houssain-barouni PR gets merged, we can rebase again and enqueue it if you're fine.

cmur2 (approved):

PR #19757, opened 2024-06-25
Add root QA modul

ChrisKujawa:
## Description

Creates a new root QA module, that should give teams a place where they can put general platform-level integration tests, that cover the whole stack. 

Move some first utilities that have been recently added to the `zeebe-qa-util` module to bootstrap the Camunda application to the new `qa/util` module. 

Furthermore, the integration test that bootstraps a Camunda application and verifies the round trip of a process instance has been moved to the new `qa/integration-test` module.

This should build the ground level where we can add and extend more tests.

## Related issues

See related discussion https://camunda.slack.com/archives/C06UKS51QV9/p1719329200944059

PR #19785, opened 2024-06-26
Add QA IT to the Unified CI

ChrisKujawa:
## Description

With #19757 we added an new QA module containing integration tests on the root level. This PR should add the execution of such tests to the unified CI. Furthermore, it adds the integrations tests from the Zeebe CI already also to the job, which can be seen as a blue print to add more such integration tests.

Probably useful for Operate, Tasklist and other components that have separate component related integration tests. 

## Related issues

related to #19757

### DIFF
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -11,8 +11,8 @@ outputs:
         github.event_name == 'push' ||
         steps.filter-common.outputs.github-actions-change == 'true'
       }}
-  java-unit-tests:
-    description: Output whether `unit-tests` (java) should be run based on GitHub event and files changed
+  java-code-changes:
+    description: Output whether `java code changes` have been detected
     value: >-
       ${{
         github.event_name == 'push' ||
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -12,7 +12,7 @@ outputs:
         steps.filter-common.outputs.github-actions-change == 'true'
       }}
   java-code-changes:
-    description: Output whether `java code changes` have been detected
+    description: Output whether jobs depending on Java code changes should be run, related checks are based on GitHub events and java files changes
     value: >-
       ${{
         github.event_name == 'push' ||
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,7 @@ jobs:
   integration-tests:
     if: needs.detect-changes.outputs.java-code-changes == 'true'
     name: "[IT] ${{ matrix.name }}"
+    needs: [detect-changes]
     timeout-minutes: 20
     outputs:
       flakyTests: ${{ steps.analyze-test-run.outputs.flakyTests }}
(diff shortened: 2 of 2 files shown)


.github/actions/paths-filter/action.yml
```diff
@@ -11,8 +11,8 @@ outputs:
         github.event_name == 'push' ||
         steps.filter-common.outputs.github-actions-change == 'true'
       }}
-  java-unit-tests:
-    description: Output whether `unit-tests` (java) should be run based on GitHub event and files changed
+  java-code-changes:
+    description: Output whether `java code changes` have been detected
```
cmur2, line 15: the reason why I included this "should be run based on GitHub event and files changed" blurb is that this condition also always is true for `push` events and people could assume that is wrong - it is probably a nitpick but just to provide context

.github/workflows/ci.yml
```diff
@@ -116,6 +116,99 @@ jobs:
...
+        if: ${{ failure() || cancelled() }}
+        with:
+          name: "[IT] ${{ matrix.name }}"
+      - name: Observe build status
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          job_name: "integration-tests/${{ matrix.group }}"
+          build_status: ${{ contains(steps.*.conclusion, 'failure') && 'failure' || 'success' }}
```
cmur2, line 207: `build_status: ${{ job.status }}` again :)

.github/actions/paths-filter/action.yml
Zelldon: Not sure what you mean with the comment :) Should I add this back again?

cmur2: I'm not sure myself either what I mean...

I think I'd prefer "Output whether jobs depending on Java code changes should be run based on GitHub event and files changed" here :)

.github/workflows/ci.yml
Zelldon: Sorry, I could swear I replaced that.

.github/actions/paths-filter/action.yml
Zelldon: How about: 

> Output whether jobs depending on Java code changes should be run, related checks are based on GitHub events and java files changes

cmur2: Fine by me :)

.github/workflows/ci.yml
```diff
@@ -92,7 +92,7 @@ jobs:
           ./mvnw -T2 -B --no-snapshot-updates
           -D skipITs -D skipQaBuild=true -D skipChecks -D surefire.rerunFailingTestsCount=3
           -D junitThreadCount=16
-          -P skip-random-tests,parallel-tests,extract-flaky-tests
+          -P skip-random-tests,parallel-tests,extract-flaky-tests,skipFrontendBuild
```
Zelldon, line 95: ℹ️  Removes around 1 min build time

cmur2 (approved):
Nice!

Issue #19482, opened 2024-06-18
Change maven cordinates and rename the project


nicpuppa:
Rename the project, change maven coordinates, the new client should be backward compatible. Explore maven redirect/mirror functionality: https://maven.apache.org/guides/mini/guide-mirror-settings.html

PR #19790, opened 2024-06-26
build: add distribution management to flattened pom

nicpuppa:
## Description

We need to add `flatten-maven-plugin` to the old `zeebe-java-client` pom. In order to add the distribution management we need for the client 

## Related issues

related #19482


korthout (approved):
LGTM 👍

PR #19793, opened 2024-06-26
Release 8.3.13

camundait:
Merge the release branch with stable. This pull request was created by Zeebe :rocket:.


korthout:
@koevskinikola Can we merge this before https://github.com/camunda/camunda/pull/19330?

koevskinikola:
Good point, I guess we should wait until that one is merged. Are there any reasons why #19098 wasn't merged until now?

korthout:
My feeling is that it doesn't matter. Let's merge this and close the older ones

korthout (approved):

koevskinikola:
Then why don't we merge the older ones first? There might be fixes that aren't present in the `stable/*` branches. 

For example, there was a discussion about fixing some issues with the CI through #19330 [here](https://camunda.slack.com/archives/C05DH1F5TAR/p1718801404149969).

PR #19804, opened 2024-06-26
chore(optimize8): fix optimize java checks

buccarel:
## Description

This fixes the following workflow:
* `Optimize Run Java checks / Run Java checks (pull_request)`

### DIFF
--- a/.github/workflows/optimize-run-java-checks.yml
+++ b/.github/workflows/optimize-run-java-checks.yml
@@ -31,7 +31,7 @@ jobs:
         uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 # v4
         with:
           distribution: "adopt"
-          java-version: "17"
+          java-version: "21"
           cache: "maven"
       - name: "Create settings.xml"
         uses: s4u/maven-settings-action@7802f6aec16c9098b4798ad1f1d8ac75198194bd # v3.0.0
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -57,6 +57,7 @@
               <exclude>clients/go/vendor/**/*.md</exclude>
               <exclude>**/node_modules/**/*.md</exclude>
               <exclude>tasklist/client/**/*.md</exclude>
+              <exclude>**/.node/**/*.md</exclude>
```
PHWaechtler, line 60: Is this also related to the java checks?

.github/workflows/optimize-run-java-checks.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 # v4
         with:
           distribution: "adopt"
-          java-version: "17"
+          java-version: "21"
```
PHWaechtler, line 34: Does this mean we're changing our java version? Would that need documented?
https://docs.camunda.io/docs/reference/supported-environments/#component-requirements

HeleneW-dot (changes requested):
Job lgtm, but I have some changes as its not clear to me what it means to change the version there. If it only affects Ci then all good to merge

.github/workflows/optimize-run-java-checks.yml
buccarel: I thought I did it here time ago with this:
https://github.com/camunda/camunda-docs/commit/f56d15e7ef9d57853f0928bd456c82818d9031ac

I don't know why the web site is not

PHWaechtler: ok all good then, just please make sure its updated

HeleneW-dot (approved):

PR #19815, opened 2024-06-27
ci: remove GH comments for Tasklist CI

ChrisKujawa:
## Description

Removing the usage of https://github.com/EnricoMi/publish-unit-test-result-action which adds GH comment for executed Tests.

**Why:**

Adding GitHub comments on the executed Test count brings no
additional value. Teams are not consuming it, and it is polluting the PR conversations with unnecessary data that is also accessible via PR status checks.

## Related issues
See slack conv https://camunda.slack.com/archives/C043W5V88M7/p1719478030055259


houssain-barouni (approved):

cmur2 (approved):
🧹

PR #19823, opened 2024-06-27
Allow disabling REST gateway security through configuration

### LINKED ISSUE #18726: `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE:none` does not disable Identity-based authentication
**Describe the bug**

It should be possible to set the env var `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE=none`, then use the Zeebe REST API without authenticating via Identity/Keycloak as mentioned in our [docs](https://docs.camunda.io/docs/self-managed/zeebe-deployment/security/client-authorization/#camunda-identity-authorization)

**To Reproduce**

- Install using [separated-ingress-values.yaml](https://github.com/daniel-ewing/camunda-platform-helm/blob/SUPPORT-22024/dev-10.0.x-tests/separated-ingress-values.yaml) and helm charts 10.0.5.
- Port forwarding the zeebe-gateway pod
- Try to execute the GET topology REST API with no bearer token: rest.zeebe.c8.dev.local/v1/topology.

The result is a 401 Unauthorized error.

**Expected behavior**

Setting `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE` to `none` should disable Identity-based authentication.

**Hint**

Customer said that: _If the spring.profiles.active: identity-auth in zeebe-gateway/configmap.yaml is commented-out/deactivated, then the Identity-based authentication is disabled._

**Environment:**

- Camunda version: 8.5
- Helm chart version: 10.0.4

Related to support case [SUPPORT-22024](https://jira.camunda.com/browse/SUPPORT-22024)

korthout:
## Description

As documented, the authentication of the Gateway can be completely disabled by setting `security.authentication.mode: none`.

>The authentication could be disabled by `setting security.authentication.mode: none` in the Gateway configuration file or via `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE=none` as environment variable.

See: https://docs.camunda.io/docs/self-managed/zeebe-deployment/security/client-authorization/#camunda-identity-authorization

For the GRPC API, this already works because we explicitly apply the security configuration:

https://github.com/camunda/camunda/blob/a5a1003acfbe210482b38eea445d880f3f0023f6/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/Gateway.java#L387

For the REST API, the security configuration is controlled through Spring magix. This made it easy to fix, but hard to test. The only solution I could think of was an integration test in the same style as `GatewayAuthenticationIdentityIT`.

https://github.com/camunda/camunda/blob/a5a1003acfbe210482b38eea445d880f3f0023f6/dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java#L62-L64

## Related issues

closes #18726

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
@@ -15,8 +15,11 @@ import jakarta.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import org.springframework.boot.actuate.autoconfigure.security.servlet.EndpointRequest;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.condition.NoneNestedConditions;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.ConfigurationCondition.ConfigurationPhase;
 import org.springframework.context.annotation.Profile;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
@@ -62,11 +65,7 @@ public final class SecurityConfiguration {
 
   @Bean
   @ConditionalOnRestGatewayEnabled
-  @ConditionalOnProperty(
-      prefix = "zeebe.gateway.security",
-      value = "authentication.mode",
-      havingValue = "identity",
-      matchIfMissing = true)
+  @Conditional(GatewaySecurityAuthenticationEnabledCondition.class)
   public SecurityFilterChain restGatewaySecurity(
       final HttpSecurity http,
       final IdentityAuthenticationManager authManager,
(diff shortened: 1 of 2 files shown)


dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -61,6 +62,11 @@ public SecurityFilterChain managementSecurity(final HttpSecurity http) throws Ex
 
   @Bean
   @ConditionalOnRestGatewayEnabled
+  @ConditionalOnProperty(
+      prefix = "zeebe.gateway.security",
+      value = "authentication.mode",
+      havingValue = "identity",
+      matchIfMissing = true)
```
romansmirnov, line 69:
```suggestion
  @ConditionalOnProperty(
      name = {"zeebe.gateway.security.authentication.mode", "zeebe.broker.gateway.security.authentication.mode"}
      havingValue = "identity",
      matchIfMissing = true)
```

:question: When the gateway runs embedded, then the authentication mode is set via `zeebe.broker.gateway.security.authentication.mode`. Am I correct?

romansmirnov (commented):
@korthout, if I am not mistaken then the used configuration property is only used when running the Gateway standalone. In the case of an embedded gateway, it would be `zeebe.broker.gateway.security.authentication.mode`. Could you please verify?

korthout:
Good point @romansmirnov 

Should we have the security enabled by default, or only when explicitly enabled through `security.authentication.mode: identity`?

korthout:
PS: commitlint is failing because of a long URL, but I don't really want to split it up into several lines :D

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -100,4 +105,28 @@ private HttpSecurity configureSecurity(final HttpSecurity http) throws Exception
         .httpBasic(HttpBasicConfigurer::disable)
         .anonymous(AnonymousConfigurer::disable);
   }
+
+  /**
+   * Condition to check if the gateway is configured to use authentication, i.e. is not explicitly
+   * set to {@code NONE}. It helps deal with the fact that the gateway can be embedded in the broker
+   * or run standalone.
+   */
+  static class GatewaySecurityAuthenticationEnabledCondition extends NoneNestedConditions {
```
npepinpe, line 114: :thought_balloon: TIL

npepinpe (approved):
:rocket: Sorry, I assumed that Roman's review was enough :sweat_smile:

PR #19830, opened 2024-06-27
feat: add initial support for the auth-basic profile in Operate

Ben-Sheppard:
## Description
This PR is the first in a series of migrating/removing the Operate security packages so that the central auth layer can be used solely.

In this first PR I wanted to start by just introducing a single class that means a user could set the profiles `operate, auth-basic, ...` and be able to use the new central auth layer. The next steps are to start migration the returned objects and eventually remove the security package from here.

I tested this change locally by starting the StanaloneCamunda application with the profiles `identity, operate, broker, auth-basic, dev, dev-data` and accessing the Operate UI, here I was able to see some instances and with write access, the option to modify them.


operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthBasicUserService.java
```diff
@@ -0,0 +1,63 @@
...
+    permissions.forEach(
+        permissionString -> {
+          if (CAMUNDA_READ_PERMISSIONS.contains(permissionString)
+              && !translatedPermissions.contains(Permission.READ)) {
+            translatedPermissions.add(Permission.READ);
+          } else if (CAMUNDA_WRITE_PERMISSIONS.contains(permissionString)
+              && !translatedPermissions.contains(Permission.WRITE)) {
+            translatedPermissions.add(Permission.WRITE);
+          }
+        });
```
Ben-Sheppard, line 59: Here I have handled the mapping of our new permissions found in `PermissionEnum` and the existing Operate permissions.

I have made assumptions here that (obviously) `*:read` maps to the Operate read permissions but also that any of our destructive permissions `*:create`, `*:update`, '*:delete` maps to the Operate write permission. It felt reasonable to make this assumption now but I realise in the near future would need to be changed.

PR #19839, opened 2024-06-27
feat: install jdk instead of jre [Operate]

ChrisKujawa:
## Description

> [!NOTE]
> This is just an proposal, feel free to reject this.

This PR changes the installation of JRE to JDK.
Installing the JDK brings us some more tools like jcmd, jmap, etc.

This can allows us during incidents, to create heap dumps, thread dumps, profile the JVM via jfr, etc. For example where it was recently needed [here](https://camunda.slack.com/archives/C077P3F763X/p1719484461257309?thread_ts=1719468455.709519&cid=C077P3F763X).

### Example:

When we have the tools installed we can do something like:

```
# find the pid of the JVM process 
jcmd

# Start flight recording; will complete after 100 seconds and write to a file
jcmd 7 JFR.start duration=100s filename=flight.jfr

# Copy from pod
k cp <pod>:/usr/local/operate/flight.jfr /tmp/flight.jfr
```

Without changing the image, nor installing separate dependencies, etc.

### Some key metadata:

 * The jdk is installed 10 mb (5 times bigger) https://pkgs.alpinelinux.org/package/edge/community/s390x/openjdk21-jdk
 * The jre which is ~2MB https://pkgs.alpinelinux.org/package/edge/community/x86_64/openjdk21-jre

#### Changing deployments

We leave it as it is, but it makes it debugging investing much harder. Especially as we have to change deployments, to get root permission to install the right tools, etc.

Example (for SM):

```json
{
    "spec": {
        "template": {
            "spec": {
                "containers": [
                    {
                        "name": "operate",
                        "securityContext": {
                            "readOnlyRootFilesystem": false,
                            "runAsNonRoot": false,
                            "runAsUser": 0
                        }
                    }
                ]
            }
        }
    }
}
```

Furthermore, we need to download then the right dependencies first:

```
# exec into the pod
k exec -it <pod> -- bash

apk upgrade # upgrade packages
apk add openjdk21-jdk # install the OpenJDK to get jcmd
```

#### Building own Image

[We can do it similar to what Zeebe does and build their own image](https://github.com/camunda/camunda/blob/main/Dockerfile)

## Related issues
See [slack](https://camunda.slack.com/archives/C077P3F763X/p1719480283965909?thread_ts=1719468455.709519&cid=C077P3F763X) and related incident

### DIFF
--- a/operate.Dockerfile
+++ b/operate.Dockerfile
@@ -60,7 +60,7 @@ LABEL io.k8s.description="Tool for process observability and troubleshooting pro
 EXPOSE 8080
 
 RUN apk update && apk upgrade
-RUN apk add --no-cache bash openjdk21-jre tzdata
+RUN apk add --no-cache bash openjdk21-jdk tzdata
 
 ENV OPE_HOME=/usr/local/operate
 


operate.Dockerfile
```diff
@@ -60,7 +60,7 @@ LABEL io.k8s.description="Tool for process observability and troubleshooting pro
 EXPOSE 8080
 
 RUN apk update && apk upgrade
-RUN apk add --no-cache bash openjdk21-jre tzdata
```
Zelldon, line 63: If we think JDK is too much we can also try to install the different tools separately.

johanwelgemoed (commented):
Do we consider potential security risks having a full SDK in the image? (eg. attacker gaining access to container, but it now also has development environment and build tool installed which which they can do funky things).
Other than that it should only be the size impact (which does not look that much in the larger scheme of things.
If we feel ok on the security front then I'd say it makes sense to have the tools to make support firefighting easier.

npepinpe:
In terms of security, the issue is not with tools like jmal and jcmd, but rather dependencies or code which the JDK would contain which provide vectors of attack not present in a JRE. I'm not aware of any, but that doesn't mean there wouldn't be in the future. I think the approach in Zeebe and Identity is a good compromise - keep the dev tools (which are Java code anyway, so same vulnerabilities as anything executed by Java) but shed a bunch of otber things.

Zelldon:
Just that i get this right @npepinpe @johanwelgemoed you both would be for the second option "the zeebe way" building the jdk image and linking all of the necessary parts?

npepinpe:
Yes for me, I think it's a good compromise  and it aligns not only with Zeebe but also Identity

Zelldon:
Thanks @npepinpe could we maybe build a common base which can be reused ? Maybe this is also something you/we can provide @cmur2 for the teams?

npepinpe:
As long as Zeebe remains on an Ubuntu or glibc base, all good :)

Zelldon:
Ok, I think this is nothing that I can do just simply now.


I think @johanwelgemoed @sdorokhova you understand best whether you would need it for your investigations (to improve/reduce root-causing. In general I think it is worth it). As you saw above you can either work with SRE together to get root access or bring it into your docker image (long-term better approach). For that, you should try to follow the Zeebe or Identities approach.

Example Zeebe docker image https://github.com/camunda/camunda/blob/main/Dockerfile

With that I will close this PR.

P.S. After 8.6 we might have the benefit of having one docker image, with all tools installed (still of course we need to support older versions).

cmur2:
> P.S. After 8.6 we might have the benefit of having one docker image, with all tools installed (still of course we need to support older versions).

Another approach projects take that care about keeping the attack surface minimal for security (and CVE reviewing reasons) is offering an additional `-debug` image with the dev tooling included - does not help the first time the problem happens, but for anticipated reproducible problems. In practice handling also depends on how C8 SaaS is working internally (how easy is it to switch to functionally identical other Docker image) which I have no insights into.

Zelldon:
> > P.S. After 8.6 we might have the benefit of having one docker image, with all tools installed (still of course we need to support older versions).
> 
> Another approach projects take that care about keeping the attack surface minimal for security (and CVE reviewing reasons) is offering an additional `-debug` image with the dev tooling included - does not help the first time the problem happens, but for anticipated reproducible problems. In practice handling also depends on how C8 SaaS is working internally (how easy is it to switch to functionally identical other Docker image) which I have no insights into.

True. This we could easily provide, just building two images one with jre base and one with the JDK 🤔 During an incident, CRD can be updated and the debug image be used if necessary. 

Drawback is that it always needs a restart of the application, and might not work for non-reproducible bugs. But I guess better than what we have now.

npepinpe:
Why can't we use a debug image with a debug container in SaaS?

Zelldon:
Due to the security policies. I can try again but last time i tried with zdb it was a pain and i was not able to see the process files etc

Zelldon:
I think even the adding of a ephemeral container was not permitted

npepinpe:
Let's check with SREs, possibly we could get a special role/permission for this. I think it's what makes most sense of that's our goal ultimately. Though we may not be able to suggest to SM customers that option...🤔

PR #19852, opened 2024-06-27
refactor: user list call to use the search endpoint

Ben-Sheppard:
## Description

With the merging of the new `/v2` user controller endpoints the Identity client needed changing as the endpoint is not `GET /v2/users` but instead is `POST /v2/users/search`. This PR updates both the endpoint used and also the data response.

### DIFF
--- a/identity/client/src/pages/groups/detail/members/AssignMembersModal.tsx
+++ b/identity/client/src/pages/groups/detail/members/AssignMembersModal.tsx
@@ -4,7 +4,7 @@ import { UseEntityModalCustomProps } from "src/components/modal";
 import { assignGroupMember } from "src/utility/api/membership";
 import useTranslate from "src/utility/localization";
 import { useApi, useApiCall } from "src/utility/api/hooks";
-import { postUserSearch, User } from "src/utility/api/users";
+import { searchUser, User } from "src/utility/api/users";
 import { TranslatedErrorInlineNotification } from "src/components/notifications/InlineNotification";
 import styled from "styled-components";
 import DropdownSearch from "src/components/form/DropdownSearch";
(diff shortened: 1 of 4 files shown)


identity/client/src/pages/groups/detail/members/AssignMembersModal.tsx
```diff
@@ -4,7 +4,7 @@ import { UseEntityModalCustomProps } from "src/components/modal";
 import { assignGroupMember } from "src/utility/api/membership";
 import useTranslate from "src/utility/localization";
 import { useApi, useApiCall } from "src/utility/api/hooks";
-import { getUsers, User } from "src/utility/api/users";
+import { postUserSearch, User } from "src/utility/api/users";
```
maryarm, line 7: Based on our conventions, it's better to be called searchUser.

identity/client/src/pages/users/List.tsx
```diff
@@ -66,7 +71,7 @@ const List: FC = () => {
     <Page>
       <EntityList
         title={t("Users")}
-        data={users}
+        data={userSearchResults?.items || []}
```
maryarm, line 74: I think here we can assume that `userSearchResults?.items` is available as we have checked the length in previous if at line 45.

maryarm (approved):
Thanks for the fix @Ben-Sheppard, the code looks good, only a few suggestions from my side.

identity/client/src/pages/users/List.tsx
Ben-Sheppard: I thought this too but when I leave out the `|| []` it complains that `User[] | undefined` is not assignable to `User[] | null`, I'll see if its me doing something wrong here

maryarm: I see, Can you try with `userSearchResults!.items! `or sth like this?

Ben-Sheppard: That done the trick :)

PR #19765, opened 2024-06-26
feat: group CRUD endpoints in unified api

ghost:
## Description
* Unified API endpoints are available for 
  * creating a group
  * updating a group
  * retrieving a group
  * retrieving all group
  * deleting a group

## Related issues

closes https://github.com/camunda-cloud/identity/issues/2862

PR #19860, opened 2024-06-28
feat: roles crud endpoints in unified api

ghost:
## Description
* Unified API endpoints are available for 
  * creating a role
  * updating a role
  * retrieving a role
  * retrieving all roles
  * deleting a role

## Related issues

closes https://github.com/camunda-cloud/identity/issues/2950


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -484,6 +484,251 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "403":
+          description: "Forbidden"
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "404":
+          description: "Bad request"
```
tmetzke, line 672: 🔧 The 404s should have "Not found" as their description. This occurs a bit more often in here.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -955,6 +1200,42 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/CamundaUserResponse"
+    AbstractRole:
+      type: object
+      properties:
+        name:
+          type: "string"
+        description:
+          type: "string"
```
tmetzke, line 1209: 🔧 We can also integrate those properties directly into the `RoleRequest`, can't we? The `AbstractRole` is only used there.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -133,4 +133,16 @@ public static <T> ResponseEntity<T> mapProblemToResponse(final ProblemDetail pro
...
+    if (e instanceof IllegalArgumentException) {
+      final ProblemDetail problemDetail =
+          ProblemDetail.forStatusAndDetail(HttpStatus.BAD_REQUEST, e.getMessage());
+      return ResponseEntity.of(problemDetail).build();
+    }
+
+    final ProblemDetail problemDetail =
+        ProblemDetail.forStatusAndDetail(HttpStatus.INTERNAL_SERVER_ERROR, e.getMessage());
+    return ResponseEntity.of(problemDetail).build();
+  }
```
tmetzke, line 147: 🙃 Just like mentioned in https://github.com/camunda/camunda/pull/19765, we can make this shorter and consistent with the other controllers by reusing some of the methods from this utility class.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/RoleControllerTest.java
```diff
@@ -0,0 +1,269 @@
...
+        .accept(MediaType.APPLICATION_JSON)
+        .exchange()
+        .expectStatus()
+        .is2xxSuccessful()
+        .expectBody(RoleResponse.class)
+        .isEqualTo(roleResponse);
+  }
+
+  @Test
+  void getRoleByIdThrowsExceptionWhenServiceThrowsException() {
```
tmetzke, line 59: 🔧 You could test the different 400 and 500 status codes for different exceptions.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/RoleControllerTest.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ProblemDetail;
+
+@WebMvcTest(RoleController.class)
+public class RoleControllerTest extends RestControllerTest {
```
tmetzke, line 33: 🔧 There are permissions in the role controller but you are not testing them at all here. There is also some validation logic in the controller that should be tested explicitly.

tmetzke (commented):
⭐ Nice job 👍 
👍 I really like seeing how straightforward it now seems to be to write such controllers 👍 
🔧 There are some suggestions you should look into.
⚠️ Be cautious with the other PR https://github.com/camunda/camunda/pull/19765 because it brings some equal changes. That might lead to merge conflicts. I recommend rebasing either PR, depending on which one is merged first.

tmetzke (approved):
👍 Looks really good, well done. Thanks for addressing the suggestions as well. 
🔧 I left some nitpicks and a suggestion regarding the error message. Nothing blocking though.

tmetzke:
The error mapper tests need the new role service now to function.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/RoleControllerTest.java
```diff
@@ -0,0 +1,357 @@
...
+        .expectBody(ProblemDetail.class)
+        .isEqualTo(expectedBody);
+  }
+
+  @Test
+  void createRoleWithUnknownPermissionShouldThrowException() {
+    final var body =
+        "{\"name\": \"test\", \"description\": \"Test role\", \"permissions\": [\"UNKNOWN\"]}";
+
+    final String message = "Failed to read request";
```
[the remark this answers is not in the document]
Ben-Sheppard: Ah thats the default Spring message AFAIK, I haven't specifically set it. I'll look though to see if we can make it nicer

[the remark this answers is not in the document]
Ben-Sheppard: Ah I've just seen the `GlobalControllerExceptionHandler` class that already handles something like this, I'll add it in there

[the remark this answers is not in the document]
Ben-Sheppard: I've added handling for the `JSON parse error` but I'm not sure if we want to be more specific or if what I've added is ok for now?

Ben-Sheppard:
Ugh, I think this is another case of being unable to merge due to the owner being "ghost". As with the other PR, I will reopen a new PR for these changes in my name

PR #19868, opened 2024-06-28
fix: scan identity packages whenever the app is started

romansmirnov:
## Description

This PR:
* Adds the Identity module scanning to the commons package to resolve issues with beans missing when running StandaloneTasklist/Operate.
* Moves the new Identity package to something that does not clash with the existing Identity packaging, this way we can remove the exclusion of the SDK packages as we are scanning on a different package path


maryarm (approved):
Looks good to me. Moving of the identity configurations to the common makes sense and I specially like the new package name of identity. :)

PR #19876, opened 2024-06-28
Stream processor failure proper rejection

### LINKED ISSUE #16180: Send rejection if partition's stream processor is not active
When a stream processor has crashed and is not processing anymore, the command API still accepts requests and writes them to the log stream. Since nothing is handling the request, clients will simply see a timeout.

Instead, the command api could reject the request if it already knows that the stream processor won't be able to process it.

This came up in https://app.incident.io/camunda/incidents/876 because the timeouts were misleading.

EuroLew:
## Description

If a partition's stream processor dies it is now set to `DEAD` and corresponding error messages now inform the user that the partition is inactive.

## Related issues

closes #16180

### DIFF
--- a/zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
+++ b/zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
@@ -214,6 +214,7 @@ final class BrokerRequestManager extends Actor {
       if (topology != null && !topology.getPartitions().contains(request.getPartitionId())) {
         throw new PartitionNotFoundException(request.getPartitionId());
       }
+      throwIfPartitionInactive(request.getPartitionId());
       // already know partition id
       return new BrokerAddressProvider(request.getPartitionId());
     } else if (request.requiresPartitionId()) {
@@ -222,10 +223,7 @@ final class BrokerRequestManager extends Actor {
       // select next partition id for request
       int partitionId = strategy.determinePartition(topologyManager);
 
-      final BrokerClusterState topology = topologyManager.getTopology();
-      if (topology != null && topology.getInactiveNodesForPartition(partitionId) != null) {
-        throw new PartitionInactiveException(partitionId);
-      }
+      throwIfPartitionInactive(partitionId);
 
       if (partitionId == BrokerClusterState.PARTITION_ID_NULL) {
         // could happen if the topology is not set yet, let's just try with partition 0 but we
@@ -239,6 +237,13 @@ final class BrokerRequestManager extends Actor {
     } else {
       // random broker
       return new BrokerAddressProvider();
+    }
+  }
+
+  private void throwIfPartitionInactive(final int partitionId) {
+    final BrokerClusterState topology = topologyManager.getTopology();
+    if (topology != null && topology.getInactiveNodesForPartition(partitionId) != null) {
+      throw new PartitionInactiveException(partitionId);
     }
   }
 
--- a/zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
+++ b/zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
@@ -222,7 +222,6 @@ final class BrokerRequestManager extends Actor {
 
       // select next partition id for request
       int partitionId = strategy.determinePartition(topologyManager);
-
       if (partitionId == BrokerClusterState.PARTITION_ID_NULL) {
         // could happen if the topology is not set yet, let's just try with partition 0 but we
         // should find a better solution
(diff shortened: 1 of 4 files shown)


zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
```diff
@@ -230,6 +240,13 @@ private BrokerAddressProvider determineBrokerNodeIdProvider(final BrokerRequest<
     }
   }
 
+  private void throwIfPartitionInactive(final int partitionId) {
+    final BrokerClusterState topology = topologyManager.getTopology();
+    if (topology != null && topology.getInactiveNodesForPartition(partitionId) != null) {
+      throw new PartitionInactiveException(partitionId);
+    }
+  }
```
deepthidevaki, line 248: :x: This is will throw even if only one replica is Inactive, but the other two replicas are healthy and there is a leader. To be more precise we can do something like : "throw if there is atleast one inactive replica AND there is no healthy leader". 

The chances of only one replica becoming inactive is very less, but we shouldn't unnecessarily reject a request if there is a healthy leader.

zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
```diff
@@ -208,13 +214,17 @@ private BrokerAddressProvider determineBrokerNodeIdProvider(final BrokerRequest<
...
+      throwIfPartitionInactive(request.getPartitionId());
       // already know partition id
       return new BrokerAddressProvider(request.getPartitionId());
     } else if (request.requiresPartitionId()) {
       final var strategy = request.requestDispatchStrategy().orElse(dispatchStrategy);
 
       // select next partition id for request
       int partitionId = strategy.determinePartition(topologyManager);
+
+      throwIfPartitionInactive(partitionId);
```
deepthidevaki, line 226: :x: Move this after the following null check.

zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
deepthidevaki: Also `topology.getInactiveNodesForPartition(partitionId) != null` this will be always true, because the returned set is never null. You should check if the set is empty.

deepthidevaki (changes requested):
It is a nice solution to do the inactive check in the BrokerClient and throw a dedicated exception. However, please see my comment to ensure that we don't unnecessarily reject a request.

zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
EuroLew: I'll also write a test for this case.

deepthidevaki:
@EuroLew Could you run a test with 3 replicas with these changes while simulating a bug in StreamProcessor? You can either run it locally, or run a benchmark. Please document the observed behavior in the issue.

zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
EuroLew: > Also `topology.getInactiveNodesForPartition(partitionId) != null` this will be always true, because the returned set is never null. You should check if the set is empty.

I thought it is possible to be null because `partitionInactiveNodes` is a map and the `.get` call can return null, also for some `BrokerClientTest` the return value  from `.getInactiveNodesForPartition(id)` will be null 
Screenshot 2024-07-01 at 09 52 59

image

EuroLew:
> @EuroLew Could you run a test with 3 replicas with these changes while simulating a bug in StreamProcessor? You can either run it locally, or run a benchmark. Please document the observed behavior in the issue.

done

zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/BrokerRequestManager.java
```diff
@@ -223,13 +230,26 @@ private BrokerAddressProvider determineBrokerNodeIdProvider(final BrokerRequest<
...
       return new BrokerAddressProvider(request.getPartitionId());
     } else {
       // random broker
       return new BrokerAddressProvider();
     }
   }
 
+  private void throwIfPartitionInactive(final int partitionId) {
+    final BrokerClusterState topology = topologyManager.getTopology();
+    final var inactiveNodes = topology.getInactiveNodesForPartition(partitionId);
```
deepthidevaki, line 244: :x: The check `topology!=null` should be before this step.

zeebe/broker-client/src/test/java/io/camunda/zeebe/broker/client/api/BrokerClientTest.java
```diff
@@ -339,6 +340,66 @@ void shouldReceiveJobAvailableNotification() {
...
+                    .isNotEmpty());
+
+    // then
+    assertThatCode(() -> client.sendRequest(request).join())
+        .isInstanceOf(CompletionException.class)
+        .hasMessageContaining("The partition " + partitionId + " is currently INACTIVE");
+  }
+
+  @Test
+  public void shouldSendRequestToLeaderIfSomePartitionReplicasInactive() {
```
deepthidevaki, line 365: :+1:

deepthidevaki (approved):
:100: 
:x: Please address my comment before merging.

PR #19882, opened 2024-06-28
ci: add custom matcher for Playwright script

vsgoulart:
We have references to the Playwright docker image on [Operate](https://github.com/camunda/camunda/blob/main/operate/client/package.json#L57) and [Tasklist](https://github.com/camunda/camunda/blob/main/tasklist/client/package.json#L53)
Because they're in the `package.json` in an NPM script Renovate can't recognize it, so I'm adding a custom matcher

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -175,7 +175,7 @@
       "customType": "regex",
       "fileMatch": ["package\\.json"],
       "matchStrings": [
-        "mcr\\.microsoft\\.com\\\/playwright:(?<currentValue>.*?) \\\/bin\\\/bash"
+        "mcr\\.microsoft\\.com\\\/playwright:(?<currentValue>.*?)\\s"
       ],
       "depNameTemplate": "mcr.microsoft.com/playwright",
       "datasourceTemplate": "docker"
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -170,6 +170,15 @@
         "renovate: datasource=(?<datasource>.*?) depName=(?<depName>.*?)( versioning=(?<versioning>.*?))?\\s.* '(?<currentValue>.*)'\\s"
       ],
       "versioningTemplate": "{{#if versioning}}{{{versioning}}}{{else}}semver{{/if}}"
+    },
+    {
+      "customType": "regex",
+      "fileMatch": ["package\\.json"],
+      "matchStrings": [
+        "mcr\\.microsoft\\.com\\\/playwright:(?<currentValue>.*?) \\\/bin\\\/bash"
```
cmur2, line 178: Should work generally but it might be more robust to change the end from ` \\\/bin\\\/bash` to `\\s` only to avoid relying on the subsequent command string 🤔

vsgoulart: @cmur2 done

cmur2 (approved):

PR #19907, opened 2024-07-01
fix: use and spring profile operator

houssain-barouni:
The AND operator in Spring profile expression is `&` and not `&&`https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Profiles.html

## Related issues

closes #


ChrisKujawa (approved):
Good catch 👍🏼

PR #19735, opened 2024-06-25
Add exists query test

ChrisKujawa:
## Description

- Simplify SearchExistsChildQuery, remove some complexity
- Add descriptive error messages for null checks
- Add parameterized tests for has child query and transformer

## Related issues

related to https://github.com/camunda/camunda/issues/19076

PR #19733, opened 2024-06-25
Add HasChild query test (ES)

ChrisKujawa:
## Description

* Simplify SearchHasChildQuery, remove some unused/complexity
* Add descriptive error messages for null checks
* Add parameterized tests for has child query and transformer

## Related issues
related to https://github.com/camunda/camunda/issues/19076

PR #19908, opened 2024-07-01
Add search client Query tests [OS]

ChrisKujawa:
## Description

Migrates the Search Client query tests from Elasticsearch to Opensearch.

Unfortunately, we can't simply use the `toString` for the Queries as Opensearch doesn't support this (or at least doesn't implement it). We have to use some JsonSerializer magic, to get the Query in a serialized format such that we can better assert the result.

This PR covers all search client query transformations:

 * https://github.com/camunda/camunda/pull/19735
 * https://github.com/camunda/camunda/pull/19733
 * https://github.com/camunda/camunda/pull/19730
 * https://github.com/camunda/camunda/pull/19724
 * https://github.com/camunda/camunda/pull/19720
 * https://github.com/camunda/camunda/pull/19682
 * https://github.com/camunda/camunda/pull/19764
 * https://github.com/camunda/camunda/pull/19736
 * https://github.com/camunda/camunda/pull/19718
 * https://github.com/camunda/camunda/pull/19682
 * https://github.com/camunda/camunda/pull/19663
 * https://github.com/camunda/camunda/pull/19678

## Related issues

### DIFF
--- a/search/search-client-opensearch/src/main/java/io/camunda/search/os/transformers/query/TermQueryTransformer.java
+++ b/search/search-client-opensearch/src/main/java/io/camunda/search/os/transformers/query/TermQueryTransformer.java
@@ -24,6 +24,10 @@ public final class TermQueryTransformer extends QueryOptionTransformer<SearchTer
     final var transformer = getFieldValueTransformer();
     final var fieldValue = value.value();
     final var tranformedFieldValue = transformer.apply(fieldValue);
-    return QueryBuilders.term().field(field).value(tranformedFieldValue).build();
+    return QueryBuilders.term()
+        .field(field)
+        .value(tranformedFieldValue)
+        .caseInsensitive(value.caseInsensitive())
+        .build();
   }
 }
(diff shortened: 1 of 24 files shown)


Zelldon:
It looks quite big, but it is essential almost the same as what you have already reviewed. I'm happy to split it up if you like me to do.

search/search-client-opensearch/src/main/java/io/camunda/search/os/transformers/query/TermQueryTransformer.java
```diff
@@ -24,6 +24,10 @@ public TermQuery apply(final SearchTermQuery value) {
     final var transformer = getFieldValueTransformer();
     final var fieldValue = value.value();
     final var tranformedFieldValue = transformer.apply(fieldValue);
-    return QueryBuilders.term().field(field).value(tranformedFieldValue).build();
+    return QueryBuilders.term()
+        .field(field)
+        .value(tranformedFieldValue)
+        .caseInsensitive(value.caseInsensitive())
```
houssain-barouni, line 30: I wonder why is we can use primitive boolean for `caseInsensitive`, to avoid handling the null case

houssain-barouni (approved):

search/search-client-opensearch/src/main/java/io/camunda/search/os/transformers/query/TermQueryTransformer.java
Zelldon: Null is fine here. This is handled by OS and ES client correctly.

PR #19568, opened 2024-06-20
Optionally limit Zeebe's overall write rate

lenaschoenburg:
## Description

Replaces the append limit with its rather ineffective default configuration with an optional, disabled by default, write rate limit. It can be configured like this:

```yaml
zeebe.broker.flowControl.write:
  enabled: false
  rampUp: 0 # ramp up time, for example 10s
  limit: 1000 # maximum number of records written per second
```

Also new is a simple to understand metric at `zeebe_flow_control_outcome`, labeled by partition, context (i.e. user command or processing result) and outcome (i.e. rejected/accepted). This counter increments by the amount of records accepted or rejected.

## Related issues

closes https://github.com/camunda/camunda/issues/19680
closes https://github.com/camunda/zeebe/issues/17669

PR #19915, opened 2024-07-01
feat: measure average exporting rate

lenaschoenburg:
## Description

Measure exporting rate so that we can later adjust write rate to match. A new metric is added but not yet visualized.

## Related issues

depends on https://github.com/camunda/camunda/pull/19568
relates to https://github.com/camunda/camunda/issues/19681


rodrigo-lourenco-lopes (approved):
LGTM! 🚀 🚀

PR #19957, opened 2024-07-02
Ensure instance migration can't detach boundary event from element instance

### LINKED ISSUE #19956: Ensure instance migration can't detach boundary event from element instance
Part of #19640 

>Map a message boundary event that is attached to a different element in the target as it was in the source.

korthout:
## Description

Adds an additional validation to instance migration.

If boundary event would become detached from element instance, the `MIGRATE` command is rejected with [`GRPC_STATUS_FAILED_PRECONDITION`](https://docs.camunda.io/docs/next/apis-tools/zeebe-api/gateway-service/#grpc_status_failed_precondition-3).

I've taken the liberty to clean up some precondition error templates.

The `gateway.proto` documentation is adjusted accordingly. We'll also need to update the docs at https://docs.camunda.io.

> [!NOTE]
> The `gateway.proto` file was not completely synced with https://docs.camunda.io. To align them, I've added:
> - a mapping instruction changes the implementation of a task

## Related issues

closes #19956

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
@@ -53,16 +53,16 @@ public final class ProcessInstanceMigrationPreconditions {
       "Expected to migrate process instance '%s' but the mapping instructions contain duplicate source element ids '%s'.";
   private static final String ERROR_SOURCE_ELEMENT_ID_NOT_FOUND =
       """
-              Expected to migrate process instance '%s' \
-              but mapping instructions contain a non-existing source element id '%s'. \
-              Elements provided in mapping instructions must exist \
-              in the source process definition.""";
+      Expected to migrate process instance '%s' \
+      but mapping instructions contain a non-existing source element id '%s'. \
+      Elements provided in mapping instructions must exist \
+      in the source process definition.""";
   private static final String ERROR_TARGET_ELEMENT_ID_NOT_FOUND =
       """
-              Expected to migrate process instance '%s' \
-              but mapping instructions contain a non-existing target element id '%s'. \
-              Elements provided in mapping instructions must exist \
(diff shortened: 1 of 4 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
```diff
@@ -494,6 +502,47 @@ private static void requireNoBoundaryEvent(
     }
   }
 
+  public static void requireMappedBoundaryEventsToStayAttachedToSameElement(
```
berkaycanbc, line 521: 🔧 This naming explains "what will happen as a result (boundary event will be detached wrongly)". A name that explains what the method does would be a better choice. E.g. `requireMappedBoundaryEventsExistInTarget` which explains what the method does. WDYT?

berkaycanbc (changes requested):
Implementation LGTM @korthout! 🚀 

I requested a naming change which I think, is still important to discuss.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
korthout: 🤔 But `requireMappedBoundaryEventsExistInTarget` is not correct though. It's not testing whether or not the mapped boundary event exists in the target. It checks that the mapped boundary event stays attached to the same element after mapping, i.e. that the boundary event is not detached from the element it was initially attached to.

Do you see any other options?

berkaycanbc: Yeah, I think, the name and the error message you chose is more user friendly. It actually indicates something that cannot be done by user when they try to move the boundary event to a different task. E.g. 
Screenshot 2024-07-04 at 19 54 51

berkaycanbc (approved):

PR #19958, opened 2024-07-02
chore(optimize8): remove qa module

buccarel:
# Description

This PR is removing the module `qa` completely from Optimize.
Together with the module, also all of its consumers are being removed.

In addition to this, it's also updating the Java version used for `spotless:check`, because `Java 17` is producing an error that does not appear to be real in the offending file.

# Issues

https://github.com/camunda/camunda-optimize/issues/13385


HeleneW-dot (approved):
lgtm

PR #19965, opened 2024-07-02
Migrate timer boundary event subscriptions

### LINKED ISSUE #15918: Migrate timer boundary event subscriptions
It should be possible to migrate timer catch event subscriptions, similar to migrating a message subscription, by providing a mapping instructions for the related message catch event.

Consider the following example, where an active user task is subscribed to a timer boundary event (interrupting). When migrating the process instance, mapping instructions are provided for both the user task and the timer boundary event. After the instance is migrated, the user task is now subscribed to the attached timer boundary event (non-interrupting).

Screenshot 2024-06-25 at 11 07 05

We can introduce a new `Timer:MIGRATED` intent to represent the subscription migration.

Subscriptions to timer boundary events and timer event subprocesses for which no mapping instructions are provided will not be migrated. Similar to these cases for message subscriptions, we can unsubscribe and subscribe to these catch events. Unsubscribing and subscribing can be left out of scope for this issue.

berkaycanbc:
## Description

Enables the specification of mapping instructions for timer boundary events during process instance migration. This feature ensures that the timer boundary event is migrated to the target process definition.

During migration, certain properties of the timer are adjusted using a new `MIGRATED` intent:

- `processDefinitionKey` is updated to the definition key of the target process
- `targetElementId` is updated to the ID of the target boundary event
- `interrupting` status is updated to match the interrupting type of the target boundary event (either interrupting or non-interrupting)
  - This adjustment ensures that when the timer is triggered, the boundary event will perform as defined in the target process.

During migration, the timer duration (due date) and expressions are **not** re-evaluated. This approach ensures alignment with [our documentation on expressions](https://docs.camunda.io/docs/next/components/concepts/process-instance-migration/#jobs-expressions-and-input-mappings) and provides a consistent experience across all migrations.

## Related issues

closes #15918

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java
@@ -404,15 +404,21 @@ public final class CatchEventBehavior {
         });
   }
 
-  private void unsubscribeFromTimerEvents(
-      final long elementInstanceKey, final Predicate<DirectBuffer> elementIdFilter) {
+  public void unsubscribeFromTimerEventsByInstanceFilter(
+      final long elementInstanceKey, final Predicate<TimerInstance> timerInstanceFilter) {
     timerInstanceState.forEachTimerForElementInstance(
         elementInstanceKey,
         timer -> {
-          if (elementIdFilter.test(timer.getHandlerNodeId())) {
+          if (timerInstanceFilter.test(timer)) {
             unsubscribeFromTimerEvent(timer);
           }
         });
+  }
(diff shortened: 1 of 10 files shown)


korthout (approved):
Thanks @berkaycanbc ⏰ 

To make progress, I've directly committed my change requests rather than provide review comments:
- 5a8b17f30c790008d19f45144d386fc21f2773f0
- a48779b9e384d5f6ea77ee282f6faf2cf48d2ed9
- 28a3ee06a6f1306d9311bd1886f8fdfc0bdba43d

LGTM 👍

berkaycanbc:
Thanks for changes @korthout! They are all makes total sense. 🚀

PR #19970, opened 2024-07-02
feat: crud roles identity unified api

Ben-Sheppard:
## Description

This is a new PR in place of https://github.com/camunda/camunda/pull/19860 as I am unable to merge that PR (potentially due to the owner being "ghost").


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
```diff
@@ -42,6 +43,11 @@ protected ProblemDetail createProblemDetail(
       // with "Required request body is missing"
       // for proper exception tracing
       detail = REQUEST_BODY_MISSING_EXCEPTION_MESSAGE;
+    } else if (isJsonParseError(ex)) {
+      // Replace detail "Failed to read request"
+      // with "JSON parse error"
+      // for proper exception tracing
+      detail = "Unable to parse JSON request body";
```
tmetzke, line 50: ❓ What's the full original error message? Could we maybe detect that we have an enum problem so we can hint at that? I guess the more detailed the information, the easier users can fix their requests. Just trying to explore what is possible here.

tmetzke (approved):
👍 Good job, code looks good.

🔧 The enum error handling is not very specific at the moment. Users will have issues figuring out what is wrong with their request. Let's see if we can improve this.

⚠️ Heads-up: the OpenAPI description is missing a lot `description` attributes for paths and data objects at the moment. While this is fine for the moment, please create a follow-up issue to tackle this. We will generate the interactive REST API explorers of the public documentation from that API file. Thus, it must contain the descriptions for all elements to provide a good user experience in the docs and avoid manual overrides when writing docs.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
Ben-Sheppard: The error that we get is
```
JSON parse error: Cannot construct instance of `io.camunda.zeebe.gateway.protocol.rest.Permission`, problem: Unexpected value 'UNKNOWN'
```
with a root cause of
```
java.lang.IllegalArgumentException: Unexpected value 'UNKNOWN'
```
I'm not sure we can confidently make assumptions here, but what if we took the original message and returned an error of something like 
```
Unexpected value 'UNKNOWN' for class 'io.camunda.zeebe.gateway.protocol.rest.Permission'
```

tmetzke: 👍 OK, I see, thanks for those details!
👍 I like the proposal in general.
🤔 Exposing the full class name including the package might not be helpful to users, though. In the REST API explorers in the public docs and the OpenAPI description, they will only see the type `Permission`, I guess? Will they even see this as a dedicated type or just an array of String options for that parameter, e.g. when using SwaggerUI to look at the OpenAPI description?

But it might become a bit error-prone if we do something like the following
```
Unexpected value 'UNKNOWN' for type 'Permission'
```

Ben-Sheppard: Looking at the openapi spec generated the user is aware of the type 
Screenshot 2024-07-03 at 08 30 34
so we can try to make this a little more intelligent and pull parts of the information out. I'll have a closer look :)

tmetzke: Nice, I see. Yeah, that would be super helpful, I think. Especially, if this is generalized in the global mapper.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
```diff
@@ -62,6 +71,33 @@ private boolean isRequestBodyMissing(final Exception ex) {
...
+      return exceptionMessage != null
+          && exceptionMessage.startsWith(JSON_CANNOT_CONSTRUCT_INSTANCE_EXCEPTION_MESSAGE)
+          && exceptionContainsEnumClass(exceptionMessage);
+    }
+
+    return false;
+  }
+
+  private boolean exceptionContainsEnumClass(final String exceptionMessage) {
+    return ENUM_CLASS_TO_PRETTY_NAME.keySet().stream().anyMatch(exceptionMessage::contains);
```
tmetzke, line 87: ❓ Could we simply use `containsKey` on the map?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/RoleController.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private PermissionEnum mapToPermissionEnum(final Permission permission) {
+    return switch (permission) {
+      case READ_ALL -> PermissionEnum.READ_ALL;
+      case CREATE_ALL -> PermissionEnum.CREATE_ALL;
+      case DELETE_ALL -> PermissionEnum.DELETE_ALL;
+      case UPDATE_ALL -> PermissionEnum.UPDATE_ALL;
+    };
+  }
```
tmetzke, line 132: ❌ This is obsolete now, I guess 🙂

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/RoleController.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private Permission mapToPermission(final PermissionEnum permissionEnum) {
+    return switch (permissionEnum) {
+      case READ_ALL -> Permission.READ_ALL;
+      case CREATE_ALL -> Permission.CREATE_ALL;
+      case DELETE_ALL -> Permission.DELETE_ALL;
+      case UPDATE_ALL -> Permission.UPDATE_ALL;
+    };
+  }
```
tmetzke, line 154: ❌ This is obsolete now, I guess 🙂

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/RoleControllerTest.java
```diff
@@ -0,0 +1,370 @@
...
+          {
+            "name": "test",
+            "description": "Test role",
+            "permissions": ["UNKNOWN"]
+          }
+        """;
+
+    final String message =
+        """
+        Invalid Enum value "UNKNOWN" for Permission, expected one of [CREATE_ALL, READ_ALL, UPDATE_ALL, DELETE_ALL]""";
```
tmetzke, line 175: 👍 Awesome ⭐

Ben-Sheppard:
I've raised https://github.com/camunda/camunda/issues/19991 to address your comments on the descriptions required for the Open API spec @tmetzke

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/RoleController.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private PermissionEnum mapToPermissionEnum(final Permission permission) {
+    return switch (permission) {
+      case READ_ALL -> PermissionEnum.READ_ALL;
+      case CREATE_ALL -> PermissionEnum.CREATE_ALL;
+      case DELETE_ALL -> PermissionEnum.DELETE_ALL;
+      case UPDATE_ALL -> PermissionEnum.UPDATE_ALL;
+    };
+  }
```
tmetzke, line 126: ❌ This one is still left over 🙂

tmetzke (approved):
👍 Nice job, I really like the final enum handling ⭐

PR #19978, opened 2024-07-03
test: simplify REST error mapper test

### LINKED ISSUE #19977: Simplify REST test setup to improve test isolation
**Description**

* The current test setup requires adjustments in the `ErrorMapperTest` with every new dependency added to REST controllers.
* That test class can be adjusted to be a `WebMvcTest` instead of a full `SpringBootTest` so only minimal context needs to be provided.
* The test application required by all REST controller tests can be externalized.

tmetzke:
## Description

Removes unnecessary context from the ErrorMapperTest in the REST gateway. This way, the test class needs no adjustments anymore for controller dependency adjustments.

Moves the test application required by all REST controller tests out of the ErrorMapperTest.

## Related issues

closes #19977

### DIFF
--- a/zeebe/gateway-rest/pom.xml
+++ b/zeebe/gateway-rest/pom.xml
@@ -146,12 +146,6 @@
       <scope>test</scope>
     </dependency>
 
-    <dependency>
-      <groupId>org.springframework.security</groupId>
-      <artifactId>spring-security-config</artifactId>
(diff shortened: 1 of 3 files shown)


Ben-Sheppard (approved):
Thank you for refactoring this, definitely a cleaner approach 👍🏻

PR #19409, opened 2024-06-16
Feat: Add endpoint to configure and get flow control

rodrigo-lourenco-lopes:
## Description

This pull request adds an endpoint to set and get the flow control configuration.

The get method returns the several flow control limits configured for each partition and broker.

The set method will use the configuration that we send in the payload in JSON to set the limits of all partitions in all brokers. To note that since the append limit cannot be null, if this is not present in the JSON payload or is set as null, then it will not be changed. In case the JSON is not parseable to the object FlowControl configuration, the error message will be return along with an example of a configuration.
## Related issues

closes https://github.com/camunda/camunda/issues/18731

PR #19980, opened 2024-07-03
refactor: improve flow control actuator serialization

lenaschoenburg:
A couple of improvements on top of https://github.com/camunda/camunda/pull/19409, mainly to improve serialization.

There are couple of remaining todos:
1. Adjust existing tests.
2. Merge the two actuator clients now that get and set both return JSON.
3. Probably more that I forgot.

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -53,6 +53,11 @@
 
     <dependency>
       <groupId>io.camunda</groupId>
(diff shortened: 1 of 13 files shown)


rodrigo-lourenco-lopes (approved):
Thanks for the great improvements in the serializer! 🚀

Issue #18378, opened 2024-05-08
Operate exporter: rework tree path fields


sdorokhova:
### Description

We need to fully get rid of tree path caches and call activity id cache.
Details are in the document: https://docs.google.com/document/d/1Nh8Jln5_aWIS1O5NnC-SGRblijizDTSujG7zIfc-GOY/edit

* [x] #24061
* [x] #24063
* [x] #25223
* [x] #24542
* [x] #24733
* [x] #25237

### Additional Information

Related Zeebe changes with description of the new incident fields: #16915 and #20293

ThorbenLindhauer:
Moving this to Blocked, because it requires Zeebe work first.

Issue #18372, opened 2024-05-08
Camunda exporter: propagate an `OperationReference` along with Zeebe commands, and use it to mark operations as completed


sdorokhova:
### Description

Currently when importing some of Zeebe records we need to complete corresponding operations (e.g. [when process instance is cancelled](https://github.com/camunda/zeebe/blob/main/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/ListViewZeebeRecordProcessor.java#L261)). For that we first [find operations to update](https://github.com/camunda/zeebe/blob/main/operate/schema/src/main/java/io/camunda/operate/util/OperationsManager.java#L94) and then send an update request. Since we must avoid any queries from Elasticsearch when performing Operate export we need to rework this logic and update operations based on an `operationReference` that operations should pass along with zeebe commands and expect it back in the exported records.
 
Depends on [#16916](https://github.com/camunda/zeebe/issues/16916).

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/issues/24027
- [ ] https://github.com/camunda/camunda/issues/24028
- [ ] https://github.com/camunda/camunda/issues/24255
- [ ] https://github.com/camunda/camunda/issues/24029
- [ ] https://github.com/camunda/camunda/issues/24030
```

mustafadagher:
I had a look at this issue partly yesterday and today, and I have a few observations:
1. The ticket title and description are outdated. As per https://github.com/camunda/camunda/pull/18669, we allowed zeebe clients/users to provide an `operationReference` to commands.
2. We do not care then about `requestId`s or `partitionId`, and we should not expect zeebe responses to contain them.

The task breakdown should now be:
1. We need to find a way to create a unique id (long), use it as a reference to send to with commands to zeebe.
2. We need to store this reference somewhere (probably in operation entity)
3. use that operation reference to co-relate between commands and records handled

The open questions here would be, 
- where (or at which point) to generate this unique key? 
- and is the operation entity the best place to store it? is it actually stored somewhere? or only in the runtime?
- When we say `mark operation as completed`, what do we actually mean? just setting the `operationEntity.state` or should we add it to a batch again? what does that entail? I'm confused because I had a quick look at the `OperationStore` in operate, and I noticed that it uses es/os. Are this going to stay the same?

CC: @Zelldon @sdorokhova

sdorokhova:
[...]

Issue #18364, opened 2024-05-08
Operate exporter: rework ProcessCache


sdorokhova:
### Description

Currently we use `ProcessCache` to store/get process definition names and flow node names. The cache has limited size and therefore the entries may be evicted and be requested from Elasticsearch again. We need to make it as performant as possible:  minimize the data stored in cache and fill it in the most performant manner.

### Tasks

Based on proposal 1 described here: https://docs.google.com/document/d/1No2WPUtLg22x9CFlw2dA8uJmk7wwNC50MsuHenAkHis/edit#heading=h.dtgya6ciylb

- [x] #24154
- [x] #24179
- [x] #24208
- [x] #24345
- [x] #24346

##### Optimizations (next iteration)

- [ ] Asynchronously fill cache on startup to reduce chances of cache miss during exporting.
- [ ] Shared cache for all partitions in the same broker.

### Additional Information

There is some description in Slack: https://camunda.slack.com/archives/C06F0GLJNFM/p1720788578770879

sdorokhova:
Will put the older (outdated) TODO items here, some of the points may be valuable during implementation: 
* rework current `ProcessCache` structure to normal cache (without size limitations and evictions)
* fill the cache on exporter startup from exporter metadata
* add data to cache during export (when new process definitions are being processed)
* [update exporter metadata](https://github.com/camunda/zeebe/blob/fa3919f106b653f02dec43365450679719edca4f/zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/context/Controller.java#L40) with every persisted entities batch
* TBD can we remove data from cache when process definitions are deleted? We did not process process definition deletion records so far. Could we start doing that to keep cache up to date?

deepthidevaki:
Following our discussion in https://docs.google.com/document/d/1No2WPUtLg22x9CFlw2dA8uJmk7wwNC50MsuHenAkHis/edit#heading=h.dtgya6ciylb, we have to provide the following interface in the ProcessCache.

```
String getProcessName(final long processDefinitionKey);
void put(final long processDefinitionKey, final String processName);
void delete(final long processDefinitionKey);
void clear();
```

In the current implementation of ProcessCache in Operate, I see the following methods. 
```
String getProcessNameOrDefaultValue(Long processDefinitionKey, String defaultValue);
String getProcessNameOrBpmnProcessId(Long processDefinitionKey, String defaultValue);
String getProcessVersionTag(Long processDefinitionKey);

[...]

PR #19983, opened 2024-07-03
feat: Exporter business logic for list-view index

### LINKED ISSUE #18908: Copy business logic for `list-view` index from importer to exporter
### Description

As a continuation for #18886 we need to copy all importer logic to corresponding `ExportHandlers`.

### TODO

* move logic from `ListViewZeebeRecordProcessor` (latest from `importer-8_6` module) to several corresponding `ExportHandlers` (e.g. `ListViewFromIncidentHandler`, `ListViewFromVariableHandler` etc.)

### AT

* when creating the process instance with variable, with incident, all the necessary data is exported to Operate `list-view` index

### Additional Information

* check `ProcessHandler` as an example

### Acceptance Criteria

ghost:
feat: Exporter business logic for list-view index

- added handlers for list-view index

Closes https://github.com/camunda/camunda/issues/18908

### DIFF
--- a/zeebe/exporters/operate-exporter/pom.xml
+++ b/zeebe/exporters/operate-exporter/pom.xml
@@ -25,6 +25,11 @@
   <dependencies>
 
     <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>io.camunda</groupId>
       <artifactId>operate-schema</artifactId>
     </dependency>
--- a/zeebe/exporters/operate-exporter/src/main/java/io/camunda/zeebe/operate/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
+++ b/zeebe/exporters/operate-exporter/src/main/java/io/camunda/zeebe/operate/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
@@ -1,0 +1,336 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.operate.exporter.handlers;
+
+import static io.camunda.operate.schema.templates.ListViewTemplate.*;
+import static io.camunda.operate.schema.templates.ListViewTemplate.STATE;
+import static io.camunda.operate.schema.templates.TemplateDescriptor.POSITION;
+import static io.camunda.zeebe.operate.exporter.util.OperateExportUtil.tenantOrDefault;
+import static io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent.*;
+
+import io.camunda.operate.cache.ProcessCache;
+import io.camunda.operate.entities.OperationType;
+import io.camunda.operate.entities.listview.ProcessInstanceForListViewEntity;
+import io.camunda.operate.entities.listview.ProcessInstanceState;
+import io.camunda.operate.exceptions.OperateRuntimeException;
+import io.camunda.operate.exceptions.PersistenceException;
+import io.camunda.operate.schema.templates.ListViewTemplate;
+import io.camunda.operate.store.FlowNodeStore;
+import io.camunda.operate.store.ListViewStore;
+import io.camunda.operate.store.elasticsearch.NewElasticsearchBatchRequest;
+import io.camunda.operate.util.*;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.value.BpmnElementType;
+import io.camunda.zeebe.protocol.record.value.ProcessInstanceRecordValue;
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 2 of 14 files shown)


mihail-ca:
The handlers are based on the following importer methods, in the importer class ListViewZeebeRecordProcessor.java:

```
ListViewFlowNodeFromIncidentHandler  				--- based on processIncidentRecord()
ListViewFlowNodeFromJobHandler  					--- based on processJobRecords()
ListViewFlowNodeFromProcesInstanceHandler  			--- based on processProcessInstanceRecord()
ListViewProcessInstanceFromProcessInstanceHandler  	--- based on processProcessInstanceRecord()
ListViewVariableFromVariableHandler  				--- based on processVariableRecords()
```

The handler `ListViewProcessInstanceFromProcessInstanceHandler` is problematic, I implemented as much as I could. But as you can see, it has multiple external dependencies, added via constructor. I am not sure the implementation is correct here, also which importer code should have been added to the method `updateEntity()`, and which to the method `flush()`. Also in the importer code, this handler was using a `ProcessCache`, which I am not sure I imported correctly. I did not add this handler to the `OperateElasticsearchExporter.java`, as it has too many dependencies, and the implementation may need rework. 
FYI @sdorokhova @ThorbenLindhauer

sdorokhova:
I will document here what we discussed about `ListViewProcessInstanceFromProcessInstanceHandler`:
* `OperationsManager` dependency. We will rework (simplify) how we complete operations in #18372. Therefore for now I propose to comment out the calls to `OperationsManager`. But the code block should be moved to `update` method, so that we don't need to have `recordsMap` local cache.
* same case with `ProcessCache`. We will rework it in #18364. So for now let's comment it out and put `bpmnProcessId` as process definition name.
* same case with `ListViewStore` and `FlowNodeStore`. We use them to build process instance tree path that we will rework in #18378. Let's again remove the corresponding methods and put null as `treePath` for now.
* dependency on `MetricsStore`. We use it to record root process instance creation event. These are business related metrics which we store in Elasticsearch as `MetricEntity`. We decided that since we have an Operate entity for this we should create a separate handler for this.
* dependency on `ImportBatch`. This is used to record Prometheus metric [counting finished process instances](https://github.com/camunda/camunda/blob/ff8dbe135d1523283ba1324ed42c98824150432d/operate/importer/src/main/java/io/camunda/operate/zeebeimport/EventsProcessedMetricsCounterImportListener.java#L36). Here we should research whether it's OK from performance point of view to record finished events one by one? Or we really need to accumulate them and flush only from time to time? In this case we can create a separate issue and address this later.

Every time when we delete/comment out some code, let's put `TODO` with the link to GitHub issue.

zeebe/exporters/operate-exporter/src/main/java/io/camunda/zeebe/operate/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
```diff
@@ -0,0 +1,348 @@
...
+  public ProcessInstanceForListViewEntity createNewEntity(String id) {
+    return new ProcessInstanceForListViewEntity().setId(id);
+  }
+
+  @Override
+  public void updateEntity(
+      Record<ProcessInstanceRecordValue> record, ProcessInstanceForListViewEntity entity) {
+
+    final var recordValue = record.getValue();
+    if (isProcessEvent(recordValue)) {
```
sdorokhova, line 117: This check can be moved to `handlesRecord` method.

mihail-ca:
Based on the comments above, I made the following changes to `ListViewProcessInstanceFromProcessInstanceHandler`:

- I removed from constructor of `ListViewProcessInstanceFromProcessInstanceHandler` references to the additional parameters: `ImportBatch ListViewStore FlowNodeStore OperationsManager ProcessCache`
- To keep the original logic from the importer, I added dummy methods at the end (marked with TODO), that define the needed references as local null variables, and return a default value, or take no action (since the references are null). I think this makes it easier to keep track of the original logic in the importer, and later we can see how to implement or replace these dummy methods, marked with TODO
- I also created a separate handler for the Operate metrics, similar to other handlers, so I removed the code related to metrics from `ListViewProcessInstanceFromProcessInstanceHandler`

FYI @sdorokhova @ThorbenLindhauer

zeebe/exporters/operate-exporter/pom.xml
```diff
@@ -24,6 +24,11 @@
 
   <dependencies>
 
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
```
sdorokhova, line 29: We have this dependency for `StringUtils` class, right? Can we may be extend our own utility classes? Would be nice not to have Spring dependency on this module.

zeebe/exporters/operate-exporter/pom.xml
```diff
@@ -84,6 +89,11 @@
       <artifactId>operate-common</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>operate-importer-common</artifactId>
```
sdorokhova, line 94: Similar question about this dependency. Why do we need it? I think it's OK to keep it for now, but would be nice to create an issue to clean this up later.

sdorokhova (approved):
I left couple of minor comments on pom, but other than that changes look good!

I will approve the PR, so that you don't need to wait for another review, but please check my comments.

zeebe/exporters/operate-exporter/pom.xml
mihail-ca: @sdorokhova I do not use this dependency, but I added it because the CI fails without it, I do not understand why:
image

mihail-ca: I will merge like this, to have the CI pass, as this dependency can be removed later.

PR #19987, opened 2024-07-03
Simplify REST data model

### LINKED ISSUE #19986: Simplify user task update data model
**Description**

The user task update REST endpoint uses a data model `UserTaskUpdateRequest` that nests another data model `Changeset` at attribute `changeset`.
The nested data model reference is consumed as an `allOf`, indicating that we extend
the original reference. However, we don't actually extend it except for adding more
description on top and making it nullable.

This can be straightened out by moving the description and nullable attributes to the
referenced data model directly and using it simply as an unaltered reference.

tmetzke:
## Description

The user task update REST endpoint uses a data model that nests another data model. The nested data model reference is consumed as an `allOf`, indicating that we extend the original reference. However, we don't actually extend it execpt for adding more description on top and making it nullable.

This can be straightened out to moving the description and nullable attributes to the referenced data model directly and using it simply as an unaltered reference.

## Related issues

closes #19986

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/UpdateUserTaskCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/UpdateUserTaskCommandImpl.java
@@ -22,8 +22,8 @@ import io.camunda.zeebe.client.api.command.UpdateUserTaskCommandStep1;
 import io.camunda.zeebe.client.api.response.UpdateUserTaskResponse;
 import io.camunda.zeebe.client.impl.http.HttpClient;
 import io.camunda.zeebe.client.impl.http.HttpZeebeFuture;
+import io.camunda.zeebe.client.protocol.rest.Changeset;
 import io.camunda.zeebe.client.protocol.rest.UserTaskUpdateRequest;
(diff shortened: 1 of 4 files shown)


npepinpe (approved):
:confused: This was painful to review not because of you, but for some reason the diff showed up as all the file was deleted and replaced, so I couldn't look at the specific changes :thinking:

tmetzke:
> 😕 This was painful to review not because of you, but for some reason the diff showed up as all the file was deleted and replaced, so I couldn't look at the specific changes 🤔

Let me check, maybe it updated the line endings on Windows and that is why.

tmetzke:
OK, that was actually it, yes, corrected it 👍

tmetzke:
Meh, the generated client DTO for the changeset is a plain Java map now. We have to adjust the way we set attributes. Not a big deal though.

tmetzke:
Rebased onto `main`, updated the new client user task command tests to consider a Java map, and successfully tested locally with `openapi-generator-maven-plugin` version `7.7.0`.

tmetzke:
@npepinpe, just a quick checkup since I changed more things in the meantime

npepinpe (approved):
:+1:

PR #19990, opened 2024-07-03
ci: stabilize `RollingUpdateTest`

### LINKED ISSUE #19910: RollingUpdateTest tests are unstable
**Summary**

We saw this test failing/timing out a lot in the latest builds. See: 

- https://github.com/camunda/camunda/actions/runs/9743991953/job/26888670280
- https://github.com/camunda/camunda/actions/runs/9741851689/job/26888388130

At the moment every merge is blocked by this test class.

**Failures**

<pre>
Paste the failure here
</pre>
</details>

**Hypotheses**

**Logs**

<details><summary>Logs</summary>
<pre>
LOGS
</pre>
</details>

lenaschoenburg:
## Description

1. Regular CI only checks first and last patch release to current version
2. Version compatibility workflow runs every day
3. Version compatibility workflow sends a message to slack when tests fail
4. Version compatibility workflow also checks all released patches of the previous minor against the current version
5. Re-enables the rolling update tests

## Related issues

closes #19910

### DIFF
--- a/.github/workflows/zeebe-version-compatibility.yml
+++ b/.github/workflows/zeebe-version-compatibility.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Test snapshot
         run: >
           ./mvnw -B --no-snapshot-updates -pl zeebe/qa/update-tests
-          -D it.test=io.camunda.zeebe.test.SnapshotTest
+          -D it.test=io.camunda.zeebe.test.RollingUpdateTest
           -D failsafe.rerunFailingTestsCount=3
           -D skipChecks
           verify
--- a/.github/workflows/zeebe-version-compatibility.yml
+++ b/.github/workflows/zeebe-version-compatibility.yml
@@ -113,3 +113,36 @@ jobs:
         with:
           name: zeebe-version-compatibility
           path: ${{ github.workspace }}/zeebe-version-compatibility
+  notify:
+    name: Notify on failure
+    runs-on: ubuntu-latest
+    needs: [released-versions, current-snapshot]
+    if: failure()
+    steps:
+      - uses: actions/checkout@v4
+      - id: slack-notify
+        uses: slackapi/slack-github-action@v1.26.0
+        with:
+          # For posting a rich message using Block Kit
+          payload: |
+            {
+              "blocks": [
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": ":warning: *Zeebe Version Compatibility checks failed:*"
+                  }
+                },
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Please check the related workflow execution: https://github.com/camunda/camunda/actions/runs/${{ github.run_id }}"
+                  }
+                }
+              ]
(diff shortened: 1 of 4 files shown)


lenaschoenburg:
Latest run here: https://github.com/camunda/camunda/actions/runs/9777054662.

npepinpe (commented):
:question: I feel dumb, but won't the daily test time out too as the number of patches increases? :sweat_smile:

lenaschoenburg:
> I feel dumb, but won't the daily test time out too as the number of patches increases?

Since we don't specify a timeout, the default of 6 hours is used. That should be plenty for now. If this becomes a problem, we can start splitting the test matrix over multiple jobs.

.github/workflows/zeebe-version-compatibility.yml
```diff
@@ -65,10 +65,38 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-zeebe
+        with:
+          secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+      - uses: ./.github/actions/build-zeebe
+      - uses: ./.github/actions/build-platform-docker
+        with:
+          version: current-test
```
npepinpe, line 82: :thought_balloon: Not really necessary. Normally you would build it however you want, and then set `ZEEBE_TEST_DOCKER_IMAGE` to the image name. It just so happens that `ZEEBE_TEST_DOCKER_IMAGE` defaults to `camunda/zeebe:current-test`. It might be better in fact to build it however and set that instead of relying on the default in the code :shrug:

lenaschoenburg: Maybe the `build-platform-docker` action could just set `ZEEBE_TEST_DOCKER_IMAGE` (or a more generic env var)?

npepinpe:
> > I feel dumb, but won't the daily test time out too as the number of patches increases?
> 
> Since we don't specify a timeout, the default of 6 hours is used. That should be plenty for now. If this becomes a problem, we can start splitting the test matrix over multiple jobs.

Where is that defined? Sorry about that :see_no_evil:

npepinpe (approved):
Considering it took ~15 minutes for 4 versions, then 6 hours should cover up to ~96 versions :smile:

lenaschoenburg:
> Where is that defined?

Here https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

.github/workflows/zeebe-version-compatibility.yml
npepinpe: That makes sense :+1:

npepinpe:
Should we backport this to 8.5? I'm having problems merging stuff there for the same reason.

npepinpe:
:see_no_evil:

lenaschoenburg:
Ah no, it was just because we never disabled `RollingUpdateTest` on 8.5

PR #19999, opened 2024-07-03
ci: automatically set zeebe test image name

lenaschoenburg:
## Description

Don't hard-code `current-test` everywhere, instead we just set the `ZEEBE_TEST_DOCKER_IMAGE` env var to whatever was built by the `build-platform-docker` action.

## Related issues
follow-up of https://github.com/camunda/camunda/pull/19990
blocked by https://github.com/camunda/camunda/issues/20001

### DIFF
--- a/.github/actions/build-platform-docker/action.yml
+++ b/.github/actions/build-platform-docker/action.yml
@@ -93,6 +93,10 @@ runs:
       shell: bash
       run: echo "result=$(echo '${{ steps.get-image.outputs.tags }}' | head -n 1)" >> $GITHUB_OUTPUT
 
+    - name: Set Zeebe test image
+      shell: bash
+      run: echo "ZEEBE_TEST_DOCKER_IMAGE=${{ steps.get-image-output.outputs.result }}" >> $GITHUB_ENV
+
     - name: Set DISTBALL path relative to the build context
       id: get-distball
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,8 +220,6 @@ jobs:
             maven-modules: "zeebe/qa/update-tests"
             maven-build-threads: 1
             maven-test-fork-count: 10
-    env:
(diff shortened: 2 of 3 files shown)


cmur2:
@lenaschoenburg seems like your fix + mine made it work?

lenaschoenburg:
I don't think so:
> Can't get Docker image: RemoteDockerImage(imageName=sha256:e71acc7cb7c23938ac092bfb49e2152a4f6aa7c1237ac4b1f63e5b20478fb3e3, imagePullPolicy=DefaultPullPolicy(), imageNameSubstitutor=org.testcontainers.utility.ImageNameSubstitutor$LogWrappedImageNameSubstitutor@2694236b)

so the imageId appears to be just the sha which is not very descriptive and apparently not usable, at least not through testcontainers.

cmur2:
It seems to work with https://github.com/camunda/camunda/pull/19999/commits/e8bb6e88b98d78baf82a3084bc92a61b190e646a but as @houssain-barouni mentioned there can be multiple tags (one per line I think) so later actions code needs to be able to deal with that eventuality.

npepinpe:
> It seems to work with [e8bb6e8](https://github.com/camunda/camunda/commit/e8bb6e88b98d78baf82a3084bc92a61b190e646a) but as @houssain-barouni mentioned there can be multiple tags (one per line I think) so later actions code needs to be able to deal with that eventuality.

Can't you pull images as `camunda/zeebe@sha256:<DIGEST>`? So we could set the image name as that with the ID? I tested it quickly with Testcontainers and it works fine.

cmur2:
> Can't you pull images as `camunda/zeebe@sha256:<DIGEST>`?

@npepinpe maybe, I didn't follow up on this approach since my first tries were unsuccessful (maybe another problem/my error) and instead went for https://github.com/camunda/camunda/pull/20087 which provides 1 tag (the first) out of several if needed.

lenaschoenburg:
Closing because I don't see that much value and I don't have time/energy to invest into this.

Issue #19591, opened 2024-06-20
Update NGINX's `proxy_buffer_size` configuration for zeebe on SaaS to match SM


mustafadagher:
In https://github.com/camunda/camunda/issues/11284, We have been noticing many ingress response header limit error thrown due to big error details being propagated from the Zeebe gRPC endpoints to the web-modeler.

While looking into this, we figured it might have been due to NGINX is configured to handle a smaller size for the header than the size of the error details from zeebe.

Looking further into the configurations we have for NGINX, we found out that while in our SM cluster we have `proxy_buffer_size` configured to be `128k` ([here](https://github.com/camunda/camunda-platform-helm/blob/fc79bdbc70475e19a22d9f7b24e11c036cea6be8/charts/camunda-platform-latest/values.yaml#L1120)), that's not the case for our Zeebe SaaS deployments ([here](https://github.com/camunda-cloud/camunda-operator/blob/main/templates/zeebe_ingress.yaml#L6)) nor for web-modeler ([here](https://github.com/camunda/web-modeler/blob/809855747b89e28865bcd986855eb80ddac9c5a8/.ci/deploy/base/modeler/webapp/ingress.yml), and [here](https://github.com/camunda-cloud/ingress-nginx-deployment/blob/9ce98e2e2f9a86e9243c6e92385cec4763c3af55/ingress-nginx/helm-values.yaml#L136))

Now it's important to notice that the default config for NGINX (if not altered) is to have `4` buffers of ` 4kb` ([ref](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#proxy-buffers-number)).

Within this issue, we aim to adjust the configuration for both SaaS Zeebe and Web-modeler to match `Self-Managed` configs.
```[tasklist]
### Tasks
- [ ] https://github.com/camunda-cloud/camunda-operator/pull/3011
```

Issue #11284, opened 2022-12-16
Considerable amount of GRPC error details can trigger ingress response header limit


megglos:
**Description**

In cases, like an invalid bpmn model, a considerable number of errors can occur. These are passed back to the client via the `grpc-status-details-bin` HTTP header, [see](https://groups.google.com/g/grpc-io/c/ZA2i0lhCBO4).
In cases where Zeebe is accessed through a reverse proxy/ingress like nginx, this behavior can exceed response header limits of the proxy, resulting in error logs as:
```
[error] 2243#2243: *1868588 upstream sent too big header while reading response header from upstream
```
leading to a 502 response to the client by the proxy.

We may consider limiting the amount of error details returned to e.g. a limit of 5 entries, followed by a summary error entry if there are more errors "and N more errors". This would allow to still enforce a response header limit on any client facing proxy while making sure the client still gets back useful information to actually fix the model errors, no matter how much errors they provoked.

[SUPPORT-15410](https://jira.camunda.com/browse/SUPPORT-15410)

```[tasklist]
### Task List
- [ ] https://github.com/camunda/camunda-bpm-platform/issues/4443
- [ ] https://github.com/camunda/camunda/issues/19595
- [ ] https://github.com/camunda/camunda/issues/19591
- [ ] https://github.com/camunda/camunda/issues/19829
```

Zelldon:
See related discussion and investigation in slack https://camunda.slack.com/archives/CT702EPFH/p1671124412586789

korthout:
@megglos Can you clarify for us what the chances are of running into this situation? And what is the impact? Does the user receive a response at all?

megglos:
@jfriedenstab 300 out of how many total deployment requests? 🤓 

Would you consider a truncated error like suggested here
> We may consider limiting the amount of error details returned to e.g. a limit of 5 entries, followed by a summary error entry if there are more errors "and N more errors"
would be a good solution to educate the user about the modeling errors in the model?

The user may at least be aware of some specific errors and once resolved may learn about the next remaining ones. On the other hand wouldn't it be better to raise those issues directly in the modeler via validation to provide more helpful errors direct in the modeler (e.g on the affected elements)? If I'm no mistaken we have such validation already and I wonder how users can deploy invalid models from the modeler still 🤔

jfriedenstab:
Thanks a lot for your quick reply, @megglos!

> 300 out of how many total deployment requests? 🤓

[...]

PR #20002, opened 2024-07-03
feat: allow overriding default `maxInboundMetadataSize` through config and change default to 16KB 

### LINKED ISSUE #19829: Extend Zeebe's java client to allow configuring `maxInboundMetadataSize `
During the testing of #11284 and after applying the camunda-operator changes from https://github.com/camunda-cloud/camunda-operator/pull/3025, while the use case we were testing had finally worked initially. 

We found out that when we increased the number of errors in the model even more, we started to see a different error `http2 exception` thrown from `zeebe-client` with an underlying error `Header size exceeded max allowed size (8192)`.

With more investigation, it turns out that `zeebe-client` doesn't set a value for `maxInboundMetadataSize` when building the channel ([here](https://github.com/camunda/camunda/blob/a5a1003acfbe210482b38eea445d880f3f0023f6/clients/java/src/main/java/io/camunda/zeebe/client/impl/ZeebeClientImpl.java#L185)).
The default value for this config is `8kb`

I'm looking making this value configurable too for zeebe's java client

mustafadagher:
## Description

Allows overriding default `maxInboundMetadataSize` through config and change default to 16KB 
This is a followup to changing default NGINX proxy buffer size to 16kb to allow returning more BPMN validation errors

## Related issues

closes #19829 
related to #11284 
related to #19591

### DIFF
--- a/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/PropertyUtil.java
+++ b/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/PropertyUtil.java
@@ -49,7 +49,6 @@ public class PropertyUtil {
       return (T) propertyValue;
     }
     T property = getPropertyFromSupplier(legacyPropertySupplier, propertyName, "legacy");
-
     if (property == null || property.equals(defaultProperty)) {
       property = getPropertyFromSupplier(propertySupplier, propertyName, "property");
     }
@@ -63,6 +62,29 @@ public class PropertyUtil {
 
     LOG.debug("Property {} set to {}", propertyName, property);
     return property;
+  }
+
+  /**
+   * Returns the property in the given relevance: property, defaultProperty
+   *
+   * @param propertyName the name of the property, used for logging
+   * @param propertySupplier a function to supply the property, may throw
+   * @param defaultProperty the default to apply if nothing else suits, may be null
+   * @param configCache the cache to save the property to, may be null
+   * @return the property resolved
+   * @param <T> the type of the property
+   */
+  public static <T> T getOrDefault(
+      final String propertyName,
+      final Supplier<T> propertySupplier,
+      final T defaultProperty,
+      final Map<String, Object> configCache) {
+    return getOrLegacyOrDefault(
+        propertyName, propertySupplier, noPropertySupplier(), defaultProperty, configCache);
+  }
+
+  private static <T> Supplier<T> noPropertySupplier() {
+    return () -> null;
   }
 
   private static <T> T getPropertyFromSupplier(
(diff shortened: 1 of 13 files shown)


spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/PropertyUtil.java
```diff
@@ -65,6 +67,24 @@ public static <T> T getOrLegacyOrDefault(
...
+   * @param configCache the cache to save the property to, may be null
+   * @return the property resolved
+   * @param <T> the type of the property
+   */
+  public static <T> T getOrOrDefault(
+      final String propertyName,
+      final Supplier<T> propertySupplier,
+      final T defaultProperty,
+      final Map<String, Object> configCache) {
+    return getOrLegacyOrDefault(propertyName, propertySupplier, null, defaultProperty, configCache);
```
korthout, line 85: 🔧 Rather than providing `null` and requiring the additional `null` check in the `getOrLegacyOrDefault` method, you could pass a null supplier `() -> null`. You could then also move that to a constant and naming it something like `NO_LEGACY_PROPERTY`:

```java
return getOrLegacyOrDefault(propertyName, propertySupplier, NO_LEGACY_PROPERTY, defaultProperty, configCache);
```

spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/PropertyUtil.java
```diff
@@ -65,6 +67,24 @@ public static <T> T getOrLegacyOrDefault(
...
+   * Returns the property in the given relevance: property, defaultProperty
+   *
+   * @param propertyName the name of the property, used for logging
+   * @param propertySupplier a function to supply the property, may throw
+   * @param defaultProperty the default to apply if nothing else suits, may be null
+   * @param configCache the cache to save the property to, may be null
+   * @return the property resolved
+   * @param <T> the type of the property
+   */
+  public static <T> T getOrOrDefault(
```
korthout, line 80: 🔧 Typo: `OrOr`

korthout (approved):
Nice work @mustafadagher 🚀 

💭 It would be good to explain in the commit message (and pull body) why we choose 16KB as the new default.

🔧 We should backport this to allow all users to avoid the bug:
- #11284 

🔧 I have a few minor suggestions

LGTM 👍

spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/PropertyUtil.java
mustafadagher: I like it! 😍

mustafadagher: Static constant won't work though because of the generic T, but will create a static method for that.

mustafadagher:
@korthout there's a problem with backporting this, the `spring-boot-starter-camunda-sdk` has severely changed

PR #20009, opened 2024-07-03
Add Search request and sort transformer tests [OS]

ChrisKujawa:
## Description

 Extend test coverage on the opensearch search client transformers. Covering now sort options, and search request transformers. 

Fixed some issues:

 * support `missing` flag in SortOptions
 * handle some null value, and/or add description

Tests are similar to before based on using the Search* Builder/Objects and then calling the transformation and verifying the resulting JSON.

## Related issues


search/search-client-opensearch/src/test/java/io/camunda/search/os/util/OSQuerySerializer.java
```diff
@@ -40,6 +41,12 @@ public String serialize(final Query query) {
     return out.toString();
   }
 
+  public String serialize(final JsonpSerializable serializable) {
```
Zelldon, line 44: 💭 I think we can remove the other methods then.

Zelldon:
Test coverage for OS search client and transformers:





The only thing we now miss is to test whether responses are correctly mapped back, you can see on the coverage of the search query hit and response. I will create an issue for this, as I will not have time for such.

romansmirnov (approved):
🚀

PR #20034, opened 2024-07-04
Add additional search client tests [ES]

ChrisKujawa:
## Description

Extend test coverage on the elasticsearch search client transformers. Covering now sort options, and search request transformers.

Fixed some issues:

- support missing flag in SortOptions
- Tests are similar to before based on using the Search* Builder/Objects and then calling the transformation and verifying the resulting JSON.

Similar to OS client test coverage no close to 100 for request, for responses we still need some tests

## Related issues

related to https://github.com/camunda/camunda/pull/20009


romansmirnov (approved):
🚀

PR #20037, opened 2024-07-04
Changing default streaming timeout

### LINKED ISSUE #19188: Set default job worker stream timeout
**Description**

Currently, job workers with streaming enabled open a long living stream with an indefinite duration. This can cause some issues to crop up over time:

- Connections may not be well balanced against all servers, as they are pinned to a specific one
- Some reverse proxies/ingresses may forcefully close idle connections early with a 504, which is not possible to distinguish between an actual 504 due to an upstream timeout, i.e. an actual error. In most cases, it's also not possible to set specific settings per call, but rather it's applied to the complete virtual host.

As a solution, we propose to set a default timeout of 1 hour. This would cause the connection to be recreated every hour regardless, and would likely cause them to be better load balanced. Combine with sane defaults for your ingresses (e.g. double that), and we can greatly mitigate the second point as well.

The setting would still be configurable as well, and it should have little impact on existing deployments.

**Context**
- https://jira.camunda.com/browse/SUPPORT-22069
- https://github.com/camunda-cloud/ingress-nginx-deployment/issues/287
- https://github.com/camunda/camunda-docs/issues/3903

ana-vinogradova-camunda:
Setting default streaming timeout to 1h 

## Description

Setting default streaming timeout to 1h to improve connections load balancing and solve [this issue](https://github.com/camunda/camunda/issues/19188).

## Related issues

closes #19188

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/worker/JobWorkerBuilderImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/worker/JobWorkerBuilderImpl.java
@@ -42,7 +42,7 @@ public final class JobWorkerBuilderImpl
 
   public static final BackoffSupplier DEFAULT_BACKOFF_SUPPLIER =
       BackoffSupplier.newBackoffBuilder().build();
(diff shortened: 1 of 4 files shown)


mustafadagher (approved):
LGTM 🚀

Issue #19984, opened 2024-07-03
Snapshot detected as corrupted because of incorrect checksum comparison


deepthidevaki:
**Describe the bug**
On restart of broker
We see 
```
Expected snapshot /usr/local/zeebe/data/raft-partition/partitions/3/snapshots/90926-1-821751-758053 to have checksum 1612185047, but the actual checksum is 3972871514; the snapshot is most likely corrupted. The startup will fail if there is no other valid snapshot and the log has been compacted.
```
then the Partition cannot startup
```
java.lang.IllegalStateException: In partition 2 expected to find a snapshot at index >= log's first index 211099, but found snapshot 0. A previous snapshot is most likely corrupted.
	at io.atomix.raft.utils.StateUtil.verifySnapshotLogConsistent(StateUtil.java:33) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.impl.RaftContext.<init>(RaftContext.java:220) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.impl.DefaultRaftServer$Builder.build(DefaultRaftServer.java:249) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.impl.DefaultRaftServer$Builder.build(DefaultRaftServer.java:223) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.partition.impl.RaftPartitionServer.buildServer(RaftPartitionServer.java:179) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.partition.impl.RaftPartitionServer.<init>(RaftPartitionServer.java:98) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.partition.RaftPartition.createServer(RaftPartition.java:133) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.partition.RaftPartition.initServer(RaftPartition.java:116) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
	at io.atomix.raft.partition.RaftPartition.bootstrap(RaftPartition.java:96) ~[zeebe-atomix-cluster-8.5.4.jar:8.5.4]
[...]

PR #20040, opened 2024-07-04
test: snapshot false negative corruption unit test

EuroLew:
## Description

When a snapshot which has been received from a leader is persisted, eventually the snapshot will have to be read say during a restore or actor start. When the snapshot is read a corruption check must occur, it therefore generates the individual file checksums and a combined checksum from the files on disk. Due to the nature of how the combined checksum is generated by reading the full file and RocksDB full file checksums does not read the file, equal combined checksums are not achieved from the same files.

The test ensures that even if combined checksums are different no errors occur (as long as the individual file checksums are equal).

The aim is to eventually remove combined checksum as it is only used in test code currently.

## Related issues

Relates to #19984


deepthidevaki (approved):
Great :tada: 

:x: Please address my comments before merging. 
Don't forget to backport to all stable branches.

deepthidevaki:
@EuroLew Please merge this PR.

deepthidevaki:
@EuroLew There are build failures due to concurrent PRs. Please rebase the branch on main and fix the conflicts.

PR #20047, opened 2024-07-04
chore: skip front-end build

OmranAbazid:
## Description
Added a profile to the client project pom to allow zeebe CI jobs to skip the front-end build like other components.
I also removed the `skip.yarn.build` since i did not find it being used anywhere. 

## Related issues

closes #


buccarel (approved):
Nice

PR #20055, opened 2024-07-04
fix: use treePathCache per partition instead one for all

ralfpuchert:
## Description

* Use one `treePathCache` for each partition 
* Initialize early the treePathCache with help from `PartitionHolder`

## Related issues

Related to #20076


ralfpuchert:
Correct one: https://github.com/camunda/camunda/pull/20077

PR #20056, opened 2024-07-04
deps: bump zeebe-test-container for compatibility with httpcore5

npepinpe:
## Description

This PR bumps zeebe-test-container to fix an issue introduced in a shared transitive dependency, httpcore5 5.2.5. This breaks the `DebugExporter` as the server receiving exported records now (since 5.2.5) checks the hostname authority, even when TLS is not enabled. This breaks several container based tests such as the disk filling ones.


npepinpe:
Renovate was taking too long to update :smile:

npepinpe:
nvm, saw the PR was created already

PR #20065, opened 2024-07-04
Threadsafe oauth credentials cache

### LINKED ISSUE #13832: Oauth token refresh can trigger rate limiting and log errors
**Describe the bug**

I have observed that my Zeebe client app will periodically emit error logs with `Failed while fetching credentials:  exception=java.io.IOException: Failed while requesting access token with status code 429 and message Too Many Requests.`

I don't see any other adverse effects besides the error logs, and the application seems to be able to continue to handle jobs afterwards.

**To Reproduce**

This is difficult to reproduce specifically, and the stack trace contains only Camunda code. It seems like this is coming from `JobPoller` and occurs when it is time to refresh the Zeebe client access token.

Perhaps something about the way that we register job handlers could be contributing to this. If we are hitting rate limits, then maybe multiple pollers are trying to refresh the access token at the same time.
```
    // WorkflowWorker and WorkerHandler are classes that our code uses to manage our various job handlers

    void register(WorkflowWorker<?>... workers) {
        Stream.of(workers).forEach(this::registerWorker);
    }

    private void registerWorker(WorkflowWorker<?> worker) {
        String jobType = worker.getJobType();
        JobHandler jobHandler = new WorkerHandler<>(worker, clientId);
        zeebeClient.newWorker()
                .jobType(jobType)
                .handler(jobHandler)
                .open();
    }
```

**Expected behavior**

I see two possible expected behaviors depending on what exactly is occurring.

1. If it is the case that these rate limit errors are causing job polling to fail, then I would expect that the token refresh implements an exponential backoff to navigate the rate limiting.
    - I saw a similar issue and fix in the community managed NodeJS client: https://github.com/camunda-community-hub/zeebe-client-node-js/issues/301
    - There is also a note on that issue that refers to the Java client implementing backoff in JobPoller, so it is possible that this error is not causing the poll to fail.

2. If it is the case that it is not causing job polling to fail, then I would expect that OauthCredentialsProvider no longer logs errors when encountering a 429.

**Log/Stacktrace**

 <p>

```
Failed while fetching credentials: 
java.io.IOException: Failed while requesting access token with status code 429 and message Too Many Requests.
	at io.camunda.zeebe.client.impl.oauth.OAuthCredentialsProvider.fetchCredentials(OAuthCredentialsProvider.java:183)
	at io.camunda.zeebe.client.impl.oauth.OAuthCredentialsProvider.refreshCredentials(OAuthCredentialsProvider.java:132)
	at io.camunda.zeebe.client.impl.oauth.OAuthCredentialsProvider.shouldRetryRequest(OAuthCredentialsProvider.java:100)
	at io.camunda.zeebe.client.impl.worker.JobPoller.onError(JobPoller.java:116)
	at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:487)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:563)
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:70)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:744)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:723)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

</p>
</details>

**Environment:**
- OS: Linux
- Zeebe Version: 8.2.11
- Configuration: _unsure_

Related:
- https://jira.camunda.com/browse/SUPPORT-18518

korthout:
## Description

In an attempt to resolve #13832, this makes the `OAuthCredentialsCache` threadsafe. It does so by locklessly replacing the concurrently accessed hashmap.

## Related issues

relates to #13832


zeebe/clients/java/src/main/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCache.java
```diff
@@ -45,38 +47,40 @@ public final class OAuthCredentialsCache {
...
-  public synchronized OAuthCredentialsCache readCache() throws IOException {
+  public OAuthCredentialsCache readCache() throws IOException {
     if (!cacheFile.exists() || cacheFile.length() == 0) {
       return this;
     }
 
     final Map<String, OAuthCachedCredentials> cache = MAPPER.readValue(cacheFile, TYPE_REFERENCE);
-    audiences.clear();
-    audiences.putAll(cache);
+    AtomicUtil.replace(audiences, current -> cache);
```
lenaschoenburg, line 64: :wrench: This could be just `audiences.set(cache)`

zeebe/clients/java/src/main/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCache.java
```diff
@@ -114,14 +118,20 @@ public synchronized <T> Optional<T> withCache(
     }
   }
 
-  public synchronized OAuthCredentialsCache put(
+  public OAuthCredentialsCache put(
       final String endpoint, final ZeebeClientCredentials credentials) {
-    audiences.put(endpoint, new OAuthCachedCredentials(credentials));
+    AtomicUtil.replace(
```
lenaschoenburg, line 123: :wrench: This can just be a normal `audiences.getAndUpdate(...)`

zeebe/clients/java/src/main/java/io/camunda/zeebe/client/impl/util/AtomicUtil.java
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * <p>If the value of the atomic reference has been replaced by another thread in the meantime,
+   * the replacer function is called again.
+   *
+   * @param ref The atomic reference that holds the value to replace
+   * @param replacer The replacer function used to provide a new value to the atomic reference
+   * @return The previous value of the atomic reference, or null if the value was not replaced
+   * @param <T> The type of the value of the atomic reference
+   */
+  public static <T> T replace(final AtomicReference<T> ref, final Function<T, T> replacer) {
```
lenaschoenburg, line 37: :thought_balloon: Unless I'm missing something, this is the same as `AtomicReference#getAndUpdate`.

korthout:
Thanks @lenaschoenburg 🙇 TIL

zeebe/clients/java/src/main/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCache.java
```diff
@@ -45,38 +46,40 @@ public final class OAuthCredentialsCache {
       new TypeReference<Map<String, OAuthCachedCredentials>>() {};
   private static final ObjectMapper MAPPER = new ObjectMapper(new YAMLFactory());
 
-  private final Map<String, OAuthCachedCredentials> audiences;
   private final File cacheFile;
+  private final AtomicReference<Map<String, OAuthCachedCredentials>> audiences;
```
mustafadagher, line 50: ❓ I'm curious why did we chose wrapping the map in an atomic reference, rather than for example using a `ConcurrentHashMap`?

`ConcurrentHashMap` supports atomic single operations on an `entry` level, rather than an `entire map` level with `AtomicReference `. It's also optimized for frequent read/write operations. Uses less memory (no need to copy the whole map with each update).

I can see that `AtomicReference<HashMap<>>` would be more suitable for cases with rare updates. 

But I'm curious to hear what was your reasoning  🙏

mustafadagher (approved):
Good job Nico! Looks good to me 🚀 
I've only one question below, curious to hear your thoughts about it :)

zeebe/clients/java/src/main/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCache.java
remcowesterhoud: @mustafadagher I understand you question and can't answer it for you. But since @korthout is on FTO this afternoon and the retreat is next week, do you mind if I go ahead and merge this without the answer? This way we can unblock the customer by giving him a snapshot version to verify against.

korthout: I initially looked into using a concurrent hashmap but noticed specific access patterns like the clear and put all that's now become a reference set. AFAIK a clear and put all cannot be done atomically with a concurrent hashmap

mustafadagher: @remcowesterhoud, I've pre-approved the PR. you can go ahead and merged :)

@korthout I see! Thanks for answering! Indeed there's not an out-of-the-box atomic `clear` and `putAll` in `ConcurrentHashMap` without locking or synchronizing.

mustafadagher: Just a tiny hint. this change doesn't ensure an atomic single (entry level) put or get on the wrapped map AFAICT. To so, we might need to wrap a `ConcurrentHashMap` within an `AtomicReference`

PR #20086, opened 2024-07-05
fix: increase Spring Boot starter pluging timeout

sdorokhova:
Increase timeout from 30 sec (default) to 40 seconds to improve E2E tests stability.


pedesen (approved):
Change looks good to me. Let's run it a few times to see if 40s are enough.

sdorokhova:
In this run Operate needed 31 seconds to start and there was no timeout error: https://github.com/camunda/camunda/actions/runs/9806161852/job/27080427391?pr=20086

```
11:02:58.507 [] [main] [] INFO  io.camunda.application.StandaloneOperate - Started StandaloneOperate in 31.434 seconds (process running for 33.419)
```

PR #19521, opened 2024-06-19
ci: use common build-platform-docker GHA

houssain-barouni:
## Description

Replace the usage of `build-{component}-docker` GHAs by `build-platform-docker`
`build-platform-docker` is already used to create the docker images for all components in release workflow
Make `build-platform-docker` support many versions in case multiple tags need to be created for the same image.
## Related issues

closes #

PR #20087, opened 2024-07-05
ci: fix build-platform-docker action to provide image output again

### LINKED ISSUE #20001: Shared `build-platform-docker` action output `image` is missing
The action [build-platform-docker](https://github.com/camunda/camunda/blob/4cc32dc75502b49b85122e401e6329481aa0fb1d/.github/actions/build-platform-docker/action.yml) say's that there should be a `image` output that contains "Fully qualified image name available in your local Docker daemon".

This does not work, the `image` output is always empty. See this example run: https://github.com/camunda/camunda/actions/runs/9777854271/job/26993894405?pr=19999

The `image` output is coming from `steps.get-image.outputs.result`. But the  `docker/metadata-action@v5`  used in that step does not define a `result` output.

This blocks https://github.com/camunda/camunda/pull/19999.

cmur2:
## Description

The `image` output of the `build-platform-docker` action is empty since the change done in https://github.com/camunda/camunda/pull/19521 which extended the action to handle multiple tags.

This PR fixes the `image` output so it provides a valid tag (if multiple are given it picks the first, all tags should be interchangeably usable in local Docker daemon) again.

## Related issues

Fixes #20001

### DIFF
--- a/.github/actions/build-platform-docker/action.yml
+++ b/.github/actions/build-platform-docker/action.yml
@@ -38,7 +38,7 @@ inputs:
 outputs:
   image:
     description: "Fully qualified image name available in your local Docker daemon"
-    value: ${{ steps.get-image.outputs.result }}
+    value: ${{ steps.get-image-output.outputs.result }}
   date:
     description: "The ISO 8601 date at which the image was created"
     value: ${{ steps.get-date.outputs.result }}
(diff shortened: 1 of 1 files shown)


houssain-barouni (approved):
Thanks!

PR #20088, opened 2024-07-05
Limit client request retries to three attempts

### LINKED ISSUE #13832: Oauth token refresh can trigger rate limiting and log errors

korthout:
## Description

There was a bug where client requests could be retried indefinitely if the request was responded to with `UNAUTHORIZED` (HTTP `401` or grpc `UNAUTHENTICATED`).

This endless retry could lead to `429 Too Many Requests`, at which point the retry stopped. However, there's no need to retry `UNAUTHORIZED` requests so often.

This pull request limits the number of retries to two. This means requests are tried a total of three times.

## Related issues

closes #13832

### DIFF
--- a/zeebe/clients/java/src/main/java/io/camunda/zeebe/client/impl/RetriableClientFutureImpl.java
+++ b/zeebe/clients/java/src/main/java/io/camunda/zeebe/client/impl/RetriableClientFutureImpl.java
@@ -19,12 +19,14 @@ import io.camunda.zeebe.client.CredentialsProvider.StatusCode;
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
 public final class RetriableClientFutureImpl<R, T> extends ZeebeClientFutureImpl<R, T> {
 
+  private final AtomicInteger retries = new AtomicInteger(2);
   private final Predicate<StatusCode> retryPredicate;
   private final Consumer<StreamObserver<T>> retryAction;
 
@@ -50,7 +52,7 @@ public final class RetriableClientFutureImpl<R, T> extends ZeebeClientFutureImpl
     final Status status = Status.fromThrowable(throwable);
     final StatusCode statusCode = new GrpcStatusCode(status.getCode());
 
-    if (retryPredicate.test(statusCode)) {
+    if (retries.getAndDecrement() > 0 && retryPredicate.test(statusCode)) {
       retryAction.accept(this);
     } else {
       super.onError(throwable);
--- a/zeebe/clients/java/src/test/java/io/camunda/zeebe/client/CredentialsTest.java
+++ b/zeebe/clients/java/src/test/java/io/camunda/zeebe/client/CredentialsTest.java
@@ -151,7 +151,8 @@ public final class CredentialsTest {
 
               @Override
               public boolean shouldRetryRequest(final StatusCode code) {
-                return retryCounter-- > 0;
+                retryCounter--;
+                return true;
               }
             });
     builder.usePlaintext().credentialsProvider(provider);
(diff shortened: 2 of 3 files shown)


remcowesterhoud (approved):
Nice find @korthout! LGTM 🚀

megglos:
@korthout following up on this one in context of https://github.com/camunda/camunda/issues/26012#issuecomment-2883775938 to better understand our retry approach.

Why did we go with 2 retries? I understand a retry on a 401 makes sense for the edge-case that the token became invalid (e.g. just in time expired after read from the cache), so doing a token refresh may resolve the situation. However from that perspective in such a case doing even another retry doesn't appear to be worth it, as I don't see a scenario where it would ever succeed on a third attempt, given on the second attempt a new token was already used. 

One reason I could see would be that within `RetriableClientFutureImpl` where the retry logic is contained, we don't have the authentication context, thus 2 retries was just a general attempt on some retry limit decoupled from the actual cause?

korthout:
I don't remember.

I could find [this comment](https://github.com/camunda/camunda/issues/13832#issuecomment-2210472885), but it doesn't explain why 2 retries (3 attempts) instead of 1 retry (2 attempts). 

Possibly, it was just a feeling for a reasonable magic number. Or something we've seen applied elsewhere. Feel free to change it if you have reason for it.

Issue #19175, opened 2024-06-07
[CPT] I can assert the process flow


saig0:
## Description

In Camunda Process Test, I want to write a process test and verify that the process took the expected path. Take inspiration from [zeebe-process-test](https://docs.camunda.io/docs/apis-tools/java-client/zeebe-process-test/#assertions). 

## Tasks
- Assert that a process is deployed
- Assert that a process instance is active/completed/terminated
- Assert that an element of a process instance is active/completed/terminated (a given number of times)
- Assert that a user task has a given property (e.g. an assignee)
- Assert that a variable of a process instance exists and has a given value
- Assert that an incident is created/resolved
- Handle the asynchronous nature of the Camunda engine transparently 

requires #18997

korthout:
@saig0 I remember you told me you want to build this using the available APIs. Does this somehow depend on:
- https://github.com/camunda/product-hub/issues/2041?

saig0:
@korthout in general, CPT depends on the API. We could use the GET endpoints in some cases instead of the search API. 

However, for the first iteration, we will use what is currently available. So, we are not blocked by the GET endpoints.

saig0:
We implemented the basic assertions for the process instance state, the element instance state, and variables. 

Let's close the issue and create follow-up issues for the other assertions.

PR #20101, opened 2024-07-05
feat: Add CPT assertions for a process instance

saig0:
## Description

- Add new assertions to:
    - Verify the state of a process instance: active/completed/terminated
    - Verify the state of BPMN elements: active/completed/terminated
- The assertions are inspired by [zeebe-process-test](https://docs.camunda.io/docs/apis-tools/java-client/zeebe-process-test/#process-instance-assertions) with noticeable differences:
    - Use different naming for `hasActiveElements()` vs. `isWaitingAtElements()` and `hasCompletedElements()` vs. `hasPassedElement()` to align the wording and distinguish completed from terminated
    - Use BPMN element names vs. element ids to make it easier/powerful under the assumption that element names are not important and visible in the BPMN. We will add support for ids later.
    - The assertions wait implicitly until the expected state is reached to ease the asynchronous of the engine
- The assertions follow the style of AssertJ 
- The implementation is based on AssertJ and [Awaitility](https://github.com/awaitility/awaitility)
    - Awaitility is used to wait until the process instance reaches the expected state 
- Build a small client to access the (old) APIs. We will replace this client later with the official SDK.   

This PR adds the first assertions to CPT. We will add more assertions later. 

## Related issues

contributes to #19175

### DIFF
--- a/testing/camunda-process-test-java/pom.xml
+++ b/testing/camunda-process-test-java/pom.xml
@@ -57,6 +57,32 @@
       <artifactId>camunda-client-java</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>compile</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.httpcomponents.core5</groupId>
+      <artifactId>httpcore5</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+
     <!--  test scope  -->
 
     <dependency>
@@ -78,12 +104,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.assertj</groupId>
-      <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
@@ -96,12 +116,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.awaitility</groupId>
-      <artifactId>awaitility</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-core</artifactId>
       <scope>test</scope>
@@ -110,18 +124,6 @@
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-slf4j2-impl</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.httpcomponents.client5</groupId>
-      <artifactId>httpclient5</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.httpcomponents.core5</groupId>
-      <artifactId>httpcore5</artifactId>
       <scope>test</scope>
     </dependency>
 
--- a/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
+++ b/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright © 2017 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.process.test.api;
+
+import io.camunda.process.test.api.assertions.ProcessInstanceAssert;
+import io.camunda.process.test.impl.assertions.CamundaDataSource;
+import io.camunda.process.test.impl.assertions.ProcessInstanceAssertj;
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.client.api.response.ProcessInstanceResult;
+import java.time.Duration;
+import org.awaitility.Awaitility;
+
+public class BpmnAssert {
+
+  public static final Duration DEFAULT_ASSERTION_TIMEOUT = Duration.ofSeconds(10);
+  public static final Duration DEFAULT_ASSERTION_INTERVAL = Duration.ofMillis(100);
+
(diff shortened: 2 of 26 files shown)


saig0:
@remcowesterhoud please have a look at this "small" PR. I will be on FTO starting from Wednesday. If possible, it would be great to review it before. Thank you. :bow:

remcowesterhoud:
I'll have a look on monday 🙂

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
```diff
@@ -0,0 +1,86 @@
...
+
+import io.camunda.process.test.api.assertions.ProcessInstanceAssert;
+import io.camunda.process.test.impl.assertions.CamundaDataSource;
+import io.camunda.process.test.impl.assertions.ProcessInstanceAssertj;
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.client.api.response.ProcessInstanceResult;
+import java.time.Duration;
+import org.awaitility.Awaitility;
+
+public class BpmnAssert {
```
remcowesterhoud, line 53: ❓ I know you aim to keep the API aligned with ZPT as much as possible. However, I wonder if it makes sense to rename this class to `CamundaAssert`. It aligns more with the naming of the rest of the project. I'm also thinking about what if we add DMN support in the future, then the name doesn't make sense. We'd have to add a `DmnAssert`.
On the flip-side, maybe adding DMN doesn't make sense anyways considering the project is called Camunda **process** test.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ProcessInstanceAssertj.java
```diff
@@ -59,6 +66,26 @@ public ProcessInstanceAssert isTerminated() {
...
   }
 
+  @Override
+  public ProcessInstanceAssert hasActiveElements(final String... elementNames) {
+    hasElementsInState(elementNames, FlowNodeInstanceState.ACTIVE, Objects::nonNull);
+    return this;
+  }
+
+  @Override
+  public ProcessInstanceAssert hasCompletedElements(final String... elementNames) {
```
remcowesterhoud, line 76: 💭 It'd be great in the future to have assertions that also verify the order in which the elements reached the state.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ProcessInstanceAssertj.java
```diff
@@ -92,19 +119,97 @@ private void hasProcessInstanceInState(
     }
   }
 
+  private void hasElementsInState(
```
remcowesterhoud, line 122: ❌ Some fun edge-case here is when there is multiple elements with the same name in the process, or if there is a loop in the process. When we pass the same element twice, I assume the endpoint returns 2 flownode instances. What will happen if the first flownode is completed and the second is active? My expectation is that the assertion will always fail 😄 

If this case is too complex for this PR just ignore it, but we should be aware of the limitation for now.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
```diff
@@ -0,0 +1,86 @@
...
+
+  public static final Duration DEFAULT_ASSERTION_TIMEOUT = Duration.ofSeconds(10);
+  public static final Duration DEFAULT_ASSERTION_INTERVAL = Duration.ofMillis(100);
+
+  private static final ThreadLocal<CamundaDataSource> DATA_SOURCE = new ThreadLocal<>();
```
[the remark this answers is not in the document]
remcowesterhoud: This may also remove the need for the initializing the data source. However, I can imagine it'd be nice to keep it as this will allow users to configure the datasource with their specific runtime.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Configures the time between two assertion attempts until the expected state is reached.
+   *
+   * @param assertionInterval time between two assertion attempts
+   * @see #DEFAULT_ASSERTION_INTERVAL
+   */
+  public static void setAssertionInterval(final Duration assertionInterval) {
+    Awaitility.setDefaultPollInterval(assertionInterval);
+  }
```
remcowesterhoud, line 88: 🔧 Maybe not fully in the scope of this PR, but it would be nice to make these configurable through the extension. E.g.:

`@CamundaProcessTest(assertionTimeout=10, assertionInterval=100)`

remcowesterhoud (changes requested):
Nice work @saig0, looks great 🚀 I have a few comments, please have a look and let me know what you think 🙂

testing/camunda-process-test-java/pom.xml
```diff
@@ -57,6 +57,32 @@
...
 
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>compile</scope>
```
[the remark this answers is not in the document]
saig0: I thought the same. :sweat_smile: If I remove scope, the maven-dependency-plugin complains that the dependency is defined for `test` scope. I guess there is a transient dependency from another dependency. Setting the scope to `compile` makes the plugin happy. :laughing:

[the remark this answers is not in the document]
remcowesterhoud: Ah, it's because the parent pom defines it with a test scope.

```xml
      <dependency>
        <groupId>org.awaitility</groupId>
        <artifactId>awaitility</artifactId>
        <version>${version.awaitility}</version>
        <scope>test</scope>
      </dependency>
```

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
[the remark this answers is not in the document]
saig0: I'm not sure about this. ZPT and the C7 assert use a ThreadLocale. It might be good to use a ThreadLocal because it is a static class. 

If you would run tests in parallel, we would need a ThreadLocal to interact with the correct runtime.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
saig0: Interesting idea. I'll rename it. :+1:

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ProcessInstanceAssertj.java
saig0: Yes!

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ProcessInstanceAssertj.java
saig0: > What will happen if the first flownode is completed and the second is active? My expectation is that the assertion will always fail

Nope. It would pass the since at lease one flow node is in the expected state. :sweat_smile: 

I would like to ignore this case for now. We could solve this case by introducing another method: `hasCompletedElements("A", 2)`, similar to ZPA.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/BpmnAssert.java
saig0: Interesting idea. I'm not sure about it. Currently, these properties are configured globally for all test cases. So, it doesn't feel correct to change them in one annotation. The other test cases would be accepted too. 

Let's think about it later again. :+1:

remcowesterhoud (approved):
That's fine with me. Thanks!

PR #20142, opened 2024-07-08
feat: optionally throttle writes when exporting is too slow

### LINKED ISSUE #19681: Zeebe can adjust write rate limit based on exporting backlog

lenaschoenburg:
## Description

When `zeebe.broker.flowControl.write.throttling.enabled=true`, the write rate is throttled, based on the observed exporting rate.

When the exporting backlog is approaching the configurable `acceptableBacklog`, the write rate limit is gradually decreased. When exporting rate continues to be slow, the write rate settles to match exporting rate exactly, maintaining a constant exporting backlog of `acceptableBacklog`.

When writes are throttled, a new annotation marks this:
exporting slowed down
exporting backlog stablilizing
backpressure increases

## Related issues

closes #19681


zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/management/FlowControlEndpointIT.java
```diff
@@ -34,105 +30,115 @@ final class FlowControlEndpointIT {
...
+        }""");
+    final var flowControlConfiguration = actuator.getFlowControlConfiguration();
 
     // then
-    assertThat(flowControlConfiguration.get(1).toString())
-        .contains(
-            "\"requestLimiter\":{\"limit\":50,\"estimatedLimit\":50.0,\"rtt_noload\":0,\"maxLimit\":1000,\"smoothing\":1.0}",
-            "\"writeRateLimit\":{\"enabled\":true,\"limit\":999,\"rampUp\":0.0}");
-  }
+    final var requestLimiter = flowControlConfiguration.get(1).get("requestLimiter");
```
rodrigo-lourenco-lopes, line 59: Thanks for the refactor of this 🙌

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/flowcontrol/RateLimit.java
```diff
@@ -25,4 +38,25 @@ public RateLimiter limiter() {
     }
     return RateLimiter.create(limit, rampUp);
   }
+
+  public record Throttling(
```
rodrigo-lourenco-lopes, line 42: ❓ should we somehow assert that the throttling min rate is not higher than the rate limit if these are enabled?

rodrigo-lourenco-lopes (approved):
Thanks for the addition! 🚀 🚀

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/flowcontrol/RateLimit.java
lenaschoenburg: Good point, I added it here: ae02649

Issue #20164, opened 2024-07-09
Create message correlation record and intents


remcowesterhoud:
<h2>Description</h2>

- Create a new `MessageCorrelationRecord` with properties:
  - `name`
  - `correlationKey`
  - `variables`
  - `tenantId`
- Create a new `MessageCorrelationIntent` with values:
  - `CORRELATE`
  - `CORRELATED`
  - `NOT_CORRELATED`

PR #20197, opened 2024-07-10
Add `MessageCorrelation` record to ES and OS exporter

### LINKED ISSUE #20186: Create template for new message correlation record
<h2>Description</h2>

We msut create templates for the new record in the ES and OS exporter. Most of this can be copied from the [message record template](https://github.com/camunda/camunda/blob/main/zeebe/exporters/opensearch-exporter/src/main/resources/zeebe-record-message-template.json).

Differences are:
1. The name
2. No TTL
3. No Deadline
4. Add correlated boolean
5. Add process instance key

<h3>Dependencies</h3>

- #20164

remcowesterhoud:
## Description

Adds the record templates and configurations to allow `MessageCorrelation` records to be exported.

Note: this PR merges into #20164 as that one has tests failing because of this.

## Related issues

closes #20186

### DIFF
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -823,6 +823,7 @@
         #     jobBatch: false
         #     message: true
         #     messageBatch: false
+        #     messageCorrelation: true
         #     messageStartSubscription: true
         #     messageSubscription: true
         #     process: true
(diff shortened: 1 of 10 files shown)


nicpuppa (approved):
LGTM

🔧  You need to add also the new index option in `broker.yaml.template` and `broker.standalone.yaml.template`

Issue #20136, opened 2024-07-08
Flaky test shouldBeThreadSafe: io.camunda.zeebe.client.impl.oauth.OAuthCredentialsCacheTest


deepthidevaki:
**Summary**

> Try to answer the following as best as possible

- How often does the test fail?
- Does it block your work?
- Do we suspect that it is a real failure?

**Failures**
```
java.lang.RuntimeException: 
java.util.concurrent.ExecutionException: com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (File); line: 1, column: 1]
	at io.camunda.zeebe.client.impl.oauth.OAuthCredentialsCacheTest.lambda$shouldBeThreadSafe$4(OAuthCredentialsCacheTest.java:267)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at io.camunda.zeebe.client.impl.oauth.OAuthCredentialsCacheTest.shouldBeThreadSafe(OAuthCredentialsCacheTest.java:262)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
Caused by: java.util.concurrent.ExecutionException: com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (File); line: 1, column: 1]
	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
	at io.camunda.zeebe.client.impl.oauth.OAuthCredentialsCacheTest.lambda$shouldBeThreadSafe$4(OAuthCredentialsCacheTest.java:265)
	... 3 more
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (File); line: 1, column: 1]
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
	at com.fasterxml.jackson.databind.ObjectMapper._initForReading(ObjectMapper.java:4996)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4898)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3733)
	at io.camunda.zeebe.client.impl.oauth.OAuthCredentialsCache.readCache(OAuthCredentialsCache.java:62)
[...]

PR #20211, opened 2024-07-11
ci: disable flaky test to avoid CI disruption until flakiness resolved

cmur2:
## Description

See https://github.com/camunda/camunda/issues/20136#issuecomment-2222231245 for reasoning. We should first seek to mitigate impact on CI health due to flaky tests and then take time to fully resolve them so other engineers in the monorepo are not blocked during the resolution process.

## Related issues

Related to #20136

### DIFF
--- a/clients/java/src/test/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCacheTest.java
+++ b/clients/java/src/test/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCacheTest.java
@@ -34,6 +34,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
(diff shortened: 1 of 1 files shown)


clients/java/src/test/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCacheTest.java
```diff
@@ -36,8 +36,11 @@
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.jupiter.api.Disabled;
 import org.junit.rules.TemporaryFolder;
 
+@Disabled(
```
lenaschoenburg, line 42: :x: This is mixing junit4 and junit5 (aka "jupiter"). The test is written in junit4, the right annotation is `org.junit.Ignore`.

:x: AFAIK, the only test that consistently fails is `shouldBeThreadSafe`, so we should not disable all tests here, just this one.

lenaschoenburg (changes requested):

clients/java/src/test/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCacheTest.java
cmur2: > This is mixing junit4 and junit5

💡 Today I learned.

lenaschoenburg (approved):
Thanks :+1:

PR #20225, opened 2024-07-11
fix: flaky shouldReportUnhealthyIfTransitionStepIsStuck()

### LINKED ISSUE #18590: Test `ZeebePartitionTest.shouldReportUnhealthyIfTransitionStepIsStuck` is unstable
**Summary**

- How often does the test fail? 
   - Once
- Does it block your work?
  - No, but it decreases confidence in the CI. 
- Do we suspect that it is a real failure?
  - Not sure. 

**Failures**

<pre>
Error:  Tests run: 23, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.838 s <<< FAILURE! -- in io.camunda.zeebe.broker.system.partitions.ZeebePartitionTest
Error:  io.camunda.zeebe.broker.system.partitions.ZeebePartitionTest.shouldReportUnhealthyIfTransitionStepIsStuck -- Time elapsed: 0.073 s <<< FAILURE!
java.lang.AssertionError: 

Expecting actual:
  "Transition from null on term 0 appears blocked, step null has been running for PT2M0.013S"
to contain:
  "Transition from FOLLOWER on term 0 appears blocked" 
	at io.camunda.zeebe.broker.system.partitions.ZeebePartitionTest.shouldReportUnhealthyIfTransitionStepIsStuck(ZeebePartitionTest.java:419)
</pre>
</details>

**Hypotheses**

**Logs**

-  [Downloaded CI logs](https://drive.google.com/file/d/1W_ZsO04nsEU7qKXhsLkU9g-S2YzdzGt_/view?usp=sharing)

rodrigo-lourenco-lopes:
## Description

As referenced in the original issue https://github.com/camunda/camunda/issues/18590#issuecomment-2124786370, the problem seems to stem possible from some reace condition on `ctx.getCurrentRole()` mock. 

On the tests that fail, this returns a null value.

To address the issue we can use the Mockito method `doReturn()` that actually does not make any call to the original method and just replaces the method with the mocked value. This could fix this race condition.

More on [doReturn()](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#doReturn(java.lang.Object)) vs [when()](https://javadoc.io/static/org.mockito/mockito-core/5.12.0/org/mockito/Mockito.html#when(T))

closes https://github.com/camunda/camunda/issues/18590

### DIFF
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/system/partitions/ZeebePartitionTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/system/partitions/ZeebePartitionTest.java
@@ -12,6 +12,7 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
@@ -396,7 +397,7 @@ public class ZeebePartitionTest {
         .thenReturn(CompletableActorFuture.completed(null));
     when(transitionStep.transitionTo(any(), anyLong(), any()))
         .thenReturn(CompletableActorFuture.completed(null));
-    when(ctx.getCurrentRole()).thenReturn(Role.FOLLOWER);
+    doReturn(Role.FOLLOWER).when(ctx).getCurrentRole();
     partition.onNewRole(Role.FOLLOWER, 0);
     schedulerRule.workUntilDone();
 


deepthidevaki (commented):
Not sure how this fixes the issue. Here using `when(..).then()` seems perfectly reasonable. 
As your comment explains, there is definitely some race-condition between mocking this method and accessing it in partition actor. But I'm not sure if this fixes the issue. 

If mocking is the issue, can we use the concrete class here? May be we can use the `TestPartitionTransitionContext` here instead of the mock?

npepinpe:
@rodrigo-lourenco-lopes - any updates here?

deepthidevaki:
Could be also fixed by #21043 ?

rodrigo-lourenco-lopes:
Maybe we can close the pr and the issue for now and see if it repeats again

rodrigo-lourenco-lopes:
Will close this for now and the original issue and reinvestigate if the flake happens again.

PR #20239, opened 2024-07-12
fix: #20199 persist tenantId on variable record store

### LINKED ISSUE #20199: Issue when importing variables in a MT environment
Support ticket: https://jira.camunda.com/browse/SUPPORT-22513

When importing a variable in a multi-tenancy environment Tasklist is not saving properly the TenantId. We can check this on VariableZeebeRecordProcessorElasticSearch as the TenantId is not being set. This is causing the /v1/variables/:variableId to fail as it will check for permissions always on the default tenant as it's not being imported correctly.

panagiotisgts:
## Description
Include tenantId field from Zeebe record to process variable index store.

## Related issues
Ref: https://github.com/camunda/tasklist/issues/4989

Closes #20199


houssain-barouni (approved):
LGTM!

PR #20251, opened 2024-07-15
ci: 8.4 and earlier supports updates to the latest patch release only

### LINKED ISSUE #20157: Exclude specific versions from rolling update test
**Description**

Rolling update tests are failing on specific combination of versions.

Root cause:

```
 io.camunda.zeebe.snapshots.impl.InvalidSnapshotChecksum: Expected snapshot /usr/local/zeebe/data/raft-partition/partitions/1/snapshots/203-1-1704-1720 to have checksum 310528106, but was 3383963713
```

The new features for snapshot - rocksdb file checksums and snapshot chunk replication - were backported to old minors. This feature was designed so that a broker on old version (old checksum logic) can replicate snapshot to new version (with new file checksum). 

RollingUpdateTest fails for example from 8.3.13 -> 8.4.0. Here 8.3.13 has the new checksum logic, and 8.4.0 has the old checksum. So the new version cannot replicate snapshot to the old version and thus the test fails. The tests fails for other version combination to 

To fix this we can ensure both versions can replicate snapshot to each other. However, this is a bit more complex and require significant amount of work. Officially we support rolling update only from 8.5. So we can conclude that the above test failures are not bugs, instead we deliberately do not support rolling upgrade between the latest pre-minor to non-latest next minor. 

**Goal**

- Ignore specific version combination (pre 8.5) where we know rolling update is not possible. Example, only test rolling update from prev-minor to latest patch of the next minor.

lenaschoenburg:
## Description

8.4 and earlier only support updates to the latest patch of the next minor version and not to an arbitrary (not-latest) patch. This adjusts the test matrix of `RollingUpdateTest` to exclude unsupported combinations.

## Related issues

closes #20157

### DIFF
--- a/zeebe/qa/update-tests/src/test/java/io/camunda/zeebe/test/VersionCompatibilityMatrix.java
+++ b/zeebe/qa/update-tests/src/test/java/io/camunda/zeebe/test/VersionCompatibilityMatrix.java
@@ -23,6 +23,9 @@ import java.net.http.HttpResponse.BodyHandlers;
 import java.util.Comparator;
 import java.util.Optional;
 import java.util.SequencedCollection;
+import java.util.function.BinaryOperator;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.junit.jupiter.params.provider.Arguments;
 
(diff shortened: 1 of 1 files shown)


lenaschoenburg:
Test run is here: https://github.com/camunda/camunda/actions/runs/9937392591

deepthidevaki (approved):
:rocket:

PR #20258, opened 2024-07-15
fix: "adress" typo

akeller:
## Description

While working on running Camunda as a plain Java app and adding the steps to the docs, I encounter a typo in multiple files - `adress` instead of `address`. I don't think this is intentional. This PR addresses the typo via find & replace.

## Related issues

No related issue.

### DIFF
--- a/dist/src/main/config/application.yaml
+++ b/dist/src/main/config/application.yaml
@@ -134,7 +134,7 @@ camunda:
     # Zeebe instance
     zeebe:
       # Gateway address
-      gatewayAdress: localhost:26500
+      gatewayAddress: localhost:26500
     # ELS instance to export Zeebe data to
     zeebeElasticsearch:
       # Cluster name
--- a/tasklist/docker-compose.yml
+++ b/tasklist/docker-compose.yml
@@ -218,7 +218,7 @@ services:
     environment:
       - camunda.tasklist.elasticsearch.url=http://elasticsearch:9200
       - camunda.tasklist.zeebeElasticsearch.url=http://elasticsearch:9200
-      - camunda.tasklist.zeebe.gatewayAdress=zeebe:26500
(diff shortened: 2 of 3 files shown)


dist/src/main/config/application.yaml
```diff
@@ -134,7 +134,7 @@ camunda:
     # Zeebe instance
     zeebe:
       # Gateway address
-      gatewayAdress: localhost:26500
+      gatewayAddress: localhost:26500
```
megglos, line 137: I'm not mistaken this config was not effective due to the typo, right? As the underlying properties class defined it as `gatewayAddress` already.

tasklist/docker-compose.yml
```diff
@@ -218,7 +218,7 @@ services:
     environment:
       - camunda.tasklist.elasticsearch.url=http://elasticsearch:9200
       - camunda.tasklist.zeebeElasticsearch.url=http://elasticsearch:9200
-      - camunda.tasklist.zeebe.gatewayAdress=zeebe:26500
+      - camunda.tasklist.zeebe.gatewayAddress=zeebe:26500
```
megglos, line 221: same here given the typo the config should have had no effect, @camunda/human-task-orchestration-dev I wonder how this was working for you?

megglos (approved):
Thanks for the fix!

Raised some questions to @camunda/human-task-orchestration-dev as I wonder how e.g. the docker config even worked before, given the config value wasn't picked up due to the typo.

dist/src/main/config/application.yaml
megglos: ok, it worked because the default value is identical https://github.com/camunda/camunda/blob/main/tasklist/common/src/main/java/io/camunda/tasklist/property/ZeebeProperties.java#L12

npepinpe:
I assume we can merge this, it seems your questions were answered :)

PR #20265, opened 2024-07-16
Create `MessageCorrelationProcessor` processor

### LINKED ISSUE #20173: Create `MessageCorrelationCorrelateProcessor`

remcowesterhoud:
## Description

This processor will handle the `MessageCorrelation.CORRELATE` command. It will write the events that are already known from the existing `Message.PUBLISH` command. What's different is that this will send a response back to the client. 

- If it finds a message start event subscription it will respond with the process instance key of the first process instance it has found.
- If it finds only a message subscription it will correlate is as usual. There is no response yet. This is part of a later issue.
- If it finds no subscriptions it will respond with a `NOT_CORRELATED` event.

Note: I worked test driven and got too excited. Some of the tests I wrote won't succeed until I've implemented some other features. I've added an ignore to those for now.

## Related issues

closes #20173

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelateBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelateBehavior.java
@@ -8,23 +8,36 @@
 package io.camunda.zeebe.engine.processing.message;
 
 import io.camunda.zeebe.engine.processing.common.EventHandle;
+import io.camunda.zeebe.engine.processing.message.command.SubscriptionCommandSender;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
 import io.camunda.zeebe.engine.state.immutable.MessageStartEventSubscriptionState;
 import io.camunda.zeebe.engine.state.immutable.MessageState;
+import io.camunda.zeebe.engine.state.immutable.MessageSubscriptionState;
+import io.camunda.zeebe.protocol.record.intent.MessageSubscriptionIntent;
 import org.agrona.DirectBuffer;
 
 public final class MessageCorrelateBehavior {
 
   private final MessageStartEventSubscriptionState startEventSubscriptionState;
+  private final MessageSubscriptionState messageSubscriptionState;
   private final MessageState messageState;
   private final EventHandle eventHandle;
+  private final StateWriter stateWriter;
+  private final SubscriptionCommandSender commandSender;
 
   public MessageCorrelateBehavior(
       final MessageStartEventSubscriptionState startEventSubscriptionState,
       final MessageState messageState,
-      final EventHandle eventHandle) {
+      final EventHandle eventHandle,
+      final StateWriter stateWriter,
+      final MessageSubscriptionState messageSubscriptionState,
+      final SubscriptionCommandSender commandSender) {
     this.startEventSubscriptionState = startEventSubscriptionState;
+    this.messageSubscriptionState = messageSubscriptionState;
     this.messageState = messageState;
     this.eventHandle = eventHandle;
+    this.stateWriter = stateWriter;
+    this.commandSender = commandSender;
   }
 
   public Subscriptions correlateToMessageStartEvents(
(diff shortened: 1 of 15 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+            startEventSubscriptionState,
+            messageState,
+            eventHandle,
+            stateWriter,
+            messageSubscriptionState,
+            commandSender);
+  }
+
+  @Override
+  public void processRecord(final TypedRecord<MessageCorrelationRecord> command) {
```
nicpuppa, line 69: 💭 I think the processor should be as simple as possible. All the logic should be moved into the Behaviour class. In this way the processor is in charge to only call the right methods, and all the logic will be in the same place. Wdyt ?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelateBehavior.java
```diff
@@ -0,0 +1,126 @@
...
+    startEventSubscriptionState.visitSubscriptionsByMessageName(
+        tenantId,
+        messageName,
+        subscription -> {
+          final var subscriptionRecord = subscription.getRecord();
+          final var bpmnProcessIdBuffer = subscriptionRecord.getBpmnProcessIdBuffer();
+
+          // create only one instance of a process per correlation key
+          // - allow multiple instance if correlation key is empty
+          if (!correlatingSubscriptions.contains(bpmnProcessIdBuffer)
```
nicpuppa, line 60: 🔧 This condition might be a bit hard to read. I'd suggest to extract the checks in variables to improve the readability. eg.

```
boolean isCorrelationKeyEmpty = correlationKey.capacity() == 0;
boolean processInstanceExists = messageState.existActiveProcessInstance(
            tenantId, bpmnProcessIdBuffer, correlationKey);
```

Maybe with better naming 😄

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/ReusableObjectList.java
```diff
@@ -75,6 +75,17 @@ public T poll() {
     return null;
   }
 
+  public T peek() {
```
nicpuppa, line 78: 🔧 I think this can be improved:

```
public T peek() {
    for (ReusableElement element : elements) {
        if (element != null && element.isSet()) {
            return element.getElement();
        }
    }
    return null;
}
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationProcessor.java
nicpuppa: Sorry maybe my explanation is not good enough. What I want to say is that we can move all the logic in the `processRecord()` method into the behaviour class, but NOT in the already existing methods.

At the end the processor should be like:

```
@Override
public void processRecord(final TypedRecord<MessageCorrelationRecord> command) {
    correlateBehaviour.doSomething(.....)
}
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/message/CorrelateMessageTest.java
```diff
@@ -0,0 +1,450 @@
...
+            MessageStartEventSubscriptionIntent.CREATED,
+            MessageCorrelationIntent.CORRELATE,
+            MessageIntent.PUBLISHED,
+            MessageStartEventSubscriptionIntent.CORRELATED,
+            MessageCorrelationIntent.CORRELATED,
+            MessageIntent.EXPIRED);
+  }
+
+  @Test
+  @Ignore("Enable in https://github.com/camunda/camunda/issues/20175")
```
nicpuppa, line 97: ❤️

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/message/CorrelateMessageTest.java
```diff
@@ -0,0 +1,450 @@
...
+  @Rule
+  public final RecordingExporterTestWatcher recordingExporterTestWatcher =
+      new RecordingExporterTestWatcher();
+
+  @Test
+  public void shouldNotCorrelateWhenNoSubscriptions() {
+    // when
+    engine
+        .messageCorrelation()
+        .withCorrelationKey("correlationKey")
```
nicpuppa, line 48: 🔧 We can move the correlationKey and the messageName outside the tests cases, make them general available. Same for the variables

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/message/CorrelateMessageTest.java
```diff
@@ -0,0 +1,450 @@
...
+            .withName(messageName)
+            .withCorrelationKey(correlationKey)
+            .correlate();
+
+    // then
+    assertMessageIsCorrelated(record);
+  }
+
+  @Test
+  public void shouldNotCorrelateMessageIfNoProcess() {
```
nicpuppa, line 195: ❓ What's the difference between this test and `shouldNotCorrelateWhenNoSubscriptions` ?

remcowesterhoud: The `shouldNotCorrelateWhenNoSubscriptions` test only checks some commands/events have been written. This one checks some more specifics about the event, such as message name and correlation key.

nicpuppa (approved):

remcowesterhoud:
Seems I broke some tests 👀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelateBehavior.java
remcowesterhoud: I ended up dropping this refactoring. The problem is that `messageState.existActiveProcessInstance(tenantId, bpmnProcessIdBuffer, correlationKey);` requires the correlation key to not be empty. This means it would result in problems when correlationkey is actually empty.

remcowesterhoud:
Having to retry a merge 5+ times per PR is driving me insane :/

nicpuppa:
I might be more lucky than you @remcowesterhoud 🤣

remcowesterhoud:
If this works I'm just gonna use you like we used Bors. 

@nicpuppa merge

nicpuppa:
nicbot @remcowesterhoud

remcowesterhoud:
I can already see that it failed again. The queue just doesn't know it yet 😞

PR #20266, opened 2024-07-16
Upgrade to rocksdb 9.4.0

### LINKED ISSUE #19981: Update RocksDB to 9.x
**Description**

RocksDB released a new major version with many features. While 8.x is still maintained, it's not clear to me for how long, so we should migrate sooner than later. However, there are some breaking changes which require manual effort, hence why we have an issue and aren't relying on Renovate here.

It should still be compatible with previous versions however, the breaking changes are in the APIs and not the database itself as far as I know - this should be verified with our compatibility tests, of course.

EuroLew:
## Description

Upgrading rocksDB to 9.4.0 and fix all broken APIs as a result of the private reflected methods being changed from non-static to static. 

## Related issues

closes #19981

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -80,7 +80,7 @@
     <version.protobuf>3.25.3</version.protobuf>
     <version.protobuf-common>2.41.0</version.protobuf-common>
     <version.micrometer>1.13.2</version.micrometer>
-    <version.rocksdbjni>8.11.4</version.rocksdbjni>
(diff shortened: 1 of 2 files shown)


deepthidevaki (approved):
Thanks. So the api change was for the internal private methods that was changed from non-static to static. Is that right? Please add this to the PR description.

EuroLew:
> Thanks. So the api change was for the internal private methods that was changed from non-static to static. Is that right? Please add this to the PR description.

Yep

Issue #17684, opened 2024-04-23
Preview Environments for the monorepo


cmur2:
## Description

Support a feature to create [Preview Environments](https://confluence.camunda.com/display/HAN/Preview+Environments) that allow engineers to test all components of the monorepo in temporary instances. This feature was available for Operate & Tasklist individually in the past. Security considerations are needed since this is a public repository, see https://github.com/camunda/connectors

The monorepo needs to be onboarded to the existing preview environment infrastructure and configuration via Helm like https://github.com/camunda/operate/tree/master/.ci/preview-environments needs to be created covering Zeebe, Operate, Tasklist. Github Actions workflows triggering on labels need to be added.

mschoe:
@cmur2 do you have a rough time line for this issue? The current situation without preview environment is slowing the QA process down and makes testing in the SaaS environment cumbersome. Furthermore the process is quite error prone as it involves a lot of manual steps.

cc// @johanwelgemoed @multani @marcosgvieira

cmur2:
> do you have a rough time line for this issue?

I cannot give out a rough timeline here yet since I'm still working on the roadmap. Tentatively there are other topics with higher priority than this also related to the monorepo migration.

> and makes testing in the SaaS environment cumbersome.

What is the relation to SaaS here? So far [Preview Environments](https://confluence.camunda.com/pages/viewpage.action?spaceKey=HAN&title=Preview+Environments) are not integrated with SaaS for most teams, so I'm wondering.

multani:
@cmur2 at the moment, engineers have to test their pull request images in SaaS: they are pulling images from registry.camunda.cloud and pushing them to our SaaS-specific registry. It was already the case before the mono-repo, for when developers wanted to test branches in our dev SaaS environment.

We could make that process a little bit easier by _not having_ to copy images between registries, but we would need to know a bit more how registry.camunda.cloud is organized and used by developers.

cmur2:
> they are pulling images from registry.camunda.cloud and pushing them to our SaaS-specific registry. It was already the case before the mono-repo

You are looking to make this workflow smoother but the request is unrelated to the monorepo transition (during which the [Preview Environments](https://confluence.camunda.com/pages/viewpage.action?spaceKey=HAN&title=Preview+Environments) functionality was lost - which this ticket is about)? If so, we can track this as a separate (sub-)ticket for integration with SaaS then to avoid mixing discussions with the Self-Managed side of things.

cmur2:
[...]

PR #20268, opened 2024-07-16
ci: remove unused CI files introduced by Optimize

buccarel:
This PR removes the entries in the .ci/ directory that were introduced by the optimize merge.
The files that were there before optimize have been left untouched.


HeleneW-dot (approved):
Lgtm as far as removing all the c7 stuff, are we sure we dont need the files in `c8saas` and `c8sm`? What do these files do? There are also failing Optimize Helm Chart Integration tests, can you please check if those are related

HeleneW-dot (changes requested):
Meant to request changes see the comment above

buccarel:
> Lgtm as far as removing all the c7 stuff, are we sure we dont need the files in `c8saas` and `c8sm`? What do these files do? There are also failing Optimize Helm Chart Integration tests, can you please check if those are related

I believe those files are under the deployment-preview directory, which is something that is not supposed to stay in the monorepo: preview environments will be added later with this https://github.com/camunda/camunda/issues/17684

PHWaechtler:
> > Lgtm as far as removing all the c7 stuff, are we sure we dont need the files in `c8saas` and `c8sm`? What do these files do? There are also failing Optimize Helm Chart Integration tests, can you please check if those are related
> 
> I believe those files are under the deployment-preview directory, which is something that is not supposed to stay in the monorepo: preview environments will be added later with this #17684

Makes sense 

> About the helm chart failures, I don't really know. They don't seem to look for any file under `.ci/`. Am I missing anything?

Are these tests also failing on the merge PR? This PR is removing files from a directory named Helm-charts. Some of the errors look like something is misconfigured or missing: `Input required and not supplied: $***name***``, some errors look like an incorrect command: `Error: flags cannot be placed before plugin name: -n`. The workflow files looks to have been introduced by Michal, maybe he is more familiar with the errors?

HeleneW-dot (changes requested):
Another question I noticed while looking into workflow removal - this isnt removing podSpecs/clusterTests. Is there. a reason for that?

misiekhardcore:
I can see that `.github/workflows/optimize-preview-env-deploy.yml` is still referencing files in `.ci`. Should we also delete/modify this workflow?

misiekhardcore (approved):

buccarel:
> I can see that `.github/workflows/optimize-preview-env-deploy.yml` is still referencing files in `.ci`.

I think that file shall be removed, yeah. Thanks for spotting it.

> Should we also delete/modify this workflow? Also there is a `.github/workflows/optimize-backend-linting.yml` workflow which also points to that directory, maybe it also need an update or maybe we need to remove it if there is a monorepo linter.

Yes, we don't need to lint that anymore. The directory .ci that is left in the repo does not belong to Optimize. Thanks

Will go ahead and address these comments, thanks

HeleneW-dot (approved):
LGTM if the rest of the pipelines wont fail (with new errors)

PR #20276, opened 2024-07-16
ci: skip optimize-upgrade docker when testing

buccarel:
## Description

With this PR, we avoid launching the `docker compose up` command that starts the elasticseach container of the optimize-upgrade module, as this conflicts with the build environment of other teams.


OmranAbazid (approved):
pre-approving. 
I think the commit message should follow the convention since it is failing here:
https://github.com/camunda/camunda/actions/runs/9955837403/job/27504437065?pr=20276

PR #20278, opened 2024-07-16
fix: Tasklist index prefixes are not correctly setup on OpenSearch

### LINKED ISSUE #19905: Tasklist index prefixes are not correctly setup on OpenSearch
### Describe the bug

When specifying the environment variable CAMUNDA_TASKLIST_OPENSEARCH_INDEXPREFIX (value: "tasklistoos"), indices are not correctly created, and index names are still set as default (tasklist-*). 
Only the 'composed_of' template is correctly set to the given name.

When executing: 
`GET /_index_template/tasklist*`

The following output will be shown:
`{
  "index_templates": [
    {
      "name": "tasklist-task-variable-8.3.0_template",
      "index_template": {
        "index_patterns": [
          "tasklist-task-variable-8.3.0_*"
        ],
        "template": {
          "mappings": {
            "dynamic": "strict",
            "properties": {
              "fullValue": {
                "ignore_above": 8191,
                "type": "keyword"
              },
              "isPreview": {
                "type": "boolean"
              },
              "partitionId": {
                "type": "integer"
              },
              "name": {
                "type": "keyword"
              },
              "tenantId": {
                "type": "keyword"
              },
              "id": {
                "type": "keyword"
              },
              "value": {
                "ignore_above": 8191,
                "type": "keyword"
              },
              "taskId": {
                "type": "keyword"
              },
              "key": {
                "type": "long"
              }
            }
          },
          "aliases": {
            "tasklist-task-variable-8.3.0_alias": {}
          }
        },
        "composed_of": [
          "tasklistoos_template"
        ]
      }
    },
[...]`

I have also tested the behavior against ES locally, where it works correctly. 

### To Reproduce

- Set up a managed OpenSearch cluster (version 2.5) on AWS.
- Utilize helm values 10.2.0 to connect to the cluster.
- Specify the CAMUNDA_TASKLIST_OPENSEARCH_INDEXPREFIX environment variable with the value "tasklistoos".
- Deploy the configuration.
- Execute GET /_index_template/tasklist* in OpenSearch and observe the output.

### Actual behavior

Indices are not created with the specified prefix (tasklistoos-*); instead, they are created with the default prefix (tasklist\*). Only the composed_of template is correctly set to the given name.

### Expected behavior

All indices should be created with the specified prefix (tasklistoos-*), allowing the use of a shared OpenSearch cluster without issues.

### Environment

- Tasklist Version: 8.5.3
- Camunda Helm: 10.2.0
- Amazon OpenSearch Service 2.5

houssain-barouni:
## Description

Fixes the following issue reported with helm deployement when setting different index prefix in Opensearch

- to get the prefix to be used for indices, Tasklist uses [TasklistOpenSearchProperties](https://github.com/camunda/camunda/blob/main/tasklist/common/src/main/java/io/camunda/tasklist/property/TasklistOpenSearchProperties.java) or [TasklistElasticSearchProperties](https://github.com/camunda/camunda/blob/main/tasklist/common/src/main/java/io/camunda/tasklist/property/TasklistElasticsearchProperties.java), depending if Opensearch database is used or not
- Tasklist [reads](https://github.com/camunda/camunda/blob/5d2d5eeb723ac946547b5b8e8548cdeba48c1250/tasklist/common/src/main/java/io/camunda/tasklist/util/TasklistPropertiesUtil.java#L19) `camunda.tasklist.database` system property or system env parameter to know if opensearch being used
- in the helm charts, `camunda.tasklist.database` is defined inside application.yml key in a configMap. Which is read a spring property and not system property.
- `getTasklistDatabase()` returns null, and Tasklist defaults [TasklistElasticSearchProperties](https://github.com/camunda/camunda/blob/main/tasklist/common/src/main/java/io/camunda/tasklist/property/TasklistElasticsearchProperties.java), in which the index prefix is the default one (tasklist) and not the one defined in `CAMUNDA_TASKLIST_OPENSEARCH_INDEXPREFIX`
- The indices are created with tasklist prefix

In this pull request:
- read `camunda.tasklist.database` as spring property
- cache `isOpensearchDatabase`, as it is static, to avoid multiple calls to `applicationContext.getEnvironment().getProperty()`
## Related issues

closes #19905 

This needs to be backported to 8.5 and 8.4 (Opensearch was officially supported since 8.4)

### DIFF
--- a/tasklist/common/pom.xml
+++ b/tasklist/common/pom.xml
@@ -275,6 +275,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+
   </dependencies>
 
   <build>
--- a/tasklist/common/src/main/java/io/camunda/tasklist/util/SpringContextHolder.java
+++ b/tasklist/common/src/main/java/io/camunda/tasklist/util/SpringContextHolder.java
@@ -28,6 +28,17 @@ public class SpringContextHolder implements ApplicationContextAware {
     return context.getBean(beanClass);
   }
 
+  /**
+   * Retrieves the property value associated with the given key from Spring's application context
+   * environment.
+   *
+   * @param key The key whose associated value is to be retrieved.
+   * @return The value associated with the specified key, or null if the key is not found.
+   */
+  public static String getProperty(String key) {
+    return context.getEnvironment().getProperty(key);
+  }
+
   @Override
   public void setApplicationContext(ApplicationContext context) throws BeansException {
     // store ApplicationContext reference to access required beans later on
--- a/tasklist/common/src/main/java/io/camunda/tasklist/util/TasklistPropertiesUtil.java
+++ b/tasklist/common/src/main/java/io/camunda/tasklist/util/TasklistPropertiesUtil.java
@@ -9,17 +9,25 @@ package io.camunda.tasklist.util;
 
 public final class TasklistPropertiesUtil {
 
-  private static final String DATABASE_PROPERTY_NAME = "camunda.tasklist.database";
+  public static final String DATABASE_PROPERTY_NAME = "camunda.tasklist.database";
+  private static volatile Boolean isOpenSearchDatabaseCache = null;
 
   private TasklistPropertiesUtil() {
     /*utility class*/
   }
 
-  public static String getTasklistDatabase() {
-    return System.getProperty(DATABASE_PROPERTY_NAME, System.getenv(DATABASE_PROPERTY_NAME));
+  public static boolean isOpenSearchDatabase() {
+    if (isOpenSearchDatabaseCache == null) {
+      synchronized (TasklistPropertiesUtil.class) {
+        if (isOpenSearchDatabaseCache == null) {
+          isOpenSearchDatabaseCache = "opensearch".equalsIgnoreCase(getTasklistDatabase());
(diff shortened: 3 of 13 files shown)


nathansandi (approved):
LGMT

panagiotisgts (commented):
Regarding the changes everything seems OK.

However, I believe the issue lies somewhere else in the configuration. The changelog seems excessive for _just reading a configuration property_. Assuming that the System property does not exist, it should be replaced by the Config Map properties set on deployment. 

Have we explored other options for this issue?

houssain-barouni:
> Regarding the changes everything seems OK.
> 
> However, I believe the issue lies somewhere else in the configuration. The changelog seems excessive for _just reading a configuration property_. 

Well, only two classes were updated, the rest is mainly about adding and fixing tests.
The main problem is that we are reading this property using static methods, which forces us to use this `ApplicationContextAware` component . Replacing static methods use by something more standard (like using `@Value` or a Spring component) would require more code change.

>Assuming that the System property does not exist, it should be replaced by the Config Map properties set on deployment.
> 
> Have we explored other options for this issue?

The "problem" comes from the fact that the helm chart sets this property as a spring configuration parameter and not as an env parameter to the container. We can change the helm to switch to the latter but it is documented at the same place where we document the  `camunda.tasklist.database`  that we supports `any standard Spring configuration method`. https://docs.camunda.io/docs/self-managed/tasklist-deployment/tasklist-configuration/

houssain-barouni:
Replaced by https://github.com/camunda/camunda/pull/20311

PR #20092, opened 2024-07-05
Rename and deprecate Zeebe classes

nicpuppa:
## Description

With this PR all the `Zeebe....` classes are replaced or deprecated in favor of more general classes. 

## Related issues

closes #19485 
closes #19488

PR #20285, opened 2024-07-16
ci: add java client reminder workflow

### LINKED ISSUE #20281: Create a GH workflow as a reminder to check Java Client if there is some changes
With this https://github.com/camunda/camunda/pull/20092 we introduce some structural changes to the java client:

In `io.camunda.client` there is the new Camunda Java Client with all the new features and fix that should be delivered in the next release
In `io.camunda.zeebe.client` there is the Zeebe Java Client (updated at the last stable version)
Why we took this decision ? Because we want to enable users to use the new client (and the new features) without changing all the codebase. Users can use the new functionalities (using the new `CamundaClient`) but they can still use the old `ZeebClient` without any problem.

This workflow is a reminder to ensure that people are informed

nicpuppa:
## Description

With this https://github.com/camunda/camunda/pull/20092 we introduce some structural changes to the java client:

In `io.camunda.client` there is the new Camunda Java Client with all the new features and fix that should be delivered in the next release
In `io.camunda.zeebe.client` there is the Zeebe Java Client (updated at the last stable version)
Why we took this decision ? Because we want to enable users to use the new client (and the new features) without changing all the codebase. Users can use the new functionalities (using the new `CamundaClient`) but they can still use the old `ZeebClient` without any problem.

This workflow is a reminder to ensure that people are informed

## Related issues

closes #20281

### DIFF
--- a/.github/workflows/java-client-reminder.yaml
+++ b/.github/workflows/java-client-reminder.yaml
@@ -1,0 +1,35 @@
+name: Camunda Java Client reminder
+
+on:
+  pull_request
+
+jobs:
+  check-path-and-comment:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Check for changes in 'clients/java'
+        id: check_changes
+        run: |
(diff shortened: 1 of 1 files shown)


megglos (approved):
Great idea! LGTM!

Issue #19384, opened 2024-06-14
[EPIC] Support `deployment` version binding for linked resources


jfriedenstab:
## Introduction
The goal of this epic is to add support for `deployment` version binding for linked resources to Zeebe, similar to what already exists in Camunda 7. This should give users a simple and safe way to reference resources that belong together in a process solution.

Currently, when a sub process, decision or form is linked from a BPMN process, Zeebe will always resolve this link to the latest known version of the target resource which might lead to unexpected behavior. With this epic, it will be possible instead to make Zeebe resolve a link reliably within the same "deployment unit".

## Task Breakdown

```[tasklist]
### BPMN Model API
- [ ] https://github.com/camunda/camunda/issues/19461
- [ ] https://github.com/camunda/camunda/issues/20054
``` 

```[tasklist]
### Engine
- [ ] https://github.com/camunda/camunda/issues/19859
- [ ] https://github.com/camunda/camunda/issues/19709
- [ ] https://github.com/camunda/camunda/issues/19710
- [ ] https://github.com/camunda/camunda/issues/19712
```

```[tasklist]
### Documentation
- [ ] https://github.com/camunda/camunda-docs/issues/4124
```

```[tasklist]
### Misc
- [ ] https://github.com/camunda/web-modeler/issues/10093
```

### Out of scope
- https://github.com/camunda/camunda/issues/20067

## Other Teams

- [x] [Desktop Modeler](https://github.com/camunda/camunda-modeler/issues):
  - https://github.com/camunda/camunda-modeler/issues/4385
  - https://github.com/camunda/camunda-modeler/issues/4386

## Links

- Product Hub (internal): https://github.com/camunda/product-hub/issues/1920

tmetzke:
Hey @jfriedenstab, I was looking at this feature in the context of the C8 REST API and I was wondering:
Is there a specific reason we don't support this for start event forms as well? We support this for user task forms but not for start events.

PR #20293, opened 2024-07-16
feat: Incident records provide tree path properties that locate the incident in a hierarchy of process instances

### LINKED ISSUE #16915: Incident records provide tree path properties that locate the incident in a hierarchy of process instances

* For the [new Elasticsearch exporter that writes Operate entities](https://github.com/camunda/product-hub/issues/2128), we need to provide the so-called tree path. It is a concept to assign entities their location in a tree of process instances and BPMN scopes, so that we can query for it and find all entities within a particular subtree. Find the full context and explanation in the [design document](https://docs.google.com/document/d/1Nh8Jln5_aWIS1O5NnC-SGRblijizDTSujG7zIfc-GOY/edit).
* With this feature, we want to provide the tree path for incident records, because that is where Operate currently needs the functionality. In the future and as the feature set evolves, we may roll it out to more records.
* [Product hub epic](https://github.com/camunda/product-hub/issues/2128)

* The following uses this example process hierarchy: 
  * Process A (2 sub processes + call activity X)
    * Process B (plain call activity Y)
      * Process C (1 sub process)

* Incident records should have three new fields:
  * BPMN scope instance path: 
    * One entry per process instance in the hierarchy of the incident
    * Each process instance entry contains all the BPMN scope and call activity instance keys
    * Example path: 101/102/103/104;105/106;107/108 (we could also add the leaf instance key)
  * Calling process definition path:
    * One entry per process definition in the hierarchy
    * Each entry is the key of the process definition the corresponding process instance
    * Example (for same hierarchy as above): A/B
    * The leaf process definition can but doesn’t have to be included
  * Calling element path:
    * One entry per call activity in the hierarchy, 
    * Each entry is a reference to the call activity in BPMN
    * Exact format to be decided during implementation. Should not be the flow node id (as that would require a lot of space), but a numerical reference into the BPMN model. Accordingly, that requires the process definition to decode.
      * Idea: could be the index of the BPMN element's position in the XML model (counted by start tag)
    * Example (for same hierarchy as above): encode(X)/encode(Y) = 3/5

* See the [design document](https://docs.google.com/document/d/1Nh8Jln5_aWIS1O5NnC-SGRblijizDTSujG7zIfc-GOY/edit) for how we got to this proposal

**Additional context**

* See the [design document](https://docs.google.com/document/d/1Nh8Jln5_aWIS1O5NnC-SGRblijizDTSujG7zIfc-GOY/edit)

mustafadagher:
## Description
For the https://github.com/camunda/product-hub/issues/2128, we need to provide the so-called tree path. It is a concept to assign entities their location in a tree of process instances and BPMN scopes, so that we can query for it and find all entities within a particular subtree.
With this feature, we want to provide the tree path for incident records, because that is where Operate currently needs the functionality. In the future and as the feature set evolves, we may roll it out to more records.

For this implementation, I've created an `ElementTreePathBuilder` that uses the `elementInstanceKey` and `elementInstanceState` to build the tree-path properties.

The changes provide 3 new properties to the `IncidentRecord`:
1. `elementInstancePath`: A list of lists, each list is contains the `elementInstanceKey`s  for all the elements contained within one process definition in the path to an incident.
2. `processDefinitionPath`: A list of `processDefinitionKey`s for each process in the path for an incident
3. `callingElementPath`: A list of encoded `ElementId`s for each `CallActivity` element in the path for an incident. ~~Since ElementIds are just a limited number of strings unique locally to the deployed model
we could use a hashing function to hash these Ids in the callingElementTreePath. Operate can use the same hashing function to hash all elementIds in a model and build a reverse `Map<Hash, ElementId>`, and use this map to look up the elementId string using the hash. I decided to use `murmur3_128` because it is a fast hashing function with low chance of collision. And I choose an implementation from `google guava`~~ 
The Element Ids are encoded by lexicographically sorting all calling activity elements in one deployment resource, and using the index of each call activity Id in this sorted list as the respective encoded value.

## Related issues

closes #16915

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
@@ -11,21 +11,31 @@ import io.camunda.zeebe.engine.state.immutable.ElementInstanceState;
 import io.camunda.zeebe.engine.state.instance.ElementInstance;
 import java.util.Deque;
 import java.util.LinkedList;
+import java.util.Objects;
 
 public class ElementTreePathBuilder {
-  private final ElementInstanceState elementInstanceState;
-  private final Long elementInstanceKey;
-  private final ElementTreePathProperties properties;
+  private ElementInstanceState elementInstanceState;
+  private Long elementInstanceKey;
+  private ElementTreePathProperties properties;
 
-  public ElementTreePathBuilder(
-      final ElementInstanceState elementInstanceState, final Long elementInstanceKey) {
+  public ElementTreePathBuilder() {}
+
+  public ElementTreePathBuilder withElementInstanceState(
+      final ElementInstanceState elementInstanceState) {
     this.elementInstanceState = elementInstanceState;
-    this.elementInstanceKey = elementInstanceKey;
-    properties =
-        new ElementTreePathProperties(new LinkedList<>(), new LinkedList<>(), new LinkedList<>());
+    return this;
   }
 
-  public ElementTreePathProperties getProperties() {
+  public ElementTreePathBuilder withElementInstanceKey(final long elementInstanceKey) {
+    this.elementInstanceKey = elementInstanceKey;
+    return this;
+  }
+
+  public ElementTreePathProperties build() {
+    Objects.requireNonNull(elementInstanceState, "elementInstanceState cannot be null");
+    Objects.requireNonNull(elementInstanceKey, "elementInstanceKey cannot be null");
+    properties =
+        new ElementTreePathProperties(new LinkedList<>(), new LinkedList<>(), new LinkedList<>());
     buildElementTreePathProperties(elementInstanceKey);
     return properties;
   }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
@@ -41,8 +41,10 @@ public class ElementTreePathBuilder {
       parentElementInstanceKey = instance.getParentKey();
     }
     properties.elementInstancePath.addFirst(elementInstancePath);
+    final var processInstanceRecord = instance.getValue();
+    properties.processDefinitionPath.addFirst(processInstanceRecord.getProcessDefinitionKey());
 
-    final long callingElementInstanceKey = instance.getValue().getParentElementInstanceKey();
+    final long callingElementInstanceKey = processInstanceRecord.getParentElementInstanceKey();
     if (callingElementInstanceKey != -1) {
       buildElementTreePathProperties(callingElementInstanceKey);
     }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
@@ -7,8 +7,12 @@
  */
 package io.camunda.zeebe.engine.processing.common;
 
+import com.google.common.hash.HashCode;
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hashing;
 import io.camunda.zeebe.engine.state.immutable.ElementInstanceState;
 import io.camunda.zeebe.engine.state.instance.ElementInstance;
+import java.nio.charset.Charset;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
@@ -17,8 +21,11 @@ public class ElementTreePathBuilder {
   private ElementInstanceState elementInstanceState;
   private Long elementInstanceKey;
   private ElementTreePathProperties properties;
+  private final HashFunction hashFunction;
 
-  public ElementTreePathBuilder() {}
+  public ElementTreePathBuilder() {
+    hashFunction = Hashing.murmur3_128();
+  }
 
   public ElementTreePathBuilder withElementInstanceState(
       final ElementInstanceState elementInstanceState) {
@@ -56,21 +63,24 @@ public class ElementTreePathBuilder {
 
     final long callingElementInstanceKey = processInstanceRecord.getParentElementInstanceKey();
     if (callingElementInstanceKey != -1) {
-      properties.callingElementPath.addFirst(getCallActivityId(callingElementInstanceKey));
+      properties.callingElementPath.addFirst(getCallActivityIdHash(callingElementInstanceKey));
       buildElementTreePathProperties(callingElementInstanceKey);
     }
   }
 
-  private String getCallActivityId(final long callingElementInstanceKey) {
+  private Long getCallActivityIdHash(final long callingElementInstanceKey) {
     final ElementInstance callActivityElementInstance =
         elementInstanceState.getInstance(callingElementInstanceKey);
     final var callActivityInstanceRecord = callActivityElementInstance.getValue();
 
-    return callActivityInstanceRecord.getElementId();
+    final HashCode hashCode =
+        hashFunction.hashString(
+            callActivityInstanceRecord.getElementId(), Charset.defaultCharset());
+    return hashCode.asLong();
   }
 
   public record ElementTreePathProperties(
       List<List<Long>> elementInstancePath,
       List<Long> processDefinitionPath,
-      List<String> callingElementPath) {}
+      List<Long> callingElementPath) {}
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
@@ -9,9 +9,10 @@ package io.camunda.zeebe.engine.processing.common;
 
 import io.camunda.zeebe.engine.state.immutable.ElementInstanceState;
 import io.camunda.zeebe.engine.state.instance.ElementInstance;
-import java.util.Deque;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Objects;
+import org.agrona.DirectBuffer;
 
 public class ElementTreePathBuilder {
   private ElementInstanceState elementInstanceState;
@@ -41,7 +42,7 @@ public class ElementTreePathBuilder {
   }
 
   private void buildElementTreePathProperties(final long elementInstanceKey) {
-    final Deque<Long> elementInstancePath = new LinkedList<>();
+    final List<Long> elementInstancePath = new LinkedList<>();
     elementInstancePath.add(elementInstanceKey);
     ElementInstance instance = elementInstanceState.getInstance(elementInstanceKey);
     long parentElementInstanceKey = instance.getParentKey();
@@ -56,12 +57,21 @@ public class ElementTreePathBuilder {
 
     final long callingElementInstanceKey = processInstanceRecord.getParentElementInstanceKey();
     if (callingElementInstanceKey != -1) {
+      properties.callingElementPath.addFirst(getCallActivityId(callingElementInstanceKey));
       buildElementTreePathProperties(callingElementInstanceKey);
     }
   }
 
+  private DirectBuffer getCallActivityId(final long callingElementInstanceKey) {
+    final ElementInstance callActivityElementInstance =
+        elementInstanceState.getInstance(callingElementInstanceKey);
+    final var callActivityInstanceRecord = callActivityElementInstance.getValue();
+
+    return callActivityInstanceRecord.getElementIdBuffer();
+  }
+
   public record ElementTreePathProperties(
-      Deque<Deque<Long>> elementInstancePath,
-      Deque<Long> processDefinitionPath,
-      Deque<Long> callingElementPath) {}
+      List<List<Long>> elementInstancePath,
+      List<Long> processDefinitionPath,
+      List<DirectBuffer> callingElementPath) {}
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
@@ -12,7 +12,6 @@ import io.camunda.zeebe.engine.state.instance.ElementInstance;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
-import org.agrona.DirectBuffer;
 
 public class ElementTreePathBuilder {
   private ElementInstanceState elementInstanceState;
@@ -62,16 +61,16 @@ public class ElementTreePathBuilder {
     }
   }
 
-  private DirectBuffer getCallActivityId(final long callingElementInstanceKey) {
+  private String getCallActivityId(final long callingElementInstanceKey) {
     final ElementInstance callActivityElementInstance =
         elementInstanceState.getInstance(callingElementInstanceKey);
     final var callActivityInstanceRecord = callActivityElementInstance.getValue();
 
-    return callActivityInstanceRecord.getElementIdBuffer();
+    return callActivityInstanceRecord.getElementId();
   }
 
   public record ElementTreePathProperties(
       List<List<Long>> elementInstancePath,
       List<Long> processDefinitionPath,
-      List<DirectBuffer> callingElementPath) {}
+      List<String> callingElementPath) {}
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
@@ -68,7 +68,7 @@ public class ElementTreePathBuilder {
     }
   }
 
-  private Long getCallActivityIdHash(final long callingElementInstanceKey) {
+  private String getCallActivityIdHash(final long callingElementInstanceKey) {
     final ElementInstance callActivityElementInstance =
         elementInstanceState.getInstance(callingElementInstanceKey);
     final var callActivityInstanceRecord = callActivityElementInstance.getValue();
@@ -76,11 +76,11 @@ public class ElementTreePathBuilder {
     final HashCode hashCode =
         hashFunction.hashString(
             callActivityInstanceRecord.getElementId(), Charset.defaultCharset());
-    return hashCode.asLong();
+    return hashCode.toString();
   }
 
   public record ElementTreePathProperties(
       List<List<Long>> elementInstancePath,
       List<Long> processDefinitionPath,
-      List<Long> callingElementPath) {}
+      List<String> callingElementPath) {}
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.common;
+
+import io.camunda.zeebe.engine.state.immutable.ElementInstanceState;
+import io.camunda.zeebe.engine.state.instance.ElementInstance;
+import java.util.Deque;
+import java.util.LinkedList;
+
+public class ElementTreePathBuilder {
+  private final ElementInstanceState elementInstanceState;
+  private final Long elementInstanceKey;
+  private final ElementTreePathProperties properties;
+
+  public ElementTreePathBuilder(
+      final ElementInstanceState elementInstanceState, final Long elementInstanceKey) {
+    this.elementInstanceState = elementInstanceState;
+    this.elementInstanceKey = elementInstanceKey;
+    properties =
+        new ElementTreePathProperties(new LinkedList<>(), new LinkedList<>(), new LinkedList<>());
+  }
+
+  public ElementTreePathProperties getProperties() {
+    buildElementTreePathProperties(elementInstanceKey);
+    return properties;
+  }
+
+  private void buildElementTreePathProperties(final long elementInstanceKey) {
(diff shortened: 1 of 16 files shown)


lenaschoenburg:
> Since ElementIds are just a limited number of strings unique locally to the deployed model
we could use a hashing function to hash these Ids in the callingElementTreePath. Operate can use the same hashing function to hash all elementIds in a model and build a reverse Map<Hash, ElementId>, and use this map to look up the elementId string using the hash. I decided to use murmur3_128 because it is a fast hashing function with low chance of collision. And I choose an implementation from google guava

Excuse the unprompted review but this caught my attention. Why rely on hashing, _hoping_ that no two element ids ever result in the same hash instead of using a unique identifier? For example the offset into a lexicographically sorted list of all element ids in the model? This would require us to also maintain that state somewhere, much like Operate would have to maintain `Map<Hash, ElementId>` in your proposal, but we completely avoid the chance for collisions.

If that's not feasible, I'd suggest a deploy time validation to check ahead of time that no two element ids result in the same hash. If they do, we could refuse the deployment to prevent inconsistent data in Operate.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private Long getCallActivityIdHash(final long callingElementInstanceKey) {
+    final ElementInstance callActivityElementInstance =
+        elementInstanceState.getInstance(callingElementInstanceKey);
+    final var callActivityInstanceRecord = callActivityElementInstance.getValue();
+
+    final HashCode hashCode =
+        hashFunction.hashString(
+            callActivityInstanceRecord.getElementId(), Charset.defaultCharset());
+    return hashCode.asLong();
```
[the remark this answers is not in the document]
mustafadagher: That's a good pint. 
Originally, I was looking to use [goodFastHash](https://guava.dev/releases/snapshot/api/docs/com/google/common/hash/Hashing.html#goodFastHash(int)) with 64 bits, to provide a balance between performance and collision safety. But I saw in the implementation it uses the `Murmur3_128HashFunction` for anything above 32 bits anyway.

I could have used `toString()` as well. And I could still change it use strings if we see a need.

mustafadagher:
> Excuse the unprompted review but this caught my attention. Why rely on hashing, _hoping_ that no two element ids ever result in the same hash instead of using a unique identifier? For example the offset into a lexicographically sorted list of all element ids in the model? This would require us to also maintain that state somewhere, much like Operate would have to maintain `Map<Hash, ElementId>` in your proposal, but we completely avoid the chance for collisions.
> 
> If that's not feasible, I'd suggest a deploy time validation to check ahead of time that no two element ids result in the same hash. If they do, we could refuse the deployment to prevent inconsistent data in Operate.

@lenaschoenburg, actually thank you for the review! All feedback is more than welcome!

Let's say we can do something like below in the [CallActivityTransformer](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java) (excuse the ugly code though)

``` java
    final List<String> allCallActivityIds =
        process.getFlowElements().stream()
            .filter(ExecutableCallActivity.class::isInstance)
            .map(ExecutableCallActivity.class::cast)
            .map(ca -> BufferUtil.bufferAsString(ca.getId()))
            .sorted()
            .toList();
    int index;
    for (index = 0; index < allCallActivityIds.size(); index++) {
      if (allCallActivityIds.get(index).equals(BufferUtil.bufferAsString(callActivity.getId()))) {
        break;
      }
    }
```
And add the index above in the [ExecutableCallActivity](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/element/ExecutableCallActivity.java) properties, to use later in the tree path.

Such piece of code will have to be done with every transformation, and has to be reversed in operate. Do you think this would be a more efficient method compared to hashing?

I'm curious also if you have simpler, more efficient ways to achieve the same, and I'm more than happy to adopt it.

korthout:
> For example the offset into a lexicographically sorted list of all element ids in the model?

A lexicographically sorted list is interesting as it resolves [the parser dependency we considered a blocker for using an index](https://camunda.slack.com/archives/C077SHZS6R0/p1721835190258389?thread_ts=1721318333.841459&cid=C077SHZS6R0).

>Such piece of code will have to be done with every transformation, and has to be reversed in operate. Do you think this would be a more efficient method compared to hashing?

We just have to do it once per transformation of the entire process, not for each element. I doubt that it's more efficient, but it would avoid any hashing collisions and be simple enough to share between zeebe and operate.

mustafadagher:
> > Excuse the unprompted review but this caught my attention. Why rely on hashing, _hoping_ that no two element ids ever result in the same hash instead of using a unique identifier? For example the offset into a lexicographically sorted list of all element ids in the model? This would require us to also maintain that state somewhere, much like Operate would have to maintain `Map<Hash, ElementId>` in your proposal, but we completely avoid the chance for collisions.
> > If that's not feasible, I'd suggest a deploy time validation to check ahead of time that no two element ids result in the same hash. If they do, we could refuse the deployment to prevent inconsistent data in Operate.
> 
> @lenaschoenburg, actually thank you for the review! All feedback is more than welcome!
> 
> Let's say we can do something like below in the [CallActivityTransformer](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java) (excuse the ugly code though)
> 
> ```java
>     final List<String> allCallActivityIds =
>         process.getFlowElements().stream()
>             .filter(ExecutableCallActivity.class::isInstance)
>             .map(ExecutableCallActivity.class::cast)
>             .map(ca -> BufferUtil.bufferAsString(ca.getId()))
>             .sorted()
>             .toList();
>     final int index = Collections.binarySearch(allCallActivityIds, element.getId());
> ```
> 
> And add the index above in the [ExecutableCallActivity](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/element/ExecutableCallActivity.java) properties, to use later in the tree path.
> 
> Such piece of code will have to be done with every transformation, and has to be reversed in operate. Do you think this would be a more efficient method compared to hashing?
> 
> I'm curious also if you have simpler, more efficient ways to achieve the same, and I'm more than happy to adopt it.

@korthout @lenaschoenburg I actually tried to see if something like the above would work, my tests (`shouldPropagateTreePathPropertiesWithIncident`) fail because in the `transformationContext` we have only information about elements contained in the processes in one single xml resource. 
Even if I tried to loop over all processes, it contains only the one.
``` java
    final List<String> allCallActivityIds =
        processes.stream()
            .flatMap(p -> p.getFlowElements().stream())
            .filter(ExecutableCallActivity.class::isInstance)
            .map(ca -> BufferUtil.bufferAsString(ca.getId()))
            .sorted()
            .toList();
    final int index = Collections.binarySearch(allCallActivityIds, element.getId());
```

We can find sometime to pair on it tomorrow, and see if there's a workable solution.

Just as a side note, I still see the chances of hash collision between callActivityElement Ids in one process extremely low.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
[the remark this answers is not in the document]
mustafadagher: switched to `HashCode.toString()` which converts the hash to hexadecimal string, using the full 128 bits to ensure less likelihood for collision.

mustafadagher:
> @korthout @lenaschoenburg I actually tried to see if something like the above would work, my tests (`shouldPropagateTreePathPropertiesWithIncident`) fail because in the `transformationContext` we have only information about elements contained in the processes in one single xml resource. Even if I tried to loop over all processes, it contains only the one.

That makes sense, since in [DeploymentTransformer](https://github.com/camunda/camunda/blob/d5f79284f690c19692339e1c9ce762ae28e556bd/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java#L107-L110) we iterate over resources in a model one by one and deploy them separately, and in [BpmnResourceTransformer](https://github.com/camunda/camunda/blob/d5f79284f690c19692339e1c9ce762ae28e556bd/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/BpmnResourceTransformer.java#L81) `bpmnTransformer.transformDefinitions(definition)` creates a new context instance for each definition.

So index in a lexicographically sorted list is not doable from there. Not without introducing bigger changes to the transformation logic. Unless there's another way I'm unaware of.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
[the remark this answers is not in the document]
mustafadagher: 💭 Just a side note:
I've asked this [here](https://camunda.slack.com/archives/C077SHZS6R0/p1719924818601759) before, and I don't think we know the answer right now.
The hexadecimal string representing `128 bit` hash will be of `32 char` length. If we (have a way to) know that most CallActivity Ids are shorter than this length, then hashing this way is working against us not for us. If it's the other way around, then that's okay.
In the case of using `long`, size wouldn't have mattered, because long is almost always more memory efficient. yet, compromising on some of the anti-collision capabilities.

The above is a trade-off I can't make my mind about without data to support, but we unfortunately, we don't have this data AFAIK.

lenaschoenburg:
> Just as a side note, I still see the chances of hash collision between callActivityElement Ids in one process extremely low.

If we decide that hashing is a practical solution I think we should do the following:
1. Verify our assumption that collisions are unlikely. For example by taking a 1M auto-generated element ids and counting how many collisions there are.
2. Prevent deployment of models with collisions.
3. Check existing deployments for collisions. But it's not clear what to do if there are any.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilder.java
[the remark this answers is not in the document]
lenaschoenburg: I think one important property of hashing (or any other form of encoding) is that it protects against pathological cases where ids are _extremely_ long. So that's something we need to consider too, not just the average case.

> The hexadecimal string representing 128 bit hash will be of 32 char length
We could also encode it as two longs instead. Or use base64 which only take 22 bytes.

In general I agree though, using a 128 bit hash is probably not very efficient for the common case.

lenaschoenburg:
I'm not convinced that the birthday paradox is a viable estimator for this because the hashing function is not truly random.

korthout:
>Prevent deployment of models with collisions.

This would be necessary, but really strange UX. User would have to change their process just because of this implementation detail. While I see that this could be a pragmatic solution, I'm not sure it's the right one. 

I feel like the simplest solution is to go for the index of the element in the process. This information is available to both Zeebe and Operate, and easy to explain.

mustafadagher:
> I feel like the simplest solution is to go for the index of the element in the process. This information is available to both Zeebe and Operate, and easy to explain. Is there any reason to not take this approach?

please check [this comment](https://github.com/camunda/camunda/pull/20293#issuecomment-2248413919) from above.

Maybe I'm missing the point of something, or maybe this is not how operate actually use it.
But a deployed process could have a call activity that calls another process that is in another deployment resource. let's assume the following example, 

we have 3 deployment resources 


- `root-process.bpmn`:
(start) -> [callActivity(callProcess1)]-> (end)
- `process1.bmpn`
(start) -> [callActivity(callProcess2)]-> (end)
- `process2.bmpn`
- (start) -> [serviceTask]-> (end)

Now when we are transforming `root-process`, we will find only one call activity and will give it index 0. and will set this index in the `ExcutableCallActivity`

Then again we come to transform `process1`, and also do the same: find only one call activity there, assign index 0 to it.

Now, when we come to construct the `calling element path`, we will have [0, 0] if an incident happened in the `serviceTask` in process2

I hope the example is clear.

korthout:
@mustafadagher Process can be deployed individually or together. It doesn't really matter for Call Activity. It will simply lookup and call the latest version of a deployed process. If that process doesn't exist at runtime, an [Incident](https://docs.camunda.io/docs/next/components/concepts/incidents/) is raised at the Call Activity.

So each process is transformed individually. Of course, you can have multiple at once:
- deploy multiple bpmn files at once in a single deployment
- deploy a single bpmn file with multiple processes

We only need to track element id index within the process definition.

mustafadagher:
> @mustafadagher Process can be deployed individually or together. It doesn't really matter for Call Activity. It will simply lookup and call the latest version of a deployed process. If that process doesn't exist at runtime, an [Incident](https://docs.camunda.io/docs/next/components/concepts/incidents/) is raised at the Call Activity.
> 
> So each process is transformed individually. Of course, you can have multiple at once:
> 
> * deploy multiple bpmn files at once in a single deployment
> * deploy a single bpmn file with multiple processes
> 
> We only need to track element id index within the process definition.
> 
> 💭 Note that we're currently working on [deployment binding for Call Activities](https://github.com/camunda/camunda/issues/19384), which will allow Call Activity to only call a process that was deployed alongside the process in which the Call Activity exists. In some cases, this makes version management easier for users.

I've the changes to element index ready, but my feeling was to have [0, 0] as callingElementPath for the use case I mentioned above is wrong, and that the index should consider all the elements in a tree path.

If @sdorokhova is fine with the caveat (from my perspective) above. I can push the changes and get rid of the caching.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/element/ExecutableCallActivity.java
```diff
@@ -15,6 +15,7 @@ public class ExecutableCallActivity extends ExecutableActivity {
 
   private boolean propagateAllChildVariablesEnabled;
   private boolean propagateAllParentVariablesEnabled;
+  private int lexicographicIndex;
```
korthout, line 18: 🔧 Let's add JavaDoc clarifying that this concerns the lexicographic index of call activities within the process it is part of.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java
```diff
@@ -53,4 +57,19 @@ private void transformProcessId(
...
     callActivity.setPropagateAllParentVariablesEnabled(propagateAllParentVariablesEnabled);
   }
+
+  private static void transformLexicographicIndex(
+      final CallActivity element,
+      final List<ExecutableProcess> processes,
+      final ExecutableCallActivity callActivity) {
+    final List<String> allCallActivityIds =
+        processes.stream()
+            .flatMap(p -> p.getFlowElements().stream())
```
korthout, line 67: ❓ This appears to be collecting all elements of **all deployed processes**, but you previously mentioned to me that only a single process is available. Did that concern this place? If so, let's add a comment to this code clarifying that this builds the lexicographic index within the process, not within all processes. 

💭 Alternatively, we could also only map the first entry, rather than flatmap all processes to clarify that there only is one that is relevant. Still this requires a clarifying comment.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java
```diff
@@ -53,4 +57,19 @@ private void transformProcessId(
...
+            .filter(ExecutableCallActivity.class::isInstance)
```
korthout, line 68: ❓ Did you consider calculating the index for all elements in the process rather than only call activities?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java
```diff
@@ -53,4 +57,19 @@ private void transformProcessId(
...
+            .map(ca -> BufferUtil.bufferAsString(ca.getId()))
+            .sorted()
+            .toList();
+    final int index = Collections.binarySearch(allCallActivityIds, element.getId());
```
korthout, line 72: 👍 Nice

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/CallActivityTest.java
```diff
@@ -766,11 +763,9 @@ public void shouldPropagateTreePathPropertiesWithIncident() {
...
     final String callActivity1Id = "callParent";
     final String callActivity2Id = "callChild";
-    final HashFunction hashFunction = Hashing.murmur3_128();
-    final var ca1Hash =
-        hashFunction.hashString(callActivity1Id, Charset.defaultCharset()).toString();
-    final var ca2Hash =
-        hashFunction.hashString(callActivity2Id, Charset.defaultCharset()).toString();
+    // every deployment resource will have indexes starting from zero
+    final var ca1Index = 0;
+    final var ca2Index = 0;
```
korthout, line 768: 🔧 These definitions were unclear to me, until I read the test another time. I suggest moving them closer to the assertion (the only place where these are used). You could also consider combining them into a more descriptive variable, e.g. `FIRST_CALL_ACTIVITY_INDEX = 0`.

~🔧 This highlights that it seems that we don't have any tests verifying the correct behavior of the lexicographic index as part of the calling element path, e.g. where the call activity is not the first one appearing in the process~ EDIT: resolved by the next commit.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilderTest.java
```diff
@@ -100,16 +113,26 @@ public void shouldIncludeProcessDefinitionAndCallingElementTreePaths() {
     final ElementInstance processB =
         elementInstanceState.newInstance(
             102, processBRecord, ProcessInstanceIntent.ELEMENT_ACTIVATING);
+    processState.putProcess(
+        processBRecord.getProcessDefinitionKey(),
+        createChildProcess(
+            processBRecord.getProcessDefinitionKey(), "B", ServiceTaskBuilder::done));
```
korthout, line 119: 💭 This resolves a problem that appears from the main change in this commit. Personally, I'd extract this into another commit to clarify intentions better, both to the reviewer as well as future readers.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/CallActivityTest.java
```diff
@@ -813,6 +813,47 @@ public void shouldPropagateTreePathPropertiesWithIncident() {
...
+  @Test
+  public void shouldPropagateCorrectIndexesInCallingElementPathWhenMultipleProcessesInSameFile() {
+    // given
+    final String rootProcessId = "root-process";
+    // call activities { "call-parent", "call-child" } when sorted wll be {"call-child",
+    // "call-parent"}
+    final var ca1Index = 1;
+    final var ca2Index = 0;
+
+    ENGINE.deployment().withXmlClasspathResource("/processes/callActivity.bpmn").deploy();
```
korthout, line 825: Generally, we prefer using the BPMN model API, as it highlights in the test what the process looks like. Navigation to the file is necessary to understand that the special thing about this test case is the order of the call activities in this process.

🔧 I suggest relying on the BPMN model API and an assertion to ensure the order of the call activities in the BPMN file.
🔧 If you don't want to rely on the model API, I suggest renaming the process file to clarify this special trait.

korthout (approved):
Love the changes @mustafadagher 🚀 

🔧 LGTM, but I have some suggestions you should consider. That should block you from merging this PR though. Please have a look at my comments.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java
mustafadagher: To clarify, I previously meant that we only have processes from one deployment resource. So if we deployed multiple processes in one file, then we have access to all of them. And that's why I'm collecting from all deployed processes here.

Check the new test case added [here](https://github.com/camunda/camunda/pull/20293/commits/31273c61efd61cbf01792acd8afe98ba53b7f95b)

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java
mustafadagher: I could. But we actually only need the indexes for call activities, so I thought this will be more performant, especially that we search the list later.

But also, this made it easier to know what to expect in tests.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/CallActivityTest.java
mustafadagher: In this test, I'm trying to test the behavior when we have multiple process deployed in the same bpmn resource (using collaboration lanes).

I couldn't find a way to build a similar model using the BPMN model API. But if there's a way to do it that I'm unaware of, I'd appreciate any help.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/common/ElementTreePathBuilderTest.java
mustafadagher: I described the intention of this change in the last line of the commit description:
> Also updated ElementTreePathBuilderTes to also populate processState since now it's needed to build callingElementPath

I could have separated the change in tests from the change in the builder, but I think they are relevant together, since the need for the processState was in the tree path builder was introduced in this commit.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java
korthout: 👍 Ah okay. Thanks for clarifying. 

🔧 Let's make this clear in the JavaDoc of the `lexicographicIndex` field. It should also be clear to Operate that this is how it works.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java
korthout: I think as well that it makes more sense, as files could have large numbers of elements. Let's ensure we document the choice somewhere (for example in the commit message).

PR #20295, opened 2024-07-17
feat: add externalFormReference to tasklist indice

nathansandi:
## Description

Add externalFormReference to the tasklist-task indice and setup importer

Not in the scope of this ticket
- Add to the response (will be added on the scope of search user task)
- Data Migration - Will be made together with variable indice (This field is not used by Tasklist FE) and will be necessary to the V2 response 
- This does not impact on V1 old endpoints

## Related issues

closes #https://github.com/camunda/tasklist/issues/4992


marcosgvieira (commented):
@nathansandi how do we know if a form is a Camunda form or "external"?

nathansandi:
> @nathansandi how do we know if a form is a Camunda form or "external"?

Hey @marcosgvieira  from Tasklist OOTB this field wont be read (over FE perspective), only on API side

From the docs:
A custom form reference can specify any custom identifier in the user task using the externalReference attribute of the task's zeebe:formDefinition extension element (see the [XML representation](https://docs.camunda.io/docs/next/components/modeler/bpmn/user-tasks/#camunda-form-linked)). How the identifier is interpreted depends on your implementation. You can use it to associate a custom form, route to a custom application, or a URL to a web page, for example. A custom form reference will not be shown in Tasklist. 

https://docs.camunda.io/docs/next/components/modeler/bpmn/user-tasks/#user-task-forms 


So at the end, over Webapp Tasklist, this field would not impact, and we dont need to worry about it, this only will be available on API response

tasklist/els-schema/src/main/resources/schema/es/create/template/tasklist-task.json
panagiotisgts, line 1: Formatter issue?

panagiotisgts (approved):

tasklist/els-schema/src/main/resources/schema/es/create/template/tasklist-task.json
nathansandi: Defntelly yes ;)

PR #20298, opened 2024-07-17
Add new camunda client properties

### LINKED ISSUE #19487: Add generic properties and support the old ones
Discover all `zeebe` related properties and duplicate them using a more generic naming. Deprecate the old properties, but they still need to be supported. Have a look at how we support old and new properties in the Spring SDK

nicpuppa:
## Description

With the new Camunda Client also properties should be renamed. Older properties `zeebe.client....` are refactored into more general ones `client.zeebe....` in this way other component can add their own properties (eg `client.operate....)`

We should ensure also that users can still use the old properties configuration, but with less priority the the new ones.
If the same property is set in the new format and old format, the new format has more priority, example:

```
client.zeebe.tenantId=default
zeebe.client.tenantId=randomId

client.getTenantId == "default"  ==> true
client.getTenantId == "randomId" ==> false
```

## Related issues

closes #19487

### DIFF
--- a/clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
+++ b/clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
@@ -15,31 +15,43 @@
  */
 package io.camunda.client.impl;
 
+import static io.camunda.client.ClientProperties.APPLY_ENVIRONMENT_VARIABLES_OVERRIDES;
 import static io.camunda.client.ClientProperties.CA_CERTIFICATE_PATH;
+import static io.camunda.client.ClientProperties.DEFAULT_JOB_POLL_INTERVAL;
+import static io.camunda.client.ClientProperties.DEFAULT_JOB_TIMEOUT;
+import static io.camunda.client.ClientProperties.DEFAULT_JOB_WORKER_NAME;
+import static io.camunda.client.ClientProperties.DEFAULT_JOB_WORKER_TENANT_IDS;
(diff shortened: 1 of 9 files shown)


nicpuppa:
Why the reminder is not working 🤔

clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
```diff
@@ -87,7 +99,7 @@ public final class CamundaClientBuilderImpl
       Collections.singletonList(CommandWithTenantStep.DEFAULT_TENANT_IDENTIFIER);
   private int jobWorkerMaxJobsActive = 32;
   private int numJobWorkerExecutionThreads = 1;
-  private String defaultJobWorkerName = DEFAULT_JOB_WORKER_NAME;
+  private String defaultJobWorkerName = DEFAULT_JOB_WORKER_NAME_VAR;
```
ana-vinogradova-camunda, line 102: 💭 Just curious why we decided to rename this variable?

ana-vinogradova-camunda (approved):
LGTM 🚀 
Thank you for the detailed description, it was very clear 🌟

clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
nicpuppa: Because otherwise it will conflict with this import `import static io.camunda.client.ClientProperties.DEFAULT_JOB_WORKER_NAME;`

PR #20316, opened 2024-07-17
Add intermediate throw event `StraightThroughProcessingLoopValidator`

### LINKED ISSUE #20308: Add Intermediate None Event to `StraightThroughProcessingLoopValidator`
**Description**

We should consinder an intermedite none event as a straight through procesing element. We can add it to [the list of elements that we consider straight through processing](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/validation/StraightThroughProcessingLoopValidator.java#L34)

Relates to https://github.com/camunda/camunda/issues/12957

remcowesterhoud:
## Description

A loop of intermediate throw events is one of the way to achieve a straight through processing loop. We should add this element type to the validator so we reject these kinds of loops.

## Related issues

closes #20308

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/validation/StraightThroughProcessingLoopValidator.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/validation/StraightThroughProcessingLoopValidator.java
@@ -42,14 +42,19 @@ public final class StraightThroughProcessingLoopValidator {
           BpmnElementType.END_EVENT,
           BpmnElementType.SUB_PROCESS,
           BpmnElementType.MULTI_INSTANCE_BODY,
-          BpmnElementType.CALL_ACTIVITY);
+          BpmnElementType.CALL_ACTIVITY,
+          BpmnElementType.INTERMEDIATE_THROW_EVENT);
 
   /**
    * Loops must contain any of the rejected element types for it to be considered a loop. This makes
    * sure we don't reject deployments containing a loop of just gateways.
    */
   private static final EnumSet<BpmnElementType> REJECTED_ELEMENT_TYPES =
-      EnumSet.of(BpmnElementType.MANUAL_TASK, BpmnElementType.TASK, BpmnElementType.CALL_ACTIVITY);
+      EnumSet.of(
+          BpmnElementType.MANUAL_TASK,
+          BpmnElementType.TASK,
+          BpmnElementType.CALL_ACTIVITY,
+          BpmnElementType.INTERMEDIATE_THROW_EVENT);
 
   /**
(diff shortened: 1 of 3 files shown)


mustafadagher:
Hi @remcowesterhoud , could you give some context or elaborate where did this requirement come from?
The changes look good to me, but I lack the domain knowledge to judge the need for it.

Thank you!

remcowesterhoud:
> Hi @remcowesterhoud , could you give some context or elaborate where did this requirement come from? The changes look good to me, but I lack the domain knowledge to judge the need for it.
> 
> Thank you!

Sure. We noticed in Zeebe that users sometimes start a process instance which contains a straight-through processing loop. An example of this could be:
image
It is straight-through as all the steps are processed within Zeebe itself. There are no steps where we wait for a Job to be completed, or a message to be send for exmaple.

The problem when this happens is that the process gets stuck in the loop, and Zeebe will keep processing the commands/events that originate from it. This will put a lot of unneccesary load on the cluster.
Generally, these loops are a bug in the users model. This validator checks a deployed process for such a loop and rejects the deployment if it contains any.

Now, when I built this validator I didn't include the intermediate throw event. However, it is a straight-through processed element. So when I was working on an incident earlier today I used it as a reproduction of the problem. Decided it was very easy to deploy such a process which overloads the cluster and that we should probably reject deployments like this as well 😄

mustafadagher:
> > Hi @remcowesterhoud , could you give some context or elaborate where did this requirement come from? The changes look good to me, but I lack the domain knowledge to judge the need for it.
> > Thank you!
> 
> Sure. We noticed in Zeebe that users sometimes start a process instance which contains a straight-through processing loop. An example of this could be: image It is straight-through as all the steps are processed within Zeebe itself. There are no steps where we wait for a Job to be completed, or a message to be send for exmaple.
> 
> The problem when this happens is that the process gets stuck in the loop, and Zeebe will keep processing the commands/events that originate from it. This will put a lot of unneccesary load on the cluster. Generally, these loops are a bug in the users model. This validator checks a deployed process for such a loop and rejects the deployment if it contains any.
> 
> Now, when I built this validator I didn't include the intermediate throw event. However, it is a straight-through processed element. So when I was working on an incident earlier this week I used it as a reproduction of the problem. Decided it was very easy to deploy such a process which overloads the cluster and that we should probably reject deployments like this as well 😄

I see! 😄  Thank you for the explanation 🙇

mustafadagher (approved):
LGTM! 🚀

remcowesterhoud:
@mustafadagher Maybe to add. The most common case we saw where trial clusters where users deployed a model with just undefined tasks to play around with Camunda. A basic example would be this:
image

It's quite a bad experience for users to try out our product. Deploy something they think is right and then have their cluster blow up because their model has a loop 😅

PR #20284, opened 2024-07-16
fix: instrument all on:push workflows to send CI health metrics

cmur2:
## Description

This PR makes all GHA workflows running `on: push` to `main` or other persistent branches integrate with [CI health metrics collection](https://github.com/camunda/camunda/wiki/CI-&-Automation#ci-health-metrics). This works towards the goal of https://github.com/camunda/camunda/issues/18210

The following BigQuery query can be used to check the availability of metrics for this specific PR:

```
SELECT *
FROM `ci-30-162810.prod_ci_analytics.build_status_v2`
WHERE ci_url="https://github.com/camunda/camunda" AND build_ref="refs/pull/20284/merge"
ORDER BY report_time DESC
LIMIT 100
```

Given those results it is working so far:

image

## Related issues

Related to #18210

Issue #18210, opened 2024-05-03
[EPIC] CI: collect baseline stability metrics from all (key) GHA workflows


cmur2:
## Description

Thanks to the Infra team we have [a framework](https://confluence.camunda.com/display/HAN/CI+Analytics) in place which allows collecting central metrics about GitHub Actions CI workflows like build failure rates (over time, per workflow). Those metrics are important to get an overall impression on stability and identify hotspots aka GHA workflows that have increased failure rates.

(Must have) The collected data [needs to be visualized](https://dashboard.int.camunda.com/d/bdmo5l8puugaoc/ci-health-camunda-camunda-monorepo) and can be either checked on regular basis or to investigate specific hypothesis about problematic GHA workflows.

This framework should be used in as many places as possible to get a good coverage for the metrics.

(Must have) GHA workflows which are in the critical path of developer experience and thus are key:

- all workflows marked as GitHub "required status checks" check ✔️
- all workflows running on `merge_queue` triggers ✔️
- all workflows running on `push` to `main`, `stable/*` or `release*` branches ✔️
- all workflows running as part of a 8.6+ release ✔️

(Optional) Nice to have to also instrument all GHA workflows that:

- run on `schedule` triggers
- run on `pull_request` triggers

See also https://github.com/camunda/camunda/issues/17603

## Task Breakdown

Phase 1:

_Focus on starting the data collection to have material to work with:_

- Data collection
  - ✔️ Setup `camunda/camunda` for metric ingestion via [CI Analytics](https://confluence.camunda.com/display/HAN/CI+Analytics)
  - ✔️ Add build status metrics for certain key workflows (e.g. `push` and/or `merge_group` triggered)
    - https://github.com/camunda/camunda/issues/18268
    - https://github.com/camunda/camunda/pull/17642

Phase 2:

_Close the feedback loop by providing a way to visualize and look at the collected data & collect more:_

- Data visualization:
  - ✔️ Chose Grafana over LookerStudio for visualization due to knowledge in some teams
  - ✔️ https://github.com/camunda/team-infrastructure/issues/623
  - ✔️ Create [CI health](https://dashboard.int.camunda.com/d/bdmo5l8puugaoc/ci-health-camunda-camunda-monorepo) dashboard for answering key questions:
    - overall job successes/failures (number + ratio) on `push` and `merge_group` on `main` and other stable branches
    - breakdown job failures by job name to identify "problematic" jobs
- Data collection:
[...]

PR #20340, opened 2024-07-18
fix: instrument release workflow to send CI health metrics

cmur2:
## Description

Continuation of https://github.com/camunda/camunda/pull/20284 to reach the goal of https://github.com/camunda/camunda/issues/18210 which is instrumenting all key workflows (includes the 8.6+ release workflows, no backports planned).

## Related issues

Related to #18210


houssain-barouni (approved):
It would be great if GHA can add, in a single place, post process steps to all the jobs in a given workflow

PR #20348, opened 2024-07-18
refactor: change props name and warnings for deprecated props

nicpuppa:
## Description

- Renamed all the properties, removed zeebe references
- Added warnings if older properties are used

## Related issues

related #19487


ana-vinogradova-camunda (approved):
LGTM 👏

PR #20349, opened 2024-07-18
Env vars renaming

nicpuppa:
## Description

Removed all the Zeebe reference from client env vars configuration. The approach used for the new env vars is the same used for the properties

## Related issues

related #19487


ana-vinogradova-camunda (approved):
LGTM 🚀

PR #20354, opened 2024-07-18
ci: make Renovate handling tasklist.Dockerfile and operate.Dockerfile

houssain-barouni:
## Description

I have noticed that Renovate did not update the alpine version in tasklist.Dockerfile and operate.Dockerfile

## Related issues

closes #

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -157,7 +157,7 @@
     }
   ],
   "dockerfile": {
-    "fileMatch": ["(^|/)([\\w-]+\\.)?[Dd]ockerfile$"],
+    "fileMatch": ["(^|\\/)([\\w-]+\\.)?[Dd]ockerfile$"],
     "ignorePaths": [
       "zeebe/benchmarks/**",
       "clients/go/vendor/**"
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -157,6 +157,7 @@
     }
   ],
   "dockerfile": {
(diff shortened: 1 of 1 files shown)


cmur2 (approved):
After merge we should check if Renovate indeed correctly recognizes the new files (while still having the old as well).

houssain-barouni:
> After merge we should check if Renovate indeed correctly recognizes the new files (while still having the old as well).

Do you have any kind of logs that shows what Renovate filters are matching?

.github/renovate.json
```diff
@@ -157,6 +157,7 @@
     }
   ],
   "dockerfile": {
+    "fileMatch": ["(^|/)([\\w-]+\\.)?[Dd]ockerfile$"],
```
cmur2, line 160: You may need to escape the `/` at position 3 in the string with `\\/`

cmur2:
> Do you have any kind of logs that shows what Renovate filters are matching?

Hi, yes you can find two possibilities in the "Related resources" on https://github.com/camunda/camunda/wiki/CI-&-Automation#renovate

* https://github.com/camunda/camunda/issues/12605 shows an abbreviated (incomplete) list, might be on there, might be not, if not go to:
* https://developer.mend.io/github/camunda/camunda

cmur2:
Seems working:

image

houssain-barouni:
> Seems working:
> 
> image

but it did not create the pull requests yet, maybe we should wait a bit

PR #20391, opened 2024-07-22
Exporters can now expose metrics through Micrometer face.

### LINKED ISSUE #18467: Provide observation context to exporters

I would like to add metrics from my own custom exporter and have them scraped along with the rest of the Zeebe metrics. However, using the default `CollectorRegistry.defaultRegistry` (or the Micrometer equivalent one) fails, as the exporter is loaded in a different class loader. This also implies that Zeebe will always rely on the default registry, which may not be true forever.

Much like the `Context` provided to the exporter on configuration has a `Logger` member, I would like the same for metrics. I think using `Micrometer` (which is to metrics what SLF4J is to logging) would make more sense, but I'm OK with Prometheus :)

You can work around this by loading the `CollectorRegistry` class directly from the system class loader, e.g.:

```java
final Class<?> registryClass =
    ClassLoader.getSystemClassLoader().loadClass(CollectorRegistry.class.getName());
final Field registryField = registryClass.getDeclaredField("defaultRegistry");
final CollectorRegistry registry = (CollectorRegistry) registryField.get(null);
Counter.build().name("test").namespace("test").register(registry);
```

As you can see, this is cumbersome, and ignores all the try-catch and possible errors. It also relies on the `defaultRegistry` global, which makes it harder if we ever more to Micrometer across the board.

EuroLew:
## Description

In the `Context` provided to exporters a `getMeterRegistry` call is now exposed. All meters registered to this meter registry will be exposed on the /actuator/prometheus endpoint.

## Related issues

closes #18467

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
@@ -14,6 +14,7 @@ import io.camunda.zeebe.exporter.api.Exporter;
 import io.camunda.zeebe.util.jar.ExternalJarLoadException;
 import io.camunda.zeebe.util.jar.ExternalJarRepository;
 import io.camunda.zeebe.util.jar.ThreadContextUtil;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -90,7 +91,8 @@ public final class ExporterRepository {
     try {
       final Exporter instance = descriptor.newInstance();
       final ExporterContext context =
-          new ExporterContext(LOG, descriptor.getConfiguration(), NULL_PARTITION_ID);
+          new ExporterContext(
+              LOG, descriptor.getConfiguration(), NULL_PARTITION_ID, new SimpleMeterRegistry());
 
       ThreadContextUtil.runCheckedWithClassLoader(
           () -> instance.configure(context), instance.getClass().getClassLoader());
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
@@ -20,6 +20,7 @@ import io.camunda.zeebe.protocol.impl.record.RecordMetadata;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
 import io.camunda.zeebe.util.jar.ExternalJarClassLoader;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
@@ -131,7 +132,8 @@ final class ExternalExporterContainerTest {
     final var descriptor = runtime.loadExternalExporter(jarFile, EXPORTER_CLASS_NAME);
     final var expectedClassLoader = descriptor.newInstance().getClass().getClassLoader();
     final var container =
-        new ExporterContainer(descriptor, 0, new ExporterInitializationInfo(0, null));
+        new ExporterContainer(
+            descriptor, 0, new ExporterInitializationInfo(0, null), new SimpleMeterRegistry());
 
     // when
     container.close();
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
@@ -151,7 +151,6 @@ final class ExternalExporterContainerTest {
     final var exporterClass = createUnloadedExporter(ExporterWithMetrics.class);
     final var jarFile = exporterClass.toJar(new File(jarDirectory, "exporter.jar"));
     final var descriptor = runtime.loadExternalExporter(jarFile, EXPORTER_CLASS_NAME);
-    final var expectedClassLoader = descriptor.newInstance().getClass().getClassLoader();
 
     final var registry = new SimpleMeterRegistry();
 
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
@@ -145,8 +145,46 @@ final class ExternalExporterContainerTest {
         .isInstanceOf(ExternalJarClassLoader.class);
   }
 
+  @Test
+  void shouldRegisterNewMeterInRegistry(final @TempDir File jarDirectory) throws Exception {
+    // given
+    final var exporterClass = createUnloadedExporter(ExporterWithMetrics.class);
+    final var jarFile = exporterClass.toJar(new File(jarDirectory, "exporter.jar"));
+    final var descriptor = runtime.loadExternalExporter(jarFile, EXPORTER_CLASS_NAME);
+    final var expectedClassLoader = descriptor.newInstance().getClass().getClassLoader();
+
+    final var registry = new SimpleMeterRegistry();
+
+    final var container =
+        new ExporterContainer(descriptor, 0, new ExporterInitializationInfo(0, null), registry);
+
+    // when
+    container.configureExporter();
+
+    // then
+    assertThat(registry.getMeters().stream())
+        .filteredOn(meter -> meter.getId().getName().contains("new.counter"))
+        .hasSize(1);
+  }
+
   private Unloaded<TclExporter> createUnloadedExporter() {
-    return new ByteBuddy().subclass(TclExporter.class).name(EXPORTER_CLASS_NAME).make();
+    return createUnloadedExporter(TclExporter.class);
+  }
+
+  private <T extends Exporter> Unloaded<T> createUnloadedExporter(final Class<T> exporterClass) {
+    return new ByteBuddy().subclass(exporterClass).name(EXPORTER_CLASS_NAME).make();
+  }
+
+  public abstract static class ExporterWithMetrics implements Exporter {
+
+    @Override
+    public void configure(final Context context) throws Exception {
+      Exporter.super.configure(context);
+      context.getMeterRegistry().counter("new.counter");
+    }
+
+    @Override
+    public void export(final Record<?> record) {}
   }
 
   // the class must be visible to the generated exporter for ByteBuddy to delegate method invocation
--- a/zeebe/exporter-test/src/main/java/io/camunda/zeebe/exporter/test/ExporterTestContext.java
+++ b/zeebe/exporter-test/src/main/java/io/camunda/zeebe/exporter/test/ExporterTestContext.java
@@ -9,6 +9,7 @@ package io.camunda.zeebe.exporter.test;
 
 import io.camunda.zeebe.exporter.api.context.Configuration;
 import io.camunda.zeebe.exporter.api.context.Context;
+import io.micrometer.core.instrument.MeterRegistry;
 import java.util.Objects;
 import net.jcip.annotations.NotThreadSafe;
 import org.slf4j.Logger;
@@ -25,6 +26,11 @@ public final class ExporterTestContext implements Context {
 
   private Configuration configuration;
   private RecordFilter recordFilter;
+
+  @Override
+  public MeterRegistry getMeterRegistry() {
+    return null;
+  }
 
   @Override
   public Logger getLogger() {
--- a/zeebe/exporter-test/src/main/java/io/camunda/zeebe/exporter/test/ExporterTestContext.java
+++ b/zeebe/exporter-test/src/main/java/io/camunda/zeebe/exporter/test/ExporterTestContext.java
@@ -10,6 +10,7 @@ package io.camunda.zeebe.exporter.test;
 import io.camunda.zeebe.exporter.api.context.Configuration;
 import io.camunda.zeebe.exporter.api.context.Context;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import java.util.Objects;
 import net.jcip.annotations.NotThreadSafe;
 import org.slf4j.Logger;
@@ -26,10 +27,11 @@ public final class ExporterTestContext implements Context {
 
   private Configuration configuration;
   private RecordFilter recordFilter;
+  private final MeterRegistry meterRegistry = new SimpleMeterRegistry();
 
   @Override
   public MeterRegistry getMeterRegistry() {
-    return null;
+    return meterRegistry;
   }
 
   @Override
--- a/zeebe/qa/integration-tests/pom.xml
+++ b/zeebe/qa/integration-tests/pom.xml
@@ -460,9 +460,23 @@
       <artifactId>zeebe-backup-store-azure</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.hawkular.agent</groupId>
+      <artifactId>prometheus-scraper</artifactId>
+      <scope>test</scope>
+      <version>0.23.0.Final</version>
+    </dependency>
+
+
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-core</artifactId>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>io.prometheus</groupId>
-      <artifactId>prometheus-metrics-exporter-common</artifactId>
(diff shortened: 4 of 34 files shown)


npepinpe:
Couple things to watch out for:

- Exporters can be instantiated multiple times, but likely they will use the same underlying Prometheus collector. That means that the metrics may end up being registered multiple times by the user - is this achievable? If not, what could be the UX here to make it simple for users? 
  - Here I'm ignoring having different people use the same metric name for different things - that's possible, but not something we can do much about. At best, we should ensure that when this happens, users know what to do about it (whether through proper logging or documentation).
- Any reason to create a new meter registry per exporter context? We lose the caching of the metrics every time the exporter is re-instantiated. Could we not simply use a global meter registry? I would propose having the dist module inject it into the broker, so we can just use the same registry used by the application in general.
  - We would need to document for exporter authors that they should name space their metrics to avoid collision, or at least how to detect such things.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
```diff
@@ -90,7 +91,8 @@ private void validate(final ExporterDescriptor descriptor) throws ExporterLoadEx
     try {
       final Exporter instance = descriptor.newInstance();
       final ExporterContext context =
-          new ExporterContext(LOG, descriptor.getConfiguration(), NULL_PARTITION_ID);
+          new ExporterContext(
+              LOG, descriptor.getConfiguration(), NULL_PARTITION_ID, new SimpleMeterRegistry());
```
npepinpe, line 95: :thought_balloon: Remember to update the documentation [here](https://docs.camunda.io/docs/self-managed/concepts/exporters/#loading) after to add a note that the meter registry you get during the loading/validation phase is just an in-memory one, and will be discarded after.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
npepinpe, line 1: :question: Should we pair on this? It would be good to see that it works with an externally loaded exporter to ensure there's no problems with the class path or anything like that.

zeebe/exporter-test/src/main/java/io/camunda/zeebe/exporter/test/ExporterTestContext.java
```diff
@@ -26,6 +27,11 @@ public final class ExporterTestContext implements Context {
   private Configuration configuration;
   private RecordFilter recordFilter;
 
+  @Override
+  public MeterRegistry getMeterRegistry() {
+    return null;
```
npepinpe, line 32: :x: Please return a `SimpleMeterRegistry`. It should be a field in the context so it's always the same returned for this particular context, not one we create on the fly here.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsIT.java
npepinpe, line 1: :x: I would only include the dependency to the Prometheus parser if we do anything more with the metric, like:

- Increment the counter and verify its value
- Check the metric type is the correct one
- Check its properties are well set (e.g. help, type, name, namespace, etc.)

If we just check for existence, then I don't think we need the dependency, and we can just look for the text in the dump and use a simple regular expression to check for it, since the format is basically:

```
metric_name [
  "{" label_name "=" `"` label_value `"` { "," label_name "=" `"` label_value `"` } [ "," ] "}"
] value [ timestamp ]
```

Meaning, for example:

```
http_requests_total{method="post",code="200"} 1027 1395066363000
```

Which as a regex would be:

```
^([a-zA-Z_:][a-zA-Z0-9_:]*)(\{.+?\})?\s+([a-zA-Z-+\.\d]+)(\s+([-\d]+))?$
```

Where match group:
1. Metric name
2. Labels (unparsed, would make it more complex to parse them)
3. Value
4. Timestamp + separator
5. Timestamp/Epoch

> [!Note]
> This is ignoring the comments with the HELP and TYPE of the metric encoded.

:question: Is there any value to asserting more properties than just the existence of the metric? If so, keep the dependency and add the assertions. If not, we drop it and just do simple text parsing to check for existence. Third option is you're not sure, we should discuss this :slightly_smiling_face:

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsTestExporter.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package io.camunda.zeebe.it.exporter;
+
+import io.camunda.zeebe.exporter.api.Exporter;
+import io.camunda.zeebe.exporter.api.context.Context;
+import io.camunda.zeebe.exporter.api.context.Controller;
+import io.camunda.zeebe.protocol.record.Record;
+import io.micrometer.core.instrument.Counter;
+
+//  do a http get curl for now to ensure that it is there, but change it to some code access
```
npepinpe, line 16: :question: I don't understand the comment. Is this a left over from development?

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/actuator/PrometheusActuator.java
```diff
@@ -0,0 +1,40 @@
...
+import io.camunda.zeebe.qa.util.cluster.TestStandaloneBroker;
+
+@SuppressWarnings({"unused", "UnusedReturnValue"})
+public interface PrometheusActuator {
+
+  static PrometheusActuator of(final TestStandaloneBroker node) {
+    return of(node.actuatorUri("prometheus").toString());
+  }
+
+  static io.camunda.zeebe.qa.util.actuator.PrometheusActuator of(final String endpoint) {
```
npepinpe, line 27: :x: Not sure why this is a fully qualified name. Can't we shorten it to just `PrometheusActuator`?

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/actuator/PrometheusActuator.java
```diff
@@ -0,0 +1,40 @@
...
+    final var target =
+        new HardCodedTarget<>(io.camunda.zeebe.qa.util.actuator.PrometheusActuator.class, endpoint);
```
npepinpe, line 29: :question: Same as above - is this some IntelliJ thing? :smile:

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/cluster/TestStandaloneBroker.java
```diff
@@ -70,6 +70,8 @@ protected SpringApplicationBuilder createSpringBuilder() {
     // we need to hook in at the last minute and set the property as it won't resolve from the
     // config bean
     withProperty("zeebe.broker.gateway.enable", config.getGateway().isEnable());
+    withProperty("management.endpoint.prometheus.enabled", true);
+    withProperty("management.prometheus.metrics.export.enabled", true);
```
npepinpe, line 74:
```suggestion
    withProperty("zeebe.broker.gateway.enable", config.getGateway().isEnable()).
      withProperty("management.endpoint.prometheus.enabled", true).
      withProperty("management.prometheus.metrics.export.enabled", true);
```

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/cluster/TestStandaloneBroker.java
```diff
@@ -70,6 +70,8 @@ protected SpringApplicationBuilder createSpringBuilder() {
     // we need to hook in at the last minute and set the property as it won't resolve from the
     // config bean
     withProperty("zeebe.broker.gateway.enable", config.getGateway().isEnable());
+    withProperty("management.endpoint.prometheus.enabled", true);
```
npepinpe, line 73: :question: Is this not enabled by default in our `application.properties` file? Do you know why it was necessary to add them here?

npepinpe (changes requested):

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsTestExporter.java
EuroLew: oops

deepthidevaki:
Since Nicolas has already reviewed this, I will remove my assignment. Please rerequest if you need my reviews after making the changes.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
EuroLew: Does the user need to know this? My understanding is if they add code in the `configure` block registering meters it will add it to the simple register, discard it, and then the second time round register it with the actual `PrometheusMeterRegistry` as intended.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExternalExporterContainerTest.java
EuroLew: Help with testing externally loaded exporters would be great, I'll set up a meeting

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/cluster/TestStandaloneBroker.java
EuroLew: it is I remember it not working without this but maybe it was due to a different test that used its own config and I assumed that was the base test config

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsIT.java
EuroLew: I guess I'm just not sure how much we would gain from not using the parsing library, It would help any tests we would do in the future which would require the actual values of the meters.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
EuroLew: https://github.com/camunda/camunda-docs/pull/4082 documentation changes are here

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsIT.java
```diff
@@ -0,0 +1,45 @@
...
+
+@ZeebeIntegration
+public class ExporterMetricsIT {
+
+  @TestZeebe
+  private final TestStandaloneBroker zeebe =
+      new TestStandaloneBroker()
+          .withExporter("foo", cfg -> cfg.setClassName(ExporterMetricsTestExporter.class.getName()))
+          .withBean(
+              "collectorRegistry", CollectorRegistry.defaultRegistry, CollectorRegistry.class);
```
npepinpe, line 30: :x: Ah, do we need to inject the registry here because the broker is not using the registry coming from Spring? Can you add a TODO to remove this here when we do the follow up of injecting the `MeterRegistry` from Spring into the broker after?

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsTestExporter.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public void open(final Controller controller) {
+    Exporter.super.open(controller);
+  }
+
+  @Override
+  public void close() {
+    Exporter.super.close();
+  }
```
npepinpe, line 36: :wrench: Can be removed since they don't override anything.

npepinpe (approved):
:+1: Please address comments before merging. I think we will also want to do the follow up now and inject the meter registry from the dist module into the broker, that way we can get rid of the todo in the QA test.

:x: Please fix the Java checks (seems like there's an unused dependency declared in one of the POMs).

EuroLew:
> 👍 Please address comments before merging. I think we will also want to do the follow up now and inject the meter registry from the dist module into the broker, that way we can get rid of the todo in the QA test.
> 
> ❌ Please fix the Java checks (seems like there's an unused dependency declared in one of the POMs).

The meter registry is now injected through the system context.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsIT.java
```diff
@@ -0,0 +1,43 @@
...
+      new TestStandaloneBroker()
+          .withExporter(
+              "foo", cfg -> cfg.setClassName(ExporterMetricsTestExporter.class.getName()));
+
+  @Test
+  public void shouldAddMeterToExporterMetrics() throws IOException {
+    // given
+    final var actuator = PrometheusActuator.of(zeebe);
+    final var metricsPlainText =
+        IOUtils.toString(actuator.metrics().body().asInputStream(), StandardCharsets.UTF_8);
```
npepinpe, line 35: :x: This isn't required anymore with newer Java versions. You can just do `new String(actuator.metrics().body().asInputStream().readAllBytes(), StandardCharsets.UTF_8);`. 

That said, Feign is pretty smart - did you try just returning `String` in the `PrometheusActuator` instead of `Response`? I have a feeling it might just do the whole thing for you.

I would try the following, assuming Feign can convert the body by itself to `String`:

```java
@Test
  public void shouldAddMeterToExporterMetrics() {
    // given
    final var actuator = PrometheusActuator.of(zeebe);

    // when
    final var metricsPlainText = actuator.metrics();

    // then
    Assertions.assertThat(metricsPlainText.lines())
        .filteredOn(line -> line.startsWith(ExporterMetricsTestExporter.REGISTERED_COUNTER))
        .hasSize(1);
  }
```

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsIT.java
```diff
@@ -0,0 +1,43 @@
...
+
+    final var metricLines = Arrays.asList(metricsPlainText.split(System.lineSeparator()));
```
npepinpe, line 37: :wrench: Since Java 11, you can just do `"foo\n".lines()` to read a String line by line.

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/actuator/PrometheusActuator.java
```diff
@@ -0,0 +1,39 @@
...
+
+  static PrometheusActuator of(final TestStandaloneBroker node) {
+    return of(node.actuatorUri("prometheus").toString());
+  }
+
+  static PrometheusActuator of(final String endpoint) {
+    final var target = new HardCodedTarget<>(PrometheusActuator.class, endpoint);
+    return Feign.builder()
+        .encoder(new JacksonEncoder())
+        .decoder(new JacksonDecoder())
```
npepinpe, line 31: :x: No need, we don't deal with JSON for encoding/decoding anything.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/exporter/ExporterMetricsIT.java
```diff
@@ -0,0 +1,43 @@
...
+public class ExporterMetricsIT {
+
+  @TestZeebe
+  private final TestStandaloneBroker zeebe =
+      new TestStandaloneBroker()
+          .withExporter(
+              "foo", cfg -> cfg.setClassName(ExporterMetricsTestExporter.class.getName()));
+
+  @Test
+  public void shouldAddMeterToExporterMetrics() throws IOException {
```
npepinpe, line 31: :x: With JUnit 5, the visibility level of classes and methods doesn't need to be public, can just be package private. Seems silly for tests, but I think as a rule of thumb, it's good to always go with the lowest visibility - you never know what someone might do with code you wrote :smile:

zeebe/qa/integration-tests/pom.xml
```diff
@@ -461,6 +461,24 @@
...
+      <groupId>io.prometheus</groupId>
+      <artifactId>simpleclient</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <scope>test</scope>
+    </dependency>
```
npepinpe, line 480: :x: I think these might not be required in the end, if you apply suggestions/changes.

npepinpe (approved):
:+1: Looks good!

Pre-approving because I don't think we need another review, but don't hesitate to ask for one if you're unsure. Do address all comments before merging though :slightly_smiling_face:

PR #20401, opened 2024-07-22
feat: add privateKeyPassword

jonathanlukas:
## Description

This PR is currently a draft for adding support for encrypted private key files.

## Related issues

Product hub issue: https://github.com/camunda/product-hub/issues/2421
Public issue: https://github.com/camunda/issues/issues/833

closes #

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
@@ -93,6 +93,7 @@ import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.agrona.CloseHelper;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


megglos:
context see https://camunda.slack.com/archives/CSQ2E3BT4/p1721899840168979?thread_ts=1721884482.711129&cid=CSQ2E3BT4

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -500,6 +501,15 @@ private CompletableFuture<Void> loadServerSslContext() {
     }
   }
 
+  private SslContextBuilder sslContextBuilder() {
```
EuroLew, line 504: This can be refactored to use the default java `.isBlank`, also could be neater to do a to only have one `SslContextBuilder.forServer` call by refactoring the private key password to either be some non blank string or null as https://netty.io/4.1/api/io/netty/handler/ssl/SslContextBuilder.html#forServer-java.io.File-java.io.File-java.lang.String- states that a null value for the password signifies it is not password protected

EuroLew:
This will also need to be documented both in the `broker.yaml.template` and the `broker.standalone.yaml.template`, In addition a `camunda-docs` PR will need to be raised denoting this configuration change.

npepinpe:
As this is quite out of date, and we ended up adding support for key stores, should we close this? It means without there is no password support for PKCS8. @jonathanlukas - considering this is still useful for some users, would you mind cleaning up the PR and applying the suggested changes? Otherwise we will simply close it and drop it.

jonathanlukas:
Hi @npepinpe if the selected solution allows to mount encryted private keys, I am good :) 

I will close it myself, thank you for picking this up!

jonathanlukas:
@npepinpe could you please add a reference to the new PR?

npepinpe:
https://github.com/camunda/camunda/pull/21220
https://github.com/camunda/camunda/pull/21375

PR #20414, opened 2024-07-22
test: move to CamundaClient

tmetzke:
## Description

Replace the usage of the ZeebeClient by the CamundaClient in integration, update, broker, and gateway tests.

## Related issues

related to #19479


zeebe/gateway-grpc/src/test/java/io/camunda/zeebe/gateway/interceptors/InterceptorIT.java
```diff
@@ -118,7 +118,7 @@ void shouldAbortDeploymentCalls() {
 
     // when
     gateway.start().join();
-    try (final var client = createZeebeClient()) {
```
nicpuppa, line 121: 🔧 ❌  The CI is failing due to a wrong import for `DeploymentEvent`

tmetzke: 🙈

tmetzke:
@nicpuppa, I think we have to employ bigger changes here if we want to adjust to the Camunda Client holistically, right? We use the Zeebe test containers library here that returns a `ZeebeClient` for the `ZeebeCluster`s that we use in integration tests.

tmetzke:
> > @nicpuppa, I think we have to employ bigger changes here if we want to adjust to the Camunda Client holistically, right? We use the Zeebe test containers library here that returns a `ZeebeClient` for the `ZeebeCluster`s that we use in integration tests.
> 
> Are this changes urgent now ? I think we can leave those tests running with zeebe client and think about a better approach without rushing now. We should remove the zeebe-test-container usage in our codebase in favor of the new CPT project. We should sync with @saig0 on this

I actually totally agree, yes 😄 Just wanted your second opinion on it and it looks like we're on the same page. I'll then revert the integration tests for now and we need to rework their setup before changing this aspect 👍

nicpuppa:
@tmetzke There are still some errors with imports, but my question is, do we need to change those tests now ? I think we can wait until we have a more clear picture on how to remove the `zeebe-test-container` dependency. Than we can switch to `CamundaClient` in all zeebe tests

tmetzke:
Yep, I wanted to see if that works now or not. Apparently, it doesn't, I'll just close this here for now 😄

PR #20434, opened 2024-07-23
refactor: revert unused identity code

Ben-Sheppard:
## Description
This PR is intended to achieve two goals:
1. Remove the existing Identity code that will not be necessary given our architectural direction change
2. As a part of 1, remove the Spring-data dependencies introduced by Identity therefore removing the H2 configuration that is causing memory issues


romansmirnov (approved):
Tested it locally, works without issues. LGTM!

PR #17581, opened 2024-04-18
feat: implementing bounded cache 

mustafadagher:
## Description
- implements bounded cache for form cache
- implements bounded cache for process caches
- allows caches capacity to be configured through broker configs
## Related issues

closes #15533

Issue #15533, opened 2023-12-08
Limit the number of element in the process caches


nicpuppa:
**Describe the bug**

When deploying a process, new models are added to the deployment cache, which is kept in memory. The deployment cache is right now unbounded which causes the broker to fail after a certain time. Recently we faced an [incident](https://camunda.slack.com/archives/C068P5KSSGP)  where a customer deployed a lot of processes and the broker was stuck in a restart loop

**Expected behaviour**

Having a bounded cache, so that the cache size does not grow indefinitely and cause the broker to crash. We should implement a similar approach to [Decisions Cache](https://github.com/camunda/zeebe/blob/6cbc4ff2500e751acf6a9b8a12e91740e6ab9bfe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbDecisionState.java#L96), by keeping a certain amount of data inside the cache, and evicting the less used elements when we reach the max size

Zelldon:
As already stated above by Nic and written by me in the incident channel:

> When deploying new models are added to the deployment cache, which is kept in memory. The deployment cache is right now unbounded which causes the broker to fail after a certain time (when enough models are deployed). The broker will start with a clean cache. Process models are then added if PI is executed or new deployment is created.

To prove this I also checked the heap.dumps which we have found in the related broker nodes. You can find one [here](https://drive.google.com/drive/folders/1ohaIrSB7fUwNUtaSDsgdOPJ6Sn-cAz_R) 

----

For example we have `processesByProcessIdAndVersion = java.util.HashMap#21089 : 3,690 elements`

![maps](https://github.com/camunda/zeebe/assets/2758593/24447f1a-a7ff-4b01-b6bb-74a976e7b9d8)

This is a two level map, first the processID, then the version and then the actual value (the process)

Already at the first entry we can see that the processIds are quite large. 
![processIds](https://github.com/camunda/zeebe/assets/2758593/78b8d19f-9329-4b16-a837-ab05d83d9d7f)


When checking the actual value we see that one of 8 has a resource with size ~40k bytes.
![deployments](https://github.com/camunda/zeebe/assets/2758593/dcddfa3b-ce9d-4ae2-9ea4-d9e7eb8cc871)

Now making the math (estimation):

If every processId has around 8 versions and each is around 40K then the size per process is: 40 * 8 = 320k (+ processId 20k) ~340k

Now 3,7k elements -> 3,7k * 340 = 1.258.000.000 bytes (1.2 gig)

[...]

PR #20452, opened 2024-07-23
7369

SpiderRus:
## Related issues

Relates to #15533 #17581


remcowesterhoud:
Hi @SpiderRus! Thanks for you contribution! Could you explain a bit what you've done and why you've done it? Is there any issue this relates to?

remcowesterhoud:
This seems to relate to #15533 which was resolved in #17581. I'll ceck why this fix wasn't backported. Backporting has preference over creating a new PR with the same fix.

PR #20457, opened 2024-07-23
refactor: scan only relevant packages with Spring

romansmirnov:
## Description

When enabling trace logging for the logger `org.springframework`, the logs show that Spring scans the packages multiple times:
1. When using the annotation `@ConfigurationPropertiesScan,` Spring will scan the defined packages to look up configuration properties. Typically, the same packages are defined with this annotation as with the annotation `@ComponentScan`.
2. Using the annotation `@EnableAutoConfiguration` all over the place. In that case, the `JacksonAutoConfiguration` will trigger another scanning of the defined base packages for all loaded classes with the annotation `@EnableAutoConfiguration` to look up classes annotated with `@JsonMixin` - a feature that we don't use of Jackson.
3. The package name `io.camunda.commons` is used in different modules, so when using `@ComponentScan` and setting the base package to `io.camunda.commons`, it will scan a few unexpected modules unnecessarily.

To resolve the scanning of packages multiple times, the following changes have been applied:
1. Remove the use of `@ConfigurationPropertierScan` and instead use `@EnableConfigurationProperties`. That way, only the required configuration properties are loaded.
2. Reduce the usage of `@EnableAutoConfiguration` to a single/common place. That way only the package `io.camunda.application.commons` is scanned for classes annotated with `@JsonMixin` (please note it is not possible to disable that scanning).
3. Repackage the `io.camunda.commons` to `io.camunda.application.commons`. That way, Spring does not unnecessarily scan modules that contain the package name `io.camunda.commons`.

## Related issues

closes #


houssain-barouni (approved):
LGTM! Great work!

PR #20465, opened 2024-07-23
refactor: refactor SearchQuerySort mapper

houssain-barouni:
## Description

Make `SearchQueryRequestMapper.toSearchQuerySort` generic to support object types other than ProcessInstance.
If we need to handle a new object type `MyObject` that supports search and sorting.
1. We need just to define `applyMyObjectSortField` that maps the supported sort fields (later, the unsupported sorted fields can be returned in a `ProblemDetail`):
```java
  private static void applyMyObjectSortField(
      final String field, final MyObjectSort.Builder builder) {
    switch (field) {
      case "foo" -> builder.foo();
      case "bar" -> builder.bar();
      default -> throw new IllegalArgumentException("Unknown sortBy: " + field);
    }
  }
```
2. Then, the `sorting` object can be obtained:
```java
    final var sorting =
        toSearchQuerySort(
                request.getSort(),
                SortOptionBuilders::myObject,
                SearchQueryRequestMapper::applyMyObjectSortField)
            .get();
```

In this pull request, I also made `field` and `order` fields mandatory under `SearchQuerySortRequest` in the rest-api.yaml. I need to check if this can be validated out-of-the-box without the need to manual nullity checks.

## Related issues

closes #

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1482,9 +1482,9 @@ components:
           type: string
         order:
           type: string
+          default: asc
       required:
         - field
-        - order
     SearchQueryResponse:
       type: object
       properties:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1482,6 +1482,9 @@ components:
           type: string
         order:
           type: string
+      required:
+        - field
+        - order
     SearchQueryResponse:
       type: object
       properties:
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
@@ -23,9 +23,12 @@ import io.camunda.zeebe.gateway.protocol.rest.SearchQueryPageRequest;
 import io.camunda.zeebe.gateway.protocol.rest.SearchQuerySortRequest;
 import io.camunda.zeebe.gateway.protocol.rest.VariableValueFilterRequest;
 import io.camunda.zeebe.util.Either;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ProblemDetail;
 
 public final class SearchQueryRequestMapper {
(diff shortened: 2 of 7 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1482,6 +1482,9 @@ components:
           type: string
         order:
           type: string
+      required:
+        - field
+        - order
```
tmetzke, line 1487: 💭 Just to check in: do we want to allow sorting with a default behavior, i.e. allow to just define the sort fields and by default sort ASC if no order is provided?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -104,32 +113,22 @@ public static Either<ProblemDetail, SearchQueryPage> toSearchQueryPage(
...
-        if ("processInstanceKey".equals(field)) {
-          builder.processInstanceKey();
-        } else if ("startDate".equals(field)) {
-          builder.startDate();
-        } else if ("endDate".equals(field)) {
-          builder.endDate();
-        } else {
-          throw new RuntimeException("unkown sortBy " + field);
+        if (sort.getField() == null) {
+          throw new IllegalArgumentException("Sort field must not be null");
```
tmetzke, line 125: ❌ We should collect validation errors like this instead of throwing an exception here. The collected errors can then be returned to the user. @nathansandi worked through this over in this PR already, maybe the two of you can align here. Let's incorporate this here so @nathansandi can consume this in his PR then.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -104,32 +113,22 @@ public static Either<ProblemDetail, SearchQueryPage> toSearchQueryPage(
...
-
-        if ("processInstanceKey".equals(field)) {
-          builder.processInstanceKey();
-        } else if ("startDate".equals(field)) {
-          builder.startDate();
-        } else if ("endDate".equals(field)) {
-          builder.endDate();
-        } else {
-          throw new RuntimeException("unkown sortBy " + field);
+        if (sort.getField() == null) {
```
tmetzke, line 124: 🔧 If we add such validation constraints, we should add unit tests in the REST gateway as well. Since we don't have any tests for the `ProcessInstanceController` yet, now might be a good time to add some.

tmetzke (changes requested):
Thanks, @houssain-barouni, I was about to look into a generalization like this as well, super happy you tackled this already! 👍 

There are some minor details to consider and I think we should add testing for the search endpoint to ensure this maps to the query API as expected, especially if we keep adding validation constraints now.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
houssain-barouni: So far, the current implementation is expecting the `order` to be not null.
I think it makes sense to default to `asc` for better DevX
I will update the api spec 👍

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
houssain-barouni: Agree, let me add these tests

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
tmetzke: Having test cases will help us validate this easily. I think you can mark fields as required in OpenAPI, generate the model, and have Spring throw an exception if the input is missing. 

It might be simpler to just incorporate validation errors here correctly in this PR, so @nathansandi's PR can be simplified a lot (just reusing your method for user task search accordingly).

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
houssain-barouni: done!

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
houssain-barouni: validationErrors are now collected

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
houssain-barouni: I created `ProcessInstanceControllerTest`

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1482,9 +1482,9 @@ components:
           type: string
         order:
           type: string
+          default: asc
```
tmetzke, line 1485: 🔧 Oh, I think you can even go for an enum and validation is done for you: https://swagger.io/docs/specification/data-models/enums/

tmetzke (approved):
👍 This is great, @houssain-barouni ⭐ 
👍 I really like the new generalization and validation error collection.

💭 I left an idea for the YAML, feel free to try that out. However, it might make the UX for users worse, regarding error messages. The enum would be validated before anything else and it might be harder to locate the error for larger requests. I'll leave it to you to decide how to progress here. Not blocking this PR 👍

houssain-barouni:
> 👍 This is great, @houssain-barouni ⭐ 👍 I really like the new generalization and validation error collection.
> 
> 💭 I left an idea for the YAML, feel free to try that out. However, it might make the UX for users worse, regarding error messages. The enum would be validated before anything else and it might be harder to locate the error for larger requests. I'll leave it to you to decide how to progress here. Not blocking this PR 👍

Thanks for the review.
I believe switching the sortOrder to an enum is beyond the initial scope of this pull request. I've created a new issue to address this change properly. https://github.com/camunda/camunda/issues/20512

PR #20470, opened 2024-07-24
ci: fix release dry-run CI

houssain-barouni:
## Description

release dry-run CI after the upgrade of alpine version in dockerfiles
- in this pull request, I update the docker verification script:
  - added BASE_IMAGE placeholder in the golden file 
  - extract the base image from the dockerfile

I also updated the CI so that it can be manually run on workflow_dispatch from any other branch and not only `main`
Slack notification is skipped in case the CI is triggered with workflow_dispatch to avoid the noise in the channel

release dry-run on this branch: https://github.com/camunda/camunda/actions/runs/10075985777

## Related issues

closes #

### DIFF
--- a/.ci/docker/test/verify.sh
+++ b/.ci/docker/test/verify.sh
@@ -119,7 +119,7 @@ expectedLabels=$(
     --arg REVISION "${REVISION}" \
     --arg DATE "${DATE}" \
     --arg DIGEST "${DIGEST}" \
-    --arg BASE_IMAGE "${BASE_IMAGE}" \
+    --arg BASE_IMAGE "docker.io/library/${BASE_IMAGE}" \
     "$(cat "${labelsGoldenFile}")"
 )
 
--- a/.ci/docker/test/verify.sh
+++ b/.ci/docker/test/verify.sh
@@ -90,6 +90,16 @@ else
     exit 1
 fi
 
+BASE_IMAGE_REGEX='ARG BASE_IMAGE="([^"]+)"'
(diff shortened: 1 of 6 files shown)


.github/workflows/camunda-platform-release-main-dry-run.yml
```diff
@@ -1,13 +1,19 @@
 name: Camunda Platform Release Dry Run from main
 on:
-  workflow_dispatch: { }
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Specifies the branch, for which the Release Dry Run should be executed'
+        default: 'main'
+        required: false
+        type: string
```
lenaschoenburg, line 9: :+1: Nice addition for future testing

.ci/docker/test/verify.sh
```diff
@@ -90,6 +90,16 @@ else
     exit 1
 fi
 
+BASE_IMAGE_REGEX='ARG BASE_IMAGE="([^"]+)"'
+DOCKERFILE=$(<"${BASH_SOURCE%/*}/../../../${DOCKERFILENAME}")
```
lenaschoenburg, line 94: :upside_down_face: We could just define this once (see line 84). Would then also be a good opportunity to gracefully handle the case that this file doesn't exist.

lenaschoenburg (approved):
Thanks! :rocket:

.ci/docker/test/verify.sh
houssain-barouni: right, came with copy pasting 😄

PR #20476, opened 2024-07-24
feat: Search user task - create new fields and refactor on Service layer

nathansandi:
## Description

- include externalFormReference processDefinitionVersion and customHeaders on Camunda Service\
- Refactor processDefinitionKey to Long
- Rename userTaskState to only State

## Related issues

closes #


service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
```diff
@@ -94,8 +93,8 @@ private SearchQuery getProcessInstanceKeysQuery(final List<Long> processInstance
     return longTerms("processInstanceId", processInstanceKeys);
   }
 
-  private SearchQuery getProcessDefinitionKeyQuery(final List<String> processDefinitionIds) {
-    return stringTerms("processDefinitionId", processDefinitionIds);
+  private SearchQuery getProcessDefinitionKeyQuery(final List<Long> processDefinitionIds) {
+    return longTerms("processDefinitionId", processDefinitionIds);
```
tmetzke, line 97: ❓ Just to understand this: on the Tasklist index, process definition key (the Long identifier internally generated by Zeebe) is named `processDefinitionId`, that is why we're mapping it here accordingly?

service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
```diff
@@ -72,13 +72,12 @@ public SearchQuery toSearchQuery(final UserTaskFilter filter) {
...
-    final var canceled = filter.canceled();
+    final var created = filter.created();
 
-    if (completed || canceled) {
-      return Arrays.asList("tasklist-task-8.5.0_alias");
-    } else {
+    if (created) {
       return Arrays.asList("tasklist-task-8.5.0_");
+    } else {
+      return Arrays.asList("tasklist-task-8.5.0_alias");
```
tmetzke, line 80: ❓ Understanding question: why do we have version numbers on those indices? Do we have to update those at some point or anything, e.g. before/after releasing version 8.6? 

💭 If not, can we remove the versions at least or go even further and establish an enum with TASKLIST-TASK, OPERATE-XYZ, or even just TASK, VARIABLE, and so on that is evaluated accordingly by the Query API later on, depending on where it goes to (ES, OS, maybe something else entirely at some point)? This looks quite error-prone and is hard to understand. It doesn't have to be in the scope of this PR here, it can be a follow-up issue (if it doesn't exist already).

tmetzke (approved):
Looks good to me, maybe @houssain-barouni can have a look as well from the HTO perspective 👍 
❓ I added some understanding questions and a suggestion regarding the referenced indices. Not blocking this PR though.

service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
nathansandi: Yes, you are right over here 
image

service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
nathansandi: We have this discussion with @ThorbenLindhauer, We cant rename them, because it would imply on data migration, as we cant just rename them over ES/OS (sadly). 

We can have another round of discussion up to the future, as other indices have similar standard .

tmetzke: I am fine with the name in general, or I don't care about it much, to be honest. But I would like to NOT have them referenced like this in the service layer code. This is leaking information from lower levels into the service layer. This could be an enum, used in the service layer and later resolved in the query API implementation accordingly.

nathansandi: I agree with your point. I will create a follow up issue, because there are other services that use this (Process Instance, Variable) that it is necessary to do this.

service/src/main/java/io/camunda/service/search/filter/UserTaskFilter.java
```diff
@@ -21,9 +21,9 @@ public final record UserTaskFilter(
     List<String> userTaskDefinitionIds,
     List<String> processNames,
     List<String> assignees,
-    List<String> userTaskState,
+    List<String> state,
```
houssain-barouni, line 24: `states` to be consistent with other fields naming

houssain-barouni, line 24: Maybe not related to this pull request, but I was wondering,  in `UserTaskFilter`, why we do have at the same time  `List<String> state`  and also 
```
    boolean created,
    boolean completed,
    boolean canceled,
```
isn't redundant ?

service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
houssain-barouni: I think we should have place where to the spec of the indices and have a mapping Index-> Index name or alias, like we already have in Tasklist
this can be done later as you said

service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
```diff
@@ -72,13 +72,12 @@ public SearchQuery toSearchQuery(final UserTaskFilter filter) {
...
   @Override
   public List<String> toIndices(final UserTaskFilter filter) {
-    final var completed = filter.completed();
-    final var canceled = filter.canceled();
+    final var created = filter.created();
 
-    if (completed || canceled) {
-      return Arrays.asList("tasklist-task-8.5.0_alias");
-    } else {
+    if (created) {
```
houssain-barouni, line 77: if the filter has `created == true` and `completed == true`, we need to use the alias.
But I said above, it is not clear when we use these booleans and when to use the `states` list in UserTasklistFilter

houssain-barouni (commented):
I have put some questions

service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
nathansandi: The boolean is just a more direct way to apply the filters (my inspiration was based on ProcessInstanceFilter implementation)

nathansandi: True, this is redundant and has better way to work on this

service/src/main/java/io/camunda/service/transformers/filter/UserTaskFilterTransformer.java
tmetzke: Thanks @nathansandi and @houssain-barouni, sound good to me for starters 👍 I agree that it can be done later in a follow-up 👍

service/src/main/java/io/camunda/service/search/filter/UserTaskFilter.java
```diff
@@ -165,11 +159,11 @@ public Builder processInstanceKeys(final List<Long> values) {
       return this;
     }
 
-    public Builder processDefinitionKeys(final String value, final String... values) {
+    public Builder processDefinitionKeys(final Long value, final Long... values) {
```
tmetzke, line 162: 🔧 I think I had a remark for Roman when we added the process instance filter: This method signature is not very common and also superficial if you ask me. The method signature `final Long... values` itself should be equivalent in behavior and is way more intuitive and common.

nathansandi: Sorry @tmetzke  it is not clear  for me.

What action I should take over here? Keep as a String or follow as Long?
Once its keys I think we should proceed with Long value to keep consistency with other keys. 
I will align this with @romansmirnov and if necessary create a further PR to fix it.

tmetzke: Sorry for not being clear enough. This would be my suggestion:
```suggestion
    public Builder processDefinitionKeys(final Long... values) {
```

tmetzke: And it would apply to all other filter methods as well.

PR #20479, opened 2024-07-24
feat: Search user tasks - Controller 

### LINKED ISSUE #19211: [C8 Unified V2 API] Implement Search Task endpoint
**Description**

This endpoint is to retrieve tasks and prepare for advanced search implementtion

Should be defined as: 
`/v2/user-tasks/`
Method: POST

We only will support equals on this iteration, use the V1 endpoint as inspiration over the Request, but equals is gonna be set as the `$eq` operator

**Acceptance Criteria**
- [ ] Integration Tests
- [ ] Be able to Get Completed/run time tasks from ES/OS.

This endpoint is not real time. 

**Note:** 
User task restrictions are not in the scope of this implementation, and will be made in a different Pull Request

nathansandi:
## Description

-  Fix the Service field names
- Move and implement UserTaskQuery class in order to put inside Camunda Controller
- Update the rest-api fields
- Unit Tests 

## Related issues

closes #https://github.com/camunda/camunda/issues/19211


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskController.java
```diff
@@ -121,4 +127,36 @@ private CompletableFuture<ResponseEntity<Object>> updateUserTask(
                 .withAuthentication(RequestMapper.getAuthentication())
                 .updateUserTask(request.userTaskKey(), request.changeset(), request.action()));
   }
+
+  @PostMapping(
+      path = "/search",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<Object> searchUserTasks(
```
tmetzke, line 135: ❌ We need to move this search functionality into a separate `UserTaskQueryController` class that is annotated with `@CamundaRestQueryController` so we can disable it by default for 8.6.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -114,27 +113,17 @@ Either<ProblemDetail, T> buildSearchQuery(
...
     if (filter != null) {
-      final var variableFiltersResult = toVariableValueFilter(filter.getVariables());
-      if (variableFiltersResult.isLeft()) {
-        validationErrors.addAll(variableFiltersResult.getLeft());
-      } else {
-        final var variableFilters = variableFiltersResult.get();
-        if (variableFilters != null) {
-          builder.variable(variableFilters);
-        }
+      final var variableFilters = toVariableValueFilter(filter.getVariables()).get();
```
tmetzke, line 118: 🔧 If there is no expected error result (Left), then simply don't use an Either. It makes the code harder to understand when an Either is involved but never used as such.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1282,8 +1305,132 @@ components:
...
+            type: string
+    DateFilter:
+      type: object
+      properties:
+        from:
+          type: string
+          format: date-time
+        to:
+          type: string
+          format: date-time
```
tmetzke, line 1433: 💭 I have been thinking about this for a while. Introducing this now creates certain expectations/standards that other query endpoints should follow so we're consistent. We'll get search options for date attributes with the advanced search topic (lower than, greater than, ...). How about we don't support date attribute filtering yet and wait for advanced search to enable this?

**Why?**: If we add `from` and `to` now, it's part of the API. We cannot simply remove this with 8.7 again in favor of the advanced search options (that will bring the same functionality but in a standardized way, as far as I can see) as this would be a breaking change in the API. Thus, I'd like to be a bit more conservative here with introducing constructs that we have also never talked about yet for the C8 REST API, to be honest. This construct has never been presented in our endpoint overview or anything. 

Let me know what you think. I assume this is something we had in the Tasklist component (V1) REST API?

tmetzke: @houssain-barouni, happy for your thoughts/input here as well so you've seen this discussion.

service/src/main/java/io/camunda/service/search/filter/UserTaskFilter.java
```diff
@@ -35,9 +35,9 @@ public final record UserTaskFilter(
     implements FilterBase {
 
   public static final class Builder implements ObjectBuilder<UserTaskFilter> {
-    private List<Long> userTaskKeys;
-    private List<String> userTaskDefinitionIds;
-    private List<String> processNames;
+    private List<Long> keys;
+    private List<String> elementIds;
+    private List<String> bpmProcessDefinitionIds;
```
tmetzke, line 40: ❓ Where is this name coming from, why are we changing to this name?

nathansandi: My idea over change this name is have consistency over the other names as well. 
Over the service layer we call userService(), and we know the key is related to the user task

related to the elementsIds and bpmProcessDefinitionIds, the goal is have the same names as the responses. 

Do you have a different aspect over here?

tmetzke: OK, I see, I was mostly wondering about the `bpmProcessDefinitionIds`, sorry for not being clear enough (I only marked that attribute but GitHub displays the lines before that as well). I tried to find the term `bpmProcessDefinitionId` somewhere else in the existing codebase but couldn't find anything. So far, I can only find `bpmnProcessId` but no `bpmProcessDefinitionId`. Can you tell me why you chose `bpmProcessDefinitionId` explicitly? Do we use this somewhere else as well?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: If we agree that we don't want to add date attribute filters yet, let's remove the code related to this from this PR, to be honest. Removing such a chunk again later from multiple places (YAML, different classes) is rather cumbersome and chances are high someones else starts using this as well.

service/src/main/java/io/camunda/service/search/filter/UserTaskFilter.java
nathansandi: You have reason over here
This should be `bpmnProcessId`, Just recheck in the document. 
I had a mistaken with the V1 Tasklist API that we were using <attribute>DefinitionId

tmetzke: OK, no worries, not a big deal and change, I just wanted to make sure I didn't miss anything 😄

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
```diff
@@ -90,4 +93,53 @@ private static DecisionDefinitionItem toDecisionDefinition(final DecisionDefinit
...
+
+    if (sortValues != null) {
+      page.setLastSortValues(Arrays.asList(sortValues));
+    }
+
+    if (items != null) {
+      response.setItems(toUserTasks(items));
+    }
+
+    return response;
```
tmetzke, line 116: 🔧 Can we align this method's code with the other methods, e.g. `toProcessInstanceSearchQueryResponse`? This code can be made more concise by reusing existing methods, I think.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
```diff
@@ -0,0 +1,63 @@
...
+              "Validation failed for UserTask Search Query");
+      return RestErrorMapper.mapProblemToResponse(problemDetail);
+    } catch (final Exception e) {
+      final var problemDetail =
+          RestErrorMapper.createProblemDetail(
+              HttpStatus.INTERNAL_SERVER_ERROR,
+              e.getMessage(),
+              "Failed to execute UserTask Search Query");
+      return RestErrorMapper.mapProblemToResponse(problemDetail);
+    }
```
tmetzke, line 61: 👍 I still really like this distinction, other search controllers can learn from this 🚀

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryControllerTest.java
```diff
@@ -0,0 +1,393 @@
...
+            ]
+        }""";
+    final var expectedResponse =
+        String.format(
+            """
+        {
+          "type": "about:blank",
+          "title": "Bad Request",
+          "status": 400,
+          "detail": "Unknown sortOrder: dsc",
```
tmetzke, line 208: ❌ Heads-up: This will need to change after you rebased your PR since we aligned the problem response handling.
```suggestion
          "type": "about:blank",
          "title": "INVALID_ARGUMENT",
          "status": 400,
          "detail": "Unknown sortOrder: dsc.",
```

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryControllerTest.java
```diff
@@ -0,0 +1,393 @@
...
+            ]
+        }""";
+    final var expectedResponse =
+        String.format(
+            """
+        {
+          "type": "about:blank",
+          "title": "Bad Request",
+          "status": 400,
+          "detail": "Unknown sortBy: unknownField",
```
tmetzke, line 250: ❌ Heads-up: This will need to change after you rebased your PR since we aligned the problem response handling.
```suggestion
          "type": "about:blank",
          "title": "INVALID_ARGUMENT",
          "status": 400,
          "detail": "Unknown sortBy: unknownField.",
```

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryControllerTest.java
```diff
@@ -0,0 +1,393 @@
...
+            ]
+        }""";
+    final var expectedResponse =
+        String.format(
+            """
+        {
+          "type": "about:blank",
+          "title": "Bad Request",
+          "status": 400,
+          "detail": "Sort field must not be null",
```
tmetzke, line 291: ❌ Heads-up: This will need to change after you rebased your PR since we aligned the problem response handling.
```suggestion
          "type": "about:blank",
          "title": "INVALID_ARGUMENT",
          "status": 400,
          "detail": "Sort field must not be null.",
```

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryControllerTest.java
```diff
@@ -0,0 +1,393 @@
...
+            }
+        }""";
+    final var expectedResponse =
+        String.format(
+            """
+        {
+          "type": "about:blank",
+          "title": "Bad Request",
+          "status": 400,
+          "detail": "Invalid date format for 'from': Text 'invalid-date' could not be parsed at index 0",
```
tmetzke, line 332: ❌ Heads-up: This will need to change after you rebased your PR since we aligned the problem response handling.
```suggestion
          "type": "about:blank",
          "title": "INVALID_ARGUMENT",
          "status": 400,
          "detail": "Invalid date format for 'from': Text 'invalid-date' could not be parsed at index 0.",
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -159,50 +159,51 @@ private static Either<List<String>, DecisionDefinitionFilter> toDecisionDefiniti
...
+      return Either.right(null);
     }
 
-    return Either.right(null);
+    final Object[] searchAfter = toArrayOrNull(requestedPage.getSearchAfter());
+    final Object[] searchBefore = toArrayOrNull(requestedPage.getSearchBefore());
+
+    if (searchAfter != null && searchBefore != null) {
+      return Either.left(
+          List.of("Error: Both searchAfter and searchBefore cannot be set at the same time."));
```
tmetzke, line 197: 👍 I like this new validation 🚀 
🔧 Let's please add all error messages to the `ErrorMessages` class so they are centralized. Please also remove the `.` at the end of the message, those will be added by default to all error messages 🙂

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryControllerTest.java
```diff
@@ -0,0 +1,393 @@
...
+            }
+        }""";
+    final var expectedResponse =
+        String.format(
+            """
+        {
+          "type": "about:blank",
+          "title": "Bad Request",
+          "status": 400,
+          "detail": "Error: Both searchAfter and searchBefore cannot be set at the same time.",
```
tmetzke, line 372: ❌ Heads-up: This will need to change after you rebased your PR since we aligned the problem response handling.
```suggestion
          "type": "about:blank",
          "title": "INVALID_ARGUMENT",
          "status": 400,
          "detail": "Error: Both searchAfter and searchBefore cannot be set at the same time.",
```

tmetzke (commented):
🚀 We're almost there, I think. 
🔧 I found some smaller things to adjust.
❌ Please also rebase your PR, the search request mapper has been adjusted a bit.

tmetzke (approved):
🚀 Thanks, @nathansandi, for sticking to this and applying incoming suggestions and upstream changes.

ℹ️ I rebased the PR onto `main`, renamed some commits to adhere to our commitlinting, and added two cleanup commits on top (nothing major, feel free to look up the changes in those last two commits).

✅ Ready to go from my end so far. Any smaller things we find we can still correct in follow-ups.

PR #20481, opened 2024-07-24
Use file based snapshots in raft randomized tests

deepthidevaki:
## Description

To cover more cases, we now use the actual snapshot store in RaftRandomizedTests instead of the in-memory ones. With this change, it was able to reproduce the bug #19984 

- Split FileBasedSnapshotStoreImpl to remove direct dependency to actors. It can now accept a ConcurrencyControl which can be used by the tests
- Added a TestFileBasedSnapshotStore which uses the TestConcurrencyControl instead of the actor.
- Use the new test store in raft randomized tests
- Added a new test `livenessTestsWithSnapshotsAndSingleRestart` which can reproduce the bug. It should have been also reproduced by the existing `livenessTestsWithSnapshotsAndRestarts`, but I couldn't. My assumption is that it was not likely to generate the sequence that reproduce it.

The usage of in-memory snapshot is only replace in RaftRandomizedTests. The other raft tests still uses the in-memory ones.

### DIFF
--- a/zeebe/atomix/cluster/src/test/java/io/atomix/raft/ControllableRaftContexts.java
+++ b/zeebe/atomix/cluster/src/test/java/io/atomix/raft/ControllableRaftContexts.java
@@ -30,8 +30,6 @@ import io.atomix.raft.partition.RaftElectionConfig;
 import io.atomix.raft.partition.RaftPartitionConfig;
 import io.atomix.raft.protocol.ControllableRaftServerProtocol;
 import io.atomix.raft.roles.LeaderRole;
-import io.atomix.raft.snapshot.InMemorySnapshot;
-import io.atomix.raft.snapshot.TestSnapshotStore;
 import io.atomix.raft.storage.RaftStorage;
 import io.atomix.raft.storage.log.IndexedRaftLogEntry;
 import io.atomix.raft.storage.log.RaftLog;
@@ -39,6 +37,8 @@ import io.atomix.raft.storage.log.RaftLogReader;
 import io.atomix.raft.zeebe.EntryValidator.NoopEntryValidator;
 import io.atomix.raft.zeebe.ZeebeLogAppender.AppendListener;
 import io.camunda.zeebe.journal.JournalException;
+import io.camunda.zeebe.scheduler.testing.TestConcurrencyControl;
+import io.camunda.zeebe.snapshots.testing.TestFileBasedSnapshotStore;
 import io.camunda.zeebe.util.FileUtil;
 import io.camunda.zeebe.util.collection.Tuple;
 import java.io.File;
@@ -62,7 +62,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.jmock.lib.concurrent.DeterministicScheduler;
@@ -88,7 +87,7 @@ public final class ControllableRaftContexts {
 
   private final int nodeCount;
   private final Map<MemberId, RaftContext> raftServers = new HashMap<>();
-  private final Map<MemberId, TestSnapshotStore> snapshotStores = new HashMap<>();
+  private final Map<MemberId, TestFileBasedSnapshotStore> snapshotStores = new HashMap<>();
   private Duration electionTimeout;
   private Duration heartbeatTimeout;
   private int nextEntry = 0;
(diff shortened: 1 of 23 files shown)


zeebe/snapshot/src/main/java/io/camunda/zeebe/snapshots/impl/FileBasedSnapshotStoreImpl.java
```diff
@@ -0,0 +1,684 @@
+/*
```
deepthidevaki, line 1: This is mainly a copy of the previous `FileBasedSnapshotStore`. The only changes are to replace the usage of actor.

zeebe/scheduler/src/test/java/io/camunda/zeebe/scheduler/testing/TestConcurrencyControl.java
```diff
@@ -44,6 +48,35 @@ public <T> void runOnCompletion(
...
+          (r, e) -> {
+            if (e != null) {
+              error.set(e);
+            }
+            if (futuresCompleted.decrementAndGet() == 0) {
+              if (error.get() != null) {
+                finalFuture.completeExceptionally(error.get());
+              } else {
+                finalFuture.complete(null);
+              }
```
lenaschoenburg, line 73: :upside_down_face: Technically not needed, you could just always complete the future successfully with whatever is in `error`.

zeebe/atomix/cluster/src/test/java/io/atomix/raft/ControllableRaftContexts.java
```diff
@@ -88,7 +87,7 @@ public final class ControllableRaftContexts {
 
   private final int nodeCount;
   private final Map<MemberId, RaftContext> raftServers = new HashMap<>();
-  private final Map<MemberId, TestSnapshotStore> snapshotStores = new HashMap<>();
+  private final Map<MemberId, TestFileBasedSnapshotStore> snapshotStores = new HashMap<>();
```
lenaschoenburg, line 90: :wrench: These should be closed in `ControllableRaftContexts#shutdown`

zeebe/snapshot/src/main/java/io/camunda/zeebe/snapshots/impl/FileBasedSnapshotStoreImpl.java
```diff
@@ -0,0 +1,684 @@
...
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class FileBasedSnapshotStoreImpl {
+  public static final String SNAPSHOTS_DIRECTORY = "snapshots";
+  public static final String PENDING_DIRECTORY = "pending";
```
lenaschoenburg, line 46: :thought_balloon: I'm wondering if these constants (more down below) shouldn't be part of the non-impl class.

zeebe/snapshot/src/test/java/io/camunda/zeebe/snapshots/testing/TestFileBasedSnapshotStore.java
```diff
@@ -0,0 +1,160 @@
...
+                throw new UncheckedIOException(e);
+              }
+            });
+      } catch (final IOException e) {
+        throw new UncheckedIOException(e);
+      }
+
+      return checksums;
+    }
+  }
```
lenaschoenburg, line 159: :thought_balloon: This shouldn't be needed, a no-op one that returns an empty map has the same behavior. A checksum provider is only useful if there are files for which we know the checksum _without_ reading all of the file contents somehow (i.e. because they are provided by RocksDB).

lenaschoenburg (approved):
Thanks @deepthidevaki, changes make sense to me :+1:

zeebe/snapshot/src/test/java/io/camunda/zeebe/snapshots/testing/TestFileBasedSnapshotStore.java
deepthidevaki: A noop one do not reproduce the bug, because it results in calculating the combined checksum as it was before. In the tests, it makes sense to keep it as close to production usage to catch such inconsistencies.

lenaschoenburg: Oh right, I missed that aspect completely. All good then :+1:

zeebe/snapshot/src/main/java/io/camunda/zeebe/snapshots/impl/FileBasedSnapshotStoreImpl.java
deepthidevaki: They are only used by the implementation classes and tests though. So I don't see a reason to put it elsewhere.

PR #20483, opened 2024-07-24
ci: don't cancel concurrent builds on `main` for full results

cmur2:
## Description

Currently very few builds in `main` are actually completely green:

image

Looking into [CI health dashboard](https://dashboard.int.camunda.com/d/bdmo5l8puugaoc/ci-health-camunda-camunda-monorepo?orgId=1) this also comes from a large percentage of canceled builds:

image

One reason for the cancelations is the [mechanism we use to save costs](https://github.com/camunda/camunda/blob/ebd6677b941cd64e8e8107bb37b9bbb7979ee882/.github/workflows/operate-ci.yml#L34-L38) by preventing concurrent builds: when a new commit comes in to a branch/PR, older builds get aborted ➡️the assumption here is that only the most recent check results are useful and relevant. While this is true for PRs it is not so true for stable branches like `main` where new changes get merged continually.

I want to adopt the solution that is also used by Web-Modeler team to prevent concurrent builds in general (to save costs) but make an exception for `main`: https://github.com/camunda/web-modeler/blob/main/.github/workflows/build-and-test.yml#L13-L19

Ultimately this allows to have more confidence in the build results on `main` ➡️failures will not be due to cancellations anymore.

## Related issues

None

### DIFF
--- a/.github/workflows/zeebe-ci.yml
+++ b/.github/workflows/zeebe-ci.yml
@@ -37,9 +37,12 @@ on:
   workflow_dispatch: { }
   workflow_call: { }
 
+# Limit workflow to 1 concurrent run per ref (branch): new commit -> old runs are canceled to save costs
+# Exception for main branch: complete builds for every commit needed for confidenence
+# Exception for deploy jobs that have to wait for each other to avoid overwriting
 concurrency:
   cancel-in-progress: true
-  group: "${{ github.workflow }}-${{ github.ref }}"
+  group: ${{ github.ref == 'refs/heads/main' && github.sha || format('{0}-{1}', github.workflow, github.ref) }}
 
 defaults:
   run:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,12 @@ on:
   merge_group: {}
   workflow_dispatch: {}
 
+# Limit workflow to 1 concurrent run per ref (branch): new commit -> old runs are canceled to save costs
+# Exception for main branch: complete builds for every commit needed for confidenence
+# Exception for deploy jobs that have to wait for each other to avoid overwriting
 concurrency:
   cancel-in-progress: true
-  group: "${{ github.workflow }}-${{ github.ref }}"
+  group: ${{ github.ref == 'refs/heads/main' && github.sha || format('{0}-{1}', github.workflow, github.ref) }}
 
 defaults:
   run:
--- a/.github/workflows/operate-a11y.yml
+++ b/.github/workflows/operate-a11y.yml
@@ -13,6 +13,12 @@ on:
       - '.github/workflows/operate-a11y.yml'
       - 'operate/client/**'
 
+# Limit workflow to 1 concurrent run per ref (branch): new commit -> old runs are canceled to save costs
+# Exception for main branch: complete builds for every commit needed for confidenence
+concurrency:
+  cancel-in-progress: true
+  group: ${{ github.ref == 'refs/heads/main' && github.sha || format('{0}-{1}', github.workflow, github.ref) }}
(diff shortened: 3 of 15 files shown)


.github/workflows/zeebe-ci.yml
```diff
@@ -37,9 +37,11 @@ on:
   workflow_dispatch: { }
   workflow_call: { }
 
+# Limit workflow to 1 concurrent run per ref (branch): new commit -> old runs are canceled to save costs
+# Exception for main branch: complete builds for every commit needed for confidenence
 concurrency:
   cancel-in-progress: true
-  group: "${{ github.workflow }}-${{ github.ref }}"
+  group: ${{ github.ref == 'refs/heads/main' && github.sha || format('{0}-{1}', github.workflow, github.ref) }}
```
lenaschoenburg, line 44: :x: This workflow deploys snapshot artifacts if it runs on main, see https://github.com/camunda/camunda/blob/ebd6677b941cd64e8e8107bb37b9bbb7979ee882/.github/workflows/zeebe-ci.yml#L601-L605. Therefor it is important that it does not run concurrently on main.

:x: Can we not achieve the same by saying `cancel-in-progress: ${{ github.ref != 'refs/heads/main'}}`? That way we cancel in-progress ones for PRs (saving resources) while queuing all runs on main, giving us safety for deploying snapshot artifacts _and_ increased test coverage.

cmur2: > Therefor it is important that it does not run concurrently on main.

Gotcha, this is a very good point! I'm not insisting on having this change in all workflows so I it is fine for me to drop it for this workflow 🙂 

> 💭 Can we not achieve the same by saying `cancel-in-progress: ${{ github.ref != 'refs/heads/main'}}`? That way we cancel in-progress ones for PRs (saving resources) while queuing all runs on main, giving us safety for deploying snapshot artifacts and increased test coverage.

Maaaaybe, I'm not 110% sure if that's possible and there are no other downsides. Going back to status quo before this PR for this 1 workflow is fine as well

lenaschoenburg: > Maaaaybe, I'm not 110% sure if that's possible and there are no other downsides. Going back to status quo before this PR for this 1 workflow is fine as well

Happy to discuss this, I currently don't see any risk with this approach. Assuming it works as intended, I think this would be the _safest_ option, especially because its future proof when one of the other workflows gets extended to also deploy artifacts or something similar.

cmur2: Actually it also seems to be possible to do this on the job level as demonstrated here by creating a different concurrency group: https://github.com/camunda/camunda/blob/edd4af9c3a512982b11ff1a5380c83c7184d3b13/.github/workflows/ci.yml#L490-L492

That should work together with not cancelling builds on `main` because the *deploy jobs specifically would wait on each other 🤔

cmur2: Zeebe CI already has that: https://github.com/camunda/camunda/blob/edd4af9c3a512982b11ff1a5380c83c7184d3b13/.github/workflows/zeebe-ci.yml#L607-L609 🤯 Sooo all good with my original change proposal?

cmur2: > But that's not a strong opinion, feel free to merge as is.

I agree and will keep it in mind for future change but right now will not do it to get the original purpose of the PR merged.

lenaschoenburg (approved):

.github/workflows/zeebe-ci.yml
cmur2: I'll also want to document the behavior at one point properly to save time for future-selves.

Issue #20176, opened 2024-07-09
Send response in `MessageSubscriptionRejectProcessor`


remcowesterhoud:
<h2>Description</h2>

- If there is another subscription then set the request data to the other message subscription record
- If there are no subscriptions then send the API response without a process instance key and NOT_CORRELATED

<h3>Dependencies</h3>

- #20174

remcowesterhoud:
Blocked by #20568

Issue #20175, opened 2024-07-09
Send response in `MessageSubscriptionCorrelateProcessor`


remcowesterhoud:
<h2>Description</h2>

If we successfully correlate the message to a subscription we need to send a `CORRELATED` response to the client.

- In the `MessageSubscriptionCorrelateProcessor ` write the `MessageCorrelation.CORRELATED` event and send it as response to the client.

⚠️ In the `CorrelateMessageTest` there are a few tests that are currently ignored. In this issue, these tests should be unignored and passing.

<h3>Dependencies</h3>

- #20174

PR #20493, opened 2024-07-24
Store message request data in state

### LINKED ISSUE #20174: Store request data in message subscription
<h2>Description</h2>

We need to be able to send a response back to the client. We can do this using the request data that comes in with the `MessageCorrelation.CORRELATE` command. Since we send the response in the `MessageSubscriptionCorrelateProcessor` we need to make sure we have the request data available here. This means we'll have to store it in the state so we can access it later.

For this we must:
- Set the request data on the `MessageSubscriptionRecord` in the `MessageCorrelatingCorrelateProcessor`
- Store the request data in the `subscriptionColumnFamily` in the `MessageSubscriptionCorrelatingApplier`

⚠️ The request data should only be stored for a single message! If the message correlated to multiple messages we can just pick the first one. We cannot send multiple responses back to the client, and with the current design it's not possible to collect all the process instances before sending the response.

<h3>Dependencies</h3>

- #20167 
- #20173

remcowesterhoud:
## Description

In order to send a response to the `MessageCorrelation.CORRELATE` command we need to have access to the request data. Since we don't always send the response in the processor we need to store it in the state so we can reaccess it when we need it (similar to create PI with result).
This PR makes it so we store the request data in the message subscription. If the message correlates with multiple subscriptions, we only store the request data for one of them, as we can only send 1 response back to the client.

⚠️ This PR contains no tests. These will added as part of #20175 and #20176

## Related issues

closes #20174

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelateBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelateBehavior.java
@@ -14,6 +14,8 @@ import io.camunda.zeebe.engine.state.immutable.MessageStartEventSubscriptionStat
 import io.camunda.zeebe.engine.state.immutable.MessageState;
 import io.camunda.zeebe.engine.state.immutable.MessageSubscriptionState;
 import io.camunda.zeebe.protocol.record.intent.MessageSubscriptionIntent;
+import java.util.Optional;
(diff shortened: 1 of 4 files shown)


nicpuppa (approved):
LGTM

nicpuppa:
🥳

remcowesterhoud:
Just when I made a script to automatically retry this for me 😄

PR #20501, opened 2024-07-24
feat: Search user task - Client

### LINKED ISSUE #19211: [C8 Unified V2 API] Implement Search Task endpoint

nathansandi:
## Description

- Client implementation
- Client API Calls Filters/Sort/Page
- unit/integration tests (Integration tests will be disable for the release) 

-Sort (Creation/Completion) Date - More parameters will add in the further in separate PRs if it is the case. 
-A previous round of reviews was done in an single PR,  this PR is a refactor 

## Related issues

closes #https://github.com/camunda/camunda/issues/19211


qa/util/src/main/java/io/camunda/qa/util/cluster/TestStandaloneCamunda.java
```diff
@@ -68,7 +72,16 @@ public final class TestStandaloneCamunda extends TestSpringApplication<TestStand
           .withEnv("xpack.security.enabled", "false")
           .withEnv("xpack.watcher.enabled", "false")
           .withEnv("xpack.ml.enabled", "false")
-          .withEnv("action.destructive_requires_name", "false");
+          .withEnv("action.destructive_requires_name", "false")
+          .withExposedPorts(9200, 9300) // Expose the ports used by Elasticsearch
```
nathansandi, line 76: You can ignore this for this review, the tests is running locally forcing on the port 9200

I am on it with @nicpuppa in order to solve the port problem on camunda client builder together.

clients/java/src/main/java/io/camunda/client/api/search/UserTaskFilter.java
```diff
@@ -0,0 +1,135 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.client.api.search;
+
+import io.camunda.client.api.search.TypedSearchQueryRequest.SearchRequestFilter;
+import io.camunda.client.protocol.rest.DateFilter;
+
+/** Interface for defining user task filters in search queries. */
+public interface UserTaskFilter extends SearchRequestFilter {
```
tmetzke, line 22: ❌ All filter attributes are already in the context of user tasks. The attributes can thus all drop the `userTask...` prefix and just be named `state`, `assignee`, and so on.

clients/java/src/main/java/io/camunda/client/api/search/UserTaskQuery.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.client.api.search;
+
+import io.camunda.client.protocol.rest.UserTaskItem;
```
tmetzke, line 33: ❌ The return type should be a specific client API class, not the REST model. See process instance search as an example.

clients/java/src/main/java/io/camunda/client/impl/search/SearchResponseMapper.java
```diff
@@ -53,4 +55,20 @@ public static SearchQueryResponse<ProcessInstance> toProcessInstanceSearchRespon
 
     return new SearchQueryResponseImpl<>(instances, page);
   }
+
+  public static SearchQueryResponse<UserTaskItem> toUserTaskSearchResponse(
+      final UserTaskSearchQueryResponse response) {
+    final SearchQueryPageResponse pageResponse = response.getPage();
+    final SearchResponsePage page = toSearchResponsePage(pageResponse);
+
+    return new SearchQueryResponseImpl<>(response.getItems(), page);
```
tmetzke, line 64: ❌ The return type of the query should be a client API class, not the REST model. Have a look at the process instance search above.

clients/java/src/main/java/io/camunda/client/impl/search/UserTaskQueryImpl.java
```diff
@@ -0,0 +1,127 @@
...
+  public UserTaskQueryImpl(final HttpClient httpClient, final JsonMapper jsonMapper) {
+    request = new UserTaskSearchQueryRequest();
+    this.jsonMapper = jsonMapper;
+    this.httpClient = httpClient;
+    httpRequestConfig = httpClient.newRequestConfig();
+  }
+
+  @Override
+  public FinalSearchQueryStep<UserTaskItem> requestTimeout(final Duration requestTimeout) {
+    return null;
```
tmetzke, line 71: ❌ The request timeout needs to be passed on to the HTTP client configuration, so it's respected. Have a look at the `ProcessInstanceQueryImpl`for reference.

clients/java/src/main/java/io/camunda/client/impl/search/UserTaskQueryImpl.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Override
+  public HttpCamundaFuture<SearchQueryResponse<UserTaskItem>> send() {
```
tmetzke, line 75: ❌ The return type of the query also needs to change here then when it's adjusted accordingly in the interface and such.

clients/java/src/main/java/io/camunda/client/impl/search/UserTaskQueryImpl.java
```diff
@@ -0,0 +1,127 @@
...
+        httpRequestConfig.build(),
+        UserTaskSearchQueryResponse.class,
+        SearchResponseMapper::toUserTaskSearchResponse,
+        result);
+    return result;
+  }
+
+  @Override
+  public UserTaskQuery filter(final UserTaskFilter value) {
+    final UserTaskFilterImpl filter = (UserTaskFilterImpl) value;
```
tmetzke, line 89: ❌ Simply doing a cast here might fail unexpectedly.b please have a look at the `ProcessInstanceQueryImpl` for reference on how to do this. There are multiple places in this class that need an adjustment accordingly.

qa/integration-tests/src/test/java/io/camunda/it/client/SearchUserTaskTest.java
```diff
@@ -0,0 +1,481 @@
...
+        camundaClient
+            .newUserTaskQuery()
+            .filter(f -> f.userTaskFollowUpDate(dateFilterOutOfRange))
+            .send()
+            .join();
+    assertThat(resultOutOfRange.items().size()).isEqualTo(0);
+  }
+
+  @Test
+  public void shouldThrowAnExceptionWhenSearchBeforeAndAfterAreApplied() {
```
tmetzke, line 315: 🔧 This can already be tested in the client itself, since it's just input validation. This doesn't have to be an integration test which simply is more costly due to its setup.

qa/integration-tests/src/test/java/io/camunda/it/client/SearchUserTaskTest.java
```diff
@@ -0,0 +1,481 @@
...
+  }
+
+  @Test
+  public void shouldSortTasksByStartDateASC() {
+    final var result =
+        camundaClient.newUserTaskQuery().sort(s -> s.creationDate().asc()).send().join();
+
+    assertThat(result.items().size()).isEqualTo(3);
+
+    // Retrieve IDs from the result
```
tmetzke, line 372: 💭 How about we just assert that creation date of item 0 is before that of item 1 and that the date of item 1 is before that of item 2? We don't have to extract and sort the whole list then. Just an idea.

tmetzke (changes requested):
👍 Thanks, @nathansandi, good job already!
🔧 There are some places that need some attention but nothing major. Please consider the existing search implementation a bit more, it serves as a good reference usually.
💭 Overall, there is a lot of "old" Java date usage in here. I would recommend looking at Java 8 date API a bit more and using that wherever possible, it's usually more convenient to use.
💭 The client API for date input (e.g. filter by creation date and such) only supports string dates, which might be a bit inconvenient for some users. Usually, we also support methods that allow to provide real date objects, e.g. an `OffsetDateTime`.

nathansandi:
> 👍 Thanks, @nathansandi, good job already! 🔧 There are some places that need some attention but nothing major. Please consider the existing search implementation a bit more, it serves as a good reference usually. 💭 Overall, there is a lot of "old" Java date usage in here. I would recommend looking at Java 8 date API a bit more and using that wherever possible, it's usually more convenient to use. 💭 The client API for date input (e.g. filter by creation date and such) only supports string dates, which might be a bit inconvenient for some users. Usually, we also support methods that allow to provide real date objects, e.g. an `OffsetDateTime`.

Thanks for the review @tmetzke, some considerations: 

The Date setup over the API is based to what was implemented to ProcessInstance Client, where the client is also using String. Should we define a convention and refactor in a different PR. Once , it would be necessary refactor on ProcessInstanceQuery as well. 

Related to the "old" Java date usage in here, I believe it is related to the Date conversion or over the Calendar in tests in order to run the Integration tests. Can you clarify this point. 

I have applied the rest of suggestions from the comments and implementations. (https://github.com/camunda/camunda/pull/20501#:~:text=%E2%80%A6-,e858c5a,-Merge%20state)

clients/java/src/main/java/io/camunda/client/api/response/UserTaskEvent.java
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.client.api.response;
+
+import java.util.List;
+import java.util.Map;
+
+public interface UserTaskEvent {
```
tmetzke, line 21: 🔧 No need to create this explicitly. The `...Event` interfaces are usually created for Zeebe interactions with the API, e.g. creating resources like a process instance. Since we don't do this for user tasks, there is no need for this. You can directly keep all this in the `UserTask` resource defined as a search response.

💭 Should we need to implement this in the future, e.g. for handling user task listeners in the client, we can create it and have the `UserTask` interface inherit from the then-created event class. But we don't need this now.

ℹ️ If you'd still like to keep this event class already, it should only contain the runtime data provided by Zeebe. Things like the `state` and `completionDate` are not provided by Zeebe for user tasks. Those two would have be removed here. The `UserTask` search response interface could add them on top of this interface here then.

clients/java/src/main/java/io/camunda/client/api/search/response/UserTask.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.camunda.client.api.response.UserTaskEvent;
+
+public interface UserTask extends UserTaskEvent {
+
+  @Override
+  String getCreationDate();
+
+  @Override
+  String getCompletionDate();
```
tmetzke, line 26: ❌ If the interface here extends the `UserTaskEvent`, it should only add new methods, not override existing ones. Otherwise, the interface here is rather useless. If you want to keep the separation between the two interfaces, remove the `String getCompletionDate();` and `String getState();` from the `UserTaskEvent` and add them here in this interface.

clients/java/src/main/java/io/camunda/client/impl/search/SearchResponseMapper.java
```diff
@@ -56,12 +57,26 @@ public static SearchQueryResponse<ProcessInstance> toProcessInstanceSearchRespon
...
-  public static SearchQueryResponse<UserTaskItem> toUserTaskSearchResponse(
+  public static SearchQueryResponse<UserTask> toUserTaskSearchResponse(
       final UserTaskSearchQueryResponse response) {
     final SearchQueryPageResponse pageResponse = response.getPage();
-    final SearchResponsePage page = toSearchResponsePage(pageResponse);
+    final SearchResponsePage page =
+        new SearchResponsePageImpl(
+            pageResponse.getTotalItems(),
+            pageResponse.getFirstSortValues(),
+            pageResponse.getLastSortValues());
```
tmetzke, line 67: ❓ Why are you not using the `toSearchResponsePage` method anymore?

tmetzke (changes requested):
👍 Great improvements, @nathansandi 🏅 

🔧 The new user task search response interfaces need some attention but nothing major.

💭 Regarding the "old" Java date usage, I recommend not using the Calendar class and related code anymore, not even in testing code. Try to familiarize yourself with the [Java 8 Date/Time API](https://www.baeldung.com/java-8-date-time-intro) and use it instead. Regard it as a learning experience and try to improve the code here.

⚠️ You must rebase your PR and add your code to the respective classes in the `io.camunda.zeebe.client` package instead of `io.camunda.client`. Except for the changes in the `CamundaClient` and `CamundaClientImpl`, all classes also exist in the `io.camunda.zeebe.client` package so that should be mostly straightforward.

clients/java/src/main/java/io/camunda/client/impl/search/SearchResponseMapper.java
nathansandi: Accident :)

tmetzke: Happens 👍 🙂

qa/integration-tests/src/test/java/io/camunda/it/client/SearchUserTaskTest.java
```diff
@@ -105,42 +104,26 @@ public void shouldRetrieveTaskByCandidateUser() {
 
   @Test
   public void shouldRetrieveTaskByCompletionDate() {
-    final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
-    dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+    final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+    final ZoneId utcZoneId = ZoneId.of("UTC");
 
-    final Date now = new Date();
+    final LocalDateTime now = LocalDateTime.now(utcZoneId);
```
tmetzke, line 110: 🔧 I think you can even do the following to obtain the current date for a specific time zone:
```java
final OffsetDateTime now = OffsetDateTime.now(ZoneId.of("UTC"))
```

qa/integration-tests/src/test/java/io/camunda/it/client/SearchUserTaskTest.java
```diff
@@ -105,42 +104,26 @@ public void shouldRetrieveTaskByCandidateUser() {
 
   @Test
   public void shouldRetrieveTaskByCompletionDate() {
-    final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
-    dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
+    final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
```
tmetzke, line 107: 🔧 If I'm not mistaken, you can skip using the formatter and simply do `dayBefore.toString`, for example. It will create a date string that can be parsed by our API again. For integration tests, that should be good enough.

💭 Input validation testing should happen in the REST unit tests and there you can also test with different String inputs to verify our API behaves as expected.

tmetzke:
@nathansandi, using Java 8 Date/Time itself saved ~70 lines of code 👍 😄

qa/integration-tests/src/test/java/io/camunda/it/client/SearchUserTaskTest.java
nathansandi: I can improve that when I move the integration tests to Camunda Client 👍 
Rest API intput tests will be made on this PR (in progress) https://github.com/camunda/camunda/pull/20479

Thats my next task today.

tmetzke:
@nathansandi, I cleaned up the mess I created with the controller PR and rebased this PR as well. Please double-check the PR but I'm confident I kept all the changes you made.

tmetzke (approved):
👍 All looks well from my end here, good job, @nathansandi

PR #20511, opened 2024-07-25
ci: observe build status for Zeebe deploy jobs

lenaschoenburg:
These jobs fail from time to time, we should observe them so that we have some data.

### DIFF
--- a/.github/workflows/zeebe-ci.yml
+++ b/.github/workflows/zeebe-ci.yml
@@ -642,6 +642,16 @@ jobs:
         env:
           MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTS_USR }}
           MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTS_PSW }}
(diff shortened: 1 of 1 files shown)


.github/workflows/zeebe-ci.yml
```diff
@@ -642,6 +642,16 @@ jobs:
         env:
           MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTS_USR }}
           MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTS_PSW }}
+      - name: Observe build status
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          build_status: ${{ contains(steps.*.conclusion, 'failure') && 'failure' || 'success' }}
```
cmur2, line 650: You can use `${{ job.status }}` here since it has the same effect (tested in the Unified CI already)

cmur2 (approved):

PR #20516, opened 2024-07-25
ci: add issues to Optimize project

misiekhardcore:
## Description

This PR adds auto labeling and project assigning functionality for Optimize

## Related issues

closes https://github.com/camunda/camunda-optimize/issues/13886

### DIFF
--- a/.github/workflows/optimize-labeled.yml
+++ b/.github/workflows/optimize-labeled.yml
@@ -1,0 +1,40 @@
+name: Assign labeled issues to Optimize project
+on:
+  issues:
+    types:
+      - labeled
+jobs:
+  add-to-projects:
+    name: Add issue to Optimize project if relevant component label added
+    runs-on: ubuntu-latest
+    if: ${{ github.event.label.name == 'component/optimize' }}
+    steps:
+      - id: get_project_count
+        uses: octokit/graphql-action@v2.3.2
+        with:
+          # API https://docs.github.com/en/graphql/reference/objects#issue
+          query: |
+            query getProjectCount($owner:String!, $repo:String!, $issue: Int!) {
+              repository(name: $repo, owner: $owner) {
+                issue: issue(number: $issue) {
+                  projectsV2 {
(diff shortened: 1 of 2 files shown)


buccarel (approved):

RomanJRW (approved):
Changes are fine for me. However, please consider:
- Target branch: This could probably already target `main`, there is no dependency on the "merge in" PR afaict
- It might be that @cmur2 also wants to review this. While I can't see an objection, it is not a bad thing to give him visibility on the changes that are being made in the mono repo workflows. Firstly, because it's his domain, and secondly because it can help him in making a more informed review on the merge-in PR. Note that you could also agree that he does this after its merged to the merge-in branch, if that is the approach you indeed take

cmur2:
I agree with Josh's comment and thanks for looping me in! PR looks good from my side (eventually we should deduplicate those workflows since the logic is always the same, except parameters).

misiekhardcore:
@cmur2 I noticed that when I wan working on this that there is a lot of duplication. If you create/have an issue to abstract this common logic into an action, I can work on that

misiekhardcore:
@buccarel how do you want this to be merged? to `main` or to `merge-optimize-in`?

.github/workflows/optimize-labeled.yml
```diff
@@ -0,0 +1,40 @@
...
+                issue: issue(number: $issue) {
+                  projectsV2 {
+                    totalCount
+                  }
+                }
+              }
+            }
+          variables: |
+            owner: "camunda"
+            repo: "zeebe"
```
cmur2, line 28: This is now `camunda´ (technically works with `zeebe` but better to update)

cmur2:
@misiekhardcore the refactoring ticket is https://github.com/camunda/camunda/issues/20768

.github/workflows/optimize-labeled.yml
misiekhardcore: renamed

PR #20518, opened 2024-07-25
ci: add optimize to labeler

misiekhardcore:
## Description

This PR add Optimize to labeler

## Related issues

closes https://github.com/camunda/camunda-optimize/issues/13887

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -54,6 +54,22 @@ component/tasklist:
           - 'tasklist/**'
           - 'tasklist.Dockerfile'
           - 'webapps-common/**'
+component/optimize:
+  - changed-files:
+      - any-glob-to-any-file:
+        - '.ci/docker/test/*'
+        - '.github/ISSUE_TEMPLATE/optimize-*'
+        - '.github/actions/**'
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -55,6 +55,22 @@ component/tasklist:
           - 'tasklist/**'
           - 'tasklist.Dockerfile'
           - 'webapps-common/**'
+component/optimize:
+  - changed-files:
+      - any-glob-to-any-file:
+        - '.ci/docker/test/*'
```
buccarel, line 61: FYI: We don't have any optimize files in that `.ci/` directory. Maybe this line is not necessary?

buccarel (approved):

.github/labeler.yml
misiekhardcore: maybe not, I just follow the list of dirs/files from Operate. I can see that in the `.ci` operate is most likely looking at `operate-docker-labels.golden.json` file. Is optimize going to have similar file there? if the answer is yes then we need to leave the config the way I made it now, otherwise we can remove it

PR #20089, opened 2024-07-05
Add a job completion REST endpoint

tmetzke:
## Description

Add an endpoint to the REST API for completing jobs.

## Related issues

closes #20080

PR #20523, opened 2024-07-25
Add a job failure endpoint

### LINKED ISSUE #20081: Migrate the Fail Job gRPC API to REST
## Description

* The Camunda 8 REST API contains a job failure endpoint at `POST /v2/jobs/{jobKey}/failure` providing the same functionality as the [gRPC endpoint](https://docs.camunda.io/docs/apis-tools/zeebe-api/gateway-service/#failjob-rpc).
* The endpoint is featured in the OpenAPI description as (internally) [described](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.wp2927rq1huh).
* The endpoint allows sending retries, errorMessage, retryBackOff, and variables (as a JSON object).

nicpuppa:
## Description

Add an endpoint to the REST API for failing a job.

## Related issues

closes #20081


zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/JobControllerRoundRobinTest.java
```diff
@@ -0,0 +1,361 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.boot.test.mock.mockito.SpyBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.http.MediaType;
+import org.springframework.util.unit.DataSize;
+
+@WebMvcTest(JobController.class)
+public class JobControllerRoundRobinTest extends RestControllerTest {
```
nicpuppa, line 47: I saw this on your PR for completed jobs endpoint @tmetzke

service/src/main/java/io/camunda/service/JobServices.java
```diff
@@ -12,13 +12,20 @@
...
 import io.camunda.zeebe.gateway.impl.job.ResponseObserver;
+import io.camunda.zeebe.protocol.impl.record.value.job.JobRecord;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
 
 public final class JobServices<T> extends ApiServices<JobServices<T>> {
 
+  public static final Integer DEFAULT_RETRIES_VALUE = 0;
```
nicpuppa, line 26: I'm not sure about this approach, I think it's possible to create those static variables in the generated class. But before doing that I want to know your opinion on this approach @tmetzke

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -182,6 +196,28 @@ private static Optional<ProblemDetail> validateJobActivationRequest(
             HttpStatus.BAD_REQUEST, String.join(". ", violations), INVALID_ARGUMENT.name()));
   }
 
+  private static Optional<ProblemDetail> validateJobFailRequest(
+      final JobFailRequest failRequest
+  ) {
+    final List<String> violations = new ArrayList<>();
+    if (failRequest != null) {
+      // retries can't be less than 0
+      if (failRequest.getRetries() < 0) {
```
nicpuppa, line 205: This is an assumption that I made, since I'm not able to find any documentation about it. But I think it make sense to not allow users to provide strange values

service/src/main/java/io/camunda/service/JobServices.java
```diff
@@ -12,13 +12,20 @@
...
+  public static final Long DEFAULT_RETRYBACKOFF_VALUE = -1L;
```
tmetzke, line 27: Is a backoff of `-1` really appropriate for a user-facing API? A backoff of `0` I can understand as "no backoff, immediately retry". How do I interpret a backoff of `-1`? 🤔

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1419,6 +1472,36 @@ components:
...
+            runs out of retries and an incident is raised, as this message can help explain why an
+            incident was raised.
+          type: string
+          nullable: true
+        retryBackOff:
+          description: >
+            The backoff timeout (in ms) for the next retry.
+          type: integer
+          format: int64
+          default: -1
```
tmetzke, line 1496: Is a backoff of `-1` really appropriate for a user-facing API? A backoff of `0` I can understand as "no backoff, immediately retry". How do I interpret a backoff of `-1`? 🤔

nicpuppa: I think you are right 😆

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/JobControllerRoundRobinTest.java
tmetzke: So this is a copy of the test created in the ["completeJob" endpoint PR](https://github.com/camunda/camunda/pull/20089)?

tmetzke: If so, let's rather complete it over at the other PR, or is there a specific need to copy it over here?

nicpuppa: No, I copied from the other PR. I saw it and use it as a starting point. What's the status of the PR ? It's ready to be reviewed ? Git should be smart enough to not create conflicts on this test class, even if we merge this PR first

tmetzke: I see. The other PR is not completely ready for review. I'll have to look at it again, I stopped close to being finished.
The test is however not required over here in this PR, is it? I don't mind in which PR it's contained but the classes are not entirely equal so git might have an issue with them. Let's either remove it here or in the other PR, I don't care much 😅

nicpuppa: I will remove the test class from this PR

service/src/main/java/io/camunda/service/JobServices.java
tmetzke: I'm not sure if we need those default values at all, actually.
1. The generated REST model class `JobFailRequest` already contains initialized attributes with the default values defined in the YAML.
2. In the `RequestMapper`, we could by default map `null` numeric values to 0, just like gRPC does. It will be hard to find a case where it should be something other than 0 by default.

nicpuppa: For number maybe you are right, but what about other types ?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
tmetzke: Well, the engine doesn't care much, it creates an incident when it's 0 or lower: https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobFailProcessor.java#L135-L137
That being said, I have no idea what people will want to use here, not sure if we need to be too strict here.

service/src/main/java/io/camunda/service/JobServices.java
tmetzke: Let's talk about that when we get there, I guess 🙂

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
nicpuppa: You are suggesting to remove the validation at all ?

tmetzke: Yep, I think we can remove it. The engine can apparently handle it and we don't know if users are actually maybe more comfortable or used to handing in -1, for example 🤷

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -121,6 +123,29 @@ public static Either<ProblemDetail, ActivateJobsRequest> toJobsActivationRequest
...
+                        jobKey,
+                        getIntOrDefault(
+                            failRequest,
+                            JobFailRequest::getRetries,
+                            JobServices.DEFAULT_RETRIES_VALUE),
+                        getStringOrEmpty(failRequest, JobFailRequest::getErrorMessage),
+                        getLongOrDefault(
+                            failRequest,
+                            JobFailRequest::getRetryBackOff,
+                            JobServices.DEFAULT_RETRYBACKOFF_VALUE),
```
tmetzke, line 145: 🔧 Let's just go for `getIntOrZero` and `getLongOrZero`. Zero will be the defacto default value for numeric values, just like for the gRPC API. We can add explicit defaults to the YAML and ensure here it's adhered to even if a default is missing in the YAML by accident.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/JobControllerTest.java
```diff
@@ -7,280 +7,141 @@
...
+  @MockBean ResponseObserverProvider responseObserverProvider;
 
   @BeforeEach
   void setup() {
-    responseObserver.reset();
+    when(jobServices.withAuthentication(any(Authentication.class))).thenReturn(jobServices);
   }
 
   @Test
-  void shouldActivateJobsImmediatelyIfAvailable() throws Exception {
```
tmetzke, line 61: ℹ️ Just for reference in case someone comes across this: the job activation tests will be added with this [JobControllerRoundRobinTest](https://github.com/camunda/camunda/pull/20089/files#diff-b172358d9f2fd6f395dfeb2ecf928264c077fe07da1ed5b9d681557a888bbd59) from https://github.com/camunda/camunda/pull/20089.

tmetzke (commented):
👍 Looks really good, nice job for the first endpoint, @nicpuppa, impressive! All integration code looks perfect, let's just tackle the default value and validation logic and we're good to go from my end ⭐

tmetzke (approved):
Whoop whoop 🚀 
Very well done, I like it 🌟

PR #20416, opened 2024-07-22
feat: create decision definition search service layer

houssain-barouni:
## Description

Create the service layer for Decision Definition search endpoint `POST /v2/decision-definitions/search` documented in [Camunda 8 REST API - Definition](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.dk8wwzs3pfou) doc
- `DecisionDefinitionFilter`:  has fields that are represented as lists, each corresponding to a supported filter.
- `DecisionDefinitionSort`: ~~as V1, supports sorting by all `DecisionDefinition` fields except       `decisionRequirementsName` and `decisionRequirementsVersion` see [v1 DecisionDefinitionController](https://github.com/camunda/camunda/blob/6512ff57906e95cbfb68bee29efbb381da1ba69d/operate/webapp/src/main/java/io/camunda/operate/webapp/api/v1/rest/DecisionDefinitionController.java#L54-L58)~~

## Related issues

relates to https://github.com/camunda/camunda/issues/20654

PR #20535, opened 2024-07-25
feat: Search decision definitions - controller implementation

houssain-barouni:
## Description

Implementation of the controller layer for the search decision definitions endpoint: `POST /v2/decision-definitions/search`. This continues from pull request [#20416](https://github.com/camunda/camunda/pull/20416). 
Assistance from an Operate engineer might be required to correctly describe the Query/Response `DecisionDefinition`
 fields in the OpenAPI spec.

## Related issues


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
```diff
@@ -7,43 +7,66 @@
...
-    response.setPage(page);
+    final var page = toSearchQueryPageResponse(result);
+    final var response =
+        new ProcessInstanceSearchQueryResponse()
+            .page(page)
+            .items(
+                ofNullable(result.items())
+                    .map(SearchQueryResponseMapper::toProcessInstances)
+                    .map(Either::get)
+                    .orElseGet(ArrayList::new));
```
tmetzke, line 40: 🙃 Just a nitpick but I guess we could also use the `Collections::emptyList` here instead of creating a new list instance? It's just a small thing though. There are more code lines like this in this PR.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
```diff
@@ -52,7 +75,7 @@ public static Either<ProblemDetail, List<ProcessInstanceItem>> toProcessInstance
             .toList());
   }
 
-  public static Either<ProblemDetail, ProcessInstanceItem> toProcessInstance(
+  private static Either<ProblemDetail, ProcessInstanceItem> toProcessInstance(
```
tmetzke, line 78: 🔧 We should only work with `Either` if it makes sense to keep the code simple.

That could be in the following cases:
1. The result can be either a success or failure (e.g. if validation takes place)
2. The specific method is used as a reference in a more generic call/structure and one of the other methods referenced could be a success or failure (e.g. `SearchQueryRequestMapper::toProcessInstanceFilter` that is used as a method reference and does itself not need the `Either` return but other filter methods might, so it's convenient to use it there).

I don't see any of the cases fulfilled here for `toProcessInstances` and `toProcessInstance`, also not for `toDecisionDefinitions` and `toDecisionDefinition`. There might be more of those here.

Please correct me if I overlooked anything.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  private ResponseEntity<DecisionDefinitionSearchQueryResponse> search(
+      final DecisionDefinitionQuery query) {
+    try {
+      final var tenantIds = TenantAttributeHolder.tenantIds();
+      final var result =
+          decisionDefinitionServices.withAuthentication((a) -> a.tenants(tenantIds)).search(query);
+      return ResponseEntity.ok(
+          SearchQueryResponseMapper.toDecisionDefinitionSearchQueryResponse(result).get());
```
tmetzke, line 47: ❌ Either `SearchQueryResponseMapper::toDecisionDefinitionSearchQueryResponse` can be a success or failure and then we should handle it accordingly, i.e. not simply return HTTP 200 and ignore a potential `Left` result. Or the method can never return a `Left` case and should thus not do it.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
```diff
@@ -0,0 +1,59 @@
...
+public class DecisionDefinitionController {
+
+  @Autowired private DecisionDefinitionServices decisionDefinitionServices;
+
+  @PostMapping(
+      path = "/decision-definitions/search",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<DecisionDefinitionSearchQueryResponse> searchDecisionDefinitions(
+      @RequestBody final DecisionDefinitionSearchQueryRequest query) {
```
tmetzke, line 35: 🔧 With this, the message body is required. That forces users to at least send an empty body. If we can work with an "undefined" query as well (and the tests imply that we can), I'd recommend setting the body to `required = false` and also testing that this works as expected.

tmetzke (changes requested):
👍 Integration into the controller structure looks good to me.
🔧 I added some smaller suggestions, mostly code style and `Either` usage.

⚠️ I haven't checked the validity of the data structure and search/filter options. I'd leave this to the feature teams to validate, e.g. @ralfpuchert.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
houssain-barouni: That's correct. The only "drawback" of using `Collections.emptyList `is its immutability. If someone tries to add an item to that list, it will cause an error. However, such an attempt would indicate a code smell. 
I'll make the change.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
houssain-barouni: I had the same thoughts when I saw that it is being used in `toProcessInstances`. I agree there is no reason to use it. I don't see a case where we would get errors when mapping the results to the DTO

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: I will remove the move of `Either` in this response mapper as discussed in the previous comment

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: it makes sense, I will update both `searchDecisionDefinitions` and `searchProcessInstances`

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
houssain-barouni: ✅

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: ✅

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: ✅

tmetzke (approved):
👍 Integration part looks good to me, pending a check of the domain-specific data (e.g. filter, request, and response attributes).

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
```diff
@@ -0,0 +1,59 @@
...
+    return SearchQueryRequestMapper.toDecisionDefinitionQuery(query)
+        .fold(this::search, RestErrorMapper::mapProblemToResponse);
+  }
+
+  private ResponseEntity<DecisionDefinitionSearchQueryResponse> search(
+      final DecisionDefinitionQuery query) {
+    try {
+      final var tenantIds = TenantAttributeHolder.tenantIds();
+      final var result =
+          decisionDefinitionServices.withAuthentication((a) -> a.tenants(tenantIds)).search(query);
```
tmetzke, line 45: 🔧 You can simplify this with the following:
```java
      final var result = decisionDefinitionServices
          .withAuthentication(RequestMapper.getAuthentication()).search(query);
```

The same can be refactored for the process instance search.

houssain-barouni: right, that's actually doing the same thing

houssain-barouni: it also creates a token, it is meant to be sent to the broker? @tmetzke

houssain-barouni: ✅

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
```diff
@@ -0,0 +1,60 @@
...
+              .search(query);
+      return ResponseEntity.ok(
+          SearchQueryResponseMapper.toDecisionDefinitionSearchQueryResponse(result));
+    } catch (final Throwable e) {
+      final var problemDetail =
+          RestErrorMapper.createProblemDetail(
+              HttpStatus.BAD_REQUEST,
+              e.getMessage(),
+              "Failed to execute Decision Definition Search Query");
+      return ResponseEntity.of(problemDetail)
```
tmetzke, line 55: 🔧 If I'm not mistaken, the REST error mapper already provides a method to do this.

tmetzke (commented):
👍 All good from my end.
🔧 Added a very small suggestion to avoid code duplication. This can also be adjusted in the process instance search.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: ✅

ralfpuchert (changes requested):
* Why aren't `decisionRequirementName` and `decisionRequirementVersion` fields included in `DefinitionDecisionSort`?
* The fields are also then missing in `SearchQueryRequestMapper::applyDecisionDefinitionSortField`

houssain-barouni:
> * Why aren't `decisionRequirementName` and `decisionRequirementVersion` fields included in `DefinitionDecisionSort`?
> * The fields are also then missing in `SearchQueryRequestMapper::applyDecisionDefinitionSortField`

Hi @ralfpuchert , as explained in https://github.com/camunda/camunda/pull/20416 : `V1` supports sorting by all DecisionDefinition fields except `decisionRequirementsName` and `decisionRequirementsVersion` see [v1 DecisionDefinitionController](https://github.com/camunda/camunda/blob/6512ff57906e95cbfb68bee29efbb381da1ba69d/operate/webapp/src/main/java/io/camunda/operate/webapp/api/v1/rest/DecisionDefinitionController.java#L54-L58)
Do you think this should different for `V2`?

ralfpuchert:
Hi @houssain-barouni, I think this was not intended in `v1`. For having a consistent behavior I think we should also have the two fields also sortable.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
```diff
@@ -0,0 +1,58 @@
...
+import io.camunda.zeebe.gateway.rest.SearchQueryRequestMapper;
+import io.camunda.zeebe.gateway.rest.SearchQueryResponseMapper;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+
+@ZeebeRestController
```
tmetzke, line 25: ❌ Needs to become a `CamundaRestController` and define a path, e.g. `@RequestMapping("/v2/)`.

tmetzke: This only works after a rebase. Otherwise, this will fail in the merge queue.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
```diff
@@ -0,0 +1,58 @@
...
+public class DecisionDefinitionController {
```
tmetzke, line 26: 🔧 For consistency, let's rename this to `DecisionDefinitionQueryController` so it's clear that this is a query controller.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionControllerTest.java
```diff
@@ -0,0 +1,272 @@
...
+import io.camunda.service.search.sort.DecisionDefinitionSort;
+import io.camunda.service.security.auth.Authentication;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.http.MediaType;
+
+@WebMvcTest(DecisionDefinitionController.class)
```
tmetzke, line 30: ❌ We need to enable the Query API for this test.
```suggestion
@WebMvcTest(
    value = DecisionDefinitionQueryController.class,
    properties = "camunda.rest.query.enabled=true"))
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: ok I will integrate this Query Controller concept

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionControllerTest.java
houssain-barouni: done here https://github.com/camunda/camunda/pull/20535/commits/d08ead22fcaead4f82cab05bfec8495a2f28534d

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: done here https://github.com/camunda/camunda/pull/20535/commits/d08ead22fcaead4f82cab05bfec8495a2f28534d

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionController.java
houssain-barouni: done here https://github.com/camunda/camunda/pull/20535/commits/d08ead22fcaead4f82cab05bfec8495a2f28534d

houssain-barouni:
> Hi @houssain-barouni, I think this was not intended in `v1`. For having a consistent behavior I think we should also have the two fields sortable.

Done in https://github.com/camunda/camunda/pull/20535/commits/3e5fdecc128cf9e4c48200be092fff9b5726407a and https://github.com/camunda/camunda/pull/20416/commits/e262f472cbb2fe1ba3f8e405fb2d619a667e0432

tmetzke (approved):
👍 Good to go from my end 🚀

ralfpuchert (approved):

Issue #20540, opened 2024-07-25
 C8 License - Create API endpoint to expose license status


matthewBearCamunda:
## Related to

Epic: https://github.com/camunda/product-hub/issues/2359

Dependent On: https://github.com/camunda/camunda/issues/18970

## Overview

A basic GET endpoint was developed with https://github.com/camunda/camunda/issues/18970. Expand on the basic structure and ensure that it adheres to the following

```
GET /license
{
   "validLicense": true,
   "environment": "self-managed"
}
```

The endpoint can be accessed through something like this `http://localhost:8080/v2/license`

- `validLicense` will be `true` or `false`. The value is determined whether the Camunda license is valid or not
- `environment` will return one of two strings, `saas` or `self-managed`. The value reflects the current running mode Camunda is in

Write tests for the `/license` endpoint to ensure its working as intended. 

Documentation for this new API endpoint will need to be made

## Actions

- The new endpoint will read from `ManagementService`, and will use that to gather license information
- Generate a new response entity for the API payload
- Extend the ZeebeClient with the new endpoint

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/21156
```

## Links

- Helpful guide for developing a new REST endpoint for the monorepo

tmetzke:
Hey @matthewBearCamunda, I have just looked at the REST API proposal.
For comparison, in C7, we have an [endpoint that fetches telemetry data](https://docs.camunda.org/rest/camunda-bpm-platform/7.21/#tag/Telemetry/operation/getTelemetryData). While this is of course way broader in scope than what we have here currently, I was wondering about the following:
- Do we only want to provide the fact that it is valid or would we want to return more details, e.g. until when the license is still valid?
- Do we want to hardwire `"selfManaged": true` or does a bit more flexibility help us in the longer run, e.g. `"environment": "self-managed"` which could also be `"environment": "saas"` or maybe `"environment": "hybrid"` at some point? That is a speculative question, I am aware. Just food for thought whether hardcoding "self-managed" is the best option here or not.
- Since this is the Camunda 8 REST API, would `GET /v2/license` suffice? Do we expect any other "license" endpoint that could otherwise collide here?

matthewBearCamunda:
@tmetzke - Going through your bullet points one by one

[...]

PR #20541, opened 2024-07-25
feat: perform license check on startup

### LINKED ISSUE #18970: C8 License - Log and Store C8 License Status
## Related to

Epic: https://github.com/camunda/product-hub/issues/2359

Dependent On: https://github.com/camunda/camunda/issues/18968

## Overview

Once the License generator and checker has been updated with the new fields, upgrade the monorepo to use the new version.

After the License checker has been upgraded, we need to log license information on startup. The license checker library will automatically log license status if it detects a C8 License. Ensure that the logging entry generated is properly outputted

The status of the license needs to be stored in a persistent variable inside `CamundaLicense`. Subsequent calls to check the status of the license should look at the persistent variable

License logging should only happen in self-managed mode, not in SaaS

Self-Managed and SaaS mode will be determined by the passed in license by environment variable. The a SaaS license will have a `licenseType` property with a value of `saas`. This SaaS license will be applied to all SaaS clusters with [to be controller created ticket]

The API endpoint will be developed with https://github.com/camunda/camunda/issues/20540

## Actions

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/20541
```

matthewBearCamunda:
## Description

On startup, perform a license check on the C8 license passed in via environment variable. Log the status of the license, and store the result of `validate()` in a persistent variable.  That variable will available to the `ManagementService` bean

The license checker library handles the logging when `validate()` is called

## Related issues

closes https://github.com/camunda/camunda/issues/18970


romansmirnov (dismissed):
@matthewBearCamunda, thanks for raising the PR. It should not be limited to environment variables to provide the license key. Instead, users should be able to provide the license key as they do with any other configuration property (i.e., inside the `application.yaml`, as an environment variable, as a system variable, etc.). Therefore, we should use the Spring capabilities to resolve the configuration property; however, the user has provided it.

This can be achieved, for example, like this:

```java
public class CamundaLicense {

  private final String license;
  private boolean isInitialized;
  private boolean isValid;

  public CamundaLicense(final String license) {
    this.license = license;
  }

  public boolean isValid() {
    initializeLicenseKey();
    return isValid;
  }

  protected void initializeLicenseKey() {
    if (!isInitialized) {
      // IMO; the implementation should be thread-safe
      synchronized (this) {
        if (!isInitialized) {
         // no need to determine it multiple times, as the license key cannot change during runtime
          isValid = determineLicenseValidity(license);
          isInitialized = true;
        }
      }
    }
  }

  protected boolean determineLicenseValidity(final String licenseStr) {
    try {
      final LicenseKey license = getLicenseKey(licenseStr);
      license.validate(); // this method logs the license status
      return true;
    } catch (final Exception e) {
      return false;
    }
  }

  protected LicenseKey getLicenseKey(final String licenseStr) throws InvalidLicenseException {
    return new LicenseKeyImpl(licenseStr);
  }
}
```

The `ManagementService` could look like this

```java
public final class ManagementService {

  private CamundaLicense camundaLicense;

  public ManagementService withLicense(final CamundaLicense license) {
    camundaLicense = license;
    return this;
  }

  public boolean isCamundaLicenseValid() {
    return camundaLicense != null ? camundaLicense.isValid() : false;
  }
}
```

and the `CamundaServicesConfiguration` could look like this

```java
@ConditionalOnProperty(
    name = {"zeebe.broker.gateway.enable", "camunda.rest.enabled"},
    havingValue = "true",
    matchIfMissing = true)
@EnableConfigurationProperties(LicenseKeyConfiguration.class)
public class CamundaServicesConfiguration {

  private final BrokerClient brokerClient;
  private final CamundaSearchClient camundaSearchClient;
  private final LicenseKeyProperties licenseKeyProperties;

  @Autowired
  public CamundaServicesConfiguration(
      final BrokerClient brokerClient, final CamundaSearchClient camundaSearchClient, final LicenseKeyProperties licenseKeyProperties) {
    this.brokerClient = brokerClient;
    this.camundaSearchClient = camundaSearchClient;
    this.licenseKeyProperties= licenseKeyProperties;
  }

  ...

  @Bean
  public ManagementService managementService(final CamundaServices camundaServices, final CamundaLicense camundaLicense) {
    final var managementService = camundaServices.managementService().withLicense(camundaLicense);
    // trigger to log if license is valid or not during startup 
    managementService.isCamundaLicenseValid();
    return managementService;
  }

  @Bean
  public CamundaLicense camundaLicense() {
    final var key = licenseKeyProperties.key();
    return new CamundaLicense(key);
  }

  @ConfigurationProperties("camunda.license")
  public static final record LicenseKeyProperties(String key) {}

}
```

That way, Spring will create the `LicenseKeyConfiguration` containing the license key, if present; otherwise, the value will be `null`. (Please note: I didn't test the code, it should show the direction to go.)

Another comment: I assume that the license check should always be done at the startup. With the current implementation, this will only occur when the REST API is started on that node. To achieve that, the license check on startup should happen in a bean/configuration that always gets initialized, independent of the management service. Maybe there could be a class `CamundaLicenseConfiguration` that creates the `CamundaLicense` key that can be used by the management service, but the initial check happens in `CamundaLicenseConfiguration`.

I haven't checked the test cases for now because I would advise changing the implementation first.

service/src/main/java/io/camunda/service/ManagementServices.java
```diff
@@ -9,15 +9,16 @@
...
-  public ManagementService() {
-    camundaLicense = new CamundaLicense();
+  public ManagementServices withLicense(final CamundaLicense license) {
+    camundaLicense = license;
+    return this;
   }
 
   public boolean isCamundaLicenseValid() {
-    return camundaLicense.isValid();
+    return camundaLicense != null && camundaLicense.isValid();
```
Zelldon, line 22: ❓ If it is null is it really valid? Do we maybe need an enum here to indicate DEV and prod? 🤔 And default to dev? 🤔

dist/src/main/java/io/camunda/application/commons/service/ManagementServicesConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+      final ClientModeProperty clientModeProperty) {
+    this.licenseKeyProperties = licenseKeyProperties;
+    this.clientModeProperty = clientModeProperty;
+  }
+
+  @Bean
+  public ManagementServices managementService(final CamundaLicense camundaLicense) {
+    final ManagementServices managementServices =
+        new ManagementServices().withLicense(camundaLicense);
+    // trigger to log if license is valid or not during startup
```
Zelldon, line 40: ❓ Is this a todo? Because the method below does just a check?

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -7,41 +7,50 @@
...
+      }
+    }
   }
 
-  private boolean isLicenseInitialized() {
-    return StringUtils.isNotBlank(licenseStr);
+  protected boolean determineLicenseValidity(final String licenseStr) {
+    try {
+      final LicenseKey license = getLicenseKey(licenseStr);
+      license.validate(); // this method logs the license status
```
Zelldon, line 46: 🔧 Sorry but I feel this is really intransparent from the caller and a bit too deep for the specific concern that we want to log the license type. Furthermore, I feel like we are coupling two things here, which speaks against [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle). To me it seems we want to validate (not sure 100% what it means here tbh) and to log specific details about the license.

I think one should do the validation (or even give back what kind of license the key is either DEV or PROD, if not given default to DEV), and then outside we could log it.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/LicenseController.java
```diff
@@ -0,0 +1,29 @@
...
+import io.camunda.zeebe.gateway.rest.ConditionalOnRestGatewayEnabled;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@ConditionalOnRestGatewayEnabled
+@RestController
+@RequestMapping(path = {"/v1", "/v2"})
```
Zelldon, line 20: ❓ Why we adding this to both versions? 🤔 Is this expected? Leaving this to @houssain-barouni I think he is more into the REST API topic.

tmetzke (commented):
Just had time for a quick review, maybe that helps already to move forward.
@houssain-barouni, would you be able to review this as well?

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -7,27 +7,38 @@
...
-  public void shouldReturnTrueWhenLicenseIsValid() {
-    // given
-    final CamundaLicense camundaLicense =
-        new CamundaLicense(new MockEnvironmentVariableReaderWithValidLicense());
+  public void shouldReturnTrueWhenLicenseIsValid() throws InvalidLicenseException {
+    final CamundaLicense testLicense = spy(CamundaLicense.class);
+    final LicenseKey mockKey = mock(LicenseKey.class);
 
-    assertTrue(camundaLicense.isValid());
+    doReturn(mockKey).when(testLicense).getLicenseKey(anyString());
```
Zelldon, line 29: 💭 🔧 Wondering whether it makes sense to test the protected methods here. I feel like this is too much implementation detail, we should rather focus on the API (public methods) and behavior of classes. 

Meaning, that if I create a CamundaLicense with a specific key and I call it valid I expect it is either valid or not (depending on the input). If I can't change the behavior of the class and test it appropriately, it might based on the class design. So we could improve the design to be able to inject dependencies, like a LicenseKey factory or something.

Just me 2ct to this.

Zelldon: Why does this matter? The test is now coupled to the structure of the class. If I change any implementation detail of the class, there is a high chance that the tests are broken, even if the behavior of the class didn't even change.

ChrisKujawa (commented):
Hey @matthewBearCamunda I took some time to read the EPICs you have sent me, to give you some review and to unblock you. 

I might missed some parts (I tried to read the important bits, and pre PRs). I had some comments, questions and suggestions please see below. 

To continue here it might also make sense, as I mentioned, to sync with @houssain-barouni as he is also more familiar with the web apps and the REST API and I'm currently focusing on other parts (Importer/Operate performance). If he has no time, I try my best to give you input (when I find the time).

I hope this helps already. 🙇🏼

service/src/main/java/io/camunda/service/license/SaaSLicense.java
```diff
@@ -7,8 +7,12 @@
  */
 package io.camunda.service.license;
 
-public class EnvironmentVariableReader {
-  public String getEnvironmentVariableValue(final String envVarName) {
-    return System.getenv().getOrDefault(envVarName, "");
+public class SaaSLicense extends CamundaLicense {
```
houssain-barouni, line 10: `SaaSLicense` has nothing in common with `CamundaLicense` expect overriding `isValid()` method
I think it is better to extract an interface `CamundaLicense` with `isValid()` method and create a `SelfManagedCamundaLicense` that implements it with all the license checks, and create a second implementation of `CamundaLicense`: `SaaSLicense` or `NoopLicense` that just implements  `isValid()`.
This way you avoid `SaaSLicense` accidentelly having access to public or protected methods of `CamundaLicense`.

service/src/main/java/io/camunda/service/ManagementServices.java
```diff
@@ -9,15 +9,16 @@
...
 
-public final class ManagementService {
+public final class ManagementServices {
 
-  private final CamundaLicense camundaLicense;
+  private CamundaLicense camundaLicense;
 
-  public ManagementService() {
-    camundaLicense = new CamundaLicense();
+  public ManagementServices withLicense(final CamundaLicense license) {
```
houssain-barouni, line 16: better use the constructor to make `ManagementServices` immutable and avoid nullity checks on `camundaLicense` (I don't know in which functional case it can be null?)

if you want to use `withLicense` pattern, you can do 
```java
public ManagementServices withLicense(final CamundaLicense license) {
return new ManagementServices(license);
}

public ManagementService() {
}

public ManagementService(CamundaLicense license) {
    camundaLicense= license;
}
```

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -7,41 +7,50 @@
...
 
-  /**
-   * Used for testing. Test can mock EnvironmentVariableReader to test out different environment
-   * variable values
-   */
-  public CamundaLicense(final EnvironmentVariableReader environmentVariableReader) {
-    this.environmentVariableReader = environmentVariableReader;
-  }
+  // for testing
+  public CamundaLicense() {}
```
houssain-barouni, line 25: you can make package visible constructor and annotate it with `@VisibleForTesting`

service/src/main/java/io/camunda/service/CamundaServices.java
```diff
@@ -52,10 +52,6 @@ public IncidentServices incidentServices() {
     return new IncidentServices(brokerClient, searchClient, transformers, authentication);
   }
 
-  public ManagementService managementService() {
```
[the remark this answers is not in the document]
matthewBearCamunda: I took it out of `CamundaServices` since `ManagementService` will not be a part of where `CamundaServices` puts services into a java beans in `CamundaServicesConfiguration`. It can't go in `CamundaServicesConfiguration` since that config class is conditional based on is REST is enabled or not. The license information needs to be accessible in any situation. This is why I have `ManagementServicesConfiguration` making the `ManagementService` bean

My thought was that by putting it in a separate configuration class, it was more consistent. The configuration classes `CamundaServicesConfiguration`and `ManagementServicesConfiguration` are responsible solely for the services that they cover. I did not want to split where `CamundaService` is but into a bean into two different config classes

service/src/main/java/io/camunda/service/license/CamundaLicense.java
matthewBearCamunda: Done! I didn't know this annotation was available

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/LicenseController.java
```diff
@@ -0,0 +1,29 @@
...
+public class LicenseController {
```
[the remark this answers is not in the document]
matthewBearCamunda: Works for me, I'll remove it for this PR!

service/src/main/java/io/camunda/service/ManagementServices.java
matthewBearCamunda: I removed this null check completely, the license itself won't ever be `null` with how `managementServices` is being created

dist/src/main/java/io/camunda/application/commons/service/ManagementServicesConfiguration.java
matthewBearCamunda: It was a comment explaining what would do the logging

I removed it since I refactored how the license key is initalized, and where the logging is coming from. Its now happening when the license is being created

service/src/main/java/io/camunda/service/ManagementServices.java
matthewBearCamunda: I also started a conversation around dev/prod licenses, there seems to be a discrepancy if they are needed or not

https://camunda.slack.com/archives/C06V2SR6U9G/p1722973721151959

service/src/main/java/io/camunda/service/license/CamundaLicense.java
matthewBearCamunda: As for the dev/prod licenses, same response as above. There seems to be a discrepancy around dev/prod licenses

https://camunda.slack.com/archives/C06V2SR6U9G/p1722973721151959

dist/src/main/java/io/camunda/application/commons/service/ManagementServicesConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+    managementServices.isCamundaLicenseValid();
+
+    return managementServices;
+  }
+
+  @Bean
+  public CamundaLicense camundaLicense() {
+    final String clientMode = clientModeProperty.key();
+    if (SELF_MANAGED_MODE.equalsIgnoreCase(clientMode)) {
```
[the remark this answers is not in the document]
matthewBearCamunda: I'll check with the PM, but I'll go with that approach for now

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
matthewBearCamunda: Makes sense! I refactored the tests to work with the public methods. What's mocked is just the call to get the license

I ran out of time for the day to finish the other two tests, but I'll get those written up tomorrow

matthewBearCamunda: Tests pushed!

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -7,41 +7,50 @@
...
+          isValid = determineLicenseValidity(license);
+          isInitialized = true;
+        }
+      }
+    }
   }
 
-  private boolean isLicenseInitialized() {
-    return StringUtils.isNotBlank(licenseStr);
+  protected boolean determineLicenseValidity(final String licenseStr) {
```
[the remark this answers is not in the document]
matthewBearCamunda: I refactored how the tests work, its testing the public method now. I also put the `@VisibleForTesting` annotation to work in `CamundaLicense`

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/LicenseController.java
tmetzke: **Note**: The controller is removed from this PR here. It will come with the next PR related to https://github.com/camunda/camunda/issues/20540.

Nonetheless, some implementation hints:
* Please only add it to V2. We should not backport this to V1 of the REST API. New endpoints and features should only be added to V2.
* Please use the `@CamundaRestController` annotation instead of `@ConditionalOnRestGatewayEnabled` and `@RestController`. If we are to change annotations for all REST controllers, we can do that centrally using `@CamundaRestController`.
* @houssain-barouni, only the C8 REST Query API (search endpoints) are not working by default with 8.6., i.e. the controllers annotated with `CamundaRestQueryController`.

dist/src/main/java/io/camunda/application/commons/service/ManagementServicesConfiguration.java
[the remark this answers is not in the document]
tmetzke: As mentioned in the Slack message, if we tie the license check to this property, all you have to do in SM to avoid a license check is set the property to `saas` 🔐

service/src/main/java/io/camunda/service/ManagementServices.java
matthewBearCamunda: From the linked slack conversation, the distinction between dev/prod licenses does not matter. The keys we will be issuing will always be production liceneses

service/src/main/java/io/camunda/service/license/CamundaLicense.java
matthewBearCamunda: After looking into this some more, I'd argue that the current approach is what we want

One of the requirements for [the epic](https://github.com/camunda/product-hub/issues/2286) is: `As a Customer, my license information is logged as soon as C8 component (One Jar, Web Modeller or Console) starts.`

So instead of each component having to implement the logging capability on their own, the idea was to include this directly in the library - this has the advantage that the logging output is always consistent across applications.

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -7,49 +7,76 @@
...
+  private boolean determineIfLicenseEnvModeIsSelfManaged(final String licenseStr) {
+    try {
+      final LicenseKey licenseKey = getLicenseKey(licenseStr);
+      return licenseKey.getProperties().entrySet().stream()
+          .noneMatch(x -> x.getKey().equals("environmentMode") && x.getValue().equals("saas"));
+    } catch (final InvalidLicenseException e) {
+      return true;
+    } catch (final Exception e) {
+      LOGGER.error("Unexpected error when determining environment mode of license", e);
+      return true;
```
Zelldon, line 57: 💭 Wondering whether we maybe should explode/fail in this case instead of continuing? Is it expected that the licenseStr is invalid? 🤔

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -7,49 +7,76 @@
...
+    }
+  }
+
+  private boolean determineLicenseValidity(final String licenseStr) {
     try {
       final LicenseKey license = getLicenseKey(licenseStr);
       license.validate(); // this method logs the license status
       return true;
+    } catch (final InvalidLicenseException e) {
+      return false;
```
Zelldon, line 67: 🤔 The same here. This would mean we would fallback to dev environment right?

Please log this case with a bit more context and the exception. It might be unexpected to the user that the license doesn't work.

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -7,49 +7,76 @@
...
 
-  protected void initializeLicenseKey() {
+  public boolean isSelfManaged() {
+    return isSelfManaged;
+  }
+
+  public void initializeWithLicense(final String license) {
     if (!isInitialized) {
       synchronized (this) {
         if (!isInitialized) {
```
Zelldon: 🔧 I think we can also simply synchronize the method, I doubt we have here a performance bottleneck (we of course have no data yet to see it). Thus avoids the boilerplate code and potential issues of double-checked locking see https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -7,49 +7,76 @@
...
+
+  private boolean determineLicenseValidity(final String licenseStr) {
     try {
       final LicenseKey license = getLicenseKey(licenseStr);
       license.validate(); // this method logs the license status
       return true;
+    } catch (final InvalidLicenseException e) {
+      return false;
     } catch (final Exception e) {
+      LOGGER.error("Unexpected error when validating license", e);
```
Zelldon, line 69: 🔧 I always recommend using our error guidelines to write good and unified error log messages https://github.com/camunda/camunda/wiki/Error-Guidelines This helps our users if they follow a similar pattern.

Example would be here: `Expected to validate SM license 'x', but was invalid. Found following errors: x, y, z.` Always good to have a reason why it is failing, or in general more context so the user knows what to do or why it is failing (error messages should at least be actionable otherwise it is just noise).

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -22,23 +24,76 @@
 public class CamundaLicenseTest {
 
   @Test
-  public void shouldReturnTrueWhenLicenseIsValid() throws InvalidLicenseException {
+  public void isValidShouldReturnTrueWhenLicenseIsValid() throws InvalidLicenseException {
```
Zelldon, line 27: 🔧 Please lets still follow our pattern of using `should*`

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -7,27 +7,93 @@
...
+    final CamundaLicense testLicense = spy(CamundaLicense.class);
+
+    Mockito.doThrow(new InvalidLicenseException("test exception"))
+        .when(testLicense)
+        .getLicenseKey(anyString());
+
+    testLicense.initializeWithLicense("whatever");
+
+    assertFalse(testLicense.isValid());
   }
```
Zelldon, line 49:
```suggestion
  @Test
  public void shouldReturnFalseWhenLicenseIsInvalid() throws InvalidLicenseException {
    // given
    final CamundaLicense testLicense = spy(CamundaLicense.class);

    Mockito.doThrow(new InvalidLicenseException("test exception"))
        .when(testLicense)
        .getLicenseKey(anyString());

    // when
    testLicense.initializeWithLicense("whatever");
    
    // then
    assertFalse(testLicense.isValid());
  }
```

Zelldon: We normally have a clearer separation of `given, when, then` via comments it helps the reader to better understand such tests.

dist/src/main/java/io/camunda/application/commons/service/ManagementServicesConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+  @Bean
+  public ManagementServices managementService(final CamundaLicense camundaLicense) {
+    return new ManagementServices(camundaLicense);
+  }
+
+  @Bean
+  public CamundaLicense camundaLicense() {
+    final CamundaLicense licence = new CamundaLicense();
+    licence.initializeWithLicense(licenseKeyProperties.key());
+    return licence;
```
Zelldon, line 38:
```suggestion
    final CamundaLicense license = new CamundaLicense();
    license.initializeWithLicense(licenseKeyProperties.key());
    return license;
```

Zelldon: ❓ Any specific reason why we have a separate method to init the license? Couldn't we just use the constructor? Is it expected to initialize the license multiple times? 🤔 What happens when I don't init the license?

service/src/main/java/io/camunda/service/license/CamundaLicense.java
Zelldon: > this has the advantage that the logging output is always consistent across applications.

I see your point. I think this can also be argued in the other direction that depending on the component you might have different logging frameworks or details you want to add to the logging, so I would expect this is done outside of a lib. 

But anyway I don't think it is a blocker, just want to raise that we seem to couple here two features and that might be sub-optimal. I will leave the decision to you.

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -7,27 +7,93 @@
...
+    final Map<String, String> testProperties = new HashMap<>();
+    testProperties.put("environmentMode", "self-managed");
+    when(mockKey.getProperties()).thenReturn(testProperties);
+
+    Mockito.doReturn(mockKey).when(testLicense).getLicenseKey(anyString());
+
+    testLicense.initializeWithLicense("whatever");
+
+    assertTrue(testLicense.isSelfManaged());
+  }
```
Zelldon, line 66:
```suggestion
  @Test
  public void shouldDetectSelfManagedWhenEnvPropertyIsSet()
      throws InvalidLicenseException {
    // given  
    final CamundaLicense testLicense = spy(CamundaLicense.class);
    final LicenseKey mockKey = mock(LicenseKey.class);

    final Map<String, String> testProperties = new HashMap<>();
    testProperties.put("environmentMode", "self-managed");
    when(mockKey.getProperties()).thenReturn(testProperties);

    Mockito.doReturn(mockKey).when(testLicense).getLicenseKey(anyString());

    // when
    testLicense.initializeWithLicense("whatever");

    // then
    assertTrue(testLicense.isSelfManaged());
  }
```

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -7,27 +7,93 @@
...
+
+    final Map<String, String> testProperties = new HashMap<>();
+    when(mockKey.getProperties()).thenReturn(testProperties);
+
+    Mockito.doReturn(mockKey).when(testLicense).getLicenseKey(anyString());
+
+    testLicense.initializeWithLicense("whatever");
+
+    assertTrue(testLicense.isSelfManaged());
+  }
```
Zelldon, line 82:
```suggestion
  public void shouldDetectSelfManagedWhenEnvPropertyDoesNotExist()
      throws InvalidLicenseException {
    // given
    final CamundaLicense testLicense = spy(CamundaLicense.class);
    final LicenseKey mockKey = mock(LicenseKey.class);

    final Map<String, String> testProperties = new HashMap<>();
    when(mockKey.getProperties()).thenReturn(testProperties);

    Mockito.doReturn(mockKey).when(testLicense).getLicenseKey(anyString());

    // when
    testLicense.initializeWithLicense("whatever");

    // then 
    assertTrue(testLicense.isSelfManaged());
  }
```

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -7,27 +7,93 @@
...
+    final Map<String, String> testProperties = new HashMap<>();
+    testProperties.put("environmentMode", "saas");
+    when(mockKey.getProperties()).thenReturn(testProperties);
+
+    Mockito.doReturn(mockKey).when(testLicense).getLicenseKey(anyString());
+
+    testLicense.initializeWithLicense("whatever");
+
+    assertFalse(testLicense.isSelfManaged());
   }
```
Zelldon, line 98:
```suggestion
  @Test
  public void shouldNotDetectSelfManagedWhenEnvIsSetToSaas() throws InvalidLicenseException {
    // given
    final CamundaLicense testLicense = spy(CamundaLicense.class);
    final LicenseKey mockKey = mock(LicenseKey.class);

    final Map<String, String> testProperties = new HashMap<>();
    testProperties.put("environmentMode", "saas");
    when(mockKey.getProperties()).thenReturn(testProperties);

    Mockito.doReturn(mockKey).when(testLicense).getLicenseKey(anyString());

    // when
    testLicense.initializeWithLicense("whatever");

    // then
    assertFalse(testLicense.isSelfManaged());
  }
```

Zelldon: i think would be good to verify that the license is valid in this case

ChrisKujawa (approved):
Thanks Matt. I just have smaller comments but I think after resolving them we should be good to go.

service/src/main/java/io/camunda/service/license/CamundaLicense.java
matthewBearCamunda: When encountering an invalid license, we don't blow up the startup process but rather continue on and mark the license as invalid. I'd want to treat an unexpected error the same, there is something wrong with the license, and we should consider it invalid

service/src/main/java/io/camunda/service/license/CamundaLicense.java
matthewBearCamunda: Log added!

And yes, when a license is invalid, its will be presented as non-production to the user

PR #20553, opened 2024-07-26
Add a job throw error endpoint

### LINKED ISSUE #20082: Migrate the Throw error gRPC API to REST
## Description

* The Camunda 8 REST API contains a job error endpoint at `POST /v2/jobs/{jobKey}/error` providing the same functionality as the [gRPC endpoint](https://docs.camunda.io/docs/apis-tools/zeebe-api/gateway-service/#throwerror-rpc).
* The endpoint is featured in the OpenAPI description as (internally) [described](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.9kvife3x49bp).
* The endpoint allows sending errorCode, errorMessage, and variables (as a JSON object).

nicpuppa:
## Description

Add an endpoint to the REST API for job throw error.

~⚠️ This PR depends on github.com/camunda/camunda/pull/20523~

## Related issues

closes #20082

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -110,7 +110,7 @@ paths:
     post:
       tags:
         - Job
-      summary: Fail job
+      summary: Error job
       description: |
         Reports a business error (i.e. non-technical) that occurs while processing a job.
       parameters:
@@ -1431,6 +1431,7 @@ components:
         worker:
           description: the name of the worker activating the jobs, mostly used for logging purposes
           type: string
+          nullable: true
         timeout:
           description: >
             a job returned after this call will not be activated by another call until the
@@ -1446,6 +1447,7 @@ components:
             a list of variables to fetch as the job variables; if empty, all visible variables at
             the time of activation for the scope of the job will be returned
           type: array
+          nullable: true
           items:
             type: string
         requestTimeout:
@@ -1456,11 +1458,18 @@ components:
             immediately, even when no job is activated.
           type: integer
           format: int64
+          default: 0
+          nullable: true
         tenantIds:
           description: a list of IDs of tenants for which to activate jobs
           type: array
           items:
             type: string
+          nullable: true
+      required:
+        - type
+        - timeout
+        - maxJobsToActivate
     JobActivationResponse:
       description: The list of activated jobs
       type: object
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1572,6 +1572,8 @@ components:
             JSON object that will instantiate the variables at the local scope of the error catch event that catches the thrown error.
           type: object
           nullable: true
+      required:
+        - errorCode
 
     ProblemDetail:
       description: >
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -106,6 +106,59 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+  /jobs/{jobKey}/error:
+    post:
+      tags:
+        - Job
+      summary: Fail job
+      description: |
+        Reports a business error (i.e. non-technical) that occurs while processing a job.
+      parameters:
+        - name: jobKey
+          in: path
+          required: true
+          description: The key of the job.
+          schema:
+            type: integer
+            format: int64
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/JobErrorRequest"
+      responses:
+        '204':
+          description: An error is thrown for the job.
+        '400':
+          description: >
+            The provided data is not valid.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        '404':
+          description: >
+            The job with the given jobKey is not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        '409':
+          description: >
+            The job with the given key is in the wrong state currently.
+            More details are provided in the response body.
+          content:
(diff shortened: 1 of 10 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1501,6 +1554,24 @@ components:
...
+          description: >
+            An error message that provides additional context.
+          type: string
+          nullable: true
+        variables:
+          additionalProperties: true
+          description: >
+            JSON object that will instantiate the variables at the local scope of the error catch event that catches the thrown error.
+          type: object
+          nullable: true
```
tmetzke, line 1574: 🔧 We can mark the `errorCode` as required as well. I think this is the syntax (please double check):
```suggestion
          nullable: true
      required:
        - errorCode
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/ErrorMessage.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.rest.validator;
+
+public final class ErrorMessage {
```
tmetzke, line 10: 🔧 Since this is a collection of error messages, should we call it `ErrorMessages`?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nicpuppa: Should we do the same also for `JobActivation` ?

tmetzke: Oh, yeah, good point 👍 I also just learned about it recently and didn't know it back then. I think the type, maxJobsToActivate, and timeout are required to be provided by the user. Any remaining numeric attributes can actually receive a default 0, I think. Feel free to make a proposal. Otherwise, we can also handle this in a follow-up issue.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/JobRequestValidator.java
```diff
@@ -0,0 +1,65 @@
...
+      final JobErrorRequest errorRequest) {
+    final List<String> violations = new ArrayList<>();
+    // errorCode can't be null or empty
+    if (errorRequest.getErrorCode() == null || errorRequest.getErrorCode().isBlank()) {
+      violations.add(ERROR_MESSAGE_EMPTY_ATTRIBUTE.formatted("errorCode"));
+    }
+    return createProblemDetail(violations);
+  }
+
+  private static Optional<ProblemDetail> createProblemDetail(final List<String> violations) {
```
tmetzke, line 58: 🔧 How about we move this helper to the general request validator class? It can be nicely reused by the specific validators then.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/RequestValidator.java
```diff
@@ -0,0 +1,65 @@
...
+import java.time.format.DateTimeParseException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ProblemDetail;
+
+public final class RequestValidator {
+
+  public static Optional<ProblemDetail> validateUpdateRequest(
```
tmetzke, line 27: 🔧 This is rather specific to user tasks, actually. I know it looks general, but it isn't. At least not yet. Let's move it to the user task validator for now.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/RequestValidator.java
```diff
@@ -0,0 +1,65 @@
...
+      final Changeset changeset = updateRequest.getChangeset();
+      validateDate(changeset.getDueDate(), "due date", violations);
+      validateDate(changeset.getFollowUpDate(), "follow-up date", violations);
+    }
+    if (violations.isEmpty()) {
+      return Optional.empty();
+    }
+    return Optional.of(
+        RestErrorMapper.createProblemDetail(
+            HttpStatus.BAD_REQUEST, String.join(" ", violations), INVALID_ARGUMENT.name()));
```
tmetzke, line 44: 🔧 How about we make the error messages consistent (e.g. always ending with a `.`), move the validation error message joining and problem detail creation to the general validator (e.g. always joining with a whitespace), and reuse the method in all request mappers and validators? The search controllers also do their own thing, I think, which makes it all inconsistent. Can also be moved into a follow up, if you think it's too much for this PR.

tmetzke (changes requested):
👍 Looks really good already, @nicpuppa, very nice job. Functionally, we're all good here already🚀 

🔧 I added some smaller suggestions and some refactoring ideas. We can add them here or create follow-up issues, as you like.

❌ The summary of the new endpoint in the YAML is not correct, let's fix that.

tmetzke (approved):
👍 Great refactorings and generalizations, job error functionality was already as expected previously 🚀

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/JobController.java
```diff
@@ -60,6 +62,16 @@ public CompletableFuture<ResponseEntity<Object>> failureJob(
         .fold(this::failJob, RestErrorMapper::mapProblemToCompletedResponse);
   }
 
+  @PostMapping(
+      path = "/jobs/{jobKey}/error",
```
tmetzke, line 66:
```suggestion
      path = "/{jobKey}/error",
```

PR #20557, opened 2024-07-26
refactor: remove import status reporting

HeleneW-dot:
## Description

- removes C7 import status reporting api and websocket from Optimize 8
- keeps status checking service related to health rest endpoint, but reduces it to only check for DB connection
- adjusted in [status api docs](https://confluence.camunda.com/pages/viewpage.action?pageId=56788227#) and health [docs](https://confluence.camunda.com/display/CO/Health)

Apologies for the style changes, had not yet disabled this.

## FE follow up

- [x] remove import status reporting in UI (already hidden in C8) with [OPT#13439](https://github.com/camunda/camunda-optimize/issues/13439) -- check with FE before merging that this can be done as a follow up (does anything in Fe still call status API in C8 despite the UI being hidden?)

## Related issues

related to [OPT-13211](https://github.com/camunda/camunda/issues/33560) part 2


optimize/backend/src/it/java/io/camunda/optimize/websocket/AssertHasChangedStatusClientSocket.java
```diff
@@ -1,52 +0,0 @@
...
-/**
- * Client class to test Web Socket implementation of status report is working. This class will
- * assert 2 properties:
- *
- * <p>1. import status has changed 2. more then one message is received
- */
-@ClientEndpoint
-@Slf4j
-@Getter
-public class AssertHasChangedStatusClientSocket {
```
RomanJRW, line 30: I am glad we get rid of this class 🙃

optimize/backend/src/main/java/io/camunda/optimize/rest/HealthRestService.java
```diff
@@ -27,8 +27,7 @@ public class HealthRestService {
 
   @GET
   public Response getConnectionStatus() {
-    if (statusCheckingService.isConnectedToDatabase()
-        && statusCheckingService.isConnectedToAtLeastOnePlatformEngineOrCloud()) {
+    if (statusCheckingService.isConnectedToDatabase()) {
```
RomanJRW, line 30: nice that you simplified the health check logic for C8 already here 👍

optimize/backend/src/main/java/io/camunda/optimize/rest/security/platform/PlatformSecurityConfigurerAdapter.java
```diff
@@ -143,7 +142,6 @@ protected SecurityFilterChain configureWebSecurity(HttpSecurity http) {
                     // common public api resources
                     .requestMatchers(
                         new AntPathRequestMatcher(createApiPath(READYZ_PATH)),
-                        new AntPathRequestMatcher(createApiPath(STATUS_PATH)),
```
RomanJRW, line 146: no problem with changing this class, but also want to make sure that it is on your radar that this class can be deleted completely?

optimize/backend/src/main/java/io/camunda/optimize/service/status/StatusCheckingServiceES.java
```diff
@@ -7,55 +7,41 @@
...
-      isConnected =
-          healthResponse.status().getStatus() == Response.Status.OK.getStatusCode()
-              && healthResponse.getStatus() != ClusterHealthStatus.RED;
+      return healthResponse.status().getStatus() == Response.Status.OK.getStatusCode()
+          && healthResponse.getStatus() != ClusterHealthStatus.RED;
     } catch (Exception ignored) {
       // do nothing
     }
-    return isConnected;
+    return false;
```
RomanJRW, line 45: should this be in the catch block?

optimize/backend/src/main/java/io/camunda/optimize/service/status/StatusCheckingServiceOS.java
```diff
@@ -7,50 +7,37 @@
...
-    boolean isConnected = false;
     try {
       final HealthResponse clusterHealthResponse =
           osClient.getOpenSearchClient().cluster().health(new HealthRequest.Builder().build());
       return clusterHealthResponse.status() != HealthStatus.Red;
     } catch (Exception ignored) {
       // do nothing
     }
-    return isConnected;
+    return false;
```
RomanJRW, line 41: same re being in the catch block

RomanJRW (changes requested):
Good progress, but there is still more to cover here:

- To delete:
  - `OptimizeRequestExecutor#buildCheckSharingStatusRequest` - two methods with this name
  - `OptimizeResourceConstants.STATUS_WEBSOCKET_PATH`, and everywhere this is called from
  - `LicenseFilter#isStatusRequest`
  - `LicenseFilter#isCloudEnvironment` - not strictly related but the answer is always true here
  - We declare the following dependencies. I do not know that they can be deleted, you might want to trace back their insertions to make sure we are not losing behaviour by removing these.
    - `spring-websocket`
    - `jetty-ee10-websocket-jakarta-server`
    - `jetty-websocket-jetty-client`
  - We have translations in each file for the key `connectionError`. Can these be deleted?
- To consider:
  - The License filter and all logic can be deleted. Do we have an issue for that?
  - The Platform Security Adaptor can be deleted too, as we never run in Platform mode here

optimize/backend/src/main/java/io/camunda/optimize/rest/security/platform/PlatformSecurityConfigurerAdapter.java
PHWaechtler: yup, anything C7 only and anything that uses CamundaPlatformCondition (incl the condition itself) will be removed ♻️ the adjustment here was just since I deleted the constant

optimize/backend/src/main/java/io/camunda/optimize/service/status/StatusCheckingServiceES.java
PHWaechtler: Adjusted this to match the OS one, can do it this way around too though

PHWaechtler:
- [ ]  OptimizeRequestExecutor#buildCheckSharingStatusRequest - two methods with this name
-- So buildCheckSharingStatusRequest is a different thing using SharingRestService and used in share IT - imo will reuse again once migrated to C8 IT. Removed the other one though

- [ ] OptimizeResourceConstants.STATUS_WEBSOCKET_PATH, and everywhere this is called from

- [ ] LicenseFilter#isStatusRequest

- [ ] LicenseFilter#isCloudEnvironment - not strictly related but the answer is always true here
--I guess isLicenseCheckNeeded will just always be false now. Matt mentioned the entire file will be thrown out anyway so left the rest as is for now (license epic [here](https://github.com/camunda/product-hub/issues/2359))

- [ ] We declare the following dependencies. I do not know that they can be deleted, you might want to trace back their insertions to make sure we are not losing behaviour by removing these.

		§ spring-websocket
		§ jetty-ee10-websocket-jakarta-server
		§ jetty-websocket-jetty-client
-- Removed and couldn’t find any usages, but am planning on spending some time smoke testing my branches before merge, but need to wait for FE changes first

- [ ] We have translations in each file for the key connectionError. Can these be deleted?
-- So for the FE side of things we also have [this](https://github.com/camunda/camunda-optimize/issues/13439) ticket, once BE is approved Omran will be adding FE removal of status checking things to this PR so that the changes don’t break anything. There will probably be more translation stuff that we can remove 
• To consider:

- [ ] 	The License filter and all logic can be deleted. Do we have an issue for that?
-- Yup, we have [this](https://github.com/camunda/product-hub/issues/2359) epic for the license logic 

- [ ] The Platform Security Adaptor can be deleted too, as we never run in Platform mode here
-- Is this related to import status reporting? We have a lot more C7 exclusive files to remove (basically anything that currently uses the platform condition, incl. the condition), but would do this on a feature by feature basis unless theyre tightly coupled, [this](https://github.com/camunda/camunda-optimize/issues/13211) is the ticket

PHWaechtler:
@RomanJRW ready for round 2 🍪 please also check [Here](https://github.com/camunda/camunda/pull/20557#issuecomment-2266685753) for replies 
Note the FE unit tests are expected to fail.

OmranAbazid:
@PHWaechtler I will add the front-end changes to your PR and ask @misiekhardcore to review it.

optimize/backend/src/main/java/io/camunda/optimize/jetty/LicenseFilter.java
```diff
@@ -90,19 +86,11 @@ private static boolean isLicenseCheckNeeded(
     return !isStaticResource(requestPath)
         && !isRootRequest(requestPath)
         && !isCloudEnvironment(applicationContext)
-        && !isExcludedApiPath(pathInfo)
-        && !isStatusRequest(requestPath);
+        && !isExcludedApiPath(pathInfo);
   }
 
   private static boolean isCloudEnvironment(ApplicationContext applicationContext) {
```
RomanJRW, line 92: rather than return true, why not just delete the method?

RomanJRW (approved):
LGTM 🚀

optimize/backend/src/main/java/io/camunda/optimize/jetty/LicenseFilter.java
PHWaechtler: When its true the method that calls it will correctly always return false, I am not sure if there's a scenario where the other checks could falsely result in true. The method that calls it is also unnecessary since it needs to be false in C8. Basically the entire logic in this class is not needed, but idk if the file will be removed or just adjusted in some way as part of the license topic and didnt want to cause conflicts.

PHWaechtler:
@OmranAbazid thats the BE changes finished, ready for FE stuff 🚀

optimize/backend/src/main/java/io/camunda/optimize/jetty/LicenseFilter.java
RomanJRW: Thanks for explaining 👍 
Ok to leave like this as a temporary state knowing that it will be deleted soon

misiekhardcore (approved):
FE changes look good 👍🏽

PR #20560, opened 2024-07-26
Dont use object mapper in DTO to ES / OS

### LINKED ISSUE #20556: Strange use of ObjectMapper in ES/OS DTO
**Description**

The following lines are used as a DTO in requests to Elasticsearch and OpenSearch:

https://github.com/camunda/camunda/blob/ff8dbe135d1523283ba1324ed42c98824150432d/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java#L243-L246

https://github.com/camunda/camunda/blob/0752796e53195c8d4eeec544925bb5cc7173b3f6/zeebe/exporters/opensearch-exporter/src/main/java/io/camunda/zeebe/exporter/opensearch/OpensearchClient.java#L301-L303

As these APIs expect an empty object `{ }` for the delete action, using `new ObjectMapper()` could eventually lead to sending a non-empty object and thus a bug. For now, it's not a bug but simply strange code.

Instead, we should pass an empty object `new Object()` or, even better, pass a `None` record.

korthout:
## Description

Use a record to represent an empty object rather than `new ObjectMapper()` in the DTO used to put the ILM/ISM for Elasticsearch and OpenSearch.

## Related issues

closes #20556


lenaschoenburg (approved):
Thank you!

PR #20575, opened 2024-07-26
refactor: remove process instance index archiving from Opt8

HeleneW-dot:
## Description

- Removes process instance archiving from Optimize 8
- Already includes index removal since it was just a small one, since the feature was never enabled this should not require a migration >>> will add one just to be safe. 
- Note that configuration changes all will be done in a separate config ticket to keep PR scopes smaller

We also want similar changes in Optimize 7, will add there too after changes are approved here. 

C7 equivalent [here](https://github.com/camunda/camunda-optimize/pull/13933)

## Related issues

related to [OPT#13211 ](https://github.com/camunda/camunda/issues/33560)part 3


grlimacan (approved):
LGTM!

optimize/upgrade/src/test/java/io/camunda/optimize/upgrade/migrate313to86/AbstractUpgrade86IT.java.out
```diff
@@ -8,33 +8,46 @@
...
-//   }
-// 
-//   protected void performUpgrade() {
-//     final List<UpgradePlan> upgradePlans =
-//         new UpgradePlanRegistry(upgradeDependencies)
-//             .getSequentialUpgradePlansToTargetVersion(TO_VERSION);
-//     upgradePlans.forEach(plan -> upgradeProcedure.performUpgrade(plan));
-//   }
-// }
+//import java.util.List;
```
grlimacan, line 11: What's the point of this file and the file below? Why is everything commented out?

grlimacan (approved):
LGTM! Just questioning whether the commented out files are senseful

optimize/upgrade/src/test/java/io/camunda/optimize/upgrade/migrate313to86/AbstractUpgrade86IT.java.out
PHWaechtler: For Optimize8, we have commented out all IT that currently only runs in C7 mode but should eventually be migrated to also run in C8 mode. The idea is to basically have all these test files we can migrate one by one, removing the C7 one and adding the C8. We've removed all IT that is C7 exclusive and does not need migrating.

PR #20576, opened 2024-07-26
Add support for REST API client middleware in the Java client

### LINKED ISSUE #19293: Provide a custom middleware mechanism for the Zeebe REST API in the Java client
### Description
The Zeebe Java client provides a custom middleware mechanism for the REST API. The new middleware mechanism works similarly to how custom gRPC interceptors are supported within the Zeebe Java client.

### AC
For the purposes of the new middleware mechanism, the following needs to be provided:
- [x] Custom middleware for the (Apache HttpClient-based) Zeeve REST API in the Zeebe Java client is implemented using the `org.apache.hc.client5.http.async.AsyncExecChainHandler` abstract class.
- [x] Custom implementations of the `AsyncExecChainHandler` class are passed in the `ZeebeClientBuilder`. The API methods should be similar to the ones related to the gRPC client interceptors.
- [x] A list of the custom `AsyncExecChainHandler` implementations is passed to the embedded Apache HttpClient instance (i.e. to it's builder).

### Additional context
- More information on the decisions can be read [here](https://github.com/camunda/camunda/issues/18446#issuecomment-2123939733)

koevskinikola:
## Description

Add support for REST API middleware in the Camunda Java client. 

## Related issues

closes #19293

### DIFF
--- a/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfigurationImpl.java
+++ b/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfigurationImpl.java
@@ -39,6 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ScheduledExecutorService;
+import org.apache.hc.client5.http.async.AsyncExecChainHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -51,6 +52,7 @@ public class ZeebeClientConfigurationImpl implements ZeebeClientConfiguration {
   private final CamundaClientProperties camundaClientProperties;
   private final JsonMapper jsonMapper;
   private final List<ClientInterceptor> interceptors;
+  private final List<AsyncExecChainHandler> chainHandlers;
   private final ZeebeClientExecutorService zeebeClientExecutorService;
 
   @Autowired
@@ -59,11 +61,13 @@ public class ZeebeClientConfigurationImpl implements ZeebeClientConfiguration {
       final CamundaClientProperties camundaClientProperties,
       final JsonMapper jsonMapper,
       final List<ClientInterceptor> interceptors,
+      final List<AsyncExecChainHandler> chainHandlers,
       final ZeebeClientExecutorService zeebeClientExecutorService) {
     this.properties = properties;
     this.camundaClientProperties = camundaClientProperties;
     this.jsonMapper = jsonMapper;
     this.interceptors = interceptors;
+    this.chainHandlers = chainHandlers;
     this.zeebeClientExecutorService = zeebeClientExecutorService;
   }
 
@@ -252,6 +256,11 @@ public class ZeebeClientConfigurationImpl implements ZeebeClientConfiguration {
   }
 
   @Override
+  public List<AsyncExecChainHandler> getChainHandlers() {
+    return chainHandlers;
+  }
+
+  @Override
   public JsonMapper getJsonMapper() {
     return jsonMapper;
   }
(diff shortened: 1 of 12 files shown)


clients/java/src/main/java/io/camunda/client/CamundaClientBuilder.java
```diff
@@ -162,6 +163,8 @@ CamundaClientBuilder jobWorkerExecutor(
 
   CamundaClientBuilder withInterceptors(ClientInterceptor... interceptor);
 
+  CamundaClientBuilder withChainHandlers(AsyncExecChainHandler... chainHandler);
```
tmetzke, line 166: 🔧 Javadoc would be super helpful here to understand what to use this for and how. We can maybe add some for the interceptors above as well.

clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
```diff
@@ -208,6 +211,11 @@ public List<ClientInterceptor> getInterceptors() {
   }
 
   @Override
+  public List<AsyncExecChainHandler> getChainHandlers() {
+    return chainHandlers;
+  }
+
+    @Override
```
tmetzke, line 218: 🔧 Indentation is a bit off here, same in other places.

tmetzke (changes requested):
👍 Really nice, good job, @koevskinikola 🚀 
🔧 I added some minor suggestions, nothing blocking. 
❌ We need to add the chain handlers to the Spring client as well, that's why the CI is failing.

clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
tmetzke: Also, imports seem to be a bit off here and in other classes, the Java checks are failing due to this.

tmetzke:
Unfortunately, we need to move this code into the `Zeebe...` classes as the Camunda Client will only come with 8.7. Let me know if you need help. You can also reach out to @nicpuppa.

clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
koevskinikola: 👍 I opened the `clients/java` module as a separate project in IntelliJ since I was having some issues with dependencies, and it seems IntelliJ doesn't pick up some of the formatting rules. I'll fix them.

koevskinikola:
> We need to add the chain handlers to the Spring client as well, that's why the CI is failing.

@tmetzke I was planning to add the Spring SDK part with a separate PR, but I guess I have to include it here for the CI.

tmetzke:
> > We need to add the chain handlers to the Spring client as well, that's why the CI is failing.
> 
> @tmetzke I was planning to add the Spring SDK part with a separate PR, but I guess I have to include it here for the CI.

It looks like it, yes 🤷

koevskinikola:
@tmetzke the feedback is implemented and the code was migrated back to the Zeebe client.

I also added the Spring SDK changes. @nicpuppa it might be good if you had a look at the code as well, in case I missed something for the Spring SDK or the migration back to the Zeebe client.

clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/zeebe/spring/client/configuration/ZeebeClientConfigurationImpl.java
```diff
@@ -39,6 +39,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ScheduledExecutorService;
+import org.apache.hc.client5.http.async.AsyncExecChainHandler;
```
tmetzke, line 42: 🔧 We need to add the `org.apache.httpcomponents.client5:httpclient5` dependency in the Spring SDK now as well since we use a class from it directly.

tmetzke (approved):
👍 Great job, @koevskinikola 🚀 
🔧 We're missing the explicit dependency to the Apache HTTP client in the Spring SDK.

Feel free to merge after adding the dependency, good to go from my end 🚢

PR #20577, opened 2024-07-26
Migrate intermediate message catch event

### LINKED ISSUE #19716: Migrate intermediate message catch event
It should be possible to migrate message intermediate catch events.

Message intermediate catch events can be migrated by providing a mapping instruction. Only active message intermediate catch events can be migrated.

Consider the following example where the process instance is waiting at the message intermediate catch event:
Screenshot 2024-06-25 at 10 59 04

berkaycanbc:
## Description

Allows for the migration of active [intermediate message catch events](https://docs.camunda.io/docs/components/modeler/bpmn/message-events/#intermediate-message-catch-events).

An active intermediate message catch event requires a mapping instruction, just like other active elements.

This explicitly does not yet support the migration of active intermediate message catch events that are attached to an [event-based gateway](https://docs.camunda.io/docs/components/modeler/bpmn/event-based-gateways/). This can could occur when the migration is requested in a race with a message correlation for that same catch event. This case is explicitly rejected for now.

## Related issues

closes #19716

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -216,7 +216,7 @@ public class ProcessInstanceMigrationMigrateProcessor
     final long processInstanceKey = elementInstanceRecord.getProcessInstanceKey();
     final var elementId = elementInstanceRecord.getElementId();
 
-    requireSupportedElementType(elementInstanceRecord, processInstanceKey);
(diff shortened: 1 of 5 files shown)


korthout:
Rebased on top of main to resolve conflicts due to merging:
- #19965

korthout (approved):
Thanks @berkaycanbc ⏰ 

To make progress, I've directly committed my change requests rather than provide review comments:
- d3d491101f4c0731f90344f2dcb9e6131604b3f4
- 65d046bc7ba763aac1f667db9ba5deb6b95ec6fe
- ccb13b6f8efa81274de3e30c1b4c71ae5712ebad
- 10a641651d10dca376ee09c6f983e6b2fcec4b2b

LGTM 👍

Issue #20287, opened 2024-07-16
Static translations of tasklist UI


Skaiir:
Product hub epic: https://github.com/camunda/product-hub/issues/2378

### Description

Offer the ability to change the language of the Tasklist UI, to accommodate strict business requirements of certain customers. For now, we'll simply do French and German statically as a POC. But it will be built future-proof. 

### User Story

```Gherkin
Scenario: As a Tasklist user
    Given I am logged in
    Then the application should be, if available, translated according to web browser preferences
    When I open the settings menu
    Then a dropdown to select of preferred language should be present, with no set default
    When I select a specific language
    Then that language should be applied and persisted across sessions (locally)
```

### Designs

https://github.com/camunda/product-hub/issues/2378#issuecomment-2206319621

### Acceptance Criteria

- [x] Tasklist language defaults to the browser default, if available, or english otherwise.
- [x] The new setting is present in the settings panel
- [x] Setting changes are persisted to localStorage
- [x] Translations for French and German are available
- [x] Feature can be easily toggled in code (as this is experimental for now)
- [ ] QA has verified the feature

### Breakdown

- [x] Install and setup i18next dependency
- [x] Catalogue and key all UI strings (following i18next key naming style)
- [x] Generate some translation files for German and French and hook them into i18next
- [x] Install and setup i18next react integration library
- [x] Add a feature flag for translations
- [x] Using i18next react integration library, replace all catalogued strings by keyed translation calls (test via manually changing the language).
- [x] Automatically detect browser language and translate to it
- [x] Add language selection entry to settings
- [x] Store to localstorage on language selection and change the application language
- [x] Load language from localstorage on application start
- [x] Add tests
- [x] Merged?

### Definition of Ready - Checklist

- [x] The issue has a meaningful title, description, and testable acceptance criteria

Skaiir:
Available in 8.6-alpha-5

PR #20581, opened 2024-07-29
Prerequisite config for tasklist internationalization

Skaiir:
## Description

Introduces the base config required to start internationalizing Tasklist and a feature flag. 

Partial PR. 

## Related issues

Related to #20287


vsgoulart (commented):
@Skaiir To fix the let's use this [method here](https://react.i18next.com/misc/testing#example-test-using-this-configuration)

I recommend we add a test to check if the checkbox is there and then we can check. We can also add a test to check if the changing the dropdown changes the value in the `l18n` instance

tasklist/client/src/Layout/Header/styles.module.scss
```diff
@@ -15,3 +15,7 @@
     color: var(--cds-link-primary);
   }
 }
+
+.languageDropdownPadding {
```
vsgoulart, line 19: @Skaiir [Please use the spacing tokens provided by Carbon](https://carbondesignsystem.com/elements/spacing/overview/)

Also remove the quotes since that's not valid CSS

vsgoulart (changes requested):

vsgoulart (approved):

PR #20584, opened 2024-07-29
Support configuring REST Query API availability

### LINKED ISSUE #20580: Configure Query API availability (disabled by default)

The C8 REST Query API needs to be disabled by default. I also want to enable it via configuration.

- There is a new configuration property `camunda.rest.query.enabled` (environment variable `CAMUNDA_REST_QUERY_ENABLED` should also work) for the REST gateway that enables the Query API controllers.
- By default, the Query API endpoints are not available.

Remove the Query API controllers from the gateway code, so they are never available.

**Additional context**

The Query API endpoints are only (planned to be) completed with 8.7. They currently lack some identity integration, which hinders a general rollout.

tmetzke:
## Description

* Aligns REST API resources regarding activation conditions, reusing `ConditionalOnRestGatewayEnabled`.
* Introduces REST Query API configuration via `camunda.rest.query.enabled` (environment variable `CAMUNDA_REST_QUERY_ENABLED` also works), so it can be disabled/enabled separately from the general REST API.
* The Query API is disabled by default.

## Related issues

closes #20580


tmetzke:
The integration tests added with https://github.com/camunda/camunda/pull/20584/commits/7f8b206538e031d97b628963b85b1f4981407b8c currently fail as mentioned in the commit. If anybody has a better idea to test this automatically, I'm happy for any input/direction 🙏

dist/src/main/java/io/camunda/application/commons/job/JobHandlerConfiguration.java
```diff
@@ -13,23 +13,20 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.util.unit.DataSize;
 
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnProperty(
-    name = {"zeebe.broker.gateway.enable", "camunda.rest.enabled"},
-    havingValue = "true",
-    matchIfMissing = true)
+@ConditionalOnRestGatewayEnabled
```
romansmirnov, line 29: This does not ensure that the beans/configurations are resolved in the expected order in Spring. This condition may be evaluated as `true` because the bean `RestGatewayDisabled` is missing at that time in moment when registering the configuration/bean, meaning Spring will initiate that configuration `JobHandlerConfiguration` eventually. But the bean `RestGatewayDisabled` might created as a next step. In that case, the bean `JobHandlerConfiguration` will be initiated even when the bean `RestGatewayDisabled` was initiated too. (In same cases, Spring might do the scanning in a different order, which will result in the expected result.)

That is why `@ConditionalOnProperty` has been used so far at these places. An alternative would be that `ConditionalOnRestGatewayEnabled` checks the configuration properties as well to have a deterministic behavior.

dist/src/main/java/io/camunda/application/commons/rest/RestApiConfiguration.java
```diff
@@ -7,14 +7,11 @@
...
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration(proxyBeanMethods = false)
 @ComponentScan(basePackages = "io.camunda.zeebe.gateway.rest")
-@ConditionalOnProperty(
-    name = {"zeebe.broker.gateway.enable", "camunda.rest.enabled"},
-    havingValue = "true",
-    matchIfMissing = true)
+@ConditionalOnRestGatewayEnabled
```
romansmirnov, line 16: The same as above.

dist/src/main/java/io/camunda/application/commons/service/CamundaServicesConfiguration.java
```diff
@@ -15,16 +15,13 @@
...
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnProperty(
-    name = {"zeebe.broker.gateway.enable", "camunda.rest.enabled"},
-    havingValue = "true",
-    matchIfMissing = true)
+@ConditionalOnRestGatewayEnabled
```
romansmirnov, line 24: The same as above.

dist/src/main/java/io/camunda/application/commons/service/SearchClientDatabaseConfiguration.java
```diff
@@ -20,11 +21,7 @@
 import org.springframework.context.annotation.Configuration;
 
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnProperty(
-    prefix = "zeebe.broker.gateway",
-    name = "enable",
-    havingValue = "true",
-    matchIfMissing = true)
+@ConditionalOnRestGatewayEnabled
```
romansmirnov, line 24: The same as above.

romansmirnov (changes requested):
Added a comment on the usage of `@ConditionalOnRestGatewayEnabled`, because it does not ensure that the beans are loaded in the expected order, hence, the respective bean might be initiated even when the "rest gateway" was disabled.

nicpuppa:
> The integration tests added with [7f8b206](https://github.com/camunda/camunda/commit/7f8b206538e031d97b628963b85b1f4981407b8c) currently fail as mentioned in the commit. If anybody has a better idea to test this automatically, I'm happy for any input/direction 🙏

Just wondering if you can use the new camunda project test to test the new configuration, creating a zeebe container should be enough. 

But if you want to configure it properly you need to use an env var, do you think it make sense to add here also the possibility to enable/disable the query api by env var ?

romansmirnov:
@tmetzke, could you please rebase your PR with `main`? I just merged https://github.com/camunda/camunda/pull/20457 which repackaged some of the commons classes (to avoid unexpected scanning of beans with Spring).

romansmirnov:
Pushed a small fix to resolve the compile issues in the test suite: https://github.com/camunda/camunda/pull/20584/commits/667391bf0300066d93d1b5ec6d6c322d94bfc72f

dist/src/main/java/io/camunda/application/commons/configuration/BrokerBasedConfiguration.java
```diff
@@ -79,12 +81,23 @@ public SchedulerConfiguration schedulerConfiguration() {
     return new SchedulerConfiguration(cpuThreads, ioThreads, metricsEnabled, "Broker", nodeId);
   }
 
-  @ConditionalOnProperty(prefix = "zeebe.broker.gateway", name = "enable", havingValue = "false")
+  @ConditionalOnExpression(
+      "'${zeebe.broker.gateway.enable:true}' == 'false' or '${camunda.rest.enabled:true}' == 'false'")
```
nicpuppa, line 85: ❌ I'm not sure about this, it should be something like:

```suggestion
      "'${zeebe.broker.gateway.enable:false}' or '${camunda.rest.enabled:false}'")
```

tmetzke: If I understand the syntax right, we provide the default values if the properties are missing after the `:` and the default value is `true` for both properties in case they are not defined. Alas, the definition should be fine in my regard. Let me know if I overlooked anything.

dist/src/main/java/io/camunda/application/commons/job/JobHandlerConfiguration.java
tmetzke: Ohhhhh, OK, I see. Yeah that makes, now that you mentioned it, I remember this general issue with Spring.

The only problem I see is that the usage of the properties is not entirely correct. The `zeebe.broker.gateway.enable` property is irrelevant for the gateway scenario. Even if I set this to `false`, the REST controllers should work in this use case but they currently don't. I guess that is what you mentioned as an edge case in the last PR you created touching this code. Now I really understand what you meant, I guess.

The inherent problem is that the `ConditionalOnRestGatewayEnabled` doesn't know which use case it is used in, gateway or broker. Thus it cannot decide which properties to check.

I guess we can live with this edge case but I find that a bit irritating since it's spread all over the code.

dist/src/main/java/io/camunda/application/commons/configuration/BrokerBasedConfiguration.java
nicpuppa: I don't think is the right way to provide the default values. You can also do like:

```suggestion
      @ConditionalOnProperty(
    name = {"zeebe.broker.gateway.enable", "camunda.rest.enabled"},
    havingValue = "false",
    matchIfMissing = true)
```

dist/src/main/java/io/camunda/application/commons/job/JobHandlerConfiguration.java
tmetzke: I will move the propery checks into the `ConditionalOnRestGatewayEnabled` and `ConditionalOnRestQueryEnabled` annotations then to at least be consistent. There have been inconsistencies already. We have a central place to determine if the REST API is enabled then.

dist/src/main/java/io/camunda/application/commons/configuration/BrokerBasedConfiguration.java
tmetzke: I have simplified the property handling, this should not be relevant anymore.

dist/src/main/java/io/camunda/application/commons/job/JobHandlerConfiguration.java
tmetzke: I provided a simplified way of evaluating the REST API properties with https://github.com/camunda/camunda/pull/20584/commits/02e8c278b2672d1a6d905be7177502c7f5fe8b24. It includes disabling the REST API if `zeebe.broker.gateway.enable` is set to false in all scenarios, which is not entirely correct but also reflects the existing behavior. In my regard, it makes our property handling more consistent and centralized, thus is should be less error-prone.

tmetzke:
I have removed the IT cases for now as they are not efficient and functional. I will tackle this in a follow-up. For now, I successfully tested all combinations of the three relevant properties in manual tests.

nicpuppa (approved):
LGTM in general 👍 

❓ Do you think make sense to also support enabling/disabling with env vars ? Not in this PR, but as a follow-up issue.

tmetzke:
> ❓ Do you think make sense to also support enabling/disabling with env vars ? Not in this PR, but as a follow-up issue.

You can already provide `CAMUNDA_REST_QUERY_ENABLED=true` as an environment variable, Spring picks this up automatically. Or do you mean something else?

nicpuppa:
> You can already provide `CAMUNDA_REST_QUERY_ENABLED=true` as an environment variable, Spring picks this up automatically. Or do you mean something else?

I meant exactly this, thanks 👍

tmetzke:
> For testing the new functionality, it might be worth to use the new camunda project test 💭

 But doesn't that use a deployed Camunda SNAPSHOT Docker image for setup? Within the same PR, that won't work since the updated Docker image doesn't exist yet, as far as I understand it. But I haven't worked with that project at all yet.

romansmirnov (approved):
LGTM! 🚀

PR #20590, opened 2024-07-29
Elasticsearch metrics now export using micrometer core library.

EuroLew:
## Description

The metrics for the elastic search exporter now use the micrometer core library so if the consumer of the metrics ever changes from prometheus it would be the work of pointing to the exported metrics.

## Related issues

relates to #18467

### DIFF
--- a/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
@@ -25,7 +25,7 @@ import io.camunda.zeebe.exporter.dto.PutIndexTemplateResponse;
 import io.camunda.zeebe.exporter.dto.Template;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.ValueType;
-import io.prometheus.client.Histogram;
+import io.micrometer.core.instrument.MeterRegistry;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Optional;
@@ -42,23 +42,20 @@ class ElasticsearchClient implements AutoCloseable {
   private final TemplateReader templateReader;
   private final RecordIndexRouter indexRouter;
   private final BulkIndexRequest bulkIndexRequest;
+  private final MeterRegistry meterRegistry;
 
   private ElasticsearchMetrics metrics;
 
-  ElasticsearchClient(final ElasticsearchExporterConfiguration configuration) {
-    this(configuration, new BulkIndexRequest());
-  }
-
   ElasticsearchClient(
-      final ElasticsearchExporterConfiguration configuration,
-      final BulkIndexRequest bulkIndexRequest) {
+      final ElasticsearchExporterConfiguration configuration, final MeterRegistry meterRegistry) {
     this(
         configuration,
-        bulkIndexRequest,
+        new BulkIndexRequest(),
         RestClientFactory.of(configuration),
         new RecordIndexRouter(configuration.index),
         new TemplateReader(configuration),
-        null);
+        null,
+        meterRegistry);
   }
 
   ElasticsearchClient(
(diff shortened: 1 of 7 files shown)


zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
```diff
@@ -42,23 +42,20 @@ class ElasticsearchClient implements AutoCloseable {
   private final TemplateReader templateReader;
   private final RecordIndexRouter indexRouter;
   private final BulkIndexRequest bulkIndexRequest;
+  private final MeterRegistry meterRegistry;
 
   private ElasticsearchMetrics metrics;
 
-  ElasticsearchClient(final ElasticsearchExporterConfiguration configuration) {
```
npepinpe, line 48: :question: Was this unused?

npepinpe (commented):
:+1: Looks good, did you test a benchmark to ensure visualizations still work? We should do that before merging.

And we should merge the first PR before this one of course ;)

npepinpe:
Starting a benchmark for this branch so we can test that the metrics are backwards compatible.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
EuroLew: It was used however since it was for an actual instantiation a meter registry is needed to be passed in so it is not fit for purpose

npepinpe:
Looks good :+1: 
 
image

npepinpe:
PR looks good, just re-running the benchmark with latest changes to check backwards compatibility with the dashboard.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchMetrics.java
```diff
@@ -7,66 +7,71 @@
...
+        Counter.builder(meterName("failed.flush"))
+            .description("Number of failed flush operations")
+            .tags(PARTITION_LABEL, partitionIdLabel)
+            .register(meterRegistry);
   }
 
-  public Histogram.Timer measureFlushDuration() {
-    return FLUSH_DURATION.labels(partitionIdLabel).startTimer();
+  public void measureFlushDuration(final Runnable flushFunction) {
+    flushDuration.record(flushFunction);
```
npepinpe, line 59: :thought_balloon: I would've liked to see the same pattern as we previously had. From what I see it sort of exists with the Timer.Sample ([see for example here](https://docs.micrometer.io/micrometer/reference/concepts/meter-provider.html)), but then you need to close it with a reference to the timer itself, so unless we build our own wrapper, I don't see this happening =/

I guess there will be a bit of a learning curve on how best to use this.

npepinpe (approved):
:+1:

npepinpe:
Can you write issues to migrate the following to use the Micrometer facade as well:
- Opensearch Exporter
- Metrics Exporter

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchMetrics.java
EuroLew: I dislike exposing the implementation details of how the metric is measured if a timer object was returned to the user as it would make it harder to change any backend logic or implementation. Is there a difference I'm not understanding between the old try-with syntax that returns a closable from the method and this method that takes a runnable? Or am I just misunderstanding and you prefer an explicit timer.start -> timer.stop

npepinpe: The pattern with an explicit start/stop - or a closeable - is generally just more flexible, especially when coupled with asynchronous code. Here in particular it's only marginally more useful, but it's good to have in general as it's just more flexible.

PR #20515, opened 2024-07-25
ci: update optimize issue templates to follow monorepo guidelines

misiekhardcore:
## Description

We need to update optimize issue templates to follow monorepo guidelines/standards

## Related issues

relates to https://github.com/camunda/camunda-optimize/issues/13885

PR #20592, opened 2024-07-29
fix: remove Optimize unused files

buccarel:
## Description

Some files, introduced by Optimize, are of no use in the monorepo


RomanJRW (changes requested):
I think the template changes here are wrong. Also they overlap with the changes from @misiekhardcore [here](https://github.com/camunda/camunda/pull/20515)

Can you remove the template changes from this PR and instead reply to @misiekhardcore to confirm that you are happy for the changes to target your merge-in branch? Your PR here can be limited to the other file removal 🙂

RomanJRW (approved):

PR #20608, opened 2024-07-30
refactor: add missing junit extension

npepinpe:
## Description

I honestly have no idea how this was working, but this adds the missing JUnit extension declaration for this class.


deepthidevaki (approved):
:see_no_evil: The test passed before because the broker was explicitly stopped and started in the test.

PR #20610, opened 2024-07-30
Fix Snyk scanning action for Zeebe

### LINKED ISSUE #20011: Snyk scan for Zeebe is broken after moving modules around
**Description**

We are not scanning Snyk vulnerabilities properly for Zeebe anymore since we moved repositories around, since we were scanning specific modules and the paths for them has changed. 

We have to fix the paths so they point to the appropriate repositories, and also fail the workflow to prevent future issues like this.

npepinpe:
## Description

Fixes the scanning of Snyk modules for Zeebe, which was broken after we moved the modules around. This also adds a check to fail the action if the project file is not found.

It also adds the Spring SDK to the scanned modules since we also want to make sure we have no vulnerabilities there.

## Related issues

closes #20011

### DIFF
--- a/.github/workflows/zeebe-snyk.yml
+++ b/.github/workflows/zeebe-snyk.yml
@@ -152,7 +152,7 @@ jobs:
           # To avoid exiting on the first failure, we instead flip this to 1 as soon as one of the
           # command fails, and return this at the end; anything non 0 will cause the step to fail
           exitCode=0
-          JAVA_PROJECTS=(bom zeebe/bpmn-model clients/java zeebe/exporter-api zeebe/gateway-protocol-impl zeebe/protocol)
+          JAVA_PROJECTS=(bom clients/java spring-boot-starter-camunda-sdk zeebe/bpmn-model zeebe/exporter-api zeebe/gateway-protocol-impl zeebe/protocol)
 
           # Remember that when called from a sub-shell, the environment/globals are different
           function output() {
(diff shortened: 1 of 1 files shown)


npepinpe:
Test run:
- https://github.com/camunda/camunda/actions/runs/10158369073/job/28090278765 - fails because projects are not found :heavy_check_mark: 
- https://github.com/camunda/camunda/actions/runs/10158475928/job/28090619235 - hopefully a successful run :)

deepthidevaki (approved):
Thanks.
Does it affect the stable branches where the old paths have to be used? I guess not because it uses the workflow from the stable branch.

npepinpe:
Correct. The only one affected is 8.5, but I will likely have to change the PR a little.

.github/workflows/zeebe-snyk.yml
```diff
@@ -152,7 +152,7 @@ jobs:
           # To avoid exiting on the first failure, we instead flip this to 1 as soon as one of the
           # command fails, and return this at the end; anything non 0 will cause the step to fail
           exitCode=0
-          JAVA_PROJECTS=(bom bpmn-model clients/java exporter-api gateway-protocol-impl protocol)
+          JAVA_PROJECTS=(bom clients/java spring-boot-starter-camunda-sdk zeebe/bpmn-model zeebe/exporter-api zeebe/gateway-protocol-impl zeebe/protocol)
```
nicpuppa, line 155: ❌ This is wrong, spring sdk is under clients/ folder now

npepinpe: Was my branch out of date? :see_no_evil: It's at the root for me

npepinpe: I'll update it in a sec

nicpuppa: Thank you 👍

PR #19381, opened 2024-06-14
build: merge optimize in

buccarel:
This PR places Optimize in the `camunda/camunda` repo, together with every other application of the Camunda stack.

**What's in scope**

* placing Optimize in the monorepo `camunda/camunda`
* making sure that the repo is healthy after the introduction of camunda

**What's not in scope**

* making Optimize as part of the single JAR

Optimize is being introduced in `camunda/camunda` as a submodule, declared as such in the main `pom.xml` file, but it's **not** using the parent pom from `parent/pom.xml`. This will be changed later, in the [context of becoming part of the same single JAR](https://github.com/camunda/product-hub/issues/2289).

**Workflow statuses**

* The workflows that are marked as "skipped" have not been checked
* At the time of marking this PR as ready for review, all of the intended workflows are healthy
* The workflow `Commitlint / lint-commits (pull_request)` currently fails, because the git history brought in by`camunda/camunda-optimize` had less strict rules than `camunda/camunda`
* There is a number of workflows that are failing. The failures are due to a known issue: https://github.com/camunda/camunda/issues/17926

**Review**

As this PR is introducing several thousands of commits in the repo, it's difficult to review it using the diff page provided by github. If a given file `$file` shall be reviewed, a visually-good way to do it would be to use the terminal as follows:

```bash
$ git checkout main
$ git pull
$ git checkout merge-optimize-in
$ git pull
$ git merge main
$ file="zeebe/sample-file.md"
$ git diff main -- "$file"
```

The content of this PR can be logically categorized as follows:

* files that introduce Optimize in the repo, under the `optimize/` directory
* project changes in the various `pom.xml`s
* changes in the CI files that were already present in the repo
* introduction of new CI files that are related to Optimize
* changes to other files that are not related to Optimize (currently only 1 test)

The first part is practically a mere copy of Optimize. It's under the directory `optimize/`. One note is that as `camunda/camunda-optimize` had different coding style rules, currently this not directory is not compliant with the rules of `camunda/camunda`. Nevertheless, as the parent `parent/pom.xml` is not being used by Optimize, this doesn't currently represent a problem.

The `pom.xml`s files that have been changed are identifiable as follows:
```bash
$ git diff --name-only main | grep 'pom.xml' | grep -v 'optimize'
dist/pom.xml
parent/pom.xml
pom.xml
testing/camunda-process-test-java/pom.xml
```
From the result, the entry `testing/camunda-process-test-java/pom.xml` can be neglected.

[...]

PR #20613, opened 2024-07-30
ci: run optimzie workflows when optimize files are changed

misiekhardcore:
## Description

These changes add conditions to optimize workflow to prevent them from running when unnecessary

## Related issues

relates to https://github.com/camunda/camunda-optimize/issues/13936


.github/workflows/optimize-ci.yml
```diff
@@ -24,6 +39,17 @@ permissions:
   pull-requests: write
 
 jobs:
+  detect-changes:
```
misiekhardcore, line 42: I decided to bring here the action I've created in [this PR](https://github.com/camunda/camunda-optimize/pull/13663) in `camunda/camunda-optimize` repo

misiekhardcore:
@RomanJRW @buccarel should we also add some configs to the main `ci.yml`? There are places where operate/tasklist/iedntity deflare their frontend builds. Second question: I've noticed that in `optimize-sync-issues.yml` we are using some `component:optimize` label in `additional_issue_labels`, shouldn't it be `component/optimize` instead?

buccarel (approved):

.github/workflows/optimize-backend-linting.yml
```diff
@@ -2,14 +2,20 @@ name: Optimize Backend Linting
 on:
   push:
     branches:
-      - master
+      - main
+      - maintenance/**
+      - release/**
+    paths:
```
buccarel, line 8: There's just this git error saying that it's not possible to use these 2 directives at the same time.
I think a correct syntax for this would be:
```
paths:
  - "optimize/**"
  - ".github/workflows/optimize-code-scanning-java.yml"
  - "!optimize/client/**"

File: .github/actions/optimize-paths-filter/action.yml [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+      with:
+        base: ${{ github.event.merge_group.base_ref || '' }}
+        ref: ${{ github.event.merge_group.head_ref || github.ref }}
+        filters: |
+          backend-change:
+            - optimize/backend/**
+            - optimize/upgrade/**
+            - optimize/util/**
+            - optimize/plugins/**
+            - optimize/qa/**
```
buccarel, line 35: The qa module is gone already ^_^

buccarel (approved):

.github/workflows/optimize-backend-linting.yml
```diff
@@ -2,14 +2,18 @@ name: Optimize Backend Linting
 on:
   push:
     branches:
-      - master
-    paths-ignore:
-      - "optimize/client/**"
-      - ".ci/**"
+      - main
+      - maintenance/**
```
RomanJRW, line 6: in the mono repo, I believe we use `stable` branches rather than `maintenance`-prefixed branches. I have created an issue for @misiekhardcore to look into this: https://github.com/camunda/camunda-optimize/issues/13938

Perhaps already changing this here makes sense

RomanJRW (approved):
Pre-approving but with a comment on future branch naming from the mono repo. Does it also make sense to get @cmur2's eyes on this?

misiekhardcore:
I decided to remove the `optimize-helm-integration.yml` and update the `camunda-helm-integration.yml` as a result of [this conversation](https://github.com/camunda/camunda/pull/19381#issuecomment-2258652841)

buccarel (approved):

.github/workflows/optimize-backend-linting.yml
cmur2: I think it does

cmur2:
> should we also add some configs to the main `ci.yml`

That would be awesome since it is the long-term goal anyways, see https://github.com/camunda/camunda/issues/17721 and [unified CI](https://github.com/camunda/camunda/wiki/CI-&-Automation#unified-ci).

But before adding **make sure** the workflows you add meet the [inclusion criteria](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria) since that is very vital for the development speed in the monorepo!

.github/workflows/optimize-backend-linting.yml
```diff
@@ -2,14 +2,18 @@ name: Optimize Backend Linting
...
+      - release/**
+    paths:
+      - "optimize/**"
+      - ".github/workflows/optimize-code-scanning-java.yml"
```
cmur2, line 10: This is not the name of this workflow file

misiekhardcore:
> > should we also add some configs to the main `ci.yml`
> 
> That would be awesome since it is the long-term goal anyways, see #17721 and [unified CI](https://github.com/camunda/camunda/wiki/CI-&-Automation#unified-ci).
> 
> But before adding **make sure** the workflows you add meet the [inclusion criteria](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria) since that is very vital for the development speed in the monorepo!

then we should not add optimize there just yet, as our pipeline definitely takes more than 10 minutes :/

cmur2:
> then we should not add optimize there just yet, as our pipeline definitely takes more than 10 minutes :/

Good to know, then some more performance optimizations/splitting/parallelism/bigger runners is probably worth looking into after the merge :+1:

.github/workflows/optimize-backend-linting.yml
misiekhardcore: I renamed the `maintenance` to `stable` in this PR and there is [another one](https://github.com/camunda/camunda/pull/20656) to rename all the remaining references to `maintenance` or `master`

misiekhardcore:
> > then we should not add optimize there just yet, as our pipeline definitely takes more than 10 minutes :/
> 
> Good to know, then some more performance optimizations/splitting/parallelism/bigger runners is probably worth looking into after the merge 👍

Yes, this is already on our radar

.github/workflows/optimize-backend-linting.yml
RomanJRW: @misiekhardcore - Please see my comment on the other PR, it is relevant here too

misiekhardcore: I replied there, see [here](https://github.com/camunda/camunda/pull/20656#issuecomment-2262190890). Also @cmur2 I think it would be good if you read that conversation

cmur2: I read it, thanks :+1:

.github/actions/optimize-paths-filter/action.yml
```diff
@@ -0,0 +1,39 @@
...
+outputs:
+  backend-changes:
+    description: Output whether any backend has changed
+    value: >-
+      ${{
+        github.event_name == 'push' ||
+        steps.filter-common.outputs.backend-change == 'true'
+      }}
+  frontend-changes:
+    description: Output whether any frontend has changed
```
RomanJRW, line 15:
```suggestion
    description: Output whether any frontend file has changed
```

.github/actions/optimize-paths-filter/action.yml
```diff
@@ -0,0 +1,39 @@
...
+      id: filter-common
+      with:
+        base: ${{ github.event.merge_group.base_ref || '' }}
+        ref: ${{ github.event.merge_group.head_ref || github.ref }}
+        filters: |
+          backend-change:
+            - optimize/backend/**
+            - optimize/upgrade/**
+            - optimize/util/**
+            - optimize/plugins/**
```
RomanJRW, line 34: I _think_ we are deleting this whole module "soon" as it is not C8. Does it make sense to already exclude it here?

.github/actions/optimize-paths-filter/action.yml
```diff
@@ -0,0 +1,39 @@
...
+
+          frontend-change:
+            - optimize/client/**
+            - optimize/c4/**
+            - optimize/backend/src/main/resources/localization/**
```
RomanJRW, line 39: in practice, does this mean the translation files? If so, I would argue that the BE workflows should also run. There are a number of ITs relating to making sure that various responses are localized/entities created correctly

.github/workflows/camunda-helm-integration.yml
```diff
@@ -25,6 +25,9 @@ jobs:
         tasklist:
           image:
             tag: SNAPSHOT
+        optimize:
+          image:
+            tag: SNAPSHOT
```
RomanJRW, line 30: didn't @OmranAbazid change this to C8-SNAPSHOT?

.github/workflows/optimize-code-scanning-java.yml
```diff
@@ -3,7 +3,7 @@ name: Optimize Java code scan
 on:
   pull_request:
     branches:
-      - "master"
+      - "main"
       - "maintenance/**"
```
RomanJRW, line 7: does this need changing too then?

RomanJRW (changes requested):

.github/actions/optimize-paths-filter/action.yml
misiekhardcore: So, to backend changes we include the whole `optimize/backend` directory, so when there are changes to translation files, backend tests will run. For the frontend changes we only care for the localization files cause changing translations might break FE tests

.github/workflows/optimize-code-scanning-java.yml
misiekhardcore: good catch

.github/workflows/camunda-helm-integration.yml
OmranAbazid: It is still on a PR here:
https://github.com/camunda/camunda/pull/20614
You can merge this and i will adjust my PR

.github/actions/optimize-paths-filter/action.yml
RomanJRW: Ah yeah of course, that's obvious now. Got it - thanks 👍

RomanJRW (approved):

.github/actions/optimize-paths-filter/action.yml
misiekhardcore: I think then we should leave it as it is and she can clean this file up as well in scope of her changes

Issue #19710, opened 2024-06-25
Resolve deployment-bound resource links within same deployment


jfriedenstab:
### Description
Resource links with `bindingType=deployment` must be resolved within the same deployment; this applies to called processes, called decisions and linked forms.

### Acceptance Criteria
- [x] When model elements are transformed, the `bindingType` is registered in `ExecutableCallActivity`/`ExecutableBusinessRuleTask`/`ExecutableUserTask` (`UserTaskProperties`).
- [x] When the target resource is resolved in `CallActivityProcessor`/`BusinessRuleTaskProcessor`/`UserTaskProcessor`, the `bindingType` is checked:
  - [x] in case of `latest`: no change; the current behavior is used
  - [x] in case of `deployment`:
    - Use the `processDefinitionKey` (available from the `BpmnElementContext`) to look up the `deploymentKey` (which will be added to a `PersistedProcess` with #19709).
    - Use the `deploymentKey` + target `processId`/`decisionId`/`formId` to retrieve the resource from the state.
    - In case the target resource cannot be found in the same deployment, return a `Failure`.
    (this case should be prevented for static IDs in the future by #19712)

### Dependencies
- requires #19461
- requires #19709

korthout:
>add deploymentKey to ProcessInstanceRecord/BpmnElementContext

I think we can avoid this by doing a lookup of the process definition of the `call activity` / `user task` / `business rule task` instance to find the `deploymentKey`. We can then use the `deploymentKey` together with the `processId` to lookup the process definition to call. This additional lookup slows down call activity execution slightly, but avoids having to add an extra field to all element instance records which.

jfriedenstab:
Thanks, @korthout, good suggestion 👍🏻. I've updated the issue description accordingly.

Issue #19709, opened 2024-06-25
Store reference to deployment for deployed resources


jfriedenstab:
### Description
To support the new [`deployment` binding type](https://github.com/camunda/camunda/issues/19384), Zeebe needs to be able to tell which resources (processes, decisions, forms) were deployed together in the same deployment. Therefore, we should store a reference to the deployment for each deployed resource.

### Acceptance Criteria
- [x] When a deployment request is processed and the contained resources are transformed, the `deploymentKey` is stored as a new property in:
  - [x] `ProcessRecord` (for all processes in a BPMN resource)
  - [x] `DecisionRecord` (for all decisions in a DMN resource)
    - _Note_: It doesn't seem to be necessary to store the deployment key in a `DecisionRequirementsRecord` as well, because only the decision ID is relevant when `deployment` binding is used in a business rule task
  - [x] `FormRecord` (for form resources)
- [x] When the records are persisted in the state, the `deploymentKey` is stored as a new property in:
  - [x] `PersistedProcess`
  - [x] `PersistedDecision`
  - [x] `PersistedForm`
- [x] New column families are added to store:
  - [x] Processes by `processId` and `deploymentKey`
  - [x] Decisions by `decisionId` and `deploymentKey`
  - [x] Forms by `formId` and `deploymentKey`
  - _Note_: To avoid taking up additional storage space, the column families could just store a foreign key (e.g. `processDefinitionKey`) instead of the whole object.
- [x] New methods are added to the state classes to look up the objects from the new column families
(these will be used in #19710)
  - The methods should use a cache to reduce latency.
- [x] The new column families are only filled for new deployments to ensure existing events are replayed deterministically. This should be done by introduction new record and event applier versions.

### Dependencies
- required by #19710

korthout:
I like the described approach. 

> [!NOTE]
> We need to carefully consider how to deal with previously deployed resources. Simply adjusting the event appliers / state classes directly can be dangerous as it may result in non-deterministic event replay. Alternatively, we can increase the version number of the appended events, to ensure newly appended events are applied differently than existing ones. Note that older events are also compacted, so a migration task may be necessary to adjust older data. However, such a task can impact startup/failover performance. All this may mean that we can only fill the proposed column family for newly deployed resources rather than for all existing ones. If so, we should clearly document it as such.

jfriedenstab:
> All this may mean that we can only fill the proposed column family for newly deployed resources rather than for all existing ones.

[...]

PR #20628, opened 2024-07-30
feat: store deployment key for deployed decisions

jfriedenstab:
## Description
Store the deployment key as a new property for deployed decisions and add the possibility to find a decision by decision id and deployment key. This is a prerequisite to be able to properly resolve called decisions when the new `deployment` binding type is used (see #19710).

## Related issues
part of #19709 (processes were already handled with #20498; forms will follow in another PR)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbDecisionState.java
```diff
@@ -348,7 +385,6 @@ public void storeDecisionRecord(final DecisionRecord record) {
     dbPersistedDecision.wrap(record);
     decisionsByKey.upsert(tenantAwareDecisionKey, dbPersistedDecision);
 
-    dbDecisionKey.wrapLong(record.getDecisionKey());
```
jfriedenstab, line 351: Decision key is already wrapped in line 347.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/CreateDeploymentMultiplePartitionsTest.java
```diff
@@ -534,6 +534,33 @@ public void shouldWriteDrgAndDecisionCreatedEventsWithSameKeys() {
...
+    // then
+    ENGINE.forEachPartition(
+        partitionId -> {
+          final var record =
+              RecordingExporter.decisionRecords()
+                  .withPartitionId(partitionId)
+                  .withIntents(DecisionIntent.CREATED)
+                  .withDecisionId(decisionId)
+                  .limit(1)
+                  .getFirst();
```
korthout, line 555: 🔧 Nit: no need to limit to 1 if we're getting the first. `getFirst()` would already short circuit if an entry is available.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/mutable/MutableDecisionState.java
```diff
@@ -28,6 +28,14 @@ public interface MutableDecisionState extends DecisionState {
    */
   void storeDecisionRequirements(DecisionRequirementsRecord record);
 
+  /**
+   * Store a reference to the decision key in the state, using the decision id and deployment key as
+   * composite database key.
+   *
+   * @param record the record of the decision
+   */
```
korthout, line 36: 👍 I like the docs

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/deployment/DecisionStateTest.java
```diff
@@ -167,6 +180,92 @@ void shouldFindDeployedDecisionByKey() {
         .isEqualTo(decisionRecord2.getDecisionKey());
   }
 
+  @DisplayName("should find deployed decision by ID and deployment key")
+  @Test
+  void shouldFindDeployedDecisionByIdAndDeploymentKey() {
```
korthout, line 185: 👍 Good test

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/deployment/DecisionStateMultiTenantTest.java
```diff
@@ -105,6 +105,47 @@ void shouldPutDecisionForDifferentTenants() {
...
+    final var drgKey = 123L;
+    final var drgId = "drgId";
+    final var decisionKey = 456L;
+    final var decisionId = "decisionId";
+    final int version = 1;
+    final var tenant1 = "tenant1";
+    final var tenant2 = "tenant2";
+    final var deploymentKey = 789L;
+    final var drg1 = sampleDecisionRequirementsRecord(tenant1, drgKey, drgId, version);
+    final var drg2 = sampleDecisionRequirementsRecord(tenant2, drgKey, drgId, version);
```
korthout, line 120: 💭 This test is a bit strange. It considers 2 DRGs belonging to different tenants, but they have the same key (`drgKey`). This already is the case in the test above (where this is likely copied from), so let's just accept it. It's not what this test focusses on, but just seems a bit strange.

korthout (approved):
Awesome @jfriedenstab 🚀 

👍 LGTM

korthout (changes requested):
I see now that there's something missing

❌ Please ensure we remove decisions from the new column family on deletion

jfriedenstab:
Thanks for the review!

> ❌ Please ensure we remove decisions from the new column family on deletion

This is actually done here:
https://github.com/camunda/camunda/blob/eba24c9386eaff49e48608dd4b1cf60aec2540a3/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbDecisionState.java#L461
(and also covered by "should not find decision after it has been deleted" in `DecisionStateTest`).

Or did I miss anything?

korthout:
Oops. I missed it. Sorry @jfriedenstab

korthout (approved):
LGTM

PR #20637, opened 2024-07-30
feat: store deployment key for deployed forms

### LINKED ISSUE #19709: Store reference to deployment for deployed resources

jfriedenstab:
## Description
Store the deployment key as a new property for deployed forms and add the possibility to find a form by form id and deployment key. This is a prerequisite to be able to properly resolve linked forms when the new `deployment` binding type is used (see #19710).

## Related issues
Closes #19709 (processes were already handled with #20498; decisions with #20628)


zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/ZbColumnFamilies.java
```diff
@@ -171,7 +171,8 @@ public enum ZbColumnFamilies implements EnumValue {
   USER_TASK_STATES(80),
   COMPENSATION_SUBSCRIPTION(81),
 
-  PROCESS_DEFINITION_KEY_BY_PROCESS_ID_AND_DEPLOYMENT_KEY(82);
+  PROCESS_DEFINITION_KEY_BY_PROCESS_ID_AND_DEPLOYMENT_KEY(82),
+  FORM_KEY_BY_FORM_ID_AND_DEPLOYMENT_KEY(84);
```
jfriedenstab, line 175: 83 was added with #20628.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/FormResourceTransformer.java
```diff
@@ -131,14 +130,16 @@ private void appendMetadataToFormRecord(
       final FormMetadataRecord formRecord,
       final String formId,
       final DeploymentResource resource,
-      final String tenantId) {
+      final DeploymentRecord deployment) {
```
korthout, line 133: 👍 I like it

korthout (approved):
Thanks @jfriedenstab 💯 

LGTM 👍

PR #20649, opened 2024-07-31
test: ensure the REST API configuration options

### LINKED ISSUE #20645: Automate tests for REST API configuration properties
## Description

Provide automated tests that ensure the C8 REST API and Query API can be enabled and disabled as expected using Spring properties. Since Spring properties can be provided as ENV variables as well, this is not explicitly tested.

Ensure that:
1. By default, REST API is enabled and Query API is disabled.
2. Providing `camunda.rest.enabled` with any value other than `true` disables the REST API (including the Query API).
3. Providing `zeebe.broker.gateway.enable` with any value other than `true` disables the REST API (including the Query API).
4. Providing `zeebe.broker.gateway.enable` and `camunda.rest.enabled` with any value other than `true` disables the REST API (including the Query API).
5. Providing `camunda.rest.query.enabled` with the value `true` enables the Query API if the REST API is enabled:
   1. Both properties mentioned above are not provided
   2. Either of the two properties mentioned above has the value `true`.
   3. Both properties mentioned above have the value `true`.
6. Providing `camunda.rest.query.enabled` with the value `true` does NOT enable the Query API if the REST API is disabled:
   1. Case 2
   2. Case 3
   3. Case 4

tmetzke:
## Description

Provide automated tests that ensure the C8 REST API and Query API can be enabled and disabled as expected using Spring properties.

## Related issues

closes #20645

### DIFF
--- a/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/RestControllerAnnotationTest.java
+++ b/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/RestControllerAnnotationTest.java
@@ -18,7 +18,7 @@ import org.springframework.web.bind.annotation.RestController;
 @AnalyzeClasses(
     packages = "io.camunda.zeebe.gateway.rest",
     importOptions = ImportOption.DoNotIncludeTests.class)
-public class RestApiDisabledTest {
+public class RestControllerAnnotationTest {
 
   /**
    * This ArchUnit test ensures that any REST API controllers, i.e. classes annotated with the
--- a/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/RestControllerTest.java
+++ b/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/RestControllerTest.java
@@ -5,7 +5,7 @@
  * Licensed under the Camunda License 1.0. You may not use this file
  * except in compliance with the Camunda License 1.0.
  */
-package io.camunda.zeebe.gateway.rest.controller;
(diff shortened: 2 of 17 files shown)


nicpuppa (approved):
In general LGTM. 💪  

❓ Why did you created separate tests class ? You can merge the different classes into one and use `@TestPropertySource` annotation on top of each tests to set the correct properties (it should work)

tmetzke:
> In general LGTM. 💪
> 
> ❓ Why did you created separate tests class ? You can merge the different classes into one and use `@TestPropertySource` annotation on top of each tests to set the correct properties (it should work)

I looked for something like that but couldn't find anything that works. I'll give that a try, nice hint, thank you! 🚀

tmetzke:
> `TestPropertySource`

OK so `TestPropertySource` only applies on class-level, not method-level. BUT I can try with [nested test classes in JUnit 5](https://github.com/spring-projects/spring-framework/issues/18951#issuecomment-1153547515).

tmetzke:
> > `TestPropertySource`
> 
> OK so `TestPropertySource` only applies on class-level, not method-level. BUT I can try with [nested test classes in JUnit 5](https://github.com/spring-projects/spring-framework/issues/18951#issuecomment-1153547515).

I played around with this for a while. It generally works to nest the different configurations and they are loaded correctly BUT each nested class loads its own application context so I'd have to duplicate the bean mocking and setup for each nested class because beans are not properly reused between the different contexts. Overall, it would probably be doable but the `@Nested` capabilities are a bit restricted which makes it not really attractive to use it here, I think.

Let me know if you think otherwise.

PR #20656, opened 2024-07-31
ci: use stable and main branch names instead of maintenance and master

misiekhardcore:
## Description

THis PR renames `master` branch to `main` and `maintenance` to `stable` in optimize files, to meet the monorepo standards

## Related issues

relates to # https://github.com/camunda/camunda-optimize/issues/13938

### DIFF
--- a/.github/ISSUE_TEMPLATE/optimize-bug_report.md
+++ b/.github/ISSUE_TEMPLATE/optimize-bug_report.md
@@ -1,10 +1,11 @@
 ---
+
 name: Optimize Bug report
 about: Report a problem about Optimize and help us fix it.
 title: ""
 labels: ["kind/bug", "qa:pendingVerification", "component/optimize"]
 assignees: ""
----
+-------------
 
 ### Describe the bug
 
@@ -60,7 +61,6 @@ Is this bug reproducible?
 #### Review Resources
 
 <!-- When in review, the resources to be used for review should be listed here) -->
-
 - Feature PR: {Link to PR targeting the main branch}
 - Preview environments: {Link(s) to preview environments}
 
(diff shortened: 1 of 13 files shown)


RomanJRW:
@misiekhardcore 
Something to note here: This change assumes that Optimize will be building releases from the mono repo from the same release/stable branches as the other components. Do we have confidence that this will be the case? Is Optimize being part of the Single Application a requirement in order for us to do so?

misiekhardcore:
> @misiekhardcore Something to note here: This change assumes that Optimize will be building releases from the mono repo from the same release/stable branches as the other components. Do we have confidence that this will be the case? Is Optimize being part of the Single Application a requirement in order for us to do so?

@RomanJRW 

For the alpha releases:
If we are going to release separately as we did two or three times already. we were using the `merge-optimize-in` branch but we could have done the same from `main` if the `merge-optimize-in` content was already merged there. Even releasing alphas from our temporary branch wasn't interfering with any other component in the monorepo, so `main` should work as well. The other thing is that the way optimize `pom.xml` file is configured now, the release branch will be named like this for optimize: `release/optimize-x.y.z-(alpha)`, so also our release branch is separated for now.

For patches:
The `stable` branch will replace our `maintenance` branch. The first `stable` branch we will need will be after 8.6.0 release. The release job only bumps the versions in `pom.xml` files, creates a tag, builds an image/artifacts and deploys them. For the tags, I configured the `pom.xml` file to make tags like: `x.y.z-(alpha)-optimize` so they also are not colliding with the monorepo main tags. These I believe are changes which even working with the same `stable` branch as others, shouldn't cause any problems (same as using `main` as mentioned above.

RomanJRW (approved):

buccarel (approved):

misiekhardcore:
@buccarel can I meerge this?

RomanJRW:
@misiekhardcore - can this now be merged?

PR #20699, opened 2024-08-01
test: fix process definition key in failing tasklist integration test

jfriedenstab:
## Description
The expected process definition key in the test has changed due to recent changes in the deployment logic: The key for the deployment itself is now generated _before_ the actual resource's key (see https://github.com/camunda/camunda/pull/20498 / https://github.com/camunda/camunda/pull/20498/commits/c26231dbd5de06a9421aacc4681673d3118072c3). That is why the expected process definition key had to be incremented by one.

## Related issues
relates to https://github.com/camunda/camunda/issues/19709


panagiotisgts (approved):
Spent way too much time looking to fix the CI on this :dragon:

jfriedenstab:
> Spent way too much time looking to fix the CI on this 🐉

Sorry about that @panagiotisgts! I think the test failure went unnoticed because the test was not executed for the original PR.

PR #20705, opened 2024-08-01
feat: add user creation

### LINKED ISSUE #20398: Camunda services support user creation
### Description
To support user creation within the existing Zeebe architecture, we need to implement several new paths of code, these are captured below.

### Acceptance criteria
- [x] A record exists containing the properties [`username`, `name`, `email`, `enabled`]
- [x] A record value exists wrapping the record for a user
- [x] A column family exists for `users`
- [x] A `DbState`, `State`, and `MutableState` exists
- [x] An Intent class exists, implementing the `Intent` class
- [x] A Processor exists, implementing the `TypedRecordProcessor` class
- [x] An Applier exists, implementing the `TypedEventApplier` class
- [x] An Event Processor class exists
  - [x] All user events are distributed to all partitions (hint: `CommandDistributionBehavior`)
- [x] A services class exists, extending the `ApiServices` class
- [x] A Broker request for user creation exists, extending the `BrokerExecuteCommand` class
  - [x] All user events are sent to partition 1 (hint: `setPartitionId(Protocol.DEPLOYMENT_PARTITION)`)
- [x] A method in the services class exists allowing the creation of a user

### Tests
- [ ] Given a user with an unknown username, when creating a user, the command is passed
- [ ] Given a user with a username already in the system, when creating a user, the command is rejected

maryarm:
## Description

To support user creation within the existing Zeebe architecture, we need to implement several new paths of code, these are captured below.

### Acceptance criteria
- [x] A record exists containing the properties [`username`, `name`, `email`, `enabled`]
- [x] A record value exists wrapping the record for a user
- [x] A column family exists for `users`
- [x] A `DbState`, `State`, and `MutableState` exists
- [x] An Intent class exists, implementing the `Intent` class
- [x] A Processor exists, implementing the `TypedRecordProcessor` class
- [x] An Applier exists, implementing the `TypedEventApplier` class
- [x] An Event Processor class exists
  - [x] All user events are distributed to all partitions (hint: `CommandDistributionBehavior`)
- [x] A services class exists, extending the `ApiServices` class
- [x] A Broker request for user creation exists, extending the `BrokerExecuteCommand` class
  - [x] All user events are sent to partition 1 (hint: `setPartitionId(Protocol.DEPLOYMENT_PARTITION)`)
- [x] A method in the services class exists allowing the creation of a user

### Tests
- [x] Given a user with an unknown username, when creating a user, the command is passed
- [x] Given a user with a username already in the system, when creating a user, the command is rejected

## Related issues

closes #20398


remcowesterhoud:
I'll have a look at this tomorrow morning 👀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
```diff
@@ -0,0 +1,74 @@
...
+import io.camunda.zeebe.engine.state.immutable.UserState;
+import io.camunda.zeebe.engine.state.mutable.MutableUserState;
+import io.camunda.zeebe.protocol.ZbColumnFamilies;
+import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
+import org.agrona.DirectBuffer;
+
+public class DbUserState implements UserState, MutableUserState {
+
+  private final UserRecordValue userRecordToRead = new UserRecordValue();
+  private final UserRecordValue userRecordToWrite = new UserRecordValue();
```
remcowesterhoud, line 25: ❌ We don't need seperate records for reading and writing. We can define a single `userRecord` and use this for both.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private final DbString username;
+  private final ColumnFamily<DbString, UserRecordValue> userColumnFamily;
```
remcowesterhoud, line 28: ❓ How does multi-tenancy fit with user management? Is it possible to have multiple users with the same name in a different tenant?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/UserRecordValue.java
```diff
@@ -0,0 +1,32 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.state.user;
+
+import io.camunda.zeebe.db.DbValue;
+import io.camunda.zeebe.msgpack.UnpackedObject;
+import io.camunda.zeebe.msgpack.property.ObjectProperty;
+import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
+
+public class UserRecordValue extends UnpackedObject implements DbValue {
```
remcowesterhoud, line 15: ❌ This naming is confusing. `RecordValue` is what we call our records in the protocol. Let's avoid using the same naming pattern here. I would opt to rename it to `StoredUser` or `PersistedUser`. This in line with `StoredMessage` or `PersistedProcess` respectively.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/UserRecordValue.java
```diff
@@ -0,0 +1,32 @@
...
+
+  private final ObjectProperty<UserRecord> recordProp =
+      new ObjectProperty<>("userRecord", new UserRecord());
```
remcowesterhoud, line 18:
```suggestion
  private final ObjectProperty<UserRecord> userProp =
      new ObjectProperty<>("user", new UserRecord());
```

🔧 Some small naming changes

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/UserRecordValue.java
```diff
@@ -0,0 +1,32 @@
...
+    declareProperty(recordProp);
+  }
+
+  public UserRecord getRecord() {
+    return recordProp.getValue();
+  }
+
+  public void setRecord(final UserRecord record) {
+    recordProp.getValue().wrap(record);
+  }
```
remcowesterhoud, line 31:
```suggestion
  public UserRecord getUser() {
    return userProp.getValue();
  }

  public void setUser(final UserRecord record) {
    userProp.getValue().wrap(record);
  }
```

🔧 More naming changes

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserCreatedApplier.java
```diff
@@ -0,0 +1,28 @@
...
+
+  private final MutableUserState userState;
+
+  public UserCreatedApplier(final MutableProcessingState processingState) {
+    userState = processingState.getUserState();
+  }
+
+  @Override
+  public void applyState(final long key, final UserRecord value) {
+    userState.create(value);
```
remcowesterhoud, line 26: ❓ We don't use the key here and we don't store it in the state. The key is fully unique to a user. Is the name unique enough so that we don't need to store the key in the state?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/CreateUserTest.java
```diff
@@ -0,0 +1,84 @@
...
+            .withEmail("bar@foo.com")
+            .withPassword("password")
+            .expectRejection()
+            .create();
+
+    final var createdUser = createdUserRecord.getValue();
+    assertThat(createdUser).isNotNull().hasFieldOrPropertyWithValue("username", username);
+
+    io.camunda.zeebe.protocol.record.Assertions.assertThat(duplicatedUserRecord)
+        .hasRejectionType(RejectionType.ALREADY_EXISTS);
```
remcowesterhoud, line 82: 🔧 Please also verify the intent is `UserIntent.CREATE` to make it more clear, and the `rejectionReason` to verify we respond with an accurate error message.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public void processRecord(final TypedRecord<UserRecord> command) {
+    final long key = command.getKey();
+    final var username = command.getValue().getUsernameBuffer();
+    final var user = userState.getUser(username);
+
+    if (user != null) {
+      rejectionWriter.appendRejection(command, RejectionType.ALREADY_EXISTS, "user already exists");
+      responseWriter.writeRejectionOnCommand(
+          command, RejectionType.ALREADY_EXISTS, "user already exists");
```
remcowesterhoud, line 51: ❌ Let's make sure the error message adheres to [our guidelines](https://github.com/camunda/camunda/wiki/Error-Guidelines):

`Expected to create user with name <name>, but a user with this name already exists`

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public void delete(final String username) {
+    delete(wrapString(username));
+  }
+
+  @Override
+  public UserRecord getUser(final DirectBuffer username) {
+    this.username.wrapBuffer(username);
+    final UserRecordValue user = userColumnFamily.get(this.username);
+    return user == null ? null : user.getRecord();
```
remcowesterhoud, line 67: ❌ Let's make sure we create a copy of the user record before we return it. This has to do with us reusing the value for each get. The object reference remains the same, so getting the second user would override the values of the first user.

It'd be like `return user == null ? null : user.getRecord().copy();`. It's also good to add a test case which verifies getting 2 users doesn't give 2 references to the same object.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/mutable/MutableUserState.java
```diff
@@ -0,0 +1,23 @@
...
+
+public interface MutableUserState extends UserState {
+
+  void create(final UserRecord user);
+
+  void update(final UserRecord user);
+
+  void delete(final DirectBuffer username);
+
+  void delete(final String username);
```
remcowesterhoud, line 22: ❌ All these methods are unused and untested. I'd suggest either removing them if you don't need them yet, or if you decide to keep them to properly test these methods.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/CreateUserTest.java
```diff
@@ -0,0 +1,84 @@
...
+    final var createdUser = createdUserRecord.getValue();
+    assertThat(createdUser)
+        .isNotNull()
+        .hasFieldOrPropertyWithValue("username", username)
+        .hasFieldOrPropertyWithValue("name", "Foo Bar")
+        .hasFieldOrPropertyWithValue("email", "foo@bar.com")
+        .hasFieldOrPropertyWithValue("password", "password");
+  }
+
+  @DisplayName("should create user throws exception when username already exists")
```
remcowesterhoud, line 53: 🔧 It's a bit nitpicky but the name says it throws an exception. This is not what really happens. Instead the command is rejected,

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/user/UserStateTest.java
```diff
@@ -0,0 +1,77 @@
...
+    userState.create(user);
+
+    // then
+    final var persistedUser = userState.getUser(user.getUsername());
+    assertThat(persistedUser).isEqualTo(user);
+  }
+
+  @DisplayName("should create user throws exception when username already exists")
+  @Test
+  void shouldThrowExceptionInCreateIfUsernameDoesNotExist() {
```
remcowesterhoud, line 63:
```suggestion
  void shouldThrowExceptionInCreateIfUsernameAlreadyExists() {
```

🔧 You forgot to rename the method fully 🙂

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/user/UserStateTest.java
```diff
@@ -0,0 +1,77 @@
...
+        new UserRecord()
+            .setUsername("username" + UUID.randomUUID())
+            .setName("U")
+            .setPassword("P")
+            .setEmail("email" + UUID.randomUUID());
+    userState.create(user);
+
+    // when/then
+    assertThatThrownBy(() -> userState.create(user))
+        .isInstanceOf(ZeebeDbInconsistentException.class);
```
remcowesterhoud, line 75: 🔧 Again it would be nice to check the error message matches with what you expect.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/UserClient.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public UserCreationClient newUser(final String username) {
+    return new UserCreationClient(writer, username);
+  }
+
+  public static class UserCreationClient {
+
+    private static final Function<Long, Record<UserRecordValue>> SUCCESS_SUPPLIER =
+        (position) -> RecordingExporter.userRecords().withSourceRecordPosition(position).getFirst();
```
remcowesterhoud, line 32: ❌ You now return the first record of the ValueType user. This will always be the `CREATE` command. If you really want to expect a success you should wait for the `CREATED` event. If that record exists you know 100% that the user is created. Right now this expectation would also work for a command rejection.

It also means that in your current test cases you are testing the data you use as input. There's no verification of the date in the `CREATED` event. It's pretty simple to change this. Changing it to this should do the trick:

```java
RecordingExporter.userRecords().withIntent(UserIntent.CREATED).withSourceRecordPosition(position).getFirst();
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/UserClient.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final Function<Long, Record<UserRecordValue>> REJECTION_SUPPLIER =
+        (position) ->
+            RecordingExporter.userRecords()
+                .onlyCommandRejections()
+                .withSourceRecordPosition(position)
+                .getFirst();
```
remcowesterhoud, line 38: 🔧 It's a bit less significant since you only find rejections here, but in general it's a good practice to also specify which command you expect to be rejected by also adding the `withIntent(UserIntent.CREATE)`

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java
```diff
@@ -86,6 +86,6 @@ String routingFor(final Record<?> record) {
   }
 
   private String valueTypeToString(final ValueType valueType) {
-    return valueType.name().toLowerCase().replace("_", "-");
+    return valueType != null ? valueType.name().toLowerCase().replace("_", "-") : "";
```
remcowesterhoud, line 89: ❓ Why is this neccessary?

zeebe/protocol-jackson/pom.xml
```diff
@@ -86,6 +86,10 @@
           <ignoredNonTestScopedDependencies>
             <ignoredNonTestScopedDependency>com.fasterxml.jackson.core:jackson-core</ignoredNonTestScopedDependency>
           </ignoredNonTestScopedDependencies>
+          <ignoredUnusedDeclaredDependencies>
+            <ignoredUnusedDeclaredDependency>io.camunda:zeebe-protocol-test-util</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>io.camunda:zeebe-protocol-asserts</ignoredUnusedDeclaredDependency>
+          </ignoredUnusedDeclaredDependencies>
```
remcowesterhoud, line 92: ❓ Why is this neccessary?

zeebe/protocol-jackson/src/test/java/io/camunda/zeebe/protocol/jackson/ZeebeProtocolModuleTest.java
```diff
@@ -7,28 +7,14 @@
...
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.MethodSource;
 
 @SuppressWarnings("unchecked")
 @Execution(ExecutionMode.CONCURRENT)
 final class ZeebeProtocolModuleTest {
-  private static final ObjectMapper MAPPER =
+  /* private static final ObjectMapper MAPPER =
```
remcowesterhoud, line 17: ❌ You commented out this entire test. It's a valid test, please revert it.

zeebe/protocol-test-util/pom.xml
```diff
@@ -131,5 +131,20 @@
...
+        <configuration>
+          <!-- dependencies only packaged but not explicitly used -->
+          <ignoredUnusedDeclaredDependencies>
+            <ignoredUnusedDeclaredDependency>org.junit.jupiter:junit-jupiter-params</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>io.camunda:zeebe-protocol-asserts</ignoredUnusedDeclaredDependency>
+          </ignoredUnusedDeclaredDependencies>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
```
remcowesterhoud, line 148: ❓ Again, why is this neccessary? I suppose it's because of the testclasses you commented out.

zeebe/protocol-test-util/src/test/java/io/camunda/zeebe/test/broker/protocol/ProtocolFactoryTest.java
```diff
@@ -7,27 +7,13 @@
...
+// import io.camunda.zeebe.protocol.record.RecordAssert;
 import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.MethodSource;
 
 @Execution(ExecutionMode.CONCURRENT)
 final class ProtocolFactoryTest {
-  private final ProtocolFactory factory = new ProtocolFactory();
+  /*private final ProtocolFactory factory = new ProtocolFactory();
```
remcowesterhoud, line 16: ❌ Please restore these tests

zeebe/exporters/opensearch-exporter/src/main/resources/zeebe-record-user-template.json
```diff
@@ -0,0 +1,39 @@
+{
+  "index_patterns": [
+    "zeebe-record_user_*"
+  ],
+  "composed_of": ["zeebe-record"],
+  "priority": 20,
+  "version": 1,
+  "template": {
+    "settings": {
+      "number_of_shards": 3,
```
remcowesterhoud, line 10: 🔧 Let's set the shards to 1 to keep it line with the other index templates.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserCreateRequest.java
```diff
@@ -0,0 +1,57 @@
...
+import io.camunda.zeebe.protocol.record.intent.UserIntent;
+import org.agrona.DirectBuffer;
+
+public class BrokerUserCreateRequest extends BrokerExecuteCommand<UserRecord> {
+
+  private final UserRecord requestDto = new UserRecord();
+
+  public BrokerUserCreateRequest() {
+    super(ValueType.USER, UserIntent.CREATE);
+    setPartitionId(Protocol.DEPLOYMENT_PARTITION);
```
remcowesterhoud, line 23: ❓ Is there a specific reason we always send the user create command to the deployment partition?

remcowesterhoud (changes requested):
Thanks for you contribution @maryarm 🚀! Please have a look at the comments I left 🙂 

@Ben-Sheppard FYI, since @maryarm is on FTO

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -820,6 +866,27 @@ components:
...
+          type: "integer"
+          format: "int64"
+        username:
+          type: "string"
+        name:
+          type: "string"
+        email:
+          type: "string"
+        enabled:
+          type: "boolean"
```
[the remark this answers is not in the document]
Ben-Sheppard: The response does get used in the future but its not visible in this PR (this reponse is used when the `/search` endpoint for users exists. I'm going to remove this for now as its not necessary in this PR

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
remcowesterhoud: Sorry, I wrote this down in a confusing way. Yes I mean we only need to define a single `UserRecordValue` instead of one for reading and one for writing.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
Ben-Sheppard: Good question, I can clarify this however my interpretation of multi-tenancy is that its the logical separation of data, and more specifically in our model data is deployed to a specific tenant and Users can then be assigned (given access) to those tenants. Based on that last part and as users are above tenants (in our implementation) its not possible for multiple users with the same name to exist

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserCreatedApplier.java
Ben-Sheppard: The `username` prop is intended to be unique so really the key shouldn't need to be stored in the state

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
Ben-Sheppard: Modified the code, need to handle the tests

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java
Ben-Sheppard: I'm not sure on the context of this change so can't speak for why its here, I would imagine that by this point we always have the value type so it shouldn't be null?

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserCreateRequest.java
Ben-Sheppard: This is something that roman and I discussed but I am unfortunately not able to find the context now. As he is currently on FTO getting that answer soon won't be possible.

My questions would be: is there anything wrong with doing this? Does t seem like a strange choice? Would you recommend a different approach?

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java
remcowesterhoud: Exactly, ValueType is never null here

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserCreateRequest.java
remcowesterhoud: Wrong is a big word. It would work fine this way. The reason I ask is because it means every user creation will be send to the same partition, instead of it being distributed amongst different partitions. This means that single partition will receive more load than others which could impact performance. I'm not too worried about this as I don't expect there'd be an insane amount of these command being sent.
Another thing is that partition 1 could be unavailable, for example due to backpressure. This means all create commands will get rejected, as opposed to another partition being able to handle them.

In general we should always try to distribute the load evenly across partitions. So even if the impact here wouldn't be that large it's still the nicer thing to do.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
Ben-Sheppard: Tests added

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserCreateRequest.java
Ben-Sheppard: I have no reason to not go with your suggestion here so lets use the preferred approach of distributing the command, if this needs changing its something that we can address later on :)

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/UserController.java
```diff
@@ -0,0 +1,51 @@
...
+
+    return RequestMapper.executeServiceMethodWithNoContentResult(
+        () ->
+            userServices
+                .withAuthentication(RequestMapper.getAuthentication())
+                .createUser(
+                    userWithPasswordDto.getUsername(),
+                    userWithPasswordDto.getName(),
+                    userWithPasswordDto.getEmail(),
+                    passwordEncoder.encode(userWithPasswordDto.getPassword())));
```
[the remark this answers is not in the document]
Ben-Sheppard: Valid concern, we should verify how this part is actually handled in the UI i.e. is a secure connection enough do we need to encrypt in the UI for the API call during user creation? For now I would propose that we make progress and I'll create a follow up action to verify moving forward

remcowesterhoud (changes requested):
Thanks for the fast changes @Ben-Sheppard! 🏅 

❌ As discussed on Slack I've added a comment on storing user keys in the state as well. Please have a look at this.

❌ Another thing is the Java client. Currently the PR introduces the REST API. We should extend the Java client so users have an easy way to call the API. You could also do this in a separate PR if you'd like. Let me know what your preference is.

🔧 Everything else looks good to me, but we are missing a few test cases:

1. We should add the `UserRecord` to the [JsonSerializableToJsonTest](https://github.com/camunda/camunda/blob/main/zeebe/protocol-impl/src/test/java/io/camunda/zeebe/protocol/impl/JsonSerializableToJsonTest.java). This test verifies the records are serialized as expected. There's plenty of examples in this test class 😄 
2. We should make sure we verify the distribution of the create command works as expected across the different partitions. We have a few examples for such a testcase available. I'd suggest taking a look at the [ResourceDeletionMultiPartitionTest](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionMultiPartitionTest.java) for inspiration. Note that you need to add a different test class for this, as the `EngineRule` needs to be configured to multiple partitions.
3. **If you add the client to this PR**: We should add an integration test to make sure the flow from client to engine and back works as expected. Take a look at the [ActivateJobsTest](https://github.com/camunda/camunda/blob/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/ActivateJobsTest.java) (you can ignore the parameterizaion and grpc part of this).

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
```diff
@@ -0,0 +1,55 @@
...
+import io.camunda.zeebe.protocol.ZbColumnFamilies;
+import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
+import org.agrona.DirectBuffer;
+
+public class DbUserState implements UserState, MutableUserState {
+
+  private final PersistedUser persistedUser = new PersistedUser();
+
+  private final DbString username;
+  private final ColumnFamily<DbString, PersistedUser> userColumnFamily;
```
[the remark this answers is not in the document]
Ben-Sheppard: That sounds fair :) I've implemented this and will push shortly

Ben-Sheppard:
Thank you for the review @remcowesterhoud - the Java client is something that @romansmirnov and I did discuss and we obviously need to address it, however, as this PR is already large I'd prefer to address it in a follow up PR.

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/ZbColumnFamilies.java
```diff
@@ -177,7 +177,8 @@ public enum ZbColumnFamilies implements EnumValue {
 
   MESSAGE_CORRELATION(85),
 
-  USERS(86);
+  USERS(86),
+  USERNAME_BY_USER_KEY(87);
```
remcowesterhoud, line 181: 🔧 The CF name is the wrong way around. The CF stores user keys by username. E.g. we use the username to get data from the CF and the data is the key 😄

zeebe/protocol-impl/src/test/java/io/camunda/zeebe/protocol/impl/JsonSerializableToJsonTest.java
```diff
@@ -2460,6 +2461,24 @@ final var record =
         }
         """
       },
+      {
```
remcowesterhoud, line 2464:
```suggestion
      /////////////////////////////////////////////////////////////////////////////////////////////
       ///////////////////////////////// UserRecord //////////////////////////////////////
       /////////////////////////////////////////////////////////////////////////////////////////////
      {
```

🔧 Let's add the "header" that we do for all other scenarios as well.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/CreateUserMultiPartitionTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    assertThat(
+            RecordingExporter.records()
+                .withPartitionId(1)
+                .limit(record -> record.getIntent().equals(CommandDistributionIntent.FINISHED)))
+        .extracting(
+            Record::getIntent,
+            Record::getRecordType,
+            r ->
+                // We want to verify the partition id where the deletion was distributing to and
```
remcowesterhoud, line 56:
```suggestion
                // We want to verify the partition id where the creation was distributing to and
```

🤡 You forgot to edit the comment

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/CreateUserMultiPartitionTest.java
```diff
@@ -0,0 +1,82 @@
...
+        .extracting(Record::getIntent)
+        .containsExactly(UserIntent.CREATE, UserIntent.CREATED);
+
+    assertThat(
+            RecordingExporter.records()
+                .withPartitionId(3)
+                .limit(record -> record.getIntent().equals(UserIntent.CREATED))
+                .collect(Collectors.toList()))
+        .extracting(Record::getIntent)
+        .containsExactly(UserIntent.CREATE, UserIntent.CREATED);
```
remcowesterhoud, line 80:
```suggestion
    assertThat(
            RecordingExporter.records()
                .withPartitionId(1)
                .limit(record -> record.getIntent().equals(CommandDistributionIntent.FINISHED)))
        .extracting(
            Record::getIntent,
            Record::getRecordType,
            r ->
                // We want to verify the partition id where the creation was distributing to and
                // where it was completed. Since only the CommandDistribution records have a
                // value that contains the partition id, we use the partition id the record was
                // written on for the other records.
                r.getValue() instanceof CommandDistributionRecordValue
                    ? ((CommandDistributionRecordValue) r.getValue()).getPartitionId()
                    : r.getPartitionId())
        .startsWith(
            tuple(UserIntent.CREATE, RecordType.COMMAND, 1),
            tuple(UserIntent.CREATED, RecordType.EVENT, 1),
            tuple(CommandDistributionIntent.STARTED, RecordType.EVENT, 1))
        .containsSubsequence(
            tuple(CommandDistributionIntent.DISTRIBUTING, RecordType.EVENT, 2),
            tuple(CommandDistributionIntent.ACKNOWLEDGE, RecordType.COMMAND, 2),
            tuple(CommandDistributionIntent.ACKNOWLEDGED, RecordType.EVENT, 2))
        .containsSubsequence(
            tuple(CommandDistributionIntent.DISTRIBUTING, RecordType.EVENT, 3),
            tuple(CommandDistributionIntent.ACKNOWLEDGE, RecordType.COMMAND, 3),
            tuple(CommandDistributionIntent.ACKNOWLEDGED, RecordType.EVENT, 3))
        .endsWith(tuple(CommandDistributionIntent.FINISHED, RecordType.EVENT, 1));

    for (int partitionId = 2; partitionId < PARTITION_COUNT; partitionId++) {
      assertThat(
              RecordingExporter.records()
                  .withPartitionId(partitionId)
                  .limit(record -> record.getIntent().equals(UserIntent.CREATED))
                  .collect(Collectors.toList()))
          .extracting(Record::getIntent)
          .containsExactly(UserIntent.CREATE, UserIntent.CREATED);
    }
```

❌ I've expanded on the assertions here. We should make sure each step of the distribution works as expected, and in the correct order. We need to make sure the other partitions actually acknowledge the distribution, and whilst theoretically just checking for the `FINISHED` should do, this is more easy to understand when this test case fails, as the message points directly to a missed acknowledgement.

I also put the user create and created assertion in a for-loop. It's not that significant but reduces some code duplication.

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/distribution/CommandDistributionRecord.java
```diff
@@ -37,6 +38,7 @@ public final class CommandDistributionRecord extends UnifiedRecordValue
     RECORDS_BY_TYPE.put(ValueType.MESSAGE_SUBSCRIPTION, MessageSubscriptionRecord::new);
     RECORDS_BY_TYPE.put(ValueType.RESOURCE_DELETION, ResourceDeletionRecord::new);
     RECORDS_BY_TYPE.put(ValueType.SIGNAL, SignalRecord::new);
+    RECORDS_BY_TYPE.put(ValueType.USER, UserRecord::new);
```
remcowesterhoud, line 41: 👍 I completely forgot we had this whitelist

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
```diff
@@ -0,0 +1,74 @@
...
+    userKey = new DbLong();
+
+    usernameToKeyColumnFamily =
+        zeebeDb.createColumnFamily(
+            ZbColumnFamilies.USERNAME_BY_USER_KEY, transactionContext, username, userKey);
+
+    keyToUserColumnFamily =
+        zeebeDb.createColumnFamily(
+            ZbColumnFamilies.USERS, transactionContext, userKey, persistedUser);
+  }
```
remcowesterhoud, line 45: ❌ Sorry, in the example I provided yesterday I forgot to add that we should use a Foreign Key in these ColumnFamilies. This ensures data consitency between the 2 Column Families. It's relatively easy to add:

```suggestion
  private final DbString username;
  private final DbLong userKey;
  private final DbForeignKey<DbLong> fkUserKey;
  private final ColumnFamily<DbString, DbForeignKey<DbLong>> usernameToKeyColumnFamily;
  private final ColumnFamily<DbKey, PersistedUser> keyToUserColumnFamily;
  
   public DbUserState(
       final ZeebeDb<ZbColumnFamilies> zeebeDb, final TransactionContext transactionContext) {
     username = new DbString();
     userKey = new DbLong();
     fkUserKey = new DbForeignKey<>(userKey, ZbColumnFamilies.USERS);

     usernameToKeyColumnFamily =
        zeebeDb.createColumnFamily(
            ZbColumnFamilies.USERNAME_BY_USER_KEY, transactionContext, username, fkUserKey);

     keyToUserColumnFamily =
        zeebeDb.createColumnFamily(
            ZbColumnFamilies.USERS, transactionContext, userKey, persistedUser);
```

So what we're saying with this is that key in the `usernameToKeyColumnFamily` is a foreign key. When we insert something into this ColumnFamily the key must exist in the `keyToUserColumnFamily`. Otherwise we're not allowed to insert it.

So this also means we need to swap the order in which we insert the data in the column families so we insert into the `keyToUserColumnFamily` first.


Sorry for not catching this sooner 🙇

remcowesterhoud (changes requested):
Great work @Ben-Sheppard! The comment I have now are mostly small. There's one bigger one.

I forgot to tell you about foreign keys in our column families yesterday. We should introduce one between the 2 column families of the user state. I've explained it in a comment. Please let me know if you have further question! Once this is fixed I reckon it's good to merge 🚀

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/CreateUserMultiPartitionTest.java
Ben-Sheppard: Haha classic copy pasta :D

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/CreateUserMultiPartitionTest.java
Ben-Sheppard: Ah thank you for this context, I agree it definitely makes sense to check all the steps

remcowesterhoud (approved):
LGTM 🏆

Issue #17019, opened 2024-03-21
Support limiting the returned attributes of responses


tmetzke:
## Description

In REST requests, I can define which attributes I want to receive in the response.

## Proposal

* This is a placeholder. A detailed proposal needs to be created.
* This must describe how the returned attributes can be configured, e.g. a plain list or also nested structures (for nested objects like variables).
* One example reference: https://jsonapi.org/format/#fetching-sparse-fieldsets

jonathanlukas:
Can this be compared to the expansion concept of Atlassian? https://developer.atlassian.com/server/jira-servicedesk/info/expansion/

tmetzke:
As far as I understand the Atlassian concept, this here is meant to be even more generic and expands to all resource attributes, not just specific ones. So you can completely state which attributes you want to have returned. By default, you get all of them (current API already). But apart from that, it is similar, I would say.

jonathanlukas:
Good to know :) 

So basically like an opt-in graphql approach, but without the graph stuff.

I really like the idea as it allows us to build larger dtos and still deliver on queries at scale.

tmetzke:
> Good to know :)
> 
> So basically like an opt-in graphql approach, but without the graph stuff.
> 
> I really like the idea as it allows us to build larger dtos and still deliver on queries at scale.

To the point, yes 👍

aleksander-dytko:
@tmetzke, what kind of use cases would this functionality help with?

tmetzke:
> [@tmetzke](https://github.com/tmetzke), what kind of use cases would this functionality help with?

As @jonathanlukas already mentioned, this is usually about data transfer optimization. The more often I interact with the API in a short time frame, the less data I would like to transport back and forth. Defining exactly those few properties I need in a response can help keep the transported data low.

It is a performance use case mostly, I would say.

aleksander-dytko:
Got it, thanks @tmetzke! 

I think this makes sense from my perspective in the long term once the V2 is at parity with V1 and covers the standard user journey well.

jonathanlukas:
Another prior art reference worth considering alongside JSON:API sparse fieldsets: the Salesforce sObject API supports a `fields` query parameter on retrieve requests (e.g. `GET /sobjects/Account/{id}?fields=Name,Industry`) to return only the requested fields instead of the full record. Could be a simpler alternative shape to evaluate for the flat (non-nested) case.

[...]

PR #20708, opened 2024-08-01
feat: add source filtering in search client

houssain-barouni:
## Description

For upcoming use cases in the C8 REST api, we need to implement source filtering in ES/OS queries to request specific fields (e.g., retrieving only the XML field of DecisionRequirement entity). 
In this pull request, we extend the search client with support for SourceConfig, which allows for field filtering by inclusion or exclusion.
For example, if you want to retrieve only ProcessInstance keys in a search query, you can build the query this way:
```java
SearchQueryBuilders.processInstanceSearchQuery()
    .responseConfig(r -> r.processInstanceKey().include());
```

**Pending**
- clean-up the code
- add unit tests

## Checklist

- [ ] for CI changes:
Pen

## Related issues

closes #


tmetzke (commented):
👍 Really cool idea, I like the general direction of getting such a filtering/projection option.

ℹ️ I only had a preliminary look at the code, I hope to have more time on Monday to give it a closer look and play around with it a bit.

Here are my initial thoughts:
💭 In REST APIs, such filtering is often done by a `fields` attribute/parameter, rather than `source` (which is an ES/OS-specific term, if I got that right?). Personally, I think `fields` is easier to relate than `source` here. But that is just a "gut feeling", take it with a grain of salt.
❓ Do we consider both inclusions and exclusions necessary from a REST API perspective? I would assume users configure their filtering and then define the fields they want returned, e.g. by simply listing the fields they want, similar to [variables to fetch](https://github.com/camunda/camunda/blob/a3510b09942f382338e18bbe349cb95cfeb94cb9/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L1651-L1658) in job activation. As an API user, that is more stable than writing "everything except attributes A and B" because the returned data schema might change over time (i.e. new fields being added). Just wondering about the use case, not saying it's wrong. 
💭 If we go for inclusions only. we could start simple with a list of attribute names to include. That would simplify the whole code a lot as well, I think. If we need more advanced mechanisms here in the future, we could still do that and treat the "simple" notation as the abbreviated form.
❓ We currently have `SearchSourceConfig` and `SearchSourceFilter` where the config only contains the filter, currently. That makes me wonder if either of them is superficial. Are they both needed or do we plan already to add more than the source filter to the source config? Otherwise, I think we can simply this for now and extend later when necessary.
💭 One or two examples of how this manifests in the REST API would be great for understanding the feature better, I think. We can also do this in the [API proposal](https://github.com/camunda/camunda/issues/17019) and look at it from the user and REST API perspective. We can then come back to this PR and see if we can simplify it based on the requirements. What do you think?

🚀 Overall, again, great direction and very important feature!
🔧 Let's however either a) go back to the drawing board and combine it with the API proposal directly so source filtering can be exposed to users as well directly or b) only build what's necessary for a specific use case internally (e.g. XML attribute fetching for an entity) and generalize later.

houssain-barouni:
> 👍 Really cool idea, I like the general direction of getting such a filtering/projection option.
> 
> ℹ️ I only had a preliminary look at the code, I hope to have more time on Monday to give it a closer look and play around with it a bit.

Thanks for the review @tmetzke 

> Here are my initial thoughts: 💭 In REST APIs, such filtering is often done by a `fields` attribute/parameter, rather than `source` (which is an ES/OS-specific term, if I got that right?). Personally, I think `fields` is easier to relate than `source` here. But that is just a "gut feeling", take it with a grain of salt. 

Yes, you’re correct that fields is also an [existing ES/OS concept](https://www.elastic.co/guide/en/elasticsearch/reference/8.14/search-fields.html#search-fields-param), and it does offer functionality similar to source filtering. However, it provides results in a different structure that requires distinct parsing.

Currently, the source field is intended primarily for internal use cases. Nevertheless, we can expose it in the REST API as fields and implement the necessary mappings between the REST layer (fields) and the service layer (source).


>❓ Do we consider both inclusions and exclusions necessary from a REST API perspective? I would assume users configure their filtering and then define the fields they want returned, e.g. by simply listing the fields they want, similar to [variables to fetch](https://github.com/camunda/camunda/blob/a3510b09942f382338e18bbe349cb95cfeb94cb9/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L1651-L1658) in job activation. As an API user, that is more stable than writing "everything except attributes A and B" because the returned data schema might change over time (i.e. new fields being added). Just wondering about the use case, not saying it's wrong. 💭 If we go for inclusions only. we could start simple with a list of attribute names to include. That would simplify the whole code a lot as well, I think. If we need more advanced mechanisms here in the future, we could still do that and treat the "simple" notation as the abbreviated form. 

This was initially thought with internal use cases in mind. Field exclusion can be useful for omitting specific fields to minimize the data overhead, especially in scenarios where certain fields contain large amounts of data. For example, in Tasklist, we have a usecase where we need to fetch variables with a preview value and exclude the full, larger value to reduce the ES network usage.

>❓ We currently have `SearchSourceConfig` and `SearchSourceFilter` where the config only contains the filter, currently. That makes me wonder if either of them is superficial. Are they both needed or do we plan already to add more than the source filter to the source config? Otherwise, I think we can simply this for now and extend later when necessary. 💭 One or two examples of how this manifests in the REST API would be great for understanding the feature better, I think. We can also do this in the [API proposal](https://github.com/camunda/camunda/issues/17019) and look at it from the user and REST API perspective. We can then come back to this PR and see if we can simplify it based on the requirements. What do you think?
> 

`SourceConfig` in ES/OS also supports disabling the source fetching , that can be used for example along with the usage of `fields` option mentioned above. But I think we won't need to include it in the `SearchSourceConfig`
❓ 
A question that now comes to my mind is: Should the SearchQuery classes in the service layer align more closely with REST usage, or should they provide a flexible abstraction of the database that accommodates various service caller needs?





> 🚀 Overall, again, great direction and very important feature! 🔧 Let's however either a) go back to the drawing board and combine it with the API proposal directly so source filtering can be exposed to users as well directly or b) only build what's necessary for a specific use case internally (e.g. XML attribute fetching for an entity) and generalize later.

tmetzke:
> This was initially thought with internal use cases in mind. Field exclusion can be useful for omitting specific fields to minimize the data overhead, especially in scenarios where certain fields contain large amounts of data. For example, in Tasklist, we have a usecase where we need to fetch variables with a preview value and exclude the full, larger value to reduce the ES network usage.
> 
> `SourceConfig` in ES/OS also supports disabling the source fetching , that can be used for example along with the usage of `fields` option mentioned above. But I think we won't need to include it in the `SearchSourceConfig`

Those are GREAT details, thanks for that! That makes a lot of sense and I guess we'll want to make use of those things sooner or later 👍 

> ❓ A question that now comes to my mind is: Should the SearchQuery classes in the service layer align more closely with REST usage, or should they provide a flexible abstraction of the database that accommodates various service caller needs?

From the top of my head, I'd say it's for abstracting from the database underneath. We can expose it to the REST API search as well but that is not required. The REST API controllers can also require certain database search functionality not exposed as a search configuration option.

Let's focus on the internal use cases first then. I'll re-evaluate it on Monday with that in mind 👍

ralfpuchert (approved):
I like the idea of projections. Besides the comments of other reviewers LGTM.

tmetzke (commented):
Had a second look, considering it only for internal use at the moment, i.e. for being used by controllers to achieve something like fetching only certain attributes of entities without the requirement of allowing this in general in the REST Query API as well already.

I like the general direction this is going into. My only advice would be: let's keep it as simple as possible and tied to a specific use case we already want to build, e.g. realizing a controller that only fetches the XML of a decision requirements entity but not the whole entity. We can take it from there to determine what we actually need. We might be already over-engineering things here otherwise.

houssain-barouni:
> Had a second look, considering it only for internal use at the moment, i.e. for being used by controllers to achieve something like fetching only certain attributes of entities without the requirement of allowing this in general in the REST Query API as well already.
> 
> I like the general direction this is going into. My only advice would be: let's keep it as simple as possible and tied to a specific use case we already want to build, e.g. realizing a controller that only fetches the XML of a decision requirements entity but not the whole entity. We can take it from there to determine what we actually need. We might be already over-engineering things here otherwise.

@tmetzke @ralfpuchert  Thanks again for the review, I have renamed the `SourceConfig` in the service layer to `QueryResultConfig` to make it less related to ES/OS and can be extended to select columns in RDBMS ;) https://github.com/camunda/camunda/pull/20708/commits/7d896aae7307b674be92203e096ea10136bb0d14

I also created this pull request https://github.com/camunda/camunda/pull/20886 to showcase how we can exclude `xml` field from DecisionRequirements search query results. The `xml` payload can be big and I think it is good to exclude from ES results especially that it is not returned in the API response. cc @nathansandi

PR #20713, opened 2024-08-01
Removed combined checksum

### LINKED ISSUE #20100: Remove references to unused combined checksum.
**Description**

The `combinedChecksum` is not used remove all references and implementations of it such as the `getCombinedValue()` and all constructors which reference it.

EuroLew:
## Description

Removed combined checksum as it is not used anymore instead where needed the actual file checksums are passed along and equality is checked. 

In addition removing any reference to it in the reading and writing of the checksum files where it mentions combined value in the comments.

For the sake of backwards compatibility, current snapshot chunks sent to other brokers must include the `snapshotChecksum` field as brokers on an older version will check this value. It will be left to a default value so that all checks on old versioned brokers pass but the field should eventually be removed.

## Related issues

closes #20100

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/snapshot/impl/SnapshotChunkImpl.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/snapshot/impl/SnapshotChunkImpl.java
@@ -34,6 +34,7 @@ public final class SnapshotChunkImpl
   private int totalCount;
   private String chunkName;
   private long checksum;
+  private long snapshotChecksum;
   private long fileBlockPosition;
   private long totalFileSize;
 
@@ -44,6 +45,7 @@ public final class SnapshotChunkImpl
     totalCount = chunk.getTotalCount();
     chunkName = chunk.getChunkName();
     checksum = chunk.getChecksum();
+    snapshotChecksum = 0;
     content.wrap(chunk.getContent());
     fileBlockPosition = chunk.getFileBlockPosition();
     totalFileSize = chunk.getTotalFileSize();
@@ -65,6 +67,7 @@ public final class SnapshotChunkImpl
 
     totalCount = SnapshotChunkDecoder.totalCountNullValue();
     checksum = SnapshotChunkDecoder.checksumNullValue();
+    snapshotChecksum = SnapshotChunkDecoder.snapshotChecksumNullValue();
     fileBlockPosition = SnapshotChunkDecoder.fileBlockPositionNullValue();
     totalFileSize = SnapshotChunkDecoder.totalFileSizeNullValue();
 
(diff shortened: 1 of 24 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/raft/snapshot/impl/SnapshotChunkImpl.java
```diff
@@ -45,7 +45,7 @@ public SnapshotChunkImpl(final SnapshotChunk chunk) {
     totalCount = chunk.getTotalCount();
     chunkName = chunk.getChunkName();
     checksum = chunk.getChecksum();
-    snapshotChecksum = chunk.getSnapshotChecksum();
+    snapshotChecksum = 0;
```
deepthidevaki, line 48: :wrench: This field can be removed. Instead directly write 0 in `write()`. Also add a comment why it is 0.

zeebe/snapshot/src/main/java/io/camunda/zeebe/snapshots/PersistedSnapshot.java
```diff
@@ -72,11 +72,11 @@ public interface PersistedSnapshot {
...
 
   /**
-   * Returns the checksum of the snapshot, which can be used to verify integrity.
+   * Returns the checksums of the snapshot files, which can be used to verify integrity.
    *
-   * @return the checksum of the snapshot
+   * @return the checksum of the snapshot files
    */
-  long getChecksum();
+  MutableChecksumsSFV getChecksums();
```
deepthidevaki, line 79: :x: This should return ImmmutableChecksum. The snapshot is already persisted that means the checksums cannot be updated anymore.

zeebe/snapshot/src/main/java/io/camunda/zeebe/snapshots/impl/SfvChecksumImpl.java
```diff
@@ -135,31 +99,25 @@ public void updateFromFile(final Path filePath) throws IOException {
...
-          combinedChecksum = new PreDefinedImmutableChecksum(crc);
-        }
-      } else {
-        final Matcher matcher = FILE_CRC_PATTERN.matcher(line);
-        if (matcher.find()) {
-          final Long crc = Long.parseLong(matcher.group(2), 16);
-          final String fileName = matcher.group(1).trim();
-          checksums.put(fileName, crc);
-        }
+        continue;
```
deepthidevaki, line 113: :question: This is for backwards compatibility so that it can still parse the old files. Is that right?

deepthidevaki (approved):
:rocket: Please address my comments before merging. 

:x: Also add a comment in the sbe definition of SnapshotChunk  (atomix/cluster/resources/snapshot-schema.xml) that snapshotChecksum is deprecated and not used anymore.

zeebe/snapshot/src/main/java/io/camunda/zeebe/snapshots/impl/SfvChecksumImpl.java
EuroLew: Yes it can still parse old files, it will ignore all commented lines (as they were just used to store combined checksum info) and read all normal checksum lines

PR #20719, opened 2024-08-02
Add an Incident Resolution REST endpoint

### LINKED ISSUE #20085: Migrate the Resolve incident gRPC API to REST
## Description

* The Camunda 8 REST API contains a incident resolution endpoint at `POST /v2/incidents/{incidentKey}/resolution` providing the same functionality as the [gRPC endpoint](https://docs.camunda.io/docs/apis-tools/zeebe-api/gateway-service/#resolveincident-rpc).
* The endpoint is featured in the OpenAPI description as (internally) [described](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.g4017mgfkpt9).

nicpuppa:
## Description

Add an endpoint to the REST API for resolving an incident.

## Related issues

closes #20085

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1351,7 +1351,10 @@ paths:
     post:
       tags:
         - Incident
-      summary: Resolves a given incident. This simply marks the incident as resolved; most likely a call to Update job will be necessary to reset the job’s retries, followed by this call.
+      summary: Resolves a given incident.
+      description: |
+        Marks the incident as resolved; most likely a call to Update job will be necessary
+        to reset the job’s retries, followed by this call.
       parameters:
         - name: incidentKey
           in: path
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1346,6 +1346,42 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+
+  /incidents/{incidentKey}/resolution:
+    post:
+      tags:
+        - Incident
+      summary: Resolves a given incident. This simply marks the incident as resolved; most likely a call to Update job will be necessary to reset the job’s retries, followed by this call.
+      parameters:
+        - name: incidentKey
+          in: path
+          required: true
+          description: Key of the incident to resolve.
+          schema:
+            type: integer
+            format: "int64"
+      responses:
+        '204':
+          description: The incident is marked as resolved.
+        "400":
+          description: The provided data is not valid.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "404":
+          description: The incident with the incidentKey is not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "500":
+          description: An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+
 components:
   responses:
     TopologyResponse:
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/IncidentController.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/IncidentController.java
@@ -34,10 +34,6 @@ public class IncidentController {
       consumes = MediaType.APPLICATION_JSON_VALUE)
   public CompletableFuture<ResponseEntity<Object>> incidentResolution(
       @PathVariable final long incidentKey) {
-    return resolveIncident(incidentKey);
-  }
-
-  private CompletableFuture<ResponseEntity<Object>> resolveIncident(final long incidentKey) {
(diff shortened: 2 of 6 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1346,6 +1346,42 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+
+  /incidents/{incidentKey}/resolution:
+    post:
+      tags:
+        - Incident
+      summary: Resolves a given incident. This simply marks the incident as resolved; most likely a call to Update job will be necessary to reset the job’s retries, followed by this call.
```
tmetzke, line 1354: 🙃 You can break this summary up into multiple lines like we do in other attributes in this document, I guess. Functionally, this is fine, of course.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/IncidentController.java
```diff
@@ -0,0 +1,47 @@
...
+    return resolveIncident(incidentKey);
+  }
+
+  private CompletableFuture<ResponseEntity<Object>> resolveIncident(final long incidentKey) {
+    return RequestMapper.executeServiceMethodWithNoContentResult(
+        () ->
+            incidentServices
+                .withAuthentication(RequestMapper.getAuthentication())
+                .resolveIncident(incidentKey));
+  }
```
tmetzke, line 46: 🙃 In this case, we might also just do the call directly in the public method since we don't parse, transform, or validate the input further. Just a nitpick though, functionally, this is fine, of course.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/IncidentControllerTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.concurrent.CompletableFuture;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.http.MediaType;
+
+@WebMvcTest(IncidentController.class)
+public class IncidentControllerTest extends RestControllerTest {
```
tmetzke, line 27: ❓ Since we only have the key here, should we add tests that break the API to ensure we're covered? Providing no key is not really possible (you'll get a 404 but, meh 🤷). We could test with a String like `abc` instead of a number? One thing we should do is test that the incident was not found and we receive an appropriate answer.

tmetzke (approved):
👍 Really cool, good job, @nicpuppa 🚀 
🔧 I added smaller nitpicks and a test suggestion, nothing that blocks the PR though.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/IncidentController.java
nicpuppa: Yeah it make sense

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/IncidentControllerTest.java
tmetzke: Yeah, I agree on "let's not test the framework" but disagree on "I trust the framework" (personally) 😄

nicpuppa: Yeah, I shouldn't 🤣

tmetzke: As long as it's Spring, you shouldn't 😆

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/IncidentControllerTest.java
```diff
@@ -54,4 +61,61 @@ void shouldResolveIncident() {
...
+        .post()
+        .uri(INCIDENT_BASE_URL + "/1/resolution")
+        .accept(MediaType.APPLICATION_JSON)
+        .contentType(MediaType.APPLICATION_JSON)
+        .exchange()
+        .expectStatus()
+        .isNoContent();
+
+    Mockito.verify(incidentServices).resolveIncident(1L);
+  }
```
tmetzke, line 82: 🔧 Copy-paste mistake? This exists twice now, if I'm not completely confused 😄

tmetzke (approved):
🚀 Thanks for applying the suggestions 👍 
🙃 One small copy-paste mistake left, I think, good to go from my end after correcting that.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/IncidentControllerTest.java
nicpuppa: 🙈

Issue #20654, opened 2024-07-31
[C8 API] Implement `Decision definition` endpoints


ralfpuchert:
### Description

Decision definition endpoints provides searching for decision definitions and the DMN description as XML.

[Issue for statistic](https://github.com/camunda/camunda/issues/21717)

* The endpoints should be implemented as described in [REST API Developer Guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)
* `{place-holder}` is the placeholder for parameters like `decisionDefinitionKey`
* Decision definitions objects are defined in
  * [Decision definition](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.ttejsuinuoa0) 

### Endpoints

- [x] https://github.com/camunda/camunda/issues/22650
- [x] https://github.com/camunda/camunda/issues/22652
- [x] https://github.com/camunda/camunda/issues/22653

### Definition of Ready - Checklist

- [ ] The issue has a meaningful title, description, and testable acceptance criteria
- [ ] If documentation needs to be updated, an issue is created in the [camunda-docs](https://github.com/camunda/camunda-docs) repo, and the issue is added to our Operate project board.
- [ ] If HELM charts need to be updated, an issue is created in the [camunda-platform-helm](https://github.com/camunda/camunda-platform-helm) repo, and the issue is added to our Operate project board.
- [ ] Potential Security risks have been considered and added to the product-hub issue risk assessment
- [ ] Cross-team dependencies have been considered

Optional:
- [ ] Design input has been collected by the assignee
### :point_right: Handover Dev to QA 

- Handy resources: 
     BPMN/DMN models, plugins, scripts, REST API endpoints + example payload, etc : 
     
- Example projects:

- Versions to validate:
> Docker file : in case that it needed to be tested via docker share the version contain the fixed along with version of other services . 

- Release version ( in which version this fixed/feature will be released)

houssain-barouni:
Hi @ralfpuchert 
I had added some comments in the API definition doc, can you please have a look
- [Decision Definition fields naming](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit?disco=AAABOxCQEw8), checkout the discussion in [Slack](https://camunda.slack.com/archives/C06UKS51QV9/p1722414816974799)
[...]

PR #20726, opened 2024-08-02
Update decision definition fields in the search endpoint

houssain-barouni:
## Description

In this pull request, I am applying the latest changes in Search Decision Definition endpoint specification.

1- renaming the `DecisionDefinition` fields in the API schema (filter, sort, response) to align them with the naming used in zeebe gateway protocol. 
* `decisionId` to `dmnDecisionId`
* `name` to `dmnDecisionName`
* `decisionRequirementsId` to `dmnDecisionRequirementsId`
* `key` to `decisionKey`
* `DecisionDefinitionEntity` reflects the ES/OS field naming as it is used to fetch the documents.
* ~~I kept the ES/OS field naming in `DecisionDefinitionFilter` and `DecisionDefinitionSort`, but I feel like this is debatable. (any thoughts @ralfpuchert @nathansandi @tmetzke )?~~

2- Removal of id field as it is just a String presentation of the decisionKey (long)

3- Removal of sorting and filtering by `decisionRequirementsName` and `decisionRequirementsVersion`: these fields do not belong the decision index but to decisionRequirements index. So either we need to:
* denormalize data and save these fields in the decision index. However, this implies some extra work in the new exporter that may not be a priority for now
* do as in V1: extra query to decisionRequirements index to fetch (with scrolling) all decisionRequirement keys that match the requested `decisionRequirementsName` and `decisionRequirementsVersion`. However, this does not fit with the current search service architecture that implies one object == one index. Also sorting will not be supported with this field as in V1.
* Create a new common index that handles child-parent relationship between decision and decisionRequirements, but this will have some overhead for migration and also handling the new index in V1 endpoints.
I opened a [new issue](https://github.com/camunda/camunda/issues/20767) to handle it.

## Checklist

- [ ] for CI changes:

## Related issues


service/src/main/java/io/camunda/service/search/filter/DecisionDefinitionFilter.java
```diff
@@ -16,65 +16,50 @@
...
-    List<String> decisionRequirementsNames,
-    List<Integer> decisionRequirementsVersions,
     List<String> tenantIds)
     implements FilterBase {
 
   public static final class Builder implements ObjectBuilder<DecisionDefinitionFilter> {
 
-    private List<String> ids;
-    private List<Long> keys;
+    private List<Long> decisionKeys;
```
nathansandi, line 30: 💭  I believe we are using just keys over here, like previous declaration.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1968,19 +1968,17 @@ components:
     DecisionDefinitionFilterRequest:
       type: object
       properties:
-        id:
-          type: string
-        key:
+        decisionKey:
```
nathansandi, line 1971: 🔨  We have been discussing over past definition meetings, once this is part of decision, the prefix decision before the key it is not necessary.

Does anything changed over here @tmetzke ?

nathansandi:
> > * I kept the ES/OS field naming in `DecisionDefinitionFilter` and `DecisionDefinitionSort`, but I feel like this is debatable. (any thoughts @ralfpuchert @nathansandi @tmetzke )?
> 
> Interesting point. No strong opinion here. There will always be different names here and there between the REST API (building on the Zeebe data model) and the indices. At some point, we have to map them accordingly. I don't really mind where that happens. We just need to align on consistent handling.
> 
> One way to do it that is easy to follow (I hope): the service layer follows the REST API data model as much as possible, in the interfaces and implementation classes. We map to the ES/OS names as late as possible.

Related to this point, even if the names of ES/OS are different, I think we should follow as much as possible the Zeebe model data names (what is defined in document). And have on the Search operations similar data models to the real time.

nathansandi (changes requested):
Good job :) 
Just one minor change, also some alignments that can be help us in the future over the comments.

houssain-barouni:
> One way to do it that is easy to follow (I hope): the service layer follows the REST API data model as much as possible, in the interfaces and implementation classes. We map to the ES/OS names as late as possible.

That would not be possible for the entity `DecisionDefinitionEntity` to use the same REST API field naming. It is used to parse the documents query response. It means that we would have discrepancy between the filter (request of the service) and the entity (response of the service) which is not great. One solution is to pass by another DTO that we map from/to `DecisionDefinitionEntity` which is overkill.

tmetzke:
> That would not be possible for the entity `DecisionDefinitionEntity` to use the same REST API field naming. It is used to parse the documents query response. It means that we would have discrepancy between the filter (request of the service) and the entity (response of the service) which is not great. One solution is to pass by another DTO that we map from/to `DecisionDefinitionEntity` which is overkill.

I see. Well, those entity classes are the latest point to map then. As I said, it's just one way. If that is the furthest "down" to the ES/OS we can go without creating additional DTOs unnecessarily, then that's it. We just need to be consistent with that then in the search endpoints. Can we work with that?

ralfpuchert (approved):
Besides the comments of the other reviewers LGTM.

houssain-barouni:
> > That would not be possible for the entity `DecisionDefinitionEntity` to use the same REST API field naming. It is used to parse the documents query response. It means that we would have discrepancy between the filter (request of the service) and the entity (response of the service) which is not great. One solution is to pass by another DTO that we map from/to `DecisionDefinitionEntity` which is overkill.
> 
> I see. Well, those entity classes are the latest point to map then. As I said, it's just one way. If that is the furthest "down" to the ES/OS we can go without creating additional DTOs unnecessarily, then that's it. We just need to be consistent with that then in the search endpoints. Can we work with that?

To sum-up
`DecisionDefinitionFilter` and `DecisionDefinitionFilter` will use the REST api field naming
`DecisionDefinitionEntity` will use ES/OS field naming

Just we need to be cautious that [DecisionDefinitionServices.search](https://github.com/camunda/camunda/blob/ebfb2bc44354410525c3ae4f9f918d0d5adb16eb/service/src/main/java/io/camunda/service/DecisionDefinitionServices.java#L45-L47) would have different field namings used in the input arguments and in the returned result.
https://github.com/camunda/camunda/blob/ebfb2bc44354410525c3ae4f9f918d0d5adb16eb/service/src/main/java/io/camunda/service/DecisionDefinitionServices.java#L45-L47

tmetzke:
> Just we need to be cautious that [DecisionDefinitionServices.search](https://github.com/camunda/camunda/blob/ebfb2bc44354410525c3ae4f9f918d0d5adb16eb/service/src/main/java/io/camunda/service/DecisionDefinitionServices.java#L45-L47) would have different field namings used in the input arguments and in the returned result.

I think that is fair and even expected. That is also what happens for Zeebe Broker calls in the services, where you put in data of one type (sometimes just plain parameters) and receive data of another type. That is what we have response mappers for, right? They map service responses (Zeebe Broker-specific, data search-layer specific) to REST responses (REST API schema).

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
houssain-barouni: for reference this are the renamed fields
* `decisionId` to `dmnDecisionId`
* `name` to `dmnDecisionName`
* `decisionRequirementsId` to `dmnDecisionRequirementsId`
* `key` to `decisionKey`

houssain-barouni: @tmetzke any thoughts on this? I don't want to go against any previous decision.

service/src/main/java/io/camunda/service/search/filter/DecisionDefinitionFilter.java
```diff
@@ -16,65 +16,50 @@
...
-    public Builder names(final List<String> values) {
+    public Builder dmnDecisionNames(final List<String> values) {
       this.names = addValuesToList(this.names, values);
       return this;
     }
 
-    public Builder names(final String value, final String... values) {
-      return names(collectValues(value, values));
+    public Builder dmnDecisionNames(final String value, final String... values) {
+      return dmnDecisionNames(collectValues(value, values));
```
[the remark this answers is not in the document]
houssain-barouni: done in https://github.com/camunda/camunda/pull/20726/commits/5fb0fac5cacf6d0b386b146fe12fa6894979bc22

houssain-barouni:
I also added support to sorting by `decisionKey` in this commit https://github.com/camunda/camunda/pull/20726/commits/0f8f09f00c733d846a73c2338f61c31a19786079

nathansandi (approved):
LGTM  💯

PR #20732, opened 2024-08-02
ci: add run assembly profile to release plugin

OmranAbazid:
## Description
We recently made the assembly plugin optional since it takes long time to run. For the release, we require the assembly plugin to run so we need to add the profile for that.

## Related issues

closes #


misiekhardcore (approved):
LGTM. I allow myself to merge this as I need that to build the release

Issue #16304, opened 2024-02-12
[EPIC] Correlate message with result


saig0:
<h3>Is your feature request related to a problem? Please describe.</h3>
When publishing a message, users need to know immediately if the publication was successful to take further action. There is currently no possibility of getting the information if the correlation was successful.

Currently, the only response for a published message is the message key ([ref](https://docs.camunda.io/docs/8.3/apis-tools/grpc/#output-publishmessageresponse)) which doesn't provide enough information to take appropriate action.

When the message is published, we need to know what was the result of this operation.

<h3>Describe the solution you'd like</h3>
Provide an API that returns:
- if the message was correlated to a process instance 
- the process instance key if the message was correlated

If the message is correlated to multiple process instances, the API returns one of the process instance keys. 

Message buffering (i.e. message TTL > 0) doesn't fit well with returning a correlation result. If there is no open message subscription during the message publication then we could return a result only within the given request timeout. Usually, the request timeout is lower than the message TTL (i.e. default is 1 hour in the Java client). To provide consistent and simple usage, we can exclude message buffering from the API.   

To separate the new API from the existing publish message API, we can use a different name. The name should reflect the intention to correlate a message to a specific process instance. This goes more in the direction of Camunda 7's message API ([ref](https://docs.camunda.org/manual/7.20/reference/bpmn20/events/message-events/#message-api)).

<details><summary><h3>Rest API</h3></summary>

<h4>Endpoint</h4>

POST `/v2/messages/correlation`

<h4>Request body</h4>

Parameter name | Type | Required | Description
-- | -- | -- | --
name | string | true | Name of the message
correlationKey | string | false | Correlation key of the message
variables | object | false | JSON object representing the variables to pass along with the message
tenantId | String | false | The tenant of the message sender

</div></b>

<h4>Response</h4>

HTTP status|Description|Response schema
--|--|--
200|The message is correlated to one or more process instances|{</br>  "key": int64, //unique ID of the published message</br>  "tenantId": string, //the tenant ID of the message</br>  "processInstanceKey": int64 //the process instance key the message was correlated to</br>}
400|The provided data is not valid.|[Problem details](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.o06ulmu7sbls)
[...]

PR #20773, opened 2024-08-05
Store message correlate request data in state

### LINKED ISSUE #20568: Store request data in state
<h2>Description</h2>

In #20196 we stored the request data in the `MessageSubscriptionRecord`. After further implementation we've come to the conclusion that this doesn't work in all cases. Specifically in the case where the `MessageSubscription` gets rejected, we have no way to access the request data we need in order to send a response.

Instead we will create a new `ColumnFamily`. This will store the request data using the key of the message. The data will be stored using the `MessageCorrelation.CORRELATING` event. It gets removed in either the `MessageCorrelation.CORRELATED` or `MessageCorrelation.NOT_CORRELATED` events.

What should we do:
- Create the new `ColumnFamily`
- Add a new `MessageCorrelation.CORRELATING` intent
- Add the request data and message key to the `MessageCorrelationRecord`
- Create the `MessageCorrelationCorrelatingApplier` and store the request data in the new `ColumnFamily`
- Revert the changes of #20196 
- Change all usages of the request data in the `MessageSubscriptionRecord` to the new `ColumnFamily` instead.

remcowesterhoud:
## Description

This PR reverts a change that was made earlier where we stored the request data on the `MessageSubscriptionRecord`. It turns out this option doesn't work properly for all cases (see [here](https://github.com/camunda/camunda/issues/16304#issuecomment-2268366752)).

The PR introduces a new `ColumnFamily`. This will be responsible for storing the request data. To store the data we also needed a new intent (`MessageCorrelation.CORRELATING`). This comes with the applier to actually store the data.

Note that I've changed the id's of the intents in this PR. As this functionality is not released yet this does't cause any issues regarding backwards compatibility.

## Checklist

- [ ] for CI changes:

## Related issues

closes #20568


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/DbMessageCorrelationState.java
```diff
@@ -0,0 +1,59 @@
...
+      final long messageKey, final long requestId, final int requestStreamId) {
+    this.messageKey.wrapLong(messageKey);
+    requestData.setRequestIdProp(requestId).setRequestStreamIdProp(requestStreamId);
+    messageCorrelationColumnFamily.insert(this.messageKey, requestData);
+  }
+
+  @Override
+  public RequestData getRequestData(final long messageKey) {
+    this.messageKey.wrapLong(messageKey);
+    return messageCorrelationColumnFamily.get(this.messageKey);
```
nicpuppa, line 51: ❓ Should we return a copy of the record instead of the record itself ? Like we do for other records

❓ What happens if the request data is not found in the state ? Is it possible ?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/RequestData.java
```diff
@@ -0,0 +1,42 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.state.message;
+
+import io.camunda.zeebe.db.DbValue;
+import io.camunda.zeebe.msgpack.UnpackedObject;
+import io.camunda.zeebe.msgpack.property.IntegerProperty;
+import io.camunda.zeebe.msgpack.property.LongProperty;
+
+public class RequestData extends UnpackedObject implements DbValue {
```
nicpuppa, line 15: 🔧 Is the name too generic ? Should we specify that this is related to messages ?

nicpuppa (commented):
Hey @remcowesterhoud great work, overall it looks good. I left some questions to discuss, please have a look

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/RequestData.java
remcowesterhoud: I don't think that's necessary. The object only stores request data, if we ever need this for a different feature there's no problem with this being reused.

nicpuppa: Ok, I see. Then should we move this class into another package ?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/DbMessageCorrelationState.java
remcowesterhoud: > Should we return a copy of the record instead of the record itself ? Like we do for other records

This isn't required, please note that the `RequestData` is not a record but a `DbValue`. We don't make copies for other `DbValues` either, e.g.:

```java
  @Override
  public StoredMessage getMessage(final long messageKey) {
    this.messageKey.wrapLong(messageKey);
    return messageColumnFamily.get(this.messageKey);
  }
```

> What happens if the request data is not found in the state ? Is it possible ?

This method will return `null` and we will likely end up in an NPE. This is why before getting it I check if it exists using the `existsRequestDataForMessageKey` method.
This does remind me that I missed to write a test for getting without value so I'll add that 😄

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/RequestData.java
remcowesterhoud: > Ok, I see. Then should we move this class into another package ?

If you have a good suggestion I'm happy to move it. I figured we could move it if we ever actually use it in multiple places.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/DbMessageCorrelationState.java
nicpuppa: > This isn't required, please note that the `RequestData` is not a record but a `DbValue`. We don't make copies for other `DbValues` either, e.g.:

This is not true:

```
  @Override
  public CompensationSubscription get(
      final String tenantId, final long processInstanceKey, final long recordKey) {
    wrapCompensationKeys(processInstanceKey, recordKey, tenantId);
    return compensationSubscriptionColumnFamily
        .get(tenantAwareProcessInstanceKeyCompensableActivityId)
        .copy();
  }
```

We had a discussion about this as a team [internal](https://camunda.slack.com/archives/C05DH1F5TAR/p1707908005239069) . `TransactionalColumnFamily.get` returns an object which gets mutated with every future get invocation

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/RequestData.java
nicpuppa: Not really. I'd suggest to leave the class there and if needed we can move it as you said 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/message/DbMessageCorrelationState.java
remcowesterhoud: I see. It's not a problem at this time as we won't get request data twice, but I'll change it to be future proof 👍

remcowesterhoud:
@nicpuppa Please have a look at the last commit that adds the copy 🙂

nicpuppa:
@remcowesterhoud are you adding also the missing test ?

remcowesterhoud:
@nicpuppa The test has arrived

nicpuppa (approved):
Thanks @remcowesterhoud

PR #20781, opened 2024-08-05
ci: improve critical issue workflow

nicpuppa:
## Description

Since the monorepo we need to differentiate critical issues related only to zeebe.

### DIFF
--- a/.github/workflows/critical-issue.yml
+++ b/.github/workflows/critical-issue.yml
@@ -10,30 +10,14 @@ jobs:
     name: Send slack notification if issue is labeled as critical
     runs-on: ubuntu-latest
     steps:
-      - id: label-check
-        name: Check if issue has component/* labels but not component/zeebe
-        run: |
-          echo "Debug: List of all labels:"
-          echo "${{ github.event.pull_request.labels.*.name }}"
-          echo "${{ github.event.issue.labels }}"
-
-          labels=$(echo "${{ github.event.pull_request.labels.*.name }}" | tr ' ' '\n' | grep '^component/' || true)
-          echo "Debug: component/* labels found:"
-          echo "$labels"
-
-          if [[ -n "$labels" ]]; then
-            if ! echo "$labels" | grep -q "component/zeebe"; then
-              echo "component/zeebe label not found among component/* labels."
-              echo "skip=true" >> $GITHUB_ENV
-            else
-              echo "component/zeebe label is present."
-            fi
-          else
-            echo "No component/* labels found."
-          fi
       - id: slack-notify
         name: Send slack notification
-        if: ${{ github.event.label.name == 'severity/critical' && env.skip != 'true' }}
+        if: >
+          contains(github.event.pull_request.labels.*.name, 'severity/critical') &&
+          (
+            contains(github.event.pull_request.labels.*.name, 'component/zeebe') ||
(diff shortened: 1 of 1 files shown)


.github/workflows/critical-issue.yml
```diff
@@ -1,17 +1,38 @@
 name: Slack notification for critical issue
 on:
   issues:
+  pull_request:
```
lenaschoenburg, line 4: :x: this would make it react on all issue events (before it was "labeled" only).

nicpuppa: Yeah, I know. But in this way I can easily test my changes 👍

nicpuppa:
🚀

nicpuppa:
I made some manual tests using this PR, I need to merge it first for testing this workflow with issues

.github/workflows/critical-issue.yml
```diff
@@ -8,10 +8,20 @@ jobs:
...
+        if: >
+          contains(github.event.issue.labels.*.name, 'severity/critical') &&
+          (
+            contains(github.event.issue.labels.*.name, 'component/zeebe') ||
+            (
+              !contains(github.event.issue.labels.*.name, 'component/operate') &&
+              !contains(github.event.issue.labels.*.name, 'component/optimize') &&
+              !contains(github.event.issue.labels.*.name, 'component/identity') &&
+              !contains(github.event.issue.labels.*.name, 'component/tasklist')
+            )
```
lenaschoenburg, line 23: Just so that we don't have to hard-code a full list of components:

```suggestion
            !contains(join(github.event.issue.labels.*.name, ", "), "component/")
```

.github/workflows/critical-issue.yml
```diff
@@ -8,10 +8,20 @@ jobs:
   slack-notify:
     name: Send slack notification if issue is labeled as critical
     runs-on: ubuntu-latest
-    if: ${{ github.event.label.name == 'severity/critical' }}
```
lenaschoenburg, line 11: :thought_balloon: The filter can remain on the job level and doesn't need to be moved to the `slack-notify` step, right?

lenaschoenburg (approved):

.github/workflows/critical-issue.yml
nicpuppa: I tried, but using a substring is not working. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#contains

.github/workflows/critical-issue.yml
nicpuppa: Oh yes 👍

.github/workflows/critical-issue.yml
lenaschoenburg: The important bit is the `join` to create a new string of all labels.

Based on https://github.com/camunda/camunda/compare/0a7156e22ad1a8d7f9b43c2b829268b0fc1cd711..ced1167fc3372734818e0ab2782dbca499ca7a0c
it looks like you only tried `startsWith(array, string)` which indeed does not work.

nicpuppa: Oh sorry I missed the join function 🙇 

This was my try with the substring https://github.com/camunda/camunda/pull/20781/commits/8a6a77b301e0dc5033ca9a04ba2a9cf7bb2404b7

PR #20792, opened 2024-08-05
fix: Fix Tasklist on iOS

vsgoulart:
## Description

Related to https://github.com/camunda/tasklist/pull/5008


Skaiir (approved):
Looks good code wise, some tests are red it seems

PR #20815, opened 2024-08-06
fix: always provide a response to command

nicpuppa:
## Description

Zeebe should always respond to a user command even if the result is that nothing changed otherwise the request will timeout.
The response to the user should be also reflected on the logstream so we are appending the event even if nothing changed

## Related issues

related to this [discussion](https://github.com/camunda/camunda/pull/20679#discussion_r1705253263)

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
@@ -43,7 +43,6 @@ public class JobUpdateProcessor implements TypedRecordProcessor<JobRecord> {
     final int retries = command.getValue().getRetries();
     final long timeout = command.getValue().getTimeout();
     final var job = jobState.getJob(key, command.getAuthorizations());
-    boolean isJobUpdated = false;
 
     if (job == null) {
       rejectionWriter.appendRejection(
@@ -57,7 +56,6 @@ public class JobUpdateProcessor implements TypedRecordProcessor<JobRecord> {
     if (retries > 0) {
       job.setRetries(retries);
       stateWriter.appendFollowUpEvent(key, JobIntent.RETRIES_UPDATED, job);
-      isJobUpdated = true;
     }
     // Handle timeout
     final long oldDeadline = job.getDeadline();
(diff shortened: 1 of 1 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
```diff
@@ -70,7 +70,7 @@ public void processRecord(final TypedRecord<JobRecord> command) {
 
     if (isJobUpdated) {
       stateWriter.appendFollowUpEvent(key, JobIntent.UPDATED, job);
-      responseWriter.writeEventOnCommand(key, JobIntent.UPDATED, job, command);
     }
+    responseWriter.writeEventOnCommand(key, JobIntent.UPDATED, job, command);
```
korthout, line 74: ❌ If we respond it to the user, we should also reflect it on the logstream with an event. Let's simply remove the `if (isJobUpdated)` statement.

korthout (commented):
Thanks @nicpuppa.

❌ I have a change request.

💭 Please describe the reasoning for making this change in the commit message, so future readers of the commit can understand why it was made without needing to navigate to github.

@tmetzke can you also have a look, so we're all aligned.

korthout:
>From the API perspective, users are not allowed to send an update request without providing at least one field between retries and timeout.

@nicpuppa I agree that users are not allowed to pass an invalid request, but we should consider idempotency of requests due to automatic retries. If a user tries to send the same update twice in a row, it's best if we OK both, rather than reject the second. The request was valid, but the data remains unchanged.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
tmetzke: Thanks, @korthout, I absolutely agree with this 👍

tmetzke:
> > From the API perspective, users are not allowed to send an update request without providing at least one field between retries and timeout.
> 
> @nicpuppa I agree that users are not allowed to pass an invalid request, but we should consider idempotency of requests due to automatic retries. If a user tries to send the same update twice in a row, it's best if we OK both, rather than reject the second. The request was valid, but the data remains unchanged.

I second this, idempotence is important here and we should OK a valid request that didn't change the internal state.

PR #20816, opened 2024-08-06
fix: Cancel active EL job on process termination

### LINKED ISSUE #20775: Cancel active EL jobs on process termination
Ensure that active Execution Listener jobs are properly cancelled when a process is terminated. 

The intended behaviour is that all EL jobs, regardless of the element type(`task, event, gateway, process instance, subprocess) they are associated with, should be cancelled to maintain consistency and proper resource management within the system. This change should prevent potential issues with orphaned jobs and ensure that the process state accurately reflects the termination.

ce-dmelnych:
## Description
This PR addresses an issue where active Execution Listener jobs were not being canceled upon process termination for some element types (events, gateways, subprocesses e.t.c). 
The fix involves updating the element termination logic to ensure that all active EL jobs are terminated when a process is canceled.

### Scope:
- moved job termination logic from individual `BpmnElementProcessor`-s to `BpmnStreamProcessor#processEvent -> case TERMINATE_ELEMENT`
- added tests to verify EL job cancellation behaviour by all element types that support EL definition:
  - tasks
  - events(start, intermediate throw, intermediate catch, end)
  - gateways
  - process instance
  - subprocesses

## Checklist

- [ ] for CI changes:

## Related issues

closes #20775

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/BpmnStreamProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/BpmnStreamProcessor.java
@@ -171,9 +171,6 @@ public final class BpmnStreamProcessor implements TypedRecordProcessor<ProcessIn
         break;
       case TERMINATE_ELEMENT:
         final var terminatingContext = stateTransitionBehavior.transitionToTerminating(context);
-        // cancel any active job associated with the element being terminated
-        // (e.g. execution listener, serviceTask job)
-        jobBehavior.cancelJob(terminatingContext);
         processor.onTerminate(element, terminatingContext);
         break;
       case COMPLETE_EXECUTION_LISTENER:
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/BpmnStreamProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/BpmnStreamProcessor.java
@@ -171,6 +171,9 @@ public final class BpmnStreamProcessor implements TypedRecordProcessor<ProcessIn
         break;
       case TERMINATE_ELEMENT:
         final var terminatingContext = stateTransitionBehavior.transitionToTerminating(context);
+        // cancel any active job associated with the element being terminated
+        // (e.g. execution listener, serviceTask job)
+        jobBehavior.cancelJob(terminatingContext);
         processor.onTerminate(element, terminatingContext);
         break;
       case COMPLETE_EXECUTION_LISTENER:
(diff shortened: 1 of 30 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/task/JobWorkerTaskProcessor.java
```diff
@@ -95,7 +95,6 @@ public Class<ExecutableJobWorkerTask> getType() {
   public void onTerminate(final ExecutableJobWorkerTask element, final BpmnElementContext context) {
     final var flowScopeInstance = stateBehavior.getFlowScopeInstance(context);
 
-    jobBehavior.cancelJob(context);
```
saig0, line 98: :wrench: We could keep the job canceling to be explicit about the behavior (here and in the other two processors). Otherwise, it may be confusing how jobs are canceled if you inspect the processor. 

For event subscriptions and incidents, we call the behavior explicitly in each processor. We should be consistent.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/BpmnStreamProcessor.java
```diff
@@ -171,6 +171,9 @@ private void processEvent(
         break;
       case TERMINATE_ELEMENT:
         final var terminatingContext = stateTransitionBehavior.transitionToTerminating(context);
+        // cancel any active job associated with the element being terminated
+        // (e.g. execution listener, serviceTask job)
+        jobBehavior.cancelJob(terminatingContext);
         processor.onTerminate(element, terminatingContext);
```
saig0, line 177: :wrench: Assuming that we keep the canceling behavior in the processors, we should flip the invocation and invoke the processor first to cancel the (service task) job within the processor. 

```suggestion
        processor.onTerminate(element, terminatingContext);
        // cancel any job associated with the element being terminated (e.g. execution listener)
        jobBehavior.cancelJob(terminatingContext);
```

saig0 (changes requested):
@ce-dmelnych looks good. :+1: 

I have one suggestion. Please have a look.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/BpmnStreamProcessor.java
ce-dmelnych: 🤔 After testing the suggested approach, I encountered an issue: invoking the job cancellation method after `processor.onTerminate` results in the following error:

```
Cannot invoke "io.camunda.zeebe.engine.state.instance.ElementInstance.getJobKey()" because "elementInstance" is null
```

This happens because `elementInstance` is deleted inside `ProcessInstanceElementTerminatedApplier#applyState`, which is triggered by `stateTransitionBehavior.transitionToTerminated` within the `processor.onTerminate` method.

Also, currently, it’s not possible to have multiple active jobs for an element; we can only have one—either a job associated with the element (e.g., a service task) or an EL job. 
We also clear `ElementInstance#jobKey` after each job completion.

---
Given these findings, @saig0 which approach would you prefer moving forward?
1. Adding `jobBehavior.cancelJob(context)` explicitly in each processor for elements that support ELs to maintain consistency.
2. Keeping the current implementation where the job canceling behavior is handled as it is now, centralised inside  `BpmnStreamProcessor`.

saig0 (approved):
@ce-dmelnych LGTM. :+1:

PR #20819, opened 2024-08-06
fix: operate indexWithRetries rebuild error

panagiotisgts:
## Description

_indexWithRetries_ function: build the IndexRequest outside of the retry function, since builder.build() leads to exception during service startup


ralfpuchert (approved):
Thank you. LGTM

PR #20827, opened 2024-08-06
refactor: job update processors

nicpuppa:
## Description

- Introduced a `JobUpdateBehaviour` to handle all the common operation between the Job Update processors to align the behaviour of the 3 different processors
- Add rejection if no updates are made in `JobUpdateProcessor`

## Related issues

follow up PR of https://github.com/camunda/camunda/pull/20679

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnBehaviors.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnBehaviors.java
@@ -12,6 +12,7 @@ import io.camunda.zeebe.engine.processing.common.CatchEventBehavior;
 import io.camunda.zeebe.engine.processing.common.ElementActivationBehavior;
 import io.camunda.zeebe.engine.processing.common.EventTriggerBehavior;
 import io.camunda.zeebe.engine.processing.common.ExpressionProcessor;
+import io.camunda.zeebe.engine.processing.job.behaviour.JobUpdateBehaviour;
 import io.camunda.zeebe.engine.processing.variable.VariableBehavior;
 
 public interface BpmnBehaviors {
@@ -55,4 +56,6 @@ public interface BpmnBehaviors {
   BpmnUserTaskBehavior userTaskBehavior();
 
   BpmnCompensationSubscriptionBehaviour compensationSubscriptionBehaviour();
+
+  JobUpdateBehaviour jobUpdateBehaviour();
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
@@ -45,7 +45,7 @@ public class JobUpdateProcessor implements TypedRecordProcessor<JobRecord> {
               jobUpdateBehaviour.updateJobRetries(jobKey, job, command).ifPresent(errors::add);
               final long timeout = command.getValue().getTimeout();
               // if no timeout is provided (the default value is -1L), no update
-              if (timeout > -1L) {
+              if (timeout > 0) {
                 jobUpdateBehaviour.updateJobTimeout(jobKey, job, command).ifPresent(errors::add);
               }
               if (errors.isEmpty()) {
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
@@ -7,70 +7,34 @@
  */
 package io.camunda.zeebe.engine.processing.job;
 
+import io.camunda.zeebe.engine.processing.job.behaviour.JobUpdateBehaviour;
 import io.camunda.zeebe.engine.processing.streamprocessor.TypedRecordProcessor;
-import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
-import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedRejectionWriter;
-import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedResponseWriter;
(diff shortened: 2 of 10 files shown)


nicpuppa:
💭 I was thinking about rejecting the command if nothing is changed. On API side, users are not allowed to send an update request without providing at least one field between retries and timeout.

It make sense to align this also in the processor

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
```diff
@@ -7,70 +7,34 @@
...
-    if (isJobUpdated) {
-      stateWriter.appendFollowUpEvent(key, JobIntent.UPDATED, job);
-      responseWriter.writeEventOnCommand(key, JobIntent.UPDATED, job, command);
+    final var updatedRetries = jobUpdateBehaviour.updateJobRetries(jobKey, jobRecord, command);
+    final var updatedTimeout = jobUpdateBehaviour.updateJobTimeout(jobKey, jobRecord, command);
+    // if retries or timeout are updated the command is rejected
+    if (updatedRetries || updatedTimeout) {
+      jobUpdateBehaviour.completeJobUpdate(jobKey, jobRecord, command);
+    } else {
+      jobUpdateBehaviour.rejectJobUpdate(command);
```
korthout, line 37: ❌ https://github.com/camunda/camunda/pull/20815#issuecomment-2271481770

tmetzke (changes requested):
👍 I like that the job update processors are aligned now, this leads to better stability and maintainability for this part of job handling.

🔧 We have all logic in the new behavior class now, even behavior specific to only one of the processors. This leads to an unnecessarily large behavior class that is harder to understand.

❌ If I'm not mistaken, we are still sending multiple rejections in case both retries and timeout are invalid. We should only send one response per command, i.e. one success or one rejection message. Furthermore, with the refactoring, I believe we currently write success responses even in rejection cases, so we write both. My impression is that we refactored this a bit "too far" so we don't easily see which responses we send. I added a proposal to hopefully overcome this.

💭 I could imagine the following base structure after this refactoring:

* The processors write responses (success or rejection), not the behavior class. The behavior class changes state and creates follow-up events.
  * Looking at a processor, it is easily visible which responses it creates. The method contract (i.e. the input and output) is simpler to understand. Otherwise, it's hidden in (potentially multiple and nested) behavior methods which makes it harder to understand.
  * The processors need to write different responses. Hiding this in the behavior class makes this harder to follow.
* The behavior class methods return the data needed for the processors to decide how to move on and respond:
  * The processors take care of finding a job in the state using the command (this could be one common method the behavior class provides, e.g. `Either<String, JobRecord> getJob(final TypedRecord<JobRecord> command)`).
  * The processors take care of writing a rejection (if no job was found) or moving on (trying to update the state).
  * Each processor invokes `updateRetries` and/or `updateTimeout` methods from the behavior class as required. The behavior class methods return an `Optional<String>` with an error message the processors can use in rejections. If an empty `Optional` is returned, the update is considered successful (no matter if the state actually changed or not).
  * The processors create their success or rejection response, depending on whether empty `Optionals` were returned by the behavior methods. The update processor simply concatenates the potential error messages from retries and timeout updates.

Let me know if this makes sense, happy to also jump on a call or contribute to the PR myself if that helps you in any way.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnBehaviors.java
```diff
@@ -55,4 +56,6 @@ public interface BpmnBehaviors {
   BpmnUserTaskBehavior userTaskBehavior();
 
   BpmnCompensationSubscriptionBehaviour compensationSubscriptionBehaviour();
+
+  JobUpdateBehaviour jobUpdateBehaviour();
```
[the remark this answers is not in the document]
tmetzke: I thought so as well but it actually doesn't really look like it, to be honest. But it might also be something we need to clean up, so those behaviors shouldn't be called `BpmnJob...Behavior`. I am not entirely sure to be honest.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
```diff
@@ -9,19 +9,59 @@
...
+        .getJob(jobKey, command)
+        .ifRightOrLeft(
+            job -> {
+              final List<String> errors = new ArrayList<>();
+              jobUpdateBehaviour.updateJobRetries(jobKey, job, command).ifPresent(errors::add);
+              final long timeout = command.getValue().getTimeout();
+              // if no timeout is provided (the default value is -1L), no update
+              if (timeout > -1L) {
+                jobUpdateBehaviour.updateJobTimeout(jobKey, job, command).ifPresent(errors::add);
+              }
```
tmetzke, line 50: 🔧 Whatever the timeout update validation logic is, should it be the same for `UPDATE` and `UPDATE_TIMEOUT` commands? Currently, we have an additional validation step here in the `UPDATE` processor.

nicpuppa: The problem is that we are telling the users that:

`Providing any of those attributes with a null value or omitting it preserves the persisted attribute’s value.`

This is not possible if we don't add this additional validation

tmetzke: From the REST controller, we send a `0` if nothing is provided by the user, right?

nicpuppa: Yes, so I need to change the validation 👍

nicpuppa:
As discussed with @tmetzke, we are going to merge this PR even if the validation in the engine is not accurate. The purpose of this PR is just refactoring/restructuring the processors. 

A follow-up PR is coming to adjust validation, both on broker and controller side 👍

tmetzke (approved):

nicpuppa:
What is going on with the go client ❓ 👀

Issue #20728, opened 2024-08-02
Retry open of exporter on failure


npepinpe:
**Description**

It's quite common for exporter authors to perform some set up during the `#open()` phase of the exporter. For example, you might set up your database schema. Unfortunately, this is not retried right now, so what happens is if it fails (e.g. database is down), the exporter is not opened, and thus nothing is exported. This prevents other exporters from also exporting.

The workaround is perform this set up in the `export` method and cache the result, since there it is retried.

What to do here:

- The `#open()` call is retried with the same logic as the `export` call, that is with a back off.

To be resolved/question:

- When we enable an exporter, we also call its open method. This is may need to be moved out, but that changes the semantics of the enable call. Alternatively, since enabling is retried, it may be that we don't need to retry this particular open call, but only the one done when initially starting the exporter director.

Tests:
- An exception thrown in `open` causes `open` to be called again
- An exporter which throws an exception in `open` once, but later does not, can export eventually

As a follow up, either as separate PRs for this issue, or as follow up issues (which you would write), please:

- Move the schema creation part of the Elasticsearch Exporter to the `open` call
- Move the schema creation part of the Opensearch Exporter to the `open` call

PR #20836, opened 2024-08-06
Retry of Exporter open

EuroLew:
## Description

This PR will only handle code path (1) referenced here https://github.com/camunda/camunda/issues/20728#issuecomment-2271370664 and make sure that the `openExporter` call is retried if it fails when initialising exporters through configuration.

## Related issues

relates to #20728

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
@@ -490,7 +490,18 @@ public final class ExporterDirector extends Actor implements HealthMonitorable, 
   private void startActiveExportingMode() {
     for (final ExporterContainer container : containers) {
       container.initMetadata();
-      container.openExporter();
+      new BackOffRetryStrategy(actor, Duration.ofSeconds(10))
+          .runWithRetry(
+              () -> {
+                try {
+                  container.openExporter();
+                  return true;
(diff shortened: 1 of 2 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
```diff
@@ -490,7 +490,18 @@ private void becomeLive() {
...
+              () -> {
+                try {
+                  container.openExporter();
+                  return true;
+                } catch (final Exception e) {
+                  LOG.error("Failed to open exporter '{}'. Retrying...", container.getId(), e);
+                  return false;
+                }
+              },
+              this::isClosed);
```
deepthidevaki, line 504: :x: As we discussed before, we have to wait until all exporters are opened before we can proceed with the next step. With the current code, we might start exporting before the open is completed. This can lead to undefined behavior. 

Please also add a test to verify this once it is fixed.

deepthidevaki (changes requested):
Please check my comments. Let me know if you want to discuss it further.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
EuroLew: done

deepthidevaki (approved):
Thanks.

PR #20837, opened 2024-08-06
Retry of exporter open during transition step

### LINKED ISSUE #20728: Retry open of exporter on failure

EuroLew:
## Description

This PR will only handle code path (3) referenced here https://github.com/camunda/camunda/issues/20728#issuecomment-2271370664 and make sure that the openExporter call is retried if it fails when initialising exporters due to the context changing between the time an asynchronous exporter director creation call is made and when the future completes.

## Related issues

closes #20728

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
@@ -259,26 +259,53 @@ public final class ExporterDirector extends Actor implements HealthMonitorable, 
       return CompletableActorFuture.completed(null);
     }
 
-    return actor.call(
-        () -> {
-          containers.stream()
-              .map(ExporterContainer::getId)
-              .filter(exporterId::equals)
-              .findFirst()
-              .ifPresentOrElse(
-                  container -> {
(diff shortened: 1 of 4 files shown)


deepthidevaki (approved):
:rocket:

PR #20885, opened 2024-08-07
ci: create workflow to push preview images to harbor

andromaqui:
## Description
 
This pr aims to:
- Add a new workflow called `preview-env-build.yml ` which is going to responsible for building the docker images for `optimize`, `operate`, `tasklist` & `zeebe` and pushing team to the harbor project `team-camunda`
- To achieve this the reusable workflows for tasklist & operate have been modified as well so that they can push to harbor based on an input variable called `previewenv`.  
- A new reusable workflow has been introduced for `optimize`
- Similarly the reusable workflow to build and push the zeebe docker image has been modified to log in to harbor in case of a preview environment build.

To be merged after:
- [x] https://github.com/camunda/infra-core/pull/8186

### Test1: Ensure the workflow pushes the images to dev harbor 
- [commit](https://github.com/camunda/camunda/pull/20885/commits/edc159d44351d4ae31111e0afb69058cb462d43b)
- Link to dev harbor project https://dev.registry.camunda.cloud/harbor/projects/91/repositories/operate/artifacts-tab 
- image tag is `pr-60433bcf6cbeff1984183668cb1a28eaa9860a77`

### Test2: Ensure nothing gets pushed to dockerhub
- https://hub.docker.com/r/camunda/tasklist/tags
- https://hub.docker.com/r/camunda/zeebe/tags
- https://hub.docker.com/r/camunda/operate/tags
## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/team-infrastructure/issues/655

### DIFF
--- a/.github/workflows/operate-ci-build-reusable.yml
+++ b/.github/workflows/operate-ci-build-reusable.yml
@@ -20,9 +20,13 @@ on:
         required: true
         type: string
       pushDocker:
-        description: "whether the built docker images are pushed to camunda registry"
+        description: "Whether the built docker images are pushed to camunda registry"
         type: boolean
         default: true
+      previewEnv:
+        description: "Defines whether this is a preview environment build or not. If true, the image will be pushed to the Harbor Docker registry."
+        type: boolean
+        default: false
 
 defaults:
   run:
@@ -110,7 +114,7 @@ jobs:
       - name: Build Docker image
         uses: ./.github/actions/build-platform-docker
         with:
-          repository: registry.camunda.cloud/team-operate/camunda-operate
+          repository: ${{ inputs.previewEnv && 'registry.camunda.cloud/team-camunda/operate' || 'registry.camunda.cloud/team-operate/camunda-operate' }}
           version: |
             ${{ env.IMAGE_TAG }}
             ${{ env.CI_IMAGE_TAG }}
--- a/.github/workflows/optimize-ci-build-reusable.yml
+++ b/.github/workflows/optimize-ci-build-reusable.yml
@@ -22,10 +22,6 @@ on:
       pushDocker:
         description: "Whether the built docker images are pushed to camunda registry"
         type: boolean
-        default: true
-      previewEnv:
-        description: "Defines whether this is a preview environment build or not. If true, the image will be pushed to the Harbor Docker registry."
-        type: boolean
         default: false
 
 defaults:
@@ -54,11 +50,7 @@ jobs:
           GIT_COMMIT_HASH=$(git rev-parse ${{ inputs.branch }})
           echo "GIT_COMMIT_HASH=$GIT_COMMIT_HASH" >> $GITHUB_ENV
           echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
-          if [ "${{ inputs.previewEnv }}" = "true" ]; then
-            echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
-          else
-            echo "IMAGE_TAG=registry.camunda.cloud/team-optimize/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
-          fi
+          echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
 
       #########################################################################
       # Setup: read java version from pom
@@ -99,4 +91,4 @@ jobs:
             VERSION=${{ steps.pom-info.outputs.project_version }}
             REVISION=${{ env.GIT_COMMIT_HASH }}
           tags: ${{ env.IMAGE_TAG }}
-          push: true
+          push: ${{ inputs.pushDocker }}
--- a/.github/workflows/optimize-ci-build-reusable.yml
+++ b/.github/workflows/optimize-ci-build-reusable.yml
@@ -50,7 +50,6 @@ jobs:
           GIT_COMMIT_HASH=$(git rev-parse ${{ inputs.branch }})
           echo "GIT_COMMIT_HASH=$GIT_COMMIT_HASH" >> $GITHUB_ENV
           echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
-          echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
 
       #########################################################################
       # Setup: read java version from pom
--- a/.github/workflows/optimize-ci-build-reusable.yml
+++ b/.github/workflows/optimize-ci-build-reusable.yml
@@ -1,0 +1,102 @@
+# This GitHub Actions workflow automates the CI build process for the 'optimize' service.
+# It triggers on a `workflow_call` event and accepts inputs for branch name[required], Java version[optional]
+#
+# It consists of a several steps:
+# 1. Setup: Checks out the branch, defines environment variables, fetches the main branch, reads Java/Maven project info, logs into the Harbor registry, and sets up Maven.
+# 2. Build: Generates a production .tar.gz file using Maven and builds a Docker image with appropriate tags.
+# 3. Upload: Pushes the Docker image to the Harbor registry tagged for the pull request.
+#
+# Environment variables are used to define the image tags used for Docker builds.
+# This workflow is designed to provide a comprehensive, automated CI process that ensures code quality, handles secrets securely,
+# and enables detailed reporting of test results.
+
+name: Optimize CI Reusable
+
+on:
+  workflow_call:
+    inputs:
+      branch:
+        description: "The branch name to be used for the workflow"
+        required: true
+        type: string
+      pushDocker:
+        description: "Whether the built docker images are pushed to camunda registry"
+        type: boolean
+        default: true
+      previewEnv:
+        description: "Defines whether this is a preview environment build or not. If true, the image will be pushed to the Harbor Docker registry."
+        type: boolean
+        default: false
+
+defaults:
+  run:
+    # use bash shell by default to ensure pipefail behavior is the default
+    # see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference
+    shell: bash
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      # Setup: checkout branch
+      - name: Checkout '${{ inputs.branch }}' branch
+        uses: actions/checkout@v4
+        with:
+          ref: refs/heads/${{ inputs.branch }}
+          fetch-depth: 0 # fetches all history for all branches and tags
+
+      #########################################################################
+      # Setup: define env variables
+      - name: Set GitHub environment variables
+        run: |
+          GIT_COMMIT_HASH=$(git rev-parse ${{ inputs.branch }})
+          echo "GIT_COMMIT_HASH=$GIT_COMMIT_HASH" >> $GITHUB_ENV
+          echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
+          if [ "${{ inputs.previewEnv }}" = "true" ]; then
+            echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
+          else
+            echo "IMAGE_TAG=registry.camunda.cloud/team-optimize/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
+          fi
+
+      #########################################################################
+      # Setup: read java version from pom
+      - name: "Read Java / Version Info"
+        id: pom-info
+        uses: YunaBraska/java-info-action@72dd3fb638d774fb4353d99d2bb3fddddd084789 # 2.1.0
+
+      #########################################################################
+      # Docker login and image tagging
+      - name: Login to Harbor registry
+        uses: ./.github/actions/login-registry
+        with:
+          secrets: ${{ toJSON(secrets) }}
+
+      #########################################################################
+      # Setup and configure Maven
+      - name: Setup Maven
+        uses: ./.github/actions/setup-maven
+        with:
+          secrets: ${{ toJSON(secrets) }}
+
+      #########################################################################
+      # Build Optimize
+      - name: Generate production .tar.gz
+        uses: ./.github/actions/run-maven
+        with:
+          parameters: -f optimize/pom.xml install -DskipTests -Dskip.docker -PrunAssembly
+
+      #########################################################################
+      # Build Docker image
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ca052bb54ab0790a636c9b5f226502c73d547a25 # v5
+        with:
+          context: .
+          file: optimize.Dockerfile
+          provenance: false
+          build-args: |
+            VERSION=${{ steps.pom-info.outputs.project_version }}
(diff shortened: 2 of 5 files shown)


.github/workflows/preview-env-build.yml
```diff
@@ -0,0 +1,71 @@
+---
+name: Preview Environment Build
+
+on:
+  pull_request:
+    types: [ labeled,synchronize ]
+
+concurrency:
+  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref_name }}
```
cmur2, line 10: What should this specific `concurrency` setting achieve for this workflow, can you describe it and put a comment into the file for future readers? [example](https://github.com/camunda/camunda/blob/67f8fbf3948880dd580c1b1fc51b3943afd7f839/.github/workflows/operate-a11y.yml#L16-L17)

.github/workflows/preview-env-build.yml
```diff
@@ -0,0 +1,71 @@
...
+      previewEnv: true
+      commitSha: ${{ github.sha }}
+
+  build-zeebe:
+    if: github.event.pull_request.state != 'closed' && (contains( github.event.label.name, 'deploy-preview') || contains( github.event.pull_request.labels.*.name, 'deploy-preview'))
+    name: Zeebe Build
+    runs-on: ubuntu-latest
+    concurrency:
+      group: deploy-docker-snapshot
+      cancel-in-progress: false
```
cmur2, line 45: I think this `concurrency` setting on the job level needs to be removed since it is redundant since this job is not pushing the `SNAPSHOT` Docker image and will block other real `deploy-docker-snapshot` jobs.

.github/workflows/preview-env-build.yml
```diff
@@ -0,0 +1,71 @@
...
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+
+      - uses: ./.github/actions/build-zeebe
+        id: build-zeebe
+        with:
+          maven-extra-args: -T1C -PskipFrontendBuild
+
+      - uses: ./.github/actions/build-platform-docker
+        id: build-zeebe-docker
```
cmur2, line 61: ID of this step not used later, can be removed

.github/actions/build-platform-docker/action.yml
```diff
@@ -98,6 +111,31 @@ runs:
...
+          secret/data/github.com/organizations/camunda NEXUS_USR;
+          secret/data/github.com/organizations/camunda NEXUS_PSW;
+
+    - name: Login to Harbor docker registry
+      uses: docker/login-action@v3
+      if: inputs.preview-env == 'true'
+      with:
+        registry: registry.camunda.cloud
+        username: ${{ steps.secrets.outputs.NEXUS_USR }}
+        password: ${{ steps.secrets.outputs.NEXUS_PSW }}
```
cmur2, line 137: This logic can be moved outside of this file since the `build-platform-docker` action does not need to know about secrets and authentication. See for example how `operate-ci-build-reusable` does the Vault secret handling and login to Harbor before calling `build-platform-docker`. The same approach should be used for `preview-env-build` where you can get secrets and login to Harbor before calling this action - this is cleaner separation of concerns.

The new inputs above are also unnecessary then.

.github/workflows/operate-ci-build-reusable.yml
```diff
@@ -53,11 +61,10 @@ jobs:
       # Setup: define env variables
       - name: Set GitHub environment variables
         run: |
-          GIT_COMMIT_HASH=$(git rev-parse ${{ inputs.branch }})
```
cmur2, line 56: Can you explain why this change is necessary and `inputs.commitSha` needs to be introduced?

cmur2 (commented):
I have some questions about the new GHA code and an architectural change we should do to make it cleaner.

.github/workflows/preview-env-build.yml
andromaqui: ah yeah, good point!

.github/actions/build-platform-docker/action.yml
andromaqui: I created a new reusable workflow called `zeebe-ci-build-reusable.yml` and another one for Optimize as well

.github/workflows/operate-ci-build-reusable.yml
andromaqui: In my understanding the `{{ github.sha }}` and  `git rev-parse branch` deliver different commit hashes. As example for instance in the last commit of this workflow I have set all workflows to use `github.sha`. However, the operate workflow uses rev parse and this results in different tags being generated (you can inspect for instance the last images pushed to [dev harbor](https://dev.registry.camunda.cloud/harbor/projects/91/repositories))

andromaqui: I was under the impression that this was a mistake in the implementation, but what is your view on this? Do you know the intention behind the `git rev-parse branch` command in this case? For the preview environmnets we definitely need the `{{ github.sha }}` as we need the hash of the latest commit.

andromaqui: Another thing that I noticed is that optimize actually tags its images with [github.event.pull_request.head_sha ](https://github.com/camunda/camunda/blob/bdc52631ca2ae1421619902e03662ab015c3ec1b/.github/workflows/tasklist-ci.yml#L33) but operate and tasklist use `git rev-parse branch` in their ci. Should these be in line? This is not directly connected to my changes but it is maybe important in order to get the logic of the reusable workflows right

cmur2: > In my understanding the {{ github.sha }} and git rev-parse branch deliver different commit hashes.

Which is the reason why I noticed this change.

> For the preview environmnets we definitely need the {{ github.sha }} as we need the hash of the latest commit.

According to [GH docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) the `github.sha` is:

> The commit SHA that triggered the workflow. The value of this commit SHA depends on the event that triggered the workflow. For more information, see "[Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows)."

So on [pull request events](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) it is

> Last merge commit on the GITHUB_REF branch

Which is not the latest commit on the branch but instead an automatically by GH created merge commit that includes the changes from `main`.

This is the reason why `git rev-parse branch` or `github.event.pull_request.head.sha` are used. [GH docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) also mention:

> Note that GITHUB_SHA for this event is the last merge commit of the pull request merge branch. If you want to get the commit ID for the last commit to the head branch of the pull request, use github.event.pull_request.head.sha instead.

cmur2: > Should these be in line?

Yes they could if the context is correct - the reusable*.yml workflows accept a `branch` input so they probably cannot use `github.event.pull_request.head.sha` since they don't know from which context they will be called.

.github/workflows/zeebe-ci-build-reusable.yml
```diff
@@ -0,0 +1,91 @@
+# This GitHub Actions workflow automates the CI build process for the 'zeebe' service.
```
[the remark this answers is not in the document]
cmur2: Same applies to `optimize-ci-build-reusable`, having a dedicated workflow that is only called by 1 place is not worth the overhead of declaring variables etc.

.github/workflows/zeebe-ci-build-reusable.yml
```diff
@@ -0,0 +1,91 @@
...
+          password: ${{ steps.secrets.outputs.NEXUS_PSW }}
+
+      - uses: ./.github/actions/build-platform-docker
+        with:
+          repository: dev.registry.camunda.cloud/team-camunda/zeebe
+          version: pr-${{ github.sha }}
+          platforms: ${{ env.DOCKER_PLATFORMS }}
+          push: true
+          distball: ${{ steps.build-zeebe.outputs.distball }}
+          secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
```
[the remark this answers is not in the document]
andromaqui: sure! I am still making changes on this pr, havent officially put it in review again thats why I havent fixed the linter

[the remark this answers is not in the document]
cmur2: ah ok no worries then

.github/workflows/zeebe-ci-build-reusable.yml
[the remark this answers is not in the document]
andromaqui: I think for Optimize the logic is being reused but not for Zeebe

.github/workflows/operate-ci-build-reusable.yml
andromaqui: sounds good!

.github/workflows/optimize-ci-build-reusable.yml
```diff
@@ -0,0 +1,102 @@
...
+name: Optimize CI Reusable
+
+on:
+  workflow_call:
+    inputs:
+      branch:
+        description: "The branch name to be used for the workflow"
+        required: true
+        type: string
+      pushDocker:
```
cmur2, line 22: This input is not used in the action code below. Please either use it or remove the input.

.github/workflows/optimize-ci-build-reusable.yml
```diff
@@ -0,0 +1,102 @@
...
+      # Setup: define env variables
+      - name: Set GitHub environment variables
+        run: |
+          GIT_COMMIT_HASH=$(git rev-parse ${{ inputs.branch }})
+          echo "GIT_COMMIT_HASH=$GIT_COMMIT_HASH" >> $GITHUB_ENV
+          echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
+          if [ "${{ inputs.previewEnv }}" = "true" ]; then
+            echo "IMAGE_TAG=registry.camunda.cloud/team-camunda/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
+          else
+            echo "IMAGE_TAG=registry.camunda.cloud/team-optimize/optimize:pr-$GIT_COMMIT_HASH" >> $GITHUB_ENV
```
cmur2, line 60: @andromaqui this logic here was carried over from the Optimize repository and Docker images called `registry.camunda.cloud/team-optimize/optimize:pr*` are only used for the preview environments. There is no need to preserve this logic in the monorepo. Outside of `inputs.previewEnv == true` there is no need to push any Docker image here.

.github/workflows/optimize-ci.yml
```diff
@@ -56,47 +56,12 @@ jobs:
...
-        provenance: false
-        build-args: |
-          VERSION=${{ steps.pom-info.outputs.project_version }}
-          REVISION=${{ github.event.pull_request.head.sha }}
-        tags: "registry.camunda.cloud/team-optimize/optimize:pr-${{ github.event.pull_request.head.sha }}"
-        push: true
+    uses: ./.github/workflows/optimize-ci-build-reusable.yml
+    secrets: inherit
+    with:
+      branch: ${{ github.head_ref }}
```
cmur2, line 64: We should make the Optimize team aware of this change when finished since it is notable.

.github/workflows/optimize-ci-build-reusable.yml
```diff
@@ -0,0 +1,102 @@
...
+        uses: docker/build-push-action@ca052bb54ab0790a636c9b5f226502c73d547a25 # v5
+        with:
+          context: .
+          file: optimize.Dockerfile
+          provenance: false
+          build-args: |
+            VERSION=${{ steps.pom-info.outputs.project_version }}
+            REVISION=${{ env.GIT_COMMIT_HASH }}
+          tags: ${{ env.IMAGE_TAG }}
+          push: true
```
cmur2, line 102: Should be taken from `inputs.pushDocker` or even better from `inputs.previewEnv` since otherwise this workflow does not need to push anything.

.github/workflows/optimize-ci-build-reusable.yml
andromaqui: I am not very about this, the [code](https://github.com/camunda/camunda/pull/20885/files#diff-9232e03df9f79b5a046a26e4ca02fa62410d4a501a9b79be5b1f577358ef3f4dL64-L99) which previously was on optimize-ci.yml was tagging and pushing this image to `registry.camunda.cloud/team-optimize/optimize:pr`. Do you think this was a mistake?

.github/workflows/optimize-ci-build-reusable.yml
andromaqui: Yeah you are right! I think it would be better to do it through the` inputs.pushDocker` variable as in my understanding the optimize-ci workflow was also [pushing the image](https://github.com/camunda/camunda/pull/20885/files#diff-9232e03df9f79b5a046a26e4ca02fa62410d4a501a9b79be5b1f577358ef3f4dL64-L99)

.github/workflows/optimize-ci-build-reusable.yml
cmur2: I think it was unused since Docker images tagged with `pr-*` are used only for preview environments, see https://github.com/search?q=repo%3Acamunda%2Fcamunda-optimize+pr-&type=code

Since the preview environments that you are building do not use `registry.camunda.cloud/team-optimize/optimize` but instead use `registry.camunda.cloud/team-camunda/optimize` the old logic is useless, yes.

andromaqui: The [optimize-ci workflow](https://github.com/camunda/camunda/blob/main/.github/workflows/optimize-ci.yml#L58) which I attached though is part of the camunda/camunda repo and runs on every pr associated with Optimize changes, is it really unused? In there there image is being tagged with [pr-${{ github.event.pull_request.head.sha }}](https://github.com/camunda/camunda/blob/main/.github/workflows/optimize-ci.yml#L98) I believe.

The [optimize-ci-build-reusable.yml](https://github.com/camunda/camunda/pull/20885/files/56ff44a9f750fdd506fa76d16a8949d936f7c57d#diff-a6380d6c48fac8323a2728ce2b1e095371ee07ab8885fd8af769b4e38f8ddedd) is being used from both the `optimize-ci `and the preview environment workflow

cmur2: The workflow is used for sure but the specific line I commented on contains logic that is unused: namely this to tag Docker images with `registry.camunda.cloud/team-optimize/optimize:pr-*` and push them. Nobody consumes them so pushing them is useless (in fact it costs Camunda money for nothing, not your fault) and if we don't need to push we also don't need that `else` condition to assign that tag.

.github/workflows/optimize-ci-build-reusable.yml
cmur2: > was also [pushing the image](https://github.com/camunda/camunda/pull/20885/files#diff-9232e03df9f79b5a046a26e4ca02fa62410d4a501a9b79be5b1f577358ef3f4dL64-L99)

Yea but nobody ever uses the image because it is meant to be used for the old Optimize 7 Preview Environments in the old `camunda/camunda-optimize` hence my reference to being "unused".

.github/workflows/optimize-ci-build-reusable.yml
andromaqui: Alrght, thank you for the explanation! That would change the logic which exists currently in the optimize-ci workflow, is this something we need to warn the Optimize team about?

.github/workflows/optimize-ci.yml
andromaqui: I will let them know, but since we dont need to push the image anymore this might not make any difference to them

.github/workflows/optimize-ci-build-reusable.yml
andromaqui: Exactly! I now understand what you meant previously :)

.github/workflows/optimize-ci-build-reusable.yml
cmur2: As we need to tell them anyways about this bigger change to their pipeline, we should mention this as well 👍🏼

cmur2 (approved):

PR #20886, opened 2024-08-07
Exclude xml field from DecisionRequirements search query

houssain-barouni:
## Description

Related to. https://github.com/camunda/camunda/pull/20708 
showcase how we can exclude xml field from DecisionRequirements search query results. The xml payload can be big and I think it is good to exclude from ES results especially that it is not returned in the API response.

Also includes https://github.com/camunda/camunda/pull/20905

## Checklist

- [ ] for CI changes:

## Related issues

closes #


ralfpuchert (approved):
Thanks for bringing this up. Am I right this is part of the projection feature where the user decides which fields should be in the result?

service/src/main/java/io/camunda/service/exception/NotFoundException.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.service.exception;
+
+public class NotFoundException extends RuntimeException {
```
tmetzke, line 10: 🔧 If we want to create a specific exception for the services (which might make sense here), we could build this on top of https://github.com/camunda/camunda/blob/main/service/src/main/java/io/camunda/service/CamundaServiceException.java. We use that exception class to determine if the error from the service was a specific broker exception or just any arbitrary exception, for example. We can also extend the `CamundaServiceException`, of course, if we want to differentiate that this was a "search exception", for example.

service/src/main/java/io/camunda/service/search/query/DecisionRequirementsQuery.java
```diff
@@ -74,7 +95,8 @@ protected Builder self() {
     public DecisionRequirementsQuery build() {
       filter = Objects.requireNonNullElse(filter, EMPTY_FILTER);
       sort = Objects.requireNonNullElse(sort, EMPTY_SORT);
-      return new DecisionRequirementsQuery(filter, sort, page());
+      resultConfig = Objects.requireNonNullElse(resultConfig, EXCLUDE_XML_RESULT_CONFIG);
```
tmetzke, line 98: ❓ I might have missed it: Do we test this default behavior explicitly already, i.e. there is no XML returned although it exists in the entity if we don't explicitly include the `xml` attribute?

tmetzke (commented):
👍 I really like how you can include/exclude attributes upon search and define them internally (for now). This is not exposed to users yet so they cannot make use of this in general search queries but I can easily see that we could support this in the future 👍

service/src/main/java/io/camunda/service/exception/NotFoundException.java
houssain-barouni: I made `NotFoundException` extending `CamundaServiceException`. It can be used in the endpoints requiring an object key to get or to mutate this object

service/src/main/java/io/camunda/service/search/query/DecisionRequirementsQuery.java
houssain-barouni: zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/DecisionRequirementsQueryControllerTest.java asserts that we always set the exclude filter on `xml` attribute

Checking that the `xml` attribute is actually filtered out in the result, require an integration test with real ES instance, however we cannot do it with client integration tests as the field is not part of the API
I will check how we can do that

PR #20888, opened 2024-08-07
test: SearchDecisionDefinitionTest integration tests

houssain-barouni:
## Checklist

- [ ] for CI changes:

## Related issues


nathansandi:
💭  Hey @houssain-baroun, thanks for add the Integration tests. 
I would suggest here add some more test cases, involving pagination, searchBefore/after and Sorting case, in order to validate the  Client & Controller e2e with those cases. I had cases over development that worked well on unit tests but discovered errors on those functionalities over the integration tests.

houssain-barouni:
> 💭 Hey @houssain-baroun, thanks for add the Integration tests. I would suggest here add some more test cases, involving pagination, searchBefore/after and Sorting case, in order to validate the Client & Controller e2e with those cases. I had cases over development that worked well on unit tests but discovered errors on those functionalities over the integration tests.

I think these case should be rather covered by unit tests, here the focus is more checking the E2E flow is working
Deploy -> Export -> Import -> Search with client.
The integration tests takes more time to run, so there is a big risk we quickly reach the CI jobs configured timeouts, the recommended CI execution time is 15 minutes max

nathansandi:
> > 💭 Hey @houssain-baroun, thanks for add the Integration tests. I would suggest here add some more test cases, involving pagination, searchBefore/after and Sorting case, in order to validate the Client & Controller e2e with those cases. I had cases over development that worked well on unit tests but discovered errors on those functionalities over the integration tests.
> 
> I think these case should be rather covered by unit tests, here the focus is more checking the E2E flow is working Deploy -> Export -> Import -> Search with client. The integration tests takes more time to run, so there is a big risk we quickly reach the CI jobs configured timeouts, the recommended CI execution time is 15 minutes max

I understand the concern regarding the CI job timeouts and the preference for keeping the integration tests focused on the E2E flow. However, I believe it is crucial to include test cases involving pagination, searchBefore/after, and sorting within the integration tests for the following reasons:
- Transformers and Integration: While unit tests are excellent for ensuring that individual components, such as the request application and mocks, work as expected, they do not fully validate the behavior of Transformers. Transformers play a critical role in converting application logic into appropriate Elasticsearch (ES) queries. If there are issues in how Transformers handle pagination, sorting, or searchBefore/after operations, these problems may go unnoticed with only unit tests. Integration tests ensure that these transformations are correct and that the system interacts with ES as expected.
- Complex Interactions: Operations like searchAfter, page limits, and sorting involve complex interactions between multiple components. For instance, a searchAfter parameter might need to be correctly applied across paginated results, and sorting must be consistent with ES's expectations.

This would imply in 2 more tests cases, or could be done in the same, and would validate the complex transformers for the ES/OS Queries, and would not impact on the IT timings significantly. Plus it is reasonable once I identified errors on IT tests already that does not happens on Unit tests.

houssain-barouni:
> > > 💭 Hey @houssain-baroun, thanks for add the Integration tests. I would suggest here add some more test cases, involving pagination, searchBefore/after and Sorting case, in order to validate the Client & Controller e2e with those cases. I had cases over development that worked well on unit tests but discovered errors on those functionalities over the integration tests.
> > 
> > 
> > I think these case should be rather covered by unit tests, here the focus is more checking the E2E flow is working Deploy -> Export -> Import -> Search with client. The integration tests takes more time to run, so there is a big risk we quickly reach the CI jobs configured timeouts, the recommended CI execution time is 15 minutes max
> 
> I understand the concern regarding the CI job timeouts and the preference for keeping the integration tests focused on the E2E flow. However, I believe it is crucial to include test cases involving pagination, searchBefore/after, and sorting within the integration tests for the following reasons:
> 
> * Transformers and Integration: While unit tests are excellent for ensuring that individual components, such as the request application and mocks, work as expected, they do not fully validate the behavior of Transformers. If there are issues in how Transformers handle pagination, sorting, or searchBefore/after operations, these problems may go unnoticed with only unit tests.
> * Complex Interactions: Operations like searchAfter, page limits, and sorting involve complex interactions between multiple components. For instance, a searchAfter parameter might need to be correctly applied across paginated results, and sorting must be consistent with ES's expectations.
> 
> This would imply in 2 more tests cases, or could be done in the same, and would validate the complex transformers for the ES/OS Queries, and would not impact on the IT timings significantly. Plus it is reasonable once I identified errors on IT tests already that does not happens on Unit tests.

It makes sense if we consider it integration test with real ES/OS, I will add couple of tests with full filtering and full sorting.
BTW, in another PR, I created an example of unit tests that checks that the transformers correctly create SearchQueryRequest of the search-client https://github.com/camunda/camunda/blob/43c1b33ec87969dede10203f0f056dfc36962fe5/service/src/test/java/io/camunda/service/DecisionDefinitionServicesTest.java

ralfpuchert (approved):

nathansandi (changes requested):
Gonna move it to Request Changes
as agreed with Houssain to add more test scenarios over here.

houssain-barouni:
Hi @nathansandi , I have added 2 integration tests
- the first one is testing the search with all filters
- second test is testing reverse sorting with version field

nathansandi (approved):
LGTM 💯

PR #16586, opened 2024-02-27
[Backport stable/8.4] Improve `Either` type

tmetzke:
# Description
Backport of #16566, #15719, and #16305 (partially).

relates to #16561 and #16207 
original authors: @tmetzke, @oleschoenburg, and @ce-dmelnych

PR #16566, opened 2024-02-26
Introduce `fold` method for `Either`

tmetzke:
## Description

Introduces the `fold` method for Either types. This allows to map lefts and rights to a common result type, providing mapping functions for each type respectively.

Example:

```java
Either<String, Integer> success = Either.right(42); // => Right(42)
Either<String, Integer> failure = Either.left("Error occurred"); // => Left("Error occurred")

var rightFn = result -> "Success: " + result;
var leftFn = error -> "Failure: " + error;

success.fold(rightFn, leftFn); // => "Success: 42"
failure.fold(rightFn, leftFn); // => "Failure: Error occurred"
```

## Related issues

closes #16561 

## Definition of Done

_Not all items need to be done depending on the issue and the pull request._

Code changes:
* [x] The changes are backwards compatibility with previous versions
* [ ] If it fixes a bug then PRs are created to [backport](https://github.com/camunda/zeebe/compare/stable/0.24...main?expand=1&template=backport_template.md&title=[Backport%200.24]) the fix to the last two minor versions. You can trigger a backport by assigning labels (e.g. `backport stable/1.3`) to the PR, in case that fails you need to create backports manually.

Testing:
* [x] There are unit/integration tests that verify all acceptance criterias of the issue
* [x] New tests are written to ensure backwards compatibility with further versions
* [ ] The behavior is tested manually
* [ ] The change has been verified by a QA run
* [ ] The impact of the changes is verified by a benchmark

Documentation:
* [ ] The documentation is updated (e.g. BPMN reference, configuration, examples, get-started guides, etc.)
* [ ] If the PR changes how BPMN processes are validated (e.g. support new BPMN element) then the Camunda modeling team should be informed to adjust the BPMN linting.

Other teams:
If the change impacts another team an issue has been created for this team, explaining what they need to do to support this change.
- [ ] [Operate](https://github.com/camunda/operate/issues)
- [ ] [Tasklist](https://github.com/camunda/tasklist/issues)
- [ ] [Web Modeler](https://github.com/camunda/web-modeler/issues)
- [ ] [Desktop Modeler](https://github.com/camunda/camunda-modeler/issues)
- [ ] [Optimize](https://github.com/camunda/camunda-optimize/issues)

Please refer to our [review guidelines](https://github.com/camunda/zeebe/wiki/Pull-Requests-and-Code-Reviews#code-review-guidelines).

PR #20889, opened 2024-08-07
refactor: Change order of `fold` method arguments in `Either` interface

### LINKED ISSUE #20588: Refactor `fold` method argument order in `Either` interface
#### Summary:
The [recently introduced `fold` method](https://github.com/camunda/camunda/pull/16566) in the `Either<L, R>` interface currently has parameters in an order that can cause confusion. The `fold` method is designed to transform an `Either` instance into a common type, regardless of whether it is `Left` or `Right`. However, the current implementation accepts the `right` (success) function as the left parameter and the `left` (failure) function as the right parameter. This ordering is inconsistent with common expectations and [Scala's convention](https://www.scala-lang.org/api/2.13.6/scala/util/Either.html#fold[C](fa:A=%3EC,fb:B=%3EC):C), where the left parameter is for `Left` and the right parameter is for `Right`. This task aims to refactor the `fold` method to match the conventional order and update associated documentation and tests accordingly.

#### Background:
In functional programming, it is a common convention that `Left` represents failure and `Right` represents success. The `fold` method should align with this convention to avoid confusion. The current implementation's parameter order contradicts this norm, leading to potential misuse and misunderstandings. The motivation for the initial implementation was to keep consistency with the existing `void ifRightOrLeft(Consumer<R> rightAction, Consumer<L> leftAction)` method, but user feedback suggests that adhering to the more intuitive convention would be beneficial.

#### Proposal:
Refactor the `fold` method in the `Either` interface to have the left parameter for transforming `Left` and the right parameter for transforming `Right`. Update all associated documentation, JavaDoc examples, and unit tests to reflect this change. Additionally, ensure this change is backported to other branches where the `fold` method was introduced.

#### Implementation steps:
1. **Refactor `fold` method**:
   Update the method signature in `Either` interface as follows:
   ```java
   <T> T fold(Function<? super L, ? extends T> leftFn, Function<? super R, ? extends T> rightFn);
   ```

2. **Update implementation & documentation**:
   - Modify the implementation in both `Left` and `Right` classes to match the new parameter order.
   - Revise & update the JavaDoc for the `fold` method to reflect the new parameter order.

3. **Update Unit tests**:
   - Adjust existing unit tests in `camunda/util/src/test/java/io/camunda/zeebe/util/EitherTest.java` to match the new parameter order.
   - Add new tests if necessary to ensure comprehensive coverage.

4. **Backport changes**:
   - Identify branches where the `fold` method was introduced.
   - Apply the same refactoring and documentation updates to these branches.

ce-dmelnych:
## Description
This PR refactors the `fold` method in the `Either` interface to align with the convention used in Scala, making it more intuitive. The `left` parameter now transforms Left, and the `right` parameter transforms Right. 
The implementation, JavaDoc, `fold` method usage and unit tests have been updated accordingly. 
This change addresses feedback on potential confusion caused by the previous parameter order. 

## Checklist

- [ ] for CI changes:

## Related issues

closes #20588

### DIFF
--- a/service/src/main/java/io/camunda/service/search/core/SearchClientBasedQueryExecutor.java
+++ b/service/src/main/java/io/camunda/service/search/core/SearchClientBasedQueryExecutor.java
@@ -45,10 +45,10 @@ public final class SearchClientBasedQueryExecutor {
     return searchClient
         .search(searchRequest, documentClass)
         .fold(
-            responseTransformer::apply,
             (e) -> {
               throw rethrowRuntimeException(e);
-            });
+            },
+            responseTransformer::apply);
   }
 
   private SearchQuery getAuthenticationCheckIfPresent() {
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionQueryController.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionQueryController.java
@@ -36,7 +36,7 @@ public class DecisionDefinitionQueryController {
   public ResponseEntity<DecisionDefinitionSearchQueryResponse> searchDecisionDefinitions(
       @RequestBody(required = false) final DecisionDefinitionSearchQueryRequest query) {
     return SearchQueryRequestMapper.toDecisionDefinitionQuery(query)
-        .fold(this::search, RestErrorMapper::mapProblemToResponse);
+        .fold(RestErrorMapper::mapProblemToResponse, this::search);
   }
 
   private ResponseEntity<DecisionDefinitionSearchQueryResponse> search(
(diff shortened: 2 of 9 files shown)


ce-dmelnych:
I'll backport these changes after the successful review of this PR.

korthout (approved):
Thanks @ce-dmelnych 💯 

LGTM 👍

korthout:
@ce-dmelnych Please consider backporting to all required branches. It's not clear to me whether or not [the original `fold` method](https://github.com/camunda/camunda/pull/16566) was backported. IMO, it would be good to have it on all stable branches, so it can be used in bugfixes.

tmetzke (approved):
🚀 Thanks, @ce-dmelnych 👍

tmetzke:
> @ce-dmelnych Please consider backporting to all required branches. It's not clear to me whether or not [the original `fold` method](https://github.com/camunda/camunda/pull/16566) was backported. IMO, it would be good to have it on all stable branches, so it can be used in bugfixes.

Was backported down to 8.2 with this PR: https://github.com/camunda/camunda/pull/16586

PR #20899, opened 2024-08-08
fix: include flag `include_global_state` when creating the snapshots

nathansandi:
## Description

`includeGlobalState(true)`: Added this method call in CreateSnapshotRequest to include global state in snapshots, ensuring that templates and ILM policies are backed up.

## Related issues

closes #https://github.com/camunda/tasklist/issues/4704


panagiotisgts (approved):
LGTM

PR #20922, opened 2024-08-08
refactor: inject engine clock

lenaschoenburg:
Before this PR, the engine was directly using `ActorClock` in various places to get the current time. The implicit assumption was that the engine would always run in an actor thread where such an `ActorClock` is available.

This PR refactors (mostly the engine) such that a `ControllableStreamClock` is injected as clock. It is injected by the `StreamProcessor` which takes the current `ActorClock` as source. Because we still use the same clock, no behavior changes are expected.

This is useful for two reasons:
1. Being able to run the engine outside of an actor thread.
2.  Injecting the clock gives us more control over the clock (e.g. for testing)

Having the engine not rely on the actor scheduler was a long-standing goal. An architecture test that was supposed to verify this goal is fixed now, proving that engine production code no longer depends on the actor scheduler.

relates to #21051

### DIFF
--- a/zeebe/engine/pom.xml
+++ b/zeebe/engine/pom.xml
@@ -25,10 +25,6 @@
       <groupId>io.camunda</groupId>
       <artifactId>zeebe-stream-platform</artifactId>
     </dependency>
-    <dependency>
-      <groupId>io.camunda</groupId>
-      <artifactId>zeebe-scheduler</artifactId>
-    </dependency>
 
     <dependency>
       <groupId>org.agrona</groupId>
@@ -116,6 +112,12 @@
     </dependency>
 
     <!-- TEST DEPENDENCIES -->
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-scheduler</artifactId>
+      <scope>test</scope>
+    </dependency>
+
     <dependency>
       <groupId>io.camunda</groupId>
       <artifactId>zeebe-logstreams</artifactId>
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
@@ -71,6 +71,18 @@ public interface StreamClock extends InstantSource {
     }
 
     /**
+     * If the clock is already offset, stacks the additional offset on top of the current offset.
+     * Otherwise, offsets the clock by the given duration, overwriting any previous modification.
+     */
+    default void stackOffset(final Duration additionalOffset) {
+      if (currentModification() instanceof Modification.Offset(final var initialOffset)) {
+        applyModification(Modification.offsetBy(initialOffset.plus(additionalOffset)));
+      } else {
+        offsetBy(additionalOffset);
+      }
+    }
+
+    /**
      * Shortcut to reset the clock to the current system time.
      *
      * @implSpec Equivalent to {@code applyModification(Modification.none())}.
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.stream.api;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.time.InstantSource;
+
+/**
+ * A clock that provides the current time. Used as a marker interface instead of using {@link
+ * InstantSource} directly to indicate that the clock is used for stream processing.
+ */
+public interface StreamClock extends InstantSource {
+
+  /**
+   * A controllable {@link StreamClock} that allows to pin the time to a specific instant or to
+   * offset the current system time by some duration.
+   */
+  interface ControllableStreamClock extends StreamClock {
+
+    /**
+     * Modifies the clock by applying the given modification. Previous modifications are overridden
+     * and do not stack up.
+     *
+     * @param modification the modification to apply. Use {@link Modification#none()} to reset the
+     *     clock to current system time.
+     */
+    void applyModification(Modification modification);
+
+    /**
+     * Returns the current modification applied to the clock. If no modification is applied, {@link
+     * Modification.None} is returned.
+     */
+    Modification currentModification();
+
+    /**
+     * Shortcut to reset the clock to the current system time. Equivalent to {@code
+     * applyModification(Modification.none())}.
+     */
+    default void reset() {
+      applyModification(Modification.none());
+    }
+
+    /**
+     * Shortcut to check if the clock is modified. Equivalent to {@code !(currentModification()
+     * instanceof Modification.None)}
+     */
+    default boolean isModified() {
+      return !(currentModification() instanceof Modification.None);
+    }
+
+    sealed interface Modification {
+      static None none() {
+        return new None();
+      }
+
+      static Pin pinAt(final Instant at) {
+        return new Pin(at);
+      }
+
+      static Offset offsetBy(final Duration by) {
+        return new Offset(by);
+      }
+
+      record None() implements Modification {}
+
+      record Pin(Instant at) implements Modification {}
+
+      record Offset(Duration by) implements Modification {
+        public Offset {
+          if (by.isZero()) {
+            throw new IllegalArgumentException("Offset must not be zero");
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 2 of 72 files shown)


lenaschoenburg:
For https://github.com/camunda/product-hub/issues/2379, we'd want to inject our own interface instead of `InstantSource` such that the engine can also _modify_ the current time.

zeebe/engine/pom.xml
```diff
@@ -116,6 +112,12 @@
     </dependency>
 
     <!-- TEST DEPENDENCIES -->
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-scheduler</artifactId>
+      <scope>test</scope>
+    </dependency>
+
```
npepinpe, line 120: :question: I thought we had removed all dependencies to the scheduler for the engine? :thinking:

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
```diff
@@ -0,0 +1,97 @@
...
+  interface ControllableStreamClock extends StreamClock {
+
+    /**
+     * Modifies the clock by applying the given modification. Previous modifications are overridden
+     * and do not stack up.
+     *
+     * @param modification the modification to apply. Use {@link Modification#none()} to reset the
+     *     clock to current system time.
+     */
+    void applyModification(Modification modification);
```
npepinpe, line 47: :question: This implies that if I say, `offset(5)`, then `offset(5)`, nothing will change? In the previous controlled clock this would actually keep incrementing by 5, as it was always a relative offset. Is this the behavior we want?

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Shortcut to check if the clock is modified. Equivalent to {@code !(currentModification()
+     * instanceof Modification.None)}
+     */
+    default boolean isModified() {
+      return !(currentModification() instanceof Modification.None);
+    }
+
+    sealed interface Modification {
```
npepinpe, line 71: :wrench: Cool use of sealed types, but I'm wondering if it's required in this instance. It's unlikely to be extended, so right now the UX is:

```java
clock.applyModification(Modification.pinAt(Instant.now()));
```

versus

```java
clock.pinAt(Instant.now());
```

While it makes it easier to extend, and easier if we keep the behavior of overriding previous modifications, the actual API is arguably worse here then just having explicit methods to pin/reset/add offset. I see you already added `reset`, which implies to me that you might have realized (i.e. `applyModification(Modification.none())` is not intuitive to reset).

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ControllableStreamClockImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+import io.camunda.zeebe.stream.api.StreamClock;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock.Modification.None;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock.Modification.Offset;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock.Modification.Pin;
+import java.time.Instant;
+import java.time.InstantSource;
+import java.util.Objects;
+
+public final class ControllableStreamClockImpl implements StreamClock.ControllableStreamClock {
```
npepinpe, line 18: :+1: Definitely easier to read in general than the current `ControlledActorClock`

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/UncontrolledStreamClock.java
```diff
@@ -0,0 +1,59 @@
...
+package io.camunda.zeebe.stream.impl;
+
+import io.camunda.zeebe.stream.api.StreamClock;
+import java.time.Clock;
+import java.time.Instant;
+import java.time.InstantSource;
+import java.time.ZoneId;
+import java.util.Objects;
+
+public final class UncontrolledStreamClock implements StreamClock {
```
npepinpe, line 17: :thought_balloon: Likely if we did not implement `InstantSource` but our own type like `EpochClock`, then we wouldn't need an implementation here but could return a constant lambda, e.g. `() -> InstantSource.system().millis()`.

npepinpe (approved):
:+1: Some suggestions, no blockers, no need for another review (but of course if you'd like). Please address comments before merging :)

:wrench: I would propose also not using `InstantSource`, but something like Agrona's `EpochClock` - we can also just write our own in `util` if we don't want to couple ourselves (though we already are :smile:) to them. My reasoning here is that we typically only work with epoch timestamps, and `InstantSource` also lets you get things like an `Instant`. Typically for the high performance stuff we'd want to avoid that, so it might just be simpler to not have it at all. It's a very minor future optimization, so I'm also fine without it.

npepinpe:
One last thing, can we verify if we need the offset thing? If it's really just pin, let's start with that.

zeebe/engine/pom.xml
lenaschoenburg: This is a test scoped dependency because various tests make use of the actor scheduler to run the engine, see for example `StreamProcessorRule`.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
lenaschoenburg: Yes, that's also part of the docstring:
```
/**
 * Modifies the clock by applying the given modification. Previous modifications are overridden
 * and do not stack up.
 *
```

> Is this the behavior we want

I don't really know yet. If not, the desired behavior can be emulated easily with first querying current offset, adding to it and then applying a new one.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
lenaschoenburg: One benefit from it is that it's a symmetric API with `currentModification` and `applyModification`. Doesn't stop us from adding convenience methods such as `pinAt` that work the same way as `reset`.

lenaschoenburg: I've added two new helpers, `pinAt` and `offsetBy`.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/api/StreamClock.java
lenaschoenburg: I've added a convenience method to do that, `stackOffset`.

lenaschoenburg:
> One last thing, can we verify if we need the offset thing? If it's really just pin, let's start with that.

I don't think we can get a better answer than the one we already have which is basically "probably pin is enough". I don't think offset is adding much complexity here so I'd just keep it for now.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/UncontrolledStreamClock.java
lenaschoenburg: Actually this implementation is already not used, maybe we can just remove it along with the interface `StreamClock`? I don't think we have a usecase where the clock should _not_ be controllable.

zeebe/engine/pom.xml
npepinpe: Oh :( I would have hoped even in the tests we didn't :sweat_smile:

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/UncontrolledStreamClock.java
npepinpe: Th only risk is in the future if we want an immutable one :shrug: I could imagine that we will have a feature flag for this, so the question is do we want to make it completely impossible, or just prevent commands from modifying the clock? I would lean towards the safer/cautious approach, but in the end with good test coverage it's likely fine either way.

lenaschoenburg: Okay then let's keep it for now, maybe we can use it to disable the entire feature :+1:

npepinpe:
Maybe I missed it, but I didn't see you address the review comment:

> 🔧 I would propose also not using InstantSource, but something like Agrona's EpochClock - we can also just write our own in util if we don't want to couple ourselves (though we already are 😄) to them. My reasoning here is that we typically only work with epoch timestamps, and InstantSource also lets you get things like an Instant. Typically for the high performance stuff we'd want to avoid that, so it might just be simpler to not have it at all. It's a very minor future optimization, so I'm also fine without it.

lenaschoenburg:
Ah yes, I didn't address it :+1: 
I'd rather continue using `InstantSource` because it's a standardized, easy to use interface that offers more functionality than something like agrona's `EpochClock`.
While it's true that we sometimes use millis since epoch instead of `Instant` or `Duration`, having the ability to get an `Instant` from the clock is still nice.

I also feel like using `InstantSource` does not add any complexity apart from having to implement two methods _if_ the `millis` code path needs to be optimized.

PR #20937, opened 2024-08-09
refactor: support newer helm chart version

ChrisKujawa:
## Description
Adjust values for helm installation to support 9x, and 10x camunda platform helm chart.

## Related issues

See https://docs.camunda.io/docs/next/self-managed/setup/upgrade/#deprecation-notes 
See https://github.com/zeebe-io/benchmark-helm/pull/171


npepinpe (approved):
:rocket:

PR #20940, opened 2024-08-09
Revert perform license check on startup

ChrisKujawa:
## Problem Description

We run into issues with the change from https://github.com/camunda/camunda/pull/20541 as we were no longer build locally, or on some CI pipelines. Thus because the license-checker dependency is not publically available. It is only in Camunda repositories available, which might also cause issues for external contributors. 

See related [slack thread conversation ](https://camunda.slack.com/archives/C06HTSPD5AP/p1723190152698859?thread_ts=1722457528.375399&cid=C06HTSPD5AP)

We will need to revert the changes, until we find another solution like making the artifacts publicly available. \cc @cmur2 might be something you can help with. Furthermore, we might want to have a check in our CI that the builds are working also with public repos only.

### What the PR does to overcome this

Reverts changes from PR https://github.com/camunda/camunda/pull/20541/commits

I tried to revert the merge commit, but [read that it might cause issues](https://stackoverflow.com/a/7100005/2165134) when we want to apply the same commits again.

> Source: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-revert.html
> Reverting a merge commit declares that you will never want the tree changes brought in by the merge. As a result, later merges will only bring in tree changes introduced by commits that are not ancestors of the previously reverted merge. This may or may not be what you want.

So I revert each commit separately (excluding merge commits).

```sh
git revert 966db932161ca6cd17b1e726008eff24cb665b39 c3855ef0facc526e66ea3bbb5788bdd13c0ca4b3 5e00a6200dfa5924e5273d10f71a07cc0591e636 62c7d21419153df062dd4095091946cd8cc897ce 51ffe8081f8f15ab7fb80d882c62ecf0979a3051 a58d6f40c04d5b967801fbb9543156a4a2cc2525 11315f771228c3e4600e0a71822fffb58baedde0 d51bde18db31220990bcd07fca8f39af36dbe949 e037f9896ff3a1f90caf57f1158163b636d6f7b1 48c1702f7e10daf2b1d5180bf402a74413ccce24 d26265e35e9f29fc3a0767ce428c15f1347ebdd7 ed356d95543c14883212d6c0f32744996cb2dc86 f02fc3fb8346fcbf25f4222f4a175afe66c860b0
```


nicpuppa (approved):
👍

PR #20950, opened 2024-08-09
Using a changeset to update a job

### LINKED ISSUE #20897: Introduce `changedAttributes` field in JobUpdateRecord to keep truck of changed fields
**Description**

Since the introduction of the new `JobUpdate` REST endpoint, users can provide different value for updating a job record, like `retries` and `timeout`. At least one of the attribute is mandatory, otherwise the request will be rejected. Providing any of those attributes with a null value or omitting it preserves the persisted attribute’s value.

To ensure this, at allow users to provide every value for the `timeout` field, we need a way to track which fields are provided by the user.

nicpuppa:
## Description

To keep track of the updated fields, a new changeset attributes is added to `JobRecord`. We thought about 2 options to implement this changeset, below I'll give a short description of the options and the final decision

### Option # 1 - Changeset as a `Map`

The first approach was to use a `Map<String, Number>` to provide the changed attributes from the gateway to the broker. Using this approach allows to have in the log a record with the older values and the new values stored into a Map. 

### Option # 2 - Changeset as a `Set` of field names

The second idea was to use a `Set` of changed field names to provide the updated fields from the gateway to the broker. This approach implies that the updated values are set into the record itself, using a `Set` field to just keep track of the changed field names

### FInal decision

We decided for `Option 2` for one main reason: in future we will support more fields to be updated, and using just a Set of names is less expensive in terms of size of the record. Using a `Map` would duplicate the record size (worst scenario, if every field is updatable)

## Related issues

closes #20897


nicpuppa:
Hey @tmetzke this is an idea on how to approach differently the problem with `JobUpdate`. Let me know what you think 💪

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobUpdateProcessor.java
```diff
@@ -42,26 +45,48 @@ public void processRecord(final TypedRecord<JobRecord> command) {
...
   }
+
+  private <T extends Number> void jobChange(
+      final Map<String, Number> changeset,
+      final String key,
+      final Function<Number, T> converter,
+      final Function<T, Optional<String>> updateFunction,
+      final List<String> errors) {
+    if (changeset.containsKey(key)) {
+      final T value = converter.apply(changeset.get(key));
```
tmetzke, line 81: 🔧 Should we handle any conversion issues that might occur? We should also test what happens when the conversion does not work properly.

service/src/main/java/io/camunda/service/JobServices.java
```diff
@@ -97,8 +97,8 @@ public CompletableFuture<JobRecord> completeJob(
   }
 
   public CompletableFuture<JobRecord> updateJob(
-      final long jobKey, final Integer retries, final Long timeout) {
-    return sendBrokerRequest(new BrokerUpdateJobRequest(jobKey, retries, timeout));
+      final long jobKey, final Map<String, Number> changeset) {
```
tmetzke, line 100: 💭 Changes might also be any arbitrary value in the future, not just number attributes. We don't have anything else so far and the service API is not considered public API, I think. So it should also be fine like this. Would we have any big downsides to using `Map<String, Object> changeset` here instead? Just wondering.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/job/JobUpdateTest.java
```diff
@@ -42,14 +43,16 @@ public void setup() {
   public void shouldUpdateJob() {
```
tmetzke, line 43: 🔧 We could add test cases for
* Conversion of the changed attribute value is failing (we accept any `Number` types currently, I think, we can probably make that fail in a test if we want to to see if our processor can handle that 🙂)
* Providing no changed attribute - what happens in our processor in this case? I know we validate this in the REST API but the processor is decoupled from that, it doesn't know that. So it needs to be prepared.
*

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -228,6 +227,15 @@ private static <R> int getIntOrZero(final R request, final Function<R, Integer>
     return value == null ? 0 : value;
   }
 
+  private static Map<String, Number> buildJobChangeset(final JobChangeset jobChangeset) {
+    final Map<String, Number> changeset = new HashMap<>();
+    Optional.ofNullable(jobChangeset.getRetries())
+        .ifPresent(retries -> changeset.put(JobRecord.RETRIES, retries));
```
tmetzke, line 233: 🔧 I'm wondering if we could use something like this Optional handling in the processor as well instead of the current routine 🤔

tmetzke (commented):
👍 Really like this direction.
🔧 Added some suggestions but nothing major. Really good job 🚀 
❓ We are currently not serializing and thus not exporting the changed attributes. Is that on purpose or did you miss it?

nicpuppa:
> ❓ We are currently not serializing and thus not exporting the changed attributes. Is that on purpose or did you miss it?

Yes, it was on purpose. This is something that maybe we need to clarify first, also for `UserTaskUpdate`

tmetzke:
> > ❓ We are currently not serializing and thus not exporting the changed attributes. Is that on purpose or did you miss it?
> 
> Yes, it was on purpose. This is something that maybe we need to clarify first, also for `UserTaskUpdate`

Well, for user task updates we already export this.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/job/JobUpdateTest.java
nicpuppa: I didn't add more tests cases yet, but thanks for the suggestions

nicpuppa:
> Well, for user task updates we already export this.

My bad, I thought was something still to clarify

tmetzke:
> > Well, for user task updates we already export this.
> 
> My bad, I thought was something still to clarify

Still correct for jobs, though. We need to decide if we want to export it. We can also move along with "no we don't" for now. This is easy to enable if we want to.

nicpuppa:
The CI is failing for an error in tasklist:

```
Caused by: org.apache.maven.plugin.compiler.CompilationFailureException: Compilation failure
/runner/_work/camunda/camunda/tasklist/importer-850/src/main/java/io/camunda/tasklist/zeebeimport/v850/record/value/JobRecordValueImpl.java:[17,8] io.camunda.tasklist.zeebeimport.v850.record.value.JobRecordValueImpl is not abstract and does not override abstract method getChangedAttributes() in io.camunda.zeebe.protocol.record.value.JobRecordValue
```

should I adjust the tasklist code ? @tmetzke

tmetzke:
> The CI is failing for an error in tasklist:
> 
> ```
> Caused by: org.apache.maven.plugin.compiler.CompilationFailureException: Compilation failure
> /runner/_work/camunda/camunda/tasklist/importer-850/src/main/java/io/camunda/tasklist/zeebeimport/v850/record/value/JobRecordValueImpl.java:[17,8] io.camunda.tasklist.zeebeimport.v850.record.value.JobRecordValueImpl is not abstract and does not override abstract method getChangedAttributes() in io.camunda.zeebe.protocol.record.value.JobRecordValue
> ```
> 
> should I adjust the tasklist code ? @tmetzke

If you feel up for it, go for it. We can also ask someone from HTO to help out and/or review.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/behaviour/JobUpdateBehaviour.java
```diff
@@ -54,28 +50,27 @@ public Either<String, JobRecord> getJob(final long jobKey, final TypedRecord<Job
   }
 
   public Optional<String> updateJobRetries(
-      final long jobKey, final JobRecord jobRecord, final TypedRecord<JobRecord> command) {
-    final int retries = command.getValue().getRetries();
+      final long jobKey,
+      final int retries,
+      final JobRecord jobRecord,
+      final TypedRecord<JobRecord> command) {
```
tmetzke, line 56: 🔧 I think we don't need the `command` anymore, do we? The retries are already provided.

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/job/JobRecord.java
```diff
@@ -71,6 +78,8 @@ public final class JobRecord extends UnifiedRecordValue implements JobRecordValu
   private final LongProperty elementInstanceKeyProp = new LongProperty("elementInstanceKey", -1L);
   private final StringProperty tenantIdProp =
       new StringProperty("tenantId", TenantOwned.DEFAULT_TENANT_IDENTIFIER);
+  private final ArrayProperty<StringValue> changedAttributesProp =
+      new ArrayProperty<>("changedAttributes", StringValue::new);
 
   public JobRecord() {
     super(20);
```
tmetzke, line 85: 🔧 If we add an attribute, we should increase the number of declared properties as well.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUpdateJobRequest.java
```diff
@@ -12,17 +12,23 @@
...
     request.setKey(jobKey);
-    requestDto.setRetries(retries);
-    requestDto.setTimeout(timeout);
+    requestDto.setChangedAttributes(changeset.keySet());
+    if (changeset.get(JobRecord.RETRIES) != null) {
+      requestDto.setRetries(changeset.get(JobRecord.RETRIES).intValue());
+    }
+    if (changeset.get(JobRecord.TIMEOUT) != null) {
+      requestDto.setTimeout(changeset.get(JobRecord.TIMEOUT).longValue());
+    }
```
tmetzke, line 31: 💭 I now remember why I also chose to use the `UserTaskRecord` as the "changeset" for the broker request; it basically is a type-safe map instead of an "anonymous" map like we have here. 

We cannot rely on the data types we receive here in `changeset`. This broker request constructor here might have to check that the data is an Integer or Long before simply using it as such. If you have the JobRecord here, it can only contain the correct types. You can then rely on it here.

tmetzke (commented):
👍 Looks really good already, the basic functionality is correct now 🚀 
🔧 There are some minor improvements that we can consider.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUpdateJobRequest.java
tmetzke: This refers to the test case if described earlier with
> Conversion of the changed attribute value is failing (we accept any Number types currently, I think, we can probably make that fail in a test if we want to to see if our processor can handle that 🙂)

nicpuppa: I'm not a big fan of using those records outside the engine honestly. I don't know what you think about this, but I'd prefer leave it like this (we are confident about data type because we get them from the request)

tmetzke: Yeah, I generally understand the point with the record but we already use the record in those broker request classes heavily already anyway 😄 
Regarding confidence, yes, currently this broker request class is only used by our REST gateway. But you know how it is, it doesn't always stay like that. The broker request doesn't know anything about its callers and it is now making assumptions on what they will provide. I simply think that is a bit dangerous and I don't want to look into an incident later trying to find out why we are running into weird JDK exceptions 🤣

nicpuppa: > Yeah, I generally understand the point with the record but we already use the record in those broker request classes heavily already anyway 😄

Yes, this should be the only place where we use them outside the engine.

But I wanto to find a solution that works for both of us, so what you suggest to do ? 😄

tmetzke: If we use the JobRecord as the changeset, we'll also use it more heavily in the REST controller to "fill it with data", right? I see your concern, might be something we also want to change for the user task record then as a follow-up so it's consistent. I totally understand this concern and I agree that keeping the records as close to the broker classes as possible only and don't leak them too much elsewhere 👍 

Looking at the issue, I see two ways currently:
- Introduce a proper DTO with specific types for the changeset attributes instead of a semi-anonymous Java `Map` with target type `Number`.
- Do validation in the broker request class for the changeset attributes.

Out of those two, I would slightly prefer the first one, I think. It relieves the broker request class from checking its inputs and puts the responsibility on the caller with an explicit contract (retries and timeout need to be proper Long or Integer values).

nicpuppa: I'll go for the first one 👍  thanks

tmetzke (approved):
🚀 🚀 Really like this, it's type-safe, it efficiently handles only the changed attributes, I think the user task update can be improved with this as a follow-up as well!

PR #20960, opened 2024-08-09
build: fix compilation issues in optimize

buccarel:
## Description

This PR fixes compilation issues detected in Optimize after the merge. In particular:
* the experimental Lombok annotation `@FieldNameConstants` was not being compiled correctly, so it is being removed in favor of its corresponding vanilla Java
* the `@Builder` annotation in one class was not being compiled, so it is being removed in favor of its corresponding vanilla Java
* the docstrings in `IgnoreDuringScan.java` had syntax issues that are being corrected

The command that I used to do the compilation is the following:
```bash
$ mvn -B -D skipTests -D skipChecks compile generate-sources source:jar javadoc:jar -P skipFrontendBuild
```

## Checklist

- [ ] for CI changes:

## Related issues

* https://github.com/camunda/camunda-optimize/issues/13962

### DIFF
--- a/optimize/backend/src/main/java/io/camunda/optimize/IgnoreDuringScan.java
+++ b/optimize/backend/src/main/java/io/camunda/optimize/IgnoreDuringScan.java
@@ -13,7 +13,8 @@ import java.lang.annotation.RetentionPolicy;
 /**
  * This annotation is used to ignore annotated classes during component-scan.
  *
- * Sample usage:
+ * <p>Sample usage:
+ *
  * <pre>{@code
  * @ComponentScan(excludeFilters = @ComponentScan.Filter(IgnoreDuringScan.class))
  * @Configuration
--- a/optimize/backend/src/main/java/io/camunda/optimize/IgnoreDuringScan.java
+++ b/optimize/backend/src/main/java/io/camunda/optimize/IgnoreDuringScan.java
@@ -18,6 +18,7 @@ import java.lang.annotation.RetentionPolicy;
  * @ComponentScan(excludeFilters = @ComponentScan.Filter(IgnoreDuringScan.class))
  * @Configuration
  * public class Main {
+ * }
  * }</pre>
  */
 @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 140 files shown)


optimize/backend/src/main/java/io/camunda/optimize/IgnoreDuringScan.java
```diff
@@ -18,6 +18,7 @@
  * @ComponentScan(excludeFilters = @ComponentScan.Filter(IgnoreDuringScan.class))
  * @Configuration
  * public class Main {
+ * }
```
buccarel, line 21: This was giving compilation error: even though it's in a comment block, it didn't close a docstrings tag well

optimize/backend/src/main/java/io/camunda/optimize/dto/optimize/importing/DecisionInstanceDto.java
```diff
@@ -43,4 +41,25 @@ public class DecisionInstanceDto implements OptimizeDto {
   private Set<String> matchedRules = new HashSet<>();
   private String engine;
   private String tenantId;
+
+  public static final class Fields {
```
PHWaechtler, line 45: ❓ How come in some classes we're using a class like this one while in others (eg IdentityLinkLogEntryDto ) we're using an enum?

HeleneW-dot (approved):
Lgtm, but it would be nice to have a consistent implementation we can share across all classes, ie either always go for the enum or always go for the static class. Can we adjust this or is there a reason we need one or the other in separate classes?

matthewBearCamunda (commented):
In general this LGTM. I didn't look through all of the classes, but they all follow the same pattern

One question though, why put the variables in a `Fields` inner class each time? Why not just have the variables in the regular class?

optimize/backend/src/main/java/io/camunda/optimize/dto/optimize/importing/DecisionInstanceDto.java
buccarel: Also replying to @matthewBearCamunda 

Well this comes from how the annotation was used before. In some cases, it had the `asEnum = true` argument, in other cases it didn't have it. The 2 cases have a different vanilla Java code, so the results are different.

I don't know why it was this way originally. CC'ing @RomanJRW to see if he happens to have more context about this

buccarel:
> In general this LGTM. I didn't look through all of the classes, but they all follow the same pattern
> 
> One question though, why put the variables in a `Fields` inner class each time? Why not just have the variables in the regular class?

@matthewBearCamunda that's how the annotation works. When you annotate a class, the class `Fields` is created. I didn't remove it now after de-lombok'ing because then I would need to change all of the consumers. Also, de-lombok'ing into a different format is not possible with automatic tools like the one I used

PR #20963, opened 2024-08-09
feat: support authorization creation

### LINKED ISSUE #20433: Camunda service supports creating authorization for a user or group or role
### Description
Camunda service allows to create an authorization for a relevant entity. Authorization entity is the holder of permissions to a specific resource.

### Acceptance Criteria
- [x] create an authorization record in zeebe with `resourceType`, `resourecKey `and a list of `permission` as fields.
- [x] create command and processor for an authorization creation
- [x] store the authorization assigned to a user.
- [x] store the authorization assigned  to a group.
- [x] store the authorization assigned  to a role.
- [x] add rest api to assign authorization (create during assignment)
- [ ] add UI for authorization creation
- [ ] add UI for assigning an authorization to a user.
- [ ] add UI for assigning an authorization to a group.
- [ ] add UI for assigning an authorization to a role.
- [x] export an authorization from logstream to elastic search

### LINKED ISSUE #21514: UpdateUserMultiPartitionTest.shouldDistributeUserUpdateCommand is flaky
**Summary**

- How often does the test fail?
- Does it block your work?
- Do we suspect that it is a real failure?

**Failures**

<pre>
java.lang.AssertionError: 

Expecting actual:
  [CREATE, UPDATE, CREATED, UPDATED]
to end with:
  [UPDATE, UPDATED]

	at io.camunda.zeebe.engine.processing.user.UpdateUserMultiPartitionTest.shouldDistributeUserUpdateCommand(UpdateUserMultiPartitionTest.java:95)
</pre>
</details>

https://github.com/camunda/camunda/actions/runs/10514546053/job/29132627947

[Uploading Test results for unit tests (2).zip…]()

Ben-Sheppard:
## Description
This PR introduces the authorization create functionality.

## Related issues

closes #20433 
closes #21514


remcowesterhoud:
Is this blocking you? If so I'd like to focus on some other topics first. I'm expecting to have this reviewed before the end of the week.

Ben-Sheppard:
Ideally I'd like to have this merged soon as it will be required before we work on the new exporters and is also required to unblock other teams :)

Ben-Sheppard:
Would a sync pair review be more efficient in this instance @remcowesterhoud ?

remcowesterhoud:
> Would a sync pair review be more efficient in this instance @remcowesterhoud ?

No, I'd like to do the review by myself. But since it's blocking I'll prioritize it over some other stuff I had planned. I'll look into it this afternoon and will probably have the review ready tomorrow by the latest 🙂

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/CreateAuthorizationTest.java
```diff
@@ -0,0 +1,99 @@
...
+            .withResourceKey("resource")
+            .withResourceType("bpmn-id")
+            .withPermissions(permissions)
+            .expectRejection()
+            .create();
+
+    final var createdAuthorization = createdAuthorizationRecord.getValue();
+    assertThat(createdAuthorization).isNotNull().hasFieldOrPropertyWithValue("ownerKey", owner);
+
+    io.camunda.zeebe.protocol.record.Assertions.assertThat(duplicatedAuthorizationRecord)
```
remcowesterhoud, line 93: 🔧 Let's create an import for this.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -0,0 +1,97 @@
...
+    ownerKey.wrapString(authorizationRecord.getOwnerKey());
+    ownerType.wrapString(authorizationRecord.getOwnerType());
+    resourceKey.wrapString(authorizationRecord.getResourceKey());
+    resourceType.wrapString(authorizationRecord.getResourceType());
+    persistedPermissions.setPermissions(authorizationRecord.getPermissions());
+    ownerAuthorizationColumnFamily.insert(ownerAndResourceCompositeKey, persistedPermissions);
+  }
+
+  @Override
+  public PersistedPermissions getAuthorization(
```
remcowesterhoud, line 82: 🔧 I feel like we should rename this to `getPermissions`. WDYT?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -0,0 +1,97 @@
...
+  private final DbLong authorizationKey;
+  private final DbString ownerKey;
+  private final DbString ownerType;
+  private final DbString resourceKey;
+  private final DbString resourceType;
+  private final DbCompositeKey<DbString, DbString> resourceCompositeKey;
+  private final DbCompositeKey<
+          DbCompositeKey<DbString, DbString>, DbCompositeKey<DbString, DbString>>
+      ownerAndResourceCompositeKey;
+  private final ColumnFamily<DbLong, PersistedAuthorization> authorizationColumnFamily;
```
remcowesterhoud, line 35: ❓ I'm a bit confused about this ColumnFamily. What will it be used for? Right now we only use it to store the key + authorization record but never read from it. Is it going to be used in a later PR?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -680,6 +680,52 @@ paths:
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/AuthorizationAssignRequest"
+        required: true
+      responses:
+        "204":
+          description: |
+            The Authorization was created successfully.
```
remcowesterhoud, line 698: ❓ There is no content in this Response. Don't you need the key Zeebe generated to update/delete the authorization later?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/dto/AuthorizationAssignRequest.java
```diff
@@ -0,0 +1,59 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.rest.controller.usermanagement.dto;
+
+import io.camunda.zeebe.protocol.impl.record.value.authorization.AuthorizationOwnerType;
+import java.util.List;
+
+public class AuthorizationAssignRequest {
```
remcowesterhoud, line 13: ❌ This DTO shouldn't exist. The request is autogenerated according to the OpenAPI spec in `rest-api.yaml`

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/AuthorizationController.java
```diff
@@ -0,0 +1,48 @@
...
+        () ->
+            authorizationServices
+                .withAuthentication(RequestMapper.getAuthentication())
+                .createAuthorization(
+                    authorizationAssignRequest.getOwnerKey(),
+                    authorizationAssignRequest.getOwnerType(),
+                    authorizationAssignRequest.getResourceKey(),
+                    authorizationAssignRequest.getResourceType(),
+                    authorizationAssignRequest.getPermissions()));
+  }
```
remcowesterhoud, line 47: ❌ We should change this slightly. I think I may have missed this in the create user PR as well as I never touched the REST API before I reviewed this.

The request that comes in is a dto. We should use the `RequestMapper` to map it to an internal object. Similar to what we do with other endpoints that have a request body. In this `RequestMapper` we can do some validations to check if the incoming request is valid. I'm not sure if there's specific validations that make sense here. Maybe the permissions need to adhere to a specific format?

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/AuthorizationControllerTest.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Test
+  void createAuthorizationShouldReturnNoContent() {
+    final var request = new AuthorizationAssignRequest();
+    request.setOwnerKey("1");
+    request.setOwnerType(AuthorizationOwnerType.USER);
+    request.setResourceKey("2");
+    request.setResourceType("resourceType");
+    request.setPermissions(List.of("permission1", "permission2"));
```
remcowesterhoud, line 52:
```suggestion
    final var request = new AuthorizationAssignRequest()
    .ownerKey("1")
    .ownerType(AuthorizationOwnerType.USER)
    .resourceKey("2")
    .resourceType("resourceType")
    .permissions(List.of("permission1", "permission2");
```

🔧 Let's chain these methods calls

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/AuthorizationControllerTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+    final var authorizationRecord = new AuthorizationRecord();
+    authorizationRecord.setOwnerKey(request.getOwnerKey());
+    authorizationRecord.setOwnerType(request.getOwnerType());
+    authorizationRecord.setResourceKey(request.getResourceKey());
+    authorizationRecord.setResourceType(request.getResourceType());
+    authorizationRecord.setPermissions(request.getPermissions());
```
remcowesterhoud, line 59:
```suggestion
    final var authorizationRecord = new AuthorizationRecord()
    .setOwnerKey(request.getOwnerKey())
    .setOwnerType(request.getOwnerType())
    .setResourceKey(request.getResourceKey())
    .setResourceType(request.getResourceType())
    .setPermissions(request.getPermissions());
```

🔧 Let's chain these method calls

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/AuthorizationControllerTest.java
```diff
@@ -0,0 +1,124 @@
...
+    request.setResourceType("resourceType");
+    request.setPermissions(List.of("permission1", "permission2"));
+
+    when(authorizationServices.createAuthorization(
+            request.getOwnerKey(),
+            request.getOwnerType(),
+            request.getResourceKey(),
+            request.getResourceType(),
+            request.getPermissions()))
+        .thenThrow(new CamundaServiceException(RejectionType.ALREADY_EXISTS.name()));
```
remcowesterhoud, line 105: 🔧 Same chaining comment 🙂

remcowesterhoud, line 105: 🔧 I think it's more realistic if this exception takes a `BrokerRejection` because if we detect an already exists we would do this on the broker side. E.g.:

```suggestion
        .thenThrow(
            new CamundaServiceException(
                new BrokerRejection(
                    AuthorizationIntent.CREATE,
                    1L,
                    RejectionType.ALREADY_EXISTS,
                    "Authorization already exists")));
```

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/AuthorizationRecord.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+import org.agrona.DirectBuffer;
+
+public final class AuthorizationRecord extends UnifiedRecordValue
+    implements AuthorizationRecordValue {
+  private final LongProperty authorizationKeyProp = new LongProperty("authorizationKey", -1L);
+  private final StringProperty ownerTypeProp = new StringProperty("ownerType");
+  private final StringProperty ownerKeyProp = new StringProperty("ownerKey");
```
remcowesterhoud, line 29: ❌ OwnerType seems to be an enum. We should turn this into an `EnumProperty`

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/AuthorizationRecord.java
```diff
@@ -0,0 +1,169 @@
...
+  private final StringProperty resourceKeyProp = new StringProperty("resourceKey");
+  private final StringProperty resourceTypeProp = new StringProperty("resourceType");
```
remcowesterhoud, line 31: ❓ I wonder if this could also be an enum? But I don't know what this would be filled with realistically.

zeebe/protocol-impl/src/test/java/io/camunda/zeebe/protocol/impl/JsonSerializableToJsonTest.java
```diff
@@ -2484,6 +2486,31 @@ final var record =
         """
       },
       /////////////////////////////////////////////////////////////////////////////////////////////
+      //////////////////////////////////// AuthorizationRecord ////////////////////////////////////
```
remcowesterhoud, line 2489: ❌ Please also add the empty `AuthorizationRecord` case. This record has optional properties so we should verify that if we don't set those, the mapping works and contains the default values.

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/intent/AuthorizationIntent.java
```diff
@@ -0,0 +1,75 @@
...
+
+public enum AuthorizationIntent implements Intent {
+  CREATE(0),
+  CREATED(1),
+  UPDATE(2),
+  UPDATING(3),
+  UPDATED(4),
+  DELETE(5),
+  DELETING(6),
+  DELETED(7);
```
remcowesterhoud, line 26: ❌ Let's remove these for now. We aren't using them. I think we don't need `UPDATING` and `DELETING` anyways but let's think about that when we implements those features.

remcowesterhoud (changes requested):
Thanks @Ben-Sheppard! I've finished my first round of review. Please have a look at my comments 👀 

❓ I am a bit confused about this feature to be honest. Maybe I don't understand how authorizations are supposed to work so please enlighten me. I would expect that we store Users in our state and that a User can be Authorized to perform certain actions. Currently I see no relationship in our state between Users and Authorizations. That seems very strange to me, because how will you check later on that User X is Authorized to do action Y?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCreateProcessor.java
```diff
@@ -0,0 +1,97 @@
...
+    rejectionWriter = writers.rejection();
+    responseWriter = writers.response();
+    this.distributionBehavior = distributionBehavior;
+  }
+
+  @Override
+  public void processNewCommand(final TypedRecord<AuthorizationRecord> command) {
+    final var authorizationToCreate = command.getValue();
+
+    // TODO check permissions of command.getAuthorizations().get("user")
```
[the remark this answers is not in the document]
Ben-Sheppard: this can be removed for now, it will be used in the future when we implement authorization checks at the command level

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCreateProcessor.java
```diff
@@ -0,0 +1,97 @@
...
+
+    newAuthorizationRecord.reset();
+    newAuthorizationRecord.wrap(command.getValue());
+    newAuthorizationRecord.setAuthorizationKey(key);
+
+    stateWriter.appendFollowUpEvent(key, AuthorizationIntent.CREATED, newAuthorizationRecord);
+    responseWriter.writeEventOnCommand(
+        key, AuthorizationIntent.CREATED, newAuthorizationRecord, command);
+
+    distributionBehavior.distributeCommand(key, command);
```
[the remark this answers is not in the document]
Ben-Sheppard: Makes sense, I will change it and then follow up at a different time with the user one

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/CreateAuthorizationTest.java
```diff
@@ -0,0 +1,99 @@
...
+            .withOwnerType(AuthorizationOwnerType.USER)
+            .withResourceKey("resource")
+            .withResourceType("bpmn-id")
+            .withPermissions(permissions)
+            .create();
+    // when
+    final var duplicatedAuthorizationRecord =
+        ENGINE
+            .authorization()
+            .newAuthorization(System.currentTimeMillis() + 1)
```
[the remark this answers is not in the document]
Ben-Sheppard: I agree, that is unexpected. I also agree that the key shouldn't be set here. I've removed this change

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
Ben-Sheppard: I agree

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
Ben-Sheppard: The resource key will come back with the search responses where the key would come from for updates/deletes. Should we also return the record we store here too?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
Ben-Sheppard: I believe that during the POC we had this as an idea of going from authorization values -> generated key -> Authorization Record. I think its not quite visible in this PR because I haven't added the update/delete side of things however, I don't think its actually needed, in the long term the Identity UI will be provided with the generated key for the record so we will be able to access the data directly instead of a hop. 

I'll remove it at least for now

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -0,0 +1,97 @@
...
+  private final ColumnFamily<
+          DbCompositeKey<DbCompositeKey<DbString, DbString>, DbCompositeKey<DbString, DbString>>,
+          PersistedPermissions>
```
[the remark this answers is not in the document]
Ben-Sheppard: Great question @remcowesterhoud, If we take the owner type as an example, the values could be any one of 
``` USER, ROLE, GROUP```
what I cannot say is if there would be a `role` with the same key as a `group`. For now I'd opt for keeping this complex key to be certain on the uniqueness and use a comment but if you feel strongly otherwise I can change it or follow up after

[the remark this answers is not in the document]
Comment (Ben-Sheppard) [reply zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java:38]: Regarding resource key, I'm not 100% sure, maybe thats something you could provide input on, resource type here is any resource in the system i.e. `authorization`, `process definition` and so on.

File: zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/AuthorizationRecord.java (hunk 11 above)
Comment (Ben-Sheppard) [reply zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/AuthorizationRecord.java:31]: I think it could be an enum, essentially it will be a defined list of the possible resources you can apply authorizations to in Camunda, for now I'd leave it as is but we can follow up is thats a better approach

File: zeebe/gateway-protocol/src/main/proto/rest-api.yaml (hunk 4 above)
Comment (remcowesterhoud) [reply zeebe/gateway-protocol/src/main/proto/rest-api.yaml:698]: No, if you don't need the key directly to store it somewhere Identity-side and it's good enough to read it from the exported records that's fine by me. Thanks!

File: zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java (hunk 17 above)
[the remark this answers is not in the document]
Comment (Ben-Sheppard) [reply zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java:39]: We can indeed, so roles and groups are created in the UI in SM, in SaaS the roles are taken from Console and Groups are created via an Identity component within Console. They have a very minimal set of fields Name, Description.

Currently the unique key of these are the name however they are unique _within_ their context (due to the table split. When an Authorization is created currently it is unique in our databases with the following fields 
```
"entityId", "entityType", "organizationId", "resource_id", "permission"
```
So here the only new field resource type (entityId/entityType maps to ownerKey/ownerType).

For this reason I believe that we need the distinction of ownerType because its very plausible that a group and role could cross over in name IMO.

Resource type is something that I'd need your advice on for the wider platform picture, we need to store the type at least so we can perform targeted `get()` calls for a specific resource type when checking authorizations at a command level however if the type needs to be in the key is unclear to me. The resourceType here can be though of _any_ resource we have in the platform so `Authorization`, `Group`, `Process Definition`, etc... so I'm not sure I can confidently say that there will never be an overlap in resource key especially if we use the name of the entity as a key and not have Zeebe generate the key.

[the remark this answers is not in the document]
Comment (remcowesterhoud) [reply zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java:39]: So, I had a good think about how we should store Authorizations in a nice way that makes it easy and hopefully performant to verify user permissions later on. Bear with me here 😄 

**ColumnFamily**
I believe we should have a ColumnFamily that allows easy searching by ownerKey. My proposal is:
`ownerKey - resourceType - permissionType -> Authorizations`

An example key would be `12345ProcessInstanceCREATE`. With this key we can do a targeted search for the authorizations of this owner, for this resource and with this permissionType. An `ownerKey` will be the key of a user, a group or a role.

The Authorizations object that's the value will contain a list of resource keys (eg: `bpmnProcessId:*`). Potentially we could even split this is another object where we store the `bpmnProcessId` and the `*` in separate fields for easier accesibility.

**How does it work?**
When we create a new Authorization we should check if we already have Authorizations for this ownerKey + resourceType + permissionType stored in the state. If we don't we create a new one. 
If we do we fetch it, check if the permission doesn't exist yet and if not add it to the Authorization object before updating it in the state.

Deleting an Authorization will be similar. We fetch the Authorization from the state by ownerKey + resourceType + permissionType. If it doesn't exist we can reject the command. If it does exist we can retrieve the permissions from it. If the permission to delete doesn't exist we reject the command. If it does, we remove it and update the Authorizations.

--------------

This does require some extra complexity on Authorization creation and deletion, but it will reduce complexity and increase performance on permission validation. In my opinion that's a good tradeoff.

Note that I did not discuss this datamodel within the team yet, as no kickoff has been held at this time.

File: zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/AuthorizationRequestValidator.java [hunk 18]
```diff
@@ -0,0 +1,43 @@
...
+        || authorizationAssignRequest.getResourceKey().isBlank()) {
+      violations.add(ERROR_MESSAGE_EMPTY_ATTRIBUTE.formatted("resourceKey"));
+    }
+
+    if (authorizationAssignRequest.getResourceType() == null) {
+      violations.add(ERROR_MESSAGE_EMPTY_ATTRIBUTE.formatted("resourceType"));
+    }
+
+    return createProblemDetail(violations);
+  }
```
remcowesterhoud, line 42: ❌ We should make sure we have unit tests that verify these validation.

zeebe/protocol-impl/src/test/java/io/camunda/zeebe/protocol/impl/JsonSerializableToJsonTest.java
```diff
@@ -2529,6 +2531,55 @@ final var record = new ClockRecord().offsetBy(30);
...
+      {
+        "Empty AuthorizationRecord",
+        (Supplier<AuthorizationRecord>)
+            () ->
+                new AuthorizationRecord()
+                    .setOwnerKey("owner")
+                    .setOwnerType(AuthorizationOwnerType.USER)
+                    .setResourceKey("resource")
+                    .setResourceType("type")
+                    .setPermissions(List.of("permission")),
```
remcowesterhoud, line 2570: 🔧 Permissions is not required in this record, so it's not truely empty atm. Let's remove this and verify permissions get serialized to an empty list.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -0,0 +1,91 @@
...
+import io.camunda.zeebe.protocol.ZbColumnFamilies;
+import io.camunda.zeebe.protocol.impl.record.value.authorization.AuthorizationRecord;
+import io.camunda.zeebe.protocol.record.value.AuthorizationOwnerType;
+
+public class DbAuthorizationState implements AuthorizationState, MutableAuthorizationState {
+  private final PersistedAuthorization persistedAuthorization = new PersistedAuthorization();
+
+  private final PersistedPermissions persistedPermissions = new PersistedPermissions();
+
+  private final DbLong authorizationKey;
```
remcowesterhoud, line 27: ❌ This is never used

remcowesterhoud (changes requested):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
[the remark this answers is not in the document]
remcowesterhoud: Regarding this, we could also fix it in a followup PR

remcowesterhoud (approved):
Nice work @Ben-Sheppard  🚀

remcowesterhoud:
I rebased and fixed the flaky test. Let's see if the CI passes now 😄

PR #21031, opened 2024-08-12
Set helm install arguments correctly

ChrisKujawa:
### Problem
Arguments for the benchmark are merged into one line, where comments, are cutting the rest.

Resulting values for an installation missing half of the args
```sh
$ helm get values ck-operate-benchmark
USER-SUPPLIED VALUES:
camunda-platform:
  zeebe:
    image:
      repository: gcr.io/zeebe-io/zeebe
      tag: ck-operate-benchmark-1ad8f375
global:
  image:
    tag: ck-operate-benchmark-1ad8f375
```

See recent run https://github.com/camunda/camunda/actions/runs/10318610814/job/28565648616#step:7:1

### PR description

This PR removes the comments, so values are correctly set.

### DIFF
--- a/.github/workflows/zeebe-benchmark.yml
+++ b/.github/workflows/zeebe-benchmark.yml
@@ -292,6 +292,7 @@ jobs:
           --namespace ${{ inputs.name }}
           --create-namespace
           --reuse-values
+          --render-subchart-notes
           --set global.image.tag=${{ needs.build-benchmark-images.outputs.image-tag }}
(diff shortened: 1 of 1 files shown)


Zelldon:
BTW This also affects the current medic benchmarks, as they have only set half of the properties. We need to re-run the medic benchmark workflow after merging this fix.

npepinpe (approved):

koevskinikola (approved):
🚀

Issue #20478, opened 2024-07-24
Add Correlate Message command to the client


remcowesterhoud:
<h2>Description</h2>

Extend the client with the new command. Follow steps in the REST controller dev guide. At the time of writing this not merged yet, but the PR can be found[ here](https://github.com/camunda/camunda/pull/20449/files)

PR #21042, opened 2024-08-12
Add message correlation REST API

### LINKED ISSUE #20477: Add new `CorrelateMessage` REST endpoint

remcowesterhoud:
## Description

This PR adds the endpoint for the Message Correlation command. The client will be implemented as part of a different issue (#20478)

This endpoint is special compared to the endpoints that already exist in two ways:

1. It takes a tenant id in the request. This tenant must be verified and we must validate the client is authenticated for this tenant.
2. It needs a response mapper. Currently only Job endpoint do response mapping, but they work in their own special way.

## Checklist

- [ ] for CI changes:

## Related issues

closes #20477

### DIFF
--- a/service/src/main/java/io/camunda/service/MessageServices.java
+++ b/service/src/main/java/io/camunda/service/MessageServices.java
@@ -36,14 +36,12 @@ public final class MessageServices<T> extends ApiServices<MessageServices<T>> {
   }
 
   public CompletableFuture<MessageCorrelationRecord> correlateMessage(
-      final CorrelateMessageRequest correlationRequest, final boolean multiTenancyEnabled) {
+      final CorrelateMessageRequest correlationRequest) {
     final var brokerRequest =
         new BrokerCorrelateMessageRequest(
                 correlationRequest.name, correlationRequest.correlationKey)
             .setVariables(getDocumentOrEmpty(correlationRequest.variables))
-            .setTenantId(
-                ensureTenantIdSet(
-                    "CorrelateMessage", correlationRequest.tenantId, multiTenancyEnabled));
+            .setTenantId(correlationRequest.tenantId);
     return sendBrokerRequest(brokerRequest);
   }
 
--- a/service/src/main/java/io/camunda/service/MessageServices.java
+++ b/service/src/main/java/io/camunda/service/MessageServices.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.service;
+
+import io.camunda.search.clients.CamundaSearchClient;
+import io.camunda.service.security.auth.Authentication;
+import io.camunda.service.transformers.ServiceTransformers;
+import io.camunda.zeebe.broker.client.api.BrokerClient;
+import io.camunda.zeebe.gateway.cmd.IllegalTenantRequestException;
+import io.camunda.zeebe.gateway.cmd.InvalidTenantRequestException;
+import io.camunda.zeebe.gateway.impl.broker.request.BrokerCorrelateMessageRequest;
+import io.camunda.zeebe.protocol.impl.record.value.message.MessageCorrelationRecord;
+import io.camunda.zeebe.protocol.record.value.TenantOwned;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
+
+public final class MessageServices<T> extends ApiServices<MessageServices<T>> {
+
+  private static final Pattern TENANT_ID_MASK = Pattern.compile("^[\\w\\.-]{1,31}$");
+
+  public MessageServices(final BrokerClient brokerClient, final CamundaSearchClient searchClient) {
+    this(brokerClient, searchClient, null, null);
+  }
+
+  public MessageServices(
+      final BrokerClient brokerClient,
+      final CamundaSearchClient searchClient,
+      final ServiceTransformers transformers,
+      final Authentication authentication) {
+    super(brokerClient, searchClient, transformers, authentication);
+  }
+
+  @Override
+  public MessageServices<T> withAuthentication(final Authentication authentication) {
+    return new MessageServices<>(brokerClient, searchClient, transformers, authentication);
+  }
+
+  public CompletableFuture<MessageCorrelationRecord> correlateMessage(
+      final CorrelateMessageRequest correlationRequest, final boolean multiTenancyEnabled) {
+    final var brokerRequest =
+        new BrokerCorrelateMessageRequest(
+                correlationRequest.name, correlationRequest.correlationKey)
+            .setVariables(getDocumentOrEmpty(correlationRequest.variables))
+            .setTenantId(
+                ensureTenantIdSet(
+                    "CorrelateMessage", correlationRequest.tenantId, multiTenancyEnabled));
+    return sendBrokerRequest(brokerRequest);
+  }
+
+  private String ensureTenantIdSet(
+      final String commandName, final String tenantId, final boolean multiTenancyEnabled) {
+    final var hasTenantId = !StringUtils.isBlank(tenantId);
+    if (!multiTenancyEnabled) {
+      if (hasTenantId && !TenantOwned.DEFAULT_TENANT_IDENTIFIER.equals(tenantId)) {
+        throw new InvalidTenantRequestException(commandName, tenantId, "multi-tenancy is disabled");
+      }
+
+      return TenantOwned.DEFAULT_TENANT_IDENTIFIER;
+    }
+
+    if (!hasTenantId) {
+      throw new InvalidTenantRequestException(
+          commandName, tenantId, "no tenant identifier was provided.");
+    }
+
+    if (tenantId.length() > 31) {
+      throw new InvalidTenantRequestException(
+          commandName, tenantId, "tenant identifier is longer than 31 characters");
+    }
+
+    if (!TenantOwned.DEFAULT_TENANT_IDENTIFIER.equals(tenantId)
+        && !TENANT_ID_MASK.matcher(tenantId).matches()) {
+      throw new InvalidTenantRequestException(
+          commandName, tenantId, "tenant identifier contains illegal characters");
+    }
+
+    final List<String> authorizedTenants;
+    try {
+      authorizedTenants = authentication.authenticatedTenantIds();
+    } catch (final Exception e) {
+      throw new InvalidTenantRequestException(
+          commandName, tenantId, "tenant could not be retrieved from the request context", e);
+    }
+    if (authorizedTenants == null) {
+      throw new InvalidTenantRequestException(
+          commandName, tenantId, "tenant could not be retrieved from the request context");
+    }
+    if (!authorizedTenants.contains(tenantId)) {
+      throw new IllegalTenantRequestException(
+          commandName, tenantId, "tenant is not authorized to perform this request");
+    }
+
+    return tenantId;
+  }
+
+  public record CorrelateMessageRequest(
+      String name, String correlationKey, Map<String, Object> variables, String tenantId) {}
+}
--- a/service/src/main/java/io/camunda/service/MessageServices.java
+++ b/service/src/main/java/io/camunda/service/MessageServices.java
@@ -16,7 +16,7 @@ import io.camunda.zeebe.protocol.impl.record.value.message.MessageCorrelationRec
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
-public final class MessageServices<T> extends ApiServices<MessageServices<T>> {
+public final class MessageServices extends ApiServices<MessageServices> {
 
   public MessageServices(final BrokerClient brokerClient, final CamundaSearchClient searchClient) {
     this(brokerClient, searchClient, null, null);
@@ -31,8 +31,8 @@ public final class MessageServices<T> extends ApiServices<MessageServices<T>> {
   }
 
   @Override
-  public MessageServices<T> withAuthentication(final Authentication authentication) {
-    return new MessageServices<>(brokerClient, searchClient, transformers, authentication);
+  public MessageServices withAuthentication(final Authentication authentication) {
+    return new MessageServices(brokerClient, searchClient, transformers, authentication);
   }
 
   public CompletableFuture<MessageCorrelationRecord> correlateMessage(
--- a/service/src/main/java/io/camunda/service/MessageServices.java
+++ b/service/src/main/java/io/camunda/service/MessageServices.java
@@ -11,20 +11,12 @@ import io.camunda.search.clients.CamundaSearchClient;
 import io.camunda.service.security.auth.Authentication;
 import io.camunda.service.transformers.ServiceTransformers;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
-import io.camunda.zeebe.gateway.cmd.IllegalTenantRequestException;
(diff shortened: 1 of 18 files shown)


service/src/main/java/io/camunda/service/MessageServices.java
```diff
@@ -0,0 +1,50 @@
...
+import io.camunda.search.clients.CamundaSearchClient;
+import io.camunda.service.security.auth.Authentication;
+import io.camunda.service.transformers.ServiceTransformers;
+import io.camunda.zeebe.broker.client.api.BrokerClient;
+import io.camunda.zeebe.gateway.impl.broker.request.BrokerCorrelateMessageRequest;
+import io.camunda.zeebe.protocol.impl.record.value.message.MessageCorrelationRecord;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public final class MessageServices<T> extends ApiServices<MessageServices<T>> {
```
tmetzke, line 19: ❓ Do we need the generic here for anything specific? Looking at the [IncidentServices](https://github.com/camunda/camunda/blob/main/service/src/main/java/io/camunda/service/IncidentServices.java), for example, we don't usually need this. The JobServices need the generic because the type of the job activation handler is generic as well. But most (if any) other services should not need this except for special cases. I don't see any reason for the `UserServices` to have it, for example. But maybe I am missing something here 🤔

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/MessageController.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @PostMapping(
+      path = "/correlation",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  public CompletableFuture<ResponseEntity<MessageCorrelationResponse>> correlateMessage(
+      @RequestBody final MessageCorrelationRequest correlationRequest) {
+    return RequestMapper.toMessageCorrelationRequest(
+            correlationRequest, multiTenancyCfg.isEnabled())
+        .fold(this::correlateMessage, RestErrorMapper::mapProblemToCompletedResponse);
```
tmetzke, line 49: 🔧 If I am not mistaken, the `fold` message has already been adjusted to swap left and right functions (left going, well, left now 🙂).

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/MessageController.java
```diff
@@ -0,0 +1,61 @@
...
+      final MultiTenancyCfg multiTenancyCfg) {
+    this.messageServices = messageServices;
+    this.multiTenancyCfg = multiTenancyCfg;
+  }
+
+  @PostMapping(
+      path = "/correlation",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  public CompletableFuture<ResponseEntity<MessageCorrelationResponse>> correlateMessage(
```
tmetzke, line 45: ❌ The return type must be `CompletableFuture<ResponseEntity<Object>>` because the response can be a `ProblemDetail` or the success type, i.e. `MessageCorrelationResponse`. This will lead to some follow-up changes in the signatures of the other methods used in this controller, the `RequestMapper`, and the `ResponseMapper`.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -125,6 +129,24 @@ public static Either<ProblemDetail, ErrorJobRequest> toJobErrorRequest(
                 getMapOrEmpty(errorRequest, JobErrorRequest::getVariables)));
   }
 
+  public static Either<ProblemDetail, CorrelateMessageRequest> toMessageCorrelationRequest(
+      final MessageCorrelationRequest correlationRequest, final boolean multiTenancyEnabled) {
+    final var validationErrorResponse = validateMessageCorrelationRequest(correlationRequest);
```
tmetzke, line 134: 🔧 How about we validate the provided tenant id here or in the `validateMessageCorrelationRequest`? That way, any issues due to invalid input data would consistently lead to a `ProblemDetail`. Currently, an exception is thrown from this method if the tenant id is not valid, if I'm not mistaken?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -154,6 +176,19 @@ public static CompletableFuture<ResponseEntity<Object>> executeServiceMethod(
                     .orElseGet(result));
   }
 
+  public static <BrokerResponseT, RestResponseT>
+      CompletableFuture<ResponseEntity<RestResponseT>> executeServiceMethod(
+          final Supplier<CompletableFuture<BrokerResponseT>> method,
+          final Function<BrokerResponseT, ResponseEntity<RestResponseT>> result) {
```
tmetzke, line 182: 💭 I haven't tried but at a first glance I would hope we can merge this method with the `executeServiceMethod` above or at least reuse one from the other.

tmetzke (changes requested):
👍 I like the overall approach here, it fits very well into what we have already. 
👍 I like the test coverage we have for the input of the endpoint
🔧 I added some suggestions, nothing major, but there are some smaller things we should correct.

service/src/main/java/io/camunda/service/MessageServices.java
remcowesterhoud: Probably not actually. I took inspiration from the JobServices so I figured it was just how things were done 😄 I'll remove it.

remcowesterhoud: I'll do it for the UserServices as well while I'm at it as I let it pass review 👍

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/MessageController.java
remcowesterhoud: @tmetzke Do you know why my tests cases succeed if this would be a problem?

tmetzke: Isn't this actually failing: https://github.com/camunda/camunda/actions/runs/10349786752/job/28644919780?pr=21042? 😄

remcowesterhoud: No not really, that's because of the Either.fold that was changed

tmetzke: Ah yeah I see 👍

tmetzke: Well, I guess, at runtime, the generic is removed by Java anyway and ignored, isn't it?

tmetzke: Well, it's going to be a `ResponseEntity` actually and not any subtype 😄 So I think `CompletableFuture<ResponseEntity<?>>` would be closest and `CompletableFuture<ResponseEntity<Object>>` is basically the same and more consistent with the other endpoints 😁

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
remcowesterhoud: 🤔 It does, but interestingly enough it does return a ProblemDetail so it seems it gets caught somewhere. Maybe some Spring magic?

But I agree we shouldn't have magic here. I wanted to put it in the `MessageCorrelateValidator` at first but ran into the problem that apart from validating, this also makes some decision on what the `tenantId` would be. So that makes it troublesome since we only return a `ProblemDetail` there. I'll see if I can think of something else.

tmetzke: > 🤔 It does, but interestingly enough it does return a ProblemDetail so it seems it gets caught somewhere. Maybe some Spring magic?

Yeah, we have a global exception handler that catches exceptions as a last resort and returns HTTP 400 problem details.

> But I agree we shouldn't have magic here. I wanted to put it in the MessageCorrelateValidator at first but ran into the problem that apart from validating, this also makes some decision on what the tenantId would be. So that makes it troublesome since we only return a ProblemDetail there. I'll see if I can think of something else.

I see, yeah. The tenant validator could return an Either with either the tenant id to use or validation errors, right?

remcowesterhoud:
@tmetzke I've addressed all your comments. Please have another look 👀

tmetzke (approved):
👍 Great work, I really like the improvements and the new tenant validation, for example 🚀 
🔧 I added some minor cleanup suggestions that don't block this.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/MultiTenancyValidator.java
```diff
@@ -7,60 +7,78 @@
...
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.regex.Pattern;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ProblemDetail;
 
 public final class MultiTenancyValidator {
+  private static final String MESSAGE_FORMAT =
+      "Expected to handle request %s with tenant identifier '%s', but %s";
```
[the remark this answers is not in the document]
remcowesterhoud: That's fair. I wasn't aware this class existed. This one is very specific to tenants though, so I doubt it'll get used by other validators. But if we want to collect all error messages in one spot that makes sense tom me.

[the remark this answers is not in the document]
tmetzke: We might revisit that idea later, let's see. For now, this is the simplest way of ensuring some consistency in REST error messages. This never works 100%, of course. We might also have to break it up again at some point if it becomes too large and there are more holistic approaches to message handling, e.g. when it comes to internationalization and such but let's not go there now 😄

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -125,6 +130,30 @@ public static Either<ProblemDetail, ErrorJobRequest> toJobErrorRequest(
...
+            .flatMap(
+                tenantId ->
+                    validateAuthorization(tenantId, multiTenancyEnabled, "Correlate Message")
+                        .map(Either::<ProblemDetail, String>left)
+                        .orElseGet(() -> Either.right(tenantId)))
+            .flatMap(
+                tenantId ->
+                    validateMessageCorrelationRequest(correlationRequest)
+                        .map(Either::<ProblemDetail, String>left)
+                        .orElseGet(() -> Either.right(tenantId)));
```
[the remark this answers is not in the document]
remcowesterhoud: Yea, it wasn't easy to write it either 😄 I will turn the var into an explicit type. As you said knowing the return type already helps a bunch in understanding what happens.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -144,19 +173,24 @@ public static Either<ProblemDetail, UpdateJobRequest> toJobUpdateRequest(
                 getLongOrZero(updateRequest, r -> updateRequest.getChangeset().getTimeout())));
   }
 
-  public static CompletableFuture<ResponseEntity<Object>> executeServiceMethod(
-      final Supplier<CompletableFuture<?>> method, final Supplier<ResponseEntity<Object>> result) {
+  public static <BrokerResponseT, RestResponseT>
+      CompletableFuture<ResponseEntity<RestResponseT>> executeServiceMethod(
```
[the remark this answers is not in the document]
remcowesterhoud: The way I see it a `ProblemDetail` is a REST response and because of that this is more specific to me than putting `Object` in there. If I put that in there I may aswell return an Either or whatever. The fact that it's not the happy-path response doesn't mean it's not a response 😄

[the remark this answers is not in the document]
remcowesterhoud: Oh hold on I misread. You're saying the target REST response. As in the function we pass returns a MessageCorrelationResponse, and that's not fair because it could be a ProblemDetail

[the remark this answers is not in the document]
tmetzke: I'm not arguing aginst that 😄 But here, the `RestResponseT` refers to the type returned by the `result` function. That function returns the "success" REST response type. Which is not the `ProblemDetail`. That might just be confusing, looking at this method and wondering how the successful REST response type will always be returned when there might be a failure case that returns something else. It's a purely theoretical discussion since Java doesn't care about the generic at runtime anyway 🤷😅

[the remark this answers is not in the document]
tmetzke: > Oh hold on I misread. You're saying the target REST response. As in the function we pass returns a MessageCorrelationResponse whcih we use as type, and that's not fair because it could be a ProblemDetail

Yeah exactly, saw that too late 😅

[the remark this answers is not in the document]
remcowesterhoud: Okay, I've fixed the generics. I probably can't merge the PR yet due to the issues with the merge queue.

tmetzke (approved):
🚀

tmetzke:
The gRPC tests need adjusting to reflect the new messaging for tenant failures.

tmetzke:
Any idea why `micrometer-commons` would be needed now as a dependency in `zeebe-gateway-rest`? 🤔

tmetzke:
> Any idea why `micrometer-commons` would be needed now as a dependency in `zeebe-gateway-rest`? 🤔

Found it, we're using `io.micrometer.common.util.StringUtils` in the `MultiTenancyValidator`. Maybe we can remove/replace that.

remcowesterhoud:
Oops, I just copied the class from gRPC, didn't realize it pulled a dependency in. I'll change it cause we can easily do this check ourselves.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/MultiTenancyValidator.java
```diff
@@ -71,7 +70,7 @@ public static Optional<ProblemDetail> validateAuthorization(
    */
   public static Either<ProblemDetail, String> validateTenantId(
       final String tenantId, final boolean multiTenancyEnabled, final String commandName) {
-    final var hasTenantId = !StringUtils.isBlank(tenantId);
+    final var hasTenantId = tenantId != null && !tenantId.trim().isEmpty();
```
tmetzke, line 73: 🙃 Isn't there an `isBlank` nowadays in the String type since Java 11, so we can write `tenantId != null && !tenantId.isBlank()`?

remcowesterhoud: Who are you, so wise in the ways of Strings

tmetzke: I think someone mentioned it in one of my PRs a couple of months back, I'm just sharing the TIL experience 😄

PR #21043, opened 2024-08-12
test: ensure that controlled actor scheduler starts idle

### LINKED ISSUE #13493: Test `ZeebePartitionTest.shouldUseCurrentTermForInactiveTransition` is flaky
**Summary**

- How often does the test fail? Once so far.
- Do we suspect that it is a real failure? Possibly?

**Failures**

<pre>
Argument(s) are different! Wanted:
partitionTransitionImpl.toInactive(
    3L
);
-> at io.camunda.zeebe.broker.system.partitions.impl.PartitionTransitionImpl.toInactive(PartitionTransitionImpl.java:51)
Actual invocations have different arguments:
partitionTransitionImpl.setConcurrencyControl(
    io.camunda.zeebe.scheduler.ActorControl@520f1862
);
-> at io.camunda.zeebe.broker.system.partitions.ZeebePartition.<init>(ZeebePartition.java:77)
partitionTransitionImpl.updateTransitionContext(
    io.camunda.zeebe.broker.system.partitions.PartitionStartupAndTransitionContextImpl@4ad0465c
);
-> at io.camunda.zeebe.broker.system.partitions.ZeebePartition.lambda$onActorStarting$0(ZeebePartition.java:121)
partitionTransitionImpl.toInactive(
    0L
);
-> at io.camunda.zeebe.broker.system.partitions.ZeebePartition.transitionToInactive(ZeebePartition.java:284)
partitionTransitionImpl.transitionTo(
    0L,
    INACTIVE
);
-> at io.camunda.zeebe.broker.system.partitions.impl.PartitionTransitionImpl.toInactive(PartitionTransitionImpl.java:51)
</pre>
</details>

Occurred in: https://github.com/camunda/zeebe/actions/runs/5553577122/jobs/10142291322

**Hypotheses**

**Logs**

[ZeebePartitionTestFlaky.zip](https://github.com/camunda/zeebe/files/12049966/ZeebePartitionTestFlaky.zip)

lenaschoenburg:
When starting the `ControlledActorScheduler`, we also need to ensure that it runs long enough to become idle initially. Without waiting, the thread might start a bit late, after initial actor tasks were submitted already. In that case the thread _should_ be idle and wait for an explicit `workUntilDone` call and not execute the task. We now ensure this by waiting for idle state after starting the thread.

This hopefully closes #13493 because the mocked/spied objects are no longer accessed concurrently.

### DIFF
--- a/zeebe/scheduler/src/test/java/io/camunda/zeebe/scheduler/testing/ControlledActorSchedulerExtension.java
+++ b/zeebe/scheduler/src/test/java/io/camunda/zeebe/scheduler/testing/ControlledActorSchedulerExtension.java
@@ -64,6 +64,7 @@ public class ControlledActorSchedulerExtension implements BeforeEachCallback, Af
     actorScheduler = builder.build();
     controlledActorTaskRunner = actorTaskRunnerFactory.controlledThread;
     actorScheduler.start();
+    controlledActorTaskRunner.waitUntilDone();
   }
 
   public ActorFuture<Void> submitActor(final Actor actor) {
--- a/zeebe/scheduler/src/test/java/io/camunda/zeebe/scheduler/testing/ControlledActorSchedulerRule.java
+++ b/zeebe/scheduler/src/test/java/io/camunda/zeebe/scheduler/testing/ControlledActorSchedulerRule.java
@@ -50,6 +50,7 @@ public final class ControlledActorSchedulerRule extends ExternalResource {
   @Override
   protected void before() {
     actorScheduler.start();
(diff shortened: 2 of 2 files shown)


npepinpe (changes requested):
:x: I think we need to add it to the `ControlledActorSchedulerExtension` as well, no?

lenaschoenburg:
Good catch, didn't think of that one :+1:

deepthidevaki (approved):
Thanks

npepinpe (approved):

PR #21058, opened 2024-08-12
fix: skip Optimize in most builds

ChrisKujawa:
## Description
Building Optimize in Zeebe and common pipelines caused most of the time failing builds due to timeouts. Likely due to longer build time, and other unknowns.

To get the team time to work and fix such issues, we ignore for now the Optimize part. The build time needs to be improved, and Optimize needs to migrate
to the common parent.

See long slack thread https://camunda.slack.com/archives/C071KP5BTHB/p1723460277287849


npepinpe (approved):
Should we add it to the `-Dquickly` flag as well for now?

Zelldon:
I think I need to merge into https://github.com/camunda/camunda/pull/20941 otherwise I might not get both merged separatelty.

Zelldon:
I merged https://github.com/camunda/camunda/pull/20941 into this PR. This reduced the build performance once more, as Zeebe ci has been reduced.

OmranAbazid (approved):

buccarel (approved):

Issue #18750, opened 2024-05-23
CI: avoid GitHub Action cache over usage errors


cmur2:
## Description

In the monorepo we encounter intermittent build failures on jobs that use the GitHub Action cache (mostly indirectly via actions like `setup-java`, etc) with error messages like `Error: Cache upload failed because file read failed with EBADF: bad file descriptor, read` or `Warning: Failed to save: Cache service responded with 429 during upload chunk.` HTTP 429 status code indicates that "the user has sent too many requests in a given amount of time", i.e. rate limiting. A quick look at GitHub [docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit) says:

> If you exceed your primary rate limit, you will receive a 403 or 429 response, and the x-ratelimit-remaining header will be 0. You should not retry your request until after the time specified by the x-ratelimit-reset header.

Unfortunately the GHA toolkit used to upload cache does not yet [handle](https://github.com/actions/toolkit/issues/1589) HTTP 429 errors, which is really annoying.

We should seek a way to avoid these errors by adjusting CI workflows to avoid exceeding the limit.

### Observations

This happens during situations where the https://github.com/camunda/zeebe/actions/caches are dramatically overused and exceeding the limit: ![image](https://github.com/camunda/zeebe/assets/800933/e2620882-1840-487d-bac2-17550b54448c)

Experience with this limit shows that over usage above 10G does not immediately lead to failures but if it gets too much, the GitHub API will prevent uploading data to the cache.

This failure is more likely in the monorepo because several CI workflows taken from other repositories have not been adjusted yet to the fact that all of them share the common 10G limit now. The cache usage details can be retrieved via the [API](https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#get-github-actions-cache-usage-for-a-repository).

Previous reports of this problem:

* https://camunda.slack.com/archives/C06HTSPD5AP/p1713969515800199
* https://camunda.slack.com/archives/C071KP5BTHB/p1716389476038289 with https://github.com/camunda/zeebe/actions/runs/9193202260/job/25283661896?pr=18647
* https://camunda.slack.com/archives/C071KP5BTHB/p1718292962257739 (problem mainly triggered by a change related to https://github.com/camunda/team-infrastructure/issues/517)

### Known Workarounds

[...]

PR #21063, opened 2024-08-12
ci: add optimize unit tests to unified CI workflow

misiekhardcore:
## Description

After merging optimize to the monorepo, there are no Optimize related checks performed to ensure that merged PRs are not breaking Optimize. This has already happened with some renovate PRs that had automerged and broke up Optimize.
 Not the whole Optimize CI workflow can be used as a requred check, cause in the current state it takes too much time to perform, but running at least the unit tests will give us more security

## Checklist

- [x] for CI changes:

## Related issues

relates to #21072


.github/workflows/ci.yml
```diff
@@ -417,6 +421,11 @@ jobs:
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
 
+  optimize-unit-tests:
+    name: Unit Tests
+    uses: ./.github/workflows/optimize-unit-tests.yml
+    secrets: inherit
```
Zelldon, line 427: ❌ Please take a look at the other jobs. We need to restrict the execution via the path filtering, otherwise it is always run.

Furthermore, right now it is not considered in the check results. That means even it is running it doesn't matter as the result it just ignored.

Example from java checks
>   java-checks:
>     if: needs.detect-changes.outputs.java-code-changes == 'true'
>     needs: [detect-changes]

RomanJRW (dismissed):

.github/workflows/ci.yml
misiekhardcore: there is a path filtering in that workflow itself so I thought it makes no sense to duplicate it. Does it make sense or should I still add another path filter nayway?

OmranAbazid: I think it should follow the pattern described here:
https://github.com/camunda/camunda/wiki/CI-&-Automation#implementation

OmranAbazid (commented):
See [comment](https://github.com/camunda/camunda/pull/21063/files#r1715408065)

.github/workflows/ci.yml
cmur2: Yes please align with and follow the pattern that Omran linked to have a common approach everywhere

cmur2 (commented):
Make sure to also follow [other inclusion criteria](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria) such as integration with Vault and emitting CI health metrics

.github/workflows/optimize-unit-tests.yml
```diff
@@ -49,6 +49,16 @@ jobs:
         retention-days: 7
         if-no-files-found: ignore
 
+    - name: Observe build status
+      if: always()
+      continue-on-error: true
+      uses: ./.github/actions/observe-build-status
+      with:
+        build_status: ${{ contains(steps.*.conclusion, 'failure') && 'failure' || 'success' }}
```
cmur2, line 57: Please use the code from https://github.com/camunda/camunda/wiki/CI-&-Automation#metrics-collection which uses `build_status: ${{ job.status }}`

.github/workflows/optimize-unit-tests.yml
```diff
@@ -49,12 +49,25 @@ jobs:
...
+        secret_vault_address: ${{ secrets.VAULT_ADDR }}
+        secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+
   unit-tests-frontend:
     name: Optimize Unit Tests - Frontend
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 10
     needs: detect-changes
     if: ${{ needs.detect-changes.outputs.frontend-changes == 'true' }}
```
cmur2, line 67: Can this change detection logic here be refactored into the common place in `.github/actions/paths-filter` to avoid 2 layers of change detection where one would be enough?

OmranAbazid (approved):

cmur2:
@misiekhardcore it could be more straightforward to implement this PR by starting with an unmodified `ci.yml` to have a clean slate and be unbiased by existing code. Then you can add the `steps` from `optimize-unit-tests.yml` into it (copying, not calling) while following [the pattern mentioned before](https://github.com/camunda/camunda/wiki/CI-&-Automation#implementation) to avoid things like having duplicate change detection.

(Maybe this hint comes too late but for future PRs it could be a better approach)

cmur2:
Don't merge this PR without addressing my comments please 🙏🏼

.github/workflows/optimize-unit-tests.yml
misiekhardcore: You mean just moved from the `optimzie-paths-filter` to `paths-filter`?

misiekhardcore:
> @misiekhardcore it could be more straightforward to implement this PR by starting with an unmodified `ci.yml` to have a clean slate and be unbiased by existing code. Then you can add the `steps` from `optimize-unit-tests.yml` into it (copying, not calling) while following [the pattern mentioned before](https://github.com/camunda/camunda/wiki/CI-&-Automation#implementation) to avoid things like having duplicate change detection.
> 
> (Maybe this hint comes too late but for future PRs it could be a better approach)
> 
> EDIT: The duplication of the change detection also means that neither Optimize FE nor BE tests are run on this PR which changes GHA code that affects them, because the 2nd layer of change detection doesn't check for GHA code changes.

I can copy over the content of the unit tests workflow, but this will cause code duplication. I pushed new commit with the `paths-filter` update and adding the job to the list of `needs` in `check-results`.

.github/workflows/optimize-unit-tests.yml
cmur2: I also meant to not execute `./.github/actions/paths-filter` twice (once in `ci.yml` and once in this file) since that is redundant and thus unexpected.

cmur2:
> I can copy over the content of the unit tests workflow, but this will cause code duplication.

I assumed that `optimize-unit-tests.yml` can be deleted afterwards (and not called anymore from `optimize-ci.yml` since it is now called from `ci.yml` instead) thus no code duplication.

misiekhardcore:
@cmur2 I moved the unit tests to `ci.yml` as suggested and removed the unit tests workflow completely. The `build-platform-frontend` fails for Optimize because of breaking renovate update. This will be fixed in #21130. When changes here will be approved, I'll merge that PR first, then rebase this one and see if the build is fixed and then we can merge this

cmur2 (approved):

misiekhardcore:
@cmur2 any idea why is this failing in the merge queue? I see that some zeebe jobs are failing and optimize backend is failing because of timing out, but in this PR, the optimize backend job takes only like 3 minutes

cmur2:
>  EDIT: I don't know how this works, but yesterday evening these BE unit tests with 4 threads took 3mins, and today they seem to run longer than 10

Just a theory/hypothesis: It might be related to caching that makes it (sometimes) run fast if e.g. a Maven cache exists. If that is true than we still need to make sure it also runs when there is no cache available to speed it up.

misiekhardcore:
There is no build output when the operation gets canceled due to the time out. Its really hard to figure out what is the bottleneck here. The only thing I've noticed is that sometimes when it starts running test command, it hands for a while on the fist step: `Scanning for projects...`. But I can't find anything about why it takes a minute or two sometimes on this, and sometimes it just takes a second

cmur2:
> java checks are failing on a file I didn't modify

Fixed in https://github.com/camunda/camunda/pull/21172

cmur2:
> There is no build output when the operation gets canceled due to the time out.

There is for the steps before the timeout happens, no? I meant to compare that timing with the timings of the runs that you mentioned finished in under 3-4 minutes and figure out what the differences are.

Chances are when merging it like this that it will block many many people when it runs into timeouts again - we need to make sure to consistenly stay below timeouts before attempting merge.

cmur2:
@misiekhardcore you can rebase on `main` now

misiekhardcore:
@cmur2 I re-ran these backend unit tests couple of times and they seem to pass each time. Maybe we should give it a try?

cmur2:
> @cmur2 I re-ran these backend unit tests couple of times and they seem to pass each time. Maybe we should give it a try?

Sorry for the late reply, was out sick. Sure lets attempt to merge once CI is green - currently failing Optimize FE unit tests (maybe due to dependency upgrade?)

misiekhardcore:
@cmur2 yes, some breaking deps updates were made. I will open a PR to fix those first

misiekhardcore:
@cmur2 @Zelldon zeebe smoke tests keep timing out, is this a known issue?

cmur2 (approved):

buccarel (approved):

cmur2:
> @cmur2 @Zelldon zeebe smoke tests keep timing out, is this a known issue?

Not to me and it does not happen on `main` either (builds there take 5mins or less, [example](https://github.com/camunda/camunda/actions/runs/10482011366/job/29032432350)). These timeouts prevent this PR from passing the merge queue successfully and I'm not sure why they happen on this PR 🤔 

What I see in the logs for this PR and what seems to take much time (9mins) is Maven messages like `Artifact com.fasterxml.jackson:jackson-bom:pom:2.17.2 is present in the local repository, but cached from a remote repository ID that is unavailable in current build context, verifying that is downloadable`...

misiekhardcore:
looks like caching issues to me, but I believe these are not related to my changes?

misiekhardcore:
@cmur2 @Zelldon any idea why this is still failing in zeebe tests? This PR is blocked for more than a week now because of those issues

cmur2:
> @cmur2 @Zelldon any idea why this is still failing in zeebe tests? This PR is blocked for more than a week now because of those issues

I'm considering to increase the timeouts for the affected Zeebe jobs since Zeebe devs also could identify another cause for the unexpected long runtimes other than caching behavior (something which is to be addressed in https://github.com/camunda/camunda/issues/18750 ),

cmur2:
So close on the last attempt but then there was a network error with DockerHub.

cmur2:
@misiekhardcore We can try again or you can increase timeouts of the smoke tests to 25mins to see if that helps

Zelldon:
> @misiekhardcore We can try again or you can increase timeouts of the smoke tests to 25mins to see if that helps

@cmur2 caches seem to be full again ~120 gig. Looks like we experience also in the smoke tests issues due to this maybe? 

Other than that I feel like it might be also related to the nodes itself. For example in the IT tests we build in ~ 1 min. In the other smoke tests it takes ~12 min.
In the failing case it was longer than 19 min

cmur2:
> caches seem to be full again ~120 gig. Looks like we experience also in the smoke tests issues due to this maybe?

Not sure since historically cache usage up to ~300 - 500 gig has been working okayish (we still need to solve this in https://github.com/camunda/camunda/issues/18750 though).

> related to the nodes itself

It is mainly the smoke tests on macOs and Windows that cause troubles due to sporadic timeouts.

misiekhardcore:
@Zelldon I can increase these timeouts, but then, will we have to revert them after this PR merges?

cmur2:
> will we have to revert them after this PR merges

First let's see if that helps, we have anyways be prepared to revert the PR in case something bad happens (maybe the failures in the merge queue indicate an underlying problem that neither of us sees). Then later we can decide about the timeouts.

cmur2:
Interesting that this worked now and would've stayed even within the old timings.

misiekhardcore:
yeah, I think it would merge by luck this time, but it was very close to timing out for macOS tests

PR #21066, opened 2024-08-12
feat: allow dns resolution to fall back to TCP

lenaschoenburg:
## Description

Enables fallback to TCP in case DNS responses are truncated, see https://github.com/netty/netty/pull/9139


zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java
```diff
@@ -386,6 +386,7 @@ public CompletableFuture<MessagingService> start() {
                               new BiDnsQueryLifecycleObserverFactory(
                                   ignored -> metrics,
                                   new LoggingDnsQueryLifeCycleObserverFactory()))
+                          .socketChannelType(clientChannelClass)
```
npepinpe, line 389: :+1: Well TIL

npepinpe (approved):
Not sure this will fix the support issue, but it makes sense to do anyway :+1:

PR #21074, opened 2024-08-13
Run Zeebe CI from Common CI

ChrisKujawa:
## Description

As we have been clogged by Renovate PRs and Zeebe CI had over 400+ jobs queued. The problem on the current Zeebe CI was that it was triggered via merge queue, which doesn't support path filtering. This means it was always triggered, even for unrelated changes, like Optimize dependency updates which we had a lot. 

We came to the conclusion that it makes sense to call Zeebe CI from common CI, such that we can use the same path filtering and disable the other github events.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


.github/actions/paths-filter/action.yml
```diff
@@ -98,3 +107,12 @@ runs:
...
           - 'identity/client/**'
+
+  - uses: dorny/paths-filter@v3
+    id: filter-zeebe
+    with:
+      base: ${{ github.event.merge_group.base_ref || '' }}
+      ref: ${{ github.event.merge_group.head_ref || github.ref }}
+      filters: |
+        zeebe-change:
+          - 'zeebe/**'
```
lenaschoenburg, line 118: changes in dist/ common/ clients/ etc. should also trigger Zeebe CI.

Zelldon: Thanks. 👍🏼  Dist is part of

lenaschoenburg (commented):
Zeebe CI is also responsible for pushing snapshot artifacts to maven and docker hub. Before this change, this would have been done for every commit on main.
If I understand correctly, this is now only done if the change detection thinks Zeebe has changed, i.e. we no longer deploy snapshot artifacts when for example Operate changes.

Zelldon:
> Zeebe CI is also responsible for pushing snapshot artifacts to maven and docker hub. Before this change, this would have been done for every commit on main. If I understand correctly, this is now only done if the change detection thinks Zeebe has changed, i.e. we no longer deploy snapshot artifacts when for example Operate changes.

Yes that is correct.

.github/actions/paths-filter/action.yml
lenaschoenburg: What about the identiy, search and service modules?

Zelldon: Integration tests are part of the common ci. Or do you mean smoke tests?

Zelldon: > What about the identiy, search and service modules?

Not sure what you mean. You we should run Zeebe ci when identity changes?

lenaschoenburg: > run Zeebe ci when identity changes

We have a dependency on identity so yes we should run Zeebe CI, otherwise it can break when Identity makes changes.

> Integration tests are part of the common ci. Or do you mean smoke tests?

Ah okay. It's not clear to me what runs where but looks like you are correct and smoke tests are missing.

lenaschoenburg:
> Yes that is correct for zeebe artifacts

It's not just Zeebe artifacts, the `deploy-snapshots` job is deploying _all_ maven modules. I don't know if it's okay to just change behavior now, there's probably something that relieson reasonably up-to-date snapshot artifacts?

Zelldon:
> > Yes that is correct for zeebe artifacts
> 
> It's not just Zeebe artifacts, the `deploy-snapshots` job is deploying _all_ maven modules. I don't know if it's okay to just change behavior now, there's probably something that relieson reasonably up-to-date snapshot artifacts?

moved via https://github.com/camunda/camunda/pull/21074/commits/7b426ce516aafaacec0a4ca9fd00629f243037a2

npepinpe:
So we were relying on Zeebe CI for the other modules? :sweat_smile:

.github/actions/paths-filter/action.yml
Zelldon: Discussed with @npepinpe. Current state, identity is not really integrated into the mono-repo. 

Zeebe is using still the old identity-spring-boot-starter https://github.com/camunda/camunda/blob/main/parent/pom.xml#L980-L984

Zelldon:
@npepinpe @lenaschoenburg please take a look again. I hope I covered everything.

npepinpe (approved):
:rocket: Thanks! I think we have the large majority of triggers covered - if we missed some edge cases, we can add them again.

npepinpe:
So this means the required check name for Zeebe changed from `Test Summary` to `zeebe-ci / Test Summary`, therefore it fails to pass the required check for the merge group. If we change the branch protection rule to go for `zeebe-ci / Test Summary`, this will require ALL open PRs to be rebased after we merge. As a migration step, we can add a dummy `Test summary` job on the main CI workflow.

In, say, 4 weeks, we can remove the dummy job and update the branch protection rule. If you haven't merged or rebased since, then I think it's safe to say the PR was stale, and you will just have to rebase.

Issue #20759, opened 2024-08-05
[Doc Handling] Introduce document API


chillleader:
As part of the document handling capability epic, we are aiming to implement several new API endpoints that will be part of the unified C8 REST API.

- API proposal: https://docs.google.com/document/d/1fe4nYImFqCqGV1y0MXG4xe4a3U7Nb3cq_1n8OgJTSjk/edit#heading=h.aesesrj8lmz8

- API endpoints: create document, get document, delete document, generate public link

- The first iteration should be backed up by a mock in-memory data store implementation to allow integration testing of other components such as [Tasklist](https://github.com/camunda/camunda/issues/20709) and Connectors.

PR #21080, opened 2024-08-13
feat: document API draft

chillleader:
## Description

This PR implements a first draft version of the Document API definition and the REST controller, backed by a mock document store implementation. This should allow integration testing of other parts of the document handling epic while the document store functionality is still under development.

* The OpenAPI definition is meant to be more or less stable, slight adjustments possible along the way
* The in-memory data store implementation will be removed when the service implementation is available

## Checklist

- [ ] for CI changes:

## Related issues

#20759

### DIFF
--- a/service/src/main/java/io/camunda/service/CamundaServices.java
+++ b/service/src/main/java/io/camunda/service/CamundaServices.java
@@ -12,6 +12,7 @@ import io.camunda.service.security.auth.Authentication;
 import io.camunda.service.transformers.ServiceTransformers;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
 import io.camunda.zeebe.gateway.impl.job.ActivateJobsHandler;
+import java.util.HashMap;
 
 public final class CamundaServices extends ApiServices<CamundaServices> {
 
@@ -63,6 +64,11 @@ public final class CamundaServices extends ApiServices<CamundaServices> {
 
   public MessageServices messageServices() {
     return new MessageServices(brokerClient, searchClient, transformers, authentication);
+  }
+
+  public DocumentServices documentServices() {
+    return new DocumentServices(
+        brokerClient, searchClient, transformers, authentication, new HashMap<>());
   }
 
   @Override
--- a/service/src/main/java/io/camunda/service/CamundaServices.java
+++ b/service/src/main/java/io/camunda/service/CamundaServices.java
@@ -12,7 +12,6 @@ import io.camunda.service.security.auth.Authentication;
 import io.camunda.service.transformers.ServiceTransformers;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
 import io.camunda.zeebe.gateway.impl.job.ActivateJobsHandler;
-import java.util.HashMap;
 
 public final class CamundaServices extends ApiServices<CamundaServices> {
 
@@ -67,8 +66,7 @@ public final class CamundaServices extends ApiServices<CamundaServices> {
   }
 
   public DocumentServices documentServices() {
-    return new DocumentServices(
-        brokerClient, searchClient, transformers, authentication, new HashMap<>());
+    return new DocumentServices(brokerClient, searchClient, transformers, authentication);
   }
 
   @Override
--- a/service/src/main/java/io/camunda/service/DocumentServices.java
+++ b/service/src/main/java/io/camunda/service/DocumentServices.java
@@ -40,6 +40,14 @@ public class DocumentServices extends ApiServices<DocumentServices> {
     this.documents = documents;
   }
 
+  public DocumentServices(
+      final BrokerClient brokerClient,
+      final CamundaSearchClient searchClient,
+      final ServiceTransformers transformers,
+      final Authentication authentication) {
+    this(brokerClient, searchClient, transformers, authentication, new HashMap<>());
+  }
+
   @Override
   public DocumentServices withAuthentication(final Authentication authentication) {
     return new DocumentServices(
--- a/service/src/main/java/io/camunda/service/DocumentServices.java
+++ b/service/src/main/java/io/camunda/service/DocumentServices.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.service;
+
+import io.camunda.search.clients.CamundaSearchClient;
+import io.camunda.service.security.auth.Authentication;
+import io.camunda.service.transformers.ServiceTransformers;
+import io.camunda.zeebe.broker.client.api.BrokerClient;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.time.ZonedDateTime;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+
+public class DocumentServices extends ApiServices<DocumentServices> {
+
+  // TODO: This is an in-memory implementation, replace it with a real document store
+  private static final String STORE_ID = "default";
+  private final Map<String, byte[]> documents;
+
+  public DocumentServices(final BrokerClient brokerClient, final CamundaSearchClient searchClient) {
+    this(brokerClient, searchClient, null, null, new HashMap<>());
+  }
+
+  public DocumentServices(
+      final BrokerClient brokerClient,
+      final CamundaSearchClient searchClient,
+      final ServiceTransformers transformers,
+      final Authentication authentication,
+      final Map<String, byte[]> documents) {
+    super(brokerClient, searchClient, transformers, authentication);
+    this.documents = documents;
+  }
+
+  @Override
+  public DocumentServices withAuthentication(final Authentication authentication) {
+    return new DocumentServices(
+        brokerClient, searchClient, transformers, authentication, documents);
+  }
+
+  public CompletableFuture<DocumentReferenceResponse> createDocument(
+      DocumentCreateRequest request) {
+    validateStoreId(request.storeId);
+    final String id =
+        request.documentId != null ? request.documentId : UUID.randomUUID().toString();
+    if (documents.containsKey(id)) {
+      throw new CamundaServiceException("Document already exists: " + id);
+    }
+    final var contentInputStream = request.contentInputStream;
+    final byte[] content;
+    try {
+      content = contentInputStream.readAllBytes();
+      contentInputStream.close();
+    } catch (IOException e) {
+      throw new CamundaServiceException("Failed to read document content", e);
+    }
+    documents.put(id, content);
+    return CompletableFuture.completedFuture(
+        new DocumentReferenceResponse(id, STORE_ID, request.metadata));
+  }
+
+  public InputStream getDocumentContent(String documentId, String storeId) {
+    validateStoreId(storeId);
+    final var content = documents.get(documentId);
+    if (content == null) {
+      throw new CamundaServiceException("Document not found: " + documentId);
+    }
+    return new ByteArrayInputStream(content);
+  }
+
+  public CompletableFuture<Void> deleteDocument(String documentId, String storeId) {
+    validateStoreId(storeId);
+    final var content = documents.remove(documentId);
+    if (content == null) {
+      throw new CamundaServiceException("Document not found: " + documentId);
+    }
+    return CompletableFuture.completedFuture(null);
+  }
+
+  private void validateStoreId(String storeId) {
+    if (storeId != null && !storeId.equals(STORE_ID)) {
+      throw new CamundaServiceException(
+          "Unsupported store id: " + storeId + ", expected: " + STORE_ID);
+    }
+  }
+
+  public record DocumentCreateRequest(
+      String documentId,
+      String storeId,
+      InputStream contentInputStream,
+      DocumentMetadataModel metadata) {}
+
+  public record DocumentMetadataModel(
+      String contentType,
+      String fileName,
+      ZonedDateTime expiresAt,
+      Map<String, Object> additionalProperties) {}
+
+  public record DocumentReferenceResponse(
+      String documentId, String storeId, DocumentMetadataModel metadata) {}
+}
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -784,6 +784,197 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+  /documents:
+    post:
+      tags:
+        - Documents
+      summary: Upload document (experimental)
+      description: |
+        Upload a document to the Camunda 8 cluster.
+
+        :::note
+        This endpoint is experimental. It currently only supports an in-memory document store, which
+        is not meant for production use.
+        :::
+      parameters:
+        - name: storeId
+          in: query
+          required: false
+          description: The ID of the document store to upload the document to.
+          schema:
+            type: string
+        - name: documentId
+          in: query
+          required: false
+          description: >
+            The ID of the document to upload. If not provided, a new ID will be generated.
+            Specifying an existing ID will result in an error if the document already exists.
+          schema:
+            type: string
+      requestBody:
+        required: true
+        content:
+          multipart/form-data:
+            schema:
+              type: object
+              properties:
+                file:
+                  type: string
+                  format: binary
+                metadata:
+                  $ref: "#/components/schemas/DocumentMetadata"
+              required:
+                - file
+            encoding:
+              metadata:
+                contentType: application/json
+      responses:
+        '201':
+          description: The document was uploaded successfully.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DocumentReference"
+        '400':
+          description: >
+            The document upload failed. More details are provided in the response body.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+  /documents/{documentId}:
+    get:
+      tags:
+        - Documents
+      summary: Download document (experimental)
+      description: |
+        Download a document from the Camunda 8 cluster.
+
+        :::note
+        This endpoint is experimental. It currently only supports an in-memory document store, which
+        is not meant for production use.
+        :::
+      parameters:
+        - name: documentId
+          in: path
+          required: true
+          description: The ID of the document to download.
+          schema:
+            type: string
+        - name: storeId
+          in: query
+          required: false
+          description: The ID of the document store to download the document from.
+          schema:
+            type: string
+      responses:
+        '200':
+          description: The document was downloaded successfully.
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+        '404':
+          description: >
+            The document with the given ID was not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        '500':
+          description: >
+            An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+    delete:
+      tags:
+        - Documents
+      summary: Delete document (experimental)
+      description: |
+        Delete a document from the Camunda 8 cluster.
+
+        :::note
+        This endpoint is experimental. It currently only supports an in-memory document store, which
+        is not meant for production use.
+        :::
+      parameters:
+        - name: documentId
+          in: path
+          required: true
+          description: The ID of the document to delete.
+          schema:
+            type: string
+        - name: storeId
+          in: query
+          required: false
+          description: The ID of the document store to delete the document from.
+          schema:
+            type: string
+      responses:
+        '200':
+          description: The document was deleted successfully.
+        '404':
+          description: >
+            The document with the given ID was not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        '500':
+          description: >
+            An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+  /document/{documentId}/links:
+    post:
+      tags:
+        - Documents
+      summary: Create document link (experimental)
+      description: |
+        Create a link to a document in the Camunda 8 cluster.
+
+        :::note
+        This endpoint is experimental. It currently only supports an in-memory document store, which
+        is not meant for production use.
+        :::
+      parameters:
+        - name: documentId
+          in: path
+          required: true
+          description: The ID of the document to link.
+          schema:
+            type: string
+        - name: storeId
+          in: query
+          required: false
+          description: The ID of the document store to link the document from.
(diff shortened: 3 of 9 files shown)


tmetzke:
I haven't reviewed properly reviewed this yet, just some preliminary pointers:

❌ You have an `ObjectMapper` in the controller which should not be necessary. The metadata attribute is a request body value, which should be mapped to the target type you defined in the OpenAPI spec. Have a look at the [REST controller guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md) and also the other controllers like the `UserTaskController`, for reference. They are also using the defined model type classes directly, Spring maps the input data to those target types directly.

🔧 We try not to throw exceptions from the controllers but rather control the `ProblemDetail` we return from the controllers specifically. The other controllers show how we do this, e.g. by validating input and using the `RestErrorMapper` if validation or processing fails.

💭 How about we move the mocked document store into the `service` module already and properly set up a `DocumentServices` class over there that uses it? The REST controller can interact with the services properly and doesn't need to change later when the implementation in the services layer changes.

chillleader:
@tmetzke thanks for the feedback! I agree with all the points you mentioned. As for using `ObjectMapper` directly, this is slightly more complex in this case because we are not consuming it directly from the request body. Instead, JSON metadata is part of a multipart request, and parsing it directly with `@RequestBody` didn't work for me. I have another approach to try, hopefully I will work around this and get rid of the object mapper here.

chillleader:
Nevermind, I got it working by using `@RequestPart`. The key to success was providing the correct content-type value for the metadata request part (`application/json`, it would default to `application/octet-stream` if not specified). The only issue is, Swagger UI doesn't seem to insert the content type, even though I specified it explicitly in the contract - but I guess we can figure it out later. Tried it via Postman and everything worked.

I also incorporated your other feedback 👍

service/src/main/java/io/camunda/service/DocumentServices.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+
+  public CompletableFuture<DocumentReferenceResponse> createDocument(
+      DocumentCreateRequest request, InputStream contentInputStream) {
+
+    validateStoreId(request.storeId);
+
+    final String id = request.documentId != null ? request.documentId : UUID.randomUUID().toString();
+    final byte[] content;
```
sbuettner, line 52: What about returning an `Either<DocumentCreationError, DocumentCreationResult>` to provide more information in case of an error?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -771,6 +771,199 @@ paths:
...
+                file:
+                  type: string
+                  format: binary
+                metadata:
+                  $ref: "#/components/schemas/DocumentMetadata"
+              required:
+                - file
+            encoding:
+              metadata:
+                contentType: application/json
```
tmetzke, line 818: 👍 This is nice, you have to appreciate the options and use cases OpenAPI already covers out of the box 🚀

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1562,9 +1755,56 @@ components:
             description: The key of the first process instance the message correlated with
             type: integer
             format: int64
+    DocumentReference:
+      type: object
+      properties:
+        $documentType:
```
tmetzke, line 1761: 🔧 As discussed, do we need this to be prefixed with `$`?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/DocumentValidator.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    if (metadata.getExpiresAt() != null) {
+      String expiresAt = metadata.getExpiresAt();
+      var dateTime = DateTimeFormatter.ISO_DATE_TIME;
+      try {
+        dateTime.parse(expiresAt);
+      } catch (Exception e) {
+        violations.add("The expiration date must be in ISO-8601 format");
+      }
```
tmetzke, line 40: 🔧 There already is a `RequestValidator::validateDate` method you can reuse here.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
```diff
@@ -0,0 +1,115 @@
...
+      final DocumentServices.DocumentCreateRequest request, final MultipartFile file, final DocumentMetadata metadata) {
+    final InputStream inputStream;
+    try {
+      inputStream = file.getInputStream();
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to read document content", e);
+    }
+    return documentServices.createDocument(request, inputStream)
+        .thenApply(resp -> toDocumentReference(resp, metadata))
+        .thenApply(reference -> new ResponseEntity<>(reference, HttpStatus.CREATED));
```
tmetzke, line 69: 🔧 I am not 100% sure but I think instead of using `thenApply` you should be able to reuse `RequestMapper::executeServiceMethod` like the `MessageController` does, for example. It takes a method that returns a `CompletableFuture<A>` and a function that maps from `A` to your response type, i.e. `ResponseEntity<Object>` and already handles exceptions coming from that service method (which you are not handling explicitly here yet, for example).

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
```diff
@@ -0,0 +1,115 @@
...
+
+  private final DocumentServices documentServices;
+
+  @Autowired
+  public DocumentController(final DocumentServices documentServices) {
+    this.documentServices = documentServices;
+  }
+
+  @PostMapping
+  public CompletableFuture<ResponseEntity<DocumentReference>> createDocument(
```
tmetzke, line 46: 🔧 We consistently return `CompletableFuture<ResponseEntity<Object>>` from controllers because we cannot know if it will be the success type or a `ProblemDetail` in the `ResponseEntity`. Please also have a look at defining the accepted and returned content types of your endpoints in this controller.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
```diff
@@ -0,0 +1,115 @@
...
+      metadata = new DocumentMetadata();
+    }
+    if (metadata.getFileName() == null) {
+      metadata.setFileName(file.getOriginalFilename());
+    }
+    if (metadata.getContentType() == null) {
+      metadata.setContentType(file.getContentType());
+    }
+    return metadata;
+  }
```
tmetzke, line 107: 🔧 This can go into the method you created in the `RequestMapper`, I think.

tmetzke (dismissed):
👍 Great stuff already and nice improvements as well 🚀 
🔧 I added a couple of suggestions on top of what was already mentioned, but nothing major really.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
```diff
@@ -0,0 +1,115 @@
...
+    final InputStream inputStream;
+    try {
+      inputStream = file.getInputStream();
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to read document content", e);
+    }
+    return documentServices.createDocument(request, inputStream)
+        .thenApply(resp -> toDocumentReference(resp, metadata))
+        .thenApply(reference -> new ResponseEntity<>(reference, HttpStatus.CREATED));
+  }
```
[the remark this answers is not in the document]
tmetzke: Yep, I agree, that part could move into the services 👍

service/src/main/java/io/camunda/service/DocumentServices.java
tmetzke: Not sure, actually. All services we've implemented so far either return the success type or fail with an exception. Callers can handle this accordingly. An `Either` would not be false per se but makes the service API too complex. A service method is fine to throw an exception since it shouldn't be the "user-facing" method, it's used by something like a REST gateway that decides how to handle the exception.

All data needed by consumers can be provided in the exception, I would say. If there is more than a message to convey, create a specific exception with additional attributes that can be formatted by the consumer accordingly. Just my 2 cents, though.

sbuettner: Ok, I thought we might want to make use of a nice and typesafe error hierarchy using `sealed` classes or interfaces. The nice thing is that this would allow clients (the web-layer) to map each error to the right response in a fully typesafe way. Introducing a new error requires the client to handle it and map it properly.

tmetzke: I see what you mean 👍 I am not completely opposed to that. Would be nice to handle this in a follow-up issue (to keep the focus here). We can explore that issue and create a PR with a proposal on how it could look. I'd like to have multiple eyes on such a decision from different teams since it affects multiple teams in how they write their services API.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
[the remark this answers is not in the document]
chillleader: I couldn't reuse the OpenAPI models directly or return them directly from the service since the service layer doesn't depend on the generated classes. But I created internal lookalike models and now the service consumes all information in one request and also produces the complete set of data required to convert it into an external reference. I agree that it's a good idea, and this will potentially allow services to add more metadata during document creation and return it to the client

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
chillleader: Right, let's drop this 👍

service/src/main/java/io/camunda/service/DocumentServices.java
```diff
@@ -0,0 +1,108 @@
...
+      DocumentCreateRequest request) {
+    validateStoreId(request.storeId);
+    final String id = request.documentId != null ? request.documentId : UUID.randomUUID().toString();
+    final var contentInputStream = request.contentInputStream;
+    final byte[] content;
+    try {
+      content = contentInputStream.readAllBytes();
+      contentInputStream.close();
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to read document content", e);
```
tmetzke, line 59: 🔧 We have a `CamundaServiceException` you can use, if you like. Not a requirement, it won't add a lot of benefit for now here, but we might extend this in the future regarding automated handling of this type of exception or subtypes.

tmetzke: But it could also be an `IllegalArgumentException`, I guess, since the user input is apparently off.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -771,6 +771,197 @@ paths:
...
+          required: false
+          description: The ID of the document store to upload the document to.
+          schema:
+            type: string
+        - name: documentId
+          in: query
+          required: false
+          description: >
+            The ID of the document to upload. If not provided, a new ID will be generated.
+            Specifying an existing ID will result in an error if the document already exists.
```
tmetzke, line 798: 🙃 We don't validate this currently, by the way. Probably not an issue and on purpose, just noticed it 😅

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -173,6 +183,25 @@ public static Either<ProblemDetail, UpdateJobRequest> toJobUpdateRequest(
...
+      final String documentId, final String storeId, final MultipartFile file,
+      final DocumentMetadata metadata) {
+    final var validationResponse = validateDocumentMetadata(metadata);
+    final var internalMetadata = toInternalDocumentMetadata(metadata, file);
+    final InputStream inputStream;
+    try {
+      inputStream = file.getInputStream();
+    } catch (IOException e) {
+      // should be handled globally as an internal server error
+      throw new RuntimeException("Failed to read document content", e);
```
tmetzke, line 196: 🔧 I would recommend handling it here explicitly instead of having it handled globally. You could either
1. Create another violation and collect it together with the ones from `validationResponse` by returning a list of violations there instead of a `ProblemDetail`.
2. Handle this with precedence and return an HTTP 500 `ProblemDetail` that "overwrites" the `validationResponse`.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
```diff
@@ -0,0 +1,99 @@
...
+            .withAuthentication(RequestMapper.getAuthentication())
+            .deleteDocument(documentId, storeId));
+  }
+
+  @PostMapping("/{documentId}/links")
+  public ResponseEntity<Void> createDocumentLink(
+      @PathVariable String documentId, @RequestBody DocumentLinkRequest linkRequest) {
+
+    // TODO: implement
+    throw new UnsupportedOperationException("Not yet implemented");
```
tmetzke, line 95: 🔧 How about we return an [HTTP 501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) `ProblemDetail` here?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -15,14 +17,18 @@
 import static io.camunda.zeebe.gateway.rest.validator.MultiTenancyValidator.validateTenantId;
 import static io.camunda.zeebe.gateway.rest.validator.UserTaskRequestValidator.validateAssignmentRequest;
 import static io.camunda.zeebe.gateway.rest.validator.UserTaskRequestValidator.validateUpdateRequest;
+import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
```
tmetzke, line 20: 🔧 Do we use this here?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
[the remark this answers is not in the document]
tmetzke: This is the way! 👨‍🚀

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
```diff
@@ -0,0 +1,99 @@
...
+public class DocumentController {
+
+  private final DocumentServices documentServices;
+
+  @Autowired
+  public DocumentController(final DocumentServices documentServices) {
+    this.documentServices = documentServices;
+  }
+
+  @PostMapping
```
tmetzke, line 45: 🔧 Can we add `produces` and `consumes` attributes to the mapping annotations (also in the other endpoints) like the other controllers do?

tmetzke (commented):
👍 Great improvements 🚀 
🔧 I added some minor suggestions here and there to make the controller consistent with the others.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
chillleader: Right, not a big deal with the current mock implementation, but I added it anyway

tmetzke:
@chillleader, general PR handling recommendation: instead of amending the initial commit, add additional `refactor` commits on top. This makes reviewing a lot easier since one can see what changed over time. You can either leave the commits and merge in the end or squash them if you feel like it.

Generally, having smaller commits with meaningful messages is more helpful than one big commit. Having multiple commits within one PR is totally fine and even encouraged.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
chillleader: I did this for upload and delete methods, but couldn't make it work for the get method. The issue is, if we want to return a `StreamingResponseBody`, we need to define it explicitly as a method return type, and not use a generic type, otherwise the response is not handled correctly.
We might work around this by implementing a custom [`HandlerMethodReturnValueHandler`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/method/support/HandlerMethodReturnValueHandler.html), but I think a nicer way is to define an exception handler for this method (see `handleDocumentContentException`), and return specific types from the method and the exception handler instead of returning Object from the method.

We can also think about defining a custom exception for this method to further limit the exception handler scope (currently it would handle methods of this controller) - but other methods are not supposed to throw an unhandled `CamundaServiceException` anyway. WDYT?

service/src/main/java/io/camunda/service/CamundaServices.java
```diff
@@ -65,6 +66,11 @@ public MessageServices messageServices() {
     return new MessageServices(brokerClient, searchClient, transformers, authentication);
   }
 
+  public DocumentServices documentServices() {
+    return new DocumentServices(
+        brokerClient, searchClient, transformers, authentication, new HashMap<>());
```
tmetzke, line 71: ❓ What is the benefit of initializing the documents with a `HashMap` here? I would argue that the document services should take care of this. Here in the config, we don't know how the documents are stored/handled/cached and shouldn't need to know. Either the document services do this themselves or inject a `documentStore` (which currently is implemented using a simple in-memory map).

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
tmetzke: Hmm, I am not too familiar with the streaming response, to be honest. What does 

>otherwise the response is not handled correctly

explicitly result in? How is it handled when the response is a `ResponseEntity<Object>`? It is not streamed back but provided in one chunk?

AFAIK, the generic is removed by Java at runtime anyway and Spring simply inspects the actual return type/body of the response and handles it accordingly.

I'd need more time to investigate options here, to be honest. I don't have that time currently and I'm on FTO for 2 weeks starting this afternoon 😅 We can also get @nicpuppa and/or @koevskinikola involved next week here to help you with that.

Regarding
> We can also think about defining a custom exception for this method to further limit the exception handler scope (currently it would handle methods of this controller) - but other methods are not supposed to throw an unhandled CamundaServiceException anyway. WDYT?

The exception handler inside the controller is not a good idea, I would say. Currently, no other controller is supposed to simply (re-)throw a `CamundaServiceException`, yes. But that might happen by accident and might result in a weird exception handling then which is not easy to track down. 

If we can't make it easily work with the streaming response, I'd rather use a specific (new) exception here at the moment so we don't introduce side effects in other controllers. You can also add a method for the exception in the `GlobalControllerExceptionHandler`.

tmetzke: Maybe using `CompletableFuture<ResponseEntity<Object>>` can help here but we'd need to see how to get the content streaming result in there. I guess it's possible but I really don't know. Just food for thought.

chillleader: > How is it handled when the response is a ResponseEntity Object? It is not streamed back but provided in one chunk?


It serializes the StreamingResponseBody object as if it was a POJO which results in an empty JSON object 😄 
I tried specifying the body content-type of the response entity wrapper as `application/octet-stream` but then it failed with an error (something like `no converter found for type <lambda> and application/octet-strteam`) which made me assume the handler is selected before the method invocation. Wrapping into a CompletableFuture didn't make a difference.

I'll continue the investigation and maybe I'll find an elegant solution, but if not - I also agree with your suggestion to introduce a specific exception for this method 👍

tmetzke: I see, OK. Thanks for looking into it, @chillleader 🙏

chillleader: Looks like it actually checks the method return value and inspects the generic type: [StreamingResponseBodyReturnValueHandler](https://github.com/spring-projects/spring-framework/blob/2eff5cb46318eab8e1ed9a7abe27428bc6fb1477/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/StreamingResponseBodyReturnValueHandler.java#L54)

tmetzke: Uh, interesting. Well, if we're going for this result type, a specific exception that is handled accordingly then might be the best way. The global handling should still be generic and only use the information from the thrown exception, if possible.

tmetzke: The global exception handler could also simply be improved a bit, for example doing something like the following (just pseudo code, will not work as is in the global handler):

```java
    if (e instanceof IllegalArgumentException) {
      final ProblemDetail problemDetail =
          ProblemDetail.forStatusAndDetail(HttpStatus.BAD_REQUEST, e.getMessage());
      problemDetail.setInstance(URI.create(requestUri));
      return ResponseEntity.of(problemDetail).build();
    }

    final ProblemDetail problemDetail =
        ProblemDetail.forStatusAndDetail(HttpStatus.INTERNAL_SERVER_ERROR, e.getMessage());
    problemDetail.setInstance(URI.create(requestUri));
    return ResponseEntity.of(problemDetail).build();
```

service/src/main/java/io/camunda/service/CamundaServices.java
chillleader: Right, no benefit here - I concealed this inside the implementation

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DocumentController.java
chillleader: It seems we can't modify this behavior unless we set up a custom `ResponseValueHandler`. I added a custom exception and kept the exception handler inside this controller - arguably, it is more readable to have it in the same class as it doesn't impact other controllers. `StreamingResponseBody` seems to be the way to go if we want to return potentially large content in a non-blocking way, so I'd stick to this right now.

Happy to adjust in the future if we find a better alternative

tmetzke: Thanks, @chillleader, sounds good, I'll have a look now.

tmetzke (commented):
👍 Great improvements, nice solutions for complex problems 🚀 
❌ The only thing we should definitely add here is tests. There are none so far. We should cover input tests for the REST controller (refer to existing controller tests). We can also add some services tests, validating the base behavior of the document services.

service/src/main/java/io/camunda/service/DocumentServices.java
sbuettner: I used this pattern in the past (using kotlin sealed classes) and it was a joy to use. As suggested lets follow-up in a separate conversation.

chillleader:
I added some tests for the REST controller, haven't covered the negative scenarios and the service layer yet as I'm planning to adjust the error handling approach in the next PRs.

sbuettner (approved):

PR #21087, opened 2024-08-13
feat: add new (dummy) camunda-exporter module

### LINKED ISSUE #20951: Create `camunda-exporter` module with a dummy exporter
Create camunda-exporter module with a dummy exporter
- [x] Add new module, setup the dependencies
- [x] Broker can export to the dummy exporter if it is configured

mustafadagher:
## Description

Created `Camunda-Exporter` module with a dummy/Noop Exporter implementation

## Related issues

closes #20951


zeebe/exporters/camunda-exporter/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-exporter-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-protocol</artifactId>
+    </dependency>
+
+    <dependency>
```
deepthidevaki, line 32: :x: Since most of these dependencies are not used now, dependency analysis fails. Let's remove these unused dependencies and add them when needed.

zeebe/exporters/camunda-exporter/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+  <parent>
+    <groupId>io.camunda</groupId>
+    <artifactId>zeebe-parent</artifactId>
+    <version>8.6.0-SNAPSHOT</version>
+    <relativePath>../../../parent/pom.xml</relativePath>
+  </parent>
+
+  <artifactId>camunda-exporter</artifactId>
+
+  <name>Camunda Apps Elasticsearch Exporter</name>
```
deepthidevaki, line 20: :wrench:  Since we will most likely have one exporter that will work with both Elasticsearch and Opensearch, may be we can remove ElasticSearch from the name.

deepthidevaki (approved):
Thanks @mustafadagher . You can merge it after fixing the build error due to unused dependencies.

PR #21088, opened 2024-08-13
feat: support user updated intent

### LINKED ISSUE #21047: Apply `User.UPDATED` event
## Description

- [x] Extend `UserIntent` with an `UPDATED` intent
- [x] Add an `updateUser` method to the `UserState`
- [x] Create and register a new `EventApplier`
    - [x] Using the data from the `UserRecord` apply the update to the `UserState`
- [x] Write a test for the new method in the state

Ben-Sheppard:
## Description

This PR adds in the updated user intent support.

## Related issues

closes #21047


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
```diff
@@ -72,4 +72,13 @@ public UserRecord getUser(final DirectBuffer username) {
   public UserRecord getUser(final String username) {
     return getUser(wrapString(username));
   }
+
+  @Override
+  public void updateUser(final UserRecord user) {
+    username.wrapBuffer(user.getUsernameBuffer());
+    final var key = userKeyByUsernameColumnFamily.get(username);
+    persistedUser.setUser(user);
```
Ben-Sheppard, line 80: Here I've set the complete user object, should this be handled in a different way @remcowesterhoud if we don't want a field updated i.e. `username`? I've seen in other places that we could potentially add another method in like `.setUserWithoutUsername` but I'm curious to hear your thoughts

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/user/UserStateTest.java
```diff
@@ -109,4 +109,29 @@ void shouldReturnCorrectUserByUsername() {
     assertThat(persistedUserOne.getUsername()).isEqualTo(usernameOne);
     assertThat(persistedUserTwo.getUsername()).isEqualTo(usernameTwo);
   }
+
+  @DisplayName("should update a user")
+  @Test
+  void shouldUpdateAUser() {
```
remcowesterhoud, line 115: 🔧  Let's test all the fields that can get updated to be complete.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/user/DbUserState.java
remcowesterhoud: I think it's fine in this case. We retrieve the persistedUser using the username. so there's no way that it could be different and accidentally updated to a different username.

remcowesterhoud (approved):
Nice work @Ben-Sheppard. I have just 1 minor suggestion about the testcase. I'll pre-approve it, feel free to merge after you've fixed it.

You might have to rebase with `main` to fix the CI.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/user/UserStateTest.java
Ben-Sheppard: Good catch, I've changed

Issue #21051, opened 2024-08-12
Stream platform time is controllable


lenaschoenburg:
To support a [clock control API](https://github.com/camunda/product-hub/issues/2379), we will extend the stream platform.
The idea is that the engine uses and controls a partition-scoped clock that is then used to assign record timestamps, trigger scheduled tasks, and generally drive all business logic.

### Acceptance Criteria
- [x] All records are written with a controllable timestamp
- [x] Engine only uses controllable clock, no other sources of time
- [x] Scheduled tasks are executed according to the controllable clock
- [x] Controllable clock is only used within one partition

```[tasklist]
### Implementation
- [x] `RecordProcessorContext` contains a controllable clock
- [x] `Engine` uses controllable clock instead of `ActorClock`
- [x] `Engine` can modify the controllable clock by pinning or applying an offset
- [x] `Sequencer` uses controllable clock to assign record timestamps
- [x] `ProcessingScheduleService` uses controllable clock
- [x] `MetricsExporter` uses controllable clock
- [ ] `ProcessingStateMachine` uses controllable clock for latency metrics
```

npepinpe:
Do we need to pin the clock, or is applying an offset sufficient? Just wondering.

lenaschoenburg:
Based on the [feedback from consulting](https://camunda.slack.com/archives/C037RS2JHB8/p1723024208080989?thread_ts=1723023003.670619&cid=C037RS2JHB8), they might only use `pin`, not offset. That's because they want exact control over the timestamps.

npepinpe:
Let's clarify this, since that's less stuff for us to do. I assume we would still need a `reset` intent or something, though.

npepinpe:
[See this issue for the persistence/synchronization part](https://github.com/camunda/camunda/issues/21068)

PR #21090, opened 2024-08-13
feat: trigger scheduled tasks according to controllable clock

lenaschoenburg:
The implementation is admittedly a bit crude but it works and I hope that it's good enough.
The idea is that `ProcessingScheduleServiceImpl` does not "forward" scheduled tasks to the underlying actor scheduler where we don't have control over the clock, but that it schedules them internally and only has the actor scheduler periodically call to check if any timers are now expired. That way, we have control over the clock we use to decide if a scheduled tasks is expired or not.

Right now the service is called every second. If a task is scheduled for less than half of that, we schedule an additional check a bit earlier, right when the task is supposed to expire. This is purely an optimization, mostly done so that we don't have to adjust a lot of tests that assume that something scheduled with zero delay will trigger more or less immediately.

Relates to https://github.com/camunda/camunda/issues/21051

### DIFF
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
@@ -18,6 +18,7 @@ import io.camunda.zeebe.stream.api.scheduling.SimpleProcessingScheduleService;
 import io.camunda.zeebe.stream.api.scheduling.Task;
 import io.camunda.zeebe.stream.impl.StreamProcessor.Phase;
 import java.time.Duration;
+import java.time.InstantSource;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
 import org.slf4j.Logger;
@@ -35,6 +36,7 @@ public class ProcessingScheduleServiceImpl
   private final BooleanSupplier abortCondition;
   private final Supplier<LogStreamWriter> writerSupplier;
   private final StageableScheduledCommandCache commandCache;
+  private final InstantSource clock;
   private LogStreamWriter logStreamWriter;
   private ActorControl actorControl;
   private AbortableRetryStrategy writeRetryStrategy;
@@ -44,11 +46,12 @@ public class ProcessingScheduleServiceImpl
       final Supplier<Phase> streamProcessorPhaseSupplier,
       final BooleanSupplier abortCondition,
       final Supplier<LogStreamWriter> writerSupplier,
-      final StageableScheduledCommandCache commandCache) {
+      final StageableScheduledCommandCache commandCache, final InstantSource clock) {
     this.streamProcessorPhaseSupplier = streamProcessorPhaseSupplier;
     this.abortCondition = abortCondition;
     this.writerSupplier = writerSupplier;
     this.commandCache = commandCache;
+    this.clock = clock;
   }
 
   @Override
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
@@ -19,6 +19,7 @@ import io.camunda.zeebe.stream.api.scheduling.Task;
 import io.camunda.zeebe.stream.impl.StreamProcessor.Phase;
 import java.time.Duration;
 import java.time.InstantSource;
+import java.util.PriorityQueue;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
```diff
@@ -31,10 +33,14 @@ public class ProcessingScheduleServiceImpl
...
   private static final ScheduledTask NOOP_SCHEDULED_TASK = () -> {};
   private static final Logger LOG = Loggers.STREAM_PROCESSING;
+  private static final int PERIODIC_CHECK_INTERVAL_MS = 1000;
+
   private final Supplier<StreamProcessor.Phase> streamProcessorPhaseSupplier;
   private final BooleanSupplier abortCondition;
   private final Supplier<LogStreamWriter> writerSupplier;
   private final StageableScheduledCommandCache commandCache;
+  private final InstantSource clock;
+  private final PriorityQueue<ScheduledTaskImpl> scheduledTasks = new PriorityQueue<>();
```
npepinpe, line 43: :question: Is this unbounded? If it is, can we either:

1. Bound it. This means somehow handling issues when it _is_ bounded...
2. Have some visibility on its size, e.g. a metric or something. This could help speed up memory issue diagnosis in the future.

npepinpe: If we think memory could be an issue, we could try and compare quickly a full `DeadlineTimerWheel` with 1 million entries memory size vs this priority queue. There are some easy utilities that tell you the size of an object (e.g. https://github.com/openjdk/jol/blob/master/README.md)

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
```diff
@@ -180,4 +209,25 @@ Runnable toRunnable(final Task task) {
           });
     };
   }
+
+  /** Note: this class has a natural ordering that is inconsistent with equals. */
```
npepinpe, line 213: :heart:

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
```diff
@@ -180,4 +209,25 @@ Runnable toRunnable(final Task task) {
...
+  private final class ScheduledTaskImpl implements ScheduledTask, Comparable<ScheduledTaskImpl> {
```
npepinpe, line 214: :question: Would making it a record reduce its memory footprint in the priority queue? I know the actual memory difference is generally small, but may be more at scale - and I believe (I'd have to look up which Java quark went into it) that the GC handles records more efficiently, so it might be a better choice overall for storing potentially many of them in an unbounded collection.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
```diff
@@ -106,6 +112,8 @@ public ActorFuture<Void> open(final ActorControl control) {
     logStreamWriter = writerSupplier.get();
     actorControl = control;
     openFuture.complete(null);
+    actorControl.runAtFixedRate(
+        Duration.ofMillis(PERIODIC_CHECK_INTERVAL_MS), this::processScheduledTasks);
```
npepinpe, line 116: :question: Will you make this configurable in a follow up PR as discussed?

npepinpe (approved):
:rocket: Please address comments, but nothing that requires a second review I think :)

My main concern as usual is the unbounded queue :sweat_smile:

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
lenaschoenburg: Bounding it would be difficult because nothing is set up to handle failure cases. Just as an example, what would it even mean if a `DueDateChecker` cannot schedule next execution?

For size comparison:
Before this PR, every scheduled timer would end up as an entry in the `DeadlineTimerWheel` (roughly two longs + inevitable overhead) _and_ a `TimerSubscription` on the actor task. Both implementations of `TimerSubscription` require 46 bytes according to JOL.

After this PR, every scheduled timer is a `ScheduledTaskImpl` in the priority queue, each taking 24 bytes according to JOL.

Overall this is a reduction in memory usage, with the same danger of unbounded growth as before.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
lenaschoenburg: I can't find a source for this. The only benefit I'm aware of is that the JVM trusts that record fields are immutable (unlike regular object fields) and can apply more optimizations that way.

I've deliberately chosen not to use a record because it's not just holding state but allows cancellation too.


I'm happy to change it to a record though if you can find a source for memory/gc benefits :)

lenaschoenburg:
I'll add metrics and the option to configure the interval in a follow-up PR :+1:

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceImpl.java
lenaschoenburg: To be fair, IntelliJ warned me about this :D

Issue #21068, opened 2024-08-12
Stream platform clock modifications are consistent


npepinpe:
**Description**

As part of https://github.com/camunda/camunda/issues/21051, we need to make sure that modifications to the stream platform clock are replicated and persisted, such that the clock modified equally across all partitions, and it is persisted across leader elections.

This means that we will:

- [x] [Introduce a new state column family for the clock](https://github.com/camunda/camunda/pull/21096)
- [x] [Introduce a new record `ValueType` to manipulate the clock](https://github.com/camunda/camunda/pull/21103)
- [x] [Introduce a new intent to pin the clock at a specified timestamp](https://github.com/camunda/camunda/pull/21106)
- ~[ ] Introduce a new intent to add offset to the current clock~
- [x] [Introduce a new intent to reset any modifications done to the current clock](https://github.com/camunda/camunda/pull/21157)

And the following properties must hold:

- [x] Any modifications done to the clock are persisted across engine restarts
- [x] Commands are broadcasted such that all clocks are modified or none are

npepinpe:
Note: before starting, verify which intents are truly needed (we possibly do not need one to add an offset).

npepinpe:
- Clocks don't need to be modified per partition, it's fine that they're all the same. 
- We will omit the offset intent for now. It should be easy enough to extend, and it may not be required by the stakeholders.

PR #21096, opened 2024-08-13
Introduce clock state to persist clock modifications across restarts

npepinpe:
## Description

This PR introduces a new engine state and column family that will allow persisting modifications to the state.

The modifications are serialized using a single data class, `DbClockModification`, which basically encodes the type of the modification and then the associated `ClockRecord`. This is required because we need to know the type of the modification when initializing the engine.

I would have liked a test to ensure it's truly persisted in the engine, but I couldn't find the right test infrastructure for such things. There are serialization tests, and persistence tests will eventually be added as engine tests once we have the processing (e.g. additions to the existing replay tests).

The state currently is only used when initializing the engine to apply any existing modification to the clock. For now, this is always none. There's also no test, as there is no real infrastructure for an `Engine` test - though if there is and I missed it, please let me know! I suspect this can be covered in the future with more complete engine tests anyway.

## Related issues

related to #21068

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockModification.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockModification.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.state.clock;
+
+import io.camunda.zeebe.db.DbValue;
+import io.camunda.zeebe.msgpack.UnpackedObject;
+import io.camunda.zeebe.msgpack.property.EnumProperty;
+import io.camunda.zeebe.msgpack.property.ObjectProperty;
+import io.camunda.zeebe.protocol.impl.record.value.clock.ClockRecord;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock.Modification;
+import java.time.Duration;
+import java.time.Instant;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class DbClockModification extends UnpackedObject implements DbValue {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DbClockModification.class);
+
+  private final EnumProperty<ModificationType> typeProperty =
+      new EnumProperty<>("type", ModificationType.class, ModificationType.NONE);
+  private final ObjectProperty<ClockRecord> modificationProperty =
+      new ObjectProperty<>("modification", new ClockRecord());
+
+  DbClockModification() {
+    super(2);
+    declareProperty(typeProperty).declareProperty(modificationProperty);
+  }
+
+  DbClockModification pinAt(final long pinnedEpoch) {
+    reset();
+    typeProperty.setValue(ModificationType.PIN);
+    modificationProperty.getValue().pinAt(pinnedEpoch);
+    return this;
+  }
+
+  DbClockModification offsetBy(final long offsetMillis) {
+    reset();
+    typeProperty.setValue(ModificationType.OFFSET);
+    modificationProperty.getValue().offsetBy(offsetMillis);
+    return this;
+  }
+
+  Modification modification() {
+    return switch (typeProperty.getValue()) {
+      case NONE -> Modification.none();
+      case PIN -> readPin();
+      case OFFSET -> readOffset();
+    };
+  }
+
+  private Modification readPin() {
+    final ClockRecord modification = modificationProperty.getValue();
+    if (!modificationProperty.isSet() || !modification.hasPinnedEpoch()) {
+      LOGGER.warn("Deserialized a PIN clock modification, but there is no pinned epoch");
+      return Modification.none();
+    }
+
+    return Modification.pinAt(Instant.ofEpochMilli(modification.getPinnedAtEpoch()));
+  }
+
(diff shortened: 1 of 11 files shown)


npepinpe:
@lenaschoenburg - please review https://github.com/camunda/camunda/pull/21103 first

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockModification.java
```diff
@@ -0,0 +1,82 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class DbClockModification extends UnpackedObject implements DbValue {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DbClockModification.class);
+
+  private final EnumProperty<ModificationType> typeProperty =
+      new EnumProperty<>("type", ModificationType.class, ModificationType.NONE);
+  private final ObjectProperty<ClockRecord> modificationProperty =
+      new ObjectProperty<>("modification", new ClockRecord());
```
lenaschoenburg, line 27: :upside_down_face: Could also just be a `LongProperty`, that way you don't use a record value in the state

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockState.java
```diff
@@ -0,0 +1,61 @@
...
+import io.camunda.zeebe.db.ColumnFamily;
+import io.camunda.zeebe.db.TransactionContext;
+import io.camunda.zeebe.db.ZeebeDb;
+import io.camunda.zeebe.db.impl.DbString;
+import io.camunda.zeebe.engine.state.mutable.MutableClockState;
+import io.camunda.zeebe.protocol.ZbColumnFamilies;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock.Modification;
+
+public class DbClockState implements MutableClockState {
+  private static final DbString KEY = new DbString("MODIFICATION");
```
lenaschoenburg, line 19: :wrench: while it's _probably_ fine for a `DbString`, I'd caution against static fields here. In the past we've had issues where writing such a static value would surprisingly mutate internal state, causing race conditions across partitions.

Maybe just do it how it's done in `DbCheckpointState`.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockState.java
```diff
@@ -0,0 +1,61 @@
...
+    value.offsetBy(offsetMillis);
+    columnFamily.upsert(KEY, value);
+    return this;
+  }
+
+  @Override
+  public MutableClockState reset() {
+    value.reset();
+    return this;
+  }
```
lenaschoenburg, line 49: :x: This needs to upsert too, right?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/clock/DbClockStateTest.java
```diff
@@ -0,0 +1,54 @@
...
+import io.camunda.zeebe.engine.state.mutable.MutableProcessingState;
+import io.camunda.zeebe.engine.util.ProcessingStateExtension;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock.Modification;
+import java.time.Duration;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(ProcessingStateExtension.class)
+final class DbClockStateTest {
```
lenaschoenburg, line 22: :wrench: Add a test for `pin`
:wrench: Add a test for `reset`

lenaschoenburg (approved):
Nice test coverage :)

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockModification.java
npepinpe: Now that we merged both properties in the previous PR, yes :smile:

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockState.java
npepinpe: Wait, if I _only_ use it in an upsert, it might be mutated? :exploding_head:

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockState.java
npepinpe: Good catch, yes!

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockState.java
lenaschoenburg: It looks like `DbString` is safe but messagepack values aren't, their `writer` is mutating internal state while writing. Writing may occur on `upsert` because RocksDB needs the raw bytes so we write.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/clock/DbClockStateTest.java
lenaschoenburg: I think they still provide a bit of value, for example they should have caught the missing upsert on reset.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/clock/DbClockState.java
npepinpe: I could make an immutable type specifically for this, but I think that's only useful if there were plans to use it in other places, and while I think it's probably useful, I can't tell for sure :sweat_smile:

PR #15783, opened 2024-01-04
Intermediate signal throw should not send response

ana-vinogradova-camunda:
## Description
Fix for this [bug](https://github.com/camunda/zeebe/issues/15649)

- Added a check to verify if the command has request metadata
- Added a unit test to verify that response is not send if the metadata is not present. Adjusted the test case from the bug above.
- Added an Integration Test to verify the response is sent when the metadata is present

## Related issues

closes #15649

## Definition of Done

_Not all items need to be done depending on the issue and the pull request._

Code changes:
* [ ] The changes are backwards compatibility with previous versions
* [ ] If it fixes a bug then PRs are created to [backport](https://github.com/camunda/zeebe/compare/stable/0.24...main?expand=1&template=backport_template.md&title=[Backport%200.24]) the fix to the last two minor versions. You can trigger a backport by assigning labels (e.g. `backport stable/1.3`) to the PR, in case that fails you need to create backports manually.

Testing:
* [ ] There are unit/integration tests that verify all acceptance criterias of the issue
* [ ] New tests are written to ensure backwards compatibility with further versions
* [ ] The behavior is tested manually
* [ ] The change has been verified by a QA run
* [ ] The impact of the changes is verified by a benchmark

Documentation:
* [ ] The documentation is updated (e.g. BPMN reference, configuration, examples, get-started guides, etc.)
* [ ] If the PR changes how BPMN processes are validated (e.g. support new BPMN element) then the Camunda modeling team should be informed to adjust the BPMN linting.

Other teams:
If the change impacts another team an issue has been created for this team, explaining what they need to do to support this change.
- [ ] [Operate](https://github.com/camunda/operate/issues)
- [ ] [Tasklist](https://github.com/camunda/tasklist/issues)
- [ ] [Web Modeler](https://github.com/camunda/web-modeler/issues)
- [ ] [Desktop Modeler](https://github.com/camunda/camunda-modeler/issues)
- [ ] [Optimize](https://github.com/camunda/camunda-optimize/issues)

Please refer to our [review guidelines](https://github.com/camunda/zeebe/wiki/Pull-Requests-and-Code-Reviews#code-review-guidelines).

Issue #15649, opened 2023-12-18
Intermediate signal throw event should not send a response out


npepinpe:
**Describe the bug**

It seems the signal broadcast command is always sending responses out. Possibly other commands do the same. This has little impact right now on Zeebe itself because the transport essentially "swallows" the error, e.g. if you send a response with a request ID and stream ID that it doesn't recognize, it simply does nothing. This is still additional, unnecessary work though, copying the response around, serializing it, etc.

This also causes an error in `zeebe-process-test`, where the transport behaves differently and _always_ expects a request for every response sent out, causing [the following error](https://github.com/camunda/zeebe-process-test/issues/1011).

In my opinion, this should be fixed in both places, although the severity is quite low in Zeebe. But the command response writer should simply not add pending responses if the associated command has no request metadata. For example, see the old `CommandProcessImpl`, which does this check.

**To Reproduce**

Add the following to the engine test `SignalIntermediateThrowEventTest.java`:

```java
  @Test
  public void shouldNotSendResponseWhenPartOfBatch() {
    // given
    final var processId = Strings.newRandomValidBpmnId();
    final var process =
        Bpmn.createExecutableProcess(processId)
            .startEvent()
            .intermediateCatchEvent(
                "start", b -> b.message(m -> m.name("start").zeebeCorrelationKey("=businessKey")))
            .intermediateThrowEvent("signal", b -> b.signal("signal"))
            .endEvent()
            .done();
    ENGINE.deployment().withXmlResource(process).deploy();
    final var piKey =
        ENGINE
            .processInstance()
            .ofBpmnProcessId(processId)
            .withVariables(Map.of("businessKey", "42"))
            .create();
    assertThat(
            RecordingExporter.processInstanceRecords(ProcessInstanceIntent.ELEMENT_ACTIVATED)
                .withProcessInstanceKey(piKey)
                .withElementId("start")
                .limit(1)
                .exists())
        .isTrue();

    // when
    ENGINE.message().withName("start").withCorrelationKey("42").publish();
    assertThat(
            RecordingExporter.signalRecords(SignalIntent.BROADCASTED)
                .withSignalName("signal")
                .limit(1)
                .exists())
        .isTrue();

    // then
    Mockito.verify(ENGINE.getCommandResponseWriter(), Mockito.never())
        .intent(SignalIntent.BROADCASTED)
        .tryWriteResponse(-1, -1);
  }
```

[...]

PR #21106, opened 2024-08-13
Introduce clock PIN command to pin the controllable stream clock

npepinpe:
## Description

This PR introduces a new PIN command processor, as well as all the wiring for the engine to process the command. It does not yet test the distribution, though the code is added. This will be added in a follow up.

Pinning is executed as a side effect to ensure we have persisted the state changes before modifying the clock. I'm not 100% sure this is required, but I figured better safe than sorry.

## Related issues

related to #21068

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockPinProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockPinProcessor.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.clock;
+
+import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.streamprocessor.DistributedTypedRecordProcessor;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.SideEffectWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedResponseWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
+import io.camunda.zeebe.protocol.impl.record.value.clock.ClockRecord;
+import io.camunda.zeebe.protocol.record.intent.ClockIntent;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock;
+import io.camunda.zeebe.stream.api.records.TypedRecord;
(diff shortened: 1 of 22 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockPinProcessor.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public void processNewCommand(final TypedRecord<ClockRecord> command) {
+    final long eventKey = keyGenerator.nextKey();
+    final var clockRecord = command.getValue();
+
+    applyClockModification(eventKey, clockRecord);
+    if (command.hasRequestMetadata()) {
+      responseWriter.writeEventOnCommand(eventKey, ClockIntent.PINNED, clockRecord, command);
+    }
```
lenaschoenburg, line 53: :thought_balloon: This looked a bit odd so I looked into it:
1. It's necessary in case the clock pin command can trigger internally, without a user request. This will most likely not be the case but we could still keep it for future-proofing.
2. The `responseWriter` could do this for us. In fact that's what you [suggested](https://github.com/camunda/camunda/issues/15649#issuecomment-1861566786) when it first came up but [the fix](https://github.com/camunda/camunda/pull/15783) didn't do that, instead adding these manual checks :shrug:

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockProcessors.java
```diff
@@ -0,0 +1,32 @@
...
+
+import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.streamprocessor.TypedRecordProcessors;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.intent.ClockIntent;
+import io.camunda.zeebe.stream.api.StreamClock.ControllableStreamClock;
+import io.camunda.zeebe.stream.api.state.KeyGenerator;
+
+public class ClockProcessors {
```
lenaschoenburg, line 18: :upside_down_face: 
```suggestion
public final class ClockProcessors {
```

lenaschoenburg (approved):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockPinProcessor.java
npepinpe: Yeah, it seems obvious to me, but :shrug: As I'm not so much in the engine anymore, I'll just follow the rules here :)

npepinpe:
@lenaschoenburg - I added tests for the distribution and correct initialization of the engine on restart.

lenaschoenburg (approved):
I reviewed the added tests now, looks good :+1:

npepinpe:
Looks like main is broken due to Optimize spotless changes? :thinking: I mean I didn't change anything, but the merge queue fails.

lenaschoenburg:
I could imagine that rebasing on main might help, that way spotless shouldn't detect that as a "changed" file.

npepinpe:
now my pr itself has the wrong check 😅

PR #21129, opened 2024-08-14
feat: allow to reuse benchmark image

ChrisKujawa:
### Problem/Challenge

Previously when running multiple benchmarks we had to rebuild the image everytime. This was an unnecessary waste of resources and time.

### Proposed solution via PR

This PR adds as functionality a specific parameter, the benchmark tag. That tag should be used for deploying the benchmarks if specified. This allows us to skip the long docker build time.

### DIFF
--- a/.github/workflows/zeebe-benchmark.yml
+++ b/.github/workflows/zeebe-benchmark.yml
@@ -280,8 +280,6 @@ jobs:
     name: Deploy
     needs:
       - calculate-image-tag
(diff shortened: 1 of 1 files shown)


Zelldon:
I was trying this approach, and it is working well. I'm able to set up a benchmark without rebuilding the complete project.

lenaschoenburg (approved):
Thanks! :+1:

Issue #20768, opened 2024-08-05
Refactor `<team>-labeled.yml` workflows to remove duplication


cmur2:
## Description

Problem: 99% redundant workflows (running on same trigger with same logic, just different GH label and GH project id)

* https://github.com/camunda/camunda/blob/main/.github/workflows/operate-labeled.yml
* https://github.com/camunda/camunda/blob/main/.github/workflows/tasklist-labeled.yml
* Optimize is also introducing one in https://github.com/camunda/camunda/pull/20516

Solution:

* composite GH action in this repository having 2 parameters (GH label, GH project id) and assigns issues matching the label to the project if the issue has no project yet
* 1 GH workflow `issues-labeled` calling this composite GH action for each `operate`, `tasklist` (, `optimize` when merged)
```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/21135
- [ ] https://github.com/camunda/camunda/pull/21176
- [ ] https://github.com/camunda/camunda/pull/21179
```

misiekhardcore:
@cmur2 when you decide this can be worked on, you can assign this task to me

cmur2:
> @cmur2 when you decide this can be worked on, you can assign this task to me

There is no dependency to other tasks so feel free to start work on this 👍🏼

misiekhardcore:
Do we reassign the issue to the reviewer when we move it to review?

cmur2:
> Do we reassign the issue to the reviewer when we move it to review?

Yes we can, I have no hard preference.

misiekhardcore:
@cmur2 can this be closed now?

cmur2:
> @cmur2 can this be closed now?

Can you please do a test if its working if you didn't already?

misiekhardcore:
it looks like I used the name of the action I had before changing it later :/ opened #21176 to fix it

cmur2:
Thanks for fixing that, can you do another (final) test to see if it works?

misiekhardcore:
I did and it seems to still be broken. I think I need to add `checkout` action to the workflow now

misiekhardcore:
It is still broken, it cant find the action and it suggests that we need to do checkout, so I opened #21179 to add it

misiekhardcore:
It finally worked https://github.com/camunda/camunda/actions/runs/10402715844/job/28807751709

PR #21135, opened 2024-08-14
ci: abstract adding issues to projects

misiekhardcore:
## Description

This PR is abstracting adding issues to projects

## Checklist

- [ ] for CI changes:

## Related issues

relates to #20768


misiekhardcore:
I'm not sure how to test this.
@cmur2 I have two questions here:
1. will the secrets be passed to that action? I think the answer is yes but better to ask
2. is there any clever way of moving the `if` statement that checks for the matching label to the action instead of keeping it in the workflow?

cmur2:
>     1. Do you have any idea how to test this?

I don't know any good way besides merging to `main` and then creating an issue. Be prepared to revert the change after the merge.

>     2. will the secrets be passed to that action? I think the answer is yes but better to ask

I don't think they will and it might be cleaner anyways to pass them explicitly from the calling workflow to show which secret is actually used. Looking at e.g. https://github.com/camunda/camunda/blob/main/.github/actions/setup-zeebe/action.yml#L25-L33 where they do this way.

>     3. is there any clever way of moving the `if` statement that checks for the matching label to the action instead of keeping it in the workflow?

Not sure but I also don't mind having the `if` statement in the workflow - the refactoring from this PR is a tremendous improvement already 👍🏼

.github/actions/add-issue-to-project/action.yml
```diff
@@ -0,0 +1,49 @@
...
+                }
+              }
+            }
+          }
+        variables: |
+          owner: "camunda"
+          repo: "camunda"
+          issue: ${{ inputs.issueNumber }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GH_PROJECT_AUTOMATION_TOKEN }}
```
cmur2, line 37: Let's add a new input `githubToken` to pass the secret explicitly from the workflow

.github/workflows/issues-labeled.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: Add ${{ matrix.teams.name }} issue to project if relevant component label added
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add labeled issue to project
+        if: ${{ github.event.label.name == matrix.teams.label }}
+        uses: ./.github/actions/add-labeled-to-project
+        with:
+          projectId: ${{ matrix.teams.projectId }}
+          issueNumber: ${{ github.event.issue.number }}
+          projectAutomationToken: ${{ secrets.GH_PROJECT_AUTOMATION_TOKEN }}
```
cmur2, line 25: Input name does not match the name defined in the action

cmur2 (approved):

PR #21138, opened 2024-08-14
feat: logstream uses a controllable clock to assign record timestamps

lenaschoenburg:
When building the logstream, we now create a `ControllableStreamClock`. It is based on the current actor clock for backwards compatibility with the old `ControllableActorClock`.

The newly created clock is part of the `PartitionContext`. It is used by the `Sequencer` to assign record timestamps and by the `StreamProcessor` so that the engine uses the same clock.

Relates to https://github.com/camunda/camunda/issues/21051


zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/TestStreams.java
```diff
@@ -95,10 +97,12 @@ public final class TestStreams {
   public TestStreams(
       final TemporaryFolder dataDirectory,
       final AutoCloseableRule closeables,
-      final ActorScheduler actorScheduler) {
+      final ActorScheduler actorScheduler,
+      final InstantSource clock) {
```
npepinpe, line 101: Hm, this is going to cause conflicts with my PR :see_no_evil: Unfortunately I'm waiting for your review, so I guess you win :smile:

npepinpe (approved):
:rocket:

PR #21139, opened 2024-08-14
Improve latency section in medic dashboard

ChrisKujawa:
## Description

Previously the latency panels just covered normal benchmarks. Instead of the actual latency related benchmarks, which caused confusion as the latency is quite high and unexpected. As the latency benchmark only run one instance per second.

Adding more context (with keeping the normal benchmarks as reference) and adding panels for the actual latency benchmarks allows to better compare and detect issues in latency.

### DIFF
--- a/monitor/grafana/medic-benchmark-dashboard.json
+++ b/monitor/grafana/medic-benchmark-dashboard.json
@@ -755,7 +755,7 @@
         "overrides": []
       },
       "gridPos": {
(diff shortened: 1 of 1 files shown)


npepinpe:
I feel like we have had that discussion before, and we'd already changed that...but I guess not? :thinking:

npepinpe (approved):

PR #21140, opened 2024-08-14
feat: use controllable clock for processing metrics

### LINKED ISSUE #21051: Stream platform time is controllable

lenaschoenburg:
## Description

This ensures that metrics published by the `ProcessingStateMachine` and the `MetricsExporter` are respecting the controllable clock.

To support this in the `MetricsExporter`, the `ExporterContext` was extended to provide an `InstantSource` that all exporters can use.

## Related issues

closes #21051

### DIFF
--- a/zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/context/Context.java
+++ b/zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/context/Context.java
@@ -32,8 +32,11 @@ public interface Context {
   Logger getLogger();
 
   /**
-   * @return a clock that provides the current time. Use this instead of system time to ensure that
-   *     time is controllable.
+   * A clock that provides the current time. Use this instead of system time to ensure that time is
+   * controllable. Especially relevant when comparing the "current" time against record timestamps.
+   *
+   * <p>This clock is not used by the {@link Controller} to trigger scheduled tasks, these still
+   * rely on uncontrollable system time.
    */
   InstantSource clock();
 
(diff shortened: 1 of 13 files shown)


npepinpe:
Does this mean that general exporters will now be able to read the engine clock? :thinking: It might be confusing since our underlying scheduler there will not be using the clock at all, but... :thinking:

zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/context/Context.java
```diff
@@ -30,6 +31,12 @@ public interface Context {
    */
   Logger getLogger();
 
+  /**
+   * @return a clock that provides the current time. Use this instead of system time to ensure that
```
npepinpe, line 35: :question: Do we want to make it explicit that this is not related to the exporter's `Controller` task scheduling capabilities? I think it might cause confusion.

npepinpe (approved):
Looks good, just one question :thinking:

zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/context/Context.java
lenaschoenburg: I'll add something :+1:

PR #21146, opened 2024-08-14
Apply spotless 

korthout:
## Description

Not sure how this got on main, but spotless check failed. Fixed by running `mvnd spotless:apply`.

## Related issues

NA


RomanJRW (approved):

RomanJRW:
@korthout 

Approved but I suspect that this will be the case in a lot of places. We aim to do a clear up and apply the mono repo spotless to the whole Optimize repo shortly, but descoped it from the initial merge-in PR

korthout:
@RomanJRW spotless doesn't require us to apply it everywhere at oncel. That's why we ratchet from `main`, so we only run it against new changes. We should just ensure that PRs can't be merged to `main` when `spotless:check` fails:

https://github.com/camunda/camunda/blob/117f3f3252771eee3d0459f783c95c1fdf40a264/bom/pom.xml#L267-L271

PR #21148, opened 2024-08-14
feat: add users search

### LINKED ISSUE #20405: Camunda services support retrieving users
### Description
To be able to return users stored in the system we need to provide paths to search for users within our storage methods.

### Acceptance criteria
- [x] A query class for users exists, implementing `TypedSearchQuery`
- [x] A filter class for users exists, implementing `FilterBase`
- [x] A transformer for the user filter exists, implementing `FilterTransformer`
- [x] A sort class for users exists, implementing `SortOption`
- [x] The `SearchQueryBuilders` class is updated to prover builder methods for user queries
- [x] The `SearchQueryRequestMapper` class is updated to support user queries
- [x] The `ServiceTransformers` class is updated to support user queries
- [x] An entity class exists representing a user to map ES hits to objects
- [x] A services class for users exists, extending `SearchQueryService` 

### Tests
- TBD

maryarm:
### Description
To be able to return users stored in the system we need to provide paths to search for users within our storage methods.

### Acceptance criteria
- [x] A query class for users exists, implementing `TypedSearchQuery`
- [x] A filter class for users exists, implementing `FilterBase`
- [x] A transformer for the user filter exists, implementing `FilterTransformer`
- [x] A sort class for users exists, implementing `SortOption`
- [x] The `SearchQueryBuilders` class is updated to prover builder methods for user queries
- [x] The `SearchQueryRequestMapper` class is updated to support user queries
- [x] The `ServiceTransformers` class is updated to support user queries
- [x] An entity class exists representing a user to map ES hits to objects
- [x] A services class for users exists, extending `SearchQueryService` 

## Related issues

closes #20405


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/UserController.java
```diff
@@ -46,4 +53,28 @@ public CompletableFuture<ResponseEntity<Object>> createUser(
                     userWithPasswordDto.getEmail(),
                     passwordEncoder.encode(userWithPasswordDto.getPassword())));
   }
+
+  @PostMapping(
```
tmetzke, line 57: ⚠️ If we add a search endpoint to a REST controller annotated with `CamundaRestController`, the endpoint will be available in a cluster by default. 

For search/query endpoints, we currently don't do this, see our [public documentation](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/camunda-api-rest-overview/#query-api) for the reasoning.

For all other search endpoints, we currently create specific controllers annotated with `CamundaRestQueryController`. Those are not enabled in a cluster by default.

It could be fine here, I just want to make you aware of this. If every authenticated user should be allowed to query all users, i.e. there are no resource permissions people can configure for this resource, we can probably keep it as is.

tmetzke:
@nathansandi, you're quite experienced with query controllers now, maybe you can have a look as well 🙂

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/UserController.java
maryarm: Good point @tmetzke, I wasn't aware of this and as I understand we shouldn't allow all authenticate user to query all users. I'm going to move this endpoint to a separate controller.

nathansandi:
I am going to review this today, just an advice for the next endpoints:
- Separate the pull requests between Service, Controller and Client: This would make it easier the review process and avoid

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -240,6 +261,29 @@ private static UserTaskFilter toUserTaskFilter(final UserTaskFilterRequest filte
     return builder.build();
   }
 
+  private static UserFilter toUserFilter(final CamundaUserFilterRequest filter) {
+    final var builder = FilterBuilders.user();
+
+    if (filter != null) {
+      // username
+      if (filter.getUsername() != null) {
```
nathansandi, line 269: 🔧  Minor suggestion, we can use Optionals here in order to refactor  
(There are some endpoints that are using already)
I am intending to refactor this later.

nathansandi (commented):
💯  Great work here, only some minor questions/suggestions
I gonna make one more review round this Tuesday

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1026,8 +1076,44 @@ components:
...
+        - $ref: "#/components/schemas/SearchQueryRequest"
+      type: object
+      properties:
+        filter:
+          allOf:
+            - $ref: "#/components/schemas/CamundaUserFilterRequest"
+    CamundaUserFilterRequest:
+      type: object
+      properties:
+        username:
```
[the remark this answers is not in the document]
maryarm: I don't think that it's useful, as within our ui we don't need to search by id and later we will add an endpoint to fetch a user by id for details.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1217,8 +1267,44 @@ components:
...
-          type: "boolean"
+    UserSearchQueryRequest:
+      allOf:
+        - $ref: "#/components/schemas/SearchQueryRequest"
+      type: object
+      properties:
+        filter:
+          allOf:
+            - $ref: "#/components/schemas/CamundaUserFilterRequest"
+    CamundaUserFilterRequest:
```
nathansandi, line 1278: One more change here
`UserFilterRequest`

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1217,8 +1267,44 @@ components:
...
+      type: object
+      properties:
+        username:
+          type: "string"
+        name:
+          type: "string"
+        email:
+          type: "string"
+    CamundaUserResponse:
```
nathansandi, line 1287: 🔨 One more update: `UserResponse`

nathansandi (changes requested):
2 more minor updates :) and we are ready to go

nathansandi (approved):
LGTM 💯

PR #21150, opened 2024-08-14
feat: add common exporter interfaces

### LINKED ISSUE #20952: Identify and create common exporter interfaces in the new camunda-exporter module
So far the common exporter interfaces identified are
- ExporterHandler
  - The concrete implementations of this interface has the business logic of transforming zeebe records to Elasticsearch entities. Operate and Identity will add the required concrete implementations.
- OperateEntity -> should be renamed to a more generic name since it's going to be used by identity too 
  - Represents the entities that are written to ElasticSearch.

We should also decide if other interfaces should be considered common and should be included

Also, to be decided whether a new module is needed as it is a common interface used by the exporter, Operate and Identity modules

mustafadagher:
## Description

Added common exporter interfaces identified.

## Related issues

closes #20952


deepthidevaki (approved):
:tada:  I'm not yet sure if all methods defined in `BatchRequest` is needed. We will see it further down the line and can revisit them as needed.

PR #21151, opened 2024-08-14
ci: disable Docker to GHA cache in PRs

cmur2:
## Description

In an effort to reduce Github Actions cache usage to safe(r) levels for #18759 (see also [CI Health dashboard](https://dashboard.int.camunda.com/d/bdmo5l8puugaoc/ci-health-camunda-camunda-monorepo) bottom row) I want to turn off caches on PR builds that don't bring much value (e.g. have high miss rate).

Looking e.g. at https://github.com/camunda/camunda/actions/caches?query=branch%3Arefs%2Fpull%2F21063%2Fmerge shows us that several large (>100MB) caches are only "used" during their creation (same timestamps) and built again in later runs:

image

The [Tasklist Docker tests workflow](https://github.com/camunda/camunda/blob/main/.github/workflows/tasklist-docker-tests.yml) is the only one that runs on PRs (and e.g. all Maven dependency update PRs from Renovate) that uses [the experimental cache feature of the docker-push-action](https://github.com/moby/buildkit/?tab=readme-ov-file#github-actions-cache-experimental), others like the comparable [Operate Docker tests workflow](https://github.com/camunda/camunda/blob/main/.github/workflows/operate-docker-tests.yml) don't and work well. [Github search](https://github.com/search?q=repo%3Acamunda%2Fcamunda+cache-to%3A+type%3Dgha&type=code)

## Checklist

- [ ] for CI changes:

## Related issues

Related to #18750

### DIFF
--- a/.github/workflows/tasklist-docker-tests.yml
+++ b/.github/workflows/tasklist-docker-tests.yml
@@ -98,9 +98,6 @@ jobs:
           context: .
           push: true
           tags: ${{ env.TASKLIST_TEST_DOCKER_IMAGE }}
-          cache-from: type=gha,ignore-error=true
-          cache-to: type=gha,mode=max,ignore-error=true
-          no-cache-filters: build,distball,dist
           file: tasklist.Dockerfile
       - name: Run Docker tests
         run: mvn -pl tasklist/qa/integration-tests -DskipChecks -Dtest=StartupIT -Dsurefire.failIfNoSpecifiedTests=false -Dspring.profiles.active=docker-test test


.github/workflows/tasklist-docker-tests.yml
```diff
@@ -98,9 +98,6 @@ jobs:
           context: .
           push: true
           tags: ${{ env.TASKLIST_TEST_DOCKER_IMAGE }}
-          cache-from: type=gha,ignore-error=true
-          cache-to: type=gha,mode=max,ignore-error=true
-          no-cache-filters: build,distball,dist
```
Zelldon, line 103: ❓ So not configuring disables the cache?

ChrisKujawa (approved):
I think the cache on your screenshot, creates a cache for each layer of the docker build process. I guess it is not really useful, as after building we not build again, and in different PRs/builds it will have a different sha hash. So I guess it makes sense to disabled it :)

.github/workflows/tasklist-docker-tests.yml
cmur2: It disables exporting the cache to an external service (S3, GHA, ...) is my understanding from https://github.com/moby/buildkit/?tab=readme-ov-file#cache - Docker will still have its layer cache file structure on disk but not upload that stuff anywhere when the build ends.

cmur2:
With some advanced statistics analysis (linear extrapolation) and a large sample size (one) while ignoring all other outside factors (developer activity etc) I determined that this PR had the desired impact on reducing the cache usage 🤓:

image

Jokes aside I'll keep monitoring the situation over the next days to determine which other measures from #18750 need to be implemented.

PR #21156, opened 2024-08-14
feat: Create /license endpoint to expose Camunda license status

matthewBearCamunda:
## Description

Create a `GET /license` endpoint to return license information

```
GET /license
{
   "validLicense": true,
   "environment": "self-managed"
}
```

The endpoint can be accessed through something like this `http://localhost:8080/v2/license`

- `validLicense` will be `true` or `false`. The value is determined whether the Camunda license is valid or not
- `environment` will return one of two strings, `saas` or `self-managed`. The value reflects the current running mode Camunda is in

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/pull/21156

### DIFF
--- a/service/src/main/java/io/camunda/service/license/CamundaLicense.java
+++ b/service/src/main/java/io/camunda/service/license/CamundaLicense.java
@@ -18,7 +18,7 @@ public class CamundaLicense {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(CamundaLicense.class);
   private boolean isValid;
-  private boolean isSelfManaged;
+  private String licenseType;
   private boolean isInitialized;
 
   @VisibleForTesting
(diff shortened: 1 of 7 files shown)


matthewBearCamunda:
@tmetzke - I'm having issues getting autowiring to work for the failing tests

`Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'io.camunda.service.ManagementServices' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}`

This error does not show when I book up any of the modules in the monorepo, it's seemingly only for testing. I can recreate the error by running `StandaloneGatewayIT` locally

Do you have any ideas as to how I can solve this error?

tmetzke (approved):
👍 Awesome, @matthewBearCamunda, looks really good 🚀 
🔧 I added a small suggestion but nothing blocking

tmetzke:
@matthewBearCamunda, not sure why you removed it from the queue. Regarding the failing checks, they are either flaky tests or flaky environments, as far as I can see. Nothing related to your PR.

matthewBearCamunda:
@tmetzke - I thought of one more test that I wanted to do locally before merging it, so I removed it to make sure that it worked

It did and I'm placing it back in the queue

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -32,39 +32,45 @@ public synchronized boolean isValid() {
...
 
   /**
-   * SaaS mode is determined through the properties of the passed in license.
+   * Camunda licenses will have a license type property, fetch that out of the license and return
+   * the value
    *
    * <p>Self-managed mode is any other possibility. (ex, blank license, prop missing, etc)
    */
-  private boolean determineIfLicenseEnvModeIsSelfManaged(final String licenseStr) {
+  private String getLicenseTypeFromProperty(final String licenseStr) {
```
nicpuppa, line 54: 🔧 We can improve a bit this method, this is a suggestion but there are other ways to do it 😄 

```
try {
    final LicenseKey licenseKey = getLicenseKey(licenseStr);
    String licenseType = licenseKey.getProperties().getOrDefault(LICENSE_TYPE_PROPERTY_NAME, UNKNOWN_LICENSE_TYPE);

    if (UNKNOWN_LICENSE_TYPE.equals(licenseType)) {
        LOGGER.error("Expected a licenseType property on the Camunda License, but none were found.");
    }

    return licenseType;

} catch (InvalidLicenseException e) {
    LOGGER.error("Encountered an invalid license when determining the license type.", e);
} catch (Exception e) {
    LOGGER.error("Unexpected error while determining the license type.", e);
}
return UNKNOWN_LICENSE_TYPE;
```

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -108,7 +107,7 @@ public void shouldReturnProperValidValuesWhenLicenseIsSaas() throws InvalidLicen
     testLicense.initializeWithLicense("whatever");
 
     // then
-    assertFalse(testLicense.isSelfManaged());
+    assertEquals("saas", testLicense.getLicenseType());
```
nicpuppa, line 110: ❓ Can we store those license type values somewhere ? I'd prefer having them in one place instead of spread those string everywhere. Wdyt ? (See also the suggestion above)

matthewBearCamunda:
@nicpuppa - I opened this PR for those changes https://github.com/camunda/camunda/pull/21236

PR #21159, opened 2024-08-14
refactor: remove optimize 7 plugins from mono repo

RomanJRW:
## Description

Removes the whole C7 plugin mechanism from Optimize 8 codebase.

Out of scope:
- Removing the config properties
- Documentation (no forked docs and the plugins are already marked as C7-only)

## Checklist

- [ ] for CI changes:

## Related issues

closes #


optimize/util/optimize-commons/pom.xml
```diff
@@ -215,6 +209,10 @@
       <artifactId>auth</artifactId>
       <version>${awssdk.version}</version>
     </dependency>
+    <dependency>
+      <groupId>jakarta.ws.rs</groupId>
+      <artifactId>jakarta.ws.rs-api</artifactId>
+    </dependency>
```
PHWaechtler, line 215: ❓ Is this intentional?

HeleneW-dot (approved):
Lgtm, small question re. one addition in a pom. Also:
- Can [this file](https://github.com/camunda/camunda/blob/13211-remove-optimize-7-plugins/optimize/backend/src/it/resources/plugin.version) also be removed? 
- is the config removal meant to become part of the other config ticket, or a separate follow up? 
- shall we also add the C7 only banner [here](https://docs.camunda.io/optimize/self-managed/optimize-deployment/plugins/plugin-system/), I think this was not done before

optimize/util/optimize-commons/pom.xml
RomanJRW: Yes, but I could have explained. The commons module declared the plugin one as a dependency and therefore pulled in this jakarta one. With the plugin module deleted, this now needs declaring here explicitly instead

RomanJRW:
> Can [this file](https://github.com/camunda/camunda/blob/13211-remove-optimize-7-plugins/optimize/backend/src/it/resources/plugin.version) also be removed?

Probably yeah, just didn't see it

> is the config removal meant to become part of the other config ticket, or a separate follow up?

I saw there was another check list item to do config, so figured I wouldn't touch it here. At the end of these clean up tasks, is there already an intention to clear up any unused config that belonged to C7?

> shall we also add the C7 only banner [h](https://docs.camunda.io/optimize/self-managed/optimize-deployment/plugins/plugin-system/)

Eventually, this will be removed from the docs. But you are right, I created https://github.com/camunda/camunda-docs/pull/4159 to do this

PR #21163, opened 2024-08-14
refactor: remove optimize 7 onboarding data modification

RomanJRW:
## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes #

PR #21164, opened 2024-08-14
refactor: remove optimize platform condition

RomanJRW:
## Description

Probably should have been built on top of https://github.com/camunda/camunda/pull/21163, but I realised the overlap a little late. Will fix this up once the other is merged

## Checklist

- [ ] for CI changes:

## Related issues

closes #


optimize/backend/src/main/java/io/camunda/optimize/service/security/SessionService.java
```diff
@@ -99,51 +80,6 @@ private boolean isValidAuthToken(final String token) {
     return isValid;
   }
 
-  public Optional<String> refreshAuthToken(final String currentToken) {
```
PHWaechtler, line 102: ❓ So jwt is still used, but we dont do invalidation or refresh anymore? May be a bit out of touch with the jwt logic here can you give me a refresher

HeleneW-dot (changes requested):
Lgtm, got one question on clarity for current session logic. Two small things I think we could add to this pr:

- I think we can remove the Cloud Condition with this one too, it will always just be true
- we can merge `UserTaskIdentityService` and `CloudUserTaskIdentityService` (unless we're expecting a saas/ccsm split some time soon?)

RomanJRW:
> I think we can remove the Cloud Condition with this one too, it will always just be true

I had added this as a new item on the clean up list, almost verbatim as what you said. As you are happy for me to incorporate it here, I will do that

> we can merge UserTaskIdentityService and CloudUserTaskIdentityService (unless we're expecting a saas/ccsm split some time soon?)

Good catch, will do that here too 👍

optimize/backend/src/main/java/io/camunda/optimize/service/security/SessionService.java
RomanJRW: In CCSM, we use the `CCSMAuthenticationCookieFilter` to handle the token refreshing. It is quite well commented in the flow.

The code being removed here was just the stuff that became unused after the deletion of the `AuthenticationCookieRefreshFilter`, which was only defined/used in the now-deleted platform web security adaptor

HeleneW-dot (approved):
🚀

PR #21165, opened 2024-08-14
refactor: remove optimize super users and super groups

RomanJRW:
## Checklist

- [ ] for CI changes:

## Related issues

closes #


optimize/util/optimize-commons/src/main/resources/service-config.yaml
```diff
@@ -607,7 +607,7 @@ panelNotification:
 
 entity:
   # which users are authorized to create/edit/delete Optimize entities outside of a collection.
-  # Available options: 'all', 'superuser', 'none'
```
PHWaechtler, line 610: wanna remove the superUserIds and superGroupIds as well while we're here? Also there's another mention in the env config `Telemetry can later be enabled/disabled in the UI by superusers` (although I guess the telemetry config will also be removed?)

HeleneW-dot (approved):
Lgtm, shall we also add a not in the [docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/#security) that this is a C7 only config?
Does this already have a FE follow up as well?

optimize/util/optimize-commons/src/main/resources/service-config.yaml
RomanJRW: Yeah I did a search on github and I think config here can be removed already, have pushed another commit doing that.

A separate commit clears up the note about telemetry in the UI

RomanJRW:
> Lgtm, shall we also add a not in the [docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/#security) that this is a C7 only config?

Hmm it's not really a section we separate at all right now. I will hold off until the later fork splitting, when I imagine this will need proper attention

> Does this already have a FE follow up as well?

What FE change would be needed here?

michaldytko:
Hi @RomanJRW , what version doesn't support superuser anymore? I just have a ticket where customer tries to enable it with 8.5.

RomanJRW:
@michaldytko: Optimize 8 never supported the concept of superusers. This PR is the change that removed it from the code base though, as they were only ever officially supported in C7 use cases

PR #21166, opened 2024-08-14
refactor: remove optimize C7 test dependencies

RomanJRW:
## Description

Initial PR to remove the remaining C7 test dependencies (excluding some of the commented out files, such as the extension).

I would like pipelines to be tested with this before merging to make sure it won't break things like the ES and OS tests, and others using the docker-compose files I am deleting here

## Checklist

- [ ] for CI changes:

## Related issues

closes #


misiekhardcore:
@RomanJRW I added one more commit with more cleanup

.github/workflows/optimize-deploy-artifacts.yml
```diff
@@ -76,15 +76,15 @@ jobs:
       with:
         compose_file: .github/actions/compose/docker-compose.smoketest.yml
         project_name: smoketest
+        timeout: 600
```
misiekhardcore, line 79: we need to reduce this timeout to something reasonable

.github/workflows/optimize-deploy-artifacts.yml
```diff
@@ -76,15 +76,15 @@ jobs:
...
       env:
-        OPTIMIZE_IMAGE_TAG: ${{ env.DOCKER_TAG }}
+        OPTIMIZE_IMAGE_TAG: ${{ env.RELEASE_VERSION }}
```
misiekhardcore, line 81: why do we do this change here?

.github/workflows/optimize-deploy-artifacts.yml
misiekhardcore: actually we can remove this param here completely, the default value of 300s works fine

matthewBearCamunda (changes requested):
The changes LGTM, however I did find references to h2 and postgres in `optimize/backend/src/it/java/io/camunda/optimize/test/it/extension/EngineDatabaseExtension.java.out`

Does that file, and possibly the others in the same directory, need adjusting too?

RomanJRW:
@matthewBearCamunda - good question. Ultimately, we are likely to delete all the IT things because of their C7-centricity, including this extension (this is even Engine specific). So it will be deleted in a separate PR. For the moment, the code is commented out and it has no impact here

matthewBearCamunda (approved):
Makes sense, approved!

PR #21169, opened 2024-08-15
ci: fix optimize issue templates

misiekhardcore:
## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

related to #21168

PR #21172, opened 2024-08-15
ci: run Java code checks incl Markdown linter on *.md change

cmur2:
## Description

This will prevent situations like in #21169. While working on this PR I also realized that using a filter expression like `*.md` with https://github.com/dorny/paths-filter only looks at files in the root folder - one has to explicitly use `**/*.md` for subpath matching.

Thus, I also fixed the Maven change filter which before this PR only detected `pom.xml` changes in the root but was actually meant to find all those changes in subpaths as well.

This PR needs https://github.com/camunda/camunda/pull/20656

## Checklist

- [x] for CI changes:

## Related issues

None

### DIFF
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -11,14 +11,15 @@ outputs:
         github.event_name == 'push' ||
         steps.filter-common.outputs.github-actions-change == 'true'
       }}
-  java-code-changes:
+  java-code-changes:  # Also runs spotless for Markdown linting
     description: Output whether jobs depending on Java code changes should be run, related checks are based on GitHub events and java files changes
     value: >-
       ${{
         github.event_name == 'push' ||
         steps.filter-common.outputs.github-actions-change == 'true' ||
         steps.filter-common.outputs.java-code-change == 'true' ||
-        steps.filter-common.outputs.maven-change == 'true'
+        steps.filter-common.outputs.maven-change == 'true' ||
+        steps.filter-common.outputs.markdown-change == 'true'
       }}
   identity-frontend-tests:
     description: Output whether Identity frontend tests should be run based on GitHub event and files changed
@@ -92,6 +93,9 @@ runs:
         dockerfile-change:
           - 'Dockerfile'
           - '*.Dockerfile'
+
+        markdown-change:
+          - '**/*.md'
 
         frontend-change:
           - 'operate/client/**'
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -84,11 +84,11 @@ runs:
           - '**/src/test/**'
 
         maven-change:
-          - 'bom/*'
+          - 'bom/**'
           - 'build-tools/**'
           - 'dist/**'
-          - 'parent/*'
-          - 'pom.xml'
+          - 'parent/**'
+          - '**/pom.xml'
 
         dockerfile-change:
           - 'Dockerfile'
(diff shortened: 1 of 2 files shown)


npepinpe (approved):

.github/actions/paths-filter/action.yml
```diff
@@ -83,16 +84,19 @@ runs:
...
 
         maven-change:
-          - 'bom/*'
+          - 'bom/**'
           - 'build-tools/**'
           - 'dist/**'
-          - 'parent/*'
-          - 'pom.xml'
+          - 'parent/**'
+          - '**/pom.xml'
```
Zelldon, line 91: I just wanted to raise that it before it was actually intended that it only affected the root pom, as the use was to run everything when parent or root changes. 

Not sure whether we still want this, but some thought. Now it will trigger even if a pom in Operate or Optimize deep down changes. 🤔

.github/actions/paths-filter/action.yml
```diff
@@ -11,14 +11,15 @@ outputs:
...
     description: Output whether jobs depending on Java code changes should be run, related checks are based on GitHub events and java files changes
     value: >-
       ${{
         github.event_name == 'push' ||
         steps.filter-common.outputs.github-actions-change == 'true' ||
         steps.filter-common.outputs.java-code-change == 'true' ||
-        steps.filter-common.outputs.maven-change == 'true'
+        steps.filter-common.outputs.maven-change == 'true' ||
+        steps.filter-common.outputs.markdown-change == 'true'
       }}
```
Zelldon, line 23: ❓ I was wondering whether we want to somehow separate this 🤔 to better react on in the ci and only run spotless markdown (if possible) and not everything else 🤔

cmur2: > whether we want to somehow separate this

That would be nice since it makes the whole setup more modular and allows us to only run the necessary functions. For this PR though I want to focus on getting the minimum done and close the loophole, so any refactoring into separate jobs (one for java checks and one for markdown checks) would have to be a follow-up 👍🏼

.github/actions/paths-filter/action.yml
cmur2: I added this filter in https://github.com/camunda/camunda/commit/d4ba3e5adb5abc8651fb2ddb9316515833c8de78 and back then my (wrong) understanding was that it would actually trigger on any `pom.xml` in the subtree. The filter was intended to capture any Maven-related change.

This filter is currently used to trigger Java compile, unit and integration tests which could be affected by the root pom or subtree poms as well right (changed properties/goals)? From my non-Java-expert POV I think it makes sense to trigger on those but happy to be overruled by you as Java experts.

I just want to avoid having another loophole - rather have execute too often.

cmur2:
> forget them in x-change parts

I'm not sure I understand what `x-change` parts refer to?

> Why don't we split off formatting to a different job that should run on any changes? Formatting checks seem like something that's cheap to run and are almost always needed anyways.

💯 percent in favor of this (@Zelldon also suggested something similar) -> see my [comment](https://github.com/camunda/camunda/pull/21172#discussion_r1718327299)

korthout:
>>Why don't we split off formatting to a different job that should run on any changes? Formatting checks seem like something that's cheap to run and are almost always needed anyways.
>
> 💯 percent in favor of this (@Zelldon also suggested something similar) -> see my https://github.com/camunda/camunda/pull/21172#discussion_r1718327299

@cmur2 Can you open an issue for this?

korthout:
@cmur2 Awesome! Thanks for considering this 🚀

PR #21174, opened 2024-08-15
test: migrating test to c8 auth error

Oleksiivanov:
This PR provides an example of how we can migrate the Optimize tests from Camunda 7 (C7) to Camunda 8 (C8). It demonstrates the necessary steps and adjustments required for the transition. However, please note that this example is not fully functional due to ongoing issues with authentication.

Key Points:

 - Authentication Issues: We are currently encountering 401 errors during user authentication.
 - Token Verification: There are problems with the token verification process, leading to authorization failures.
 - Model Adjustments: Introduced the new BpmnModelsC8 class to handle structural changes in BPMN models between C7 and C8.
 
Further investigation and resolution of the authentication issues are required to achieve a fully working migration.

Related : 
 - https://github.com/camunda/camunda-optimize/pull/13512
 - https://github.com/camunda/camunda-optimize/pull/13662
 - https://github.com/camunda/team-data-engine-maintenance/issues/14
 - https://github.com/camunda/camunda-optimize/pull/13971


RomanJRW:
As this was never intended to be merged, but rather investigate and prototype a potential approach, I will go ahead and close this PR but leave the branch in tact.

Feel free to reopen if you were still committing here

PR #21182, opened 2024-08-15
ci: disable Docker to GHA cache for Zeebe Benchmark

### LINKED ISSUE #18750: CI: avoid GitHub Action cache over usage errors

cmur2:
## Description

Produces large cache which is probably not useful. See #18750 for motivation to get GHA cache size down to less dangerous levels. Looking at the history of https://github.com/camunda/camunda/actions/workflows/zeebe-benchmark.yml it is usually invoked only a couple of times per week so user impact should be small.

## Checklist

- [ ] for CI changes:

## Related issues

closes #18750

### DIFF
--- a/.github/workflows/zeebe-benchmark.yml
+++ b/.github/workflows/zeebe-benchmark.yml
@@ -218,8 +218,6 @@ jobs:
           context: .
           push: true
           tags: gcr.io/zeebe-io/operate:${{ inputs.operate-tag }}
-          cache-from: type=gha,ignore-error=true
-          cache-to: type=gha,mode=max,ignore-error=true
           file: operate.Dockerfile
 
   build-benchmark-images:


npepinpe (approved):

Zelldon:
Might conflict with https://github.com/camunda/camunda/pull/21189

ChrisKujawa (approved):

PR #21189, opened 2024-08-15
ci: improve benchmark workflow

lenaschoenburg:
- Don't measure network latency by default, no one wants that
- Use the shared maven cache
- Only build the dist module and its dependencies
- Skip building test sources, the frontend and anything needing Go

Reduces the time it takes to run this workflow from ~10 minutes to ~ 4 minutes: https://github.com/camunda/camunda/actions/runs/10404685239

Also removes a lot of flakiness by not building the frontend.

### DIFF
--- a/.github/workflows/zeebe-benchmark.yml
+++ b/.github/workflows/zeebe-benchmark.yml
@@ -129,24 +129,22 @@ jobs:
           token_format: 'access_token'
           workload_identity_provider: 'projects/628707732411/locations/global/workloadIdentityPools/zeebe-gh-actions/providers/gha-provider'
           service_account: 'zeebe-gh-actions@zeebe-io.iam.gserviceaccount.com'
-      - name: Setup BuildKit
-        uses: docker/setup-buildx-action@v3
-      - name: Login to GCR
-        uses: docker/login-action@v3
+      - uses: ./.github/actions/setup-zeebe
         with:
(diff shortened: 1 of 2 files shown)


npepinpe (approved):

ChrisKujawa (approved):
Thanks 🙇🏼

.github/workflows/zeebe-benchmark.yml
```diff
@@ -123,30 +123,36 @@ jobs:
...
         with:
-          context: .
-          tags: gcr.io/zeebe-io/zeebe:${{ needs.calculate-image-tag.outputs.image-tag }}
+          repository: 'gcr.io/zeebe-io/zeebe'
+          revision: ${{ inputs.ref }}
           push: true
-          cache-from: type=gha,ignore-error=true
-          cache-to: type=gha,mode=max,ignore-error=true
-          build-args: DIST=build
-          provenance: false
```
cmur2, line 148: Why was this option removed? It might not be needed for images that are internal-use only but just fyi it can cause problems on older registries: https://github.com/camunda/camunda/commit/1a63c420e952e9fe09e5f6612b349bf820d0f44f

lenaschoenburg: This workflow only pushes to gcr.io/zeebe-io which apparently supports provenance.

PR #21192, opened 2024-08-15
ci: don't build frontend to test docker image

lenaschoenburg:
These tests don't need a working frontend. Removing these steps should make the workflow much faster.

Before it took 15 minutes: https://github.com/camunda/camunda/actions/runs/10405673664/job/28817024677
After it took 8 minutes: https://github.com/camunda/camunda/actions/runs/10405712323/job/28817102079


ChrisKujawa (approved):
We are not using the docker images later for snapshot right? SO I guess should be fine then?

lenaschoenburg:
No, they are not reused :+1:

Zelldon:
Thanks for doing it btw @lenaschoenburg 🙇🏼

cmur2:
Seen and appreciated 🙇🏼

PR #21195, opened 2024-08-15
feat: scheduling service is configurable and observable

lenaschoenburg:
Follow up after https://github.com/camunda/camunda/issues/21051.

Makes the micrometer registry available to the stream processor and uses it to publish new metrics for the scheduling service:

- Number of scheduled tasks
- Delay between actual task execution and the requested time

These metrics are not visualized because they are usually not relevant. The number of scheduled tasks is usually 6 because all of the scheduled tasks try hard to only schedule their next execution and nothing more.
The delay might be slightly more interesting but we don't often investigate support cases where this information is needed.

I've also added a configuration option for the interval at which we check for expired scheduled tasks: `zeebe.broker.processing.scheduledTaskCheckInterval`. By default it's still 1 second.

### DIFF
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -968,6 +968,11 @@
       # This setting can also be overridden using the environment variable ZEEBE_BROKER_PROCESSING_SKIPPOSITIONS
       # skipPositions: ""
 
+      # Configures the rate at which a partition leader checks for expired scheduled tasks such as the due date checker.
+      # The default value is 1 second. Use a lower interval to potentially decrease delays between requested and actual execution of scheduled tasks.
+      # This setting can also be overridden using the environment variable ZEEBE_BROKER_PROCESSING_SCHEDULEDTASKCHECKINTERVAL
+      # scheduledTaskCheckInterval: 1s
+
     # experimental
       # Be aware that all configuration's which are part of the experimental section
       # are subject to change and can be dropped at any time.
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -970,6 +970,7 @@
 
       # Configures the rate at which a partition leader checks for expired scheduled tasks such as the due date checker.
       # The default value is 1 second. Use a lower interval to potentially decrease delays between requested and actual execution of scheduled tasks.
+      # Using a low interval will result in unnecessary load while idle. We recommend to benchmark any changes to this setting.
       # This setting can also be overridden using the environment variable ZEEBE_BROKER_PROCESSING_SCHEDULEDTASKCHECKINTERVAL
       # scheduledTaskCheckInterval: 1s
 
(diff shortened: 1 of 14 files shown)


dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -968,6 +968,11 @@
       # This setting can also be overridden using the environment variable ZEEBE_BROKER_PROCESSING_SKIPPOSITIONS
       # skipPositions: ""
 
+      # Configures the rate at which a partition leader checks for expired scheduled tasks such as the due date checker.
+      # The default value is 1 second. Use a lower interval to potentially decrease delays between requested and actual execution of scheduled tasks.
+      # This setting can also be overridden using the environment variable ZEEBE_BROKER_PROCESSING_SCHEDULEDTASKCHECKINTERVAL
```
npepinpe, line 973: :question: Should we put a warning about a possible performance issue if you put it too low? Or in the docs? :thinking: Essentially, I'm trying to anticipate a user saying "oh well, I don't want delays right?" :smile:

npepinpe: Alternative is putting an enforced lower bound, e.g. 100ms, 200ms, something.

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceTest.java
```diff
@@ -67,7 +68,9 @@ void before() {
             lifecycleSupplier,
             () -> testWriter,
             commandCache,
-            actorScheduler.getClock());
+            actorScheduler.getClock(),
+            Duration.ofSeconds(1),
+            ScheduledTaskMetrics.noop());
```
npepinpe, line 73: :wrench: With Micrometer, we can now also inject our `SimpleMeterRegistry` to effectively test metrics properly. I think a simple test would suffice. I understand this isn't the most important metric, but I would like to encourage us starting, and this seems as good a place to start as any :)

npepinpe (approved):
:rocket:

npepinpe:
I'm thinking of trying out using a [TimeGauge](https://docs.micrometer.io/micrometer/reference/concepts/gauges.html#_timegauge) to actually view the current clock in the dashboard. [There's an official clock panel](https://raw.githubusercontent.com/grafana/clock-panel/06ecf59c191db642127c6153bc3145e93a1df1f8/src/img/screenshot-clocks.png). Maybe not the most useful, but just a way to play with Micrometer :smile:

dist/src/main/config/broker.standalone.yaml.template
lenaschoenburg: I've added a small note for this :+1:

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/ProcessingScheduleServiceTest.java
lenaschoenburg: The `SimpleMeterRegistry` really lends itself well for this :+1: 
Added a couple of basic tests now.

PR #21197, opened 2024-08-15
Support REST for Job Completion

### LINKED ISSUE #21167: Support the Complete Job API over REST

nicpuppa:
## Description

Support REST for Job Completion in the java client

## Related issues

closes #21167

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CompleteJobCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CompleteJobCommandImpl.java
@@ -135,13 +135,13 @@ public final class CompleteJobCommandImpl extends CommandWithVariables<CompleteJ
   @Override
   protected CompleteJobCommandStep1 setVariablesInternal(final String variables) {
     grpcRequestObjectBuilder.setVariables(variables);
-    // this check is mandatory, otherwise gRPC requests will fail
-    // gRPC and REST are handling setting variables in a different way:
-    // - for gRPC commands we will check only if is a valid JSON and forward it to the engine.
-    //    is the engine that check if the provided string can be transformed into a Map, if not it
-    //    it will throw an error
-    // - for REST commands instead if users provides a not valid map string, an exception will be thrown
-    //    from the client
+    // This check is mandatory. Without it, gRPC requests can fail unnecessarily.
+    // gRPC and REST handle setting variables differently:
+    // - For gRPC commands, we only check if the JSON is valid and forward it to the engine.
+    //    The engine checks if the provided String can be transformed into a Map, if not it
+    //    throws an error.
+    // - For REST commands, users have to provide a valid JSON Object String.
+    //    Otherwise, the client throws an exception already.
     if (useRest) {
       httpRequestObject.setVariables(jsonMapper.fromJsonAsMap(variables));
     }
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CompleteJobCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CompleteJobCommandImpl.java
@@ -22,7 +22,10 @@ import io.camunda.zeebe.client.api.command.CompleteJobCommandStep1;
 import io.camunda.zeebe.client.api.command.FinalCommandStep;
 import io.camunda.zeebe.client.api.response.CompleteJobResponse;
 import io.camunda.zeebe.client.impl.RetriableClientFutureImpl;
+import io.camunda.zeebe.client.impl.http.HttpClient;
+import io.camunda.zeebe.client.impl.http.HttpZeebeFuture;
 import io.camunda.zeebe.client.impl.response.CompleteJobResponseImpl;
+import io.camunda.zeebe.client.protocol.rest.JobCompletionRequest;
 import io.camunda.zeebe.gateway.protocol.GatewayGrpc.GatewayStub;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.CompleteJobRequest;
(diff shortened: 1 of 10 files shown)


clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CompleteJobCommandImpl.java
```diff
@@ -85,7 +134,17 @@ private void send(
...
   protected CompleteJobCommandStep1 setVariablesInternal(final String variables) {
-    builder.setVariables(variables);
+    grpcRequestObjectBuilder.setVariables(variables);
+    // this check is mandatory, otherwise gRPC requests will fail
+    // gRPC and REST are handling setting variables in a different way:
+    // - for gRPC commands we will check only if is a valid JSON and forward it to the engine.
+    //    is the engine that check if the provided string can be transformed into a Map, if not it
+    //    it will throw an error
+    // - for REST commands instead if users provides a not valid map string, an exception will be thrown
+    //    from the client
```
tmetzke, line 144: 🔧 Just some wording suggestions.

```suggestion
    // This check is mandatory. Without it, gRPC requests can fail unnecessarily.
    // gRPC and REST handle setting variables differently:
    // - For gRPC commands, we only check if the JSON is valid and forward it to the engine.
    //    The engine checks if the provided String can be transformed into a Map, if not it
    //    throws an error.
    // - For REST commands, users have to provide a valid JSON Object String.
    //    Otherwise, the client throws an exception already.
```

service/src/main/java/io/camunda/service/JobServices.java
```diff
@@ -52,18 +52,24 @@ public JobServices<T> withAuthentication(final Authentication authentication) {
...
-      final ActivateJobsRequest request,
+      final String type,
+      final int maxJobsToActivate,
+      final List<String> tenantIds,
+      final long timeout,
+      final String worker,
+      final List<String> variables,
+      final long requestTimeout,
       final ResponseObserver<T> responseObserver,
       final Consumer<Runnable> cancelationHandlerConsumer) {
```
tmetzke, line 63: ❓ Any specific reason to move to this long list of parameters instead of a parameter object? This is usually not recommended for an API since a long list of method parameters becomes hard to read and understand really quick.

tmetzke (commented):
👍 This is AMAZING, great job, love the test changes and coverage 🚀 
❓ The only thing I don't understand is the motivation to change the job services method for job activation 😅 Please enlighten me 🦕

service/src/main/java/io/camunda/service/JobServices.java
nicpuppa: Yep, in the service module we don't have access to records from `RequestMapper`. I thought about it a bit, and I prefer having all the records in the same place, even if this means to have a methods with (maybe) too much args

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CompleteJobCommandImpl.java
nicpuppa: Thanks 🙏

service/src/main/java/io/camunda/service/JobServices.java
tmetzke: I'm not sure if I agree here, to be honest. I think this makes the service methods worse, which is not favorable. Let's rather have the services ALWAYS define the records they consume/produce then as a rule.

tmetzke: The services know best what they need/create, so they should handle this --> separation of concerns

tmetzke: If a service method doesn't justify a parameter object (e.g. just max. 3 parameters), it doesn't expose one.

nicpuppa: I mean, having the records in the same place doesn't mean to have them in the Mapper 🤣 
I can move them into the services, so they can be also separated by type (Job and User Task)

tmetzke: Well, I think (but I might be wrong) the records in the mapper are only used in the REST controllers, they are not necessarily sent to the services as well. So, the services declaring them but not using them makes not much sense.

We have helper records in the mapper to create request data we can easily pass around in the REST classes until we forward some of their data to the services.

We also have helper records in some services if the method signatures would otherwise become too long. The REST mapper reuses those if they are already there.

I personally think this is actually OK, but that might just be me 😄

nicpuppa: I mean, with my changes also `ActivateJobsRequest ` record is used only in the controller 😄 I don't get your point here

tmetzke: The way to fix that is by introducing a parameter object that the service class can use. Which brings us back to the code without the adjustments here. 🙂

nicpuppa: I'll roll back the changes 🛩️ 

But I'll keep an eye on this, because I'd like to have all the records in the same place 👍

tmetzke: Thanks 👍
Please don't be disappointed if that is not possible or leads to an over-engineered solution 😅 I understand the drive behind it but the REST mapper simply is an intersection place for different "worlds" (REST data and services data). It cannot contain all possible data structures it deals with, mapping back and forth. Some of them it handles on its own (the records defined in it only needed for REST), some of them come from the services API (either records created there or even those it simply passes along from the broker results).

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/JobController.java
```diff
@@ -101,7 +101,8 @@ private CompletableFuture<ResponseEntity<Object>> activateJobs(
     final var result = new CompletableFuture<ResponseEntity<Object>>();
     final var responseObserver = responseObserverProvider.apply(result);
     jobServices.activateJobs(
-        activationRequest, responseObserver, responseObserver::setCancelationHandler);
+        activationRequest, responseObserver,
+        responseObserver::setCancelationHandler);
```
tmetzke, line 105: ❌ This needs to be reverted as well, the Java checks fail otherwise.

tmetzke (approved):
🚀

PR #21199, opened 2024-08-15
feat: Optimize - log license information on startup and add license info to the API

matthewBearCamunda:
## Description

 Read the `CAMUNDA_LICENSE_KEY` environment variable on startup, and validate the license. It should log on startup like https://github.com/camunda/camunda/issues/18970 and have API data returned like https://github.com/camunda/camunda/issues/20540. Instead of a dedicated API endpoint to return license information, for Optimize the license information is added to the existing `/ui-configuration` endpoint

Optimize requires its own service and separate implementation because Optimize is not part of the single application inside the monorepo. [See this thread for more
info](https://camunda.slack.com/archives/CUZUYDMMG/p1723814110562649)

## Checklist

- [ ] for CI changes:
- [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI"
requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes #

(cherry picked from commit a974fe7c1ef73efa0fe3be443db084fb44574510)

## Checklist

- [ ] for CI changes:

## Related issues

closes #


optimize/backend/pom.xml
```diff
@@ -398,6 +398,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
RomanJRW, line 401: do you think it would be worth adding a requirement to @buccarel's single application issue that explicitly states that this dependency should be removed?

optimize/backend/src/main/java/io/camunda/optimize/service/CamundaLicenseService.java
```diff
@@ -0,0 +1,30 @@
...
+import io.camunda.service.license.CamundaLicense;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CamundaLicenseService {
+
+  private final CamundaLicense license;
+
+  public CamundaLicenseService() {
+    final String camundaLicense = System.getenv("CAMUNDA_LICENSE_KEY");
```
RomanJRW, line 19: Let's have a constant for this, if one doesn't already exist

optimize/backend/src/main/java/io/camunda/optimize/service/UIConfigurationService.java
```diff
@@ -102,6 +105,14 @@ public UIConfigurationResponseDto getUIConfiguration() {
     return uiConfigurationDto;
   }
 
+  private boolean isCamundaLicenseValid() {
+    return camundaLicenseService.isCamundaLicenseValid();
+  }
+
+  private String getCamundaEnvironment() {
```
RomanJRW, line 112: Why is this called environment? It is not really returning an environment is it?

Also, do we really need it here for Optimize? We already return the Optimize profile on line 66, why not reuse that?

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/query/ui_configuration/UIConfigurationResponseDto.java
```diff
@@ -37,6 +37,8 @@ public class UIConfigurationResponseDto {
   private int maxNumDataSourcesForReport;
   private Integer exportCsvLimit;
   private DatabaseType optimizeDatabase;
+  private boolean validLicense;
+  private String environment;
```
RomanJRW, line 41: I know we didn't on the other PR, so maybe this question is broader than Optimize here - but why did we not have an enum for the environments?

RomanJRW (changes requested):

optimize/backend/src/main/java/io/camunda/optimize/service/CamundaLicenseService.java
matthewBearCamunda: One was added with https://github.com/camunda/camunda/pull/21236

I've taken out the `String` and am using the constant now

optimize/backend/src/main/java/io/camunda/optimize/service/UIConfigurationService.java
matthewBearCamunda: `environment` came from the API design discussion here: https://github.com/camunda/camunda/issues/20540. Right now it's either be `saas`,`self-managed`, and `unknown`, but a possible future state could include `hybrid`.

I'd prefer to keep it, as these two fields are an exact mirror of what the Unified REST API `/license` endpoint is. The `/license` endpoint will be used in the future once Optimize is part of the single-application. I'd want to keep the transition from these temporary Optimize-specifc fields to the `/license` endpoint as painless as possible

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/query/ui_configuration/UIConfigurationResponseDto.java
matthewBearCamunda: I added a License enum to both Optimize and the other part of the monorepo. It makes sense

matthewBearCamunda: Adding a new `enum` caused build issues with finding the new class. I put the new `enum` inside `CamundaLicense` to fix the build issue

optimize/backend/pom.xml
```diff
@@ -398,6 +398,21 @@
...
+      <groupId>io.camunda</groupId>
+      <artifactId>camunda-service</artifactId>
+      <version>${project.version}</version>
+      <exclusions>
```
matthewBearCamunda, line 405: These exclusions were added to fix build issues. There were errors related to not being able to find the libraries

optimize/backend/src/main/java/io/camunda/optimize/service/UIConfigurationService.java
```diff
@@ -70,6 +72,8 @@ public UIConfigurationResponseDto getUIConfiguration() {
     uiConfigurationDto.setMaxNumDataSourcesForReport(
         configurationService.getUiConfiguration().getMaxNumDataSourcesForReport());
     uiConfigurationDto.setOptimizeDatabase(ConfigurationService.getDatabaseType(environment));
+    uiConfigurationDto.setValidLicense(isCamundaLicenseValid());
+    uiConfigurationDto.setLicenseType(getLicenseType().getName());
```
PHWaechtler, line 76: Do we already have a FE follow up to use these?
Also lets add these fields to our r[est api docs](https://confluence.camunda.com/display/CO/UI+Configuration)

optimize/backend/src/test/java/io/camunda/optimize/service/uiconfiguration/UIConfigurationServiceTest.java
```diff
@@ -164,5 +167,7 @@ private void initializeMocks() {
     when(identityUsers.isAvailable()).thenReturn(true);
     when(environment.getProperty(CAMUNDA_OPTIMIZE_DATABASE, ELASTICSEARCH_DATABASE_PROPERTY))
         .thenReturn(DatabaseType.ELASTICSEARCH.toString());
+    when(camundaLicenseService.getCamundaLicenseType()).thenReturn(LicenseType.SAAS);
+    when(camundaLicenseService.isCamundaLicenseValid()).thenReturn(true);
```
PHWaechtler, line 171: Lets also add some assertions on these to asser the correct returned uiConfig

HeleneW-dot (changes requested):
Lgtm, got some small questions re follow ups, one suggestion for testing and one additional q: the commit message mentions logging license info on startup, but I dont see any code changes that do this. Is this already implemented>

optimize/backend/src/main/java/io/camunda/optimize/service/UIConfigurationService.java
matthewBearCamunda: We do! Tnose tickets are linked on the epic https://github.com/camunda/product-hub/issues/2359

Added!

optimize/backend/pom.xml
matthewBearCamunda: I just removed this. After chatting with Mattia, Optimize cannot use any monorepo dependencies until it is part of the single application

matthewBearCamunda:
Still working on addressing the build issues. I chatted with Mattia and I cannot use any monorepo dependecies in Optimize. I will have to copy the code used in the mono-repo into Optimize and use that

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/query/ui_configuration/UIConfigurationResponseDto.java
matthewBearCamunda: I was able to figure out and solve the build issues, the `enum` is back to being it own class

matthewBearCamunda:
Optimize cannot use any monorepo dependencies. I copied the license implementation from the monorepo and used it in the optimize module. 

Once Optimize is a part of the single application, these duplicate Optimize classes can be removed, and the monorepo implementation can be used

optimize/backend/pom.xml
```diff
@@ -398,20 +398,11 @@
...
-          <artifactId>zeebe-scheduler</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>io.camunda</groupId>
-          <artifactId>zeebe-gateway-protocol-impl</artifactId>
-        </exclusion>
-      </exclusions>
+      <groupId>org.camunda.bpm</groupId>
+      <artifactId>camunda-license-check</artifactId>
+      <version>2.9.0</version>
```
PHWaechtler, line 405: This version is accessible for everyone right?

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,112 @@
...
+public class CamundaLicense {
+
+  public static final String CAMUNDA_LICENSE_ENV_VAR_KEY = "CAMUNDA_LICENSE_KEY";
+  private static final Logger LOGGER = LoggerFactory.getLogger(CamundaLicense.class);
+  private static final String UNKNOWN_LICENSE_TYPE = "unknown";
+  private boolean isValid;
+  private LicenseType licenseType;
+  private boolean isInitialized;
+
+  // @VisibleForTesting
```
PHWaechtler, line 25: 👀 whats this one

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,112 @@
...
+    if (license != null && !license.isBlank()) {
+      isValid = determineLicenseValidity(license);
+      licenseType = getLicenseTypeFromProperty(license);
+    } else {
+      isValid = false;
+      licenseType = LicenseType.UNKNOWN;
+      LOGGER.error(
+          "No license detected when one is expected. Please provide a license through the "
+              + CAMUNDA_LICENSE_ENV_VAR_KEY
+              + " environment variable.");
```
PHWaechtler, line 54: Nit
Can be one line

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    isInitialized = true;
+  }
+
+  /**
+   * Camunda licenses will have a license type property, fetch that out of the license and return
+   * the value
+   *
+   * <p>Self-managed mode is any other possibility. (ex, blank license, prop missing, etc)
```
PHWaechtler, line 64: I dont understand this sentence 🤔

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,112 @@
...
+    } catch (final InvalidLicenseException e) {
+      LOGGER.error(
+          "Expected a valid license when determining the type of license, but encountered an invalid one instead. ",
+          e);
+    } catch (final Exception e) {
+      LOGGER.error(
+          "Expected to determine the license type of the license, but the following unexpected error was encountered: ",
+          e);
+    }
+    return LicenseType.UNKNOWN;
```
PHWaechtler, line 87: ❓ we still just return the unknown type and would validate with`initializeWithLicense` as long as the license itself was valid. Should this be blocking validation or is a valid license with unknown type still fine (or maybe it would never be valid in that case)?

optimize/backend/src/main/java/io/camunda/optimize/license/LicenseType.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package io.camunda.optimize.license;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public enum LicenseType {
+  SAAS("saas"),
+  PRODUCTION("production"),
```
PHWaechtler, line 16: ❓ probably not your creation, but why these types? I would have expected either differentiation like saas/ccsm or trial/production.

optimize/pom.xml
```diff
@@ -691,6 +690,30 @@
...
     </repository>
+
+    <repository>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>camunda-cpm</id>
```
PHWaechtler, line 701: should this be `camunda-bpm`?

optimize/pom.xml
```diff
@@ -691,6 +690,30 @@
       </snapshots>
       <layout>default</layout>
     </repository>
+
```
PHWaechtler, line 693: ❓ Why are we adding the bpm repo? I think I'm missing some context here

service/src/main/java/io/camunda/service/license/LicenseType.java
```diff
@@ -0,0 +1,41 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.service.license;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public enum LicenseType {
```
PHWaechtler, line 14: ❓ Are these the duplicates you mentioned here `Once Optimize is a part of the single application, these duplicate Optimize classes can be removed, and the monorepo implementation can be used` ? So basically both are our addition, but rn we need a temporary backup in optimize too?

HeleneW-dot (dismissed):
While I dont have important change suggestions, I think I have too many questions to confidently approve for now. Quite possibly I'm just a bit out of the loop for the context of these changes but would appreciate clarification 🙏

service/src/main/java/io/camunda/service/license/LicenseType.java
```diff
@@ -0,0 +1,41 @@
...
+  SAAS("saas"),
+  PRODUCTION("production"),
+  UNKNOWN("unknown");
```
RomanJRW, line 17: so does unknown actually mean nothing has been provided? Like it is unlicensed?

RomanJRW, line 17: Hmm it feels like these are not exclusive. Production doesn't imply Self-Managed to me. I don't feel so strongly about this and it's not a new thing introduced by this Enum, but the design feels a bit off to me. Also if my understanding of Unknown is correct, then it's not really a "type" of license, but rather an absence of a license

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -16,11 +16,11 @@
 
 public class CamundaLicense {
 
+  public static final String CAMUNDA_LICENSE_ENV_VAR_KEY = "CAMUNDA_LICENSE_KEY";
   private static final Logger LOGGER = LoggerFactory.getLogger(CamundaLicense.class);
   private static final String UNKNOWN_LICENSE_TYPE = "unknown";
```
RomanJRW, line 21: if you proceed with enums, I don't think this would be necessary (also happy with a non-enum solution)

service/src/main/java/io/camunda/service/license/LicenseType.java
```diff
@@ -0,0 +1,41 @@
...
+    this.name = name;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public static LicenseType get(final String name) {
+    return ENUM_MAP.getOrDefault(name.toLowerCase(), LicenseType.UNKNOWN);
+  }
```
RomanJRW, line 40: could this not be simplified with a wrapped Enum.valueOf?

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,112 @@
...
+      return false;
+    } catch (final Exception e) {
+      LOGGER.error(
+          "Expected to validate a the Camunda license, but the following unexpected error was encountered: ",
+          e);
+      return false;
+    }
+  }
+
+  // @VisibleForTesting
```
RomanJRW, line 108: I don't think we need this as there is no test for it anyway

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,112 @@
...
+          e);
+    } catch (final Exception e) {
+      LOGGER.error(
+          "Expected to determine the license type of the license, but the following unexpected error was encountered: ",
+          e);
+    }
+    return LicenseType.UNKNOWN;
+  }
+
+  private boolean determineLicenseValidity(final String licenseStr) {
```
RomanJRW, line 90: this signature doesn't read well to me. Determining something doesn't suggest you would necessarily get a boolean response from it. Rather something like `isLicenseValid(...)` would suggest a boolean response

optimize/backend/src/main/java/io/camunda/optimize/license/LicenseType.java
```diff
@@ -0,0 +1,41 @@
...
+  LicenseType(final String name) {
+    this.name = name;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public static LicenseType get(final String name) {
+    return ENUM_MAP.getOrDefault(name.toLowerCase(), LicenseType.UNKNOWN);
```
RomanJRW, line 39: this is handled in two places, the defaulting to Unknown. It feels like the design/usage is wrong

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Camunda licenses will have a license type property, fetch that out of the license and return
+   * the value
+   *
+   * <p>Self-managed mode is any other possibility. (ex, blank license, prop missing, etc)
+   */
+  private LicenseType getLicenseTypeFromProperty(final String licenseStr) {
+    try {
+      final LicenseKey licenseKey = getLicenseKey(licenseStr);
```
RomanJRW, line 68: this is all duplicated no? We do the exact same in the method below, including the error logging etc.

RomanJRW (changes requested):
A little confused by some of the changes here, we can discuss sync if it's easier/faster too

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: `@ VisibleForTesting` is a Zeebe annotation that's used to mark methods that have a non-`private` level of visibility because of testing reasons

Since Optimize cannot use any zeebe dependencies, I commented it out. This class is a temporary copy of the monorepo implementation of `CamundaLicense`

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: The automated style perfers the multi-line solution here, its not accepting one line 🤷

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: I reworded the method description here, I think its more clear

optimize/backend/pom.xml
matthewBearCamunda: Yeah, this is the version that was placed in a public repo

optimize/pom.xml
matthewBearCamunda: I thought it would be that too, but `camunda-cpm` is used for this repository in the monorepo, `camunda-bpm` is not used

optimize/pom.xml
matthewBearCamunda: The BPM repo is public and is where the license checker library lives. The previous repo for the license checker was not public and people could not pull down the dependency

Placing it in the bpm repo makes that checker available to download

service/src/main/java/io/camunda/service/license/LicenseType.java
matthewBearCamunda: The `service` implementations are the ones that should be kept around in the future, the Optimize implementations are the duplicates that will be removed when Optimize is placed in the single application architecture

service/src/main/java/io/camunda/service/license/LicenseType.java
matthewBearCamunda: Correct, `unknown` is the error state. No license has been provided, the license is invalid, or an error occured when processing the license

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: Since this class is a copy of the monorepo implementation I kept the annotations to try to make the copy as close as possible to the original. I'll get these removed and add a blurb that the class will need to be deleted in the future

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: I've refactored how the two methods work, its now all in one

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: I've refactored, and this method was combined with the other. The name of this method will no longer exists. It is now a `void` method, `validateLicense()`

optimize/backend/src/main/java/io/camunda/optimize/license/LicenseType.java
matthewBearCamunda: I consolidated the code into the `get()` method here

service/src/main/java/io/camunda/service/license/CamundaLicense.java
matthewBearCamunda: I like the idea of enums now that I have it implemented, its guaranteeing that valid types are going to be used, and all else gets bucketed into `unknown`

Since this string is used in multiple places, I'm inclined to keep it here

service/src/main/java/io/camunda/service/license/LicenseType.java
matthewBearCamunda: I'm getting more safety from the `get()` method I implemented. `Enum.valueOf()` is case-sensitive, whereas the method I implemented can handle any case. The `get()` method can also handle the case of `null` and will return an `unknown` license type

service/src/main/java/io/camunda/service/license/LicenseType.java
RomanJRW: Ok understood 👍 
Up to you, but perhaps constants are a better fit than an enum here then

optimize/backend/src/main/java/io/camunda/optimize/license/LicenseType.java
```diff
@@ -0,0 +1,51 @@
...
+  UNKNOWN("unknown");
+  private static final Map<String, LicenseType> ENUM_MAP;
+
+  static {
+    final Map<String, LicenseType> map = new HashMap<>();
+    for (final LicenseType instance : LicenseType.values()) {
+      map.put(instance.getName().toLowerCase(), instance);
+    }
+    ENUM_MAP = Collections.unmodifiableMap(map);
+  }
```
RomanJRW, line 33: I feel like all of this is redundant and could be better handled with a much simplified `get` implementation. Example:

```
public static LicenseType get(final String name) {
  Arrays.stream(LicenseType.values())
    .filter(e -> e.name().equalsIgnoreCase(dayName)).findAny().orElse(LicenseType.UNKNOWN);
}
```

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
```diff
@@ -114,7 +117,7 @@ public void shouldReturnProperValidValuesWhenLicenseIsSaas() throws InvalidLicen
     testLicense.initializeWithLicense(TEST_LICENSE);
 
     // then
-    assertEquals(SAAS_LICENSE_TYPE, testLicense.getLicenseType());
+    assertEquals(LicenseType.SAAS, testLicense.getLicenseType());
```
RomanJRW, line 120: Is this not a breaking API change? Previously, we would get a lowercased string, as per the API specification. Now with these enums, would it not be the upper-case? Please correct me if I am wrong here

RomanJRW (changes requested):

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public class CamundaLicense {
+
+  public static final String CAMUNDA_LICENSE_ENV_VAR_KEY = "CAMUNDA_LICENSE_KEY";
+  private static final Logger LOGGER = LoggerFactory.getLogger(CamundaLicense.class);
+  private boolean isValid;
+  private LicenseType licenseType;
+  private boolean isInitialized;
+
+  protected CamundaLicense() {}
```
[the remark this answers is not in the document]
matthewBearCamunda: Not for the Optimize implementation. It's needed for the monorepo implmentation for testing. I've removed it here

optimize/backend/src/main/java/io/camunda/optimize/license/LicenseType.java
matthewBearCamunda: I see what you're doing here, and agree - this is simpler. The `ENUM_MAP` here is a overkill for something that isn't called a whole lot and doesn't have a lot of `enum` values

service/src/test/java/io/camunda/service/license/CamundaLicenseTest.java
matthewBearCamunda: A lower case is still returned in the API. If you look at `LicenseController`, its setting the returned value with the `getName()` method which is returning the lower-case representation of the enum

`LicenseControllerTest` is also testing that a lower case is being returned

RomanJRW: Thanks for explaining! 👍

matthewBearCamunda:
I changed the logging levels for license errors from `error` to `warn` to address this: https://github.com/camunda/camunda/issues/21566

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -0,0 +1,95 @@
...
+    initializeWithLicense(license);
+  }
+
+  public synchronized boolean isValid() {
+    return isValid;
+  }
+
+  public synchronized LicenseType getLicenseType() {
+    return licenseType;
+  }
```
RomanJRW, line 41: out of interest, why do these need to be `synchronized`?

RomanJRW (approved):
LGTM 🚀

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: This was actually a weird one for me to figure out. I was getting checkstyle errors in the `initializeWithLicense()` method related to deadlocks

I played around with the method to try and solve it, but the issue was that since these two methods are being called by `initializeWithLicense()` they need to be `synchronized` as well

PR #21201, opened 2024-08-15
Expose stream clock via partitions actuator to ease debugging

npepinpe:
## Description

This PR exposes the stream clock (immutably, aka in a thread safe manner) via the partitions actuator. This then allows anyone to quickly debug if everything is working as expected when they make clock modifications.

As part of it, I migrated the old broker admin service integration test to use our new testing infrastructure, which is closer to "production" usage.

I also fixed one issue where I forgot to allow the clock commands to be accepted by the command API.

## Related issues

related to #21068


lenaschoenburg (approved):
Nice, I like the idea of  `ImmutableStreamClock` :+1:

Issue #19176, opened 2024-06-07
[CPT] I can trigger a BPMN timer event


saig0:
## Description

In Camunda Process Test, I want to trigger a BPMN timer event. Assuming that a timer event has a duration of 1 hour, I want to trigger the timer event immediately instead of waiting until the due date is reached.  

## Tasks
- Provide a test API to trigger a BPMN timer event  

requires #18997   
requires #18999 
related to #18815

saig0:
## Solution idea

- Find the timer entity related to the given BPMN timer event (selected by process instance key and the element name/id)
- Modify the timer's due date to the current time
- As a result, the engine triggers the BPMN timer event

## Requirements

- New API to modify the due date: https://github.com/camunda/camunda/issues/18815
- New API to query timer entities

marcellobarile:
To confirm, this task requires https://github.com/camunda/camunda/issues/18815, which isn't targeted for 8.10. Is this a mismatch, or are we not committed?

saig0:
@marcellobarile you're right. Currently, this issue is not committed for 8.10, but it is a potential issue. We will revisit the target version during planning. If we decide on it, we will contribute the API.

PR #21203, opened 2024-08-16
[CPT] I can modify the clock for process tests

saig0:
## Description

Allows modifying the clock for process tests. This enables the common use case to trigger a BPMN timer event without waiting until its due date is reached. 

The changes don't fulfill #19176 completely because it doesn't trigger a single BPMN timer event explicitly (and isolated). However, it's a common feature that could be used also for other purposes, for example, job backoff timeout and message TTL. 

Verify the behavior in integration tests only. The implementation is simple and I don't expect changes in the clock API. 

## Related issues

related to #19176


remcowesterhoud (approved):
Nothing to add. Really like how CPT is turning out so loosely coupled to the engine. This makes it way more maintainable compared to ZPT 😄

PR #21204, opened 2024-08-16
Don't build frontends during the zeebe version compatibility checks

remcowesterhoud:
## Description

We've seen these checks fail due to failures building the frontend. The check doesn't require any frontends so we can skip building these and prevent such failures in the future. Added bonus, it's less CI time, thus also less CI costs.

## Checklist

- [ ] for CI changes:

## Related issues

N/A

### DIFF
--- a/.github/workflows/zeebe-version-compatibility.yml
+++ b/.github/workflows/zeebe-version-compatibility.yml
@@ -43,6 +43,8 @@ jobs:
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
       - uses: ./.github/actions/build-zeebe
+        with:
+          maven-extra-args: -PskipFrontendBuild
       - name: Test rolling update
         run: >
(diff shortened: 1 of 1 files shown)


.github/workflows/zeebe-version-compatibility.yml
```diff
@@ -43,6 +43,8 @@ jobs:
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
       - uses: ./.github/actions/build-zeebe
+        with:
+          maven-extra-args: -T1C -PskipFrontendBuild
```
lenaschoenburg, line 47: :upside_down_face: `-T1C` is the default already

lenaschoenburg (approved):
Thanks! :+1:

.github/workflows/zeebe-version-compatibility.yml
remcowesterhoud: I just copy-pasted it from another workflow 😄 But you're right, I'll remove it.

PR #21213, opened 2024-08-16
Avoid concurrent database writes in `MigrateProcessInstanceConcurrentNoBatchingTest`

### LINKED ISSUE #21196: `MigrateProcessInstanceConcurrentNoBatchingTest` is flaky
**Summary**

- How often does the test fail? at least [once](https://github.com/camunda/camunda/actions/runs/10405955810/job/28817973356)
- Does it block your work? it may block merges
- Do we suspect that it is a real failure? no, it's a concurrent modification in rocksdb by the test

**Failures**

<pre>
Error:  io.camunda.zeebe.engine.processing.processinstance.migration.MigrateProcessInstanceConcurrentNoBatchingTest.shouldRejectMigrationForActiveCatchEventAttachedToEventBasedGateway -- Time elapsed: 0.120 s <<< ERROR!
java.lang.RuntimeException: Unexpected error occurred during RocksDB transaction.
	at io.camunda.zeebe.db.impl.rocksdb.transaction.DefaultTransactionContext.runInTransaction(DefaultTransactionContext.java:42)
	at io.camunda.zeebe.db.impl.rocksdb.transaction.TransactionalColumnFamily.ensureInOpenTransaction(TransactionalColumnFamily.java:343)
	at io.camunda.zeebe.db.impl.rocksdb.transaction.TransactionalColumnFamily.get(TransactionalColumnFamily.java:139)
	at io.camunda.zeebe.stream.impl.state.NextValueManager.getCurrentValue(NextValueManager.java:63)
	at io.camunda.zeebe.stream.impl.state.NextValueManager.getCurrentValue(NextValueManager.java:54)
	at io.camunda.zeebe.stream.impl.state.NextValueManager.getNextValue(NextValueManager.java:38)
	at io.camunda.zeebe.stream.impl.state.DbKeyGenerator.nextKey(DbKeyGenerator.java:41)
	at io.camunda.zeebe.engine.processing.processinstance.migration.MigrateProcessInstanceConcurrentNoBatchingTest.shouldRejectMigrationForActiveCatchEventAttachedToEventBasedGateway(MigrateProcessInstanceConcurrentNoBatchingTest.java:982)
Caused by: org.rocksdb.RocksDBException: WriteBatch has wrong count
	at org.rocksdb.Transaction.commit(Native Method)
	at org.rocksdb.Transaction.commit(Transaction.java:216)
	at io.camunda.zeebe.db.impl.rocksdb.transaction.ZeebeTransaction.commitInternal(ZeebeTransaction.java:147)
	at io.camunda.zeebe.db.impl.rocksdb.transaction.DefaultTransactionContext.runInNewTransaction(DefaultTransactionContext.java:62)
	at io.camunda.zeebe.db.impl.rocksdb.transaction.DefaultTransactionContext.runInTransaction(DefaultTransactionContext.java:33)
	... 8 more
</pre>
</details>

**Hypotheses**

The test case modifies RocksDB by asking for a new key from the keygenerator.

https://github.com/camunda/camunda/blob/d7f9b87f79454b43beb59eb2b685f40224336552/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateProcessInstanceConcurrentNoBatchingTest.java#L979-L982

**Logs**

korthout:
## Description

This unflakes `MigrateProcessInstanceConcurrentNoBatchingTest` by pausing the processing before generating a record key.

This test needs to generate a record key for use in some events that it appends to the log as part of the set up of this test case.

Generating a next key with `DbKeyGenerator` is not a safe operation for use in tests. It reads and changes a value in the state. The state is also accessed concurrently by the stream processor. This can lead to errors in committing the database transaction when both are trying to do so simultaneously.

We can avoid this by pausing processing before generating the key.

An alternative approach would be to stop the engine first (which we do after generating the key so we can write events that are applied during replay), but this is not possible. If the engine is stopped, the state is inaccessible.

## Related issues

closes #21196

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateProcessInstanceConcurrentNoBatchingTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateProcessInstanceConcurrentNoBatchingTest.java
@@ -977,6 +977,8 @@ public class MigrateProcessInstanceConcurrentNoBatchingTest {
         .await();
 
     // we generate a key for the intermediate catch event so we can activate it using events
+    // this requires us to pause processing to avoid concurrent database transaction modification
+    ENGINE.pauseProcessing(1);
     final var keyGenerator =
         ((MutableProcessingState) ENGINE.getProcessingState()).getKeyGenerator();
     final var intermediateCatchEventKey = keyGenerator.nextKey();


berkaycanbc (approved):
Nice catch, LGTM! 🚀

Issue #16912, opened 2024-03-15
🎯 Export directly to the Elasticsearch Operate indexes


ThorbenLindhauer:
### Description

* In the Zeebe Elasticsearch exporter, transform Zeebe records into Operate entities and write them into the Operate Elasticsearch indexes
* Keep the current exporting behavior, i.e. writing to Zeebe Elasticsearch indexes, as Tasklist and Optimize still require them
* See the [prototype ticket](https://github.com/camunda/operate/issues/6031) and its implementation to get started
* [Product hub epic](https://github.com/camunda/product-hub/issues/2128)

### Included tasks

- [ ] #18304
- [x] #19006
- [ ] https://github.com/camunda/camunda/issues/23190
- [ ] https://github.com/camunda/camunda/issues/22899
- [ ] #20191
- [ ] #19388
- [ ] #18378 (depends on #16915)
- [ ] #18372 (depends on #16916)
- [ ] #18364 
- [ ] #18306
- [ ] #16920
- [ ] #22747

Langleu:
Hey @ThorbenLindhauer,
any updates on the topic for 8.6 and to aim for a specific alpha?

This change would essentially mean that I could easily run Operate in highly available mode since we're moving the impoter to the exporter layer in Zeebe, right?
I looked for similar epics concerning Tasklist, anything we plan there as well or is it a follow up to however this epic goes?

ThorbenLindhauer:
Hi @Langleu,

> any updates on the topic for 8.6 and to aim for a specific alpha?

Due to the complexity of the implementation we are not targeting a particular alpha at the moment. I'd expect this only to be done with alpha5.

> This change would essentially mean that I could easily run Operate in highly available mode since we're moving the impoter to the exporter layer in Zeebe, right?

I'm not sure what our definition of *highly available mode* is, but assuming that it requires us to have a set of pods where no single pods have special roles: I agree, this should largely resolve with removing the importer. Note however two things:

* We still have the [archiver component](https://docs.camunda.io/docs/self-managed/operate-deployment/importer-and-archiver/) that has similar limitations as the importer (can have only one archiver per Zeebe partition on one pod); this may be less critical though when it comes to availability (the archiver stopping for a while is not noticeable for the end user, unlike the importer)
[...]

PR #21297, opened 2024-08-19
feat: add exporting metrics

ChrisKujawa:
## Description

To get a better understanding where time is spent, adding new metrics for the exporting.

 * new metric for exporting, time from record written to exporting
 * new metric for exporter latency, time how long it takes for each exporter to accept an record

This allows us to already visualize latency like this:

Missing:

 * elasticsearch exporter - exporting latency
 * importer latency
 * ...
 
## Related issues

related to https://github.com/camunda/camunda/issues/16912

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
@@ -28,11 +28,11 @@ public final class ExporterMetrics {
           .labelNames(LABEL_NAME_PARTITION, LABEL_NAME_VALUE_TYPE)
           .register();
 
-  private static final Histogram EXPORTER_LATENCY =
+  private static final Histogram EXPORTER_EXPORTING_DURATION =
       Histogram.build()
           .namespace(NAMESPACE_ZEEBE)
-          .name("exporter_latency")
-          .help("Time exporter needs to export certain record (in seconds)")
+          .name("exporter_exporting_duration")
+          .help("The time an exporter needs to export certain record (duration in seconds)")
           .labelNames(LABEL_NAME_PARTITION, LABEL_NAME_EXPORTER, LABEL_NAME_VALUE_TYPE)
           .register();
 
@@ -116,8 +116,11 @@ public final class ExporterMetrics {
         .observe((exporting - written) / 1000f);
   }
 
-  public Histogram.Timer startExportLatencyTimer(final ValueType valueType, final String exporter) {
-    return EXPORTER_LATENCY.labels(partitionIdLabel, exporter, valueType.name()).startTimer();
+  public Histogram.Timer startExporterExportingTimer(
+      final ValueType valueType, final String exporter) {
+    return EXPORTER_EXPORTING_DURATION
+        .labels(partitionIdLabel, exporter, valueType.name())
+        .startTimer();
   }
 
   public void initializeExporterState(final ExporterPhase state) {
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
@@ -10,6 +10,7 @@ package io.camunda.zeebe.broker.exporter.stream;
 import io.camunda.zeebe.protocol.record.ValueType;
 import io.prometheus.client.Counter;
 import io.prometheus.client.Gauge;
+import io.prometheus.client.Histogram;
 
 public final class ExporterMetrics {
 
@@ -18,6 +19,23 @@ public final class ExporterMetrics {
   private static final String LABEL_NAME_ACTION = "action";
   private static final String LABEL_NAME_VALUE_TYPE = "valueType";
   private static final String NAMESPACE_ZEEBE = "zeebe";
+
+  private static final Histogram EXPORTING_LATENCY =
+      Histogram.build()
+          .namespace(NAMESPACE_ZEEBE)
+          .name("exporting_latency")
+          .help("Time between a record is written until it is picked up for exporting (in seconds)")
+          .labelNames(LABEL_NAME_PARTITION, LABEL_NAME_VALUE_TYPE)
+          .register();
+
+  private static final Histogram EXPORTER_LATENCY =
+      Histogram.build()
+          .namespace(NAMESPACE_ZEEBE)
+          .name("exporter_latency")
+          .help("Time exporter needs to export certain record (in seconds)")
+          .labelNames(LABEL_NAME_PARTITION, LABEL_NAME_EXPORTER, LABEL_NAME_VALUE_TYPE)
+          .register();
+
   private static final Counter EXPORTER_EVENTS =
       Counter.build()
           .namespace(NAMESPACE_ZEEBE)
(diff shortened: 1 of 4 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/RecordExporter.java
```diff
@@ -52,18 +54,25 @@ boolean export() {
       return true;
     }
 
+    final ValueType valueType = typedEvent.getValueType();
+    exporterMetrics.exportingLatency(
+        valueType, typedEvent.getTimestamp(), ActorClock.currentTimeMillis());
```
lenaschoenburg, line 59: :x: this must use the new `StreamClock` to prevent wrong metrics in case the engine clock was modified. This clock is already available (as an `InstantSource`) in the `ExporterDirector`, you just need to forward it so that we can use it here instead of `ActorClock`.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/RecordExporter.java
```diff
@@ -52,18 +54,25 @@ boolean export() {
       return true;
     }
 
+    final ValueType valueType = typedEvent.getValueType();
+    exporterMetrics.exportingLatency(
```
lenaschoenburg, line 58: :wrench: Shouldn't we record the latency _after_ it is exported? Or do you really want to measure the time between writing until _starting_ to export?

lenaschoenburg: Also, this would measure repeatedly when exporting is retried. More reason to move this to _after_ the record was successfully exported.

lenaschoenburg (changes requested):
:upside_down_face: Would have been a good chance to refactor to use Micrometer instead. The registry is already available.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
```diff
@@ -18,6 +19,23 @@ public final class ExporterMetrics {
...
+
+  private static final Histogram EXPORTING_LATENCY =
+      Histogram.build()
+          .namespace(NAMESPACE_ZEEBE)
+          .name("exporting_latency")
+          .help("Time between a record is written until it is picked up for exporting (in seconds)")
+          .labelNames(LABEL_NAME_PARTITION, LABEL_NAME_VALUE_TYPE)
+          .register();
+
+  private static final Histogram EXPORTER_LATENCY =
```
deepthidevaki, line 31: :wrench:  The term `latency` is more suitable to represent queuing time. So here it is better to use a different term, preferably `duration`. So may be something like `RECORD_EXPORTING_DURATION`.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/RecordExporter.java
deepthidevaki: I think the metrics should use System clock, because we need the exact duration. We had this issue in StreamProcessor metrics where the processing duration was always 0 because  the ActorClock is not updated with in a task.

Zelldon: Interesting 🤔 Thanks for pointing out.

Zelldon: Maybe I just use the timer api, instead of calculating on my onw.

lenaschoenburg: I don't think that works, you need to use the record timestamp and compare it to the "current" clock. And the "current" clock needs to be the controllable one you get as `InstantSource` in the director.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/RecordExporter.java
deepthidevaki: :thought_balloon: Calculating latency until the start of the export makes sense to me. But we have to consider retries as well. So may be calculate it before the start, but only record after it is successfully exported?

Zelldon: Yes this was intended. It is similar to the overall processing metric/ panel (in our dashboard)

I can build then something like this 




This allows me to better understand where time is spent, especially when I have the export time separately.

Zelldon:
> 🙃 Would have been a good chance to refactor to use Micrometer instead. The registry is already available.

Yeah, I was expecting this comment. I wanted to focus first on some benchmarks and adding metrics. Likely I will iterate over it, so we can migrate them.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/RecordExporter.java
Zelldon: Ah yeah, was already somewhere else with my head :) to many (benchmarks/metrics) at once.

lenaschoenburg: > I think the metrics should use System clock, because we need the exact duration

Not if we use the record timestamp as reference. Because that timestamp is not given by system time but the controllable clock

Zelldon: BTW I haven't looked into the feature, but is this controllable clock still enhancing after set/moved? Or is it pinned?

lenaschoenburg: The implementation supports pinning and offset but the rest (like state, API etc.) will most likely only support pinning.

Zelldon: @lenaschoenburg I was a bit confused. In my branch the clock was not even available. :D I had to rebase now an found it. 👍🏼 

I used it in the record exporter now. Be aware I still collect the metric before, so even if it is failing as otherwise I might have no metric at all. In this case I increasing latency which is what I want. I added a comment in code as well.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/RecordExporter.java
Zelldon: > But we have to consider retries as well. So may be calculate it before the start, but only record after it is successfully exported?

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
Zelldon: Yes I was also unsure during writing tbh. Found a good/simple enough definition: `latency is how long it takes to start the behavior and duration is how long the behavior lasts.`

Zelldon:
Updated cutout from the latency panels of my dashboard

lenaschoenburg (approved):

Zelldon:
@lenaschoenburg @deepthidevaki any thoughts regarding backporting?

lenaschoenburg:
If you have the patience for it :)
Except for the clock bit, it should be pretty straight-forward.

PR #21298, opened 2024-08-19
Identity/21048 process user update command

### LINKED ISSUE #21048: Process `User.UPDATE` command
### Description

- [x] Extend `UserIntent` with an `UPDATE` intent (ordering-wise it's nice to put it above the UPDATED event)
- [x] Create a `UserUpdateProcessor`
    - Very similar to the `UserCreateProcessor`
    - Verify a user with this username exists:
        - [x]  If not, write a command rejection with a proper error message indicating the user doesn't exist
        - [x] If yes, write a `UserIntent.UPDATED` event and distribute the command
    - [x] In the `processDistributedCommand` just write the `UserIntent.UPDATED` event and acknowledge the command
- [x] Write tests that verify processing the command works as expected. Verify the rejection, a successful update, and the distribution acknowledgement.

Ben-Sheppard:
## Description

This PR adds the next part of the user update flow which is processing the user update intent.

## Related issues

closes #21048


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/UpdateUserMultiPartitionTest.java
```diff
@@ -0,0 +1,99 @@
...
+                .setEmail("foo@bar.com")
+                .setPassword("password"))
+        .withUpdatedName("Bar Foo")
+        .withUpdatedEmail("bar@foo.com")
+        .update();
+
+    assertThat(
+            RecordingExporter.records()
+                .withPartitionId(1)
+                .limit(record -> record.getIntent().equals(CommandDistributionIntent.FINISHED)))
```
remcowesterhoud, line 64: ❌ This is the reason the test is flaky. What we say is here is to get all the records up until the first record with the intent `CommandDistributionIntent.FINISHED`. As we create the user before the update, and a user creationg also gets distributed this isn't correct. We must wait until found a second record with this intent. We can easily do this with:

```suggestion
                .limitByCount(record -> record.getIntent().equals(CommandDistributionIntent.FINISHED), 2))
```

After this it may still be flaky. Ideally we wait after the user creation and before the update that the distribution of the create command is finished:

```java
    RecordingExporter.commandDistributionRecords(CommandDistributionIntent.FINISHED)
        .withDistributionIntent(UserIntent.CREATE)
        .await();
```

remcowesterhoud (changes requested):
Thanks for the changes @Ben-Sheppard. I have a few comments that should be easy to fix 🙂

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserUpdateProcessor.java
```diff
@@ -0,0 +1,70 @@
...
+      final var rejectionMessage =
+          "Expected to update user with username %s, but a user with this username does not exist"
+              .formatted(command.getValue().getUsername());
+
+      rejectionWriter.appendRejection(command, RejectionType.NOT_FOUND, rejectionMessage);
+      responseWriter.writeRejectionOnCommand(command, RejectionType.NOT_FOUND, rejectionMessage);
+      return;
+    }
+
+    stateWriter.appendFollowUpEvent(command.getKey(), UserIntent.UPDATED, command.getValue());
```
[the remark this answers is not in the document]
Ben-Sheppard: I think this actually makes sense especially given the idea that we will use a PATCH approach for the update. I have two questions:
1. What are reasonable default values
2. With the change to line 45 I assume this is where it handles only updating the values we set and leaving the others?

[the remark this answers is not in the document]
remcowesterhoud: > What are reasonable default values

I'd go with empty strings. In the REST API we should validate that if these properties are provided, they should be non-empty. So if we make the default empty we know the user didn't provide the values.

> With the change to line 45 I assume this is where it handles only updating the values we set and leaving the others?

Line 45 gets the existing record from the state. That's good and shouldn't change. But yes, after this we must check which properties were provided in the command and set those on that existing record.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserUpdateProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+      return;
+    }
+
+    final var updatedUser = overlayUser(persistedUser, command.getValue());
+
+    final long key = keyGenerator.nextKey();
+    stateWriter.appendFollowUpEvent(key, UserIntent.UPDATED, updatedUser);
+    responseWriter.writeEventOnCommand(key, UserIntent.UPDATED, updatedUser, command);
+
+    distributionBehavior.distributeCommand(command.getKey(), command);
```
remcowesterhoud, line 67: ❌ Please use the generated key here as well. The command key is set to `-1L`. Command distribution won't work with that 🙂

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserUpdateProcessor.java
```diff
@@ -0,0 +1,92 @@
...
+    stateWriter = writers.state();
+    rejectionWriter = writers.rejection();
+    responseWriter = writers.response();
+    this.distributionBehavior = distributionBehavior;
+  }
+
+  @Override
+  public void processNewCommand(final TypedRecord<UserRecord> command) {
+    final var username = command.getValue().getUsernameBuffer();
+    final var persistedUser = userState.getUser(username);
```
remcowesterhoud, line 49: 🧠 My goal in the end would be to work with keys instead of usernames. In general keys are used as identifiers throughout Zeebe and I'd prefer to keep it consistent. This means that in the command we'd need a `userKey`, which of course gets generated by Zeebe upon user creation. The sender of this command should consume the exported `User.CREATE` record to know what the `userKey` is.

If you want we can follow up on that later in a separate PR.

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/user/UserRecord.java
```diff
@@ -17,10 +17,10 @@
 import org.agrona.DirectBuffer;
 
 public final class UserRecord extends UnifiedRecordValue implements UserRecordValue {
-  private final StringProperty usernameProp = new StringProperty("username");
-  private final StringProperty nameProp = new StringProperty("name");
-  private final StringProperty emailProp = new StringProperty("email");
-  private final StringProperty passwordProp = new StringProperty("password");
+  private final StringProperty usernameProp = new StringProperty("username", "");
```
remcowesterhoud, line 20: ❌ I don't think username should be optional. As-is we can't do anything without having a username set, as it's the current identifier for a user.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/UpdateUserMultiPartitionTest.java
remcowesterhoud: ❌ I saw you've added the await, which is good. But you forgot to change the `limit` to `limitByCount`. Without this the test is still flaky 😉

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/UpdateUserTest.java
```diff
@@ -0,0 +1,79 @@
...
+        ENGINE
+            .user()
+            .updateUser(username)
+            .withName("Bar Foo")
+            .withEmail("foo@bar.blah")
+            .withPassword("Foo Bar")
+            .update();
+
+    final var createdUser = updatedUser.getValue();
+    assertThat(createdUser)
```
remcowesterhoud, line 50:
```suggestion
    final var updatedUser =
        ENGINE
            .user()
            .updateUser(username)
            .withName("Bar Foo")
            .withEmail("foo@bar.blah")
            .withPassword("Foo Bar")
            .update()
            .getValue();

    assertThat(updatedUser)
```

🔧 The naming is a bit strange as it reads as if we're asserting over the createdUser instead of the updatedUser.

remcowesterhoud (changes requested):
Thanks for the changes @Ben-Sheppard! I've got a few more comments but it's already looking good!

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/UpdateUserTest.java
Ben-Sheppard: Its definitely very strange, it should be the updated user :D

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserUpdateProcessor.java
Ben-Sheppard: I think I'd be keen to follow up in a smaller and more focused PR for that change

remcowesterhoud (approved):
LGTM! 🚀

PR #21311, opened 2024-08-19
refactor: record value and metadata length is always non-zero

### LINKED ISSUE #19225: `Sequencer#isEntryValid` is slow because it has to calculate the full record size
This method checks that record value and metadata don't have size 0. This validation was copied over from the dispatcher.
The method shows up in profiling because calculating the length of `UnifiedRecordValue` is a fairly expensive operation. The length is calculated by recursively walking the object graph until we find primitive types such as `long`.

We should be able to remove this checks because `RecordMetadata` can never have length 1 and `UnifiedRecordValue` also offers an optimized `isEmpty` method.

lenaschoenburg:
This is a performance improvement to avoid calculating the length of record value and metadata for every entry that's being appended.

Record values inherit from `ObjectValue` which always has non-zero length due to the map header.
Record metadata is serialized via SBE which also requires a header and thus always has non-zero length.

closes #19225


deepthidevaki (approved):
:rocket:

PR #21313, opened 2024-08-19
ci: ensure independent workflows on main don't cancel each other

cmur2:
## Description

This is a fixup to https://github.com/camunda/camunda/pull/20483 which was reported in https://camunda.slack.com/archives/C071KP5BTHB/p1723798638842329 to have unintended side-effects due to a faulty new `concurrency` configuration for `main` branch. The missing piece here was that the new `concurrency` configuration works only if used by only one workflow at a time since it doesn't discriminate between workflow names.

Example for this is https://github.com/camunda/camunda/actions/runs/10450995603 which says `Canceling since a higher priority waiting request for '53d5cfaf9b1473bb3c300d150fae1260222944f2' exists` so no workflow name was involved that would allow distinction.

This PR adds that distinction to make sure that on `main` the concurrency group will always be `workflow_name-github_sha` which should be unique and still (as intended with #20483) allow all workflows to run til their end.

## Checklist

- [x] for CI changes:

## Related issues

None

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ on:
 # Exception for deploy jobs that have to wait for each other to avoid overwriting
 concurrency:
   cancel-in-progress: true
-  group: ${{ github.ref == 'refs/heads/main' && github.sha || format('{0}-{1}', github.workflow, github.ref) }}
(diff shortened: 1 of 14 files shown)


Langleu (approved):
nice work 🚀

Issue #16029, opened 2024-01-22
CI: Performance build fails


Zelldon:
[Build fails in the journal module ](https://github.com/camunda/zeebe/actions/runs/7612190618/job/20729263287)with:

```
Error:  Failed to execute goal org.codehaus.mojo:exec-maven-plugin:3.1.1:java (default) on project zeebe-journal: An exception occurred while executing the Java class. Message header not found in map for key: messageHeader -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.codehaus.mojo:exec-maven-plugin:3.1.1:java (default) on project zeebe-journal: An exception occurred while executing the Java class. Message header not found in map for key: messageHeader
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:193)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:180)
    at java.util.concurrent.FutureTask.run (FutureTask.java:317)
    at java.util.concurrent.Executors$RunnableAdapter.call (Executors.java:572)
    at java.util.concurrent.FutureTask.run (FutureTask.java:317)
    at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1144)
[...]

PR #21320, opened 2024-08-19
bugfix: set a limit to evaluationFailureMessage size

### LINKED ISSUE #20248: Exporting decision evaluation records breaks with large `evaluationFailureMessage` values
**Describe the bug**

When decision evaluation fails with a long error message, Zeebe fails to export the record to ES. It get's rejected because the `evaluationFailureMessage` is supposed to be parsed as `keyword` but ElasticSearch fails to do so.
The impact is that exporting is fully blocked until a temporary workaround is applied.

**To Reproduce**

Use a large FEEL expression that fails for some input. The automatically generated error message includes the entire FEEL expression and more, making it too large for ES to handle.

**Workaround**

Modify the template ad-hoc:
```
PUT zeebe-record-decision-evaluation/_mapping
{
  "properties": {
    "value.evaluationFailureMessage": {
      "type": "keyword",
      "ignore_above": 1024
    }
  }
}
```

**Expected behavior**

Exporting should never fail. Comparable fields such as `exceptionMessage` from the error record are configured to ignore values above a certain size.

**Log/Stacktrace**

 <p>

```
io.camunda.zeebe.exporter.ElasticsearchExporterException: Failed to flush bulk request: [Failed to flush 1 item(s) of bulk request [type: document_parsing_exception, reason: [1:99] failed to parse field [value.evaluationFailureMessage] of type [keyword] in document with id '1-1019381702'.
```

</p>
</details>

**Related**

Incident: https://app.incident.io/camunda/incidents/1557

berkaycanbc:
## Description

There was a bug where `evaluationFailureMessage` was failed to be indexed when it goes over `keyword` data type's size. A limit is set to the related field to ignore the remaining part of the keyword when it is above the size 1024. I chose number as 1024 because evaluation failure message typically contains an error from a FEEL expression which is not expected to be too long.

## Related issues

closes #20248

### DIFF
--- a/zeebe/exporters/elasticsearch-exporter/src/main/resources/zeebe-record-decision-evaluation-template.json
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/resources/zeebe-record-decision-evaluation-template.json
@@ -61,7 +61,8 @@
               "type": "long"
             },
             "evaluationFailureMessage": {
-              "type": "keyword"
+              "type": "keyword",
+              "ignore_above": 1024
             },
             "failedDecisionId": {
               "type": "keyword"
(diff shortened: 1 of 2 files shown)


mustafadagher (approved):

korthout:
Error was unrelated to the changes:
- https://github.com/camunda/camunda/issues/16029

PR #21325, opened 2024-08-19
operate: include priority field for user tasks

### LINKED ISSUE #21293: Include Priority field in Operate User Task indices
Operate should import the new Priority field for Zeebe User Tasks in it's indices

panagiotisgts:
## Description

Include priority field in user-task index of Operate for future usage.

## Related issues
part of: https://github.com/camunda/product-hub/issues/217
closes #21293

### DIFF
--- a/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/UserTaskZeebeRecordProcessor.java
+++ b/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/UserTaskZeebeRecordProcessor.java
@@ -97,7 +97,9 @@ public class UserTaskZeebeRecordProcessor {
             "dueDate",
             new Tuple<>(UserTaskTemplate.DUE_DATE, userTaskEntity::getDueDate),
             "followUpDate",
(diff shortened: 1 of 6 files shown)


ralfpuchert (approved):
Thank you @panagiotisgts for contributing this!
Does the new field needs some kind of default value for existing
documents? For that we have migration scripts like this
https://github.com/camunda/camunda/blob/581617f05b90c9327973df7c5f17ff73e11620b5/operate/schema/src/main/resources/schema/elasticsearch/change/8.3.1-1_incident_tenantId.json#L7-L7

panagiotisgts:
> Thank you @panagiotisgts for contributing this! Does the new field needs some kind of default value for existing documents? For that we have migration scripts like this
> 
> https://github.com/camunda/camunda/blob/581617f05b90c9327973df7c5f17ff73e11620b5/operate/schema/src/main/resources/schema/elasticsearch/change/8.3.1-1_incident_tenantId.json#L7-L7

It would not be required to run any migration. The field has been added for future use cases in Operate. Zeebe is responsible for providing the value for this field, where defaults are also set .

PR #21333, opened 2024-08-19
fix: add separate ilm requests for indices

nathansandi:
Closes #https://github.com/camunda/tasklist/issues/5024 

## Description

During the investigation of the incident [#inc-support-23157-tasklist-pod-crashes-continuously](https://camunda.slack.com/archives/C07G9CX5DKL) 

It was discovered the follow bug:
- When archiver is enabled for such a time and indexes was already created...
- after ilm is enabled
- The request for apply ILM updates on OpenSearch does not work

The reason:
- Updates for templates settings on OpenSearch uses one type of request
- Updates for index settings on Opensearch uses different ones


panagiotisgts (approved):
LGTM

PR #21350, opened 2024-08-20
Allways send user create command to deployment partition

remcowesterhoud:
## Description

At first I thought this was not neccessary. However, with the current state of Zeebe it's better to have one partition be the single source of truth for user management. If we don't we could run into conflicts due to the distribution of the commands.
For example, if 2 user create commands are sent shortly after another it could occur that they are both processed on a different partition. If the processing happens before the distribution both partition will accept the command, create a user and respond to the client that creation was successful. However, distribution will get stuck as both partitions cannot create the user from the other partition, as a user with this specific username already exists in the state.

## Checklist

- [ ] for CI changes:

## Related issues

Relates to #20398

### DIFF
--- a/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserCreateRequest.java
+++ b/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserCreateRequest.java
@@ -8,6 +8,7 @@
 package io.camunda.zeebe.gateway.impl.broker.request;
 
 import io.camunda.zeebe.broker.client.api.dto.BrokerExecuteCommand;
+import io.camunda.zeebe.protocol.Protocol;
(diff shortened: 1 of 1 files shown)


remcowesterhoud:
@Ben-Sheppard FYI, I made a mistake during the user creation PR review. We do need to always create users on the deployment partition. See the PR description as to why this is 🙂

Ben-Sheppard (approved):
Thank you for sorting this and thank you for the context as to why, is this also the case with other actions i.e. updates and deletes?

remcowesterhoud:
> Thank you for sorting this and thank you for the context as to why, is this also the case with other actions i.e. updates and deletes?

For Update and Delete we should work with a userKey (eg the key that Zeebe creates for the User). In this key we have encoded the partition the user is created on and we can use that by setting it as `key` on the broker request. Let's discuss it in our meeting today.

PR #21351, opened 2024-08-20
fix: request timeout property accept millis

nicpuppa:
## Description

Property `requestTimeout` should accept millis NOT seconds


clients/java/src/test/java/io/camunda/zeebe/client/ZeebeClientTest.java
```diff
@@ -957,4 +958,19 @@ public void shouldOverrideDefaultRetryPolicyWithProperty() {
     // then
     assertThat(builder.useDefaultRetryPolicy()).isFalse();
   }
+
+  @Test
+  public void shouldSetTimeoutInMillis() {
```
megglos, line 963: ❤️

megglos (approved):
🚀

PR #21360, opened 2024-08-20
Migrate intermediate timer catch events

### LINKED ISSUE #20911: Migrate intermediate timer catch events
It should be possible to migrate timer intermediate catch events.

Timer intermediate catch events can be migrated by providing a mapping instruction. Only active timer intermediate catch events can be migrated.

Screenshot 2024-08-08 at 15 22 24

berkaycanbc:
## Description

With this PR, migration of active timer catch events will be supported. Since we already support migrating timer events in the migration processor with `handleCatchEvents()` method, no additional change is needed in the processor.

## Related issues

closes #20911


korthout (approved):
Nice work @berkaycanbc 🚀 

👍 Easy to review

🔧 Let's add a test case to `MigrateIntermediateCatchEventTest` that shows that the timer can trigger after the migration, similar to what `shouldCorrelateMessageSubscriptionAfterMigration()` does for message correlation.

LGTM 👍

berkaycanbc:
> 🔧 Let's add a test case to MigrateIntermediateCatchEventTest that shows that the timer can trigger after the migration, similar to what shouldCorrelateMessageSubscriptionAfterMigration() does for message correlation.
@korthout it is already there https://github.com/camunda/camunda/pull/21360/files#diff-d6de90060f540671ab7ded1e8ad58e0aa76910fad21b73ebe013cee45628f559R255-R268 👀

PR #21363, opened 2024-08-20
refactor: remove CDN font-path

pedesen:
## Description

This PR removes `fonts.camunda.io` as font resolver for `carbon/react`. Instead fonts like IBM Plex will now be resolved with a bundled version of the font. This ensures that the correct font is loaded even if Operate runs in self managed setups without internet access.

This also fixes an issue with resolving the IBM Plex font correctly after a [dependency update](https://github.com/camunda/camunda/pull/21287/files#diff-ab72112d545f504654de5bfd7703fdb6ea4528138712f46c0f22b554026ace97R1334) of `carbon/styles` to version 1.63.0. In this version Carbon changed the way how the IBM Plex font is imported (see https://github.com/carbon-design-system/carbon/pull/17025), which broke the font resolution to `fonts.camunda.io`. This caused rendering a fallback font which then caused all visual regression fonts to fail. 

~~In this PR there are still some visual regression snapshots updated. These are related to minor changes like:~~

- ~~updated background color of date picker input fields~~
- ~~minor change to the IBM Plex font itself~~

After rebasing on `stable/operate-8.5` these problems resolved itself.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/operate/client/src/index.scss
+++ b/operate/client/src/index.scss
@@ -5,10 +5,7 @@
  * except in compliance with the proprietary license.
  */
 
-@use '@carbon/react' as * with (
-  $font-path: 'https://fonts.camunda.io'
-);
-
+@use '@carbon/react' as *;
 @use '@carbon/react/scss/themes';
 @use '@carbon/react/scss/theme';
 


operate/client/e2e-playwright/visual/decisions.spec.ts-snapshots/decisions-page-optional-filters-visible---light-1-chromium-linux.png
vsgoulart, line 1: Something broke the layers here which is not related to the fonts

operate/client/src/index.scss
```diff
@@ -5,10 +5,7 @@
  * except in compliance with the proprietary license.
  */
 
-@use '@carbon/react' as * with (
-  $font-path: 'https://fonts.camunda.io'
-);
-
+@use '@carbon/react' as *;
```
vsgoulart, line 8: The correct config is like this: https://github.com/carbon-design-system/carbon/blob/main/examples/vite/src/index.scss

vsgoulart: For Tasklist it's the opposite, it doesn't work if we don't like that. I guess it's because of Vite

We'll have to change this when we finally manage to migrate Operate

vsgoulart (approved):
@pedesen Shouldn't we do it on `main` too?

Another question is where did this request come from? Should we do this fix on Tasklist too?

pedesen:
> @pedesen Shouldn't we do it on `main` too?

Yes, I think so.

> Another question is where did this request come from? Should we do this fix on Tasklist too?

It looks like it got fixed in another dependency update (see [visual regression workflows runs](https://github.com/camunda/camunda/actions/workflows/operate-playwright.yml?query=branch%3Astable%2Foperate-8.5)). So I don't see a strong need for Tasklist to fix this.

I still like to merge this PR, because I think bulding the font is the better way, because it shows the correct font even in sealed self managed environments without internet access.

vsgoulart:
@pedesen can we merge this?

pedesen:
@vsgoulart Yes, I somehow missed this.

PR #21375, opened 2024-08-20
Enable TLS Configuration for gRPC server with keystore file

EuroLew:
## Description

This will also configure the gRPC server with TLS using the keystore (PKCS12 format) file, there has also been minor refactoring since both the broker TLS config and gateway TLS config share code (in that they both validate the config to be mutually exclusive between the new keystore file and old key + cert chain file, and both require retrieval of private keys and cert chains from the keystore file when configuring the ssl context for netty).

## Related issues

relates to https://github.com/camunda/issues/issues/833


zeebe/util/src/main/java/io/camunda/zeebe/util/TlsConfigUtil.java
```diff
@@ -0,0 +1,107 @@
...
+    if ((certificateChain != null || privateKey != null) && keyStore != null) {
+      throw new IllegalArgumentException(
+          String.format(
+              """
+                      Cannot provide both separate certificate chain and or private key along with a
+                      keystore file, use only one approach.
+                      certificateChainPath: %s
+                      privateKeyPath: %s
+                      OR
+                      keyStorePath: %s""",
```
npepinpe, line 32: :wrench: Something I missed last time, we have some conventions for error messages: https://github.com/camunda/camunda/wiki/Error-Guidelines

So this would be reworded to something like `Expected to configure with a certificate and private key pair, or with a key store and password, but both were provided. Please select only one approach.`

zeebe/util/src/main/java/io/camunda/zeebe/util/TlsConfigUtil.java
```diff
@@ -0,0 +1,107 @@
...
+import java.io.FileInputStream;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.X509Certificate;
+import java.util.Arrays;
+
+public class TlsConfigUtil {
```
npepinpe, line 20: :wrench: Just a heads up, since utility classes aren't meant to be instantiated, you would typically also make it final, and provide a private default constructor. This makes the intent of the class clearer (i.e. just a collection of utility methods).

npepinpe (approved):
:rocket: No need for another review, but please address comments before merging.

- :clown_face: I'm not big on utility classes in general, and I wish we would've found a better way to share this configuration, especially since a lot of the properties themselves are also duplicated/shared. But I also don't have a really strong alternative to suggest OTOH, so let's go with that. 
- :question: I forgot to ask, can I use a passwordless key store with this? I assume not right now, and it might create a NullPointerException, right? If effort is < 2 hours to add support for passwordless, then let's do it as a follow up PR. If not, then let's at least not throw a NullPointerException but instead a proper exception if no password is given in the validation. :thinking:

zeebe/util/src/main/java/io/camunda/zeebe/util/TlsConfigUtil.java
EuroLew: done

EuroLew:
> 🚀 No need for another review, but please address comments before merging.
> 
> * 🤡 I'm not big on utility classes in general, and I wish we would've found a better way to share this configuration, especially since a lot of the properties themselves are also duplicated/shared. But I also don't have a really strong alternative to suggest OTOH, so let's go with that.
> * ❓ I forgot to ask, can I use a passwordless key store with this? I assume not right now, and it might create a NullPointerException, right? If effort is < 2 hours to add support for passwordless, then let's do it as a follow up PR. If not, then let's at least not throw a NullPointerException but instead a proper exception if no password is given in the validation. 🤔

Agree ideally the configs would get merged but for now its a stop gap (hopefully). Also now supports passwordless keystores

PR #21392, opened 2024-08-20
feat: support user record in exporter

### LINKED ISSUE #21230: An implementation of `ExportHandler` exists to support Identity related records

Ben-Sheppard:
## Description

This PR adds in the two implementations required to handle user records within the new camunda-exporter. 

Things to think about:

1. I have used the record key as the ID, I'm not sure if this is the correct approach but it felt like the best option for now
2. I've set the index at the moment to `users`, do we have any preferences here? For example should our indexes be prefixes with something that avoids clashes with potentially other indexes in ES?
3. Tests do not exist yet so I haven't added anything there yet

## Related issues

closes #21230

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/entities/UserEntity.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/entities/UserEntity.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.entities;
+
+public class UserEntity implements ExporterEntity<UserEntity> {
+  private String id;
+  private String username;
+  private String name;
+  private String email;
+  private String password;
+
+  @Override
+  public String getId() {
+    return id;
+  }
+
+  @Override
+  public UserEntity setId(final String id) {
+    this.id = id;
+    return this;
+  }
+
+  public String getUsername() {
+    return username;
+  }
+
+  public UserEntity setUsername(final String username) {
+    this.username = username;
+    return this;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public UserEntity setName(final String name) {
+    this.name = name;
+    return this;
+  }
+
+  public String getEmail() {
+    return email;
+  }
+
+  public UserEntity setEmail(final String email) {
+    this.email = email;
+    return this;
+  }
+
+  public String getPassword() {
+    return password;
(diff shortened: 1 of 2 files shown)


mustafadagher:
Hi Ben, 
I'm currently busy with implementing a similar exporter for Operate. I've opened a [draft PR here](https://github.com/camunda/camunda/pull/21395) for your benefit. You can have a look, but keep in mind this is still a **WIP**

Things to notice:
1. I've created a parent Entity for all operate entities `OperateExporterEntity`. I'm not sure that will continue forward, but I thought it's a good idea to have separate parents for separate exporters.
2. I'm also creating a sub-package for every operate functionality (e.g. entities, exporters, etc.)
3. I'll try to keep the es/os connectors common if possible
4. I'm also keeping common utils to be used by all exporters

I hope this gives you a bit of an idea about the direction I'm trying to take.

Ben-Sheppard:
> Hi Ben, I'm currently busy with implementing a similar exporter for Operate. I've opened a [draft PR here](https://github.com/camunda/camunda/pull/21395) for your benefit. You can have a look, but keep in mind this is still a **WIP**
> 
> Things to notice:
> 
> 1. I've created a parent Entity for all operate entities `OperateExporterEntity`. I'm not sure that will continue forward, but I thought it's a good idea to have separate parents for separate exporters.
> 2. I'm also creating a sub-package for every operate functionality (e.g. entities, exporters, etc.)
> 3. I'll try to keep the es/os connectors common if possible
> 4. I'm also keeping common utils to be used by all exporters
> 
> I hope this gives you a bit of an idea about the direction I'm trying to take.

Thank you for this context @mustafadagher :) - seems like, with the exception of a couple of small changes I'm on the right path (which is nice to see!). I think given that we want to move away from the idea of component ownership and instead a resource based split I'd opt not to introduce structures such as `OperateExporterEntity` or `IdentityExporterEntity` but thats just down to some implementation details.

One question I have would be, do we need another layer between the `ExporterEntity` and the implementation? Can we just have a single base class with the partition methods there too? I'm not sure what we gain from have another parent entity.

Regarding the sub packages, could we maybe just subpackage by entity? once again, not sure if we gain much by that though.

mustafadagher:
Hi @Ben-Sheppard! Sorry, this has slipped away from my list.
 
> I think given that we want to move away from the idea of component ownership and instead a resource based split I'd opt not to introduce structures such as `OperateExporterEntity` or `IdentityExporterEntity` but thats just down to some implementation details.
>
> One question I have would be, do we need another layer between the `ExporterEntity` and the implementation? Can we just have a single base class with the partition methods there too? I'm not sure what we gain from have another parent entity.

Good point, agreed!
 
> Regarding the sub packages, could we maybe just subpackage by entity? once again, not sure if we gain much by that though.

I also agree. No need to sub-package by entity as well honestly.

One thing to mention, I've abandoned my draft PR for now to focus more on the common exporter functionality, rather than operate specifics. But you shall go ahead with your entities and handlers

mustafadagher (approved):
LGTM! 🚀

PR #21417, opened 2024-08-21
Qualify operateObjectMapper injections

panagiotisgts:
## Description
Qualify Jackson ObjectMapper with _operateObjectMapper_ in Operate due to conflicts in _StandaloneCamunda_ runner and the _camunda/camunda_ docker image conflicting with other modules such as Tasklsit.


ralfpuchert (approved):
Thanks for catching this!

PR #21426, opened 2024-08-21
Support REST in Throw error Job command

### LINKED ISSUE #21412: Support the Throw error API over REST

nicpuppa:
## Description

Rest support for Job throw error command

## Related issues

closes #21412

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ThrowErrorCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ThrowErrorCommandImpl.java
@@ -22,6 +22,9 @@ import io.camunda.zeebe.client.api.command.FinalCommandStep;
 import io.camunda.zeebe.client.api.command.ThrowErrorCommandStep1;
 import io.camunda.zeebe.client.api.command.ThrowErrorCommandStep1.ThrowErrorCommandStep2;
 import io.camunda.zeebe.client.impl.RetriableClientFutureImpl;
+import io.camunda.zeebe.client.impl.http.HttpClient;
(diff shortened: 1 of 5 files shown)


clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ThrowErrorCommandImpl.java
```diff
@@ -30,69 +33,127 @@
...
-    builder = ThrowErrorRequest.newBuilder();
-    builder.setJobKey(key);
+    grpcRequestObjectBuilder = ThrowErrorRequest.newBuilder();
+    grpcRequestObjectBuilder.setJobKey(key);
+    this.httpClient = httpClient;
+    httpRequestConfig = httpClient.newRequestConfig();
+    httpRequestObject = new JobErrorRequest();
+    useRest = preferRestOverGrpc;
+    jobKey = key;
+    this.jsonMapper = jsonMapper;
```
koevskinikola, line 71: 🔧 Same as with the `FailJobCommandImpl`, why do we need to have a second property for the ` jsonMapper` when the base class already provides it through the `objectMapper` property?

koevskinikola (changes requested):
Nice job @nicpuppa, I only have a minor clarification before we can merge this PR. 🚀

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ThrowErrorCommandImpl.java
nicpuppa: Good catch 👍

koevskinikola (approved):
Awesome! Looks great now @nicpuppa 🚀

Issue #18212, opened 2024-05-03
[EPIC] CI: use Infra self-hosted runners everywhere


cmur2:
## Description

Several GHA workflows of the monorepo run on GHA runners managed and maintained by the Zeebe team in a separate Google Cloud project. Find [current usages via this GitHub search](https://github.com/search?q=repo%3Acamunda%2Fcamunda+runs-on%3A+self-hosted+linux+amd64&type=code). The **Infra** team is offering **an [Engineering-wide service](https://confluence.camunda.com/display/HAN/Github+Actions+Self-Hosted+Runners) for the same use case**. The **Zeebe** team should be able to **focus on their primary development activities** and not have to maintain CI infrastructure that is redundant with other Engineering, see e.g. https://github.com/camunda/zeebe/issues/17090#issuecomment-2037059935.

**Goal:** all GHA workflows of the monorepo should be using the [Infra self-hosted GHA runners](https://confluence.camunda.com/display/HAN/Github+Actions+Self-Hosted+Runners). If there are impediments that prevent this migration, those have to be resolved first.

Known impediments:

- ✅ waiting for https://github.com/camunda/team-infrastructure/issues/517 to increase isolation giving more stability
- ✅ Infra runners do not offer `root` access for security reasons thus kernel settings required in https://github.com/camunda/zeebe/blob/51a9d1abfe801dad77e8b3175c0e7950b0b6bb1c/.github/workflows/zeebe-ci.yml#L167 cannot be changed (workaround: use Github's runners for special test cases)
- ✅ Code like [this](https://github.com/camunda/camunda/blob/3065a7ced57dda640693f7c63e76d8321e133b55/.github/actions/setup-zeebe/action.yml#L154) checks for self-hosted runners using outdated/insufficient name prefix comparison and needs to be adjusted. Fixed in https://github.com/camunda/camunda/pull/21429

Adoption ratio/status on 2024-08-07:

![Image](https://github.com/user-attachments/assets/0954f324-38c0-4fd0-8786-464fd7bf5aec)

## Task Breakdown

* https://github.com/camunda/team-infrastructure/issues/517
* solve known impediments:
    * https://github.com/camunda/camunda/pull/21429
* get Infra team help for incremental migration (not all at once to avoid capacity bottlenecks, monitor success via CI health metrics): https://github.com/camunda/team-infrastructure/issues/671

cmur2:
[...]

PR #21429, opened 2024-08-21
ci: use `runner.environment` variable instead of custom script

cmur2:
## Description

Relying on `runner.name` prefix is fragile and the list is outdated currently anyways (Infra runners currently have the prefix `camunda-`). This PR removes an impediment for #18212.

⚠️ Merging this PR will lead to one-time Maven cache recreation on `main`.

See also [docs 1](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context) or [docs 2](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context)

## Checklist

- [ ] for CI changes:

## Related issues

None

### DIFF
--- a/.github/actions/setup-zeebe/action.yml
+++ b/.github/actions/setup-zeebe/action.yml
@@ -144,18 +144,6 @@ runs:
         -D aether.syncContext.named.time=120
         -D maven.artifact.threads=32
         EOF
-    - name: Determine if running on GH infra or self-hosted
-      if: inputs.java == 'true'
-      id: runner-env
-      shell: bash
-      # it matters for caching as absolute paths on self-hosted and Github runners differ
-      # self-hosted: `/runner/` vs gh-hosted: `/home/runner`
-      run: |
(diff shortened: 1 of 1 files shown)


lenaschoenburg (approved):
Well that's a much better way to do it :facepalm: I could swear this wasn't available when I wrote this :sweat_smile: 

Since the naming is almost the same, only caches from github-hosted runners will need to be recreated.

Thanks :+1:

cmur2:
> Well that's a much better way to do it 🤦 I could swear this wasn't available when I wrote this 😅

It wasn't documented for more than a year, I just learned about it today when looking for a better way! See https://github.com/orgs/community/discussions/48359 and https://github.com/orgs/community/discussions/118413

> Since the naming is almost the same, only caches from github-hosted runners will need to be recreated.

💯

Issue #21038, opened 2024-08-12
Resolve called process to correct version for call activities with binding type `versionTag`


jfriedenstab:
### Description
Call activities that use the new binding type `versionTag` must call exactly the version of the target process referenced by the specified version tag.

_Note_: The same version tag can be assigned to multiple deployed versions of a process. In this case, the latest of these versions must be used.

### Acceptance Criteria
- [x] During model transformation, the `versionTag` from the `ZeebeCalledElement` extension element of a call activity is added to the `ExecutableCallActivity`.
- [x] When the process to be called is resolved in `CallActivityProcessor` and the binding type is `versionTag`:
    - Use the target `processId` and `versionTag` to retrieve the deployed process from the state (using the method introduced in #21035).
    - In case a process with the given `processId` and `versionTag` cannot be found, return a `Failure` with a meaningful error message.

### Dependencies
- requires #21034
- requires #21035

PR #21435, opened 2024-08-21
feat: store deployed processes by id and version tag

### LINKED ISSUE #21035: Store deployed processes by id and version tag
### Description
To support the new binding type `versionTag` for call activities, Zeebe needs to be able to resolve a called process by process id and version tag.

### Acceptance Criteria
- [x] When a BPMN process is deployed, the (optional) version tag is read from the XML (`zeebe:versionTag` extension element) and stored in a new property `versionTag` in the created `ProcessRecord` and `PersistedProcess`.
- [x] A new column family is added to store a reference to the respective `processDefinitionKey` by `processId` and `versionTag`.
- [x] A new method is added to the `ProcessState` that allows to find a deployed process by `processId` and `versionTag`.

### Dependencies
- requires https://github.com/camunda/camunda/issues/21124
- required by #21038

jfriedenstab:
## Description
Store the version tag as a new property for deployed processes and add the possibility to find a process by process id and version tag. This is a prerequisite to be able to properly resolve called processes when the new `versionTag` binding type is used (see #21038).

## Related issues
closes #21035


zeebe/exporters/elasticsearch-exporter/src/main/resources/zeebe-record-process-template.json
```diff
@@ -45,6 +45,9 @@
             },
             "deploymentKey": {
               "type": "long"
+            },
+            "versionTag": {
+              "type": "keyword"
```
jfriedenstab, line 50: FYI: I was a bit unsure if I should use type `keyword` or `text`, but from https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html it sounded like `keyword` would be more appropriate:

> used for structured content such as IDs, email addresses, hostnames, status codes, zip codes, or **tags**.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbProcessState.java
```diff
@@ -94,6 +94,19 @@ public final class DbProcessState implements MutableProcessState {
...
       processDefinitionKeyByProcessIdAndDeploymentKeyColumnFamily;
 
+  private final DbString versionTag;
+  private final DbTenantAwareKey<DbCompositeKey<DbString, DbString>>
+      tenantAwareProcessIdAndVersionTagKey;
+
+  /**
+   * <b>Note</b>: Will only be filled with processes deployed from 8.6 onwards that have a version
+   * tag assigned (which is an optional property).
+   */
```
koevskinikola, line 104: ❓ Does this mean that we don't need to perform a migration on already deployed processes?

🤔 I wonder if it would be good to have a migration test to confirm that there are no migration issues due to this change.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionDeleteProcessor.java
koevskinikola, line 1: 🔧 It would be good to put the resource deletion changes in a separate commit for better readability of the changes.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionTest.java
koevskinikola, line 1: 🔧 It would be good to put the resource deletion changes in a separate commit for better readability of the changes.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessCreatedV2Applier.java
```diff
@@ -25,5 +25,6 @@ public ProcessCreatedV2Applier(final MutableProcessingState state) {
   public void applyState(final long processDefinitionKey, final ProcessRecord value) {
     processState.putProcess(processDefinitionKey, value);
     processState.storeProcessDefinitionKeyByProcessIdAndDeploymentKey(value);
+    processState.storeProcessDefinitionKeyByProcessIdAndVersionTag(value);
```
koevskinikola, line 28: 💭 Just commenting for clarity:

This class was added with PR #20498 so any records that are processed with this applier must be created by Zeebe `8.6.0`, i.e. they will be `versionTag`-aware.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbProcessState.java
```diff
@@ -231,6 +269,10 @@ public void deleteProcess(final ProcessRecord processRecord) {
     processDefinitionKeyByProcessIdAndDeploymentKeyColumnFamily.deleteIfExists(
         tenantAwareProcessIdAndDeploymentKey);
 
+    // use deleteIfExists as no entry exists for processes that do not have a version tag assigned
+    processDefinitionKeyByProcessIdAndVersionTagColumnFamily.deleteIfExists(
+        tenantAwareProcessIdAndVersionTagKey);
+
```
koevskinikola, line 275: 🔧 It would be good to put the process deletion changes in a separate commit for better readability.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbProcessState.java
```diff
@@ -372,6 +414,22 @@ public DeployedProcess getProcessByProcessIdAndVersion(
...
 
+  @Override
+  public DeployedProcess getProcessByProcessIdAndVersionTag(
+      final DirectBuffer processId, final String versionTag, final String tenantId) {
+    tenantIdKey.wrapString(tenantId);
+    this.processId.wrapBuffer(processId);
+    this.versionTag.wrapString(versionTag);
+    final var foreignKey =
+        processDefinitionKeyByProcessIdAndVersionTagColumnFamily.get(
+            tenantAwareProcessIdAndVersionTagKey);
```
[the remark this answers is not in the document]
koevskinikola: I agree, I'm not sure how cache invalidation would work. All the process caches until now add a unique combination (ex. tenant and key, process id and version, etc) that may push out an older entry from the cache, but not invalidate it.

Maybe you can create a follow-up "enhancement" issue for this? That way, if we notice performance issues with the look-up by process id and version tag, we can add them to that issue and prioritize it.

koevskinikola (changes requested):
Nice job @jfriedenstab! 👏 

🙃 There are a lot of comments, but most of them are nitpicks for improving the readability of this PR or the code in some of the classes.

:x: I noted some (probably) required changes for the Tasklist classes.

❓ I'm also wondering if a migration test would be useful here, to ensure that no existing deployed processes are impacted by this change.

koevskinikola:
🔧 @jfriedenstab the `MultiPartitionDeploymentLifecycleTest` is also failing, it might be related to the changes.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbProcessState.java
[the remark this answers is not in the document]
jfriedenstab: Thanks, good suggestion! 👍🏻 I'll create an issue.

tasklist/importer-850/src/main/java/io/camunda/tasklist/zeebeimport/v850/record/value/deployment/DeployedProcessImpl.java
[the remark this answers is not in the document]
jfriedenstab: Ah, thanks, I totally missed those methods 🤦🏻.

jfriedenstab:
> `MultiPartitionDeploymentLifecycleTest` is also failing, it might be related to the changes.

My assumption is that it's because of the new `versionTag` property in the record that is being written (which makes the record slightly bigger and causes an `EXCEEDED_BATCH_RECORD_SIZE` error). It can be fixed by generating a slightly smaller resource in `shouldDeployIfResourceIsLargeButNotTooMuch`.

zeebe/exporters/elasticsearch-exporter/src/main/resources/zeebe-record-deployment-template.json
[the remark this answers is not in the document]
jfriedenstab: Good suggestion! I've split the commits.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbProcessState.java
jfriedenstab: I'm not really familiar with the migrations, but in my opinion, we don't need to perform one, because already deployed processes won't have a version tag anyway.

By the way, I also had a similar discussion with Nico in the context of "deployment version binding" and we concluded that a migration would not be necessary (see https://github.com/camunda/camunda/issues/19709#issuecomment-2245723417 and the comments above).

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessCreatedV2Applier.java
jfriedenstab: Yes, I figured that it would be a similar case as above where processes are stored by id and deployment key.

jfriedenstab:
@koevskinikola Thanks a lot for the review! I've addressed all your comments.

jfriedenstab:
@camunda/hto: I needed to make some changes in the `tasklist` module (see `DeployedProcessImpl` classes) due to a new field in the `ProcessMetadataValue` interface, maybe one of you could have a quick look?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbProcessState.java
koevskinikola: Ah okay, that's good to see. Then consider this discussion resolved 😄

koevskinikola (approved):
All good from my side @jfriedenstab! 🚀

marcosgvieira (approved):
Everything looking good from Tasklist perspective, thanks @jfriedenstab

PR #21441, opened 2024-08-21
Reject decimal from task priority and throw incidents

panagiotisgts:
## Description

Validator was mapping doubles as integers. Provide a fix to only accept integers and not decimal numbers.

Adjust expression evaluation to throw incidents when priority expression is not in the [0,100] range.

Add test cases for the above two scenarios

## Related issues

closes #21434


korthout (changes requested):
Thanks @panagiotisgts 🚀 

👍 The changes look good

❌ Please add a test case to show that this works

korthout (approved):
Love it!

LGTM 👍

PR #21442, opened 2024-08-21
fix: prevent completing multi-instance body until the completion/termination of all its children

### LINKED ISSUE #12918: NullPointerException in ActivateProcessInstanceBatchProcessor.createChildInstanceRecord
**Describe the bug**

NPE is thrown when creating instance of process with multiinstance and a large input collection. 

Instance is created:

```
2023-05-31 14:41:06.222  INFO 36794 --- [pool-6-thread-1] i.c.cloud.gameday.ProcessApplication     : Created process instances with large collection. [order-id: '7c7014f4-16fe-4609-9ef2-d017546416af']
```

But can't be executed. 
[It looks like the instance get banned.](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22camunda-saas-int-chaos%22%0Aresource.labels.location%3D%22europe-west1%22%0Aresource.labels.cluster_name%3D%22worker-chaos-1%22%0Aresource.labels.namespace_name%3D%2229c46b01-ab7a-4b96-9262-10a28257dfe5-zeebe%22%0Aresource.labels.pod_name:%22zeebe-%22%0A-resource.labels.container_name%3D%22zeebe-analytics%22%0A-resource.labels.container_name%3D%22zeebe-gateway%22;timeRange=2023-05-31T11:29:31.241992Z%2F2023-05-31T13:36:09.546Z;pinnedLogId=2023-05-31T12:36:40.822878656Z%2F8qiahd3sixyvvbf4;cursorTimestamp=2023-05-31T12:36:40.822878656Z?project=camunda-saas-int-chaos)

**To Reproduce**

Follow https://github.com/camunda-cloud/game-day-cookbook/tree/main/mains/large-multi-instance and change 

```diff
--- a/mains/large-multi-instance/src/main/java/io/camunda/cloud/gameday/ProcessApplication.java
+++ b/mains/large-multi-instance/src/main/java/io/camunda/cloud/gameday/ProcessApplication.java
@@ -49,15 +49,15 @@ public class ProcessApplication implements CommandLineRunner {
   public void run(final String... args) {
     final var executor = Executors.newSingleThreadScheduledExecutor();
     executor.scheduleAtFixedRate(
            () -> createInstanceOfProcess(10), 1, 30, TimeUnit.SECONDS);

     executor.scheduleAtFixedRate(
         () -> {
-          final var orderId = createInstanceOfProcess(20_000);
+          final var orderId = createInstanceOfProcess(2_000_000);
           LOG.info("Created process instances with large collection. [order-id: '{}']", orderId);
 
-        }, 5, 15, TimeUnit.MINUTES);
+        }, 0, 15, TimeUnit.MINUTES);
   }

          final var orderId = createInstanceOfProcess(2_000_000);
```

**Expected behavior**

Correct batching and not failing of creating such instance, or if possible rejection.

**Log/Stacktrace**

 <p>

```
java.lang.NullPointerException: Cannot invoke "io.camunda.zeebe.engine.state.instance.ElementInstance.getValue()" because "parentElementInstance" is null
	at io.camunda.zeebe.engine.processing.processinstance.ActivateProcessInstanceBatchProcessor.createChildInstanceRecord(ActivateProcessInstanceBatchProcessor.java:66) ~[zeebe-workflow-engine-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.engine.processing.processinstance.ActivateProcessInstanceBatchProcessor.processRecord(ActivateProcessInstanceBatchProcessor.java:45) ~[zeebe-workflow-engine-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.engine.Engine.process(Engine.java:142) ~[zeebe-workflow-engine-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.stream.impl.ProcessingStateMachine.batchProcessing(ProcessingStateMachine.java:346) ~[zeebe-stream-platform-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.stream.impl.ProcessingStateMachine.lambda$processCommand$2(ProcessingStateMachine.java:268) ~[zeebe-stream-platform-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.db.impl.rocksdb.transaction.ZeebeTransaction.run(ZeebeTransaction.java:84) ~[zeebe-db-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.stream.impl.ProcessingStateMachine.processCommand(ProcessingStateMachine.java:268) ~[zeebe-stream-platform-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.stream.impl.ProcessingStateMachine.tryToReadNextRecord(ProcessingStateMachine.java:227) ~[zeebe-stream-platform-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.stream.impl.ProcessingStateMachine.readNextRecord(ProcessingStateMachine.java:203) ~[zeebe-stream-platform-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.scheduler.ActorJob.invoke(ActorJob.java:92) [zeebe-scheduler-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.scheduler.ActorJob.execute(ActorJob.java:45) [zeebe-scheduler-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.scheduler.ActorTask.execute(ActorTask.java:119) [zeebe-scheduler-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.scheduler.ActorThread.executeCurrentTask(ActorThread.java:106) [zeebe-scheduler-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.scheduler.ActorThread.doWork(ActorThread.java:87) [zeebe-scheduler-8.2.5.jar:8.2.5]
	at io.camunda.zeebe.scheduler.ActorThread.run(ActorThread.java:198) [zeebe-scheduler-8.2.5.jar:8.2.5]

```

</p>
</details>

**Environment:**
- OS: 
- Zeebe Version: 8.2.x
- Configuration: G3-S

### LINKED ISSUE #20958: Multi instance body finishes with less inner activities than expected
**Describe the bug**

Multi instance body finishes with less inner activities than expected.

**To Reproduce**
Use the below processes ti reproduce the issue:

[MultiInstanceLoop-NOT WORKING.bpmn.zip](https://github.com/user-attachments/files/16562463/MultiInstanceLoop-NOT.WORKING.1.bpmn.zip) - multi instance body finishes with less inner activities than expected, the last multi-instance body finishes with just one activity.

[MultiInstanceLoop-fixed.bpmn.zip](https://github.com/user-attachments/files/16562465/MultiInstanceLoop-NOT.WORKING-fixed.bpmn.zip) - adding a wait time resolves the issue and the last multi-instance body finishes with 5 activities

**Expected behavior**

Both processes should have the same number of inner activities.

**Log/Stacktrace**
[raft-partition.zip](https://github.com/user-attachments/files/16562530/raft-partition.zip)
Process instance key with less inner activities: `2251799813687853` 

**Environment:**
- Zeebe Version: 8.6.0-SNAPSHOT, likely affects older releases down to 8.2 as well

Linked to https://jira.camunda.com/browse/SUPPORT-23118

mustafadagher:
## Description

Prevent completing multi-instance body until the completion/termination of all its children

## Related issues

closes #20958
closes #12918

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
@@ -237,6 +237,17 @@ public final class MultiInstanceBodyProcessor
     }
 
     if (!childInstanceCreated && stateBehavior.canBeCompleted(childContext)) {
+      final var inputCollectionOrFailure = readInputCollectionVariable(element, flowScopeContext);
+      final var multiInstanceElementInstance = stateBehavior.getElementInstance(flowScopeContext);
+      if (inputCollectionOrFailure.isRight()) {
+        final var inputCollectionSize = inputCollectionOrFailure.get().size();
+        if (inputCollectionSize
+            != (multiInstanceElementInstance.getNumberOfCompletedElementInstances()
+                + multiInstanceElementInstance.getNumberOfTerminatedElementInstances())) {
+          throw new IllegalStateException("kill bill");
+        }
+      }
+
       stateTransitionBehavior.completeElement(flowScopeContext);
     }
   }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
@@ -183,10 +183,6 @@ public final class MultiInstanceBodyProcessor
       return satisfiesCompletionConditionOrFailure;
     }
 
-    if (!element.getLoopCharacteristics().isSequential()) {
-      return Either.right(satisfiesCompletionConditionOrFailure.get());
-    }
-
     // test that input collection variable can be evaluated correctly
     return readInputCollectionVariable(element, flowScopeContext)
         .map(ok -> satisfiesCompletionConditionOrFailure.get());
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
@@ -20,6 +20,7 @@ import io.camunda.zeebe.engine.processing.bpmn.behavior.MultiInstanceOutputColle
 import io.camunda.zeebe.engine.processing.common.ExpressionProcessor;
 import io.camunda.zeebe.engine.processing.common.Failure;
 import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableMultiInstanceBody;
+import io.camunda.zeebe.engine.state.instance.ElementInstance;
 import io.camunda.zeebe.msgpack.spec.MsgPackHelper;
 import io.camunda.zeebe.msgpack.spec.MsgPackWriter;
 import io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent;
@@ -213,19 +214,19 @@ public final class MultiInstanceBodyProcessor
       }
       return;
     }
+    final var inputCollectionOrFailure = readInputCollectionVariable(element, flowScopeContext);
+    if (inputCollectionOrFailure.isLeft()) {
+      // this incident is un-resolvable
+      incidentBehavior.createIncident(inputCollectionOrFailure.getLeft(), childContext);
+      return;
+    }
+
+    final ElementInstance multiInstanceElementInstance =
+        stateBehavior.getElementInstance(flowScopeContext);
 
     if (loopCharacteristics.isSequential()) {
-
-      final var inputCollectionOrFailure = readInputCollectionVariable(element, flowScopeContext);
-      if (inputCollectionOrFailure.isLeft()) {
-        // this incident is un-resolvable
-        incidentBehavior.createIncident(inputCollectionOrFailure.getLeft(), childContext);
-        return;
-      }
-
       final var inputCollection = inputCollectionOrFailure.get();
-      final var loopCounter =
-          stateBehavior.getElementInstance(flowScopeContext).getMultiInstanceLoopCounter();
+      final var loopCounter = multiInstanceElementInstance.getMultiInstanceLoopCounter();
 
       if (loopCounter < inputCollection.size()) {
         createInnerInstance(element, flowScopeContext);
@@ -237,18 +238,11 @@ public final class MultiInstanceBodyProcessor
     }
 
     if (!childInstanceCreated && stateBehavior.canBeCompleted(childContext)) {
-      final var inputCollectionOrFailure = readInputCollectionVariable(element, flowScopeContext);
-      final var multiInstanceElementInstance = stateBehavior.getElementInstance(flowScopeContext);
-      if (inputCollectionOrFailure.isRight()) {
-        final var inputCollectionSize = inputCollectionOrFailure.get().size();
-        if (inputCollectionSize
-            != (multiInstanceElementInstance.getNumberOfCompletedElementInstances()
-                + multiInstanceElementInstance.getNumberOfTerminatedElementInstances())) {
-          throw new IllegalStateException("kill bill");
-        }
+      final int inputCollectionSize = inputCollectionOrFailure.get().size();
+      if (isAllChildrenHasCompletedOrTerminated(
+          multiInstanceElementInstance, inputCollectionSize)) {
+        stateTransitionBehavior.completeElement(flowScopeContext);
       }
-
-      stateTransitionBehavior.completeElement(flowScopeContext);
     }
   }
 
@@ -266,6 +260,14 @@ public final class MultiInstanceBodyProcessor
       // all remaining child instances were terminated.
       stateTransitionBehavior.completeElement(flowScopeContext);
     }
+  }
+
+  private static boolean isAllChildrenHasCompletedOrTerminated(
+      final ElementInstance multiInstanceElementInstance, final int inputCollectionSize) {
+    final int completedOrTerminatedChildren =
+        multiInstanceElementInstance.getNumberOfCompletedElementInstances()
+            + multiInstanceElementInstance.getNumberOfTerminatedElementInstances();
+    return inputCollectionSize == completedOrTerminatedChildren;
   }
 
   private void activate(
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/multiinstance/MultiInstanceBatchedSubProcessesTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/multiinstance/MultiInstanceBatchedSubProcessesTest.java
@@ -18,13 +18,13 @@ import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 
-// TODO rename
-public class NewMultiInstanceTest {
+public class MultiInstanceBatchedSubProcessesTest {
 
   @ClassRule
   public static final EngineRule ENGINE = EngineRule.singlePartition().maxCommandsInBatch(47);
 
   private static final String PROCESS_ID = "process";
+  private static final String SUBPROCESS_2_START_EVENT = "subprocess2StartEvent";
   private static final BpmnModelInstance PROCESS =
       Bpmn.createExecutableProcess(PROCESS_ID)
           .startEvent()
@@ -45,7 +45,7 @@ public class NewMultiInstanceTest {
                           b.zeebeInputCollectionExpression("line.orders")
                               .zeebeInputElement("order")))
           .embeddedSubProcess()
-          .startEvent("subprocess2StartEvent")
+          .startEvent(SUBPROCESS_2_START_EVENT)
           .exclusiveGateway("exclusive-gateway")
           .defaultFlow()
           .exclusiveGateway()
@@ -61,19 +61,18 @@ public class NewMultiInstanceTest {
       new RecordingExporterTestWatcher();
 
   @Test
-  public void test() {
+  public void shouldCompleteAllChildInstancesBeforeCompletingTheMultiInstanceBody() {
     // given
     ENGINE.deployment().withXmlResource(PROCESS).deploy();
-    final var processInstanceKey = ENGINE.processInstance().ofBpmnProcessId(PROCESS_ID).create();
+    final int orderCount = 50;
 
     // when
-    final int orderCount = 50;
-    final String startEventToLookFor = "subprocess2StartEvent";
+    ENGINE.processInstance().ofBpmnProcessId(PROCESS_ID).create();
 
     // then
     Assertions.assertThat(
             RecordingExporter.processInstanceRecords(ProcessInstanceIntent.ELEMENT_ACTIVATED)
-                .withElementId(startEventToLookFor)
+                .withElementId(SUBPROCESS_2_START_EVENT)
                 .limit(orderCount)
                 .toList())
         .hasSize(orderCount);
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/multiinstance/MultiInstanceBatchedSubProcessesTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/multiinstance/MultiInstanceBatchedSubProcessesTest.java
@@ -21,37 +21,23 @@ import org.junit.Test;
 public class MultiInstanceBatchedSubProcessesTest {
 
   @ClassRule
-  public static final EngineRule ENGINE = EngineRule.singlePartition().maxCommandsInBatch(47);
+  public static final EngineRule ENGINE = EngineRule.singlePartition().maxCommandsInBatch(6);
 
   private static final String PROCESS_ID = "process";
-  private static final String SUBPROCESS_2_START_EVENT = "subprocess2StartEvent";
+  private static final String SUB_PROCESS_START = "sub-process-start";
   private static final BpmnModelInstance PROCESS =
       Bpmn.createExecutableProcess(PROCESS_ID)
           .startEvent()
-          .zeebeOutputExpression(getProcessVariable(), "bill")
+          .zeebeOutputExpression("= [1,2]", "inputCollection")
           .subProcess(
               "subprocess1",
               s ->
                   s.multiInstance(
                       b ->
-                          b.zeebeInputCollectionExpression("bill.lines").zeebeInputElement("line")))
+                          b.zeebeInputCollectionExpression("inputCollection")
+                              .zeebeInputElement("input")))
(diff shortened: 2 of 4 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/multiinstance/MultiInstanceBatchedSubProcessesTest.java
```diff
@@ -0,0 +1,401 @@
...
+    // then
+    Assertions.assertThat(
+            RecordingExporter.processInstanceRecords(ProcessInstanceIntent.ELEMENT_ACTIVATED)
+                .withElementId(SUBPROCESS_2_START_EVENT)
+                .limit(orderCount)
+                .toList())
+        .hasSize(orderCount);
+  }
+
+  private static String getProcessVariable() {
```
remcowesterhoud, line 81: ❓ Do you think you could simplify this test case now that we understand the real problem? I believe we should able to reproduce with a single (non-nested) multi-instance subprocess with an input collection of only 2 elements 🙂

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
```diff
@@ -213,19 +214,19 @@ public void afterExecutionPathCompleted(
...
+    final var inputCollectionOrFailure = readInputCollectionVariable(element, flowScopeContext);
+    if (inputCollectionOrFailure.isLeft()) {
+      // this incident is un-resolvable
+      incidentBehavior.createIncident(inputCollectionOrFailure.getLeft(), childContext);
+      return;
+    }
 
-    if (loopCharacteristics.isSequential()) {
-
-      final var inputCollectionOrFailure = readInputCollectionVariable(element, flowScopeContext);
```
korthout, line 219: ❓  This may come with a performance hit for parallel multi-instance, as it now reads the input collection variable on each inner instance completion. Did we check the performance of this?

💭 Note that reading the variable here means evaluating the input collection expression again. Most cases will probably do a direct lookup of a list variable, but theoretically this expression could require multiple variable lookups.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/multiinstance/MultiInstanceBatchedSubProcessesTest.java
mustafadagher: I actually tried simplifying to `1 non-nested MI with 5 items` and `nested multi instance of two lines each with two orders` and adjust the batch size, and in both cases I failed to reproduce.

Which probably says it has something to do with the size of the input collection too 😕 🤔

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
mustafadagher: I see your point. No, we didn't check any performance implications, we focused more on resolving the bug.

We indeed `readInputCollectionVariable` at `onActivate` and `onActivating` and we could use these times to get the input collection size, I've looked around now and I could't decide where/how to store it that it keeps it relevant to an element. 

The elementInstance state could be one candidate, but I feel like the `MultiInstanceBodyProcessor` is not the right place to alter the state.

Maybe I could use `stateBehavior.setLocalVariable()`? But I'm also not sure if this is a correct use case to use it.

Any suggestions?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/multiinstance/MultiInstanceBatchedSubProcessesTest.java
remcowesterhoud: Try this one 😉 

```java
/*
 * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
 * one or more contributor license agreements. See the NOTICE file distributed
 * with this work for additional information regarding copyright ownership.
 */
package io.camunda.zeebe.engine.processing.bpmn.multiinstance;

import io.camunda.zeebe.engine.util.EngineRule;
import io.camunda.zeebe.model.bpmn.Bpmn;
import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
import io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent;
import io.camunda.zeebe.test.util.record.RecordingExporter;
import io.camunda.zeebe.test.util.record.RecordingExporterTestWatcher;
import org.assertj.core.api.Assertions;
import org.junit.ClassRule;
import org.junit.Rule;
import org.junit.Test;

public class MultiInstanceBatchedSubProcessesTest {

  @ClassRule
  public static final EngineRule ENGINE = EngineRule.singlePartition().maxCommandsInBatch(6);

  private static final String PROCESS_ID = "process";
  private static final BpmnModelInstance PROCESS =
      Bpmn.createExecutableProcess(PROCESS_ID)
          .startEvent()
          .zeebeOutputExpression("= [1,2]", "inputCollection")
          .subProcess(
              "subprocess1",
              s ->
                  s.multiInstance(
                      b ->
                          b.zeebeInputCollectionExpression("inputCollection")
                              .zeebeInputElement("input")))
          .embeddedSubProcess()
          .startEvent("sub-process-start")
          .subProcessDone()
          .endEvent()
          .done();

  @Rule
  public final RecordingExporterTestWatcher recordingExporterTestWatcher =
      new RecordingExporterTestWatcher();

  @Test
  public void shouldCompleteAllChildInstancesBeforeCompletingTheMultiInstanceBody() {
    // given
    ENGINE.deployment().withXmlResource(PROCESS).deploy();
    final int count = 2;

    // when
    ENGINE.processInstance().ofBpmnProcessId(PROCESS_ID).create();

    // then
    Assertions.assertThat(
            RecordingExporter.processInstanceRecords(ProcessInstanceIntent.ELEMENT_ACTIVATED)
                .withElementId("sub-process-start")
                .limit(count)
                .toList())
        .hasSize(count);
  }
}

```

The reason you couldn't reproduce it is because the amount of commands processed in the subprocess exceeded the maximum command batch size. Because of that the subprocess didn't complete fully. By reducing the amount of commands written in the subprocess we can reproduce it with a more minimal example 🙂 

I'd also suggest to link the issue in the test case to specify it's a regression test for this bug. That gives some more context to understand it later on.

mustafadagher: Thank you! 🙇

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
korthout: It would be new information for an `ElementInstance` that is not yet tracked. Something along the lines of `childPlannedCount` or `multiInstanceInputSize`, which we could add next to the existing `childCount`, `childActivatedCount`, `childCompletedCount`, and `childTerminatedCount`:

https://github.com/camunda/camunda/blob/2d623b82dee3dfb8ee3b8c8fe89d03acccf5696b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/ElementInstance.java#L28-L34

We can store this on `ProcessInstance:ELEMENT_ACTIVATING` or `ProcessInstance:ELEMENT_ACTIVATED` by adding a new property on it. However, this property would be specific to multi-instance body element instances, so all other records always get the same value (1) which is quite wasteful (at least a few extra bytes for each process instance record).

Alternatively, we can introduce a new record/event to store this specific data, e.g. something along the lines of `MultiInstance:BODY_ACTIVATING`. It can reference the multi-instance body by key, and provide some additional information:
- input collection size: important for this
- input collection itself: I don't see a use for this yet, but might be helpful later (I'd not add this if we have no need for it yet)

Perhaps there are other records we can use to store this data. But this is what comes to mind. Curious to your thoughts about this @mustafadagher @remcowesterhoud.

mustafadagher: @korthout Do you mean as in to add it in the scope of `EventAppliers`? I don't think we have the `expressionBehavior` or any of the `BpmnBehaviors` available for us in this phase to evaluate the expression and store the result.

korthout: >>Alternatively, we can introduce a new record/event to store this specific data, e.g. something along the lines of MultiInstance:BODY_ACTIVATING. It can reference the multi-instance body by key, and provide some additional information

>Fetching data from the state versus evaluating an expression. I wonder what's more performant 🤔 We do already evaluate an input expression if the MI-body is sequential as opposed to parallel. But I suppose performance may be less of an issue if the MI is sequential anyways.

I meant that the event can be used to fill the data in the `ElementInstance`. It would not require an additional lookup on completion of an inner instance as the `ElementInstance` of the multi-instance body is already looked up.

https://github.com/camunda/camunda/blob/2d623b82dee3dfb8ee3b8c8fe89d03acccf5696b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java#L226-L228

> >We can store this on ProcessInstance:ELEMENT_ACTIVATING or ProcessInstance:ELEMENT_ACTIVATED by adding a new property on it. However, this property would be specific to multi-instance body element instances, so all other records always get the same value (1) which is quite wasteful (at least a few extra bytes for each process instance record).

>I would prefer this. We already store the other properties (childActivatedCount, childCompletedCount) which are also not relevant for many of the elemant instances

These are incremented by the `ELEMENT_ACTIVATED` and `ELEMENT_COMPLETED` appliers of the inner instances. No additional information is stored for them in the `ProcessInstance` records.

> I would argue the default value would be 0 as for most elements no children are planned.

👍 You're right.

mustafadagher: One other way to do it could be: the output collection has the same size as the input collection, we already initialize it and store it as a `Local Variable` in `state behavior` when activating the multi-instance body.

We could query the output variable and rely on its size. Or store the input collection size similarly as a `local variable`, we wouldn't then need to change anything to the element instance.

korthout: >We could query the output variable and rely on its size

@mustafadagher While unexpected, and probably leading to problems, users could update the output collection variable themselves. We should not rely on process variables to determine the execution here anymore.

mustafadagher: @korthout please check again now

korthout: I've discussed this solution idea with @saig0 and we concluded that:
- Regarding performance:
  - we don't actually know the performance impact the initial fix has and whether that is a problem
  - the new `MultiInstance` record theoretically also comes at the cost of some performance, e.g. serialization/deserialization of msgpack, writing and reading from RocksDB, and additional exporter load for the new record
  - there are other solutions that may be better suited to improve the performance, like caching the input collection in memory
- Regarding concurrent modification issues for the input collection:
  - there are other solutions that we anyways want to add at some point, like the introduction of immutable variables
- It's generally not a good idea to optimize prematurely

@mustafadagher With this in mind, I would actually suggest that we don't take this optimization now. Instead, let's just fix the bug with the additional read of the collection, and take our learnings by documenting them in issues:
- one to improve the performance of multi-instance
- one to avoid problems in the case of concurrent modification of the input collection

mustafadagher: @korthout 
For the first point, I think I presented the same argument of not knowing if saving and loading instance in the state is more performant than evaluating the expression. But your counter argument was that some expressions might need to render other expressions too to evaluate their values and that might be very slow to do multiple times.

If I understood the conclusion correctly, I will revert my last few commits saving the input collection to the event, and then I can merge this PR, is that tight?

korthout: >If I understood the conclusion correctly, I will revert my last few commits saving the input collection to the event, and then I can merge this PR, is that right?

👍 Yes, I can provide a new review once the changes are ready

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
```diff
@@ -213,19 +214,19 @@ public void afterExecutionPathCompleted(
       }
       return;
     }
+    final var inputCollectionOrFailure = readInputCollectionVariable(element, flowScopeContext);
+    if (inputCollectionOrFailure.isLeft()) {
+      // this incident is un-resolvable
+      incidentBehavior.createIncident(inputCollectionOrFailure.getLeft(), childContext);
+      return;
```
korthout, line 221: ❌ As this incident is unresolveable, we must protect against it earlier. This requires checking the input collection in the `beforeExecutionPathCompleted` method also for the non-sequential (parallel) case.

korthout (changes requested):
Thanks @mustafadagher 🚀 

👍 Changes look good in general

❌ Just one change request to ensure that a specific incident can be resolved

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java
mustafadagher: I introduced no change here, this has been here from before, and I'm not comfortable changing this behavior in the scope of this PR.

mustafadagher: And just as an FYI: we do create an incident  at `onChildActivating`

mustafadagher: if what you mean is to remove the lines 
https://github.com/camunda/camunda/blob/008d8b9106088d19d4895936b6846254d019b1c9/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/MultiInstanceBodyProcessor.java#L185-L187

I shall do so, but I'm not sure if it's necessary

korthout (approved):
Love it! Thanks! 🚀

mustafadagher:
will rebase to resolve failing operate CI errors

PR #21445, opened 2024-08-21
Add ES exporter metrics

ChrisKujawa:
## Description
Add an approximation of how long a record has to spent in the exporter buffer, before flushing. It is an approximation or a max calculation as the first record in the buffer is only considered, but it still already shows some interesting insights.

You could also name it as how long the buffer is open (alive) and collects new records before it gets flushed. 

Allows to build the following panel:

## Related issues

related to https://github.com/camunda/camunda/issues/16912

### DIFF
--- a/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchMetrics.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchMetrics.java
@@ -24,7 +24,7 @@ public class ElasticsearchMetrics {
   private final Timer flushDuration;
   private final DistributionSummary bulkSize;
   private final Counter failedFlush;
-  private final Timer recordBufferLifetime;
+  private final Timer flushLatency;
 
   public ElasticsearchMetrics(final int partitionId, final MeterRegistry registry) {
     partitionIdLabel = String.valueOf(partitionId);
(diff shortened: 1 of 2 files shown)


zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchMetrics.java
```diff
@@ -53,6 +54,14 @@ public ElasticsearchMetrics(final int partitionId, final MeterRegistry registry)
             .description("Number of failed flush operations")
             .tags(PARTITION_LABEL, partitionIdLabel)
             .register(meterRegistry);
+
+    recordBufferLifetime =
+        Timer.builder(meterName("record.buffer.lifetime"))
```
Zelldon, line 59: 🙃 Naming is hard. Of course the name is discussable and happy for any input.

Zelldon: Thinking about this we might want to call it flush latency 🤔 thinking about the discussion here https://github.com/camunda/camunda/pull/21297#discussion_r1725204578

deepthidevaki: Flush latency sounds good to me. Alternate: record buffering latency.

deepthidevaki (approved):
Thanks.

Zelldon:
@deepthidevaki any thoughts regarding backporting? I guess the metric facade refactoring was not backported?

deepthidevaki:
> @deepthidevaki any thoughts regarding backporting? I guess the metric facade refactoring was not backported?

I wouldn't say it is necessary. Other metrics such as "number of records not exported" helps with debugging during incidents. That said, if it is not too much effort to backport it is nice to have it in atleast one previous minor.

Issue #21040, opened 2024-08-12
Resolve called decision to correct version for business rule tasks with binding type `versionTag`


jfriedenstab:
### Description
Business rule tasks that use the new binding type `versionTag` must call exactly the version of the target decision referenced by the specified version tag.

_Note_: The same version tag can be assigned to multiple deployed versions of a decision. In this case, the latest of these versions must be used.

### Acceptance Criteria
- [x] During model transformation, the `versionTag` from the `ZeebeCalledDecision` extension element of a business rule task is added to the `ExecutableBusinessRuleTask`.
- [x] When the decision to be called is resolved in `BpmnDecisionBehavior` and the binding type is `versionTag`:
    - Use the target `decisionId` and `versionTag` to retrieve the deployed decision from the state (using the method introduced in #21036).
    - In case a decision with the given `decisionId` and `versionTag` cannot be found, return a `Failure` with a meaningful error message.

### Dependencies
- requires #21034
- requires #21036

PR #21450, opened 2024-08-21
feat: store deployed decisions by id and version tag

### LINKED ISSUE #21036: Store deployed decisions by id and version tag
### Description
To support the new binding type `versionTag` for business rule tasks, Zeebe needs to be able to resolve a called decision by decision id and version tag.

### Acceptance Criteria
- [x] When a DMN decision is deployed, the (optional) version tag is read from the XML (`zeebe:versionTag` extension element) and stored in a new property `versionTag` in the created `DecisionRecord` and `PersistedDecision`.
- [x] A new column family is added to store a reference to the respective `decisionKey` by `decisionId` and `versionTag`.
- [x] A new method is added to the `DecisionState` that allows to find a deployed decision by `decisionId` and `versionTag`.

### Dependencies
- requires https://github.com/camunda/camunda/issues/21124
- required by #21040

jfriedenstab:
## Description
Store the version tag as a new property for deployed decisions and add the possibility to find a decision by decision id and version tag. This is a prerequisite to be able to properly resolve a called decision when the new `versionTag` binding type is used in a business rule task (see #21040).

## Related issues
closes #21036


zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/ZbColumnFamilies.java
```diff
@@ -180,7 +180,9 @@ public enum ZbColumnFamilies implements EnumValue {
   USERS(86),
   USER_KEY_BY_USERNAME(87),
 
-  CLOCK(88);
+  CLOCK(88),
+
+  DMN_DECISION_KEY_BY_DECISION_ID_AND_VERSION_TAG(90);
```
jfriedenstab, line 185: FYI: `89` was added with #21435.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionDeleteProcessor.java
koevskinikola, line 1: 🔧 It would be better to put resource deletion changes in a separate commit.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbDecisionState.java
```diff
@@ -452,13 +501,17 @@ public void deleteDecision(final DecisionRecord record) {
...
 
     decisionKeyByDecisionRequirementsKey.deleteExisting(dbDecisionRequirementsKeyAndDecisionKey);
     decisionsByKey.deleteExisting(tenantAwareDecisionKey);
     decisionKeyByDecisionIdAndVersion.deleteExisting(tenantAwareDecisionIdAndVersion);
 
     // use deleteIfExists as no entry exists for older records that do not have a deployment key yet
     decisionKeyByDecisionIdAndDeploymentKey.deleteIfExists(tenantAwareDecisionIdAndDeploymentKey);
+
+    // use deleteIfExists as no entry exists for decisions that do not have a version tag assigned
+    decisionKeyByDecisionIdAndVersionTag.deleteIfExists(tenantAwareDecisionIdAndVersionTag);
```
koevskinikola, line 514: 🔧 It would be better to put resource deletion changes in a separate commit.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionTest.java
koevskinikola, line 1: 🔧 It would be better to put resource deletion changes in a separate commit.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DmnDeploymentTest.java
koevskinikola, line 1: 🔧 I think it would be useful if we also add a test that confirms that a new DMN deployment with the same version tag overwrites the existing DMN entry with that version tag in the state.

koevskinikola (approved):
Looks good @jfriedenstab 👍 

I only have some minor feedback that I think would be good to be implemented.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DmnDeploymentTest.java
jfriedenstab: I've added additional tests now to ensure that:
- you can deploy a new decision with an already existing version tag
- storing the decision by ID and version tag in the state overwrites an existing entry

jfriedenstab:
@koevskinikola I've addressed all your comments. Could you have a quick look again? :)

koevskinikola (approved):
The changes look good now @jfriedenstab! 🚀

PR #21474, opened 2024-08-22
Update tasklist translations based on the feedback from native speakers

### LINKED ISSUE #21473: Update Tasklist translations based on the feedback from native speakers
Product hub epic: https://github.com/camunda/product-hub/issues/2378

### Description

Update Tasklist translations to DE, FR, and ES based on the feedback from native speakers.

### TODO

- [x] Update translations for German
- [x] Update translations for Spanish
- [x] Update translations for French
- [x] Fix issue with truncated labels in task sorting popover

volodymyr-melnykc:
## Description

Fixes https://github.com/camunda/camunda/issues/21473

## Checklist

- [ ] for CI changes:

## Related issues

closes #


volodymyr-melnykc:
Lint commits failed because I used the keyword "chore" for some commits. It worked when Tasklist was a separate repo, but now apparently, it is not allowed.

vsgoulart:
> Lint commits failed because I used the keyword "chore" for some commits. It worked when Tasklist was a separate repo, but now apparently, it is not allowed. Is there a way to fix it?

@volodymyr-melnykc The only fix is to use another prefix, in this case `fix: ` is probably better

tasklist/client/src/modules/internationalization/locales/de.json
```diff
@@ -149,7 +149,7 @@
     "customFiltersModalConfirmDeletionButton": "Löschen bestätigen",
     "customFiltersModalAriaLabel": "Benutzerdefinierte Filtermodale",
     "customFiltersModalApplyFiltersTitle": "Filter anwenden",
-    "customFiltersModalCloseButton": "Schließen",
+    "optionsModalCloseButton": "Schließen",
```
Skaiir, line 152: Should we instead simply split into two keys?

tasklist/client/src/index.scss
```diff
@@ -70,3 +70,7 @@ body,
 svg {
   fill: currentcolor;
 }
+
+.cds--overflow-menu-options {
+  width: unset;
+}
```
vsgoulart, line 76: @volodymyr-melnykc Why is this necessary?

Skaiir: Right this is a fix to allow the sorting dropdown to expand. The default carbon component is super limited in width and they just don't support it being wider.

volodymyr-melnykc: @vsgoulart The width of the popover was fixed which caused the issue that the text in other languages wasn't fully displayed (truncated).

image

Skaiir: With this, it adapts to the width of its contents. I wonder if we want to improve upon this, but I think solving it the proper way (which is to ask carbon for a wider one lol) would probably take too long. We can follow up IMO.

tasklist/client/src/modules/internationalization/locales/de.json
volodymyr-melnykc: @Skaiir I renamed it because the close action is used in multiple dialogs (and can have the same tooltip).
Alternatively, we can split into separate keys.

Skaiir: Probably not a huge deal.

Skaiir (approved):

Skaiir:
Since we're happy about the width change, I'm happy to push this through. @vsgoulart any objections?

tasklist/client/src/index.scss
vsgoulart: @Skaiir  @volodymyr-melnykc We still had some ellipsis in Spanish for me. I fixed it now and made the style overwrite localized to that component only instead of globally like it was before.

Please check my last commit

vsgoulart (approved):

Issue #21427, opened 2024-08-21
CI: refactor all workflows to use new composite Yarn cache setup action


cmur2:
## Description

Part of #18750. The goal of this ticket is to ensure all GHA workflows using Yarn comply with the [caching strategy](https://github.com/camunda/camunda/wiki/CI-&-Automation#caching-strategy) by using the composite action created in #21424.

This can mostly be achieved by identifying all GHA workflows using Yarn, removing the `cache: "yarn"` and `cache-dependency-path: ${{ inputs.directory }}/yarn.lock` options from `setup-node` and inserting an additional step that uses #21424.

cmur2:
The change is effective and overall GHA cache usage is trending down. No NodeJS caches will be created on new PRs.

Issue #21424, opened 2024-08-21
CI: create composite action to set up Yarn cache


cmur2:
## Description

Part of #18750. Create a composite GHA that can be reused from other GHA workflows which sets up Yarn (v1, global cache) to use GHA cache according to [our caching strategy](https://github.com/camunda/camunda/wiki/CI-&-Automation#caching-strategy). That means the action should always try to restore from cache but only save the cache `on: push` to `main` or `stable*` branch builds.

See #21423 for how it is done on the backend side with Maven. See https://github.com/actions/setup-node/blob/26961cf329f22f6837d5f54c3efd76b480300ace/src/cache-utils.ts#L39-L66 for how the official `setup-node` action does set up Yarn to GHA cache. The official `setup-node` action always saves caches on PR builds, there is an [open feature request](https://github.com/actions/setup-node/issues/663) related to it.

cmur2:
Done in https://github.com/camunda/camunda/pull/21475 + https://github.com/camunda/camunda/pull/21495, caches are now getting created and used in few example builds:

![image](https://github.com/user-attachments/assets/3c2042dd-71d3-4440-9435-3dd4554f686d)

PR #21475, opened 2024-08-22
ci: add composite action to set up Yarn cache

cmur2:
## Description

This PRs adds a custom action that uses similar logic like [Zeebe has for Maven](https://github.com/camunda/camunda/blob/fd2e86a59785408bd39daa1134aa1d292281c390/.github/actions/setup-zeebe/action.yml#L113-L180) combined with Yarn-specific paths from https://github.com/actions/setup-node/blob/26961cf329f22f6837d5f54c3efd76b480300ace/src/cache-utils.ts#L39-L66 ([setup-node cannot do it](https://github.com/actions/setup-node/issues/663)) to implement our [caching strategy](https://github.com/camunda/camunda/wiki/CI-&-Automation#caching-strategy).

To demonstrate that this causes no problems I changed only 1 job to use it, see https://github.com/camunda/camunda/actions/runs/10508214639/job/29111637484?pr=21475 -- remaining jobs will be refactored in https://github.com/camunda/camunda/issues/21427

Old usage:

```yaml
    - name: Setup yarn cache
      uses: actions/setup-node@v4
      with:
        node-version: "20"
        cache: "yarn"
        cache-dependency-path: project-path/yarn.lock
```

New usage after this PR:

```yaml
    - name: Setup yarn cache
      uses: actions/setup-node@v4
      with:
        node-version: "20"
    - uses: ./.github/actions/setup-yarn-cache
      with:
        directory: project-path
```

As a test in https://github.com/camunda/camunda/pull/21475/commits/d4324461c70c9b2086034d3442b3b395bb2c31b6 I replaced all usages like this and indeed no **node** caches were created on the PR:

image

## Checklist

- [x] for CI changes:

## Related issues

Related to #21424

### DIFF
--- a/.github/actions/build-frontend/action.yml
+++ b/.github/actions/build-frontend/action.yml
@@ -19,8 +19,9 @@ runs:
       uses: actions/setup-node@v4
       with:
         node-version: "20"
-        cache: "yarn"
-        cache-dependency-path: ${{ inputs.directory }}/yarn.lock
+    - uses: ./.github/actions/setup-yarn-cache
+      with:
+        directory: ${{ inputs.directory }}
     - name: Install node dependencies
       working-directory: ${{ inputs.directory }}
       shell: bash
--- a/.github/actions/setup-yarn-cache/action.yml
+++ b/.github/actions/setup-yarn-cache/action.yml
@@ -1,0 +1,47 @@
+---
+name: Setup Yarn Cache
+
+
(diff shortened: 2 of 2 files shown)


vsgoulart (approved):
Overall it looks good to me. The only problem I see is that if any team decides to use a different package manager we will have to implement a special case for it, but we don't often change package managers anyway

I guess we can merge this test it

cmur2:
> if any team decides to use a different package manager

Fair point but I hope to foster some standardization here in the monorepo implicitly so this becomes rather unlikely 🤞🏼

PR #21487, opened 2024-08-22
feat: feature flag to disable partition scaling by default

### LINKED ISSUE #21463: Feature flag to disable partition scaling
Introduce a new configuration option to disable partition scaling by default.
A good place to put this might be `zeebe.broker.experimental.partitionScaling`. The default value is false.

When this feature flag is disabled, partition scaling must not leave any persisted state that could later pose issues for backwards compatibility.
In particular, we should not add new data to the `ClusterConfiguration`. Engine state is similar but there it might be acceptable to already keep some state because it might be easier to deal with backwards compatibility.

lenaschoenburg:
Very boring PR but the first one for partition scaling :tada: 

Adds a new experimental feature under `zeebe.broker.experimental.features.enablePartitionScaling` that is disabled by default. It's currently not used at all but will be soon.

Closes https://github.com/camunda/camunda/issues/21463


deepthidevaki (approved):
:rocket: First PR for the feature is to disable the feature :smile:

PR #21489, opened 2024-08-22
refactor: remove variable cleanup mode

HeleneW-dot:
## Description
Removes `variable` cleanup mode from Optimize C8.
This config is not mentioned in camunda-operator or camunda-platform-helm.

Please also review the follow up [docs PR.](https://github.com/camunda/camunda-docs/pull/4184)

## Checklist

- [ ] for CI changes:

## Related issues

related to [OPT#13211](https://github.com/camunda/camunda/issues/33560)


RomanJRW (changes requested):
I didn't actually review this - but why are we removing this? I don't see this as a C7 feature, and can even imagine some customers in SM using this configuration in C8

PHWaechtler:
> I didn't actually review this - but why are we removing this? I don't see this as a C7 feature, and can even imagine some customers in SM using this configuration in C8

This came from the removal tasks initially synced with Omran when we started working on the fork, I was informed this is not a C8 feature since in C8 we always clean up all instance data. Is this supposed to be available in C8?

PHWaechtler:
> > a C8 feature since in C8 we always clean up all instance data
> 
> Is this not only true in SaaS? I can imagine SM customers using the variable cleanup

The info I had was that its C7 only, but I'm following up to see if there was a specific source for that or if there was a misalignment somewhere. If its C8 too we can close this pr.

PHWaechtler:
> > > a C8 feature since in C8 we always clean up all instance data
> > 
> > 
> > Is this not only true in SaaS? I can imagine SM customers using the variable cleanup
> 
> The info I had was that its C7 only, but I'm following up to see if there was a specific source for that or if there was a misalignment somewhere. If its C8 too we can close this pr.

@RomanJRW [This](https://docs.google.com/spreadsheets/d/1MF_LUb5NFqx8kreYsxXLK3LM2KGXndun5LplUK0EZ6s/edit?gid=0#gid=0) is the sheet this info came from (see line 100), I will ping tobias to ask if thats outdated/incorrect.

RomanJRW:
@PHWaechtler - Sure, sounds good. Note that if we proceed with deleting, I think it would also need to be something we communicate in the upgrade guide as it would be a breaking change for anyone using it 👍. Please ensure Tobias is aware of this when making this decision

PHWaechtler:
Closing as we will keep this in C8. C8 feature sheet updated accordingly [here](https://docs.google.com/spreadsheets/d/1MF_LUb5NFqx8kreYsxXLK3LM2KGXndun5LplUK0EZ6s/edit?gid=0#gid=0). I think the same will apply to the process data cleanup config (line 99)

PR #21492, opened 2024-08-22
feat: show license status in identity UI

### LINKED ISSUE #20739: C8 License - Show C8 License status in Identity
## Related to

Epic: https://github.com/camunda/product-hub/issues/2359

Dependent On: https://github.com/camunda/camunda/issues/20540

## Overview

Once the License endpoint in the monorepo is available, the front end needs to consume data from the endpoint and display it to the user. Use the `C3Navigation` component to display the status. The label will only show in Self Managed, it will not show in SaaS

The endpoint: 
```
GET /license
{
   "validLicense": true,
   "licenseType": "self-managed"
}
```
`validLicense` will be `true` or `false`. The value is determined whether the Camunda license is valid or not
`licenseType` will return one of three strings, `saas`, `self-managed`, or `unknown`. The value reflects the current running mode Camunda is in. `unknown` is returned when a license is invalid or missing

The C3 Navbar component contains a feature which can easily display the license status. [See here for the prop to use](https://camunda-composite-components.pages.dev/?path=/docs/uishell-c3navigation--docs#navbar-prop). The license prop is available in `@camunda/camunda-composite-components` version `0.7.0` and higher

The value of `validLicense` maps directly to the `isProductionLicense` prop for the header. `show` will be `true` when `environment` is equal to `self-managed`

Console Self-Managed has implemented a license status already: https://github.com/camunda-cloud/camunda-cloud-management-apps/blob/ff518c29429d34363e9fd678b66da7762cadb7d2/apps/console-sm-frontend/src/components/navbar.tsx#L104

## Actions

- Consume the `camunda-license` endpoint and read the license status. The status is then displayed on the front end for the user

```[tasklist]
### Pull Requests
- [ ] tbd
```

## Example Screenshot with Navbar props

Screenshot 2024-08-07 at 7 25 17 AM

maryarm:
## Description
Once the License endpoint in the monorepo is available, the front end needs to consume data from the endpoint and display it to the user. Use the `C3Navigation` component to display the status. The label will only show in Self Managed, it will not show in SaaS

The endpoint: 
```
GET /license
{
   "validLicense": true,
   "licenseType": "self-managed"
}
```
`validLicense` will be `true` or `false`. The value is determined whether the Camunda license is valid or not
`licenseType` will return one of three strings, `saas`, `self-managed`, or `unknown`. The value reflects the current running mode Camunda is in. `unknown` is returned when a license is invalid or missing.

## Related issues

closes #20739

### DIFF
--- a/identity/client/src/components/layout/AppHeader.tsx
+++ b/identity/client/src/components/layout/AppHeader.tsx
@@ -1,10 +1,20 @@
 import { C3Navigation } from "@camunda/camunda-composite-components";
 import { useGlobalRoutes } from "src/components/global/useGlobalRoutes";
 import { useNavigate } from "react-router";
+import { useApi } from "src/utility/api";
+import { checkLicense } from "src/utility/api/headers";
 
 const AppHeader = () => {
   const routes = useGlobalRoutes();
   const navigate = useNavigate();
+  const { data: license } = useApi(checkLicense);
+  const showLicense: boolean =
+    license == null
+      ? false
+      : license.licenseType !== undefined &&
+        license.licenseType == "self-managed";
+  const isProductionLicense: boolean =
+    license == null ? false : license.validLicense;
 
   return (
     <C3Navigation
@@ -26,6 +36,10 @@ const AppHeader = () => {
             onClick: () => navigate(route.key),
           },
         })),
+        licenseTag: {
+          show: showLicense,
+          isProductionLicense: isProductionLicense,
+        },
       }}
     />
   );
--- a/identity/client/src/components/layout/AppHeader.tsx
+++ b/identity/client/src/components/layout/AppHeader.tsx
@@ -11,7 +11,7 @@ const AppHeader = () => {
   const showLicense: boolean =
     license == null
       ? true
-      : license.licenseType === undefined && license.licenseType != "saas";
+      : license.licenseType === undefined || license.licenseType != "saas";
   const isProductionLicense: boolean =
     license == null ? false : license.validLicense;
 
--- a/identity/client/src/components/layout/AppHeader.tsx
+++ b/identity/client/src/components/layout/AppHeader.tsx
@@ -10,8 +10,8 @@ const AppHeader = () => {
   const { data: license } = useApi(checkLicense);
   const showLicense: boolean =
     license == null
-      ? false
-      : license.licenseType !== undefined && license.licenseType != "saas";
+      ? true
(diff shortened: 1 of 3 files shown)


identity/client/src/components/layout/AppHeader.tsx
```diff
@@ -1,10 +1,20 @@
...
 
 const AppHeader = () => {
   const routes = useGlobalRoutes();
   const navigate = useNavigate();
+  const { data: license } = useApi(checkLicense);
+  const showLicense: boolean =
+    license == null
+      ? false
+      : license.licenseType !== undefined &&
+        license.licenseType == "self-managed";
```
matthewBearCamunda, line 15: `showLicense` should be looking for `saas`, that's the only case where it the license would be hidden. Otherwise, it would show

So, a small change is needed here

```
const showLicense: boolean =
    license == null
      ? false
      : license.licenseType !== undefined &&
        license.licenseType != "saas";
```

identity/client/src/pages/users/List.tsx
```diff
@@ -71,7 +71,7 @@ const List: FC = () => {
     <Page>
       <EntityList
         title={t("Users")}
-        data={userSearchResults!.items}
+        data={userSearchResults == null ? [] : userSearchResults.items}
```
matthewBearCamunda, line 74: Is this change needed for the license call to work?

matthewBearCamunda (changes requested):
One small change and one question!

identity/client/src/pages/users/List.tsx
maryarm: It's not directly related as it fixes another bug I discovered during the test, but without this the page can't be rendered and just shows error page.

identity/client/src/components/layout/AppHeader.tsx
```diff
@@ -1,10 +1,19 @@
...
+import { checkLicense } from "src/utility/api/headers";
 
 const AppHeader = () => {
   const routes = useGlobalRoutes();
   const navigate = useNavigate();
+  const { data: license } = useApi(checkLicense);
+  const showLicense: boolean =
+    license == null
+      ? true
+      : license.licenseType === undefined && license.licenseType != "saas";
```
matthewBearCamunda, line 14: This should have a `||` operator rather than `&&`

matthewBearCamunda (changes requested):

matthewBearCamunda (approved):
LGTM

PR #21496, opened 2024-08-22
build: optimize dependencies comply with dependency:analyze

### LINKED ISSUE #21630: Cleanup Optimize Maven dependencies
### Context

Currently, Optimize is present within the monorepo but it's not inheriting from the zeebe parent POM (i.e., `parent/pom.xml`). Soon, this change will need to happen and one of its implications is that Optimize will need to comply with the rules the maven-dependency-plugin is configured with, in such parent POM.

With this task, we're going to make Optimize compliant with these rules by:

* removing dependencies that are declared but not used by some of the Optimize's submodules
* correct the declaration of transient dependencies (i.e., dependencies that are used by some submodules but are not declared)
* restrict the scope to `test` for the dependencies that are used for tests only but have been declared with a wider scope

### Acceptance Criteria

* The refactoring should be applied as much as possible and the warnings/errors, that are given by the dependencies that - for some reason - are impossible to refactor, should be suppressed, to make the build green.
*  in practice, the command`mvn dependency:analyze`, executed from within the `optimize/` directory, shall not give warnings or errors and Optimize shall keep working as expected.

### Links

* https://github.com/camunda/product-hub/issues/2289

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/21496
```

## Task Lifecycle

#### Review Resources

- Feature PR: {Link to PR targeting the main branch}
- Preview environments: {Link(s) to preview environments}

#### Engineering Review

#### QA Review

### Completion

buccarel:
## Description

This PR fixes the problems related with the way some Optimize dependencies are declared.
For context, see https://github.com/camunda/camunda/issues/21630

The set of issues can be summarized as follows:
* some dependencies were imported as indirect, without being explicitly declared
* some dependencies were declared without being used
* some dependencies were declared too far from the  module that was actually using them, causing them to be imported by the intermediate modules too

Common test dependencies can be excluded from the analysis, as their `test` scope prevents them from being packed in the release artifact anyways.

In addition to the changes, as Optimize does not include the zeebe-parent configuration yet, a pipeline is being added to perform the dependency analysis on PR.

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/21630

### DIFF
--- a/.github/workflows/optimize-dependency-check.yml
+++ b/.github/workflows/optimize-dependency-check.yml
@@ -1,0 +1,50 @@
+# TODO: Remove this pipeline once Optimize inherits from zeebe-parent,
+#       with https://github.com/camunda/camunda/issues/22271
+name: Optimize Dependency Analysis
+env:
+  JAVA_VERSION: "21"
+on:
+  push:
+    branches:
+      - main
+  pull_request: { }
+  merge_group: { }
+  workflow_dispatch: { }
+  workflow_call: { }
+jobs:
+  dependency-analysis:
+    name: Dependency Analysis
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - name: Setup Java
+        uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73 # v4
+        with:
+          distribution: "adopt"
+          java-version: "21"
+          cache: "maven"
+      - name: Import Secrets
+        id: secrets
+        uses: hashicorp/vault-action@d1720f055e0635fd932a1d2a48f87a666a57906c # v3.0.0
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          method: approle
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secrets: |
+            secret/data/github.com/organizations/camunda NEXUS_USR;
+            secret/data/github.com/organizations/camunda NEXUS_PSW;
+      - name: "Create settings.xml"
+        uses: s4u/maven-settings-action@7802f6aec16c9098b4798ad1f1d8ac75198194bd # v3.0.0
+        with:
+          githubServer: false
+          servers: |
+            [{
+               "id": "camunda-nexus",
+               "username": "${{ steps.secrets.outputs.NEXUS_USR }}",
+               "password": "${{ steps.secrets.outputs.NEXUS_PSW }}"
+             }]
+          mirrors: '[{"url": "https://repository.nexus.camunda.cloud/content/groups/internal/", "id": "camunda-nexus", "mirrorOf": "*", "name": "camunda Nexus"}]'
+      - name: Run dependency analysis
+        run: mvn -f optimize/pom.xml dependency:analyze -DskipTests -DskipDocker -DdependencyAnalyzeFailOnWarning
--- a/optimize/backend/pom.xml
+++ b/optimize/backend/pom.xml
@@ -22,11 +22,15 @@
 
   <dependencies>
     <!-- Logging dependencies -->
+
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>net.logstash.logback</groupId>
       <artifactId>logstash-logback-encoder</artifactId>
       <version>8.0</version>
     </dependency>
+
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>org.codehaus.janino</groupId>
       <artifactId>janino</artifactId>
@@ -34,12 +38,6 @@
     </dependency>
 
     <!-- Optimize dependencies -->
-    <dependency>
-      <groupId>io.camunda.optimize</groupId>
-      <artifactId>client</artifactId>
-      <version>${project.version}</version>
-      <type>pom</type>
-    </dependency>
     <dependency>
       <groupId>io.camunda.optimize</groupId>
       <artifactId>optimize-commons</artifactId>
@@ -100,6 +98,8 @@
       <groupId>org.eclipse.jetty.http2</groupId>
       <artifactId>jetty-http2-server</artifactId>
     </dependency>
+
+    <!-- as per documentation, we keep these for http/2. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-alpn-java-client</artifactId>
@@ -108,6 +108,7 @@
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-alpn-java-server</artifactId>
     </dependency>
+
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-alpn-server</artifactId>
@@ -115,10 +116,6 @@
     <dependency>
       <groupId>org.eclipse.jetty.ee10</groupId>
       <artifactId>jetty-ee10-servlet</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.eclipse.jetty.ee10</groupId>
-      <artifactId>jetty-ee10-servlets</artifactId>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
@@ -134,6 +131,7 @@
     </dependency>
 
     <!-- IDENTITY-->
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>io.camunda</groupId>
       <artifactId>identity-spring-boot-starter</artifactId>
@@ -151,10 +149,14 @@
       <groupId>org.springframework</groupId>
       <artifactId>spring-context-support</artifactId>
     </dependency>
+
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-tx</artifactId>
     </dependency>
+
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-web</artifactId>
@@ -166,16 +168,21 @@
         </exclusion>
       </exclusions>
     </dependency>
+
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-jetty</artifactId>
       <version>${spring.boot.version}</version>
     </dependency>
+
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot</artifactId>
       <version>${spring.boot.version}</version>
     </dependency>
+
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-jersey</artifactId>
@@ -187,11 +194,7 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-actuator</artifactId>
-      <version>${spring.boot.version}</version>
-    </dependency>
+
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-web</artifactId>
@@ -217,11 +220,6 @@
       <artifactId>spring-security-oauth2-resource-server</artifactId>
       <version>${spring.security.version}</version>
     </dependency>
-    <dependency>
-      <groupId>com.nimbusds</groupId>
-      <artifactId>nimbus-jose-jwt</artifactId>
-      <version>${nimbus.jose.jwt.version}</version>
-    </dependency>
 
     <!-- Authentication -->
     <dependency>
@@ -232,18 +230,6 @@
         <exclusion>
           <artifactId>jackson-databind</artifactId>
           <groupId>com.fasterxml.jackson.core</groupId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-
-    <!-- Jackson/JSON -->
-    <dependency>
-      <groupId>com.fasterxml.jackson.jaxrs</groupId>
-      <artifactId>jackson-jaxrs-json-provider</artifactId>
-      <exclusions>
-        <exclusion>
-          <groupId>com.fasterxml.jackson.module</groupId>
-          <artifactId>jackson-module-jaxb-annotations</artifactId>
         </exclusion>
       </exclusions>
     </dependency>
@@ -259,18 +245,6 @@
           <groupId>org.slf4j</groupId>
         </exclusion>
       </exclusions>
-    </dependency>
-
-    <!-- Email framework -->
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-mail</artifactId>
-      <version>${spring.boot.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>jakarta.mail</groupId>
-      <artifactId>jakarta.mail-api</artifactId>
-      <version>2.1.3</version>
     </dependency>
 
     <dependency>
@@ -312,9 +286,11 @@
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
       <version>${commons.io.version}</version>
+      <scope>test</scope>
     </dependency>
 
     <!-- Micrometer Prometheus registry  -->
+    <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>io.micrometer</groupId>
       <artifactId>micrometer-registry-prometheus</artifactId>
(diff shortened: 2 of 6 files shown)


RomanJRW:
@buccarel - do you need this PR?

optimize/backend/pom.xml
```diff
@@ -21,25 +21,38 @@
...
   <dependencies>
+    <!-- fix build errors -->
+    <dependency>
+      <groupId>io.github.classgraph</groupId>
+      <artifactId>classgraph</artifactId>
+      <version>4.8.175</version>
+    </dependency>
+
     <!-- Logging dependencies -->
+    <!-- unused/declared
```
RomanJRW, line 32: I think this is actually used to support the `JSON_LOGGING` property in the `logback.xml` file 🤔 

I can see why this would be detected as unused though. Maybe a candidate for an exclusion?

optimize/backend/pom.xml
```diff
@@ -100,14 +113,18 @@
...
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-alpn-java-client</artifactId>
     </dependency>
+    -->
+    <!-- unused/declared
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-alpn-java-server</artifactId>
     </dependency>
+    -->
```
RomanJRW, line 127: I think these were added for HTTP2 support. It would be good to confirm that their exclusion won't break things. Maybe @matthewBearCamunda can provide more context here

optimize/backend/pom.xml
```diff
@@ -116,10 +133,12 @@
       <groupId>org.eclipse.jetty.ee10</groupId>
       <artifactId>jetty-ee10-servlet</artifactId>
     </dependency>
+    <!-- unused/declared
     <dependency>
       <groupId>org.eclipse.jetty.ee10</groupId>
       <artifactId>jetty-ee10-servlets</artifactId>
     </dependency>
+    -->
```
RomanJRW, line 141: I wonder what the difference between this and the one above is

optimize/backend/pom.xml
```diff
@@ -166,16 +190,20 @@
...
       </exclusions>
     </dependency>
+    -->
+    <!-- unused/declared
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-jetty</artifactId>
       <version>${spring.boot.version}</version>
     </dependency>
+    -->
```
RomanJRW, line 200: Removing all of these Spring ones are interesting. We should heavily test this to make sure we aren't breaking things

optimize/backend/pom.xml
```diff
@@ -187,11 +215,14 @@
...
       </exclusions>
     </dependency>
+    -->
+    <!-- unused/declared
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-actuator</artifactId>
       <version>${spring.boot.version}</version>
     </dependency>
+    -->
```
RomanJRW, line 225: I would be surprised if this could be removed, I think it is used to expose our Metrics. However, the metrics IT is currently in a state of skipped. Would be nice to reenable and get green again

optimize/backend/pom.xml
```diff
@@ -217,11 +248,13 @@
       <artifactId>spring-security-oauth2-resource-server</artifactId>
       <version>${spring.security.version}</version>
     </dependency>
+    <!-- unused/declared
     <dependency>
       <groupId>com.nimbusds</groupId>
       <artifactId>nimbus-jose-jwt</artifactId>
       <version>${nimbus.jose.jwt.version}</version>
     </dependency>
+    -->
```
RomanJRW, line 257: Hmm this might be used for C8SaaS auth. We wouldn't have an automated test for this case, but we should at least test it manually to be sure we aren't breaking SaaS auth

optimize/backend/pom.xml
```diff
@@ -308,18 +345,22 @@
...
+      <scope>test</scope>
     </dependency>
 
     <!-- Micrometer Prometheus registry  -->
+    <!-- unused/declared
     <dependency>
       <groupId>io.micrometer</groupId>
       <artifactId>micrometer-registry-prometheus</artifactId>
       <version>1.13.4</version>
     </dependency>
```
RomanJRW, line 362: same as the actuator one, I think we would actually need this for metrics

optimize/backend/pom.xml
```diff
@@ -385,19 +427,23 @@
...
       <version>${apache.http5-core.version}</version>
     </dependency>
-
+    -->
+    <!-- unused/declared
     <dependency>
       <groupId>org.apache.httpcomponents.core5</groupId>
       <artifactId>httpcore5-h2</artifactId>
       <version>${apache.http5-core.version}</version>
     </dependency>
```
RomanJRW, line 445: possibly also http2 related

optimize/backend/pom.xml
```diff
@@ -416,6 +462,397 @@
       <artifactId>camunda-license-check</artifactId>
       <version>2.9.0</version>
     </dependency>
+
+    <!-- used/undeclared -->
+    <dependency>
+      <groupId>org.mock-server</groupId>
+      <artifactId>mockserver-core</artifactId>
+      <version>5.15.0</version>
```
RomanJRW, line 470: I think this, and many of the ones below, already have properties defined for their version. I know this is WIP but we should use them where possible instead of having their version here

optimize/backend/pom.xml
```diff
@@ -416,6 +462,397 @@
...
+      <groupId>io.netty</groupId>
+      <artifactId>netty-common</artifactId>
+      <version>4.1.112.Final</version>
+      <scope>compile</scope>
+    </dependency>
+
+    <!-- unused/declared
+    <dependency>
+      <groupId>org.eclipse.angus</groupId>
+      <artifactId>jakarta.mail</artifactId>
```
RomanJRW, line 788: This feels like possibly something that could be deleted, can't recall us using this

optimize/backend/pom.xml
```diff
@@ -416,6 +462,397 @@
...
+      <version>4.4</version>
+      <scope>compile</scope>
+    </dependency>
+    <!-- test-only -->
+    <dependency>
+      <groupId>commons-collections</groupId>
+      <artifactId>commons-collections</artifactId>
+      <version>3.2.2</version>
+      <scope>test</scope>
+    </dependency>
```
RomanJRW, line 806: I wonder if this can be deleted and we just start using the collections4 version

optimize/backend/pom.xml
```diff
@@ -416,6 +462,397 @@
...
+      <version>3.1.8</version>
+      <scope>compile</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-actuator</artifactId>
+      <version>3.3.3</version>
+      <scope>compile</scope>
+    </dependency>
```
RomanJRW, line 834: I realised that I made a lot of comments above that relate to dependencies that have been moved rather than removed. Feel free to ignore any of the above where that is the case

optimize/qa/schema-integrity-tests/pom.xml
```diff
@@ -46,9 +46,71 @@
...
+      <version>4.5.14</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.elasticsearch</groupId>
+      <artifactId>elasticsearch</artifactId>
+      <version>7.17.24</version>
+      <scope>test</scope>
+    </dependency>
```
RomanJRW, line 78: What is this dependency? This is not even the client?

optimize/util/optimize-commons/pom.xml
```diff
@@ -77,25 +77,31 @@
       <artifactId>json-path</artifactId>
       <version>${jsonpath.version}</version>
     </dependency>
+    <!-- unused/declared
     <dependency>
       <groupId>org.yaml</groupId>
       <artifactId>snakeyaml</artifactId>
       <version>2.3</version>
     </dependency>
```
RomanJRW, line 85: pretty sure that this was added to fix a CVE in the past. If this does not expose us (we could check via Snyk/trivy), then happy to remove

optimize/util/optimize-commons/pom.xml
```diff
@@ -77,25 +77,31 @@
...
+    -->
 
     <!-- Logging -->
+    <!-- unused/declared
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>jul-to-slf4j</artifactId>
       <version>${slf4j.version}</version>
     </dependency>
```
RomanJRW, line 103: I would like us to understand whether this can actually be removed. It might be to allow a flexible log adaptor

optimize/util/optimize-commons/pom.xml
```diff
@@ -112,11 +118,13 @@
...
     <!-- needed for elasticsearch logs to be available over slf4j -->
+    <!-- unused/declared
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-to-slf4j</artifactId>
```
RomanJRW, line 124: same as above

optimize/util/optimize-commons/pom.xml
```diff
@@ -134,22 +142,26 @@
       <artifactId>commons-lang3</artifactId>
       <version>3.17.0</version>
     </dependency>
+    <!-- unused/declared
     <dependency>
       <groupId>com.ethlo.time</groupId>
       <artifactId>itu</artifactId>
       <version>1.10.2</version>
     </dependency>
```
RomanJRW, line 150: I think this is used but it's in a commented out test. This scenario makes things a little more difficult because it's also probably true for a bunch of other tests

optimize/util/optimize-commons/pom.xml
```diff
@@ -199,16 +211,20 @@
...
       <version>${awssdk.version}</version>
     </dependency>
+    -->
+    <!-- unused/declared
     <dependency>
       <groupId>software.amazon.awssdk</groupId>
       <artifactId>sts</artifactId>
       <version>${awssdk.version}</version>
     </dependency>
+    -->
```
RomanJRW, line 227: I think this was added with OpenSearch in mind recently. Would be good to check with Giuliano before we delete this

RomanJRW (commented):
Some initial notes on some of the dependencies, mostly for more context for now.

Regarding the tests:
- The e2e test failures are probably because Optimize is not building as a result of these dependency changes. It's not super clear what the issue is, but we can look together later
- The Schema Integrity tests can be ignored, they need fixing on main too (WIP)
- The Zeebe IT failures look legit. Also not clear at a glance what the issue is, but perhaps something we can look at together in more detail. We can run these locally and debug if needed

optimize/backend/pom.xml
buccarel: If that's the case, these are very much the perfect candidates for the warning/error suppression indeed.

optimize/backend/pom.xml
buccarel: My understanding is that the one with the extra 's' extends the one without the extra 's' with more filters and functionalities. Link: https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/ee10/servlets/package-summary.html

optimize/backend/pom.xml
buccarel: Maybe we could leave it commented out, for the moment, instead of deleting it and with a TODO?

optimize/util/optimize-commons/pom.xml
buccarel: I just tried to run Snyk and I got this output with and without that dependency:
```
~/workspace/camunda/optimize/util/optimize-commons (optimize-dependencies-bis) » snyk test       mattia.buccarella@Mattias-MacBook-Pro

Testing /Users/mattia.buccarella/workspace/camunda/optimize/util/optimize-commons...

Organization:      buccarel
Package manager:   maven
Target file:       pom.xml
Project name:      io.camunda.optimize:optimize-commons
Open source:       no
Project path:      /Users/mattia.buccarella/workspace/camunda/optimize/util/optimize-commons
Licenses:          enabled

✔ Tested 105 dependencies for known issues, no vulnerable paths found.

Next steps:
- Run `snyk monitor` to be notified about new related vulnerabilities.
- Run `snyk test` as part of your CI/test.


~/workspace/camunda/optimize/util/optimize-commons (optimize-dependencies-bis) »
```
Even the number of tested dependencies stays the same at 105.

optimize/util/optimize-commons/pom.xml
buccarel: I think we can still leave them commented out with a TODO.
Once we re-enable the tests, they would stop the compilation if the dependency is not present, so the problem would be visible

optimize/util/optimize-commons/pom.xml
buccarel: Tagging @grlimacan

optimize/backend/pom.xml
RomanJRW: Or we just test metrics on this branch before merging. One way to do this would be reenabling the MetricsITs (might require finding out what the initial problem with it was anyway)

optimize/util/optimize-commons/pom.xml
RomanJRW: ok nice, perhaps the upstream dependency (I can't remember what it was) finally updated to a patched version of snakeyaml. In that case, this is probably good to rip out

optimize/util/optimize-commons/pom.xml
RomanJRW: I think we are more likely to delete the tests than reenable them, and build again from scratch. The delta between Optimize 7 and 8 ITs is quiet great, and reenabling/fixing them is difficult

optimize/util/optimize-commons/pom.xml
grlimacan: This needs to stay unfortunately, this is the weird way that the AWS integration works, which is that if those packages are present it knows it needs to use it. I am not sure how exactly this is handled in the background by AWS

optimize/backend/pom.xml
matthewBearCamunda: This dependency was not added with my http2 work for Optimize. @grlimacan - I saw you added this dependency for OPT-7226, OPT-7195 and OPT-7197 for OpenSearch work

optimize/backend/pom.xml
matthewBearCamunda: These were indeed added for http2 support

However, I just did a test where I removed these dependencies, and I was still able to get http2 to work. These should be safe for removal

As part of the testing process for this ticket, I think we should also do some https & http2 tests just to confirm

optimize/util/optimize-commons/pom.xml
buccarel: Ok then I'll re-include it and I'll add it to the ignore set.
Thanks.

optimize/backend/pom.xml
buccarel: As we agreed on slack, we're keeping these as per documentation's suggestion

optimize/backend/pom.xml
grlimacan: Yep, we need that for the OpenSearch client. Is thee a problem?

.github/workflows/optimize-dependency-check.yml
```diff
@@ -0,0 +1,49 @@
+# TODO: Remove this pipeline once Optimize inherits from zeebe-parent
```
grlimacan, line 1: Please add a ticket number here where this gets implemented, so that we don't overlook a TODO

optimize/backend/pom.xml
```diff
@@ -100,6 +98,8 @@
       <groupId>org.eclipse.jetty.http2</groupId>
       <artifactId>jetty-http2-server</artifactId>
     </dependency>
+
+    <!-- as per docomentation, we keep these for http/2. Dependency:analyze error suppressed -->
```
grlimacan, line 102:
```suggestion
    <!-- as per documentation, we keep these for http/2. Dependency:analyze error suppressed -->
```

optimize/pom.xml
```diff
@@ -444,6 +434,28 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-dependency-plugin</artifactId>
           <version>3.8.0</version>
+          <configuration>
+            <failOnWarning>${dependencyAnalyze.failOnWarning}</failOnWarning>
+            <ignoredUnusedDeclaredDependencies combine.children="append">
+              <!-- TODO: remove once Lombok is removed from the project -->
```
grlimacan, line 440: Please add a ticket number here where this TODO will get solved

optimize/qa/schema-integrity-tests/pom.xml
```diff
@@ -20,14 +20,16 @@
     <dependency>
       <groupId>org.elasticsearch.client</groupId>
       <artifactId>elasticsearch-rest-high-level-client</artifactId>
-      <version>7.17.24</version>
+      <version>${elasticsearch.client.test.version}</version>
```
grlimacan, line 23: I don't think this is correct. The integrity tests are the only part of the code that are still using elastic search 7, I believe the value you are referencing is an elastic search 8, this will likely lead to the test failing

optimize/util/optimize-commons/pom.xml
```diff
@@ -242,9 +214,115 @@
...
   <build>
+     <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-dependency-plugin</artifactId>
+          <version>3.8.0</version>
+          <configuration>
+            <ignoredUnusedDeclaredDependencies combine.children="append">
+              <!-- TODO: remove once Zeebe's parent is Optimize's parent -->
```
grlimacan, line 314: Add reference to a ticket number

grlimacan (approved):
Pre-approving, just some minor suggestions

PR #21497, opened 2024-08-22
feat: resolve called processes for binding type `versionTag`

### LINKED ISSUE #21038: Resolve called process to correct version for call activities with binding type `versionTag`

jfriedenstab:
## Description
Call activities that use the new binding type `versionTag` now call the version of the target process referenced by the specified version tag.

## Related issues
closes #21038

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/CallActivityProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/CallActivityProcessor.java
@@ -76,7 +76,8 @@ public final class CallActivityProcessor
     return evaluateProcessId(context, element)
         .flatMap(
             processId ->
-                getProcessForProcessIdAndBindingType(processId, element.getBindingType(), context))
+                getCalledProcess(
+                    processId, element.getBindingType(), element.getVersionTag(), context))
         .flatMap(this::checkProcessHasNoneStartEvent)
         .flatMap(p -> eventSubscriptionBehavior.subscribeToEvents(element, context).map(ok -> p))
         .thenDo(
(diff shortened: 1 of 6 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/CallActivityProcessor.java
```diff
@@ -206,15 +207,16 @@ private Either<Failure, DirectBuffer> evaluateProcessId(
...
       final ZeebeBindingType bindingType,
+      final String versionTag,
       final BpmnElementContext context) {
     return switch (bindingType) {
       case deployment -> getProcessVersionInSameDeployment(processId, context);
       case latest -> getLatestProcessVersion(processId, context.getTenantId());
-      // will be implemented with https://github.com/camunda/camunda/issues/21038
-      case versionTag -> Either.left(new Failure("Binding type 'versionTag' not supported"));
+      case versionTag ->
+          getLatestProcessVersionWithVersionTag(processId, versionTag, context.getTenantId());
```
koevskinikola, line 219: ❓ Should we have a `default` case that throws an error since users may manually modify a BPMN XML and provide a faulty value for the binding type?

koevskinikola (changes requested):
The code looks good @jfriedenstab 🎉  I just have a single question regarding error handling.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/container/CallActivityProcessor.java
jfriedenstab: Actually, that case should already be covered by the `ZeebeBindingTypeValidator`, so users would not be able to deploy a model with an invalid binding type in the first place 🙂.

koevskinikola (approved):
All questions answered @jfriedenstab, 👍 from me!

PR #21498, opened 2024-08-22
feat: resolve called decisions for binding type `versionTag`

### LINKED ISSUE #21040: Resolve called decision to correct version for business rule tasks with binding type `versionTag`

jfriedenstab:
## Description
Business rule tasks that use the new binding type `versionTag` now call the version of the target decision referenced by the specified version tag.

## Related issues
closes #21040

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnDecisionBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnDecisionBehavior.java
@@ -78,7 +78,7 @@ public final class BpmnDecisionBehavior {
 
     final var decisionId = decisionIdOrFailure.get();
     final var decisionOrFailure =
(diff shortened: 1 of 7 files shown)


koevskinikola (approved):
LGTM @jfriedenstab 👍

Issue #21472, opened 2024-08-22
Make benchmarks more realistic


Zelldon:
## Description

> As of now, we use quite simple processes with only one service task. We normally just increase the  load by increasing the process instance creation/completion per second (PI/s).
> 
> In the past, we discussed several times to improve our coverage and use different process models. What we did is to run on a weekly basis a benchmark with a higher load, with still the one task process, but also a mixed benchmark which has a process with a message start event and a task, plus a process with a timer. Of course likely not really covering the reality, but at least covering some more symbols.
> 
> We want to further improve and make the benchmarks more realistic, by using a process model that covers the average process orchestration use case and a payload that reflects the reality.
> 
> This is not only useful for this EPIC or topic, but also in general to improve our benchmarking approach and to cover more components in the future (the ultimate target is to cover the Camunda Platform as a whole). Results will be more understandable for Customers/Users and will help in Consulting to point to different results

related to https://github.com/camunda/camunda/issues/21462

### Break down

```[tasklist]
### Pre-requisites
- [x] [Reached out to consulting](https://app.slack.com/client/T0PM0P1SA/C01Q05D6YUU) for input to model a more realistic (Avg) process model
- [x] [Modeled more realistic model](https://camunda.slack.com/archives/C01Q05D6YUU/p1724306840107089?thread_ts=1724138721.573489&cid=C01Q05D6YUU)
- [x] [Found a good enough payload for the process instance](https://camunda.slack.com/archives/C01Q05D6YUU/p1724312774772399?thread_ts=1724138721.573489&cid=C01Q05D6YUU)
```

```[tasklist]
### Todo's
- [x] https://github.com/camunda/camunda/pull/21501
- [x] Add complex process model to benchmark application https://github.com/camunda/camunda/pull/21501
- [x] Add suitable payload to benchmark application, generate a payload that allows a multi-instance [of around 50 elements] (https://camunda.slack.com/archives/C01Q05D6YUU/p1724312774772399?thread_ts=1724138721.573489&cid=C01Q05D6YUU) https://github.com/camunda/camunda/pull/21501
- [x] Extend starter to generate business key, if configured (variable name should be configurable to be used as `customerId`) https://github.com/camunda/camunda/pull/21501
- [x] Extend starter to deploy extra process models, extending to be backwards compatible https://github.com/camunda/camunda/pull/21501
- [ ] <Optional> (Not clear yet) Extend starter to generate different payload?
[...]

Issue #21462, opened 2024-08-22
[EPIC]: Extend benchmarking approach to measure Operate import performance


Zelldon:
## Description

As part of https://github.com/camunda/camunda/issues/16912, we want to tackle an ultimate user problem, reducing the Operate import delay. This means the time until the user sees data in Operate.

The current idea, which is followed in #16912, is to reduce the import latency, by moving the importing logic into the exporting, reducing additional index hops and unnecessary data conversions. 

As this is an important performance topic, we want to make sure that we are on the right track and guide the implementation phase with benchmarks, which are realistic and which include Operate measurements (especially import latency measurements and other measures).

This topic is divided into the following sub-topics:

1. improve benchmarks in general
2. make the benchmarking more realistic
3. improve measurements (via adding more metrics and dashboards)
4. execute benchmarks

### Improved benchmarks

We need to revamp the benchmarks, especially our [Helm charts](https://github.com/zeebe-io/benchmark-helm), upgrading to the newest versions and making them more extensible.

```[tasklist]
### Benchmark improvements
- [x] https://github.com/zeebe-io/benchmark-helm/pull/171 - Migrate to the last helm chart version (`10.2.1`)
- [ ] https://github.com/camunda/camunda-platform-helm/pull/2183
- [ ] https://github.com/camunda/camunda-platform-helm/issues/2197
- [x] https://github.com/camunda/camunda/pull/20937 - Fix issues in benchmark workflows, for new chart version
- [x] https://github.com/camunda/camunda-platform-helm/pull/2223 - To fix configuration for Operate/Tasklist
- [x] https://github.com/zeebe-io/benchmark-helm/pull/176 - Alpha version already contained cfg fixes
- [x] https://github.com/zeebe-io/benchmark-helm/pull/173
- [ ] https://github.com/camunda/camunda/pull/21031
- [ ] https://github.com/camunda/camunda/pull/21129
- [x] Make starter extensible https://github.com/zeebe-io/benchmark-helm/pull/184
- [x] Make worker extensible https://github.com/zeebe-io/benchmark-helm/pull/183
```

### Realistic benchmarks

As of now, we use quite simple processes with only one service task. We normally just increase the load by increasing the process instance creation/completion per second (PI/s).

[...]

PR #21501, opened 2024-08-22
Extend benchmark application to make it more realistic

ChrisKujawa:
### Context
In context of https://github.com/camunda/camunda/issues/21472 and https://github.com/camunda/camunda/issues/21462 we want to improve our current benchmarks and make them more realistic

> As of now, we use in our benchmarks quite simple processes with only one service task. We normally just increase the load by increasing the process instance creation/completion per second (PI/s).
> 
> In the past, we discussed several times to improve our coverage and use different process models. What we did is to run on a weekly basis a benchmark with a higher load, with still the one task process, but also a mixed benchmark which has a process with a message start event and a task, plus a process with a timer. Of course likely not really covering the reality, but at least covering some more symbols.
> 
> We want to further improve and make the benchmarks more realistic, by using a process model that covers the average process orchestration use case and a payload that reflects the reality.

For that, I reached out to consulting [here](https://camunda.slack.com/archives/C01Q05D6YUU/p1724226341620789), see also a similar slack thread from @felix-mueller [here](https://camunda.slack.com/archives/C01Q05D6YUU/p1724138721573489).
After this, I paired with @ingorichtsmeier to model a more realistic process model that we can use in our benchmarks. 

The process model is based on a blueprint we provide in the marketplace, see [here](https://marketplace.camunda.com/en-US/apps/449510/credit-card-fraud-dispute-handling). Created by @till-stadtler. We extended it a bit with multi-instances and a call activity. 

After extending it looks like this now:

Bank_customer_complaint_dispute_handling

The called process is quite simple and looks like this:

refundingProcess

The DMN hasn't been changed and looks like this:

### What the PR contains: 

The PR:

* adds the process models from above
* adds a corresponding payload
* extends the starter to support multiple process models
* extends the starter to generate a business key

The payload has been taken from the blueprint, extended and adjusted. For example the `disputPositions` has been generated (via https://json-generator.com/), such that we run around 50 multi-instance instances, as discussed on [slack](https://camunda.slack.com/archives/C01Q05D6YUU/p1724312774772399?thread_ts=1724138721.573489&cid=C01Q05D6YUU).

To deploy multiple process models via the starter, a new property has been added. That allows to configure "extra" process models. This has been done like this, to be backwards compatible with the previous benchmarks.

To generate a business key for each process instance a new property has been added to the starter. That allows to configure the business key variable name. This will allow us to generate a unique `customerId`, which needs to be used for the message correlation keys, etc.

## Related issues

related to https://github.com/camunda/camunda/issues/21472


Zelldon:
The changes have been tested with a recent benchmark and the following configuration for the starter:

```
- name: JDK_JAVA_OPTIONS
          value: -Dconfig.override_with_env_vars=true -Dapp.brokerUrl=zeebe-benchmark-test-zeebe-gateway:26500
            -Dapp.starter.rate=1 -Dapp.starter.durationLimit=0 -Dzeebe.client.requestTimeout=62000
            -Dapp.starter.processId="bankDisputeHandling" -Dapp.starter.bpmnXmlPath="bpmn/realistic/bankCustomerComplaintDisputeHandling.bpmn"
            -Dapp.starter.extraBpmnModels.0="bpmn/realistic/determineFraudRatingConfidence.dmn"
            -Dapp.starter.extraBpmnModels.1="bpmn/realistic/refundingProcess.bpmn"
            -Dapp.starter.businessKey="customerId" -XX:+HeapDumpOnOutOfMemoryError
```

Next step is here to make it more configurable in the Benchmark helm charts.


We can see that the Process model is deployed and instances are created.




The decision is also successfully created.




In our new dashboard we can follow the constant small load
Peek 2024-08-23 12-31

npepinpe:
@rodrigo-lourenco-lopes - no need for you to do a super thorough review (but you're encouraged of course), but as this is likely something you will run, it's good if you familiarize yourself with it already.

npepinpe (approved):
:rocket: 

- :question: What's the purpose of this benchmark? Will it replace our standard single task benchmark? Will it replace the tracked latencies in the overall benchmark dashboard? Or will it just be an additional type of benchmark we now run?
- :question: We were thinking of using this for the benchmarking effort to find new cluster sizes. Do you think this is appropriate?
- :wrench: Now that we have a more complex process, should we add process tests for it as well to make sure it doesn't break on new changes?

Zelldon:
> 🚀
> 
> * ❓ What's the purpose of this benchmark? 

Please see PR description. In essence: `We want to further improve and make the benchmarks more realistic, by using a process model that covers the average process orchestration use case and a payload that reflects the reality.` This should allow us to run benchmarks which are comparable with real world scenarios and users. Especially useful in communicating with users and customer may it be in blog posts, support or consulting.


> Will it replace our standard single task benchmark? 

That is my ultimate goal yes. But not directly, as I have to run them first and to get a better understand how it works and performs, furthermore of course I wanted to sync with the Zeebe team first before completely replacing it.



> Will it replace the tracked latencies in the overall benchmark dashboard? Or will it just be an additional type of benchmark we now run?

See above. 
> * ❓ We were thinking of using this for the benchmarking effort to find new cluster sizes. Do you think this is appropriate?

Yes I would love if we would reuse this for this. Because I'm pretty sure that with that we easily come to our current limits (speaking of resources), please make sure to enable all components to have a more realistic picture. (not only zeebe)

> * 🔧 Now that we have a more complex process, should we add process tests for it as well to make sure it doesn't break on new changes?

Changes on the process model you mean ?

ingorichtsmeier:
Hi @Zelldon,

I checked the generated payload and have two remarks:
1. The `index` element is not required, as Zeebe will generate them with the multi instance. The upper object is an array with a given sequence, no extra index needed here.
1. The `_id` element could be a real world transaction ID, to enable idempotency for each single payment. To emphasize it, please remove the leading underscore.

ingorichtsmeier:
Hi @Zelldon,

I don't know the internal structure of the benchmarks, but for a realistic example, it would be good to be able to package the process diagrams together with the workers and starter logic into a single process application, as we describe it in our best practices here: https://docs.camunda.io/docs/next/components/best-practices/development/connecting-the-workflow-engine-with-your-world/#programming-glue-code.

It is OK, if you run your benchmarks in a different from the proposed setup. But for others, to run a single process, it should end up in a single docker container containing all required libraries. 

If you like, we can talk about the internals of a benchmark and a real world process application and how to support both approaches.

Zelldon:
> Hi @Zelldon,
> 
> I checked the generated payload and have two remarks:


Hey @ingorichtsmeier sorry for the delay. Last week I had no time to respond or work on this.

> 
> 1. The `index` element is not required, as Zeebe will generate them with the multi instance. The upper object is an array with a given sequence, no extra index needed here.


Thanks for your input. I don't see this as an issue. The index I added on purpose is to see how many elements we have (on generation time but also later), especially when looking at it during development or in later blog posts, etc. (to not always count). I don't think that this is so unrealistic to have.


> 2. The `_id` element could be a real world transaction ID, to enable idempotency for each single payment. To emphasize it, please remove the leading underscore.

Exactly. I think underscore is also fine here.

Zelldon:
> Hi @Zelldon,
> 
> I don't know the internal structure of the benchmarks, but for a realistic example, it would be good to be able to package the process diagrams together with the workers and starter logic into a single process application, as we describe it in our best practices here: https://docs.camunda.io/docs/next/components/best-practices/development/connecting-the-workflow-engine-with-your-world/#programming-glue-code.


Not sure what you exactly mean, but yes the starter is bundled together with the process model. It will use it to deploy the configured process model and start an instance. 


> 
> It is OK, if you run your benchmarks in a different from the proposed setup. But for others, to run a single process, it should end up in a single docker container containing all required libraries.

Not sure whether I would agree with this. Performance wise it makes much more sense to scale out workers, and separate them. Maybe we need to change our recommendations then.

> 
> If you like, we can talk about the internals of a benchmark and a real world process application and how to support both approaches.

Zelldon:
>It is OK, if you run your benchmarks in a different from the proposed setup. But for others, to run a single process, it should end up in a single docker container containing all required libraries.

PR #21524, opened 2024-08-23
Migrate timer event subprocess

### LINKED ISSUE #21218: Migrate timer event subprocess
## Description

### Migrate active timer event subprocess (interrupting or non-interrupting)
Screenshot 2024-08-23 at 12 18 50

For interrupting active timer event subprocess mappings should be provided between event subprocesses (sub1 -> sub2) and active elements inside event subprocesses (B -> B). If the start event of timer event subprocess is non-interrupting a mapping should be provided for active elements outside of the event subprocess as well (A -> A).

### Migrate timer event subprocess start event subscription
Screenshot 2024-08-23 at 12 27 33

To migrate the start event subscription of a timer event subprocess a mapping needs to be provided for active elements (A-> A) and between timer start events (start1 -> start2). This case is especially useful when user doesn't want the timer start event to be re-created because when a timer event is re-created its timer is reset.

### Subscribe to timer event subprocess start event subscription
Screenshot 2024-08-23 at 12 30 38

In this case, a mapping should be provided for the active element (A -> A). The timer event subscription will be created after the migration for timer start event (start2).

### Unsubscribe from timer event subprocess start event subscription
Screenshot 2024-08-23 at 12 32 42

In this case, a mapping should be provided for the active element (A -> A). The timer event subscription will be deleted after the migration for timer start event (start1).

berkaycanbc:
## Description

Please review this PR after #21217 because this PR is created on top of that PR.

## Related issues

closes #21218

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -223,12 +223,14 @@ public class ProcessInstanceMigrationMigrateProcessor
     requireUnchangedFlowScope(
         elementInstanceState, elementInstanceRecord, targetProcessDefinition, targetElementId);
     requireNoEventSubprocessInSource(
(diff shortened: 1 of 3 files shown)


berkaycanbc:
@korthout I rebased changes and send commits based on the review comments in #21217. Please take a look. 👀

korthout (approved):
Thanks @berkaycanbc 💯 

LGTM 👍

PR #21525, opened 2024-08-23
ci: build frontend for operate benchmarks

ChrisKujawa:
## Description

When benchmarking with Operate we want to have the frontend, previously we skipped this which doesn't allow us to investigate Operate UI during benchmarks.


lenaschoenburg (approved):

pedesen:
Hey @Zelldon it is preferred to build the Operate frontend separately instead of building it during the maven step. We changes this recently, because it caused some issue in the CI (see https://github.com/camunda/camunda/pull/20880).

You can use the build-operate-fe action for this:

```
- name: Build frontend
    uses: ./.github/actions/build-operate-fe
```

Zelldon:
Thanks @pedesen  :) 

True I remember that we changed that. I will do a follow up :)

Issue #21465, opened 2024-08-22
`ClusterConfiguration` contains routing info


lenaschoenburg:
For partition scaling, the `ClusterConfiguration` must contain information that helps the gateway route requests to the right partitions.

1. Partitions which are available for round-robin requests such as starting new process instances, activating jobs etc.
2. How to route messages

We need to distinguish them because message relocation will be implemented later and partitions become available for round-robin requests much earlier than they participate in messages.

This state should be initialized once from static configuration.
If the feature flag for partition scaling is disabled, the state should _not_ be set and remain empty.

lenaschoenburg:
We can hopefully copy the approach from https://github.com/camunda/camunda/pull/19329 to initialize this new field for the first time.

PR #21529, opened 2024-08-23
feat: add optional routing state to cluster configuration

lenaschoenburg:
This PR adds a new optional field `routingState` to the dynamic cluster configuration.
This field is not yet initialized in any way, this will be done as a follow-up.

Routing state is versioned. We expect that it will be the coordinators responsibility to update it as new partitions are added and become ready. The version is useful for other members to merge updates received out of order.

The routing state itself consists of two fields, `activePartitions` and `messageCorrelation`.

`activePartitions` is a set of partition ids. This will be used later on by the gateway to route requests such as creating new process instances, activating jobs etc.

`messageCorrelation` describes the current strategy for routing messages. The only supported strategy right now is our default "hash then mod over partition count".

The routing state is pure data, it does not implement any interfaces that would help the gateway or engine to actually make use of that data yet.

Relates to https://github.com/camunda/camunda/issues/21465

### DIFF
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/serializer/ProtoBufSerializer.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/serializer/ProtoBufSerializer.java
@@ -49,9 +49,12 @@ import io.camunda.zeebe.dynamic.config.state.ExporterState;
 import io.camunda.zeebe.dynamic.config.state.ExportersConfig;
 import io.camunda.zeebe.dynamic.config.state.MemberState;
 import io.camunda.zeebe.dynamic.config.state.PartitionState;
+import io.camunda.zeebe.dynamic.config.state.RoutingState;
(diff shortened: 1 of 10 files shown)


deepthidevaki (approved):
:rocket: 

:wrench: It would be good to verify encoding/decoding of routingState in `ProtoBufSerializerTest`. 
:thought_balloon: An update method to update routingState should be added in ClusterConfiguration at some point. But this can be added later when we need it.
:question: Should RoutingState be part of the query response `GET actuator/cluster`? May be not now.  But you can create a follow up issue to add it in the response.

lenaschoenburg:
> 🔧 It would be good to verify encoding/decoding of routingState in ProtoBufSerializerTest.

I've added a test case now. The serialization is already tested through the property test though.

>  💭 An update method to update routingState should be added in ClusterConfiguration at some point. But this can be added later when we need it.

Yep, I will add it when needed :+1: 

>  ❓ Should RoutingState be part of the query response GET actuator/cluster? May be not now. But you can create a follow up issue to add it in the response.

Good point, I haven't thought of that. I'll add it to the tasklist.

PR #21532, opened 2024-08-23
feat: ExporterBatchWriter can register handlers, handles records, and flush records to a batch

### LINKED ISSUE #21531: Exporter can register handlers, handles records, and flush records to a batch

mustafadagher:
## Description

ExporterBatchWriter can register handlers, handles records, and flush records to a batch

## Related issues

closes #21531


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/store/ExporterBatchWriterTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ExporterBatchWriterTest {
+  private ExporterBatchWriter batchWriter;
+  private TestExportHandler handler;
+
+  @BeforeEach
+  void setUp() {
+    handler = mock(TestExportHandler.class);
```
deepthidevaki, line 36: :wrench: If we are only using the mock, then there is no need to define a new class `TestExportHandler`.

I suppose you can just use the base class as follows.
```
 private ExportHandler<TestExporterEntity, TestValue> handler;
 handler = mock(ExportHandler.class);
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Caches exporter entities of different types and provide the method to flush them in a batch. */
+@SuppressWarnings({"rawtypes", "unchecked"})
+public class ExporterBatchWriter {
+  private final Map<ValueType, List<ExportHandler>> handlers = new HashMap<>();
+  private final Map<EntityIdAndEntityType, EntityAndHandler> cachedEntities = new HashMap<>();
```
deepthidevaki, line 26: :question: Why should EntityType be part of the map key? Isn't `EntityId` unique?

deepthidevaki (approved):
:rocket: Thanks for breaking it down into smaller PR.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/store/ExporterBatchWriterTest.java
mustafadagher: agreed, I created the class before I decided to use mockito. will remove it.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java
mustafadagher: That's a good question, I also shared the same concern when looking at the draft PR. 
From what I could see in the implemented handlers there: I assume it should be unique, but it's done differently for each of the implemented handler and I wasn't quite sure of this assumption.

The way @sdorokhova  implemented this key was a `final Tuple<String, Class<?>> cacheKey =  new Tuple<>(id, handler.getEntityType());`. And I just replaced the `Tuple` with a record. I thought maybe she anticipates some cases where this id would be repeated for different record types but the entity type is what separates them.

E.g:  multiple records all has their id's as the `processInstanceKey` but one is a `ProcessInstanceRecord` and the other is a `SequenceFlow` for example. (this is not an accurate example, it's just for elaboration)

deepthidevaki: Thanks. We can leave it as it is now and revisit later if we learn that this is not required. :+1:

PR #21534, opened 2024-08-23
feat: initialize routing state when partition scaling feature is enabled

### LINKED ISSUE #21465: `ClusterConfiguration` contains routing info

lenaschoenburg:
:warning: Depends on https://github.com/camunda/camunda/pull/21529

This adds a new `ClusterConfigurationModifier` to initialize the routing state once on startup if the feature flag is enabled.
If it is enabled, all members will initialize the routing state the same way (assuming that they are all configured with the same partition count, otherwise all bets are off). Because all members will initialize it the same way, there will be no merge conflicts as long as the routing state is not further modified.

closes #21465

### DIFF
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/RoutingStateInitializer.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/RoutingStateInitializer.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.dynamic.config;
+
+import io.camunda.zeebe.dynamic.config.state.ClusterConfiguration;
+import io.camunda.zeebe.dynamic.config.state.RoutingState;
+import io.camunda.zeebe.scheduler.future.ActorFuture;
+import io.camunda.zeebe.scheduler.future.CompletableActorFuture;
+import java.util.Optional;
+
+/**
+ * Initializes the routing state of the cluster configuration if the partition scaling feature is
+ * enabled. All members will initialize the same routing state (as long as their statically
+ * configured partition counts match).
+ */
+public class RoutingStateInitializer implements ClusterConfigurationModifier {
+
+  private final boolean enablePartitionScaling;
+  private final int staticPartitionCount;
+
+  public RoutingStateInitializer(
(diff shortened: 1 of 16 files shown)


zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/RoutingStateInitializer.java
```diff
@@ -0,0 +1,48 @@
...
+  public RoutingStateInitializer(
+      final boolean enablePartitionScaling, final int staticPartitionCount) {
+    this.enablePartitionScaling = enablePartitionScaling;
+    this.staticPartitionCount = staticPartitionCount;
+  }
+
+  @Override
+  public ActorFuture<ClusterConfiguration> modify(final ClusterConfiguration configuration) {
+    if (configuration.routingState().isPresent() || !enablePartitionScaling) {
+      return CompletableActorFuture.completed(configuration);
```
deepthidevaki, line 35: :thought_balloon: Does using `CompletableActorFuture` causes any problems in tests? The other classes uses the provided `ConcurrencyControl` to create the future, so that in tests we can use `TestConcurrencyControl` and `TestActorFuture`. Probably not an issue any more because now CompletableActorFuture can deal with non-actors in most cases.

deepthidevaki (approved):
:rocket: 

:wrench: Now, we only initialize the `RoutingState` in the `RoutingStateInitializer`. But ideally, if it is a new cluster and partition scaling is enabled, we can initialize it in the StaticInitializer itself. `RoutingStateInitializer` should initialize it only when an existing cluster is updated to the newer version that adds this feature. We should be able to eventually remove `RoutingStateInitializer` once the feature is stable and rolled out to alteast one release.  What do you think?

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/RoutingStateInitializer.java
lenaschoenburg: Ah, I was wondering why we were using the executor to create futures in the other initializer :+1:
It hasn't caused any issues yet so I'm inclined to leave it as is.

lenaschoenburg:
> 🔧 Now, we only initialize the RoutingState in the RoutingStateInitializer. But ideally, if it is a new cluster and partition scaling is enabled, we can initialize it in the StaticInitializer itself.

That makes sense, I'll add this :+1:

PR #21539, opened 2024-08-23
Migrate send task instances

### LINKED ISSUE #21538: Migrate send task instances
### Description

A send task should be migrated. Since send task works same as service tasks (only visual representation is different see [docs](https://docs.camunda.io/docs/next/components/modeler/bpmn/send-tasks/)), adding `SEND_TASK` to supported types should be enough to make it available for migration.

berkaycanbc:
## Description

Added send task to supported types. It should be enough to support migrating send tasks because send
tasks works the same as service tasks in the Engine.

## Related issues

closes #21538


koevskinikola (approved):
LGTM @berkaycanbc 👍

PR #21552, opened 2024-08-24
feat: C8 API - Implement Process instance search endpoint

### LINKED ISSUE #21837: [C8 API] Implement Process instance search - controller and service parts
### Description

Controller and service implementation for `/v2/process-instances/search` endpoint. Part of https://github.com/camunda/camunda/issues/20660

- [ ] Potential Security risks have been considered and added to the product-hub issue risk assessment

Optional:

- Handy resources: 
     
- Example projects:

- Versions to validate:
> Docker file : in case that it needed to be tested via docker share the version contain the fixed along with version of other services . 

- Release version ( in which version this fixed/feature will be released)

ghost:
feat: C8 API - Implement Process instance search endpoint

- updated old implementation
- implemented controller and service
- added unit tests
- fixed client compilation errors (client is not implemented yet)

Closes https://github.com/camunda/camunda/issues/21837

## Description

Limitations:
- Filtering by multiple values (IN operation) is not supported, so the fields of type array, `processInstanceKeys`, `processDefinitionKeys`, `excludeKeys`, were not added to the model for `ProcessInstanceQuery`
- The response does not set calculated fields `hasActiveOperation` and `rootInstanceId`
- The response does not set fields of type nested object list, `operations` and `callHierarchy` (required data is either calculated, or stored in other indices).  These fields should probably be returned in their own endpoints, and here return only their ids.
- Client and integration tests are not implemented yet

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/ProcessInstanceFilter.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/ProcessInstanceFilter.java
@@ -16,20 +16,9 @@
 package io.camunda.zeebe.client.api.search.filter;
 
 import io.camunda.zeebe.client.api.search.query.TypedSearchQueryRequest.SearchRequestFilter;
-import java.util.List;
-import java.util.function.Consumer;
 
 public interface ProcessInstanceFilter extends SearchRequestFilter {
 
-  /** Filter by process instance keys. */
-  ProcessInstanceFilter processInstanceKeys(final Long... values);
-
-  /** Filter by process instance keys. */
-  ProcessInstanceFilter processInstanceKeys(final List<Long> values);
-
-  /** Filter by variable values. */
-  ProcessInstanceFilter variable(final VariableValueFilter filter);
-
-  /** Filter by variable values. */
-  ProcessInstanceFilter variable(final Consumer<VariableValueFilter> fn);
+  /** Filter by bpmnProcessId */
+  ProcessInstanceFilter bpmnProcessId(final String bpmnProcessId);
 }
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java
@@ -30,8 +30,8 @@ public class OperationImpl implements Operation {
   public OperationImpl(final OperationItem item) {
     id = item.getId();
     batchOperationId = item.getBatchOperationId();
-    type = item.getType();
-    state = item.getState();
+    type = item.getType().toString();
+    state = item.getState().toString();
     errorMessage = item.getErrorMessage();
     completedDate = item.getCompletedDate();
   }
--- a/service/src/main/java/io/camunda/service/search/filter/ProcessInstanceFilter.java
+++ b/service/src/main/java/io/camunda/service/search/filter/ProcessInstanceFilter.java
@@ -7,8 +7,7 @@
  */
 package io.camunda.service.search.filter;
 
-import static io.camunda.util.CollectionUtil.addValuesToList;
-import static io.camunda.util.CollectionUtil.collectValues;
+import static io.camunda.util.CollectionUtil.*;
 
 import io.camunda.util.ObjectBuilder;
 import java.util.Collections;
@@ -16,8 +15,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.function.Function;
 
-public final record ProcessInstanceFilter(
-    List<Long> processInstanceKeys,
+public record ProcessInstanceFilter(
     boolean running,
     boolean active,
     boolean incidents,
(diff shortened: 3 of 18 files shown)


service/src/main/java/io/camunda/service/search/filter/ProcessInstanceFilter.java
```diff
@@ -7,103 +7,136 @@
...
+      finished = false;
+      return this;
+    }
+
+    public Builder completed(final boolean completed) {
+      this.completed = completed;
       return this;
     }
 
     public Builder completed() {
```
ralfpuchert, line 103: I think this is confusing. The user doesn't know what happens in this case. Maybe it is better to have only one method like the 
```java
public Builder completed(final boolean completed) {
```
Same for all other boolean builder methods.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -147,18 +128,41 @@ private static ProcessInstanceFilter toProcessInstanceFilter(
...
 
     if (filter != null) {
-      final var variableFilters = toVariableValueFilter(filter.getVariables());
-      if (variableFilters != null) {
-        builder.variable(variableFilters);
-      }
-      if (filter.getKey() != null && !filter.getKey().isEmpty()) {
-        builder.processInstanceKeys(filter.getKey());
-      }
+      ofNullable(filter.getRunning()).ifPresentOrElse(builder::running, builder::running);
```
ralfpuchert, line 131: This has to do with the other review comment (FilterBuilder). Can this be expressed in more clearer way?

ralfpuchert (changes requested):
Thanks, see my minor comments on the files.

I think we need unit tests due to the complex logic especially in `ProcessInstanceFilterTransformer` before we merge it into `main`.
WDYT about a "feature" branch where we merge this branch and the "test" branch. The "feature" branch will then merged into `main' if everything is done.

ralfpuchert (approved):

nathansandi (approved):
LGTM (left only a small question/not a blocker)

clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/ProcessInstanceFilter.java
```diff
@@ -16,20 +16,9 @@
...
-  /** Filter by process instance keys. */
-  ProcessInstanceFilter processInstanceKeys(final List<Long> values);
-
-  /** Filter by variable values. */
-  ProcessInstanceFilter variable(final VariableValueFilter filter);
-
-  /** Filter by variable values. */
-  ProcessInstanceFilter variable(final Consumer<VariableValueFilter> fn);
+  /** Filter by bpmnProcessId */
+  ProcessInstanceFilter bpmnProcessId(final String bpmnProcessId);
```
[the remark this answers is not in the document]
mihail-ca: I mentioned in the Description - I did not implement the Client part. These are leftovers from the PoC. I will implement the  Client in a separate commit, to avoid making this commit to large.

clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java
```diff
@@ -30,8 +30,8 @@ public class OperationImpl implements Operation {
   public OperationImpl(final OperationItem item) {
     id = item.getId();
     batchOperationId = item.getBatchOperationId();
-    type = item.getType();
-    state = item.getState();
+    type = item.getType().toString();
+    state = item.getState().toString();
```
mihail-ca, line 34: @ralfpuchert OperationImpl was added by you, however after rebase with main, I got errors below:

```
[ERROR] COMPILATION ERROR :
[ERROR] /clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java:[33,24] incompatible types: io.camunda.zeebe.client.protocol.rest.OperationItem.TypeEnum cannot be converted to java.lang.String
[ERROR] /clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java:[34,26] incompatible types: io.camunda.zeebe.client.protocol.rest.OperationItem.StateEnum cannot be converted to java.lang.String
```

This is because I declared type and state as enums for the Open API model, same as in Operate code:
https://github.com/camunda/camunda/pull/21552/files#diff-a18635d76659a781a6e8d301bd0fed7c77fd6cc5c8cfc57e11c5bbc19c78c9cdR1589

So I had to add conversion using toString(). What is the correct way to handle such enums in client code ? With toString() as I did, or a different way ? Because I could not find any examples, thank you.

mihail-ca: FYI this will be fixed as part of https://github.com/camunda/camunda/issues/21836

PR #21555, opened 2024-08-25
fix: add identity to the list of healthchecks in the initializer

Ben-Sheppard:
## Description

The controller team asked about health checks for the Identity component, after investigating this for them I noticed that Identity is not added to the health check initializer.


romansmirnov (approved):
LGTM.

PR #21571, opened 2024-08-26
Support job update REST API

nicpuppa:
## Description

With this PR `JobUpdateRetries` and `JobUpdateTimeout` commands support REST.
A new command `JobUpdate` is introduced, with this command clients can provide a `JobChangeset` containing the new values for retries and timeout, or can provide those fields separately. This new command support only REST

## Related issues

closes #21523


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/JobRequestValidator.java
```diff
@@ -60,7 +60,7 @@ public static Optional<ProblemDetail> validateJobUpdateRequest(
       final JobUpdateRequest updateRequest) {
     final List<String> violations = new ArrayList<>();
     final JobChangeset changeset = updateRequest.getChangeset();
-    if (changeset.getRetries() == null && changeset.getTimeout() == null) {
+    if (changeset == null || (changeset.getRetries() == null && changeset.getTimeout() == null)) {
```
koevskinikola, line 63: 🔧 This validation seems wrong to me. It will add an error message if even one of the parameters (retries and timeout) are `null`. I think we only need to trigger it when both are missing.

koevskinikola (changes requested):
The code looks good @nicpuppa 👍 

I just have some questions to clarify and some minor suggestions for (maybe) better UX.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/JobRequestValidator.java
nicpuppa: We add the violations if the changeset is null OR if both fields are null. There are also tests that verify this

clients/java/src/main/java/io/camunda/zeebe/client/api/command/UpdateJobCommandStep1.java
```diff
@@ -0,0 +1,58 @@
...
+   *
+   * <p>Timeout value passed as a duration is used to calculate a new job deadline. This will happen
+   * when the command is processed. The timeout value will be added to the current time then.
+   *
+   * @param retries retries the retries of this job, if null it will be not updated
+   * @param timeout the time the timeout of this job, if null it will be not updated
+   * @return the builder for this command. Call {@link #send()} to complete the command and send it
+   *     to the broker.
+   */
+  UpdateJobCommandStep2 update(Integer retries, Long timeout);
```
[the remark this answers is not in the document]
nicpuppa: This is a nice suggestion, I think I prefer to introduce the single command steps instead of allowing user to provide null values. Thanks 💪

nicpuppa:
Hey @koevskinikola I addressed all the comments, please take another look 👀 🙏

koevskinikola (approved):
Nice job @nicpuppa 🎉

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/JobRequestValidator.java
koevskinikola: Yup, I saw them in the IT tests. I had my logic in reverse, good point.

Issue #16181, opened 2024-02-01
Log an easy to understand message when Zeebe broker is starting from scratch because the disk is empty


lenaschoenburg:
We've had several cases where users failed to recognize that restoring from backup was not successful. For everyone involved, from user, over support agent to Zeebe engineer, it would have been helpful if Zeebe would log an easy to understand message during startup if the data directory is empty and we are starting from scratch.

Related to https://app.incident.io/camunda/incidents/914

npepinpe:
ZDP: let's do it next week as the assumption is it would cut down a lot on diagnosis time.

Zelldon:
@npepinpe i will not come to it, removing my assignment.

rodrigo-lourenco-lopes:
Closed with https://github.com/camunda/camunda/pull/21589

PR #21589, opened 2024-08-26
Check if partitions root directory remains empty after restore.

rodrigo-lourenco-lopes:
## Description

This PR adds a warning if data directory remains empty after restore.

The request for this warning came from a support case where the customer failed to recognize that restoring from backup was not successful. This was due to a mistake in our internal documentation which we shared with the customer. As a result, their restore wasn't effective because it didn't restore data to the Zeebe disks but to temporary disks.

## Checklist

- [ ] for CI changes:

## Related issues

closes #https://github.com/camunda/camunda/issues/16181

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/partitioning/startup/RaftPartitionFactory.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/partitioning/startup/RaftPartitionFactory.java
@@ -24,10 +24,12 @@ import java.io.UncheckedIOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class RaftPartitionFactory {
   public static final String GROUP_NAME = "raft-partition";
-
+  private static final Logger LOG = LoggerFactory.getLogger(RaftPartitionFactory.class);
   private final BrokerCfg brokerCfg;
 
   public RaftPartitionFactory(final BrokerCfg brokerCfg) {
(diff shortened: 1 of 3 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/partitioning/startup/RaftPartitionFactory.java
```diff
@@ -24,10 +24,12 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class RaftPartitionFactory {
   public static final String GROUP_NAME = "raft-partition";
-
+  private static final Logger LOG = LoggerFactory.getLogger(RaftPartitionFactory.class);
```
deepthidevaki, line 32: :wrench: May be we can use `io.camunda.zeebe.broker.Loggers.SYSTEM_LOGGER`

zeebe/broker/src/main/java/io/camunda/zeebe/broker/partitioning/startup/RaftPartitionFactory.java
```diff
@@ -41,6 +43,11 @@ public RaftPartition createRaftPartition(final PartitionMetadata partitionMetada
             .resolve("partitions")
             .resolve(partitionMetadata.id().id().toString());
     try {
+      if (FileUtil.isEmpty(partitionDirectory)) {
+        LOG.info(
+            "Root directory for partition {} is empty or does not exist, creating it if does not exist.",
```
deepthidevaki, line 48: Would it be better to make the message more explicit as follows? 
```suggestion
            "Root directory '{}' for partition '{}' is empty or does not exist. The partition `{}` is starting with no pre-existing data.",
```

deepthidevaki (approved):
Thanks :rocket:  Just some minor comments. Please check them before merging.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/partitioning/startup/RaftPartitionFactory.java
rodrigo-lourenco-lopes: Good catch 🙌

PR #21591, opened 2024-08-26
Migrate process instance with an active exclusive gateway

### LINKED ISSUE #21580: Migrate exclusive gateway instances
A process instance may be waiting at an Exclusive Gateway if there is an Incident on it. In that case, it should be possible to migrate the active exclusive gateway over to the target process definition. 

This allows resolving modeling mistakes in the Conditions of the outgoing sequence flows.

korthout:
## Description

This pull request makes it possible to fix mistakes in the Condition expression of sequence flows outgoing from an Exclusive Gateway.

- When the Exclusive Gateway activates, the Conditions of the outgoing Sequence Flows are evaluated. If the evaluation fails, an Incident is raised on the Exclusive Gateway. If the evaluation fails because of a mistake in the Condition expression itself, this could not be resolved previously.

- Users can now fix the mistake in the model, redeploy it, and migrate the process instance to this model. After migrating the process instance, the incident can be resolved.

> [!NOTE]
> Exclusive Gateways are activated and completed atomically (when there are no incidents). So, there is no other case to test or migrate.

> [!NOTE]
> I've taken the liberty to refactor some Exclusive Gateway Processor code 

## Related issues

closes #21580

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateExclusiveGatewayTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateExclusiveGatewayTest.java
@@ -66,10 +66,11 @@ public class MigrateExclusiveGatewayTest {
     final long processInstanceKey =
         ENGINE.processInstance().ofBpmnProcessId(sourceProcessId).create();
 
-    RecordingExporter.incidentRecords(IncidentIntent.CREATED)
-        .withProcessInstanceKey(processInstanceKey)
-        .withElementId("xor")
-        .await();
+    final var incident =
+        RecordingExporter.incidentRecords(IncidentIntent.CREATED)
+            .withProcessInstanceKey(processInstanceKey)
+            .withElementId("xor")
+            .getFirst();
 
     // when
     ENGINE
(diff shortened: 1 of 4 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateExclusiveGatewayTest.java
```diff
@@ -0,0 +1,95 @@
...
+    // when
+    ENGINE
+        .processInstance()
+        .withInstanceKey(processInstanceKey)
+        .migration()
+        .withTargetProcessDefinitionKey(targetProcessDefinitionKey)
+        .addMappingInstruction("xor", "xor2")
+        .migrate();
+
+    // then
```
berkaycanbc, line 83: ❌ We should await migration completion to avoid flakiness.

berkaycanbc (changes requested):
@korthout thanks for the improvements in exclusive gateway implementation. 🚀 

I added one suggestion only. 👀

berkaycanbc (approved):
@korthout Pre-approving it to save time once you tackle my comment.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateExclusiveGatewayTest.java
korthout: 💭 The `.migrate()` actually awaits the successful migration already, and the Incident is migrated as part of it. However, there's something to say for being explicit in tests, so we can add it. It's not necessary to avoid flakiness though.

berkaycanbc: Ahh, we wait for successful expectation completion, right. Okay. 👍

korthout:
Rebased on `main` to resolve a conflict with another PR that added an enum value to the supported elements (e36abaa)

PR #21594, opened 2024-08-26
feat: add user create request to java client

### LINKED ISSUE #21561: Add User Create request to Java Client
### Description

- [x] Support User Create commands in the Java client, [following the guidelines](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md#camunda-client-extension)

### LINKED ISSUE #21695: Add User Create integration tests
<h2>Description</h2>

- [x] Add integration tests verifying User Creation works as expected (refer to the [CorrelateMessageTest](https://github.com/camunda/camunda/blob/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/CorrelateMessageTest.java) for an example)
    - [x] Verify a successful create scenario
    - [x] Verify an invalid request scenario

maryarm:
### Description

- [x] Support User Create commands in the Java client, [following the guidelines](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md#camunda-client-extension)

## Related issues

closes #21561
closes #21695


clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateUserCommandImpl.java
```diff
@@ -0,0 +1,88 @@
...
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<CreateUserResponse> send() {
+    ArgumentUtil.ensureNotNull("username", request.getUsername());
+    ArgumentUtil.ensureNotNull("email", request.getEmail());
+    ArgumentUtil.ensureNotNull("name", request.getName());
+    ArgumentUtil.ensureNotNull("password", request.getPassword());
+    final HttpZeebeFuture<CreateUserResponse> result = new HttpZeebeFuture<>();
```
Ben-Sheppard, line 56: ❓ in the UserController for the `/users` endpoint we return with `executeServiceMethodWithNoContentResult`, will there be a user response here to map?

Ben-Sheppard (commented):
It certainly looks aligned with other Zeebe client implementations (maybe @remcowesterhoud is best to review fully) though I have just a couple of questions :)

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateUserCommandImpl.java
maryarm: It had also confused me but I checked the user-tasks endpoint and client and it seems like it's the way to go. And as I understand the fact that response is empty reflected in CreateUserResponse being an empty interface.

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateUserCommandImpl.java
```diff
@@ -0,0 +1,80 @@
...
+  private final UserWithPasswordRequest request;
+  private final JsonMapper jsonMapper;
+  private final HttpClient httpClient;
+  private final RequestConfig.Builder httpRequestConfig;
+
+  public CreateUserCommandImpl(final HttpClient httpClient, final JsonMapper jsonMapper) {
+    this.jsonMapper = jsonMapper;
+    this.httpClient = httpClient;
+    httpRequestConfig = httpClient.newRequestConfig();
+    request = new UserWithPasswordRequest();
```
remcowesterhoud, line 41: 💭 I'm not thinkin about the `UserWithPasswordRequest` name. Will there be a user request without password in it? I feel like a better name would be `UserCreationReqeust` as that seems more in line with the naming we have for other requests. WDYT?

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateUserCommandImpl.java
remcowesterhoud: Correct, it works because it's an empty response. Otherwise there'd have to be a response mapping in the controller

clients/java/src/test/java/io/camunda/zeebe/client/user/CreateUserTest.java
```diff
@@ -0,0 +1,75 @@
...
+        .name("name")
+        .email("email@example.com")
+        .password("password")
+        .send()
+        .join();
+
+    // then
+    final UserWithPasswordRequest request =
+        gatewayService.getLastRequest(UserWithPasswordRequest.class);
+    assertThat(request.getUsername()).isEqualTo("username");
```
remcowesterhoud, line 44: ❌ Let's make sure we verify all the fields are set in the request properly, instead of just the username.

clients/java/src/test/java/io/camunda/zeebe/client/user/CreateUserTest.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Test
+  void shouldRaiseExceptionOnNullUsername() {
```
remcowesterhoud, line 48: ❌ The other fields are also required. Let's add test scenarios that verify these must be set as well, instead of just for the username.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/util/ZeebeAssertHelper.java
```diff
@@ -330,4 +331,24 @@ public static void assertVariableDocumentUpdated(
...
+      final String username, final Consumer<UserRecordValue> consumer) {
+    final UserRecordValue user =
+        RecordingExporter.userRecords()
+            .filter(
+                userRecordValueRecord ->
+                    userRecordValueRecord.getValue().getUsername().equals(username))
+            .limit(1)
+            .map(Record::getValue)
+            .toList()
+            .getLast();
```
remcowesterhoud, line 349: ❌ This filter should be on the `UserRecordStream`. That way it's easier to use and reuse. You can just say `.withUsername(username)` here.

I noticed on the `UserRecordStream` we already have this filter, it's just named `withUsernameKey`. Let's rename that to `withUsername`. The key part is not relevant and would only cause confusion as it's not a key.


You also don't filter for intent. This means that you are searching for all user record, not just the created ones. It works because you pick the last record, which is the created record in this case. But it's a bit fault-sensitive when we make changes in the future.


In the end I'd expect it to look like this:

```suggestion
    final var user =
        RecordingExporter.userRecords()
            .withIntent(UserIntent.CREATED)
            .withUsername(username)
            .getFirst()
            .getValue();
```

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/CreateUserTest.java
```diff
@@ -0,0 +1,105 @@
...
+                    client
+                        .newUserCreateCommand()
+                        .username("username")
+                        .name("name")
+                        .email("email@example.com")
+                        .password("password")
+                        .send()
+                        .join())
+            .hasCauseInstanceOf(ProblemException.class)
+            .hasMessageContaining("Failed with code 400: 'Bad Request'")
```
remcowesterhoud, line 86: 💭 IMO this should be a `409 Conflict` instead of a `400 Bad Request`. I've [raised this with the rest api team](https://camunda.slack.com/archives/C06UKS51QV9/p1724920818935409). Nothing you need to do about this, just wanted to share.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/CreateUserTest.java
```diff
@@ -0,0 +1,105 @@
...
+        .newUserCreateCommand()
+        .username("username")
+        .name("name")
+        .email("email@example.com")
+        .password("password")
+        .send()
+        .join();
+
+    // when / then
+    final var problem =
```
remcowesterhoud, line 74: 🙃 We can get rid of this variable, it's not used.

remcowesterhoud (changes requested):
Nice work @maryarm 🚀 I only have a few comments, mostly related to tests. Please have a look at them and let me know if you have any questions!

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateUserCommandImpl.java
maryarm: Naming, the most famous problem with devs :D I agree to change to be more consistent but it's not only UserCreationRequest, we were using the same structure for updating a user.

remcowesterhoud: That's fair. We can call it a UserRequest?

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/CreateUserTest.java
remcowesterhoud: @maryarm FYI a PR to change the status code of ALREADY_EXISTS is getting merged to main. So you'll have to modify this assertion to fix the CI on this one before you can merge the PR

remcowesterhoud (approved):
LGTM 🚀

remcowesterhoud:
@maryarm fyi, there's compilation errors. You'll have to rebase with main and fix them before the merge will succeed.

PR #21595, opened 2024-08-26
feat: use dynamic routing info for message request dispatch

### LINKED ISSUE #21468: Use routing info to dispatch message requests
If `ClusterConfiguration` contains routing information, the gateway should use that information instead of `SubscriptionUtil` to determine where to send new messages.
If the routing information is not present because the feature flag is disabled, the gateway should fall back to static configuration and `SubscriptionUtil`.

lenaschoenburg:
:warning:  Depends on https://github.com/camunda/camunda/pull/21590

Updates the `PublishMessageDispatchStrategy` to prefer the message correlation strategy of the current routing state. This ensures that after partition scaling, messages will be routed to the old partitions until we support message relocation.

closes #21468

### DIFF
--- a/zeebe/gateway/pom.xml
+++ b/zeebe/gateway/pom.xml
@@ -110,6 +110,18 @@
     </dependency>
 
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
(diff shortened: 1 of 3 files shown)


deepthidevaki (approved):
:rocket:

PR #21596, opened 2024-08-26
docs: Incorporate OpenAPI spec changes from docs reviews

pepopowitz:
## Description

Incorporates changes into the OpenAPI spec which were proposed in https://github.com/camunda/camunda-docs/pull/4166.

This should prevent us in the future from introducing regressions to the docs for these descriptions.

Collaboration with @christinaausley.

## Related issues

Follow-up from https://github.com/camunda/camunda-docs/pull/4166


remcowesterhoud (approved):
@pepopowitz Thanks for fixing these things 🙇

PR #21613, opened 2024-08-27
ci: skip optimize from platform release

OmranAbazid:
## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/.github/workflows/camunda-platform-release.yml
+++ b/.github/workflows/camunda-platform-release.yml
@@ -155,6 +155,7 @@ jobs:
             -DpushChanges=${PUSH_CHANGES} \
             -DremoteTagging=${PUSH_CHANGES} \
             -DcompletionGoals="spotless:apply" \
+            -D skipOptimize \
             -P-autoFormat \
             -DpreparationGoals="" \
             -Darguments=''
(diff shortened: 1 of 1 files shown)


OmranAbazid:
Not sure if we should also add it on the `mvnw -B versions:set-property` in the workflow file. 
I did not do that but let me know if this is necessary.

nicpuppa (approved):

buccarel (approved):

.github/workflows/camunda-platform-release.yml
```diff
@@ -155,6 +155,7 @@ jobs:
             -DpushChanges=${PUSH_CHANGES} \
             -DremoteTagging=${PUSH_CHANGES} \
             -DcompletionGoals="spotless:apply" \
+            -D skipOptimize \
```
buccarel, line 158: I might be wrong, but I think the recommended syntax is without space, i.e.,
```
-DskipOptimize
```
I think in this case it's equivalent though

nicpuppa: AFAIK it can be a problem on windows OS. Should be good in our case

PR #21616, opened 2024-08-27
build: replace protolock with buf CLI

### LINKED ISSUE #18821: CI build failed on protocol reading due to lock
## Description

[Merge build failed](https://github.com/camunda/camunda/actions/runs/9226120982/job/25385192985) with obscure error:

```
  Failed to execute goal com.salesforce.servicelibs:proto-backwards-compatibility:1.0.7:backwards-compatibility-check (default) on project zeebe-gateway-protocol: An error occurred while running protolock: Cannot run program "/home/runner/work/camunda/camunda/zeebe/gateway-protocol/target/protolock-bin/protolock" (in directory "/home/runner/work/camunda/camunda/zeebe/gateway-protocol"): error=26, Text file busy -> [Help 1]
Error:  
```

Looks like there were multiple builds referencing the same file?

npepinpe:
## Description

This PR replaces the outdated, unmaintained protolock tool (and Maven plugin) for [Buf CLI](https://buf.build/docs/ecosystem/cli-overview).

The main reason behind this is that we were having some issues with the old tool, and as it's unmaintained, it's likely these will be solved. Furthermore, the original maintainers suggest migrating to Buf instead of using their tool.

Since this is a new tool, the approach is a little different: we compare the protocol files directly between the PR/merge group and the base branch HEAD.

We will check explicitly two modules: `zeebe/dynamic-config`, and `zeebe/gateway-protocol`. The reason I set it up like this is because if we check the whole repository, then it also picks up things like proto files in the vendored modules (e.g. Go client), unfortunately. It seems (or I couldn't figure out) that you can exclude paths, but not patterns - you'd have to exclude specific paths. Locally, for example, this will also fail due to the `target` folders which also contain some of the .proto files.

I couldn't figure out how to make it work with a single `buf.yaml` file with multiple modules, primarily because it fails to resolve relative imports.

Each module is configured differently:
- The `zeebe/dynamic-config` module is configured to only check for WIRE breaking changes (i.e. binary incompatibility), not source changes.
- The `zeebe/gateway-protocol` module is configured to check for both binary and source code breaking changes.

In order to allow breaking changes, you must add BREAKING CHANGE: in the pull request description. This will cause the check to be skipped on pull_request, merge_group, and push triggers.

This is now documented in the Wiki: https://github.com/camunda/camunda/wiki/Breaking-Changes#protobuf

## Related issues

closes #18821


npepinpe:
I'm not super keen on merging binary files though. That would mean the reviewer needs to double check the file locally if it gets modified. An alternative is instead copying the proto files to a hidden directoy `.proto.lock`, which would contain the original `buf.yaml` and our proto files. This way it might be clearer to see the changes?

npepinpe:
Ah, it looks like it's not fetching the base reference, so I'd need to do that first.

npepinpe:
Here's an example of a breaking source change: https://github.com/camunda/camunda/actions/runs/10594107376/job/29357095400
And here's passing when I set the breaking level to WIRE (i.e. only break if binary format is incompatible, ignore source code changes): https://github.com/camunda/camunda/actions/runs/10594192091/job/29357380276?pr=21616
And here's an example failure if there's a WIRE breaking change (i.e. binary incompatibility): https://github.com/camunda/camunda/actions/runs/10594239344/job/29357650307?pr=21616

npepinpe:
One small weird behavior is the comment by the action is edited by the last job to run. So one run may fail, but the other may set it to green. A little weird, but in the end you can still see the failed check :shrug:

npepinpe:
@lenaschoenburg Happy to block some time and see if we can't figure out a better way to bypass breaking changes than having the last commit of the PR have the specific message =/

.github/workflows/ci.yml
```diff
@@ -7,7 +7,13 @@ on:
...
       - stable/*
       - release-*
-  pull_request: {}
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
```
lenaschoenburg, line 16: :question: What's the purpose behind that change?

.github/workflows/zeebe-ci.yml
```diff
@@ -282,6 +282,33 @@ jobs:
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+  protobuf-checks:
+    name: "Protobuf Backwards Compatibility"
+    timeout-minutes: 25
+    runs-on: ubuntu-latest
+    if: |
+      contains(fromJSON('["pull_request","push"]'), github.event_name) &&
+      !contains(github.event.pull_request.labels.*.name, 'breaking/protobuf')
```
lenaschoenburg, line 291: :question: This last expression for the PR label will just treat `push` events as not labeled with `breaking/protobuf`, right? So the check runs for push events.

lenaschoenburg (approved):
Just a couple of questions, nothing blocking the merge. Nice work :rocket:

lenaschoenburg:
❌ this should be added to the test-summary job do that it actually fails the build.

🔧 perhaps we should add it to the common ci (including change detection for .proto and buf.yaml fiöes) instead?

npepinpe:
Good point :+1: Now shoo, go on your FTO!

.github/workflows/ci.yml
npepinpe: The default types for PR are `opened`, `synchronize`, and `reopened`. So I added these to keep the behavior as before. The idea is we now need `labeled` and `unlabeled`, which are not triggers, if we want to also trigger CI to run again when we add/remove the bypass label.

The downside is that now every time you label your PR, the CI will run...I will look with Christian how we can find a better middle ground.

.github/workflows/zeebe-ci.yml
npepinpe: Yes, it will evaluate to false in the case of push, then inverted to true :) So for push it should just work. Of course, this is assumption as I couldn't test yet with push.

npepinpe:
Moved to a common CI check, and only if we have `**/*.proto` or `buf.yaml` file changes.

.github/workflows/ci.yml
```diff
@@ -500,6 +507,36 @@ jobs:
...
 
+  protobuf-checks:
+    name: "Protobuf Backwards Compatibility"
+    timeout-minutes: 10
+    needs: [detect-changes]
+    runs-on: ubuntu-latest
+    if: |
+      needs.detect-changes.outputs.protobuf-changes == 'true' &&
+      contains(fromJSON('["pull_request","push"]'), github.event_name) &&
+      !contains(github.event.pull_request.labels.*.name, 'breaking/protobuf')
```
npepinpe, line 518: @cmur2 - the assumption here is that this would evaluate to true when the event is not a pull_request, basically. Let me know if you think that's wrong :+1:

npepinpe:
Successful run with some backwards compatible proto changes: https://github.com/camunda/camunda/actions/runs/10611525030/job/29411255025?pr=21616
Failing run with some non backwards compatible proto changes: https://github.com/camunda/camunda/actions/runs/10611579232/job/29411657190

npepinpe:
Failing check on breaking changes: https://github.com/camunda/camunda/actions/runs/10612870987/job/29415581718?pr=21616

I added the `BREAKING CHANGE:` in the PR body, but as we can see it's not automatically re-triggered. But! Re-running the job should fetch the latest, and then skip.

npepinpe:
So that didn't work :smile:

npepinpe:
I'm realizing we may want to allow something more unique than `BREAKING CHANGE:`, but I guess we can iterate on that.

.github/workflows/ci.yml
npepinpe: In the end, we will go with `BREAKING CHANGE:` must be in the last commit (for merge_group and push), and in the PR body for a PR. The PR body is then the message of the merge commit, so it will be the last commit when merging/pushing.

npepinpe:
@cmur2 - as discussed, just review the CI triggers, no need to check the actual protobuf stuff.

.github/workflows/ci.yml
```diff
@@ -500,6 +501,49 @@ jobs:
...
+          lint: false
+          push: false
+          archive: false
+          breaking: true
+      - name: Observe build status
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          build_status: ${{ contains(steps.*.conclusion, 'failure') && 'failure' || 'success' }}
```
cmur2, line 542: Please use `build_status: ${{ job.status }}` (see [wiki](https://github.com/camunda/camunda/wiki/CI-&-Automation#metrics-collection) or how the other jobs in this file do it)

cmur2 (commented):
Please change the `build_status` part that I remarked on, afterwards free to merge from my POV without additional review 👍🏼

.github/workflows/ci.yml
npepinpe: I copy pasted this from elsewhere, I guess we have outdated definitions :smile:

cmur2: Yea Chris removed those from the Zeebe CI 2-3 days ago :)

npepinpe:
Backport will fail for 8.4, but I'll edit it and then make that backport to 8.3. As with the other buf PRs, we'll skip 8.2 since it's going out of support at the end of September, so no worth the effort.

Issue #20661, opened 2024-07-31
[C8 API] Implement `Incident` endpoints


ralfpuchert:
### Description

Incident endpoints provides searching for incidents.

[Issue for statistic endpoints](https://github.com/camunda/camunda/issues/21721)

* The endpoints should be implemented as described in [REST API Developer Guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)
* Incident objects are defined in
  * [Incident](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.62bj2m285hxd) 
 
### Endpoints

- [x] Search incident - `/v2/incidents/search` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.8xthsdktfnb2)
- [x] Get incident - `GET /v2/incidents/{incidentKey}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.d152kb3ezmjd)

### Acceptance Criteria

- [x] Unit tests
- [x] Integration tests

### Definition of Ready - Checklist

- [ ] The issue has a meaningful title, description, and testable acceptance criteria
- [ ] If documentation needs to be updated, an issue is created in the [camunda-docs](https://github.com/camunda/camunda-docs) repo, and the issue is added to our Operate project board.
- [ ] If HELM charts need to be updated, an issue is created in the [camunda-platform-heml](https://github.com/camunda/camunda-platform-helm) repo, and the issue is added to our Operate project board.
- [ ] Potential Security risks have been considered and added to the product-hub issue risk assessment
- [ ] Cross-team dependencies have been considered

Optional:
- [ ] Design input has been collected by the assignee
### :point_right: Handover Dev to QA 

- Handy resources: 
     BPMN/DMN models, plugins, scripts, REST API endpoints + example payload, etc : 
     
- Example projects:

- Versions to validate:
> Docker file : in case that it needed to be tested via docker share the version contain the fixed along with version of other services . 

- Release version ( in which version this fixed/feature will be released)

tmetzke:
Added `Get incident` endpoint

Issue #20657, opened 2024-07-31
[C8 API] Implement `Flownode instance` endpoints


ralfpuchert:
### Description

Flow node instance endpoints provides searching for flow node instances and metadata.

[Issue for statistic endpoint](https://github.com/camunda/camunda/issues/21719)

* The endpoints should be implemented as described in [REST API Developer Guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)
* Flow node instance objects are defined in
  * [Flow node instance](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.wcm9xzmzyzor) 
 
### Endpoints

- [x] Search Flow node instances - `/v2/flownode-instances/search` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.gkjsn89wc5pe)
- [x] Get flow node instance - `GET /v2/flownode-instances/{flowNodeInstanceKey}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.ku4woahmdfen)
   https://github.com/camunda/camunda/pull/22453
~~- [ ] Search Flow node metadata  - `/v2/flownode-instances/metadata` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.q5ljdfnepda4)~~ 8.8

### Acceptance Criteria

- [ ] Unit tests
- [ ] Integration tests

### Definition of Ready - Checklist

- [ ] The issue has a meaningful title, description, and testable acceptance criteria
- [ ] If documentation needs to be updated, an issue is created in the [camunda-docs](https://github.com/camunda/camunda-docs) repo, and the issue is added to our Operate project board.
- [ ] If HELM charts need to be updated, an issue is created in the [camunda-platform-heml](https://github.com/camunda/camunda-platform-helm) repo, and the issue is added to our Operate project board.
- [ ] Potential Security risks have been considered and added to the product-hub issue risk assessment
- [ ] Cross-team dependencies have been considered

Optional:
- [ ] Design input has been collected by the assignee
### :point_right: Handover Dev to QA 

- Handy resources: 
[...]

PR #21624, opened 2024-08-27
feat: add implementation for flownode instance entity, filter, sort, transformer

ralfpuchert:
## Description

* Add implementation for entity, filter, sort, transformer
* Test PR will follow
* Is rebased on [operate/c8-flownode-instances-endpoint](https://github.com/camunda/camunda/tree/operate/c8-flownode-instances-endpoint) 

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/service/src/main/java/io/camunda/service/entities/FlownodeInstanceEntity.java
+++ b/service/src/main/java/io/camunda/service/entities/FlownodeInstanceEntity.java
@@ -7,4 +7,20 @@
  */
 package io.camunda.service.entities;
 
-public record FlownodeInstanceEntity() {}
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record FlownodeInstanceEntity(
+    Long key,
+    Long processInstanceKey,
+    Long processDefinitionKey,
+    String startDate,
+    String endDate,
+    String flowNodeId,
+    String flowNodeName,
+    String treePath,
+    String type,
+    String state,
+    Boolean incident,
+    Long incidentKey,
+    String tenantId) {}
--- a/service/src/main/java/io/camunda/service/search/filter/FlownodeInstanceFilter.java
+++ b/service/src/main/java/io/camunda/service/search/filter/FlownodeInstanceFilter.java
@@ -7,15 +7,151 @@
  */
 package io.camunda.service.search.filter;
 
-import io.camunda.util.ObjectBuilder;
+import static io.camunda.util.CollectionUtil.addValuesToList;
+import static io.camunda.util.CollectionUtil.collectValuesAsList;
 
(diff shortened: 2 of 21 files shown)


service/src/main/java/io/camunda/service/entities/FlownodeInstanceEntity.java
```diff
@@ -7,4 +7,20 @@
  */
 package io.camunda.service.entities;
 
-public record FlownodeInstanceEntity() {}
```
houssain-barouni, line 10: The term “FlowNode” is a compound word, and using proper camel case would improve readability.
I suggest using`FlowNodeInstanceEntity` which is consistent with existing naming in Operate
Same in the other classes

service/src/main/java/io/camunda/service/search/filter/FlownodeInstanceFilter.java
```diff
@@ -7,15 +7,151 @@
...
+import static io.camunda.util.CollectionUtil.collectValuesAsList;
+
 import io.camunda.util.ObjectBuilder;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
 
-public record FlownodeInstanceFilter() implements FilterBase {
+public record FlownodeInstanceFilter(
+    List<Long> keys,
```
houssain-barouni, line 19: I wonder if we should use `flowNodeKeys` to be consistent with the other filters
We have used `decisionRequirementsKeys` for instance to be consistent with Zeebe API namings, but it seems that FlowNodeInstance is not part of the Zeebe API

houssain-barouni (commented):
Looks good 👍 , I have 2 comments about namings

service/src/main/java/io/camunda/service/entities/FlownodeInstanceEntity.java
nathansandi, line 10: I agree with @houssain-barouni ove renaming this class to `FlowNodeInstanceEntity`.

nathansandi (commented):
Great work here,
I would just check the comments by @houssain-barouni

Also related to the key, in order to check if this entity also already exists over zeebe records to follow the same naming standard

ghost (approved):

mihail-ca:
@ralfpuchert could you please make sure that **all** PRs are linked to the [corresponding issue](https://github.com/camunda/camunda/issues/20657), for easier tracking ? Same for [incidents endpoint issue](https://github.com/camunda/camunda/issues/20661), thank you
image

service/src/main/java/io/camunda/service/search/filter/FlownodeInstanceFilter.java
ralfpuchert: Good point, it would then `flowNodeInstanceKeys` etc.

PR #21626, opened 2024-08-27
Define a cluster change operation for adding a new partition

### LINKED ISSUE #21478: New cluster change operation to bootstrap a new partition
Define a new `ClusterConfigurationChangeOperation` that bootstrap a _new_ partition.
Defining an applier for this operation is out of scope, this will be added later.

## Acceptance criteria

- [ ] Define a record carrying necessary information
- [ ] Define a protobuf message for serialization
- [ ] Extend the `ProtoBufSerializer` to serialize the operation
- [ ] Extend the `ProtoBufSerializer` to deserialize the operation

deepthidevaki:
## Description

This PR adds a new `ClusterConfigurationChangeOperation` to bootstrap a new partition. On bootstrap, the partition will be started with only a single replica. More replicas should be added in subsequent operation using the existing `PartitionJoinOperation`. 

## Related issues

closes #21478


deepthidevaki:
@lenaschoenburg I didn't extend `ProtobufSerializerTest`, eventhough I suggested it in you PR :) . I was thinking to remove it because the randomized test already covers it. What do you think?

lenaschoenburg (approved):
Nice :rocket: 

I think removing the explicit `ProtobufSerializerTest` when we already have the property tests is fine :+1: But the one test in there is about _updating_ the persisted file. Perhaps we could add another similar one that just asserts that any `ClusterConfiguration` can be serialized and deserialized back to an equal one again?

deepthidevaki:
> Nice 🚀
> 
> I think removing the explicit `ProtobufSerializerTest` when we already have the property tests is fine 👍 But the one test in there is about _updating_ the persisted file. Perhaps we could add another similar one that just asserts that any `ClusterConfiguration` can be serialized and deserialized back to an equal one again?

:+1: Will do in a followup PR.

PR #21640, opened 2024-08-27
fix: scan for the identity index controller to serve the Identity UI

Ben-Sheppard:
## Description

I noticed that when starting the single image with the profile `identity` there is no UI for Identity as it returns a 404, further investigation of this lead me to find that the `IdentityIndexController` bean is not being picked up during scans and therefore there is no URLs to catch the webapp redirect.

This PR adds in a config to scan the controller and therefore serve the Identity UI.


romansmirnov (approved):
LGTM

PR #21660, opened 2024-08-28
Add property tests for protobuf serializer

deepthidevaki:
## Description

This PR adds a property test for ProtobufSerializer for `ClusterConfiguration`. It reuses the domain defined for the existing property test for PersistedClusterConfiguration.

The property tests covers all possible values for a `ClusterConfiguration`. So there is no need to explicitly add a new test when ever new fields are added to `ClusterConfiguration`.


lenaschoenburg (approved):
Thanks! :rocket:

PR #21670, opened 2024-08-28
Migrate business rule task instances

### LINKED ISSUE #21588: Migrate business rule task instances
## Description

Migrating a business rule task should be possible. There can be two different implementations of a business rule task:
- A called decision
- Job worker implementation

### Called Decision

In normal circumstances the called decision is evaluated immediately after the business rule task is activated. Still, a process instance may be waiting on a business rule task if there is an incident on it.

**Note:** There is an edge case where the migration command can come after the activation of the business rule task. When the `ACTIVATE_ELEMENT` command for the business rule task is the last command in a command batch (we have a limit for number of commands in a batch), `MIGRATE` command can come right after the activation but before completion of the business rule task as follows:

- ACTIVATE_ELEMENT business rule task
- MIGRATE process instance
- COMPLETE_ELEMENT business rule task

But the case mentioned above is not easy to test because the commands for business rule tasks should be written manually and that is not a trivial operation to do with our current test framework.

### Job Worker Implementation
The second case is already handled here:

https://github.com/camunda/camunda/blob/3ee2a8cf7b1490eb58570727f7dd289630b11a3c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java#L255-L273

### Test Cases

Following test cases should be added to verify possible use cases:

- Migrating decision based business rule task with an incident
- Migrating job worker based business rule task

berkaycanbc:
## Description

Migrating a business rule task should be possible. There can be two different implementations of a business rule task:
- A called decision
- Job worker implementation

### Called Decision

In normal circumstances the called decision is evaluated immediately after the business rule task is activated. Still, a process instance may be waiting on a business rule task if there is an incident on it.

**Note:** There is an edge case where the migration command can come after the activation of the business rule task. When the `ACTIVATE_ELEMENT` command for the business rule task is the last command in a command batch (we have a limit for number of commands in a batch), `MIGRATE` command can come right after the activation but before completion of the business rule task as follows:

- ACTIVATE_ELEMENT business rule task
- MIGRATE process instance
- COMPLETE_ELEMENT business rule task

But the case mentioned above is not easy to test because the commands for business rule tasks should be written manually and that is not a trivial operation to do with our current test framework.

### Job Worker Implementation
The second case is already handled here:

https://github.com/camunda/camunda/blob/3ee2a8cf7b1490eb58570727f7dd289630b11a3c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java#L255-L273

### Test Cases

Following test cases should be added to verify possible use cases:

- Migrating decision based business rule task with an incident
- Migrating job worker based business rule task

## Related issues

closes #21588

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateBusinessRuleTaskTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateBusinessRuleTaskTest.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.processinstance.migration;
+
+import static io.camunda.zeebe.engine.processing.processinstance.migration.MigrationTestUtil.extractProcessDefinitionKeyByProcessId;
+import static io.camunda.zeebe.protocol.record.Assertions.assertThat;
+
+import io.camunda.zeebe.engine.util.EngineRule;
+import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.protocol.record.Assertions;
+import io.camunda.zeebe.protocol.record.intent.DecisionEvaluationIntent;
+import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
+import io.camunda.zeebe.protocol.record.intent.JobIntent;
+import io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent;
+import io.camunda.zeebe.protocol.record.intent.ProcessInstanceMigrationIntent;
+import io.camunda.zeebe.protocol.record.value.BpmnElementType;
+import io.camunda.zeebe.test.util.BrokerClassRuleHelper;
+import io.camunda.zeebe.test.util.record.RecordingExporter;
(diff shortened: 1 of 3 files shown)


korthout:
>But the case mentioned above is not easy to test because the commands for business rule tasks should be written manually and that is not a trivial operation to do with our current test framework.

Regarding this edge case, I think it's okay to have no test for it. The only thing that must be migrated is the element instance itself, similar to all other active elements during migration. There will not be any difference apart from the element type. As this case (an element instance is migrated when that element is in the ACTIVATED state) is already covered by so many other element migrations, I'm fine with not having this specific case tested.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateBusinessRuleTaskTest.java
```diff
@@ -0,0 +1,187 @@
...
+        .processInstance()
+        .withInstanceKey(processInstanceKey)
+        .migration()
+        .withTargetProcessDefinitionKey(targetProcessDefinitionKey)
+        .addMappingInstruction("businessRuleTask1", "businessRuleTask2")
+        .migrate();
+
+    RecordingExporter.processInstanceMigrationRecords(ProcessInstanceMigrationIntent.MIGRATED)
+        .withProcessInstanceKey(processInstanceKey)
+        .await();
```
korthout, line 93: 🔧 This is redundant. The call to `.migrate()` (the last line above this) already awaits this exact event. 
🔧 We can of course be more explicit by awaiting the Incident to be migrated, see: d90e8d4986254bbb542ba5415aa153c6b1bdc8d4

korthout (approved):
Great PR 🚀 

👍 I really that you've added the reasoning in the commit messages and a description of the bigger picture in the pull request.

🔧 I have a minor suggestion

👍 LGTM

PR #21685, opened 2024-08-28
Create composite action to cache maven.

### LINKED ISSUE #21423: CI: create composite action to set up Maven cache
## Description

Part of #18750. Create a composite GHA that can be reused from other GHA workflows which sets up Maven (wrapper) to use GHA cache according to [our caching strategy](https://github.com/camunda/camunda/wiki/CI-&-Automation#caching-strategy). That means the action should always try to restore from cache but only save the cache `on: push` to `main` or `stable*` branch builds.

Existing logic (relevant parts) from Zeebe team in https://github.com/camunda/camunda/blob/fd2e86a59785408bd39daa1134aa1d292281c390/.github/actions/setup-zeebe/action.yml#L113-L180 can be reused and moved to a new composite GHA stored in `.github/action/setup-maven-cache`.

rodrigo-lourenco-lopes:
## Description

Create composite action to cache maven. wraps the logic in setup-zeebe action in a composite action.

Usage of this composite action requires as inputs the maven cache key modifier.

```yaml
    - name: Configure Maven
      uses: ./.github/actions/setup-maven-cache
      with:
        maven-cache-key-modifier: ${{ inputs.maven-cache-key-modifier }}
```

To test the viability of the composite it is now used in setup-zeebe action.

## Checklist

- [x] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/21423

### DIFF
--- a/.github/actions/setup-maven-cache/action.yaml
+++ b/.github/actions/setup-maven-cache/action.yaml
@@ -48,7 +48,7 @@ runs:
         EOF
     - name: Cache local Maven repository
       # Only use the full cache action if we're on main or stable/* branches
-      if: ${{ startsWith(github.ref_name, 'stable/') || github.ref_name == 'main' }}
+      if: ${{ !(startsWith(github.ref_name, 'stable/') || github.ref_name == 'main') }}
       uses: actions/cache@v4
       with:
         # This is the path used by the `enhancedLocalRepository` set up in the 'Configure Maven' step.
@@ -57,9 +57,9 @@ runs:
         path: ~/.m2/repository/cached/releases/
         # it matters for caching as absolute paths on self-hosted and GitHub runners differ
         # self-hosted: `/runner/` vs gh-hosted: `/home/runner`
-        key: ${{ runner.environment }}-${{ runner.os }}-mvn-${{ inputs.maven-cache-key-modifier }}-${{ hashFiles('**/pom.xml') }}
+        key: ${{ runner.environment }}-${{ runner.os }}-mvn-1-${{ inputs.maven-cache-key-modifier }}-${{ hashFiles('**/pom.xml') }}
         restore-keys: |
-          ${{ runner.environment }}-${{ runner.os }}-mvn-${{ inputs.maven-cache-key-modifier }}
+          ${{ runner.environment }}-${{ runner.os }}-mvn-1-${{ inputs.maven-cache-key-modifier }}
     - name: Restore maven cache
       # Restore cache (but don't save it) if we're not on main or stable/* branches
       if: ${{ !(startsWith(github.ref_name, 'stable/') || github.ref_name == 'main') }}
@@ -67,6 +67,6 @@ runs:
       with:
         # This has to match the 'Cache local Maven repository' step above
         path: ~/.m2/repository/cached/releases/
-        key: ${{ runner.environment }}-${{ runner.os }}-mvn-${{ inputs.maven-cache-key-modifier }}-${{ hashFiles('**/pom.xml') }}
+        key: ${{ runner.environment }}-${{ runner.os }}-mvn-2-${{ inputs.maven-cache-key-modifier }}-${{ hashFiles('**/pom.xml') }}
         restore-keys: |
-          ${{ runner.environment }}-${{ runner.os }}-mvn-${{ inputs.maven-cache-key-modifier }}
+          ${{ runner.environment }}-${{ runner.os }}-mvn-2-${{ inputs.maven-cache-key-modifier }}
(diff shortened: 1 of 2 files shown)


.github/actions/setup-maven-cache/action.yaml
```diff
@@ -0,0 +1,77 @@
...
+name: Setup Maven Cache
+
+
+description: Configured GHA cache for Maven global cache dir (no save on PRs), see https://github.com/camunda/camunda/wiki/CI-&-Automation#caching-strategy
+
+inputs:
+  java:
+    description: If true, will set up Java; defaults to true
+    required: false
+    default: "true"
```
cmur2, line 11: This input is unneeded and should be removed. All steps in this action are conditional on this input so it is easier to just not call the action if the caller identifies that no Java setup is desired.

.github/actions/setup-zeebe/action.yml
```diff
@@ -111,63 +111,10 @@ runs:
...
-      if: inputs.java == 'true' && !(startsWith(github.ref_name, 'stable/') || github.ref_name == 'main')
-      uses: actions/cache/restore@v4
+      uses: ./.github/actions/setup-maven-cache
       with:
-        # This has to match the 'Cache local Maven repository' step above
-        path: ~/.m2/repository/cached/releases/
-        key: ${{ runner.environment }}-${{ runner.os }}-mvn-${{ inputs.maven-cache-key-modifier }}-${{ hashFiles('**/pom.xml') }}
-        restore-keys: |
-          ${{ runner.environment }}-${{ runner.os }}-mvn-${{ inputs.maven-cache-key-modifier }}
+        java: ${{ inputs.java }}
```
cmur2, line 116: Convert this to an `if: {{ inputs.java == 'true' }}` conditional for the step that calls the setup-maven-cache action.

cmur2:
@rodrigo-lourenco-lopes since you asked how to test this PR:

I looked at the CI runs and can see e.g. in https://github.com/camunda/camunda/actions/runs/10598267866/job/29370652128?pr=21685 that the new composite action is used as expected and restoring a cache from the `main` branch successfully. That means to me that the cache restoring part works fine and needs no further testing.

It should also be tested if a cache can be created successfully (to avoid roundtrip of merging this PR to `main` and then realizing it doesn't work). This can be done by temporarily tweaking/inverting the `if` conditions in the new composite action so that they allow creating a cache on a PR.

rodrigo-lourenco-lopes:
> It should also be tested if a cache can be created successfully (to avoid roundtrip of merging this PR to `main` and then realizing it doesn't work). This can be done by temporarily tweaking/inverting the `if` conditions in the new composite action so that they allow creating a cache on a PR.

Thanks for the tip! 🙌 I will add a commit to test it in this PR while it is in draft, and revert it later once it is ready to merge and if the test is successful.

rodrigo-lourenco-lopes:
It seems that the test was is working by observing [here](https://github.com/camunda/camunda/actions/runs/10614567736/job/29442431525?pr=21685). The cache operation was run creating one with the key `github-hosted-macOS-mvn-shared-b8f0f396e1172a55c103b6acfb8bc8217f2fef46e76fe9156890ec1d0b4e4745
` and restoring from the same key right after successfully.

cmur2:
> It seems that the test was is working by observing [here](https://github.com/camunda/camunda/actions/runs/10614567736/job/29442431525?pr=21685). The cache operation was run creating one with the key `github-hosted-macOS-mvn-shared-b8f0f396e1172a55c103b6acfb8bc8217f2fef46e76fe9156890ec1d0b4e4745 ` and restoring from the same key right after successfully.

I don't think that is what happened. In general in the beginning of a job only restores happen, creating a cache is happening in a "Post Run" step. See [this line](https://github.com/camunda/camunda/actions/runs/10614567736/job/29442431525?pr=21685#step:3:183) which restores the cache and then [this line](https://github.com/camunda/camunda/actions/runs/10614567736/job/29442431525?pr=21685#step:3:216) which restores it again.

Creating/saving the cache would happen in [this line](https://github.com/camunda/camunda/actions/runs/10614567736/job/29442431525?pr=21685#step:15:3) but doesn't since a cache was already found (and could be restored).

To test that effectively you probably have to change the cache key temporarily to something unique so no cache restore is possible -> means that cache creation will be triggered at the end.

cmur2:
> To test that effectively you probably have to change the cache key temporarily to something unique so no cache restore is possible -> means that cache creation will be triggered at the end.

@rodrigo-lourenco-lopes can you try this please?

cmur2:
> [here](https://github.com/camunda/camunda/actions/runs/10666794638/job/29563145662?pr=21685#step:15:5) we can see that the cache was created successfully in the post run step like you suggested.
> 
> The test commit had the condition inverted to allow the creation of the cache on PRs and different cache key names so that we would have a cache miss.

Thank you very much for testing this scenario 👍🏼

cmur2 (approved):
LGTM

PR #21691, opened 2024-08-28
Migrate script task instances

### LINKED ISSUE #21671: Migrate script task instances
## Description

Migrating a Script Task should be possible. There can be two different implementations of a Script Task:
- A FEEL expression
- Job worker implementation

### FEEL expression

The FEEL expression is evaluated immediately after the Script Task is activated. Still, a process instance may be waiting on a Script Task if there is an incident on it.

### Job Worker Implementation
A Script Task can be based on job worker. This case is already handled here:

https://github.com/camunda/camunda/blob/3ee2a8cf7b1490eb58570727f7dd289630b11a3c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java#L255-L273

### Test Cases

Following test cases should be added to verify possible use cases:

- Migrating Script Task with an Incident
- Migrating Job Worker based Script Task

berkaycanbc:
## Description

Migrating a Script Task should be possible. There can be two different implementations of a Script Task:
- A FEEL expression
- Job worker implementation

### FEEL expression

The FEEL expression is evaluated immediately after the Script Task is activated. Still, a process instance may be waiting on a Script Task if there is an incident on it.

### Job Worker Implementation
A Script Task can be based on job worker. This case is already handled here:

https://github.com/camunda/camunda/blob/3ee2a8cf7b1490eb58570727f7dd289630b11a3c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java#L255-L273

### Test Cases

Following test cases should be added to verify possible use cases:

- Migrating Script Task with an Incident
- Migrating Job Worker based Script Task

## Related issues

closes #21671

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateScriptTaskTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateScriptTaskTest.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.processinstance.migration;
+
+import static io.camunda.zeebe.engine.processing.processinstance.migration.MigrationTestUtil.extractProcessDefinitionKeyByProcessId;
+import static io.camunda.zeebe.protocol.record.Assertions.assertThat;
+
+import io.camunda.zeebe.engine.util.EngineRule;
+import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.protocol.record.Assertions;
+import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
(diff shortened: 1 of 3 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateScriptTaskTest.java
```diff
@@ -0,0 +1,168 @@
...
+        .addMappingInstruction("scriptTask1", "scriptTask2")
+        .migrate();
+
+    RecordingExporter.processInstanceMigrationRecords(ProcessInstanceMigrationIntent.MIGRATED)
+        .withProcessInstanceKey(processInstanceKey)
+        .await();
+
+    ENGINE.incident().ofInstance(processInstanceKey).resolve();
+
+    Assertions.assertThat(
```
korthout, line 86: 🔧 To show that the Script Task was successfully we can also verify that the result variable was created:

```java
Assertions.assertThat(
        RecordingExporter.variableRecords(VariableIntent.CREATED)
            .withName("result")
            .getFirst()
            .getValue())
    .describedAs("Expected to successfully evaluate the feel expression")
    .hasValue("true");
```

korthout (approved):
Thanks @berkaycanbc 📜 Another useful migration supported!

🔧 I have a few suggestions, but nothing blocking. Please have a look.

LGTM 👍

Issue #21376, opened 2024-08-20
Zeebe exporting latency can increase significantly without clear root cause


Zelldon:
**Describe the bug**

I have observed several times in my recent benchmarks that the exporting latency, the time from writing a record to start exporting the record, has increased significantly. See new metric [PR](https://github.com/camunda/camunda/pull/21297)

Normally it is expected that this latency is similar to the processing around some milliseconds, but on higher load benchmarks it has been seen that this latency can increase to minutes.

![2024-08-20_14-11](https://github.com/user-attachments/assets/03653b57-f9a5-4797-820a-7fc6da07c93c)

This obviously also impacts the importing latency and import throughput.

![2024-08-20_13-28](https://github.com/user-attachments/assets/390f6307-6199-41c9-9854-7eaf0175f424)

Right now it is not clear whether the impact is due to ES and the Elasticsearch exporter flush duration or something else.

During the benchmark, several metrics have been investigated. For example, no pods have been restarted, the disk usage was around 30% (for ES).

The only interesting was that a ES seem to create a new instance around 11, when the latency was degraded. 

![2024-08-20_14-55](https://github.com/user-attachments/assets/39aaba78-04d6-45a2-882a-361bb8a06312)

Furthermore, in [ES logs I can see that some queries take quite a while. 
](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22zeebe-io%22%0Aresource.labels.location%3D%22europe-west1-b%22%0Aresource.labels.cluster_name%3D%22zeebe-cluster%22%0Aresource.labels.namespace_name%3D%22ck-20-08-high-load-base%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fcomponent%3D%22master%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Finstance%3D%22zeebe-benchmark-test%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fname%3D%22elasticsearch%22;pinnedLogId=2024-08-20T08:59:30.135105502Z%2Favew5v3u11wpuiib;cursorTimestamp=2024-08-20T08:59:33.341274202Z;startTime=2024-08-20T08:56:00.000Z;endTime=2024-08-20T09:08:00.000Z?project=zeebe-io)

![2024-08-20_14-56](https://github.com/user-attachments/assets/61119fe6-f6c6-437e-9841-b160b7f26250)

[...]

PR #21699, opened 2024-08-28
feat: add cpu throttling panel

ChrisKujawa:
## Description

Adds a CPU throttling panel to the Zeebe dashboard to better detect whether pod/container is throttled and can't run/execute its work

### Normal benchmark:

Here we can see that the gateway is for example throttled

### Mix benchmark 

In our mixed benchmarks with lower load all seem to be fine

### DIFF
--- a/monitor/grafana/zeebe.json
+++ b/monitor/grafana/zeebe.json
@@ -6295,7 +6295,7 @@
                 "uid": "${DS_PROMETHEUS}"
               },
               "editorMode": "code",
-              "expr": "sum(rate(container_cpu_cfs_throttled_periods_total{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$__rate_interval])) by (pod) / sum(rate(container_cpu_cfs_periods_total{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$__rate_interval])) by (pod)",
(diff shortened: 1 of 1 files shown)


monitor/grafana/zeebe.json
```diff
@@ -6236,6 +6204,107 @@
...
+            }
+          },
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${DS_PROMETHEUS}"
+              },
+              "editorMode": "code",
+              "expr": "sum(rate(container_cpu_cfs_throttled_periods_total{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$__rate_interval])) by (pod) / sum(rate(container_cpu_cfs_periods_total{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$__rate_interval])) by (pod)",
```
deepthidevaki, line 6298: :wrench: It would be good to also filter by pod. Sometimes, if I want to only inspect zeebe nodes finding and filtering it from legend when all components are running is too much :) .

deepthidevaki (approved):
:rocket: 

Does high throttling means, it can benefit from allocating more cpu?

Zelldon:
> Does high throttling means, it can benefit from allocating more CPU?

In general, it indicates that it doesn't get as much CPU as it needs to run. The resolution could be more CPU allocated or just better scheduling (on a different node), I would say depends on each case. 

For example, it can be that we have a pod that has request 1 and limit 2, means it can run up to 2 CPUs but there is no guarantee for it. In some cases, it would like to, but it can't as the node has no free CPU so it is throttled. That is at least my understanding. CPU scheduling depends on prioritization (pods with higher request/limits will have higher priority). 

Depending on the case, we could change the requests or the anti-affinity to have a better scheduling on different nodes. For example, I also observed that even putting a request and limit to the same value like 2, doesn't necessarily mean you will 2 CPUs all the time. See for example  https://github.com/camunda/camunda/issues/21376#issuecomment-2301317822

I hope this makes sense?

PR #21703, opened 2024-08-28
feat: create migration app skeleton

### LINKED ISSUE #21282: A skeleton Spring boot based Migration app exists for Identity in the monorepo
## Description
Migration app will be deployed on each execution cluster and have a connection to management identity and execution identity. The goal of this application is migrating the existing data on Postgres to the new architecture.

## Tasks

- [x] Create a spring module called migration in camunda/camunda repo.
- [x] Active profile for the migration app's beans is **migration**.
- [x] root package name is: io.camunda.identity.migration
- [x] Create an spring boot application in dist called MigrationApplication.
- [x] Migration application implements CommandLineRunner to get the args and pass it to the Migration management bean that reside in migration module. These args can be: 
   - mode: resume or restart (default to resume)
   - entities: tenant, user, .. (default to all)
   - command: migrate: do the migration or status: to print the latest statistics and status
- [x] Add dependency to camunda-service so then migration app can use identity related services from zeebe.
- [x] Add dependency to WebClient so then migration app can call management identity migration endpoints.

maryarm:
## Description
Migration app will be deployed on each execution cluster and have a connection to management identity and execution identity. The goal of this application is migrating the existing data on Postgres to the new architecture.

## Tasks

- [x] Create a spring module called migration in camunda/camunda repo.
- [x] Active profile for the migration app's beans is **migration**.
- [x] root package name is: io.camunda.identity.migration
- [x] Create an spring boot application in dist called MigrationApplication.
- [x] Migration application implements CommandLineRunner to get the args and pass it to the Migration management bean that reside in migration module. These args can be: 
   - mode: resume or restart (default to resume)
   - entities: tenant, user, .. (default to all)
   - command: migrate: do the migration or status: to print the latest statistics and status
- [x] Add dependency to camunda-service so then migration app can use identity related services from zeebe.
- [x] Add dependency to WebClient so then migration app can call management identity migration endpoints.
## Related issues

closes #21282


dist/src/main/java/io/camunda/application/IdentityMigrationApplication.java
```diff
@@ -0,0 +1,30 @@
...
+import io.camunda.identity.migration.MigrationRunner;
+import io.camunda.zeebe.gateway.GatewayModuleConfiguration;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.ImportAutoConfiguration;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.core.env.AbstractEnvironment;
+
+@SpringBootApplication(scanBasePackages = {"io.camunda.identity.migration"})
+@ImportAutoConfiguration(
```
romansmirnov, line 20: ❓ Would `@Import` be sufficient?

identity/migration/src/main/java/io/camunda/identity/migration/MigrationRunner.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Override
+  public void run(final ApplicationArguments args) throws Exception {
+    final String command =
+        args.containsOption("command") ? args.getOptionValues("command").getFirst() : "migrate";
+    final String mode =
+        args.containsOption("mode") ? args.getOptionValues("mode").getFirst() : "resume";
+    final String entities =
+        args.containsOption("entities") ? args.getOptionValues("entities").getFirst() : null;
```
Ben-Sheppard, line 45: 💭 as discussed, my understanding of the migration app is that it will run in a way that is designed to pick up from where it was during a pod restart and if there is any issue then we will rollback the cluster. 

Having thought more about it I'm not convinced these options will be used so I would prefer if they were removed for now (and we can add them in later on is something like this is needed) as it makes the code simpler and cleaner.

Ben-Sheppard (commented):
Just one comment about the necessity of the args but beyond that things look good to me :)

identity/migration/src/main/java/io/camunda/identity/migration/MigrationRunner.java
```diff
@@ -0,0 +1,52 @@
...
+    this.userService = userService;
+    this.authorizationServices = authorizationServices;
+    restTemplate = new RestTemplate();
+  }
+
+  @Override
+  public void run(final ApplicationArguments args) throws Exception {
+
+    final String command =
+        args.containsOption("command") ? args.getOptionValues("command").getFirst() : "migrate";
```
Ben-Sheppard, line 44: ❓ I know that the other command here is `status` would we just include logging to provide updates on where we are in the flow itself? It would also be worth confirming with distribution and controller if we can provide the app with these command line args in their deployment styles :)

Ben-Sheppard (approved):
One remaining comment but I'm going to approve because the question I have is something that we can also iterate on and it should block progress - if you feel it can be addressed then please do, otherwise we can think about it as we go :)

identity/migration/src/main/java/io/camunda/identity/migration/MigrationRunner.java
maryarm: Ok, good idea. if they can't handle command line args, we can provide a rest endpoint or a special health indicator to return status and stats.

Issue #21559, opened 2024-08-26
Productize search plugin architecture


npepinpe:
**Description**

[We recently merged the first iteration of the search client plugins](https://github.com/camunda/camunda/pull/20794). This allows users to define plugins which will intercept every request to our search database (e.g. Elasticsearch, Opensearch) in Operate, Tasklist, and Zeebe, and apply custom headers to these request.

As we wanted something that users can test in 8.6.0-alpha5, we ended up going with a very broad, hard to maintain approach, and lower test coverage than what we would usually accept.

The goal here is:

- Improve test coverage.
  - Lack of integration tests ensuring that the autowiring of the properties produces an appropriate connector.
- Centralize plugin loading/architecture to avoid code that is pasted six times in different parts.
  - This makes maintenance harder, as if you find a bug somewhere, you _need_ to fix it everywhere, and it's easy to forget some place.
- Integrate with new search-client architecture.

To do so, we will work on the following items:

- [x] In `ExternalJarClassLoader`, only warn if calling the default close. Add a new method to close explicitly if we want that will not warn.
  - This will be used further down the line.
  - That said, I already checked, and it's not strictly required to close the `ExternalJarClassLoader`, as any files opened for it are cleaned up during garbage collection. Furthermore, opening multiple ones for the same JAR does not consume more file descriptors, as there is some caching involved. So it's safe to open and forget to close, but this relies on undocumented implementation behavior, so we should look into closing those which will not live for the duration of the application lifetime (e.g. exporters).
- [x] Rename `camunda-search-plugins` to `camunda-search-client-plugin` module, and move it to a submodule of `camunda-search`, and set that as the parent.
  - It seems to me in the future, all search related modules would go there, so it makes sense to add our plugin one there as well. 
- [x] #21562
  -  This eases maintenance as we can now test the loading and intercepting via unit tests, making it easier to write controlled tests, and improving our test coverage.
  - I would advocate a similar `Repository` pattern as we have with the interceptors and exporters as well. I will write more in the issue dedicated to this.
- [x] #21563
- [x] Add missing integration tests (Operate ES/OS connectors, Tasklist ES/OS connectors).
  - Note that Zeebe seems to have no Elasticsearch/Opensearch QA tests, but just exporter integration tests. We can stick with that for now, as Zeebe exporter configuration is not instantiated through Spring anyway.
- [x] Update sample(s)
- [x] All required documentation changes

I will write follow up issues for the more "complex" action items.

[...]

PR #21731, opened 2024-08-29
feat: moved and renamed camunda-search-client-plugin

igpetrov:
## Description

feat: moved and renamed camunda-search-client-plugin

## Checklist

- [ ] for CI changes:

## Related issues

Related to https://github.com/camunda/camunda/issues/21559


search/pom.xml
```diff
@@ -19,6 +19,7 @@
     <module>search-client-elasticsearch</module>
     <module>search-client-opensearch</module>
     <module>search-client-connect</module>
+    <module>search-plugins</module>
```
npepinpe, line 22: :x: Please rename it like the others, i.e. `search-client-plugin`.

search/search-plugins/pom.xml
```diff
@@ -5,16 +5,16 @@
...
 
-  <artifactId>camunda-search-plugins</artifactId>
+  <artifactId>camunda-search-client-plugin</artifactId>
   <version>8.6.0-SNAPSHOT</version>
   <packaging>jar</packaging>
 
-  <name>Camunda Search Plugins</name>
+  <name>Camunda Search Client Plugin</name>
   <description>${project.name}</description>
   <inceptionYear>2024</inceptionYear>
```
npepinpe, line 19: :wrench: I think this can be removed.

npepinpe (approved):
:rocket: Please address all comments before merging, but no need for another review.

:wrench: One question - did we decide on a license for the `camunda-search-client-plugin`? As this is something customers need to include in their own code, the license may or may not be an issue. For example, for the `exporter-api`, we use Apache. 

I think if we don't specify one here, then it will be Camunda License v1.0. I just want to make sure we thought about it and decided for that here.

Issue #19006, opened 2024-05-31
Camunda exporter creates and updates Operate schema


sdorokhova:
### Description

Since Operate exporter is filling the Operate indices with data, exporter will also be responsible for creating and updating Operare indices. 

### TODO

- [x] create new Maven module that will store entities and index descriptor classes for all web apps #22274
- [x] move Operate entities to the new module: #22310
- [ ] move index and template descriptors as well as schema mapping JSONs to the new module: https://github.com/camunda/camunda/pull/22802
- [ ] ~~make sure that exporter creates proper schema on startup~~ -> another issue #22804
---------------------------------------------------------------------------------------------

### Acceptance Criteria

- [ ] Checked if the specific data needs to be provided for the QA process

### Definition of Ready - Checklist

- [ ] The issue has a meaningful title, description and testable acceptance criteria
- [ ] If documentation needs to be updated, an issue is created in the [camunda-docs](https://github.com/camunda/camunda-docs) repo, and the issue is added to our Operate project board.
- [ ] If HELM charts need to be updated, an issue is created in the [camunda-platform-heml](https://github.com/camunda/camunda-platform-helm) repo, and the issue is added to our Operate project board.
- [ ] Potential Security risks have been considered
- [ ] Cross-team dependencies have been considered

RomanJRW:
@sdorokhova - bumping here in case you missed the above in your absence

sdorokhova:
Hi @RomanJRW , sorry, missed your question (two times 🙈 )

We're implementing the same functionality that we have in Operate right now, which will include configuring shards and replica counts (one for all indices, or per index) and ILM.

RomanJRW:
@sdorokhova - all good, thanks for the response! 👍

sdorokhova:
One note or retry logic. In Operate we executed both schema creation and update with retry logic so that if Elastic is not available Operate won't fail to startup but rather will wait for some time for Elastic to come back.
In Camunda exporter we're not doing this in schema manager logic, but instead `CamundaExporter#open` method will be retried in case of failure (see https://github.com/camunda/camunda/issues/20728).

PR #21733, opened 2024-08-29
Refactor component descriptor

EuroLew:
## Description

Much easier to handle with the json string then a file as adding files to the class loader during run time is bad.

Also some refactoring to avoid repeated code with regards to `TemplateDetails`

## Related issues

relates to #19006

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/ComponentTemplateDescriptor.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/ComponentTemplateDescriptor.java
@@ -8,7 +8,7 @@
 package io.camunda.exporter.schema.descriptors;
 
 public interface ComponentTemplateDescriptor {
-  String getTemplateClasspathFileName();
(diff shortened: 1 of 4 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/ComponentTemplateDescriptor.java
```diff
@@ -8,7 +8,7 @@
 package io.camunda.exporter.schema.descriptors;
 
 public interface ComponentTemplateDescriptor {
-  String getTemplateClasspathFileName();
+  String getTemplateJson();
```
mustafadagher, line 11: 💭 Let's think together here about this a bit

Let's say Operate will implement a `ComponentTemplateDescriptor` (say `OperateComponentTemplateDescriptor`), you want them in this implementation to write formatted json representation of this component template?

I don't think that's advisable. 

For me, we either keep it as we did (representing it in json files next to index templates) or revert it to how operate did it to begin with (filling it from configs), but not halfway.

PR #21737, opened 2024-08-29
Validate user creation request

### LINKED ISSUE #21672: Validate user creation request in REST API
<h2>Description</h2>

The user creation REST endoint currently has no validations. We should make sure we validate the data provided by users is valid.

1. Username must be set and not empty
2. Password must be set and not empty
3. Name must be set and not empty
4. Email must be set and not empty
5. Email must be in a valid format

remcowesterhoud:
## Description

Adds a validation and mapping from the UserWithPasswordRequest DTO to an internal object (CreateUserRequest). When we have a CreateUserRequest the request is validated and we can assume all fields are correct.

The validation is quite straightforward. It checks for all fields that they are not empty or blank. Email validation is
a bit more tricky. It's always hard to determine if an email is valid or not. Instead of coming up with something myself
I used Apache's commons-validator. This has a built-in email validator. I did have to pull in a new dependency for this.

## Checklist

- [ ] for CI changes:

## Related issues

closes #21672

### DIFF
--- a/service/src/main/java/io/camunda/service/UserServices.java
+++ b/service/src/main/java/io/camunda/service/UserServices.java
@@ -52,4 +52,6 @@ public class UserServices extends SearchQueryService<UserServices, UserQuery, Us
             .setEmail(email)
             .setPassword(password));
   }
+
+  public record CreateUserRequest(String username, String name, String email, String password) {}
 }
--- a/service/src/main/java/io/camunda/service/UserServices.java
+++ b/service/src/main/java/io/camunda/service/UserServices.java
@@ -19,7 +19,7 @@ import io.camunda.zeebe.gateway.impl.broker.request.BrokerUserCreateRequest;
 import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
 import java.util.concurrent.CompletableFuture;
 
-public class UserServices<T> extends SearchQueryService<UserServices<T>, UserQuery, UserEntity> {
+public class UserServices extends SearchQueryService<UserServices, UserQuery, UserEntity> {
 
   public UserServices(final BrokerClient brokerClient, final CamundaSearchClient dataStoreClient) {
     this(brokerClient, dataStoreClient, null, null);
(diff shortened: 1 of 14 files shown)


remcowesterhoud:
@maryarm Pinging you to make sure you've seen I requested your review for this PR 🙂

service/src/main/java/io/camunda/service/UserServices.java
```diff
@@ -39,8 +39,8 @@ public SearchQueryResult<UserEntity> search(final UserQuery query) {
   }
 
   @Override
-  public UserServices<T> withAuthentication(final Authentication authentication) {
-    return new UserServices<>(brokerClient, searchClient, transformers, authentication);
+  public UserServices withAuthentication(final Authentication authentication) {
+    return new UserServices(brokerClient, searchClient, transformers, authentication);
   }
 
   public CompletableFuture<UserRecord> createUser(
```
maryarm, line 46: ❓As now we have a record for CreateUserRequest, isn't it better to change the signature of this method and pass the  CreateUserRequest as the arg?

service/src/main/java/io/camunda/service/UserServices.java
```diff
@@ -52,4 +52,6 @@ public CompletableFuture<UserRecord> createUser(
             .setEmail(email)
             .setPassword(password));
   }
+
+  public record CreateUserRequest(String username, String name, String email, String password) {}
```
maryarm, line 56: ❗Although we still don't have the update user in service/controller but I'm wondering if we need the same dto for it, or does it look different. (We were using the PUT method before and in that case they were same but I'm not sure about dto when using PATCH)

maryarm (commented):
Overall it looks good to me, I just have added a few comments/questions.

service/src/main/java/io/camunda/service/UserServices.java
remcowesterhoud: We might reuse it. But we can rename it once we implement that. No need to consider it in this PR 🙂

remcowesterhoud:
@maryarm I've processed your review comments. Please have another look 👀

maryarm (approved):
LGTM

PR #21750, opened 2024-08-29
Initial schema manager resources

### LINKED ISSUE #21528: Ability to create all initial schemas for schema manager
**Description**

The schema manager will be provided (in a way to be fully decided in the future), with a list of indices and templates to be instantiated in the search engine (Elasticsearch/Opensearch) and after all dependency injection has occurred it should create the schemas in the search engine.

EuroLew:
## Description

It creates the skeleton for a schema manager which will instantiate all indices and index templates provided at startup (not necessarily in the constructor it has not been fully decided yet). In addition it allows the configuration of the shards and replicas for the indices and templates.

## Related issues

closes #21528


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ElasticsearchClientFactory.java
```diff
@@ -0,0 +1,33 @@
...
+import co.elastic.clients.elasticsearch.ElasticsearchClient;
+import co.elastic.clients.json.jackson.JacksonJsonpMapper;
+import co.elastic.clients.transport.ElasticsearchTransport;
+import co.elastic.clients.transport.rest_client.RestClientTransport;
+import io.camunda.exporter.NoopExporterConfiguration.ElasticsearchConfig;
+import io.camunda.exporter.schema.ElasticsearchEngineClient;
+import org.apache.http.HttpHost;
+import org.elasticsearch.client.RestClient;
+
+public final class ElasticsearchClientFactory {
```
mustafadagher, line 19: is this being used anywhere in the scope of this PR?

mustafadagher, line 19: is this being used anywhere in the scope of this PR?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/NoopExporterConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public final class NoopExporterConfiguration {
```
mustafadagher, line 13: 💭 I understand you needed to make an example, but just an FYI, the NoopExporter is going to be removed once we have a concrete implementation available.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/NoopExporterConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+  public final ElasticsearchConfig elasticsearch = new ElasticsearchConfig();
+
+  public static final class ElasticsearchConfig {
+    public String url;
+    public IndexSpecificSettings defaultSettings = new IndexSpecificSettings();
+    public Map<String, String> replicasByIndexName = new HashMap<>();
+    public Map<String, String> shardsByIndexName = new HashMap<>();
+  }
+
+  public static final class IndexSpecificSettings {
```
mustafadagher, line 23: 🔧 the name could be better, but that's just my taste.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/NoopExporterConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+    public String numberOfShards = "1";
```
mustafadagher, line 24: we expect these to be passed as integers from the configuration, exposing them as strings might lead to miss-use.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
```diff
@@ -155,7 +157,13 @@ private PutIndexTemplateRequest putIndexTemplateRequest(
...
                       .mappings(
                           deserializeJson(IndexTemplateSummary._DESERIALIZER, templateMappings)
-                              .mappings()))
+                              .mappings())
+                      .settings(
+                          s ->
+                              s.index(
+                                  i ->
+                                      i.numberOfShards(settings.numberOfShards)
+                                          .numberOfReplicas(settings.numberOfReplicas))))
```
mustafadagher, line 166: what operate currently does with these is it uses these settings to create a component template with a default name (in runtime) before creating index templates. All index templates will be `composedOf` this component template. then, we don't have to pass these configurations around

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchSchemaManager.java
```diff
@@ -0,0 +1,54 @@
...
+      final ElasticsearchConfig elasticsearchConfig) {
+    this.elasticsearchClient = elasticsearchClient;
+    this.indexDescriptors = indexDescriptors;
+    this.indexTemplateDescriptors = indexTemplateDescriptors;
+    this.elasticsearchConfig = elasticsearchConfig;
+  }
+
+  @Override
+  public void initialiseResources() {
+    indexTemplateDescriptors.forEach(this::createIndexTemplate);
```
mustafadagher, line 35: related to the comment above:
we have the configs available here. before this line we can create a component template using the settings from configs

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/ComponentTemplateDescriptor.java
```diff
@@ -7,18 +7,6 @@
  */
 package io.camunda.exporter.schema.descriptors;
 
-public interface ComponentTemplateDescriptor {
+public interface ComponentTemplateDescriptor extends TemplateDetails {
```
mustafadagher, line 10: is this currently used anywhere?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/TemplateDetails.java
```diff
@@ -0,0 +1,22 @@
...
+public interface TemplateDetails {
+  String getTemplateName();
+
+  /**
+   * Maps to the create query parameter, if true the corresponding request cannot replace or update
+   * an existing component template, defaults to false
+   *
+   * @return The component template request is create only
+   */
+  default Boolean create() {
```
mustafadagher, line 19: I don't understand how is this going to be used

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/TestUtil.java
```diff
@@ -0,0 +1,54 @@
...
+
+import io.camunda.exporter.schema.descriptors.IndexDescriptor;
+import io.camunda.exporter.schema.descriptors.IndexTemplateDescriptor;
+import java.util.List;
+
+public final class TestUtil {
+
+  private TestUtil() {}
+
+  public static IndexTemplateDescriptor createIndexTemplate(
```
mustafadagher, line 22: this could be renamed to `mockIndexTemplate `

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/TestUtil.java
```diff
@@ -0,0 +1,54 @@
...
+    doReturn(alias).when(descriptor).getAlias();
+    doReturn(composedOf).when(descriptor).getComposedOf();
+
+    doReturn(templateName).when(descriptor).getTemplateName();
+    doReturn(mappingsFileName).when(descriptor).getMappingsClasspathFilename();
+
+    return descriptor;
+  }
+
+  public static IndexDescriptor createIndex(
```
mustafadagher, line 41: same

mustafadagher (changes requested):
Thanks @EuroLew! 
I made a few suggestions below.

❌  Also one observation: the commit history is a bit confusing, it cost me time trying to follow and find the relevant changes. I suggest to reset and recommit in an order that tells your story

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ElasticsearchClientFactory.java
EuroLew: my bad this was a WIP I forgot to delete

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
EuroLew: I don't understand the need to have a component template, yes every index template composed_of the component template for settings (and the config doesn't have to be "passed around") 

but the same applies here, the config is not passed around. An index/templates replicas/shards are calculated once in the schema manager and then passed to the request.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/ComponentTemplateDescriptor.java
EuroLew: Actually no, I added it because I thought component templates would be created, but if they aren't (in that settings don't require them) then this and `createComponentTemplate` in the `SearchEngineClient` can also be removed so its pending on the other point regarding component templates

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/TemplateDetails.java
EuroLew: The old operate client uses a custom `overwrite` when making index template requests, overwrite=true means you make the request and it overwrites the current template, overwrite=false means you only make the request if the template doesn't already exist.

This is the exact same behaviour as create, create=true == overwrite=false and vice versa.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/NoopExporterConfiguration.java
EuroLew: Yea I understand, but not sure where else to place it, maybe an CamundaExporterConfiguration? But eventually it'll need to get merged with whatever implementation you add anyway

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
EuroLew: At some point we need to check if there is some per index replicas/shards settings, and if not we set it to some default I feel adding a component template is just redundant.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchSchemaManager.java
EuroLew: same point as above

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/NoopExporterConfiguration.java
mustafadagher: That was just an FYI. No need to change anything right now

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
mustafadagher: we can leave it as is right now, and change later if needed.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/ComponentTemplateDescriptor.java
mustafadagher: you can remove it now. we shall recreate it in the future if we revisit the decision about component templates

EuroLew: removed

EuroLew:
Cleaned up the commit history here, https://github.com/camunda/camunda/pull/21871

PR #21767, opened 2024-08-30
feat: support authorization query in service layer

### LINKED ISSUE #21338: Implement `AuthorizationQuery` class

### LINKED ISSUE #21340: Update `AuthorizationServices` class to support search
### Description
- [ ] Add `search` method to `Authorization Services
  - [ ] Should return `SearchQueryResult<AuthorizationEntity>`

Ben-Sheppard:
## Description

This PR adds in the authorization query and also the relevant search method in the authorization service class. The functionality here forms the foundations of being able to apply authorizations to other queries in an ES context.

## Related issues

closes #21338, closes #21340


service/src/test/java/io/camunda/service/query/filter/AuthorizationFilterTest.java
```diff
@@ -0,0 +1,107 @@
...
+            SearchTermQuery.class,
+            t -> {
+              assertThat(t.field()).isEqualTo(column);
+              assertThat(t.value().stringValue()).isEqualTo(value);
+            });
+  }
+
+  public static Stream<Arguments> queryFilterParameters() {
+    return Stream.of(
+        Arguments.of(
```
maryarm, line 88: ❓In my idea, it's ok to search by resourceType or ownerType or resourceKey+resourceType and ownerType+ownerKey and combination of them, but it's better to not allow to search by ownerKey without ownerType and also same for resourceKey. wdyt?

maryarm (commented):
It looks good to me 👍, I have added a question regarding allowed filters.

service/src/test/java/io/camunda/service/query/filter/AuthorizationFilterTest.java
Ben-Sheppard: Good question, so keys in Zeebe are unique so there wouldn't be a case where the same key is used for both an owner or a resource so its fine here to allow searching by one without a contextual type filter :)

maryarm (approved):
💯! Looks good.

PR #21783, opened 2024-08-30
feat: add flow node instance search

ralfpuchert:
## Description

* Add Filter, Sort for FlownodeInstance
* Add tests
* Based on https://github.com/camunda/camunda/pull/21624

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/ZeebeClient.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/ZeebeClient.java
@@ -44,6 +44,7 @@ import io.camunda.zeebe.client.api.fetch.DecisionDefinitionGetXmlRequest;
 import io.camunda.zeebe.client.api.response.ActivatedJob;
 import io.camunda.zeebe.client.api.search.query.DecisionDefinitionQuery;
 import io.camunda.zeebe.client.api.search.query.DecisionRequirementsQuery;
+import io.camunda.zeebe.client.api.search.query.FlownodeInstanceQuery;
 import io.camunda.zeebe.client.api.search.query.IncidentQuery;
 import io.camunda.zeebe.client.api.search.query.ProcessInstanceQuery;
 import io.camunda.zeebe.client.api.search.query.UserTaskQuery;
@@ -700,6 +701,31 @@ public interface ZeebeClient extends AutoCloseable, JobClient {
    */
   @ExperimentalApi("https://github.com/camunda/camunda/issues/20596")
   ProcessInstanceQuery newProcessInstanceQuery();
+
+  /**
+   * Executes a search request to query flow node instances.
+   *
+   * <pre>
+   * long flownodeInstanceKey = ...;
+   *
+   * zeebeClient
+   *  .newFlownodeInstanceQuery()
+   *  .filter((f) -> f.processInstanceKeys(processInstanceKey))
+   *  .sort((s) -> s.flowNodeName().asc())
+   *  .page((p) -> p.limit(100))
+   *  .send();
+   * </pre>
+   *
+   * <p><strong>Experimental: This method is under development, and as such using it may have no
+   * effect on the client builder when called. The respective API on compatible clusters is not
+   * enabled by default. Thus, this method doesn't work out of the box with all clusters. Until this
+   * warning is removed, anything described below may not yet have taken effect, and the interface
+   * and its description are subject to change.</strong>
+   *
+   * @return a builder for the process instance query
+   */
+  @ExperimentalApi("https://github.com/camunda/camunda/issues/20596")
+  FlownodeInstanceQuery newFlownodeInstanceQuery();
 
   /**
    * Executes a search request to query user tasks.
--- a/clients/java/src/main/java/io/camunda/zeebe/client/api/search/SearchRequestBuilders.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/api/search/SearchRequestBuilders.java
@@ -17,24 +17,28 @@ package io.camunda.zeebe.client.api.search;
 
 import io.camunda.zeebe.client.api.search.filter.DecisionDefinitionFilter;
 import io.camunda.zeebe.client.api.search.filter.DecisionRequirementsFilter;
+import io.camunda.zeebe.client.api.search.filter.FlownodeInstanceFilter;
 import io.camunda.zeebe.client.api.search.filter.IncidentFilter;
 import io.camunda.zeebe.client.api.search.filter.ProcessInstanceFilter;
 import io.camunda.zeebe.client.api.search.filter.UserTaskFilter;
 import io.camunda.zeebe.client.api.search.filter.VariableValueFilter;
 import io.camunda.zeebe.client.api.search.sort.DecisionDefinitionSort;
 import io.camunda.zeebe.client.api.search.sort.DecisionRequirementsSort;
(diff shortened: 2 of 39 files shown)


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/FlownodeInstanceQueryController.java
```diff
@@ -51,11 +52,18 @@ private ResponseEntity<FlownodeInstanceSearchQueryResponse> search(
               .search(query);
       return ResponseEntity.ok(
           SearchQueryResponseMapper.toFlownodeInstanceSearchQueryResponse(result));
-    } catch (final Throwable e) {
+    } catch (final ValidationException e) {
```
houssain-barouni, line 55: I am wondering what can throw `jakarta.validation.ValidationException` in the try block 🤔 , are we using some constraint annotations ?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/FlownodeInstanceQueryController.java
```diff
@@ -51,11 +52,18 @@ private ResponseEntity<FlownodeInstanceSearchQueryResponse> search(
...
       final var problemDetail =
           RestErrorMapper.createProblemDetail(
               HttpStatus.BAD_REQUEST,
               e.getMessage(),
+              "Validation failed for Flownode Instance Search Query");
+      return RestErrorMapper.mapProblemToResponse(problemDetail);
+    } catch (final Throwable e) {
```
houssain-barouni, line 62: I suggest we log this unexpected exception stack trace. This can be helpful in troubleshooting, and `e.getMessage `may not be explicit enough
```
REST_LOGGER.warn("An exception occurred in searchFlowNodeInstances.", e);
```

houssain-barouni (commented):
The client code LGTM (other than the Flow**N**odeInstance naming that I asked about in another PR)
I had few comments about the changes in the controller code

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/FlownodeInstanceQueryController.java
nathansandi: Hey @houssain-barouni, I gonna answer this one as I am proposed this in the past 

Possible Sources of ValidationException in the Code:

	1.	Bean Validation on Request Objects - (@Valid, @NotNull, etc.) can trigger validation before the method is invoked. However, in this code, there is no explicit validation annotation like @Valid on the query parameter of the searchFlownodeInstances method. This reduces the likelihood that the exception is triggered directly by this controller method.
	2.	Manual Validation within Services: - The most likely place where the ValidationException could be thrown is within the InstanceServices class or any related service that handles the Query object.
	3.	Search Query Validation: If Query (or its related data structures) has any validation annotations and those are programmatically validated within the service layer, a ValidationException could be thrown if the data does not meet the expected constraints.

 The `ValidationException` is caught to handle scenarios where the query or any of its fields fail these validation checks, ensuring the client receives a meaningful error response.
And the goal is not return only 500 Status code, in case of any error, but separate the validation code between Request/Internal server.

clients/java/src/main/java/io/camunda/zeebe/client/ZeebeClient.java
```diff
@@ -590,6 +591,31 @@ static ZeebeClientCloudBuilderStep1 newCloudClientBuilder() {
...
+   * <p><strong>Experimental: This method is under development, and as such using it may have no
+   * effect on the client builder when called. The respective API on compatible clusters is not
+   * enabled by default. Thus, this method doesn't work out of the box with all clusters. Until this
+   * warning is removed, anything described below may not yet have taken effect, and the interface
+   * and its description are subject to change.</strong>
+   *
+   * @return a builder for the process instance query
+   */
+  @ExperimentalApi("https://github.com/camunda/camunda/issues/20596")
+  FlownodeInstanceQuery newFlownodeInstanceQuery();
```
nathansandi, line 617: I would suggest naming as newFlowNodeInstanceQuery (camel case as suggested on previous PR)

clients/java/src/main/java/io/camunda/zeebe/client/api/search/SearchRequestBuilders.java
```diff
@@ -156,4 +160,25 @@ public static DecisionRequirementsSort decisionRequirementsSort(
     fn.accept(sort);
     return sort;
   }
+
+  public static FlownodeInstanceFilter flownodeInstanceFilter() {
```
nathansandi, line 164: 🔧  Same change as previous one

clients/java/src/main/java/io/camunda/zeebe/client/impl/search/query/FlownodeInstanceQueryImpl.java
```diff
@@ -0,0 +1,120 @@
...
+  public FinalSearchQueryStep<FlownodeInstance> requestTimeout(final Duration requestTimeout) {
+    httpRequestConfig.setResponseTimeout(requestTimeout.toMillis(), TimeUnit.MILLISECONDS);
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<SearchQueryResponse<FlownodeInstance>> send() {
+    final HttpZeebeFuture<SearchQueryResponse<FlownodeInstance>> result = new HttpZeebeFuture<>();
+    httpClient.post(
+        "/user-tasks/search",
```
nathansandi, line 71: 🔧  I think here is  "/flownode-instances/search" right?

nathansandi (dismissed):
Great work over here, some minor changes only

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/FlownodeInstanceQueryController.java
houssain-barouni: The caller will send a query that will be validated either by the annotations generated in the API data model or through manual validations, both of which occur outside this try block. If any other validation fails within this try block, it is beyond the caller’s control, meaning they cannot adjust the query to make it work. So, the 400 error in this case is kind of useless.

ghost (approved):

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1831,6 +1873,85 @@ components:
...
 
+    FlowNodeInstanceSearchQueryRequest:
+      allOf:
+        - $ref: "#/components/schemas/SearchQueryRequest"
+      type: object
+      properties:
+        filter:
+          allOf:
+            - $ref: "#/components/schemas/FlowNodeInstanceFilterRequest"
+    FlowNodeInstanceFilterRequest:
```
houssain-barouni, line 1884: I think we should add description to all these fields

ralfpuchert: Thanks! `state` and `type` can also be enums. I propose to tackle this in another PR I need to revise the Operate related API definitions anyway. So there might be more changes to come.

houssain-barouni (approved):

PR #21795, opened 2024-08-30
feat: create default user in application startup

### LINKED ISSUE #21421: Camunda should support creation of a default user to be authorized to access endpoints when auth-basic is enabled
### Description
After enabling the basic authentication, camunda needs at least one user to access endpoints.

### Scope
In this issue we just allow creating a default user to pass the basic auth and permissions come later with next issues.

maryarm:
## Description

After enabling the basic authentication, camunda needs at least one user to access endpoints.

## Related issues

closes #


service/src/main/java/io/camunda/service/ApiServices.java
```diff
@@ -55,7 +55,9 @@ public T withAuthentication(
   }
 
   protected <R> CompletableFuture<R> sendBrokerRequest(final BrokerRequest<R> brokerRequest) {
-    brokerRequest.setAuthorization(authentication.token());
+    if (authentication != null) {
```
Ben-Sheppard, line 58: ❓ We should validate this change with someone more familiar with the wider code than me, it seems fine but I wonder if maybe @houssain-barouni could see something here that may cause an issue?

Ben-Sheppard (commented):
Some comments from my side :)

service/src/main/java/io/camunda/service/ApiServices.java
houssain-barouni: It seems this change was necessary because the broker calls to `UserServices.search` and `UserServices.createUser` within `DataInitializer` without passing the authentication.

When the gateway operates in standalone mode and needs to communicate with a remote broker, authorization may be needed. You should use `userServices.withAuthorization(RequestMapper.getAuthentication()).createUser(..)` for these calls.

Since data initialization occurs outside the scope of an HTTP request, `RequestMapper.getAuthentication()` might fail to retrieve the authorized tenants list (`List<String> authorizedTenants = TenantAttributeHolder.tenantIds()`), which needs to be confirmed. If this is the case, you may need to create a new method that generates the authentication object without tenants, as they might not be necessary for this use case.

maryarm: Thanks @houssain-barouni for clarification, what I face here is that it's the calls to create a user and before that we don't have any user available to be able to pass the basic auth. Do you know any workaround in this case?

houssain-barouni, line 58: @maryarm with your changes, this null check is not needed anymore isn't?

remcowesterhoud:
@maryarm Sorry for not looking at this PR sooner. Is this supposed to be a temporary solution? The way I see it the default user must be created within Zeebe itself. As Zeebe is going to take care of user management it should be Zeebe's responsibility to create the default user.

With the solution you've implemented here you send a user create command to Zeebe. What will the authorization be on this request? There is no user, so Zeebe will likely reject the create command sying it's unauthorized to create users.

maryarm:
> @maryarm Sorry for not looking at this PR sooner. Is this supposed to be a temporary solution? The way I see it the default user must be created within Zeebe itself. As Zeebe is going to take care of user management it should be Zeebe's responsibility to create the default user.
> 
> With the solution you've implemented here you send a user create command to Zeebe. What will the authorization be on this request? There is no user, so Zeebe will likely reject the create command sying it's unauthorized to create users.



> @maryarm Sorry for not looking at this PR sooner. Is this supposed to be a temporary solution? The way I see it the default user must be created within Zeebe itself. As Zeebe is going to take care of user management it should be Zeebe's responsibility to create the default user.
> 
> With the solution you've implemented here you send a user create command to Zeebe. What will the authorization be on this request? There is no user, so Zeebe will likely reject the create command sying it's unauthorized to create users.

It's a workaround for now to have some default users, for authentication as @houssain-barouni has suggested we used a m2m JWT Token. As we are not checking user's specific permissions for now, it will work, but I think we need to do some hacks later if we add permission check to `UserCreateProcessor`. I hope until that time we find a way to do it the correct way and Zeebe itself manage it.

remcowesterhoud:
> It's a workaround for now to have some default users

Can I propose we don't close #21421 with this PR in that case? That way we can use it for the proper solution. I [made a PoC for it today](https://github.com/camunda/camunda/pull/22198) which seems to work fine.

PR #21832, opened 2024-09-02
Export `userKey` in UserRecord

### LINKED ISSUE #21830: Export userKey in UserRecord
<h2>Description</h2>

We need to export the userKey in the UserRecord. This key will be used by the webapps to perform further actions on the user.

remcowesterhoud:
## Description

We must export the userKey so consumers of this record are aware which user the events relate to.

## Checklist

- [ ] for CI changes:

## Related issues

closes #21830

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
@@ -63,6 +63,8 @@ public class UserCreateProcessor implements DistributedTypedRecordProcessor<User
     }
 
     final long key = keyGenerator.nextKey();
+    command.getValue().setUserKey(key);
+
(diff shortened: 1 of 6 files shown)


Ben-Sheppard (approved):

Ben-Sheppard:
I think its NABD for now @remcowesterhoud but with https://github.com/camunda/camunda/pull/21392/files we added in the start of the new exporters where there is now an entity class for the user object, it has an `id` field so maybe that would be `key` here, but maybe that class should also have key as a specific field?

remcowesterhoud:
> I think its NABD for now @remcowesterhoud but with https://github.com/camunda/camunda/pull/21392/files we added in the start of the new exporters where there is now an entity class for the user object, it has an `id` field so maybe that would be `key` here, but maybe that class should also have key as a specific field?

Yes, I would propose to name it key, or userKey to keep the naming across components aligned. Otherwise we will get confusion again with regards to what an id is, and what a key is.

PR #21841, opened 2024-09-02
feat: create initial version of ElasticsearchExporter

### LINKED ISSUE #21838: Create initial version of ElasticsearchExporter

mustafadagher:
## Description

create initial version of ElasticsearchExporter

## Related issues

closes #21838

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ElasticsearchExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ElasticsearchExporter.java
@@ -11,6 +11,7 @@ import static io.camunda.zeebe.protocol.record.ValueType.USER;
 
 import co.elastic.clients.elasticsearch.ElasticsearchClient;
 import co.elastic.clients.elasticsearch.core.BulkRequest;
+import io.camunda.exporter.clients.elasticsearch.ElasticsearchClientFactory;
 import io.camunda.exporter.config.ElasticsearchExporterConfiguration;
 import io.camunda.exporter.exceptions.ElasticsearchExporterException;
 import io.camunda.exporter.exceptions.PersistenceException;
@@ -90,9 +91,8 @@ public class ElasticsearchExporter implements Exporter {
     }
   }
 
-  private static ElasticsearchClient createClient() {
-    // FIXME call a client factory to instantiate `client` here
-    return new ElasticsearchClient(null);
+  private ElasticsearchClient createClient() {
+    return ElasticsearchClientFactory.INSTANCE.create(configuration.elasticsearch);
   }
 
   private boolean shouldFlush() {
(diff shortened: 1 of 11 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ElasticsearchExporter.java
```diff
@@ -0,0 +1,156 @@
...
+    } catch (final PersistenceException ex) {
+      throw new ElasticsearchExporterException(ex.getMessage(), ex);
+    }
+  }
+
+  private void updateLastExportedPosition() {
+    controller.updateLastExportedRecordPosition(lastPosition, null);
+  }
+
+  private record ElasticsearchRecordFilter() implements RecordFilter {
```
mustafadagher, line 142: It's not obvious to me if the supported `RecordTypes` and `ValueTypes` will need to be configurable at some point

mustafadagher:
The PR is missing some Integration tests for the exporter, could be added in a followup PR.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/clients/elasticsearch/ElasticsearchClientFactory.java
```diff
@@ -0,0 +1,136 @@
...
+    } catch (final Exception e) {
+      throw new ElasticsearchExporterException("Couldn't connect to Elasticsearch. Abort.", e);
+    }
+  }
+
+  private HttpAsyncClientBuilder configureHttpClient(
+      final HttpAsyncClientBuilder httpAsyncClientBuilder,
+      final ElasticsearchProperties elsConfig) {
+    setupAuthentication(httpAsyncClientBuilder, elsConfig);
+    // TODO setup SSLContext if configured
```
mustafadagher, line 97: This is still missing.

deepthidevaki:
I made the following changes to the PR

1. Removed the retry for health check when creating the ElasticsearchClient. The retries will block the Zeebe exporter thread for the whole time. It is enough to just check the health once and log the warning. We are not failing the client creation anyway. So there is no value in retrying for such a long time.
2. Renamed it to `CamundaExporter` to avoid confusion with the existing ElasticsearchExporter. My assumption is we will extend this to also support OpenSearch. If we end up having another implementation for Opensearch we can rename it then.
3. Added a single acceptance test to verify that everything works. More tests need to be added. 

I will add followup issues for the following:
Tests:
* Verify that basic authentication using configured username + password works
* Verify flush behavior
    * Should flush periodically
    * Should use flush configuration
* Verify position is updated after flushing
* Verify that record is exported when elastic is not initially reachable, but is reachable later.

Configuration:
* Configure bulksize if needed, and use it for flush
* Validate configuration

@EuroLew Could you do a second round of review?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ElasticsearchExporterConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+  public String toString() {
+    return "ElasticsearchExporterConfiguration{"
+        + "bulk="
+        + bulk
+        + ", elasticsearch="
+        + elasticsearch
+        + '}';
+  }
+
+  public static class BulkConfiguration {
```
EuroLew, line 24: If we're going to use getters and setters for config do it here as well?

EuroLew (approved):

PR #21844, opened 2024-09-02
Add support to text/xml response contentType in REST client

houssain-barouni:
## Description

Currently the REST client supports only JSON response content types, it does not support endpoints that returns plain text,
for now I added support to `text/xml` mime type but it can be extended to other mime types

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/ApiEntityConsumer.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/ApiEntityConsumer.java
@@ -70,8 +70,9 @@ final class ApiEntityConsumer<T> extends AbstractBinAsyncEntityConsumer<ApiEntit
       entityConsumer = new JsonApiEntityConsumer<>(json, type, false);
     } else {
       final boolean isResponse =
-          SUPPORTED_TEXT_CONTENT_TYPES.stream().anyMatch(t -> t.isSameMimeType(contentType));
-      entityConsumer = new RawApiEntityConsumer(isResponse);
+          String.class.equals(type)
+              && SUPPORTED_TEXT_CONTENT_TYPES.stream().anyMatch(t -> t.isSameMimeType(contentType));
+      entityConsumer = new RawApiEntityConsumer<>(isResponse);
     }
   }
 
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/ApiEntityConsumer.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/ApiEntityConsumer.java
@@ -23,6 +23,7 @@ import io.camunda.zeebe.client.protocol.rest.ProblemDetail;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.List;
 import org.apache.hc.core5.http.ContentType;
 import org.apache.hc.core5.http.nio.entity.AbstractBinAsyncEntityConsumer;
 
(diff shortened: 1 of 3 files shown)


clients/java/src/main/java/io/camunda/zeebe/client/impl/http/ApiEntityConsumer.java
```diff
@@ -87,13 +92,16 @@ protected ApiEntity<T> generateContent() throws IOException {
       if (nonJsonBody == null || bufferedBytes == 0) {
         return null;
       }
+      if (responseType == ResponseType.TEXT) {
+        return (ApiEntity<T>) ApiEntity.of(new String(nonJsonBody, 0, bufferedBytes));
+      }
 
       return ApiEntity.of(ByteBuffer.wrap(nonJsonBody, 0, bufferedBytes));
```
tmetzke, line 99: 🔧 The `ResponseType` could have a `wrapValue` method or similar that returns the wrapped `nonJsonBody` when it's handed in. That way, we don't have to `if-return` here again, we already have a type and it should know how to return the wrapped value.

💭 It might actually make more sense to create a response object that encapsulates the logic of the underlying content/response type. Otherwise, this method here will become hard to understand really quickly if we add more supported response types. For example, the response object can hold a parser and buffer if it's a JSON response (JSON or JSON_PROBLEM). For XML text, the response object wouldn't have such objects at all. I can also create a commit to visualize what I have in mind later, if you like.

tmetzke (commented):
👍 I like the general direction of this and that it's apparently not too complex to support other content types.
🔧 We must add test cases for this, of course, when we've settled for a direction.
💭 How about we use the momentum and clean that `ApiEntityConsumer` class up a bit if we want to handle different content types that wrap the response string differently?

houssain-barouni:
Hi @tmetzke , I introduced `TypedApiEntityConsumer` interface and its implementations (`JsonApiEntityConsumer` and `RawApiEntityConsumer`) into the `ApiEntityConsumer` which makes it more modular.
I also created a class of tests `ApiEntityConsumerTest`.

Please let me know if you have any concerns/feedback

tmetzke (commented):
👍 Functionally, this looks good to me.

💭 Conceptually, we introduce a typed API entity consumer that might only fail at runtime when I provide an expected type `MyClass` and receive a `String` instead. We must use this class correctly to avoid it, which is error-prone moving forward. I have played around with the code locally a bit. I think there are ways to avoid this, e.g. by not using an `ApiEntityConsumer` in the `ApiResponseConsumer` when we only expect a plain text response instead of a JSON-parseable entity. I think we can look into that in a follow-up to make progress here.

koevskinikola:
❓ @houssain-barouni do you still need my input here, or is @tmetzke's review enough?

houssain-barouni:
> 👍 Functionally, this looks good to me.
> 
> 💭 Conceptually, we introduce a typed API entity consumer that might only fail at runtime when I provide an expected type `MyClass` and receive a `String` instead. We must use this class correctly to avoid it, which is error-prone moving forward. I have played around with the code locally a bit. I think there are ways to avoid this, e.g. by not using an `ApiEntityConsumer` in the `ApiResponseConsumer` when we only expect a plain text response instead of a JSON-parseable entity. I think we can look into that in a follow-up to make progress here.

I can rely on the `type` field in `ApiEntityConsumer` which is the expected class of the response. If it is `String.class`, I consider the entity as real response otherwise it is considered as 'unknown'.
We can even skip the check on the MimeType in this case.

houssain-barouni:
> ❓ @houssain-barouni do you still need my input here, or is @tmetzke's review enough?

Unless you see a blocker, you are free to go 😄

tmetzke:
> I can rely on the `type` field in `ApiEntityConsumer` which is the expected class of the response. If it is `String.class`, I consider the entity as real response otherwise it is considered as 'unknown'. We can even skip the check on the MimeType in this case.

Yeah, that is also something I considered 👍

tmetzke (approved):
👍 Nice refactoring, @houssain-barouni, I like it 🚀

houssain-barouni:
@tmetzke I applied the requested changes in  https://github.com/camunda/camunda/pull/21844/commits/eb4621b69068578528a0310473ac24509e27e9ef
I kept the checks on the mimeType to prevent returning strange Strings for some unexpected responses. We can revisit it in the future

Issue #21722, opened 2024-08-29
Unstable Optimize backend tests


Zelldon:
Engineering DRI: @matthewBearCamunda

## Description

It seems that the `optimize-backend-unit-tests` is unstable in performance. It seems to timeout more often in the last 2 days, based on metrics and observations of failing CI builds. It would be good to look into this further. 

**Why is this important:**

Unstable/unreliable CI builds affect the effectiveness and volatility of engineers. It is harder to merge PRs, may it be handcrafted or automated dependency updates. 

Furthermore, it causes Medics (like Zeebe medic) look into CI failures, see [#zeebe-ci](https://app.slack.com/client/T0PM0P1SA/C013MEVQ4M9).

**Example failed builds** (from last days):

* https://github.com/camunda/camunda/actions/runs/10592814828/job/29352967793
* https://github.com/camunda/camunda/actions/runs/10604303486/job/29390660322
* https://github.com/camunda/camunda/actions/runs/10605037217/job/29393018948

**Statistics**

![2024-08-29_09-26](https://github.com/user-attachments/assets/1b1cae68-b9ed-4254-af6d-608b327fcc4d)

misiekhardcore:
it looks like these tests are timing out because they are downloading dependencies instead of using cache. Is there a known solution to this problem? I know that these kind of caching issues occurred in other workflows as well, how were they solved?

cmur2:
More examples after the merge of https://github.com/camunda/camunda/commit/1168140edfa0d991851b69559eedc607ff20be65

* https://github.com/camunda/camunda/actions/runs/10625631897/job/29455916164
* https://github.com/camunda/camunda/actions/runs/10630718220/job/29470197481

cmur2:
@grlimacan I think the fix in https://github.com/camunda/camunda/commit/1168140edfa0d991851b69559eedc607ff20be65 didn't (fully?) help

grlimacan:
> @grlimacan I think the fix in [1168140](https://github.com/camunda/camunda/commit/1168140edfa0d991851b69559eedc607ff20be65) didn't (fully?) help

Hmm, it looks at least more stable than before. What do you think about reducing it to 0.5 as Chris hinted at yesterday?

cmur2:
> Hmm, it looks at least more stable than before.

[...]

PR #21852, opened 2024-09-03
ci: add Optimize migration IT to CI

HeleneW-dot:
## Description

- Adds Migration IT to Optimize CI
- Fixes template deletion step, the issue was that it was using the rollover index name (with rollover suffix) as template name. Due to that and since the index is also deleted in the same migration, I have adjusted the logic to build it name manually and not rely on existing aliases or indices. >> moved to [separate PR](https://github.com/camunda/camunda/pull/21935)
- Spotless adjusted the spacing in the CI file a bunch, odd it didnt do this for your PR. Do you think this is an incorrect adjustment?

## Checklist

- [x] for CI changes:

## Related issues

related to https://github.com/camunda/camunda-optimize/issues/14144


.github/workflows/optimize-ci.yml
```diff
@@ -70,73 +69,73 @@ jobs:
...
+
+      - name: "Read Java / Version Info"
+        id: "pom-info"
+        uses: YunaBraska/java-info-action@main
+
+      - name: Expose Parsed Elastic Version
+        run: |
+          echo "ELASTIC_VERSION=${{ steps.pom-info.outputs.x_elasticsearch_test_version }}" >> "$GITHUB_ENV"
+
+      - name: Start Elastic
```
RomanJRW, line 101:
```suggestion
      - name: Start Elasticsearch
```

.github/workflows/optimize-ci.yml
```diff
@@ -297,17 +296,79 @@ jobs:
         with:
           archive_name: integration-tests-opensearch-dev-${{ matrix.includedGroups }}-docker
 
+  migration:
+    name: Migration
```
RomanJRW, line 300: I think we should call this `Upgrade`, especially now the other is called schema migration

.github/workflows/optimize-ci.yml
```diff
@@ -297,17 +296,79 @@ jobs:
         with:
           archive_name: integration-tests-opensearch-dev-${{ matrix.includedGroups }}-docker
 
+  migration:
```
RomanJRW, line 299:
```suggestion
  upgrade:
```

.github/workflows/optimize-ci.yml
```diff
@@ -297,17 +296,79 @@ jobs:
...
+      - name: Login to Harbor registry
+        uses: ./.github/actions/login-registry
+        with:
+          secrets: ${{ toJSON(secrets) }}
+
+      - name: "Read Java / Version Info"
+        id: "pom-info"
+        uses: YunaBraska/java-info-action@main
+
+      - name: Start Elastic
```
RomanJRW, line 320:
```suggestion
      - name: Start Elasticsearch
```

.github/workflows/optimize-ci.yml
```diff
@@ -297,17 +296,79 @@ jobs:
...
+        if: always()
+        with:
+          archive_name: migration-docker
+
   # Only deploy artifacts on push event, which in this case can only be triggered by main and stable
   deploy-artifacts:
     if: github.event_name == 'push' && (github.ref_name == 'main' || startsWith(github.ref_name, 'stable/'))
     name: Deploy Artifacts
-    needs: [integration-tests]
+    needs: [ integration-tests, migration ]
```
RomanJRW, line 364: same here

.github/workflows/optimize-ci.yml
```diff
@@ -297,17 +296,79 @@ jobs:
...
     uses: ./.github/workflows/optimize-deploy-artifacts.yml
     secrets: inherit
 
   # Rerun failed jobs running on self-hosted runners in case of network issues or node preemption
   rerun-failed-jobs:
     needs:
+      - migration
```
RomanJRW, line 371: here too

.github/workflows/optimize-ci.yml
```diff
@@ -297,17 +296,79 @@ jobs:
...
     if: github.event_name == 'push' && (github.ref_name == 'main' || startsWith(github.ref_name, 'stable/'))
     name: Deploy Artifacts
-    needs: [integration-tests]
+    needs: [ integration-tests, migration ]
     uses: ./.github/workflows/optimize-deploy-artifacts.yml
     secrets: inherit
 
   # Rerun failed jobs running on self-hosted runners in case of network issues or node preemption
   rerun-failed-jobs:
     needs:
```
RomanJRW, line 370: what do you think about adding these to the Unified CI? They run quickly and are stable, do they add value?

optimize/upgrade/src/test/java/io/camunda/optimize/upgrade/AbstractUpgradeIT.java
```diff
@@ -0,0 +1,352 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.optimize.upgrade;
```
RomanJRW, line 8: are there any changes in these test files that I should be aware of? Or are they all just uncommented?

RomanJRW (commented):
Not approving or rejecting right now, just want to clarify the scope of the review

optimize/upgrade/src/test/java/io/camunda/optimize/upgrade/AbstractUpgradeIT.java
PHWaechtler: Just commenting them back in except for the template one where I also fixed the assertion.

.github/workflows/optimize-ci.yml
PHWaechtler: Yeah that sounds reasonable. Guidelines say there must have been 10 green runs, but I can retrigger this one. I would then also remove them from optimize-ci so theyre not run twice on PRs right?
Wdyt about when to run? on upgrade changes obviously but there could also be backend changes that could influence the upgrade, eg if an index file is changed in a way that doesnt match the defined upgrade plan and we happen to have tests touching this, we would catch it if we run upgrade IT on general backend changes. I"m thinking of scenarios like having a new introduced field and then renaming it without adjusting the migration. Do you think its worth running them on all backend changes even though in most cases running them on all backend changes would be unnecessary?

PHWaechtler: Added this as a separate PR, after you re-review I'll ping the CI DRI as well.

.github/workflows/ci.yml
```diff
@@ -484,7 +484,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4
         with:
-          name: unit-tests-backend-junit
+          name: optimize-unit-tests-backend-junit
```
RomanJRW, line 487: is this an unrelated fix?

.github/workflows/ci.yml
```diff
@@ -446,7 +446,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4
         with:
-          name: unit-tests-frontend-junit
+          name: optimize-unit-tests-frontend-junit
```
RomanJRW, line 449: is this an unrelated fix?

.github/workflows/ci.yml
```diff
@@ -499,6 +499,70 @@ jobs:
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
 
+  optimize-upgrade-tests:
+    if: needs.detect-changes.outputs.optimize-backend-changes == 'true'
```
RomanJRW, line 503: I think this makes sense 👍

.github/workflows/optimize-ci.yml
```diff
@@ -301,13 +300,14 @@ jobs:
   deploy-artifacts:
     if: github.event_name == 'push' && (github.ref_name == 'main' || startsWith(github.ref_name, 'stable/'))
     name: Deploy Artifacts
-    needs: [integration-tests]
+    needs: [ integration-tests, upgrade ]
```
RomanJRW, line 303: The action warning here looks legit

.github/workflows/optimize-ci.yml
```diff
@@ -301,13 +300,14 @@ jobs:
...
     uses: ./.github/workflows/optimize-deploy-artifacts.yml
     secrets: inherit
 
   # Rerun failed jobs running on self-hosted runners in case of network issues or node preemption
   rerun-failed-jobs:
     needs:
+      - upgrade
```
RomanJRW, line 310: the warning here looks legit too. This should not be here if we end up having this on the unified status checks

RomanJRW (changes requested):

.github/workflows/optimize-ci.yml
RomanJRW: Triggered on backend changes only seems reasonable to me. It could be a changed index file, or even anything implementing the `OptimizeDTO` interface that would break things if no migration has been written

I would be ok with this in either the unified status checks, or on the Optimize CI. It meets the status check criteria and would give us some additional confidence if there is a backend change that erroneously amended the Index DTOs, for example.

That being said, changes outside of Optimize are unlikely to break Optimize in such a way that a test failure here would be helpful. So I am also ok keeping this as part of the Optimize CI if there is resistance when you get a CI review

.github/workflows/ci.yml
PHWaechtler: Sorry - not a fix and not related, just noticed the name for this was a bit unspecific

.github/workflows/optimize-ci.yml
PHWaechtler: Yup overlooked to remove

.github/workflows/optimize-ci.yml
PHWaechtler: Ah, I figured we were also thinking along the lines of having this in the unified to add more checks to renovate. I'll ping for a Ci review on this anyway.

.github/workflows/ci.yml
```diff
@@ -499,6 +499,70 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+      - name: Setup Maven
+        uses: ./.github/actions/setup-maven
+        with:
+          secrets: ${{ toJSON(secrets) }}
+      - name: Login to Harbor registry
+        uses: ./.github/actions/login-registry
+        with:
+          secrets: ${{ toJSON(secrets) }}
```
cmur2, line 517: I'd like to assess if this step is needed: Which of the subsequent steps uses Docker images that need to be downloaded from registry.camunda.cloud?

In https://github.com/camunda/camunda/blob/main/.github/actions/compose/docker-compose.elasticsearch.yml the ElasticSearch image is downloaded from the official ES registry, not from registry.camunda.cloud.

I'm not sure what the other steps in this job do but if none of them needs registry.camunda.cloud you should remove this login step.

cmur2 (commented):
Scope of my review is `ci.yml` only since Helene asked me.

I see that the new `optimize-upgrade-test` job is passing in https://github.com/camunda/camunda/actions/runs/10764699672/job/29848814627?pr=21852 but would like to establish confidence that it is reliable over multiple executions (see the [docs about this](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)). Can you please rerun it a couple of times to ensure that?

We currently have the case of https://github.com/camunda/camunda/issues/21722 where an Optimize-related job initially seemed stable but has very slow executions in ~15% of the cases. Just want to make sure this new job does not have the same behavior.

RomanJRW (approved):
Approving from an Optimize perspective. However, if we do not have confidence and evidence that this can run as part of the unified status checks, then I am also totally fine with this being in the Optimize CI only.

In any case, please either:
- Demonstrate that there is confidence that it can be on the unified CI checks
- Otherwise (or if the above proves flakiness or unreasonable risk of breaks), please move it to the Optimize CI instead

.github/workflows/ci.yml
PHWaechtler: Thanks, looks like I copied this over without needing to - removed now. Will do a bit more testing to see if the IT are too flaky for unified CI.

PHWaechtler:
Thanks @cmur2 for review, I removed the unnecessary step and ran it 20 times which was green throughout so on that front we're good. However, Josh and I discussed this again earlier and were uncertain on whether our current release process could lead to a temporary phase where these IT would fail during the release week. We will keep these IT in our separate optimize IT for now and once we've done one release and can see whether they are stable then we can move them to the unified CI. I will make the required adjustments to this PR tomorrow once I have more time.

PR #21853, opened 2024-09-03
21614 handle error on decimals

panagiotisgts:
## Description

Use strict ObjectMapper with disabled `ACCEPT_FLOAT_AS_INT` for specific DTOs, `UserTaskUpdateRequest` and `UserTaskSearchQueryRequest` in order to disallow decimal numbers in input payload for `priority`.

## Related issues

closes #21614

### DIFF
--- a/zeebe/gateway-rest/pom.xml
+++ b/zeebe/gateway-rest/pom.xml
@@ -169,6 +169,18 @@
     </dependency>
 
     <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-core</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
       <groupId>jakarta.servlet</groupId>
       <artifactId>jakarta.servlet-api</artifactId>
       <scope>provided</scope>
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
@@ -10,6 +10,7 @@ package io.camunda.zeebe.gateway.rest;
 import jakarta.servlet.http.HttpServletRequest;
 import java.net.URI;
 import java.util.Objects;
+import org.springframework.core.annotation.Order;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.HttpStatusCode;
 import org.springframework.http.ProblemDetail;
@@ -22,6 +23,7 @@ import org.springframework.web.context.request.WebRequest;
 import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
 
 @ControllerAdvice(annotations = RestController.class)
+@Order
 public class GlobalControllerExceptionHandler extends ResponseEntityExceptionHandler {
 
   private static final String REQUEST_BODY_MISSING_EXCEPTION_MESSAGE =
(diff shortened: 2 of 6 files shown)


tmetzke (changes requested):
Thanks, @panagiotisgts, for providing this.

I feel that the changes are quite intrusive regarding the REST application overall, introducing ordered exception handlers and a new object mapper.

I would like us to explore less intrusive and more focused solution approaches first, e.g.
- Can the specific field in the OpenAPI spec be defined differently so it only accepts integers?
- Should we rather validate the specific user input in the respective controller manually and create a problem response if it's not an integer?

Those approaches would be more focused on the actual field that needs this, rather than introducing new structures for the whole application.

Let me know what you think. Maybe you have looked into such approaches already but I couldn't see any mention of that in the PR or the issue.

panagiotisgts:
Hey @tmetzke, I have explored other options to achieve this behavior for the specific endpoints and field requirements. Some of them have been discussed [here](https://camunda.slack.com/archives/CSQ2E3BT4/p1725033194459079)

> - Can the specific field in the OpenAPI spec be defined differently so it only accepts integers?
> - Should we rather validate the specific user input in the respective controller manually and create a problem response if it's not an integer?

The ObjectMapper is rounding all decimal inputs when the field is specified as `integer` before reaching the respective Spring Controller. Another approach would be to mark the field as `BigDecimal` and use the `@jakarta.validation.constraint.Digits` annotation in the generator to Validate the input or do a custom validation logic in the Controller itself. This would be confusing in my opinion, since the input is never a BigDecimal actually but an Integer and this will be shown in the API spec also.

I have also tried adding custom validation annotations to the OpenAPI codegen tool, similar to [this](https://www.baeldung.com/java-swagger-custom-validation) approach, but again, Jackson transforms it into an Integer trimming the decimals by the time it reaches the scope of the validator. Retrieving the original value from the ValidationContext is tedious and requires even more 'bloat' to be added to the Context.

> I feel that the changes are quite intrusive regarding the REST application overall, introducing ordered exception handlers and a new object mapper.

The ObjectMapper is a copy of the one already injected in the context having the same options except the disable feature to achieve the desired outcome, and applied only to the specific Models we want this behavior to be present. This was the only option I found to be applicable without having to add a more "complex" solution. I am open to suggestions :) .

Other options discussed:
- Create a custom Deserializer for the specific Models -> Will need maintenance on model changes and would not prefer to have custom SERDEs overall
- Use a WebFilter to unwrap the HTTPRequest and then wrap it again to proceed through the chain -> Has a high performance cost for such a simple functionality
- Move the `@ExceptionHandler` over the respective Controller class, would require duplication on any other controller that might have such a scenario

tmetzke:
Thanks, @panagiotisgts, for the detailed answer, that helps a lot 👍
Since we have a lot of `integer` attributes and parameters in the OpenAPI spec, this needs consolidated handling to make the API behave consistently. Currently, it behaves consistently for all such fields. It silently accepts and transforms the values.

Is this ideal? Probably not. However, introducing this handling now for only a handful of models and attributes makes this a bit worse, to be honest.

I believe we should treat all `integer` fields the same regarding decimal input handling - either silently transform them or validate them and return an error response. Whether that happens by generating `BigDecimal` for `integer` parameters or by not accepting floats for integers in the REST API's object mapper in general, can be discussed.

I would generally prefer the object mapper way that you started here, I think. We can then maybe integrate this into the global exception handler that we already have by default (not entirely sure about that part). There might be caveats that I am currently missing here, but it appears like the more straightforward solution to me.

panagiotisgts:
@tmetzke Since this is no blocker in any way and the current implementation is aligned with the way all `Integers` are treated in the API I will leave this PR open for followup discussions for further alignment as a PoC and update the issue.

tmetzke:
Thanks, @panagiotisgts, I agree, let's use this as a discussion and playground to test the idea. I would really like to have a general integer validation in the REST API and your code seems like the right way forward if we can generalize it 👍

panagiotisgts:
Closing this one since it's outdated and another approach has been taken.

Issue #21843, opened 2024-09-02
Create tasklist-listview indice and importers


nathansandi:
**Description**

In order to procceed with the join relations, we wanna have an indice that contains

- Process as a parent
- Task as a child from Process
- taskVariable as a child from Task
- processVariable as a child from Process

Subprocesses does not need to persist, but their variables yes.
- TaskVariable associated with the task inside subprocess
- Subprocess variables with the parent node, but not used in futher search for user task because of limitations mentioned already

The Subprocess tasks will be child from a Process, as well as the variables, once the task is completed

To remind the limitations
- It is not possible retrieve tasks present in subprocess by subprocess variables, but it is possible by input (This limitation is also over C7, as we can have a unlimited nested subprocess, and it could cause follow loops going over the subprocesses, this also does not work on V1). But still possible retrieve subprocess tasks by the processVariable or inputVariable on the task.
- We will persist the subprocess variables (even if we wont use in the search), once we can reuse this indice to retrieve variables as well)
- When a subprocess is completed, subprocess variables will become a process variable. So it will be possible research the tasks by this variable
- When a task inside a subprocess is completed the current behaviour is the output of task became also a process variable (Current behaviour on tasklist) 

Acceptance Criteria
- Indices tasklist-listview created
- Importers working accordindly to the POC
- entity and parsers working on the same way as User Task
- Only User Task should be imported - Variables can be imported all, but when persist job workers, the variables from job workers can be deleted

PR #21868, opened 2024-09-03
feat: add tasklist-listview template and follow up entity template classes

### LINKED ISSUE #21845: [Root Task] Search Task by Variable
```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/issues/21843
- [ ] https://github.com/camunda/camunda/issues/21846
- [ ] https://github.com/camunda/camunda/issues/21848
- [ ] https://github.com/camunda/camunda/issues/21849
```

nathansandi:
## Description

- Create tasklist-list-view template for ES/OS
- Create TasklistListView entity
- Create TasklistListViewTemplate Class

This contains
- Simple process flow node to set as parent
- Variable data
- User Task Data (As it is made for tasklist-task) but considering only V2 Necessary data (the same field names were used as tasklist-task, considering hybrid approach in 8.7 (to use both `tasklist-task` and `tasklist-listview` to retrieve task when variables are included in search) in the future. 

The importers will be created in a follow up Pull Request 

**Validation:**

image

## Related issues

closes #https://github.com/camunda/camunda/issues/21843


houssain-barouni (commented):
Hi @nathansandi,

❓ I am wondering if we can split `TasklistListViewEntity` into 3 classes (Task, Variable and Process) like it is done in [Operate](https://github.com/camunda/camunda/tree/main/operate/schema/src/main/java/io/camunda/operate/entities/listview) and create a `ListViewJoinRelation` class to define the relations between the classes. 
This should improve the separation of concerns and can make the code easier to manage and extend.

nathansandi:
> Hi @nathansandi,
> 
> ❓ I am wondering if we can split `TasklistListViewEntity` into 3 classes (Task, Variable and Process) like it is done in [Operate](https://github.com/camunda/camunda/tree/main/operate/schema/src/main/java/io/camunda/operate/entities/listview) and create a `ListViewJoinRelation` class to define the relations between the classes. This should improve the separation of concerns and can make the code easier to manage and extend.

Hello @houssain-barouni 
I have studied this possibility, but it would create confusion with the existent classes, and the fact we have V1 data from those classes, I believe when we do not have v1 inconsistencies and differences, this would be possible, because we wont have Task data with v1 fields (like formEmbedded).

If the fields are exactly the same, this would be possible, but it is not the case. We can split this in the further, once we have consistency over the classes. For now, to avoid this complexity over edge cases, and increase the complexity, I would prefer to keep as is (it is not a problem revisit this in the future)

houssain-barouni:
> Hello @houssain-barouni, thanks for bringing this up.
> 
> I have considered this approach before, but it might introduce confusion given the existing classes and the differences in fields between versions. Currently, we have some fields in V1 that won’t be present in the ListView for V2 (e.g., formEmbedded, formVersion). Splitting the entity now could lead to having two TaskEntity classes, which could increase complexity, especially with edge cases. Also, another point the Entities are used over SchemaValidators, and the fields must be the same (not more because the dynamic is set as strict)
> 
> However, once we achieve consistency across these classes and no longer support Job Worker, this refactoring could become more feasible and beneficial. For now, I believe it’s best to keep the current structure to avoid unnecessary complexity. That said, I’m open to revisiting this idea in the future when the conditions are more favorable.

What about the V2 stack? Are you planning to create new entity classes under `camunda-service` module?

nathansandi:
> > Hello @houssain-barouni, thanks for bringing this up.
> > I have considered this approach before, but it might introduce confusion given the existing classes and the differences in fields between versions. Currently, we have some fields in V1 that won’t be present in the ListView for V2 (e.g., formEmbedded, formVersion). Splitting the entity now could lead to having two TaskEntity classes, which could increase complexity, especially with edge cases. Also, another point the Entities are used over SchemaValidators, and the fields must be the same (not more because the dynamic is set as strict)
> > However, once we achieve consistency across these classes and no longer support Job Worker, this refactoring could become more feasible and beneficial. For now, I believe it’s best to keep the current structure to avoid unnecessary complexity. That said, I’m open to revisiting this idea in the future when the conditions are more favorable.
> 
> What about the V2 stack? Are you planning to create new entity classes under `camunda-service` module?

No, I am going to reuse the existent ones already created there (UserTaskEntity) in V2. Only the query will be updated (it will be clear when the PR is there)

houssain-barouni:
> Hey @houssain-barouni No, I am going to reuse the existent ones already created there (UserTaskEntity) in V2, which is already with the fields in V2. Only the query will be updated (it will be clear when the PR is there).

Will you also reuse the existing VariableEntity ? It has different naming for the fields (example `varValue` vs `value`)

nathansandi:
> > Hey @houssain-barouni No, I am going to reuse the existent ones already created there (UserTaskEntity) in V2, which is already with the fields in V2. Only the query will be updated (it will be clear when the PR is there).
> 
> Will you also reuse the existing VariableEntity ? It has different naming for the fields (example `varValue` vs `value`)

I have no answer for this question now, once it is not in the scope of the Search Task by Variable (We wont retrieve variables on user-task). So I wont use `VariableEntity` for this query. 
But I understand your point here, and this decision can be made on the development of Search Variables.

nathansandi:
Hello @houssain-barouni  I have just pushed the changes as we discussed on Slack

houssain-barouni (commented):
Thanks @nathansandi for the sync explanation of the unique identifier of the new entities
as discussed, we need to add the `partitionId` and `tenantId to ProcessInstance and Variable entities

houssain-barouni (approved):

Issue #21480, opened 2024-08-22
Cluster API has endpoints for adding partitions


lenaschoenburg:
Extend the Cluster API available at `actuator/cluster` with new endpoints to add partitions to a cluster.
When making such a request, an appropriate `ScaleRequest` is sent to the `ClusterConfigurationManagementApi`.

## Acceptance Criteria

- [ ] The OpenAPI spec is extended
- [ ] The `ClusterActuator` is extended
- [ ] The `ClusterActuatorAssert` is extended

PR #21869, opened 2024-09-03
docs: add spec for api to change cluster configuration

deepthidevaki:
## Description

This PR adds the Open API spec for reconiguring the cluster. The api allows to add brokers, remove brokers, change partitions count and replication factor. This API will replace the existing broker scaling API `POST actuator/cluster/brokers/`. 

The PR do not add the implementation. 

## Related issues

related #21480

### DIFF
--- a/dist/src/main/resources/api/cluster/cluster-api.yaml
+++ b/dist/src/main/resources/api/cluster/cluster-api.yaml
@@ -20,7 +20,7 @@ paths:
     post:
       summary: Add a broker to the cluster
       description: Add a broker with the given brokerId to the cluster. The broker must be running
-                   to complete the operation.
+        to complete the operation.
       parameters:
         - $ref: '#/components/parameters/BrokerId'
         - $ref: '#/components/parameters/DryRunParameter'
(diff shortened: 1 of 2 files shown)


dist/src/main/resources/api/cluster/cluster-api.yaml
```diff
@@ -103,6 +103,34 @@ paths:
         '504':
           $ref: 'components.yaml#/responses/TimeoutError'
 
+    patch:
+      summary: Reconfigure cluster
+      description: Reconfigure cluster by adding or removing brokers, adding more partitions, or
+                   changing the replicationFactor. Note that adding partitions is an experimental
+                   feature and not fully supported.
```
lenaschoenburg, line 110: :upside_down_face: This is deeply indented. Above you have changed another description to be less indented. Not sure how this will look in docs etc.

lenaschoenburg (approved):
Thanks!

dist/src/main/resources/api/cluster/cluster-api.yaml
deepthidevaki: Not sure though. It was auto-formatted. :thinking:

PR #21871, opened 2024-09-03
Initial schema manager resources

### LINKED ISSUE #21528: Ability to create all initial schemas for schema manager

EuroLew:
## Description

It creates the skeleton for a schema manager which will instantiate all indices and index templates provided at startup (not necessarily in the constructor it has not been fully decided yet). In addition it allows the configuration of the shards and replicas for the indices and templates.

## Related issues

closes https://github.com/camunda/camunda/issues/21528


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/IndexTemplateDescriptor.java
```diff
@@ -16,4 +16,14 @@ public interface IndexTemplateDescriptor extends IndexDescriptor {
...
+
+  /**
+   * Maps to the create query parameter, if true the corresponding request cannot replace or update
+   * an existing component template, defaults to false
+   *
+   * @return The component template request is create only
+   */
+  default Boolean create() {
+    return false;
+  }
```
mustafadagher, line 28: I'm still unsure about this. If my understanding is correct, it's not up to the `IndexTemplate` to signal if it could be overwritten or not, but it's more about if we are creating or updating schema.

I think this doesn't belong here, and it should be passed as a parameter in the call stack to `PutIndexTemplate`

mustafadagher (approved):
Thanks! 🚀 
I left only one comment below

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/descriptors/IndexTemplateDescriptor.java
EuroLew: Moved the create option to be passed in when creating and sending PutIndexTemplateRequest

PR #21874, opened 2024-09-03
Updateable indices and index templates

### LINKED ISSUE #21762: Updating of index and index templates can be handled by schema manager
**Description**

When a schema manager is started up if the mappings described are different it should be able to update both the index and index template with the new mappings. Currently no validation will be done if the update is valid as that will be handled later.

Success Criteria
- [ ] Index template mappings can be updated by providing the new mappings block the template will have.
- [ ] Index mappings can be updated by providing new properties to append.

EuroLew:
## Description

When a schema manager is started up if the mappings described are different it should be able to update both the index and index template with the new mappings. Currently no validation will be done if the update is valid as that will be handled later.

Success Criteria

- [x] Index template mappings can be updated by providing the new mappings block the template will have.
- [x] Index mappings can be updated by providing new properties to append.

## Related issues

closes #21762


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SearchEngineClient.java
```diff
@@ -20,20 +21,9 @@ void createIndexTemplate(
       final Boolean create);
```
deepthidevaki, line 21: :wrench: You can use primitive `boolean` instead.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
```diff
@@ -68,35 +68,28 @@ public void createIndexTemplate(
...
+      final IndexDescriptor indexDescriptor, final Set<IndexMappingProperty> newProperties) {
+    final PutMappingRequest request = putMappingRequest(indexDescriptor, newProperties);
 
     try {
       client.indices().putMapping(request);
       LOG.debug("Mapping in [{}] was successfully updated", indexDescriptor.getIndexName());
     } catch (final IOException e) {
       final var errMsg =
-          String.format("Mapping in [{}] was NOT updated", indexDescriptor.getIndexName());
+          String.format("Mapping in [%s] was NOT updated", indexDescriptor.getIndexName());
```
deepthidevaki, line 80: :+1:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SearchEngineClient.java
```diff
@@ -20,20 +21,9 @@ void createIndexTemplate(
...
-   * }
-   * }</pre>
-   *
    * @param indexDescriptor Representing index of which to update the mappings
-   * @param propertiesJson New properties to be appended to the index
+   * @param newProperties New properties to be appended to the index
    */
-  void putMapping(final IndexDescriptor indexDescriptor, final String propertiesJson);
+  void putMapping(
+      final IndexDescriptor indexDescriptor, final Set<IndexMappingProperty> newProperties);
```
deepthidevaki, line 28: :wrench: Why don't we just use `Map<String, Object>` instead of `Set<IndexMappingProperty`. To serialize it to json, you are building a map anyway.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/ElasticsearchSchemaManagerIT.java
```diff
@@ -115,7 +122,72 @@ void shouldUseIndexSpecificSettingsIfSpecified() throws IOException {
...
         elsClient.indices().get(req -> req.index("full_name")).get("full_name");
 
-    Assertions.assertThat(retrievedIndex.settings().index().numberOfReplicas()).isEqualTo("5");
-    Assertions.assertThat(retrievedIndex.settings().index().numberOfShards()).isEqualTo("5");
+    assertThat(retrievedIndex.settings().index().numberOfReplicas()).isEqualTo("5");
+    assertThat(retrievedIndex.settings().index().numberOfShards()).isEqualTo("5");
+  }
+
+  @Test
+  void shouldOverwriteIndexTemplateIfMappingsFileChanged() throws IOException {
```
deepthidevaki, line 130: :+1:

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/ElasticsearchSchemaManagerIT.java
```diff
@@ -115,7 +122,72 @@ void shouldUseIndexSpecificSettingsIfSpecified() throws IOException {
...
+            .getIndexTemplate(req -> req.name("template_name"))
+            .indexTemplates()
+            .getFirst();
+
+    assertThat(template.indexTemplate().template().mappings().properties().get("foo").isText())
+        .isTrue();
+  }
+
+  @Test
+  void shouldAppendToIndexMappingsWithNewProperties() throws IOException {
```
deepthidevaki, line 166: :+1:

zeebe/exporters/camunda-exporter/src/test/resources/mappings-deleted-property.json
```diff
@@ -0,0 +1,13 @@
+{
```
deepthidevaki, line 1: :question: Where is this file used?

zeebe/exporters/camunda-exporter/src/test/resources/mappings-deleted-different-property.json
```diff
@@ -0,0 +1,13 @@
+{
```
deepthidevaki, line 1: :question: Where is this file used?

deepthidevaki (commented):
:rocket: Great.

I have one main comment regarding the interface for `updateSchema`. Please share your thoughts. I will approve the PR after that.

zeebe/exporters/camunda-exporter/src/test/resources/mappings-deleted-property.json
EuroLew: oops I'll remove

zeebe/exporters/camunda-exporter/src/test/resources/mappings-deleted-different-property.json
EuroLew: same accidently put in this branch

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SearchEngineClient.java
EuroLew: I was trying to keep it as compatible with the current operate code as possible to save time making things work. The `IndexSchemaValidator` in the `validateIndexMappings()` function returns for every index descriptor a `Set<IndexMappingProperty>` detailing any additional properties to add. Doing it this way requires no marshalling of different types although I agree with your point regarding the serialisation. 

In conclusion the way the difference is calculated as per `IndexMappingDifference` will return a `Set<IndexMappingProperty>` at some point in time this will need to be serialised to json or converted to an els/opensearch data structure to be passed to the request, and in the request will hide implementation details to users of the client.

deepthidevaki: :+1: Since this do not affect the actual Operate code, I think it is ok to change it. But let's leave it like this and consider refactoring later when implementing the validation.

deepthidevaki (approved):
Sorry. Forgot to approve. :rocket:

PR #21879, opened 2024-09-03
fix: skip initialization of partition ring until topology is initialized

lenaschoenburg:
This fixes a bug where the partition ring was incorrectly initialized from the topology. Checking for `null` is not enough, we must also account for the situation where cluster topology is available but not yet initialized with cluster size and partition count.

Without this fix, we'd initialize with partition count 0 and then never update it again, eventually causing `ArithmeticException: / by zero` when trying to look up a partition that wasn't known before.

Observed here: https://console.cloud.google.com/errors/detail/CIONl6PFkr2UsQE;service=zeebe;version=ck-platform-benchmark;filter=%5B%5D;time=P7D;locations=global?project=zeebe-io

### DIFF
--- a/zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/RoundRobinDispatchStrategy.java
+++ b/zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/RoundRobinDispatchStrategy.java
@@ -31,7 +31,8 @@ public final class RoundRobinDispatchStrategy implements RequestDispatchStrategy
   public int determinePartition(final BrokerTopologyManager topologyManager) {
     final BrokerClusterState topology = topologyManager.getTopology();
 
-    if (topology == null) {
+    if (topology == null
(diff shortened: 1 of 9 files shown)


zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/impl/RoundRobinDispatchStrategy.java
```diff
@@ -31,7 +31,8 @@ public final class RoundRobinDispatchStrategy implements RequestDispatchStrategy
   public int determinePartition(final BrokerTopologyManager topologyManager) {
     final BrokerClusterState topology = topologyManager.getTopology();
 
-    if (topology == null) {
+    if (topology == null
+        || topology.getClusterSize() == BrokerClusterStateImpl.UNINITIALIZED_CLUSTER_SIZE) {
```
deepthidevaki, line 35: :wrench: Can be moved to a `topology.isInitialized()` method.

deepthidevaki (approved):
:rocket:

PR #21893, opened 2024-09-03
feat: C8 API - Implement Process instance search endpoint

### LINKED ISSUE #21836: [C8 API] Implement Process instance search - client part
## Description

Client implementation for `/v2/process-instances/search` endpoint. Part of https://github.com/camunda/camunda/issues/20660

ghost:
feat: C8 API - Implement Process instance search endpoint

- implemented client part
- added unit tests

Closes https://github.com/camunda/camunda/issues/21836

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java
@@ -30,8 +30,8 @@ public class OperationImpl implements Operation {
   public OperationImpl(final OperationItem item) {
     id = item.getId();
     batchOperationId = item.getBatchOperationId();
-    type = item.getType().toString();
-    state = item.getState().toString();
+    type = (item.getType() == null) ? null : item.getType().toString();
+    state = (item.getState() == null) ? null : item.getState().toString();
     errorMessage = item.getErrorMessage();
     completedDate = item.getCompletedDate();
   }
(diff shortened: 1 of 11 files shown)


clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java
```diff
@@ -30,8 +30,8 @@ public class OperationImpl implements Operation {
   public OperationImpl(final OperationItem item) {
     id = item.getId();
     batchOperationId = item.getBatchOperationId();
-    type = item.getType().toString();
-    state = item.getState().toString();
+    type = (item.getType() == null) ? null : item.getType().toString();
```
houssain-barouni, line 33: We should use enums to be consistent with the API spec

nathansandi, line 33: 💭 minor/suggestion: 
We can use here: 
`Optional.ofNullable(item.getType()).map(Enum::toString).orElse(null)` 

for setting `type` and `state`.

nathansandi (approved):
Great work here 💯 🥇

houssain-barouni (commented):
One comment regarding the use of enums: it relates to the discussion we had [here](https://camunda.slack.com/archives/C06UKS51QV9/p1725288373831619)

clients/java/src/main/java/io/camunda/zeebe/client/impl/search/response/OperationImpl.java
mihail-ca: I did not add enums anymore, because in the model we return strings, so there is not need to convert from enum to enum, just from enum to string.

Issue #21562, opened 2024-08-26
Centralize implementation of search plugin loading


npepinpe:
**Description**

To simplify maintenance and reduce risks of bugs, we will replace the current interceptor plugin loading code (which is sextuplicated) and centralize it in a new `io.camunda.search.connect.plugin` package in the `camunda-search-connect` module.

This will then be called by the Operate ES/OS connectors, the Tasklist ES/OS connectors, and the Zeebe ES/OS exporters. The idea is that eventually it will _only_ be called by the shared common connector code (i.e. what's in the `camunda-search-connect` module), and indirectly used in the applications above.

The following is a proposal of what it could look like, feel free to challenge it.

I would propose following the `*Repository` pattern, as we did with the gateway interceptors/filters, and the Zeebe exporters. This is because in most cases, the plugins only need to be loaded once, and should live for the complete application lifetime. Only in the case of exporters are they loaded on a per-exporter basis (note: this further highlights the exporter pattern is not the right one for our search storage, but that's out of scope here).

So, for example, you would have the following repository:

<details><summary>PluginRepository.java</summary>

```java
/** Loads and holds references to the configured plugins. */
public final class PluginRepository implements AutoCloseable {
  private static final Logger LOGGER = LoggerFactory.getLogger(PluginRepository.class);

  private final ExternalJarRepository jarRepository;
  private final SequencedMap<String, Class<? extends DatabaseCustomHeaderSupplier>> plugins;
  private final Path basePath;

  public PluginRepository() {
    this(
        new LinkedHashMap<>(),
        new ExternalJarRepository(),
        Paths.get(Optional.ofNullable(System.getProperty("basedir")).orElse(".")));
  }

  PluginRepository(
      final SequencedMap<String, Class<? extends DatabaseCustomHeaderSupplier>> plugins,
      final ExternalJarRepository jarRepository,
      final Path basePath) {
    this.plugins = plugins;
    this.jarRepository = jarRepository;
    this.basePath = basePath;
  }

  public SequencedMap<String, Class<? extends DatabaseCustomHeaderSupplier>> getPlugins() {
    return Collections.unmodifiableSequencedMap(plugins);
  }

  public boolean isEmpty() {
    return plugins.isEmpty();
  }

  public CompatHttpRequestInterceptor asRequestInterceptor() {
    return PluginRepositoryInterceptor.ofRepository(this);
  }

  public PluginRepository load(final SequencedCollection<PluginConfiguration> configs) {
[...]

PR #21904, opened 2024-09-04
feat: centralize search plugin loading

igpetrov:
## Description

feat: centralize search plugin loading

The implementation will be extended with more improvements and actual wiring with ES/OS clients. This will be done in another PRs to keep this one small and easier to review.

## Checklist

- [ ] for CI changes:

## Related issues

Related to: https://github.com/camunda/camunda/issues/21562

### DIFF
--- a/search/search-client-connect/src/main/java/io/camunda/search/connect/plugin/PluginRepository.java
+++ b/search/search-client-connect/src/main/java/io/camunda/search/connect/plugin/PluginRepository.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.search.connect.plugin;
+
+import io.camunda.plugin.search.header.DatabaseCustomHeaderSupplier;
+import io.camunda.zeebe.util.ReflectUtil;
+import io.camunda.zeebe.util.VisibleForTesting;
+import io.camunda.zeebe.util.jar.ExternalJarLoadException;
+import io.camunda.zeebe.util.jar.ExternalJarRepository;
+import io.camunda.zeebe.util.jar.ThreadContextUtil;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Optional;
+import java.util.SequencedCollection;
+import java.util.SequencedMap;
+import java.util.stream.Stream;
+import org.agrona.LangUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Loads and holds references to the configured plugins. */
+public final class PluginRepository implements AutoCloseable {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PluginRepository.class);
+
+  private final ExternalJarRepository jarRepository;
+  private final LinkedHashMap<String, Class<? extends DatabaseCustomHeaderSupplier>> plugins;
+  private final Path basePath;
+
+  public PluginRepository() {
+    this(
+        new LinkedHashMap<>(),
+        new ExternalJarRepository(),
+        Paths.get(Optional.ofNullable(System.getProperty("basedir")).orElse(".")));
+  }
+
+  PluginRepository(
+      final LinkedHashMap<String, Class<? extends DatabaseCustomHeaderSupplier>> plugins,
+      final ExternalJarRepository jarRepository,
+      final Path basePath) {
+    this.plugins = plugins;
+    this.jarRepository = jarRepository;
+    this.basePath = basePath;
+  }
+
+  @VisibleForTesting
+  SequencedMap<String, Class<? extends DatabaseCustomHeaderSupplier>> getPlugins() {
+    return Collections.unmodifiableSequencedMap(plugins);
+  }
+
+  public boolean isEmpty() {
+    return plugins.isEmpty();
+  }
+
+  public CompatHttpRequestInterceptor asRequestInterceptor() {
+    return PluginRepositoryInterceptor.ofRepository(this);
+  }
+
+  public PluginRepository load(final SequencedCollection<PluginConfiguration> configs) {
+    if (configs == null || configs.isEmpty()) {
+      return this;
+    }
+
+    configs.forEach(
+        config -> {
+          try {
+            load(config);
+          } catch (final Exception e) {
+            LOGGER.debug("Failed to load interceptor {} with config {}", config.id(), config);
+            LangUtil.rethrowUnchecked(e);
+          }
+        });
+
+    return this;
+  }
+
+  @Override
+  public void close() throws Exception {
+    try {
+      jarRepository.close();
+    } catch (final Exception e) {
+      LOGGER.warn(
+          "Failed to close external JAR repository; may result in dangling file descriptors", e);
+    }
+  }
+
+  private void load(final PluginConfiguration config) throws ExternalJarLoadException {
+    final ClassLoader classLoader;
+    final Class<? extends DatabaseCustomHeaderSupplier> interceptorClass;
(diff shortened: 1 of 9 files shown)


search/search-client-connect/src/main/java/io/camunda/search/connect/plugin/PluginRepository.java
```diff
@@ -0,0 +1,142 @@
...
+
+  public PluginRepository() {
+    this(
+        new LinkedHashMap<>(),
+        new ExternalJarRepository(),
+        Paths.get(Optional.ofNullable(System.getProperty("basedir")).orElse(".")));
+  }
+
+  PluginRepository(
+      final SequencedMap<String, Class<? extends DatabaseCustomHeaderSupplier>> plugins,
```
npepinpe, line 43: :x: I think I made a mistake, and this should be the concrete type, `LinkedHashMap`. Here and for the field. The reason for this is `LinkedHashMap` defines the ordering of the map as the order in which keys were inserted, which matches with the order in which the plugins will be configured. Using the interface type, `SequencedMap`, we could technically pass a different implementation with a different ordering.

So we should ensure the maps used here are _always_ `LinkedHashMap` (unless there's an interface type I'm unaware of which guarantees the above semantics).

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+  @Test
+  void shouldJustReturnSelfWhenLoadIfConfigsWereEmpty() {
+    final var repo = new PluginRepository();
+
+    final var repoRef = repo.load(new ArrayList<>());
+
+    assertThat(repoRef.getPlugins()).isEmpty();
+    assertThat(repo).isSameAs(repoRef);
+  }
```
npepinpe, line 64: :wrench: imo those tests are not useful, the semantics that it returns itself as nice, but even if you break them, would not cause errors.

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
npepinpe, line 1: :x: Please avoid using mocks for easily stubbed or constructed objects, e.g `Path`

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
```diff
@@ -0,0 +1,179 @@
...
+    final var basePath = Mockito.mock(Path.class);
+    final var repo = new PluginRepository(initialPlugins, jarRepository, basePath);
+
+    final var plugins = repo.getPlugins();
+
+    assertThat(plugins).hasSize(1);
+  }
+
+  @Test
+  void shouldLoadExternalPlugin() throws ExternalJarLoadException, ClassNotFoundException {
```
npepinpe, line 83: :x: Please use actual objects and not so many mocks.

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
npepinpe, line 1: :x: Generally the tests and logic are good, but let's use concrete objects instead of mocks. It simplifies debugging and diagnosing issues, and it means the test is much closer to the production code, which we should always try to have whenever possible (unless major cons).

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
```diff
@@ -0,0 +1,179 @@
...
+    final var jarRepository = Mockito.mock(ExternalJarRepository.class);
+    final var basePath = Mockito.mock(Path.class);
+    try (final var repo = new PluginRepository(new LinkedHashMap<>(), jarRepository, basePath)) {
+      repo.getPlugins(); // no-op
+    }
+
+    verify(jarRepository).close();
+  }
+
+  public static final class DatabaseCustomHeaderSupplierTestImpl
```
npepinpe, line 171: :wrench: Prefix with Test instead of suffix, it's a convention we tend to use to indicate clearly that this is a test impl and not production (helps, for example, when looking up indexed classes). Additionally, the `Test` suffix tends to be reserved for test classes.

npepinpe (dismissed):
:rocket: 

As I'm on FTO, @megglos will do/support the second round of review. The main issue I see an overuse of mocks for objects which are easily created for the test, or easily stubbed. We should avoid mocks unless necessary, as they add another layer of abstraction in your test - makes it harder to ensure we're running as close to production as possible, makes it harder to debug, and harder to diagnose problems.

npepinpe:
@megglos - sorry, I requested your review, but I did the first round. Igor will request from you for the second round.

search/search-client-connect/src/main/java/io/camunda/search/connect/plugin/PluginRepository.java
igpetrov: Good point, as long as nobody decides they want to implement their own version of `LinkedHashMap` 🙃 with whatever order.
I'll change that in the next revision.

search/search-client-connect/src/main/java/io/camunda/search/connect/plugin/PluginRepository.java
```diff
@@ -0,0 +1,142 @@
...
+      final ExternalJarRepository jarRepository,
+      final Path basePath) {
+    this.plugins = plugins;
+    this.jarRepository = jarRepository;
+    this.basePath = basePath;
+  }
+
+  public SequencedMap<String, Class<? extends DatabaseCustomHeaderSupplier>> getPlugins() {
```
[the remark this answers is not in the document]
igpetrov: Testing/debugging only. Will add `@VisibleForTesting`.

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryInterceptorTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
+
+import io.camunda.plugin.search.header.CustomHeader;
+import io.camunda.plugin.search.header.DatabaseCustomHeaderSupplier;
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+
+public class PluginRepositoryInterceptorTest {
```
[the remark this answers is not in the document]
igpetrov: I can do that. I usually prefer unit-tests to be as isolated as possible without involving other real objects but see no problem following your suggestion.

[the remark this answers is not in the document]
igpetrov: Done

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
igpetrov: Replaced most of the mocks

igpetrov: Replaced most of the mocks with concrete objects

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
igpetrov: Removed

search/search-client-connect/src/test/java/io/camunda/search/connect/plugin/PluginRepositoryTest.java
igpetrov: Done + moved in separate class

igpetrov:
Hi @megglos

Ready for the second round of review. @npepinpe already left comments, and, while I addressed them all, I am not resolving those for visibility.

Thank you!

megglos (approved):
LGTM!

PR #21910, opened 2024-09-04
Restructure authorization rest api

### LINKED ISSUE #21792: Restructure Authorization REST API
<h2>Description</h2>

The current REST API that's implemented doesn't match the design that was made.
Adding/removing permissions from Authorizations will work using a `PATCH` request. The full design can be found in the [kickoff document](https://docs.google.com/document/d/1CMaeHXh-sDfab-bc2xohLLd8cTTOlOBqbg79bQi8dMs/edit#heading=h.rv9tapyimlq2).

remcowesterhoud:
## Description

This restructures the Authorization endpoint to match what was designed during the kickoff. It's quite a few changes. For easy review I've tried to split it into clear commits.

## Checklist

- [ ] for CI changes:

## Related issues

closes #21792

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -804,25 +804,35 @@ paths:
           description: >
             The decision requirements search query failed.
             More details are provided in the response body.
-  /authorizations:
-    post:
+
+  /authorizations/{ownerKey}:
+    patch:
       tags:
         - Authorization
-      summary: Create an authorization
-      operationId: createAuthorization
+      summary: Patch an authorization
+      description: Add or remove permissions to an authorization
+      operationId: patchAuthorization
+      parameters:
+        - name: ownerKey
+          in: path
+          required: true
+          description: The key of the owner of the authorization.
+          schema:
+            type: integer
+            format: int64
       requestBody:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/AuthorizationAssignRequest"
+              $ref: "#/components/schemas/AuthorizationPatchRequest"
         required: true
       responses:
-        "204":
+        "202":
           description: |
-            The Authorization was created successfully.
+            The Authorization was patched successfully.
         "400":
           description: |
-            The Authorization could not be created.
+            The Authorization could not be patched.
             More details are provided in the response body.
           content:
             application/problem+json:
@@ -830,7 +840,7 @@ paths:
                 $ref: "#/components/schemas/ProblemDetail"
         "401":
           description: |
-            The request to create an authorization was unauthorized.
+            The request to patch the authorization was unauthorized.
             More details are provided in the response body.
           content:
             application/problem+json:
@@ -838,18 +848,23 @@ paths:
                 $ref: "#/components/schemas/ProblemDetail"
         "403":
           description: |
-            The request to create an authorization was denied.
+            The request to patch an authorization was denied.
             More details are provided in the response body.
           content:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+        "409":
+          description: |
+            The request to add or remove permissions to an authorization was in conflict.
+            More details are provided in the response body.
         "500":
           description: An internal error occurred while processing the request.
           content:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+
   /users:
     post:
       tags:
@@ -1697,26 +1712,45 @@ components:
           format: int64
         tenantId:
           type: string
-    AuthorizationAssignRequest:
+    AuthorizationPatchRequest:
       type: object
       properties:
-        ownerKey:
-          type: integer
-          format: int64
-        ownerType:
+        action:
+          description: Indicates if permissions should be added or removed.
           type: string
           enum:
+            - ADD
+            - REMOVE
+        resourceType:
+          description: The type of resource to add/remove perissions to/from.
+          enum:
+            - AUTHORIZATION
+            - MESSAGE
+            - JOB
+            - APPLICATION
+            - TENANT
+            - DEPLOYMENT
+            - PROCESS_DEFINITION
+            - USER_TASK
+            - DECISION_REQUIREMENTS_DEFINITION
+            - DECISION_DEFINITION
+            - USER_GROUP
             - USER
             - ROLE
-            - GROUP
-        resourceKey:
-          type: string
-        resourceType:
-          type: string
         permissions:
           type: array
           items:
-            type: string
+            properties:
+              permissionType:
+                enum:
+                  - CREATE
+                  - READ
+                  - UPDATE
+                  - DELETE
+              resourceIds:
+                type: array
+                items:
+                  type: string
     UserWithPasswordRequest:
       type: "object"
       properties:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -810,7 +810,7 @@ paths:
       tags:
         - Authorization
       summary: Patch an authorization
-      description: Add or remove permissions to an authorization
+      description: Manage the permissions assigned to the authorization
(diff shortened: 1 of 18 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1706,26 +1721,45 @@ components:
...
+        action:
+          description: Indicates if permissions should be added or removed.
           type: string
           enum:
+            - ADD
+            - REMOVE
+        resourceType:
+          description: The type of resource to add/remove perissions to/from.
+          enum:
+            - AUTHORIZATION
```
Ben-Sheppard, line 1727: ❓ I like the addition of this as an enum, its a known set of values from us so its suited to an enum, my question is if I think about how authorizations are applied at a processor level, does this match with what we would use (ValueType) to determine the authorizations to fetch?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1706,26 +1721,45 @@ components:
...
           type: array
           items:
-            type: string
+            properties:
+              permissionType:
+                enum:
+                  - CREATE
+                  - READ
+                  - UPDATE
+                  - DELETE
```
Ben-Sheppard, line 1749: ❓ As per the [migration doc](https://docs.google.com/document/d/1yShs_FRhrFv8NmyO94bElZk9UPuiMYGfbKUhn5vxkDE/edit#heading=h.jyh8xap11g2r) should we have a wildcard here?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1706,26 +1721,45 @@ components:
...
+              resourceIds:
```
Ben-Sheppard, line 1750: ❓ This is something that we discussed in the kick off and I'm still not quite sure about how we phrase this, is it Ids, are they keys? Should we have a term that is different to either of them? I'm not sure first had of the usage of IDs and Keys in the system so it may be fine that ID is used here.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -194,11 +198,27 @@ public static Either<ProblemDetail, UpdateJobRequest> toJobUpdateRequest(
...
+              .forEach(
+                  permission -> {
+                    permissions.put(
+                        PermissionType.valueOf(permission.getPermissionType().name()),
+                        permission.getResourceIds());
+                  });
+
+          return new PatchAuthorizationRequest(
+              ownerKey,
+              PermissionAction.valueOf(authorizationPatchRequest.getAction().name()),
```
Ben-Sheppard, line 218: 🔧 I'm not fond of this naming here, PermissionAction implies to me that its something to do direction with the permission, really this is the action of the request. I think this could be generated and therefore out of our hands but if not I'd really prefer something like `AuthorizationRequestAction` so its clear that the action relates to the request as a whole

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -804,52 +804,67 @@ paths:
...
-    post:
+
+  /authorizations/{ownerKey}:
+    patch:
       tags:
         - Authorization
-      summary: Create an authorization
-      operationId: createAuthorization
+      summary: Patch an authorization
+      description: Add or remove permissions to an authorization
```
Ben-Sheppard, line 813: 💭 just an option on wording, the add part aligns with `to an authorization` but the `remove permissions` doesn't
```suggestion
      description: Manage the permissions assigned to the authorization
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/AuthorizationRequestValidator.java
```diff
@@ -8,34 +8,45 @@
...
+              permission -> {
+                if (permission.getPermissionType() == null) {
+                  violations.add(ERROR_MESSAGE_EMPTY_ATTRIBUTE.formatted("permissionType"));
+                } else if (permission.getResourceIds() == null
+                    || permission.getResourceIds().isEmpty()) {
+                  violations.add(
+                      ERROR_MESSAGE_EMPTY_NESTED_ATTRIBUTE.formatted(
+                          "resourceIds", permission.getPermissionType()));
+                }
+              });
```
Ben-Sheppard, line 49: 🔧  There is quite some nesting here, some is avoidable of course but I wonder if we could introduce some helper methods to make this block easier to understand, for example a method such as `validateResourceIds` that lifts the else if block out.

Ben-Sheppard (commented):
Just a few comments from my side, there are q few changes here but it seems good on the whole

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
remcowesterhoud: Most of them would match a ValueType, but not all. For example, we don't have an `APPLICATION` ValueType. All the others would match. I'd also argue `USER_GROUP` could be called just `GROUP`. But for now I just took what was listed in the document.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
remcowesterhoud: This is the first time I'm seeing this doc. You tell me if we need a wildcard 😄 I've been using [this doc](https://docs.google.com/document/d/1ikxw9ARCv-MBwASnj7WhW5uHCndCRk2B35xi-cB2N0M/edit#heading=h.25pdsyjx8m33) as my guideline.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
remcowesterhoud: The conclusion in the kickoff was that we should Ids instead of keys. A key is a long value that's generated within Zeebe. An Id is generally a String identifier.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
remcowesterhoud: In my mind this action does have something to do directly to do with the permissions inside of the requests. It refers to whether the permissions in the requests are added, or removed.

The way the request works is we patch an Authorization by adding or removing permissions.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
Ben-Sheppard: Thats ok with me, I just wanted to check if it impacted how the authorizations are applied lower down, but you are aware of that functionality so I will leave it to you :)

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
Ben-Sheppard: Ah, I think for now we will leave the wildcard out, it adds complexity and can just be achieved right now by selecting all the permissions. Down the line we may want to add one, but lets leave it

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
Ben-Sheppard: I see, fair point, I don't feel strongly enough about this to push for any change as you've given me a valid viewpoint :)

Ben-Sheppard (approved):
Thank you for answering my questions and applying changes where necessary, I can't see anything fundamentally wrong here so I'm going to approve so we can progress on!

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
Ben-Sheppard: Agreed :)

PR #21917, opened 2024-09-04
feat: engine uses dynamic routing info

### LINKED ISSUE #21476: Engine uses routing info from state instead of `SubscriptionUtil`
Currently, the engine uses `SubscriptionUtil` and statically configured partition count to figure out to which partition messages and subscriptions belong.

Because the number of partitions may now change, the engine must use the routing information from the state instead.

We should define an interface that allows the engine to map correlation key to partition and inject an implementation everywhere where needed.

## Acceptance Criteria
- [ ] Engine no longer uses `SubscriptionUtil`
- [ ] Engine can use routing info even when the feature is disabled
- [ ] Routing info is read from state and updates when state changes

lenaschoenburg:
This replaces all usages of static partition count from the engine. Instead, we now inject `RoutingInfo`.
This new interface is implemented naively at first, by looking up the current routing info from the state or falling back to static partition count. The fallback is needed to not break all engine tests where migrations don't run and `DbRoutingState` is never initialized.
Later on, we might want to optimize the implementation to cache some information and not look up from state every time.
Because caching is difficult to get right (there's no callback to tell us whether the latest state changes were committed or rolled back), I'll leave this for later.

Closes https://github.com/camunda/camunda/issues/21476

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/routing/PersistedRoutingInfo.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/routing/PersistedRoutingInfo.java
@@ -18,7 +18,7 @@ import io.camunda.zeebe.msgpack.value.IntegerValue;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public class RoutingInfo extends UnpackedObject implements DbValue {
+final class PersistedRoutingInfo extends UnpackedObject implements DbValue {
   private final ArrayProperty<IntegerValue> partitions =
       new ArrayProperty<>("partitions", IntegerValue::new);
   private final EnumProperty<MessageCorrelationStrategy> messageCorrelationStrategy =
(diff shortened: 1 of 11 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/routing/PersistedRoutingInfo.java
```diff
@@ -18,7 +18,7 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public class RoutingInfo extends UnpackedObject implements DbValue {
+final class PersistedRoutingInfo extends UnpackedObject implements DbValue {
```
deepthidevaki, line 21: :+1:

deepthidevaki (approved):
:rocket:

lenaschoenburg:
This is causing flaky tests that assume a specific order in which we iterate through partitions, for example `CommandDistributionBehaviorTest`: https://github.com/camunda/camunda/actions/runs/10701636922

PR #21928, opened 2024-09-04
feat: add routing info to cluster endpoint response

### LINKED ISSUE #21920: Cluster actuator response includes routing info
The actuator should return the current routing info from the `GET` endpoint.

lenaschoenburg:
This extends the OpenAPI spec with the new routing state and defines the necessary mapping.

closes https://github.com/camunda/camunda/issues/21920

### DIFF
--- a/dist/src/main/resources/api/cluster/components.yaml
+++ b/dist/src/main/resources/api/cluster/components.yaml
@@ -93,6 +93,55 @@ schemas:
         $ref: "components.yaml#/schemas/CompletedChange"
       pendingChange:
         $ref: "components.yaml#/schemas/TopologyChange"
+      routing:
+        $ref: "components.yaml#/schemas/RoutingState"
+
+  RoutingState:
+    description: The routing state, i.e. where requests are sent to and how messages are correlated
+    type: object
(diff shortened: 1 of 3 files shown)


dist/src/main/resources/api/cluster/components.yaml
```diff
@@ -93,6 +93,55 @@ schemas:
...
+          format: int32
+      messageCorrelation:
+        $ref: "components.yaml#/schemas/MessageCorrelation"
+
+
+  MessageCorrelation:
+    description: The message correlation strategy
+    oneOf:
+      - $ref: "components.yaml#/schemas/MessageCorrelationHashMod"
+    discriminator:
```
deepthidevaki, line 127: :+1: TIL

deepthidevaki (approved):
:rocket:

dist/src/main/resources/api/cluster/components.yaml
lenaschoenburg: Not suuuper ergonomic because we still have to know about this value and set it when building the response but otherwise quite nice.

Issue #21862, opened 2024-09-03
Migrate signal boundary event subscriptions


berkaycanbc:
It should be possible to migrate signal boundary event subscriptions, similar to migrating a timers, by providing a mapping instructions for the related signal catch event.

Consider the following example, where an active user task is subscribed to a signal boundary event (interrupting). When migrating the process instance, mapping instructions are provided for both the user task and the signal boundary event. After the instance is migrated, the user task is now subscribed to the attached signal boundary event (non-interrupting).

<img width="389" alt="Screenshot 2024-09-03 at 12 33 12" src="https://github.com/user-attachments/assets/175bba04-7c03-44ed-b307-5b25a211faa0">

We will introduce a new `Signal:MIGRATED` intent to represent the subscription migration.

Subscriptions to signal boundary events for which no mapping instructions are provided will not be migrated. Similar to these cases for timers, we can unsubscribe and subscribe to these catch events.
```[tasklist]
### Tasks
- [x] Subscribe to signal event subscriptions
- [x] Unsubscribe from signal event subscriptions
```

Issue #21447, opened 2024-08-21
Optionally ordered command distribution


lenaschoenburg:
Extend the `CommandDistributionBehavior` and related classes to optionally support ordered distribution.
This is useful when distributing multiple commands that require a specific order, for example to prevent https://github.com/camunda/camunda/issues/21357 and to support [partition scaling](https://github.com/camunda/camunda/issues/21439).

Currently the command distribution relies on a unique `distributionKey` per distributed command. A command can be distributed over one or multiple partitions. Because sending commands to remote partitions is not reliable or ordered, and all distributions are started immediately, the order in which commands are distributed is arbitrary. It follows that these commands are received, written and processed in arbitrary order too.

To support ordered distribution, users of `CommandDistributionBehavior` can provide a `queueId`. All distributions submitted with the same `queueId` are distributed (and thus received, written and processed) in the order they were submitted. The `queueId` works per-partition, meaning that **ordering is maintained within but not across partitions**. For example if C1 and C2 are submitted for queue X to be distributed to partition 1, 2 and 3, each partition will receive C1 first and then C2. But one partition might already receive C2 while another partition has not even received C1 yet.

## Acceptance Criteria

- [ ] Unordered command distribution is not affected, works the same way as before
- [ ] No data migrations needed to support queueing
- [ ] Rolling updates are not affected, ordering is maintained even when some brokers are old
- [ ] Commands submitted with the same `queueId` are distributed in the order they were submitted
- [ ] One partition cannot block distribution to another partition

PR #21932, opened 2024-09-04
Refactor handle catch event migrations

### LINKED ISSUE #21924: Fix and refactor handle catch event migrations
## Description

During the implementation of https://github.com/camunda/camunda/issues/21862 we encountered with an issue where signal subscriptions with the same signal name cannot be re-subscribed (no mapping instruction provided for signal boundary event during the migration) within the same migration. Because in the state, we use `signal name` and `subscription key` (instance key of the element that the signal event belongs to) as the column key here:
https://github.com/camunda/camunda/blob/ebe00fc9390669a7d079d883b884dafe3a3eb315/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/signal/DbSignalSubscriptionState.java#L24-L69

In the migration processor, we first subscribe to events and then unsubscribe from them later. Since we run the subscribing first, when the signal name is the same, we override the existing signal subscription in the state which is a wrong behaviour. To fix this we can unsubscribe first and subscribe later.

As a part of this issue, we can do an overall refactoring for `handleCatchEvents()` method. The plan is to change the order of execution as follows:
1) Unsubscribe from process instance events that are not part of the provided mapping instruction
2) Subscribe to target process definition events that are not part of the provided mapping instruction
3) Migrate events that are provided in the mapping instruction

In that way, we do the separation by execution instead of per event type.

berkaycanbc:
## Description

During the implementation of https://github.com/camunda/camunda/issues/21862 we encountered with an issue where signal subscriptions with the same signal name cannot be re-subscribed (no mapping instruction provided for signal boundary event during the migration) within the same migration. Because in the state, we use `signal name` and `subscription key` (instance key of the element that the signal event belongs to) as the column key here:
https://github.com/camunda/camunda/blob/ebe00fc9390669a7d079d883b884dafe3a3eb315/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/signal/DbSignalSubscriptionState.java#L24-L69

In the migration processor, we first subscribe to events and then unsubscribe from them later. Since we run the subscribing first, when the signal name is the same, we override the existing signal subscription in the state which is a wrong behaviour. To fix this we can unsubscribe first and subscribe later.

As a part of this issue, we can do an overall refactoring for `handleCatchEvents()` method. The plan is to change the order of execution as follows:
1) Unsubscribe from process instance events that are not part of the provided mapping instruction
2) Subscribe to target process definition events that are not part of the provided mapping instruction
3) Migrate events that are provided in the mapping instruction

In that way, we do the separation by execution instead of per event type.

## Related issues

closes #21924

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
@@ -1,0 +1,390 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.processinstance;
+
+import static io.camunda.zeebe.engine.processing.processinstance.ProcessInstanceMigrationPreconditions.requireNoPendingMsgSubMigrationDistribution;
+
+import io.camunda.zeebe.engine.processing.bpmn.BpmnElementContextImpl;
+import io.camunda.zeebe.engine.processing.common.CatchEventBehavior;
+import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableCatchEventSupplier;
+import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.processinstance.ProcessInstanceMigrationPreconditions.ProcessInstanceMigrationPreconditionFailedException;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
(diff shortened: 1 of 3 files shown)


koevskinikola:
@berkaycanbc can you re-base the changes so it's easier to complete the review? There are also some merge conflicts since the initial PR has already been merged.

berkaycanbc:
@koevskinikola resolved merge conflicts and rebased the branch ✅

koevskinikola (approved):
LGTM 👍

korthout (commented):
Great job @berkaycanbc and @koevskinikola 👏 

👍 The refactoring greatly improved readability / understandability!

💭 I went back all the way to 96e1b895c043b4b6ffd9afaeb6638f1d9e57e664 to understand why we subscribed before unsubscribing. The reason was that we lacked information about which catch events to subscribe to after having unsubscribed from some. However, with the current solution this is covered quite elegantly:
- unsubscribe from all, except for those that are mapped and keep track of these to migrate later
- subscribe to all, except those that are mapped
- migrate the mapped ones that we tracked earlier

berkaycanbc:
@korthout Additionally, after this refactoring and with the help of new feature [Optionally ordered command distribution](https://github.com/camunda/camunda/issues/21447) we can resolve the issue of re-subscribing to the same message name. Because we will be unsubscribing first and subscribing later to the same message name but this time one after another in the message subscription partition.

PR #21934, opened 2024-09-04
Generate operations for scale request to add new partitions

deepthidevaki:
## Description

This PR 
* Defines three for patching the cluster configuration. 
    * The two requests for scaling allows to 
        * scale brokers by either providing the new clusterSize or providing a list of broker ids to add or remove. 
        * Scale partitions by providing a new partition count
        * Change replication factor
        * It is possible to change multiple parameters at the same time. For example, add partitions and brokers in the same request.
   * A new force remove brokers request allows to specify set of the brokers to remove, instead of the members to retain.  
   * These requests can eventually replace the existing `ScaleRequest`. But we keep it for backwards compatibility.
* Added the request transformer to generate operations.

This PR do not add the request handling on the coordinator. This will be done in a follow up PR to reduce the size of this PR. 

## Related issues

related #21482

### DIFF
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationManagementRequest.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationManagementRequest.java
@@ -42,6 +42,24 @@ public sealed interface ClusterConfigurationManagementRequest {
     }
   }
 
+  record ClusterScaleRequest(
+      Optional<Integer> newClusterSize,
+      Optional<Integer> newPartitionCount,
+      Optional<Integer> newReplicationFactor,
+      boolean dryRun)
+      implements ClusterConfigurationManagementRequest {}
+
+  record ClusterPatchRequest(
+      Set<MemberId> membersToAdd,
+      Set<MemberId> membersToRemove,
+      Optional<Integer> newPartitionCount,
+      Optional<Integer> newReplicationFactor,
+      boolean dryRun)
+      implements ClusterConfigurationManagementRequest {}
+
+  record ForceRemoveBrokersRequest(Set<MemberId> membersToRemove, boolean dryRun)
+      implements ClusterConfigurationManagementRequest {}
+
   record ExporterDisableRequest(String exporterId, boolean dryRun)
       implements ClusterConfigurationManagementRequest {}
 
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterPatchRequestTransformer.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterPatchRequestTransformer.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.dynamic.config.api;
+
+import io.atomix.cluster.MemberId;
+import io.camunda.zeebe.dynamic.config.changes.ConfigurationChangeCoordinator.ConfigurationChangeRequest;
+import io.camunda.zeebe.dynamic.config.state.ClusterConfiguration;
+import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation;
+import io.camunda.zeebe.util.Either;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+public final class ClusterPatchRequestTransformer implements ConfigurationChangeRequest {
+
+  private final Set<MemberId> membersToAdd;
+  private final Set<MemberId> membersToRemove;
+  private final Optional<Integer> newPartitionCount;
+  private final Optional<Integer> newReplicationFactor;
+
+  public ClusterPatchRequestTransformer(
+      final Set<MemberId> membersToAdd,
+      final Set<MemberId> membersToRemove,
+      final Optional<Integer> newPartitionCount,
+      final Optional<Integer> newReplicationFactor) {
+    this.membersToAdd = membersToAdd;
+    this.membersToRemove = membersToRemove;
+    this.newPartitionCount = newPartitionCount;
+    this.newReplicationFactor = newReplicationFactor;
+  }
+
+  @Override
+  public Either<Exception, List<ClusterConfigurationChangeOperation>> operations(
+      final ClusterConfiguration clusterConfiguration) {
+    // if membersToAdd and membersToRemove have common items, reject the request
+    if (membersToAdd.stream().anyMatch(membersToRemove::contains)) {
+      return Either.left(
+          new ClusterConfigurationRequestFailedException.InvalidRequest(
+              new IllegalArgumentException(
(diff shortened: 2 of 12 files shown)


zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationManagementRequest.java
```diff
@@ -42,6 +42,16 @@ public ScaleRequest(final Set<MemberId> members, final boolean dryRun) {
     }
   }
 
+  record ClusterPatchRequest(
+      Optional<Integer> newClusterSize,
+      Set<MemberId> membersToAdd,
+      Set<MemberId> membersToRemove,
```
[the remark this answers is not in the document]
lenaschoenburg: Oh I see you added that as explicit validation in the transformer. Using the record constructor instead might be a bit nicer but that's more of a :upside_down_face: and not a :wrench:

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterPatchRequestTransformer.java
```diff
@@ -0,0 +1,92 @@
...
+          new ClusterConfigurationRequestFailedException.InvalidRequest(
+              "Provide only one of 'newClusterSize' or 'membersToAdd'/'membersToRemove'"));
+    }
+
+    if (request.force()
+        && (request.newClusterSize().isPresent() || !request.membersToAdd().isEmpty())) {
+      return Either.left(
+          new ClusterConfigurationRequestFailedException.InvalidRequest(
+              "Force operation is only allowed when removing members"));
+    }
```
lenaschoenburg, line 88: :wrench: Similarly, changing the replication factor or partition count is also not supported.

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/api/ClusterPatchRequestTransformerTest.java
```diff
@@ -0,0 +1,412 @@
...
+    // when
+    final var result =
+        new ClusterPatchRequestTransformer(patchRequest, MemberId.from("0"))
+            .operations(oldClusterTopology);
+    assertThat(result).isRight();
+    final var operations = result.get();
+
+    // apply operations to generate new topology
+    final ClusterConfiguration newTopology =
+        TestTopologyChangeSimulator.apply(oldClusterTopology, operations);
```
lenaschoenburg, line 386: :thought_balloon: just in theory, we could do the same in the request transformer to validate the request, right? That way we'd ensure that even in production, the generated operations always lead to the expected outcome.

lenaschoenburg (approved):
Looks good :rocket: 

:thought_balloon: I'm a bit concerned about the long term maintainability and extensibility of the patch request. It's very powerful (that's the whole point) so handling it correctly feels tricky, requiring lots of validation and careful programming to always generate the right operations. I don't have anything concrete to improve, just half baked ideas. For example, could we keep the flexibility in the API but generate more specific requests from it, where it would then be easier to validate and transform them?

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/api/ClusterPatchRequestTransformerTest.java
deepthidevaki: In the tests we can verify it against an expected outcome. But in production, we don't know the expected outcome. 

We are validating generated operation in the `ConfigurationChangeCoordinator` by a simulator. May be that's what you meant.

lenaschoenburg: Oh yeah of course, we only have the patch request and not the full information to calculate the expected topology at the end. Nevermind then :+1:

deepthidevaki:
> For example, could we keep the flexibility in the API but generate more specific requests from it, where it would then be easier to validate and transform them?

Do you mean keep a single API in the gateway, but internally it is mapped to more than one request based on the provided parameters?  So something like: 

1. ScaleBrokerCount(newClusterSize, optionalPartitionCount, optionalReplicationFactor)
2. ChangeBrokerConfig(brokersToAdd, brokersToRemove, optionalPartitionCount, optionalReplicationfactor)
3. ForceRemoveBrokers(brokersToRemove)

lenaschoenburg:
> > For example, could we keep the flexibility in the API but generate more specific requests from it, where it would then be easier to validate and transform them?
> 
> Do you mean keep a single API in the gateway, but internally it is mapped to more than one request based on the provided parameters? So something like:
> 
>     1. ScaleBrokerCount(newClusterSize, optionalPartitionCount, optionalReplicationFactor)
> 
>     2. ChangeBrokerConfig(brokersToAdd, brokersToRemove, optionalPartitionCount, optionalReplicationfactor)
> 
>     3. ForceRemoveBrokers(brokersToRemove)

Yes, exactly! Feels like more work now but I think it'd simplify the transformer quite a bit and help us in the future.

deepthidevaki:
Let us not merge this PR then. I will change it to add three different requests.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationManagementRequest.java
```diff
@@ -42,6 +42,24 @@ public ScaleRequest(final Set<MemberId> members, final boolean dryRun) {
     }
   }
 
+  record ClusterScaleRequest(
```
lenaschoenburg, line 45: :upside_down_face: It should be possible to rename `ScaleRequest` to the more accurate `BrokerScaleReqest` or something similar. Just to distinguish it from the new and more generic `ClusterScaleRequest`. As far as I remember the name is not serialized anywhere.

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/api/ClusterScaleRequestTransformerTest.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import net.jqwik.api.ForAll;
+import net.jqwik.api.Property;
+import net.jqwik.api.constraints.IntRange;
+import org.assertj.core.api.Assertions;
+
+final class ClusterScaleRequestTransformerTest {
```
lenaschoenburg, line 32: :upside_down_face: The name should end with `RandomizedPropertyTest` because we use that to identify property tests: https://github.com/camunda/camunda/blob/180aa200ddbebfacf1d82c45543f28bae349bedb/parent/pom.xml#L2833-L2834

lenaschoenburg: Looks like we didn't do that for a few other tests too so feel free to ignore

lenaschoenburg (approved):
Nice, thanks for the refactoring! :rocket: 

:question: The mapping of the generic `PATCH` request on the API to one of the three new `ClusterConfigurationManagementRequest`s will be done later in `ClusterConfigurationManagementRequestsHandler` I guess?

deepthidevaki:
> ❓ The mapping of the generic `PATCH` request on the API to one of the three new `ClusterConfigurationManagementRequest`s will be done later in `ClusterConfigurationManagementRequestsHandler` I guess?

Either there or in the gateway itself.

PR #21965, opened 2024-09-05
feat: document store implementation (GCP)

### LINKED ISSUE #22010: [Doc Handling] GCP document store
**Description**

We should introduce a GCP-based document store as a default option for C8 SaaS offering.

Epic link: https://github.com/camunda/issues/issues/841

chillleader:
## Description

This PR will introduce the document store implementation based on the Google Cloud Storage.

I also introduced a new module structure for document-related code and refactored the service module, moving the hard-coded in-memory store implementation into a new package.

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/22010


document/api/src/main/java/io/camunda/document/api/DocumentMetadataModel.java
```diff
@@ -0,0 +1,18 @@
...
+package io.camunda.document.api;
+
+import java.time.ZonedDateTime;
+import java.util.Map;
+
+public record DocumentMetadataModel(
+    String contentType,
+    String fileName,
+    ZonedDateTime expiresAt,
+    long size,
```
sbuettner, line 17: Lets go for a `Long` in case its optional? WDYT?

document/api/src/main/java/io/camunda/document/api/DocumentMetadataModel.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+package io.camunda.document.api;
+
+import java.time.ZonedDateTime;
+import java.util.Map;
+
+public record DocumentMetadataModel(
+    String contentType,
+    String fileName,
+    ZonedDateTime expiresAt,
```
johnBgood, line 16: Any reason not to use a timestamp (long) here?

document/api/src/main/java/io/camunda/document/api/DocumentOperationResponse.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  enum DocumentErrorCode {
+    DOCUMENT_NOT_FOUND,
+    DOCUMENT_ALREADY_EXISTS,
+    INVALID_INPUT,
+    OPERATION_NOT_SUPPORTED,
+    UNKNOWN_ERROR
+  }
```
sbuettner, line 44: We could go with a sealed class hierarchy here instead of an enum. This would allow individual errors to return different properties
DOCUMENT_NOT_FOUND: The id of the requested document.
DOCUMENT_NOT_FOUND: The operation that is not supported.
DOCUMENT_ALREADY_EXISTS: Potentially the document that exists already.
INVALID_INPUT: Details about the invalid input. For example validation errors.

document/api/src/main/java/io/camunda/document/api/DocumentStoreClient.java
```diff
@@ -0,0 +1,24 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.api;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+public interface DocumentStoreClient {
+
+  String resolveStoreId(String storeId);
```
sbuettner, line 15: Curious what the purpose of this is? Do we have different ids per store?

document/store/src/main/java/io/camunda/document/store/InMemoryDocumentStore.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+
+public class InMemoryDocumentStore implements DocumentStore {
+
+  private final Map<String, byte[]> documents;
+
+  public InMemoryDocumentStore() {
+    documents = new HashMap<>();
```
sbuettner, line 31: What about using a `ConcurrentHashMap`?

document/api/src/main/java/io/camunda/document/api/DocumentStore.java
```diff
@@ -0,0 +1,21 @@
...
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.api;
+
+import java.io.InputStream;
+
+public interface DocumentStore {
```
sbuettner, line 12: If an implementation itself is async, wouldnt it be better to support an async store interface and let the implementation handle it (for example the S3 client library?). It looks like we have a blocking interface and an async client wrapper to call it. Curious about your thoughts on this.

sbuettner: Async S3 Client interface: https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/S3AsyncClient.html

lenaschoenburg: :+1: I can recommend to check out the backup store API which is very similar. All methods return futures here: https://github.com/camunda/camunda/blob/main/zeebe/backup/src/main/java/io/camunda/zeebe/backup/api/BackupStore.java

document/api/src/main/java/io/camunda/document/api/DocumentMetadataModel.java
chillleader: As per the [API guidelines](https://docs.google.com/document/d/1G9AmmNac-4QLGZ0LXQXa3FyeCrSJdIboaPt3-R6dWNw/edit#heading=h.yilyqujdhsix) we cannot return a timestamp from the API - consequently, returning it as a Date object here saves us an extra conversion. I wasn't sure about `ZonedDateTime` vs `LocalDateTime` but all other endpoints seem to use `ZonedDateTime`.

document/api/src/main/java/io/camunda/document/api/DocumentStoreClient.java
chillleader: The request may use a non-specific store ID such as `null` or `"default"` that is technically resolvable by the store wrapper, but the document reference should contain a specific store ID.

I'll try to provide the resolved ID as part of the return value instead, this should look nicer, wdyt?

document/api/src/main/java/io/camunda/document/api/DocumentStore.java
chillleader: Good idea, initially I didn't like how the method signature was becoming too long like `CompletableFuture<DocumentOperationResponse<DocumentReference>>` if we used the async store implementation. But I can clearly see the benefit of having an async interface to support async store clients directly. I'll give it another try.

document/api/src/main/java/io/camunda/document/api/DocumentMetadataModel.java
sbuettner: Somehow it got mixed up. I meant the filesize. 😄

document/api/src/main/java/io/camunda/document/api/DocumentOperationResponse.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.api;
+
+public sealed interface DocumentOperationResponse<T> {
```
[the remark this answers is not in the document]
tmetzke: Happy to help but I didn't have time yet to look into this deeper. Let me know what you want to achieve and I can tell you if I know of anything 🙂

document/api/src/main/java/io/camunda/document/api/DocumentMetadataModel.java
chillleader: Replied re. filesize [here](https://github.com/camunda/camunda/pull/21965#discussion_r1751590270), long story short I think it will always be present in the payload so no need to make it nullable

document/api/src/main/java/io/camunda/document/api/DocumentMetadataModel.java
sbuettner: Ok, in general I would recommend to use the object and not the primitive type for this.

document/api/src/main/java/io/camunda/document/api/DocumentError.java
```diff
@@ -0,0 +1,29 @@
...
+  record UnknownError(String message, Throwable cause) implements DocumentError {
+
+    public UnknownError(final String message) {
+      this(message, null);
+    }
+
+    public UnknownError(final Throwable cause) {
+      this(null, cause);
+    }
+  }
```
sbuettner, line 28: Thanks for considering the feedback. Looks very clean and readable.

document/api/src/main/java/io/camunda/document/api/DocumentStoreRegistry.java
```diff
@@ -0,0 +1,15 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.api;
+
+public interface DocumentStoreRegistry {
+
+  DocumentStoreRecord getDocumentStore(final String id);
+
+  DocumentStoreRecord getDefaultDocumentStore();
```
sbuettner, line 14: We could also consider a sealed class that represents either the default store or a store with an idea. As the default store could be disabled. Might also want to consider an either here is well in case a store isnt found.

sbuettner (approved):
Only a minor modeling comment is left which is not a blocker. 

Great work and a good way to move forward!

Lets see how this evolves when we add an additional store like S3.

document/api/src/main/java/io/camunda/document/api/DocumentStoreRegistry.java
chillleader: Good point, I'm not sure at the moment if we will want to disable the default store (assumption is: one of the stores should be nominated as default anyway). But I added a todo to consider this later when we have more stores

document/store/src/main/java/io/camunda/document/store/InMemoryDocumentStore.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class InMemoryDocumentStore implements DocumentStore {
```
johnBgood, line 26: @chillleader do we want to mark this class as @NotMultiInstanceSafe (adding a comment would work) as a reminder?

johnBgood (approved):

PR #18669, opened 2024-05-21
feat: clients can provide a reference to a command that can be observed in the resulting event

mustafadagher:
## Description
Allow zeebe clients to send an `operation reference` to be included in the resulted record metadata.

- Created a new interface `CommandWithOperationReferenceStep` that could be used to extend any existing command steps to have operation reference be passed to them.
- Extended the following commands to have the `operationReference`: 
  - CancelProcessInstanceCommand
  - DeleteResourceCommand 
  - MigrateProcessInstanceCommand
  - ModifyProcessInstanceCommand
  - ResolveIncidentCommand
  - SetVariablesCommand
  - UpdateRetriesJobCommand
  - UpdateTimeoutJobCommand
- Updated `zeebe-record-template` to include the `operationReference` in elasticsearch and opensearch indexes

- An example of how to use it with the `CancelProcessInstanceCommand` would look like 
```java
client
  .newCancelInstanceCommand(processInstanceKey)
  .operationReference(OPERATION_REFERENCE)
  .send()
  .join();
```
## Related issues

closes #16916

## Context
I initially implemented https://github.com/camunda/zeebe/pull/17877 as a solution to the same problem, yet, after a discussion with @korthout, we agreed that having zeebe generate the requestId and send it response could result in edge cases where a request is processed but a network partition (or similar issue) happens before a client gets a response, and hence missing the `requestId` correlated for their operation.

While the above problem should necessarily be a rare edge case, and it's not very important to cover right now for our main stakeholder (operate), I thought I'll create another POC PR quickly and compare the two solutions afterwards.

PR #17805, opened 2024-04-25
ci: add spring zeebe sdk separate CI

nicpuppa:
## Description

Added a new workflow for Spring Zeebe SDK

## Related issues

closes #17767

PR #21967, opened 2024-09-05
Add set variables endpoint

### LINKED ISSUE #21057: Migrate the Set Variables gRPC API to REST
## Description

* The Camunda 8 REST API contains a variables update endpoint at `PATCH /v2/element-instances/{elementInstanceKey}/variables` providing the same functionality as the [gRPC endpoint](https://docs.camunda.io/docs/next/apis-tools/zeebe-api/gateway-service/#setvariables-rpc).
* The endpoint is featured in the OpenAPI description as (internally) [described](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.atrvej6tl897).
* Variables can be updated for one element instance (e.g. process instance, flow element instance).
* The provided variables will be updated, other variables stay untouched.

nicpuppa:
## Description

This PR add the Set Variables REST endpoint

## Related issues

closes #21057

### DIFF
--- a/service/src/main/java/io/camunda/service/ElementServices.java
+++ b/service/src/main/java/io/camunda/service/ElementServices.java
@@ -34,7 +34,7 @@ public class ElementServices extends ApiServices<ElementServices> {
   public CompletableFuture<VariableDocumentRecord> setVariables(final SetVariablesRequest request) {
     final var brokerRequest =
         new BrokerSetVariablesRequest()
-            .setElementInstanceKey(request.elementKey())
+            .setElementInstanceKey(request.elementInstanceKey())
             .setVariables(getDocumentOrEmpty(request.variables()))
             .setLocal(request.local());
 
@@ -45,5 +45,8 @@ public class ElementServices extends ApiServices<ElementServices> {
   }
 
   public record SetVariablesRequest(
-      long elementKey, Map<String, Object> variables, Boolean local, Long operationReference) {}
+      long elementInstanceKey,
+      Map<String, Object> variables,
+      Boolean local,
+      Long operationReference) {}
 }
--- a/service/src/main/java/io/camunda/service/ElementServices.java
+++ b/service/src/main/java/io/camunda/service/ElementServices.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.service;
+
+import io.camunda.search.clients.CamundaSearchClient;
+import io.camunda.service.security.auth.Authentication;
+import io.camunda.service.transformers.ServiceTransformers;
+import io.camunda.zeebe.broker.client.api.BrokerClient;
+import io.camunda.zeebe.gateway.impl.broker.request.BrokerSetVariablesRequest;
+import io.camunda.zeebe.protocol.impl.record.value.variable.VariableDocumentRecord;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class ElementServices extends ApiServices<ElementServices> {
+
+  public ElementServices(
+      final BrokerClient brokerClient,
+      final CamundaSearchClient searchClient,
+      final ServiceTransformers transformers,
+      final Authentication authentication) {
+    super(brokerClient, searchClient, transformers, authentication);
+  }
+
+  @Override
+  public ElementServices withAuthentication(final Authentication authentication) {
+    return new ElementServices(brokerClient, searchClient, transformers, authentication);
+  }
+
+  public CompletableFuture<VariableDocumentRecord> setVariables(final SetVariablesRequest request) {
+    final var brokerRequest =
+        new BrokerSetVariablesRequest()
+            .setElementInstanceKey(request.elementKey())
+            .setVariables(getDocumentOrEmpty(request.variables()))
+            .setLocal(request.local());
+
+    if (request.operationReference() != null) {
+      brokerRequest.setOperationReference(request.operationReference());
+    }
+    return sendBrokerRequest(brokerRequest);
+  }
+
+  public record SetVariablesRequest(
+      long elementKey, Map<String, Object> variables, Boolean local, Long operationReference) {}
+}
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1320,14 +1320,14 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
-  /elements/{elementKey}/variabels:
+  /elements/{elementKey}/variables:
     post:
       tags:
         - Variables
+        - Elements
       summary: Updates element variables
       description: |
-        Updates all the variables of a particular scope (e.g. process instance, flow element
-        instance) with the given variable data.
+        Updates all the variables of a particular scope (e.g. process instance, flow element instance) with the given variable data.
       parameters:
         - name: elementKey
           in: path
@@ -2763,7 +2763,7 @@ components:
           additionalProperties: true
           type: object
         local:
-          description: >
+          description: |
             If set to true, the variables will be merged strictly into the local scope
             (as indicated by elementKey).
             Otherwise, the variables are propagated to upper scopes and set at the outermost one.
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1320,19 +1320,18 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
-  /elements/{elementKey}/variables:
+  /element-instances/{elementInstanceKey}/variables:
     post:
       tags:
-        - Variables
-        - Elements
-      summary: Updates element variables
+        - Element instance
+      summary: Updates element instance variables
       description: |
         Updates all the variables of a particular scope (e.g. process instance, flow element instance) with the given variable data.
       parameters:
-        - name: elementKey
+        - name: elementInstanceKey
           in: path
           required: true
-          description: The key of the element to which to update the variables.
+          description: The key of the element instance to which to update the variables.
           schema:
             type: integer
             format: int64
@@ -2764,8 +2763,7 @@ components:
           type: object
         local:
           description: |
-            If set to true, the variables will be merged strictly into the local scope
-            (as indicated by elementKey).
+            If set to true, the variables will be merged strictly into the local scope (as indicated by elementInstanceKey).
             Otherwise, the variables are propagated to upper scopes and set at the outermost one.
 
             Let’s consider the following example:
@@ -2775,7 +2773,7 @@ components:
             1 => { "foo" : 2 }
             2 => { "bar" : 1 }
 
-            An update request with elementKey as ‘2’, variables { "foo" : 5 }, and local set
+            An update request with elementInstanceKey as ‘2’, variables { "foo" : 5 }, and local set
             to true leaves scope ‘1’ unchanged and adjusts scope ‘2 ‘ to { "bar" : 1, "foo" 5 }.
 
             By default, with local set to false, scope ‘1’ will be { "foo": 5 }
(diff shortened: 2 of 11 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1272,6 +1272,44 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
+  /elements/{elementKey}/variabels:
```
nicpuppa, line 1275: I implemented straight forward my idea, this can be easily changed

pepopowitz, line 1275:
```suggestion
  /elements/{elementKey}/variables:
```

❌ Spelling error. From my very non-Java-engineer perspective, I _think_ this is spelled correctly in the corresponding Java class, so the spelling correction is not needed elsewhere.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1272,6 +1272,44 @@ paths:
...
+    post:
+      tags:
+        - Variables
+      summary: Updates element variables
+      description: |
+        Updates all the variables of a particular scope (e.g. process instance, flow element
+        instance) with the given variable data.
```
pepopowitz, line 1282:
```suggestion
        Updates all the variables of a particular scope (e.g. process instance, flow element instance) with the given variable data.
```

:x: This is the issue I've mentioned about our docs generator choosing exactly the first line for the page's meta description, which results in an incomplete sentence in cases like this. 

💭 I don't like how frequently this idiosyncrasy of our docs generator is already making an appearance in my reviews, especially since I suspect you all are just using your editor to auto-wrap these descriptions. See my & @tmetzke's related conversation at https://github.com/camunda/camunda/pull/21787#discussion_r1745757112.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1272,6 +1272,44 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
+  /elements/{elementKey}/variabels:
+    post:
+      tags:
+        - Variables
```
pepopowitz, line 1278:
```suggestion
        - Elements
```

🔧 is this endpoint about the "Variables" resource, or about the "Elements" resource? To me it sounds like Elements, and you're setting Variables on an Element.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2550,6 +2588,42 @@ components:
...
 
+    VariableRequest:
+      type: object
+      properties:
+        variables:
+          description: JSON object representing the variables to set in the element’s scope.
+          additionalProperties: true
+          type: object
+        local:
+          description: >
```
pepopowitz, line 2599:
```suggestion
          description: |
```
:x: I think we want to preserve the line-breaks in this comment. 

TIL [the difference between `|` and `>` for multi-line YAML](https://yaml-multiline.info/#block-scalars) 😄 


Here's what it looks like in a Swagger UI with `>` (and similar in the generated docs):

image

And here's what it looks like with `|`:

image

pepopowitz (changes requested):
Great docs, a few things to clean up from my perspective! 

Also, I realized I wasn't using [the emoji code](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md#review-emoji-code) in my previous reviews of this sort. Please let me know if I'm using it incorrectly!

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nicpuppa: Yes, it's for element resources. We can use both tags, wdyt ?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nicpuppa: I need to pay more attention, thanks 🙇

nicpuppa:
> Also, I realized I wasn't using [the emoji code](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md#review-emoji-code) in my previous reviews of this sort. Please let me know if I'm using it incorrectly!

Great, using the emoji help us to understand more your suggestions. 

In my [previous comment](https://github.com/camunda/camunda/pull/21787#issuecomment-2331075483) I was referring at those examples ([1#](https://github.com/camunda/camunda/pull/17805#pullrequestreview-2038808451) [2#](https://github.com/camunda/camunda/pull/20391#pullrequestreview-2211798497) [3#](https://github.com/camunda/camunda/pull/18669#pullrequestreview-2093354305)). 
Pre-approving PRs can be risky (auto merge for example), but I think is a good approach when we think that the suggested changes are easy to address. 
Pre-approving PRs will be part of Zeebe review guidelines very soon.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: For each tag, we add a navigation category in our reference docs. The endpoint is added to all categories then. This is a bit weird, I guess, so we should focus on one tag, I think. Voting for "Element instances" since this is the resource we're working on.

tmetzke: See [here](https://github.com/camunda/camunda-docs/pull/4217#discussion_r1742669720) what happens with multiple tags

tmetzke: @pepopowitz, just noticed, we currently use the singular form for the tags, not the plural. This is (mostly) what Operate, Optimize, and Tasklist did as well in their REST APIs so far.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Looking at [the discussion](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit?disco=AAABTUdE7BY) results, let's go for `element-instances` instead of `elements`. Wdyt?

nicpuppa: I'll change the path accordingly

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nicpuppa: I'd go for element instance

service/src/main/java/io/camunda/service/ElementServices.java
```diff
@@ -0,0 +1,52 @@
...
+import io.camunda.search.clients.CamundaSearchClient;
+import io.camunda.service.security.auth.Authentication;
+import io.camunda.service.transformers.ServiceTransformers;
+import io.camunda.zeebe.broker.client.api.BrokerClient;
+import io.camunda.zeebe.gateway.impl.broker.request.BrokerSetVariablesRequest;
+import io.camunda.zeebe.protocol.impl.record.value.variable.VariableDocumentRecord;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class ElementServices extends ApiServices<ElementServices> {
```
tmetzke, line 19: 🔧 Should we rename this to `ElementInstanceServices` as well to reflect the change? Makes it a bit more relatable as well since "Element" could be anything 😅

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2717,6 +2754,42 @@ components:
           description: All the variables visible in the root scope.
           type: object
 
+    VariableRequest:
```
tmetzke, line 2757: 🔧 Looking at the other endpoints, `SetVariablesRequest` might be more appropriate. Unfortunately, this collides with the same-name record from the element instance services. But the REST API model should be as consistent as possible (we're not gonna be able to hold this up forever, I fear, but let's try our best).

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1320,6 +1320,43 @@ paths:
...
+      tags:
+        - Element instance
+      summary: Updates element instance variables
+      description: |
+        Updates all the variables of a particular scope (e.g. process instance, flow element instance) with the given variable data.
+      parameters:
+        - name: elementInstanceKey
+          in: path
+          required: true
+          description: The key of the element instance to which to update the variables.
```
tmetzke, line 1334: 🔧 Let's try to provide more hints on where this can be fetched from as a user. The gRPC API gives more details here that we should reuse.

> can be the process instance key (as obtained during instance creation), or a given element, such as a service task (see elementInstanceKey on the job message)

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2717,6 +2754,42 @@ components:
...
+            There are two scopes '1' and '2'.
+            Scope ‘1’ is the parent scope of ‘2’. The effective variables of the scopes are:
+            1 => { "foo" : 2 }
+            2 => { "bar" : 1 }
+
+            An update request with elementInstanceKey as ‘2’, variables { "foo" : 5 }, and local set
+            to true leaves scope ‘1’ unchanged and adjusts scope ‘2 ‘ to { "bar" : 1, "foo" 5 }.
+
+            By default, with local set to false, scope ‘1’ will be { "foo": 5 }
+            and scope ‘2’ will be { "bar" : 1 }.
```
tmetzke, line 2780: 🔧 Some of the `'` are off, I think they are different characters. Probably an issue with the original text and copy-pasting them into the OpenAPI spec.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/ElementInstanceController.java
```diff
@@ -0,0 +1,54 @@
...
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  public CompletableFuture<ResponseEntity<Object>> setVariables(
+      @PathVariable final long elementInstanceKey,
+      @RequestBody final VariableRequest variableRequest) {
+    return RequestMapper.toVariableRequest(variableRequest, elementInstanceKey)
+        .fold(RestErrorMapper::mapProblemToCompletedResponse, this::variables);
+  }
+
+  private CompletableFuture<ResponseEntity<Object>> variables(
```
tmetzke, line 46: 🔧 Should we go for `setVariables` as the method name?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/ElementRequestValidator.java
```diff
@@ -0,0 +1,35 @@
...
+        violations -> {
+          if (variableRequest.getVariables() == null || variableRequest.getVariables().isEmpty()) {
+            violations.add(ERROR_MESSAGE_EMPTY_ATTRIBUTE.formatted("variables"));
+          }
+          if (variableRequest.getOperationReference() != null
+              && variableRequest.getOperationReference() < 1) {
+            violations.add(
+                ERROR_MESSAGE_INVALID_ATTRIBUTE_VALUE.formatted(
+                    "operationReference", variableRequest.getOperationReference(), "> 0"));
+          }
```
tmetzke, line 32: 🔧 I guess we can extract this to the general request validator, passing in the operation reference value.

tmetzke (approved):
👍 Really nice, great job 🚀 
🔧 Added some smaller suggestions, nothing blocking, functionally this is done from my perspective 🧑‍🚀

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/ElementRequestValidator.java
nicpuppa: Yes, I want to use a general approach for validate `operationReference` fields. I'll do it in a separate task if you don't mind

mesellings:
@nicpuppa I will review this for you 👍

pepopowitz (approved):
Thanks for making changes!

I'll pre-approve in the future for small change requests.

mesellings (changes requested):
@nicpuppa I've made some suggestions for you, some of which I think are blocking (from a writing perspective). I tried using the review emoji code, hope I'm doing it correctly? Great work by the way! 👍

mesellings (approved):

PR #21970, opened 2024-09-05
ci: only add issues to zeebe projects if component matches

megglos:
## Description

Up until now all non task issues were added to the Zeebe project boards, this was too verbose given the activity of multiple teams on the repo. Now only issues labeled as zeebe will get added, also on changing labels (in case an issue was lacking as label on initial creation).

### DIFF
--- a/.github/workflows/add-to-projects.yml
+++ b/.github/workflows/add-to-projects.yml
@@ -37,7 +37,7 @@ jobs:
           project-url: https://github.com/orgs/camunda/projects/92
           github-token: ${{ secrets.GH_PROJECT_AUTOMATION_TOKEN }}
           labeled: component/zeebe
-      - id: add-to-zda
+      - id: add-to-zpa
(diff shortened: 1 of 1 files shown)


.github/workflows/add-to-projects.yml
```diff
@@ -36,17 +36,15 @@ jobs:
         with:
           project-url: https://github.com/orgs/camunda/projects/92
           github-token: ${{ secrets.GH_PROJECT_AUTOMATION_TOKEN }}
-          labeled: kind/task
-          label-operator: NOT
+          labeled: component/zeebe
       - id: add-to-zda
```
remcowesterhoud, line 40: 🔧 I know you didn't change it, but you would mind changing this typo? 😄 

```suggestion
      - id: add-to-zpa
```

remcowesterhoud (approved):
Very happy with this one 🥳

PR #21971, opened 2024-09-05
Add publish message endpoint

### LINKED ISSUE #21056: Migrate the Publish Message gRPC API to REST
## Description

* The Camunda 8 REST API contains a message publication endpoint at `POST /v2/messages/publication` providing the same functionality as the [gRPC endpoint](https://docs.camunda.io/docs/next/apis-tools/zeebe-api/gateway-service/#publishmessage-rpc).
* The endpoint is featured in the OpenAPI description as (internally) [described](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.110qcwu7c17f).

nicpuppa:
## Description

This PR add a new Publish Message REST endpoint.

## Related issues

closes #21056


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -898,6 +898,42 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+
+  /messages/pubblication:
```
pepopowitz, line 902:
```suggestion
  /messages/publication:
```

:x: Spelling error.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -898,6 +898,42 @@ paths:
...
+    post:
+      tags:
+        - Message
+      summary: Publish a message
+      description: |
+        Publishes a single message. Messages are published to specific partitions computed from
+        their correlation keys.
```
pepopowitz, line 909:
```suggestion
        Publishes a single message. 
        Messages are published to specific partitions computed from their correlation keys.
```

:x: Endpoint `description`s require a complete sentence on the first line.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -898,6 +898,42 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/MessagePublicationResponse"
+        "400":
+          description: The provided data is not valid.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "500":
+          description: Internal Server Error.
```
pepopowitz, line 932:
```suggestion
          description: Internal server error.
```

:x: Case consistency.

pepopowitz (changes requested):
A few small changes in the spec.

pepopowitz (approved):
👍 thanks for the updates!

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/MessageRequestValidator.java
```diff
@@ -0,0 +1,42 @@
...
+
+import static io.camunda.zeebe.gateway.rest.validator.ErrorMessages.ERROR_MESSAGE_EMPTY_ATTRIBUTE;
+import static io.camunda.zeebe.gateway.rest.validator.RequestValidator.validate;
+
+import io.camunda.zeebe.gateway.protocol.rest.MessageCorrelationRequest;
+import io.camunda.zeebe.gateway.protocol.rest.MessagePublicationRequest;
+import java.util.Optional;
+import org.springframework.http.ProblemDetail;
+
+public final class MessageRequestValidator {
```
tmetzke, line 18: 👍 Good call to create this

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2460,6 +2498,51 @@ components:
...
+    MessagePublicationRequest:
+      type: object
+      properties:
+        name:
+          description: The name of the message.
+          type: string
+        correlationKey:
+          description: The correlation key of the message.
+          type: string
+          default: ""
```
tmetzke, line 2510: ❓ Do we need this? We could also remove the default, remove it from the required list, and form an empty String by default in the backend, right?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2460,6 +2498,51 @@ components:
...
+          type: object
+          nullable: true
+        tenantId:
+          description: The tenant of the message sender.
+          type: string
+          nullable: true
+      required:
+        - name
+        - correlationKey
+        - timeToLive
```
tmetzke, line 2533: ❓ Is this really required? If this is not provided in gRPC, it is `0` by default, simply leading to the message not being buffered on the broker. That should be fine, right?

tmetzke (commented):
👍 Looks really good 🚀 
❓ Only some smaller questions regarding some parameters and their default values.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nicpuppa: There is no difference on setting the default here or in the service IMO, using the specification to underline this behaviour make it more clear also for the users

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nicpuppa: We can argue here, we can also remove `timeToLive` from the required field add the default and minimum value (I don't think a negative value can be provided here)

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Yeah, I don't have a strong opinion here, to be honest. Both is fine for me 👍

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: A negative value can actually be provided, the backend is resilient enough and does the same thing for any value `<=0` 🤷

tmetzke: Removing it from the required fields allows for easier calls for users since it's actually not required for us, it can be left out and has a proper default behavior then.

tmetzke (approved):
Pre-approving since open discussions are only about cosmetic things, functionality is provided already.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nicpuppa: I'd say, let's remove it from the required field, and add the default value to 0, wdyt ? @tmetzke

tmetzke: Sounds good to me 👍

PR #21978, opened 2024-09-05
refactor: use ES module format in dependency script

OmranAbazid:
## Description
When i tried to use this module to generate the dependencies, it did not work since we migrated optimize to es module format.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/client/scripts/dependencies.js
+++ b/optimize/client/scripts/dependencies.js
@@ -5,16 +5,15 @@
  * Licensed under the Camunda License 1.0. You may not use this file
  * except in compliance with the Camunda License 1.0.
  */
+import fs from 'fs';
+import {resolve as _resolve, dirname} from 'path';
+import {execFileSync} from 'child_process';
+import {fileURLToPath} from 'url';
 
-const fs = require('fs');
-const path = require('path');
-const {execFileSync} = require('child_process');
-const url = require('url');
-const toPairs = require('lodash/toPairs');
-
-const licenseChecker = path.resolve(__dirname, '..', 'node_modules', '.bin', 'license-checker');
-const depJsonFile = path.resolve(__dirname, '..', 'dependencies.json');
-const depMdFile = path.resolve(__dirname, '..', 'frontend-dependencies.md');
+const __dirname = dirname(fileURLToPath(import.meta.url));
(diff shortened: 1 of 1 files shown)


optimize/client/scripts/dependencies.js
```diff
@@ -34,7 +33,7 @@ const lines = toPairs(dependencies)
   }, []);
 
 function getLibraryLink(name, rawUrl, repository) {
-  const parsedUrl = rawUrl ? url.parse(rawUrl) : {};
+  const parsedUrl = rawUrl ? new URL(rawUrl.startsWith('http') ? rawUrl : `https://${rawUrl}`) : {};
```
misiekhardcore, line 36: and what if the url starts with `https`? then still the `startsWith('http')` will be true. Do you think it can cause issues? What was the reason for this change? Didn't the `url.parse` method work?

optimize/client/scripts/dependencies.js
```diff
@@ -5,16 +5,15 @@
  * Licensed under the Camunda License 1.0. You may not use this file
  * except in compliance with the Camunda License 1.0.
  */
+import fs from 'fs';
```
misiekhardcore, line 8: We can destructure these too if we are destructuring everything else
```suggestion
import {readFileSync, writeFileSync} from 'fs';
```

optimize/client/scripts/dependencies.js
```diff
@@ -63,3 +62,15 @@ menu:
...
+function toPairs(obj) {
+  const pairs = [];
+
+  for (const key in obj) {
+    if (obj.hasOwnProperty(key)) {
+      pairs.push([key, obj[key]]);
+    }
+  }
+
+  return pairs;
```
misiekhardcore, line 75: Won't this do the same thing?
```suggestion
  return Object.entries(obj);
```

misiekhardcore (approved):
Pre-approving. See comments

OmranAbazid:
@misiekhardcore I will close this PR since we need to remove this file anyway as mentioned in https://github.com/camunda/camunda-optimize/issues/14139#issuecomment-2397331021

misiekhardcore:
@OmranAbazid agree. I opened this [PR](https://github.com/camunda/camunda/pull/23528) to remove this script instead

Issue #21980, opened 2024-09-05
Provide a way to register indices and index templates for expoirter


EuroLew:
**Description**

Other teams should be able to define indices and index templates which are then available in the `CamundaExporter` for using by the schema manager to create/update the resources.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/21981
- [ ] https://github.com/camunda/camunda/issues/21982
```

EuroLew:
Talking with @sdorokhova the way `IndexDescriptors` and `IndexTemplateDescriptors` are created is hard coded, in the future a possible refactor is move everything into the `_meta` block in the index json. For now however since it is hard coded this can be marked as completed

PR #21981, opened 2024-09-05
Skeleton of schema resource provider 

EuroLew:
## Description

A skeleton for how teams will provide schema resources such as indices and index templates to the `CamundaExporter` so it can be used by the schema manager to create/update resources.

## Related issues

relates to #21980

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterComponentsProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterComponentsProvider.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter;
+
+import io.camunda.exporter.schema.descriptors.IndexDescriptor;
+import io.camunda.exporter.schema.descriptors.IndexTemplateDescriptor;
+import java.util.List;
+
+public class DefaultExporterComponentsProvider implements ExporterComponentsProvider {
+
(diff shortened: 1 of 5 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterComponentsProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter;
+
+import io.camunda.exporter.schema.descriptors.IndexDescriptor;
+import io.camunda.exporter.schema.descriptors.IndexTemplateDescriptor;
+import java.util.List;
+
+public class DefaultExporterComponentsProvider implements ExporterComponentsProvider {
```
deepthidevaki, line 14: :wrench: Add java docs explaining this is the "production" code where the teams should add their "components" like handlers, index descriptors etc.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterComponentsProvider.java
```diff
@@ -0,0 +1,18 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter;
+
+import io.camunda.exporter.schema.descriptors.IndexDescriptor;
+import io.camunda.exporter.schema.descriptors.IndexTemplateDescriptor;
+import java.util.List;
+
+public interface ExporterComponentsProvider {
```
deepthidevaki, line 14: :x: Add java docs for the interface and the methods.

deepthidevaki: :wrench: Also if you can think of a better name that would be great :smile: I suggested this without thinking much.

deepthidevaki (approved):
:rocket: Thanks.

PR #21985, opened 2024-09-05
Removed NoopExporter

EuroLew:
## Description

The `NoopExporter` was a place holder now that the `CamundaExporter` exists it has been deleted and the `NoopExporterConfiguration` has been moved to `ElasticsearchProperties`


deepthidevaki (approved):
Thanks.

PR #21986, opened 2024-09-05
ci: Reduce maven's memory use when building optimize-backend

matthewBearCamunda:
## Description

Increase the build timeout to 15 minutes, set a cap on the memory used by `optimize-backend`, and do not allow forking on the same module

The goal is to reduce memory use as that module has been the one that has been failing on timeouts

These changes have had >10 successful CI runs. Most of them complete in a couple of minutes, some a little less than 10, and rarely over 10 minutes

## Checklist

- [x] for CI changes:

## Related issues

closes #

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -465,7 +465,7 @@ jobs:
     if: needs.detect-changes.outputs.optimize-backend-changes == 'true'
     needs: [detect-changes]
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
     env:
       LIMITS_CPU: 4
     steps:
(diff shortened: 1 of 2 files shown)


cmur2 (approved):
Let's merge this to mitigate the impact of the optimize-backend-unit-tests job being sometimes slow.

Given the timings you mentioned for this PR I still think we didn't find the root cause since the execution time spread for this job seems to be quite large from 2 minutes (fastest) to 12 minutes (slowest) which is factor 6.

In genera,l I want to take 2 steps:
* visualize and analyze build execution times over jobs in the CI at larger scale to identify trends, maybe this can be correlated to the runner type etc.
* tackle https://github.com/camunda/camunda/issues/22204 so we have easier means to look at resource usage (when running on Camunda's machines)

PR #22009, opened 2024-09-06
refactor: remove async completion

ChrisKujawa:
## Description

Feature was not used and added unnecessary complexity to the worker.

Note: Discussed this previously already with Nicolas, and he was fine with removing it.

## Related issues

related #21472


deepthidevaki (approved):
:rocket: Makes sense. 

:question: Can we increase the number of worker threads if needed? I don't think it matters for our normal benchmarks. But in case we see a case where the workers are the bottleneck, it might help.

Zelldon:
> 🚀 Makes sense.
> 
> ❓ Can we increase the number of worker threads if needed? I don't think it matters for our normal benchmarks. But in case we see a case where the workers are the bottleneck, it might help.

Yes it is an existing config already :) https://github.com/camunda/camunda/blob/main/zeebe/benchmarks/project/src/main/resources/application.conf#L25

PR #22011, opened 2024-09-06
refactor: user query from indexes

Ben-Sheppard:
## Description

With the exporter changes in https://github.com/camunda/camunda/pull/21841 users are now exported to the correct indexes in a structure that is flat. This means that we can update the existing code to use the new indexes and flatter structure.

Along with this, I have also set the user key to be exported on the new entities and added in the required filter/sort/transformer support


remcowesterhoud (approved):
All good 👍

PR #22016, opened 2024-09-06
feat: partition scoped metric registrity

### LINKED ISSUE #22017: ElasticSearch and OpenSearch exporter metrics are missing after role changes
Take this screenshot as proof: 
image

As we can see, only the metrics for partition 2 from broker 1 are published, the others are missing. Presumably, this was introduced by https://github.com/camunda/camunda/pull/20590 and https://github.com/camunda/camunda/pull/21029 because micrometer does not handle re-registration of gauges nicely.

lenaschoenburg:
This solves three problems:
1. How to remove meters when partitions are removed
2. How to reset meters when partitions change roles
3. How to set the partition id label for every meter

We do this by creating a new registry on every partition transition.
This allows us to define common labels (i.e. partition id) and to remove all meters specific to that partition and role.

The partition registry is wrapped over the general broker registry such that all meters are forwarded to the correct
backend automatically.

This happens to close https://github.com/camunda/camunda/issues/22017 which I only discovered while testing this PR.


lenaschoenburg:
See this as an example for automatic meter removal: 
image

Leadership for p3 changed from broker 0 to broker 2. After the role change, broker 0 does not even publish the meter anymore.

EuroLew (approved):
LGTM

PR #22021, opened 2024-09-06
feat: new camunda performance dashboard

ChrisKujawa:
## Description

Contribute new Camunda performance dashboard this dashboard should allow to give a holistic overview of the general performance of the Camunda platform.

Currently, this involves, Zeebe, Operate, and Elasticsearch mainly. Other components need to be added in the future.

Several best-practices have been followed to build this dashboard:

For example:
 * Collapse all rows per default
 * Add variables to select namespaces, pods, etc.
 * Use a common time interval for range vectors, for example `$_rate_interval` which is four times as big as the scrape interval (to have enough data points), and make all panels comparable
 * Add descriptions to most of the panels, to clarify the usage/shown metrics

### Details

You can find it here https://grafana.dev.zeebe.io/d/camunda-benchmarks/camunda-performance?orgId=1

It gives a general overview of the performance and healthiness of running components in the first section.

general

In the later sections, panels show latency and throughput-related metrics.

latency

> [!Note]
>
> Not all Zeebe-related latency and throughput metrics/panels have been added, as Zeebe already provides a great dashboard to investigate such if necessary.

## Checklist

- [ ] for CI changes:

## Related issues

related to https://github.com/camunda/camunda/issues/21462

### DIFF
--- a/monitor/grafana/camunda-performance.json
+++ b/monitor/grafana/camunda-performance.json
@@ -1,0 +1,3649 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__elements": {},
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "10.1.5"
+    },
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    },
+    {
+      "type": "panel",
+      "id": "stat",
+      "name": "Stat",
+      "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "timeseries",
+      "name": "Time series",
+      "version": ""
+    }
+  ],
+  "annotations": {
+    "list": [
+      {
+        "$$hashKey": "object:216",
+        "builtIn": 1,
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 1,
+  "id": null,
+  "links": [],
+  "liveNow": false,
+  "panels": [
+    {
+      "collapsed": true,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 27,
+      "panels": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$DS_PROMETHEUS"
+          },
+          "description": "Shows readiness of pods in the namespace ",
+          "fieldConfig": {
+            "defaults": {
+              "color": {
+                "mode": "palette-classic"
+              },
+              "custom": {
+                "axisCenteredZero": false,
+                "axisColorMode": "series",
+                "axisLabel": "",
+                "axisPlacement": "auto",
+                "axisSoftMax": 2,
+                "axisSoftMin": 0,
+                "barAlignment": 0,
+                "drawStyle": "line",
+                "fillOpacity": 0,
+                "gradientMode": "none",
+                "hideFrom": {
+                  "legend": false,
+                  "tooltip": false,
+                  "viz": false
+                },
+                "insertNulls": false,
+                "lineInterpolation": "linear",
+                "lineWidth": 1,
+                "pointSize": 5,
+                "scaleDistribution": {
+                  "type": "linear"
+                },
+                "showPoints": "auto",
+                "spanNulls": false,
+                "stacking": {
+                  "group": "A",
+                  "mode": "none"
+                },
+                "thresholdsStyle": {
+                  "mode": "off"
+                }
+              },
+              "decimals": 0,
+              "links": [],
+              "mappings": [
+                {
+                  "options": {
+                    "0": {
+                      "color": "dark-green",
+                      "index": 1,
+                      "text": "ready"
+                    },
+                    "1": {
+                      "color": "dark-red",
+                      "index": 0,
+                      "text": "unready"
+                    }
+                  },
+                  "type": "value"
+                }
+              ],
+              "max": 1,
+              "thresholds": {
+                "mode": "absolute",
+                "steps": [
+                  {
+                    "color": "green",
+                    "value": null
+                  },
+                  {
+                    "color": "red",
+                    "value": 80
+                  }
+                ]
+              },
+              "unit": "none"
+            },
+            "overrides": []
+          },
+          "gridPos": {
+            "h": 6,
+            "w": 24,
+            "x": 0,
+            "y": 1
+          },
+          "id": 26,
+          "options": {
+            "legend": {
+              "calcs": [
+                "lastNotNull"
+              ],
+              "displayMode": "table",
+              "placement": "right",
+              "showLegend": true,
+              "sortBy": "Last *",
+              "sortDesc": true
+            },
+            "tooltip": {
+              "mode": "single",
+              "sort": "none"
+            }
+          },
+          "pluginVersion": "10.1.5",
+          "targets": [
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "exemplar": false,
+              "expr": "1-kube_pod_container_status_ready{namespace=~\"$namespace\",pod!~\"leader.*\"}",
+              "format": "time_series",
+              "hide": false,
+              "instant": false,
+              "interval": "",
+              "legendFormat": "{{pod}}",
+              "range": true,
+              "refId": "A"
+            }
+          ],
+          "title": "Pod readiness",
+          "type": "timeseries"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "description": "Shows the processing queue size over time. Processing queue defines, what is already appended but not yet processed by the stream processor.",
+          "fieldConfig": {
+            "defaults": {
+              "color": {
+                "mode": "thresholds"
+              },
+              "mappings": [],
+              "min": 0,
+              "thresholds": {
+                "mode": "absolute",
+                "steps": [
+                  {
+                    "color": "green",
+                    "value": null
+                  }
+                ]
+              },
+              "unit": "short"
+            },
+            "overrides": []
+          },
+          "gridPos": {
+            "h": 5,
+            "w": 8,
+            "x": 0,
+            "y": 7
+          },
+          "id": 38,
+          "options": {
+            "colorMode": "value",
+            "graphMode": "area",
+            "justifyMode": "auto",
+            "orientation": "auto",
+            "reduceOptions": {
+              "calcs": [
+                "lastNotNull"
+              ],
+              "fields": "",
+              "values": false
+            },
+            "textMode": "auto"
+          },
+          "pluginVersion": "10.1.5",
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${DS_PROMETHEUS}"
+              },
+              "editorMode": "code",
+              "exemplar": false,
+              "expr": "max(zeebe_log_appender_last_appended_position{namespace=~\"$namespace\", partition=~\"$partition\"}) - max (zeebe_stream_processor_last_processed_position{namespace=~\"$namespace\", partition=~\"$partition\"})",
+              "instant": false,
+              "interval": "",
+              "legendFormat": "Partition {{partition}}",
+              "range": true,
+              "refId": "A"
+            }
+          ],
+          "title": "Number of records not processed",
+          "type": "stat"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "description": "Shows how many records are not exported yet.",
+          "fieldConfig": {
+            "defaults": {
+              "color": {
+                "mode": "thresholds"
+              },
+              "mappings": [],
+              "thresholds": {
+                "mode": "absolute",
+                "steps": [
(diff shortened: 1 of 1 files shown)


monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+            }
+          },
+          "pluginVersion": "10.4.0",
+          "targets": [
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "expr": "sum by (pod, container, service) (rate(container_cpu_usage_seconds_total{container!~\"\", namespace=~\"$namespace\"}[$__rate_interval]))",
```
lenaschoenburg, line 600: :question: Why sum by container and service? Especially when the label only shows you the pod. I've tried only summing by pod and it looked like we'd get the same result

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+          },
+          "pluginVersion": "10.1.5",
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${DS_PROMETHEUS}"
+              },
+              "editorMode": "code",
+              "expr": "max (zeebe_log_appender_last_committed_position{namespace=~\"$namespace\", partition=~\"$partition\"})  - max (zeebe_exporter_last_exported_position{namespace=~\"$namespace\", partition=~\"$partition\", exporter=\"elasticsearch\"}) ",
```
lenaschoenburg, line 322: :x: This is misleading because it mixes partitions. One partition can be very far in commit position, another partition very far in export position. Then the number of records not exported would be small even though the backlog of a different partition might be very large.

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+          },
+          "pluginVersion": "10.1.5",
+          "targets": [
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "exemplar": true,
+              "expr": "avg(rate(elasticsearch_indices_indexing_index_total{namespace=~\"$namespace\"}[$__rate_interval])) by (name)",
```
lenaschoenburg, line 702: :question: Why average by name? Name should be unique anyway, right?

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+          },
+          "pluginVersion": "10.1.5",
+          "targets": [
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "exemplar": true,
+              "expr": "avg(rate(elasticsearch_index_stats_indexing_index_total{namespace=~\"$namespace\"}[5m])) by (index)",
```
lenaschoenburg, line 944: :question: Why average by index when the index name has to be unique anyway?

lenaschoenburg: I suppose in this case it's useful for observing multiple namespaces at once, for example to observe that a certain index is always slow.

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+              "hide": false,
+              "interval": "",
+              "legendFormat": "{{index}}",
+              "range": true,
+              "refId": "A"
+            }
+          ],
+          "title": "indexing",
```
lenaschoenburg, line 952: :question: What's the difference to the "Documents Indexed Rate" panel? The answer should probably be part of the panel description.

:wrench: Naming wise this one sticks out. Lower case, not very descriptive etc.

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+            }
+          },
+          "pluginVersion": "10.4.0",
+          "targets": [
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "expr": "(kubelet_volume_stats_used_bytes{namespace=~\"$namespace\", persistentvolumeclaim=~\".*zeebe.*\"} / kubelet_volume_stats_capacity_bytes{namespace=~\"$namespace\", persistentvolumeclaim=~\".*zeebe.*\"})",
```
lenaschoenburg, line 1045: :wrench: We should just combine this with the panel for ES. You can define multiple queries for each panel if that's needed.

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+            }
+          },
+          "pluginVersion": "10.1.5",
+          "targets": [
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "expr": "sum(rate(operate_import_time_seconds_sum{namespace=~\"$namespace\", partition=~\"$partition\", pod=~\"$pod\"}[$__rate_interval])) / sum(rate(operate_import_time_seconds_count{namespace=~\"$namespace\", partition=~\"$partition\", pod=~\"$pod\"}[$__rate_interval]))",
```
lenaschoenburg, line 1415: :upside_down_face: Using percentile 50 (i.e. the median instead of mean) would result in more elegant query (less repetition) and might be more useful anyway because it handles outliers better.

lenaschoenburg: Same argument for a bunch of other queries

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+              "editorMode": "code",
+              "expr": "histogram_quantile(0.90, sum(increase(zeebe_stream_processor_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "hide": false,
+              "instant": false,
+              "legendFormat": "p90",
+              "range": true,
+              "refId": "C"
+            }
+          ],
+          "title": "Write-to-Processing latency in seconds (avg)",
```
lenaschoenburg, line 1828: :wrench: Title is not correct, it shows multiple percentiles, not the average.

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "expr": "histogram_quantile(0.5, sum(increase(zeebe_stream_processor_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "format": "time_series",
+              "interval": "",
+              "intervalFactor": 1,
+              "legendFormat": "p50 (avg)",
```
lenaschoenburg, line 1797: :wrench: Technically not correct, p50 is the median not average

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+          "targets": [
+            {
+              "datasource": {
+                "uid": "$DS_PROMETHEUS"
+              },
+              "editorMode": "code",
+              "expr": "histogram_quantile(0.5, sum(increase(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "format": "time_series",
+              "intervalFactor": 1,
+              "legendFormat": "p50 (avg)",
```
lenaschoenburg, line 1936: :wrench: Technically not correct, p50 is the median not the average

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+              "editorMode": "code",
+              "expr": "histogram_quantile(0.90, sum(increase(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "hide": false,
+              "instant": false,
+              "legendFormat": "p90",
+              "range": true,
+              "refId": "C"
+            }
+          ],
+          "title": "Write-to-Exporting latency in seconds (avg)",
```
lenaschoenburg, line 1967: :wrench: This shows multiple percentiles, not just the average

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+              "editorMode": "code",
+              "expr": "histogram_quantile(0.90, sum(increase(zeebe_stream_processor_processing_duration_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "hide": false,
+              "instant": false,
+              "legendFormat": "p90",
+              "range": true,
+              "refId": "C"
+            }
+          ],
+          "title": "Processing duration in seconds (avg)",
```
lenaschoenburg, line 2219: :wrench: This shows multiple percentiles but not the average

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+              "editorMode": "code",
+              "expr": "sum(rate(zeebe_element_instance_events_total{namespace=~\"$namespace\",partition=~\"$partition\", action=\"completed\", type=\"PROCESS\"}[$__rate_interval]))",
+              "format": "time_series",
+              "intervalFactor": 1,
+              "legendFormat": "{{type}} {{action}}",
+              "range": true,
+              "refId": "A"
+            }
+          ],
+          "title": "Completed Processes in avg overtime per sec",
```
lenaschoenburg, line 3134: :wrench: This and other panel titles in this row feel a bit verbose/confusing. It's clear that something is shown over time, IMO that doesn't need to be specified in the title. Also, the stat panel shows more than just the average so I'd suggest to remove that too.
I'd suggest to just use "Completed Processes" or if you think it's necessary "Completed Processes per second"

lenaschoenburg (approved):
Really nice work @Zelldon :star_struck: 

I have lots of comments, some marked as :x: but feel free to merge anyway as it clearly provides value already.

:x: There's many places where `increase` over the `$__rate_interval` is used. The result is misleading and we should always use `rate` instead. `increase` returns the total increase over the specified interval. But `$__rate_interval` is dynamic and not visible to the user, so they cannot interpret the value meaningfully. Using `rate` ensures that all values are _per second_ and is much more consistent and easier to understand.

lenaschoenburg:
One more thing:

:wrench: You might want to include the Scaling and Throttling annotations from the Zeebe dashboard because those are highly relevant to performance too.

Zelldon:
Thank you @lenaschoenburg for the thorough review! I appreciate this. ❤️  I will try to follow up on this on Monday. :)

monitor/grafana/camunda-performance.json
Zelldon: This is actually a copy from our Zeebe Dashboard, and came from previous observations where we had multiple services (multiple prometheus scraping data) causing to show wrong values. I don't think it hurts.

monitor/grafana/camunda-performance.json
Zelldon: Good catch, just a copy from the Zeebe dashboard :)

monitor/grafana/camunda-performance.json
Zelldon: Because it is an aggregration, if we don't specify the name we just get one value.

monitor/grafana/camunda-performance.json
Zelldon: See https://github.com/camunda/camunda/pull/22021#discussion_r1760870810

monitor/grafana/camunda-performance.json
lenaschoenburg: I see, I didn't realize the `avg` was on purpose. Usually we just use the rate directly and I though the `avg by (name)` was done to merge multiple timeseries.

monitor/grafana/camunda-performance.json
```diff
@@ -0,0 +1,3649 @@
...
+              "editorMode": "code",
+              "expr": "",
+              "hide": false,
+              "instant": false,
+              "legendFormat": "__auto",
+              "range": true,
+              "refId": "B"
+            }
+          ],
+          "title": "Number of records not imported",
```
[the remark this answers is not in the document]
Zelldon: I checked again yes it is. I created this as reminder for myself to add such metric 😅 I can also remove the panel for now

monitor/grafana/camunda-performance.json
Zelldon: Actually, we can't right now. 

This is on my agenda, but Operate currently only exports sum and count. They do not expose histograms :/ Weird thing with micrometer...

monitor/grafana/camunda-performance.json
Zelldon: Actually I removed this in a newer version already, but good point

monitor/grafana/camunda-performance.json
Zelldon: fixed

monitor/grafana/camunda-performance.json
Zelldon: fixed

Zelldon:
I iterated again over the dashboard (already after the PR creation due to my own usage, but also several things I applied based on your great Review. Thanks again @lenaschoenburg )

I would like to merge it make it public available so we can also reference it in SaaS. 

I will anyway continue to improve and iterate over it. Besides that I also improve a bit the Operate Dashboard #22201 




Screenshots <current-state>:


general
general-2
latency
latency-2
throughput

lenaschoenburg:
Did you consider the feedback regarding the use of `increase`? https://github.com/camunda/camunda/pull/22021#pullrequestreview-2286093840
> ❌ There's many places where increase over the $__rate_interval is used. The result is misleading and we should always use rate instead. increase returns the total increase over the specified interval. But $__rate_interval is dynamic and not visible to the user, so they cannot interpret the value meaningfully. Using rate ensures that all values are per second and is much more consistent and easier to understand.

Zelldon:
> Did you consider the feedback regarding the use of `increase`? [#22021 (review)](https://github.com/camunda/camunda/pull/22021#pullrequestreview-2286093840)



Thanks for the reminder. I missed that.

monitor/grafana/camunda-performance.json
Zelldon: Fixed

monitor/grafana/camunda-performance.json
Zelldon: I removed this

monitor/grafana/camunda-performance.json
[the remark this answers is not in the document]
Zelldon: I removed this

PR #22022, opened 2024-09-06
ci: allow to trigger realistic benchmarks from GHA

ChrisKujawa:
## Description

Reference the values file from the benchmark repo to deploy a realistic benchmark

\cc @rodrigo-lourenco-lopes likely useful for your size testing.

## Related issues

related to https://github.com/camunda/camunda/issues/21462

### DIFF
--- a/.github/workflows/zeebe-benchmark.yml
+++ b/.github/workflows/zeebe-benchmark.yml
@@ -70,11 +70,6 @@ on:
         description: 'Specifies which benchmark components to deploy. `starter`, `timer` and `publisher` can be assigned with the rate at which they publish. Allows arbitrary helm arguments, like --set starter.rate=100'
         type: string
         required: false
-      measure:
-        description: 'Measure impact of network latency'
-        type: boolean
-        required: false
-        default: true
       publish:
         description: 'Where to publish the results, can be "slack" or "comment"'
         default: ""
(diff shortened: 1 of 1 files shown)


ChrisKujawa:
@npepinpe previously we had a long running GKE, which is why we had this config:

```
      cluster:
        description: 'Specifies which cluster to deploy the benchmark on'
        default: 'zeebe-cluster'
        required: false
      cluster-region:
        description: 'Specifies the cluster region. Needed to retrieve cluster credentials'
        default: europe-west1-b
        required: false
```

Can we get rid of this? This would free up two options, and we could add the realistic option.

npepinpe:
I rebased on main :+1: 

Re your question, I would leave the cluster and cluster-region. Reason for this is we will slowly move to the infra maintained GKE, and while this is happening, it may be useful to support changing these values.

Once we're fully off of the zeebe-cluster and we decide what to do with it (e.g. kill it), we can remove it.

npepinpe:
Ah, I see, it's blocking. OK, let's remove it, we will figure out something :smile:

npepinpe (approved):

npepinpe:
Test run: https://github.com/camunda/camunda/actions/runs/14463312155

npepinpe:
How is this making Tasklist fail? I suspect there's something funky with the Tasklist tests...

PR #22023, opened 2024-09-06
Migrate error boundary events

### LINKED ISSUE #21866: Migrate error boundary events
## Description
It should be possible to migrate tasks with error boundary events. Error catch event doesn't have a subscription record but its **static** error code stored in the catch event element itself. When an error is thrown, the error catch event is accessed through the process instance itself.

Given above, there is no need (or no way) to introduce a migration intent for error catch event. Consider the following example, where an active service tasks has an error boundary event attached. When migrating the process instance the error boundary event will be migrated to the target process **regardless of mapping instruction is provided for the error boundary element or not**. After the instance is migrated, **the error code will be updated** as in the target process since it is static value.

Screenshot 2024-09-06 at 13 23 27

Two things needs to be implemented:
- Add `BpmnEventType.ERROR` to supported boundary element types
- Verify a task can be migrated with a error boundary event

berkaycanbc:
## Description
It should be possible to migrate tasks with error boundary events. Error catch event doesn't have a subscription record but its **static** error code stored in the catch event element itself. When an error is thrown, the error catch event is accessed through the process instance itself.

Given above, there is no need (or no way) to introduce a migration intent for error catch event. Consider the following example, where an active service tasks has an error boundary event attached. When migrating the process instance the error boundary event will be migrated to the target process **regardless of mapping instruction is provided for the error boundary element or not**. After the instance is migrated, **the error code will be updated** as in the target process since it is static value.

Screenshot 2024-09-06 at 13 23 27

Two things needs to be implemented:
- Add `BpmnEventType.ERROR` to supported boundary element types
- Verify a task can be migrated with a error boundary event

## Related issues

closes #21866

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -214,12 +214,14 @@ public class ProcessInstanceMigrationMigrateProcessor
     requireNoBoundaryEventInSource(
         sourceProcessDefinition,
         elementInstanceRecord,
-        EnumSet.of(BpmnEventType.MESSAGE, BpmnEventType.TIMER, BpmnEventType.SIGNAL));
+        EnumSet.of(
+            BpmnEventType.MESSAGE, BpmnEventType.TIMER, BpmnEventType.SIGNAL, BpmnEventType.ERROR));
     requireNoBoundaryEventInTarget(
         targetProcessDefinition,
         targetElementId,
         elementInstanceRecord,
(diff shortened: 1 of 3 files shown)


koevskinikola (approved):
LGTM @berkaycanbc 👍

koevskinikola (changes requested):
@berkaycanbc I just checked the CI results.

The `MigrateProcessInstanceRejectionTest.shouldRejectCommandWhenTheMigratedProcessInstanceHasATaskSubscribedToAnErrorBoundaryEvent` is failing, I'm guessing because error boundary events can be migrated now. 

So I guess we need to remove this test now.

berkaycanbc:
@koevskinikola yes, missed that one. Removed the test now. ✅

koevskinikola (approved):
❌ @berkaycanbc there seems to be some more "migration rejection" test failures and some format violations as well.

However, it's clear that the main functionality of this PR is correct, so I'm pre-approving it. ✅

PR #22030, opened 2024-09-06
feat: Add validation for non-negative value in `pin` clock command

### LINKED ISSUE #21964: Add validation for pinning time to ensure non-negative values
###Description:

To ensure that time manipulation commands in the Zeebe system are safe and avoid causing instability, we need to validate the values provided when pinning the clock. Specifically, the `pin` clock command should not allow negative timestamps or instants before the Unix epoch `1970-01-01T00:00:00Z`. This task involves adding validation to both the REST endpoint and the Zeebe client to enforce this rule and prevent issues such as repeated errors or system crashes observed durning manual QA testing with negative timestamp values.

1. Add validation for non-negative time values in `ClockProcessor` for `PIN` command.
   - If a negative time is provided, reject this command with appropriate rejection message: 
      `"Expected pin time to be not negative but it was %d"`

2. Add validation for non-negative timestamp values in the `PUT /v2/administration/clock` REST endpoint:
   - Ensure that when the clock is pinned using a timestamp, the value must be **greater than or equal to 0** .
   - If a negative timestamp is provided, return an appropriate error message: 
     `"The value for timestamp is '%d' but must be non-negative."`.

3. Update `ZeebeClient.newClockPinCommand().time(long timestamp)`:
   - Add a validation that ensures the `timestamp` is **non-negative**.
   - If a negative value is provided, throw an `IllegalArgumentException` with the message:
     `"timestamp must be not negative"`.

4. Update `ZeebeClient.newClockPinCommand().time(Instant instant)`:
   - Add a validation to check that the provided `Instant` is **not before `Instant.EPOCH`**.
   - If the `Instant` is before the epoch, throw an `IllegalArgumentException` with the message:
     `"instant must be equal to or after '1970-01-01T00:00:00Z'"`

ce-dmelnych:
## Description

This PR adds validations for clock pinning commands to ensure the use of valid non-negative time values. Negative timestamps and instants before the Unix epoch (`1970-01-01T00:00:00Z`) will now be rejected in both the REST endpoint and the Zeebe client.

- Validation added in `ClockProcessor` to reject PIN command with negative pin time
- Added validation to `PUT /v2/administration/clock` REST endpoint to accept only non-negative timestamps
- Updated `ClockPinCommandImpl#time(long timestamp)` to ensure non-negative timestamps
- Updated `ClockPinCommandImpl#time(Instant instant)` to ensure that the `Instant` is not before `Instant.EPOCH`

## Checklist

- [ ] for CI changes:

## Related issues

closes #21964

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ArgumentUtil.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ArgumentUtil.java
@@ -51,9 +51,7 @@ public final class ArgumentUtil {
   }
 
   public static void ensureNotNegative(final String property, final Duration testValue) {
-    if (testValue.isNegative()) {
-      throw new IllegalArgumentException(String.format("%s must be not negative", property));
-    }
+    ensureNotNegative(property, testValue.toMillis());
   }
 
   public static void ensureNotZero(final String property, final Duration testValue) {
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ArgumentUtil.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ArgumentUtil.java
@@ -16,6 +16,7 @@
 package io.camunda.zeebe.client.impl.command;
 
 import java.time.Duration;
+import java.time.Instant;
 
 public final class ArgumentUtil {
 
@@ -43,6 +44,12 @@ public final class ArgumentUtil {
     }
   }
 
+  public static void ensureNotNegative(final String property, final long testValue) {
+    if (testValue < 0) {
+      throw new IllegalArgumentException(String.format("%s must be not negative", property));
+    }
+  }
+
   public static void ensureNotNegative(final String property, final Duration testValue) {
     if (testValue.isNegative()) {
       throw new IllegalArgumentException(String.format("%s must be not negative", property));
@@ -58,5 +65,13 @@ public final class ArgumentUtil {
   public static void ensurePositive(final String property, final Duration testValue) {
     ensureNotNegative(property, testValue);
     ensureNotZero(property, testValue);
+  }
+
+  public static void ensureNotBefore(
+      final String property, final Instant testValue, final Instant otherInstant) {
+    if (testValue.isBefore(otherInstant)) {
+      throw new IllegalArgumentException(
+          String.format("%s must be equal to or after %s", property, otherInstant));
+    }
   }
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockProcessor.java
@@ -49,10 +49,8 @@ public final class ClockProcessor implements DistributedTypedRecordProcessor<Clo
 
   @Override
   public void processNewCommand(final TypedRecord<ClockRecord> command) {
-    final long eventKey = keyGenerator.nextKey();
-    final var clockRecord = command.getValue();
     final var intent = (ClockIntent) command.getIntent();
-    final var resultIntent = followUpIntent(intent);
+    final var clockRecord = command.getValue();
 
     if (intent == ClockIntent.PIN && clockRecord.getTime() < 0) {
       final var rejectionMessage =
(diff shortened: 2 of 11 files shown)


clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ArgumentUtil.java
```diff
@@ -43,6 +44,12 @@ public static void ensureGreaterThan(
     }
   }
 
+  public static void ensureNotNegative(final String property, final long testValue) {
```
nicpuppa, line 47: 🔧 Can you refactor this method and the one below ? They are basically doing the same thing

```
public static void ensureNotNegative(final String property, final long testValue) {
  ........
}

public static void ensureNotNegative(final String property, final Duration testValue) {
    ensureNotNegative(property, testValue.getSeconds())
  }
```
💭 Using `testValue.toMillis()` should be fine as well 🤔

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockProcessor.java
```diff
@@ -50,6 +54,16 @@ public void processNewCommand(final TypedRecord<ClockRecord> command) {
     final var intent = (ClockIntent) command.getIntent();
     final var resultIntent = followUpIntent(intent);
 
+    if (intent == ClockIntent.PIN && clockRecord.getTime() < 0) {
```
nicpuppa, line 57: 🔧 You can move this validation check some lines above, you need only `clockRecord` and `intent` as variable, you can define the others variable after this check

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -603,7 +603,7 @@ paths:
           description: >
             The clock was successfully pinned to the specified time in epoch milliseconds.
         "400":
-          description: The required timestamp parameter is missing.
+          description: The required timestamp parameter is missing or if a negative timestamp value is provided.
```
nicpuppa, line 606:
```suggestion
          description: The required timestamp parameter is missing or it is negative.
```

❌ Can you also add in `ClockPinRequest` schema, under `timestamp` field, the property `minimun: 0` ?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/ClockValidator.java
```diff
@@ -19,8 +20,13 @@ public class ClockValidator {
   public static Optional<ProblemDetail> validateClockPinRequest(final ClockPinRequest pinRequest) {
     return validate(
         violations -> {
-          if (pinRequest.getTimestamp() == null) {
+          final Long timestamp = pinRequest.getTimestamp();
```
nicpuppa, line 23: 🔧 To avoid if/else statement you can refactor the validation like this:

```
Optional.ofNullable(pinRequest.getTimestamp())
            .ifPresentOrElse(
                timestamp -> {
                    if (timestamp < 0) {
                        violations.add(ERROR_MESSAGE_INVALID_ATTRIBUTE_VALUE
                            .formatted("timestamp", timestamp, "not negative"));
                    }
                },
                () -> violations.add(ERROR_MESSAGE_EMPTY_ATTRIBUTE.formatted("timestamp"))
            );
```

nicpuppa (changes requested):
Hey @ce-dmelnych looks great 🚀 

I have some small suggestions and also a requested change, please have a look 👀

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/ArgumentUtil.java
ce-dmelnych: Sure, update to reuse the code!

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/clock/ClockProcessor.java
ce-dmelnych: Yeah, moved validation upper

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
ce-dmelnych: Wording improved, added minimum prop for `ClockPinRequest#timestamp` field.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/validator/ClockValidator.java
ce-dmelnych: Nice suggestion!
Refactored this validation check.

ce-dmelnych:
@nicpuppa, thank you for your suggestions! The improvements have been applied.

nicpuppa (approved):
Thanks @ce-dmelnych

PR #22035, opened 2024-09-06
Migrate error event subprocess

### LINKED ISSUE #22032: Migrate error event subprocess
## Description

### Migrate active error event subprocess

Screenshot 2024-09-10 at 16 04 08

Mappings should be provided between event subprocesses (sub1 -> sub2) and active elements inside event subprocesses (B -> B). 

### Migrate error event subprocess start event subscription

Screenshot 2024-09-10 at 16 02 05

To migrate the start event subscription of a error event subprocess a mapping needs to be provided for active elements (A-> A) and between error start events (start1 -> start2). This case currently doesn't provide much advantage since error code will not change during the migration, but will be useful when we introduce the feature for re-evaluating error codes.

### Subscribe to error event subprocess start event subscription

Screenshot 2024-09-10 at 16 00 04

In this case, a mapping should be provided for the active element (A -> A). The error event subscription will be created after the migration for error start event (start2).

### Unsubscribe from error event subprocess start event subscription

Screenshot 2024-09-10 at 15 58 13

In this case, a mapping should be provided for the active element (A -> A). The error event subscription will be deleted after the migration for error start event (start1).

berkaycanbc:
## Description

### Migrate active error event subprocess

Screenshot 2024-09-10 at 16 04 08

Mappings should be provided between event subprocesses (sub1 -> sub2) and active elements inside event subprocesses (B -> B). 

### Migrate error event subprocess start event subscription

Screenshot 2024-09-10 at 16 02 05

To migrate the start event subscription of a error event subprocess a mapping needs to be provided for active elements (A-> A) and between error start events (start1 -> start2). This case currently doesn't provide much advantage since error code will not change during the migration, but will be useful when we introduce the feature for re-evaluating error codes.

### Subscribe to error event subprocess start event subscription

Screenshot 2024-09-10 at 16 00 04

In this case, a mapping should be provided for the active element (A -> A). The error event subscription will be created after the migration for error start event (start2).

### Unsubscribe from error event subprocess start event subscription

Screenshot 2024-09-10 at 15 58 13

In this case, a mapping should be provided for the active element (A -> A). The error event subscription will be deleted after the migration for error start event (start1).

## Related issues

closes #22032

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateErrorEventSubprocessTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateErrorEventSubprocessTest.java
@@ -1,0 +1,351 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.processinstance.migration;
+
+import static io.camunda.zeebe.engine.processing.processinstance.migration.MigrationTestUtil.extractProcessDefinitionKeyByProcessId;
+import static io.camunda.zeebe.protocol.record.Assertions.assertThat;
+
+import io.camunda.zeebe.engine.util.EngineRule;
+import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.protocol.record.Assertions;
+import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
+import io.camunda.zeebe.protocol.record.intent.JobIntent;
+import io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent;
+import io.camunda.zeebe.protocol.record.value.BpmnElementType;
+import io.camunda.zeebe.protocol.record.value.ErrorType;
+import io.camunda.zeebe.test.util.BrokerClassRuleHelper;
+import io.camunda.zeebe.test.util.record.RecordingExporter;
+import io.camunda.zeebe.test.util.record.RecordingExporterTestWatcher;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestWatcher;
+
+public class MigrateErrorEventSubprocessTest {
+
+  @ClassRule public static final EngineRule ENGINE = EngineRule.singlePartition();
+
+  @Rule public final TestWatcher watcher = new RecordingExporterTestWatcher();
+  @Rule public final BrokerClassRuleHelper helper = new BrokerClassRuleHelper();
+
+  @Test
+  public void shouldMigrateActiveErrorEventSubprocess() {
+    // given
+    final String processId = helper.getBpmnProcessId();
+    final String targetProcessId = helper.getBpmnProcessId() + "2";
+    final var deployment =
+        ENGINE
+            .deployment()
+            .withXmlResource(
+                Bpmn.createExecutableProcess(processId)
+                    .eventSubProcess(
+                        "sub1",
+                        s ->
+                            s.startEvent("start1").error("error1").userTask("userTask1").endEvent())
+                    .startEvent("start")
+                    .serviceTask("A", t -> t.zeebeJobType("A"))
+                    .endEvent("end")
+                    .done())
+            .withXmlResource(
+                Bpmn.createExecutableProcess(targetProcessId)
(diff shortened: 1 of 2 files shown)


korthout:
PR not yet ready for review @berkaycanbc

berkaycanbc:
> PR not yet ready for review @berkaycanbc

Converted to draft, I will fix it tomorrow morning 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateErrorEventSubprocessTest.java
```diff
@@ -0,0 +1,351 @@
...
+        .hasProcessDefinitionKey(targetProcessDefinitionKey)
+        .describedAs("Expect that bpmn process id and element id changed")
+        .hasBpmnProcessId(targetProcessId)
+        .hasElementId("B")
+        .describedAs("Expect that version number did not change")
+        .hasVersion(1);
+
+    // note that the error code is changed since the error code is a static value set inside the
+    // element itself
+    ENGINE.job().ofInstance(processInstanceKey).withType("A").withErrorCode("error2").throwError();
```
korthout, line 188: 🔧 We need to clarify this to users to avoid surprises, I suggest one of:
- document the different behavior of errors compared to other event types like messages (static message name changes are left unchanged)
  - this kinda binds us to this behavior, as it may be hard to change later
- open a bug report to describe the different (unexpected) behavior
- open a bug report and highlight it in the docs

korthout (approved):
Nice work @berkaycanbc

🔧 We need to consider that the behavior for mapped error catch events differs from that of mapped catch events of other event types. Please open an issue to ensure we document this clearly.

👍 LGTM

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateErrorEventSubprocessTest.java
berkaycanbc: I would go with the first option as we plan to support re-evaluating expressions on other elements.

berkaycanbc:
@korthout created the related issue https://github.com/camunda/camunda/issues/22180.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateErrorEventSubprocessTest.java
korthout: >I would go with the first option as we plan to support re-evaluating expressions on other elements.

@berkaycanbc How do you see that working?

Re-evaluating expressions will be something the user has to enable through the API (something like `.addMappingInstruction("source", "target", i -> i.evaluateMessageNameExpression())`). The default behavior for errors still differs from other event types. I've asked [here](https://camunda.slack.com/archives/CSQ2E3BT4/p1726043974128849) to see if someone knows why we don't just have error subscriptions.

berkaycanbc: I concluded this one as I assumed re-evaluating expressions will be the default behaviour unless user specifies the otherwise. If it won't be the default behaviour, I think, we should fix it in this iteration to keep user experience consistent across element migrations.

korthout: > I assumed re-evaluating expressions will be the default behaviour unless user specifies the otherwise

We've already [documented the behavior](https://docs.camunda.io/docs/components/concepts/process-instance-migration/#jobs-expressions-and-input-mappings), we can't/shouldn't change the behavior in a breaking manner. So, adding re-evaluation of expression will be something users have to specifically choose.

korthout: Re-reading that docs, we could expand upon it clarifying that this only concerns the parts of the element that have already been used. Output mappings, and End-type execution listeners will use the new process definition.

korthout: >I've asked [here](https://camunda.slack.com/archives/CSQ2E3BT4/p1726043974128849) to see if someone knows why we don't just have error subscriptions.

If we have a good reason that errors are truly different. Then, we can just document it as a special behavior to errors. Otherwise, I suggest we consider this a bug.

berkaycanbc: Created a related issue here https://github.com/camunda/camunda/issues/22180

PR #22037, opened 2024-09-06
docs: bring up downstream changes found in C8 REST API's generated documentation

pepopowitz:
## Description

This PR addresses many small C8 REST API spec issues identified in https://github.com/camunda/camunda-docs/pull/4217.

See that PR for original conversation on these changes. The changes can be categorized as:

* Case consistency
* Shorter path `summary`
* Missing `description`

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -565,7 +565,7 @@ paths:
                 $ref: "#/components/schemas/UserTaskSearchQueryResponse"
         "400":
           description: >
-            The user task search Query failed.
+            The user task search query failed.
             More details are provided in the response body.
           content:
             application/problem+json:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -558,14 +558,14 @@ paths:
       responses:
         "200":
           description: >
-            The User Task Search successful response.
+            The user task search successful response.
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/UserTaskSearchQueryResponse"
         "400":
           description: >
-            The User Task Search Query failed.
+            The user task search Query failed.
             More details are provided in the response body.
           content:
             application/problem+json:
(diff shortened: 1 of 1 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -558,14 +558,14 @@ paths:
...
-            The User Task Search successful response.
+            The user task search successful response.
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/UserTaskSearchQueryResponse"
         "400":
           description: >
-            The User Task Search Query failed.
+            The user task search Query failed.
```
tmetzke, line 568: ❓ Should `Query` also be lowercase here? I can see the other query endpoints also use uppercase but I was wondering why that is. In @christinaausley's [original commit](https://github.com/camunda/camunda-docs/pull/4178/commits/45b9e61a03d355cde8de48139f1f1e273db7000a), it was all lowercase.

tmetzke (approved):
👍 Love it, thanks for feeding this back! 🚀 
❓ There is one question left open from my end otherwise all changes look good.

❓ Understanding question: there are a couple of similar issues still around for other endpoints, especially other search endpoints. Are you addressing those with the original authors or how do we handle them?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
pepopowitz: Yes, this is a mistake on my part! Fixing.

pepopowitz, line 568:
```suggestion
            The user task search query failed.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1245,14 +1245,14 @@ paths:
...
-            The Incident Search successful response.
+            The incident search successful response.
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/IncidentSearchQueryResponse"
         '400':
           description: >
-            The Incident Search Query failed.
+            The incident search Query failed.
```
pepopowitz, line 1255:
```suggestion
            The incident search query failed.
```

❌ Oops! Another case issue I missed.

pepopowitz:
> ❓ Understanding question: there are a couple of similar issues still around for other endpoints, especially other search endpoints. Are you addressing those with the original authors or how do we handle them?

My intention with this PR was to capture exactly what had been noticed in https://github.com/camunda/camunda-docs/pull/4217, so that I didn't end up down a rabbit hole fixing more issues. 

We should probably do a sweep of the spec for issues that we missed before we were thoroughly reviewing from the DevEx side. I'll bring that to our team meeting next week to talk about when/how we will approach it. What do you think?

tmetzke:
> My intention with this PR was to capture exactly what had been noticed in [camunda/camunda-docs#4217](https://github.com/camunda/camunda-docs/pull/4217), so that I didn't end up down a rabbit hole fixing more issues.
> 
> We should probably do a sweep of the spec for issues that we missed before we were thoroughly reviewing from the DevEx side. I'll bring that to our DevEx team meeting next week to talk about when/how we will approach it. What do you think?

That makes sense, yes, I understand your concern 😅

tmetzke:
It's a GO from my end 👍

PR #22042, opened 2024-09-06
test: make integration tests resilient to retries

houssain-barouni:
## Description

Fixes issues observed when test retries are triggered at the test class level with Maven Surefire configuration. The static resources (`TestApplication` or `TestCluster` annotated with `@TestZeebe`) in the test classes are not reinitialized after a retry. As a result, certain configurations, such as the Raft working directory, were not reset, leading to unexpected data from previous test runs being loaded, which caused test failures.

### Reproduction Steps:
To reproduce this issue, configure a test class (e.g., `DecisionQueryTest`) to run multiple times in your IDE. Tests begin to fail after the second retry due to stale data in Zeebe from previous runs.

### Changes in This Pull Request:
1. **Added `initMethod()` attribute in `@TestZeebe`:**
   - Introduced a new attribute `initMethod()` in the `@TestZeebe` annotation, allowing the specification of a method to initialize static resources.
   - If `initMethod` is defined, it will be invoked via reflection before the test class execution. This ensures static resources like `TestApplication` and `TestCluster` are properly reinitialized on each test class retry, avoiding issues with stale data.

   **Example:**
   ```java
   @TestZeebe(initMethod = "initTestCluster")
   private static TestCluster cluster;

   static void initTestCluster() {
     cluster = TestCluster.builder().withGatewaysCount(2).withEmbeddedGateway(false).build();
   }
   ```

2. **Stopping ES Container `TestStandaloneCamunda`:**
   - Ensured that the Elasticsearch container is stopped after all tests requiring this resource have been executed. Each test class using a static `TestStandaloneCamunda` instance has its own ES container. Stopping the container after the test calls prevents multiple ES containers from unnecessarily running after tests.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


qa/integration-tests/src/test/java/io/camunda/it/client/DecisionQueryTest.java
```diff
@@ -34,21 +34,27 @@
...
   private static ZeebeClient zeebeClient;
 
-  @TestZeebe
-  private static TestStandaloneCamunda testStandaloneCamunda = new TestStandaloneCamunda();
+  @TestZeebe(initMethod = "initTestStandaloneCamunda")
+  private static TestStandaloneCamunda testStandaloneCamunda;
+
+  @SuppressWarnings("unused")
+  static void initTestStandaloneCamunda() {
+    testStandaloneCamunda = new TestStandaloneCamunda();
```
mihail-ca, line 53: I haven't tested, but wouldn't it work if we just initialized `testStandaloneCamunda` in `beforeAll` method instead ?

houssain-barouni: No, I already tried it but it does not work.
`ZeebeIntegrationExtension.beforeAll` is invoked before an expects to find a value for the fields annotated with `@TestZeebe`

ghost (approved):

PR #22054, opened 2024-09-08
deps: exclude jetty depenendencies from spring boot

RomanJRW:
## Description

e2e and ITs are failing since the upgrade to Jetty 12.0.13. The reason is because the Spring Boot Jetty library still pulls in Jetty 12.0.12 dependencies

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/backend/src/it/java/io/camunda/optimize/test/it/extension/ZeebeExtension.java
+++ b/optimize/backend/src/it/java/io/camunda/optimize/test/it/extension/ZeebeExtension.java
@@ -283,7 +283,9 @@ public class ZeebeExtension implements BeforeEachCallback, AfterEachCallback {
   }
 
   private void destroyClient() {
-    zeebeClient.close();
-    zeebeClient = null;
+    if (zeebeClient != null) {
+      zeebeClient.close();
+      zeebeClient = null;
+    }
   }
 }
--- a/optimize/pom.xml
+++ b/optimize/pom.xml
@@ -35,6 +35,7 @@
     <!-- https://www.elastic.co/guide/en/elasticsearch/client/java-api/6.5/_using_another_logger.html -->
     <elasticsearch.log4j.version>2.24.0</elasticsearch.log4j.version>
 
+    <jetty.version>12.0.13</jetty.version>
     <jackson.version>2.17.2</jackson.version>
     <jsonpath.version>2.9.0</jsonpath.version>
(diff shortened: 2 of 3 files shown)


optimize/pom.xml
```diff
@@ -108,6 +109,56 @@
...
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-rewrite</artifactId>
+        <version>${jetty.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-util</artifactId>
+        <version>${jetty.version}</version>
+      </dependency>
```
RomanJRW, line 156: Note that this is a newly declared dependency. Ultimately, we were getting 12.0.12 from Spring, and the changes in this [Jetty commit](https://github.com/jetty/jetty.project/commit/1726c8778c0c1a2040d5aafeb265c59ad927e599#diff-642ee93cf6ea50eebacf1719c77826adeb4ccc4c2f87570d3d348fe4ef026c24) meant that our ITs were breaking when trying to setup the connection to Zeebe. This is one of the fixes in this PR, to make sure that the util version we use matches that of the actual jetty version we use

optimize/backend/src/it/java/io/camunda/optimize/test/it/extension/ZeebeExtension.java
```diff
@@ -283,7 +283,9 @@ private void setZeebeRecordPrefixForTest() {
   }
 
   private void destroyClient() {
-    zeebeClient.close();
-    zeebeClient = null;
+    if (zeebeClient != null) {
+      zeebeClient.close();
+      zeebeClient = null;
+    }
```
RomanJRW, line 289: I noticed that we were trying to close the client, even in the case when tests weren't set up correctly in the first place. While it wasn't really hurting us, this is just a simple check to avoid the NPE case

optimize/pom.xml
```diff
@@ -108,6 +109,56 @@
         <scope>import</scope>
         <type>pom</type>
       </dependency>
+      <dependency>
```
RomanJRW, line 112: this is the main fix, to move jetty dependencies to the parent dependency management. This allows us to declare the jetty version explicitly and not have spring jetty take precedence

HeleneW-dot (approved):

PR #22061, opened 2024-09-09
fix: check for isEmbeddedForm to handle customFormKey user tasks

### LINKED ISSUE #21811: Job Worker User Task with `Custom Form Key` not being handled properly
### Describe the bug

When creating a Job Worker UserTask with a `Custom Form Key` reference, accessing the UserTask via Tasklist UI (even though this is not the normal use case for this type of UserTask forms), Tasklist backend tries to retrieve the Form unsuccessfully returning 404 to the UI client. 

The user can `Assign/Unassign` the User Task normally, however the `Complete` operation does not conclude, instead it waits for the action to timeout, without displaying any information. Also, no variables are allowed to be added in Tasklist.

In contrast, for Zeebe User Tasks with `External Form Reference` :
- Tasklist UI is **not** trying to fetch the Form
- Variables can be added in the User Task
- Task can be completed normally from Tasklist

### Additional context

Support Issue: https://jira.camunda.com/browse/SUPPORT-23394
Slack Discussion Thread: https://camunda.slack.com/archives/C046XS3GGCR/p1725016863154569

### Affected Versions
- 8.6 
- 8.5

panagiotisgts:
## Description
Adjust the condition to set the `isFormEmbedded` flag for User Tasks to match the pattern set by Zeebe for such forms `^camunda-forms:bpmn:.*` .

closes #21811

### DIFF
--- a/tasklist/client/src/Tasks/Task/index.tsx
+++ b/tasklist/client/src/Tasks/Task/index.tsx
@@ -156,7 +156,7 @@ const Task: React.FC = observer(() => {
   }
 
   const isDeployedForm = typeof formId === 'string';
-  const isEmbeddedForm = typeof formKey === 'string' && !isDeployedForm;
+  const isEmbeddedForm = typeof formKey === 'string' && task.isFormEmbedded;
   if (isEmbeddedForm || isDeployedForm) {
     return (
       <FormJS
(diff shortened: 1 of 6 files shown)


houssain-barouni (changes requested):
LGTM! Can you add integration tests with a bpmn with custom form, to make sure this does not break in the future?

houssain-barouni (approved):
Thanks!

PR #22063, opened 2024-09-09
fix: ensure access to raft role is thread safe

### LINKED ISSUE #21759: io.atomix.raft.ReconfigurationTest$Joining.shouldRequireAdjustedQuorum is Flaky
https://github.com/camunda/camunda/actions/runs/10450168824
https://github.com/camunda/camunda/actions/runs/10451253160/job/28937304725

[Test results for unit tests (1).zip](https://github.com/user-attachments/files/16658554/Test.results.for.unit.tests.1.zip)

_Originally posted by @nicpuppa in https://github.com/camunda/camunda/issues/18054#issuecomment-2296130567_

deepthidevaki:
## Description

The flaky test in #21759 failed because it expectes there is no leader with in a specific timeout. But it found a leader.  However, the logs indicate that there was no leader. 

```
08:59:43.737 [raft-server-0] WARN  io.atomix.raft.roles.LeaderAppender - RaftServer{3} - Suspected network partition after 1008 failures from 5 over a period of time 5038 > 1000, stepping down
08:59:43.737 [raft-server-0] TRACE io.atomix.raft.impl.RaftContext - RaftServer{3} - Set leader null
08:59:43.737 [raft-server-0] INFO  io.atomix.raft.impl.RaftContext - RaftServer{3} - Transitioning to FOLLOWER
```

After this we see several PollRequest failures because there is no quorum. So there was no leader elected after this point.

I suspect the test failed because it checks the role in RaftContext to determine the current leader. It is not accessed or updated in a thread-safe manner resulting in a race-condition where the test always sees the old state. 

We are also exposing the RaftContext#getRaftRole in production code which is accessed by other threads. To make it thread-safe, the update and external access to the role is now protected via a `synchronized` block. Internal access to role can be left unprotected because, all access goes through raft threadcontext. That is why there is no need to make it a volatile variable.

## Related issues

closes #21759

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
@@ -130,6 +130,9 @@ public class RaftContext implements AutoCloseable, HealthMonitorable {
   private final ReceivableSnapshotStore persistedSnapshotStore;
   private final LogCompactor logCompactor;
   private volatile State state = State.ACTIVE;
+  // Some fields are read by external threads. To ensure thread-safe access, we can use the lock for
+  // synchronizing write and reads on such fields.
+  private final Object externalAccessLock = new Object();
   private RaftRole role = new InactiveRole(this);
   private volatile MemberId leader;
   private volatile long term;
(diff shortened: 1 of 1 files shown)


lenaschoenburg (commented):
To be honest I don't really understand how adding these two `synchronized` blocks helps. Both reading from and writing to a field reference should be atomic. If a concurrent thread reads it, it either gets the old or the new value. Adding a `synchronized` to this doesn't really change it, right?

deepthidevaki:
Ah.. right :see_no_evil: Since the `role` gets overwritten, there is no point in synchronizing over it. So we should "lock" on a different object.

lenaschoenburg:
But why do we need to lock at all? If I understand the issue correctly, we only need to make sure that some concurrent thread is able to read the current role. But that should already be the case with the code we have.

deepthidevaki:
> But why do we need to lock at all? If I understand the issue correctly, we only need to make sure that some concurrent thread is able to read the current role. But that should already be the case with the code we have.

Since `role` is not a volatile field, there is no guarantee when a write to it are visible to other threads. Wrapping access to it via a synchronized block or a lock guarantees the reading threads always see the latest the state due to happens-before ordering guarantees.  Alternate it is to make `role` a volatile fields which could be an overhead because it is accessed frequently in raft-thread.

lenaschoenburg (approved):
Thanks!

PR #22071, opened 2024-09-09
Schema manager exporter integration

### LINKED ISSUE #21956: Schema manager initialises and updates schemas on exporter open
**Description**

- [ ] On exporter open, schema manager is invoked to create or update the schema
- [ ] Provide configuration to enable/disable schema creation on startup

EuroLew:
## Description

When the exporter opens it will 1 validate the new schemas represented by descriptors and make sure all mapping changes are valid. If not it will throw an exception.

If changes to the schemas are valid it will align the search engine schemas with the descriptors by
- creating any schemas which don't exist
- updating any schemas which do exist

Also added a `createSchema` field to the index which if true allows schemas to be mutable and if false forbids any writing.

## Related issues

closes #21956


deepthidevaki (commented):
Looks good to me. 

:wrench: I was wondering if the IT tests should be split into two - one for schema management test and one for other tests, especially considering the merge conflicts with the main.

EuroLew:
> Looks good to me.
> 
> 🔧 I was wondering if the IT tests should be split into two - one for schema management test and one for other tests, especially considering the merge conflicts with the main.

It's a an issue of the exporter doing a lot (in this case both exporting records and schema initialisation), ideally all tests related to the exporter is in one class, I do agree with you on the delineation between exporting records/schema management.

For now I think it's ok but if it balloons with more use cases in the future then maybe then we split.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -66,7 +79,11 @@ public void configure(final Context context) {
   public void open(final Controller controller) {
     this.controller = controller;
     client = createClient();
-    // TODO createSchemaWhenNeeded();
+    searchEngineClient = new ElasticsearchEngineClient(client);
+    schemaManager = createSchemaManager();
+    schemaValidator = new IndexSchemaValidator(schemaManager);
```
deepthidevaki, line 84: :wrench: No need to keep them as the members of the class. You can directly pass it to `schemaStartup`.

deepthidevaki (approved):
Thanks.

PR #22137, opened 2024-09-10
Add description to AuthorizationPatchRequest properties

remcowesterhoud:
## Description

These properties didn't have a description. Following our guidelines we should make sure all properties have a description.

## Checklist

- [ ] for CI changes:

## Related issues

N/A

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -585,8 +585,8 @@ paths:
         - Clock
       summary: Pin internal clock (experimental)
       description: |
-        Set a precise, static time for the Zeebe engine’s internal clock. 
-        When the clock is pinned, it remains at the specified time and does not advance. 
+        Set a precise, static time for the Zeebe engine’s internal clock.
+        When the clock is pinned, it remains at the specified time and does not advance.
         To change the time, the clock must be pinned again with a new timestamp.
 
         :::note
@@ -620,7 +620,7 @@ paths:
         - Clock
       summary: Reset internal clock (experimental)
       description: |
-        Resets the Zeebe engine’s internal clock to the current system time, enabling it to tick in real-time. 
+        Resets the Zeebe engine’s internal clock to the current system time, enabling it to tick in real-time.
         This operation is useful for returning the clock to
         normal behavior after it has been pinned to a specific time.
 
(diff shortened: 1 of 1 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1852,16 +1852,19 @@ components:
...
               permissionType:
+                description: Specifies the type of permissions.
                 enum:
                   - CREATE
                   - READ
                   - UPDATE
                   - DELETE
               resourceIds:
                 type: array
+                description: A list of resource ids the permission relates to.
```
pepopowitz, line 1867:
```suggestion
                description: A list of resource IDs the permission relates to.
```

💭 I can't find anything in our writing style guide, but @camunda/tech-writers it seems to me that we probably have a standard for the plural form of "IDs"?

pepopowitz (approved):
👍 Only a very small question on the proper capitalization of IDs, which I've tagged the technical writers on.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
remcowesterhoud: I actually capitalized it first and removed it due to this part of the guidelines:

Follow these rules for casing:
Use "Sentence case" (with no period) for path tags properties.
Use "Sentence case" (with no period) for path summary properties.
Use "Sentence case." (with a period) for all other descriptive text.
Use "lower case" to reference API resources in descriptive text.
Example: use "The decision definition search query failed." instead of "The Decision Definition Search Query failed."

pepopowitz: ah, thank you for mentioning that! Let's see what the technical writers think about this, and then I'll work with them to update that guidance to be more clear.

conceptualshark: My preference here would be `IDs`, but I don't think we have a consensus on this yet (it's currently [up for discussion](https://github.com/camunda/developer-experience/issues/197)). As long as we're consistent within this PR/in general for the spec changes, I think it should be fine.

remcowesterhoud: I have changed it to `IDs` as in other parts of the spec we also use `ID` 🙂

PR #22157, opened 2024-09-10
Re-enable process instance migration tests

korthout:
## Description

This unignores some tests for process instance migration that were previously ignored. They are now supported functionality, so we can re-enable them.

## Related issues

NA


berkaycanbc (approved):
Thanks @korthout!

PR #22166, opened 2024-09-10
Migrate element instances marked as parallel multi instance

### LINKED ISSUE #22129: Migrate parallel multi-instance element instances
## Description

Following cases should be supported:

### Migrate multi-instance service tasks

Screenshot 2024-09-10 at 18 19 45

A multi-instance service task itself, its instances and created jobs (per instance) should be migrated. A mapping should be provided between A->B.

### Migrate multi-instance receive tasks

Screenshot 2024-09-10 at 18 20 33

A multi-instance receive task itself, its instances and created message subscriptions (per instance) should be migrated. A mapping should be provided between A->B.

### Migrate multi-instance embedded subprocess

Screenshot 2024-09-10 at 18 23 30

A multi-instance embedded subprocess itself, its instances and any active element inside instances should be migrated. A mapping should be provided between sub1->sub2 and A->B.

### Migrate multi-instance call activities

Screenshot 2024-09-10 at 18 24 21

A multi-instance call activity should be migrated. A mapping should be provided between A->B. 

Please note that it is not possible to migrate each call activity instance because they belong to another process instance. A new migration should be defined per called activity to migrate each child instance because they are separate process instances.

####  Notes
- Migrating multi-instance body with some of its instances are completed should be possible as well. For example, if 2 instances out of 3 is completed for a multi-instance service task, remaining active instance should be migrated.
- Migrating large set of multi instances (exceeding batch size) is left out of scope. Migrating such multi instances will exceed `MAX_MESSAGE_SIZE` and fail.

berkaycanbc:
## Description

Following cases should be supported:

### Migrate multi-instance service tasks

Screenshot 2024-09-10 at 18 19 45

A multi-instance service task itself, its instances and created jobs (per instance) should be migrated. A mapping should be provided between A->B.

### Migrate multi-instance receive tasks

Screenshot 2024-09-10 at 18 20 33

A multi-instance receive task itself, its instances and created message subscriptions (per instance) should be migrated. A mapping should be provided between A->B.

### Migrate multi-instance embedded subprocess

Screenshot 2024-09-10 at 18 23 30

A multi-instance embedded subprocess itself, its instances and any active element inside instances should be migrated. A mapping should be provided between sub1->sub2 and A->B.

### Migrate multi-instance call activities

Screenshot 2024-09-10 at 18 24 21

A multi-instance call activity should be migrated. A mapping should be provided between A->B. 

Please note that it is not possible to migrate each call activity instance because they belong to another process instance. A new migration should be defined per called activity to migrate each child instance because they are separate process instances.

####  Notes
- Migrating multi-instance body with some of its instances are completed should be possible as well. For example, if 2 instances out of 3 is completed for a multi-instance service task, remaining active instance should be migrated.
- Migrating large set of multi instances (exceeding batch size) is left out of scope. Migrating such multi instances will exceed `MAX_MESSAGE_SIZE` and fail.

## Related issues

closes #22129

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
@@ -55,7 +55,8 @@ public final class ProcessInstanceMigrationPreconditions {
           BpmnElementType.EVENT_BASED_GATEWAY,
           BpmnElementType.BUSINESS_RULE_TASK,
           BpmnElementType.SCRIPT_TASK,
-          BpmnElementType.SEND_TASK);
+          BpmnElementType.SEND_TASK,
+          BpmnElementType.MULTI_INSTANCE_BODY);
   private static final Set<BpmnElementType> UNSUPPORTED_ELEMENT_TYPES =
       EnumSet.complementOf(SUPPORTED_ELEMENT_TYPES);
   private static final Set<BpmnEventType> SUPPORTED_INTERMEDIATE_CATCH_EVENT_TYPES =
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
@@ -561,11 +561,7 @@ public final class ProcessInstanceMigrationPreconditions {
           sourceFlowScopeElement.getValue().getElementIdBuffer();
       final AbstractFlowElement targetFlowElement =
           targetProcessDefinition.getProcess().getElementById(targetElementId);
-      DirectBuffer actualFlowScopeId = targetFlowElement.getFlowScope().getId();
-
-      if (expectedFlowScopeId.equals(actualFlowScopeId)) {
-        return;
-      }
+      DirectBuffer actualFlowScopeId;
 
       // if target element is a multi instance body, we should check the inner activity flow scope
       // because the inner activity of the multi instance body can still match the source element's
@@ -580,6 +576,11 @@ public final class ProcessInstanceMigrationPreconditions {
         if (expectedFlowScopeId.equals(actualFlowScopeId)) {
           return;
         }
+      }
+
+      actualFlowScopeId = targetFlowElement.getFlowScope().getId();
+      if (expectedFlowScopeId.equals(actualFlowScopeId)) {
+        return;
       }
 
       final String reason =
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
@@ -162,6 +162,13 @@ public final class ProcessInstanceMigrationPreconditions {
       but a concurrent command was executed on the process instance. \
       Please retry the migration.""";
 
+  private static final String ERROR_UPDATED_LOOP_CHARACTERISTICS =
+      """
+      Expected to migrate process instance '%s' \
+      but active element with id '%s' has a different loop characteristics \
+      than the target element with id '%s'. \
+      Both elements must have either sequential or parallel loop characteristics.""";
+
   private static final String ZEEBE_USER_TASK_IMPLEMENTATION = "zeebe user task";
   private static final String JOB_WORKER_IMPLEMENTATION = "job worker";
 
@@ -445,7 +452,7 @@ public final class ProcessInstanceMigrationPreconditions {
    *
    * @param targetProcessDefinition target process definition to retrieve the target element type
    * @param targetElementId target element id
-   * @param elementInstanceRecord element instance to do the check
+   * @param elementInstance element instance to do the check
    * @param processInstanceKey process instance key to be logged
    */
   public static void requireSameElementType(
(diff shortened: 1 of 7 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateMultiInstanceBodyTest.java
```diff
@@ -0,0 +1,704 @@
...
+                Bpmn.createExecutableProcess(processId)
+                    .startEvent()
+                    .serviceTask(
+                        "serviceTask1",
+                        t ->
+                            t.zeebeJobTypeExpression("jobType")
+                                .multiInstance(
+                                    b ->
+                                        b.zeebeInputCollectionExpression("jobTypes")
+                                            .zeebeInputElement("jobType")))
```
korthout, line 59: 💭 Interesting construct to create jobs with types based on the input collection. This is not the common usage though. Most often, users simply need to a single task multiple times. Cool that this works though! And I can see that it makes the test easier to verify

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateMultiInstanceBodyTest.java
```diff
@@ -0,0 +1,704 @@
...
+    engine.job().ofInstance(childProcessInstanceKey).withType("A").complete();
+
+    assertThat(
+            RecordingExporter.processInstanceRecords(ProcessInstanceIntent.ELEMENT_ACTIVATED)
+                .withProcessInstanceKey(childProcessInstanceKey)
+                .withElementType(BpmnElementType.END_EVENT)
+                .withElementId("multi_instance_child_process_end")
+                .findAny())
+        .describedAs("Expect that the process instance is continued in the target process")
+        .isPresent();
```
korthout, line 566: ❌ This verifies that the end event of the child process instance has activated. However, the child process instance was not migrated. We need to verify that the parent process instance continues in the target process, e.g.:
- after all jobs are completed, the multi-instance body is completed and flows to a different element (one that didn't exist in the source).

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
```diff
@@ -451,20 +452,37 @@ public static void requireSameElementType(
...
+    if (targetElementType == BpmnElementType.MULTI_INSTANCE_BODY) {
+      final ExecutableMultiInstanceBody targetElement =
+          targetProcessDefinition
+              .getProcess()
+              .getElementById(targetElementId, ExecutableMultiInstanceBody.class);
+      targetElementType = targetElement.getInnerActivity().getElementType();
+      if (elementInstanceRecord.getBpmnElementType() == targetElementType) {
+        return;
+      }
+    }
```
korthout, line 474: ❌ Please add a test case that shows this case. E.g. a multi-instance marked service task changed to a multi-instance marked user task.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
```diff
@@ -540,24 +558,38 @@ public static void requireUnchangedFlowScope(
...
+        final ExecutableMultiInstanceBody targetElement =
+            targetProcessDefinition
+                .getProcess()
+                .getElementById(targetElementId, ExecutableMultiInstanceBody.class);
+
+        actualFlowScopeId = targetElement.getInnerActivity().getFlowScope().getId();
+        if (expectedFlowScopeId.equals(actualFlowScopeId)) {
+          return;
+        }
+      }
```
korthout, line 582: ❌ Let's add a test case that cover this.

korthout (changes requested):
Great work @berkaycanbc 🚧 

❌ There are some edge cases that are not covered yet. Please add tests for them to show that we can handle them correctly.

For example, we need to add a test that showcases that you can't change the element type of the multi-instance.

We also need to cover rejections for the following edge cases:
- a sequential multi-instance marked activity cannot be changed to parallel multi-instance
- a parallel multi-instance marked activity cannot be changed to sequential multi-instance
- a multi-instance marked activity cannot be unmarked as multi-instance
- a non-multi-instance activity cannot be marked as multi-instance

Lastly, we need to show that incidents caused by the `output element`, `output collection`, and `completion condition` expressions can be resolved through instance migration.

💭 We've already discussed that sequential multi-instance will be covered separately.

korthout (commented):
Great stuff @berkaycanbc 

🙇 I was unable to complete the comprehensive review

I have left a few comments for you to consider during my short absence. I'll complete my comprehensive review Tuesday. Please have a look at my comments in the meantime, if you find the time.

💭 I haven't covered all tests in my review
💭 I haven't checked whether all edge cases are now covered by tests

👍 However, generally this looks good

🔧 You could consider testing migrations that map elements nested inside a multi-instance embedded sub-process changing the flow scope. While this should probably already be covered by our checks, I have a feeling that our checks may not cover them well in the same way that you found in 4f11ab6c94dbdd4827f717d6b6e27f2ca218b528:
- a user task inside a multi-instance embedded sub-process is mapped to a user task inside a doubly nested embedded sub-process (i.e. a sub-process is added between the inner instance and the user task), we should reject this case as the flow scope has changed
- a user task inside a multi-instance embedded sub-process is mapped to a user task outside of the embedded sub-process (i.e. it is moved out of the subprocess), we should reject this case as the flow scope has changed

korthout:
Rebased on `main` (resolving a small conflict in 08d976e6ac80040bcb15fe26eb881ded2e6ff91b) and added the two requested test cases.

korthout:
Pair review with @berkaycanbc:
- one more test needed for changing output collection during migration
- one concurrent test: activate parallel multi-instance body, followed by migration, before `ProcessInstanceBatch:ACTIVATE_ELEMENT`

korthout:
@berkaycanbc I had to add a fix in 3c0fa11 due to the output collection changed case. For the rest all is as expected and discussed.

korthout (approved):

PR #22205, opened 2024-09-11
feat: write continuation commands at the end of distribution queues

### LINKED ISSUE #21449: Command distribution can write a custom continuation command
Command distribution is fire-and-forget in the sense that processors that submit records for distribution don't get a notification when distribution is completed. While command distribution itself does write `ACKNOWLEDGE` and `FINISHED` events, these are not observable by other processors.

For scaling partitions, we need to redistribute all resources to new partitions. Because resources tend to be large, this needs to be incremental. Additionally, redistributing resources is just the first of many steps, so we need a way to _continue_ after distribution of all resources is finished.

To support this, we propose the ability to register a _continuation command_ for a [queue](https://github.com/camunda/camunda/issues/21447). The command can be registered at any time, not only when the queue is "created" for the first time.

When a continuation command is registered and the queue is already empty or later becomes empty, the continuation command is written to the log.

Depends on https://github.com/camunda/camunda/issues/21447.

## Acceptance Criteria
- [ ] Any arbitrary command can be registered as continuation
- [ ] The command is persisted in the state and survives a restart
- [ ] On registration, if the queue is empty, the command is written immediately
- [ ] If the queue is not empty on registration, the command is written when the queue becomes empty

lenaschoenburg:
This adds the concept of _Continuation Commands_ to the command distribution behavior. In particular, we allow registering an arbitrary amount of commands that will be written once a queue is or becomes empty.
 
```java
commandDistributionBehavior
  .withKey(123L)
  .afterQueue("my-custom-queue")
  .continueWith(myCallbackCommand);
```
Requesting a continuation command on an empty queue will just immediately write the command and nothing else.

To do this, I've added a new column family to persist continuation commands as well as two events to register a new continuation command and to remove it again after it is written.

The commands and events are written as follow up commands/events for a given _continuation key_.

:warning: As usual, tests are missing because behaviors are not really testable with the usual `EngineRule` infrastructure. The first use case will have to add required tests.

closes https://github.com/camunda/camunda/issues/21449

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
@@ -57,7 +57,9 @@ public class CommandDistributionAcknowledgeProcessor
     stateWriter.appendFollowUpEvent(
         distributionKey, CommandDistributionIntent.ACKNOWLEDGED, recordValue);
 
-    commandDistributionBehavior.distributeNextInQueue(distributionKey, partitionId);
+    distributionState
+        .getQueueIdForDistribution(distributionKey)
+        .ifPresent(queueId -> commandDistributionBehavior.advanceQueue(queueId, partitionId));
 
     if (!distributionState.hasPendingDistribution(distributionKey)) {
       // We write an empty command here as a distribution could contain a lot of data. Because of
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
@@ -9,6 +9,7 @@ package io.camunda.zeebe.engine.processing.distribution;
 
 import io.camunda.zeebe.engine.processing.streamprocessor.TypedRecordProcessor;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedCommandWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedRejectionWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
 import io.camunda.zeebe.engine.state.immutable.DistributionState;
@@ -27,6 +28,7 @@ public class CommandDistributionAcknowledgeProcessor
 
   private final CommandDistributionBehavior commandDistributionBehavior;
   private final DistributionState distributionState;
+  private final TypedCommandWriter commandWriter;
   private final StateWriter stateWriter;
   private final TypedRejectionWriter rejectionWriter;
 
@@ -36,6 +38,7 @@ public class CommandDistributionAcknowledgeProcessor
       final Writers writers) {
     this.commandDistributionBehavior = commandDistributionBehavior;
     this.distributionState = distributionState;
+    commandWriter = writers.command();
     stateWriter = writers.state();
     rejectionWriter = writers.rejection();
   }
@@ -57,21 +60,20 @@ public class CommandDistributionAcknowledgeProcessor
     stateWriter.appendFollowUpEvent(
         distributionKey, CommandDistributionIntent.ACKNOWLEDGED, recordValue);
 
-    distributionState
-        .getQueueIdForDistribution(distributionKey)
-        .ifPresent(queueId -> commandDistributionBehavior.advanceQueue(queueId, partitionId));
+    final var queueId = distributionState.getQueueIdForDistribution(distributionKey);
+    queueId.ifPresent(
+        queue -> commandDistributionBehavior.distributeNextInQueue(queue, partitionId));
 
     if (!distributionState.hasPendingDistribution(distributionKey)) {
-      // We write an empty command here as a distribution could contain a lot of data. Because of
-      // this we could exceed the max message size. As we only need the distributionKey in the
-      // FINISHED event applier an empty record will suffice here.
-      stateWriter.appendFollowUpEvent(
-          distributionKey,
-          CommandDistributionIntent.FINISHED,
+      final var finishRecord =
           new CommandDistributionRecord()
               .setPartitionId(record.getPartitionId())
               .setValueType(recordValue.getValueType())
-              .setIntent(recordValue.getIntent()));
+              .setIntent(recordValue.getIntent());
+      queueId.ifPresent(finishRecord::setQueueId);
+
+      commandWriter.appendFollowUpCommand(
+          distributionKey, CommandDistributionIntent.FINISH, finishRecord);
     }
   }
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
@@ -275,12 +275,10 @@ public final class CommandDistributionBehavior {
   private <T extends UnifiedRecordValue> void requestContinuation(
       final String queue,
       final long key,
-      final boolean awaitNonEmptyQueue,
       final ValueType valueType,
       final Intent intent,
       final T value) {
-    final var writeImmediately =
-        !awaitNonEmptyQueue && !distributionState.hasQueuedDistributions(queue);
+    final var writeImmediately = !distributionState.hasQueuedDistributions(queue);
 
     if (writeImmediately) {
       commandWriter.appendFollowUpCommand(key, intent, value);
@@ -325,17 +323,17 @@ public final class CommandDistributionBehavior {
   }
 
   public interface ContinuationRequestBuilder {
-    /**
-     * Specifies whether to wait for the queue to be non-empty before writing the continuation
-     * command. If set to false, the continuation command will be written immediately if the queue
-     * is currently empty.
-     */
-    ContinuationRequestBuilder awaitNonEmptyQueue(boolean awaitNonEmptyQueue);
 
-    /** Write this command once the queue is empty. */
+    /**
+     * Write this command once the queue is empty. If the queue is already empty, the command will
+     * be written immediately.
+     */
     <T extends UnifiedRecordValue> void continueWith(TypedRecord<T> command);
 
-    /** Write this command once the queue is empty. */
+    /**
+     * Write this command once the queue is empty. If the queue is already empty, the command will
+     * be written immediately.
+     */
     <T extends UnifiedRecordValue> void continueWith(
         final ValueType valueType, final Intent intent, final T value);
   }
@@ -345,7 +343,6 @@ public final class CommandDistributionBehavior {
     final long key;
     String queue;
     Set<Integer> partitions = routingInfo.partitions();
-    private boolean awaitNonEmptyQueue;
 
     public DistributionRequest(final long key) {
       this.key = key;
@@ -406,26 +403,15 @@ public final class CommandDistributionBehavior {
     }
 
     @Override
-    public ContinuationRequestBuilder awaitNonEmptyQueue(final boolean awaitNonEmptyQueue) {
-      this.awaitNonEmptyQueue = awaitNonEmptyQueue;
-      return this;
-    }
-
-    @Override
     public <T extends UnifiedRecordValue> void continueWith(final TypedRecord<T> command) {
       requestContinuation(
-          queue,
-          key,
-          awaitNonEmptyQueue,
-          command.getValueType(),
-          command.getIntent(),
-          command.getValue());
+          queue, key, command.getValueType(), command.getIntent(), command.getValue());
     }
 
     @Override
     public <T extends UnifiedRecordValue> void continueWith(
         final ValueType valueType, final Intent intent, final T value) {
-      requestContinuation(queue, key, awaitNonEmptyQueue, valueType, intent, value);
+      requestContinuation(queue, key, valueType, intent, value);
     }
   }
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
@@ -83,7 +83,7 @@ public final class CommandDistributionBehavior {
    *     completed. Additionally, the key determines the order in which the commands are distributed
    *     if they are distributed ordered, i.e. a queue is specified.
    */
-  public CommandDistributionRequestBuilder withKey(final long distributionKey) {
+  public RequestBuilder withKey(final long distributionKey) {
     return new DistributionRequest(distributionKey);
   }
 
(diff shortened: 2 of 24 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
```diff
@@ -252,6 +252,8 @@ public interface RequestBuilder {
     DistributionRequestBuilder unordered();
 
     DistributionRequestBuilder inQueue(String queue);
+
+    ContinuationRequestBuilder afterQueue(String queue);
```
remcowesterhoud, line 306: ❓ Why did you decide to separate the continuation from the disribution itself? Do you think there'll be use-cases where we want to add a continuation without distributing something to a queue? In my mind the api would've been something like:

```java
commandDistributionBehavior
  .withKey(123L)
  .inQueue("queueId")
  .withContinuation(continuationCommand)
  .distribute()
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
```diff
@@ -272,10 +274,28 @@ <T extends UnifiedRecordValue> void distribute(
...
   }
 
-  private class DistributionRequest implements RequestBuilder, DistributionRequestBuilder {
+  public interface ContinuationRequestBuilder {
+    /**
+     * Specifies whether to wait for the queue to be non-empty before writing the continuation
+     * command. If set to false, the continuation command will be written immediately if the queue
+     * is currently empty.
+     */
+    ContinuationRequestBuilder awaitNonEmptyQueue(boolean awaitNonEmptyQueue);
```
remcowesterhoud, line 333: ❌ I don't like this option to be honest. It makes the assumption that something will be distributed within the queue. Do you see a specific need for this already? Like I said in my previous comment, I believe it should only possible to add a continuation when actually distributing a command within the same queue.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/distribution/DbDistributionState.java
```diff
@@ -55,6 +55,12 @@ public class DbDistributionState implements MutableDistributionState {
   private final DbCompositeKey<DbString, DbCompositeKey<DbInt, DbForeignKey<DbLong>>>
       queuedDistributionKey;
 
+  /** [queue id | key] => persisted continuation command */
+  private final ColumnFamily<DbCompositeKey<DbString, DbLong>, PersistedCommandDistribution>
+      continuationCommandColumnFamily;
```
remcowesterhoud: 🔧 By using the queue id and distributionKey as the key of this CF you add a limitation that we can only add 1 continuation command per distribution. Is this intended? In the PR description you mention:

> In particular, we allow registering an arbitrary amount of commands that will be written once a queue is or becomes empty.

This is only true across different distributions now.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
```diff
@@ -57,7 +57,9 @@ public void processRecord(final TypedRecord<CommandDistributionRecord> record) {
     stateWriter.appendFollowUpEvent(
         distributionKey, CommandDistributionIntent.ACKNOWLEDGED, recordValue);
 
-    commandDistributionBehavior.distributeNextInQueue(distributionKey, partitionId);
+    distributionState
+        .getQueueIdForDistribution(distributionKey)
+        .ifPresent(queueId -> commandDistributionBehavior.advanceQueue(queueId, partitionId));
```
remcowesterhoud, line 62: ❌ / 🔧  It feels a bit strange to me to try and continue in the Acknowledge processor. The acknowledgement comes from a single partition. Most of the times we'd still need to wait for the acknowledgement of the other partition before we can continue with the continuation command.

I find it cleaner to have a `CommandDistribution.FINISH` command that takes care of this, as well as writing the `FINISHED` event. What do you think?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
```diff
@@ -187,6 +191,23 @@ void distributeNextInQueue(final long finishedDistributionKey, final int partiti
...
+      return;
+    }
+    distributionState.forEachContinuationCommand(queue, this::handleContinuationCommand);
+  }
+
+  private void handleContinuationCommand(final long key, final CommandDistributionRecord record) {
+    final var intent = record.getIntent();
+    final var value = record.getCommandValue();
+
+    commandWriter.appendFollowUpCommand(key, intent, value);
```
remcowesterhoud, line 205: ❌ We don't know how much continuation commands are in our state and how large the record values are. This seems like a big risk to exceed the maximum batch size. I think we'd be better off writing a `CONTINUE` command with just the queueId and the key. That way we can write all the actual commands in separate batches.

remcowesterhoud: Writing this comment makes me realize that the continuation naming still confuses me a bit. If we add a  `CommandDistributionIntent.CONTINUE` command, what will it continue. Will it write the next command in the queue, or will it take care of the callback. Hence why I think callback is a better name.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
```diff
@@ -187,6 +191,23 @@ void distributeNextInQueue(final long finishedDistributionKey, final int partiti
...
+
+    record.setPartitionId(currentPartitionId);
+    stateWriter.appendFollowUpEvent(key, CommandDistributionIntent.CONTINUATION_COMPLETED, record);
```
remcowesterhoud, line 208: ❌ We shouldn't use the full record here since we can easily run out of the maximum batch size this way. As you only need the key and queueId we should create an empty record with only those properties (and other non-optional properties) set.

remcowesterhoud (changes requested):
Thanks @lenaschoenburg! 🚀  I have a few remarks on the proposed changes. Please have a look and let me know if you have any questions 🙂 

❌ I miss unit tests of the new methods that were added in the DistributionState class.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: Yeah exactly. I think it's likely that adding distributions and registering continuations will be two separate steps. One loop is just adding new distributions and something else is registering the continuation. That way we don't have to add logic to the loop to check if it's the last distribution etc.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: That's fair :+1: The use case I had in mind was first registering a continuation and then adding a bunch of distributions to the queue. However, I'm not yet sure if it will be needed or not. If you prefer, I can remove it for now until we know that we need it.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/distribution/DbDistributionState.java
lenaschoenburg: The key is not a distribution key, it can be user provided. That allows registering multiple continuations on the same queue.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
lenaschoenburg: That would only shift things around slightly, no? Right now we check if the queue is done and if it is we write the continuation command. What you are proposing is to still check if the queue is done but then first write a `CommandDistribution.FINISH` command that then writes the continuation command.  What do we gain from that?
Perhaps a bit besides the point but I'd find `CommandDistribution.FINISH` confusing because it'd only happen when the queue becomes empty and `CommandDistribution.FINISHED` has an entirely different meaning and is written for each individual distribution.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: True, I haven't thought about that :+1: 
To avoid confusing naming we could do this:
1. On ACK, when the queue is found to be empty, write `CommandDistributionIntent.FINISH_QUEUE` command.
2. When processing `CommandDistributionIntent.FINISH_QUEUE` write x continuation commands and matching `CommandDistributionIntent.CONTINUATION_COMPLETED` events.
3. If we still have remaining continuation commands at the end of handling `CommandDistributionIntent.FINISH_QUEUE`, write another `CommandDistributionIntent.FINISH_QUEUE` command.
4. If we don't have continuation commands at the end of handling `CommandDistributionIntent.FINISH_QUEUE`, write a no-op `CommandDistributionIntent.FINISHED_QUEUE` event.

Something like that?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: Fixed in 950df4b

lenaschoenburg:
> ❌ I miss unit tests of the new methods that were added in the DistributionState class.

I added some in a680c8d16e4c1b4435d90cab551b2c37f5dbf705

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: On the other hand this complicates the implementation a good bit. I'm almost inclined to just remove the feature to register multiple continuation commands for that reason. If we only have one continuation command, we don't need to care about batch size at all.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
remcowesterhoud: Let's remove it for now and see if we'll need it in the future 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/distribution/DbDistributionState.java
remcowesterhoud: Okay, so we'll have to generate a new key fo reach continuation. That sounds good to me. The naming in the builder confused me here. Could we rename this?

```java
    @Override
    public <T extends UnifiedRecordValue> void continueWith(final TypedRecord<T> command) {
      requestContinuation(
          queue,
          distributionKey, <--------
          awaitNonEmptyQueue,
          command.getValueType(),
          command.getIntent(),
          command.getValue());
    }

    @Override
    public <T extends UnifiedRecordValue> void continueWith(
        final ValueType valueType, final Intent intent, final T value) {
      requestContinuation(queue, distributionKey, <------ awaitNonEmptyQueue, valueType, intent, value);
    }
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
remcowesterhoud: > What do we gain from that?

Semantically it makes sense to check for callbacks and trigger them after something has finished. So instead of the log being `Distribute -> Acknowledge -> Continuation Completed -> Finished`

It'd be `Distribute -> Acknowledge -> Finished -> Continuation Completed`
In my opinion the second option is more correct. After a distribution is finished you trigger the continuation, as the continuation is a final step.

You'd also be doing less useless checks. Within the acknowledge processor you know >50% of the processed commands aren't going to trigger a continuation, since other partitions haven't acknowledged yet. If you do it in a `FINISH` command you know all the partitions are done with the distribution and there's a fair chance it was the last distribution of the queue.

> Perhaps a bit besides the point but I'd find CommandDistribution.FINISH confusing because it'd only happen when the queue becomes empty and CommandDistribution.FINISHED has an entirely different meaning and is written for each individual distribution.

I could say the same for CommandDistribution.ACKNOWLEDGED 😄 This processor is even more atomic as it's written for each partial distribution.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
remcowesterhoud: > On the other hand this complicates the implementation a good bit. I'm almost inclined to just remove the feature to register multiple continuation commands for that reason. If we only have one continuation command, we don't need to care about batch size at all.

I think that's fair, let's not over-engineer this until we actually need it. 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/distribution/DbDistributionState.java
lenaschoenburg: Ah yeah my bad, this is leftover from refactoring :+1:

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionAcknowledgeProcessor.java
remcowesterhoud: Exactly 👍

lenaschoenburg: Done, please check the latest commits :)

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/distribution/DbDistributionState.java
lenaschoenburg: Fixed in f389974ef272f99f3016feff097ed9eea12ce706

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: If I remove this option, what should the behavior be if the queue is currently empty? Write the command immediately or just request it for later?
I slightly leaning towards just keeping this option because it makes this behavior explicit and doesn't really add much complexity.

remcowesterhoud: This is why I expected a continuation command to only be avialable when you actaully distribute something 😄 

Personally I would say it makes sense to immediately write the command. When you add a continuation you are basically saying, please write this command when queue `X` is empty. If it's already empty I don't see a reason to wait for it 🤷

remcowesterhoud: We can clarify this happens on the javadoc

remcowesterhoud (commented):
I looked at the changes so far, and it looks good to me 👍 There will likely be some test failures considering the distirbution lifecycle changed now that there is a `FINISHED` command.

lenaschoenburg:
Rebased to resolve minor merge conflicts

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: :+1: This option is removed now.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionBehavior.java
lenaschoenburg: Because removing this functionality made implementation slightly more difficult, I've opted for you suggestion to first write small `CONTINUE` commands which in turn write the continuation command and the `CONTINUATION_COMPLETED` event. This should avoid the batch size issue except in pathological cases where there's thousands of continuation commands and we can't even write all of the `CONTINUE` commands.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionContinueProcessor.java
```diff
@@ -0,0 +1,47 @@
...
+    final var distributionRecord = record.getValue();
+    final var queue = distributionRecord.getQueueId();
+
+    final var continuationRecord = distributionState.getContinuationRecord(queue, key);
+    final var intent = continuationRecord.getIntent();
+    final var continuationCommand = continuationRecord.getCommandValue();
+    commandWriter.appendFollowUpCommand(key, intent, continuationCommand);
+
+    stateWriter.appendFollowUpEvent(
+        key, CommandDistributionIntent.CONTINUATION_COMPLETED, distributionRecord);
```
remcowesterhoud, line 45: 🔧 Not a must but I wonder if it's nicer to rename `CONTINUATION_COMPLETED` to `CONTINUED`. It's not really completed, as we don't know what happened with the continued command. Maybe it got rejected 🤷

remcowesterhoud (approved):
Thanks for the changes @lenaschoenburg! I have one smaller comment, please have a look at that and feel free to merge after 🙂

PR #22211, opened 2024-09-11
fix: introduce timeout when waiting for backup snapshot to finish

### LINKED ISSUE #18049: Operate can't complete backup snapshot
### Describe the bug

When Operate takes a backup it schedules several snapshots (currently 6) for it. Then it checks for every snapshot the state for progress. In case that the snapshot state doesn't change for any reason (e.g. it's running exceptionally long; bug in Elasticsearch), then the check is running forever without any further information for the user.

### To Reproduce

* We cannot fully reproduce this scenario at the moment
* See [Support 21530](https://jira.camunda.com/browse/SUPPORT-21530)

### Current behavior

* After the create snapshot request initially times out, we check the snapshot status periodically via ES API and enter into a loop until the snapshot status `"IN PROGRESS"` resolves (see https://github.com/camunda/camunda/blob/operate-8.5.6/operate/webapp/src/main/java/io/camunda/operate/webapp/elasticsearch/backup/ElasticsearchBackupRepository.java#L405)
* More details in [Support 21530](https://jira.camunda.com/browse/SUPPORT-21530)

### Expected behavior

* We make transparent to the user what is happening: add logs to periodically report that the snapshot is still in progress (TODO: decide on log level and frequency); this should include the snapshot ID so that the user can double-check against the Elasticsearch API
* We introduce a configurable timeout so that
  * the snapshot status polling ends once exceeded
  * the snapshot request fails, so that the user can clean up and decide if they want to try it again
  * The default value for that timeout is 0 (unbounded; users need to opt in) to avoid that we introduce regressions for users where the snapshot procedure takes long but is generally successful

### Environment
Seen in versions 8.3.2 and above

- OS: [e.g. MacOS]
- Browser: [e.g. chrome, safari]
- Operate Version: [e.g. 8.1]

### Additional context

- [ ] The issue has been labeled with an appropriate `Bug-area` label

- Changed components:
- Side effects on other components:
- Handy resources:
  
- Example projects:

- Commands/Steps needed to test; Versions to validate:

- Docker file / HELM chart : in case that it needed to be tested via docker share the version contain the fixed along with version of other services .

- Release version ( in which version this fixed/feature will be released):

sdorokhova:
## Description

* introduce timeout when waiting for backup snapshot to finish
* add configuration parameter `camunda.operate.backup.snapshotTimeout` to set the timeout (default no timeout).

## Related issues

closes #18049

### DIFF
--- a/operate/common/src/main/java/io/camunda/operate/property/BackupProperties.java
+++ b/operate/common/src/main/java/io/camunda/operate/property/BackupProperties.java
@@ -11,12 +11,23 @@ public class BackupProperties {
 
   private String repositoryName;
 
+  private long snapshotTimeoutMillis = 0;
+
   public String getRepositoryName() {
     return repositoryName;
   }
 
-  public BackupProperties setRepositoryName(String repositoryName) {
+  public BackupProperties setRepositoryName(final String repositoryName) {
     this.repositoryName = repositoryName;
+    return this;
+  }
+
+  public long getSnapshotTimeoutMillis() {
+    return snapshotTimeoutMillis;
+  }
+
+  public BackupProperties setSnapshotTimeoutMillis(final long snapshotTimeoutMillis) {
+    this.snapshotTimeoutMillis = snapshotTimeoutMillis;
     return this;
   }
 }
--- a/operate/common/src/main/java/io/camunda/operate/property/BackupProperties.java
+++ b/operate/common/src/main/java/io/camunda/operate/property/BackupProperties.java
@@ -11,7 +11,7 @@ public class BackupProperties {
 
   private String repositoryName;
 
-  private long snapshotTimeoutMillis = 0;
+  private int snapshotTimeout = 0;
 
   public String getRepositoryName() {
     return repositoryName;
@@ -22,12 +22,12 @@ public class BackupProperties {
     return this;
   }
 
-  public long getSnapshotTimeoutMillis() {
-    return snapshotTimeoutMillis;
+  public int getSnapshotTimeout() {
+    return snapshotTimeout;
   }
 
-  public BackupProperties setSnapshotTimeoutMillis(final long snapshotTimeoutMillis) {
-    this.snapshotTimeoutMillis = snapshotTimeoutMillis;
+  public BackupProperties setSnapshotTimeout(final int snapshotTimeout) {
+    this.snapshotTimeout = snapshotTimeout;
     return this;
   }
 }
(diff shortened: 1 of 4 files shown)


operate/common/src/main/java/io/camunda/operate/property/BackupProperties.java
```diff
@@ -11,12 +11,23 @@ public class BackupProperties {
 
   private String repositoryName;
 
+  private long snapshotTimeoutMillis = 0;
```
mihail-ca, line 14: Would it make more sense to express this in seconds ? I don't think someone will ever use sub-second timeout for backups. It removes 3 unnecessary trailing zeroes

operate/webapp/src/main/java/io/camunda/operate/webapp/elasticsearch/backup/ElasticsearchBackupRepository.java
```diff
@@ -307,7 +313,66 @@ private List<SnapshotInfo> findSnapshots(String repositoryName, Long backupId) {
...
+      if (currentSnapshot.state() == IN_PROGRESS) {
+        ThreadUtil.sleepFor(100);
+        count++;
+        if (count % 600 == 0) { // approx. 1 minute, depending on how long findSnapshots takes
+          LOGGER.info(String.format("Waiting for snapshot [%s] to finish.", snapshotName));
+        }
+      } else {
+        return handleSnapshotReceived(currentSnapshot);
+      }
+    } while (snapshotTimeout == 0 || System.currentTimeMillis() - startTime <= snapshotTimeout);
```
mihail-ca, line 346: I think we can replace `do {} while` with just `while()`

operate/webapp/src/main/java/io/camunda/operate/webapp/elasticsearch/backup/ElasticsearchBackupRepository.java
```diff
@@ -307,7 +313,66 @@ private List<SnapshotInfo> findSnapshots(String repositoryName, Long backupId) {
     }
   }
 
-  private GetBackupStateResponseDto getBackupResponse(Long backupId, List<SnapshotInfo> snapshots) {
+  protected boolean waitForFinishedSnapshotWithTimeout(
```
mihail-ca, line 316: I think the method name is confusing, because it return boolean, but we always wait, whether the output is true or false. I would rename this method to something like `isSnapshotFinishedWithinTimeout()`

operate/webapp/src/main/java/io/camunda/operate/webapp/elasticsearch/backup/ElasticsearchBackupRepository.java
```diff
@@ -375,59 +440,41 @@ private GetBackupStateResponseDto getBackupResponse(Long backupId, List<Snapshot
...
     }
 
     @Override
-    public void onFailure(Exception ex) {
+    public void onFailure(final Exception ex) {
       if (ex instanceof SocketTimeoutException) {
         // This is thrown even if the backup is still running
         LOGGER.warn(
             String.format(
                 "Timeout while creating snapshot [%s] for backup id [%d]. Need to keep waiting with polling...",
```
mihail-ca, line 471: I would rename this `Timeout` text in the warning to `SocketTimeout`, to have a clear distinction between the SocketTimeout and the Polling Timeout.

mihail-ca:
LGTM, just small changes I would recommend.

ghost (approved):

sdorokhova:
Backup/restore tests are green on the branch: https://github.com/camunda/camunda/actions/runs/10938463159/job/30366552167

PR #22212, opened 2024-09-11
feat: POC - centralized resource authorization for search services

houssain-barouni:
## Description

This pull request introduces a centralized resource-based authorization filtering for entities search.

1.	`@ResourceAuthorizationKey` annotation:
  •	A new annotation used to mark fields in resource entity classes that are used by authorization filtering.
  •	The annotation includes a `forResourceType()` attribute used to match user authorizations.

Example
```java
public record ProcessInstanceEntity(
    Long key,
    String processName,
    Integer processVersion,
    String bpmnProcessId,
    @ResourceAuthorizationKey(forResourceType = "process-definition") String  String bpmnProcessId)
```

2.	`ResourceAuthorizationSearchChecker`:
  •	Search for fields annotated with `@ResourceAuthorizationKey` in the entity class.
  •	Constructs a search query to fetch the authorizations of the authenticated user for the resource type specified by the annotated fields.
  •	If no authorized resource keys are found, the checker generates `matchNone()` query.
  •	If the wildcard resource key "*" is found, the user is granted access to all resources `matchAll()`.
  •	Otherwise, a `stringTerms()` query is constructed to filter based on the resource key field.

`ResourceAuthorizationSearchChecker` is then called after building the main query, along with authentication query (tenants filter query)
`ResourceAuthorizationSearchChecker` can be enabled at service level (for example, it is disabled for Authorizations and Users search services)

This still miss checks over the permission `READ` or `CREATE` ... and also supports filtering on string keys only

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/service/src/main/java/io/camunda/service/AuthorizationServices.java
+++ b/service/src/main/java/io/camunda/service/AuthorizationServices.java
@@ -37,7 +37,7 @@ public class AuthorizationServices<T>
       final CamundaSearchClient searchClient,
       final ServiceTransformers transformers,
       final Authentication authentication) {
-    super(brokerClient, searchClient, transformers, authentication);
+    super(brokerClient, searchClient, transformers, authentication, false);
   }
 
   @Override
--- a/service/src/main/java/io/camunda/service/UserServices.java
+++ b/service/src/main/java/io/camunda/service/UserServices.java
@@ -30,7 +30,7 @@ public class UserServices extends SearchQueryService<UserServices, UserQuery, Us
       final CamundaSearchClient searchClient,
       final ServiceTransformers transformers,
       final Authentication authentication) {
-    super(brokerClient, searchClient, transformers, authentication);
+    super(brokerClient, searchClient, transformers, authentication, false);
   }
 
   @Override
--- a/service/src/main/java/io/camunda/service/entities/ProcessInstanceEntity.java
+++ b/service/src/main/java/io/camunda/service/entities/ProcessInstanceEntity.java
@@ -8,6 +8,7 @@
 package io.camunda.service.entities;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import io.camunda.service.security.auth.ResourceAuthorizationKey;
 import java.util.List;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
@@ -15,7 +16,7 @@ public record ProcessInstanceEntity(
     Long key,
     String processName,
     Integer processVersion,
-    String bpmnProcessId,
+    @ResourceAuthorizationKey(forResourceType = "process-definition") String bpmnProcessId,
     Long parentProcessInstanceKey,
     Long parentFlowNodeInstanceKey,
     String startDate,
--- a/service/src/main/java/io/camunda/service/security/auth/ResourceAuthorizationSearchChecker.java
+++ b/service/src/main/java/io/camunda/service/security/auth/ResourceAuthorizationSearchChecker.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.service.security.auth;
+
+import static io.camunda.search.clients.query.SearchQueryBuilders.matchAll;
+import static io.camunda.search.clients.query.SearchQueryBuilders.matchNone;
+import static io.camunda.search.clients.query.SearchQueryBuilders.stringTerms;
+import static io.camunda.service.search.query.SearchQueryBuilders.authorizationSearchQuery;
+
+import io.camunda.search.clients.CamundaSearchClient;
+import io.camunda.search.clients.query.SearchQuery;
+import io.camunda.search.clients.query.SearchQueryBuilders;
+import io.camunda.service.entities.AuthorizationEntity;
+import io.camunda.service.search.core.SearchClientBasedQueryExecutor;
+import io.camunda.service.transformers.ServiceTransformers;
+import java.util.Arrays;
+import java.util.List;
+
+public class ResourceAuthorizationSearchChecker {
+
+  private final SearchClientBasedQueryExecutor searchExecutor;
+
+  public ResourceAuthorizationSearchChecker(
+      final CamundaSearchClient camundaSearchClient,
+      final ServiceTransformers serviceTransformers) {
+    searchExecutor =
+        new SearchClientBasedQueryExecutor(camundaSearchClient, serviceTransformers, null);
+  }
+
+  /**
+   * Creates a search query that checks the user's authorization for accessing the resource entity.
+   *
+   * <p>This method looks for fields in the resource entity class that are annotated with
+   * {@code @ResourceAuthorizationKey}. For each such field, it generates a query that ensures the
+   * authenticated user has permission to access the resource.
+   */
+  public SearchQuery getResourceAuthorizationCheck(
+      final Authentication authentication, final Class<?> resourceEntityClass) {
+    return Arrays.stream(resourceEntityClass.getFields())
+        .filter(field -> field.isAnnotationPresent(ResourceAuthorizationKey.class))
+        .map(
+            field ->
+                createResourceAuthorizationQuery(
+                    field.getAnnotation(ResourceAuthorizationKey.class),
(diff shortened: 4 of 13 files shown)


service/src/main/java/io/camunda/service/AuthorizationServices.java
```diff
@@ -37,7 +37,7 @@ public AuthorizationServices(
       final CamundaSearchClient searchClient,
       final ServiceTransformers transformers,
       final Authentication authentication) {
-    super(brokerClient, searchClient, transformers, authentication);
+    super(brokerClient, searchClient, transformers, authentication, false);
```
Ben-Sheppard, line 40: 💭 Authorizations are a tricky one, in one instance the authorizations needs to be retrieved without checks so they can be applied effectively but in the other instance when I want to view authorizations via REST then I should only see them if I am authorized to, I think this is likely something that we need to address in an alternate way, perhaps with a different service class or handler

service/src/main/java/io/camunda/service/UserServices.java
```diff
@@ -30,7 +30,7 @@ public UserServices(
       final CamundaSearchClient searchClient,
       final ServiceTransformers transformers,
       final Authentication authentication) {
-    super(brokerClient, searchClient, transformers, authentication);
+    super(brokerClient, searchClient, transformers, authentication, false);
```
Ben-Sheppard, line 33: ❓ I think we would want to apply authorizations to a user search query because I should only be able to see users if I am authorized to do so.

service/src/main/java/io/camunda/service/entities/ProcessInstanceEntity.java
```diff
@@ -8,14 +8,15 @@
...
+import io.camunda.service.security.auth.ResourceAuthorizationKey;
 import java.util.List;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public record ProcessInstanceEntity(
     Long key,
     String processName,
     Integer processVersion,
-    String bpmnProcessId,
+    @ResourceAuthorizationKey(forResourceType = "process-definition") String bpmnProcessId,
```
Ben-Sheppard, line 19: 🔧  I like the approach here, I would instead rely on the `AuthorizationResourceType` class to give you the resource type that maps to the authorizations

service/src/main/java/io/camunda/service/security/auth/ResourceAuthorizationSearchChecker.java
```diff
@@ -0,0 +1,81 @@
...
+    if (authorizedResourceKeys.isEmpty()) {
+      return matchNone();
+    }
+    if (authorizedResourceKeys.contains("*")) {
+      return matchAll();
+    }
+    return stringTerms(fieldName, authorizedResourceKeys);
+  }
+
+  private List<AuthorizationEntity> getAuthorizationEntities(
```
Ben-Sheppard, line 75: 💭 just something to bear in mind for the future, authorizations can be assigned in different ways i.e. via groups, so this method would expand to also include the authorizations assigned in other ways.

service/src/main/java/io/camunda/service/entities/ProcessInstanceEntity.java
houssain-barouni: yes, good point, I was about to create a new enum 😄 
I wonder why `AuthorizationResourceType` is not used in `AuthorizationEntity` ? 🤔

service/src/main/java/io/camunda/service/security/auth/ResourceAuthorizationSearchChecker.java
houssain-barouni: we can iteratively implement that.
I guess we need to change the filters below to include the authorization of the user's group

service/src/main/java/io/camunda/service/AuthorizationServices.java
houssain-barouni: Yes, good point. 
I believe we need to differentiate between “business” resources (such as processes and decisions) and infrastructure resources (like authorizations, users, and groups). 
While infrastructure resources should be accessible to the applications, we need to apply the authorization checks when these resources are exposed through the API.

service/src/main/java/io/camunda/service/UserServices.java
houssain-barouni: same as above

service/src/main/java/io/camunda/service/security/auth/ResourceAuthorizationSearchChecker.java
Ben-Sheppard: Yea absolutely, I just wanted to note it down that in the future this set of filters will expand to be more than just the user Id :)

service/src/main/java/io/camunda/service/entities/ProcessInstanceEntity.java
Ben-Sheppard: Good question, I think thats just an oversight on my part to be honest and something that I should update

service/src/main/java/io/camunda/service/AuthorizationServices.java
Ben-Sheppard: The differentiation sounds fair to me there, I guess really its down to the context i.e. when I get the authorizations on an infra level I don't want the checks, but when its at a business level then I do. That feels like something that we can just handle as a different service I think so the service for the business level stuff has this enabled, but the other service for internal infra stuff does not

romansmirnov: The boolean flag might become a potential source of issues. When do I set to `true` when to `false`, does the service that I have right now does authorization checks or not, etc.

Can't we make it more explicit? [For example in C7](https://github.com/camunda/camunda-bpm-platform/blob/64274723bb94a91343e0acba1b08920459439f6c/engine/src/main/java/org/camunda/bpm/engine/impl/interceptor/CommandContext.java#L540), there is a method called `runWithoutAuthorization()`which must be explicitly used to execute something without any authorization check. That way, it is also explicit in the code of a service and clear what is happening.

service/src/main/java/io/camunda/service/entities/ProcessInstanceEntity.java
romansmirnov: Hm... it implies that the field name is also called `bpmnProcessId,` which might be okay in that case, but I could imagine that other fields are named differently depending on the index.

romansmirnov:
I went through the PR. In general, it seems okay, and in the very first version of authorization checks it might be sufficient. But I think, it wouldn't be enough looking forward. By that, I don't mean we have to build everything from beginning, but at least there should be a structure that allows us to extend it in the future.

1. How to check for specific permission? Currently the proposal does not take into account which permissions to check. It loads them all for a specific resource, but does not take into account if the user has a read permission.
3. How to construct a (disjunctive) composite permission check? For example, `(bpmnProcessId == foo && permission == READ_INSTANCE) OR (processInstanceKey == 1234 && permission == READ)`
4. What happens when the authorization check happens on a field that is actually not part of the entity?
5. How can we ensure that the service layer defines the authorizations to check, and then the transformers (or the search client) applies them according to the used datastorage? (Background of my question is the RDBMS project).

I think, we should re-iterate on this POC.

houssain-barouni:
> I went through the PR. In general, it seems okay, and in the very first version of authorization checks it might be sufficient. But I think, it wouldn't be enough looking forward. By that, I don't mean we have to build everything from beginning, but at least there should be a structure that allows us to extend it in the future.
> 
> 1. How to check for specific permission? Currently the proposal does not take into account which permissions to check. It loads them all for a specific resource, but does not take into account if the user has a read permission.
> 2. How to construct a (disjunctive) composite permission check? For example, `(bpmnProcessId == foo && permission == READ_INSTANCE) OR (processInstanceKey == 1234 && permission == READ)`
> 3. What happens when the authorization check happens on a field that is actually not part of the entity?
> 4. How can we ensure that the service layer defines the authorizations to check, and then the transformers (or the search client) applies them according to the used datastorage? (Background of my question is the RDBMS project).
> 
> I think, we should re-iterate on this POC.

Thanks @romansmirnov for the review. I agree with your points. The aim of this POC was to demonstrate that we can centralize authorization checks in one place, although it's still far from perfect. 

You may already be aware, but starting from iteration 2 (mid-October), the C8 REST API team will take over handling permission checks in query services. 
I'll be the DRI for this, and I plan to come up with an improved solution, breaking it down into smaller tasks. 
I will also check what is being done regarding authorization checks for commands services done by @remcowesterhoud. Your feedback and questions provide a great starting point. I'll keep you and @Ben-Sheppard looped in and we will do a kick-off. cc @tmetzke.

PR #22215, opened 2024-09-11
Do not update treePath on FNI completed or terminated

### LINKED ISSUE #22075: Do not query for the parent tree path when an element instance completes or terminates
### Description

* When a flow node instance is imported, the importer determines its tree path. If the tree path value is not in the cache, then we fetch it from Elasticsearch.
* We do this on every type of event for the flow node instance. However, on `ELEMENT_COMPLETED` and `ELEMENT_TERMINATED`, the tree path will already exist in Elasticsearch and we don't need to overwrite it (we need to do it for `ELEMENT_MIGRATED` though)

### TODO

* Do not include the tree path in flow node instance upsert requests, when the processed record had the intent `ELEMENT_COMPLETED` or `ELEMENT_TERMINATED`
* Note: Must still work correctly when `ELEMENT_TERMINATED` and `ELEMENT_TERMINATED` are in the same batch (so that we only make one request for both events)!
* Backport to 8.5

### Additional Information

### Acceptance Criteria

ChrisKujawa:
## Description

To reduce the load on our cache and respectively our backend (ES or OS) we do not resolve the treePath on COMPLETED nor TERMINATED. Furthermore, we do no update the property in the index, as we expect it to be set earlier. 

This not resolving of the treePath is especially interesting when Operate has been restarted, the cache is empty and we are importing an COMPLETED event. This can happen in cases where instances are long-living. See related comment https://github.com/camunda/camunda/issues/22075#issuecomment-2338272792

### What the PR does

First, we had to extract the logic to transform a Zeebe flow node instance record
to an Operate Flow node instance entity into a separate class.

This encapsulates the transformation's logic and improves
maintainability and testability.

As a second step the bug has been fixed, that when FNI is in state COMPLETED or TERMINATED, we do not need to resolve nor update the treePath. With skipping this step we can reduce
the load on ES/OS.

## Related issues

closes #22075

### DIFF
--- a/operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/FlowNodeInstanceZeebeRecordProcessor.java
+++ b/operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/FlowNodeInstanceZeebeRecordProcessor.java
@@ -19,26 +19,24 @@ package io.camunda.operate.zeebeimport.v8_4.processors;
 import static io.camunda.operate.zeebeimport.util.ImportUtil.tenantOrDefault;
 import static io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent.*;
 
+import io.camunda.operate.Metrics;
 import io.camunda.operate.entities.FlowNodeInstanceEntity;
-import io.camunda.operate.entities.FlowNodeState;
-import io.camunda.operate.entities.FlowNodeType;
 import io.camunda.operate.exceptions.PersistenceException;
 import io.camunda.operate.property.OperateProperties;
 import io.camunda.operate.schema.templates.FlowNodeInstanceTemplate;
 import io.camunda.operate.store.BatchRequest;
 import io.camunda.operate.store.FlowNodeStore;
 import io.camunda.operate.util.ConversionUtils;
-import io.camunda.operate.util.DateUtil;
 import io.camunda.operate.zeebe.PartitionHolder;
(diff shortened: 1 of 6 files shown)


romansmirnov (approved):
LGTM.

Zelldon:
@romansmirnov @ralfpuchert I realized i haven't changed the 8.4 importer before. Done now with https://github.com/camunda/camunda/pull/22215/commits/1b2c473c0357df3ff2aaee0e737d7e2f33570bbc 

I had to copy the transformer and respective tests, as we use different record versions as far as I understood. Is this correct? And expected to do it like this, or is there another way?

Let me know.

Zelldon:
> as we use different record versions as far as I understood. Is this correct? And expected to do it like this, or is there another way?

Actually this doesn't seem to be true? Looking at the POMs 🤔 So I could move the transformer into common and reuse it or? Why do we have the duplication?

romansmirnov:
@Zelldon,

For reference, that's the code that decides which version of the processor to use:

https://github.com/camunda/camunda/blob/07b46b3e08b2b9af1a82543ff588d079cee8a338/operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportJob.java#L151-L154

> Furthermore, I believe might make sense to have the transformer per version as we might change/add properties? 🤔 But I think in the last versions we never did or? Maybe tenants

Tenants have been introduced with 8.3 😄 However, I think the idea is to duplicate the processors/transformers even when there is no change between the versions. (It also means that potential bugs must be fixed in two places.) By that, I think it is fine as done in this PR.

Issue #22076, opened 2024-09-09
Do not add an element instance tree path to the cache if the element isn't a container node


ThorbenLindhauer:
### Description

* We use the tree path cache only to select tree path values of parent flow node instances. Accordingly, we should only cache values for flow node instances that can have child instances (process, sub process, event sub process, multi-instance body)
* The current implementation puts unnecessarily many values into the tree path cache, leading to higher memory consumption and earlier eviction of values that are important to be cached

### TODO

* Only add tree path values to the cache for instances of: process, sub process, event sub process multi-instance body

### Additional Information

---------------------------------------------------------------------------------------------

### Acceptance Criteria

- [ ] Checked if the specific data needs to be provided for the QA process

### Definition of Ready - Checklist

- [ ] The issue has a meaningful title, description and testable acceptance criteria
- [ ] If documentation needs to be updated, an issue is created in the [camunda-docs](https://github.com/camunda/camunda-docs) repo, and the issue is added to our Operate project board.
- [ ] If HELM charts need to be updated, an issue is created in the [camunda-platform-heml](https://github.com/camunda/camunda-platform-helm) repo, and the issue is added to our Operate project board.
- [ ] Potential Security risks have been considered
- [ ] Cross-team dependencies have been considered

PR #22219, opened 2024-09-11
Do not store leaf FNI in treePath Cache

ChrisKujawa:
## PR Description

Be aware that this PR bases on changes from https://github.com/camunda/camunda/pull/22215

**Problem description:**

We want to reduce the load on the cache and unnecessarily evictions of important values.

> * We use the tree path cache only to select tree path values of parent flow node instances. Accordingly, we should only cache values for flow node instances that can have child instances (process, sub process, event sub process, multi-instance body)
> * The current implementation puts unnecessarily many values into the tree path cache, leading to higher memory consumption and earlier eviction of values that are important to be cached

### What the PR contains

 * Introduces a new interface for TreePathCache
 * Extract storing of treePath as separate method of the cache
   * give control to outside when to cache treePath and corresponding key 
 * Adjust FNI transformer such that the treePath is only cached for container FNIs
   * This change should allow to reduce load on the cache and potential evictions
 * Add several different tests related to that

## Related issues

closes https://github.com/camunda/camunda/issues/22076

### DIFF
--- a/operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/FlowNodeInstanceZeebeRecordProcessor.java
+++ b/operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/FlowNodeInstanceZeebeRecordProcessor.java
@@ -73,7 +73,7 @@ public class FlowNodeInstanceZeebeRecordProcessor {
             flowNodeTreeCacheSize,
             flowNodeStore::findParentTreePathFor,
             new TreePathCacheMetricsImpl(partitionIds, metrics));
-    fniTransformer = new FNITransformer(treePathCache::resolveTreePath);
+    fniTransformer = new FNITransformer(treePathCache);
   }
 
   public void processIncidentRecord(final Record record, final BatchRequest batchRequest)
--- a/operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/fni/FNITransformer.java
+++ b/operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/fni/FNITransformer.java
@@ -1,9 +1,18 @@
 /*
- * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
- * one or more contributor license agreements. See the NOTICE file distributed
- * with this work for additional information regarding copyright ownership.
- * Licensed under the Zeebe Community License 1.1. You may not use this file
- * except in compliance with the Zeebe Community License 1.1.
+ * Copyright Camunda Services GmbH
+ *
+ * BY INSTALLING, DOWNLOADING, ACCESSING, USING, OR DISTRIBUTING THE SOFTWARE (“USE”), YOU INDICATE YOUR ACCEPTANCE TO AND ARE ENTERING INTO A CONTRACT WITH, THE LICENSOR ON THE TERMS SET OUT IN THIS AGREEMENT. IF YOU DO NOT AGREE TO THESE TERMS, YOU MUST NOT USE THE SOFTWARE. IF YOU ARE RECEIVING THE SOFTWARE ON BEHALF OF A LEGAL ENTITY, YOU REPRESENT AND WARRANT THAT YOU HAVE THE ACTUAL AUTHORITY TO AGREE TO THE TERMS AND CONDITIONS OF THIS AGREEMENT ON BEHALF OF SUCH ENTITY.
+ * “Licensee” means you, an individual, or the entity on whose behalf you receive the Software.
+ *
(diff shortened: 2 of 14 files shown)


operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/fni/FNITransformer.java
```diff
@@ -33,10 +34,18 @@ public class FNITransformer {
   private static final Set<String> FINISH_STATES =
       Set.of(ELEMENT_COMPLETED.name(), ELEMENT_TERMINATED.name());
   private static final Set<String> START_STATES = Set.of(ELEMENT_ACTIVATING.name());
-  private final Function<FNITreePathCacheCompositeKey, String> treePathResolver;
 
-  public FNITransformer(final Function<FNITreePathCacheCompositeKey, String> treePathResolver) {
-    this.treePathResolver = treePathResolver;
+  private static final Set<BpmnElementType> CONTAINER_TYPES =
+      Set.of(
+          BpmnElementType.CALL_ACTIVITY,
```
romansmirnov, line 40: If I am not mistaken, then a call activity doesn't have a "child" flow node instance (i.e., element instance). So, in that sense, it is not a container type. It might only be relevant if that cache is used to store the tree path containing the call hierarchy.

Zelldon: Yeah here I was not sure, and wanted to first add all I know. Happy to remove it.

Zelldon: Should I remove it? I think it will be added to the cache at least before 🤔 The only point where call activity are mentioned is here https://github.com/camunda/camunda/blob/main/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/ListViewZeebeRecordProcessor.java#L722 but maybe I missed something

romansmirnov: Yes, let's remove 😄 Besides that, the PR looks good.

romansmirnov (approved):
LGTM.

operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/fni/FNITransformer.java
```diff
@@ -18,11 +18,12 @@
...
+import io.camunda.operate.zeebeimport.cache.TreePathCache;
 import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.value.BpmnElementType;
 import io.camunda.zeebe.protocol.record.value.ProcessInstanceRecordValue;
 import java.time.Instant;
 import java.util.Set;
-import java.util.function.Function;
 
 /**
  * Transfomer to transform a given Zeebe flow node instance record to a {@link
```
ralfpuchert, line 29: Small typo: Transformer

operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/FlowNodeInstanceZeebeRecordProcessor.java
```diff
@@ -75,7 +75,7 @@ public FlowNodeInstanceZeebeRecordProcessor(
             flowNodeTreeCacheSize,
             flowNodeStore::findParentTreePathFor,
```
ralfpuchert, line 76: Do we need `flowNodeStore` as a field, it is only used here?

ralfpuchert (approved):
LGTM, only 2 minor things.

operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/FlowNodeInstanceZeebeRecordProcessor.java
Zelldon: True, missed that. I will remove this.

Zelldon: Already done on other PR which hit stable :)

operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/fni/FNITransformer.java
Zelldon: Done 👍🏼

PR #22234, opened 2024-09-12
Add request handlers for cluster scale request for scaling brokers and partitions

### LINKED ISSUE #21482: Scale request adds operations to bootstrap new partitions

deepthidevaki:
## Description

This PR adds the request handlers and client sender for 
* ClusterPatchRequest that can change cluster configuration by providing a set of members to add or remove
* ClusterScaleRequest that can change cluster configuration by providing a new cluster size
* ForceRemoveBroker that can force remove the given set of brokers.

On the gateway endpoint, we will have just one request exposed to the client. The gateway will map it to one of these three request based on the provided input. Alternative was to define one request in the protobuf and let `ClusterConfigurationManagementRequestsHandler` map it. However, this felt unnecessary as this way we can define the user facing endpoint independent of the internal implementation. 

The tests verifies the integration of the API client and server. The test covering different combinations of the input are already covered by the unit tests for the corresponding request transformers.

## Related issues

closes #21482

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
@@ -9,11 +9,11 @@ package io.camunda.zeebe.shared.management;
 
 import io.atomix.cluster.MemberId;
 import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.AddMembersRequest;
+import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.BrokerScaleRequest;
 import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.CancelChangeRequest;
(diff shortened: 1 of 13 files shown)


deepthidevaki:
@lenaschoenburg I was wondering if we should consider the "feature flag" in this PR. Should we reject request to change partition count if the feature is disabled?

lenaschoenburg (approved):
> Should we reject request to change partition count if the feature is disabled?

Hm, I guess we'd have to do that, yes. Since we are planning for a flexible REST api, it's probably best to check the feature flag one level deeper. Perhaps `ClusterConfigurationManagementRequestsHandler` is a good place for it?

deepthidevaki:
> > Should we reject request to change partition count if the feature is disabled?
> 
> Hm, I guess we'd have to do that, yes. Since we are planning for a flexible REST api, it's probably best to check the feature flag one level deeper. Perhaps `ClusterConfigurationManagementRequestsHandler` is a good place for it?

Will do a follow up PR.

PR #22244, opened 2024-09-12
Benchmark: Worker use natural retry mechanism for message publish

ChrisKujawa:
## Problem Description

When a worker fails to publish a message, it fails the corresponding job, allowing immediate retry and sending the message again. In general, this seems to be a good approach, but in practice, it seems to lead to some problems especially if the system shortly struggles to handle requests and we are sending messages always to the same partition that struggles.

<details>

fail-retry-issue-general

When broker restart and partition role changes happen there is a short time when requests might be rejected, this can cause a spike in failed messages. If we now fail the job with immediate retry, and the send-back of the job is quick (due to job push and chances are hight the job exists on a different partition), then this makes the issues more severe. 
As it produces more and more load on the same partition retrying (which currently struggles), causing more to be rejected, etc.

We can see that the system is in the end overloaded with requests due to retry loops, causing to drop the throughput heavily.

</details>

## Solution

Instead of failing the job, we simply let the job time out, so someone else has to pick up the job later. This might delay the individual process instance, but overall it has a lesser impact, as we can work on a different job in the meantime, keeping up the throughput.

natural-retry-general
natural-retry

**Why does it work for a different job?** 

It might be that one partition has currently some struggle due to restarts or role changes, chances are low that this affects all partitions.

This might cause issues for the current job to publish a message, but we are sending messages via correlation key, based on the process instance payload. 

On the next job/message published the chances are `partition count - 1 / partition count` that we hit another partition where it works without issues. 

## Finding the solution

I tried several approaches before finding the solution described above. In the following, I want to briefly summarize what I have tried.

#### Approach 1 

**Approach:** Specify a retry delay on the fail job, after message publish has failed. We used an exponential delay to not retry always at the same time. 

Image: gcr.io/zeebe-io/worker:ck-worker-retry-benchmark-f1f49f53
Benchmark: [Dashboard](https://grafana.dev.zeebe.io/d/zeebe-dashboard/zeebe?orgId=1&var-DS_PROMETHEUS=prometheus&var-cluster=All&var-namespace=ck-base-85-benchmark&var-pod=All&var-partition=All&from=1726130186773&to=1726210819379)

**Problem(s):** 
<details>
first-retry-approach
first-retry-approach-jobs

1. Using one Exponential delay on Worker creation, causes concurrency issues as all accessing the same strategy. See [first commit](https://github.com/camunda/camunda/pull/22244/commits/f1f49f53e35c2e8046edb84aa1843627b4f436c1).
2. No measurable improvement. The system was still overloaded, with retry-loops. Fail and publish message requests flooding the system. This was causing high backpressure and low throughput.
</details>

#### Approach 2

**Approach change:** To make the retry strategy available per thread, will not relate to a corresponding job and its retry. So I just randomize the retry delay. See [commit](https://github.com/camunda/camunda/pull/22244/commits/855efb9b9ced4196afb5610101d5a1863dc6b4dc)

Image: gcr.io/zeebe-io/starter:ck-worker-retry-benchmark-855efb9b

**Problem(s):** 

<details>

random-retry

random-retry-job

1. Similar to the above, the system was overloaded. The throughput was not stable and we can see spikes of retry loops.

</details>

#### Approach 3

**Approach change:** c4d9670fd72e06bbc72450b901122b6e705a9631 

Retry it directly, and fail it when time is up. This was more complex, as it involved loops, etc. There were some mistakes in the commit, as the changes were done with lack of focus. The results were misleading. I skipped this and tried another simpler approach, which resulted in the final solution.

Image: gcr.io/zeebe-io/starter:ck-worker-retry-benchmark-c4d9670f
Benchmark: [Dashboard](https://grafana.dev.zeebe.io/d/zeebe-dashboard/zeebe?orgId=1&var-DS_PROMETHEUS=prometheus&var-cluster=All&var-namespace=ck-base-86-benchmark&var-pod=All&var-partition=All&from=1726152289159&to=1726155817987)

**Problem(s):** 

<details>

direct-retry
direct-retry-jobs

The throughput was much lower than in other benchmarks, when we think about this it also makes sense as we retry for awhile instead of making progress on other things until the related job is unblocked.

</details>

## Related issues

related thread https://camunda.slack.com/archives/C07J9URA0B1/p1726123410255959

### DIFF
--- a/zeebe/benchmarks/project/src/main/java/io/camunda/zeebe/Worker.java
+++ b/zeebe/benchmarks/project/src/main/java/io/camunda/zeebe/Worker.java
@@ -98,15 +98,20 @@ public class Worker extends App {
 
         final boolean messagePublishedSuccessfully = publishMessage(client, correlationKey);
         if (!messagePublishedSuccessfully) {
-          // - On issues with publishing a message we need to retry the job
-          // - thus is to make sure our message gets published
-          // - otherwise our process might get stuck
-          // - failing the job makes it immediately available again
-          jobClient
-              .newFailCommand(job)
-              .retries(job.getRetries())
-              .errorMessage("Message publish failed.")
-              .send();
+          // Instead of failing the job, we simply let the job time out, so someone else has to
+          // pick up the job later. This might delay the individual process instance, but overall it
+          // has a lesser impact, as we can work on a different job in the meantime, keeping up the
+          // throughput.
+          //
+          // It might be that one partition has currently some struggle due to restarts or role
+          // changes, chances are low that this affects all partitions.
+          //
+          // This might cause issues for the current job to publish a message, but we are sending
+          // messages via correlation key,   based on the process instance payload.
+          //
+          // On the next job/message published the chances are (partition count - 1 / partition
+          // count) that we hit another partition where it works without issues.
+
           return;
         }
       }


rodrigo-lourenco-lopes (approved):
Thanks for the investigation on this issue! 🚀 
Letting the job time out seems like a good strategy and should reduce the resource exhausted problems that we were seeing before. Curious to try this with the config that was giving us issues.

PR #22270, opened 2024-09-12
feat: support basic auth login webapps

Ben-Sheppard:
## Description

I created a [PR](https://github.com/camunda/camunda/pull/19830) previously that looked to support basic auth in operate, that PR went stale and things have changed since then so I've opened this PR to complete the circle.

Some things to note:

1. With this PR I can start the single application with `auth-basic, operate, tasklist` and log in using a user in my ES
2. The UI correctly shows the name of the user
3. Tasklist functioned without alterations
4. I've extended the base Spring class because I think this is the best approach here to add our own fields in, and support a full transition from the web app security classes to our own (as there are more custom fields needed)

### DIFF
--- a/operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthBasicUserService.java
+++ b/operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthBasicUserService.java
@@ -7,42 +7,31 @@
  */
 package io.camunda.operate.webapp.security.auth;
 
-import io.camunda.operate.OperateProfileService;
 import io.camunda.operate.webapp.rest.dto.UserDto;
 import io.camunda.operate.webapp.security.AbstractUserService;
-import org.springframework.beans.factory.annotation.Autowired;
+import io.camunda.operate.webapp.security.Permission;
+import java.util.List;
 import org.springframework.context.annotation.Profile;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.stereotype.Component;
 
 @Component
-@Profile({
-  "!"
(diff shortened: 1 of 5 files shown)


operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthBasicUserService.java
```diff
@@ -0,0 +1,37 @@
...
+public class AuthBasicUserService extends AbstractUserService<UsernamePasswordAuthenticationToken> {
+  @Override
+  public UserDto createUserDtoFrom(final UsernamePasswordAuthenticationToken authentication) {
+    final var camundaUser = (CamundaUser) authentication.getPrincipal();
+
+    return new UserDto()
+        .setUserId(String.valueOf(camundaUser.getUserKey()))
+        .setDisplayName(camundaUser.getName())
+        .setCanLogout(false)
+        .setPermissions(List.of(Permission.READ, Permission.WRITE));
```
Ben-Sheppard, line 30: 🔧 at the moment the user logged in has access to everything, this should be taken from their access permissions in authorizations, may be fine for a first step until we have the new permission format in place for app access

remcowesterhoud (commented):
I don't think this is a PR I can/should review as a Zeebe engineer. The code looks alright to me though. I'm missing tests, but I don't know if it makes sense here or if it will be covered later.

maryarm (commented):
Code looks good to me, I just have a comment about missing tests, like something to check the type of the user detail and it's attribute that set in context.

Ben-Sheppard:
Hey @maryarm - I altered some tests to confirm new fields/the new object, hopefully this is what you were after :)

maryarm (approved):
Looks good.

PR #22274, opened 2024-09-12
feat: extract interfaces to separate `webapps-schema` module

### LINKED ISSUE #19006: Camunda exporter creates and updates Operate schema

sdorokhova:
## Description

Created the module `webapps-schema` and moved schema related interfaces in it. I would propose to put put implementations also in this module. We could separate different webapps by package names.


deepthidevaki (approved):
:rocket: Looks good to me.

Do you envision adding concrete implementations of both IndexDescriptors and the entities in this module? If so, should we also move the UserEntity from the exporter to here? No need to do it in this PR.

sdorokhova:
> Do you envision adding concrete implementations of both IndexDescriptors and the entities in this module? If so, should we also move the UserEntity from the exporter to here? No need to do it in this PR.

@deepthidevaki Yes, I propose to add entities and index descriptors in this module in sub packages for every webapp module.
I tried to move both `UserEntity` and `AuthorizationEntity`, but `AuthorizationEntity` depends on `zeebe-protocol` classes. I did not decide how to resolve that :) Do you have an opinion? Should we move only `UserEntity`? Or should we add `zeebe-protocol` dependency to the new module?

deepthidevaki:
> > Do you envision adding concrete implementations of both IndexDescriptors and the entities in this module? If so, should we also move the UserEntity from the exporter to here? No need to do it in this PR.
> 
> @deepthidevaki Yes, I propose to add entities and index descriptors in this module in sub packages for every webapp module. I tried to move both `UserEntity` and `AuthorizationEntity`, but `AuthorizationEntity` depends on `zeebe-protocol` classes. I did not decide how to resolve that :) Do you have an opinion? Should we move only `UserEntity`? Or should we add `zeebe-protocol` dependency to the new module?

I'm not sure about that. We can do it in a separate PR. So let it not block this one.

PR #22289, opened 2024-09-13
fix: Fix bugs related with upgrade es to new lib

KarpovIvan:
## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/operate/schema/src/main/java/io/camunda/operate/store/opensearch/client/sync/OpenSearchDocumentOperations.java
+++ b/operate/schema/src/main/java/io/camunda/operate/store/opensearch/client/sync/OpenSearchDocumentOperations.java
@@ -111,10 +111,23 @@ public class OpenSearchDocumentOperations extends OpenSearchRetryOperation {
       final Consumer<HitsMetadata<R>> hitsMetadataConsumer,
       final Class<R> clazz)
       throws IOException {
+    SearchResponse<R> response = openSearchClient.search(request, clazz);
+    return scrollWith(
+        request, response, hitsConsumer, hitsMetadataConsumer, clazz, Integer.MAX_VALUE);
+  }
+
+  public <R> Map<String, Aggregate> scrollWith(
+      final SearchRequest request,
+      SearchResponse<R> response,
+      final Consumer<List<Hit<R>>> hitsConsumer,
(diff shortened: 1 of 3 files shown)


grlimacan (changes requested):
Code itself looks fine, but there seems to be a linting issue.

oleksandr-kriuchenko-lohika (approved):

RomanJRW (approved):

operate/schema/src/main/java/io/camunda/operate/store/opensearch/client/sync/OpenSearchDocumentOperations.java
```diff
@@ -137,8 +124,7 @@ public <R> Map<String, Aggregate> scrollWith(
       scrollId = response.scrollId();
       List<Hit<R>> hits = response.hits().hits();
 
-      while (!hits.isEmpty() && scrollId != null && count < limit) {
-        count += response.hits().hits().size();
+      while (!hits.isEmpty() && scrollId != null) {
```
RomanJRW, line 127: same as other PR, I don't think we should touch Operate files here

KarpovIvan: I reverted all changes as we have in master because this changes affect java check style build

RomanJRW: even though this was not a newly changed/committed file? This is strange, it should only look at new/changed files I believe.

The change here also looks like a _logical_ change, no?

KarpovIvan: If right now its important I will create clean PR. And Push changes related with operate directly without PR for fix Operate class. But for me its strange.

RomanJRW: Why strange? The point here is that there is nothing that requires us to touch operate code in the context of these changes. This is not code that the Optimize team wrote or maintains, and is not in the scope of this PR or the changes being brought to the mono repo from Optimize.

If we have different understandings, or if there is something I am overlooking, please let me know :)

PR #22291, opened 2024-09-13
[Backport main]: Do not update treePath on FNI completed or terminated

ChrisKujawa:
## Description
Backport #22215

> To reduce the load on our cache and respectively our backend (ES or OS) we do not resolve 
> the treePath on COMPLETED nor TERMINATED. Furthermore, we do no update the property in the index, as we expect it to be set earlier.
> 
> This not resolving of the treePath is especially interesting when Operate has been restarted, the cache is empty and we are importing an COMPLETED event. This can happen in cases where instances are long-living. See related comment https://github.com/camunda/camunda/issues/22075#issuecomment-2338272792

### DIFF
--- a/operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/fni/FNITransformer.java
+++ b/operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/fni/FNITransformer.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright Camunda Services GmbH
+ *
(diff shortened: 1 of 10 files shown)


operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/fni/FNITransformer.java
```diff
@@ -0,0 +1,112 @@
...
+import io.camunda.zeebe.protocol.record.value.ProcessInstanceRecordValue;
+import java.time.Instant;
+import java.util.Set;
+import java.util.function.Function;
+
+/**
+ * Transfomer to transform a given Zeebe flow node instance record to a {@link
+ * FlowNodeInstanceEntity}.
+ */
+public class FNITransformer {
```
romansmirnov, line 31: The entire `importer-8_4` can be deleted because it is not required on `main`.

romansmirnov (approved):
The `importer-8_4` can be deleted. Besides that, LGTM 🚀

operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/fni/FNITransformer.java
Zelldon: Ah these duplications driving me nuts :D Sorry for that

PR #22296, opened 2024-09-13
feat: implement resource authorizations migration logic

### LINKED ISSUE #22295: Implement AuthorizationMigrationHandler to handle the migration logic
### Description 
As we have a new structure for Authorisations in standalone application, we have to migrate the existing Resource Authorizations to this new structure.

### Tasks
- [x] call management identity client to get a batch of resource auhtorization for migration
- [x] group by authorizations and convert values
- [x]  create/update Authorization
- [x] call management identity client to update status of migrated items

maryarm:
## Description 
As we have a new structure for Authorisations in standalone application, we have to migrate the existing Resource Authorizations to this new structure.

## Tasks
- [x] call management identity client to get a batch of resource auhtorization for migration
- [x] group by authorizations and convert values
- [x]  create/update Authorization
- [x] call management identity client to update status of migrated items

## Related issues

closes #22295


identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
```diff
@@ -0,0 +1,134 @@
...
+  public void migrate() {
+    UserResourceAuthorization lastRecord = null;
+    while (true) {
+      final List<UserResourceAuthorization> authorizations =
+          managementIdentityProxy.fetchUserResourceAuthorizations(lastRecord, 100);
+      if (authorizations.isEmpty()) {
+        return;
+      }
+      lastRecord = authorizations.getLast();
+      final Map<String, Map<String, Map<String, List<String>>>> authorizationMap =
```
Ben-Sheppard, line 47: ❓ I'm concerned at the complexity here (and in the following `forEach`) but I think that we are somewhat limited in our options so it may be that we actually have no other way.

I wonder if we can build a composite key (represented by a internal record) that maps to the list of resource IDs, so instead of 
```
{"ben", {"process-definition", {"write", ["resource-one", "resource-two"]}}}
```
we could have 
```
{<composite key object>, ["resource-one", "resource-two"]}
```

Composite key represented by the class 
```
  private record AuthorizationCompositeKey(
      String username, String resourceType, String permission) {}
```

I think that would drop the stream to a shorter size because we would say
```
      final Map<AuthorizationCompositeKey, List<String>> authorizationsMap =
          authorizations.stream()
              .collect(
                  Collectors.groupingBy(
                      auth ->
                          new AuthorizationCompositeKey(
                              auth.username(), auth.resourceType(), auth.permission()),
                      Collectors.mapping(
                          UserResourceAuthorization::resourceId, Collectors.toList())));
```

This also means the stream below with the conversion simplifies because we have the composite key giving us data and the linked resources i.e. 
```
      authorizationsMap.forEach(
          (key, value) -> {
            final long ownerKey = getOwnerKeyForUsername(key.username());
            authorizationService.patchAuthorization(
                new PatchAuthorizationRequest(
                    ownerKey,
                    PermissionAction.ADD,
                    convertResourceType(key.resourceType()),
                    //helper method to expand key.permission after conversion));
          });
```

identity/migration/src/test/java/io/camunda/identity/migration/AuthorizationMigrationHandlerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    MockitoAnnotations.openMocks(this).close();
+    migrationHandler =
+        new AuthorizationMigrationHandler(authorizationServices, managementIdentityProxy);
+  }
+
+  @AfterEach
+  void tearDown() {}
+
+  @Test
+  void stopWhenNorMoreRecords() {
```
Ben-Sheppard, line 47:
```suggestion
  void stopWhenNoMoreRecords() {
```

Ben-Sheppard (commented):
Left a couple of comments :)

identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
maryarm: I like the idea to use the CompositKey but for our case the composite key should n't contains permissions as we are patching all permissions + (resourceIds) which belong to a owner/resource in one call, so we can just decrease one level of the map and put owner + resource Type together as a key.

identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
```diff
@@ -0,0 +1,134 @@
...
+          authorizations.stream()
+              .collect(
+                  Collectors.groupingBy(
+                      UserResourceAuthorization::username,
+                      Collectors.groupingBy(
+                          UserResourceAuthorization::resourceType,
+                          Collectors.groupingBy(
+                              UserResourceAuthorization::permission,
+                              Collectors.mapping(
+                                  UserResourceAuthorization::resourceId, Collectors.toList())))));
```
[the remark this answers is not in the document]
maryarm: I'm not sure if I understand your comment completely. Management identity supports resource authorization for process definition and decision definition types, so my understanding is in then new authorization system these two would be the resourceTypes and resourceId will be the same resourceId in management identity.

identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
```diff
@@ -0,0 +1,134 @@
...
+
+    if (e.getKey().startsWith("delete")) {
+      return List.of(PermissionType.DELETE);
+    }
+
+    throw new IllegalArgumentException("Unsupported permission type: " + e.getKey());
+  }
+
+  private long getOwnerKeyForUsername(final String owner) {
+    return 0;
```
Ben-Sheppard, line 132: 💭  I think we should leave a todo here so we don't leave the default value in place :)

identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
```diff
@@ -0,0 +1,134 @@
...
+                                  SimpleEntry::getKey,
+                                  Collectors.flatMapping(
+                                      e -> e.getValue().stream(), Collectors.toList())));
+                  final long ownerKey = getOwnerKeyForUsername(owner);
+                  authorizationService.patchAuthorization(
+                      new PatchAuthorizationRequest(
+                          ownerKey,
+                          PermissionAction.ADD,
+                          convertResourceType(resourceType),
+                          permissions));
```
Ben-Sheppard, line 80: 🔧 The intention is that this migration app will use the java client to interact with the API, so here instead of using the service classes directly we will use a similar method to the user create via java client you implemented.

I think for now I'm fine if we want to merge the general structure but this will need to change soon.

Ben-Sheppard (commented):
Sorry for not spotting it sooner but I've just spotted something regarding the use of the service classes that goes against my current understanding of the migration flow

identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
maryarm: you'r right, I didn't notice that we AuthorizationClient ready.

identity/migration/pom.xml
```diff
@@ -39,6 +39,27 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-protocol</artifactId>
+      <version>8.6.0-SNAPSHOT</version>
```
Ben-Sheppard, line 55: 🔧 These versions should be lifted up into a property :)

identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
Ben-Sheppard: Hmmm I was sure that there was a patch method in the java client, ok something to refactor then, I wonder if a TODO would be good here to not forget NABD though

identity/migration/src/main/java/io/camunda/identity/migration/ManagementIdentityProxy.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.List;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+
+@Component
+public class ManagementIdentityProxy {
+
+  private final RestTemplate restTemplate = new RestTemplate();
+
+  public List<UserResourceAuthorization> fetchUserResourceAuthorizations(
```
Ben-Sheppard, line 22: 🔧 For SaaS we need to also supply the organization ID (which can be passed in via configuration) but I think this can be followed up with

Ben-Sheppard (approved):
I've got a few more comments but I think I'm keen to get the base in and follow up with further changes and tweaks

identity/migration/src/main/java/io/camunda/identity/migration/AuthorizationMigrationHandler.java
Ben-Sheppard: Ah! Its here https://github.com/camunda/camunda/blob/e5c32ef9b87a58085d5197710ec330b96642775d/clients/java/src/main/java/io/camunda/zeebe/client/impl/ZeebeClientImpl.java#L627

PR #22086, opened 2024-09-09
refactor: bring leftovers from optimize repo

OmranAbazid:
## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes #

PR #22299, opened 2024-09-13
feat: add plugins on Optimize OS

igpetrov:
## Description

feat: add plugins on Optimize OS

This will be rebased after https://github.com/camunda/camunda/pull/22086 merged.

Solution discussed in internal slack: https://camunda.slack.com/archives/C076RT81VM2/p1726169685111989

## Checklist

- [ ] for CI changes:

## Related

Depends on https://github.com/camunda/camunda/pull/22086

### DIFF
--- a/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/upgrade/os/OpenSearchClientBuilder.java
+++ b/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/upgrade/os/OpenSearchClientBuilder.java
@@ -30,6 +30,7 @@ import io.camunda.optimize.service.db.os.ExtendedOpenSearchClient;
 import io.camunda.optimize.service.exceptions.OptimizeConfigurationException;
 import io.camunda.optimize.service.exceptions.OptimizeRuntimeException;
 import io.camunda.optimize.service.util.configuration.ConfigurationService;
+import io.camunda.optimize.service.util.configuration.OpenSearchConfiguration;
 import io.camunda.optimize.service.util.configuration.elasticsearch.DatabaseConnectionNodeConfiguration;
 import io.camunda.optimize.service.util.mapper.CustomAuthorizedReportDefinitionDeserializer;
 import io.camunda.optimize.service.util.mapper.CustomCollectionEntityDeserializer;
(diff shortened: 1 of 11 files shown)


optimize/util/optimize-commons/src/main/java/io/camunda/optimize/upgrade/os/OpenSearchClientBuilder.java
```diff
@@ -91,12 +96,31 @@ public class OpenSearchClientBuilder {
 
   public static ExtendedOpenSearchClient buildOpenSearchClientFromConfig(
       final ConfigurationService configurationService) {
-    return new ExtendedOpenSearchClient(buildOpenSearchTransport(configurationService));
+    final var pluginRepo = new PluginRepository();
```
oleksandr-kriuchenko-lohika, line 99: IMO better design would be to move PluginRepository instantiation outside of those client factory methods and pass it as a parameter, similar to how configurationService is passed. This should make OpenSearchClientBuilder better testable and less dependent on how PluginRepository is instantiated and initialized.

oleksandr-kriuchenko-lohika (approved):

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/upgrade/os/OpenSearchClientBuilder.java
igpetrov: That was my consideration too, but didn't do it in the first place because I was thinking that creating passing object as an argument in long chain might not have the best look. 

Do you suggest to create it in the [client factory create](https://github.com/camunda/camunda/blob/main/optimize/backend/src/main/java/io/camunda/optimize/service/db/os/OptimizeOpenSearchClientFactory.java#L32) or one level  higher in the `OptimizeOpenSearchClientConfiguration`?

oleksandr-kriuchenko-lohika: Looking into current design the best place to me seems to be io.camunda.optimize.service.db.os.OptimizeOpenSearchClientConfiguration#createOptimizeOpenSearchClient - this is the topmost place where all the dependencies are collected and passed as parameters. Yeah, I agree that passing dependencies as parameters in long chains is not the most convinient approach, but for static factory methods there is no other option, I suppose.

igpetrov: Alright, I can do it

PR #22312, opened 2024-09-13
Create a default user

### LINKED ISSUE #21421: Camunda should support creation of a default user to be authorized to access endpoints when auth-basic is enabled

remcowesterhoud:
## Description

This PR adds a scheduled task upon broker start. This task is only scheduled a single time and is responsible for creating the default user. This user will only be created once, and only on the Deployment partition. It's created via a command and will be distributed to other partitions accordingly.

## Checklist

- [ ] for CI changes:

## Related issues

closes #21421

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/DefaultUserCreator.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/DefaultUserCreator.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.user;
+
+import static java.util.Optional.*;
+
+import io.camunda.zeebe.engine.Loggers;
+import io.camunda.zeebe.engine.state.mutable.MutableProcessingState;
+import io.camunda.zeebe.protocol.Protocol;
+import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
+import io.camunda.zeebe.protocol.record.intent.UserIntent;
+import io.camunda.zeebe.protocol.record.value.UserType;
+import io.camunda.zeebe.stream.api.ReadonlyStreamProcessorContext;
+import io.camunda.zeebe.stream.api.StreamProcessorLifecycleAware;
+import io.camunda.zeebe.stream.api.scheduling.Task;
+import io.camunda.zeebe.stream.api.scheduling.TaskResult;
+import io.camunda.zeebe.stream.api.scheduling.TaskResultBuilder;
+import org.slf4j.Logger;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+
+public final class DefaultUserCreator implements StreamProcessorLifecycleAware, Task {
+  public static final String DEFAULT_USER_USERNAME = "demo";
+  public static final String DEFAULT_USER_PASSWORD = "demo";
+  public static final String DEFAULT_USER_EMAIL = "demo@demo.com";
+  private static final Logger LOG = Loggers.PROCESS_PROCESSOR_LOGGER;
+  private final MutableProcessingState processingState;
+  private final BCryptPasswordEncoder passwordEncoder;
+
+  public DefaultUserCreator(final MutableProcessingState processingState) {
+    this.processingState = processingState;
+    passwordEncoder = new BCryptPasswordEncoder();
+  }
+
+  @Override
+  public void onRecovered(final ReadonlyStreamProcessorContext context) {
+    if (context.getPartitionId() != Protocol.DEPLOYMENT_PARTITION) {
+      // We should only create users on the deployment partition. The command will be distributed to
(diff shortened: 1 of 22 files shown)


remcowesterhoud:
@nicpuppa Review is probably easiest to do by commit

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -273,15 +274,15 @@ public static Either<ProblemDetail, DocumentCreateRequest> toDocumentCreateReque
...
+      final UserRequest request, final PasswordEncoder passwordEncoder) {
     return getResult(
         validateUserCreateRequest(request),
         () ->
             new CreateUserRequest(
                 request.getUsername(),
                 request.getName(),
                 request.getEmail(),
-                request.getPassword()));
+                passwordEncoder.encode(request.getPassword())));
```
nicpuppa, line 285: 💭 I remember a discussion about it. I thought we said that Zeebe should not be in charge of encoding the password, it should just store and export the password as it is. Identyty should be in charge of provide an encrypted password

nicpuppa (approved):
Great job @remcowesterhoud, I left some suggestions/comments, please have a look 👀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/DefaultUserCreator.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.Logger;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+
+public final class DefaultUserCreator implements StreamProcessorLifecycleAware, Task {
+  public static final String DEFAULT_USER_USERNAME = "demo";
+  public static final String DEFAULT_USER_PASSWORD = "demo";
+  public static final String DEFAULT_USER_EMAIL = "demo@demo.com";
+  private static final Logger LOG = Loggers.PROCESS_PROCESSOR_LOGGER;
+  private final MutableProcessingState processingState;
+  private final BCryptPasswordEncoder passwordEncoder;
```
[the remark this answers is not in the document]
remcowesterhoud: With choice you mean why we use the `BCryptPasswordEncoder`? It's because Identity uses this to encrypt passwords before sending them to Zeebe. We have to make sure we use the same encryption as Identity for this default user, as they need to match the password when a user logs in.

So the flow is:

1. Zeebe creates default user with BCrypt encoded password
2. Zeebe exports User.CREATED event containing this encoded password
3. A user logs in in Identity
4. Identity pulls the exported user from elastic
5. Identity matches the entered password with the encoded password

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/DefaultUserCreator.java
```diff
@@ -0,0 +1,80 @@
...
+                      .setName(DEFAULT_USER_USERNAME)
+                      .setEmail(DEFAULT_USER_EMAIL)
+                      .setPassword(passwordEncoder.encode(DEFAULT_USER_PASSWORD))
+                      .setUserType(UserType.DEFAULT);
+
+              taskResultBuilder.appendCommandRecord(UserIntent.CREATE, defaultUser);
+              LOG.info(
+                  "Created default user with username '{}' and password '{}'",
+                  DEFAULT_USER_USERNAME,
+                  DEFAULT_USER_PASSWORD);
```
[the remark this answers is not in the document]
remcowesterhoud: It's to let users know what the credentials for the default user are. They are not configurable at the moment. I took inspiration here from how PostgreSQL does it. They also create a default user and log the credentials of this.

I will leave this as-is for now. We'll probably change this by allowing users to specify their own default user credentials and at that point we don't need to log it.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
nicpuppa: But if you encode the password here, you encode it for every request. The default user is not created via REST API

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/DefaultUserCreator.java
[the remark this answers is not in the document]
nicpuppa: Yes, I was wondering why `BCryptPasswordEncoder ` instead of other options, thanks

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
remcowesterhoud: Oh sorry, I thought this comment was in the `DefaultUserCreator` 🤦 Forgot I had to add this here as well. I had to add this back since I accidentally removed it in a refactoring.

I'll clarify further with Identity, but will merge this in the meantime. I think it makes sense to do it here as this is a public endpoint. Users can call this without going through Identity. As a result, we cannot guarantee they will use the same encoding we need. If they'd use something different from us we cannot match the password upon login.

nicpuppa: Ok got it

PR #22338, opened 2024-09-16
docs: link to official release policy

MaxTru:
* Simplifies the maintenance process (one source of truth)
* Also see https://camunda.slack.com/archives/C05DH1F5TAR/p1726068183693239

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,35 +30,6 @@ Using Camunda 8, you can:
 
 [Learn more at camunda.com](https://camunda.com/platform/).
 
-## Release Lifecycle
-
-Our release cadence within major releases is a minor release every six months, with an alpha release on each of the five months between minor releases. Releases happen on the second Tuesday of the month, Berlin time (CET).
-
-Minor releases are supported with patches for eighteen months after their release.
-
-Here is a diagram illustrating the lifecycle of minor releases over a 27-month period:
-
-```
-2024                       2025                                2026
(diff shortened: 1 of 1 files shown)


MaxTru:
Fixing spotless issues....

korthout (approved):
This change makes sense to me 👍 Thanks @MaxTru 

🔧 I suggest to move the release policy section somewhere above the license section. I feel it fits better there, but I don't have arguments why. I guess it's something that I've seen other projects do.

🔧 I want to note that the release policy doesn't use the term `alpha` anywhere. Instead it mentions preview releases. This may be confusing when users find alpha. rather than preview releases. We could consider an additional mention in the text.

MaxTru:
Thanks @korthout 

> 🔧 I suggest to move the release policy section somewhere above the license section. I feel it fits better there, but I don't have arguments why. I guess it's something that I've seen other projects do.

Agreed, and done.

> 🔧 I want to note that the release policy doesn't use the term `alpha` anywhere. Instead it mentions preview releases. This may be confusing when users find alpha. rather than preview releases. We could consider an additional mention in the text.

True. I will do a thorough review of the release policy (see https://camunda.slack.com/archives/C02E8BHB7EW/p1726470283167709). Will involve you there.

lenaschoenburg (approved):

PR #22340, opened 2024-09-16
Stable/Operate-8.5: Fix integration tests - lazy creation FNI transformer

ChrisKujawa:
## Description

The partition IDs are not necessarly available on start, especially in IT. To avoid bootstrap issues we create the FNI transformer lazy on first use.

The chances are high on the first import that the partitions are already available.

## Related issues

On main we detected that IT tests were always failing https://github.com/camunda/camunda/pull/22203 and @ThorbenLindhauer realized that the problem is that the integration tests can't start Operate, as Zeebe partitions can't be resolved before the ApplicationContext is created.


operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/FlowNodeInstanceZeebeRecordProcessor.java
```diff
@@ -74,6 +93,7 @@ public FlowNodeInstanceZeebeRecordProcessor(
             flowNodeStore::findParentTreePathFor,
             new TreePathCacheMetricsImpl(partitionIds, metrics));
     fniTransformer = new FNITransformer(treePathCache);
+    return fniTransformer;
```
ThorbenLindhauer, line 96: Should we synchronize the creation of the transformer? Otherwise there is a race condition that we create more than one, although I don't see an immediate negative side effect of it. But is a future bug potential.

operate/importer-8_4/src/main/java/io/camunda/operate/zeebeimport/v8_4/processors/FlowNodeInstanceZeebeRecordProcessor.java
```diff
@@ -64,7 +69,21 @@ public FlowNodeInstanceZeebeRecordProcessor(
...
+  }
+
+  private FNITransformer lazyLoadFNITransformer() {
+    if (fniTransformer != null) {
+      return fniTransformer;
+    }
+
+    // We create the FNITransformer lazy, as partition holder didn't hold the partitionIds right on
+    // start. The chances are higher that when we start importing we can also request the
+    // partitions.
```
ThorbenLindhauer, line 85: I think that's not only more likely but a given. When scheduling the records readers (via `io.camunda.operate.zeebeimport.RecordsReaderHolder.getAllRecordsReaders()` and calling code), then we already create records readers per partition. Not batch makes it into processing before that.

ThorbenLindhauer (changes requested):
See the comments please.

operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/FlowNodeInstanceZeebeRecordProcessor.java
```diff
@@ -53,27 +57,59 @@ public class FlowNodeInstanceZeebeRecordProcessor {
   private static final Logger LOGGER =
       LoggerFactory.getLogger(FlowNodeInstanceZeebeRecordProcessor.class);
   private static final Set<String> AI_START_STATES = Set.of(ELEMENT_ACTIVATING.name());
-
+  private final ConcurrentInitializer<FNITransformer> fniTransformerLazy;
```
ThorbenLindhauer, line 60: Implementation looks good to me. Must also be applied to the 8.4 `FlowNodeInstanceZeebeRecordProcessor` class.

operate/importer-8_5/src/main/java/io/camunda/operate/zeebeimport/processors/FlowNodeInstanceZeebeRecordProcessor.java
```diff
@@ -53,27 +57,59 @@ public class FlowNodeInstanceZeebeRecordProcessor {
...
+  }
+
+  private FNITransformer getFNITransformer() {
+    try {
+      return fniTransformerLazy.get();
+    } catch (final ConcurrentException e) {
+      // we do not expect any exception as our instance supplier defined is not throwing any
+      // exception. If we catch this here there is something weird going on, so passing it on
+      throw new RuntimeException(
+          "Expected to retrieve FNITransformer without an error, bot caught one.", e);
```
ThorbenLindhauer, line 111: Type in the error message, should be `but` instead of `but`

ThorbenLindhauer (changes requested):

Zelldon:
@ThorbenLindhauer can you re-check plese

ralfpuchert (approved):
Besides other reviewer comments I have nothing to add.

ThorbenLindhauer (approved):
Looks all good to me, too!

PR #22343, opened 2024-09-16
deps: fix CVE-2024-38816

npepinpe:
## Description

This PR fixes a high severity CVE: https://spring.io/security/cve-2024-38816

While we shouldn't be affected by this pre-REST API, it's still good to update.


npepinpe:
Oh, 6.0.x is not supported anymore unless it's for enterprise, so we need to update to 6.1.x to fix this CVE.

If it's not much effort, I'll do that - but if it is, considering 8.2 is running out of support soon, we'll drop it.

megglos (approved):
👍

PR #22354, opened 2024-09-16
Benchmark:  Use different job types

ChrisKujawa:
## Description

Currently investigating why the benchmarks at some point spiking with a lot of publish messages, and we were realizing that several tasks share the same job type.

Whis caused to send messages multiple times, unnecessarily. The new job type is used to 'get the vendor info', needs to be adjusted on the deployed benchmark config as well.

This doesn't necessarily the problem (not clear yet), but it reduces the amount of not useful work.

### DIFF
--- a/zeebe/benchmarks/project/src/main/resources/bpmn/realistic/bankCustomerComplaintDisputeHandling.bpmn
+++ b/zeebe/benchmarks/project/src/main/resources/bpmn/realistic/bankCustomerComplaintDisputeHandling.bpmn
@@ -51,7 +51,7 @@
         <zeebe:ioMapping>
           <zeebe:input source="=&#34;dispute_unsuccessful&#34;" target="type" />
         </zeebe:ioMapping>
-        <zeebe:taskDefinition type="customer_notification" />
+        <zeebe:taskDefinition type="inform_about_failed_claim" />
       </bpmn:extensionElements>
       <bpmn:incoming>Flow_0cceqh3</bpmn:incoming>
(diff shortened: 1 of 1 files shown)


Zelldon:
Fixed also the customer_notification job types which were overused https://camunda.slack.com/archives/C07J9URA0B1/p1726491504557209?thread_ts=1726434007.020899&cid=C07J9URA0B1

rodrigo-lourenco-lopes (approved):
💭 I know that the task "Send document request reminder" is not being activated at the moment, but should we also change the task here, just in case the benchmark gets updated in the future? 

Otherwise looks good! Thanks for the fix 🚀 
We should not have to increase the replicas much more, except perhaps on the tasks "Request proof from vendor", or "Get vendor info", due to the high number of parallel requests here.

Zelldon:
> 💭 I know that the task "Send document request reminder" is not being activated at the moment, but should we also change the task here, just in case the benchmark gets updated in the future?
> 
> Otherwise looks good! Thanks for the fix 🚀 We should not have to increase the replicas much more, except perhaps on the tasks "Request proof from vendor", or "Get vendor info", due to the high number of parallel requests here.



> 💭 I know that the task "Send document request reminder" is not being activated at the moment, but should we also change the task here, just in case the benchmark gets updated in the future?

Here we need to same type as it needs to send the same message (it might got lost (which I doubt 😀 ))
> 
> Otherwise looks good! Thanks for the fix 🚀 We should not have to increase the replicas much more, except perhaps on the tasks "Request proof from vendor", or "Get vendor info", due to the high number of parallel requests here.

👍🏼

PR #22370, opened 2024-09-16
fix: exclude collapsed subprocess nodes from heatmap reports

RomanJRW:
## Description

related to [OPT#11198](https://github.com/camunda/camunda-optimize/issues/11198)

## Checklist

- [ ] for CI changes:

## Related issues

closes #


optimize/backend/src/it/java/io/camunda/optimize/service/db/es/report/process/single/flownode/frequency/groupby/flownode/FlowNodeFrequencyByFlowNodeReportEvaluationIT.java.out
```diff
@@ -105,6 +105,86 @@
 //   }
 //
 //   @Test
+//   public void testCollapsedNodesAreExcludedWhenUnsavedReportViewedViewedAsHeatmap()
```
PHWaechtler, line 108: We could consider adding a test case for a non heatmap report to ensure this filtering does not happen there, though that wouldnt cover all possible types anywway so I would also be ok not including this.

optimize/backend/src/it/java/io/camunda/optimize/service/db/es/report/process/single/flownode/frequency/groupby/flownode/FlowNodeFrequencyByFlowNodeReportEvaluationIT.java.out
```diff
@@ -105,6 +105,86 @@
...
+//     reportData.setVisualization(ProcessVisualization.HEAT);
+//     final ReportResultResponseDto<List<MapResultEntryDto>> result =
+//         reportClient.evaluateMapReport(reportData).getResult();
+//
+//     // then
+//     final List<MapResultEntryDto> resultData = result.getFirstMeasureData();
+//     // The flow nodes from the collapsed subprocess are not included
+//     assertThat(resultData).hasSize(7);
+//     final List<String> resultLabels = resultData.stream().map(MapResultEntryDto::getLabel).toList();
+//     assertThat(resultLabels)
```
PHWaechtler, line 135: Q: Are there any result metrics that could theoretically be influenced by the flownode count? Eg relative frequency, things like that where we might use the flownode count for calculations. This is not affected by the filtering correct?

HeleneW-dot (approved):
Lgtm, just have some small suggestions and questions.
One tangentially related question I have is: when adding flownode selection filters and selecting this collapsed subprocess, does that also select every ode within the subprocess? Same the other way around for deselection.

optimize/backend/src/main/java/io/camunda/optimize/service/identity/CollapsedSubprocessNodesService.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * This cache extracts all flow node IDs from a process XML that are considered hidden from view.
+ * This refers to the flow nodes within a collapsed subprocess
+ */
+@Component
+public class CollapsedSubprocessNodesService {
+
+  public static final int MAX_CACHE_SIZE = 10;
+  public static final int CACHE_EXPIRY_MINS = 10;
```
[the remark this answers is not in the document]
RomanJRW: I considered it but realised this is probably too low level for customers to care about. I don't see a risk here, given the small size and lifecycle of the cache

optimize/backend/src/main/java/io/camunda/optimize/service/identity/CollapsedSubprocessNodesService.java
```diff
@@ -0,0 +1,54 @@
...
+    } else {
+      final Set<String> collapsedSubprocessNodeIds =
+          BpmnModelUtil.getCollapsedSubprocessElementIds(reportDataDto.getConfiguration().getXml());
+      collapsedSubprocessNodesCache.put(cacheKey, collapsedSubprocessNodeIds);
+      return collapsedSubprocessNodeIds;
+    }
+  }
+
+  private static String buildCacheEntryKey(final ProcessReportDataDto reportDataDto) {
+    return reportDataDto.getDefinitionKey() + "_v" + reportDataDto.getDefinitionVersions();
```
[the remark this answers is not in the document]
RomanJRW: Q1 - I hadn't thought of this but also can't think of a good solution for it. Given that we are just hiding flow nodes though, I don't see it as a huge problem if the rendering is slightly off for a period of time. No evaluation within a ten minute period would be enough for this to correct itself.

I considered Q2 while implementing this and good find a particularly tidy solution to the problem. I went for the lazy way and instead assumed that each cache entry is likely to exist for a given report, in which the values of this list are unlikely to change frequently. If another report has slight variation of versions, then yes it would take another cache entry.

I am not planning any action given the above. But please let me know if you feel like something _does_ need doing and I could incorporate an easy fix if that is the case

optimize/backend/src/test/resources/test/data/visible_and_collapsed_subprocess.bpmn
```diff
@@ -0,0 +1,134 @@
...
+      <bpmn:sequenceFlow id="Flow_11x1roq" sourceRef="Event_1qhi2im" targetRef="Activity_009pyxh" />
+      <bpmn:sequenceFlow id="Flow_1ao92b6" sourceRef="Activity_009pyxh" targetRef="Event_08998nd" />
+      <bpmn:userTask id="Activity_009pyxh" name="visibleUserTask">
+        <bpmn:incoming>Flow_11x1roq</bpmn:incoming>
+        <bpmn:outgoing>Flow_1ao92b6</bpmn:outgoing>
+      </bpmn:userTask>
+    </bpmn:subProcess>
+    <bpmn:sequenceFlow id="Flow_0nbrned" sourceRef="StartEvent_1" targetRef="Activity_0saxiqa" />
+    <bpmn:sequenceFlow id="Flow_11a1exz" sourceRef="Activity_0saxiqa" targetRef="Activity_01fq1lc" />
+    <bpmn:subProcess id="Activity_01fq1lc" name="Collapsed Subprocess">
```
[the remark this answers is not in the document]
RomanJRW: When modelling, the user ultimately decides whether the subprocess is collapsed or expanded. If expanded, it contains an attribute `isExpanded` in the xml (scroll down to the diagrams part), which Optimize can use to determine whether or not to hide the data on heatmap reports.

The name within the BPM I chose is just to indicate that this is the subprocess I chose to collapse when modelling it

optimize/backend/src/it/java/io/camunda/optimize/service/db/es/report/process/single/flownode/frequency/groupby/flownode/FlowNodeFrequencyByFlowNodeReportEvaluationIT.java.out
RomanJRW: Fair point, I kind of went for the easy option here because of time, but agree that your suggested test case would make sense. I will not do this here (also because of time) unless you feel strongly, but it sounds like this is not the case

optimize/backend/src/it/java/io/camunda/optimize/service/db/es/report/process/single/flownode/frequency/groupby/flownode/FlowNodeFrequencyByFlowNodeReportEvaluationIT.java.out
RomanJRW: I actually quite deliberately did not want to modify anything other than the view here. My reasoning is that the user could quite easily switch to another visualisation, such as a bar chart, and I don't think that any data itself should actually be different in this case. To me it would be odd behaviour for the user to change a visualisation and get a different result

RomanJRW:
> when adding flownode selection filters and selecting this collapsed subprocess, does that also select every ode within the subprocess? Same the other way around for deselection

I don't know the answer to this, but is it related to these changes? The hidden ones should always be hidden, regardless of the filtering

optimize/backend/src/it/java/io/camunda/optimize/service/db/es/report/process/single/flownode/frequency/groupby/flownode/FlowNodeFrequencyByFlowNodeReportEvaluationIT.java.out
PHWaechtler: Should be fine if we cover this during QA imo

optimize/backend/src/it/java/io/camunda/optimize/service/db/es/report/process/single/flownode/frequency/groupby/flownode/FlowNodeFrequencyByFlowNodeReportEvaluationIT.java.out
PHWaechtler: Yeah I agree I would also not want to have the metrics affected, I think my line of thinking was more to evaluate whether it makes sense to assert this in the test as well. The way I currently understand it though the filtering of the flownodes in the results is logically far enough removed from the calculation of the other metrics that an assertion isnt necessary

PR #22371, opened 2024-09-16
Optimize upgrade log fixes

RomanJRW:
## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/upgrade/src/main/java/io/camunda/optimize/upgrade/steps/UpgradeStep.java
+++ b/optimize/upgrade/src/main/java/io/camunda/optimize/upgrade/steps/UpgradeStep.java
@@ -27,7 +27,7 @@ public abstract class UpgradeStep {
   // failing conversion of test indices that don't have lookups
   @VisibleForTesting @Setter protected boolean skipIndexConversion = false;
 
-  protected UpgradeStep(IndexMappingCreator index) {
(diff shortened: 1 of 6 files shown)


optimize/upgrade/src/main/java/io/camunda/optimize/upgrade/steps/UpgradeStep.java
```diff
@@ -27,16 +27,16 @@ public abstract class UpgradeStep {
...
   }
 
   public abstract UpgradeStepType getType();
 
   protected abstract void performUpgradeStep(SchemaUpgradeClient<?, ?> schemaUpgradeClient);
 
-  public void execute(SchemaUpgradeClient<?, ?> schemaUpgradeClient) {
-    if (!skipIndexConversion) {
+  public void execute(final SchemaUpgradeClient<?, ?> schemaUpgradeClient) {
+    if (!skipIndexConversion && index != null) {
```
RomanJRW, line 39: this is the fix. There are two subclasses that do not set an index property because they relate to index/template deletion. This then fails on execution. For such cases, we do not want to try to convert an index because no index exists.

The other commit in this PR is unrelated, but uses the updated logback property since the version has been updated

igpetrov (approved):
Thank you for the prompt fix

PR #22406, opened 2024-09-17
[CPT] Use a single Camunda Docker container

saig0:
## Description

Replace the Zeebe, Operate, and Tasklist containers with the single Camunda Docker container that bundles all images together. 

- Use the Docker image `camunda/camunda`
- Set all env vars from Zeebe, Operate, and Tasklist
- Use the API of the Camunda Docker container

Why?
- Remove code
- Prepare for multi-tenancy and v2 REST API

**Reviewer note:**
Most changes are only refactorings to rename the container/properties/methods/fields and to remove the containers. Instead of going over all commit, it is better to check only the result.


remcowesterhoud (approved):
Changes look good to me. Sorry for taking too long 😉

PR #22410, opened 2024-09-17
fix: merge concurrent resolves for up to 128 hostnames

lenaschoenburg:
We make many requests, each of which requires a resolved hostname. To avoid flooding the DNS server with many redundant queries, we can enable request that netty merges the requests to resolve hostnames. A second request will wait for and use the result of the first request.

I've chosen 128 as a magic number because it's a somewhat realistic upper bound on the number of hosts we'd look up (brokers + gateways) and is still low enough to not waste too much memory.

Relates to https://jira.camunda.com/browse/SUPPORT-22744


npepinpe:
I didn't get from their docs whar's the cache policy, but simplifiex, it means after 128 in flight requests for different hosts, it would stop caching and do every other request without consolidation? (until there is spacd in the cache)

lenaschoenburg:
Yes, that's my understanding too.

npepinpe (approved):

PR #22411, opened 2024-09-17
Fix realistic benchmark process

ChrisKujawa:
## Description

The process changes we recently did, caused PI didn't complete anymore. This caused at some point the clusters where crash due to their large state. 

This PR adjusts the process model to make more sense in general to the used workers and calculation.

Changes:

 * Multi instance on outer sub process doesn't make much sense, as
DMN is executed multiple times, overwriting the respective value.
Reduces also the load.
* To run the right path in the process some variables have been adjusted
  * To calculate a better rating adjust the vendor claim frequency
  * To have a better rating, and cover the bigger dispute imputs we adjust the values
* Add input mappings to the sub processes to avoid concurrent writes
  * Without the definition as input mappings, the variables are created on the root scope, causing to overwrite each other.

Running with the changes we can see that instances are completed again:

Metrics:

## Related issues

related #21472 
related [thread](https://camunda.slack.com/archives/C037W9NMATG/p1726587900515399)

### DIFF
--- a/zeebe/benchmarks/project/src/main/resources/bpmn/realistic/bankCustomerComplaintDisputeHandling.bpmn
+++ b/zeebe/benchmarks/project/src/main/resources/bpmn/realistic/bankCustomerComplaintDisputeHandling.bpmn
@@ -57,6 +57,11 @@
       <bpmn:outgoing>Flow_1cvob8y</bpmn:outgoing>
     </bpmn:sendTask>
     <bpmn:subProcess id="Activity_199882v" name="Document Request Process">
+      <bpmn:extensionElements>
+        <zeebe:ioMapping>
+          <zeebe:input source="=null" target="correlationKey" />
+        </zeebe:ioMapping>
+      </bpmn:extensionElements>
       <bpmn:incoming>Flow_15wpx65</bpmn:incoming>
       <bpmn:outgoing>Flow_06bdzt7</bpmn:outgoing>
       <bpmn:startEvent id="Event_1g0dyzc" name="documents required">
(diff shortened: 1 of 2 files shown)


Zelldon:
You can watch it [here](https://grafana.dev.zeebe.io/d/zeebe-dashboard/zeebe?orgId=1&var-DS_PROMETHEUS=prometheus&var-cluster=All&var-namespace=ck-realistic-benchmark-main&var-pod=All&var-partition=All&from=now-30m&to=now)

Zelldon:
Results look good (created yesterday at 6 PM CET, and still running with a good load).

benchmark-general

The load is pretty stable. We are able to complete a lot of jobs per second 
pi-jobs.

I realized that our dashboard in the general section is only looking at service tasks

servicetask

When we look at all tasks then this matches with the job count as well.

task


The snapshot size is not increasing, we seem to be able to complete all instances (no left overs, messages are correlated, etc.)
snapshot


We can see that we need to adjust a bit the worker configurations, as mentioned in slack [here](https://camunda.slack.com/archives/C07J9URA0B1/p1726586834503369?thread_ts=1726434007.020899&cid=C07J9URA0B1). I will do this next.

log

We have a lot of job completion rejections because likely they timed out before and have been completed twice. I will reduce the queue size, and maybe scale the workers or increase the thread count.

Zelldon:
Based on the metrics I would say the workers are not really busy, so I guess increasing threads would make sense here

Zelldon:
@deepthidevaki do you have maybe time to review/approve this? So I can move forward with the benchmarks

rodrigo-lourenco-lopes (approved):
Thanks for the fix! The results look promising, lets give it a try 🚀

PR #22434, opened 2024-09-18
feat: validate deployment-bound resource links at deploy time

### LINKED ISSUE #19384: [EPIC] Support `deployment` version binding for linked resources

### LINKED ISSUE #19712: Validate deployment-bound resource links at deploy time
### Description
Validate at deploy time that resource links with `bindingType=deployment` can actually be resolved within the same deployment. If at least one resource is missing, the deployment request should fail.

The goal is to inform users early about problems with their deployment and prevent later incidents at runtime (that would occur if a deployment-bound link could not be resolved due to a missing resource).

_Note_: The ID of a target resource (e.g. a [called process ID](https://docs.camunda.io/docs/next/components/modeler/bpmn/call-activities/#defining-the-called-process)) could be an expression (which is only evaluated at runtime); such a link cannot be validated at deploy time.

#
This is a _should-have_ requirement; see the following user story from https://github.com/camunda/product-hub/issues/1920:
  > (S) [Missing Resource] When using a Zeebe client to deploy a process and its dependent resources, the deployment fails if a dependent resource is not available

### Acceptance Criteria
- [x] For every BPMN resource in a deployment request, all contained resource links with `bindingType=deployment` are validated, i.e. it is checked that the target resource is part of the same deployment. This applies to:
  - [x] called processes in call activities
  - [x] called decisions in business rule tasks
  - [x] linked forms in user tasks
- [x] If at least one resource link cannot be resolved, the deployment request is rejected.
- [x] For each unresolvable link, a meaningful error is added to the response.
  - Example:
  ```
  'test.bpmn':
  - Element: myCallActivity > extensionElements > calledElement
      - ERROR: Expected to find process with id 'test-process' in the current deployment, but not found.
  - Element: myBusinessRuleTask > extensionElements > calledDecision
      - ERROR: Expected to find decision with id 'test-decision' in the current deployment, but not found.
  ```

### Dependencies
- relates to https://github.com/camunda/camunda/issues/19859
- relates to https://github.com/camunda/camunda/issues/19710

jfriedenstab:
## Description
When a BPMN process gets deployed, validate that all linked resources (e.g. a called process in a call activity) are part of the same deployment if the binding type `deployment` is used. If at least one resource is missing, the deployment will be rejected.

## Related issues
closes #19712
closes #19384

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
@@ -28,8 +28,6 @@ import java.time.InstantSource;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.function.BiFunction;
-import java.util.function.Function;
 import org.agrona.DirectBuffer;
 import org.slf4j.Logger;
 
@@ -113,27 +111,18 @@ public final class DeploymentTransformer {
     // records are being written yet)
     while (resourceIterator.hasNext()) {
       final DeploymentResource deploymentResource = resourceIterator.next();
-      success &=
-          transformResource(
-              deploymentEvent,
-              errors,
-              deploymentResource,
-              transformer -> transformer::createMetadata);
+      success &= createMetadata(deploymentResource, deploymentEvent, errors);
     }
 
     // step 2: update metadata (optionally) and write actual event records
     if (success) {
       for (final DeploymentResource deploymentResource : deploymentEvent.resources()) {
-        success &=
-            transformResource(
-                deploymentEvent,
-                errors,
-                deploymentResource,
-                transformer -> transformer::writeRecords);
+        final var transformer = getResourceTransformer(deploymentResource.getResourceName());
+        transformer.writeRecords(deploymentResource, deploymentEvent);
       }
-    }
+      return Either.right(null);
 
-    if (!success) {
+    } else {
       rejectionType = RejectionType.INVALID_ARGUMENT;
       rejectionReason =
           String.format(
@@ -141,24 +130,17 @@ public final class DeploymentTransformer {
 
       return Either.left(new Failure(rejectionReason));
     }
-
-    return Either.right(null);
   }
 
-  private boolean transformResource(
-      final DeploymentRecord deploymentEvent,
-      final StringBuilder errors,
+  private boolean createMetadata(
       final DeploymentResource deploymentResource,
-      final Function<
-              DeploymentResourceTransformer,
-              BiFunction<DeploymentResource, DeploymentRecord, Either<Failure, Void>>>
-          transformation) {
+      final DeploymentRecord deploymentEvent,
+      final StringBuilder errors) {
     final var resourceName = deploymentResource.getResourceName();
     final var transformer = getResourceTransformer(resourceName);
 
     try {
-      final var result =
-          transformation.apply(transformer).apply(deploymentResource, deploymentEvent);
+      final var result = transformer.createMetadata(deploymentResource, deploymentEvent);
 
       if (result.isRight()) {
         return true;
@@ -196,20 +178,13 @@ public final class DeploymentTransformer {
     @Override
     public Either<Failure, Void> createMetadata(
         final DeploymentResource resource, final DeploymentRecord deployment) {
-      return createUnknownResourceTypeFailure(resource);
-    }
-
-    @Override
-    public Either<Failure, Void> writeRecords(
-        final DeploymentResource resource, final DeploymentRecord deployment) {
-      return createUnknownResourceTypeFailure(resource);
-    }
-
-    private Either<Failure, Void> createUnknownResourceTypeFailure(
-        final DeploymentResource resource) {
       final var failureMessage =
           String.format("%n'%s': unknown resource type", resource.getResourceName());
       return Either.left(new Failure(failureMessage));
     }
+
+    @Override
+    public void writeRecords(
+        final DeploymentResource resource, final DeploymentRecord deployment) {}
   }
 }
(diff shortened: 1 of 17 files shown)


jfriedenstab:
Hi @korthout,
I've assigned this PR for review directly to you, because you're probably most familiar with the deployment binding feature, but feel free to re-assign it to someone else from your team 🙂.

megglos:
@berkaycanbc and @ana-vinogradova-camunda  will do a pair review this week.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
```diff
@@ -111,54 +111,68 @@ public Either<Failure, Void> transform(final DeploymentRecord deploymentEvent) {
...
     if (success) {
       for (final DeploymentResource deploymentResource : deploymentEvent.resources()) {
-        success &=
-            transformResource(
-                deploymentEvent,
-                errors,
-                deploymentResource,
-                transformer -> transformer::writeRecords);
+        final var transformer = getResourceTransformer(deploymentResource.getResourceName());
+        transformer.writeRecords(deploymentResource, deploymentEvent);
```
ana-vinogradova-camunda, line 147: ❓ It is not entirely clear why we don't need a transformation here, could you please explain this change in more details? Was it already unnecessary in the previous implementation?

ana-vinogradova-camunda (commented):
Thank you for the changes and refactoring 🚀 
🔧 We've written down some suggestions, could you please take a look?

❓ There is also one clarification question, could you please assist with that?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
jfriedenstab: Good question :). I tried to explain the reasoning behind this change in the description of https://github.com/camunda/camunda/pull/22434/commits/d9c86ffbfe05632d479542c9c0509142606b68da.

The `writeRecords` methods were always returning `Either.right(null)` (so never a `Failure`) already before which means that they could never change `success` to `false` anyway. So it seemed fine to make them just return `void`.

Another problem was that `DeploymentResourceTransformer.createMetadata` now has three method parameters instead of two, so the `BiFunction` here did not work any more for both `createMetadata` and `writeRecords`:

https://github.com/camunda/camunda/blob/59b0b3e0e845b11a32e36c0e33c23b17878ccf2d/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java#L148-L155

I think it would not have been easily possible to make the `transformResource` method work for both steps (`createMetadata` and `writeRecords`).

jfriedenstab: What I just realized is that the code does not handle any unexpected errors that might occur in `writeRecords` any more (only for `createMetadata`):

https://github.com/camunda/camunda/blob/ad07c9d152d7ec346f552d9cbdca8ac8444cd1e9/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java#L185-L187

I'll fix this and add the same kind of exception handling.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DeploymentRejectionTest.java
```diff
@@ -274,4 +277,133 @@ public void shouldDoAtomicDeployments() {
...
+        Bpmn.createExecutableProcess("process-2")
+            .startEvent()
+            .userTask(
+                "userTask",
+                builder ->
+                    builder
+                        .zeebeFormBindingType(ZeebeBindingType.deployment)
+                        .zeebeFormId("test-form"))
+            .endEvent()
+            .done();
```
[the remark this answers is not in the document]
jfriedenstab: Good suggestion 👍🏻. I'll split up the test.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
ana-vinogradova-camunda: @jfriedenstab Thank you for the explanation and for adding error handling 👍

ana-vinogradova-camunda (approved):
LGTM, approving the PR, to the best of my knowledge - I see no further questions here 👍

PR #22456, opened 2024-09-18
feat: improve java client exceptional completion

### LINKED ISSUE #21173: Improve Java client with REST API exceptional flow

As a user, if I'm using the REST API it's possible that my request gets denied for several reasons. Using the Java client the way I'd make my request is like this:

```java
client
    .newFooCommand()
    .propertyBar(bar)
    .send()
```

This `send()` method will return me a `ZeebeFuture<FooResponse>`. There's no guarantee my request succeeded so I need to do some exception handling when joining this future:

```java
try {
    final var response = future.join();
} catch (ProblemException e) {
   // handle unhappy path
}
```

The above is what I'd expect to be able to do. However, the exception thrown by the future is not a `ProblemException`. Instead I get a `CompletionException` with the `ProblemException` as a cause. That means if I as a user want to know what the problem was, I need to handle the exceptional future like this:

```java
try {
    final var response = future.join();
} catch (CompletionException e) {
    if (e.getCause() instanceof ProblemException) {
    final ProblemException acutalException = (ProblemException) e.getCause();
        // handle unhappy path
    }
}
```

It's quite tedious for users to get to the original failure of the request.

Joining the future should return the `ProblemException` directly in cases like this.

<details><summary><b>Example test case</b></summary>

```java
  @Test
  void shouldNotCorrelateWhenNoExistingSubscription() {
    // given
    final var messageName = "messageName";
    final var correlationKey = "correlationKey";

    // when
    final var responseFuture =
        client
            .newCorrelateMessageCommand()
            .messageName(messageName)
            .correlationKey(correlationKey)
            .send();

    //     then
    assertThatThrownBy(responseFuture::join)
        .isInstanceOf(ProblemException.class)
        .hasMessageContaining("title: NOT_FOUND")
        .hasMessageContaining("status: 404")
        .hasMessageContaining(
            "Expected to find subscription for message with name '%s' and correlation key '%s', but none was found."
                .formatted(messageName, correlationKey));
  }
```

</details>

koevskinikola:
## Description

Improve the Zeebe Java client handling of failures during REST API calls.

With the addition of REST API support for the Java client, we added a new `HttpZeebeFuture` that extends `CompletableFuture`. We directly use the `CompletableFuture#join()` there, and it throws `CompletionExceptions`. In contrast, when using gRPC, the java client calls `CompletableFuture#get()` instead, and it throws `ExecutionExceptions` when a request fails.

We unwrap `ExecutionExceptions`, but don't touch `CompletionExceptions`.

This PR aligns the behavior and handling of exceptional REST API calls with gRPC calls in the Zeebe Java client.

## Checklist

- [ ] for CI changes:

## Related issues

closes #21173

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/HttpZeebeFuture.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/HttpZeebeFuture.java
@@ -60,7 +60,7 @@ public class HttpZeebeFuture<RespT> extends CompletableFuture<RespT> implements 
     try {
       return super.get(timeout, unit);
     } catch (final ExecutionException e) {
-      throw new ClientException(e);
+      throw unwrapExecutionException(e);
     } catch (final InterruptedException e) {
       Thread.currentThread().interrupt();
       throw new ClientException("Failed: interrupted while awaiting response", e);
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/HttpZeebeFuture.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/HttpZeebeFuture.java
@@ -17,6 +17,7 @@ package io.camunda.zeebe.client.impl.http;
 
 import io.camunda.zeebe.client.api.ZeebeFuture;
 import io.camunda.zeebe.client.api.command.ClientException;
+import io.camunda.zeebe.client.api.command.ProblemException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
@@ -34,6 +35,27 @@ public class HttpZeebeFuture<RespT> extends CompletableFuture<RespT> implements 
   private volatile Future<?> transportFuture;
 
   @Override
+  public RespT join() {
+    try {
+      return get();
+    } catch (final ExecutionException e) {
+      throw unwrapExecutionException(e);
+    } catch (final InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new ClientException("Failed: interrupted while awaiting response", e);
+    }
+  }
+
+  @Override
+  public boolean cancel(final boolean mayInterruptIfRunning) {
+    if (transportFuture != null) {
+      transportFuture.cancel(mayInterruptIfRunning);
+    }
+
+    return super.cancel(mayInterruptIfRunning);
+  }
+
+  @Override
   public RespT join(final long timeout, final TimeUnit unit) {
     try {
       return super.get(timeout, unit);
@@ -47,21 +69,21 @@ public class HttpZeebeFuture<RespT> extends CompletableFuture<RespT> implements 
     }
   }
 
-  @Override
-  public boolean cancel(final boolean mayInterruptIfRunning) {
-    if (transportFuture != null) {
-      transportFuture.cancel(mayInterruptIfRunning);
-    }
-
-    return super.cancel(mayInterruptIfRunning);
-  }
-
   public void transportFuture(final Future<?> httpFuture) {
     transportFuture = httpFuture;
 
     // possibly we were already cancelled between calls
     if (isCancelled()) {
       httpFuture.cancel(true);
+    }
+  }
+
+  private ProblemException unwrapExecutionException(final ExecutionException e) {
+    final Throwable cause = e.getCause();
+    if (cause instanceof ProblemException) {
+      throw (ProblemException) cause;
+    } else {
+      throw new ClientException(cause);
     }
   }
 }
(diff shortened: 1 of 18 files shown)


koevskinikola:
Some more details on the internals:

In the Zeebe Java client `ZeebeClientFutureImpl#join()` we call `super.get()` instead of `super.join()` ([src](https://github.com/camunda/camunda/blob/1376ef75260543ddb917e165f5c02d5a4c331f73/clients/java/src/main/java/io/camunda/zeebe/client/impl/ZeebeClientFutureImpl.java#L52))

The differences between CompletableFuture#get() and CompletableFuture#join() seem to be that:
* get()
   * Failures result in an ExecutionException
   * Waits if necessary for the future to complete
   * Can be interrupted
* join()
   * Failures result in a CompletionException
   * Can't be interrupted

When we added REST API support for the Java client, we added a new `HttpZeebeFuture` that extends `CompletableFuture`. We directly use the `CompletableFuture#join()` there, but we override the `CompletableFuture#join(timeout, unit)` and call `super.get(timeout, unit)` within.

**Decision**

We decided to align the `ZeebeClientFutureImpl` and `HttpZeebeFuture` implementations to both call `get()` internally when `join()` is invoked. The reasons are the following:
* It aligns with the gRPC behavior in regards to interuptability and exception handling.
* Both classes implement the `ZeebeFuture`, i.e. the `Future` interface which only defines the `get()`/`get(timeout, timeunit)` methods.

koevskinikola:
@tmetzke just something to raise with these changes:

⚠️ This PR will introduce breaking changes, i.e the Topology and UserTask Java client commands that were already introduced with Zeebe `8.5.0` were returning `CompletionException`. They will now return a `ProblemException`. This change can be seen with commit [16949b3](https://github.com/camunda/camunda/pull/22456/commits/16949b386849a6e6811d850a7cf1ab4b8cfc5968) as I needed to refactor several tests due to the changes.

This might break the code of some users that already decided to use our REST API.

Are we okay with that?
- I think this should be acceptable as with Camunda `8.6.0` we will release `v2` of the REST API spec.

tmetzke:
> @tmetzke just something to raise with these changes:
> 
> ⚠️ This PR will introduce breaking changes, i.e the Topology and UserTask Java client commands that were already introduced with Zeebe `8.5.0` were returning `CompletionException`. They will now return a `ProblemException`. This change can be seen with commit [16949b3](https://github.com/camunda/camunda/pull/22456/commits/16949b386849a6e6811d850a7cf1ab4b8cfc5968) as I needed to refactor several tests due to the changes.
> 
> This might break the code of some users that already decided to use our REST API.
> 
> Are we okay with that?
> 
> * I think this should be acceptable as with Camunda `8.6.0` we will release `v2` of the REST API spec.

As discussed offline:

* Yes, this is breaking user space in exceptional cases.
* I agree that the breaking change is never ideal but it is acceptable in this case.
* Let's document the behavior change in the migration guide.

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/HttpZeebeFuture.java
```diff
@@ -56,6 +55,20 @@ public boolean cancel(final boolean mayInterruptIfRunning) {
     return super.cancel(mayInterruptIfRunning);
   }
 
+  @Override
+  public RespT join(final long timeout, final TimeUnit unit) {
+    try {
+      return super.get(timeout, unit);
+    } catch (final ExecutionException e) {
+      throw new ClientException(e);
```
tmetzke, line 63: 🔧 Don't we need to unwrap the exception here as well using `unwrapExecutionException`?

tmetzke: That is also what we do in the `ZeebeClientFutureImpl` ([src](https://github.com/camunda/camunda/blob/1376ef75260543ddb917e165f5c02d5a4c331f73/clients/java/src/main/java/io/camunda/zeebe/client/impl/ZeebeClientFutureImpl.java#L74)).
I'll adjust this, add a test case, and merge this PR if it succeeds.

tmetzke:
```suggestion
      throw unwrapExecutionException(e);
```

clients/java/src/test/java/io/camunda/zeebe/client/ExceptionHandlingRestTest.java
```diff
@@ -0,0 +1,55 @@
...
+        () -> new ProblemDetail().title("Invalid request").status(400));
+
+    // when / then
+    assertThatThrownBy(() -> client.newTopologyRequest().useRest().send().join())
+        .isInstanceOf(ProblemException.class)
+        .hasMessageContaining("Invalid request");
+  }
+
+  @Test
+  public void shouldProvideProblemExceptionOnFailedRequest() {
```
tmetzke, line 43:
```suggestion
  public void shouldProvideProblemExceptionOnFailedRequestWithTimeout() {
```

tmetzke (approved):

PR #22458, opened 2024-09-18
fix: use new micrometer metric names

### LINKED ISSUE #22164: ExecutionLatencyMetrics are broken
There are no values shown for the latest benchmark: 
image

Likely broken due to https://github.com/camunda/camunda/pull/21029 or https://github.com/camunda/camunda/pull/22016

ChrisKujawa:
## Description

Metric names have been changed, as we migrated the MetricsExporter to micrometer. Histograms and timers now contain always `_seconds` in their name.

Adding additional queries to panels, to support both metrics (for backward compatibility)

fix-execution-metrics

closes https://github.com/camunda/camunda/issues/22164

### DIFF
--- a/monitor/grafana/zeebe.json
+++ b/monitor/grafana/zeebe.json
@@ -6251,8 +6251,7 @@
                 "mode": "absolute",
                 "steps": [
                   {
-                    "color": "green",
-                    "value": null
+                    "color": "green"
                   },
                   {
                     "color": "red",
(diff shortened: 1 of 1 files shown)


lenaschoenburg:
This would close #22164 but i think @npepinpe was looking for a better solution where micrometer would not generate new names.

Zelldon:
Hf+gl;

https://github.com/micrometer-metrics/micrometer/issues/1063

Zelldon:
Seems to be possible with https://docs.micrometer.io/micrometer/reference/implementations/prometheus.html#_the_prometheus_rename_filter


But not sure about that, as we couple our self again to prometheus which we wanted to reduce/remove afaik 🤔

npepinpe (approved):
:+1: I think @Zelldon raises a fair point. We can move forward with this, so long as the dashboard supports both versions.

PR #22459, opened 2024-09-18
feat: load authenticated user on the authorizations

### LINKED ISSUE #22452: Authenticated user is loaded into the command authentication
### Description
To be able to support the application of authorizations at a command level the processors need to be able to access the authentication data supplied with the request.

In the instance of basic authentication we retrieve the user in the security classes and have access to this on the request classes, the user key should be included in the `command.getAuthorizations()` method.

### Acceptance criteria
- [ ] The user key of the user making the request is set on the `authorizations` field of a command

Ben-Sheppard:
## Description

To be able to support the application of authorizations at the processor level, we need to load specific authentication data onto the `Authentication` object that is passed through on the command.

In this PR I have used the existing methods to get the data down into the command (the crafted JWT) however soon we will want this to be the users token directly to support OIDC implementations (and mapping rules) but having looked there would be a few more changes and I want to make progress here so we at least have the data in some form.

The changes in this PR mean that you are able to access the authenticated user key in the commands via:
```
command.getAuthorizations().getOrDefault(Authorization.AUTHORIZED_USER_KEY, -1L);
```
Which then supports the retrieval of the authorizations for this user from state to be applied to the specific command.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22452


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -373,14 +375,21 @@ public static Either<ProblemDetail, BroadcastSignalRequest> toBroadcastSignalReq
...
-            .encode();
-    return new Builder().token(token).tenants(authorizedTenants).build();
+            .withClaim(Authorization.AUTHORIZED_TENANTS, authorizedTenants);
+
+    final var authenticatedUser =
+        (CamundaUser) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+
+    if (authenticatedUser != null) {
+      token.withClaim(Authorization.AUTHORIZED_USER_KEY, authenticatedUser.getUserKey());
+    }
```
remcowesterhoud: ❌  I believe we can make this method return an `Either<ProblemDetail, Authentication>`. If authorization is enabled and we don't have an authenticated user we can fail early with a 403 Forbidden. That makes more sense than sending it to Zeebe who will respond with a 401 Unauthorized.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -373,14 +375,21 @@ public static Either<ProblemDetail, BroadcastSignalRequest> toBroadcastSignalReq
...
+
+    return new Builder().token(token.build()).tenants(authorizedTenants).build();
```
remcowesterhoud, line 393: ❓ Do we have a followup issue to not put this in a JWT anymore?

remcowesterhoud (changes requested):
Thanks for prioritizing this @Ben-Sheppard 🙇 I have a few comments, please have a look and let me know what you think!

zeebe/auth/src/test/java/io/camunda/zeebe/auth/JwtAuthorizationTest.java
```diff
@@ -110,6 +110,7 @@ public void shouldFailJwtTokenValidationWithNoAuthorizedTenants() {
             .withIssuer(DEFAULT_ISSUER)
             .withAudience(DEFAULT_AUDIENCE)
             .withSubject(DEFAULT_SUBJECT)
+            .withClaim(Authorization.AUTHORIZED_USER_KEY, 1234L)
```
remcowesterhoud, line 113: 🔧 I'm a bit surprised to see this in this test case. According to the name this verifies `shouldFailJwtTokenValidationWithNoAuthorizedTenants`.

It's more clear to add a new test case to try adding the authorized user key.

Ben-Sheppard: This was accidentally left in for this test, I need to add a separate one for the user key as you suggest :)

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
Ben-Sheppard: There isn't a dedicated issue at the moment but I will sort that out

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
remcowesterhoud: I can live with that for now. I do think we should improve it later on 👍

remcowesterhoud (approved):
Thanks @Ben-Sheppard. Looks good to me 🍪

PR #22462, opened 2024-09-18
feat: Implemented ExporterFactories feature #22446

### LINKED ISSUE #22446: Refactor Exporter Creation Handling to be able to use Spring Context
Because our RDBMS Exporter (Other Branch - https://github.com/camunda/product-hub/issues/2439) needs access to the Spring Context, we need to refactor / extend the process of how exporters get created. 

Idea is to have a new ExporterFactory Interface which can be implemented and defined as Spring Bean. 
Those Factory should then be used in the ExporterDescriptor to create new instances. 
Since the ExporterFactory is a Spring Bean, it can forward needed Beans to the created instance. 

ExporterFactory drawio

p-wunderlich:
- This is an alternative way to create new instances of exporters

## Description

When zeebe creates exporters, it now first looks if there is an ExporterFactory in the spring context for the exporter id. 
If yes, it uses newInstance on this factory. This makes it possible to pass spring beans to the exporter during creation. 
If no, the old default behaviour is to simply create the exporter by the provided className.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22446

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
+++ b/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
@@ -11,6 +11,8 @@ import io.atomix.cluster.AtomixCluster;
 import io.camunda.application.commons.configuration.BrokerBasedConfiguration;
 import io.camunda.identity.sdk.IdentityConfiguration;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
+import io.camunda.zeebe.broker.exporter.repo.ExporterDescriptor;
+import io.camunda.zeebe.broker.exporter.repo.ExporterRepository;
 import io.camunda.zeebe.broker.system.SystemContext;
 import io.camunda.zeebe.scheduler.ActorScheduler;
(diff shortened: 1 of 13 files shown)


romansmirnov (changes requested):
Thanks for raising the PR. In general, I like the factory pattern that allows exporters to be created as needed, enabling them to inject beans.

Instead of passing the factories to the repository, can we leverage the `ExporterDescriptor` and pass a list of pre-configured descriptors to it? That way, the repository behavior can remain the same, and it does not deal with resolving whether a factory or a class must be used.

The descriptor accepts an exporter class, which could be wrapped into a default factory, or a factory that is created outside of the exporter.

I implemented a first idea therefore, however, I can park it on another branch.

Another problem to solve is to enable an exporter by initializing it from another exporter's state. It is currently expected that both importers will be of the same type. This needs to be solved with the factory approach as well.

Let's follow up on these topics.

romansmirnov (approved):
Thanks for the contribution. I like the approach of passing `ExporterDescriptor`s to the repository. It keeps the implementation simple. I have tested it locally, it works! LGTM! 🚀

PR #22463, opened 2024-09-18
feat: add key to the tasklist-form

nathansandi:
## Description

- Add `key` to `tasklist-form` indice
- We wont migrate data now, this is in order to cover the default sorting for V2
- we still going to use `id` for Get Form by Key , once all data is there (but the research only will cover linked forms)
- I added the key as Long, and setup the importers to import the formKey for 8.6.0 

For embeeded form new data, and linked old data, for now, key will be set as a null 

image

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/tasklist/common/src/main/java/io/camunda/tasklist/entities/FormEntity.java
+++ b/tasklist/common/src/main/java/io/camunda/tasklist/entities/FormEntity.java
@@ -13,6 +13,7 @@ import java.util.Objects;
 
 public class FormEntity extends TenantAwareTasklistEntity<FormEntity> {
 
+  private Long key;
   private String bpmnId;
   private String processDefinitionId;
   private String schema;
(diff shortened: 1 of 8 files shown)


panagiotisgts (approved):
LGTM!

One nit pick, in the `tasklist-task` index we are using the term `formKey` for forms that are embedded.
It could be confusing since for embedded forms in the `form` index the `formKey` is NULL in the database and a String in the form of `camunda-forms:bpmn:{ID}`.

In the rest of the cases the `formKey` field match the `formKey` in the `form` index.

:grey_question: Maybe something to be discussed in order to streamline the meaning of these fields.

nathansandi:
> LGTM!
> 
> One nit pick, in the `tasklist-task` index we are using the term `formKey` for forms that are embedded. It could be confusing since for embedded forms in the `form` index the `formKey` is NULL in the database and a String in the form of `camunda-forms:bpmn:{ID}`.
> 
> In the rest of the cases the `formKey` field match the `formKey` in the `form` index.
> 
> ❔ Maybe something to be discussed in order to streamline the meaning of these fields.

I agree with you, this will be changed in tasklist-list-view, This was a decision in the past and the embedded forms wont be supported in the new V2 API, which would solve this problem 

For now, on tasklist-task we had to support embedded and linked ones (thats the reason we may have 2 different formats). You can check more details over the Form Importer. 

So far my expectation as soon as we remove job worker from the modeler, we will clean this from the importer. Also further version of data ingestions.

PR #22475, opened 2024-09-19
fix: use initial deployment key in metadata of duplicate resources

### LINKED ISSUE #22473: Duplicate resources refer to current `deploymentKey` instead of the one from initial deployment
-- more context: [Slack thread](https://camunda.slack.com/archives/C075XEP2XGU/p1722866181089019) --

**Describe the bug**
The metadata of a duplicate resource in a `DeploymentRecord` (e.g. `ProcessMetadata`) refers to the `deploymentKey` of the current deployment rather than the one it was initially deployed in. This is somehow inconsistent as duplicate resources are not actually changed with the current deployment and the `deploymentKey` in the persisted resource record (e.g. `ProcessRecord`) also contains the initial deployment key.

**To Reproduce**
1. Deploy a BPMN resource -> The `ProcessMetadata` in the `DeploymentRecord` will contain the `deploymentKey` of this deployment.
2. Deploy the same resource again (without changes):
    - The resource will be marked as a duplicate.
    - The `ProcessMetadata` in the `DeploymentRecord` will contain the `deploymentKey` of the second deployment instead of the initial one.

The behavior is the same for BPMN, DMN, and Forms.

**Expected behavior**
For resource versions marked as duplicate, the `deploymentKey` in the metadata always refers to the deployment the version of the resource was initially deployed in.

jfriedenstab:
## Description
For resource versions marked as duplicates, use the deployment key of the deployment they were initially deployed with in the metadata that is added to the current deployment record.

## Related issues
closes #22473

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/BpmnResourceTransformer.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/BpmnResourceTransformer.java
@@ -133,7 +133,8 @@ public final class BpmnResourceTransformer implements DeploymentResourceTransfor
                     .setVersion(
                         processState.getNextProcessVersion(
                             metadata.getBpmnProcessId(), deployment.getTenantId()))
-                    .setDuplicate(false);
+                    .setDuplicate(false)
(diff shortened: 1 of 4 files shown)


korthout (approved):
Thanks! Looks good to me 🚀

PR #22480, opened 2024-09-19
feat: get task by key - endpoint

### LINKED ISSUE #22481: [C8-Rest-API] Implement Get Task by Key
### Description
Implement the `GET /v2/user-tasks/{userTaskKey}` endpoint for retrieving a User Task by its key. This endpoint should return detailed User Task Item (Entity returned in Search User Task

---

### Endpoint
`GET /v2/user-tasks/{userTaskKey}`

#### Request
- **Path Parameter:**
  - `userTaskKey` (Long): The unique key for the User Task that should be retrieved.

#### Response
- **Success (200 OK):**
  - Returns a JSON object representing User Task 

Same Response as Search User Task, but a single item, instead of list

nathansandi:
## Description

Implement the Service and Gateway Rest for Get Task by Key

GET:
`/v2/user-task/{userTaskKey}`

## Related issues

closes #https://github.com/camunda/camunda/issues/22481


ralfpuchert (approved):
LGTM :)

christinaausley (approved):

tmetzke:
Are we adding the endpoint to the OpenAPI spec as well? I cannot see that it was added there. Am I missing something?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
```diff
@@ -60,4 +68,29 @@ private ResponseEntity<Object> search(final UserTaskQuery query) {
...
+      return RestErrorMapper.mapProblemToResponse(problemDetail);
+    } catch (final Exception e) {
+      REST_LOGGER.warn("An exception occurred in Get User Task by key.", e);
+      final var problemDetail =
+          RestErrorMapper.createProblemDetail(
+              HttpStatus.INTERNAL_SERVER_ERROR,
+              e.getMessage(),
+              "Failed to execute Get User Task by key.");
+      return RestErrorMapper.mapProblemToResponse(problemDetail);
+    }
```
tmetzke, line 94: 🔧 My recommendation here would be to use `RestErrorMapper::mapErrorToProblem` for any exception that occurs when calling the `UserTaskServices::getByKey`. We just need to add the `NotFoundException` to the switch statement in the `RestErrorMapper`. The generic exception case will change a little bit regarding messaging when using the mapper instead (it runs into the default case there then). We can either live with that or adjust the `RestErrorMapper::mapErrorToProblem` by adding a String parameter that allows defining the message in the generic error case.

nathansandi: Great catch 💯  @tmetzke  :) 
I just updated the Controller with your suggested change

I am going to apply the same update on Form ASAP as I merge this PR.

nathansandi:
FWY: There is a problem on the CI in FE part, I am waiting to be handled by FE team, to merge the main and update this PR

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
nathansandi: @tmetzke  I have applied the follow suggestions. Please check if the `NotFoundException` is according to the guidelines

Following the implementation we will have the follow error for Not found

```
                {
                  "type": "about:blank",
                  "status": 404,
                  "title": "NOT_FOUND",
                  "detail": "User Task with key 999 not found"
                }
```

tmetzke (commented):
👍 Looks good already.
🔧 Some minor things need adjustment or should be looked at.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
tmetzke: 👍 Very nice

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
```diff
@@ -60,4 +63,21 @@ private ResponseEntity<Object> search(final UserTaskQuery query) {
...
   }
+
+  @GetMapping(
+      path = "/{userTaskKey}",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE})
+  public ResponseEntity<?> getByKey(@PathVariable("userTaskKey") final Long userTaskKey) {
+    try {
+      // Success case: Return the left side with the UserTaskItem wrapped in ResponseEntity
+      return ResponseEntity.ok()
+          .contentType(new MediaType(MediaType.APPLICATION_JSON, StandardCharsets.UTF_8))
```
tmetzke, line 74: ❓ Just saw this now, sorry. As far as I can see, the other endpoints don't define the content type explicitly. Do we need to do this? Just wondering.

nathansandi: We dont need to define this explicity
I accidently commited this with my copilot 
Will remove it

Thank you for checking this

tmetzke (approved):
Looks good to me 👍

Issue #22204, opened 2024-09-11
CI: bot command to retrieve details about failed PR builds


cmur2:
## Description

When engineers encounter slow, aborted or failing CI checks on their PRs they need to perform some manual tasks (or get help to do so) to figure out why those checks are problematic, especially if those CI checks are required to merge:

* for failing CI checks open the logs + parse them to find a relevant error message + understand the error message to categorize if it is related to changes on the PR or an external problem (network error, authentication problem etc.)
    * retry in case of external problems
* for aborted CI checks find the reason + understand the reason to categorize if it is related to changes on the PR or an external problem (runner problem)
    * retry in case of external problems
* for slow builds on self-hosted runners manually navigate to [the dashboard to check resource usage](https://dashboard.int.camunda.com/d/000000019/pods)

The goal of this ticket to improve this situation. As a first step we should establish a bot command that automatically collects relevant info like in https://github.com/camunda/team-infrastructure/issues/459 for the above cases (error messages, K8s logs for self-hosted runners, resource usage dashboard for self-hosted runners) to eliminate manual labor and uses a database of well-known errors to make classifying error messages easier.

Steps:

* create new (or reuse existing) command parse+dispatch workflow like https://github.com/camunda/camunda/blob/main/.github/workflows/optimize-commands-dispatch.yml on PR comments, based on https://github.com/peter-evans/slash-command-dispatch/
    * needs a GitHub app token
    * prefer reuse if possible to avoid many parallel implementations of the same thing
* listen to command `/ci-problems`
    * future commands can be `/ci-*` like `/ci-retry` etc
* identify PR and execute custom script that collects relevant info (extensible) and generates a result
* post result via https://github.com/peter-evans/create-or-update-comment

cmur2:
POC in https://github.com/cmur2/poc-debug-cmd/pull/1

cmur2:
PR in https://github.com/camunda/camunda/pull/22483

cmur2:
Docs at https://github.com/camunda/camunda/wiki/CI-&-Automation#chatops and announcement in https://camunda.slack.com/archives/C071KP5BTHB/p1727259126657949

cmur2:
Will keep this ticket open until tomorrow for any immediate feedback.

PR #22483, opened 2024-09-19
ci: add generalized chatops command for GHA CI problem analysis

cmur2:
## Description

This PR has **three aspects** that are connected:

1. **The analysis script:** The goal of this PR is to implement the first iteration of https://github.com/camunda/camunda/issues/22204 providing a semi-automated way of analyzing (failed) GHA runs in this repository and providing actionable advice and links in case of detected problems.
This is implemented with a Bash script which I considered fine for the first iteration but might have to be rewritten (+ unit tests) if it grows much more. It might also become useful outside the monorepo.
Right now the script detects self-hosted runner problems, GHA timeouts, DockerHub connectivity problems and general "Process exited with error code XY" messages. This can and should be extended in the future to cover even more cases.

2. **PR comments trigger:** The script is invoked when someone comments with `/ci-problems` under a PR in this monorepo which uses https://github.com/peter-evans/slash-command-dispatch to trigger the script and post a new comment on that PR with Markdown output. Example can be seen in this POC: https://github.com/cmur2/poc-debug-cmd/pull/1

3. **Refactoring Optimize ChatOps:** We already had an existing ChatOps mechanism in the monorepo since the Optimize source code merge in https://github.com/camunda/camunda/blob/main/.github/workflows/optimize-commands-dispatch.yml - I chose to rename and reuse this workflow instead of creating a duplicate to handle both cases:
    * Optimize commands on issues - that might honestly be useful also for other Engineering teams in the monorepo, have no command prefix
    * CI commands on PRs - the first one is `/ci-problems` that runs the analysis script

image

This PR can only be tested in the monorepo after merge to `main` due to the command dispatch restrictions using the default branch.

_So far I didn't create a dedicated GitHub App for the CI commands and rather reused the Optimize GH app which is not perfectly clean but good enough for a first iteration._

## Checklist

- [x] for CI changes:

## Related issues

Related to #22204


cmur2:
Pinging @misiekhardcore @RomanJRW as a FYI that I touched the Optimize ChatOps mechanism and plan to generalize it to avoid duplication across several teams/use cases. Please chime in if you have concerns or other feedback.

.ci/chatops-commands/ci-problems-analyze.sh
```diff
@@ -0,0 +1,86 @@
...
+    >&2 echo "Usage: $0 PR_NUMBER"
+    exit 1
+fi
+
+set -e
+
+OWNER_NAME="camunda"
+REPO_NAME="camunda"
+PR_NUMBER="$1"
+PR_HEAD_SHA=$(gh api -X GET "repos/$OWNER_NAME/$REPO_NAME/pulls/$PR_NUMBER" --jq '.head.sha')
```
andromaqui, line 13: NABD: Feel free to ignore this comment, but In this repository, sometimes the bash variables in scripts are enclosed in {} and other not. I know both ways are correct but I just wanted to point out that in infra-core most of the times we enclose these variables in {} and I personally find it cleaner.

.ci/chatops-commands/ci-problems-analyze.sh
```diff
@@ -0,0 +1,86 @@
...
+      echo "  * GitHub prematurely lost connection to the runner which can happen due to high job CPU load (try reducing load), network issues or hardware failures (try rerunning)."
+      echo "  * Check [resource usage for runner \`$runner_name\`]($pods_dashboard_url) and [Kubernetes logs for runner \`$runner_name\`]($gke_logs_url)."
+    fi
+
+    ############################################################################
+    timeout_annotations=$(echo "$job_annotations" | jq '.[] | select(.message | contains("has exceeded the maximum execution time"))')
+    if echo "$timeout_annotations" | grep -q message; then
+      runner_name=$(gh api "repos/$OWNER_NAME/$REPO_NAME/actions/jobs/$job_id" --jq '.runner_name' | tr '[:upper:]' '[:lower:]')
+
+      >&2 echo "Job $job_id got cancelled due to timeout."
```
andromaqui, line 56: Do you think it is a good idea to print also the workflow name for better UX?

andromaqui (commented):
I just left 3 remarks but I am mainly not sure if the jq requests are being handled properly.

.ci/chatops-commands/ci-problems-analyze.sh
```diff
@@ -0,0 +1,86 @@
...
+PR_NUMBER="$1"
+PR_HEAD_SHA=$(gh api -X GET "repos/$OWNER_NAME/$REPO_NAME/pulls/$PR_NUMBER" --jq '.head.sha')
+
+echo "## 🔎 GHA problems for PR #$PR_NUMBER"
+echo ""
+echo "For most recent commit $PR_HEAD_SHA:"
+echo ""
+
+gh api -X GET "repos/$OWNER_NAME/$REPO_NAME/actions/runs?sha=$PR_HEAD_SHA" --jq '.workflow_runs[]' | while read -r workflow_run; do
+  workflow_run_id=$(echo "$workflow_run" | jq -r '.id')
```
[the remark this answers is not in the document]
cmur2: My plan to catch errors is by exiting the Bash script when any command returns a non-zero exit code. This is achieved by the `set -e` at the beginning. This protects against failing GitHub API requests since the `gh` tool then exists non-zero.

For `jq` this also happens correctly as I tested locally (exit code 4):

```
$ echo "some error message from previous command" | jq -r '.id'
parse error: Invalid numeric literal at line 1, column 5
```

.ci/chatops-commands/ci-problems-analyze.sh
cmur2: This line specifically prints to stderr and is only for debugging where the ID is needed for unique identification (copy&paste) but your remark in general is right! I adjusted the end-user-facing parts of the script to output `<workflow_name> / <job_name>` for better readability.

.ci/chatops-commands/ci-problems-analyze.sh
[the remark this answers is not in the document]
andromaqui: Ah you are right! I hadnt enabled `set -e` when testing

andromaqui (approved):
LGTM!

cmur2:
I'll merge this PR when I'm back on 2024-09-24 since it also needs testing after merge + if successful an announcement to be useful to other engineers.

PR #22485, opened 2024-09-19
feat: add support for oidc authentication

### LINKED ISSUE #19741: Camunda can be configured to use an OIDC Identity Provider
### Description

Camunda supports OIDC authentication via the profile `auth-oidc`.

Where possible we should support existing configuration options documented in https://docs.camunda.io/docs/next/self-managed/setup/guides/connect-to-an-oidc-provider/?optionsType=env.

```[tasklist]
### Breakdown
- [ ] Authentication with an OIDC provider can be configured via application properties and env variables
```

maryarm:
## Description

Camunda supports OIDC authentication via the profile auth-oidc.

Where possible we should support existing configuration options documented in https://docs.camunda.io/docs/next/self-managed/setup/guides/connect-to-an-oidc-provider/?optionsType=env.

## Related issues

closes #19741

### DIFF
--- a/dist/src/main/resources/application-auth-oidc.yaml
+++ b/dist/src/main/resources/application-auth-oidc.yaml
@@ -1,0 +1,17 @@
+spring:
+  security:
+    oauth2:
+      client:
+        registration:
+          oidcclient:
+            client-id:
+            client-secret:
+            authorization-grant-type: authorization_code
+            redirect-uri:
+            provider: oidcclient
+            scope:
+              - "openid"
+              - "profile"
+        provider:
+          oidcclient:
+            issuer-uri:
--- a/authentication/pom.xml
+++ b/authentication/pom.xml
@@ -81,6 +81,16 @@
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-oauth2-client</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.nimbusds</groupId>
+          <artifactId>oauth2-oidc-sdk</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-oauth2-core</artifactId>
     </dependency>
   </dependencies>
 </project>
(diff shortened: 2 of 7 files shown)


dist/src/main/resources/application-auth-oidc.yaml
```diff
@@ -0,0 +1,17 @@
+spring:
```
Ben-Sheppard, line 1: 🔧 I think that the typical place for these values is in `application-<profile>.properties` - I don't think there is anything functionally different between the two but I think its good to use a single approach here if possible.

dist/src/main/resources/application-auth-oidc.yaml
```diff
@@ -0,0 +1,17 @@
...
+    oauth2:
+      client:
+        registration:
+          oidcclient:
+            client-id:
+            client-secret:
+            authorization-grant-type: authorization_code
+            redirect-uri:
+            provider: oidcclient
+            scope:
```
Ben-Sheppard, line 12: 💭 Something for the future but we've had instance in the past where customers have wanted configurable scopes, I guess we can add this in later though if necessary

Ben-Sheppard (approved):
LGTM - I can't really see any tests that we can add here that bring value or that aren't just testing the sub-systems

dist/src/main/resources/application-auth-oidc.yaml
maryarm: Do you mean to switch from yaml to properties? I find it more readable in cases like this that there is a multi level hierarchy.

dist/src/main/resources/application-auth-oidc.yaml
maryarm: I see. Doesn't the default way of property overriding for this case help? for example customer can override the scope property here with their new list of profile.

Ben-Sheppard: Good point, it does, but i'm not sure if we want to make it a nicer experience down the line i.e. an env variable. Not critical for now at all but wanted to mention it

dist/src/main/resources/application-auth-oidc.yaml
Ben-Sheppard: Yea so elsewhere in the code we have things like `application-identity.properties` for all the identity related things. NABD but I wanted to raise that it goes against the current approach

maryarm: I'll apply your comment in my next PR (loading claims in authentication object when login with oidc)

PR #22576, opened 2024-09-23
feat: Introduce support for Task Listeners on UserTask `complete` operation

### LINKED ISSUE #21977: Introduce Task Listener support for `complete` operation
## Description:
Introduce task listener support for `complete` operation as proposed in [PoC](https://github.com/camunda/camunda/pull/21467).

### Scope:
- Extend `UserTaskIntent` with a new `COMPLETE_TASK_LISTENER` command.
- Update `JobCompleteProcessor` to write `COMPLETE_TASK_LISTENER` on successful Task Listener job completion.
- Create `UserTaskListenerCompleteProcessor` to process the `COMPLETE_TASK_LISTENER` command. 
- Register the new command and processor in the `UserTaskEventProcessors` class.
- Add `taskListeners` property for the `ExecutableUserTask` element.
- Update `BpmnJobBehavior` to create a Task Listener job for defined listeners.
- Introduce `TaskListenerIndicesRecord` to track the index of the Task Listener under execution.

ce-dmelnych:
## Description
This PR introduces support for task listeners on the UserTask `complete` operation. It builds on top of the previous [refactoring](https://github.com/camunda/camunda/pull/22484) by adding the ability to create and handle task listener jobs defined for UserTask element with `complete` event type.

Key changes include:
- Implementation
  - Refactored `BpmnJobBehavior` to handle the creation of task listener jobs.
  - Introduced `COMPLETE_TASK_LISTENER` intent in `UserTaskIntent` to signal about the completion of task listener job.
  - Extended logic in `UserTaskProcessor` to create listener job and react on `COMPLETE_TASK_LISTENER` either by creation of the next task listener or the finalization of the original user task command.
  - Introduced a new error type `TASK_LISTENER_NO_RETRIES` that will be used as a type for incidents when task listener job have no retries, this work will be done in the scope of #23492 task.
  - Introduced `UserTaskIntermediateStateValue` to persist intermediate user task data after command invocation, until all task listeners are executed and then use it to finalize the original command with unmodified data as originally intended by the user.
    - Updated `DbUserTaskState` to manage intermediate state storage, retrieval and deletion.
  - Added support for merging task listener variables into the local user task scope, ensuring that subsequent listeners have access to variables produced by previous listeners.
  - Populated task listener job headers with relevant user task properties (e.g., `assignee`, `dueDate`, etc.).
  - Added `TaskListenerIndicesRecord` to `ElementInstance` state to track the task listener index that was executed for `complete` UserTask operation.

- Tests and documentation:
   - Added tests to ensure correct behavior for task listeners during the `complete` operation.
   - Added JavaDocs for `COMPLETE_TASK_LISTENER` and `COMPLETE_EXECUTION_LISTENER` intents to explain their roles.
   - Improved naming consistency and added clarification comments throughout the code.

## Checklist

- [ ] for CI changes:

## Related issues

closes #21977


megglos:
@ce-dmelnych what's due to get this out of the draft status?

ce-dmelnych:
> @ce-dmelnych what's due to get this out of the draft status?

Hi @megglos,

To move this PR out of draft status, I need a few more days to:
- Find a way to persist data that can be set while completing the task (like `variables` and `action`) when complete task listeners are defined. This involves storing this data until all task listeners have executed so it can be properly persisted after on the `UserTask.COMPLETED` event.

- Evaluate options how to handle recovery after an incident caused by TL property expression evaluation issues (unless we decide to extract it to a separate task).

- Enhance the tests to ensure comprehensive coverage, as this is the task with initial TL support, unexpected issues may arise.

I'm syncing with Nico and Ana today where we can discuss some of these topics.
I'll update the PR accordingly after our meeting.

Thanks for your patience!

ce-dmelnych:
@megglos,

Since my last comment, I had a sync with Nico and Ana where we evaluated the options for handling incident recovery related to task listener property expression evaluation failures. We found that both options (re-invoking the failed user task command or re-executing only the failed task listener) have similar technical complexity.
As a result, we've decided to separate this issue into a dedicated [subtask](https://github.com/camunda/camunda/issues/22717) to allow the business to decide on the preferred approach.

For this PR, what's left is to implement the logic to persist the user task state before invoking the task listeners. This involves storing any data (like `variables` and `action`) that can be set while completing the task, so it can be properly persisted after all task listeners have executed and the `UserTask.COMPLETED` event is triggered.
I'm currently working on this.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobCompleteProcessor.java
```diff
@@ -60,29 +63,48 @@ public void afterAccept(
...
+            return;
+          }
+        case TASK_LISTENER:
+          {
+            // to store the variable for merge, to handle concurrent commands
+            eventHandle.triggeringProcessEvent(value);
 
-      final long scopeKey = serviceTask.getValue().getFlowScopeKey();
-      final ElementInstance scopeInstance = elementInstanceState.getInstance(scopeKey);
+            // TODO explain here why we retrieve intermediate state and not the `regular` user task
```
ana-vinogradova-camunda, line 88: ❓ Probably we should resolve this TODO before the PR merge.

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
```diff
@@ -33,5 +33,39 @@ public enum JobListenerEventType {
...
+   * the listener should be triggered when a user task is completed.
+   */
+  COMPLETE,
+
+  /**
+   * Represents the `cancel` event for a task listener. This event type is used to indicate that the
+   * listener should be triggered when a user task is canceled or terminated before completion. This
+   * can occur due to workflow interruptions or exceptions.
+   */
+  CANCEL
```
korthout, line 70: 🔧 If you want to reduce the size of the PR, you could extract these and add them later again

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
```diff
@@ -33,5 +33,39 @@ public enum JobListenerEventType {
...
    */
-  END
+  END,
+
+  /**
+   * Represents the `create` event for a task listener. This event type is used to indicate that the
+   * listener should be triggered when a user task is created. This occurs when the task becomes
+   * available in the workflow for the first time.
+   */
+  CREATE,
```
korthout, line 43: ❓ Why did we choose `CREATE` over `CREATING`, `CREATED`, or `CREATION`? 

To me, `create` is a command or a request. It does not describe an event. Events are things that have happened. They are typically written in the past-tense. As this feature allows blocking the operation, you could argue that the event that has happened is `CREATING` as the listener still has the chance to block the task from actually being `CREATED`. 

An alternative naming could be to use the concept that it refers to: `CREATION`. This aligns with `START` and `END` event types.

I want us to carefully consider the naming here, as this is user facing.

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
```diff
@@ -33,5 +33,39 @@ public enum JobListenerEventType {
    * the listener should be triggered at the end of an execution, such as the completion of a
    * process instance, sub-process or element (task, event, gateway).
    */
-  END
+  END,
+
+  /**
+   * Represents the `create` event for a task listener. This event type is used to indicate that the
+   * listener should be triggered when a user task is created. This occurs when the task becomes
+   * available in the workflow for the first time.
```
korthout, line 41: 🔧 Let's refer to process > workflow

```suggestion
   * available in the process for the first time.
```

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/intent/UserTaskIntent.java
```diff
@@ -40,7 +40,9 @@ public enum UserTaskIntent implements ProcessInstanceRelatedIntent {
   UPDATING(12),
   UPDATED(13),
 
-  MIGRATED(14);
+  MIGRATED(14),
+
+  COMPLETE_TASK_LISTENER(15);
```
korthout, line 45: ❓ Did we consider alternative names? For example, `CONTINUE_AFTER_TASK_LISTENER_COMPLETED`

korthout (commented):
Great work @ce-dmelnych 👏 

Here's my first findings. I sadly haven't had a chance to review everything yet. I'll continue reviewing, but wanted to share these thoughts already

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/ProcessElementProperties.java
```diff
@@ -0,0 +1,90 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.bpmn;
+
+import static io.camunda.zeebe.util.buffer.BufferUtil.bufferAsString;
+
+import io.camunda.zeebe.protocol.impl.record.value.usertask.UserTaskRecord;
+
+public final class ProcessElementProperties {
```
korthout, line 14: ❓ What's the reason for introducing this new properties class? It seems to cover the same data as the `BpmnElementContext` and has a similar name.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/UserTaskIntermediateState.java
```diff
@@ -0,0 +1,47 @@
...
+package io.camunda.zeebe.engine.state.instance;
+
+import io.camunda.zeebe.db.DbValue;
+import io.camunda.zeebe.msgpack.UnpackedObject;
+import io.camunda.zeebe.msgpack.property.ObjectProperty;
+import io.camunda.zeebe.protocol.impl.record.value.usertask.UserTaskRecord;
+
+/**
+ * This class holds the necessary user task data (including variables) to finalize commands such as
+ * `COMPLETE`, `UPDATE`, `ASSIGN` etc. mainly when task listeners are defined for these operations.
```
korthout, line 17: 🔧 This introduction could use an example for clarification. Something like:

> For example, when the user task is being assigned for the first time, the user task must remain unassigned while the task listener blocks the assignment. In the meantime, this tracks the new `assignee` which can be used when the task listener completes.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserTaskCompletingV2Applier.java
```diff
@@ -0,0 +1,31 @@
...
+  private final MutableUserTaskState userTaskState;
+
+  public UserTaskCompletingV2Applier(final MutableProcessingState processingState) {
+    userTaskState = processingState.getUserTaskState();
+  }
+
+  @Override
+  public void applyState(final long key, final UserTaskRecord value) {
+    userTaskState.updateUserTaskLifecycleState(key, LifecycleState.COMPLETING);
+    userTaskState.createUserTaskIntermediateState(value);
```
korthout, line 29: 🔧 I wonder whether we can improve the understandability of this User Task Intermediate State somewhat by changing this method name. For example, `storeIntermediateState(value, LifecycleState.COMPLETING)` might help clarify that this is storing intermediate data for the completing state.

korthout (commented):
Some more questions/suggestions

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/ProcessElementProperties.java
ce-dmelnych: Thank you, @korthout, for pointing this out! 🙏
I wasn't entirely happy with this solution either, and your comment prompted me to rethink it.
Initially, I introduced `ProcessElementProperties` because it wasn't obvious that I could simply initialize `BpmnElementContext` inside `UserTaskProcessor` at the places where I needed it :sweat_smile:

After your feedback, I took a closer look and found a better approach:
now I'm building the `BpmnElementContext` directly inside the `UserTaskProcessor` from user task `ElementInstance`, which made the `ProcessElementProperties` class redundant.
So, removed it as part of this [commit](https://github.com/camunda/camunda/pull/22576/commits/b0fc3e250481ccba37e8467db63a063958e186ab).

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
ce-dmelnych: Thanks! I've removed the other job listener event types except for `COMPLETE`, as they aren't required at this stage. We'll reintroduce them in dedicated PRs when we'll work on support for `update/assign/etc.` task listeners.

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
ce-dmelnych: Thanks for raising this point, Nico!

When I initially chose the names for `JobListenerEventType` values and [ZeebeTaskListenerEventType](https://github.com/camunda/camunda/blob/main/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeTaskListenerEventType.java), I followed the same naming conventions for task listener events that we have in Camunda 7 (as seen [here](https://docs.camunda.org/manual/7.21/user-guide/process-engine/delegation-code/#task-listener-event-lifecycle)).
In Camunda 7, we use terms like `create`, `update`, `assignment`, `complete`, etc., which are already familiar to users.

However, since we're building Camunda 8, we have the opportunity to establish more consistent and clearer naming for these event types. I agree with your suggestion that naming them in a way that reflects the event's lifecycle (e.g., `CREATION ` or `COMPLETION`) could improve clarity.

We should also ensure that the naming aligns with the values in [ZeebeTaskListenerEventType](https://github.com/camunda/camunda/blob/main/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/instance/zeebe/ZeebeTaskListenerEventType.java), so that it consistent across the modules.

Happy to discuss this further and refine the naming to make it more intuitive for users!

I’ll raise this topic for discussion in the **#usertask-listeners** channel ✍️.

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
ce-dmelnych: Got it :write

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/intent/UserTaskIntent.java
ce-dmelnych: I followed the same naming convention that we use for execution listeners,
specifically [`ProcessInstanceIntent.COMPLETE_EXECUTION_LISTENER`](https://github.com/camunda/camunda/blob/7add7a0dc05d08d11b07f8b6b1764501b9a9c169/zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/intent/ProcessInstanceIntent.java#L39), which was introduced to serve the same purpose —> freezing the process execution until this command is written.

That’s why I opted for `COMPLETE_TASK_LISTENER` to maintain consistency.

However, I agree that your option, `CONTINUE_AFTER_TASK_LISTENER_COMPLETED`, sounds really self-explanatory and may actually suit this case better! 👍

I’m open to discussing if we decide to go with a more descriptive naming!

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
korthout: I've started a discussion [here](https://camunda.slack.com/archives/C0622NWC9LK/p1728996247956779). Let's take the decision along here

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/intent/UserTaskIntent.java
korthout: 💭 Although possible, it's a bit hard to change this. You'd need to add the new intent name, mark the old as deprecated, and rename the processor and register it for both the deprecated and the new intent (for backward compatibility). I'm not sure whether it's worth it. I'll leave that up to you. As an alternative, we could consider documenting the meaning in JavaDoc on the intent.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobCompleteProcessor.java
ce-dmelnych: Thanks Ana for highlighting this!
I've added the explanation comment instead of // TODO in the scope of this [commit](https://github.com/camunda/camunda/pull/22576/commits/65e06b31728360a30fcf5bc68a10bac29483f3fb).

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/UserTaskIntermediateState.java
ce-dmelnych: Thanks for the suggestion, Nico!
I've added the example as per your feedback in this [commit](https://github.com/camunda/camunda/pull/22576/commits/65e06b31728360a30fcf5bc68a10bac29483f3fb#diff-495cacb2dfd86d4f86cddc70e7ce10b65ef46c2716d87dbbcc8e288f495a02d1R34-R37).

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserTaskCompletingV2Applier.java
ce-dmelnych: Thanks for the suggestion, Nico!
I had a similar thought about linking the `user task record` with the `user task lifecycle`, which could also help with debugging. I've applied your suggestion, including the naming update, in this [commit](https://github.com/camunda/camunda/pull/22576/commits/65e06b31728360a30fcf5bc68a10bac29483f3fb#diff-495cacb2dfd86d4f86cddc70e7ce10b65ef46c2716d87dbbcc8e288f495a02d1).

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
```diff
@@ -0,0 +1,496 @@
...
+        .fail();
+
+    // then: incident created
+    final Record<IncidentRecordValue> incident =
+        RecordingExporter.incidentRecords(IncidentIntent.CREATED)
+            .withProcessInstanceKey(processInstanceKey)
+            .getFirst();
+    Assertions.assertThat(incident.getValue())
+        .hasProcessInstanceKey(processInstanceKey)
+        .hasErrorType(ErrorType.JOB_NO_RETRIES)
```
korthout, line 187: ~~❌ I believe we decided to use `TASK_LISTENER_NO_RETRIES`~~

EDIT: It seems we'll tackle this elsewhere: 
- https://github.com/camunda/camunda/issues/23492

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
```diff
@@ -0,0 +1,496 @@
...
+    final long processInstanceKey =
+        createProcessInstance(
+            createProcessWithZeebeUserTask(
+                t ->
+                    t.zeebeTaskListener(
+                        l ->
+                            l.complete()
+                                .typeExpression("\"listener_1_\"+my_var")
+                                .retriesExpression("5+3"))),
+            Map.of("my_var", "abc"));
```
korthout, line 239: 🔧 It may be slightly clearer that the `Map.of(...)` are variables if the method name is changed to `createProcessInstanceWithVariables`

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
```diff
@@ -0,0 +1,496 @@
...
+        .complete();
+
+    // then: assert the variable 'userTaskOutput' is accessible in the subsequent element
+    final var subsequentServiceTaskJob = activateJob(processInstanceKey, "subsequent_service_task");
+    assertThat(subsequentServiceTaskJob.getVariables()).containsEntry("userTaskOutput", "bar_abc");
+    completeJobs(processInstanceKey, "subsequent_service_task");
+  }
+
+  @Test
+  public void shouldIncludeConfiguredUserTaskDataInCompleteTaskListenerJobHeaders() {
```
korthout, line 351: 💭 This is fine for now. It's possible that we remove this later again if we support explicit user data on the job record. While that's still up to debate, I think it's great that we already support some form of providing this data to the listeners.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
```diff
@@ -0,0 +1,496 @@
...
+    assertThat(activatedListenerJob.getCustomHeaders())
+        .containsOnly(
+            entry(
+                Protocol.USER_TASK_CANDIDATE_GROUPS_HEADER_NAME,
+                "[\"group_A\",\"group_C\",\"group_F\"]"),
+            entry(Protocol.USER_TASK_CANDIDATE_USERS_HEADER_NAME, "[\"user_A\",\"user_B\"]"),
+            entry(Protocol.USER_TASK_ASSIGNEE_HEADER_NAME, "admin"),
+            entry(Protocol.USER_TASK_DUE_DATE_HEADER_NAME, "2095-09-18T10:31:10+02:00"),
+            entry(Protocol.USER_TASK_FORM_KEY_HEADER_NAME, Objects.toString(form.getFormKey())),
+            entry(USER_TASK_KEY_HEADER_NAME, String.valueOf(userTaskRecordValue.getKey())));
```
korthout, line 379: ❓ Why is the `USER_TASK_KEY_HEADER_NAME` not added to `Protocol`?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
```diff
@@ -7,26 +7,32 @@
  */
 package io.camunda.zeebe.engine.processing.bpmn.behavior;
 
+import static io.camunda.zeebe.util.Either.right;
+import static java.util.Optional.of;
+import static java.util.Optional.ofNullable;
```
korthout, line 12: 🔧 Static members should be accessed via their class according to our code style

https://google.github.io/styleguide/javaguide.html#s6.3-static-members

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserTaskCompletingV2Applier.java
```diff
@@ -0,0 +1,31 @@
...
+  private final MutableUserTaskState userTaskState;
+
+  public UserTaskCompletingV2Applier(final MutableProcessingState processingState) {
+    userTaskState = processingState.getUserTaskState();
+  }
+
+  @Override
+  public void applyState(final long key, final UserTaskRecord value) {
+    userTaskState.updateUserTaskLifecycleState(key, LifecycleState.COMPLETING);
+    userTaskState.storeIntermediateState(value, LifecycleState.COMPLETING);
```
korthout, line 29: 💭 I was thinking about this naming and what we really mean by intermediate state. IMO, we could also name it lifecycle transition. But I'm not sure whether that actually clarifies anything :)

korthout (approved):
Thanks @ce-dmelnych 🚀 

👍 I really like tests and documentation you added

💭 I noticed that this does not include the ability for listeners to prevent the completion. I think that's good and we should work on that separately. The same goes for the ability to change the user task data in the listener.

💭 Regarding the potential name change for event types, I don't see it as a reason to block this PR. If we have to, we can adjust the code later.

👍 LGTM

ana-vinogradova-camunda (approved):
LGTM 🚀 
I like the code clarity and test coverage 👍 

💭 The only thing - it might be a good idea to keep PRs to a bit smaller size in the future

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
ce-dmelnych: 👍 Updated the code to refer to static methods using their class name.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
ce-dmelnych: I'm not sure that we have a strong decision yet on whether `USER_TASK_KEY_HEADER_NAME` should be part of the custom headers. Or it could be included as a separate property in the `JobRecord`.
Once that's decided, we can make it part of the public `Protocol` API.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
ce-dmelnych: 👍 Thanks for suggestion, applied ✅

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserTaskCompletingV2Applier.java
ce-dmelnych: Yeah, it's often tricky to nail down the perfect name right away... 🤔
Let's see how we end up using this "intermediate" state in practice, and we can adjust the naming if needed for better clarity.

PR #22609, opened 2024-09-23
ci: use infra github-managed runners for `ci.yml/java-unit-tests` workflow

clementnero:
## Description

Related to https://github.com/camunda/team-infrastructure/issues/671.

This PR moves Java unit tests requiring Strace, and the kernel parameter `kernel.yama.ptrace_scope=0`, to a dedicated job in `zeebe-ci.yml` using a GitHub-managed runner.

The security risk is thus outsourced, and the `ci.yml/java-unit-tests` job can be migrated to the Infra Self-Hosted runners. safely.

A new `strace` test group has been created for strace-related tests. It is excluded by default, and can be included using the `include-strace-tests` profile.

## Checklist

- [x] for CI changes:

## Related issues

closes #

### DIFF
--- a/.github/workflows/zeebe-ci.yml
+++ b/.github/workflows/zeebe-ci.yml
@@ -186,6 +186,57 @@ jobs:
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+  strace-tests:
+    # Zeebe tests requiring strace
+    name: Strace Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install and allow strace tests
+        run: |
+          sudo apt-get -qq update && sudo apt-get install -y strace
+          sudo sysctl -w kernel.yama.ptrace_scope=0
+      - uses: ./.github/actions/setup-zeebe
+        with:
+          secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+      - uses: ./.github/actions/build-zeebe
+        with:
+          maven-extra-args: -T1C -PskipFrontendBuild
+      - name: Create build output log file
+        run: echo "BUILD_OUTPUT_FILE_PATH=$(mktemp)" >> $GITHUB_ENV
+      - name: Maven Test Build
+        run: >
+          ./mvnw -T1C -B --no-snapshot-updates
+          -P include-strace-tests
+          -D junitThreadCount=2
+          -D skipChecks
+          -f zeebe
+          test
+          | tee "${BUILD_OUTPUT_FILE_PATH}"
+      - name: Analyze Test Runs
+        if: always()
+        uses: ./.github/actions/analyze-test-runs
+        with:
+          buildOutputFilePath: ${{ env.BUILD_OUTPUT_FILE_PATH }}
+          # workaround to avoid https://github.com/camunda/camunda/issues/16604
+          skipSummary: true
+      - name: Upload test artifacts
+        uses: ./.github/actions/collect-test-artifacts
+        if: ${{ failure() || cancelled() }}
+        with:
+          name: Strace Tests
+      - name: Observe build status
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          build_status: ${{ job.status }}
+          secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
   docker-checks:
     name: Docker checks
     runs-on: ubuntu-latest
@@ -260,6 +311,7 @@ jobs:
       - property-tests
       - performance-tests
       - docker-checks
+      - strace-tests
     steps:
       - uses: actions/checkout@v4
       - run: exit ${{ ((contains(needs.*.result, 'cancelled') || contains(needs.*.result, 'failure')) && 1) || 0 }}
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -2342,8 +2342,8 @@
           <artifactId>maven-surefire-plugin</artifactId>
           <version>${plugin.version.surefire}</version>
           <configuration>
-            <!-- by default, exclude performance tests -->
-            <groups>!performance</groups>
+            <!-- by default, exclude performance and strace tests -->
+            <excludedGroups>performance, strace</excludedGroups>
             <failIfNoTests>false</failIfNoTests>
             <trimStackTrace>false</trimStackTrace>
             <redirectTestOutputToFile>true</redirectTestOutputToFile>
@@ -2876,6 +2876,24 @@
     </profile>
 
     <profile>
+      <!-- Dedicated profile to run ONLY tests requiring strace. -->
+      <id>include-strace-tests</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <groups>strace</groups>
+              <!-- Reset <excludedGroups> as it take precedence over <groups> -->
+              <excludedGroups combine.self="override"></excludedGroups>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
+    <profile>
       <!-- Dedicated profile to run ONLY performance tests. The include config overwrites the default pattern-->
       <id>include-performance-tests</id>
       <build>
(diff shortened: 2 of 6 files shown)


.github/workflows/zeebe-ci.yml
```diff
@@ -186,6 +186,57 @@ jobs:
           secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+  strace-tests:
```
cmur2, line 189: Job name needs to be added to the `needs` list of the `test-summary` to make it wait for this new job and fail the build if the new job fails.

.github/workflows/zeebe-ci.yml
```diff
@@ -186,6 +186,57 @@ jobs:
...
+    # Zeebe tests requiring strace
+    name: Strace Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
cmur2, line 193: Probably from copy&paste, let's use a timeout closer to the actual runtime in the range of 5-10 minutes.

cmur2: Ok! We can also settle on 15min orso but 30min seems too much..

clementnero: It [took](https://github.com/camunda/camunda/actions/runs/11050067884/job/30697060018) ~14min, I prefer to set timeout to 20min to have some margin.

cmur2 (approved):

cmur2 (approved):

npepinpe (approved):
The usual way of excluding tests with junit 5 is instead through tags. See the `include-performance-tests` profile, for example. The randomized tests are a bit of an exception because they're still on junit 4 (for some of them) =/

I'm not putting this as a blocker - I can do it in a follow up, or if you're keen to learn, we can pair it. But generally using tags/groups is better as it means we can add tests in the future to the group without having to change the build files.

npepinpe:
One quick question, are we backporting the migration to CI runners to older stable branches? It would help us by allowing us to remove our infrastructure entirely. At the same time, considering stable branches are relatively, well, stable, I understand the trade off is not that great for you (other than my eternal gratitude ;)).

clementnero:
> The usual way of excluding tests with junit 5 is instead through tags. See the `include-performance-tests` profile, for example. The randomized tests are a bit of an exception because they're still on junit 4 (for some of them) =/
> 
> I'm not putting this as a blocker - I can do it in a follow up, or if you're keen to learn, we can pair it. But generally using tags/groups is better as it means we can add tests in the future to the group without having to change the build files.

Thank you for clarifying this point, it is indeed more convenient. I didn't know this was a new Junit5 feature. I had looked at the `include-performance-tests` profile and `@Tag(“performance”), I will try to use tags the same way. I will let you know :)

parent/pom.xml
```diff
@@ -2883,11 +2883,7 @@
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
-              <excludes combine.children="append">
-                <exclude>**/SnapshotChecksumTest.java</exclude>
-                <!-- Re-add the default exclude for inner classes, see https://github.com/camunda/camunda/issues/8637 -->
-                <exclude>**/*$*.java</exclude>
-              </excludes>
+              <groups>!strace</groups>
```
npepinpe, line 2886: :x: You can exclude them by default in the main configuration. See where we set `!performance` on line 2346 of the parent/pom. Then we don't need the skip-strace-tests profile at all.

zeebe/snapshot/src/test/java/io/camunda/zeebe/snapshots/impl/SnapshotChecksumTest.java
```diff
@@ -25,10 +25,12 @@
 import org.agrona.IoUtil;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 import org.junit.jupiter.api.io.TempDir;
 
+@Tag("strace")
```
npepinpe, line 33: :wrench: I would normally add a marker annotation to avoid everyone knowing the magical constant, and ease refactoring (you can easily refactor types via your IDE, not so much strings).

e.g. (a file which lives in the `zeebe/test-util` module next to, say, `@JMHTest`)

```java
package io.camunda.zeebe.test.util.junit;

import java.lang.annotation.Documented;
import java.lang.annotation.ElementType;
import java.lang.annotation.Inherited;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;
import org.junit.jupiter.api.Tag;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;

/**
 * Marker annotation indicating the following test requires strace to be installed on the machine.
 * By default, such tests are run only in CI where we can guarantee strace is installed. Of course
 * the test will also run when running it explicitly via your IDE or command line.
 *
 * @see io.camunda.zeebe.test.util.STracer
 */
@Target({ElementType.METHOD})
@Documented
@Inherited
@Tag("strace")
@EnabledIfEnvironmentVariable(named = "CI", matches = "true")
@Test
public @interface StraceTest {}
```

Then in the test where we need it (`SnapshotChecksumTest.java`), you replace:

```java
  @Test
  void shouldFlushOnPersist() throws Exception {
```

With

```java
  @StraceTest
  void shouldFlushOnPersist() throws Exception {
```

npepinpe (changes requested):

parent/pom.xml
clementnero: Indeed, and I would have been blocked to concat/merge `<groups>!performance</groups>` and `<groups>!strace</groups>` anyway.

npepinpe (approved):
:tada:

npepinpe:
Hm, seems like the unit tests are still executing the strace tests :thinking:

parent/pom.xml
clementnero: Thx for pairing on this 👍

npepinpe:
I have to log off for the night, but if it doesn't work, we can do a quick sync pair call to clarify, and just fall back to excluding files if we can't make it work (though it worked for performance tests, so I don't get why it wouldn't here :smile:)

clementnero:
> Hm, seems like the unit tests are still executing the strace tests 🤔 We could try the following to exclude, since I'm not sure the `!performance,!strace` syntax works:
> 
> ```
> <excludedGroups>performance, strace</excludedGroups>
> ```

And I am also wondering if `excludedGroups` takes precedence over `groups`, I will do a couple of tests.

clementnero:
> Hm, it's failing with a familiar error, but one we thought we had solved 🤔 [ebfb2bc](https://github.com/camunda/camunda/commit/ebfb2bc44354410525c3ae4f9f918d0d5adb16eb)

According to https://man7.org/linux/man-pages/man1/strace.1.html

> If a system call is being executed and meanwhile another one is being called from a different thread/process then strace will try to preserve the order of those events and mark the ongoing call as being unfinished.  When the call returns it will be marked as resumed.

[Pattern](https://github.com/camunda/camunda/commit/ebfb2bc44354410525c3ae4f9f918d0d5adb16eb#diff-b495bc2bf35d1fa769918d096eb247d76c9b770b47fd81fd1afb47c67e09cf00R140) has been updated for ` <unfinished...>` call, but should also match when resumed ? 

> Expected line to match format of 'PID fsync(FD<PATH>) = RESULT', but '4407  <... fsync resumed>)

Could it be that this case is happening because of the GitHub-managed runners ? I don't think so, but lack context here

clementnero:
The [retry](https://github.com/camunda/camunda/actions/runs/11056372830/job/30745481888) worked.

cmur2:
> One quick question, are we backporting the migration to CI runners to older stable branches? It would help us by allowing us to remove our infrastructure entirely.

I initially didn't consider this in the scope but given the goal of #18212 is to remove the maintenance burden from the Zeebe team this makes sense. But we have to be careful since stable branches e.g. lack the CI Health monitoring so we have less visibility.

npepinpe:
Thanks, I'll reassign someone to look into the flakiness, possibly the pattern just needs to be updated.

As for backporting, consider it a best of effort :) If it ends up being too complex/time consuming, I would be OK with not doing it. We could consider that newer stable branches are more important (since they are supported longer) than older, so if, say, stable/8.3 is too complex to backport, just drop it (since we'll drop support for it in April, so it may be acceptable to live with for a while). You can already 100% ignore stable/8.2 since we will drop support for it in a few weeks.

clementnero:
> Thanks, I'll reassign someone to look into the flakiness, possibly the pattern just needs to be updated.

Good, I will put this PR on hold until then.

npepinpe:
It sounds like the flakiness is likely already present without this PR honestly, so I would be OK with merging it now, but if we're not pressed by time, we can also wait :)

npepinpe:
@EuroLew will look at it, time-boxed 1-2 hours today. If it takes longer, we'll split it off as a follow up and I think it's likely safe to merge, as imo the flakiness is probably there anyway.

EuroLew:
the fsync flake shouldn't happen anymore, https://github.com/camunda/camunda/pull/22609/commits/69068d7a29cd933c392df493b38a4e920c08ff41

When a fsync call is interrupted it will sometimes result in a resumed call we can safely ignore this because we know the initial unfinished trace line exists, and it will be processed and pass the test

clementnero:
Thx @EuroLew for the fix. @npepinpe it seems we are good go here  :)

npepinpe (approved):

PR #22632, opened 2024-09-24
Add deploy resource command REST support

### LINKED ISSUE #22191: Support the Deploy Resource API over REST

nicpuppa:
## Description

This PR add to the deploy resource command the REST support

## Related issues

closes #22191

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeployResourceCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeployResourceCommandImpl.java
@@ -20,6 +20,7 @@ import static io.camunda.zeebe.client.impl.command.ArgumentUtil.ensureNotNull;
 import com.google.protobuf.ByteString;
 import io.camunda.zeebe.client.CredentialsProvider.StatusCode;
 import io.camunda.zeebe.client.ZeebeClientConfiguration;
+import io.camunda.zeebe.client.api.JsonMapper;
 import io.camunda.zeebe.client.api.ZeebeFuture;
 import io.camunda.zeebe.client.api.command.ClientException;
 import io.camunda.zeebe.client.api.command.CommandWithTenantStep;
@@ -28,7 +29,10 @@ import io.camunda.zeebe.client.api.command.DeployResourceCommandStep1.DeployReso
 import io.camunda.zeebe.client.api.command.FinalCommandStep;
 import io.camunda.zeebe.client.api.response.DeploymentEvent;
 import io.camunda.zeebe.client.impl.RetriableClientFutureImpl;
+import io.camunda.zeebe.client.impl.http.HttpClient;
+import io.camunda.zeebe.client.impl.http.HttpZeebeFuture;
 import io.camunda.zeebe.client.impl.response.DeploymentEventImpl;
+import io.camunda.zeebe.client.protocol.rest.ResourceResponse;
 import io.camunda.zeebe.gateway.protocol.GatewayGrpc.GatewayStub;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.DeployResourceRequest;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.DeployResourceResponse;
@@ -46,23 +50,42 @@ import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
+import org.apache.hc.client5.http.config.RequestConfig;
+import org.apache.hc.client5.http.entity.mime.MultipartEntityBuilder;
+import org.apache.hc.core5.http.ContentType;
 
 public final class DeployResourceCommandImpl
     implements DeployResourceCommandStep1, DeployResourceCommandStep2 {
 
+  private static final String RESOURCES_FIELD_NAME = "resources";
+  private static final String TENANT_FIELD_NAME = "tenantId";
+
+  private final MultipartEntityBuilder multipartEntityBuilder =
+      MultipartEntityBuilder.create().setContentType(ContentType.MULTIPART_FORM_DATA);
   private final DeployResourceRequest.Builder requestBuilder = DeployResourceRequest.newBuilder();
   private final GatewayStub asyncStub;
   private final Predicate<StatusCode> retryPredicate;
   private Duration requestTimeout;
+  private final HttpClient httpClient;
+  private final RequestConfig.Builder httpRequestConfig;
+  private boolean useRest;
+  private final JsonMapper jsonMapper;
 
   public DeployResourceCommandImpl(
       final GatewayStub asyncStub,
       final ZeebeClientConfiguration config,
-      final Predicate<StatusCode> retryPredicate) {
+      final Predicate<StatusCode> retryPredicate,
+      final HttpClient httpClient,
+      final boolean preferRestOverGrpc,
+      final JsonMapper jsonMapper) {
     this.asyncStub = asyncStub;
     requestTimeout = config.getDefaultRequestTimeout();
     this.retryPredicate = retryPredicate;
     tenantId(config.getDefaultTenantId());
+    this.httpClient = httpClient;
+    httpRequestConfig = httpClient.newRequestConfig();
+    useRest = preferRestOverGrpc;
+    this.jsonMapper = jsonMapper;
   }
 
   /**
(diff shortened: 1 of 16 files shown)


clients/java/src/test/java/io/camunda/zeebe/client/resource/rest/DeployResourceRestTest.java
```diff
@@ -0,0 +1,473 @@
...
+        .addResourceFromClasspath(filename2)
+        .tenantId(tenantId)
+        .send()
+        .join();
+
+    // then
+    LoggedRequestAssert.assertThat(RestGatewayService.getLastRequest())
+        .hasMethod(RequestMethod.POST)
+        .hasUrl(RestGatewayPaths.getDeploymentsUrl());
+  }
```
tmetzke, line 199: ❓ Are we able to assert the resource data bytes that we send from the client like we do for gRPC?

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/CreateProcessInstanceTest.java
```diff
@@ -338,12 +338,12 @@ private CreateProcessInstanceCommandStep1 getCommand(
     return useRest ? createInstanceCommand.useRest() : createInstanceCommand.useGrpc();
   }
 
-  private void deployProcesses(final TestInfo testInfo) {
+  private void deployProcesses(final TestInfo testInfo, final boolean useRest) {
```
tmetzke, line 341: 🔧 Neat trick but I think we need to adjust the `CreateDeployment...Test` classes to thoroughly check all deployment use cases work equally well.

tmetzke (commented):
👍 Looks really good already, nice work!
🔧 Tenant handling and testing needs another close look I think to ensure we don't introduce unwanted behavior.

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeployResourceCommandImpl.java
```diff
@@ -173,31 +208,66 @@ public DeployResourceCommandStep2 addProcessModel(
...
+      return sendRestRequest();
+    } else {
+      return sendGrpcRequest();
+    }
+  }
+
+  @Override
+  public DeployResourceCommandStep2 tenantId(final String tenantId) {
+    requestBuilder.setTenantId(tenantId);
+    multipartEntityBuilder.addTextBody(TENANT_FIELD_NAME, tenantId);
```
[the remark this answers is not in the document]
nicpuppa: Yes you are right. I fixed it, but I lost the changes while refactoring 😄

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/CreateProcessInstanceTest.java
nicpuppa: I missed those tests 🤔

clients/java/src/test/java/io/camunda/zeebe/client/resource/rest/DeployResourceRestTest.java
nicpuppa: I'll have a look

tmetzke: Synced offline:
* The gRPC request simply holds the request data separately which is why it's easily accessible there for tests.
* In the REST request, this is encoded inside the multipart form data request, which makes it less accessible in the tests.
* We'll create a follow-up to improve the tests. We're confident that with integration tests that use the client and verify the content arrives well we're well covered for now nonetheless.

dist/src/main/resources/application.properties
```diff
@@ -61,3 +61,7 @@ camunda.identity.init.users[0].username=demo
 camunda.identity.init.users[0].password=demo
 spring.jpa.generate-ddl=false
 spring.jpa.hibernate.ddl-auto=validate
+
+# Multipart file uploads
+spring.servlet.multipart.max-file-size=4MB
+spring.servlet.multipart.max-request-size=4MB
```
nicpuppa, line 67: @tmetzke this two properties should have the same value of the `maxMessageSize` configured in zeebe, I think we should document this somewhere. Informing the customer to keep them aligned

nicpuppa: The default values are:

```
spring.servlet.multipart.max-file-size=1MB
spring.servlet.multipart.max-request-size=10MB
```

tmetzke: Hmm, interesting. Is this configuring the server to allow those sizes?

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/HttpClient.java
```diff
@@ -187,7 +187,10 @@ private <HttpT, RespT> void sendRequest(
         final HttpEntity entity = (HttpEntity) body;
         final byte[] entityBytes;
         try {
-          entityBytes = EntityUtils.toByteArray(entity);
+          final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+          entity.writeTo(byteArrayOutputStream);
+          entityBytes = byteArrayOutputStream.toByteArray();
+          byteArrayOutputStream.close();
         } catch (final IOException e) {
```
tmetzke: 🔧 `ByteArrayOutputStream` is an auto-closable, so it should be safer to write a try-with-resources statement.
```suggestion
        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
          entityBytes = EntityUtils.toByteArray(entity);
          entity.writeTo(byteArrayOutputStream);
          entityBytes = byteArrayOutputStream.toByteArray();
        } catch (final IOException e) {
```

tmetzke (approved):
👍 Pre-approving because the code looks good and works correctly.
🔧 One smaller suggestion, nothing big.
❓ I am not 100% sure about the properties and how to document them. Do we need to adjust any downstream components that take care of configuration (helm, or anything)? I am not too familiar with that, to be honest. I was also wondering if we could set those by default in our application code and allow users to override it like they can override the max message size.

dist/src/main/resources/application.properties
nicpuppa: Yes, but they still need to be the same value configured in zeebe for the maxMessageSize. Otherwise we will face this [error](https://github.com/camunda/camunda/blob/4d9013e77cdc01678e59a66c9643492d0a0d6e9d/zeebe/broker/src/main/java/io/camunda/zeebe/broker/transport/commandapi/CommandApiRequestHandler.java#L151-L154)

nicpuppa:
> ❓ I am not 100% sure about the properties and how to document them. Do we need to adjust any downstream components that take care of configuration (helm, or anything)? I am not too familiar with that, to be honest. I was also wondering if we could set those by default in our application code and allow users to override it like they can override the max message size.

I think helm charts should be adjusted, I'm not aware of any other components. 
Regarding the default values, they already have a default value. We are overriding it from Spring, I guess users can do the same (?). But I'm not sure if that `application.properties` file can be overridden or is it more an internal config file

dist/src/main/resources/application.properties
tmetzke: But only if users put up a payload that is too big, right? If that is the case, it's totally fine that the defaults of the servlet stay untouched, I think. The message size is the limitation inside the broker and if it cannot handle the request, it will be rejected. The REST request should fail with a reasonable error, explaining that the payload was too big.

Is there any behavior that is unexpected in that regard so we should change those properties?

nicpuppa: If the customers use a lower `maxMessageSize` (our default is 4MB) there is no problem, cause the request will be rejected by the broker. 
But customers can use also an higher `maxMessageSize` (> 4MB), in this case the request will be rejected by Spring itself

tmetzke: Yeah, I see what you mean. OK, let's adjust this here and think about proper documentation and configuration downstream in a follow-up issue 👍

tmetzke (approved):

tmetzke:
> > ❓ I am not 100% sure about the properties and how to document them. Do we need to adjust any downstream components that take care of configuration (helm, or anything)? I am not too familiar with that, to be honest. I was also wondering if we could set those by default in our application code and allow users to override it like they can override the max message size.
> 
> I think helm charts should be adjusted, I'm not aware of any other components. Regarding the default values, they already have a default value. We are overriding it from Spring, I guess users can do the same (?). But I'm not sure if that `application.properties` file can be overridden or is it more an internal config file
> 
> I don't know also how it works for Saas, and how configure new properties there

You're right 👍 Can you create a follow-up for this please so we can figure out what else to adjust so it works by default well for users? Worst case is it only works for small files by default. We should also document this well, not sure yet where though.

PR #22655, opened 2024-09-24
[POC] Permission verification on command processor level

remcowesterhoud:
# POC, Do not merge

**Proposed implementation on a technical level:**
We should aim to largely abstract the permission verification away. Especially with the trend of receiving more contributions within Zeebe we should aim to create an interface that easily enables developers to do permission checks, without worrying too much about the implementation details. However, we should still force those contributors to consider authorizations, so we don’t miss permission checks when new processors get implemented.

As of now I can see 5 different scenarios in command processing that we need to differentiate between:
1. We need to check permissions in a TypedRecordProcessor for a wildcard or specific resource id (eg. bpmnProcessId)
2. We need to check permissions in a DistributedTypedRecordProcessor for a wildcard or specific resource id (eg. bpmnProcessId)
3. We need to check permissions in a TypedRecordProcessor for a wildcard
4. We need to check permissions in a DistributedTypedRecordProcessor for a wildcard
5. We need to check permissions in a CommandProcessor for a wildcard or specific resource id (eg. bpmnProcessId)
    - This case is only relevant for:
        - Process instance creation
        - Process instance creation with result
        - Job cancellation
        - Job completion
        - Job error throwing
        - Ideally we get rid of this difference in our processors and align them all. For now, we need to work around this problem.

My proposed implementation is to use a Decorator pattern which will push the permission checks to a separate class. There would be 5(?) new TypedProcessor implementations, 1 for each of the above scenarios (names are subject to change):
1. AuthorizableResourceProcessor
2. AuthorizableDistributionResourceProcessor
3. AuthorizableProcessor
4. AuthorizableDistributionProcessor
5. AuthorizablCommandProcessor

When registering a processor the above processors are used to wrap the existing processor. The wrapped processor will no longer be a TypedRecordProcessor. Instead, they will implement a new interface. Depending on the scenario the interface will provide the wrapped processor with methods (names subject to change):
- processRecord(command, resource)
    - Used for commands that need to verify permissions on a resource specific level
- processNewCommand(command, resource) & processDistributedCommand(command, resource)
    - Used for commands that need to verify permissions on a resource specific level, and need to be distributed
- getResource()
    - Returns some wrapper object. This object contains the date needed to verify the permissions, such as:
        - ResourceType
        - PermissionType
        - Resource (mainly used to prevent duplicate state lookups)
        - ResourceIds (a Map of resourceIds, such as {bpmnProcessId: foo})

To make sure we don’t forget adding permissions verification to new processors that are created we will implement an ArchUnit test. In this test any new processor must be explicitly excluded. In a time where more people without in-depth Zeebe knowledge will be contributing, this test forces the dev and reviewer to think whether permissions need to be checked on their new command.

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/BpmnProcessors.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/BpmnProcessors.java
@@ -11,6 +11,7 @@ import io.camunda.zeebe.engine.metrics.ProcessEngineMetrics;
 import io.camunda.zeebe.engine.processing.bpmn.BpmnStreamProcessor;
 import io.camunda.zeebe.engine.processing.bpmn.behavior.BpmnBehaviors;
 import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.identity.behavior.AuthorizationCheckBehavior;
 import io.camunda.zeebe.engine.processing.message.PendingProcessMessageSubscriptionChecker;
 import io.camunda.zeebe.engine.processing.message.ProcessMessageSubscriptionCorrelateProcessor;
 import io.camunda.zeebe.engine.processing.message.ProcessMessageSubscriptionCreateProcessor;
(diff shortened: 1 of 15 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/BpmnProcessors.java
```diff
@@ -119,7 +121,11 @@ private static void addProcessInstanceCommandProcessor(
     typedRecordProcessors.onCommand(
         ValueType.PROCESS_INSTANCE,
         ProcessInstanceIntent.CANCEL,
-        new ProcessInstanceCancelProcessor(processingState, writers));
+        new AuthorizableProcessDefinitionProcessor<>(
+            new AuthorizationCheckBehavior(
+                processingState.getAuthorizationState(), processingState.getUserState()),
+            writers,
+            new ProcessInstanceCancelProcessor(processingState, writers)));
   }
```
korthout: 🔧 You could probably move the construction of `AuthorizationCheckBehavior` into the constructor of `AuthorizableProcessDefinitionProcessor` to clarify that the `ProcessInstanceCancelProcessor` is decorated.

Issue #22689, opened 2024-09-25
Replace reports endpoint with entities in create new button component


misiekhardcore:

OmranAbazid:
This issue require a bit more work from my side. Since it is alow priority, i will keep it open until i am back from my leave.

RomanJRW:
@OmranAbazid - should we close the PRs for now too?

OmranAbazid:
@RomanJRW I closed the PRs. I will reopen them when i continue working on this.

PR #22693, opened 2024-09-25
refactor: replace reports endpoint with entites one

misiekhardcore:
## Checklist

- [ ] for CI changes:

## Related issues

related to https://github.com/camunda/camunda/issues/22689

### DIFF
--- a/optimize/client/src/components/Dashboards/AddButton/CreateTileModal.tsx
+++ b/optimize/client/src/components/Dashboards/AddButton/CreateTileModal.tsx
@@ -12,9 +12,9 @@ import {Button, ComboBox, Form, TextInput, TextInputSkeleton} from '@carbon/reac
 import {SerializedEditorState} from 'lexical';
 
 import {Modal, Tabs, TextEditor} from 'components';
-import {getCollection, isTextTileValid, loadReports} from 'services';
+import {getCollection, isTextTileValid, loadEntities} from 'services';
 import {t} from 'translation';
-import {DashboardTile, GenericReport} from 'types';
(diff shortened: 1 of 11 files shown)


optimize/client/src/components/Dashboards/AddButton/CreateTileModal.tsx
```diff
@@ -34,7 +34,7 @@ export default function CreateTileModal({close, confirm}: CreateTileModalProps)
   useEffect(() => {
     (async () => {
       const collection = getCollection(pathname);
-      const availableReports = await loadReports(collection);
+      const availableReports = await loadEntities(collection);
```
OmranAbazid, line 37: Since we are not filtering for reports here, this is currently listing other entities.

optimize/client/src/components/Home/Home.js
```diff
@@ -61,7 +61,7 @@ export function Home({mightFail, user}) {
       setIsLoading(true);
       setSorting({key: sortBy, order: sortOrder});
       mightFail(
-        loadEntities(sortBy, sortOrder),
+        loadEntities(null, sortBy, sortOrder),
```
OmranAbazid, line 64: I do not like this as it is not very readable for someone looking at this the first time.
I suggest creating a function called `loadAllEntities` and in `loadEntities` function we can either call the `loadAllEntities` one or the `loadCollectionEntities` one if a collection is provided.

misiekhardcore: WHYT if we change the order of arguments so the collection ID comes last? This way we have just one function still, but we dont have to pass `null` as the first argument?

PR #22697, opened 2024-09-25
feat: added plugins at shared ES/OS clients

igpetrov:
## Description

feat: added plugins at shared ES client

Similar to this: https://github.com/camunda/camunda/issues/21559

## Checklist

- [ ] for CI changes:

## Related issues

Related to: https://github.com/camunda/camunda/issues/21559

### DIFF
--- a/search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
+++ b/search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
@@ -14,6 +14,7 @@ import io.camunda.search.connect.configuration.ConnectConfiguration;
 import io.camunda.search.connect.configuration.SecurityConfiguration;
 import io.camunda.search.connect.jackson.JacksonConfiguration;
 import io.camunda.search.connect.os.json.SearchRequestJacksonJsonpMapperWrapper;
+import io.camunda.search.connect.plugin.PluginRepository;
 import io.camunda.search.connect.util.SecurityUtil;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -26,6 +27,7 @@ import org.apache.hc.client5.http.impl.nio.PoolingAsyncClientConnectionManagerBu
 import org.apache.hc.client5.http.ssl.ClientTlsStrategyBuilder;
 import org.apache.hc.client5.http.ssl.NoopHostnameVerifier;
 import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.HttpRequestInterceptor;
 import org.apache.hc.core5.util.Timeout;
 import org.opensearch.client.opensearch.OpenSearchAsyncClient;
 import org.opensearch.client.opensearch.OpenSearchClient;
@@ -35,6 +37,7 @@ import org.opensearch.client.transport.aws.AwsSdk2TransportOptions;
 import org.opensearch.client.transport.httpclient5.ApacheHttpClient5TransportBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
 import software.amazon.awssdk.http.nio.netty.NettyNioAsyncHttpClient;
 import software.amazon.awssdk.regions.Region;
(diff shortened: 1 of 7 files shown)


igpetrov:
Hi @romansmirnov 

Similar to https://github.com/camunda/camunda/issues/21559, I implemented plugin support for shared ES/OS clients. I know that they are not currently used, but I did it anyway so that in future we don't forget about this feature.

Please take a look and let me know what you think.

search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
```diff
@@ -45,24 +48,44 @@ public final class OpensearchConnector {
...
+        new JacksonConfiguration(configuration).createObjectMapper(),
+        DefaultCredentialsProvider.create(),
+        new PluginRepository());
   }
 
   public OpensearchConnector(
-      final ConnectConfiguration configuration, final ObjectMapper objectMapper) {
+      final ConnectConfiguration configuration,
+      final ObjectMapper objectMapper,
+      final AwsCredentialsProvider credentialsProvider,
```
romansmirnov, line 66: ❓ Why was this change necessary?

romansmirnov (commented):
In general, it looks good, pretty straightforward. However, I have a question because I don't understand why the `OpensearchConnector` now also accepts a `AwsCredentialsProvider`.

search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
igpetrov: It was very hard to test/mock: because method `boolean shouldCreateAWSBasedTransport()` explicitly invokes `DefaultCredentialsProvider.create()`. See, the  `DefaultCredentialsProvider` initializes in chain `EnvironmentVariableCredentialsProvider`, `SystemPropertyCredentialsProvider`, and few others. This means, that people (like myself), who have a set of configs for AWS accounts as env vars, will have AWS initialized, thus making impossible to test local instance of OpenSearch.

So I decided that since there are multiple implementations of the `AwsCredentialsProvider`, it might not hurt to move it into constructor. Please keep in mind, that the backwards compatibility is respected, and when not defined, setting `DefaultCredentialsProvider.create()`.

WDYT?

igpetrov:
Adding @npepinpe since @romansmirnov seems to be on FTO

igpetrov:
Also adding @mustafadagher who has some context on this topic

search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
```diff
@@ -45,24 +48,44 @@ public final class OpensearchConnector {
 
   private final ConnectConfiguration configuration;
   private final ObjectMapper objectMapper;
+  private final PluginRepository pluginRepository;
+
+  private AwsCredentialsProvider credentialsProvider;
```
mustafadagher, line 53: 🔧 this could be final, since it's either passed to or initialized in a constructor.

search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
```diff
@@ -125,14 +152,21 @@ private HttpHost getHttpHost(final ConnectConfiguration osConfig) {
...
+    } catch (final URISyntaxException e) {
       throw new SearchClientConnectException("Error in url: " + osConfig.getUrl(), e);
     }
   }
 
   protected HttpAsyncClientBuilder configureHttpClient(
-      final HttpAsyncClientBuilder httpAsyncClientBuilder, final ConnectConfiguration osConfig) {
+      final HttpAsyncClientBuilder httpAsyncClientBuilder,
+      final ConnectConfiguration osConfig,
+      final HttpRequestInterceptor... interceptors) {
```
mustafadagher, line 163: ❓ why an array of interceptors here, when we only ever pass one?

mustafadagher (approved):
Thanks @igpetrov! LGTM 🚀 

I've only left a few questions/suggestions in comments below.

search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
igpetrov: Good point. Will make it!

search/search-client-connect/src/main/java/io/camunda/search/connect/os/OpensearchConnector.java
igpetrov: There was a case when configuration required multiple plugin repositories / interceptors, so we decided that we will support an array of interceptors. In addition to that, the plugin functionality will grow in similar way as Optimize C7 had.

search/search-client-connect/src/test/java/io/camunda/search/connect/os/OpensearchConnectorTest.java
```diff
@@ -0,0 +1,131 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.mockito.Mockito;
+import org.opensearch.client.transport.httpclient5.ApacheHttpClient5Transport;
+
+class OpensearchConnectorTest {
+
+  @RegisterExtension
+  // Native OS clients refuse to work while
+  // actual HTTP(S) connection is available
```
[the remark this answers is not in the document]
igpetrov: So, if I try executing any non-existing HTTP endpoint with `org.opensearch.client.transport.httpclient5.ApacheHttpClient5Transport` (e.g., `localhost:1234`, assuming it is unavailable), the HTTP client won't even do anything, just silently fail. Therefore in this case, I have to actually kick on a wiremock. And then it does perform an HTTP call, including adding headers.

This is different to ES client, where header will be added to whatever non-existing endpoint: see similar `ElasticsearchConnectorTest`.

PR #22702, opened 2024-09-25
update rejection message for subscribing to the same message

### LINKED ISSUE #20906: Update rejection message for resubscribing to the same message name during instance migration
## Description

Error message for rejection and the explanation in the document should be aligned. E.g. use closing and opening instead of re-subscribing.

berkaycanbc:
## Description

The rejection message is updated to make the error reason more clear. Also, the workaround is changed to be more exact.

## Related issues

closes #20906


berkaycanbc:
@ThorbenLindhauer I assigned you as the reviewer based on our discussion under the related issue #20906. Please let me know if you would consider to modify the rejection message further.

ThorbenLindhauer (approved):

PR #22735, opened 2024-09-26
Raise incident at specific called process depth

### LINKED ISSUE #16410: Banning escape hatch doesn't work as expected on termination for recursive process models
**Describe the bug**

When canceling an instance we do this until a certain depth (until we [reach the stack limit](https://github.com/camunda/zeebe/blob/b703e3b09f97ce84dbd6393de54aaeab7eb88fa8/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnStateTransitionBehavior.java#L438)), if we reach that we stop terminating (as we are not able to write all the terminations) and ban the last instance.

There are two problems with this:

 1. All other instances in this hierarchy are not affected by this. They can still continue/complete.
 2. In the case of a recursion loop (which we had several time now) it is likely that this will not stop the recursion. The problem here is that the creation (or the recurison) is done in batches, if we try to cancel and a batch comes afterwards that continues the recursion the banning will only affect on instance in between the recursion.

See related game day https://confluence.camunda.com/display/HAN/2024-02-15+Game+Day and incident https://app.slack.com/client/T0PM0P1SA/C06JV3B18S1

* Put severity high as it is not resolving the issue, when the customer/user wants to cancel the faulty process model
* Likelihood low as we don't expect that deploying such models will happen that often. But if it does then likelihood is high with a recursion loop deployed

**To Reproduce**
Run a recursion loop
recursion-loop

[recursion-loop.bpmn.txt](https://github.com/camunda/zeebe/files/14310625/recursion-loop.bpmn.txt)

After creating some instances try to cancel the root instance.

**Expected behavior**
Either canceling works (use a loop or batching) or banning (no longer execution of child ) 

**Log/Stacktrace**

 <p>

```
io.camunda.zeebe.engine.processing.bpmn.behavior.BpmnStateTransitionBehavior$ChildTerminationStackOverflowException: Process instance `2251799821027706` has too many nested child instances and could not be terminated. The deepest nested child instance has been banned as a result.
	at io.camunda.zeebe.engine.processing.bpmn.behavior.BpmnStateTransitionBehavior.lambda$onElementTerminated$5(BpmnStateTransitionBehavior.java:437) ~[zeebe-workflow-engine-8.4.2.jar:8.4.2]
	at 

Remove several lines; it was way too long.
```

</p>
</details>

**Environment:**
- OS: SaaS
- Zeebe Version: 8.4.2 
- Configuration: 

https://jira.camunda.com/browse/SUPPORT-23526

korthout:
## Description

This is the effort we made during our mob programming session. We added a test case with a failsafe to ensure the process instance doesn't overload the test engine. We also added a simple implementation to check the called process depth.

Still to do:
- actually increment the depth
- make the depth configurable
- consider how to deal with existing child instances
- configure the depth in the test for performance

## Related issues

closes #16410

### DIFF
--- a/dist/src/main/config/broker.yaml.template
+++ b/dist/src/main/config/broker.yaml.template
@@ -1258,6 +1258,18 @@
           # This setting can also be overridden using the environment variable ZEEBE_BROKER_EXPERIMENTAL_ENGINE_AUTHORIZATIONS_ENABLEAUTHORIZATION
           # enableAuthorization: false
 
+        # Allows to configure the maximum depth of child process instances that can be created.
+        # The root process instance is considered depth `1`, and each called instance is considered
+        # to be one level deeper. If a call activity attempts to create a child process instance
+        # that would exceed the configured max depth, an incident is raised at the call activity.
+        # Note that the depth is only tracked for process instances created since 8.3.22, 8.4.18,
+        # 8.5.17, 8.6.12, and 8.7. Any child process instances created before 8.7 (or any of these
+        # patches) will have a depth of 1 rather than the correct depth. For existing child process
+        # instances, the limit is thus applied only from the current child instance onwards.
+        # This setting can also be overridden using the environment variable
+        # ZEEBE_BROKER_EXPERIMENTAL_ENGINE_MAXPROCESSDEPTH
+        # maxProcessDepth: 1000
+
       # Allows to configure feature flags. These are used to test new features in dev and int environments prior
       # to rolling them out to production
       # features:
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/EngineConfiguration.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/EngineConfiguration.java
@@ -144,7 +144,8 @@ public final class EngineConfiguration {
     return maxProcessDepth;
   }
 
-  public void setMaxProcessDepth(final int maxProcessDepth) {
+  public EngineConfiguration setMaxProcessDepth(final int maxProcessDepth) {
     this.maxProcessDepth = maxProcessDepth;
+    return this;
   }
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/EngineConfiguration.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/EngineConfiguration.java
@@ -28,6 +28,8 @@ public final class EngineConfiguration {
   public static final int DEFAULT_VALIDATORS_RESULTS_OUTPUT_MAX_SIZE = 12 * 1024;
   public static final boolean DEFAULT_ENABLE_AUTHORIZATION_CHECKS = false;
 
+  public static final int DEFAULT_CALL_ACTIVITY_MAX_DEPTH = 1000;
+
   private int messagesTtlCheckerBatchLimit = DEFAULT_MESSAGES_TTL_CHECKER_BATCH_LIMIT;
   private Duration messagesTtlCheckerInterval = DEFAULT_MESSAGES_TTL_CHECKER_INTERVAL;
   private int drgCacheCapacity = DEFAULT_DRG_CACHE_CAPACITY;
@@ -41,6 +43,8 @@ public final class EngineConfiguration {
   private int validatorsResultsOutputMaxSize = DEFAULT_VALIDATORS_RESULTS_OUTPUT_MAX_SIZE;
 
   private boolean enableAuthorization = DEFAULT_ENABLE_AUTHORIZATION_CHECKS;
+
+  private int defaultCallActivityMaxDepth = DEFAULT_CALL_ACTIVITY_MAX_DEPTH;
 
   public int getMessagesTtlCheckerBatchLimit() {
     return messagesTtlCheckerBatchLimit;
@@ -134,5 +138,13 @@ public final class EngineConfiguration {
   public EngineConfiguration setEnableAuthorization(final boolean enableAuthorization) {
     this.enableAuthorization = enableAuthorization;
     return this;
+  }
+
+  public int getDefaultCallActivityMaxDepth() {
+    return defaultCallActivityMaxDepth;
+  }
+
+  public void setDefaultCallActivityMaxDepth(final int defaultCallActivityMaxDepth) {
+    this.defaultCallActivityMaxDepth = defaultCallActivityMaxDepth;
   }
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/EngineConfiguration.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/EngineConfiguration.java
@@ -28,7 +28,7 @@ public final class EngineConfiguration {
   public static final int DEFAULT_VALIDATORS_RESULTS_OUTPUT_MAX_SIZE = 12 * 1024;
   public static final boolean DEFAULT_ENABLE_AUTHORIZATION_CHECKS = false;
 
-  public static final int DEFAULT_CALL_ACTIVITY_MAX_DEPTH = 1000;
+  public static final int DEFAULT_MAX_PROCESS_DEPTH = 1000;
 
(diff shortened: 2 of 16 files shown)


korthout:
Considerations regarding existing process instances:
- they either don't have the depth property set, or we need a migration task (which may be slow)
- if we use a default of `0` or `-1`, we can allow incrementing from now on. This will at least restrict existing process instances to some limit (current depth + configured limit), but this means that the configured limit is not exactly applied to existing process instances
- alternatively, we can avoid incrementing for `-1`. This allows us to either: accept that the safeguard doesn't work at all on existing process instances, or to recursively look up the depth on any call activity activation. This prior option means we don't win anything for existing process instances, and the latter option can come with a performance hit to existing process instances.
- alternatively, we can look up the entire depth recursively once when we encounter `-1` and apply the `depth + 1` to the next level (e.g. `0 > 0 > 0 > 0 > 4`). This slows down a single call activity activation, but at least guarantees we apply the configured limit more truly. One example where this can hurt is an existing process instance with multi-instance call activity.

remcowesterhoud:
Update from the mob-programming session of Oct 3rd:

We focussed on actually increment the depth. We did so by incrementing the depth on the `ElementInstance` in the `ProcessInstanceElementActivatingApplier`. We managed to make it work, but the code can probably be cleaned up a bit still.

What's left:
- Set the depth to 0 for initial process instances.
- make the depth configurable
- consider how to deal with existing child instances
- configure the depth in the test for performance

korthout:
I've rebased this on `main` +7000 commits :D

koevskinikola:
Update from the mob programming session on Jan 9th, 2025:

* We noticed that having a depth of 1010 in the test case was triggering a timeout in the `RecordingExporter`. Increasing the timeout value to 10000 (10secs) didn't help. However, reducing the CallActivity depth (in the corresponding processor) for which an Incident is raised got the test case to pass.
* We focused on making the default call activity depth configurable through the `EngineConfiguration` class. This is now completed.
* Furthermore, we expanded the `EngineRule` class to allow for modifications of the `EngineConfiguration` class. This allowed us to set a custom depth of `10` in the test case for test performance reasons.

Participants in the mob programming session: @koevskinikola, @remcowesterhoud, @berkaycanbc, @ana-vinogradova-camunda, @ce-dmelnych, @eppdot 

What's left:

* Set the depth to 0 for initial process instances.
* consider how to deal with existing child instances

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceElementActivatingApplier.java
```diff
@@ -122,16 +124,22 @@ private void moveVariablesToNewEventScope(
...
       final var parentElementType = parentElementInstance.getValue().getBpmnElementType();
       if (parentElementType == BpmnElementType.CALL_ACTIVITY) {
-        parentElementInstance.setCalledChildInstanceKey(elementInstanceKey);
+        parentElementInstance.setCalledChildInstanceKey(elementInstance.getKey());
         elementInstanceState.updateInstance(parentElementInstance);
+
+        final var parentProcessInstance =
+            elementInstanceState.getInstance(
+                elementInstance.getValue().getParentProcessInstanceKey());
+        elementInstance.setCalledProcessDept(parentProcessInstance.getCalledProcessDepth() + 1);
```
korthout, line 141:
```suggestion
        elementInstance.setCalledProcessDepth(parentProcessInstance.getCalledProcessDepth() + 1);
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/EngineConfiguration.java
```diff
@@ -135,4 +139,12 @@ public EngineConfiguration setEnableAuthorization(final boolean enableAuthorizat
...
     return this;
   }
+
+  public int getDefaultCallActivityMaxDepth() {
+    return defaultCallActivityMaxDepth;
+  }
+
+  public void setDefaultCallActivityMaxDepth(final int defaultCallActivityMaxDepth) {
+    this.defaultCallActivityMaxDepth = defaultCallActivityMaxDepth;
+  }
```
korthout, line 149: 🔧 My feeling is that the wording default does not make sense for this configuration option.

```suggestion
  public int getCallActivityMaxDepth() {
    return callActivityMaxDepth;
  }

  public void setCallActivityMaxDepth(final int callActivityMaxDepth) {
    this.callActivityMaxDepth = callActivityMaxDepth;
  }
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/ElementInstance.java
```diff
@@ -52,6 +52,15 @@ public final class ElementInstance extends UnpackedObject implements DbValue {
   private final ObjectProperty<TaskListenerIndicesRecord> taskListenerIndicesRecordProp =
       new ObjectProperty<>("taskListenerIndicesRecord", new TaskListenerIndicesRecord());
 
+  /**
+   * This value is added in 8.7, any child process instances created before 8.7 will have a depth of
+   * 0.
+   *
+   * @since 8.7
```
korthout, line 59: 🔧 Child process instances created after 8.7, but as part of root process instance that was created prior to 8.7 would not have a correct depth. Their depth would be counted as the depth since 8.7, i.e. the first child process instance since 8.7 will have depth `1`.

🔧 We should document this point in the description of the configuration option as well for correctness.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/ElementInstance.java
```diff
@@ -301,4 +301,27 @@ public void resetTaskListenerIndices() {
...
+   *     depth of 1 + the depth of the parent instance. Therefore, child instances created on or
+   *     after 8.7 that are part of a root process instance created prior to 8.7, will not have a
+   *     correct depth.
+   */
+  public int getProcessDepth() {
+    final int processDepth = getValue().getProcessDefinitionPath().size();
+    if (processDepth == 0) {
+      // This process instance was created prior to 8.7 and its process definition path has not been
+      // tracked yet. As we cannot determine the depth, we return 1 as a fallback.
+      return 1;
```
korthout, line 323: 🔧 We need to verify how the process definition path behaves on instances created prior to 8.7.

korthout: I've been thinking more about this approach, and I don't think it's a good idea to use it. Instead, I will revert to tracking the `processDepth` independently from the process definition path.

- it's not available on versions prior to 8.7 and so we cannot backport it, that's not the case when we track the processDepth independently
- tracking the processDepth has some performance impact to call activity activation (check before activation, and track on activation), but this it's not likely that this will greatly impact general execution performance (throughput/latency)
- using the same solution before and after 8.7 makes it easy to backport any changes we make to this in the future

dist/src/main/config/broker.yaml.template
```diff
@@ -1258,6 +1258,18 @@
...
+        # The root process instance is considered depth `1`, and each called instance is considered
+        # to be one level deeper. If a call activity attempts to create a child process instance
+        # that would exceed the configured max depth, an incident is raised at the call activity.
+        # Note that the depth is only tracked for process instances created since 8.3.22, 8.4.18,
+        # 8.5.17, 8.6.12, and 8.7. Any child process instances created before 8.7 (or any of these
+        # patches) will have a depth of 1 rather than the correct depth. For existing child process
+        # instances, the limit is thus applied only from the current child instance onwards.
+        # This setting can also be overridden using the environment variable
+        # ZEEBE_BROKER_EXPERIMENTAL_ENGINE_MAXPROCESSDEPTH
+        # maxProcessDepth: 1000
```
ana-vinogradova-camunda, line 1271: Very useful and detailed comments, thank you 👍

ana-vinogradova-camunda (approved):
Great PR, thank you @korthout 🚀  🎆 
I am really happy to see this issue resolved. Great team effort and thank you for your work @korthout 🏅

PR #22750, opened 2024-09-26
feat: implement get form for user task

### LINKED ISSUE #22449: Implement `Get Form for User Task`
### Description
Implement the `GET /v2/user-tasks/{userTaskKey}/form` endpoint for retrieving a form by user task key. This endpoint should return detailed form information including tenant ID, BPMN ID, schema, version, and more.

---

### Endpoint
`GET /v2/user-tasks/{userTaskKey}/form`

#### Request
- **Path Parameter:**
  - `userTaskKey` (Long): The unique key for User Task Key 

#### Response
- **Success (200 OK):**
  - Returns a JSON object representing the form with the following fields:
    - `formKey`: The unique key of the form.
    - `tenantId`: The tenant ID associated with the form.
    - `bpmnId`: The BPMN ID of the form.
    - `schema`: The schema representing the form.
    - `version`: The version of the form.

  **Example Response:**
  ```json
  {
    "formKey": 1,
    "tenantId": "tenant-1",
    "bpmnId": "bpmn-1",
    "schema": {},
    "version": 1
  }

Tasks:
- Implement the Service
- Implement the Rest endpoint
- Implement the Client
- Implement the integration Tests

nathansandi:
## Description

- Remove Form Controller and respective unit tests
- Add `/user-tasks/{userTaskKey}/form` controller endpoint
- Add Unit tests for the controller endpoint
- Client Implementation
- Integration Tests 

## Related issues

https://github.com/camunda/camunda/issues/22449


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
```diff
@@ -78,4 +82,27 @@ public ResponseEntity<Object> getByKey(@PathVariable("userTaskKey") final Long u
...
+  @GetMapping(
+      path = "/{userTaskKey}/form",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE})
+  public ResponseEntity<Object> getFormByUserTaskKey(
+      @PathVariable("userTaskKey") final Long userTaskKey) {
+    try {
+      final Long formKey = userTaskServices.getByKey(userTaskKey).formKey();
+
+      if (formKey == null) {
+        throw new NotFoundException(
```
houssain-barouni, line 95: I wonder if we should rather return HTTP 204, to indicate that the request was processed successfully, but there is no form to return.
I think we should reserve the 404 to when:
- the task with the key provided by the caller does not exist.
- the task has a formKey but the form cannot be found

houssain-barouni (commented):
I had one comment regarding the return HTTP code

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -542,6 +542,63 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+  /user-tasks/{userTaskKey}/form:
+    get:
+      tags:
+        - User Task
+      summary: Return form by a user task key.
```
tmetzke, line 549:
```suggestion
      summary: Return user task form
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -542,6 +542,63 @@ paths:
...
+      description: |
+        Get the form by the user task key.
```
tmetzke, line 551:
```suggestion
        Get the form of a user task.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -542,6 +542,63 @@ paths:
...
+
+        :::note
+        This endpoint will return only linked forms, Embedded forms are not retrieved for this endpoint.
```
tmetzke, line 554:
```suggestion
        This endpoint will only return linked forms. This endpoint does not support embedded forms.
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
```diff
@@ -81,4 +84,26 @@ public ResponseEntity<Object> getByKey(@PathVariable("userTaskKey") final Long u
...
+  @GetMapping(
+      path = "/{userTaskKey}/form",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE})
+  public ResponseEntity<Object> getFormByUserTaskKey(
+      @PathVariable("userTaskKey") final Long userTaskKey) {
+    try {
+      final Long formKey = userTaskServices.getByKey(userTaskKey).formKey();
+
+      if (formKey == null) {
+        return ResponseEntity.noContent().build();
```
tmetzke, line 97: 🔧 Let's also add this response code to the OpenAPI spec so people are aware that this is what happens if the user task has no form.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -542,6 +542,66 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+  /user-tasks/{userTaskKey}/form:
+    get:
+      tags:
+        - User Task
+      summary: Return user task form
```
ralfpuchert, line 549: I think we need a period at the end.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UserTaskQueryController.java
```diff
@@ -29,6 +31,7 @@
 public class UserTaskQueryController {
 
   private final UserTaskServices userTaskServices;
+  @Autowired private FormServices formServices;
```
ralfpuchert, line 34: I would inject the `FormServices` also in the constructor like in the other controllers.

ralfpuchert (approved):
Thank you @nathansandi! I found only 2 minor things. See comments.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
pepopowitz: No, we prefer no period at the end of the endpoint `summary` property. This line meets our recommendations without changes. 

See item 3.iv in https://github.com/camunda/camunda/blob/main/docs/rest-controller.md#openapi-extension for casing recommendations. We don't want periods on the `summary` because it is used in the left navigation of the docs: 

image

pepopowitz (approved):
👍 The spec looks good to me!

PR #22766, opened 2024-09-27
fix: remove base url property

jonathanlukas:
## Description

This PR deprecates the `baseUrl` property for the zeebeClientProperties so that `restAddress` and `grpcAddress` can be used instead.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


ana-vinogradova-camunda (approved):
Thanks, that makes perfect sense. I am just wondering if there should be any changes in documentation/config examples that we provide? I think that not, but it might be worth checking.

PR #22718, opened 2024-09-26
[Backport 8.6.0] fix: stop zeebe task completion variables from being serialized

panagiotisgts:
Backport of https://github.com/camunda/camunda/pull/21764 for release 8.6.0 from main.

PR #22771, opened 2024-09-27
fix: zeebe task completion variable import

panagiotisgts:
Port changes from https://github.com/camunda/camunda/pull/22718
- Change variable serialization on v850 importer
- Fix OS variable mapping


houssain-barouni (approved):
LGTM!

PR #22775, opened 2024-09-27
Disable Identity Execution features for the 8.6 release

remcowesterhoud:
## Description

User creation, authorization patching and default user creation are all part of the 8.7 release. We should disabled them in 8.6 so we can guarantee our state is fresh for the 8.7 release.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


tmetzke (commented):
👍 Changes look good to me.
Can we add a test case for each controller to verify what the response looks like in 8.6?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserProcessors.java
```diff
@@ -32,7 +32,7 @@ public static void addUserProcessors(
         .onCommand(
             ValueType.USER,
             UserIntent.UPDATE,
-            new UserUpdateProcessor(keyGenerator, processingState, writers, distributionBehavior))
-        .withListener(new DefaultUserCreator(processingState, config));
+            new UserUpdateProcessor(keyGenerator, processingState, writers, distributionBehavior));
+    //        .withListener(new DefaultUserCreator(processingState, config));
```
tmetzke, line 36: Good catch, I hadn't thought about that one 🕵️

remcowesterhoud:
@tmetzke I've added the tests, they just return a regular ProblemDetail now 🙂

tmetzke:
> @tmetzke I've added the tests, they just return a regular ProblemDetail now 🙂

Awesome, thanks! I assume our global exception handler catches this now and turns this into a 400. Usually, this would be a [501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501).

remcowesterhoud:
> > @tmetzke I've added the tests, they just return a regular ProblemDetail now 🙂
> 
> Awesome, thanks! I assume our global exception handler catches this now and turns this into a 400. Usually, this would be a [501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501).

That's fair. Do you think we should change this now? Should be relatively easy, but then again we should add that to `main` as well and we don't want to forward port this PR I suppose.

tmetzke (approved):
👍 This let's me sleep better now, thanks! 😄

tmetzke:
> That's fair. Do you think we should change this now? Should be relatively easy, but then again we should add that to `main` as well and we don't want to forward port this PR I suppose.

Nah, good enough in my view.

remcowesterhoud:
OH no merge queue here I see 😅

tmetzke:
YOLO

remcowesterhoud:
If the release branch breaks you know why 👍

PR #22800, opened 2024-09-27
feat: switch to use CamundaUser as principal

### LINKED ISSUE #22728: Existing component security classes use the central user classes
### Description
To support a migration from the existing security modules in Operate and Tasklist to the central module Camunda Authentication, the existing security modules should use the new centralised representations of our user objects i.e. `CamundaUser` for basic auth.

This is important as it will enable us to have a single approach to all authentication handling and processing of the principal on an authenticated request.

```[tasklist]
### Implementation
- [x] `CamundaUser` object is updated to support the user fields required by the existing component security classes (i.e. `c8Links`
- [x] Existing security classes return a `CamundaUser` object when loading the user into Security context
```

maryarm:
## Description

To support a migration from the existing security modules in Operate and Tasklist to the central module Camunda Authentication, the existing security modules should use the new centralised representations of our user objects i.e. CamundaUser for basic auth.

This is important as it will enable us to have a single approach to all authentication handling and processing of the principal on an authenticated request.

## Related issues

closes #22728


operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthUserService.java
```diff
@@ -23,21 +24,21 @@
       + OperateProfileService.SSO_AUTH_PROFILE
       + " & !"
       + OperateProfileService.IDENTITY_AUTH_PROFILE
-      + " & !"
```
marcosgvieira, line 26: @maryarm can you help me to understand this part of removing the profile for Operate? 

We have something similar here on Tasklist: https://github.com/camunda/camunda/blob/fc225e2f15a9d08ad519f4edd26b0b709b0a1154/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserDetailsService.java#L35

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserReader.java
```diff
@@ -32,13 +33,15 @@ public class SearchEngineUserReader implements UserReader {
...
-    if (principal instanceof User) {
-      final User user = (User) principal;
+    if (principal instanceof CamundaUser) {
+      final CamundaUser user = (CamundaUser) principal;
       return Optional.of(
           new UserDTO()
               .setUserId(user.getUserId())
               .setDisplayName(user.getDisplayName())
-              .setPermissions(rolePermissionService.getPermissions(user.getRoles()))
+              .setPermissions(
```
nathansandi, line 42: Question - On this case, what is the difference between permissions and roles?

nathansandi (commented):
Good work here, left some minor questions in order to understand the changes

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthUserService.java
maryarm: @marcosgvieira, I saw in the history that The AUTH_BASIC profile has added to AuthUserService by Ben. The reason is: after we have enabled auth basic in camunda-authentication, principal type is CamundaUser. As now I changed the code to always return CamundaUser as principal no matter which UserDetailService is used, I removed the profile exclusion here.

authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
```diff
@@ -8,25 +8,75 @@
...
+
+  public List<String> getRoles() {
+    return getAuthorities().stream().map(GrantedAuthority::getAuthority).toList();
+  }
+
+  public boolean isCanLogout() {
+    return canLogout;
   }
 
   public static final class CamundaUserBuilder {
```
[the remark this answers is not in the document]
maryarm: Not a very strong reason ;) but I have added the new fields to make operate and tasklist UserDetailService prinicpal, compatible with CamundaUserDetailService. As I think we have to eventually remove these UserDetailServices and CamundaUserDetailService be the only provider, I skipped adding these new fields to builder (that we may remove with new permission and authorizations strategy)

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserReader.java
maryarm: I don't know in detail as I only refactored the code. but with my understanding about permissions' in identity I can guess that permissions are like "write:*" but roles are like "OWNER", "USER" which has a set of permissions assigned to it.

operate/qa/integration-tests/src/test/java/io/camunda/operate/webapp/security/es/ElasticSearchUserDetailsServiceIT.java
```diff
@@ -9,16 +9,18 @@
...
 import io.camunda.operate.util.OperateAbstractIT;
 import io.camunda.operate.util.SearchTestRule;
 import io.camunda.operate.util.TestApplication;
 import io.camunda.operate.util.searchrepository.TestSearchRepository;
 import io.camunda.operate.webapp.security.auth.OperateUserDetailsService;
-import io.camunda.operate.webapp.security.auth.User;
+import io.camunda.operate.webapp.security.auth.Role;
 import java.io.IOException;
 import java.util.Map;
+import org.elasticsearch.core.List;
```
houssain-barouni, line 23: use `java.util.List` and remove the dependency to `org.elasticsearch:elasticsearch-core`

tasklist/qa/integration-tests/src/test/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserDetailsServiceIT.java
```diff
@@ -23,6 +24,7 @@
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import org.elasticsearch.core.List;
```
houssain-barouni, line 27: same thing here

authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
```diff
@@ -8,25 +8,75 @@
...
+    this.displayName = displayName;
+  }
+
+  public CamundaUser(
+      final String displayName,
+      final String username,
+      final String password,
+      final List<String> roles) {
+    super(username, password, roles.stream().map(SimpleGrantedAuthority::new).toList());
+    userKey = null;
```
houssain-barouni, line 42: This constructor is used to create a `CamundaUser` from `UserEntity`. Why not setting the `userKey` from the `UserEntity.key` ?

authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
```diff
@@ -8,25 +8,75 @@
...
   public CamundaUser(
-      final Long userKey, final String name, final String username, final String password) {
+      final Long userKey, final String displayName, final String username, final String password) {
     super(username, password, Collections.emptyList());
     this.userKey = userKey;
-    this.name = name;
+    this.displayName = displayName;
+  }
+
+  public CamundaUser(final String displayName, final String username, final String password) {
```
houssain-barouni, line 30: This constructor is not used anywhere

authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
```diff
@@ -8,25 +8,75 @@
...
+      final String displayName,
+      final String username,
+      final String password,
+      final List<String> roles) {
+    super(username, password, roles.stream().map(SimpleGrantedAuthority::new).toList());
+    userKey = null;
+    this.displayName = displayName;
+  }
+
+  public CamundaUser(
```
houssain-barouni, line 46: Same, this constructor is not used

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthUserService.java
houssain-barouni: so, we need to remove `!AUTH_BASIC` from `SearchEngineUserDetailsService` in Tasklist module as well?

houssain-barouni (changes requested):
Some comments and questions

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthUserService.java
maryarm: no, don't think so, I removed the AUTH_BASIC from AuthUserService in operate not form UserDetailsServices, so it could keep the profile, I have just changed loadByUsername to return CamundaUser.

authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
maryarm: That would be nice if there was a key but this constructor is used in OperateUserDetailsService and SearchEngineUserDetailsService. and in none of them I can find the key in the Entity class. (UserEntity or TasklistEntity).

houssain-barouni: oh my mistake! I thought is was the new UserEntity

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/AuthUserService.java
houssain-barouni: I see, it should be kept for the Components extending `UserDetailsService`

houssain-barouni (approved):
LGTM

PR #22803, opened 2024-09-27
fix: user details clickthrough in Identity

Ben-Sheppard:
## Description

This PR fixes the Identity UI to support the click through of a user due to a change in the `id` field (now `key`)


ralfpuchert (commented):
LGTM so far. Will tests for Controller and Service follow?

Ben-Sheppard:
They will indeed @ralfpuchert - I will add them in shortly

emulbreh:
I fixed this independently in https://github.com/camunda/camunda/pull/24570

PR #22818, opened 2024-09-27
Verify permissions on deploy resource commands

### LINKED ISSUE #22585: Verify permissions in `DeploymentCreateProcessor`
<h2>Description</h2>

⚠️ Note this is the first processor for which permissions will be checked. This means we need to setup an entire permission verification behavior. The default user is always authorized to perform any action.

- **Resource type**: `DEPLOYMENT`
- **Possible permissions**:
    -  `*`

-----------

**Out of scope permissions**:
- `deploymentKey:123456789`

remcowesterhoud:
## Description

Users should only be allowed to deploy a resource if they have a `CREATE` permission for the `DEPLOYMENT` resource.
This also includes an IT to verify the full stack.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22585

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.identity;
+
+import io.camunda.zeebe.auth.impl.Authorization;
+import io.camunda.zeebe.engine.EngineConfiguration;
+import io.camunda.zeebe.engine.state.immutable.AuthorizationState;
+import io.camunda.zeebe.engine.state.immutable.UserState;
+import io.camunda.zeebe.protocol.impl.record.UnifiedRecordValue;
+import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
+import io.camunda.zeebe.protocol.record.value.PermissionType;
+import io.camunda.zeebe.protocol.record.value.UserType;
+import io.camunda.zeebe.stream.api.records.TypedRecord;
+import java.util.List;
+import java.util.Map;
+
+public final class AuthorizationCheckBehavior {
+
+  public static final String DELIMITER = ":";
+  private static final String WILDCARD_PERMISSION = "*";
+  private final AuthorizationState authorizationState;
+  private final UserState userState;
+  private final EngineConfiguration engineConfig;
+
+  public AuthorizationCheckBehavior(
+      final AuthorizationState authorizationState,
+      final UserState userState,
+      final EngineConfiguration engineConfig) {
+    this.authorizationState = authorizationState;
+    this.userState = userState;
+    this.engineConfig = engineConfig;
+  }
+
+  /**
+   * Checks if a user is Authorized to perform an action on a resource. The user key is taken from
+   * the authorizations of the command.
+   *
+   * <p>This method does not take a Map of resource identifiers to check for. The user is considered
+   * authorized if it is the default user, or if it has a wildcard permission for the provided
+   * resource type and permission type. If you want to check for specific resource identifiers, use
+   * {@link #isAuthorized(TypedRecord, AuthorizationResourceType, PermissionType, Map)}
+   *
+   * @param command the command to check authorization for
+   * @param resourceType the type of resource to check authorization for
+   * @param permissionType the type of permission to check authorization for (CRUD)
+   * @return true if the user is authorized, false otherwise
+   */
+  public <T extends UnifiedRecordValue> boolean isAuthorized(
+      final TypedRecord<T> command,
+      final AuthorizationResourceType resourceType,
+      final PermissionType permissionType) {
+    return isAuthorized(command, resourceType, permissionType, Map.of());
+  }
+
+  /**
+   * Checks if a user is Authorized to perform an action on a resource. The user key is taken from *
+   * the authorizations of the command.
+   *
+   * <p>The caller of this method should provide a Map of resource identifiers. Examples of this
+   * are:
+   *
+   * <ul>
+   *   <li>Key: bpmnProcessId, Value: myProcess
+   *   <li>Key: processInstanceKey, Value: 1234567890
+   * </ul>
+   *
+   * @param command the command to check authorization for
+   * @param resourceType the type of resource to check authorization for
+   * @param permissionType the type of permission to check authorization for (CRUD)
+   * @param requiredResourceIdentifiers the resource identifiers to check for
+   * @return true if the user is authorized, false otherwise
+   */
+  public <T extends UnifiedRecordValue> boolean isAuthorized(
+      final TypedRecord<T> command,
+      final AuthorizationResourceType resourceType,
+      final PermissionType permissionType,
+      final Map<String, String> requiredResourceIdentifiers) {
+    final var userKey = (Long) command.getAuthorizations().get(Authorization.AUTHORIZED_USER_KEY);
+    if (userKey == null) {
+      return false;
+    }
+    return isAuthorized(userKey, resourceType, permissionType, requiredResourceIdentifiers);
+  }
+
+  private boolean isAuthorized(
+      final long userKey,
+      final AuthorizationResourceType resourceType,
+      final PermissionType permissionType,
+      final Map<String, String> requiredResourceIdentifiers) {
+
+    // If authorization is disabled, everyone is authorized
+    if (!engineConfig.isEnableAuthorization()) {
+      return true;
+    }
+
+    final var userOptional = userState.getUser(userKey);
+    if (userOptional.isEmpty()) {
+      return false;
+    }
+    final var user = userOptional.get();
+
+    // The default user has all permissions
+    if (user.getUserType().equals(UserType.DEFAULT)) {
+      return true;
+    }
+
+    final var authorizedResourceIdentifiers =
+        getAuthorizedResourceIdentifiers(userKey, resourceType, permissionType);
+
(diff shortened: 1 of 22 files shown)


remcowesterhoud:
@reviwers, reviewing this by commit is likely easiest 😄

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -59,6 +59,8 @@ public class RestErrorMapper {
             yield RestErrorMapper.createProblemDetail(HttpStatus.CONFLICT, message, title);
           case INVALID_ARGUMENT:
             yield RestErrorMapper.createProblemDetail(HttpStatus.BAD_REQUEST, message, title);
+          case UNAUTHORIZED:
+            yield RestErrorMapper.createProblemDetail(HttpStatus.UNAUTHORIZED, message, title);
```
remcowesterhoud, line 63: @tmetzke FYI, I've added unauthorized to this error mapper.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -0,0 +1,149 @@
...
+        .map(
+            resourceId -> {
+              final var splitResourceId = resourceId.split(DELIMITER, 2);
+              return new ResourceIdentifier(splitResourceId[0], splitResourceId[1]);
+            })
+        .anyMatch(
+            resourceId ->
+                requiredResourceIdentifiers
+                    .get(resourceId.resourceName)
+                    .equals(resourceId.resourceValue));
```
remcowesterhoud, line 145: ❌ We must revisit this as we've decided we won't use a delimiter anymore.

remcowesterhoud: I will tackle this on a followup issue to keep this PR comprehensible

remcowesterhoud: https://github.com/camunda/camunda/issues/22858

remcowesterhoud:
@tmetzke sorry for already assigning a PR to you 😅 There is a slight urgency behind it and Ben mentioned to me he would prefer a 2nd Zeebe reviewer.

tmetzke:
> @tmetzke sorry for already assigning a PR to you 😅 There is a slight urgency behind it and Ben mentioned to me he would prefer a 2nd Zeebe reviewer. Happy to talk you through it if it speeds things up.

No worries, will look at it tomorrow morning CEST 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/streamprocessor/AuthorizableDistributionProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * This processor decorates processors where authorization checks are required, taking care of the
+ * authorization checks for them. If the authorization check fails, the processor will reject the
+ * command and write a rejection response back to the client.
+ *
+ * <p>This decorator should be used for processors that do not require a resource from the state to
+ * determine which resourceIds it is permitted for.
+ */
+public final class AuthorizableDistributionProcessor<T extends UnifiedRecordValue>
+    implements DistributedTypedRecordProcessor<T> {
```
tmetzke, line 31: 💭 I am wondering if we need to create such a hierarchical structure for commands. As an alternative, each command could simply call a utility method from an authorization behavior to check for specific required permissions "ad hoc". If the check is negative, another utility method writes the rejection and response and the command is done. I see the benefit of "hiding" this complexity a bit in a wrapping command but this also makes all the commands more complex to understand.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/immutable/UserState.java
```diff
@@ -16,5 +17,13 @@ public interface UserState {
 
   UserRecord getUser(final String username);
 
+  /**
+   * Returns a user by its key. If no user was found, an empty optional is returned.
+   *
+   * @param userKey the key of the user
+   * @return An optional containing the user if it was found, otherwise an empty optional
+   */
+  Optional<UserRecord> getUser(final long userKey);
```
tmetzke, line 26: 👍 I like this better than returning `null`.
🔧 I'd propose to make this class consistent though regarding fetching users and the response you receive, either the object directly or an `Optional`.

tmetzke (commented):
👍 Great work, @remcowesterhoud, I really like the initial idea of authorizing commands.
💭 I've added two alternative ideas that I'm happy to discuss further if needed.
🔧 There is one minor code change I'd propose for better consistency.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/immutable/UserState.java
tmetzke: To be more precise, we fetch by username and potentially return a `null` instead of an empty `Optional`.

remcowesterhoud: Good call 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -0,0 +1,151 @@
...
+    final var userOptional = userState.getUser(userKey);
+    if (userOptional.isEmpty()) {
+      return false;
+    }
+    final var user = userOptional.get();
+
+    // The default user has all permissions
+    if (user.getUserType().equals(UserType.DEFAULT)) {
+      return true;
+    }
```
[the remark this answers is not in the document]
remcowesterhoud: We could, but in the end we want customers to be able to create admin users as well. It seems to me that assigning a set of default permissions is very maintenance heavy, since any new permissions would need to be added here as well. Additionally it may require state migrations to get these permissions itno existing admin users.

I'd be curious to hear what struggles you've encountered with this in C7. For now, let's not let this discussion block this PR from merging so we can keep progressing.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/streamprocessor/AuthorizableDistributionProcessor.java
remcowesterhoud: I think there's multiple advantages to this approach:

1. I love the fact that it's hiding this complexity 😄 Especially with the plans of having more contributors to Zeebe. I think this provides a clear interface for them to work with without having to know the details of how authorization works. They just provide some properties in a method they are forced to implement and everything gets taken care of.
2. What you see in this PR is one of the simpler scenarios we have. In my first POC I called a behavior class directly from the processor. It worked, but imo it made the processor harder to understand as there'd be a bunch of code regarding authorizations in each of the processors. Besides that it would introduce some duplication. I never commited that version of the POC unfortunately 🤦 
3. It makes it possible to write an ArchUnit test. We can check in this test if the processors are implementing an Authorizable interface. If it doesn't the test will fail, unless the processor is explicitly listed as an exception. Why would we want that? Anytime someone adds a new processor this test will fail, forcing them to think about authorizations and preventing processors that require authorizations to be added without authorization checks.
4. If you recall your groups subject in the team event a few weeks ago you remember that we sometimes fetch resources multiple times from the state. With this approach, once we implement processors that actually have a resource to check we will provide this resource immediately to the `processRecord()` method. You can see an example in [the POC here](https://github.com/camunda/camunda/pull/22655/files#diff-ed88b0209fb1a4cc1a1ea77527fccdca919a00b01b76096804343339d51a3a60R55-R56). This hopefully prevents people from fetching resources from the state multiple times.

remcowesterhoud: Mainly point 3 is great imo.

remcowesterhoud: >I see the benefit of "hiding" this complexity a bit in a wrapping command but this also makes all the commands more complex to understand.

Maybe to also add to this just to be sure. We don't wrap the command. We only wrap the command processor

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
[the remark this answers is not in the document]
tmetzke: > We could, but in the end we want customers to be able to create admin users as well
Which you still could if you define a set of permissions internally that represent admin users, or even create a role for that with a set of permission and add admin users to that role.

> It seems to me that assigning a set of default permissions is very maintenance heavy, since any new permissions would need to be added here as well. Additionally it may require state migrations to get these permissions into existing admin/default users.

That is true, yes. It's the biggest downside of this approach, especially when we're in a developing phase where we add new permission types over a longer period of time.

> I'd be curious to hear what struggles you've encountered with this in C7

The biggest issue is that you must consider this specific role additionally, in command processing, Query API, and web apps. Whenever you touch on permission checks, you need to bake in the admin role on top. If it's just a user with a set of permissions, you don't. It makes all permission handling logic more complex by default. We have forgotten that role here and there sometimes when building new features and had to patch the admin role back in later. Contributing a feature to the platform that touches on permissions in any way is definitely more complex for anyone, especially externals. The admin role becomes "hidden knowledge" quite easily that way.

> For now, let's not let this discussion block this PR from merging so we can keep progressing.

I agree 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/streamprocessor/AuthorizableDistributionProcessor.java
tmetzke: Good points, definitely 👍 
Thanks for sharing in that detail, I also took away some nice learnings, point 3 is really nice for sure 🙂
Let's roll with this then and see how it goes. Happy to try that out 👍

tmetzke (approved):

PR #22848, opened 2024-09-30
Schema json files now supports settings

### LINKED ISSUE #22804: Operate schema creation is not working from Camunda exporter
### Description

When working on #22802 I noticed that [ElasticsearchEngineClient#putIndexTemplateRequest](https://github.com/camunda/camunda/blob/6bcbab0ed0e50d3e45a8d1742703622f7d5af629/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java#L195) expects json schema files to contain only template mappings, but some of them contain also settings. This is the reason that e.g. `list-view` is failing to be created currently.

To run exporter with Operate schema creation you need to uncomment descriptors initialization in [DefaultExporterResourceProvider](https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java) (after #22802 is merged) and run StandaloneBroker with following parameters:

```
ZEEBE_BROKER_EXPORTERS_OPERATE_ARGS_ELASTICSEARCH_CREATESCHEMA=true;ZEEBE_BROKER_EXPORTERS_OPERATE_ARGS_ELASTICSEARCH_INDEXPREFIX=newoperate;ZEEBE_BROKER_EXPORTERS_OPERATE_ARGS_ELASTICSEARCH_URL=http://localhost:9200;ZEEBE_BROKER_EXPORTERS_OPERATE_CLASSNAME=io.camunda.exporter.CamundaExporter
```

### TODO

*  fix template creation to be able to read both mappings and settings from json file

EuroLew:
## Description

The json files which contain the contents for index templates now support a setting fields which will be reflected in the created index template. This does not conflict with the default replicas/shards settings for templates.

## Related issues

closes #22804

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
@@ -26,9 +26,11 @@ import io.camunda.exporter.exceptions.ElasticsearchExporterException;
 import io.camunda.exporter.exceptions.IndexSchemaValidationException;
 import io.camunda.webapps.schema.descriptors.IndexDescriptor;
 import io.camunda.webapps.schema.descriptors.IndexTemplateDescriptor;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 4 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/ElasticsearchEngineClient.java
```diff
@@ -258,25 +259,27 @@ private PutIndexTemplateRequest putIndexTemplateRequest(
...
           .name(indexTemplateDescriptor.getTemplateName())
           .indexPatterns(indexTemplateDescriptor.getIndexPattern())
           .template(
               t ->
                   t.aliases(indexTemplateDescriptor.getAlias(), Alias.of(a -> a))
-                      .mappings(
-                          deserializeJson(IndexTemplateSummary._DESERIALIZER, templateMappings)
-                              .mappings())
+                      .mappings(templateFields.mappings())
                       .settings(
```
sdorokhova, line 275: @EuroLew Did I get it right, that we can't simply use `.settings(templateSettings)` here because we need to supplement settings read from file with number of shards and number of replicas?
I'm worried here, that we read settings from file as they are but take only couple of properties from them (analisys and index). Could be error prone in the future if we decide to use other setting "types" also. Do you think we can somehow use the full settings object from the file? 🤔

EuroLew: yes we can't use .settings due to the need to supplement, and yes you are correct it is error prone in the future as it will not pick up other settings types. The solutions are 
- capture every settings type this would be like a dozen+ lines? I could add this just not pretty
- add an endpoint to specifically update the settings of templates, I don't like this solution because this is what PUT index templates is meant to support but that currently derives from the .json file so it would require deserialisation into a json type/map and edit that map then serialise back in an input stream for the request

I

EuroLew: I resolved it by deserialising, adding the desired fields, serialising back, its a costly operate but we won't need to do it often (only on creation) so should be fine.

sdorokhova: I would still try to manipulate Json in some way. Like reading `settings` part to map, putting number of shards/replicas and then somehow convert it to `IndexSettings` object. It will be ugly probably, but should be safe. What do you think?

If not, I'm fine with the 1st solution.

sdorokhova (approved):
I like how it looks at the end!

PR #22849, opened 2024-09-30
Schema manager rolling update tests

### LINKED ISSUE #22758: Schema manager rolling update tests
**Description**

The schema manager must support rolling update scenarios where some exporters are on a new version (having triggered new schema updates) while old exporters still exist and can restart (trigger schema update attempts on the old exporters).

Due to the design of how schema validation works rolling updates should be supported this issue is to write tests to verify such behaviour.

Scenario 1:

In an update one of the brokers will update with new schemas which will change the resources on ELS to match the new schemas, the other brokers will also eventually update and their schema updates will be idempotent as ELS is already in the state of the new schemas. This represents the normal scenario.

Scenario 2:

In a situation where a exporter has updated and the ELS resources have changed to reflect the new schemas, however there are still old exporters running, and one of these old exporter restarts (calling exporter.open which updates the schema). This will pass validation and no update requests are made to the ELS instance (as there are no new fields to add). Operation will continue as normal. This is the expected (and wanted) behaviour and no further action is needed.

EuroLew:
## Description

Verify schema manager supports rolling updates for the following scenarios:

Scenario 1:

In an update one of the brokers will update with new schemas which will change the resources on ELS to match the new schemas, the other brokers will also eventually update and their schema updates will be idempotent as ELS is already in the state of the new schemas. This represents the normal scenario.

Scenario 2:

In a situation where a exporter has updated and the ELS resources have changed to reflect the new schemas, however there are still old exporters running, and one of these old exporter restarts (calling exporter.open which updates the schema). This will pass validation and no update requests are made to the ELS instance (as there are no new fields to add). Operation will continue as normal. This is the expected (and wanted) behaviour and no further action is needed.

## Related issues

closes #22758

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -141,7 +141,7 @@ public class CamundaExporter implements Exporter {
     if (configuration.elasticsearch.getRetention().isEnabled()) {
       searchEngineClient.putIndexLifeCyclePolicy(
           configuration.elasticsearch.getRetention().getPolicyName(),
-          configuration.elasticsearch.getRetention().getMininumAge());
+          configuration.elasticsearch.getRetention().getMinimumAge());
 
       final var lifecycleUpdate =
           Map.of(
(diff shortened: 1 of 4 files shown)


mustafadagher (commented):
The test cases seem fine, however, how the second case `shouldNotErrorIfOldExporterRestartsWhileNewExporterHasAlreadyStarted` works is not completely clear to me.

I've been debugging it, and I feel unsure how it works right now.

Another side note, if you change the config in line 76 to set an index prefix (like below), 4 test cases start to fail. I'm not sure if that's about the test setup, or an issue in the logic.

```java
    config.elasticsearch.setIndexPrefix("camunda-record-");
```

Let's catch up more about it.

mustafadagher (approved):
LGTM! 🚀

PR #22853, opened 2024-09-30
feat: add jwt token claims in to authentication

### LINKED ISSUE #22732: Token claims are loaded into the command authentication
### Description
To be able to support the application of authorizations at a command level the processors need to be able to access the authentication data supplied with the request.

In the instance of OIDC authentication we need to provide the processors with the claims and values on the token such that the assigned mapping rules can be retrieve by Zeebe.

### Acceptance criteria
- [x] The claims and their values on the token on the request are set on the command

maryarm:
## Description

To be able to support the application of authorizations at a command level the processors need to be able to access the authentication data supplied with the request.

In the instance of OIDC authentication we need to provide the processors with the claims and values on the token such that the assigned mapping rules can be retrieve by Zeebe.
## Related issues

closes #22732


zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/RequestMapperTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.jwt.Jwt;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+
+class RequestMapperTest {
```
Ben-Sheppard, line 29: 💯 nice addition here :)

Ben-Sheppard (commented):
Some small comments on my side by largely this seems fine as a first version (until we refactor to remove the crafted JWT :)

Ben-Sheppard (approved):
thank you for addressing my comments, the changes seem fine to me, I wonder if @remcowesterhoud has any thoughts here too as he will need them down in the processors.

For now though I'm going to approve because I think that we can make any necessary tweaks later (if any are needed)

zeebe/auth/src/main/java/io/camunda/zeebe/auth/impl/JwtAuthorizationDecoder.java
```diff
@@ -91,6 +90,13 @@ public Map<String, Object> decode() {
           decodedJWT.getClaim(Authorization.AUTHORIZED_USER_KEY).asLong());
     }
 
+    claimMap.putAll(
+        decodedJWT.getClaims().entrySet().stream()
+            .filter(entry -> entry.getKey().startsWith(USER_TOKEN_CLAIM_PREFIX))
+            .collect(
+                Collectors.toMap(
+                    Map.Entry::getKey, claimEntry -> claimEntry.getValue().as(Object.class))));
+
```
remcowesterhoud, line 99: @maryarm @Ben-Sheppard This is for the Mapping rules cases, right? I feel we're going the wrong way by putting this in the JWT. We talked before about we wanted to get rid of this JWT as it provides no value for Zeebe. By putting everything in this JWT now it's going to be difficult to differentiate when we need to check the authorized user key, or when we need to check mapping rules.

I'd very much prefer these to be different [AuthDataFormats](https://github.com/camunda/camunda/blob/main/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/encoding/AuthInfo.java#L97-L106). One specific for user keys and one specific for user tokens claims.

Ben-Sheppard: Absolutely agree here @remcowesterhoud - the addition of them here is just to keep the tasks moving and have some way to iterate over. The eventual idea (as we spoke about and mentioned [here](https://github.com/camunda/camunda/pull/22853#pullrequestreview-2341212539)) is to not use this JWT.

remcowesterhoud: Do we have an issue to refactor this already?

Ben-Sheppard: I can't see one yet, but I will create an issue for it

PR #22857, opened 2024-09-30
feat: document API methods in the Zeebe Java client

chillleader:
## Description

This PR will add the new commands to interact with the Document API to the Zeebe Java client.

It can be reviewed by commits:

**Part 1: Tweaks to the Document API**

1. [refactor: switch to TTL for link creation](https://github.com/camunda/camunda/pull/22857/commits/4876070fc3dd7ce1148fb8b6200d05e6baccfb5f) - changes the create link command in the document API to accept a TTL for convenience. Also switches to millis for uniformity with other API methods.
2. [refactor: change the document discriminator property name](https://github.com/camunda/camunda/pull/22857/commits/82ddf2f058f8832875c6fe0f67cb315a2e41418b) - changes the discriminator field name in the document reference based on the new agreement with the implementation team
3. [fix: in-memory document store must be a singleton](https://github.com/camunda/camunda/pull/22857/commits/1c87ed6503a35710f2c797f6d7365629f0aa1bea) - fixes the lifecycle of the document store registry to ensure that the in-memory store impl is always a singleton
4. [feat: update dist module](https://github.com/camunda/camunda/pull/22857/commits/6c649dd088d9b9a0c767be1e63551225db150165) - adds the document store module to dist

**Part 2: Zeebe Client changes**

1. [feat: get document content command](https://github.com/camunda/camunda/pull/22857/commits/8cdc20ec8e45913cbbf21047c8bc60ec3dddb7c9) - "Get document" command with the new consumer implementation for non-blocking streaming data reading
2. [feat: create document command](https://github.com/camunda/camunda/pull/22857/commits/f7fc2a1bbf71cd0735267e94e9e60e7d2387da6d) - "Create document" command - multipart post
3. [feat: create document link command](https://github.com/camunda/camunda/pull/22857/commits/9117830aac9855ce9bd698a96557d135bf521bee)
4. [feat: delete document command](https://github.com/camunda/camunda/pull/22857/commits/125c72b290de28000462ed76736ff8db023f3e1f)
5. [feat: update the zeebe client interface](https://github.com/camunda/camunda/pull/22857/commits/cb221e3f06b8c565fdad80f5a7bbadefa4eb1282) - adds all commands to the `ZeebeClient` interface and its implementation.

## Checklist

- [ ] for CI changes:

## Related issues

related: https://github.com/camunda/team-connectors/issues/873


clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateDocumentLinkCommandImpl.java
```diff
@@ -0,0 +1,93 @@
...
+        requestTimeout.toMillis(), java.util.concurrent.TimeUnit.MILLISECONDS);
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<DocumentLinkResponse> send() {
+    final DocumentLinkRequest documentLinkRequest = new DocumentLinkRequest();
+    documentLinkRequest.setTimeToLive(timeToLive.toMillis());
+
+    final Map<String, String> queryParams = new HashMap<>();
```
nicpuppa, line 79: ❌ this is not used

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeleteDocumentCommandImpl.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Override
+  public ZeebeFuture<DeleteDocumentResponse> send() {
+    final Map<String, String> queryParams = new HashMap<>();
+    if (storeId != null) {
+      queryParams.put("storeId", storeId);
+    }
+    final HttpZeebeFuture<DeleteDocumentResponse> result = new HttpZeebeFuture<>();
+    client.delete("/documents", queryParams, requestConfig.build(), result);
```
nicpuppa, line 65: ❓ Are we delete all the documents in the store ? I think you missed the `documentId` in the path

nicpuppa (commented):
Hey @chillleader , great work overall. I had a first review and left some comments/suggestions 👍 

❌ All QA IT tests for the new commands are missing, please have a look

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeleteDocumentCommandImpl.java
chillleader: Good catch :)

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateDocumentLinkCommandImpl.java
chillleader: It should be used, added it to the call below 👍

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeployResourceCommandImpl.java
```diff
@@ -249,6 +250,7 @@ private ZeebeFuture<DeploymentEvent> sendRestRequest() {
     final HttpZeebeFuture<DeploymentEvent> result = new HttpZeebeFuture<>();
     httpClient.postMultipart(
         "/deployments",
+        Collections.emptyMap(),
```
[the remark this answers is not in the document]
chillleader: Added the wrapper, agree that it probably looks cleaner

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4017,7 +4017,7 @@ components:
     DocumentReference:
       type: object
       properties:
-        documentType:
+        camunda.document.type:
```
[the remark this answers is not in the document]
chillleader: > Why this naming format ?

We use this field to detect that a process variable contains a document and apply custom deserialization logic. The strange naming format is chosed deliberately to avoid accidental clashes with potential customer data - e.g. there is a high chance that any business payload contains something like `documentType`, so we shouldn't rely on a discriminator field with such a common name. Dots are meant to represent the "namespace". More details on the doc reference [here](https://docs.google.com/document/d/1fe4nYImFqCqGV1y0MXG4xe4a3U7Nb3cq_1n8OgJTSjk/edit#heading=h.ydnzsxjw8ev1).

> Why we need this enum ? Since is always set to camunda

There is a plan to support another document type `external` in the future - for documents that are not managed by Camunda but can be resolved/accessed by Camunda.

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateDocumentCommandImpl.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  @Override
+  public CreateDocumentCommandStep2 content(final InputStream content) {
+    this.content = content;
+    return this;
+  }
+
+  @Override
+  public CreateDocumentCommandStep2 content(final byte[] content) {
```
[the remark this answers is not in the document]
chillleader: I think the role of the file name in the document creation process is a bit different. For the document object, a file name is just an optional part of the metadata - we are anyway generating an ID for each file and not using the file name as an identifier. We have to add it to the multipart request too, though, because Spring MVC doesn't support working with request parts as files if they have no filename. But we want to allow the users to create documents without a file name.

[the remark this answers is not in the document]
nicpuppa: Thanks for sharing, it make sense. Do we need to close the `InputStream` then ? We can add a try/finally state in the send command and close the stream there

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateDocumentLinkCommandImpl.java
```diff
@@ -44,12 +47,15 @@ public CreateDocumentLinkCommandImpl(
       final String documentId,
       final String storeId,
       final JsonMapper jsonMapper,
-      final HttpClient httpClient) {
+      final HttpClient httpClient,
+      final ZeebeClientConfiguration configuration) {
+    ensureNotNull("documentId", documentId);
```
nicpuppa, line 52: 🔧 I think this check here is not needed, as we instantiate the class in `ZeebeClientImpl`.

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateDocumentCommandImpl.java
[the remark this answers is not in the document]
chillleader: Good point!

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/CircularBufferInputStream.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  public synchronized void signalError(final IOException e) {
+    exception = e;
+    notifyAll();
+  }
+
+  @Override
+  public synchronized int read() throws IOException {
+    while (available == 0) {
```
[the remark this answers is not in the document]
chillleader: Makes sense - I went for a slightly different refactoring option, extracting the duplicating preconditions check and wait. I also optimized the buffer operations to avoid looping through single bytes and use array copying instead.

chillleader:
Regarding `CircularBufferInputStream` - it would have been really nice to avoid synchronization at all, because we have a single producer - single consumer set up. We can't just drop synchronization because the `read` and `write` methods can still be accessed in parallel, and they share the same state. We could potentially use a `LinkedBlockingQueue` for this, but here are some considerations:
- We probably want to avoid storing single bytes in the queue due to boxing overhead and overall inefficiency of single-byte operations
- Therefore we would need to store byte arrays in the queue. For that we'd have to implement a custom capacity check (queue size limit would not work, as it checks for the number of entries, and we need to check the total byte number). This would, again, require some form of synchronization or a custom implementation of the queue.

With that in mind, I don't think it makes sense to go deep into this right now - the current approach should work fine with our use cases.

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/DocumentDataConsumer.java
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * @param <T> the type of the successful response body, always an {@link InputStream}
+ */
+public class DocumentDataConsumer<T>
+    implements AsyncEntityConsumer<ApiEntity<T>>, CapacityCallback {
+
+  private final CircularBufferInputStream inputStream;
+  private FutureCallback<ApiEntity<T>> resultCallback;
+  private final int maxCapacity;
+  private CapacityChannel capacityChannel;
```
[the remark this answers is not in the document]
chillleader: Yes, that's a good idea

clients/java/src/main/java/io/camunda/zeebe/client/impl/response/DocumentMetadataImpl.java
```diff
@@ -17,10 +17,21 @@
...
 public class DocumentMetadataImpl implements DocumentMetadata {
 
+  private static final String EXPIRES_AT =
+      io.camunda.zeebe.client.protocol.rest.DocumentMetadata.JSON_PROPERTY_EXPIRES_AT;
+  private static final String SIZE =
+      io.camunda.zeebe.client.protocol.rest.DocumentMetadata.JSON_PROPERTY_SIZE;
+  private static final String FILE_NAME =
+      io.camunda.zeebe.client.protocol.rest.DocumentMetadata.JSON_PROPERTY_FILE_NAME;
+  private static final String CONTENT_TYPE =
+      io.camunda.zeebe.client.protocol.rest.DocumentMetadata.JSON_PROPERTY_CONTENT_TYPE;
```
nicpuppa, line 34: ❓ Any specific reason for this ? Can we simply use the variables from `DocumentMetadata` ?

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/CircularBufferInputStream.java
```diff
@@ -64,15 +64,27 @@ public synchronized void write(final ByteBuffer data) throws IOException {
...
+    availableData += firstCopyLength;
+
+    final int remaining = dataSize - firstCopyLength;
+    if (remaining > 0) {
+      // Wrap around
+      data.get(buffer, writePos, remaining);
+      writePos = (writePos + remaining) % capacity;
+      availableData += remaining;
     }
+
```
nicpuppa, line 85: 🔧 If I'm not wrong `availableData += firstCopyLength;` and `availableData += remaining;` can be removed in favour of `availableData += dataSize;` just after this if statement

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/CircularBufferInputStream.java
```diff
@@ -123,39 +128,58 @@ public synchronized int read(@Nonnull final byte[] b, final int off, final int l
...
+    // Inform capacity callback
+    if (capacityCallback != null) {
+      capacityCallback.onCapacityAvailable(bytesRead);
+    }
+
+    return bytesRead;
+  }
+
+  private boolean canConsumeData() throws IOException {
```
nicpuppa, line 163: 🔧 This can be refactored as follow if I'm not wrong:

```
private boolean canConsumeData() throws IOException {
    while (availableData == 0 && exception == null && !endOfStream) {
        try {
            wait();
        } catch (final InterruptedException e) {
            Thread.currentThread().interrupt();
            throw new IOException("Interrupted while reading", e);
        }
    }

    if (exception != null) {
        IOException ex = exception;
        exception = null; 
        throw ex;
    }
    return availableData == 0 && endOfStream;
}
```
Adding also `exception = null; ` to avoid repeated exceptions being thrown

nicpuppa:
> Regarding `CircularBufferInputStream` - it would have been really nice to avoid synchronization at all, because we have a single producer - single consumer set up. We can't just drop synchronization because the `read` and `write` methods can still be accessed in parallel, and they share the same state. We could potentially use a `LinkedBlockingQueue` for this, but here are some considerations:
> 
> * We probably want to avoid storing single bytes in the queue due to boxing overhead and overall inefficiency of single-byte operations
> * Therefore we would need to store byte arrays in the queue. For that we'd have to implement a custom capacity check (queue size limit would not work, as it checks for the number of entries, and we need to check the total byte number). This would, again, require some form of synchronization or a custom implementation of the queue.
> 
> With that in mind, I don't think it makes sense to go deep into this right now - the current approach should work fine with our use cases.

Thanks a lot for this deep reflection, I'm in favour of keep it simple at the moment 👍

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/CircularBufferInputStream.java
chillleader: Also the `endOfStream` check in the last line seems redundant as we are already checking it in the first condition

clients/java/src/main/java/io/camunda/zeebe/client/impl/response/DocumentMetadataImpl.java
chillleader: I added this because I encountered problems with how additional properties are handled in the generated classes. 

`DocumentMetadata` is generated from the OpenAPI spec. This class allows arbitrary additional properties, and the generator produces an extension of `HashMap` with some custom fields and getters/setters. When we deserialize the class with Jackson, it always tries to resolve the hash map properties first and all the times I ran this my custom setters returned nulls, and `contentType` and other fields ended up in the HashMap as regular entries.

To make things worse, I didn't have such an issue on the server side because we are using different code generators for the client and for the server (`JavaCodegen` in the client, `SpringCodegen` in the gateway). `SpringCodegen` apparently uses a different generation strategy and doesn't produce a subclass of `HashMap` - so the properties are resolved correctly there.

I think this might be a bug in the `JavaCodegen`, but I'm not sure if we can rely on this behavior so I tried to handle both options - so that it doesn't break if this behavior changes.

This also applies to the ❓ above about `Long` vs `Integer` - hash map entries are deserialized as integer by default.

nicpuppa: Why you need `additionalProperties` ? Can you simply use the properties that you define in the specification ?
It might be a stupid question but I'm missing the reason

chillleader: We want to allow the users to store any arbitrary properties in the metadata, along with a set of pre-defined ones like contentType etc.

chillleader: Maybe to avoid these weird transformations we could deserialize the metadata as a `HashMap` (i.e. remove the dependency on the generated class). Then `DocumentMetadataImpl` could be a simple wrapper on top of the map where methods like `getContentType()` can simply return the value from the map using a predefined key. WDYT?

nicpuppa: In this way users can send every type of value for the predefined key, right ? 
e.g suppose to have a defined key SIZE, users can set for this key a String, when we are going to get this value from the map we should handle all this edge cases, I don't like it tbh.

What if in the spec you define the `DocumentMetadata` with the pre-defined property and then you add a new field `customProperties` that has a reference to a sub-object with only additional properties.

```
{
   size: 1,
   expireAt: "",
   fileName: "filename",
   contentType: "type",
   customProperties: {
      "foo": "bar"
   }
}
```

chillleader: I agree this is a potential flaw and I think this approach could be a solution. We already assumed the previous API design everywhere so I'll check if there are any issues with changing it at this stage. But in general I think it will work 👍

chillleader: Moved all custom field to `customProperties`

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateDocumentCommandImpl.java
```diff
@@ -155,33 +153,39 @@ public CreateDocumentCommandStep2 storeId(final String storeId) {
...
-    metadata.put(DocumentMetadata.JSON_PROPERTY_EXPIRES_AT, expiresAt.toString());
+    metadata.setExpiresAt(expiresAt.toString());
     return this;
   }
 
   @Override
   public CreateDocumentCommandStep2 customMetadata(final String key, final Object value) {
     ensureNotNull("key", key);
-    metadata.put(key, value);
+    if (metadata.getCustomProperties() == null) {
```
nicpuppa, line 176: ❓ Do you think we should check also for null `value` ?

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/CreateDocumentCommandImpl.java
```diff
@@ -155,33 +153,39 @@ public CreateDocumentCommandStep2 storeId(final String storeId) {
...
+      metadata.setCustomProperties(new HashMap<>());
+    }
+    metadata.getCustomProperties().put(key, value);
     return this;
   }
 
   @Override
   public CreateDocumentCommandStep2 customMetadata(final Map<String, Object> customMetadata) {
-    metadata.putAll(customMetadata);
+    if (metadata.getCustomProperties() == null) {
```
nicpuppa, line 185: 🔧 You can check if the provided `customMetadata` is not null, if not simply set it into `customProperties`

```
    ensuroNotNull("customeMetadata", customMetadata)
    metadata.setCustomProperties(customMetadata);
    return this;
```

nicpuppa (approved):
Thanks for all the effort 💪 

I just left a couple of small comments, please have a look 👀

nicpuppa (approved):
The changes look good 👍 

But I think with the latest null check we broke an IT test, please double check 💪

Issue #22832, opened 2024-09-29
Refactor distinction between LogStream and SyncLogStream


npepinpe:
**Description**

With the unified flow control epic, the `LogStream` is now fully synchronous. This means the distinction between `SyncLogStream` (the test implementation of `LogStream`) and `LogStream` is not about async/sync, but rather that the test one has some extra helper functionality for writing tests.

The goal here is to refactor it so the distinction is clearer, and also remove any unnecessary fields/methods which may remain.

> [!Note]
> It's likely you can/should split up the resolution of this issue in multiple, bite-sized PRs.

### Acceptance criteria

- [ ] Remove reference to actor/scheduler in the `LogStream`
  - [ ] `LogStreamBuilder#withActorSchedulingService` method is removed
    - [ ] Calls to the method are removed
    - [ ] All implementation of this interface method are removed
  - [ ] Any dead code, fields, etc., due to the previous removal is also removed (e.g. `LogStreamBuilderImpl#actorSchedulingService` field, and its usage in the validate method)
- [ ] Remove `SyncLogStream#getAsyncLogStream` interface method
  - [ ] Remove implementations of this method
- [ ] Rename `SynchronousLogStreamWriter` to `BlockingLogStreamWriter`
  - [ ] Rename `SyncLogStream#newSyncLogStreamWriter` to `SyncLogStream#newBlockingLogStreamWriter`
- [ ] Rename `SyncLogStream` to `TestLogStream`
  - [ ] Rename related interfaces/classes appropriately (e.g. `SyncLogStreamBuilder` -> `TestLogStreamBuilder`)

PR #22870, opened 2024-09-30
refactor: Remove reference to actor/scheduler in LogStream class

entangled90:
## Description

First part of https://github.com/camunda/camunda/issues/22832. Removed reference to ActorSchedulingService from LogStream & LogStreamBuilder.

ControlledActorClock is removed from class LogStreamRule.

## Related issues
related to #22832

### DIFF
--- a/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/clock/ControlledActorClock.java
+++ b/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/clock/ControlledActorClock.java
@@ -44,13 +44,10 @@ public final class ControlledActorClock implements ActorClock {
 
   public void setCurrentTime(final long currentTime) {
     this.currentTime = currentTime;
-    if (usesPointInTime()) {
-      update();
-    }
   }
 
   public void setCurrentTime(final Instant currentTime) {
-    setCurrentTime(currentTime.toEpochMilli());
+    this.currentTime = currentTime.toEpochMilli();
   }
 
   private boolean usesPointInTime() {
--- a/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/clock/ControlledActorClock.java
+++ b/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/clock/ControlledActorClock.java
@@ -44,10 +44,13 @@ public final class ControlledActorClock implements ActorClock {
 
   public void setCurrentTime(final long currentTime) {
     this.currentTime = currentTime;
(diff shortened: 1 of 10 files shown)


zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/clock/ControlledActorClock.java
npepinpe, line 1: :x: The change itself is not incorrect, but I don't think we need any changes here, as it may also break other places where the behavior is unexpected. It's likely not worth it here, especially because relying on the old `ActorClock` was likely incorrect in the first place once we removed actors from the log stream.

npepinpe (changes requested):
:x: The issue comes with the usage of the `ActorClock` as our `LogStreamBuilder#withClock`, aka the stream clock (see the interface `StreamClock` in `zeebe-stream-platform` for the concept reference). Basically, every stream processor (one per partition) has its own logical clock which is used only there, but does not affect the actual thread pool scheduling. But since the stream processor uses the log stream to write things, and we append the timestamp there, then the log stream has an indirect reference to the clock.

So here, in the failing test, we can just pass our own controllable `InstantSource` as the clock, e.g.

<details><summary>LogStreamWriterTest.java</summary>

```java
public final class LogStreamWriterTest {
  private final ControlledClock clock = new ControlledClock();
  private final LogStreamRule logStreamRule = LogStreamRule.startByDefault(b -> b.withClock(clock));

  // ...

  @Test
  public void shouldWriteEventWithTimestampFromClock() {
    // given
    clock.delegate = () -> Instant.ofEpochMilli(123456789L);
    final long position = tryWrite(TestEntry.ofDefaults());

    // when
    final LoggedEvent event = getWrittenEvent(position);

    // then
    assertThat(event.getTimestamp()).isEqualTo(123456789L);
  }

  // ...

  private static final class ControlledClock implements InstantSource {
    private InstantSource delegate;

    @Override
    public Instant instant() {
      return delegate.instant();
    }
  }
}
```

</details>

> [!Note]
> We likely wouldn't need a mutable `ControlledClock` class here if the test was a junit 5 test, and the `LogStreamRule` an extension, because then we could write nested tests and change the setup. But that's likely time consuming, and I would only propose this if you're familiar with junit 5 and writing extensions, otherwise let's skip it for now.

npepinpe (approved):
Assuming once you've made the suggested changes or similar (and revert the ones in `ControlledActorClock`), and everything is green, we can go ahead and merge :)

Oh, and for the change log tool, in the PR description it should be `related to`, not `relates`

PR #22872, opened 2024-09-30
Opensearch BatchRequest should build batch operations and executes them

### LINKED ISSUE #21839: Opensearch BatchRequest should build batch operations and executes them

mustafadagher:
## Description

Implemented BatchRequest for Opensearch.
Exporter handlers should use it to flush their entities into the Batch request, and the ExporterBatchWriter should use execute it to flush all batched requests

## Related issues

closes #21839

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
@@ -1,0 +1,248 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.store;
+
+import io.camunda.exporter.exceptions.OpensearchExporterException;
+import io.camunda.exporter.exceptions.PersistenceException;
+import io.camunda.exporter.utils.OpensearchScriptBuilder;
+import io.camunda.webapps.schema.entities.ExporterEntity;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.client.opensearch._types.OpenSearchException;
+import org.opensearch.client.opensearch._types.Refresh;
+import org.opensearch.client.opensearch.core.BulkRequest;
+import org.opensearch.client.opensearch.core.BulkResponse;
+import org.opensearch.client.opensearch.core.bulk.BulkResponseItem;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@SuppressWarnings("rawtypes")
+public class OpensearchBatchRequest implements BatchRequest {
+  public static final int UPDATE_RETRY_COUNT = 3;
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(OpensearchBatchRequest.class);
+  private final OpenSearchClient osClient;
+  private final BulkRequest.Builder bulkRequestBuilder;
+  private final OpensearchScriptBuilder scriptBuilder;
+
+  public OpensearchBatchRequest(
+      final OpenSearchClient osClient,
+      final BulkRequest.Builder bulkRequestBuilder,
+      final OpensearchScriptBuilder scriptBuilder) {
+    this.osClient = osClient;
+    this.bulkRequestBuilder = bulkRequestBuilder;
+    this.scriptBuilder = scriptBuilder;
+  }
+
+  @Override
+  public BatchRequest add(final String index, final ExporterEntity entity) {
+    return addWithId(index, entity.getId(), entity);
+  }
+
+  @Override
+  public BatchRequest addWithId(final String index, final String id, final ExporterEntity entity) {
+    LOGGER.debug("Add index request for index {} id {} and entity {} ", index, id, entity);
+    bulkRequestBuilder.operations(op -> op.index(idx -> idx.index(index).id(id).document(entity)));
+    return this;
+  }
+
+  @Override
+  public BatchRequest addWithRouting(
+      final String index, final ExporterEntity entity, final String routing) {
+    LOGGER.debug(
+        "Add index request with routing {} for index {} and entity {} ", routing, index, entity);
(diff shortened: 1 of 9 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
```diff
@@ -0,0 +1,248 @@
...
+      final OpenSearchClient osClient,
+      final BulkRequest.Builder bulkRequestBuilder,
+      final OpensearchScriptBuilder scriptBuilder) {
+    this.osClient = osClient;
+    this.bulkRequestBuilder = bulkRequestBuilder;
+    this.scriptBuilder = scriptBuilder;
+  }
+
+  @Override
+  public BatchRequest add(final String index, final ExporterEntity entity) {
```
EuroLew, line 45: I don't know how I feel about having this add when `addWithId` exists, ideally there should be as little interface methods as possible. I would prefer `add` to be removed and `addWithId` become `add`.

To be honest I would prefer you have a single `add` function and you pass in some object you built with a builder 
e.g.
`var BatchItem item = new BatchItem.Builder().index(...).id(...).entity(...).routing(...)`
and then you can have a single `add(BatchItem item)` where you can do a single `bulkRequestBuilder.operations`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
```diff
@@ -0,0 +1,248 @@
...
+      final String index,
+      final String id,
+      final ExporterEntity entity,
+      final String script,
+      final Map<String, Object> parameters) {
+    return upsertWithScriptAndRouting(index, id, entity, script, parameters, null);
+  }
+
+  @Override
+  public BatchRequest upsertWithScriptAndRouting(
```
EuroLew, line 116: Similiar feedback to add, I feel like you could have a single `upsert` and it takes an example `BatchItem` (or other name) with all the fields you need (instantiated with a builder), and you could condense all the upserts .

This constant drilling through with nulls just makes me think a builder would be better suited.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
```diff
@@ -0,0 +1,248 @@
...
+        op ->
+            op.update(
+                upd ->
+                    upd.index(index).id(id).document(entity).retryOnConflict(UPDATE_RETRY_COUNT)));
+
+    return this;
+  }
+
+  @Override
+  public BatchRequest updateWithScript(
```
EuroLew, line 178: I feel like you could also condense the update to one method call similar to the above approaches.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
```diff
@@ -0,0 +1,248 @@
...
+    return this;
+  }
+
+  @Override
+  public BatchRequest upsertWithScript(
+      final String index,
+      final String id,
+      final ExporterEntity entity,
+      final String script,
+      final Map<String, Object> parameters) {
```
EuroLew, line 111: Also since `parameters` is required with the script maybe have a check that both are not null? Instead of getting a less descriptive error message

EuroLew (commented):
if the builder changes for requests do happen it will also require refactoring of the interface and the elasticsearch batch request, but you should be able to reuse any new classes as the code looks the same to me.

mustafadagher:
> if the builder changes for requests do happen it will also require refactoring of the interface and the elasticsearch batch request, but you should be able to reuse any new classes as the code looks the same to me.

I don't think I fully understand this comment, could you maybe rephrase?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
mustafadagher: so about this interface in general, I didn't invent it, it was there, and I'm not completely aware of any historic reasons it was build this way for.

I agree with your sentiment in general, but I also don't want to change too many things at once. 

Also, in the bulk operations builders, they have `index`, `update`, `create` and so on, and those should be separated.
You could say the passed object could also contain an operation, yet then I have to keep a list of valid operation, decide upon them and use different methods for different operations. which might in the end make the interface less cluttered, yet the code itself might be more complex.

again, I agree with your sentiment, and I think your idea is worth considering, but I'd say let's see more use cases when more handlers and entities are ported here first and then decide for that

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
mustafadagher: I'm not sure if parameters is required to be honest. but I can double check

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
EuroLew: ok sure we can wait and see if this changes in the future

EuroLew:
> > if the builder changes for requests do happen it will also require refactoring of the interface and the elasticsearch batch request, but you should be able to reuse any new classes as the code looks the same to me.
> 
> I don't think I fully understand this comment, could you maybe rephrase?

yea sorry, if we change the upsert/update methods to instead take an object that we instantiate using a builder this will require a change in interface and you need to do the same refactoring for the elasticsearch batch request, but since we're leaving it for now just a thought.

EuroLew:
if you just add a parameters not null check and maybe any other checks required (as I feel it would result in more accurate error messages) then LGTM.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/OpensearchBatchRequest.java
mustafadagher: it would indeed in the script builder.

mustafadagher:
> if you just add a parameters not null check and maybe any other checks required (as I feel it would result in more accurate error messages) then LGTM.

I'll add a `requireNonNull` for the parameters.

EuroLew (approved):

PR #22914, opened 2024-10-01
feat: add process definition search service, controller

ralfpuchert:
## Description

* Refactor from `service` to `search` package structure
* Rename fields according to naming convention
* Client implementation: https://github.com/camunda/camunda/pull/22919

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

Related to #22643

PR #22919, opened 2024-10-01
feat: add process definition search to client

ralfpuchert:
## Description

* Add process definition search to client
* Based on https://github.com/camunda/camunda/pull/22914 (service, controller)

## Related issues

closes #


clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/ProcessDefinitionFilter.java
```diff
@@ -0,0 +1,77 @@
...
+
+public interface ProcessDefinitionFilter extends SearchRequestFilter {
+
+  /**
+   * Filters process definitions by the specified process definition key.
+   *
+   * @param processDefinitionKey the key of the process definition
+   * @return the updated filter
+   */
+  ProcessDefinitionFilter processDefinitionKey(final Long processDefinitionKey);
```
houssain-barouni, line 28: Minor: can we use a primitive long ? that avoids to handle the null case

clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/ProcessDefinitionFilter.java
```diff
@@ -0,0 +1,77 @@
...
+   */
+  ProcessDefinitionFilter resourceName(final String resourceName);
+
+  /**
+   * Filters process definitions by the specified version.
+   *
+   * @param version the version of the process definition
+   * @return the updated filter
+   */
+  ProcessDefinitionFilter version(final Integer version);
```
houssain-barouni, line 52: same here, I think it is better to use a primitive int. It indicates to the user that null is not a valid value here.

clients/java/src/main/java/io/camunda/zeebe/client/api/search/response/ProcessDefinition.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.zeebe.client.api.search.response;
+
+public interface ProcessDefinition {
```
houssain-barouni, line 18: This looks very similar to `io.camunda.zeebe.client.api.response.Process` that is used in the command clients. Should we use 2 different datamodels?

houssain-barouni (approved):
LGTM overall,
I had some questions , these questions apply to other client implementations as well.
Also there is a TODO [here](https://github.com/camunda/camunda/blob/8c04534ab1d0912150b332f2f28311dfa6058dda/qa/integration-tests/src/test/java/io/camunda/it/client/ProcessQueryTest.java#L497) that was expecting the changes in your pull request, so maybe we can fix it here.

nathansandi (approved):
Besides the minor suggestions from @houssain-barouni  which I am agree 
LGTM, good job 👍

clients/java/src/main/java/io/camunda/zeebe/client/api/search/response/ProcessDefinition.java
ralfpuchert: Good catch! Will be handled in https://github.com/camunda/camunda/issues/23120

PR #22948, opened 2024-10-01
refactor: remove getAsyncLogStream from SynchronousLogStream and rena…

### LINKED ISSUE #22832: Refactor distinction between LogStream and SyncLogStream

entangled90:
## Description

refactor: remove getAsyncLogStream from SynchronousLogStream and rename SyncLogStream

Method getAsyncLogStream is removed from SynchronousLogStream.
Class SyncLogStream renamed to TestLogStream and delegate previously unsupported methods to field logStream.
Interface [SynchronousLogStream.java](https://github.com/camunda/camunda/pull/22948/files#diff-852ff4f97f701b4e29dd38d7735e022d1e92c46920e7b0162e879018468283fe) has been removed as it only had one implementation and only used in tests.

closes #22832


zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/ProcessingExporterTransistor.java
```diff
@@ -45,8 +45,7 @@ public void onRecovered(final ReadonlyStreamProcessorContext context) {
         () -> {
           final int partitionId = context.getPartitionId();
           typedEvent = new TypedRecordImpl(partitionId);
-          final var asyncLogStream = synchronousLogStream.getAsyncLogStream();
-          asyncLogStream.registerRecordAvailableListener(this::onNewEventCommitted);
+          synchronousLogStream.registerRecordAvailableListener(this::onNewEventCommitted);
```
npepinpe, line 48: :wrench: Since we renamed the class too, might as well rename the field/variable.

```suggestion
          logStream.registerRecordAvailableListener(this::onNewEventCommitted);
```

zeebe/logstreams/src/test/java/io/camunda/zeebe/logstreams/util/SynchronousLogStream.java
npepinpe, line 1: :thought_balloon: I guess my issue was not clear enough :smile: 

:x: Please rename the interface as well. So I guess we go for `TestLogStream`, and `TestLogStreamImpl` for the implementation.

npepinpe: To be honest, I would question why we need the interface in the first place for a test utility :thinking:

:question: Are there more implementations of `SynchronousLogStream`? Because if not, then I would propose maybe we just merge it with the implementation. I don't expect we would create more implementations of this interface honestly.

:x: And one last thing, could you also update the Javadocs? It still refers to `Synchronous methods` and the actor model.

npepinpe (approved):
:rocket: 

Please address all comments before merging, but unless you want to challenge them, I don't think we need another review pass here :+1:

zeebe/logstreams/src/test/java/io/camunda/zeebe/logstreams/util/SynchronousLogStream.java
entangled90: Yes, it only has one implementation, therefore I merged it with the implementation.

PR #22994, opened 2024-10-02
Extended tests for camunda exporter

### LINKED ISSUE #22854: Extended tests for camunda exporter
Extend existing unit and integration test to verify basic functionality of `CamundExporter` independent of the business logic of the webapps.

ElasticSearch
 - [x] Verify that basic authentication using configured username + password works
 - [x] Verify that record is exported when elastic is not initially reachable, but is reachable later.
 
General exporter behavior
 - [x] Verify periodic flush based on the configuration
 - [x] Verify exporter position is updated after flushing

EuroLew:
## Description
Also refactored camunda tests into nested blocks for a logical grouping
ElasticSearch
 - [x] Verify that basic authentication using configured username + password works
 - [x] Verify that record is exported when elastic is not initially reachable, but is reachable later.
 
General exporter behavior
 - [x] Verify periodic flush based on the configuration
 - [x] Verify exporter position is updated after flushing

## Related issues

closes #22854

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
@@ -20,6 +20,10 @@ import static org.mockito.Mockito.when;
 
 import co.elastic.clients.elasticsearch.ElasticsearchClient;
 import co.elastic.clients.elasticsearch.core.GetResponse;
+import com.github.dockerjava.api.model.ExposedPort;
+import com.github.dockerjava.api.model.HostConfig;
+import com.github.dockerjava.api.model.PortBinding;
+import com.github.dockerjava.api.model.Ports;
 import io.camunda.exporter.config.ElasticsearchExporterConfiguration;
 import io.camunda.exporter.entities.AuthorizationEntity;
 import io.camunda.exporter.entities.UserEntity;
@@ -493,6 +497,7 @@ final class CamundaExporterIT {
       exporter.open(controller);
 
       // when
+      final var currentPort = CONTAINER.getFirstMappedPort();
       CONTAINER.stop();
       Awaitility.await().until(() -> !CONTAINER.isRunning());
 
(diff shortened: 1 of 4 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -105,6 +114,45 @@ void beforeEach() throws IOException {
...
+        new ExporterTestContext()
+            .setConfiguration(new ExporterTestConfiguration<>("elastic", config));
+
+    exporter.configure(context);
+
+    // when
+    final var spiedController = spy(controller);
+    exporter.open(spiedController);
+
+    spiedController.runScheduledTasks(Duration.ofSeconds(duration));
```
mustafadagher, line 135: ❓ I'm curious what do we need this call for here?

mustafadagher (changes requested):
Thanks for the extra tests @EuroLew, I like it 👍  Please, have a look at the extra suggestions below. I also left one question at one of the new tests.

🔧 Should we add a test to verify that `export()` only flushes when we reach the configured `configuration.bulk.size` ?

🔧 Should we add a test to verify that exporter should flush on `close()`?

💭 The last commit included a removal of some test(s) which made me confused what was added and what was removed and had to double check the previous `restructure` commit. I'd have amended  this removal to the previous restructure commit and left the last one to just add the new tests. That's just thinking out loud in retrospect, no need to change anything for now (unless you personally like to)

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
EuroLew: We don't use a real `controller` which submits jobs to the `actor` instead its a test controller where we need to manually call for the schduled tasks to run

EuroLew:
> 🔧 Should we add a test to verify that `export()` only flushes when we reach the configured `configuration.bulk.size` ?

Given the lack of dependency injection as resources are created in `open` to take advantage of retry (and just how exporters work natively), it's hard to verify that the `writer` does specific actions when we reach the bulk size, the best we can do is verify that the last exported record position is the nth record exported where n = bulk size. This would technically test the above, however it depends on implementation of how record position is updated.

I could `@VisibleForTesting` block so we can inject the writer so we can verify calls but to be honest I think `@VisibleForTesting` is a code smell for new code, it should be used for legacy code where we can't work around it. 

My thoughts are either add a VisibleForTesting to inject the writer or no test, thoughts?

mustafadagher:
> This would technically test the above, however it depends on implementation of how record position is updated.

💭  Calling `updateLastExportedRecordPosition` after guaranteeing a record is flushed is a documented behavior of `Exporter.export(record)` in the interface. Check [Exporter.java#L61-L63](https://github.com/camunda/camunda/blob/fa3919f106b653f02dec43365450679719edca4f/zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/Exporter.java#L61-L63)

❓ Would that still be testing implementation details in your opinion? (to for example configure the exporter to only flush after `x` record, and verify the position is only updated after the record number `x` is exported?)

I'll approve the PR, and leave that up to your judgement to decide if it's a `good to have`

mustafadagher (approved):

PR #23012, opened 2024-10-02
move es/os specific queries

houssain-barouni:
## Description

This pull request aims to introduce a unified [SearchClients](https://github.com/camunda/camunda/blob/607b3dbcf911be81f3c404d04788c153708d659f/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java) class for both ES and OS search clients. The primary objective is to eliminate code duplication between ES and OS, particularly for any additional logic that we need to add in the implemented search clients, like the addition of authorization checks...

In the `search-client` module, we now expose only the interfaces for the different entity search clients. The service module will depend solely on this module, with the implementation of [SearchClients](https://github.com/camunda/camunda/blob/607b3dbcf911be81f3c404d04788c153708d659f/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java) being injected into various services that require any of the entity search clients.

Kudos to @romansmirnov for this contribution :) 

## Checklist

- [ ] for CI changes:

## Related issues

closes #


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java
houssain-barouni, line 1: @romansmirnov I had to move the packages under `search-client-query-transformer` to avoid package + class name collisions with `search-domain` module. For example, they both had `io.camunda.search.query.SearchClientBuilders` classes

romansmirnov (approved):
Thanks for taking care of that topic!! I tested it locally, it works. LGTM!

PR #23031, opened 2024-10-03
feat: Allow configuring timeouts in cluster topology gossipper

### LINKED ISSUE #15793: Allow configuring cluster topology gossipper.

entangled90:
## Description
ClusterConfigurationGossiperConfig can be configured in both broker & gateway profiles. The current hardcoded configuration is used as default in case it's not configured.

A new namespace in the configuration has been created under `[zeebe.broker|zeebe.gateway].cluster.dynamicConfig`. At the moment it contains only one field `gossip`, but it can be used if we need to extend the configuration.

The configuration is validated using the following criteria:
- syncDelay must be strictly positive
- syncRequestTimeout must be strictly positive
- gossipFanout must be greater or equal than 2
- if enableSync is false, syncDelay & syncRequestTimeout are not validated

❓ Should enableSync be removed? If yes, should we remove directly in this PR?

❓ When the broker is started in standalone mode, the gateway will not start a gossiping the dynamic config as it's already done by the broker, hence I did not add a specific config in the template under `zeebe.broker.gateway.cluster.dynamicConfig`.
Is this correct?

## Checklist
- [X] Issue created: camunda/camunda-docs#4395

## Related issues

closes #15793

### DIFF
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -617,6 +617,24 @@
         # This setting can also be overridden using the environment variable ZEEBE_BROKER_CLUSTER_MEMBERSHIP_SYNCINTERVAL
         # syncInterval: 10s
 
+      # Configure the parameters used to propagate the dynamic cluster configuration across brokers and gateways
+      # configManager:
+        # Configure the parameters used for gossiping the dynamic cluster configuration.
+        # Note that this is not related to cluster membership gossip.
+        # gossip:
+          # Enable the synchronization of the gossip instead of only relying on gossip
+          # This setting can also be overridden using the environment variable ZEEBE_GATEWAY_CLUSTER_CONFIGMANAGER_GOSSIP_ENABLESYNC
+          # enableSync: true
+          # Sets the interval between two synchronization requests to other members of the cluster
+          # This setting can also be overridden using the environment variable ZEEBE_BROKER_CLUSTER_CONFIGMANAGER_GOSSIP_SYNCDELAY
+          # syncDelay: 10s
+          # Sets the timeout for the synchronization requests
+          # This setting can also be overridden using the environment variable ZEEBE_BROKER_CLUSTER_CONFIGMANAGER_GOSSIP_SYNCREQUESTTIMEOUT
+          # syncRequestTimeout: 2s
+          # Sets the number of cluster members the configuration is gossiped to.
+          # This setting can also be overridden using the environment variable ZEEBE_BROKER_CLUSTER_CONFIGMANAGER_GOSSIP_GOSSIPFANOUT
+          # gossipFanout: 2
+
       # Configure compression algorithm for all message sent between the brokers and between the broker and
       # the gateway. Available options are NONE, GZIP and SNAPPY.
       # This feature is useful when the network latency between the brokers is very high (for example when the brokers are deployed in different data centers).
--- a/dist/src/main/config/broker.yaml.template
+++ b/dist/src/main/config/broker.yaml.template
@@ -495,6 +495,24 @@
         # This setting can also be overridden using the environment variable ZEEBE_BROKER_CLUSTER_MEMBERSHIP_SYNCINTERVAL
         # syncInterval: 10s
 
+      # Configure the parameters used to propagate the dynamic cluster configuration across brokers and gateways
+      # configManager:
+        # Configure the parameters used for gossiping the configuration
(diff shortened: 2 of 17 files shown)


dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -617,6 +617,23 @@
         # This setting can also be overridden using the environment variable ZEEBE_BROKER_CLUSTER_MEMBERSHIP_SYNCINTERVAL
         # syncInterval: 10s
 
+      # Configure the parameters used to propagate the dynamic cluster configuration across brokers and gateways
+      # dynamicConfig:
+        # Configure the parameters used for gossiping the configuration
+        # gossip:
```
deepthidevaki, line 623: :wrench: May be we could add a note to clarify that this is not the same as membership gossip?

deepthidevaki (commented):
> ❓ Should enableSync be removed? If yes, should we remove directly in this PR?

This can be removed in my opinion. It is ok to do it in a follow up PR to limit the scope of this one. 

> ❓ When the broker is started in standalone mode, the gateway will not start a gossiping the dynamic config as it's already done by the broker, hence I did not add a specific config in the template under zeebe.broker.gateway.cluster.dynamicConfig.
Is this correct?

This is correct.

:thought_balloon: I was wondering if `dynamicConfig` is a reasonable name for the new configuration. I think it is not meaningful for the user. May be something like `configManager` or `configManagment` is better?

dist/src/main/config/broker.standalone.yaml.template
entangled90: Ok I'll modify it to: 
` # Configure the parameters used for gossiping the dynamic cluster configuration. Note that this is not related to cluster membership gossip`

entangled90:
> > ❓ Should enableSync be removed? If yes, should we remove directly in this PR?
> 
> This can be removed in my opinion. It is ok to do it in a follow up PR to limit the scope of this one.
> 
> > ❓ When the broker is started in standalone mode, the gateway will not start a gossiping the dynamic config as it's already done by the broker, hence I did not add a specific config in the template under zeebe.broker.gateway.cluster.dynamicConfig.
> > Is this correct?
> 
> This is correct.
> 
> 💭 I was wondering if `dynamicConfig` is a reasonable name for the new configuration. I think it is not meaningful for the user. May be something like `configManager` or `configManagment` is better?

Regarding the naming, I add a quick discussion with @npepinpe. I'm happy to change it to whatever you prefer. I see that in the code is usually referred as `dynamicClusterConfig`, so that could be another name in the list.

npepinpe:
I didn't even think of `configManager` :smile: 

This was our discussion:

> options:
> - zeebe.broker.cluster.dynamicConfig...
> - zeebe.broker.cluster.dynamicConfig.gossip...
> - zeebe.broker.cluster.gossip...
> - zeebe.broker.dynamicConfig.gossip...
> - zeebe.broker.dynamicConfig...
>
> :mega-thinking:

(all proposed by me, but you already know I'm [bad at naming](https://martinfowler.com/bliki/TwoHardThings.html) :smile:)

I'm fine with configManager or configManagement. The latter is a bit long, but :shrug:

entangled90:
is `configManager` ok for everyone?

deepthidevaki:
> is `configManager` ok for everyone?

:+1: I'm not fully happy with that. But I think it is better than `dynamicConfig`. Let's go with it now.

deepthidevaki (approved):
Great :rocket: Please rename the environment variables in the config templates before merging.

dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -617,6 +617,24 @@
         # This setting can also be overridden using the environment variable ZEEBE_BROKER_CLUSTER_MEMBERSHIP_SYNCINTERVAL
         # syncInterval: 10s
 
+      # Configure the parameters used to propagate the dynamic cluster configuration across brokers and gateways
+      # configManager:
+        # Configure the parameters used for gossiping the dynamic cluster configuration.
+        # Note that this is not related to cluster membership gossip.
+        # gossip:
+          # Enable the synchronization of the gossip instead of only relying on gossip
+          # This setting can also be overridden using the environment variable ZEEBE_GATEWAY_CLUSTER_DYNAMICCONFIG_GOSSIP_ENABLESYNC
```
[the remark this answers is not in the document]
entangled90: Good catch, thanks!

PR #23035, opened 2024-10-03
feat: support user update in camunda services

### LINKED ISSUE #21677: Update a user using the userKey instead of the username
<h2>Description</h2>

At this moment the UserUpdateProcessor uses a username to find a user in the state and update it. We should change this to work with keys, which are the identifiers used in Zeebe.

1. We need add a userKey to the UserRecord
2. We need get the PersistedUser from the UserState using the userKey, instead of using the username

### LINKED ISSUE #22569: Distribute User.UPDATE command in the Identity distribution queue
<h2>Description</h2>

With the implementation of #21447 we should make sure we distribute all identity related commands in the identity queue. This guarantees the other partitions will process these commands in the same order.

In the `UserUpdateProcessor` change the distribution from `unordered()` to `inQueue(DitributionQueue.IDENTITY.getQueueId())`

Ben-Sheppard:
## Description

This PR refactors some of the code related to updating a user specifically:
- Use the user key when fetching a user instead of the username
- Distribute the user update command onto the Identity queue

And also adds in:
- A service method to trigger an update user command

I have not added tests at this stage because the existing ITs require the API to be in place with the client for ease.

## Related issues

closes https://github.com/camunda/camunda/issues/21677, closes https://github.com/camunda/camunda/issues/22569

### DIFF
--- a/service/src/main/java/io/camunda/service/UserServices.java
+++ b/service/src/main/java/io/camunda/service/UserServices.java
@@ -15,6 +15,7 @@ import io.camunda.search.security.auth.Authentication;
 import io.camunda.service.search.core.SearchQueryService;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
 import io.camunda.zeebe.gateway.impl.broker.request.BrokerUserCreateRequest;
+import io.camunda.zeebe.gateway.impl.broker.request.BrokerUserUpdateRequest;
 import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
 import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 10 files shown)


remcowesterhoud (commented):
Changes look good, but there's a bunch of failing tests. Let me know if you need help with those.

It'd be good to also add a test to verify the queue id like we do in the [AddPermissionAuthorizationMultiPartitionTest](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AddPermissionAuthorizationMultiPartitionTest.java#L111-L159)

remcowesterhoud (approved):
Thanks @Ben-Sheppard! I'm going to assume the actual rest endpoint gets added in a later PR?

Ben-Sheppard:
Correct @remcowesterhoud - I didn't want to add too much to this PR but the API endpoint is in progress locally :)

Ben-Sheppard:
Added here @remcowesterhoud https://github.com/camunda/camunda/issues/23093

PR #23036, opened 2024-10-03
Verify permissions on process instance creation

### LINKED ISSUE #22587: Verify permissions in `ProcessInstanceCreationCreateProcessor`
<h2>Description</h2>

⚠️ Note this is the first processor for which permissions will be checked on a resource level (eg. by `bpmnProcessId`). This means we need to setup our permission verification behaviour to support this.

- **Possible permissions**:
    -  `*`
    - `bpmnProcessId:Foo`

remcowesterhoud:
## Description

Adds permission verification on process instance create (with result) commands and adds an IT to verify it works as expected. There's also some refactorings in this PR to extract some code duplication which we had in combination with the Deployment create authorization checks.

Merge blocked by #23034 

## Checklist

- [ ] for CI changes:

## Related issues

closes #22587

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
@@ -16,7 +16,9 @@ import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
 import io.camunda.zeebe.protocol.record.value.PermissionType;
 import io.camunda.zeebe.protocol.record.value.UserType;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
+import java.util.HashSet;
 import java.util.Set;
+import java.util.Optional;
 
 public final class AuthorizationCheckBehavior {
 
@@ -49,7 +51,7 @@ public final class AuthorizationCheckBehavior {
    * @return true if the user is authorized, false otherwise
    */
   public <T extends UnifiedRecordValue> boolean isAuthorized(
-      final TypedRecord<T> command, final AuthorizationRequest authorizationRequest) {
+      final TypedRecord<T> command, final AuthorizationRequest<?> authorizationRequest) {
     return isAuthorized(command, authorizationRequest, Set.of());
   }
 
@@ -73,7 +75,7 @@ public final class AuthorizationCheckBehavior {
    */
   public <T extends UnifiedRecordValue> boolean isAuthorized(
       final TypedRecord<T> command,
-      final AuthorizationRequest authorizationRequest,
+      final AuthorizationRequest<?> authorizationRequest,
       final Set<String> requiredResourceIdentifiers) {
     if (!engineConfig.isEnableAuthorization()) {
       return true;
@@ -88,7 +90,7 @@ public final class AuthorizationCheckBehavior {
 
   private boolean isAuthorized(
       final long userKey,
-      final AuthorizationRequest authorizationRequest,
+      final AuthorizationRequest<?> authorizationRequest,
       final Set<String> requiredResourceIdentifiers) {
     final var userOptional = userState.getUser(userKey);
     if (userOptional.isEmpty()) {
@@ -103,7 +105,9 @@ public final class AuthorizationCheckBehavior {
 
     final var authorizedResourceIdentifiers =
         getAuthorizedResourceIdentifiers(
-            userKey, authorizationRequest.resourceType(), authorizationRequest.permissionType());
+            userKey,
+            authorizationRequest.getResourceType(),
+            authorizationRequest.getPermissionType());
 
     // Check if authorizations contain a resource identifier that matches the required resource
     // identifiers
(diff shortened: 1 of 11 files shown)


remcowesterhoud:
@tmetzke This one is very similar to the previous PR you reviewed for me. I just need some special support for processors implement the `CommandProcessor` interface .... I hate that interface

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCreationCreateProcessor.java
```diff
@@ -97,17 +102,30 @@ public ProcessInstanceCreationCreateProcessor(
...
 
   @Override
-  public boolean onCommand(
+  public Either<Rejection, AuthorizationRequest<DeployedProcess>> getAuthorizationRequest(
       final TypedRecord<ProcessInstanceCreationRecord> command,
       final CommandControl<ProcessInstanceCreationRecord> controller) {
+    return getProcess(command.getValue())
+        .map(
+            process ->
+                new AuthorizationRequest<DeployedProcess>(
```
tmetzke, line 111: 💭 Adding the specific target resource as a generic to the authorization request might limit us very soon. Currently, we only work with what is called ["atomic checks" in C7](https://github.com/camunda/camunda-bpm-platform/blob/master/engine/src/main/java/org/camunda/bpm/engine/impl/db/PermissionCheckBuilder.java#L34). For those, a resource can be determined and is required so we can map it against the permissions defined internally. However, from experience, I'm relatively certain we'll want to allow "combined" permission checks soon as well, e.g. 
```
CREATE on any PROCESS_INSTANCE
OR
CREATE_INSTANCE on PROCESS_DEFINITION with bpmnProcessId=xyz
```
What is the resource generic of the `AuthorizationRequest` in those cases?
I see that it is appealing to fetch the target resource already in the `getAuthorizationRequest` and hand it over to the `onCommand` method but my gut feeling is that this will not hold very long. It looks like a different concern that we're patching onto the authorization request. At least that is what it looks like to me 🙂

That being said, it might still all work well and we're saving on fetching data multiple times. But it feels like a bandaid somehow 😄 And I believe the problem stems from the wrapper class, where we split up the processing logic now and thus need to fetch the target resource twice (once for getting the auth request, once for executing the command). That's why I suggested previously to keep those things within the original `onCommand` method, fetch the resource once, check the authorization using a behavior, and execute the logic if all is fine.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/streamprocessor/AuthorizableCommandProcessor.java
```diff
@@ -0,0 +1,117 @@
...
+import io.camunda.zeebe.stream.api.records.TypedRecord;
+import io.camunda.zeebe.util.Either;
+
+/**
+ * This processor decorates processors where authorization checks are required, taking care of the
+ * authorization checks for them. If the authorization check fails, the processor will reject the
+ * command and write a rejection response back to the client.
+ *
+ * <p>This decorator should be used for processors that do not require a resource from the state to
+ * determine which resourceIds it is permitted for.
```
tmetzke, line 29: ❓ Is that really true? Because the process instance create processor actually fetches it from the state.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -122,13 +95,50 @@ private Set<String> getAuthorizedResourceIdentifiers(
...
+    private final PermissionType permissionType;
+    private Resource resource;
+    private final Set<String> resourceIds;
+
+    public AuthorizationRequest(
+        final AuthorizationResourceType resourceType, final PermissionType permissionType) {
+      this.resourceType = resourceType;
+      this.permissionType = permissionType;
+      resourceIds = new HashSet<>();
+      resourceIds.add(WILDCARD_PERMISSION);
```
tmetzke, line 115: 👍 Adding this automatically makes a lot more sense, yes. We're always checking for this then when checking existing permissions 💡

tmetzke (commented):
👍 Really nice refactorings of the authorization check behavior and the ITs.
💭 I added some thoughts on the generic of the auth request and tried to think ahead a bit to see if this will stay valid. I cannot completely judge if it's simpler like this still or will cause issues. But the generic feels a bit "wrong" in that place somehow.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/streamprocessor/AuthorizableCommandProcessor.java
remcowesterhoud: No it's not. Good catch, I missed that during the refactoring 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCreationCreateProcessor.java
remcowesterhoud: I get what you're saying. When I [checked previously with PM](https://docs.google.com/document/d/1ikxw9ARCv-MBwASnj7WhW5uHCndCRk2B35xi-cB2N0M/edit?disco=AAABTovC3AY) it was mentioned that they don't see this as a requirements for 8.8 and don't have any feature requests that require this.

This isn't something that's unchangeable. We don't break backwards compatibility so if the request comes in the future we can solve it then 🤷 

I don't see it as a big problem personally. If you need multiple resources nothing is stopping you from adding a record `Resources(ResourceA resourceA, ResourceB resourceB)` and passing that object along.

tmetzke: I agree, we can still change this later if we a) discover it's hindering us more than it helps and b) need to handle the target resource(s) differently.

The generic on the distributed command wrapper is still weird though, I have to say 😄 I am quite certain that this will lead to some confusion in the future as to why this is needed.

remcowesterhoud: Yeah I understand why you'd think that. If you think it's more clear we could also remove the generic and add a field to the classes to prevent fetching data multiple times? But my concern is that this will be even more confusing 😄

tmetzke: Well, you could split up the commands further into `Resource getTargetResource()` (or whatever), `getAuthorizationRequest(Resource)`, and `onCommand(..., Resource)` (or whatever the method is depending on the type).

tmetzke: Meh, yeah, right 😅

tmetzke (approved):
👍 Implementation looks good to me 🚀 
💭 We can follow up on the target resource handling later.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCreationCreateProcessor.java
tmetzke: Let's keep it as is for now since it's working. I'm still not entirely sure whether the wrapper is more harm or help. Let's see where this is going.

PR #23039, opened 2024-10-03
NOT TO BE MERGED: Java Client PoC for Task Listeners support

### LINKED ISSUE #22398: Prepare a POC for User Task Listener dedicated client
Prepare a PoC of TLs support in Java Client for the below listed Option 1 and Option 2:

**Option 1.** 
[PoC code is available here](https://github.com/camunda/camunda/pull/23691)
A Task Listener job is treated as a regular job.
The activated job has a getUserTaskData() method that provides a typed object UserTaskJobData with the known user task attributes. This will only be populated for Task Listener jobs and will be empty or null for other jobs.
Below is the example of how this approach could look like:

```
zeebeClient
   .newWorker()
   .jobType("listener-1")
   .handler((client, job) -> {
       LOGGER.info("Assignee for the job: {}", job.getUserTaskData().getAssignee());

       client.newCompleteCommand(job.getKey()).send();
       LOGGER.info("Element with id:'{}' processing completed.", job.getElementId());
   })
   .open();
```

**Option 2.** 
[PoC for this approach can be found here](https://github.com/camunda/camunda/pull/23039)
Java Client provides a dedicated Task Listener handler, that is different from a job handler.
The handler would accept a callback that would provide the client and Task Listener job for the execution. User Task properties will be exposed by Task Listener job.
We might want to consider the event type to be optional to allow for for assigning the same job worker for all Task Listener events.
Below is the example of how this approach could look like:

```
zeebeClient
   .newWorker()
   .taskListener()
   .ofTaskListenerEventType(COMPLETE)
   .jobType("listener-1")
   .handler((client, taskListenerJob) -> {
        LOGGER.info("Assignee for the job: {}", taskListenerJob.getAssignee());

        client.newCompleteCommand(taskListenerJob.getJob().getKey()).send();
        LOGGER.info("Element with id:'{}' processing completed.", taskListenerJob.getJob().getElementId());
    })
    .open();
```

**Option 3.** A Task Listener is different from a job worker.
This will provide independent configuration options from job workers. This option would also allow to do different things specific to user task listeners.
Below is the example of how this approach could look like:

```
zeebeClient
   .newUserTaskListener("listener-1")
   .handler((client, taskListenerJob) -> {
      LOGGER.info("Assignee for the job: {}", taskListenerJob.getAssignee());

      client.newCompleteCommand(taskListenerJob.getJob().getKey()).send();
      LOGGER.info("Element with id:'{}' processing completed.", taskListenerJob.getJob().getElementId());
    })
   .open();
```

ana-vinogradova-camunda:
## Description

This is a PoC for Task Listeners support in Java Client based on `complete` Task Listener implementation provided by @ce-dmelnych 

**Current approach:**

- Extend existing Job Worker Builder Step1 interface to accept two types of handlers:
```
JobWorkerBuilderStep3 handler(JobHandler handler);
UserTaskListenerJobWorkerBuilderStep3 handler(UserTaskListenerJobHandler handler); // for User Task Listener jobs
```
- Introduce UserTaskListenerJobWorkerBuilderStep3 to set an `eventType` (mandatory for User Task Listener jobs)
```
 interface UserTaskListenerJobWorkerBuilderStep3 {
    JobWorkerBuilderStep3 eventType(String eventType);
  }
```
- Use custom headers to access Task User properties and expose these to the Client
- Add interface `UserTaskListenerJob` that the Client will receive as part of a Task Listener callback. It contains properties that are specific for User Task and a reference to a usual Job
- Opening a JobWorker will check the type of a handler provided and execute custom logic if `userTaskListenerJobHandler` is provided.

**This approach was validated from a test application with a local deployment:**
```
        final BpmnModelInstance process =
                Bpmn.createExecutableProcess("processId_Ana")
                        .startEvent()
                        .userTask(
                                "my_user_task",
                                t -> t.zeebeUserTask().zeebeAssignee("Ana"))
                        .zeebeTaskListener(l -> l.complete().type("listener-1"))
                        .endEvent()
                        .done();
```

```
 JobWorker tlWorker = zeebeClient
                .newWorker()
                .jobType("listener-1")
                .handler(((tlJob) -> {
                    // Log the fields
                    LOGGER.info("User Task Key: {}", tlJob.getUserTaskKey());
                    LOGGER.info("Assignee for the job: {}", tlJob.getAssignee());

                    zeebeClient.newCompleteCommand(tlJob.getJob().getKey()).send();
                    LOGGER.info("Element with id:'{}' processing completed.", tlJob.getJob().getElementId());
                }))
                .eventType("COMPLETE")
                .open();
```
Resulted in:
```
2024-10-03 11:53:14 INFO Connected to Camunda 8.
2024-10-03 11:53:28 INFO User Task Key: 2251799813685389
2024-10-03 11:53:28 INFO Assignee for the job: demo
2024-10-03 11:53:28 INFO Element with id:'my_user_task' processing completed.
```

closes: #22398


nicpuppa:
I had a quick look, and my first question is:

❓ Why we are re-using jobWorkerBuilder ? At a first look it seems that we can create another ad-hoc builder to Task Listeners like JobWorkerWithTaskListener or similar name

ana-vinogradova-camunda:
> I had a quick look, and my first question is:
> 
> ❓ Why we are re-using jobWorkerBuilder ? At a first look it seems that we can create another ad-hoc builder to Task Listeners like JobWorkerWithTaskListener or similar name

That was the initial approach, but that introduced quite a lot of code duplication.

nicpuppa:
> That was the initial approach, but that introduced quite a lot of code duplication.

An new interface and the implementation are needed right ? You can also extract the common parts into a common abstract class

tmetzke (commented):
👍 Thanks for providing this E2E, from gateway to client. This helps a lot to discuss on a practical level ⭐ 

Some thoughts from my end here:
* Re-using the job custom headers is fine, it introduces less infrastructure work for us which is good. We need to be mindful of at least the following things though:
  * The custom headers can theoretically collide with user-defined custom headers. We need to document the ones we reserve internally so users have a chance to know which ones not to use.
  * We also need to transport the `customHeaders` of the user task itself in the job custom headers, e.g. by serializing them to JSON in the gateway and deserializing them to a Map in the client. This is where custom headers become a small liability but I think it's good enough still.
  * Alternative: make the user task data a dedicated object property in the proto and OpenAPI spec
    * The attributes of that object are the user task attributes.
    * You can access typed data in the gateway and client without manual transformations.
    * The respective framework and protocol handle the serialization.
    * For most jobs (all but TL jobs), the user task data is empty.
* The client UX is not ideal yet regarding the following:
  * You define a handler in a lambda style without really "knowing" which case you are handling (TL jobs or anything else). This leaves some uncertainties and makes this prone to misunderstandings and errors for users.
  * You define a job type and an event type while a job type should be sufficient for working on a job. We identify jobs by job type, the event type check should not be necessary. As a user, I might want to assign the same job worker for all TL events (to put the event data into a local Kafka queue or whatever). With the event type check, I cannot easily do this. I have to create a worker for each event type (I can reuse the handler implementation but still have multiple workers). I think we don't have to enforce this.
* I can see the following access patterns that are consistent and simple to understand from a user perspective:
  1. A TL job is a regular job. I define a job type, nothing more. The activated job has a `getUserTaskData()` method that provides a typed object `UserTaskJobData` (or similar) with the known user task attributes. This can be empty, of course. Users need to check that there is user task data.
  2. A TL gets a dedicated handler, different from a job handler. I can call `JobWorkerBuilder::userTaskListener(UserTaskListenerJobHandler)`. It does what you have already implemented, providing a `UserTaskListenerJob` object containing user task data. We can decide how we handle users invoking `handler` and `userTaskListener` on a job worker builder, either throwing an exception or silently overriding the other (we need to clearly document this in the API and public docs). With this, it would be OK to check for the job being of a task listener type, I guess, since you explicitly want to handle one. Only task listener jobs will be passed through to be handled.
  3. A TL is a dedicated listener, different from a job worker. I can call `zeebeClient.newUserTaskListener(type).handler(UserTaskListenerHandler)`. Internally, it might just work like a job worker. It would allow to do different things though specific to user task listeners, e.g. handle that the user wants to adjust user task data in the handler that needs to be fed back to the gateway somehow. With this, it would be OK to check for the job being of a task listener type, I guess, since you explicitly want to handle one. Only task listener jobs will be passed through to be handled.

There might be more viable options here in the client of course. I believe that we should either treat all jobs the same (option i) or provide a clear separation for TL handlers (options ii and iii).

PR #23074, opened 2024-10-04
fix: Throw IllegalArgumentException when non-existing exporters are configured

### LINKED ISSUE #12678: Unhelpful error message when setting configuration for non-existing exporter
**Describe the bug**

When I set configuration for an exporter that does not exist, for example because I make a typo in the exporter name, Zeebe fails to start and logs a very unhelpful error message that gives me no indication on what went wrong or how to fix it.

**To Reproduce**

``` shell
$ docker run --rm -e ZEEBE_BROKER_EXPORTERS_OOPS_ARGS_INDEX_NUMBEROFREPLICAS=1 camunda/zeebe:latest
...
java.lang.IllegalStateException: Failed to execute CommandLineRunner
        at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:772) [spring-boot-3.0.5.jar:3.0.5]
        at org.springframework.boot.SpringApplication.callRunners(SpringApplication.java:753) [spring-boot-3.0.5.jar:3.0.5]
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:317) [spring-boot-3.0.5.jar:3.0.5]
        at io.camunda.zeebe.broker.StandaloneBroker.main(StandaloneBroker.java:82) [camunda-zeebe-8.2.4.jar:8.2.4]
Caused by: java.lang.NullPointerException
        at java.util.concurrent.ConcurrentHashMap.putVal(Unknown Source) ~[?:?]
        at java.util.concurrent.ConcurrentHashMap.putIfAbsent(Unknown Source) ~[?:?]
        at java.lang.ClassLoader.getClassLoadingLock(Unknown Source) ~[?:?]
        at jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(Unknown Source) ~[?:?]
        at jdk.internal.loader.BuiltinClassLoader.loadClass(Unknown Source) ~[?:?]
        at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(Unknown Source) ~[?:?]
        at java.lang.ClassLoader.loadClass(Unknown Source) ~[?:?]
        at io.camunda.zeebe.broker.exporter.repo.ExporterRepository.load(ExporterRepository.java:79) ~[zeebe-broker-8.2.4.jar:8.2.4]
        at io.camunda.zeebe.broker.Broker.buildExporterRepository(Broker.java:147) ~[zeebe-broker-8.2.4.jar:8.2.4]
        at io.camunda.zeebe.broker.Broker.<init>(Broker.java:70) ~[zeebe-broker-8.2.4.jar:8.2.4]
        at io.camunda.zeebe.broker.Broker.<init>(Broker.java:49) ~[zeebe-broker-8.2.4.jar:8.2.4]
        at io.camunda.zeebe.broker.StandaloneBroker.run(StandaloneBroker.java:91) ~[camunda-zeebe-8.2.4.jar:8.2.4]
        at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:769) ~[spring-boot-3.0.5.jar:3.0.5]
        ... 3 more
...
```

**Expected behavior**

This should result in an easy to understand error message or be ignored silently.

filipecampos:
Validates configured exporters and throws exception in case there are any invalid exporters (i.e. without a defined `className`)

## Description

We now verify if the configured exporters exist during the configuration validation step in `SystemContext`.

## Related issues

closes #12678

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/SystemContext.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/SystemContext.java
@@ -21,6 +21,7 @@ import io.camunda.zeebe.broker.system.configuration.ClusterCfg;
 import io.camunda.zeebe.broker.system.configuration.DataCfg;
 import io.camunda.zeebe.broker.system.configuration.DiskCfg.FreeSpaceCfg;
 import io.camunda.zeebe.broker.system.configuration.ExperimentalCfg;
+import io.camunda.zeebe.broker.system.configuration.ExporterCfg;
 import io.camunda.zeebe.broker.system.configuration.SecurityCfg;
 import io.camunda.zeebe.broker.system.configuration.backup.AzureBackupStoreConfig;
 import io.camunda.zeebe.broker.system.configuration.backup.BackupStoreCfg;
@@ -38,6 +39,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import org.slf4j.Logger;
 
@@ -110,9 +112,26 @@ public final class SystemContext {
 
     validateExperimentalConfigs(cluster, brokerCfg.getExperimental());
 
+    validateExporters(brokerCfg.getExporters());
+
     final var security = brokerCfg.getNetwork().getSecurity();
     if (security.isEnabled()) {
       validateNetworkSecurityConfig(security);
+    }
+  }
+
+  private void validateExporters(final Map<String, ExporterCfg> exporters) {
+    final Set<Entry<String, ExporterCfg>> entries = exporters.entrySet();
+    final var badExportersNames =
+        entries.stream()
+            .filter(entry -> entry.getValue().getClassName() == null)
+            .map(Entry::getKey)
+            .toList();
+
+    if (!badExportersNames.isEmpty()) {
+      throw new IllegalArgumentException(
+          "Expected to find a 'className' configured for the exporter. Couldn't find a valid one for the following exporters "
+              + badExportersNames);
     }
   }
 
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/system/SystemContextTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/system/SystemContextTest.java
@@ -14,16 +14,21 @@ import static org.mockito.Mockito.mock;
 import io.atomix.cluster.AtomixCluster;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
 import io.camunda.zeebe.broker.system.configuration.BrokerCfg;
+import io.camunda.zeebe.broker.system.configuration.ExporterCfg;
 import io.camunda.zeebe.broker.system.configuration.backup.BackupStoreCfg.BackupStoreType;
 import io.camunda.zeebe.broker.system.configuration.partitioning.FixedPartitionCfg;
 import io.camunda.zeebe.broker.system.configuration.partitioning.FixedPartitionCfg.NodeCfg;
 import io.camunda.zeebe.broker.system.configuration.partitioning.Scheme;
 import io.camunda.zeebe.scheduler.ActorScheduler;
+import io.camunda.zeebe.test.util.junit.RegressionTest;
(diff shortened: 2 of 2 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/SystemContext.java
```diff
@@ -110,12 +114,32 @@ private void validateConfiguration() {
...
+                    entry.getValue().getClassName() == null
+                        || entry.getValue().getJarPath() == null);
+    final Optional<Entry<String, ExporterCfg>> optExporter = badExporters.findFirst();
+
+    if (optExporter.isPresent()) {
+      final Entry<String, ExporterCfg> exporter =
+          optExporter.get(); // TODO: change to get more exporters
+      throw new IllegalArgumentException(
+          "Couldn't find a valid exporter for " + exporter.getKey() + "!");
+    }
```
deepthidevaki, line 140: :x: It would be good to get all misconfigured exporters in the error message. Please remove the TODO as well once it is done.

:wrench: The error message can be more descriptive - "Expected to find a 'className' or 'jarPath" configured for the exproter. But couldn't find a valid one for the following exporters []".

deepthidevaki (changes requested):
Thanks. I like this approach where the configuration is validated before hand instead of handling the NPE where it occurs. 

:x: Please see my comment. Also update the PR title to be concise and update the PR description to follow the format.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/SystemContext.java
filipecampos: Thanks for the suggestions!

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/SystemContext.java
```diff
@@ -110,12 +113,31 @@ private void validateConfiguration() {
...
   }
 
+  private void validateExporters(final Map<String, ExporterCfg> exporters) {
+    final Set<Entry<String, ExporterCfg>> entries = exporters.entrySet();
+    final Stream<Entry<String, ExporterCfg>> badExporters =
+        entries.stream()
+            .filter(
+                entry ->
+                    entry.getValue().getClassName() == null
+                        || entry.getValue().getJarPath() == null);
```
deepthidevaki, line 131: :question: I missed this in the previous review. Is `jarPath` mandatory configuration? It can be null, right?

filipecampos: By reading the exporters section in `broker.yaml.template` and `broker.standalone.yaml.template` I would say so by looking at the elasticsearch and opensearch examples that only state the `className`. I'll remove the references to `jarPath`

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/SystemContext.java
```diff
@@ -110,12 +113,27 @@ private void validateConfiguration() {
...
     if (security.isEnabled()) {
       validateNetworkSecurityConfig(security);
     }
   }
 
+  private void validateExporters(final Map<String, ExporterCfg> exporters) {
+    final Set<Entry<String, ExporterCfg>> entries = exporters.entrySet();
+    final Stream<Entry<String, ExporterCfg>> badExporters =
+        entries.stream().filter(entry -> entry.getValue().getClassName() == null);
+    final var badExportersNames = badExporters.map(Entry::getKey).sorted().toList();
```
deepthidevaki, line 128: :wrench: can be combined.
```suggestion
     final var badExportersNames = exporters.entrySet().stream().filter(entry -> entry.getValue().getClassName() == null).map(Entry::getKey).sorted().toList();
```

Also, as I discussed before, there is no need to sort here.

deepthidevaki (commented):
:rocket:

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/SystemContext.java
filipecampos: Thanks. I've applied your suggestion and also removed the sorts.

deepthidevaki (approved):
Great :rocket:  I thought I already approved this yesterday :)

PR #23080, opened 2024-10-04
feat: provide custom method security

### LINKED ISSUE #22731: Camunda Authentication module provides security method implementations using the new authorization implementation
### Description
To support V1 API endpoint authorizations, the Camunda Authentication module should provide an implementation of the security methods that accept a resource type and permission to check. The implementation can then handle the logic between how the authorizations are checked (OIDC vs Basic Auth).

```[tasklist]
### Implementation
- [ ] Add a class that implements MethodSecurityExpressionOperations to handle v1 security expressions
- [ ] Configure spring to use this class to resolve security expression
- [ ] Right unit tests to check the logic with new Authorizations structure
```

maryarm:
## Description

To support V1 API endpoint authorizations, the Camunda Authentication module should provide an implementation of the security methods that accept a resource type and permission to check. The implementation can then handle the logic between how the authorizations are checked (OIDC vs Basic Auth).

## Related issues

closes #22731


service/src/main/java/io/camunda/service/AuthorizationServices.java
```diff
@@ -46,6 +50,29 @@ public SearchQueryResult<AuthorizationEntity> search(final AuthorizationQuery qu
     return authorizationSearchClient.searchAuthorizations(query, authentication);
   }
 
+  public Set<String> fetchAssignedPermissions(
+      final String ownerType,
```
Ben-Sheppard, line 54: 🔧 NABD but I believe as `ownerId` is an entirely unique field we can make this method simpler by removing the `ownerType`. I think this works in all instances, for example if I am using basic-auth then the authentication object contains my user key which is the value stored on the authorization as the owner.

service/src/main/java/io/camunda/service/AuthorizationServices.java
```diff
@@ -46,6 +50,29 @@ public SearchQueryResult<AuthorizationEntity> search(final AuthorizationQuery qu
...
+      final String resourceId) {
+    final SearchQueryResult<AuthorizationEntity> result =
+        search(
+            SearchQueryBuilders.authorizationSearchQuery(
+                fn ->
+                    fn.filter(
+                            f ->
+                                f.resourceType(resourceType.name())
+                                    .resourceKey(
+                                        StringUtils.isNoneEmpty(resourceId) ? resourceId : null)
```
Ben-Sheppard, line 66: ❓ instead of adding a dependency here would something like this also have the same outcome that you're after (checking for empty)
```suggestion
                                        resourceId.isEmpty() ? null : resourceId)
```

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
```diff
@@ -34,6 +38,12 @@ public class WebSecurityConfig {
       new String[] {"/login**", "/logout**", "/error**", "/actuator**"};
   private static final Logger LOG = LoggerFactory.getLogger(WebSecurityConfig.class);
 
+  @Bean
+  public MethodSecurityExpressionHandler methodSecurityExpressionHandler(
```
Ben-Sheppard, line 42: 🔧 When I tried to start the code locally with the profiles `operate, identity, broker, auth-basic` Spring complained of multiple beans, with this change we should make sure that the old classes are also either updated to be set on profiles or maybe this one is primary.

It may also be that with this change and for 8.7 the existing method security classes can be removed actually, maybe someone from one of the component teams or API team could confirm (@tmetzke do you have thoughts?)

authentication/src/main/java/io/camunda/authentication/handler/CustomMethodSecurityExpressionRoot.java
```diff
@@ -0,0 +1,111 @@
...
+  public boolean hasReadAccess(final String resourceType, final String entity) {
+    return hasPermissions(resourceType, entity, READ_ACCESS_AUTHORITIES);
+  }
+
+  public boolean hasWriteAccess(final String resourceType) {
+    return hasWriteAccess(resourceType, "*");
+  }
+
+  public boolean hasWriteAccess(final String resourceType, final String entity) {
+    return hasPermissions(resourceType, entity, WRITE_ACCESS_AUTHORITIES);
```
Ben-Sheppard, line 55: ❓ with this approach I would need all permissions that are considered write (CREATE, UPDATE, DELETE) to have access (based on the `.containsAll` check [here](https://github.com/camunda/camunda/pull/23080/files#diff-0880de59b39250a94703e490fd9e900610f62bf91daca18442e1351b90c73093R69)) is that correct or would it be _any_ of those permissions and I have access? In an ideal world its on an endpoint basis so I'm targetting a specific permission but wanted to check

Ben-Sheppard (commented):
Some early comments but in general this seems as expected :)

authentication/src/main/java/io/camunda/authentication/handler/CustomMethodSecurityExpressionRoot.java
maryarm: It's also one of my confusion point, to be strict write means all this permission because we are not sure if it's going to be a delete or create. I think in reality we will use hasPermissions directly and pass the required permissions to the method.

Ben-Sheppard: I agree, based on a kick off I attended yesterday we will be more specific with the permission, for now I think I would remove these combined methods and just provide the `hasPermission` method. The others can always be added later on if we need them

Ben-Sheppard (approved):
As spoken about on Slack, before 8.7 we'd want this Method Security class to be the single one (instead of Operate/Tasklist) but that can change as part of the work to migrate the V1 endpoints to the new authroization structure.

Otherwise we can get this merged and make progress. I think in the future we'd want to refactor the combined methods i.e. write and read but thats not something to block this PR IMO

PR #23090, opened 2024-10-06
Metrics for Camunda exporter

### LINKED ISSUE #22867: Add metrics for CamundaExporter
- [x] Identify relevant metrics that helps us with performance tuning and debugging 
- [x] Add the metrics to the exporter
- [x] Make the new metrics available in grafana dashboard

EuroLew:
## Description

Micrometer metrics for Camunda exporter, currently tracks
- Failed flushes
- Flush duration
- Flush latency (time from first record in batch to flush command)
- Bulk size (number of entities)
- Bulk size memory

Currently there is no implementation of bulk size memory as the exporter cannot currently track the memory usage of records currently as not implemented.

Screenshots of grafana metrics at https://github.com/camunda/camunda/issues/22867#issuecomment-2397375347

**Success Criteria:**

- [x] Identify relevant metrics that helps us with performance tuning and debugging 
- [x] Add the metrics to the exporter
- [x] Make the new metrics available in grafana dashboard

## Related issues

closes #22867

### DIFF
--- a/monitor/grafana/zeebe.json
+++ b/monitor/grafana/zeebe.json
@@ -18217,6 +18217,551 @@
     },
     {
       "collapsed": true,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 16
+      },
+      "id": 615,
+      "panels": [
+        {
+          "datasource": {
+            "default": false,
+            "type": "prometheus",
+            "uid": "$DS_PROMETHEUS"
+          },
+          "cards": {},
+          "color": {
+            "cardColor": "#ef843c",
+            "colorScale": "sqrt",
+            "colorScheme": "interpolateSpectral",
+            "exponent": 0.5,
+            "mode": "spectrum"
+          },
+          "dataFormat": "tsbuckets",
+          "fieldConfig": {
+            "defaults": {
+              "custom": {
+                "scaleDistribution": {
+                  "type": "linear"
+                },
+                "hideFrom": {
+                  "tooltip": false,
+                  "viz": false,
+                  "legend": false
+                }
+              }
+            },
+            "overrides": []
+          },
+          "gridPos": {
+            "h": 10,
+            "w": 12,
+            "x": 0,
+            "y": 17
+          },
+          "heatmap": {},
+          "hideZeroBuckets": true,
+          "highlightCards": true,
+          "id": 616,
+          "legend": {
+            "show": false
+          },
+          "options": {
+            "calculate": false,
+            "yAxis": {
+              "axisPlacement": "left",
+              "reverse": false,
+              "unit": "dtdurations"
+            },
+            "rowsFrame": {
+              "layout": "auto"
+            },
+            "color": {
+              "mode": "scheme",
+              "fill": "#ef843c",
+              "scale": "exponential",
+              "exponent": 0.5,
+              "scheme": "Spectral",
+              "steps": 128,
+              "reverse": false
+            },
+            "cellGap": 2,
+            "filterValues": {
+              "le": 1e-9
+            },
+            "tooltip": {
+              "mode": "single",
+              "yHistogram": false,
+              "showColorScale": false
+            },
+            "legend": {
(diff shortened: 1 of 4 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -53,6 +55,15 @@ public class CamundaExporter implements Exporter {
...
   private final ExporterResourceProvider provider;
+  private CamundaExporterMetrics metrics;
+
+  /**
+   * Sample to measure the flush latency of the current bulk request.
+   *
+   * <p>Time of how long an export bulk request is open and collects new records before flushing,
+   * meaning latency until the next flush is done.
+   */
+  private Sample flushLatencyMeasurement;
```
mustafadagher, line 66: 💭   Could we keep track of this instance inside the `CamundaExporterMetrics `?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -108,10 +120,21 @@ public void close() {
...
+    if (writer.getBatchSize() == 0) {
+      flushLatencyMeasurement = metrics.startFlushLatencyMeasurement();
+    }
+
     writer.addRecord(record);
     lastPosition = record.getPosition();
+
     if (shouldFlush()) {
-      flush();
+      try (final var ignored = metrics.measureFlushDuration()) {
```
mustafadagher, line 131: 👍 this looked weird at first look, until I went through the micrometer codebase to find that it returns an `AutoCloseable` and only records the duration in the `close()` method

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/metrics/CamundaExporterMetrics.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public void recordBulkSize(final int bulkSize) {
+    DistributionSummary.builder(meterName("bulk.size"))
+        .description("Exporter bulk size")
+        .serviceLevelObjectives(10, 100, 1_000, 10_000, 100_000)
+        .register(meterRegistry)
+        .record(bulkSize);
+  }
+
+  public void recordBulkMemorySize(final int bulkMemorySize) {
```
mustafadagher, line 52: 💭  We are not using this right no (since we actually don't do much yet with the memory size). It doesn't harm to still keep the method until then, so I think it fine

monitor/grafana/zeebe.json
mustafadagher, line 1: 💭 I couldn't review the content of this file for good or bad, but I trust you have tested it to work.

mustafadagher (approved):

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
EuroLew: Yes if we do not ever want to track more than 1 instance of flush latency at a time, do you think this will ever occur?

mustafadagher: I'm not sure I completely understand your point. so let me rephrase

We have one instance of `CamundaExporterMetrics` per instance of `exporter`, right? and basically `flushLatencyMeasurement` is necessarily one per exporter instance too. How is having here different than there? are we sharing this `exporter metrics` instance somehow?

Please elaborate to me if there's something I might be missing 🙏

EuroLew: I was thinking of a scenario where we would have multiple batch requests (per exporter) at once  (for whatever reason), and we would want to track the latencies of each. But this probably won't happen so I'll move it to metrics

PR #23110, opened 2024-10-07
refactor: Refactor MDC property actor scheduler name setting on gossip threads

### LINKED ISSUE #14648: Set MDC property actor-scheduler/actor-name for non actor threads
**Description**

As a way to unify logging about partition ID/node ID, we should set the MDC properties on `actor-scheduler` and `actor-name` where it makes sense.

- [x] Set `actor-scheduler` to `Broker-<nodeId>` on the Raft threads
- [x] Set `actor-name` to `raft-<threadType>-<partitionId>` on the Raft threads
- [ ] Set `actor-scheduler` to `Broker-<nodeId>` or `Gateway-<nodeId>` on the gossip threads
- [ ] Set `actor-scheduler` to `Broker-<nodeId>` or `Gateway-<nodeId>` on the messaging threads

The left overs are the gRPC threads from the gateway, which we don't have a whole lot of control on. It's likely possible to set some properties on these anyway, but I didn't dig too much into it.

filipecampos:
## Description
Refactored the initialization of the `ExecutorService`s in `SwimMembershipProtocol` to set the actor scheduler value.

## Related issues

closes #14648

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/AtomixCluster.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/AtomixCluster.java
@@ -104,12 +104,7 @@ public class AtomixCluster implements BootstrapService, Managed<Void> {
 
   public AtomixCluster(
       final ClusterConfig config, final Version version, final String actorSchedulerName) {
-    this(
-        config,
-        version,
-        buildMessagingService(config, actorSchedulerName),
-        buildUnicastService(config, actorSchedulerName),
-        actorSchedulerName);
+    this(config, version, null, null, actorSchedulerName);
   }
 
   protected AtomixCluster(
@@ -118,7 +113,6 @@ public class AtomixCluster implements BootstrapService, Managed<Void> {
       final ManagedMessagingService messagingService,
       final ManagedUnicastService unicastService,
       final String actorSchedulerName) {
-
     this.messagingService =
         messagingService != null
             ? messagingService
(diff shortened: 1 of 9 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/cluster/AtomixCluster.java
```diff
@@ -118,19 +113,15 @@ protected AtomixCluster(
       final ManagedMessagingService messagingService,
       final ManagedUnicastService unicastService,
       final String actorSchedulerName) {
-
+    config.getProtocolConfig().setActorSchedulerName(actorSchedulerName);
```
deepthidevaki, line 116: :thought_balloon:  It is not really a configuration for the protocol. I was thinking of passing it directly to the constructor the membership protocol. So in `buildMembershipProtocol`:

```
config.getProtocolConfig().getType().newProtocol(config.getProtocolConfig(), actorSchedulerName);
```

What do you think?

deepthidevaki: >  I had tried as you suggested but that introduces more changes due to the modification of GroupMembershipProtocol.newProtocol 

Is that an issue? The only main change is in the interface `newProtocol` and the two implementations of it, right? 

`buildClusterMembership` and `buildMessagingService` are internal methods in `AtomixCluster`. So I guess we are not getting much benefit by removing the parameter from there.

filipecampos: I agree with that view since those methods are internal.

deepthidevaki (approved):
:rocket:  Thanks.

Issue #22868, opened 2024-09-30
Schema management works with OpenSearch


deepthidevaki:
CamundaExporter can create, update and validate mappings for the indices in OpenSearch. 
```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/issues/23163
- [ ] https://github.com/camunda/camunda/issues/23164
- [ ] https://github.com/camunda/camunda/issues/23166
```

PR #23125, opened 2024-10-07
Support opensearch in camunda exporter

### LINKED ISSUE #21840: Create initial version of OpensearchExporter
Currently `CamundaExporter` only supports exporting to `Elasticsearch`, yet it should be able to export to both `Elasticsearch` or `Opensearch` based on the configuration.

In this issue, the goal is to allow `CamundaExporter` to create and use opensearch based on configuration passed to it, and add the needed functionality to connect, export and flush to opensearch.

The task might require some refactoring to the current Exporter implementation to make it `vendor` agnostic.

Anything related to schemas or indexes that might (or might not) be different in opensearch than elasticsearch is out of scope of this issue.

mustafadagher:
## Description

- Refactor camunda-exporter to extract out any vendor specific logic
- Simplify exporter configs and remove several layers of hierarchy
- Allow `camunda-exporter` to export to either ES or OS based on `connect` config

This PR still misses the opensearch implementations of `SearchEngineClient` and `SchemaManager`. I was waiting to have them ready before opening the PR. But then I thought this already brings and added value and the rest can be addressed in #22868 

## Related issues

closes #21840


mustafadagher:
rebased to resolve conflicts from https://github.com/camunda/camunda/pull/23146

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ElasticsearchAdapter.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public void createClient(final ConnectConfiguration config) {
+    final var connector = new ElasticsearchConnector(config);
+    client = connector.createClient();
+  }
+
+  @Override
+  public SearchEngineClient createSearchEngineClient() {
+    Objects.requireNonNull(
```
EuroLew, line 38: Creating client in the constructor also allows removal of this check

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ElasticsearchAdapter.java
```diff
@@ -0,0 +1,71 @@
...
+    return new ElasticsearchSchemaManager(
+        searchEngineClient,
+        provider.getIndexDescriptors(),
+        provider.getIndexTemplateDescriptors(),
+        configuration);
+  }
+
+  @Override
+  public BatchRequest createBatchRequest() {
+    Objects.requireNonNull(
```
EuroLew, line 60: same for removal of this check, if client is removed in constructor it can be removed

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ElasticsearchAdapter.java
```diff
@@ -0,0 +1,71 @@
...
+        client, "ElasticsearchClient must be created before creating search engine client");
+    searchEngineClient = new ElasticsearchEngineClient(client);
+    return searchEngineClient;
+  }
+
+  @Override
+  public SchemaManager createSchemaManager(
+      final ExporterResourceProvider provider, final ExporterConfiguration configuration) {
+    if (searchEngineClient == null) {
```
EuroLew, line 47: If the client is created in the constructor, why not also create the searchEngineClient there eagerly, this would allow removal of this check and maybe rename the method to `getSearchEngineClient`?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -65,9 +63,9 @@ public CamundaExporter(final ExporterResourceProvider provider) {
 
   @Override
   public void configure(final Context context) {
-    configuration =
-        context.getConfiguration().instantiate(ElasticsearchExporterConfiguration.class);
+    configuration = context.getConfiguration().instantiate(ExporterConfiguration.class);
     provider.init(configuration);
+    createClientAdapter();
```
EuroLew, line 68: maybe a better name `setClientAdapter`? Or you can do a refactor with the enhanced switch to return values/throw and do a `clientAdapater = createClientAdapter()`

EuroLew (commented):
Really nice making the config implementation agnostic, this will make the Opensearch schema manager a lot easier 💯, just some thoughts on the adapter mainly.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ClientAdapter.java
```diff
@@ -0,0 +1,29 @@
...
+
+import io.camunda.exporter.ExporterResourceProvider;
+import io.camunda.exporter.config.ExporterConfiguration;
+import io.camunda.exporter.schema.SchemaManager;
+import io.camunda.exporter.schema.SearchEngineClient;
+import io.camunda.exporter.store.BatchRequest;
+import io.camunda.search.connect.configuration.ConnectConfiguration;
+import java.io.IOException;
+
+public interface ClientAdapter {
```
EuroLew, line 18: To be honest I'm a massive fan of a static factory e.g. `ClientAdapter.of(config)` (or any appropriate name), this lets us hide details about whether it is an opensearch/elasticsearch in the config we can place the `createClientAdapter` logic in this `of` function

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ClientAdapter.java
```diff
@@ -0,0 +1,29 @@
...
+  void createClient(final ConnectConfiguration config);
```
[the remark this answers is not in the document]
mustafadagher: I thought about that too. Yet, I'm creating the Adapter in `configure()` but the client itself is created in the `open`

I want to create the adapter in configure to validate that we have the right connection type at this point, or fail configure. and the client creation at `open()` maintains the old behavior (which I think has a point)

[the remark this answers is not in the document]
mustafadagher: I will separate the config validation from this creation and do it this way!

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ClientAdapter.java
```diff
@@ -8,20 +8,25 @@
...
 import io.camunda.exporter.store.BatchRequest;
-import io.camunda.search.connect.configuration.ConnectConfiguration;
 import java.io.IOException;
 
 public interface ClientAdapter {
-  void createClient(final ConnectConfiguration config);
 
-  SearchEngineClient createSearchEngineClient();
+  static ClientAdapter of(final ExporterConfiguration configuration) {
+    return switch (ConnectionTypes.from(configuration.getConnect().getType())) {
```
EuroLew, line 21: We can have a default: throw which would validate the connection type, the difference being it would throw in open (which would retry, and the exporter would be stuck in a retry loop), verses throwing in configure (which I think just crashes and doesn't retry? Correct me if I'm wrong).

EuroLew (approved):
It's not a blocking change so I'll approve.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ClientAdapter.java
mustafadagher: I think configuration validation should be done in `config` anyway, we have a whole ticket to start introducing it.

Also, there won't be a case for default, since converting to the enum type will fail in case of a wrong value. (even if I removed my throws exception statement there, it will throw `IllegalArgumentException`)

EuroLew: I was just thinking of having `configuration.getConnect().getType()` in the switch statement to catch it but if we're doing validation in `configure` then keep it that way to be consitent.

mustafadagher:
rebased again to resolve conflicts

PR #23162, opened 2024-10-08
Add new RoleRecord and RoleIntent

### LINKED ISSUE #22912: A role record exists with relevant intents
<h2>Description</h2>

Please refer to our [developer handbook](https://github.com/camunda/camunda/blob/main/zeebe/docs/developer_handbook.md#how-to-create-a-new-record) for an in-detail description

We need to create a new `RoleRecord` with properties:
- `roleKey`
    - type: Long
    - Optional
    - The key generated b yZeebe when processing the Role Create command. It is the unique identifier of the role within our system.
- `name`
    -  type: String
    - Optional
    - The name of the role. Not used within Zeebe itself
- `entityKey`
    - type: Long
    - Optional
    - Key of a user/mapping to assign/remove from a role

We need to introduce a `RoleIntent` enum with the following values:
- CREATE
- CREATED
- UPDATE
- UPDATED
- DELETE
- DELETED
- ADD_ENTITY
- ENTITY_ADDED
- REMOVE_ENTITY
- ENTITY_REMOVED

nicpuppa:
## Description

This PR creates new `RoleRecord` and `RoleIntent` classes. Exporting the new record will be handled in a separate issue

## Related issues

closes #22912

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/appliers/EventAppliersTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/appliers/EventAppliersTest.java
@@ -20,6 +20,7 @@ import io.camunda.zeebe.engine.state.mutable.MutableProcessingState;
 import io.camunda.zeebe.protocol.record.RecordValue;
 import io.camunda.zeebe.protocol.record.intent.Intent;
 import io.camunda.zeebe.protocol.record.intent.ProcessIntent;
+import io.camunda.zeebe.protocol.record.intent.RoleIntent;
 import io.camunda.zeebe.protocol.record.intent.management.CheckpointIntent;
 import java.util.Arrays;
 import org.junit.jupiter.api.Assertions;
@@ -167,7 +168,9 @@ public class EventAppliersTest {
             .flatMap(c -> Arrays.stream(c.getEnumConstants()))
             .filter(Intent::isEvent)
             // CheckpointIntent is not handled by the engine
-            .filter(intent -> !(intent instanceof CheckpointIntent));
+            .filter(intent -> !(intent instanceof CheckpointIntent))
+            // todo delete this filter after all the appliers are implemented
+            .filter(intent -> !(intent instanceof RoleIntent));
 
     // when
     eventAppliers.registerEventAppliers(mock(MutableProcessingState.class));
--- a/zeebe/exporters/elasticsearch-exporter/src/test/java/io/camunda/zeebe/exporter/TestSupport.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/test/java/io/camunda/zeebe/exporter/TestSupport.java
@@ -132,7 +132,9 @@ final class TestSupport {
             ValueType.SBE_UNKNOWN,
             ValueType.NULL_VAL,
             ValueType.PROCESS_INSTANCE_RESULT,
-            ValueType.CLOCK);
+            ValueType.CLOCK,
+            // these are not yet supported
+            ValueType.ROLE);
     return EnumSet.complementOf(excludedValueTypes).stream();
   }
 }
--- a/zeebe/exporters/opensearch-exporter/src/test/java/io/camunda/zeebe/exporter/opensearch/TestSupport.java
+++ b/zeebe/exporters/opensearch-exporter/src/test/java/io/camunda/zeebe/exporter/opensearch/TestSupport.java
@@ -120,7 +120,9 @@ final class TestSupport {
             ValueType.SBE_UNKNOWN,
             ValueType.NULL_VAL,
             ValueType.PROCESS_INSTANCE_RESULT,
-            ValueType.CLOCK);
+            ValueType.CLOCK,
+            // these are not yet supported
+            ValueType.ROLE);
     return EnumSet.complementOf(excludedValueTypes).stream();
   }
 }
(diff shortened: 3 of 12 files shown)


nicpuppa:
Nevermind I cannot decouple the exporting part, I will add the implementation

zeebe/exporters/elasticsearch-exporter/src/test/java/io/camunda/zeebe/exporter/TestSupport.java
```diff
@@ -132,7 +132,9 @@ static Stream<ValueType> provideValueTypes() {
             ValueType.SBE_UNKNOWN,
             ValueType.NULL_VAL,
             ValueType.PROCESS_INSTANCE_RESULT,
-            ValueType.CLOCK);
+            ValueType.CLOCK,
+            // these are not yet supported
+            ValueType.ROLE);
```
nicpuppa, line 137: Adding this here as I'm not sure if following the old approach is still valid

zeebe/exporters/opensearch-exporter/src/test/java/io/camunda/zeebe/exporter/opensearch/TestSupport.java
```diff
@@ -120,7 +120,9 @@ static Stream<ValueType> provideValueTypes() {
             ValueType.SBE_UNKNOWN,
             ValueType.NULL_VAL,
             ValueType.PROCESS_INSTANCE_RESULT,
-            ValueType.CLOCK);
+            ValueType.CLOCK,
+            // these are not yet supported
+            ValueType.ROLE);
```
nicpuppa, line 125: Adding this here as I'm not sure if following the old approach is still valid

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/appliers/EventAppliersTest.java
```diff
@@ -167,7 +168,9 @@ void shouldRegisterApplierForAllIntents() {
             .flatMap(c -> Arrays.stream(c.getEnumConstants()))
             .filter(Intent::isEvent)
             // CheckpointIntent is not handled by the engine
-            .filter(intent -> !(intent instanceof CheckpointIntent));
+            .filter(intent -> !(intent instanceof CheckpointIntent))
+            // todo delete this filter after all the appliers are implemented
+            .filter(intent -> !(intent instanceof RoleIntent));
```
nicpuppa, line 173: Added this otherwise the test will fail for missing appliers

zeebe/protocol/src/main/resources/protocol.xml
```diff
@@ -58,6 +58,7 @@
       <validValue name="USER">41</validValue>
       <validValue name="CLOCK">42</validValue>
       <validValue name="AUTHORIZATION">43</validValue>
+      <validValue name="ROLE">44</validValue>
```
remcowesterhoud, line 61: 💭 This will conflict with another PR. Guess we'll see which merges first 😄

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/RoleRecord.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.camunda.zeebe.msgpack.property.LongProperty;
+import io.camunda.zeebe.msgpack.property.StringProperty;
+import io.camunda.zeebe.protocol.impl.record.UnifiedRecordValue;
+import io.camunda.zeebe.protocol.record.value.RoleRecordValue;
+import io.camunda.zeebe.util.buffer.BufferUtil;
+
+public class RoleRecord extends UnifiedRecordValue implements RoleRecordValue {
+
+  private final LongProperty roleKeyProp = new LongProperty("roleKey");
```
remcowesterhoud, line 18: ❌ This should default to -1. If we create a new role we won't have the key available yet.

zeebe/exporters/elasticsearch-exporter/src/test/java/io/camunda/zeebe/exporter/TestSupport.java
remcowesterhoud: I informed about this [here](https://camunda.slack.com/archives/C05QBRGMA9F/p1728383396104949?thread_ts=1728374201.862919&cid=C05QBRGMA9F). We don't need to support these records in the old exporters as long as they don't get consumed. The only component still using the old exporters is Optimize and since they won't consume these records we can safely ignore them and don't need to support them in the ES/OS exporters.

zeebe/exporters/opensearch-exporter/src/test/java/io/camunda/zeebe/exporter/opensearch/TestSupport.java
remcowesterhoud: Same comment as vor the ES exporter

remcowesterhoud (changes requested):
Thanks @nicpuppa! Just some small stuff 👍

zeebe/protocol/src/main/resources/protocol.xml
nicpuppa: Yes, saw that

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/RoleRecord.java
nicpuppa: Right 👍

remcowesterhoud (approved):
🏆

PR #23197, opened 2024-10-09
feat: provide isolate CompositeMeterRegistry for each ExporterContainer

### LINKED ISSUE #22834: Provide isolated composite meter registry per exporter container
**Description**

Now that we expose a `MeterRegistry` to the exporter API, exporters can easily add metrics to Zeebe. That's great, but there is a drawback to exposing the actual global meter registry: metrics live forever (even if they are ephemeral) unless the exporter explicitly removes them (which it likely won't), you can't share common tags (e.g. partition ID, exporter ID). The first one in particular can be problematic if you have dynamic sets of metrics based on, say, a dynamic set of tags. It will also cause problem with "restarting" values on transitions, for example, for certain kinds of metrics (e.g. gauges).

The proposal is to create a new `CompositeMeterRegistry` per `ExporterContainer`, where we register the global one, with the common tags `partitionId` and `exporterId`, and close the `CompositeMeterRegistry` when the exporter container is closed.

### Acceptance

- [ ] Create a new `CompositeMeterRegistry` per `ExporterContainer`
  - [ ] Register the `brokerMeterRegistry` to the newly created `CompositeMeterRegistry`
- [ ] Add common tags `partitionId` and `exporterId` with correct values
- [ ] Close the composite meter registry when closing the exporter container, _after the exporter is closed of course_.
- [ ] On top of any required unit/integration tests, test a live benchmark with our dashboard on https://grafana.dev.zeebe.io to make sure the metrics still work (e.g. Elasticsearch exporter metrics)
  - [ ] Make sure to test with node restarts and transitions to make sure we don't get strange values.

entangled90:
## Description

A dedicated CompositeMeterRegistry is created in the ExporterContext, instantiated by ExporterContainer.
When the context is closed, its metrics are cleared and its unregistered from "global" meterRegistry.

## Checklist
[x] check that metrics are registered/unregistered correctly using a benchmark

Released branch as benchmark: See [grafana](https://grafana.dev.zeebe.io/d/zeebe-dashboard/zeebe?orgId=1&var-DS_PROMETHEUS=prometheus&var-cluster=All&var-namespace=isolated-meter-registery-exporter&var-pod=All&var-partition=All&from=1728478214000&to=1728479125000). At 14:55 a leadership change occurred and metrics from one exporter "stopped" 

## Related issues

closes #22834

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java
@@ -13,17 +13,20 @@ import io.camunda.zeebe.protocol.record.RecordType;
 import io.camunda.zeebe.protocol.record.ValueType;
 import io.camunda.zeebe.util.EnsureUtil;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 import java.time.InstantSource;
 import org.slf4j.Logger;
 
-public final class ExporterContext implements Context {
+public final class ExporterContext implements Context, AutoCloseable {
 
   private static final RecordFilter DEFAULT_FILTER = new AcceptAllRecordsFilter();
 
   private final Logger logger;
   private final Configuration configuration;
   private final int partitionId;
-  private final MeterRegistry meterRegistry;
+  private final CompositeMeterRegistry meterRegistry;
+  private final MeterRegistry underlyingMetricRegistry;
   private final InstantSource clock;
 
   private RecordFilter filter = DEFAULT_FILTER;
@@ -37,7 +40,18 @@ public final class ExporterContext implements Context {
     this.logger = logger;
     this.configuration = configuration;
     this.partitionId = partitionId;
-    this.meterRegistry = meterRegistry;
+    underlyingMetricRegistry = meterRegistry;
+    this.meterRegistry = new CompositeMeterRegistry();
+    // meterRegistry is null in tests
+    if (meterRegistry != null) {
+      this.meterRegistry.add(meterRegistry);
+    }
+    this.meterRegistry
+        .config()
+        .commonTags(
+            Tags.of(
+                "partition", Integer.toString(partitionId),
+                "exporterId", configuration.getId()));
     this.clock = clock;
   }
 
@@ -74,6 +88,18 @@ public final class ExporterContext implements Context {
   public void setFilter(final RecordFilter filter) {
     EnsureUtil.ensureNotNull("filter", filter);
     this.filter = filter;
+  }
+
+  @Override
+  public void close() {
+    // Clear the registry, so the metrics created are deleted:
+    // it must be done before removing the parent registry
+    meterRegistry.clear();
+    // remove the parentMeterRegistry so it's not closed when we close the composite.
+    if (underlyingMetricRegistry != null) {
+      meterRegistry.remove(underlyingMetricRegistry);
+    }
+    meterRegistry.close();
   }
 
   private static final class AcceptAllRecordsFilter implements RecordFilter {
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
@@ -97,16 +97,19 @@ public final class ExporterRepository {
   private void validate(final ExporterDescriptor descriptor) throws ExporterLoadException {
     try {
       final Exporter instance = descriptor.newInstance();
-      final ExporterContext context =
+      try (final var context =
           new ExporterContext(
               LOG,
               descriptor.getConfiguration(),
               NULL_PARTITION_ID,
               new SimpleMeterRegistry(),
-              InstantSource.system());
+              InstantSource.system())) {
 
-      ThreadContextUtil.runCheckedWithClassLoader(
-          () -> instance.configure(context), instance.getClass().getClassLoader());
+        ThreadContextUtil.runCheckedWithClassLoader(
+            () -> instance.configure(context), instance.getClass().getClassLoader());
+      } finally {
+        instance.close();
+      }
     } catch (final Exception ex) {
       throw new ExporterLoadException(descriptor.getId(), "failed validation", ex);
     }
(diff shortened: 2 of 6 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/metrics/MetricsTags.java
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.broker.exporter.metrics;
+
+/** Collection of tags used in micrometer metrics */
+public class MetricsTags {
```
lenaschoenburg, line 11: :upside_down_face: can be `final` and declare a private empty constructor to prevent useless instantiation.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
```diff
@@ -97,16 +97,19 @@ public ExporterDescriptor load(final String id, final ExporterCfg config)
...
-              InstantSource.system());
+              InstantSource.system())) {
 
-      ThreadContextUtil.runCheckedWithClassLoader(
-          () -> instance.configure(context), instance.getClass().getClassLoader());
+        ThreadContextUtil.runCheckedWithClassLoader(
+            () -> instance.configure(context), instance.getClass().getClassLoader());
+      } finally {
+        instance.close();
+      }
```
lenaschoenburg, line 112: :thought_balloon: Huh, interesting that we didn't close the instance before. How did you notice this?

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/metrics/MetricsTags.java
lenaschoenburg: :wrench: This mixes two different types of tags: the generic ones set on the registry (partitionId, exporterId) and ones specific to the `MetricsExporter` (jobs, type). This, and that this class is defined in the package used for the `MetricsExporter` is a bit confusing IMO. Instead of splitting this class accordingly into two which each define just two constants, I'd just go back to inlining the label names again.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/metrics/MetricsExporterTest.java
```diff
@@ -96,6 +124,67 @@ void shouldObserveJobLifetime() {
         .describedAs("Expected the correct job_life_time bucket to have counted the event");
   }
 
+  @Test
+  void shouldCleanupMetricsWhenContextIsClosed() throws Exception {
```
lenaschoenburg, line 128: :wrench: I'd suggest making this a generic test for the behavior of `ExporterContext` instead of a test specific to the `MetricsExporter`. Something like: 
1. Create a context
2. Set some metrics
3. Assert that metrics are set with partitionId, exporterId labels

And then a second one to assert that the metrics are removed when the context is closed.

lenaschoenburg (changes requested):
Basically works! :rocket: 

But I have a couple of requested changes, please take a look.

:wrench: I'd suggest to split up any additional changes over multiple commits.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/repo/ExporterRepository.java
entangled90: I noticed because I made the ExporterContext Autocloseable and surrounded its constructor with try-with-resources.
But I think in this we don't need to close it explicitly as the Exporter instance itself will close the context. 
So I will leave only the `instance.close()`

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java
```diff
@@ -37,7 +40,18 @@ public ExporterContext(
...
+    this.meterRegistry = new CompositeMeterRegistry();
+    // meterRegistry is null in tests
+    if (meterRegistry != null) {
+      this.meterRegistry.add(meterRegistry);
+      this.meterRegistry
+          .config()
+          .commonTags(
+              Tags.of(
+                  "partition", Integer.toString(partitionId),
+                  "exporterId", configuration.getId()));
```
lenaschoenburg, line 53: :wrench: doesn't need to be within the null check

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java
```diff
@@ -76,6 +90,18 @@ public void setFilter(final RecordFilter filter) {
     this.filter = filter;
   }
 
+  @Override
+  public void close() {
+    if (meterRegistry != null) {
```
lenaschoenburg, line 95: :wrench: Null check not needed anymore

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java
```diff
@@ -37,7 +40,18 @@ public ExporterContext(
     this.logger = logger;
     this.configuration = configuration;
     this.partitionId = partitionId;
-    this.meterRegistry = meterRegistry;
+    underlyingMetricRegistry = meterRegistry;
+    this.meterRegistry = new CompositeMeterRegistry();
+    // meterRegistry is null in tests
+    if (meterRegistry != null) {
+      this.meterRegistry.add(meterRegistry);
```
lenaschoenburg, line 47: :x: Should be 
```suggestion
      this.meterRegistry.add(underlyingMetricRegistry);
```

lenaschoenburg: Let's add a test for that too, checking that metrics published by an exporter are forwarded to the underlying registry

lenaschoenburg (changes requested):

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java
lenaschoenburg: Ah no wait, this is correct because `meterRegistry` is the param, not the field. My bad!

Test could be useful anyway.

lenaschoenburg (approved):

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java
entangled90: I'll add assertions that in both registry we can see the same metrics

Issue #22866, opened 2024-09-30
Allow configuring the CamundaExporter


deepthidevaki:
Most configurable parameters are already added as part of the respective PRs that add the feature. But there might be missing ones. This issue is to review and add any necessary configuration.
- [x] https://github.com/camunda/camunda/pull/23198
- [x] Review what is configurable in Operate and if we need to port this
- [x] Review currently implemented properties and if we should consolidate them.
- [x] Add validations for the configuration if applicable
- [x] Provide the config to enable the Exporter in the Helm charts - see https://github.com/camunda/camunda-platform-helm/issues/2294
- [x] Provide the config to enable the Exporter in the SaaS controller - see https://github.com/camunda-cloud/camunda-operator/issues/3258

mustafadagher:
I reviewed the configuration used in Operate as much as my knowledge could help (and in zeebe exporters too), and I couldn't find any configs that we missed except for two main things that exist in zeebe elastic search exporter
1. `indexSuffixDatePattern` and I don't think we need this one now, unless you think otherwise.
2. `bulk.memoryLimit` which we have now, yet left unused and I'll delete it according to the discussion in #22869 

I'll open a PR in a bit with some config validations that are ported from zeebe (I couldn't find anything relevant in operate code base).

@sdorokhova please advise if you see something that I might have missed.

sdorokhova:
> **EDIT:** Operate also uses a `refreshInterval` in the indexSettings, I don't know if we should have this in the exporter as well.
> 
> @sdorokhova please advise if you see something that I might have missed.

I don't think that we need it.

Going over Operate code, it looks like we use it only in migration, which we won't have.

PR #23198, opened 2024-10-09
Refactor exporter config

mustafadagher:
## Description

Refactor exporter configs to match zeebe configs as possible

The goal is to avoid confusion when configuring the new exporter and make it more fluent
- rename IndexSettings var to index
- move indexPrefix to be part of IndexSettings
- adjust usages

In addition, also made appendToFileSchemaSettings more concise

## Related issues

relates to #22866


EuroLew (approved):
LGTM

PR #23205, opened 2024-10-09
Remove authorization wrapper classes

remcowesterhoud:
## Description

The wrapper classes became too complex, introduced some duplication and were difficult to generalise for all the cases we've got to support. Instead of these wrappers we can just call the `AuthorizationCheckBehavior` from our processors directly and reject commands inside of the processors.

You win @tmetzke 😉 

## Checklist

- [ ] for CI changes:

## Related issues

N/A


tmetzke (approved):
👍 This is GREAT 🚀 
🌱 If I am not mistaken, we just saved 200 lines of code and are still not fetching data twice due to auth checks, right?

💭 Follow-up thought: we could move some duplicated code into the auth check behavior methods, e.g. error message creation. The auth check methods could return `Optional<Rejection>` that contains the rejection to use already.

remcowesterhoud:
> 💭 Follow-up thought: we could move some duplicated code into the auth check behavior methods, e.g. error message creation. The auth check methods could return Optional<Rejection> that contains the rejection to use already.

That's true, but I could imagine we need to personalize the error messages still. For example, creating a PI now only says you're not allowed to create this PI, instead of mentioning the `processId` as well. But I want to align these message with the error messages given by search requests and haven't talked about this yet.

PR #23220, opened 2024-10-09
fix: migration issue for passwords on ES/OS

### LINKED ISSUE #23213: Password encoder change is not backward compatible
### Describe the bug

After upgrading from 8.5 to 8.6 passwords store in Elastic in BCrypt format are not working any more. The reason is changed password encoder: https://github.com/camunda/camunda/pull/21039/files.

The workaround is to append all password values with `{bcrypt}` directly in Elasticsearch.

### To Reproduce

1. Start Operate/Tasklist 8.5.x with default authentication (no Identity or Auth0).
2. Log in with demo/demo user and make sure that it's working.
3. Upgrade Operate/Tasklist to 8.6.0 and try to log in.

### Current behavior

Login is not possible:
```
2024-10-09 12:15:14.474 [] [http-nio-0.0.0.0-8080-exec-8] [] ERROR
      org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/].[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception
java.lang.IllegalArgumentException: You have entered a password with no PasswordEncoder. If that is your intent, it should be prefixed with `{noop}`.
        at org.springframework.security.crypto.password.DelegatingPasswordEncoder$UnmappedIdPasswordEncoder.matches(DelegatingPasswordEncoder.java:296) ~[spring-security-crypto-6.3.3.jar:6.3.3] ....
```
### Expected behavior

Login still works with the same passwords.

### Environment

- Operate Version: 8.6.0

### Additional context

https://camunda.slack.com/archives/CRQAL5A1M/p1728476689754559

- [ ] The issue has been labeled with an appropriate `Bug-area` label

- Changed components:
- Side effects on other components:
- Handy resources:
  
- Example projects:

- Commands/Steps needed to test; Versions to validate:

- Docker file / HELM chart : in case that it needed to be tested via docker share the version contain the fixed along with version of other services .

- Release version ( in which version this fixed/feature will be released):

marcosgvieira:
## Description

Reverting change around Encryptor as it's causing problems on the migrations - it's not backward compatible so when coming from 8.5, password cannot be checked.

## Checklist

- [ ] for CI changes:

## Related issues

closes #23213

### DIFF
--- a/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserDetailsService.java
+++ b/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserDetailsService.java
@@ -26,7 +26,7 @@ import org.springframework.context.annotation.Primary;
 import org.springframework.context.annotation.Profile;
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
-import org.springframework.security.crypto.factory.PasswordEncoderFactories;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.stereotype.Component;
 
@@ -50,7 +50,7 @@ public class SearchEngineUserDetailsService implements UserDetailsService {
   @Bean
   @Primary
   public PasswordEncoder getPasswordEncoder() {
-    return PasswordEncoderFactories.createDelegatingPasswordEncoder();
+    return new BCryptPasswordEncoder();
   }
 
   public void initializeUsers() {
@@ -86,10 +86,10 @@ public class SearchEngineUserDetailsService implements UserDetailsService {
     }
   }
 
-  private boolean userExists(final String userId) {
+  private boolean userExists(String userId) {
     try {
       return userStore.getByUserId(userId) != null;
-    } catch (final Exception t) {
+    } catch (Exception t) {
       return false;
     }
   }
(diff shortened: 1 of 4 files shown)


tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserDetailsService.java
```diff
@@ -86,10 +86,10 @@ public void initializeUsers() {
     }
   }
 
-  private boolean userExists(final String userId) {
+  private boolean userExists(String userId) {
```
nathansandi, line 89: Minor question, any reason for this not being final for userId?

nathansandi (approved):
LGTM, Just a minor question

houssain-barouni (approved):
LGTM, just one remark, the users created in 8.6.0 may not work after the upgrade to 8.6.1. But I think this is less "critical"

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/se/SearchEngineUserDetailsService.java
sdorokhova: I think it's just an automatic revert of previous commits. That's why final modifier is also reverted.

sdorokhova (approved):

sdorokhova:
> LGTM, just one remark, the users created in 8.6.0 may not work after the upgrade to 8.6.1. But I think this is less "critical"

@nathansandi I thought about that also. Since we have quite easy workaround - remove `{bcrypt}` prefix - should be fine IMO. If they come with such support question we will guide them how to fix it.

marcosgvieira:
> LGTM, just one remark, the users created in 8.6.0 may not work after the upgrade to 8.6.1. But I think this is less "critical"

Good point @houssain-barouni! I did a test creating the users with the `PasswordEncoderFactories.createDelegatingPasswordEncoder();` and trying to login using `BCryptPasswordEncoder()` and it worked, so I believe we should be ok

PR #23225, opened 2024-10-09
fix: c8run enable v2 apis and add test for APIs

jessesimpson36:
## Description

As part of our discussion in the c8run-sdk-auth-issue slack channel, we learned that some of the v2 api's are locked behind a feature flag. This PR proposes enabling those by default inside c8run, as well as disabling CSRF prevention which causes issues when interacting with the v1 API's in c8run.

Here are the 3 flags changed:

```diff
+export CAMUNDA_REST_QUERY_ENABLED=true
+export CAMUNDA_OPERATE_CSRFPREVENTIONENABLED=false
+export CAMUNDA_TASKLIST_CSRFPREVENTIONENABLED=false

```

And this PR also introduces a bash script unit test for validating that these API's are operational. One thing I'd like to note is that the windows version does not execute this bash API integration test because windows doesn't use bash.  This was simply the quickest / easiest thing I could hack together. 

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/.github/workflows/c8run-build.yaml
+++ b/.github/workflows/c8run-build.yaml
@@ -70,6 +70,10 @@ jobs:
         run: npx playwright test
         working-directory: ./c8run/e2e_tests
 
+      - name: Run v2 API endpoint test
+        run: ./api_tests.sh
+        working-directory: ./c8run/e2e_tests
+
       - uses: actions/upload-artifact@v4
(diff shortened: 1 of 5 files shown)


hisImminence (approved):
lgtm, thanks @jessesimpson36

PR #23227, opened 2024-10-09
Support delete operations in the batch request

Ben-Sheppard:
## Description

Identity has a requirement that we perform deletes by doc ID, this PR introduces the base `delete` method in the BatchRequest class such that it can be used later in the handlers to process deletes.


mustafadagher (approved):
LGTM! 🚀

PR #23254, opened 2024-10-10
Add RoleCreatedApplier

### LINKED ISSUE #22916: Apply `Role.CREATED` event
<h2>Description</h2>

In Zeebe we modify our state by applying events. We need to create a new `RoleCreatedApplier`:

3. Add a method to to the `MutableRoleState` interface to create a role and implement this method in the `DbRoleState`:
    - Wrap the `roleKey` and set the record in the `persistedRole`
    - Wrap the `roleName`
    - Insert into the `ROLES` column family
    - Insert into the `ROLE_BY_NAME` column family
    - Unit test this method
4. In the `RoleCreatedApplier` call this create method of the `DbRoleState`

nicpuppa:
## Description

This PR adds the `RoleCreatedApplier`. It also add `create`, `get` and `getByName` method in the `RoleState` to create and retrieve roles.

## Related issues

closes #22916

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbRoleState.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbRoleState.java
@@ -17,6 +17,7 @@ import io.camunda.zeebe.db.impl.DbString;
 import io.camunda.zeebe.engine.state.mutable.MutableRoleState;
 import io.camunda.zeebe.protocol.ZbColumnFamilies;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
+import java.util.Optional;
 
 public class DbRoleState implements MutableRoleState {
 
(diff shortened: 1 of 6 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbRoleState.java
```diff
@@ -55,4 +56,28 @@ public DbRoleState(
...
+    roleKey.wrapLong(roleRecord.getRoleKey());
+    persistedRole.setRole(roleRecord);
+    roleColumnFamily.insert(roleKey, persistedRole);
+
+    roleName.wrapString(roleRecord.getName());
+    roleByNameColumnFamily.insert(roleName, fkRoleKey);
+  }
+
+  @Override
+  public RoleRecord getRole(final long roleKey) {
```
remcowesterhoud, line 71: 🔧 WDYT about returning an Optional instead of null?

remcowesterhoud (approved):
You are flying through these 😄 I've pre-approved but please have a look at my comment.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbRoleState.java
nicpuppa: 👍

PR #23277, opened 2024-10-10
Template descriptor creates indices

EuroLew:
## Description

Template descriptors should create the index template then create the matching index so that it is visible during startup to operate.

**Success Criteria:**
- [x] Template descriptors create template and then matching index.


sdorokhova (approved):
LGTM

PR #23305, opened 2024-10-10
refactor: remove serialization logic from SegmentDescriptor

### LINKED ISSUE #7650: Extract reading and writing of segment descriptor out of the JournalSegmentDescriptor
Extract reading and writing of segment descriptor out of the JournalSegmentDescriptor. We can also think of having a `SegmentDescriptorLoader` which knows how to read,write and serialize segment descriptors. This is only needed when loading or creating a segment. Once the segment is created we always refer to the in memory `JournalSegmentDescriptor`. 

Follow up #6221 
Related #7604

entangled90:
## Description

Following what has been done with JournalRecordSerializer, the logic related to serializing/deserializing a SegmentDescriptor has been moved to a dedicated interface: SegmentDescriptorSerializer. Currently only one implementation (valid for versions 2.x) has been created using SBE.
In the future, additional implementation can be added in case of breaking changes of the schema.

SegmentDescriptor has been converted to a record, which means it is now immutable.
To limit the memory allocations, the memory layout of the record has been tweaked by using byte or short for some fields (version, activeSchemaVersion & encodingLength).

The diff looks huge, but most of the changes are an effect of moving code from one file to another one.

## Related issues

closes #7650


zeebe/journal/src/main/java/io/camunda/zeebe/journal/file/SegmentDescriptorSerializerSbe.java
```diff
@@ -0,0 +1,249 @@
...
+      } else {
+        throw new UnknownVersionException(
+            String.format(
+                "Expected version to be one (%d %d] but read %d instead.",
+                META_VERSION, CUR_VERSION, version));
+      }
+    } catch (final IndexOutOfBoundsException error) {
+      // Previously SegmentLoader checks if the file has sufficient size for the descriptor. But
+      // it
+      // is not checked anymore because the encoded length is not known before reading.
```
deepthidevaki, line 117: :wrench: Can format this :slightly_smiling_face:

deepthidevaki (approved):
Great :100:

Issue #21387, opened 2024-08-20
Update Operate docs screenshots


ThorbenLindhauer:
### Description

With https://github.com/camunda/camunda/issues/19554 we have changed how the executed path is visualized. This affects a lot of docs screenshots of Operate. Let's update them with [our automation](https://github.com/camunda/camunda/actions/workflows/operate-update-docs-screenshots.yml).

### TODO

* [x] Fix the screenshot workflow
* [x] Run it so that all Operate screenshots in the docs are updated for 8.6

### Additional Information

---------------------------------------------------------------------------------------------

### Acceptance Criteria

- [ ] Checked if the specific data needs to be provided for the QA process

### Definition of Ready - Checklist

- [ ] The issue has a meaningful title, description and testable acceptance criteria
- [ ] If documentation needs to be updated, an issue is created in the [camunda-docs](https://github.com/camunda/camunda-docs) repo, and the issue is added to our Operate project board.
- [ ] If HELM charts need to be updated, an issue is created in the [camunda-platform-heml](https://github.com/camunda/camunda-platform-helm) repo, and the issue is added to our Operate project board.
- [ ] Potential Security risks have been considered
- [ ] Cross-team dependencies have been considered

ThorbenLindhauer:
Once the workflow is fixed, let's share the PR with @christinaausley so that she can take it as an example on how to work with the automation.

pedesen:
@ThorbenLindhauer @christinaausley I merged the PR and run the screenshot generator on the main branch. It works! The screenshots can be downloaded as part of the playwright report here: https://github.com/camunda/camunda/actions/runs/11292047862

christinaausley:
Wahoo! Great news. I'll launch a docs PR. Thank you!

PR #18604, opened 2024-05-16
refactor: add base path in Operate and Tasklist apps urls

houssain-barouni:
## Description
The Camunda single application should serve Zeebe, Operate and Tasklist using one running application.

To dispatch between Operate and Tasklist UIs, we need to add `/operate` and `/tasklist` base paths to the URLs respectively used by Operate and Tasklist UIs
For consistency, the same base paths will be used when Operate and Tasklist are run in standalone mode.

- requests to `/tasklist` and `/operate` are redirected to Tasklist or Operate index file
- requests to the root domain are redirect to either `/tasklist` or `/operate` depending on the used profile in standalone mode, it defaults to `/operate` in single application mode.
- backend API urls do not change and do not contain the base paths.
- old app URLs not containing the base path (example `http://localhost:8080/processes`) will not be working, the app base path should be provided in the URL (example `http://localhost:8080/tasklist/processes`) 
- Tasklist public forms URLs (`/new/{bpmnProcessId}`do not require the base path and will continue to be working as they are served by a [dedicated endpoint in backend](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/api/rest/v1/controllers/PublicProcessController.java).

## Related issues

relates to https://github.com/camunda/zeebe/issues/18535

PR #23309, opened 2024-10-10
fix: correct broken test steps and create new version of process to enable valid migration

vitorwtavares:
## Description

- Fixes domain path problems which had roots related to #18604 
- Fixes overlay test ids that were changed in a different PR for E2E tests but not updated for the update docs screenshots workflow
- Creates a new version of the `orderProcess` process in order to have a valid mapping for the instance migration screenshots

## Related issues

relates to #21387 (closes part of the ticket)

### DIFF
--- a/operate/client/playwright.config.ts
+++ b/operate/client/playwright.config.ts
@@ -26,11 +26,7 @@ const getPort = () => {
 };
 
 const getBasePath = () => {
-  if (IS_CI) {
-    return IS_E2E ? '/operate/' : '';
-  }
-
-  return '/operate/';
+  return IS_CI && !IS_E2E ? '' : '/operate/';
 };
 
 /**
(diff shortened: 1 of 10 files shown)


operate/client/playwright.config.ts
```diff
@@ -26,7 +26,11 @@ const getPort = () => {
 };
 
 const getBasePath = () => {
-  return IS_E2E ? '/operate/' : '';
+  if (IS_CI) {
+    return IS_E2E ? '/operate/' : '';
+  }
+
+  return '/operate/';
```
pedesen, line 33: What do you think about this? Might make it a bit more readable:

```suggestion
  if (IS_CI && !IS_E2E) {
    return '';
  }

  return '/operate/';
```

pedesen (commented):
Looks good (besides the failing e2e tests). I also left a suggestion, which is "take it or leave it". Will approve when the e2e tests are fixed.

pedesen (approved):

PR #23320, opened 2024-10-10
build: Add script for v2 API env

vsgoulart:


tasklist/Makefile
```diff
@@ -16,6 +16,11 @@ env-up: $(DATABASE) setup-maven-context
 	$(eval SERVICES := zeebe)
 	$(call run-env)
 
+.PHONY: v2-up
+v2-up: $(DATABASE) setup-maven-context
```
panagiotisgts, line 20: I believe the V2 API params are already present in the [env](https://github.com/camunda/camunda/blob/c16758258ff00040054dd476152ffe6e99229502/tasklist/config/envs/.env.database.elasticsearch) file for ES/OS. Is there something else missing?

vsgoulart: @panagiotisgts You mean I should use these vars instead of hard-coding it?

panagiotisgts: They are always being added to the containers/runtime when using the Makefile or the docker-compose

panagiotisgts: Ah, I see it's the `elasticsearch` URL reference probably. Can you try adding
`./config/write_or_replace_var.sh "CAMUNDA_DATABASE_URL" "$DATABASE_URL" $MAVEN_CONTEXT`
in the `tasklist/config/env_setup.sh` along the other replacements that happen and try again the `make env-up` goal?

vsgoulart: @panagiotisgts it works, thanks

panagiotisgts (approved):
Thanks for taking care of the env setup :rocket:

PR #23353, opened 2024-10-11
Check permissions in `MessageCorrelationCorrelateProcessor`

### LINKED ISSUE #22885: Verify permissions in `MessageCorrelationCorrelateProcessor`
<h2>Description</h2>

Message correlate is special in the sense that it requires 2 different permission types, depending on if the message creates a PI, or correlates to an existing PI.

Creates a PI:
- **Possible permissions**:
    -  `bpmnProcessId - Foo`

Correlated to existing PI:
- **Permission type**: `UPDATE`
- **Possible permissions**:
    -  `bpmnProcessId - Foo`

The command is processed for all or for none. If the user is not permitted to correlate the message to one or more of the subscriptions, none of the correlations are correlated and the command should get rejected.

remcowesterhoud:
## Description

It's a bit of an odd one since there's different permissions to check here. The current solution is likely not the most performant, but it is the easiest way to implement it at this time. If it becomes a problem we can improve upon this later.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22885

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationCorrelateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationCorrelateProcessor.java
@@ -169,19 +169,14 @@ public final class MessageCorrelationCorrelateProcessor
     final var isAuthorized =
         correlatingSubscriptions.visitSubscriptions(
             subscription -> {
-              if (subscription.isStartEventSubscription()) {
-                request.set(
-                    new AuthorizationRequest(
-                        command,
-                        AuthorizationResourceType.PROCESS_DEFINITION,
-                        PermissionType.CREATE));
-              } else {
-                request.set(
-                    new AuthorizationRequest(
-                        command,
-                        AuthorizationResourceType.PROCESS_DEFINITION,
-                        PermissionType.UPDATE));
-              }
+              final PermissionType permissionType =
+                  subscription.isStartEventSubscription()
+                      ? PermissionType.CREATE
+                      : PermissionType.UPDATE;
+
+              request.set(
+                  new AuthorizationRequest(
+                      command, AuthorizationResourceType.PROCESS_DEFINITION, permissionType));
 
               request.get().addResourceId(bufferAsString(subscription.getBpmnProcessId()));
               return authCheckBehavior.isAuthorized(request.get());
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationCorrelateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationCorrelateProcessor.java
@@ -7,8 +7,14 @@
  */
 package io.camunda.zeebe.engine.processing.message;
 
+import static io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.UNAUTHORIZED_ERROR_MESSAGE;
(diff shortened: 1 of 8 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationCorrelateProcessor.java
```diff
@@ -138,4 +160,42 @@ private MessageData createMessageData(
         messageCorrelationRecord.getVariablesBuffer(),
         messageCorrelationRecord.getTenantId());
   }
+
+  private Optional<Rejection> isAuthorizedForAllSubscriptions(
```
nicpuppa, line 164: 🔧 This method can be refactored a bit:

```
private Optional<Rejection> checkAuthorizationForSubscriptions(
      final TypedRecord<MessageCorrelationRecord> command,
      final Subscriptions correlatingSubscriptions) {

    AuthorizationRequest authRequest = null;

    final var isAuthorized = correlatingSubscriptions.visitSubscriptions(
        subscription -> {

          PermissionType permission = subscription.isStartEventSubscription()
              ? PermissionType.CREATE
              : PermissionType.UPDATE;

          authRequest = new AuthorizationRequest(
              command, AuthorizationResourceType.PROCESS_DEFINITION, permission);
          authRequest.addResourceId(bufferAsString(subscription.getBpmnProcessId()));

          return authCheckBehavior.isAuthorized(authRequest);
        },
        true
    );

    if (!isAuthorized) {
      String errorMessage = UNAUTHORIZED_ERROR_MESSAGE.formatted(
          authRequest.getPermissionType(),
          authRequest.getResourceType()
      );
      return Optional.of(new Rejection(RejectionType.UNAUTHORIZED, errorMessage));
    }

    return Optional.empty();
  }
```

nicpuppa (approved):
💭 Another approach will be to use the AuthCheck behaviour inside the message correlation behaviour and check the authorization there, in both `correlateToMessageStartEvents` and `correlateToMessageEvents`. In this way we can remove looping again all the subscriptions, and also exit early while looping in the behaviour.
This approach will hide the auth check inside the behaviour of the message correlation, but we can underline this maybe changing the methods name.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/MessageCorrelationCorrelateProcessor.java
remcowesterhoud: That doesn't work. You cannot assign a value to `authRequest` from within a lambda.

remcowesterhoud: I'll take the permission type part though

nicpuppa: Oh yeah, sorry

remcowesterhoud:
> 💭 Another approach will be to use the AuthCheck behaviour inside the message correlation behaviour and check the authorization there, in both correlateToMessageStartEvents and correlateToMessageEvents. In this way we can remove looping again all the subscriptions, and also exit early while looping in the behaviour.
This approach will hide the auth check inside the behaviour of the message correlation, but we can underline this maybe changing the methods name.

I did consider this. The problem is this behavior is used by message publish as well as message correlation. I don't see a very straightforward solution to do this in a nice way currently. Let's revisit if it becomes a problem.

PR #23362, opened 2024-10-11
Add RoleEntityAddedApplier

### LINKED ISSUE #22923: Apply `Role.ENTITY_ADDED` event
<h2>Description</h2>

In Zeebe we modify our state by applying events. We need to create a new `RoleEntityAddedApplier`:

3. Add a method to to the `MutableRoleState` interface to add an entity to a role and implement this method in the `DbRoleState`:
    - Wrap the `roleKey` and `entityKey`
    - Insert into the `ENTITY_BY_ROLE` column family
    - Unit test this method
4. Add a method to the `MutableUserState` interface to add a role to a user
     - Wrap the `userKey`
     - Get the `PersistedUser` from `USERS` ColumnFamily
     - Add the `roleKey` to the `PersistedUser`
     - Update the `PersistedUser` in the ColumnFamily
5. Add a method to the `MutableMappingState` interface to add a role to a mapping
     - Wrap the `mappingKey`
     - Get the `PersistedMapping` from `MAPPINGS` ColumnFamily
     - Add the `roleKey` to the `PersistedMapping`
     - Update the `PersistedMapping ` in the ColumnFamily 
4. In the `RoleEntityAddedApplier `:
    - Call the new method in the `MutableRoleState` to store the entity key by role key
    - If `entityType` = USER, call the method in the `MutableUserState` to add the role to the user
    - if `entityType` = MAPPING, call the method in the `MutableMappingState` to add the role to the mapping
5. Unit test all the new state methods

nicpuppa:
## Description

This PR:
1. Adds a method to to the `MutableRoleState` interface to add an entity to a role
2. Adds a method to the MutableUserState interface to add a role to a user
3. Adds RoleEntityAddedApplier

## Related issues

closes #22923


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbRoleState.java
```diff
@@ -68,6 +69,20 @@ public void createRole(final RoleRecord roleRecord) {
...
+    roleKey.wrapLong(roleRecord.getRoleKey());
+    entityKey.wrapLong(roleRecord.getEntityKey());
+    entityTypeValue.setEntityType(roleRecord.getEntityType());
+    entityTypeByRoleColumnFamily.insert(fkRoleKeyAndEntityKey, entityTypeValue);
+
+    // update role record
+    final var persistedRole = roleColumnFamily.get(roleKey);
+    persistedRole.getRole().setEntityKey(roleRecord.getEntityKey());
+    persistedRole.getRole().setEntityType(roleRecord.getEntityType());
+    roleColumnFamily.update(roleKey, persistedRole);
```
nicpuppa, line 83: 💭 I'm not sure if this update is needed, it might be useful

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/RoleRecord.java
```diff
@@ -18,6 +20,9 @@ public class RoleRecord extends UnifiedRecordValue implements RoleRecordValue {
   private final LongProperty roleKeyProp = new LongProperty("roleKey", -1L);
   private final StringProperty nameProp = new StringProperty("name", "");
   private final LongProperty entityKeyProp = new LongProperty("entityKey", -1L);
+  private final EnumProperty<EntityType> entityTypeProp =
+      new EnumProperty<>("entityType", EntityType.class, EntityType.USER);
+  private final LongProperty userKeyProp = new LongProperty("userKey", -1L);
```
remcowesterhoud, line 25: ❌ We don't need a `uerKeyProp` this is what the `entityKeyProp` is for 🙂

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/RoleRecord.java
```diff
@@ -18,6 +20,9 @@ public class RoleRecord extends UnifiedRecordValue implements RoleRecordValue {
   private final LongProperty roleKeyProp = new LongProperty("roleKey", -1L);
   private final StringProperty nameProp = new StringProperty("name", "");
   private final LongProperty entityKeyProp = new LongProperty("entityKey", -1L);
+  private final EnumProperty<EntityType> entityTypeProp =
+      new EnumProperty<>("entityType", EntityType.class, EntityType.USER);
```
remcowesterhoud, line 24: 🔧 I think it'd be good to have an `EntityType.UNSPECIFIED` we can use as default, so we don't accidentally store entities with the wrong entity type.

remcowesterhoud: ❌ We must declare this property in the constructor and in the wrap method.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbRoleState.java
remcowesterhoud: ❌ We don't lets remove it. I already mentioned this in the other PR you created but now I'm certain we should change it. Let's define the `roleKey` and the `roleName` in the `PersistedRole` with properties, instead of one property with the `RoleRecord`. That way we don't store the `entityKey` and the `entityType` in there.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbRoleState.java
```diff
@@ -82,4 +96,11 @@ public long getRoleKeyByName(final String roleName) {
     final var fkRoleKey = roleByNameColumnFamily.get(this.roleName);
     return fkRoleKey != null ? fkRoleKey.inner().getValue() : -1;
   }
+
+  @Override
+  public EntityType getEntityType(final long roleKey, final long entityKey) {
+    this.roleKey.wrapLong(roleKey);
+    this.entityKey.wrapLong(entityKey);
+    return entityTypeByRoleColumnFamily.get(fkRoleKeyAndEntityKey).getEntityType();
```
remcowesterhoud, line 104: ❌ This could result in an NPE. Let's return an `Optional<EntityType>` in this method

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/user/UserRecord.java
```diff
@@ -27,15 +27,17 @@ public final class UserRecord extends UnifiedRecordValue implements UserRecordVa
   private final StringProperty passwordProp = new StringProperty("password", "");
   private final EnumProperty<UserType> userTypeProp =
       new EnumProperty<>("userType", UserType.class, UserType.REGULAR);
+  private final LongProperty roleKeyProp = new LongProperty("roleKey", -1L);
```
remcowesterhoud, line 30: ❌ A user can have multiple roles assigned, so we should make this an `ArrayProperty`

remcowesterhoud (changes requested):
Nice work @nicpuppa 💪  It seems I didn't explain it enough during the onboarding, but a user can have multiple roles. So some changes are needed.

remcowesterhoud (approved):
LGTM 🏆

Issue #23163, opened 2024-10-08
Opensearch client for schema manager


EuroLew:
**Description**

The schema manager needs to support Opensearch, and this will involve implementing the `SearchEngineClient` interface which contains all the actions required to create/update schema resources.

**Success Criteria:**
- [ ] Can create indices in Opensearch given an index descriptor and settings.
- [ ] Can create/update index templates given a descriptor and settings.
- [ ] Can update mappings of indices and index templates.
- [ ] Can get mapping of indices and templates.
- [ ] Can update settings of indices/templates
- [ ] Can create index state management resources (Opensearch equivalent of Elasticsearch index lifecycle management).

PR #23378, opened 2024-10-11
Refactoring SearchEngineClient to be more generic.

EuroLew:
## Description

The opensearch client has some limitations in that you cannot build requests using json, instead you must pass in implementations of `Settings` or `Mappings`. 

In addition there is shared functionality that will also be used by the opensearch client so functions such as `appendToFileSchemaSettings` has been moved to the interface.

**Success Criteria:**
- [x] Common functionality such as `appendToFileSchemaSettings`, `listIndices` are accessible by any `SearchEngineClient` implementation.
- [x] Any requests built using json such as PUT settings and PUT Mapping now instead pass an implementation specific object to the request builders as opensearch does not support request building with json.
- [x] Any builders or functions which create objects required by requests are accessible by any `SearchEngineClient` implementation, in this case `mapToSettings` and `toPropertiesmap`  

## Related issues

Relates to #23163

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
@@ -22,13 +22,13 @@ import org.apache.commons.io.IOUtils;
 
 public record IndexMappingProperty(String name, Object typeDefinition) {
   private static final ObjectMapper MAPPER = new ObjectMapper();
-  private static final JsonpMapper jsonpMapper = new JacksonJsonpMapper(MAPPER);
+  private static final JsonpMapper JSONP_MAPPER = new JacksonJsonpMapper(MAPPER);
 
   public Entry<String, Property> toElasticsearchProperty() {
     try {
       final var typeDefinitionParser = getTypeDefinitionParser();
       final var elasticsearchProperty =
-          Property._DESERIALIZER.deserialize(typeDefinitionParser, jsonpMapper);
+          Property._DESERIALIZER.deserialize(typeDefinitionParser, JSONP_MAPPER);
 
       return entry(name(), elasticsearchProperty);
     } catch (final IOException e) {
@@ -44,7 +44,7 @@ public record IndexMappingProperty(String name, Object typeDefinition) {
     final var typeDefinitionJson =
         IOUtils.toInputStream(MAPPER.writeValueAsString(typeDefinition()), StandardCharsets.UTF_8);
 
-    return jsonpMapper.jsonProvider().createParser(typeDefinitionJson);
+    return JSONP_MAPPER.jsonProvider().createParser(typeDefinitionJson);
   }
 
   public static IndexMappingProperty createIndexMappingProperty(
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
@@ -9,39 +9,42 @@ package io.camunda.exporter.schema;
 
 import static java.util.Map.entry;
 
+import co.elastic.clients.elasticsearch._types.mapping.Property;
+import co.elastic.clients.json.JsonpMapper;
+import co.elastic.clients.json.jackson.JacksonJsonpMapper;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.InputStream;
+import jakarta.json.stream.JsonParser;
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collection;
-import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 import org.apache.commons.io.IOUtils;
 
 public record IndexMappingProperty(String name, Object typeDefinition) {
+  private static final ObjectMapper MAPPER = new ObjectMapper();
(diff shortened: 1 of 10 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
```diff
@@ -7,32 +7,40 @@
...
+  public static <T> Map<String, T> toPropertiesMap(
+      final Set<IndexMappingProperty> properties,
+      final ObjectMapper mapper,
+      final Function<InputStream, T> typeDeserializer) {
+
+    return properties.stream()
+        .map(
+            prop -> {
+              try {
+                final var typeJson =
```
mustafadagher, line 34: ❓ where did these variable names (typeJson, type) come from? what do we mean by type here? feels confusing to me

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
```diff
@@ -7,32 +7,40 @@
...
+                    IOUtils.toInputStream(
+                        mapper.writeValueAsString(prop.typeDefinition()), StandardCharsets.UTF_8);
+                final var type = typeDeserializer.apply(typeJson);
+                return entry(prop.name(), type);
+              } catch (final JsonProcessingException e) {
+                throw new RuntimeException(e);
```
mustafadagher, line 40: ❌ Should we throw an exporter specific Exception, with a proper message? 

❓ How are we handling exceptions in such cases too? this now will fail the schema create or the indexing, right? is that an accepted behavior?

mustafadagher (changes requested):
Thanks @EuroLew, Good work 🚀 

I've left a few comments with possible improvements

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -7,14 +7,43 @@
...
 public interface SchemaManager {
   void initialiseResources();
 
   void updateSchema(final Map<IndexDescriptor, Set<IndexMappingProperty>> newFields);
 
-  IndexMapping readIndex(final IndexDescriptor indexDescriptor);
+  @SuppressWarnings("unchecked")
+  static IndexMapping readIndex(final IndexDescriptor indexDescriptor, final ObjectMapper mapper) {
+    try (final var mappingsStream =
+        SchemaManager.class.getResourceAsStream(indexDescriptor.getMappingsClasspathFilename())) {
```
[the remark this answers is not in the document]
EuroLew: I don't know why but if we preface the files with a `/` everything works fine, e.g. `/mappings.json`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
EuroLew: This is actually out of date, I made the serialisation generic, it was changed here https://github.com/camunda/camunda/pull/23393/files, It would be easier to review that separately, in the mean time I will rebase it in to here.

EuroLew: This has been refactored into `IndexMappingProperty.toElasticsearchProperty` (with the opensearch implementation to be created in the next PR) which converts an IndexMappingProperty back into a elasticsearch index mapping property. 

The new variable names and function names should be a lot clearer

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
EuroLew: This now throws an IllegalStateException given it is a class method of `IndexMappingProperty` it is ok for us to throw an exception as it will crash the exporter because the schema update failed, this is accepted behaviour.

mustafadagher: Yes, we shall throw an exception if that's expected, but shouldn't we throw `ExporterException` with a proper message then? I just don't like the plain `RuntimeException`.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -7,14 +7,43 @@
...
+import java.util.stream.Collectors;
 
 public interface SchemaManager {
   void initialiseResources();
 
   void updateSchema(final Map<IndexDescriptor, Set<IndexMappingProperty>> newFields);
 
-  IndexMapping readIndex(final IndexDescriptor indexDescriptor);
+  @SuppressWarnings("unchecked")
+  static IndexMapping readIndex(final IndexDescriptor indexDescriptor, final ObjectMapper mapper) {
```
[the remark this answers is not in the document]
mustafadagher: May I hear your thoughts about this point?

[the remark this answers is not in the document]
mustafadagher: sorry, seems like you addressed it. pull was failing locally probably due to rebase.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/IndexMappingProperty.java
EuroLew: I've changed this now so it's IllegalStateException since it's an instance method on IndexMappingProperty which I think is a more valid and granular than exporterexception

mustafadagher (approved):
LGTM right now. Thanks Euro!

PR #23385, opened 2024-10-11
feat: extend user task sort fields

nathansandi:
## Description

Include the follow options for sort on user task
- dueDate
- folloUpDate
- priority

Fix
- rename `tenantIds `to `tenantId `on Search User Task request/response

In order to cover some Tasklist OOTB requirements


ralfpuchert (approved):
LGTM!

clients/java/src/main/java/io/camunda/zeebe/client/impl/search/filter/UserTaskFilterImpl.java
```diff
@@ -84,7 +84,7 @@ public UserTaskFilter processInstanceKey(final Long processInstanceKey) {
 
   @Override
   public UserTaskFilter tentantId(final String tenantId) {
```
houssain-barouni, line 86: this is not related to this PR by `tentantId` typo needs to be fixed

houssain-barouni (changes requested):
You need also to update the request mapper in https://github.com/camunda/camunda/blob/df399a3f3420f180b299660a5befa7bcd7fb79ea/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java#L616-L630

clients/java/src/main/java/io/camunda/zeebe/client/impl/search/filter/UserTaskFilterImpl.java
nathansandi: I had included on the PR description

" Fix - rename tenantIds to tenantId on Search User Task request/response"

I can either rename the PR title or create a new one for that

But I dont believe is necessary a new PR to apply this fix.

nathansandi:
Thank you @houssain-barouni , just fixed :)

houssain-barouni (approved):

PR #23388, opened 2024-10-11
feat: Harmonize Process and ProcessDefinition interfaces

### LINKED ISSUE #23120: Harmonize `Process` and `ProcessDefinition` interfaces in `io.camunda.zeebe.client.api.response`
### Description

[Process](https://github.com/camunda/camunda/blob/53af5bb30977b0939a2ca27971e10273c3f0b136/clients/java/src/main/java/io/camunda/zeebe/client/api/response/Process.java#L18-L18) and `ProcessDefinition` interfaces in `io.camunda.zeebe.client.api.response` describe the same entity but with slightly different fields.  

### TODO

- [ ] Use only one interface in client 

### Additional Information

### Acceptance Criteria

ghost:
feat: Harmonize Process and ProcessDefinition interfaces

- merge interfaces `Process` and `ProcessDefinition` into a single interface - `ProcessDefinition`
- replaced field `bpmnProcessId` with `processDefinitionId`

Closes https://github.com/camunda/camunda/issues/23120


clients/java/revapi.json
```diff
@@ -91,6 +91,19 @@
           "code": "java.class.removed",
           "old": "class io.camunda.zeebe.client.protocol.rest.UserTaskUpdateRequestChangeset",
           "justification": "The classes/methods are only internal and are not user-facing, and as such does not need to maintain any backwards compatibility guarantees."
+        },
+        {
+          "ignore": true,
+          "code": "java.method.returnTypeTypeParametersChanged",
+          "old": "method java.util.List<io.camunda.zeebe.client.api.response.Process> io.camunda.zeebe.client.api.response.DeploymentEvent::getProcesses()",
+          "new": "method java.util.List<io.camunda.zeebe.client.api.response.ProcessDefinition> io.camunda.zeebe.client.api.response.DeploymentEvent::getProcesses()",
+          "justification": "The classes/methods are only internal and are not user-facing, and as such does not need to maintain any backwards compatibility guarantees."
```
tmetzke, line 100: 💭 Unfortunately, this is not true. This IS the result of a client command and definitely user-facing. This breaks user space. Anyone using the result of a "deploy resources" client command will have to adjust their code with this change.

tmetzke (commented):
👍 Thanks, @mihail-ca, for tackling this.
⚠️ This change will break users' code. While that might be OK, we still need to consider it and make a conscious decision here and not take it lightly.
💭 We could also have the `ProcessDefinition` as a search result extend the `Process` interface. Otherwise, we'll now also have a result of the "deploy resources" client command that always contains `null` for `name` and `versionTag`. Maybe we either do this E2E and return those two attributes upon deploying resources as well (slightly bigger change) or have one interface per command because they simply have different results.

Happy for any other thoughts here, of course.

clients/java/revapi.json
mihail-ca: Hello @tmetzke does this ticket still make sense ? As you said, one option is _to have one interface per command because they simply have different results_. which I assume, is the current implementation (interfaces Process and ProcessDefinition). So do we still need to change ? Thank you.

tmetzke: Yeah, that is generally true. I think we can also simply not do this change at all 👍

Issue #23040, opened 2024-10-03
Move Tasklist entities to the webapps-schemas


panagiotisgts:
### Description
Migrate the existing `tasklist-form` and adapt existing UserTask indices to accomodate Variables from Tasklist's schema defintion to the common WebApps schema module .

### TODO
- [x] `tasklist-task`
- [x] `tasklist-form`
- [x] `tasklist-metric`

### Additional Information

The `tasklist-form` index has properties related to, now retired, Job Worker User Tasks. The fields `processDefintiionId` and `isEmbedded` are irrelevant for Zeebe User Tasks.

The current `tasklist-task` index will be enhanced in the following structure, similar to the previous `tasklist-list-view` and it will also include _processVariables, taskVariables_ and _process_ information.

<b style="font-weight:normal;" id="docs-internal-guid-076c12db-7fff-1443-d6ae-86ecf7c2d64b"><div dir="ltr" style="margin-left:0pt;" align="center">
 New | Field | Data Type | Meta | Notes
-- | -- | -- | -- | --
No | formId | keyword |  |  Maintained from tasklist-task (deprecated)
No | formVersion | integer |  |  Maintained from tasklist-task (deprecated)
No | isFormEmbedded | boolean |  |  Maintained from tasklist-task (deprecated)
 :heavy_check_mark:   | action | keyword |   | Will not migrate data  from operate-user-task
No | assignee | keyword |   |  
No | bpmnProcessId | keyword |   |  
No | candidateGroups | keyword |   |  
No | candidateUsers | keyword |   |  
:heavy_check_mark: | changedAttributes | text |   | Will not migrate data  from operate-user-task
No | dueDate | date | "format": "date_time \|\| epoch_millis" |  
No | flowNodeBpmnId | keyword |   |  
No | flowNodeInstanceId | keyword |   |  
No | externalFormReference | keyword |   |  
No | followUpDate | date | "format": "date_time \|\| epoch_millis" |  
No | formKey | keyword |   |  
No | id | keyword |   | Same as flowNodeInstanceId
No | key | long |   | Refers to variableKey, userTaskKey, processKey
No | partitionId | integer |   |  
No | position | long |   |  
No | priority | integer |   |  
No | processDefinitionId | keyword |   |  
No | processDefinitionVersion | integer |   |  
No | processInstanceId | keyword |   |  
No | tenantId | keyword |   |  
:heavy_check_mark:  | name | keyword |   | variables
[...]

PR #23393, opened 2024-10-11
getMappings now retrieves all property fields.

EuroLew:
## Description

Previously when serialising mappings from Elasticsearch(ELS) only the `type` field would be captured.

This cause an issue where:
- If there is a schema file with complex property fields e.g. type, index, format etc... This resource would be correctly created in ELS with all the property fields. 
- When retrieving mappings and converting to an implementation agnostic object we would only serialize the `type` field missing the other fields this would cause an error during schema validation as it thinks the fields are different. 
  - As an example `[name=creationTime, leftValue=IndexMappingProperty[name=creationTime, typeDefinition={format=date_time || epoch_millis, type=date}], rightValue=IndexMappingProperty[name=creationTime, typeDefinition={type=date}]],` here we see creation time has 2 fields in its properties but for the `rightValue` only the `type` value was serialised.

In addition there is an error case where elasticsearch recognises a setting say `"index": "false"` as a boolean false and in the index sets the value to false. Then when comparing the schema to the state in the database we get a difference and validation fails as `"index": "false"`(in json file) != `"index": false`(in els database). This error should not occur as long as fields are given their correct type but if it does the error should be recognised, the default tostring method hides this mismatch so a more detailed one is to be implemented.

**Success Criteria:**
- [x] For a property field with multiple fields `getMappings` will capture all fields.
- [x] More descriptor IndexMappingProperty toString 

## Related issues

relates to  #23040

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/SchemaResourceSerializer.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/SchemaResourceSerializer.java
@@ -11,7 +11,6 @@ import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import io.camunda.exporter.exceptions.ElasticsearchExporterException;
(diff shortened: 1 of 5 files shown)


panagiotisgts (approved):
LGTM! Tried the scenario with the `JOIN` field and the diff aswell, works as expected. :100:

PR #23457, opened 2024-10-14
test: fix qa generation naming for 8.6

lenaschoenburg:
The naming schema changed from "Camunda 8.6.0" to "Camunda 8.6-gen1". This fix is not very smart, we'll have to redo it until all supported minor versions use the new naming scheme.

See https://camunda.slack.com/archives/CKZK2E7RP/p1728898155269379?thread_ts=1728897908.564069&cid=CKZK2E7RP


lenaschoenburg:
That fix alone seems to be not enough, the process fails in the mapping names to uuid task now.

lenaschoenburg:
Nevermind, I just needed to add the generation to the right channel.

berkaycanbc (approved):
LGTM 👍 

Our daily QA on minor release is failing every day due to this issue and requires manual fix every day. Approving it to unblock QA process.

PR #23439, opened 2024-10-14
Add exporter handler for Decision records with integration tests

deepthidevaki:
## Description

* Move `DecisionHandler` from branch `feat-operate-exporter` to `camunda-exporter` in `main`.
* Refactored added unit test for the handler. Mostly removed the mocks and instead used the protocol factory to generate the records. 
* Added an integration test for Decision records in `qa/integration-tests`. The test can be easily extended to run for Opensearch when it is fully supported. This test serve as a reference for other integration tests. 

ps:- the index prefix has been changed to "operate" because this is hard-corded in the rest api. This should be updated based on the decision from [ongoing discussion](https://camunda.slack.com/archives/C06F0GLJNFM/p1728568094544759)

## Related issues
 
related to  https://github.com/camunda/camunda/issues/22642

PR #23463, opened 2024-10-14
Exporter/migrate existing identity exporter entities to common schema

Ben-Sheppard:
## Description

With the addition of the new webapps-schema module to manage the schemas of our entities in ES/OS, the existing implementation of the `Authorization` and `User` entities needed to move. As part of this refactor the existing Identity based entities lost some access to the Zeebe housed enums so I also needed to make some changes there. Essentially this PR:

1. Moves the `Authorization` and `User` entity to `webapps-schema`
2. Adds in the schema definitions for both entities
3. Updates the entities given the restriction on dependencies
4. Updates the `Authorization` entity to better support searching (see below)

### Authorization entity update
When looking to implement support for adding a filter to retrieve based on permission (i.e. `CREATE`, `READ`) I realised that the existing structure was awkward to work with as we had dynamic keys. Looking for inspiration online lead me to see that a slight change with named keys would help us search in a clearer way. The changes in this PR to the `Authorization.permission` object results in a record in ES/OS with the following structure:

```
               "_index": "authorizations",
               ...,
                "_source": {
                    ...,
                    "permissions": [
                        {
                            "type": "CREATE",
                            "resourceIds": [
                                "foo"
                            ]
                        },
                        {
                            "type": "READ",
                            "resourceIds": [
                                "bar",
                                "baz"
                            ]
                        }
                    ]
                }
         }
```

this then means that our filters for searching by permission can effectively be `permissions.type === term("CREATE")` 

Questions I have:

1. Do I need template files for anything? I could derive their use within my context
2. Should I add to the `descriptors` package too?


mustafadagher:
Hi @Ben-Sheppard, I didn't get a chance to fully review this PR yet. But for migrating handlers, I'd suggest you take this PR https://github.com/camunda/camunda/pull/23439 as a reference (also in the kind of tests to include for each handler).

I'll give it another look when I have more time, but can't promise this week.

webapps-schema/src/main/resources/schema/elasticsearch/create/index/authorization.json
```diff
@@ -0,0 +1,29 @@
+{
```
Zelldon, line 1: TBH I have no enough knowledge to judge this. But from what I can tell looks ok to me.

ChrisKujawa (approved):
Looks good to me @Ben-Sheppard 

> 1. Moves the Authorization and User entity to webapps-schema
> 2. Adds in the schema definitions for both entities
> 3. Updates the entities given the restriction on dependencies

These parts look good to me.

> 4. Updates the Authorization entity to better support searching (see below)

Here I can't really judge, but makes sense to me. How was it before? A comparison would help here :) 

Regarding

> 1. Do I need template files for anything? I could derive their use within my context

No. Operate uses template files for indices that need historic clean up afaik. So you don't need this.


> 2. Should I add to the descriptors package too?

Not sure for what this is. Maybe @sdorokhova or @EuroLew can tell more about this.

Ben-Sheppard:
Thank you @mustafadagher and @Zelldon. @Zelldon in reference to your question regarding a comparison structure the old format meant the records were stored as such
```
                    "permissions": [
                        {
                            "CREATE": [
                                "foo"
                            ]
                        },
                        {
                            "READ": [
                                "bar",
                                "baz"
                            ]
                        }
                    ]
```

Which whilst is still possible to work with, made it a bit more involved as we would need to try and target `READ` or `CREATE` for example in the permissions. The new structure
```
                    "permissions": [
                        {
                            "type": "CREATE",
                            "resourceIds": [
                                "foo"
                            ]
                        },
                        {
                            "type": "READ",
                            "resourceIds": [
                                "bar",
                                "baz"
                            ]
                        }
                    ]
```
means we can target specific fields i.e. `permissions.type` and check a value

Issue #23365, opened 2024-10-11
Add client integration tests for API endpoints


ralfpuchert:
### Description

Some API endpoints need client integration tests.

### TODO

- [x] Process definition: search, get: https://github.com/camunda/camunda/pull/23461
- [x] Flow node instance: search, get: 
      https://github.com/camunda/camunda/pull/23525
      https://github.com/camunda/camunda/pull/23488
- [x] Incident: search, get: https://github.com/camunda/camunda/pull/23567
- [x] User task: search, get

Issue #22850, opened 2024-09-30
Solution design for deleting historic process resources via REST API


tmetzke:
# Introduction

The C8 REST API must support the deletion of history data. This includes the following:
 
* Resources like process definitions and all related data, such as finished process instances and their variables
* Specific process instances and all related data, such as their variables

# Concept

* Through C8 REST endpoints, I can do the following:
  * Delete history data related to a single process instance identified by a key
  * Delete history data related to process instances that fulfill certain filter criteria
  * Delete history data related to a single resource identified by a key, such as a process definition (e.g., extend [this endpoint](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/delete-resource/) to also support history deletion optionally or create a new endpoint)
  * Delete history data related to resources, such as process definitions, that fulfill certain filter criteria
* All endpoints must also consider related data, such as all process instances for a process definition and all variables of a process instance.
* Running process instances must have been canceled ([REST API](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/cancel-process-instance/)) or completed (by bringing the instance to a defined end) before being deleted.
* The data is deleted asynchronously in the background through some batching mechanism, so the REST requests return within an appropriate time.
* All endpoints must return a unique identifier of the background (batch) operation that deletes the data. This allows for querying the operation's status with subsequent calls to another dedicated query endpoint.
* The REST gateway issues one command to the brokers containing the data needed to identify and delete the requested data.
* The broker command creates the required background (batch) operation to delete the data in chunks and returns with the unique identifier of that operation.
* The background operation handles the querying and deleting of all related data in chunks asynchronously.
* The background operation is observable on request, yielding its current status (e.g., in progress, failed, completed).

## Deletion "algorithm"

* As we are deleting data chunk by chunk and the data is related to each other, we should follow a clear deletion order that leaves as few unexpected inconsistent data views to users as possible.
* For example, it could make sense to delete "leaf" elements first, e.g., children of a process instance that are unrelated to other data or not parents for other data points themselves, such as variables.
* The deletion strategy could also differ per database type, e.g., deleting leaf entities first by key in RDBMS and deleting by query in ES/OS.
  * For the batch processor, this might change the number of follow-up batch executions to run and the keys contained in there.
[...]

PR #23488, opened 2024-10-14
feat: add integration tests for flow node instance client api

ralfpuchert:
## Description

Get flow node instance method is missing in client and will be provided in follow up PR.

- Add integration tests for flow node instance client api
- Use processDefinitionId instead of bpmnProcessId
- Use flowNodeInstanceKey instead of key

After review hints:
- Refactor flow node instance integration tests into `ProcessInstanceAndFlowNodeInstanceQueryTest`
- Factor out utility methods into `QueryTest`

## Related issues

Related to #23365


qa/integration-tests/src/test/java/io/camunda/it/client/FlowNodeInstanceQueryTest.java
```diff
@@ -0,0 +1,492 @@
...
+        .ignoreExceptions() // Ignore exceptions and continue retrying
+        .untilAsserted(
+            () -> {
+              final var result = zeebeClient.newProcessInstanceQuery().send().join();
+              assertThat(result.items().size()).isEqualTo(expectedProcessInstances);
+            });
+  }
+
+  private static void waitForProcessInstancesToExecute() throws InterruptedException {
+    Thread.sleep(15000);
```
nathansandi, line 464: ❓  Is it possible use Awaitility to check test conditions? 
`Awaitility.await().atMost(Duration.ofSeconds(15)).until(() -> checkCondition());` as it used in other integration tests?

ralfpuchert: Yes, thanks. Copy-paste error.

nathansandi (approved):
LGTM

houssain-barouni (commented):
One comment regarding the location of the integration tests

qa/integration-tests/src/test/java/io/camunda/it/client/FlowNodeInstanceQueryTest.java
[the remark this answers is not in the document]
ralfpuchert: Ok, then I would also rename this.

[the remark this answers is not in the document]
ralfpuchert: Should we move other tests, from ProcessDefinitionQueryTest etc , also? Then I propose to create a follow up PR for that (except the tests here).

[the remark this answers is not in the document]
houssain-barouni: I don t see any `ProcessDefinitionQueryTest`, there is only `ProcessQueryTest` that contains ProcessDefinition/ProcessInstance tests

ralfpuchert:
@tmetzke Can you review the part for `IncidentQueryTest`: https://github.com/camunda/camunda/pull/23488/commits/0b367acf539b88e9aa46fa4864cf6c6b8a8a4bdb?

I refactored the code in this PR already so this should also solve https://github.com/camunda/camunda/issues/22850.
The preconditions are now in [QueryTest](https://github.com/camunda/camunda/pull/23488/commits/3f3a1e927bc6d4ceb578d9faa8ca95a6695666f4#diff-dd0714b02dd533ec2dc75cc68a139a9c58f91dbfa2df8e446aa2efb79277c622) and should same setups for the tests.

tmetzke:
> @tmetzke Can you review the part for `IncidentQueryTest`: [0b367ac](https://github.com/camunda/camunda/commit/0b367acf539b88e9aa46fa4864cf6c6b8a8a4bdb)?
> 
> I refactored the code in this PR already so this should also solve #22850. The preconditions are now in [QueryTest](https://github.com/camunda/camunda/pull/23488/commits/3f3a1e927bc6d4ceb578d9faa8ca95a6695666f4#diff-dd0714b02dd533ec2dc75cc68a139a9c58f91dbfa2df8e446aa2efb79277c622) and should ensure same setups for the tests.

Did you really mean #22850 or rather https://github.com/camunda/camunda/issues/23706? 😅

ralfpuchert:
Oh sorry, I meant #23706.

qa/integration-tests/src/test/java/io/camunda/it/client/QueryTest.java
```diff
@@ -0,0 +1,128 @@
...
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import io.camunda.zeebe.client.api.search.response.SearchQueryResponse;
+import java.time.Duration;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Function;
+import org.awaitility.Awaitility;
+
+abstract class QueryTest {
```
tmetzke, line 24: 🔧 We use this as a utility class (static methods only). Let's make it one then instead of it being an abstract class that should be implemented. Did you switch between making this a base class the other query ITs should extend and this being a utility class instead, maybe? 😅 
I would stick to a utility class like you already use 👍

tmetzke (approved):
👍 Looks good, thanks for looking into the incident flakiness and refactoring the ITs.
🔧 There is a small improvement idea, nothing blocking though 🙂

qa/integration-tests/src/test/java/io/camunda/it/client/QueryTest.java
ralfpuchert: Thanks. Will use it as utility class.

PR #23489, opened 2024-10-14
build: use reflective equals and hashCode in Optimize

buccarel:
## Description

This is a prototype: the implementation is currently complete and the build is green, but it's missing unit tests.

With the de-lombok'ification of the monorepo, the removal of the annotation `@EqualsAndHashCode` generated a lot of code. Such code, not only is big in terms of number of lines, but it's also static (e.g., if we add a variable to a class that implements `hashCode()`, we also need to change the implementation of such method to also include the new variable).

With this prototype, I am using an implementation of equals and hashCode that works reflectively when applied to any class, with any variables in it. It's provided by an apache library and it is my understanding that it doesn't need the "call-super" special case, when used reflectively.

This PR has been generated programmatically with a script that replaces the equals() and hashCode() methods, when hashCode() is present in the class.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


optimize/util/optimize-commons/src/main/java/io/camunda/optimize/util/DynamicEqualsAndHashCode.java
```diff
@@ -0,0 +1,130 @@
...
+import io.camunda.optimize.service.exceptions.OptimizeRuntimeException;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.Objects;
+
+public class DynamicEqualsAndHashCode {
+
+  public static boolean computeEquals(Object object1, Object object2) {
+    // case 1: at least 1 object is null
```
RomanJRW, line 19: this would already be covered by case 2?

matthewBearCamunda (changes requested):
The code changes look good to me, I think the approach of using the library from apache commons is a cleaner and simpler way to go

Before I give the green approval check, would you be able to test the performance of this change on a large dataset? The use of reflection in both the hash code and equals method makes me think that a test would be worthwhile

matthewBearCamunda (approved):
Approving! The performance test was helpful, and the performance hit we would take would be acceptable

https://camunda.slack.com/archives/CUZUYDMMG/p1729258751439819?thread_ts=1729082399.343169&cid=CUZUYDMMG

optimize/upgrade/pom.xml
```diff
@@ -188,6 +188,10 @@
       <version>2.12.0</version>
       <scope>runtime</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
```
korthout, line 194: ❌ This change broke building with `mvnd` (Maven Daemon)

```
mvnd clean install -Dquickly -Dskip.fe.build -Dspotless.skip
[INFO] Processing build on daemon 60b1894c
[INFO] Scanning for projects...
[ERROR] Some problems were encountered while processing the POMs
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project io.camunda.optimize:upgrade-optimize:8.7.0-SNAPSHOT (/Users/korthout/gh/camunda/camunda/optimize/upgrade/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.apache.commons:commons-lang3:jar -> version 3.14.0 vs (?) @ line 191, column 5
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
```

@buccarel Can you revert this change?

PR #23494, opened 2024-10-14
Opensearch client for camunda exporter

### LINKED ISSUE #23163: Opensearch client for schema manager

EuroLew:
**Description**

The schema manager needs to support Opensearch, and this will involve implementing the `SearchEngineClient` interface which contains all the actions required to create/update schema resources.

**Success Criteria:**
- [x] Can create indices in Opensearch given an index descriptor and settings.
- [x] Can create/update index templates given a descriptor and settings.
- [x] Can update mappings of indices and index templates.
- [x] Can get mapping of indices and templates.
- [x] Can update settings of indices/templates
- [x] Can create index state management resources (Opensearch equivalent of Elasticsearch index lifecycle management).

closes #23163

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/OpensearchEngineClient.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/OpensearchEngineClient.java
@@ -18,11 +18,14 @@ import java.io.InputStream;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
 import org.opensearch.client.json.JsonpDeserializer;
 import org.opensearch.client.json.jsonb.JsonbJsonpMapper;
 import org.opensearch.client.opensearch.OpenSearchClient;
 import org.opensearch.client.opensearch.indices.CreateIndexRequest;
 import org.opensearch.client.opensearch.indices.PutIndexTemplateRequest;
+import org.opensearch.client.opensearch.indices.PutMappingRequest;
 import org.opensearch.client.opensearch.indices.put_index_template.IndexTemplateMapping;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -82,8 +85,19 @@ public class OpensearchEngineClient implements SearchEngineClient {
 
   @Override
   public void putMapping(
-      final IndexDescriptor indexDescriptor,
-      final Collection<IndexMappingProperty> newProperties) {}
+      final IndexDescriptor indexDescriptor, final Collection<IndexMappingProperty> newProperties) {
+    final PutMappingRequest request = putMappingRequest(indexDescriptor, newProperties);
+
+    try {
+      client.indices().putMapping(request);
+      LOG.debug("Mapping in [{}] was successfully updated", indexDescriptor.getIndexName());
+    } catch (final IOException e) {
+      final var errMsg =
+          String.format("Mapping in [%s] was NOT updated", indexDescriptor.getIndexName());
+      LOG.error(errMsg, e);
+      throw new OpensearchExporterException(errMsg, e);
+    }
+  }
 
   @Override
   public Map<String, IndexMapping> getMappings(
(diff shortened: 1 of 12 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/TestSupport.java
```diff
@@ -18,8 +19,27 @@ public final class TestSupport {
...
+   *
+   * <p>The container is configured to use 512m of heap and 512m of direct memory. This is required
+   * because OpenSearch, by default, will grab all the RAM available otherwise.
+   *
+   * <p>Additionally, security is explicitly disabled to avoid having tons of warning printed out.
+   */
+  @SuppressWarnings("resource")
+  public static OpensearchContainer<?> createDefaultOpensearchContainer() {
+    return new OpensearchContainer<>(OPENSEARCH_IMAGE)
+        .withEnv("OPENSEARCH_JAVA_OPTS", "-Xms256m -Xmx512m -XX:MaxDirectMemorySize=536870912")
```
mustafadagher, line 39: 🔧 the comment above says `The container is configured to use 512m of heap and 512m of direct memory` while here I see 256mb 😄

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/OpensearchEngineClient.java
```diff
@@ -150,7 +158,53 @@ public void putSettings(
   }
 
   @Override
-  public void putIndexLifeCyclePolicy(final String policyName, final String deletionMinAge) {}
+  public void putIndexLifeCyclePolicy(final String policyName, final String deletionMinAge) {
+    final var request = createIndexStateManagementPolicy(policyName, deletionMinAge);
+
+    try (final var response = client.generic().execute(request)) {
```
mustafadagher, line 164: 💭 I think this is a valid way around executing raw REST requests here. 
I know that operate had an `OpenSearchISMOperations` implementation with various methods to deal with ISM, would we need to port this here instead at some point?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/OpensearchEngineClient.java
mustafadagher, line 1: ❌ I know all opensearch client operations throw `OpenSearchException` too (next to the `IOException`), shouldn't we also catch this in the current catch blocks?

PS: I think we missed this for the elasticsearch client as well.

mustafadagher (approved):
Great job Euro, Thank you!

Nothing really big in the comments, so I'll pre-approve, but please address the comments before moving forward.

mustafadagher:
Also, one final comment:
🔧  let's also use the client in the `OpensearchAdapter` and remove the FIXME and TODOs around it.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/OpensearchEngineClient.java
```diff
@@ -0,0 +1,344 @@
...
+import org.opensearch.client.opensearch.indices.PutMappingRequest;
+import org.opensearch.client.opensearch.indices.get_index_template.IndexTemplateItem;
+import org.opensearch.client.opensearch.indices.put_index_template.IndexTemplateMapping;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OpensearchEngineClient implements SearchEngineClient {
+  private static final Logger LOG = LoggerFactory.getLogger(OpensearchEngineClient.class);
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+  private static final String OPERATE_DELETE_ARCHIVED_POLICY =
```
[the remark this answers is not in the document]
EuroLew: Is there a better place? It will have to be somewhere, I assumed it would always be here unless there's like a constants file.

Issue #23133, opened 2024-10-08
Create an advanced search concept for the REST API


tmetzke:
## Description

* We need advanced querying capabilities for the C8 REST API with 8.7 as outlined in https://github.com/camunda/camunda/issues/22584
* There are two ways to achieve this at least:
  * Build proper Advanced Search (in iterations) into the REST API
  * Build the capabilities specifically for certain fields and have a non-breaking integration concept for Advanced Search for 8.8+

## Acceptance criteria

* We have a concept for Advanced Search in the C8 REST API
* We know whether we integrate it directly in 8.7 (with a first set of supported operations) or 8.8+ in a non-breaking way
```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/23530
```

tmetzke:
## Findings (wip)

* A `oneOf` definition for filter attributes, allowing either a primitive type (integer, string, boolean) or an object (advanced operator notation), would generally be possible in OpenAPI. This would allow adding Advanced Search with 8.8+ on spec level without breaking the 8.7 API.
```yaml
  UserTaskFilterRequest:
      description: User task filter request.
      type: object
      properties:
        userTaskKey:
          oneOf:
            - type: integer
              format: int64
            - $ref: "#/components/schemas/LongFilterProperty"
```
* Such a `oneOf` definition can, as of our current knowledge, **NOT** be generated into functional Java code. The `oneOf` definition leads to an interface type for the attributes. The implementation classes of this interface represent the concrete types the REST controller can receive and deserialize into. For the primitive type, no interface implementation is generated though.
  * We could invest time into finding a way to enable/work around this by configuring the generation or providing a custom type ourselves.
  * We could define all properties as plain `object` types and do validation manually in the controllers. However, this leaves no structured information for API users and generators. But it would allow current shortened equals notation (`"field": "value"`) and operator notation (`"field": { "$eq": "value"}`) for the attributes.
* Creating an Advanced Search notation in the spec could be done with `oneOf` definitions as well, e.g.
```yaml
    LongEqFilter:
      description: Equality filter for Long values.
      type: object
      properties:
        $eq:
          type: integer
          format: int64
    LongGtFilter:
      description: Greater than filter for Long values.
      type: object
      properties:
        $gt:
          type: integer
          format: int64
    LongFilterProperty:
      description: Query filter property for Long values.
      oneOf:
        - $ref: "#/components/schemas/LongEqFilter"
[...]

PR #23530, opened 2024-10-15
feat: advanced search api poc

danielkelemen:
## Description

Proof of concept for the advanced search API proposal.
- OpenAPI specification with complete filter properties and primitive value for Integer, Long and String filters
- `JsonDeserializer`s for implicit equals
- Tests for all filters
- Java client adjustment to new APIs

## Related issues

https://github.com/camunda/camunda/issues/23133


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java
```diff
@@ -264,6 +273,50 @@ public static <C extends Collection<Integer>> SearchQuery intTerms(
     }
   }
 
+  public static List<SearchQuery> intTerms(
```
tmetzke, line 276: 🔧 We can probably rather call those `integerOperations`, `longOperations`, and so on. If I am not mistaken, the word "term" is used for equality checks.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/deserializer/FilterDeserializer.java
```diff
@@ -0,0 +1,26 @@
...
+package io.camunda.zeebe.gateway.rest.deserializer;
+
+import com.fasterxml.jackson.core.TreeNode;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
+
+public abstract class FilterDeserializer<T> extends JsonDeserializer<T> {
+
+  protected static final ObjectMapper OM = new ObjectMapper();
```
tmetzke, line 17: 🔧 For later: I think the mapper needs to come from the implementing class because those will sit in the Spring config classes and inject the mapper. Could just be an instance attribute that is passed in a constructor.

tmetzke (commented):
POC looks really nice already, I like the deserializer solution and how straightforward it is 👍

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/deserializer/FilterDeserializer.java
danielkelemen: One thing is important: we need a mapper that doesn't have the same deserializer registered because that would cause a loop.

tmetzke: OK, I see. Alternative idea:
* We DO use interfaces for `oneOf`.
* We register the custom deserializers on the `IntegerFilter`.
* If it's a primitive, we return an instance of the `AdvancedIntegerFilter` with `$eq` set.
* Otherwise, we use the one general OM to automatically create an `AdvancedIntegerFilter`.

Assumptions:
* The OM with the registered custom deserializers works fine with implementing classes, not running into a loop because it's registered on the interface.
* The interface usage doesn't run into any other issues (but we handle the deserialization ourselves, so maybe that is fine).

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java
danielkelemen: Yes, I agree.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/deserializer/FilterDeserializer.java
danielkelemen: Tried two solutions they're both in the commits and they both could work:

1. Define two object mappers. Register deserializers only with one and pass the basic OM as parameter. It works but would need a cleaner solution with the configuration and due to multiple OMs available.
2. Generate interface and implementation with openapi-generator. Register deserializers for the interfaces only and deserialize into the implementation's class. Inject ObjectMapper and use that for deserializing explicit use-cases automatically. It works, changes are also quite miminal.

Decision: 2nd, unless something pops up in CI.

tmetzke: Both approaches seem to have some issues still, looking at the CI results, unfortunately.

For the first one, I thinks it's mostly the setup in Spring. We already have a definition for the OM in Spring and we should simply extend that one with the code you defined in the new configuration, I believe.

For the interface approach, I am not 100% sure yet what fails there. Debugging this on the server side would be interesting, I guess, to see where exactly this fails.
Here are the server logs for this: 
[Test results for [IT] Root Integration Tests.zip](https://github.com/user-attachments/files/17480018/Test.results.for.IT.Root.Integration.Tests.zip)

(For future reference, you can access them from the failed [workflow run summary](https://github.com/camunda/camunda/actions/runs/11463393731), scrolling to the bottom and downloading from the **Artifacts**)

tmetzke: This is what I can see in the logs there
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `io.camunda.zeebe.gateway.protocol.rest.LongFilter` (no Creators, like default constructor, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 33] (through reference chain: io.camunda.zeebe.gateway.protocol.rest.ProcessInstanceSearchQueryRequest["filter"]->io.camunda.zeebe.gateway.protocol.rest.ProcessInstanceFilterRequest["processInstanceKey"])
	at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:67) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.DeserializationContext.reportBadDefinition(DeserializationContext.java:1888) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.DatabindContext.reportBadDefinition(DatabindContext.java:414) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1375) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.AbstractDeserializer.deserialize(AbstractDeserializer.java:274) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:399) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:185) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:399) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:185) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:342) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.ObjectReader._bindAndClose(ObjectReader.java:2125) ~[jackson-databind-2.18.0.jar:2.18.0]
	at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:1501) ~[jackson-databind-2.18.0.jar:2.18.0]
	at org.springframework.http.converter.json.AbstractJackson2HttpMessageConverter.readJavaType(AbstractJackson2HttpMessageConverter.java:395) ~[spring-web-6.1.13.jar:6.1.13]
```

I think the registration of the deserializers in the setup didn't really work. It looks like they are not picked up, maybe.

danielkelemen: Interesting, the setup is exactly the same as previously without interfaces and that was green. I suspect it's due to the interface & implementation. It does work fine in `ProcessInstanceQueryControllerTest` though. I'll investigate it a bit.

tmetzke: Yeah, I was also confused by that a bit. Let me know if you'd like to pair on this, happy to jump on a call if you need a second pair of eyes.

tmetzke: Ohhh, OK 😄

danielkelemen: For now I took out the `MappingJackson2HttpMessageConverter` from operate. I suspect this will trigger other errors.

db/rdbms/src/main/resources/mapper/ProcessInstanceMapper.xml
```diff
@@ -69,20 +69,53 @@
...
+        &lt;= #{operation.value}
+      </when>
+      <when test="operation.operator.name().equals('IN')">
+        IN <foreach collection="operation.values" item="value" open="(" separator=", " close=")">#{value}</foreach>
+      </when>
+      <when test="operation.operator.name().equals('LIKE')">
+        LIKE #{operation.value}
+      </when>
+    </choose>
+  </sql>
```
tmetzke, line 100: @danielkelemen, have you missed writing mapping files? 😆

danielkelemen: I mean, I kind of like it 😄

tmetzke: Well, you only have to write it once, ideally. Unless you have to write specific mapping files for different databases 😆

PR #23535, opened 2024-10-15
test: Update Tasklist mock data to Zeebe user tasks

vsgoulart:
## Description

Changes the mock data on Tasklist to use only Zeebe user tasks


houssain-barouni (commented):
`ZeebeImportMultipleProcessesIT` is using `two_processes.bpmn` and it seems to be failing
`

houssain-barouni (approved):
I fixed ZeebeImportMultipleProcessesIT, it was checking the embedded forms imported from the extensionElements of the process, no need to check that anymore

PR #23545, opened 2024-10-15
feat: store commitIndex in raft MetaStore

### LINKED ISSUE #13790: Store commit index in raft MetaStore
Currently, on restart a partition's commitIndex is initialized to 0. This is perfectly fine, as we can reliably re-built the latest commitIndex from the quorum. 

In practice though, there can be disaster scenarios where data loss occur in one or more replica. In this cases there are chances that the replicas with data loss forms a quorum and overwrites data in the replica that actually has the correct latest data. We cannot 100% prevent data corruption or loss in case quorum of nodes lose their data. But we might be able to detect or prevent some inconsistencies.

The goal is to persist commitIndex in MetaStore. On restart a replica initialize commitIndex from the MetaStore. If the replicas with data loss forms a quorum, then the node with the latest committed data can detect this when it tries to overwrite its entries by the entries received from the new leader. This will still require manual intervention to recover, and might still result in some data loss. However, it is better than the current scenario, where we silently overwrite correct data.

Caveats: Writing commitIndex to MetaStore can add more I/O. Delayed and less frequent flushing would be acceptable in this case.

entangled90:
## Description
The raft commitIndex is saved in the MetaStore, when lastFlushedIndex is written to disk.

A new version of the SBE schema in raft-entry-schema.xml has been created to accomodate a new field.
On startup the MetaStore reads the file and rewrites it using the current version of the message to be able to update it in place.

## Related issues
closes #13790

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/PriorityElectionTimer.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/PriorityElectionTimer.java
@@ -77,6 +77,7 @@ public class PriorityElectionTimer implements ElectionTimer {
                 log.debug("Failed to poll a majority of the cluster in {}", pollTimeout);
                 reset();
               });
+      log.trace("Node triggering an election");
       triggerElection.run();
     } else {
       log.debug(
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
@@ -158,6 +158,9 @@ public class RaftContext implements AutoCloseable, HealthMonitorable {
   private final RaftPartitionConfig partitionConfig;
   private final int partitionId;
 
+  // after firstCommitIndex is set it will be null
+  private AwaitingReadyCommitListener awaitingReadyCommitListener;
+
   public RaftContext(
       final String name,
       final int partitionId,
@@ -204,7 +207,6 @@ public class RaftContext implements AutoCloseable, HealthMonitorable {
     // Load the current term and last vote from disk.
     term = meta.loadTerm();
     lastVotedFor = meta.loadVote();
-
     // Construct the core log, reader, writer, and compactor.
     raftLog =
         storage.openLog(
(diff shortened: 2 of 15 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/PriorityElectionTimer.java
```diff
@@ -77,6 +77,7 @@ private void onElectionTimeout() {
                 log.debug("Failed to poll a majority of the cluster in {}", pollTimeout);
                 reset();
               });
+      log.debug("Node triggering an election");
```
deepthidevaki, line 80: :wrench: This can be removed or set it to trace level.

zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
```diff
@@ -251,7 +253,20 @@ public RaftContext(
     registerHandlers(protocol);
     started = true;
 
-    addCommitListener(new AwaitingReadyCommitListener());
+    if (meta.hasCommitIndex()) {
+      log.debug("MetaStore has commit index: {}", meta.commitIndex());
```
deepthidevaki, line 257: :wrench: This log can be removed, because it is logged right after.

zeebe/atomix/cluster/src/main/java/io/atomix/raft/storage/log/RaftLog.java
```diff
@@ -175,7 +185,9 @@ public void deleteAfter(final long index) {
     if (index < commitIndex) {
       throw new IllegalStateException(
           String.format(
-              "Expected to delete index after %d, but it is lower than the commit index %d. Deleting committed entries can lead to inconsistencies and is prohibited.",
+              """
+                 Expected to delete index after %d, but it is lower than the commit index %d.\
+                 Deleting committed entries can lead to inconsistencies and is prohibited.""",
```
deepthidevaki, line 190: :wrench: Here also we can inlcude the full message added in `reset`.

zeebe/atomix/cluster/src/test/java/io/atomix/raft/RaftFailOverIT.java
```diff
@@ -196,6 +250,7 @@ public void shouldReplicateEntriesAfterSnapshotOnJoin() throws Exception {
     raftRule.takeCompactingSnapshot(200);
 
     // when
+    System.out.println("Follower %s joining cluster".formatted(follower));
```
deepthidevaki, line 253: :x: Replace this with logger.

zeebe/atomix/cluster/src/test/java/io/atomix/raft/RaftFailOverIT.java
```diff
@@ -66,6 +71,55 @@ public void shouldCommitEntriesAfterFollowerShutdown() throws Throwable {
...
+        .getContext()
+        .addStateChangeListener(
+            state -> {
+              final var commitIdx = server.getContext().getCommitIndex();
+              states.put(commitIdx, state);
+            });
+
+    bootstrapFuture.join();
+    assertThat(states)
+        .isEqualTo(Map.ofEntries(Map.entry(1L, State.ACTIVE), Map.entry(33L, State.READY)));
```
deepthidevaki, line 120: :+1: This is a nice way to test this scenario.

deepthidevaki (approved):
Great job :rocket:

PR #23548, opened 2024-10-15
Add RoleUpdateProcessor

### LINKED ISSUE #22930: Process `Role.UPDATE` command
<h2>Description</h2>

image

1. Create new class `RoleUpdateProcessor` and implement `AuthorizableDistributionProcessor.Authorizable<RoleRecord>`
    - Check if the role already exists. If no, reject the command with `NOT_FOUND`. If yes continue.
    - If the role name changed, check if the new role name already exists. If yes reject the command with `ALREADY_EXISTS`. If no continue.
    - Write a `Role.UPDATED` event and respond to the client
    - Write a `Role.UPDATED` event
        - Resource type: `ROLE`
        - Permission type: `UPDATE`
        - Resource: `RoleRecord`
5. Unit test the Role.UPDATED command
6. Unit test the Role.UPDATED command distribution

nicpuppa:
## Description

This PR adds the `RoleUpdateProcessor`

## Related issues

closes #22930

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleUpdateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleUpdateProcessor.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.identity;
+
+import static io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.UNAUTHORIZED_ERROR_MESSAGE;
+
+import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.AuthorizationRequest;
+import io.camunda.zeebe.engine.processing.streamprocessor.DistributedTypedRecordProcessor;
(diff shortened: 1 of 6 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleUpdateProcessor.java
```diff
@@ -0,0 +1,106 @@
...
+    this.commandDistributionBehavior = commandDistributionBehavior;
+  }
+
+  @Override
+  public void processNewCommand(final TypedRecord<RoleRecord> command) {
+    final var record = command.getValue();
+    final var persistedRecord = roleState.getRole(record.getRoleKey());
+    if (persistedRecord.isEmpty()) {
+      final var errorMessage =
+          "Expected to update role with key '%s', but a role with this key does not exists"
```
koevskinikola, line 60: 🔧 Small typo:

```suggestion
          "Expected to update role with key '%s', but a role with this key does not exist."
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/RoleTest.java
```diff
@@ -52,4 +52,61 @@ public void shouldNotDuplicate() {
...
+    assertThat(notPresentUpdateRecord)
+        .hasRejectionType(RejectionType.NOT_FOUND)
+        .hasRejectionReason(
+            "Expected to update role with key '"
+                + notPresentRoleKey
+                + "', but a role with this key does not exists");
+  }
+
+  @Test
+  public void shouldRejectIfRoleWithSameNameIsPresent() {
```
koevskinikola, line 94: 🔧 Should we also test for a rejection if a user is not authorized? Or is it enough that the authorization behavior class is tested?

koevskinikola (approved):
👍 @nicpuppa the code looks good. I only found a minor typo and a possible test case that we can cover, so I'm pre-approving this PR.

PR #23576, opened 2024-10-15
Opensearch schema manager

### LINKED ISSUE #23164: Schema manager can use opensearch client
**Description**

The schema manager is implementation agnostic so it should be able to simply take a `SearchEngineClient` instance and execute normally (blind to the exact implementation of the client ELS/Opensearch). 

Write tests to confirm schema creation/update actions work as expected.

**Success Criteria:**

- [x] Verify initialisation of resources given a list of index and index template descriptors, creating new resources and skipping resources which already exist.
- [x] Verify update of index and index template mapping fields given a list of new fields.

EuroLew:
## Description

This will add opensearch support for the schema manager.

**Success Criteria:**
- [x] Schema manager can be created with opensearch client.
- [x] Schema manager functionality is as intended when using the opensearch client

This PR involves
- Moving the `SearchEngineClient` implementations into their own packages
- Converting `SchemaManager` into a class since the implementations as an interface are the exact same only the `SearchEngineClient` changes but since an interface is passed in all the details are abstracted.
- Refactored `SchemaManagerIT` to support both ELS and opensearch tests without copy and pasting the exact same tests just changing `OpensearchClient` for `ElasticsearchClient`

## Related issues

closes #23164


mustafadagher (approved):
All looks good, thank you! 🎉 

I have only one suggestion below.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -81,7 +81,12 @@ public void open(final Controller controller) {
...
     final var searchEngineClient = clientAdapter.getSearchEngineClient();
-    final var schemaManager = clientAdapter.createSchemaManager(provider);
+    final var schemaManager =
+        new SchemaManager(
+            searchEngineClient,
+            provider.getIndexDescriptors(),
+            provider.getIndexTemplateDescriptors(),
+            configuration);
 
     schemaStartup(schemaManager, searchEngineClient);
```
mustafadagher, line 91: 🔧 another non-related suggestion/question: could this `schemaStartup` be part of the schema manager class?
like `SchemaManager.startup()` 
I see that it doesn't need anything from the exporter that it doesn't have access too already.

EuroLew: sounds like a good idea

PR #23584, opened 2024-10-15
Identity/process user delete command

### LINKED ISSUE #21679: Process `User.DELETE` command
<h2>Description</h2>

- [x] Extend `UserIntent` with a `DELETE` intent (ordering-wise it's nice to put it above the `DELETED` event)
- [x] Create a `UserDeleteProcessor` 
    - Verify if a user with this `username` exists
        - [x] If not, write a rejection with a proper eror message indicating the user doesn't exist
        - If yes
            - [x] Write `AuthorizationIntent.DELETED` events for each of the user's authorizations
            - [x] Write a `UserIntent.DELETED` event and distirbute the command on the Identity `DistributionQueue`
- [x] Write tests that verify processing the command works as expected. Verify the rejection, a success case, and the distribution acknowledgement.

Ben-Sheppard:
## Description

This PR adds in the support for the user delete processor, I would like to highlight:

1. In this PR we don't emit an event for each authorization belonging to the user, instead we handle the changes to the authorization state within the user delete applier. This approach was suggested by @remcowesterhoud 
2. We also remove the owner type from the authorization state when the user is deleted to ensure the ownerType -> ownerKey mapping is clean.

## Related issues

closes #21679

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/DeleteUserMultiPartitionTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/DeleteUserMultiPartitionTest.java
@@ -14,13 +14,17 @@ import io.camunda.zeebe.engine.state.distribution.DistributionQueue;
 import io.camunda.zeebe.engine.util.EngineRule;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.RecordType;
+import io.camunda.zeebe.protocol.record.ValueType;
 import io.camunda.zeebe.protocol.record.intent.CommandDistributionIntent;
 import io.camunda.zeebe.protocol.record.intent.UserIntent;
 import io.camunda.zeebe.protocol.record.value.CommandDistributionRecordValue;
 import io.camunda.zeebe.test.util.record.RecordingExporter;
 import io.camunda.zeebe.test.util.record.RecordingExporterTestWatcher;
+import java.time.Duration;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
+import org.assertj.core.api.Assertions;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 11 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/DeleteUserTest.java
```diff
@@ -0,0 +1,89 @@
...
+        .hasFieldOrPropertyWithValue("email", "foo@bar.blah")
+        .hasFieldOrPropertyWithValue("password", "Foo Bar");
+  }
+
+  @DisplayName("should reject the command if user does not exist")
+  @Test
+  public void shouldRejectTheCommandIfUserDoesNotExist() {
+    final var username = UUID.randomUUID().toString();
+    final var userRecord =
+        ENGINE
```
nicpuppa, line 65: 🔧 This can be removed, as you are fetching the user by key. You can just try to delete a user providing a reasonable `userKey` (instead of `-1`)

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/appliers/UserDeletedApplierTest.java
```diff
@@ -0,0 +1,95 @@
...
+import io.camunda.zeebe.engine.util.ProcessingStateExtension;
+import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
+import io.camunda.zeebe.protocol.record.value.AuthorizationOwnerType;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@ExtendWith(ProcessingStateExtension.class)
+public class UserDeletedApplierTest {
```
nicpuppa, line 28: 🔧 This is great, but I'd suggest you to add also unit tests for the new state methods in the `UserStateTest` class.
Since the appliers should be simple in terms of logic (just calling some methods from the state class) we prefer testing the single methods, only when there is some logic in the appliers then it makes sense to create a specific test for the applier

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/user/DeleteUserMultiPartitionTest.java
```diff
@@ -0,0 +1,132 @@
...
+import io.camunda.zeebe.test.util.record.RecordingExporter;
+import io.camunda.zeebe.test.util.record.RecordingExporterTestWatcher;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestWatcher;
+
+public class DeleteUserMultiPartitionTest {
```
nicpuppa, line 29: ❌ We should add a test to verify distribution ordering within the queue, you can see an example [here](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AddPermissionAuthorizationMultiPartitionTest.java)

nicpuppa (changes requested):
Hey @Ben-Sheppard great job, please have a look at my comments 👀 🙏

nicpuppa (approved):
Thanks 👍

PR #23598, opened 2024-10-15
fix: apply enum for task states and fix candidate group and user

nathansandi:
## Description

- Rename `candidateGroup` to `candidateGroups` 
- Rename `candidateUser` to `candidateUsers`
- Create `enums` for User Task State 
- Update Client and Controller tests


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
```diff
@@ -345,10 +345,10 @@ public static UserTaskItem toUserTask(final UserTaskEntity t) {
         .processDefinitionKey(t.processDefinitionId())
         .elementInstanceKey(t.flowNodeInstanceId())
         .processDefinitionId(t.bpmnProcessId())
-        .state(t.state())
+        .state(UserTaskItem.StateEnum.fromValue(String.valueOf(t.state())))
```
tmetzke, line 348: 🔧 I haven't looked into this in my IDE but it feels like this could be done in a simpler way. What does `.state()` expect as a value and what is the type of `t.state()`?

nathansandi: Yes it possible to make it simpler, this expects receive a UserItem.StateEnum

I have updated the code on this way now: `.state(UserTaskItem.StateEnum.fromValue(t.state().name()))`

tmetzke: The type of `t.state()` is a different enum class, I assume?

nathansandi: Yes ;) It is from `UserItem`, now its like in the same way as the others.

tmetzke: OK, I figured that, yeah. Good enough for me 👍

search/search-domain/src/main/java/io/camunda/search/entities/UserTaskEntity.java
```diff
@@ -19,7 +19,7 @@ public final record UserTaskEntity(
     String creationTime,
     String completionTime,
     String assignee,
-    String state,
+    UserTaskState state,
```
tmetzke, line 22: ❓ Is this change really necessary? In the search domain, it could also just stay a String, right? This should only be about the REST API, where more "type safety" would be nice for users.

nathansandi: It works without this change as well, I follow up what was develop in another entities as:
- ProcessInstanceEntity
- FlowNodeInstanceEntity
I have applied on UserTaskEntity to keep the consistency

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2583,18 +2588,23 @@ components:
           format: int64
         state:
           type: string
```
tmetzke, line 2590: 🔧 By the way, all of the user task search components don't have descriptions. That makes it harder for users to understand the properties.

search/search-domain/src/main/java/io/camunda/search/entities/UserTaskEntity.java
tmetzke: Interesting, I wasn't aware 👍 
I'll have a look at that as well. This is non-blocking 👍

tmetzke (approved):
👍 The changes are valid, thanks, @nathansandi.
🔧 There are some smaller things we can follow-up on if we don't want to tackle them right away, like adding more descriptions in the OpenAPI spec.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nathansandi: Sounds great :) 
I have added the more descriptions here https://github.com/camunda/camunda/pull/23598#:~:text=on%20rest%2Dapi-,33b21de,-Merge%20state 

I am going to tag a tech written for review ;).

christinaausley (approved):
Uppercased ID, otherwise looks good!

PR #23604, opened 2024-10-16
Add RoleAddEntityProcessor

### LINKED ISSUE #22938: Process `Role.ADD_ENTITY` command
<h2>Description</h2>

image

1. Create new class `RoleAddEntityProcessor` and implement `AuthorizableDistributionProcessor.Authorizable <RoleRecord>`
    - Check if the role exists. If no, reject the command with `NOT_FOUND`. If yes continue.
    - Check if the entity exists. If no, reject the command with `NOT_FOUND`. If yes continue.
    - Write a `Role.ENTITY_ADDED` event and respond to the client
    - Write a `Role.ENTITY_ADDED` event
        - Resource type: `ROLE`
        - Permission type: `UPDATE`
        - Resource: `RoleRecord`
5. Unit test the Role.ADD_ENTITY command
6. Unit test the Role.ADD_ENTITY command distribution

nicpuppa:
## Description

This PR adds the `RoleAddEntityProcessor`

## Related issues

closes #22938

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
@@ -62,7 +62,7 @@ public class RoleAddEntityProcessor implements DistributedTypedRecordProcessor<R
     final var persistedRecord = roleState.getRole(record.getRoleKey());
     if (persistedRecord.isEmpty()) {
       final var errorMessage =
-          "Expected to update role with key '%s', but a role with this key does not exists"
+          "Expected to update role with key '%s', but a role with this key does not exist."
               .formatted(record.getRoleKey());
       rejectionWriter.appendRejection(command, RejectionType.NOT_FOUND, errorMessage);
       responseWriter.writeRejectionOnCommand(command, RejectionType.NOT_FOUND, errorMessage);
@@ -71,7 +71,7 @@ public class RoleAddEntityProcessor implements DistributedTypedRecordProcessor<R
 
     final var authorizationRequest =
         new AuthorizationRequest(command, AuthorizationResourceType.ROLE, PermissionType.UPDATE)
-            .addResourceId(record.getName());
+            .addResourceId(persistedRecord.get().getName());
     if (!authCheckBehavior.isAuthorized(authorizationRequest)) {
       final var errorMessage =
           UNAUTHORIZED_ERROR_MESSAGE.formatted(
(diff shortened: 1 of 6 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/RoleTest.java
```diff
@@ -52,4 +54,81 @@ public void shouldNotDuplicate() {
...
+    assertThat(notPresentUpdateRecord)
+        .hasRejectionType(RejectionType.NOT_FOUND)
+        .hasRejectionReason(
+            "Expected to update role with key '"
+                + notPresentRoleKey
+                + "', but a role with this key does not exists");
+  }
+
+  @Test
+  public void shouldRejectIfEntityIsNotPresent() {
```
koevskinikola, line 109: ❓ Similar to my question with the "role update" PR: does it make sense to add a test for rejection if unauthorized?

Or would that go in the authorization unit, or IT tests?

koevskinikola (changes requested):
@nicpuppa the code looks great. 👍 

I have some small suggestions for typos in the error messages, and a possible optimization for the state access. Let me know what you think.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+      rejectionWriter.appendRejection(command, RejectionType.UNAUTHORIZED, errorMessage);
+      responseWriter.writeRejectionOnCommand(command, RejectionType.UNAUTHORIZED, errorMessage);
+      return;
+    }
+
+    final var entityKey = record.getEntityKey();
+    final var entityType = record.getEntityType();
+    if (!isEntityPresent(entityKey, entityType)) {
+      final var errorMessage =
+          "Expected to add entity with key '%s' and type '%s' to role with key '%s', the entity does not exists"
```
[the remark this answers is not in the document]
nicpuppa: In the future we will have different entity types (e.g mapping), so the number of reads to the states will increase in this check. For this reason is better to check the auth first and then check the entities

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/RoleTest.java
nicpuppa: We have IT tests to verify this behaviour of the processors (see [here](https://github.com/camunda/camunda/tree/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/authorization)). Since I'm working on refactor them and the behaviour class is tested, I'd say is out of scope for this PR

koevskinikola (approved):
All good @nicpuppa 👍

PR #23612, opened 2024-10-16
Add RoleRemoveEntityProcessor

### LINKED ISSUE #22940: Process `Role.REMOVE_ENTITY` command

nicpuppa:
## Description

This PR adds the `RoleRemoveEntityProcessor`

## Related issues

closes #22940 

blocked by https://github.com/camunda/camunda/pull/23604

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleRemoveEntityProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleRemoveEntityProcessor.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.identity;
+
+import static io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.UNAUTHORIZED_ERROR_MESSAGE;
+
+import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.AuthorizationRequest;
+import io.camunda.zeebe.engine.processing.streamprocessor.DistributedTypedRecordProcessor;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedRejectionWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedResponseWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
+import io.camunda.zeebe.engine.state.distribution.DistributionQueue;
+import io.camunda.zeebe.engine.state.immutable.RoleState;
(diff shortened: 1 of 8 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleRemoveEntityProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+    this.commandDistributionBehavior = commandDistributionBehavior;
+  }
+
+  @Override
+  public void processNewCommand(final TypedRecord<RoleRecord> command) {
+    final var record = command.getValue();
+    final var persistedRecord = roleState.getRole(record.getRoleKey());
+    if (persistedRecord.isEmpty()) {
+      final var errorMessage =
+          "Expected to update role with key '%s', but a role with this key does not exists"
```
koevskinikola, line 65: 🔧 Small typo:

```suggestion
          "Expected to update role with key '%s', but a role with this key does not exist"
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleRemoveEntityProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+      rejectionWriter.appendRejection(command, RejectionType.UNAUTHORIZED, errorMessage);
+      responseWriter.writeRejectionOnCommand(command, RejectionType.UNAUTHORIZED, errorMessage);
+      return;
+    }
+
+    final var entityKey = record.getEntityKey();
+    final var entityType = record.getEntityType();
+    if (!isEntityPresent(entityKey, entityType)) {
+      final var errorMessage =
+          "Expected to remove entity with key '%s' and type '%s' to role with key '%s', the entity does not exists"
```
koevskinikola, line 88: 🙃  Some small suggestions:

```suggestion
          "Expected to remove an entity with key '%s' and type '%s' from role with key '%s', but the entity doesn't exist"
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleRemoveEntityProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+          UNAUTHORIZED_ERROR_MESSAGE.formatted(
+              authorizationRequest.getPermissionType(), authorizationRequest.getResourceType());
+      rejectionWriter.appendRejection(command, RejectionType.UNAUTHORIZED, errorMessage);
+      responseWriter.writeRejectionOnCommand(command, RejectionType.UNAUTHORIZED, errorMessage);
+      return;
+    }
+
+    final var entityKey = record.getEntityKey();
+    final var entityType = record.getEntityType();
+    if (!isEntityPresent(entityKey, entityType)) {
```
koevskinikola, line 86: 🔧 It might be good to perform the entity check first, so the number of reads to the states is reduced if the entity doesn't exist.

koevskinikola (changes requested):
@nicpuppa the code looks good. I just have some suggestions for some of the processor's error messages, and a possible optimization for the number of reads to the states.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleRemoveEntityProcessor.java
nicpuppa: In the future we will have different entity types (e.g mapping), so the number of reads to the states will increase in this check. For this reason is better to check the auth first and then check the entities

koevskinikola (approved):
LGTM @nicpuppa 🎉

PR #23626, opened 2024-10-16
fix: set isTruncated as a Boolean instead of primitive 

nathansandi:
## Description

- VariableFilter `isTruncated` should set as object and not primitive on FilterRequests

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/VariableFilterTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/VariableFilterTransformer.java
@@ -83,7 +83,10 @@ public class VariableFilterTransformer implements FilterTransformer<VariableFilt
     return stringTerms("tenantId", tenant);
   }
 
-  private SearchQuery getIsTruncatedQuery(final boolean isTruncated) {
+  private SearchQuery getIsTruncatedQuery(final Boolean isTruncated) {
+    if (isTruncated == null) {
+      return null;
+    }
     return term("isPreview", isTruncated);
   }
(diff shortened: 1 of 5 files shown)


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/VariableFilterTransformer.java
```diff
@@ -83,7 +83,10 @@ private SearchQuery getTenantIdQuery(final List<String> tenant) {
     return stringTerms("tenantId", tenant);
   }
 
-  private SearchQuery getIsTruncatedQuery(final boolean isTruncated) {
+  private SearchQuery getIsTruncatedQuery(final Boolean isTruncated) {
+    if (isTruncated == null) {
+      return null;
+    }
     return term("isPreview", isTruncated);
```
panagiotisgts, line 90: Does the `isPreview` match the ES/OS field property? If so just giving you a reminder that it will be changed to `isTruncated` for consistency when the schema is moved over the exporter :smiling_face_with_tear:

panagiotisgts (approved):
LGTM! One comment/reminder only.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/VariableFilterTransformer.java
nathansandi: It is :) For now the mapping is still isPreview, so for avoid breaking changes, kets keep it as is, and change when the exporters are released and indexes updates

Thanks for raise this point @panagiotisgts

PR #23627, opened 2024-10-16
build: remove unused dependencies

Ben-Sheppard:
## Description

It was noticed that the H2 dependency in use introduces a CVE, after looking further, this dependency was in Identity from a time that predates the new architectural approach of using Zeebe.

As we don't use H2 actively I believe its safe to remove (we also don't use Postgres in Identity for the same reason so I have also removed that dependency too)

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -262,11 +262,6 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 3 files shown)


maryarm (approved):

Ben-Sheppard:
Closing due to being stale

PR #23629, opened 2024-10-16
fix: use prefix for templates

sdorokhova:


sdorokhova:
Was fixed within global prefix work: https://github.com/camunda/camunda/commit/d741901d4f3d9524bf0ad933e48c26a28b296b7f

PR #23631, opened 2024-10-16
Camunda exporter opensearch tests

### LINKED ISSUE #23166: Exporter is integrated with opensearch schema manager
**Description:**

All exporter actions should work normally when using an opensearch backed schema manager.

**Success Criteria:**
- [x] Exporter can export records to opensearch

EuroLew:
## Description

Verify functionality of the Camunda exporter when it comes to exporting records works with an opensearch backing engine.

**Success Criteria:**
- [x] Exporter can export records to opensearch 

## Related issues

closes #23166

### DIFF
--- a/zeebe/exporters/camunda-exporter/pom.xml
+++ b/zeebe/exporters/camunda-exporter/pom.xml
@@ -144,12 +144,6 @@
     </dependency>
 
     <dependency>
-      <groupId>com.github.docker-java</groupId>
-      <artifactId>docker-java-api</artifactId>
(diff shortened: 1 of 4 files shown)


mustafadagher (approved):
This looks good to me! 🚀 

💭 I was thinking if there was a way to have an abstract class defining all the test methods, with 2 implementations which implement the params or the methods specific to the search engine (or its client). But the I don't think it's worth the investment right now. This is more than good enough :)

Issue #23630, opened 2024-10-16
Move Optimize footer information to sidebar and remove footer component


misiekhardcore:
Engineering DRI: @misiekhardcore

### Context

Following other components Optimize needs to remove page footer and use info sidebar for the info displayed now in the footer

### Acceptance Criteria

- [x] move footer information into info sidebar (same as other components do)
- [x] remove footer component

### Testing Notes

- page footer is gone
- version information is displayed in user side bar

![Screenshot from 2024-10-16 15-21-00](https://github.com/user-attachments/assets/2164708d-ffec-4c8d-90b2-77af19980ef0)

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/23639
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [x] The change is implemented as described on all target environments/versions/modes
- [x] The documentation changes are as expected

### Completion

- [x] All Review stages are successfully completed
- [x] All associated PRs are merged to the main branch(es) and stable branches
- [x] The correct version labels are applied to the issue

RobertDorlan:
Successfully tested agains C8-SNAPSHOT

 move footer information into info sidebar (same as other components do) ✅ 
 remove footer component ✅ 
![image](https://github.com/user-attachments/assets/1a9b49d1-05a1-4fe9-abbd-e7527d4c4c24)

PR #23639, opened 2024-10-16
refactor: move footer data to side menu and remove footer

misiekhardcore:
## Checklist

- [ ] for CI changes:

## Related issues

related to https://github.com/camunda/camunda/issues/23630


optimize/client/src/components/Footer/Footer.tsx
```diff
@@ -1,31 +0,0 @@
...
-  const {optimizeVersion} = useUiConfig();
-
-  const timezoneInfo =
-    t('footer.timezone') + ' ' + Intl.DateTimeFormat().resolvedOptions().timeZone;
-
-  return (
-    <footer className="Footer">
-      <div title={timezoneInfo} className="timezone">
-        {timezoneInfo}
-      </div>
```
OmranAbazid, line 24: The timezone information is an important thing in Optimize and should not be removed.
I think it should be moved to the user menu as well. We can try adding a seperator and putting it below the copyright info.

OmranAbazid (commented):
The line height is too big for us for some reason
Screenshot 2024-10-21 at 11 31 41 AM

Other webapps use a much smaller line height. See console for example:
Screenshot 2024-10-21 at 11 32 29 AM

Can you take a look and see if it is a problem in our side?

misiekhardcore:
@OmranAbazid the line heights are different due to the carbon reset we are disabling

OmranAbazid:
> @OmranAbazid the line heights are different due to the carbon reset we are disabling 

@misiekhardcore lets add a temporary reset in the header with a todo to remove it later

optimize/client/src/components/Footer/Footer.tsx
misiekhardcore: I cant see any place where we can pass this. C3 navigation has defined props and its not really allowing to pass anything more then those props. There is `bottomElements` section, but it renders buttons and takes props to use with buttons.

misiekhardcore:
@OmranAbazid C3 navigation is styled wirh styled-components so I can't really target that element with version info to style it. I think we need to keep it as is.
FYI Out of curiosity I enabled the carbon style reset and Optimize still looks alright, so maybe it is a good time to consider enabling that reset for good.

optimize/client/src/components/Footer/Footer.tsx
misiekhardcore: I will go for it and use that `bottomElements` for this, but it is really hacky

OmranAbazid:
> @OmranAbazid C3 navigation is styled wirh styled-components so I can't really target that element with version info to style it. I think we need to keep it as is. FYI Out of curiosity I enabled the carbon style reset and Optimize still looks alright, so maybe it is a good time to consider enabling that reset for good.

We will have issue with p and bold tags across the application. We have an issue to fix those and enable the css reset.
Let's keep this issue on hold until we fix that css reset. I will try to do it as soon as possible and afterwards, i will look into merging this.

optimize/client/src/components/Footer/Footer.tsx
OmranAbazid: @misiekhardcore can we try the suggested solution in the slack thread above?

misiekhardcore:
@OmranAbazid I've seen [this convo](https://camunda.slack.com/archives/C049YH1C202/p1729768509111639) and I moved the timezone info to that section

OmranAbazid (approved):

PR #23655, opened 2024-10-16
feat: support user delete in camunda-services

Ben-Sheppard:
## Description

This PR introduces a method in the service classes to support user deletion


matthewBearCamunda (approved):
LGTM 🚀

PR #23665, opened 2024-10-16
Exporter/usermanagement index descriptor

Ben-Sheppard:
## Description

This PR achieves a couple of tasks:

1. Adds in the required index descriptors based on the suggestion [here](https://camunda.slack.com/archives/C05QBRGMA9F/p1729083388177189?thread_ts=1728936008.994909&cid=C05QBRGMA9F)
2. Refactors the existing handlers just to make them neater/refined

Something I've also noticed which we may want to consider in the future, we are opinionated about there being a `component` tied to an index, in the new streamlined architecture the intention is not to think in terms of components but instead resources - for now I've used `identity` to avoid issues.


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/AuthorizationFilterTransformer.java
```diff
@@ -19,14 +19,14 @@ public class AuthorizationFilterTransformer implements FilterTransformer<Authori
   @Override
   public SearchQuery toSearchQuery(final AuthorizationFilter filter) {
     return and(
-        filter.ownerKey() == null ? null : term("value.ownerKey", filter.ownerKey()),
-        filter.ownerType() == null ? null : term("value.ownerType", filter.ownerType()),
-        filter.resourceKey() == null ? null : term("value.resourceKey", filter.resourceKey()),
-        filter.resourceType() == null ? null : term("value.resourceType", filter.resourceType()));
+        filter.ownerKey() == null ? null : term("ownerKey", filter.ownerKey()),
```
houssain-barouni, line 22: @Ben-Sheppard can you please update [AuthorizationEntity](https://github.com/camunda/camunda/blob/main/search/search-domain/src/main/java/io/camunda/search/entities/AuthorizationEntity.java) at the same time and remove the `value` intermediate field ?

Ben-Sheppard: I can indeed, but there are some other changes that I think will be needed with changing that class so it would be best if I follow up with a PR for those (to keep things smaller, this PR has already bloated a bit!)

mustafadagher (approved):
Hi @Ben-Sheppard , I've reviewed the changes only in the `camunda-exporter` module, and they LGTM!

EuroLew (approved):
Just minor changes, otherwise LGTM.

PR #23675, opened 2024-10-17
feat: introduce Tasklist's metric index

panagiotisgts:
## Description

Introduce Tasklist's Metric index and related entity in the webapps-schema

Part of: #23040
## Checklist

- [ ] for CI changes:

## Related issues

closes #


webapps-schema/src/main/resources/schema/elasticsearch/create/index/tasklist-metric.json
```diff
@@ -0,0 +1,23 @@
...
+        "format": "date_time || epoch_millis",
+        "type": "date"
+      },
+      "event": {
+        "type": "keyword"
+      },
+      "value": {
+        "type": "keyword"
+      },
+      "tenantId": {
```
Zelldon, line 18: ❓ Why do we have this?

Zelldon: Not sure how useful the tenant is for the metrics 🤔

ChrisKujawa (approved):
Thanks @panagiotisgts I just had two smaller comments.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/tasklist/MetricEntity.java
```diff
@@ -0,0 +1,69 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.webapps.schema.entities.tasklist;
+
+import java.time.OffsetDateTime;
+import java.util.Objects;
+
+public class MetricEntity extends TasklistEntity<MetricEntity> {
```
mustafadagher, line 13: I just added a similar one for operate in https://github.com/camunda/camunda/pull/23690, do we need two of the same?

Zelldon: So what we need is having two indices. Entities we might be able to streamline?

Zelldon: in general can we maybe also just use one entity base class?

webapps-schema/src/main/resources/schema/elasticsearch/create/index/tasklist-metric.json
panagiotisgts: It's the same as in Operate's metric index. Not really why we would need the tenant there as well.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/tasklist/MetricEntity.java
panagiotisgts: Yeah probably we can use the same entity, since the index name is being passed to the handler directly. I will reiterate on this one, once @mustafadagher  's PR is merged.

mustafadagher (approved):
LGTM

PR #23680, opened 2024-10-17
Add RoleDeletedApplier

### LINKED ISSUE #22920: Apply `Role.DELETED` event
<h2>Description</h2>

In Zeebe we modify our state by applying events. We need to create a new `RoleDeletedApplier`:

3. Iterate over all entities of the `ENTITIES_BY_ROLE` ColumnFamily that have this `roleKey`
    - For each entity that is a USER:
        - Get the user by its key from `UserState`
        - Remove the Role from the `PersistedUser`
        - Update the PersistedUser in the state
     - For each entity that is a MAPPING:
         - Get the mapping by its key from the `MappingState`
         - Remove the Role from the `PersistedMapping`
         - Update the `PeristedMapping` in the state
4. Iterate over all Authorizations in the `AuthorizationState` where the `ownerKey` is this `roleKey`
    - Delete any that are found 
5. Remove this Role from the `OWNER_TYPE_BY_OWNER_KEY` ColumnFamily in the `AuthorizationState`
6. Add a method to to the `MutableRoleState` interface to delete a role and implement this method in the `DbRoleState`:
    - Wrap the `roleKey`
    - Wrap the `roleName`
    - Delete from `ROLE_BY_NAME` ColumnFamily
    - Delete all entires for this role from the `ENTITY_BY_ROLE` ColumnFamily
    - Delete role from the `ROLES` ColumnFamily
    - Unit test this method
7. In the `RoleUpdatedApplier` call this update method of the `DbRoleState`
8. Unit test all the new state methods

ℹ️ Note that we don't send individual events to remove entities from the role when we delete it. While this may be nice, this can quickly get out of hand if there's a lot of entities assigned to this role. There could be thousands of follow-up events which significantly increases the likelihood of running into `ExceededBatchRecordSizeException`

nicpuppa:
## Description

This PR adds the `RoleDeletedApplier`

## Related issues

closes #22920

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/EventAppliers.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/EventAppliers.java
@@ -466,6 +466,10 @@ public final class EventAppliers implements EventApplier {
     register(
         RoleIntent.ENTITY_REMOVED,
         new RoleEntityRemovedApplier(state.getRoleState(), state.getUserState()));
+    register(
+        RoleIntent.DELETED,
+        new RoleDeletedApplier(
+            state.getRoleState(), state.getUserState(), state.getAuthorizationState()));
   }
 
   private void registerScalingAppliers(final MutableProcessingState state) {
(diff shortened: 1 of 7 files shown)


koevskinikola (approved):
LGTM 👍

PR #23687, opened 2024-10-17
spike: archiver with in exporter

deepthidevaki:
## Description

A skeleton of how archiver will be integrated with the exporter

Main things to make it work with the exporter:
1. Implementation for Opensearch and Elasticsearch interface. The existing implementation uses different clients than what is available in the exporter. However, I assume this is not a blocker - just some busy work to adapt the apis. 
2. Injecting dependencies needs to be done manually instead of using spring magic. 
3. `BatchOperationTemplate` should be added to common `webapps-schema`. It was not added before because it is not needed by the exporter, but only the webapps. But the archiver needs some info of it (only index name and id). 

Note: This PR won't run because Opensearch/Elasticsearch implementation does not exist.

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
@@ -7,6 +7,12 @@
  */
 package io.camunda.exporter;
 
+import io.camunda.exporter.archiver.Archiver;
+import io.camunda.exporter.archiver.ArchiverJob;
+import io.camunda.exporter.archiver.ArchiverUtil;
+import io.camunda.exporter.archiver.BatchOperationArchiverJob;
+import io.camunda.exporter.archiver.ElasticsearchArchiverRepository;
+import io.camunda.exporter.archiver.ProcessInstancesArchiverJob;
 import io.camunda.exporter.config.ConnectionTypes;
 import io.camunda.exporter.config.ExporterConfiguration;
 import io.camunda.exporter.handlers.AuthorizationRecordValueExportHandler;
@@ -26,8 +32,10 @@ import io.camunda.webapps.schema.descriptors.operate.template.ListViewTemplate;
 import io.camunda.webapps.schema.descriptors.operate.template.VariableTemplate;
 import io.camunda.webapps.schema.descriptors.tasklist.index.FormIndex;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 
 /**
  * This is the class where teams should make their components such as handlers, and index/index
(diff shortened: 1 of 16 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
```diff
@@ -98,4 +106,28 @@ public Set<ExportHandler> getExportHandlers() {
...
     return exportHandlers;
   }
+
+  @Override
+  public Archiver getArchiver(final int partitionId) {
+    final ElasticsearchArchiverRepository archiverRepository =
+        new ElasticsearchArchiverRepository();
+    // Can provide more threads if needed. TODO: consider whether this thread pool should be shared
+    // among all partitions. Might require some refactoring, so probably not worth it for now.
+    final ScheduledThreadPoolExecutor archiverExecutor = new ScheduledThreadPoolExecutor(1);
```
Zelldon, line 116: I think this is good for the first iteration

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ElasticsearchAdapter.java
```diff
@@ -22,10 +22,13 @@ class ElasticsearchAdapter implements ClientAdapter {
   private final ElasticsearchClient client;
   private final ElasticsearchEngineClient searchEngineClient;
 
+  // TODO: private final ElasticsearchAsyncClient asyncClient;
+
   ElasticsearchAdapter(final ExporterConfiguration configuration) {
     final var connector = new ElasticsearchConnector(configuration.getConnect());
     client = connector.createClient();
     searchEngineClient = new ElasticsearchEngineClient(client);
+    // TODO: asyncClient = connector.createAsyncClient(); // for Archiver
```
Zelldon, line 31: The search client we can use?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/AbstractArchiverJob.java
```diff
@@ -0,0 +1,83 @@
...
+    errorStrategy = new BackoffIdleStrategy(100, 1.2f, 10_000);
+
+    rolloverBatchSize = 100; // TODO operateProperties.getArchiver().getRolloverBatchSize();
+    delayBetweenRuns =
+        Duration.ofMinutes(1)
+            .toMillis(); // TODO: operateProperties.getArchiver().getDelayBetweenRuns();
+  }
+
+  @Override
+  public void run() {
```
Zelldon, line 43: What I thought was whether we also could trigger the archiver from the Exporter (without having the searching and retrigger approach). E.g if we reach a PROCESS complete event or something 🤔 But I guess this would make the solution more the complex?

deepthidevaki: It is more complex. We don't want to block exporter thread. So the archiver has to run in background or async. So, how do you ensure archiving is triggered again, incase the exporter closed before archiving completes, and the new leader do not re-export the record. There would be several such "new" cases that we should consider. Given the time constraint, I would like to keep it simple and just port what we have already.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ElasticsearchAdapter.java
deepthidevaki: That is just implementation details. If we want to wrap asyncClient with in the searchclient, we can do that. Not sure if that has any benefit.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/AbstractArchiverJob.java
Zelldon: Was more an afterthought (improvement) nothing I would do now.

deepthidevaki: > What I thought is that the exporter can submit a job on the other thread pool to archive the respective instance.

Ya. Same issue as I mentioned before. There are several edge-cases to consider.

PR #23690, opened 2024-10-17
Move MetricFromProcessInstanceHandler Exporter handler to main

### LINKED ISSUE #23624: Move MetricFromProcessInstanceHandler

mustafadagher:
## Related issues

closes #23624


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromProcessInstanceHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override
+  public List<String> generateIds(final Record<ProcessInstanceRecordValue> record) {
+    return List.of();
+  }
+
+  @Override
+  public MetricEntity createNewEntity(final String id) {
+    return new MetricEntity();
+  }
```
mustafadagher, line 60: PS: these two functions (not generating Id, and not setting it on entity) were like this on https://github.com/camunda/camunda/blob/feat-operate-exporter/zeebe/exporters/operate-exporter/src/main/java/io/camunda/zeebe/operate/exporter/handlers/MetricFromProcessInstanceHandler.java.

I assumed this was intentional, right?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -175,7 +176,14 @@ private void updateLastExportedPosition() {
   private record CamundaExporterRecordFilter() implements RecordFilter {
     // TODO include other value types to export
     private static final Set<ValueType> VALUE_TYPES_2_EXPORT =
-        Set.of(USER, AUTHORIZATION, DECISION, PROCESS_INSTANCE, VARIABLE, INCIDENT);
+        Set.of(
+            USER,
+            AUTHORIZATION,
+            DECISION,
+            DECISION_REQUIREMENTS,
```
mustafadagher, line 183: This change has nothing to do with this PR. but I've missed adding `DECISION_REQUIREMENTS` to the values to export in the previous one, and I didn't think it needed a separate PR.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromProcessInstanceHandler.java
panagiotisgts: Since this is a metric index probably ES/OS create random document IDs for them? cc @sdorokhova maybe you know better regarding this

panagiotisgts (approved):
LGTM I'll wait for Svetlana to check the comment regarding the ID and then proceed to merge it.

P.S: @mustafadagher  I'll fix the checkstyle issue on the rebase

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromProcessInstanceHandler.java
mustafadagher: I had a discussion with @sdorokhova about this.
If ids were not generated the entities won't be updated or cached, and hence not picked up for flushing ([https://github.com/camunda/camunda/blob/e78dd201e46fbd3897e44b64bd78b7693916dd46/z[…]rc/main/java/io/camunda/exporter/store/ExporterBatchWriter.java](https://github.com/camunda/camunda/blob/e78dd201e46fbd3897e44b64bd78b7693916dd46/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java#L37))

We agreed to use processInstanceKey as an id here.

PR #23692, opened 2024-10-17
El 23638 global prefix camunda exporter

### LINKED ISSUE #23638: Support global prefix
It has been [decided](https://docs.google.com/document/d/1w1VzwyvvBzu0vnjCUj4UcLxTCVae8iMXAMA5_m6xs1s/edit#bookmark=id.wq9h75bo7w08) that we will support one global prefix. 

This means:

- The current prefixes `operate` and `tasklist` will be part of the index names. 
- The global prefix will be by default empty. 
- Setting a prefix will not remove the `operate` or `tasklist` parts, as it did before.

This involves that we need to make sure that the prefix is prepended to current index names (in descriptors).

Furthermore, we need to make sure that all indices that are defined via descriptors [here](https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java#L60) are created/validated/updated.

AT: 

 1. I can define a prefix that will be prepended to index names -
     1. Operate, Tasklist are still part of the name
 3. If no prefix is defined indices are created with their default name - defined in the descriptor
 4. When a descriptor is defined for an index/template it will be used to create/validate and update an index
     1. This means indices for Tasklist, Operate and Identity are always created no matter of prefix, etc.

EuroLew:
## Description

Add support for a global prefix for every index/index template.

**Success Critieria:**

- [x] _I can define a prefix that will be prepended to index names (Operate, Tasklist are still part of the name):_  This will now be set through the `getComponentName()` defined in the `AbstractIndexDescriptor`.
- [x] _If no prefix is defined indices are created with their default name - defined in the descriptor_: Done
- [x] _When a descriptor is defined for an index/template it will be used to create/validate and update an index
This means indices for Tasklist, Operate and Identity are always created no matter of prefix, etc._: Now that there is a global prefix, searches will now catch all indices with this prefix instead of excluding based on a component prefix where a prefix of `operate` would exclude `tasklist` indices.

## Related issues

closes #23638

### DIFF
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
@@ -26,19 +26,7 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   @Override
   public String getFullQualifiedName() {
-
-    //    legacy support for descriptors in operate where the index prefix contains the component
-    // name e.g. indexPrefix = "operate"
-    if (getIndexPrefix() != null && getIndexPrefix().contains(getComponentName())) {
-      return String.format("%s-%s-%s_", getIndexPrefix(), getIndexName(), getVersion());
-    }
-
-    if (getIndexPrefix() == null || getIndexPrefix().isEmpty()) {
-      return String.format("%s-%s-%s_", getComponentName(), getIndexName(), getVersion());
-    } else {
-      return String.format(
-          "%s-%s-%s-%s_", getIndexPrefix(), getComponentName(), getIndexName(), getVersion());
-    }
+    return String.format("%s-%s-%s_", getPrefixAndComponentName(), getIndexName(), getVersion());
   }
 
   @Override
@@ -55,15 +43,7 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   @Override
   public String getAllVersionsIndexNameRegexPattern() {
-    if (getIndexPrefix() != null && getIndexPrefix().contains(getComponentName())) {
-      return String.format("%s-%s-\\d.*", getIndexPrefix(), getIndexName());
-    }
-
-    if (getIndexPrefix() == null || getIndexPrefix().isEmpty()) {
-      return String.format("%s-%s-\\d.*", getComponentName(), getIndexName());
-    } else {
-      return String.format("%s-%s-%s-\\d.*", getIndexPrefix(), getComponentName(), getIndexName());
-    }
+    return String.format("%s-%s-\\d.*", getPrefixAndComponentName(), getIndexName());
   }
 
   @Override
@@ -73,6 +53,16 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   public String getIndexPrefix() {
     return indexPrefix;
+  }
+
+  private String getPrefixAndComponentName() {
+    // Cannot start index with "-" so must not append "-" for empty prefix
+    final var prefix = getIndexPrefix().isBlank() ? getIndexPrefix() : getIndexPrefix() + "-";
+
+    // Legacy descriptors have the same index prefix as component name this avoids duplication.
+    return getIndexPrefix().contains(getComponentName())
+        ? getIndexPrefix()
+        : prefix + getComponentName();
   }
 
   public abstract String getComponentName();
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
@@ -48,7 +48,11 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   @Override
   public String getAllVersionsIndexNameRegexPattern() {
-    return String.format("%s-%s-\\d.*", getIndexPrefix(), getIndexName());
+    if (getIndexPrefix().isEmpty()) {
+      return String.format("%s-%s-\\d.*", getComponentName(), getIndexName());
+    } else {
+      return String.format("%s-%s-%s-\\d.*", getIndexPrefix(), getComponentName(), getIndexName());
+    }
   }
 
   @Override
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
@@ -28,7 +28,7 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
   public String getFullQualifiedName() {
 
     //    legacy support for descriptors in operate where the index prefix contains the component
-    // nam e.g. indexPrefix = "operate"
+    // name e.g. indexPrefix = "operate"
     if (getIndexPrefix() != null && getIndexPrefix().contains(getComponentName())) {
       return String.format("%s-%s-%s_", getIndexPrefix(), getIndexName(), getVersion());
     }
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
@@ -26,7 +26,12 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   @Override
   public String getFullQualifiedName() {
-    return String.format("%s-%s-%s_", getIndexPrefix(), getIndexName(), getVersion());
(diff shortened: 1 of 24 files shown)


webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
```diff
@@ -26,7 +26,19 @@ public AbstractIndexDescriptor(final String indexPrefix, final boolean isElastic
...
+    if (getIndexPrefix() != null && getIndexPrefix().contains(getComponentName())) {
+      return String.format("%s-%s-%s_", getIndexPrefix(), getIndexName(), getVersion());
+    }
+
+    if (getIndexPrefix() == null || getIndexPrefix().isEmpty()) {
+      return String.format("%s-%s-%s_", getComponentName(), getIndexName(), getVersion());
+    } else {
+      return String.format(
+          "%s-%s-%s-%s_", getIndexPrefix(), getComponentName(), getIndexName(), getVersion());
+    }
```
Zelldon, line 41: Couldn't we just do:

```suggestion
      return String.format(
          "%s-%s-%s-%s_", getIndexPrefix(), getComponentName(), getIndexName(), getVersion());
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
```diff
@@ -99,10 +99,8 @@ public String toString() {
 
   public static final class IndexSettings {
     public static final int DEFAULT_VARIABLE_SIZE_THRESHOLD = 8191;
-    private String prefix = "operate";
+    private String prefix = "";
 
-    // TODO consolidate tasklist and operate prefixes
-    private String tasklistPrefix = "tasklist";
```
Zelldon, line 105: 👍🏼

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -123,20 +140,35 @@ private IndexSettings getIndexSettings(final String indexName) {
...
+            config.getIndex().getPrefix() + "*", MappingSource.INDEX_TEMPLATE);
 
     return schemaValidator.validateIndexMappings(
         currentTemplates,
         indexTemplateDescriptors.stream()
             .map(IndexDescriptor.class::cast)
             .collect(Collectors.toSet()));
   }
+
+  private String alLIndexNames() {
```
Zelldon, line 169: 🔧 I guess this is a typo?
```suggestion
  private String allIndexNames() {
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -123,20 +140,35 @@ private IndexSettings getIndexSettings(final String indexName) {
...
+    return indexDescriptors.stream()
+        .map(descriptor -> descriptor.getFullQualifiedName() + "*")
```
Zelldon, line 171: I think the start we don't need

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/DefaultExporterResourceProviderTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    provider
+        .getIndexDescriptors()
+        .forEach(
+            descriptor -> {
+              final var name = descriptor.getFullQualifiedName();
+              final var nameWithoutPrefix = name.substring("global-".length());
+              assertThat(name.matches(descriptor.getAllVersionsIndexNameRegexPattern())).isTrue();
+              assertThat(name.startsWith("global-")).isTrue();
+              assertThat(startsWithValidComponent(nameWithoutPrefix)).isTrue();
```
Zelldon, line 59: 🔧 This will fails as soon we add Identity I guess. The templates have no prefix.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -123,20 +140,35 @@ private IndexSettings getIndexSettings(final String indexName) {
...
 
   private Map<IndexDescriptor, Collection<IndexMappingProperty>> validateIndexTemplates(
       final IndexSchemaValidator schemaValidator, final SearchEngineClient searchEngineClient) {
-    final var currentTemplates = searchEngineClient.getMappings("*", MappingSource.INDEX_TEMPLATE);
+    if (indexTemplateDescriptors.isEmpty()) {
+      return Map.of();
+    }
+
+    final var currentTemplates =
+        searchEngineClient.getMappings(
```
Zelldon, line 159: ❓ Here we should also use the template descriptors no?

ChrisKujawa (commented):
Thanks @EuroLew I had some comments see below. We can also sync on them f2f if you want.

@mustafadagher @sdorokhova @panagiotisgts please take a look at the changes, as this affected also your area you're working on. 

We planned to have only one global index, and the previous prefixes (operate/tasklist) are component names that are part of the actual full name.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
EuroLew: for legacy descriptors in operate the index prefix will contain the string `operate` or `tasklist` causing the indices to be for example `prefix-operate-operate-index-8.x.x` and if we provide an empty prefix we will have our names start with a `-` e.g. `-operate-index-8.x.x`

Zelldon: So like?
```
var prefix = isBlank() ? getComponent() : getIndexPrefix() + "-" + getComponent();
return String.format("%s-%s-%s_", prefix, getComponentName(), getIndexName(), getVersion());
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
EuroLew: we need the wildcard, without it the requests will throw an exception if one of the values doesn't exist, e.g. `getIndices("a,b,c")` if only a and b exist it will throw but if `getIndices(a*,b*c*)` then it will return the a and b indices.

There isn't to my knowledge a better way to check which indices from a list exist in ELS/OS (with regards to doing it in only 1 request). There's [this](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-template-exists-v1.html) which tells you about existence but it only returns a true/false and not more information (which would be ideal)

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/DefaultExporterResourceProviderTest.java
EuroLew: maybe we could create a enum and have all `getComponentName` instances reference this enum but that seems a bit overkill and it would be easy to fix the test.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
EuroLew: template descriptors do not support a comma seperated list of values, just a singular value https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-template.html. 

This is as granular was we can get I think.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
EuroLew: this would break for legacy descriptors where 
prefix = operate
component name = operate
index name = index
verison = 8.0.0

it would return operate-operate-index-8.0.0, take a look at `OperationTemplate.java` as an example

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -123,20 +140,40 @@ private IndexSettings getIndexSettings(final String indexName) {
...
         indexTemplateDescriptors.stream()
             .map(IndexDescriptor.class::cast)
             .collect(Collectors.toSet()));
   }
+
+  private String allIndexNames() {
+
+    // The wildcard is required as without it, requests would fail if the index didn't exist.
+    // this way all descriptors can be retrieved in one request without errors due to not created
+    // indices
```
Zelldon, line 173: 👍🏼

ChrisKujawa (approved):
Awesome thanks 🚀

PR #23700, opened 2024-10-17
Move event related handlers to main

### LINKED ISSUE #23621: Move event related handlers
Move event related handlers: EventFromIncidentHandler, EventFromJobHandler, EventFromProcessInstanceHandler, EventFromProcessMessageSubscriptionHandler

mustafadagher:
## Description

Introducing the following event related handlers to camunda exporter:
- EventFromIncidentHandler
- EventFromJobHandler
- EventFromProcessInstanceHandler
- EventFromProcessMessageSubscriptionHandler

## Related issues

closes #23621


mustafadagher:
💭 I noticed that all these event handlers share many common logic. like how they persist, and the es scripts are exactly the same.

PS: I converted the scripts from `String.format` to `Block Texts`, now they're more concise and easier to read.

sdorokhova:
@mustafadagher I see now the duplications that you mentioned. And they are actually excessive. In importer [we have only one method to create the update field map](https://github.com/camunda/camunda/blob/main/operate/importer-8_7/src/main/java/io/camunda/operate/zeebeimport/processors/EventZeebeRecordProcessor.java#L413) and reuse it for all cases. It looks like we multiplicated the method when moving to export handlers. We can improve that to at least update only those fields that make sense for specific record type, e.g. message event [may only bring two metadata fields](https://github.com/camunda/camunda/pull/23700/files#diff-a549dd76df21afdc902ef5d7c07cd2655edab49f7386ddea09445c6be5c2b77dR104-R105): message name and correlation key. So we need to include only those in update fields map and in script. 

Do you think you could analyze and clean up this within current PR? Otherwise we could create an issue to look into this later.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/EventFromIncidentHandler.java
```diff
@@ -0,0 +1,222 @@
...
+    final Map<String, Object> jsonMap = new HashMap<>();
+    jsonMap.put(KEY, entity.getKey());
+    jsonMap.put(EVENT_SOURCE_TYPE, entity.getEventSourceType());
+    jsonMap.put(EVENT_TYPE, entity.getEventType());
+    jsonMap.put(DATE_TIME, entity.getDateTime());
+    jsonMap.put(PROCESS_KEY, entity.getProcessDefinitionKey());
+    jsonMap.put(BPMN_PROCESS_ID, entity.getBpmnProcessId());
+    jsonMap.put(FLOW_NODE_ID, entity.getFlowNodeId());
+    jsonMap.put(positionFieldName, positionFieldValue);
+    if (entity.getMetadata() != null) {
```
sdorokhova, line 136: I don't see unit tests for updating the metadata fields. Could we may be add couple of fields for every record type?

sdorokhova (approved):
Looks good to me! Thank you for reworking the scripts, looks much better!

Just one minor comment in code about unit tests.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/EventFromIncidentHandler.java
mustafadagher: it is there, the last few lines in `testUpdateEntity` in each handler test

mustafadagher:
> @mustafadagher I see now the duplications that you mentioned. And they are actually excessive. In importer [we have only one method to create the update field map](https://github.com/camunda/camunda/blob/main/operate/importer-8_7/src/main/java/io/camunda/operate/zeebeimport/processors/EventZeebeRecordProcessor.java#L413) and reuse it for all cases. It looks like we multiplicated the method when moving to export handlers. We can improve that to at least update only those fields that make sense for specific record type, e.g. message event [may only bring two metadata fields](https://github.com/camunda/camunda/pull/23700/files#diff-a549dd76df21afdc902ef5d7c07cd2655edab49f7386ddea09445c6be5c2b77dR104-R105): message name and correlation key. So we need to include only those in update fields map and in script.
> 
> Do you think you could analyze and clean up this within current PR? Otherwise we could create an issue to look into this later.

@sdorokhova I just pushed a commit to remove duplication, can you please check?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/EventFromIncidentHandler.java
mustafadagher: or do you mean on flush?

mustafadagher: done

sdorokhova (approved):
Thank you! Looks perfect! Will merge it

Issue #23127, opened 2024-10-07
Configuration parameter to enable resource permissions checks


houssain-barouni:
### Description

We need to a configuration parameter that allows an admin to enable/disable resource permission checks

In V1, `CAMUNDA_TASKLIST_IDENTITY_RESOURCE_PERMISSIONS_ENABLED` and `CAMUNDA_OPERATE_IDENTITY_RESOURCE_PERMISSIONS_ENABLED` are respecively used for Tasklist and Operate

In V2, `ZEEBE_BROKER_EXPERIMENTAL_ENGINE_AUTHORIZATIONS_ENABLEAUTHORIZATION` was [added](https://github.com/camunda/camunda/pull/22312/commits/ba4485628a428fb268a65ceb386799ff6bcd69db) to check the authorization

### Breakdown
- [x] new configuration parameter ~~that can be used E2E in V2, Zeebe engine and Gateway and everywhere else~~ #23719
- [x] https://github.com/camunda/camunda-platform-helm/pull/2593
- [x] Update camunda-operator: https://github.com/camunda-cloud/camunda-operator/pull/3401

PR #23719, opened 2024-10-18
Configuration parameter to enable resource permissions checks

houssain-barouni:
## Description

This PR introduces a new configuration parameter, `camunda.security.authorizations.enabled`, which enables or disables the permission checks feature (disabled by default). 

In the SaaS, this setting will be controlled by the cluster admin via a cluster-level configuration (see image attached). Updates to the Camunda Operator and Helm charts will follow later.
image

In V1, this setting sets the `CAMUNDA_TASKLIST_IDENTITY_RESOURCE_PERMISSIONS_ENABLED` and `CAMUNDA_OPERATE_IDENTITY_RESOURCE_PERMISSIONS_ENABLED` environment variables.

I would like to generalize this parameter to also cover the Zeebe engine (which is currently managed through `ZEEBE_BROKER_EXPERIMENTAL_ENGINE_AUTHORIZATIONS_ENABLEAUTHORIZATION`). This would streamline configuration for customers by providing a unified parameter for the full application (especially for the C8 single-app).
I still need to evaluate the feasibility of this change.

For easier review, you can refer to this commit: https://github.com/camunda/camunda/pull/23719/commits/1b61244d8dd5ba40e74c8027cd56a3b4f5d5e06c.

The other commit includes the addition of the configuration across all services and the removal of an unnecessary generic type in `AuthorizationServices`.

## Checklist

- [ ] for CI changes:

## Related issues

relates to #23127

### DIFF
--- a/authentication/pom.xml
+++ b/authentication/pom.xml
@@ -102,10 +102,6 @@
     </dependency>
     <dependency>
       <groupId>io.camunda</groupId>
-      <artifactId>zeebe-protocol-impl</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>io.camunda</groupId>
       <artifactId>camunda-security-core</artifactId>
     </dependency>
     <dependency>
--- a/authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
+++ b/authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
@@ -14,7 +14,6 @@ import io.camunda.authentication.handler.AuthFailureHandler;
 import io.camunda.authentication.handler.CustomMethodSecurityExpressionHandler;
 import io.camunda.service.AuthorizationServices;
 import io.camunda.service.UserServices;
-import io.camunda.zeebe.protocol.impl.record.value.authorization.AuthorizationRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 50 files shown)


remcowesterhoud (approved):
Changes look good to me 👍 

You mention:
>I would like to generalize this parameter to also cover the Zeebe engine (which is currently managed through ZEEBE_BROKER_EXPERIMENTAL_ENGINE_AUTHORIZATIONS_ENABLEAUTHORIZATION). This would streamline configuration for customers by providing a unified parameter for the full application (especially for the C8 single-app).

I agree we should streamline this. Will you also take care of replacing the configuration on the Zeebe-side?

houssain-barouni:
> Changes look good to me 👍
> 
> You mention:
> 
> > I would like to generalize this parameter to also cover the Zeebe engine (which is currently managed through ZEEBE_BROKER_EXPERIMENTAL_ENGINE_AUTHORIZATIONS_ENABLEAUTHORIZATION). This would streamline configuration for customers by providing a unified parameter for the full application (especially for the C8 single-app).
> 
> I agree we should streamline this. Will you also take care of replacing the configuration on the Zeebe-side?

I think this is a broader topic that can also include other areas: like unifying ES/OS configuration used in the services and also by the exporter. I think it should be handled as an epic by itself. WDYT @romansmirnov @tmetzke ?

Issue #23735, opened 2024-10-21
[Bug] Set ExternalFormReference as formKey for Zeebe User Taks - V1 Tasklist


nathansandi:
**Description**

On Job workers, the V1 endpoint has the behaviour to set as formKey the formReference on the importers, this should be the same behaviour for the Zeebe User Tasks on the persistence.

Expected behaviour:
- Model a Zeebe User Task with `externalFormReference`
- Make a call on V1
- `externalFormReference` should be showed in the response

Incident
- `externalFormReference` is not shown as formKey for V1 - Zeebe User Tasks

As for the old data this is not imported, so to adjust V1, only set on the response the externalFormReference from the entity to the response. Once for Zeebe User tasks the behaviour

PR #23750, opened 2024-10-21
fix: externalFormReference on zeebe user tasks

nathansandi:
## Description

- recordValue was bringing externalFormReference as empty string and persisted as empty string on ES/OS, the fix set the importer to set as null for Zeebe User Tasks (This will be fixed on Data exporters level)
- Set  externalFormReference as a formKey for Zeebe User Task, when existent, reproducing the behaviour of V1

## Related issues

closes #https://github.com/camunda/camunda/issues/23735


tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/mapper/TaskMapper.java
```diff
@@ -43,7 +43,11 @@ public TaskSearchResponse toTaskSearchResponse(final TaskDTO taskDTO) {
             .setTaskState(taskDTO.getTaskState())
             .setSortValues(taskDTO.getSortValues())
             .setIsFirst(taskDTO.getIsFirst())
-            .setFormKey(taskDTO.getFormKey())
+            .setFormKey(
+                (taskDTO.getExternalFormReference() == null
+                        || taskDTO.getExternalFormReference().isBlank())
+                    ? taskDTO.getFormKey()
+                    : taskDTO.getExternalFormReference())
```
houssain-barouni, line 50: the reason of not having a dedicated field for `externalFormReference` is to have something similar to JobWorkers ? @nathansandi

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/mapper/TaskMapper.java
```diff
@@ -43,7 +43,11 @@ public TaskSearchResponse toTaskSearchResponse(final TaskDTO taskDTO) {
             .setTaskState(taskDTO.getTaskState())
             .setSortValues(taskDTO.getSortValues())
             .setIsFirst(taskDTO.getIsFirst())
-            .setFormKey(taskDTO.getFormKey())
+            .setFormKey(
```
houssain-barouni, line 46: the same needs to be done in `TaskResponse toTaskResponse(TaskDTO)` below

houssain-barouni (changes requested):

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/mapper/TaskMapper.java
nathansandi: On V2 yes, on V1 I am reproducing the same behaviour as Job Workers (where set the external form on formKey), where if there is external form, this is set as the form key. Otherwise the developer would need create a condition on their side to check it, considering in 8.6 he has support still for both Job Workers, and Zeebe User Tasks.

houssain-barouni (approved):

PR #23754, opened 2024-10-21
docs: add camunda exporter configs to `broker template` and `broker standalone template`

### LINKED ISSUE #23633: Document CamundaExporter config in Broker config
## Description

We maintain all possible configurations in our template files in the dist module. This allows users to easily understand what is possible to configure and provides a good reference.

We are currently missing this part.

## AT: 

All possible configuration should be reflected and documented here 

- https://github.com/camunda/camunda/blob/main/dist/src/main/config/broker.standalone.yaml.template
- https://github.com/camunda/camunda/blob/main/dist/src/main/config/broker.yaml.template

## Example

Example for Elasticsearch https://github.com/camunda/camunda/blob/main/dist/src/main/config/broker.standalone.yaml.template#L835

mustafadagher:
## Description

Added camunda exporter configs to `broker template` and `broker standalone template`.

**PS:** I only added the main configs that should be needed or the ones that can at least have a default value. Some configs like `index.replicasByIndexName`, `index.shardsByIndexName` and `connect.interceptorPlugins` are not documented there since it's difficult to come up with an example

## Related issues

closes #23633

### DIFF
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -1015,6 +1015,7 @@
         #     prefix:
         #     numberOfShards: 3
         #     numberOfReplicas: 0
+        #     variableSizeThreshold: 8191
         #
         #   retention:
         #     enabled: false
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -983,6 +983,46 @@
         #     variable: true
         #     variableDocument: true
 
+      # Camunda Exporter ----------
+      # An example configuration for the camunda exporter:
+      #
+      # These setting can also be overridden using the environment variables "ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_..."
+      #
+      # camundaExporter:
+        # className: io.camunda.exporter.CamundaExporter
+        #
+        # args:
+        #   connect:
+        #     type: elasticsearch
+        #     url: http://localhost:9200
+        #     clusterName: elasticsearch
+        #     dateFormat: yyyy-MM-dd'T'HH:mm:ss.SSSZZ
+        #     socketTimeout: 1000
+        #     connectTimeout: 1000
+        #     username: elastic
+        #     password: changeme
+        #     security:
+        #       enabled: false
+        #       certificatePath: /path/to/certificate
+        #       verifyHostname: true
+        #       selfSigned: false
+        #
+        #   bulk:
+        #     delay: 5
+        #     size: 1000
+        #
+        #   index:
+        #     prefix:
+        #     numberOfShards: 3
+        #     numberOfReplicas: 0
+        #
+        #   retention:
+        #     enabled: false
+        #     minimumAge: 30d
+        #     policyName: camunda-retention-policy
+        #
+        #   createSchema: true
+
     # processing:
       # Sets the maximum number of commands that processed within one batch.
       # The processor will process until no more follow up commands are created by the initial command
--- a/dist/src/main/config/broker.yaml.template
+++ b/dist/src/main/config/broker.yaml.template
@@ -858,6 +858,47 @@
         #     variable: true
         #     variableDocument: true
 
+      # Camunda Exporter ----------
+      # An example configuration for the camunda exporter:
+      #
+      # These setting can also be overridden using the environment variables "ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_..."
+      #
+      # camundaExporter:
+        # className: io.camunda.exporter.CamundaExporter
+        #
+        # args:
+        #   connect:
+        #     type: elasticsearch
+        #     url: http://localhost:9200
+        #     clusterName: elasticsearch
+        #     dateFormat: yyyy-MM-dd'T'HH:mm:ss.SSSZZ
+        #     socketTimeout: 1000
+        #     connectTimeout: 1000
+        #     username: elastic
+        #     password: changeme
+        #     security:
+        #       enabled: false
+        #       certificatePath: /path/to/certificate
+        #       verifyHostname: true
+        #       selfSigned: false
+        #
+        #   bulk:
+        #     delay: 5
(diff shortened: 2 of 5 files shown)


dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -983,6 +983,47 @@
...
+        #     delay: 5
+        #     size: 1000
+        #
+        #   index:
+        #     prefix:
+        #     numberOfShards: 3
+        #     numberOfReplicas: 0
+        #     variableSizeThreshold: 8191
+        #
+        #   retention:
```
Zelldon, line 1020: ❓ Is it expected that we don't have a policy description? 

See for OS exporter https://github.com/camunda/camunda/pull/23754/files#diff-6783f1e71cace45bffaa5446fc6bde1f63deb6ae0ff99fd28eeae7e659881253L930

dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -983,6 +983,47 @@
...
+        #     enabled: false
+        #     minimumAge: 30d
+        #     policyName: camunda-retention-policy
+        #
+        #   createSchema: true
+
```
Zelldon, line 1026: ❓ In the Opensearch exporter we have a `aws` section do you know the use case for such?

https://github.com/camunda/camunda/pull/23754/files#diff-6783f1e71cace45bffaa5446fc6bde1f63deb6ae0ff99fd28eeae7e659881253R936-L939

Do we need this?

dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -983,6 +983,47 @@
...
+        # className: io.camunda.exporter.CamundaExporter
+        #
+        # args:
+        #   connect:
+        #     type: elasticsearch
+        #     url: http://localhost:9200
+        #     clusterName: elasticsearch
+        #     dateFormat: yyyy-MM-dd'T'HH:mm:ss.SSSZZ
+        #     socketTimeout: 1000
+        #     connectTimeout: 1000
```
Zelldon, line 1001: ❓ What is the difference here?

What is the difference to the request timeout we had before? https://github.com/camunda/camunda/pull/23754/files#diff-6783f1e71cace45bffaa5446fc6bde1f63deb6ae0ff99fd28eeae7e659881253L845

Can we also specify the unit in the name? Otherwise it is not fully clear.

Like socketTimeoutMs

dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -983,6 +983,47 @@
...
+        #     password: changeme
+        #     security:
+        #       enabled: false
+        #       certificatePath: /path/to/certificate
+        #       verifyHostname: true
+        #       selfSigned: false
+        #
+        #   bulk:
+        #     delay: 5
+        #     size: 1000
```
Zelldon, line 1012: It would be good if we rename it to something like `delaySeconds` and `sizeBytes` as otherwise it is not clear what unit we expect here.

ChrisKujawa (commented):
Hey @mustafadagher thanks!

I know some comments might be quite late in the process, but I hadn't a chance to look at the code and configs earlier. Would be great if you can take a look at them. 

I know when have not really documentation/explanation on the ES/OS exporter, but I think would be good to have a bit better description what each property does. WDYT?

dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -983,6 +983,47 @@
...
+      #
+      # These setting can also be overridden using the environment variables "ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_..."
+      #
+      # camundaExporter:
+        # className: io.camunda.exporter.CamundaExporter
+        #
+        # args:
+        #   connect:
+        #     type: elasticsearch
+        #     url: http://localhost:9200
```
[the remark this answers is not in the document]
mustafadagher: yes, if `security. enabled` is set to `true` below

dist/src/main/config/broker.standalone.yaml.template
mustafadagher: this os not `bytes` by the way, it's the number of records, and I inherited the old exporter naming convention as is here.

dist/src/main/config/broker.standalone.yaml.template
mustafadagher: right now in the `search-client-connect` they initialize the open search connector with a `DefaultCredentialsProvider.create()`.
We could extend our configs to provide custom aws configurations and provide our own `AwsCredentialsProvider` too. 
I just followed their example in `OpenSearchSearchClientConfiguration` for simplicity for now. Yet, I'm not sure if the default credential provider should always work out of the box.

mustafadagher:
> I know when have not really documentation/explanation on the ES/OS exporter, but I think would be good to have a bit better description what each property does. WDYT?

@Zelldon I know that such descriptions exist for other configs indeed. But to the contrary, I think otherwise. It's an overkill to document each field where should be a config template, when we actually have a documentation page with detailed descriptions for the same in a much cleaner format.

Zelldon:
> > I know when have not really documentation/explanation on the ES/OS exporter, but I think would be good to have a bit better description what each property does. WDYT?
> 
> @Zelldon I know that such descriptions exist for other configs indeed. But to the contrary, I think otherwise. It's an overkill to document each field where should be a config template, when we actually have a documentation page with detailed descriptions for the same in a much cleaner format.

Thanks @mustafadagher. Yeah in the past these templates were the actual documentation and the single point of truth. I think we should at least have one place to document them. My thought was we could do it here, and then copy it over. Personally I normally check in the templates what config exists and how the ENV name is, etc. Not sure whether this is the "normal". 

>when we actually have a documentation page with detailed descriptions for the same in a much cleaner format.

Zelldon:
Ah I saw now 👍🏼 https://github.com/camunda/camunda-docs/pull/4500

ChrisKujawa (approved):
🚀 Thanks @mustafadagher

PR #23764, opened 2024-10-21
feat: support session persistence with ES

### LINKED ISSUE #23760: Implement Session Entity to store session information

maryarm:
## Description

In the current implementation, the Operate and Tasklist components use ES to support sharing sessions for persistent sessions. This is to support a pod shutting down and not forcing the user to login again. We need to be able to consider this within the Camunda Authentication module such that if a user has an existing session then this is loaded into the security context.

## Related issues

closes #23760 #23761 #23762 #23763


maryarm:
Hey @camunda/hto-backend, I've added you as the reviewer because this task is to move the existing mechanism for web session storage to new arch. Please take a look and also feel free to add relevant interviewers.

authentication/src/main/java/io/camunda/authentication/handler/session/SessionDocumentMapper.java
```diff
@@ -0,0 +1,156 @@
...
+      if (attributesObject instanceof Map) {
+        final Map<String, Object> attributes = (Map<String, Object>) document.get(ATTRIBUTES);
+        attributes
+            .keySet()
+            .forEach(
+                name -> {
+                  if (attributes.get(name) instanceof String) {
+                    session.setAttribute(
+                        name,
+                        deserialize(Base64.getDecoder().decode((String) attributes.get(name))));
```
nathansandi, line 105: 🔧  Should we add a try/catch in order to catch possible deserialisation errors over here and log that?

search/search-client-elasticsearch/src/main/java/io/camunda/search/es/clients/ElasticsearchSessionDocumentClient.java
```diff
@@ -0,0 +1,66 @@
...
+  private final String indexName = String.format("%s-%s-%s_", "camunda", "web-session", "1.1.0");
+
+  public ElasticsearchSessionDocumentClient(final RetryElasticsearchClient client) {
+    this.client = client;
+  }
+
+  public void setup() {
+    final Map<String, Object> indexDescription =
+        ElasticsearchJSONUtil.readJSONFileToMap("/camunda-web-session.json");
+    client.createIndex(
```
nathansandi, line 37: 🔧  Should we add try/catch in order to log possible errors over the **_ES operations_** (in general) ?

authentication/src/main/java/io/camunda/authentication/handler/session/NoOpSessionDocumentStorageClient.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class NoOpSessionDocumentStorageClient implements SessionDocumentStorageClient {
+  private final Map<String, Map<String, Object>> sessionDocuments =
+      Collections.synchronizedMap(new HashMap<>());
+
+  @Override
+  public void consumeSessions(final Consumer<Map<String, Object>> sessionConsumer) {
```
nathansandi, line 21: 💯 Nice

.idea/codeStyles/Project.xml
```diff
@@ -454,7 +454,6 @@
       </indentOptions>
     </codeStyleSettings>
     <codeStyleSettings language="Python">
-      <option name="KEEP_BLANK_LINES_IN_CODE" value="1" />
```
nathansandi, line 457: ❓ Is this change will be bring on this PR?

nathansandi (approved):
Overall, looks good to me
I left some suggestions and questions related to the implementation, and potential improvements.

dist/src/main/java/io/camunda/application/commons/search/ElasticSearchSearchClientConfiguration.java
```diff
@@ -31,4 +37,30 @@ public ElasticsearchSearchClient elasticsearchSearchClient(
     final var elasticsearch = connector.createClient();
     return new ElasticsearchSearchClient(elasticsearch);
   }
+
+  @Bean(destroyMethod = "close")
+  public RestHighLevelClient restHighLevelClient(final SearchClientProperties configuration) {
```
houssain-barouni, line 42: we are actually trying to get rid of ElasticsearchClient 7 dependency (which brings the `RestHighLevelClient`), is it possible to use `co.elastic.clients.elasticsearch.ElasticsearchClient`?

houssain-barouni (changes requested):
❌ Including the `elasticsearch-rest-high-level-client` dependency in the search module is not an option. 
Ideally, `SessionDocumentStorageClient` should have an implementation that is abstracted from ES/OS, though I understand this may require additional effort. 
That said, we should avoid reusing `RetryElasticsearchClient` and instead use `co.elastic.clients.elasticsearch.ElasticsearchClient` from `elasticseach-java` dependency.

search/search-client-elasticsearch/src/main/resources/camunda-web-session.json
```diff
@@ -0,0 +1,23 @@
+{
```
Zelldon, line 1: ❓ Q: Shouldn't this be part of the webapps-schema module ?

https://github.com/camunda/camunda/tree/main/webapps-schema/src/main/resources/schema/elasticsearch/create/index

houssain-barouni (changes requested):
Hi @maryarm 
I am not fully convinced that having a separate Elasticsearch client within the search module solely for session storage is the best approach. I see two alternative options:
1. Manage the session index like any other index, with index creation handled by the Camunda exporter or through a shared module with the exporter for schema management.
2. Treat this index as an internal asset of the authentication module. So you create, inside the authentication module, an ES client that allows you to create the index, and read, write documents in it

@romansmirnov I would like to have your opinion about these kind of indices that need to be managed and fed outside the exporter?

maryarm:
> Hi @maryarm I am not fully convinced that having a separate Elasticsearch client within the search module solely for session storage is the best approach. I see two alternative options:
> 
> 1. Manage the session index like any other index, with index creation handled by the Camunda exporter or through a shared module with the exporter for schema management.
> 2. Treat this index as an internal asset of the authentication module. So you create, inside the authentication module, an ES client that allows you to create the index, and read, write documents in it
> 
> @romansmirnov I would like to have your opinion about these kind of indices that need to be managed and fed outside the exporter?

The first option sounds like a good idea to me (move all schema creation stuff to a shared module), but I find it a big bigger than scope of this PR. 
For the second approach should then we also create an OpensearchClient within authentication module? I mean in my design we have a simple in-memory implementation in authentication module but the other implementation comes in search module based on the configured storage. Maybe the search is not exactly the match but in existing modules I can't find any other relevant module.
that would be nice to hear your ideas here @romansmirnov.

search/search-client-elasticsearch/src/main/java/io/camunda/search/es/clients/ElasticsearchSessionDocumentClient.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.function.Consumer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ElasticsearchSessionDocumentClient implements SessionDocumentStorageClient {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(ElasticsearchSessionDocumentClient.class);
+  private final RetryElasticsearchClient client;
+  private final String indexName = String.format("%s-%s-%s_", "camunda", "web-session", "1.1.0");
```
Zelldon, line 24: ❌ We must support with 8.7 one global prefix https://github.com/camunda/camunda/issues/23807

Zelldon:
> 1. Manage the session index like any other index, with index creation handled by the Camunda exporter or through a shared module with the exporter for schema management.

I think it makes sense to use the Camunda Exporter to do the schema management (creation and validation), I think it is good to have this in one place only. So we don't have this kindof logic everywhere distributed.

romansmirnov:
> I think it makes sense to use the Camunda Exporter to do the schema management (creation and validation), I think it is good to have this in one place only. So we don't have this kindof logic everywhere distributed.

To avoid doing the same multiple times and to avoid having the logic spread across multiple modules, I agree with @Zelldon and @houssain-barouni to have the index creation as part of the schema management in the camunda exporter.

authentication/src/main/java/io/camunda/authentication/config/SessionRepositoryConfig.java
```diff
@@ -0,0 +1,54 @@
...
+import jakarta.servlet.http.HttpServletRequest;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+import org.springframework.session.SessionRepository;
+import org.springframework.session.config.annotation.web.http.EnableSpringHttpSession;
+
+@Configuration
```
romansmirnov, line 25: All Spring configurations must be centralized in the `dist` module.

dist/src/main/java/io/camunda/application/commons/search/SearchClientDatabaseConfiguration.java
```diff
@@ -38,11 +46,40 @@ public class SearchClientDatabaseConfiguration {
...
+
+  @Bean
+  @ConditionalOnProperty(
+      prefix = "camunda.database",
+      name = "type",
+      havingValue = "elasticsearch",
+      matchIfMissing = true)
+  @ConditionalOnSessionPersistence
+  @Primary
+  public ElasticsearchSessionDocumentClient elasticsearchSessionDocumentClient(
```
romansmirnov, line 74: Why do we need this separation? Why do we need a `RetryElasticsearchClient`?

authentication/src/main/java/io/camunda/authentication/config/SessionRepositoryConfig.java
maryarm: If it's the standard then Ok, but as we have WebSecurityConfig in authentication module for a long time, should it be moved either?

search/search-client-elasticsearch/src/main/java/io/camunda/search/es/clients/ElasticsearchSessionDocumentClient.java
maryarm: Good point, I have renamed the index to start with identity as I didn't want to add a new component.

PR #23767, opened 2024-10-21
ci: check for merge commits in PRs

tmetzke:
## Description

Check PRs to not contain any merge commits.

## Checklist

- [x] for CI changes:

## Related issues

n/a

### DIFF
--- a/.github/workflows/no-merge-commits.yml
+++ b/.github/workflows/no-merge-commits.yml
@@ -1,0 +1,17 @@
+name: No merge commits
+on: [pull_request]
+
+jobs:
+  check-for-merge-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - id: depth
+        name: Determine fetch depth for checkout
+        run: echo "depth=$(expr ${{ github.event.pull_request.commits }} + 1)" >> $GITHUB_OUTPUT
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


korthout (changes requested):
❌This would also fail when two branches are merged together and then need to be merged to main. For example, hot fixes on the release branch before it is merged back to main (or the relevant stable branch). I believe that this is a blocker. 

We should somehow check whether the parents exist anywhere on the base branch, which may require an unlimited depth.

To speed up the checkout without the ability to limit the depth, we could ensure that it fetches only the commit tree, and not the actual changes.

I had tried something along those lines already https://github.com/camunda/camunda/pull/23274#issuecomment-2404506465

tmetzke:
> ❌This would also fail when two branches are merged together and then need to be merged to main. For example, hot fixes on the release branch before it is merged back to main (or the relevant stable branch). I believe that this is a blocker.
> 
> We should somehow check whether the parents exist anywhere on the base branch, which may require an unlimited depth.
> 
> To speed up the checkout without the ability to limit the depth, we could ensure that it fetches only the commit tree, and not the actual changes.
> 
> I had tried something along those lines already [#23274 (comment)](https://github.com/camunda/camunda/pull/23274#issuecomment-2404506465)

Very good insights, @korthout, thanks a lot for that 👍
I also wasn't aware that you already gave it a try 💡
Happy to join forces here at some point but maybe not now then.

PR #23768, opened 2024-10-21
[CPT] Print process instance data if a test fails

### LINKED ISSUE #23092: [CPT] Print all relevant process instance data if a test fails
## Description

If a process test fails, the failed assertion prints specific information, for example, the state of a process instance (e.g. `expected completed but was active`). I need more information to debug the failed test cases, such as the process instance variables and open incidents.

## Tasks
- Print all process instance variables (global)
- Print open incidents

For example:
```
Process test results:
=====================

Process instance: 123 [process-id: 'process-a']

Variables:
- 'a': 1
- 'b': "foo"

Open incidents:
<None>
---------------------

Process instance: 46 [process-id: 'process-b']

Variables:
<None>

Open incidents:
- 'task-a' [type: JOB_NO_RETRIES] "No retries left."
=====================
```

saig0:
## Description

If a test fails, the extension/execution listener prints the following data:
- All process instances with key and process-id
- All (global) process instance variables 
- Open incidents

Reviewer notes:
- The JUnit TestWatcher could be a good fit but this doesn't work with the Spring execution listener.
- We can use the process test result later to generate the process coverage report.  

Test output:

```
08:12:56.349 [main] INFO  io.camunda.process.test.impl.runtime.CamundaContainerRuntime - Starting Camunda container runtime
08:13:30.884 [main] INFO  io.camunda.process.test.impl.runtime.CamundaContainerRuntime - Camunda container runtime started in PT34.528894661S
08:13:42.657 [main] INFO  io.camunda.process.test.impl.runtime.CamundaContainerRuntime - Stopping Camunda container runtime
08:13:43.878 [main] INFO  io.camunda.process.test.impl.runtime.CamundaContainerRuntime - Camunda container runtime stopped in PT1.221388236S
Process test results:
=====================

Process instance: 2251799813685251 [process-id: 'process']

Variables:
- 'status': "active"

Open incidents:
- 'task' [type: EXTRACT_VALUE_ERROR] "Expected result of the expression 'due_date' to be one of '[DURATION, PERIOD, STRING]', but was 'NULL'. The evaluation reported the following warnings:
[NO_VARIABLE_FOUND] No variable found with name 'due_date'"
=====================


java.lang.AssertionError: Process instance [key: 2251799813685251] should be completed but was active.

```

## Related issues

closes #23092

### DIFF
--- a/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/client/CamundaApiClient.java
+++ b/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/client/CamundaApiClient.java
@@ -35,8 +35,10 @@ public class CamundaApiClient {
 
   // Operate v1 endpoints
   private static final String PROCESS_INSTANCE_GET_ENDPOINT = "/v1/process-instances/%d";
+  private static final String PROCESS_INSTANCE_SEARCH_ENDPOINT = "/v1/process-instances/search";
   private static final String FLOW_NODE_INSTANCES_SEARCH_ENDPOINT = "/v1/flownode-instances/search";
   private static final String VARIABLES_SEARCH_ENDPOINT = "/v1/variables/search";
+  private static final String INCIDENTS_GET_ENDPOINT = "/v1/incidents/%d";
 
   private final ObjectMapper objectMapper =
       new ObjectMapper().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
@@ -100,6 +102,14 @@ public class CamundaApiClient {
     return objectMapper.readValue(body, ProcessInstanceDto.class);
   }
 
+  public SearchProcessInstanceResponseDto findProcessInstances() throws IOException {
+    ensureAuthenticated();
+
+    final String requestBody = "{}";
+    final String responseBody = sendPostRequest(PROCESS_INSTANCE_SEARCH_ENDPOINT, requestBody);
+    return objectMapper.readValue(responseBody, SearchProcessInstanceResponseDto.class);
+  }
+
   public FlowNodeInstancesResponseDto findFlowNodeInstancesByProcessInstanceKey(
       final long processInstanceKey) throws IOException {
     ensureAuthenticated();
@@ -120,6 +130,13 @@ public class CamundaApiClient {
             processInstanceKey, processInstanceKey);
     final String responseBody = sendPostRequest(VARIABLES_SEARCH_ENDPOINT, requestBody);
     return objectMapper.readValue(responseBody, VariableResponseDto.class);
+  }
+
+  public IncidentDto getIncidentByKey(final long incidentKey) throws IOException {
+    ensureAuthenticated();
+
+    final String body = sendGetRequest(String.format(INCIDENTS_GET_ENDPOINT, incidentKey));
+    return objectMapper.readValue(body, IncidentDto.class);
   }
 
   private String sendGetRequest(final String endpoint) throws IOException {
--- a/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/testresult/CamundaProcessTestResultPrinter.java
+++ b/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/testresult/CamundaProcessTestResultPrinter.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright © 2017 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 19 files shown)


testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/client/CamundaApiClient.java
```diff
@@ -35,8 +35,10 @@ public class CamundaApiClient {
 
   // Operate v1 endpoints
   private static final String PROCESS_INSTANCE_GET_ENDPOINT = "/v1/process-instances/%d";
+  private static final String PROCESS_INSTANCE_SEARCH_ENDPOINT = "/v1/process-instances/search";
```
nicpuppa, line 38: ❌  Why `v1` ? It should be `v2`. If I'm not wrong the search endpoint are available only under `v2`. Can we also update the other path to use the new API version (not sure if it's going to be a breaking change)

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/client/CamundaApiClient.java
```diff
@@ -100,6 +102,14 @@ public ProcessInstanceDto getProcessInstanceByKey(final long processInstanceKey)
     return objectMapper.readValue(body, ProcessInstanceDto.class);
   }
 
+  public SearchProcessInstanceResponseDto findProcessInstances() throws IOException {
+    ensureAuthenticated();
+
+    final String requestBody = "{}";
+    final String responseBody = sendPostRequest(PROCESS_INSTANCE_SEARCH_ENDPOINT, requestBody);
+    return objectMapper.readValue(responseBody, SearchProcessInstanceResponseDto.class);
```
nicpuppa, line 110: ❓ Why did you map the response into another DTO object ? Can we re-use the one available from the client ? In this case will be `ProcessInstanceSearchQueryResponse`

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/testresult/CamundaProcessTestResultPrinter.java
```diff
@@ -0,0 +1,103 @@
...
+    this.printStream = printStream;
+  }
+
+  public void print(final ProcessTestResult result) {
+    final String formattedResult = formatResult(result);
+    printStream.accept(formattedResult);
+  }
+
+  private String formatResult(final ProcessTestResult result) {
+    return "Process test results:\n"
```
nicpuppa, line 41: ❓ Do you think `StringUtils` can be helpful here ? Improving maintainability.

nicpuppa (changes requested):
Great job, I left some questions and a change request. Please have a look 👍

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/client/CamundaApiClient.java
saig0: Currently, CPT uses the Operate endpoint (v1): https://docs.camunda.io/docs/next/apis-tools/operate-api/specifications/search-1/.

When the new query API (v2) is working, I would love to switch to the new API (v2) and use the SDK. However, the auth issue is blocking us at the moment. 

For now, let's stay at the old v1 API. :sweat_smile:

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/client/CamundaApiClient.java
saig0: At the moment, CPT invokes the v1 Operate API. The client DTOs are part of the v2 API.

When we switch to the new API and use the SDK, we can remove the DTOs from CPT. The new `SearchProcessInstanceResponseDto` uses the already existing `ProcessInstanceDto`. 

For now, let's keep the internal CPT DTOs.

nicpuppa: Yep, sorry for it

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/testresult/CamundaProcessTestResultPrinter.java
saig0: I have no idea how to utilize `StringUtils` here. I'm open to your ideas. :+1:

nicpuppa: Sorry, I meant `StringBuilder` 🙏

nicpuppa (approved):

PR #23778, opened 2024-10-21
Add RoleDeleteProcessor

### LINKED ISSUE #22934: Process `Role.DELETE` command
<h2>Description</h2>

image

1. Create new class `RoleDeleteProcessor` and implement `AuthorizableDistributionProcessor.Authorizable <RoleRecord>`
    - Check if the role exists. If no, reject the command with `NOT_FOUND`. If yes continue.
    - Write a `Role.DELETED` event and respond to the client
    - Write a `Role.DELETED` event
        - Resource type: `ROLE`
        - Permission type: `DELETE`
        - Resource: `RoleRecord`
5. Unit test the Role.DELETED command
6. Unit test the Role.DELETED command distribution

nicpuppa:
## Description

This PR adds the `RoleDeleteProcessor`

## Related issues

closes #22934

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
@@ -41,7 +41,7 @@ public class DeleteRoleMultiPartitionTest {
     // when
     final var name = UUID.randomUUID().toString();
     final var roleKey = engine.role().newRole(name).create().getValue().getRoleKey();
-    engine.role().deleteRole(roleKey).withName(name).delete();
+    engine.role().deleteRole(roleKey).delete();
 
     assertThat(
             RecordingExporter.records()
@@ -88,7 +88,7 @@ public class DeleteRoleMultiPartitionTest {
     // when
     final var name = UUID.randomUUID().toString();
     final var roleKey = engine.role().newRole(name).create().getValue().getRoleKey();
-    engine.role().deleteRole(roleKey).withName(name).delete();
+    engine.role().deleteRole(roleKey).delete();
 
     // then
     assertThat(
@@ -107,7 +107,7 @@ public class DeleteRoleMultiPartitionTest {
       interceptRoleCreateForPartition(partitionId);
     }
     final var roleKey = engine.role().newRole(name).create().getValue().getRoleKey();
-    engine.role().deleteRole(roleKey).withName(name).delete();
+    engine.role().deleteRole(roleKey).delete();
 
     // Increase time to trigger a redistribution
     engine.increaseTime(Duration.ofMinutes(1));
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.authorization;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.tuple;
+
+import io.camunda.zeebe.engine.state.distribution.DistributionQueue;
+import io.camunda.zeebe.engine.util.EngineRule;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.RecordType;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.intent.AuthorizationIntent;
+import io.camunda.zeebe.protocol.record.intent.CommandDistributionIntent;
+import io.camunda.zeebe.protocol.record.intent.RoleIntent;
+import io.camunda.zeebe.protocol.record.value.AuthorizationOwnerType;
+import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
+import io.camunda.zeebe.protocol.record.value.CommandDistributionRecordValue;
+import io.camunda.zeebe.protocol.record.value.PermissionAction;
+import io.camunda.zeebe.test.util.record.RecordingExporter;
+import io.camunda.zeebe.test.util.record.RecordingExporterTestWatcher;
+import java.time.Duration;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestWatcher;
+
+public class DeleteRoleMultiPartitionTest {
+
+  private static final int PARTITION_COUNT = 3;
+
+  @Rule public final EngineRule engine = EngineRule.multiplePartition(PARTITION_COUNT);
+
+  @Rule public final TestWatcher testWatcher = new RecordingExporterTestWatcher();
+
+  @Test
+  public void shouldDistributeRoleDeleteCommand() {
+    // when
+    final var name = UUID.randomUUID().toString();
+    final var roleKey = engine.role().newRole(name).create().getValue().getRoleKey();
+    engine
+        .authorization()
+        .permission()
+        .withOwnerKey(roleKey)
+        .withOwnerType(AuthorizationOwnerType.ROLE)
+        .withResourceType(AuthorizationResourceType.ROLE)
+        .withAction(PermissionAction.REMOVE)
+        .add();
+    engine.role().deleteRole(roleKey).withName(name).delete();
+
+    assertThat(
+            RecordingExporter.records()
+                .withPartitionId(1)
+                .limitByCount(
+                    record -> record.getIntent().equals(CommandDistributionIntent.FINISHED), 3))
+        .extracting(
+            Record::getIntent,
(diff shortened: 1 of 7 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/RoleTest.java
```diff
@@ -273,4 +281,47 @@ public void shouldRejectIfEntityIsNotPresentEntityRemoval() {
...
+
+    final var createdRole = roleRecord.getValue();
+    Assertions.assertThat(createdRole).isNotNull().hasFieldOrPropertyWithValue("name", name);
+
+    assertThat(notPresentUpdateRecord)
+        .hasRejectionType(RejectionType.NOT_FOUND)
+        .hasRejectionReason(
+            "Expected to delete role with key '"
+                + notPresentRoleKey
+                + "', but a role with this key doesn't exist.");
```
remcowesterhoud, line 325:
```suggestion
            "Expected to delete role with key '%s', but a role with this key doesn't exist."
                .formatted(notPresentRoleKey));
```

🔧

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/RoleTest.java
```diff
@@ -273,4 +281,47 @@ public void shouldRejectIfEntityIsNotPresentEntityRemoval() {
...
+
+    final var deletedRecord = roleState.getRole(roleKey);
+    assertTrue(deletedRecord.isEmpty());
+  }
+
+  @Test
+  public void shouldRejectIfRoleIsNotPresentOnDeletion() {
+    // given
+    final var name = UUID.randomUUID().toString();
+    final var roleRecord = ENGINE.role().newRole(name).create();
```
remcowesterhoud, line 310: ❌ I don't see a reason to create a role here. The test wants to verify the command is rejected if the role doesn't exist. Why do we need this role?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
```diff
@@ -0,0 +1,166 @@
...
+    final var name = UUID.randomUUID().toString();
+    final var roleKey = engine.role().newRole(name).create().getValue().getRoleKey();
+    engine
+        .authorization()
+        .permission()
+        .withOwnerKey(roleKey)
+        .withOwnerType(AuthorizationOwnerType.ROLE)
+        .withResourceType(AuthorizationResourceType.ROLE)
+        .withAction(PermissionAction.REMOVE)
+        .add();
```
remcowesterhoud, line 55: ❓ Why do we need to add a permission to the role? It shouldn't be required for deleting a role. Removing this also means we need to change the `limitByCount`

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
```diff
@@ -0,0 +1,166 @@
...
```
remcowesterhoud, line 110: ❓ Why do we need to add a permission to the role? It shouldn't be required for deleting a role. Removing this also means we need to change the `limitByCount`

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    engine
+        .authorization()
+        .permission()
+        .withOwnerKey(roleKey)
+        .withOwnerType(AuthorizationOwnerType.ROLE)
+        .withResourceType(AuthorizationResourceType.ROLE)
+        .withAction(PermissionAction.REMOVE)
+        .add();
```
remcowesterhoud, line 138: ❓ Why do we need to add a permission to the role? It shouldn't be required for deleting a role. Removing this also means we need to change the `limitByCount`

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/RoleTest.java
```diff
@@ -273,4 +281,47 @@ public void shouldRejectIfEntityIsNotPresentEntityRemoval() {
...
+        .authorization()
+        .permission()
+        .withOwnerKey(roleKey)
+        .withOwnerType(AuthorizationOwnerType.ROLE)
+        .withResourceType(AuthorizationResourceType.ROLE)
+        .withAction(PermissionAction.REMOVE)
+        .add();
+
+    // when
+    ENGINE.role().deleteRole(roleKey).withName(name).delete();
```
remcowesterhoud, line 300: ❌ Let's remove the `withName` part. We shouldn't need to provide a name upon deleting a Role.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
```diff
@@ -0,0 +1,166 @@
...
+    final var roleKey = engine.role().newRole(name).create().getValue().getRoleKey();
+    engine
+        .authorization()
+        .permission()
+        .withOwnerKey(roleKey)
+        .withOwnerType(AuthorizationOwnerType.ROLE)
+        .withResourceType(AuthorizationResourceType.ROLE)
+        .withAction(PermissionAction.REMOVE)
+        .add();
+    engine.role().deleteRole(roleKey).withName(name).delete();
```
remcowesterhoud, line 56: ❌ Let's remove the `withName` part. We shouldn't need to provide a name upon deleting a Role.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
```diff
@@ -0,0 +1,166 @@
...
```
remcowesterhoud, line 111: ❌ Let's remove the `withName` part. We shouldn't need to provide a name upon deleting a Role.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  @Test
+  public void distributionShouldNotOvertakeOtherCommandsInSameQueue() {
+    // when
+    final var name = UUID.randomUUID().toString();
+    final var roleKey = engine.role().newRole(name).create().getValue().getRoleKey();
+    for (int partitionId = 2; partitionId <= PARTITION_COUNT; partitionId++) {
+      interceptRoleCreateForPartition(partitionId);
+    }
```
remcowesterhoud, line 129: ❌ You need to swap the role creation with the interceptor, otherwise you create a race condition. So step 1 is saying intercept these things when you encounter them. Step 2 is actually creating the role, which results in the inter partition commands being send, and thus also intercepted.

remcowesterhoud (changes requested):
Thanks @nicpuppa 🙇 I've got a few comments in your tests, otherwise it looks good.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/RoleTest.java
nicpuppa: It's just testing purpose, but yes it can be removed 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
nicpuppa: It's required, because in the applier we are performing this operation on the authorization state:

```
    authorizationState.deleteAuthorizationsByOwnerKeyPrefix(roleKey);
    authorizationState.deleteOwnerTypeByKey(roleKey);
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
nicpuppa: At the moment the name is mandatory, because in the `delete` in the role state we are removing also the role from the column name. 
We can retrieve this information from the state in the processor, I wasn't sure about this approach tho

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
remcowesterhoud: I don't see why that would make adding a permission be required? `authorizationState.deleteAuthorizationsByOwnerKeyPrefix(roleKey);` is in a `whileEqualPrefix`. If there's no permissions added it will do nothing, which is fine.
`authorizationState.deleteOwnerTypeByKey(roleKey);` will work anyway as this CF is populated upon role creation. No need for a permission to be added here either.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/DeleteRoleMultiPartitionTest.java
remcowesterhoud: Oh I see where it went wrong. In the processor you are getting the role by the key. You can use this to set the name on the record. 

`record.setName(persistedRole.getName());`

Then you no longer need to provide the name in the tests. We need to change this, because in the REST API we won't get anything but a key, so we need to get the name ourselves.

remcowesterhoud (changes requested):
Please look at the comment I replied to

remcowesterhoud (approved):

PR #23784, opened 2024-10-21
test: when append fails a response is sent to the leader immediately

### LINKED ISSUE #8612: Follower doesn't send response back if append fails
**Describe the bug**

Taking a look at the code I saw that if the follower is failing to append it just flush the appended records and stops with the appending 

https://github.com/camunda-cloud/zeebe/blob/develop/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java#L548-L552

But not error or anything is returned to the leader, which means it has to timeout on the leader first and then the append is resend. This produces unnecessary delay.

**To Reproduce**

Cause a failure on append on the follower side.

**Expected behavior**

The follower sends an error back and the leader reacts appropriate with that.

**Environment:**
 - Zeebe Version: 1.3.0, 1.2.x, 1.1.x

entangled90:
## Description
Adds a test case to check that a negative response is sent back to the leader when append fails

## Related issues

closes #8612

### DIFF
--- a/zeebe/atomix/cluster/src/test/java/io/atomix/raft/roles/PassiveRoleTest.java
+++ b/zeebe/atomix/cluster/src/test/java/io/atomix/raft/roles/PassiveRoleTest.java
@@ -237,4 +237,18 @@ public class PassiveRoleTest {
     // then
     assertThat(response.succeeded()).isTrue();
   }
+
+  @Test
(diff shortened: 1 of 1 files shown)


deepthidevaki (approved):
:rocket:

Issue #19222, opened 2024-06-10
CI: operate and tasklist CI not working on PRs from forks


cmur2:
## Description

The GHA workflows from Operate & Tasklist project were developed in private repositories and never designed to be used on Pull Requests from forks by external contributors which can now happen in `camunda/camunda`. This manifests e.g. now in https://github.com/camunda/camunda/pull/18885 by instructions like https://github.com/camunda/camunda/blob/95e929f1720cb99cb0199e5cd9c41a4d6f695407/.github/workflows/tasklist-ci-build-reusable.yml#L47-L49 and also no access to secrets.

The Zeebe CI GHA workflows deal with that gracefully.

### Impact

Since Operate & Tasklist GHA workflows are part of the merge queue they represent GitHub required status checks that cannot be circumvented, so **since April 2024 no PRs from external contributors can be merged**.

#### Workaround

If a Camundi wants to merge the changes from an external contributor, the Camundi can:

1. Explain the current technical limitation to the external contributor and refer to _this_ ticket for context
2. Make sure the external contributor understands this that they will be properly credited _(Camunda does not want to steal any code, this is just for technical reasons and will be done on behalf of the contributor)_
3. Create a new PR in `camunda/camunda` explicitly crediting the external contributor, making sure to follow the [Contribution Guidelines](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md) and use the code from the original PR and backlink to it
4. Check CI status of the new PR and merge if reviewed & green

#### Possible solutions

* adjust Operate & Tasklist CI GHA workflows (minimum: those marked as GitHub required status checks) to be able to run from forks
* wait for https://github.com/camunda/camunda/issues/17721

korthout:
For bookkeeping, this contribution is blocked by this:
- https://github.com/camunda/camunda/pull/17358#issuecomment-2157559337

cmur2:
The 2nd contribution in June 2024 that I'm aware of and is blocked by this:

* https://github.com/camunda/camunda/pull/18885

lenaschoenburg:
And another one: https://github.com/camunda/camunda/pull/19286

We don't want to discourage external contributions so I'd suggest to prioritize this if possible.

saig0:
And another contribution: https://github.com/camunda/camunda/pull/16846

korthout:
I followed the suggestion, but there are still two CI jobs failing:
- https://github.com/camunda/camunda/pull/20075

[...]

PR #23785, opened 2024-10-21
ci: introduce a new composite action setup-build

clementnero:
## Description

Related to https://github.com/camunda/team-infrastructure/issues/675.

Replace `build-operate-setup`,  `build-tasklist-setup`,  `setup-maven` and `setup-zeebe` with a new generic `setup-build` action, integrating the same features:
- Install Java (version and distribution)
- Eventually set Maven version to be used by the wrapper (`mvnw`)
- Configure Camunda Nexus and extra Maven mirrors and servers
- Set Maven cache
- Set a Time Zone (Linux only)
- Logs on to Dockerhub

#### Fork PRs

If workflows are run for a PR Fork, the DockerHub login and Camunda Nexus configuration are ignored, as the GitHub secrets needed to retrieve Vault secrets are not accessible. The equivalent logic in `setup-zeebe` is therefore preserved and is now used by default for all other workflows.

#### Camunda Nexus and DockerHub accounts

A common (and unique) machine account is used for Nexus.

For DockerHub, the team dedicated account is used if the caller workflow is prefixed with the team name (e.g. `optimize-*.(yml|yaml)`), otherwise the `camundacicamunda` account dedicated to monorepo is used. Permissions to the `camunda8` DockerHub team have been updated accordingly.

### After merging
- [ ] check that snapshot Docker images are pushed as expected when running on `main`
- [ ] remove this list

## Checklist

- [x] for CI changes:

## Related issues

closes #


.github/actions/setup-zeebe/action.yml
```diff
@@ -1,113 +0,0 @@
...
-        username: ${{ steps.secrets.outputs.REGISTRY_HUB_DOCKER_COM_USR }}
-        password: ${{ steps.secrets.outputs[inputs.docker-token] }}
-    # Use CI Nexus as co-located pull-through cache for Maven artifacts via ~/.m2/settings.xml
-    - name: 'Create settings.xml'
-      uses: s4u/maven-settings-action@v3.0.0
-      if: |
-        inputs.java == 'true'
-        && inputs.secret_vault_address != ''
-        && inputs.secret_vault_roleId != ''
-        && inputs.secret_vault_secretId != ''
```
cmur2, line 99: I think this and similar logic for other steps of the old `setup-zeebe` action got lost in the refactoring in this PR. This logic allows running CI on repositories forked from camunda/camunda that don't have access to Vault/secrets by falling back to Maven Central.

This feature is important for external contributors, see also https://github.com/camunda/camunda/issues/19222

This is also the reason why `setup-zeebe` handles secrets retrieval from Vault itself to be able to conditionally skip that and other related steps requiring secrets when detecting that run on a fork. Can you find a way to add this logic back to `setup-build` (implicitly also adding it to Operate/Tasklist etc)?

clementnero: Good catch!!
I can skip `camunda-nexus`config when running on forks, but I will probably have to introduce a new `setup-secrets` action to not break this logic. `setup-secrets` should not fail on forks and print a message that secrets cannot be retrieved. Not sure how to deal with expected (empty) outputs.

clementnero: As discussed, I have updated `steup-build` with same logic for fork PRs and will be used by default for all workflows. Team-dedicated accounts are used to preserve a similar quota distribution, but implemented to pave the way to a smooth transition to a unique account.

clementnero: For a reusable workflow, `github.workflows_ref` returns the name of the caller workflow. `camundacicamunda` thus needs read/write access to the `camunda/zeebe` repository as `zeebe-ci.yml` is called from `ci.yml` (zeebe DockerHub account will not be used in this case).

`preview-env-build-and-deploy.yml` results in deploying images to Harbor not DockerHub.

cmur2: > zeebe DockerHub account will not be used in this case
Makes sense to me 👍🏼

cmur2:
@clementnero there are some commits from Thorben on this branch which shouldn't be here, right?

clementnero:
> @clementnero there are some commits from Thorben on this branch which shouldn't be here, right?

Strange 🤔 I don't know what happened, just rebased now.

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,233 @@
...
+      if [[ "${REPO_OWNED_BY_CAMUNDA}" != "true" ]]; then
+        echo "This repository does not belong to Camunda, and is probably an external fork."
+        echo "As a result, Vault secrets cannot be recovered, Camunda Nexus and DockerHub login will be disabled."
+      fi
+
+      # Check if DockerHub and DockerHub-readonly are both enabled
+      if [[ "${DOCKERHUB}" == "true" && "${DOCKERHUB_READONLY}" == "true" ]]; then
+        echo "Both dockerhub and dockerhub-readonly inputs are enabled. Only one can be enabled at a time."
+        exit 1
+      fi
```
cmur2, line 111: ❓ Both variables could be combined into one 3-modal variable (`Dockerhub: yes/readonly/no`) to simplify/avoid this check without other downsides afaik.

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,233 @@
...
+            "path": "secret/data/products/optimize/ci/camunda-optimize"
+          },
+          "tasklist": {
+            "path": "secret/data/products/tasklist/ci/tasklist"
+          },
+          "zeebe": {
+            "path": "secret/data/products/zeebe/ci/zeebe"
+          }
+        }
+      REPO_OWNED_BY_CAMUNDA: ${{ github.repository_owner == 'camunda' }}
```
cmur2, line 96: ❓ This expression is very lean and elegant way to check. Is it available in all contexts (on fork PRs etc)? The setup-zeebe action used a different expression for this kind of check (`github.event.name == 'pull_request' && github.event.pull_request.head.repo.full_name == 'camunda/camunda'`) that reads more (over-?)complicated.

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,233 @@
...
+  - name: Setup Maven
+    if: inputs.maven-version != ''
+    uses: ./.github/actions/setup-maven-dist
+    with:
+      maven-version: ${{ inputs.maven-version }}
+      set-mvnw: true
+  # - name: Configure Maven cache
+  #   uses: ./.github/actions/setup-maven-cache
+  #   with:
+  #     maven-cache-key-modifier: ${{ inputs.maven-cache-key-modifier }}
```
cmur2, line 155: ❓ Is this commented intentionally?

.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,233 @@
...
+      TZ_IDENTIFIER: ${{ inputs.time-zone }}
+    run: |
+      # tzdata is not installed in self-hosted runners
+      sudo apt-get update && sudo apt-get install -y tzdata
+      echo ${TZ_IDENTIFIER} | sudo tee /etc/timezone 
+      sudo rm -rf /etc/localtime
+      sudo ln -s /usr/share/zoneinfo/${TZ_IDENTIFIER} /etc/localtime
+    shell: bash
+  - name: Login to Docker Hub
+    # Camunda Nexus is enabled only if repo is owned by Camunda (not for external forks)
```
cmur2, line 223: ❓ "Camunda Nexus" should be "DockerHub" I guess

.github/workflows/tasklist-ci-build-reusable.yml
```diff
@@ -77,16 +77,19 @@ jobs:
...
 
       #########################################################################
       # Setup and configure Maven, Docker, and settings
-      - uses: ./.github/actions/build-tasklist-setup
+      - uses: ./.github/actions/setup-build
         name: Build setup
         with:
-          nexusUsername: ${{ steps.secrets.outputs.NEXUS_USR }}
-          nexusPassword: ${{ steps.secrets.outputs.NEXUS_PSW }}
+          dockerhub: true
```
cmur2, line 86: Only needed on default branch, right?

cmur2 (commented):
Looking good so far! I left some questions on some details

cmur2:
There is a Git conflict because of a recent change by Optimize team to one of their CI jobs

.github/actions/setup-build/action.yml
cmur2: Nvm, addressed in follow-up PR https://github.com/camunda/camunda/pull/24409/commits/a7435821ad4261309351823d958ad09793398015?diff=unified&w=0

clementnero: Yes indeed :)

.github/actions/setup-build/action.yml
clementnero: Introduced by  a temporary [commit](https://github.com/camunda/camunda/pull/23785/commits/49f75af51f03abc379840b9aa68093b293e2e967) to test with cache disable, and will be removed.

.github/actions/setup-build/action.yml
clementnero: I was inspired by https://github.com/orgs/community/discussions/26409, i did not tested it on a real fork but I guess it should work for all type contexts.

The original check from `setup-zeebe` was to fail if an internal PR does not define GitHub secrets for Vault. I think it is not necessary (but still provide nice error log) because it will fail anyway on the Vault step ... .
```yaml
    - name: Secrets check
      if: |
        github.event.name == 'pull_request'
        && github.event.pull_request.head.repo.full_name == 'camunda/camunda'
        && (inputs.secret_vault_address == ''
        || inputs.secret_vault_roleId == ''
        || inputs.secret_vault_secretId == '')
      shell: bash
      run: |
        echo "We will fail the action if it's an internal PR without the secrets. If this occurs it indicates a configuration failure that needs to be fixed."
        exit 1
```

I will check different contexts to validate the approach.

.github/actions/setup-build/action.yml
clementnero: Good question.
On subsequent PR the `harbor` input of boolean type is introduced.  I am wondering it is not a bit confusing to have  two different type of variable ? e.g.:

```yaml
dockerhub: readonly
harbor: true
```

What do you htink ?

.github/workflows/tasklist-ci-build-reusable.yml
clementnero: Good catch!

clementnero:
> There is a Git conflict because of a recent change by Optimize team to one of their CI jobs

And `rdbms-integration-test.yml` has been introduced in the meantime ... .

.github/actions/setup-build/action.yml
cmur2: I'm also fine with the current solution

.github/actions/setup-build/action.yml
cmur2: Sounds good to me 👍🏼

cmur2 (approved):
Code changes look good now

.github/actions/setup-build/action.yml
clementnero: It does not work as I expected 😢 

I created a test fork, see this workflow run and job https://github.com/camunda/infra-global-github-actions/actions/runs/11727212631/job/32667727156

image

I need to rely on  `github.event.pull_request.head.repo.full_name` to detect workflow runs for fork PRs.
Really good catch!!

clementnero: I switch from a repo-owned-by-camunda check to a is-fork check which is more clear btw.

cmur2 (approved):
Thanks

cmur2:
There is another merge conflict created in the meantime Oo

clementnero:
Workflow run on main branch https://github.com/camunda/camunda/actions/runs/11742137770 worked as expected, publishing snapshot docker images.

PR #23788, opened 2024-10-21
ci: run operate update screenshots workflow on PRs

vitorwtavares:
## Description

- Runs `Operate update screenshots` workflow on pull requests to keep screenshots updated alongside E2E tests.

## Checklist

- [ ] for CI changes:

### DIFF
--- a/.github/workflows/operate-update-docs-screenshots.yml
+++ b/.github/workflows/operate-update-docs-screenshots.yml
@@ -1,5 +1,22 @@
 name: Operate Update screenshots
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+    paths:
+      - ".github/workflows/operate-update-docs-screenshots.yml"
+      - "operate/client/**"
+  pull_request:
+    paths:
+      - ".github/workflows/operate-update-docs-screenshots.yml"
+      - "operate/client/**"
+
+# Limit workflow to 1 concurrent run per ref (branch): new commit -> old runs are canceled to save costs
+# Exception for main branch: complete builds for every commit needed for confidenence
+concurrency:
+  cancel-in-progress: true
+  group: ${{ format('{0}-{1}', github.workflow, github.ref == 'refs/heads/main' && github.sha || github.ref) }}
 
 jobs:
   test:


.github/workflows/operate-update-docs-screenshots.yml
```diff
@@ -1,5 +1,38 @@
 name: Operate Update screenshots
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+      - "stable/**"
+      - "release/**"
```
pedesen, line 8: I think it's enough that the workflow only runs on `main` branch, because we would never use screenshots from stable or release branches.

.github/workflows/operate-update-docs-screenshots.yml
```diff
@@ -1,5 +1,38 @@
...
+    paths:
+      - ".ci/**"
+      - ".github/actions/**"
+      - ".github/workflows/operate-*"
+      - "bom/*"
+      - "distro/**"
+      - "operate.Dockerfile"
+      - "operate/**"
+      - "parent/*"
+      - "pom.xml"
```
pedesen, line 18: The Operate backend is mocked in this workflow ([example](https://github.com/camunda/camunda/blob/main/operate/client/e2e-playwright/docs-screenshots/process-instance-modification.spec.ts#L28-L40)), which makes it independent from changes in zeebe or Operate backend. I think it should only be triggered when there are frontend changes (or to the yaml file itself), same as for [frontend tests](https://github.com/camunda/camunda/blob/main/.github/workflows/operate-frontend.yml). 

Same for the pull_request trigger below.

pedesen (changes requested):

.github/workflows/operate-update-docs-screenshots.yml
vitorwtavares: Makes sense, I didn't realize the BE was mocked and was thinking like the visual regression tests where a "external" change (BE, changes in carbon) would result in changes and it would be good for us to catch it. I've fixed it to match the FE tests. Good catch!

.github/workflows/operate-update-docs-screenshots.yml
```diff
@@ -1,5 +1,22 @@
 name: Operate Update screenshots
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "main"
+    paths:
+      - ".github/workflows/operate-frontend.yml"
```
pedesen, line 8:
```suggestion
      - ".github/workflows/operate-update-docs-screenshots.yml"
"
```

.github/workflows/operate-update-docs-screenshots.yml
```diff
@@ -1,5 +1,22 @@
...
+      - "operate/client/**"
+  pull_request:
+    paths:
+      - ".github/workflows/operate-frontend.yml"
```
pedesen, line 12:
```suggestion
      - ".github/workflows/operate-update-docs-screenshots.yml"
```

pedesen (changes requested):

pedesen (approved):

Issue #23483, opened 2024-10-14
Create single user selection and use it for selecting process owner


OmranAbazid:
Engineering DRI: @misiekhardcore

### Context

We currently use the multi-user selection for selecting a single owner. We prevent the owner from selecting additional owners by closing the selection after selecting an owner. This is confusing to the customer, and instead we should implement a single user selection input.

### Acceptance Criteria

- Implement a single user selection input
- Use the new single user selection input for the process owner selection.

### Testing Notes

1. go to Dashboards page
2. on the processes list click the settings icon to open process owner modal
3. the user selector should allow to select one user only

![Screenshot from 2024-10-24 16-06-07](https://github.com/user-attachments/assets/cd170806-4bca-44bf-a53c-91db5f4863a8)

![Screenshot from 2024-10-24 16-06-18](https://github.com/user-attachments/assets/198cf18c-8e89-4d0d-8f97-7e7dfd3e5e17)

### Links

related to https://jira.camunda.com/browse/SUPPORT-23995

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/23827
- [ ] https://github.com/camunda/camunda-optimize/pull/14645
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [x] All code targeting the main branch has been reviewed by at least one Engineer
- [x] The PR targeting the main branch has been approved by the reviewing Engineer
- [x] If the API has changed, the API documentation is updated
- [x] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [ ] The change is implemented as described on all target environments/versions/modes
- [ ] The documentation changes are as expected

### Completion

- [ ] All Review stages are successfully completed
- [ ] All associated PRs are merged to the main branch(es) and stable branches
- [ ] The correct version labels are applied to the issue

OmranAbazid:
@misiekhardcore from looking at this i imagine that we could build a SingleUserInput component that uses the carbon combo box component.
Feel free to do something else if you see a better solution though.

misiekhardcore:
@OmranAbazid should this be done in both C7 and C8? should it be backported?

OmranAbazid:
@misiekhardcore can you take a look at the support ticket and figure out the version we need to have this.
Since this is a support issue, it is important that customers do not face the issue again.
Feel free to fix it where you see is necessary.

[...]

PR #23827, opened 2024-10-22
feat: add single user selection option to user typeahead

misiekhardcore:
## Checklist

- [ ] for CI changes:

## Related issues

related to https://github.com/camunda/camunda/issues/23483


misiekhardcore:
@OmranAbazid I am not opening PR for C7 repo yet until we agree that the solution here is good

OmranAbazid (commented):
I like the change. Everything is easier to read now with the refactoring.

One thing i noticed in the UI is that we favor the email over the user name:

Screenshot 2024-10-24 at 11 59 08 AM

This looks a bit weird imo as the name is more important that the email.
If possible, we should display both like we do in the multi user selection:

Screenshot 2024-10-24 at 12 16 17 PM

misiekhardcore:
> One thing i noticed in the UI is that we favor the email over the user name:
>
> This looks a bit weird imo as the name is more important that the email. If possible, we should display both like we do in the multi user selection:

These both were there but the styling was broken. I will fix that

OmranAbazid (approved):

Issue #23215, opened 2024-10-09
`ScaleUp` command is written by new cluster change operation


lenaschoenburg:
Add a new cluster change operation that uses the broker client to write the newly introduced `ScaleUp` command to partition 1.

PR #23837, opened 2024-10-22
feat: change applier for starting partition scale up

lenaschoenburg:
This adds a new applier, `StartPartitionScaleUpApplier` that is supposed to initiate the scale up within the engine.
It does this by updating the routing state, incrementing version and marking new partitions as initially inactive, and then calling out to a new, not yet implemented, `PartitionScalingChangeExecutor`.

Depends on https://github.com/camunda/camunda/pull/23777
Relates to https://github.com/camunda/camunda/issues/23215

### DIFF
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplier.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplier.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.dynamic.config.changes;
+
+import io.camunda.zeebe.dynamic.config.changes.ConfigurationChangeAppliers.ClusterOperationApplier;
+import io.camunda.zeebe.dynamic.config.state.ClusterConfiguration;
+import io.camunda.zeebe.dynamic.config.state.RoutingState;
+import io.camunda.zeebe.dynamic.config.state.RoutingState.RequestHandling;
+import io.camunda.zeebe.dynamic.config.state.RoutingState.RequestHandling.ActivePartitions;
+import io.camunda.zeebe.dynamic.config.state.RoutingState.RequestHandling.AllPartitions;
+import io.camunda.zeebe.protocol.Protocol;
+import io.camunda.zeebe.scheduler.future.ActorFuture;
+import io.camunda.zeebe.scheduler.future.CompletableActorFuture;
+import io.camunda.zeebe.util.Either;
+import io.camunda.zeebe.util.Either.Left;
+import io.camunda.zeebe.util.Either.Right;
+import java.util.Set;
+import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Starts the process of scaling up partitions. The routing state version is incremented and the
+ * request handling strategy is updated from {@link AllPartitions} to {@link ActivePartitions} one,
+ * with new partitions starting off as inactive.
+ *
+ * <p>We use the request handling strategy to detect whether scaling up is possible or not - we
+ * expect {@link AllPartitions} at the beginning, update it to {@link ActivePartitions} and finish
+ * the scaling by going back to {@link AllPartitions}. If we find anything but {@link
+ * AllPartitions}, we assume that we are already scaling up and refuse the operation.
+ */
+final class StartPartitionScaleUpApplier implements ClusterOperationApplier {
+
+  private final PartitionScalingChangeExecutor partitionScalingChangeExecutor;
+  private final int desiredPartitionCount;
+
+  StartPartitionScaleUpApplier(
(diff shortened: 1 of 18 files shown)


npepinpe (commented):
:question: I have to admit I'm not following what this is supposed to do. So the applier checks that the current cluster configuration already contains the partitions we want to scale up to. Who is updating this configuration, where is the increase in partitions actually happening? Because with the current applier, if I have 3 partitions and want 4, it will reject it unless somehow the current topology already contains 4 (?).

So what would be the expected sequence of operations? Something like:

- `PartitionBootstrapOperation` (bootstraps the new partition in the primary)
- `PartitionJoinOperation` (one for each replica)
- `PartitionJoinOperation`
- `StartScaleUpOperation` (to redistribute resources)

Something like that? Would we have one operation per scaled partition? I'm sorry if this was all documented somewhere else and I missed it. Maybe we should do a quick sync for onboarding :see_no_evil:

lenaschoenburg:
Exactly, the `StartScaleUpOperation` comes after these partitions are already bootstrapped and all replicas have joined. It is only used to start the scale up within the engine.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplier.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    if (desiredPartitionCount > Protocol.MAXIMUM_PARTITIONS) {
+      return new Left<>(
+          new IllegalArgumentException(
+              "Desired partition count must not exceed %d".formatted(Protocol.MAXIMUM_PARTITIONS)));
+    }
+
+    for (int partition = Protocol.START_PARTITION_ID;
+        partition <= desiredPartitionCount;
```
npepinpe, line 67: :upside_down_face: Super minor, but this condition relies on the fact `START_PARTITION_ID = 1`. If it's not (e.g. 2), then this fails. So could be `partition < desiredPartitionCount + Protocol.START_PARTITION_ID` instead. Super nit picky, sorry, since when would we ever change the start partition ID? (never say never :smile:)

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplierTest.java
```diff
@@ -0,0 +1,223 @@
...
+        .left()
+        .asInstanceOf(InstanceOfAssertFactories.throwable(IllegalArgumentException.class))
+        .hasMessageContaining("Desired partition count must be greater than 1");
+
+    EitherAssert.assertThat(
+            new StartPartitionScaleUpApplier(executor, 100000).init(initialConfiguration))
+        .left()
+        .asInstanceOf(InstanceOfAssertFactories.throwable(IllegalArgumentException.class))
+        .hasMessageContaining("Desired partition count must not exceed 8192");
+  }
```
npepinpe, line 66: :wrench: Would split this in two tests :shrug: There's no shared set up or anything anyway.

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplierTest.java
npepinpe, line 1: :thought_balloon: This is a new pattern where you assert the preconditions beforehand, before asserting the actual behavior. Did this come from some experience, or just a new thing you're trying?

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplierTest.java
```diff
@@ -0,0 +1,223 @@
...
+
+    // then - init fails with a useful exception
+    EitherAssert.assertThat(result)
+        .left()
+        .asInstanceOf(InstanceOfAssertFactories.throwable(IllegalStateException.class))
+        .hasMessageContaining("Partition 4 is not known");
+  }
+
+  @Test
+  void shouldFailWhenRequestHandlingIsUnexpected() {
```
npepinpe, line 99: :wrench: For completeness, we should also test the third case (`Mixed`). Can be a parameterized test (though unlike enums, I don't think we can ensure exhaustive use of sealed interface types here)

npepinpe (approved):
:rocket: Contingent on approval of #23777 of course

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplier.java
lenaschoenburg: Good catch. I can check `partition < Protocol.START_PARTITION_ID + desiredPartitionCount` instead :+1:

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplierTest.java
lenaschoenburg: Just something I was trying out. I generally dislike shared test setup because each test may have subtle assumptions that could break when making changes to it. So I wanted to try and codify these assumptions.

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/changes/StartPartitionScaleUpApplierTest.java
lenaschoenburg: Mixed is removed now :tada:

PR #23860, opened 2024-10-22
fix: compare with base branch or previous commit

npepinpe:
## Description

This PR updates the Go compat check by comparing using the default from the action: the base branch (for PRs), or the target commit for merge queues. This reuses the same logic to pick the base SHA as we use to check for Protobuf compatibility.

Without this fix, when we delete old branches, nothing can be merged on the next minor branch.


megglos (approved):
Thanks for taking action on this.

npepinpe:
timing out building the Docker image on ARM64 :sob:

PR #23865, opened 2024-10-22
Full qualified name fix

EuroLew:
## Description

The full qualified name(FQN)  now maintains full backwards compatibility, in that operate FQN stay the same and camunda exporter FQN now always use the format `<prefix>-<component-name>-<index name>-<version>`

### DIFF
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
@@ -19,9 +19,6 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
   protected String indexPrefix;
   protected boolean isElasticsearch;
 
-  // Will not use global prefix unless this value is true
-  protected boolean hasGlobalPrefix = false;
-
   public AbstractIndexDescriptor(final String indexPrefix, final boolean isElasticsearch) {
     this.indexPrefix = indexPrefix;
     this.isElasticsearch = isElasticsearch;
@@ -29,12 +26,8 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   @Override
   public String getFullQualifiedName() {
-    if (hasGlobalPrefix) {
-      return String.format(
-          "%s%s-%s-%s_", formattedIndexPrefix(), getComponentName(), getIndexName(), getVersion());
-    } else {
-      return String.format("%s%s-%s_", formattedIndexPrefix(), getIndexName(), getVersion());
-    }
+    return String.format(
+        "%s%s-%s-%s_", formattedIndexPrefix(), getComponentName(), getIndexName(), getVersion());
   }
 
   @Override
@@ -51,12 +44,8 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   @Override
   public String getAllVersionsIndexNameRegexPattern() {
-    if (hasGlobalPrefix) {
-      return String.format(
-          "%s%s-%s-\\d.*", formattedIndexPrefix(), getComponentName(), getIndexName());
-    } else {
-      return String.format("%s%s-\\d.*", formattedIndexPrefix(), getIndexName());
-    }
+    return String.format(
+        "%s%s-%s-\\d.*", formattedIndexPrefix(), getComponentName(), getIndexName());
   }
 
   @Override
@@ -70,11 +59,6 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
 
   public String getIndexPrefix() {
     return indexPrefix;
-  }
-
-  public AbstractIndexDescriptor withHasGlobalPrefix(final boolean isGlobalPrefix) {
-    hasGlobalPrefix = isGlobalPrefix;
-    return this;
   }
 
   public abstract String getComponentName();
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
@@ -33,7 +33,9 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
       return String.format(
           "%s%s-%s-%s_", formattedIndexPrefix(), getComponentName(), getIndexName(), getVersion());
     } else {
-      return String.format("%s-%s-%s_", getIndexPrefix(), getIndexName(), getVersion());
+      return getIndexPrefix() != null && !getIndexPrefix().isBlank()
+          ? String.format("%s-%s-%s_", getIndexPrefix(), getIndexName(), getVersion())
+          : String.format("%s-%s_", getIndexName(), getVersion());
     }
   }
 
@@ -55,7 +57,9 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
       return String.format(
           "%s%s-%s-\\d.*", formattedIndexPrefix(), getComponentName(), getIndexName());
     } else {
-      return String.format("%s-%s-\\d.*", getIndexPrefix(), getIndexName());
+      return getIndexPrefix() != null && !getIndexPrefix().isBlank()
+          ? String.format("%s-%s-\\d.*", getIndexPrefix(), getIndexName())
+          : String.format("%s-\\d.*", getIndexName());
     }
   }
 
@@ -65,7 +69,7 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
   }
 
   private String formattedIndexPrefix() {
-    return getIndexPrefix().isBlank() ? "" : getIndexPrefix() + "-";
+    return getIndexPrefix() != null && !getIndexPrefix().isBlank() ? getIndexPrefix() + "-" : "";
   }
 
   public String getIndexPrefix() {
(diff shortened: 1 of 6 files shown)


webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
```diff
@@ -19,14 +19,22 @@ public abstract class AbstractIndexDescriptor implements IndexDescriptor {
   protected String indexPrefix;
   protected boolean isElasticsearch;
 
+  // Will not use global prefix unless this value is true
+  protected boolean isGlobalPrefix = false;
```
mustafadagher, line 23: ❓ why protected?

mustafadagher, line 23: 🔧 `hasGlobalPrefix`?

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
```diff
@@ -43,26 +51,29 @@ public String getMappingsClasspathFilename() {
...
 
-  private String getPrefixAndComponentName() {
-    // Cannot start index with "-" so must not append "-" for empty prefix
-    final var prefix = getIndexPrefix().isBlank() ? getIndexPrefix() : getIndexPrefix() + "-";
-
-    // Legacy descriptors have the same index prefix as component name this avoids duplication.
-    return getIndexPrefix().contains(getComponentName())
-        ? getIndexPrefix()
-        : prefix + getComponentName();
+  public void setIsGlobalPrefix(final boolean isGlobalPrefix) {
```
mustafadagher, line 75: 🔧 💭 A builder-style `withHasGlobalPrefix()` that returns an instance (i.e. `return `this`) could help getting rid of the list, then loop then map in the ResourceProvider class

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
```diff
@@ -78,45 +83,49 @@ public void init(final ExporterConfiguration configuration) {
...
+            entry(
+                SequenceFlowTemplate.class,
+                new SequenceFlowTemplate(globalPrefix, isElasticsearch)),
+            entry(
+                DecisionInstanceTemplate.class,
+                new DecisionInstanceTemplate(globalPrefix, isElasticsearch)),
+            entry(EventTemplate.class, new EventTemplate(globalPrefix, isElasticsearch)),
+            entry(TaskTemplate.class, new TaskTemplate(globalPrefix, isElasticsearch)));
+
+    templateDescriptors.forEach(entry -> entry.getValue().setIsGlobalPrefix(true));
```
mustafadagher, line 106: 🔧 the suggestion above about the builder-style with can help get rid of this iteration, and then the collection to a map later and make it more explicit. (if you don't want to include the field on all constructors)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
```diff
@@ -78,45 +83,49 @@ public void init(final ExporterConfiguration configuration) {
...
+                new DecisionRequirementsIndex(globalPrefix, isElasticsearch)),
+            entry(MetricIndex.class, new MetricIndex(globalPrefix, isElasticsearch)),
+            entry(ProcessIndex.class, new ProcessIndex(globalPrefix, isElasticsearch)),
+            entry(FormIndex.class, new FormIndex(globalPrefix, isElasticsearch)),
+            entry(
+                TasklistMetricIndex.class, new TasklistMetricIndex(globalPrefix, isElasticsearch)),
+            entry(UserIndex.class, new UserIndex(globalPrefix, isElasticsearch)),
+            entry(AuthorizationIndex.class, new AuthorizationIndex(globalPrefix, isElasticsearch)));
+
+    indexDescriptors.forEach(entry -> entry.getValue().setIsGlobalPrefix(true));
```
mustafadagher, line 125: same.

Zelldon:
I think what would be benefical to add some unit tests about the prefix logic. 

Still, i dont think we need to make it so complex. It should just set the prefix if some is given otherwise component name and index name with version.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
EuroLew: I thought about this there are two solutions and I think this is probably the cleanest. If I want the `withHasGlobalPrefix` to be in the `AbstractIndexDescriptor` when I `return this` it will return an instance of `AbstractIndexDescriptor` instead of the implementation type. 

Possible solutions
- Make `AbstractIndexDescriptor` a generic class such that each descriptor class is now `Descriptor extends AbstractIndexDescriptor<Descriptor>` not that clean and requires a lot of change.
- Current approach, not ideal as the descriptor is now not immutable.

If there's a better way to implement the `with` syntax other than generics I'd be open to it.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
```diff
@@ -43,26 +51,30 @@ public String getMappingsClasspathFilename() {
...
-    // Cannot start index with "-" so must not append "-" for empty prefix
-    final var prefix = getIndexPrefix().isBlank() ? getIndexPrefix() : getIndexPrefix() + "-";
-
-    // Legacy descriptors have the same index prefix as component name this avoids duplication.
-    return getIndexPrefix().contains(getComponentName())
-        ? getIndexPrefix()
-        : prefix + getComponentName();
+  public AbstractIndexDescriptor withHasGlobalPrefix(final boolean isGlobalPrefix) {
+    hasGlobalPrefix = isGlobalPrefix;
+    return this;
```
Zelldon, line 77: ❓ Couldn't this be derived from `getIndexPrefix() != null && !getIndexPrefix().isBlank()`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
```diff
@@ -81,42 +81,42 @@ public void init(final ExporterConfiguration configuration) {
...
+            new SequenceFlowTemplate(globalPrefix, isElasticsearch).withHasGlobalPrefix(true),
             DecisionInstanceTemplate.class,
-            new DecisionInstanceTemplate(globalPrefix, isElasticsearch),
+            new DecisionInstanceTemplate(globalPrefix, isElasticsearch).withHasGlobalPrefix(true),
             EventTemplate.class,
-            new EventTemplate(globalPrefix, isElasticsearch),
+            new EventTemplate(globalPrefix, isElasticsearch).withHasGlobalPrefix(true),
             TaskTemplate.class,
-            new TaskTemplate(globalPrefix, isElasticsearch));
+            new TaskTemplate(globalPrefix, isElasticsearch).withHasGlobalPrefix(true));
```
Zelldon, line 100: 🙃 Feels duplicated, as we specify the prefix in the constructor 🤔

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java
EuroLew: I think we decided to always have the global prefix + component name even if they repeat stuff, I think this will cause issues in usages in the `operate` module as the FQN will change. 

The answer is it cannot. if it's not clear we can talk more.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
EuroLew: there's no way to know (to my knowledge) based on the value of the prefix whether we are using a global prefix or a local prefix (like in the operate module) I think this is the most succinct solution without adding a new constructor to dozens of classes.

Zelldon: Lets sync

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/DefaultExporterResourceProviderTest.java
```diff
@@ -28,52 +34,57 @@ void shouldHaveCorrectDetailsForDefaultEmptyPrefix() {
...
   }
 
-  @Test
-  void shouldUseGivenIndexPrefixInRegexAndNameDetails() {
-    final var config = new ExporterConfiguration();
-    config.getIndex().setPrefix("global");
-    final var provider = new DefaultExporterResourceProvider();
+  static Stream<ExporterConfiguration> configProvider() {
+    final var configNoPrefix = new ExporterConfiguration();
+    final var configWithPrefix = new ExporterConfiguration();
```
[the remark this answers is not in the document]
Zelldon: This was actually the problem we wanted to fix :)

ChrisKujawa (approved):
Thanks Euro 🙇🏼

PR #23912, opened 2024-10-23
feat: added submitCallable to ActorControl: tasks can be scheduled from outside actor

entangled90:
## Description
Added `submitCallable` method to `ActorControl`. While adding a simple test I noticed that both `submit` and `submitCallable` are not in `ConcurrencyControl`.
I  temporarily added them as it made testing it easier.

The implementation is mostly taken from the method `submit`, adjusting it to fill the `ActorJob` with a Callable instead.

### DIFF
--- a/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/Actor.java
+++ b/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/Actor.java
@@ -147,6 +147,16 @@ public abstract class Actor implements AutoCloseable, AsyncClosable, Concurrency
     return actor.schedule(delay, runnable);
   }
 
+  @Override
+  public void submit(final Runnable action) {
+    actor.submit(action);
+  }
+
+  @Override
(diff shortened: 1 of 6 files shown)


npepinpe:
Shall we close this? Or do you still see some value?

entangled90:
I did not need it for the PR I originally created, but there are some places in the codebase where we could use it potentially.
Basically everywhere we write:
```java
final var future = new CompletableFuture<A>();
actor.submit(() -> {
 var taskResult = task();
future.complete(taskResult);
});
```

I think that the above pattern does not complete the future with an error in case there is an exception in the runnable or in `actor.submit` future (less likely). So it's potentially more error prone than what proposed in the PR.

npepinpe:
You're right, but in those cases you can use `call`, right? I still don't see the big different between call and `submitCallable` :thinking:

entangled90:
Yes, sorry I got confused again that `call` could only be used from *inside* the actor.

PR #23846, opened 2024-10-22
feat: introduce form related fields in process index

panagiotisgts:
## Description

Introduce the `formId` and the `isPublic` fields over the `operate-process` index along with the respective chagnes to the ProcessHandler

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes #23043

PR #23923, opened 2024-10-23
feat: get process form - controller

### LINKED ISSUE #22733: Implement `Get Form for Process Definition`
### Description
Implement the `GET  /v2/process-definitions/{processDefinitionKey}/form` endpoint for retrieving a form by process definition key. This endpoint should return detailed form information including tenant ID, BPMN ID, schema, version, and more.

It is important to note that in contrast to User Tasks, for Processes it should retrieve the highest version of the deployed form, which means `bpmnId` from the form should be added to `process-definitions`.

---

### Endpoint
`GET /v2/process-definitions/{processDefinitionKey}/form`

#### Request
- **Path Parameter:**
  - `processDefinitionKey` (Long): The unique key for the process where the form that should be retrieved.

#### Response
- **Success (200 OK):**
  - Returns a JSON object representing the form with the following fields:
    - `formKey`: The unique key of the form.
    - `tenantId`: The tenant ID associated with the form.
    - `bpmnId`: The BPMN ID of the form.
    - `schema`: The schema representing the form.
    - `version`: The version of the form.

  **Example Response:**
  ```json
  {
    "formKey": 1,
    "tenantId": "tenant-1",
    "bpmnId": "bpmn-1",
    "schema": {},
    "version": 1
  }

Tasks:
- Implement the Service
- Implement the Rest endpoint
- Implement the Client
- Implement the integration Tests

nathansandi:
## Description

- Adjust Filter capabilities on Forms to filter formId, and sort by Version
- Implement get latest version by formId on formServices
- Include Controller Get on Process-definitions to get the form
- Unit tests 

PR to include formId on Operate-process index:
https://github.com/camunda/camunda/pull/23846

- This is only possible to test with exporters.

## Related issues

https://github.com/camunda/camunda/issues/22733


zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/ProcessDefinitionQueryControllerTest.java
```diff
@@ -206,4 +223,75 @@ public void shouldGetProcessDefinitionXmlHasNoXml() {
...
+        .json(FORM_ITEM_JSON);
+
+    verify(processDefinitionServices, times(1)).getByKey(1L);
+    verify(formServices, times(1)).getLatestVersionByFormId("formId");
+  }
+
+  @Test
+  public void shouldReturn404ForFormInvaliProcessKey() throws Exception {
+    when(processDefinitionServices.getByKey(999L))
+        .thenThrow(new NotFoundException("Process Definition with key 999 not found"));
```
christinaausley, line 253: This is a nitpick but I would lowercase definition, so "Process definition".

nathansandi: Hi @christinaausley, Thanks for the suggestion, this is a test class and this behaviour is only threw on tests. I was checking the another test classes, and they are following this standard, for example Process Instance on `ProcessInstanceQueryControllerTest`

Same standard on all services, So this would require a change in general. @ralfpuchert  which standard should we follow over the error messages? 

I will keep as is for now and we can change all together in a different issue if it is the case @christinaausley

ralfpuchert: I prefer lowercase definition, in this case "Process definition". We can change this step by step in all beginning here.

ralfpuchert (approved):
LGTM, see my comment regarding messages.

PR #23929, opened 2024-10-23
feat: HealthMonitor reports arbitrarily nested component in a tree-like structure

entangled90:
## Description
HealthMonitor reports arbitrarily nested component in a tree-like structure

CriticalHealthMonitor recursively updates registered components in order to aggregate their health status into a single recursive data-type

HealthReport has been refactored by adding a children field that makes it recursive. Small refactoring has been done, primarily moving some functions from CriticalComponentHealthMonitor into HealthReport.

HealthStatus has been extended with a comparator, so that it's possible to take the "maximum" of a collection of HealthStatus in a way that returns the "worst" HealthStatus.

## Related issues
relates #16454


zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/ActorControl.java
```diff
@@ -240,6 +266,30 @@ public void submit(final Runnable action) {
     }
   }
 
+  // TODO check it
+  @SuppressWarnings("unchecked")
+  public <T> ActorFuture<T> submitCallable(final Callable<T> callable) {
```
entangled90, line 271: it already includes draft changes in #23912

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/StreamProcessor.java
```diff
@@ -464,19 +466,24 @@ public ActorFuture<Long> getLastWrittenPositionAsync() {
...
+      return HealthReport.unhealthy(this)
+          .withMessage("actor is closed", streamProcessorContext.getClock().instant());
     }
 
     if (processingStateMachine != null && !processingStateMachine.isMakingProgress()) {
       return HealthReport.unhealthy(this)
-          .withMessage("Processing not making progress. It is in an error handling loop.");
+          .withMessage(
+              "Processing not making progress. It is in an error handling loop.",
+              streamProcessorContext.getClock().instant());
```
npepinpe, line 477: :thinking: Interesting. The `StreamClock` is controllable and used mostly for externally visible things - e.g. writing when a command was processed on the log, for example, or for scheduling of BPMN timers. This allows users to "time travel" in their processes, which is very useful for testing (e.g. "trigger this timer on November 24th 2031", and the corresponding events and so on are at the time). It's divorced from the actor clock or system clock to avoid triggering timeouts and other time-sensitive things in other parts, where we need wall time/logical time (e.g. Raft leader election).

In this case, we would likely want to show the actor clock, I think? This is more about internal failures than user processing, so I would think that's the right clock to use. `ActorClock.instant()`.

npepinpe (changes requested):
:rocket: First, I generally really like the changes and how you approached the PR. I saw you started initially with recomputing on every update, and then changed that down the line :+1: It's a lot of comments, but I want to highlight generally I like the PR :slightly_smiling_face: 

:question: My main question here is why are we separating it from the partition status? Do we think it would pollute the result of the partition status call, and we want to keep it separate? That's a fair point, but I'm also not sure about it honestly. I would poll the other engineers in our team channel on it (I can do that for you so you see how we do it via slack :))

I'm also not super into adding a new endpoint `/actuator/partitionHealth`. It's additional endpoints for little benefit I think, and makes it harder to discover partition related things. Why can't we use `/actuator/partitions/health`? We can simply add a new operation in the existing actuator endpoint (`BrokerAdminServiceEndpoint`). If it's complicated because of how it's set up now (as a plain `Endpoint`), one option is migrating it to a `WebEndpoint`).

One thing we've been wanting to do is de-Springify the modules, and move all Spring magic into a single module (`dist`), or leave it in Spring specific modules (e.g. the REST API). This is to avoid complex object graphs which become hard to track (if anything can provide beans), and minimize in general the Spring magic. So one opportunity here is moving the existing `BrokerAdminServiceEndpoint` to the `dist` module, for example. There's also some possible refactoring as follow up we can do in the admin service itself from what I can see :smile: (but we can discuss that separately).

zeebe/scheduler/src/test/java/io/camunda/zeebe/scheduler/health/CriticalComponentsHealthMonitorTest.java
```diff
@@ -165,10 +179,61 @@ public void shouldMonitorComponentDeath() {
...
+    waitUntilAllDone();
+    final var root = parentComponents[0];
+    // when
+    // set a child unhealthy at level +1: all its "parents" will be unhealthy
+    final var unhealthyFrom = levels - 2;
+    components[unhealthyFrom][0].setUnhealthy();
+    waitUntilAllDone();
+    final var report = root.getHealthReport();
+    var parentAtLevel = report;
+    // then
```
npepinpe, line 218: :upside_down_face: 

```suggestion

    // then
    final var report = root.getHealthReport();
    var parentAtLevel = report;
```

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/management/BrokerAdminServiceEndpointTest.java
```diff
@@ -0,0 +1,69 @@
...
+            .setContentType(ContentType.JSON)
+            // set URL explicitly since we want to ensure the mapping is correct
+            .setBaseUri("http://localhost:" + broker.mappedPort(TestZeebePort.MONITORING) + "/foo")
+            .addFilter(new ResponseLoggingFilter())
+            .addFilter(new RequestLoggingFilter())
+            .build();
+  }
+
+  @Test
+  void shouldReturnPartitions() {
```
npepinpe, line 52: :wrench: Could we test an expected response? Just a basic response I think is fine. That way if we also change the output, we would notice it could be breaking.

npepinpe (approved):
:rocket: nice

Please address all comments, but I'm pre-approving because I doubt we need another review (though if you want, you can ofc request one!)

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/management/BrokerAdminServiceEndpointTest.java
```diff
@@ -0,0 +1,69 @@
...
+public class BrokerAdminServiceEndpointTest {
+
+  static RequestSpecification brokerServerSpec;
+
+  @TestZeebe(initMethod = "initTestStandaloneBroker")
+  private static TestStandaloneBroker broker;
+
+  @SuppressWarnings("unused")
+  static void initTestStandaloneBroker() {
+    broker = new TestStandaloneBroker().withProperty("management.server.base-path", "/foo");
```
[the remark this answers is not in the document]
entangled90: I tried what you suggested, but the test fails. I don't see the broker starting up in time. I see that the annotation is used in quite a few tests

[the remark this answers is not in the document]
npepinpe: Let's do it together tomorrow

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/management/BrokerAdminServiceEndpointTest.java
entangled90: I added an expected json for both routes. Unfortunately there is a timestamp in the response, so I had to replace that timestamp to a "static" value before checking.

PR #23981, opened 2024-10-24
feat: Implement extended support for variables to rdbms module #23273

### LINKED ISSUE #23273: Implement Exporter and SearchClient for Variables
The RDBMS Camunda 8 REST API V2 should support querying for Variables.

**AcceptanceCriteria:**
- the RDBMS module fully implements the VariableSearchClient with all filters and fields
- the RDBMS module can process all relevant Zeebe records for the Variables

p-wunderlich:
## Description

Implement extended support for variables to rdbms module

- Integrate search client
- Sorting of variables
- Filter incl complex filtering of variable value

## Checklist

- [ ] for CI changes:

## Related issues

closes #23273

### DIFF
--- a/db/rdbms/pom.xml
+++ b/db/rdbms/pom.xml
@@ -36,11 +36,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-lang3</artifactId>
-    </dependency>
-
-    <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
       <scope>test</scope>
--- a/db/rdbms/pom.xml
+++ b/db/rdbms/pom.xml
@@ -1,4 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+  ~ one or more contributor license agreements. See the NOTICE file distributed
+  ~ with this work for additional information regarding copyright ownership.
+  ~ Licensed under the Camunda License 1.0. You may not use this file
+  ~ except in compliance with the Camunda License 1.0.
+  -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
@@ -26,6 +33,11 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
     </dependency>
 
     <dependency>
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/VariableReader.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/VariableReader.java
@@ -32,7 +32,7 @@ public class VariableReader {
             new VariableQuery(
                 new Builder().variableKeys(key).build(),
                 VariableSort.of(b -> b),
-                SearchQueryPage.of(b -> b.from(0).size(10))))
+                SearchQueryPage.of(b -> b.from(0).size(1))))
         .hits
         .getFirst();
   }
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/VariableReader.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/VariableReader.java
@@ -8,9 +8,18 @@
 package io.camunda.db.rdbms.read.service;
 
 import io.camunda.db.rdbms.sql.VariableMapper;
-import io.camunda.db.rdbms.write.domain.VariableDbModel;
+import io.camunda.search.entities.VariableEntity;
+import io.camunda.search.filter.VariableFilter.Builder;
+import io.camunda.search.page.SearchQueryPage;
+import io.camunda.search.query.VariableQuery;
+import io.camunda.search.sort.VariableSort;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class VariableReader {
+
+  private static final Logger LOG = LoggerFactory.getLogger(VariableReader.class);
 
   private final VariableMapper variableMapper;
 
@@ -18,7 +27,22 @@ public class VariableReader {
     this.variableMapper = variableMapper;
   }
 
-  public VariableDbModel findOne(final Long key) {
-    return variableMapper.findOne(key);
+  public VariableEntity findOne(final Long key) {
+    return search(
+            new VariableQuery(
+                new Builder().variableKeys(key).build(),
+                VariableSort.of(b -> b),
+                SearchQueryPage.of(b -> b.from(0).size(10))))
+        .hits
+        .getFirst();
   }
+
+  public SearchResult search(final VariableQuery filter) {
+    LOG.trace("[RDBMS DB] Search for variables with filter {}", filter);
+    final var totalHits = variableMapper.count(filter);
+    final var hits = variableMapper.search(filter);
+    return new SearchResult(hits, totalHits.intValue());
+  }
+
+  public record SearchResult(List<VariableEntity> hits, Integer total) {}
 }
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/VariableDbModel.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/VariableDbModel.java
@@ -7,11 +7,143 @@
  */
 package io.camunda.db.rdbms.write.domain;
 
+import org.apache.commons.lang3.math.NumberUtils;
+
 public record VariableDbModel(
     Long key,
-    Long processInstanceKey,
-    Long scopeKey,
     String name,
+    ValueTypeEnum type,
+    Double doubleValue,
+    Long longValue,
     String value,
-    boolean truncated,
-    String tenantId) {}
+    String fullValue,
+    boolean isPreview,
+    Long scopeKey,
+    Long processInstanceKey,
+    String tenantId) {
+
+  public static final int DEFAULT_VARIABLE_SIZE_THRESHOLD = 8191; // TODO make configurable
+
+  public static class VariableDbModelBuilder {
+
+    private Long key;
+    private String name;
+    private String value;
+    private Long scopeKey;
+    private Long processInstanceKey;
+    private String tenantId;
+
+    // Public constructor to initialize the builder
+    public VariableDbModelBuilder() {}
+
+    // Builder methods for each field
+    public VariableDbModelBuilder key(final Long key) {
+      this.key = key;
+      return this;
+    }
+
+    public VariableDbModelBuilder value(final String value) {
+      this.value = value;
+      return this;
+    }
+
+    public VariableDbModelBuilder name(final String name) {
+      this.name = name;
+      return this;
+    }
+
+    public VariableDbModelBuilder scopeKey(final Long scopeKey) {
+      this.scopeKey = scopeKey;
+      return this;
+    }
+
+    public VariableDbModelBuilder processInstanceKey(final Long processInstanceKey) {
+      this.processInstanceKey = processInstanceKey;
+      return this;
+    }
+
+    public VariableDbModelBuilder tenantId(final String tenantId) {
+      this.tenantId = tenantId;
+      return this;
+    }
+
+    // Build method to create the record
+    public VariableDbModel build() {
+      if (value != null && value.length() > DEFAULT_VARIABLE_SIZE_THRESHOLD) {
+        return getModelWithPreview();
+      } else {
+        if (!NumberUtils.isParsable(value)) {
+          return getModel();
+        } else {
+          return getNumericModel();
+        }
+      }
+    }
+
+    private VariableDbModel getModel() {
+      return new VariableDbModel(
+          key,
+          name,
+          ValueTypeEnum.STRING,
+          null,
+          null,
+          value,
+          null,
+          false,
+          scopeKey,
+          processInstanceKey,
+          tenantId);
+    }
+
+    private VariableDbModel getModelWithPreview() {
+      return new VariableDbModel(
+          key,
+          name,
+          ValueTypeEnum.STRING,
+          null,
+          null,
+          value.substring(0, DEFAULT_VARIABLE_SIZE_THRESHOLD),
+          value,
+          true,
+          scopeKey,
+          processInstanceKey,
+          tenantId);
+    }
+
+    private VariableDbModel getNumericModel() {
+      if (value.contains(".")) {
+        return new VariableDbModel(
+            key,
+            name,
+            ValueTypeEnum.DOUBLE,
+            Double.parseDouble(value),
+            null,
+            value,
+            null,
+            false,
+            scopeKey,
+            processInstanceKey,
+            tenantId);
+      } else {
+        return new VariableDbModel(
+            key,
+            name,
+            ValueTypeEnum.LONG,
+            null,
+            Long.parseLong(value),
+            value,
+            null,
+            false,
+            scopeKey,
+            processInstanceKey,
+            tenantId);
+      }
+    }
+  }
+
+  enum ValueTypeEnum {
+    STRING,
+    LONG,
+    DOUBLE
+  }
+}
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/VariableDbModel.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/VariableDbModel.java
@@ -7,7 +7,10 @@
  */
 package io.camunda.db.rdbms.write.domain;
 
-import org.apache.commons.lang3.math.NumberUtils;
+import static io.camunda.util.ValueTypeUtil.mapBoolean;
+
+import io.camunda.search.entities.ValueTypeEnum;
+import io.camunda.util.ValueTypeUtil;
 
 public record VariableDbModel(
     Long key,
@@ -72,15 +75,16 @@ public record VariableDbModel(
       if (value != null && value.length() > DEFAULT_VARIABLE_SIZE_THRESHOLD) {
         return getModelWithPreview();
       } else {
-        if (!NumberUtils.isParsable(value)) {
-          return getModel();
-        } else {
-          return getNumericModel();
-        }
+        return switch (ValueTypeUtil.getValueType(value)) {
+          case LONG -> getLongModel();
+          case DOUBLE -> getDoubleModel();
+          case BOOLEAN -> getModel(mapBoolean(value));
+          default -> getModel(value);
+        };
       }
     }
 
-    private VariableDbModel getModel() {
+    private VariableDbModel getModel(final String value) {
       return new VariableDbModel(
           key,
           name,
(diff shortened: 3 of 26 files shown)


db/rdbms/src/main/resources/db/changelog/rdbms-support/changesets/create_engine.xml
```diff
@@ -60,14 +60,19 @@
...
 
     <createIndex tableName="VARIABLE" indexName="IDX_VARIABLE_PROCESS_INSTANCE_KEY">
       <column name="PROCESS_INSTANCE_KEY" />
     </createIndex>
+    <createIndex tableName="VARIABLE" indexName="IDX_VARIABLE_SCOPE_KEY">
+      <column name="SCOPE_KEY" />
+    </createIndex>
+    <createIndex tableName="VARIABLE" indexName="IDX_VARIABLE_TENANT_ID">
+      <column name="TENANT_ID" />
+    </createIndex>
```
stoerti, line 75: Maybe not create too many indexes from the beginning on? Especially VARIABLES table could get slow on insert.

db/rdbms/src/main/resources/mapper/VariableMapper.xml
```diff
@@ -9,30 +9,123 @@
...
+  <!-- default search statement for databases supporting LIMIT/OFFSET-->
+  <select id="search" parameterType="io.camunda.db.rdbms.read.domain.VariableDbQuery" resultMap="io.camunda.db.rdbms.sql.VariableMapper.searchResultMap">
+    SELECT
+      VAR_KEY,
+      PROCESS_INSTANCE_KEY,
+      SCOPE_KEY,
+      VAR_NAME,
+      VAR_VALUE,
+      TENANT_ID,
+      false as IS_PREVIEW
```
stoerti, line 27: I see this isPreview in the entity but not in any filter.

Do we need a follow-up ticket for clarification/implementation?

qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/util/CamundaRdbmsTestApplication.java
```diff
@@ -45,9 +45,7 @@ public CamundaRdbmsTestApplication start() {
           .withProperty("spring.datasource.url", jdbcDatabaseContainer.getJdbcUrl())
           .withProperty("spring.datasource.username", jdbcDatabaseContainer.getUsername())
           .withProperty("spring.datasource.password", jdbcDatabaseContainer.getPassword())
-          .withProperty("spring.datasource.password", jdbcDatabaseContainer.getPassword())
-          .withProperty("logging.level.io.camunda.db.rdbms", "DEBUG")
-          .withProperty("logging.level.org.mybatis", "DEBUG");
+          .withProperty("spring.datasource.password", jdbcDatabaseContainer.getPassword());
```
stoerti, line 48: Double password config?

stoerti (dismissed):

db/rdbms/src/main/resources/db/changelog/rdbms-support/changesets/create_engine.xml
p-wunderlich: I oriented myself on C7, but maybe at least we can kick the Scope Key for now?

db/rdbms/src/main/resources/mapper/VariableMapper.xml
p-wunderlich: Yea I think we need to discuss this. Since we do NOT save truncated date, it's never a preview ... I'm not sure if the webapps need a preview ... I'll clarify this.

qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/util/CamundaRdbmsTestApplication.java
p-wunderlich: “Better safe than sorry”? :D  ("Doppel hält besser")

search/search-domain/src/main/java/io/camunda/search/entities/VariableEntity.java
```diff
@@ -15,7 +15,7 @@ public final record VariableEntity(
     String name,
     String value,
     String fullValue,
-    boolean isPreview,
+    Boolean isPreview,
```
danielkelemen, line 18: ❓Why can it not be a primitive boolean?

danielkelemen (approved):
👍 Looks good! Added a few suggestions and a question.
Approving it to not block you, this is my last day on this week.

db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/VariableReader.java
```diff
@@ -7,18 +7,43 @@
...
+  }
+
+  public SearchResult search(final VariableDbQuery filter) {
+    LOG.trace("[RDBMS DB] Search for variables with filter {}", filter);
+    final var totalHits = variableMapper.count(filter);
+    final var hits = variableMapper.search(filter);
+    return new SearchResult(hits, totalHits.intValue());
   }
+
+  public record SearchResult(List<VariableEntity> hits, Integer total) {}
```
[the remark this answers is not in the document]
stoerti: In #23997 in will remove this SearchResult record anyway and use the SearchQueryResult<> from search-domain.

search/search-domain/src/main/java/io/camunda/search/entities/VariableEntity.java
p-wunderlich: Because MyBatis always uses the Wrapper type in the resultMap. With the primitiv type, I get this error: 

`### SQL: SELECT       VAR_KEY,       PROCESS_INSTANCE_KEY,       SCOPE_KEY,       VAR_NAME,       VAR_VALUE,       TENANT_ID,       false as IS_PREVIEW     FROM VARIABLE           WHERE 1 = 1                        AND SCOPE_KEY IN        (  ? )                                               LIMIT ? OFFSET ?
### Cause: org.apache.ibatis.reflection.ReflectionException: Error instantiating class io.camunda.search.entities.VariableEntity with invalid types (Long,String,String,String,Boolean,Long,Long,String) or values (2,variable-name-48,variable-value-367,variable-value-367,false,1,3,tenant-937). Cause: java.lang.NoSuchMethodException: io.camunda.search.entities.VariableEntity.<init>(java.lang.Long,java.lang.String,java.lang.String,java.lang.String,java.lang.Boolean,java.lang.Long,java.lang.Long,java.lang.String)
`

Also when I use this as arg: 

`<arg column="IS_PREVIEW" javaType="boolean"/>`

db/rdbms/src/main/resources/mapper/VariableMapper.xml
p-wunderlich: See here: https://github.com/camunda/camunda/issues/24230

db/rdbms/src/main/java/io/camunda/db/rdbms/read/domain/VariableDbFilter.java
```diff
@@ -0,0 +1,76 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.db.rdbms.read.domain;
+
+import io.camunda.search.filter.VariableFilter;
+import io.camunda.search.filter.VariableValueFilter;
+import java.util.List;
+
+/**
+ * Since the operator fields in the VariableValueFilter are reserved keywords in mybatis (OGNL
```
romansmirnov, line 15: ❓ I understand that operators like `eq,` `neq,` etc. are reserved keywords, but I don't fully understand why it is necessary to wrap them into another filter object. Additionally, couldn't we do something similar - what is done with `VariableValueDbFilter` - in the layer above already? So instead of having class properties like `eq` (and others) in the `VariableValueFilter`, wrap them in a record/tuple that contains the name, operator, and the value. Would that allow us to avoid having that additional filter`?

db/rdbms/src/main/java/io/camunda/db/rdbms/read/domain/VariableDbQuery.java
```diff
@@ -0,0 +1,77 @@
...
+import io.camunda.search.filter.VariableFilter;
+import io.camunda.search.page.SearchQueryPage;
+import io.camunda.search.query.SearchQueryBase;
+import io.camunda.search.sort.SortOptionBuilders;
+import io.camunda.search.sort.VariableSort;
+import io.camunda.util.ObjectBuilder;
+import java.util.Objects;
+import java.util.function.Function;
+
+public record VariableDbQuery(VariableDbFilter filter, VariableSort sort, SearchQueryPage page) {
```
romansmirnov, line 20: ❓ Why do we need to wrap them once again in another `VariableDbQuery` object? Putting the `VariableDbFilter` aside, why can't the `VariableQuery` be used instead? I noticed that the same is done in other places as well, like with `ProcessInstanceQuery`, but I couldn't find a reason for it.

romansmirnov (commented):
I have two questions to better understand why certain things are done that way.

db/rdbms/src/main/java/io/camunda/db/rdbms/read/domain/VariableDbFilter.java
p-wunderlich: Sure, this would be much better in the layer above. :) I just don't wanted to changed existing behaviour since this also effects the ES/OS Layer. I can do this directly with this PR.

db/rdbms/src/main/java/io/camunda/db/rdbms/read/domain/VariableDbQuery.java
p-wunderlich: Hm, I'm not really sure, I think we did this to have the default values there. When I get rid of the VariableDbFilter, I can try to do it also without this wrapper class. But then I maybe have to set the Defaults in some layer above.

db/rdbms/src/main/java/io/camunda/db/rdbms/read/domain/VariableDbFilter.java
p-wunderlich: This class is not longer needed by using the new Operation Strategy which is already in the VariableFilter included. 

private Map<String, List<Operation<Object>>> variableOperations;

db/rdbms/src/main/java/io/camunda/db/rdbms/read/domain/VariableDbQuery.java
p-wunderlich: Removed it and we now work directly with the VariableQuery

db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/VariableDbModel.java
```diff
@@ -7,11 +7,94 @@
...
 
+import org.apache.commons.lang3.math.NumberUtils;
+
 public record VariableDbModel(
     Long key,
-    Long processInstanceKey,
-    Long scopeKey,
     String name,
+    double varDouble,
+    long varLong,
```
danielkelemen, line 16: 🔧 I feel like `<type>Value` naming is better.
❓ Shouldn't these fields be defined boxed primitives as one of them will always be null?

```suggestion
    Double doubleValue,
    Long longValue,
```

db/rdbms/src/main/resources/mapper/VariableMapper.xml
```diff
@@ -9,30 +9,162 @@
...
+        VAR_VALUE != #{operation.value}
+      </when>
+      <when test="operation.operator.name().equals('EXISTS')">
+        VAR_VALUE IS NOT NULL
+      </when>
+      <when test="operation.operator.name().equals('NOT_EXISTS')">
+        VAR_VALUE IS NULL
+      </when>
+      <when test="operation.operator.name().equals('GREATER_THAN')">
+        VAR_LONG &gt; #{operation.value} OR VAR_DOUBLE &gt; #{operation.value}
```
danielkelemen, line 100: ❌ These could yield wrong results with the comparison and `or` condition if we use primitives with 0.

I think we need to either:
- Use null instead of 0 for _empty_ values and add null checks here.
- Add another column where we store the type of the variable and use the corresponding value for comparisons. This might be the cleanest solution. You might be able to reuse the `operationCondition` snippet.

danielkelemen (dismissed):

db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/VariableDbModel.java
p-wunderlich: Yes, good idea, I will change it

db/rdbms/src/main/resources/mapper/VariableMapper.xml
p-wunderlich: You are right, I added a type column and use it here for the conditions

db/rdbms/src/main/java/io/camunda/db/rdbms/write/utils/NumberChecker.java
```diff
@@ -0,0 +1,22 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.db.rdbms.write.utils;
+
+public class NumberChecker {
+
+  private static final String INTEGER_REGEX = "^-?\\d+$";
+  private static final String DOUBLE_REGEX = "^-?\\d*[.,]?\\d+$";
```
stoerti, line 13: This does not cover scientific notation of Doubles like "1.1e7". DO we need to cover it?

p-wunderlich: @romansmirnov @danielkelemen ?

p-wunderlich: A, I dont use this class. :D Was my first IMPL before using apache commons.

db/rdbms/src/main/resources/mapper/VariableMapper.xml
```diff
@@ -9,30 +9,172 @@
...
+      </foreach>
+      false )
+    </if>
+
+  </sql>
+
+  <sql id="variableOperationCondition">
+    <choose>
+      <when test="operation.operator.name().equals('EQUALS')">
+        VAR_VALUE = #{operation.value}
```
romansmirnov, line 88: ❓ Does it work in all cases with numeric and non-numeric values?

db/rdbms/src/main/resources/mapper/VariableMapper.xml
```diff
@@ -9,30 +9,172 @@
...
+      </when>
+      <when test="operation.operator.name().equals('NOT_EQUALS')">
+        VAR_VALUE != #{operation.value}
+      </when>
+      <when test="operation.operator.name().equals('EXISTS')">
+        VAR_VALUE IS NOT NULL
+      </when>
+      <when test="operation.operator.name().equals('NOT_EXISTS')">
+        VAR_VALUE IS NULL
+      </when>
```
romansmirnov, line 98: ❓ The semantic of `exists` and `not_exists` is not clear when searching for variables. Probably, they are not needed and can be removed for now.

db/rdbms/src/main/resources/mapper/VariableMapper.xml
```diff
@@ -9,30 +9,172 @@
...
+      <when test="operation.operator.name().equals('GREATER_THAN')">
+        (TYPE = 'LONG' AND LONG_VALUE &gt; #{operation.value})
+        OR
+        (TYPE = 'DOUBLE' AND DOUBLE_VALUE &gt; #{operation.value})
```
romansmirnov, line 102: ❓ Can the type already be determined in the application before selecting?

romansmirnov (commented):
As discussed some comments.

search/search-domain/src/main/java/io/camunda/search/filter/Operation.java
```diff
@@ -10,13 +10,18 @@
 import io.camunda.util.CollectionUtil;
 import java.util.List;
 
-public record Operation<T>(Operator operator, List<T> values) {
+public record Operation<T>(Operator operator, List<T> values, boolean numericNeeded) {
```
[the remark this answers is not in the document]
p-wunderlich: No, but I have to change the Operation Model anyway because of other suggestions Roman made.

db/rdbms/src/main/resources/mapper/VariableMapper.xml
p-wunderlich: Fixed in newest commit

db/rdbms/src/main/resources/mapper/VariableMapper.xml
p-wunderlich: For search queries where we just search for the name "EXISTS" is used. So it makes no sense here, but we still "get" this operation. I now simply ignore ist.

db/rdbms/pom.xml
```diff
@@ -1,4 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+  ~ one or more contributor license agreements. See the NOTICE file distributed
+  ~ with this work for additional information regarding copyright ownership.
+  ~ Licensed under the Camunda License 1.0. You may not use this file
+  ~ except in compliance with the Camunda License 1.0.
+  -->
```
danielkelemen, line 8: 🔧 I don't think copyright is needed in pom.

db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/VariableReader.java
```diff
@@ -8,17 +8,41 @@
...
   }
 
-  public VariableDbModel findOne(final Long key) {
-    return variableMapper.findOne(key);
+  public VariableEntity findOne(final Long key) {
+    return search(
+            new VariableQuery(
+                new Builder().variableKeys(key).build(),
+                VariableSort.of(b -> b),
+                SearchQueryPage.of(b -> b.from(0).size(10))))
```
danielkelemen, line 35: ❓ Why does `findOne` need to search for a page of 10 records? Can it be just 0,1?

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java
```diff
@@ -475,7 +476,9 @@ public static <C extends List<Operation<Object>>> List<SearchQuery> variableOper
                   case LOWER_THAN_EQUALS ->
                       range(q -> q.field(varValue).lte(op.value())).toSearchQuery();
                   case IN -> objectTerms(varValue, op.values());
-                  default -> throw unexpectedOperation("Variable", op);
+                  default ->
+                      throw new IllegalStateException(
+                          "Unexpected %s operation: %s".formatted("Variable", op.operator()));
```
danielkelemen, line 481: 🔧 Method does the same, let's keep that.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java
```diff
@@ -453,7 +454,7 @@ public static <C extends List<Operation<OffsetDateTime>>> List<SearchQuery> date
     }
   }
 
-  public static <C extends List<Operation<Object>>> List<SearchQuery> variableOperations(
+  public static <C extends List<UntypedOperation>> List<SearchQuery> variableOperations(
```
danielkelemen, line 457: 💭 This will probably reverted back to just string instead of object. At least for the `variables/search` this is not needed yet.

danielkelemen (approved):
Looks good! Added 1-2 small notes.

db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/VariableReader.java
p-wunderlich: You are right!

db/rdbms/pom.xml
p-wunderlich: Removend

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java
p-wunderlich: Nope, the method can't deal with the new UntypedOperation which we have here.

p-wunderlich: But I just saw, that just the operator is used there. I reused the method there, and changed the method to just need the operator instead the whole operation.

PR #23988, opened 2024-10-24
Always upload artifacts on flaky test encounter

### LINKED ISSUE #23986: Upload test artifacts on flaky tests
## Description

With the changes in #21674 the CI was changed to not fail the build on flaky tests. An unexpected side-effect of this is that we no longer upload the test artifacts when we encounter a flake. As a result it becomes a lot harder to root-cause and fix them. We should make sure we always upload the artifacts when a test is flaky, regardless if the build failed or succeeds.

remcowesterhoud:
## Description

The job doesn't fail when a flaky test is encountered. As a result these artifacts didn't get uploaded anymore. Adding this condition makes it so the artifacts are always uploaded if a flaky tests is encountered.

Example build where the artifacts are uploaded on a successful job: https://github.com/camunda/camunda/actions/runs/11499226356

## Checklist

- [ ] for CI changes:

## Related issues

closes #23986

### DIFF
--- a/.github/workflows/zeebe-ci.yml
+++ b/.github/workflows/zeebe-ci.yml
@@ -54,6 +54,8 @@ jobs:
           distball: ${{ steps.build-zeebe.outputs.distball }}
           platforms: linux/${{ matrix.arch }}
           push: false
+      - name: Create build output log file
+        run: echo "BUILD_OUTPUT_FILE_PATH=$(mktemp)" >> $GITHUB_ENV
       - name: Run smoke test on ${{ matrix.arch }}
         env:
           # For non Linux runners there is no container available for testing, see build-zeebe-docker job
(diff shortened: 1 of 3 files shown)


.github/workflows/zeebe-ci.yml
```diff
@@ -66,9 +68,15 @@ jobs:
...
+      - name: Analyze Test Runs
+        id: analyze-test-run
+        if: always()
+        uses: ./.github/actions/analyze-test-runs
+        with:
+          buildOutputFilePath: ${{ env.BUILD_OUTPUT_FILE_PATH }}
       - name: Upload test artifacts
         uses: ./.github/actions/collect-test-artifacts
-        if: ${{ failure() || cancelled() }}
+        if: ${{ failure() || cancelled() || steps.analyze-test-run.outputs.flakyTests != '' }}
```
remcowesterhoud, line 79: ❌ I struggle getting this one to work. The CI complains that the `BUILD_OUTPUT_FILE_PATH` doesn't exist. I've added the step to create this file, but it still complains about it. Could 1 of the reviewers have a look and see if I missed anything?

remcowesterhoud: Nvm I got it

kellervater (approved):
With still limited knowledge I'd say it's looking good.
CI is green, no structural changes, so I'd say LGTM

PR #23991, opened 2024-10-24
Map PartitionInactiveException and NoTopologyFoundException to UNAVAILABLE

### LINKED ISSUE #23446: `PartitionInactiveException` is not handled gracefully
When a `PartitionInactiveException` occurs (introduced in https://github.com/camunda/camunda/pull/19876), the `GrpcErrorMapper` does not handle this exception gracefully, eventually leading to an error log about an unexpected error.

Error group: https://console.cloud.google.com/errors/detail/CLLEq6jc8LnF6QE;service=zeebe;time=P7D;locations=global?project=camunda-saas-dev

npepinpe:
## Description

This PR ensures we map `PartitionInactiveException` and `NoTopologyFoundException` errors to UNAVAILABLE (gRPC) and 503 (SERVICE_UNAVAILABLE for REST) respectively. This avoids logging them as internal server errors, which creates alerts and noise, as these issues are typically transient.

As a side effect, I realized `NoTopologyFoundException` was not mapped anywhere in the REST API, so I added this as well.

Additionally, it seems `InvalidBrokerRequestArgumentException` was not used anymore (except to map the error), so I removed it.

I did not add unit tests for the `GrpcErrorMapper` because the logic is really quite simple. I also did not add unit tests for the `RestErrorMapper` because it seems there are none, and I assume this is a conscious decision.

## Related issues

closes #23446


entangled90:
I checked if all `BrokerResponseException` are handled and I think these are missing:
- `UnsupportedBrokerResponseException`
- `IllegalBrokerResponseException`

Is it worth fixing them all?
:wrench:  should make `BrokerResponseException` sealed?

npepinpe:
IIRC `BrokerResponseException` are used when something unexpected happened and we couldn't parse the response from the broker. In this case, the default mapping (`INTERNAL` or 500) is appropriate, since retrying wouldn't make sense, and there's a bug in the server communication.

entangled90 (approved):

PR #24003, opened 2024-10-24
feat: add user's authorizations search endpoint

### LINKED ISSUE #23800: Camunda rest api can retrieve list of authorizations assigned to a user
### Description
- [x] Add `/users/{key}/authorizations/search` endpoint following [the guidelines](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)

maryarm:
## Description

Add` /users/{id}/authorizations/search` endpoint following [the guidelines](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)

## Related issues

closes #23800


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2405,6 +2405,72 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
+  /users/{userKey}/authorizations/search:
+    post:
+      tags:
+        - Authorization
+      summary: "Query a User's authorizations (alpha)"
```
christinaausley, line 2412:
```suggestion
      summary: "Query a user's authorizations (alpha)"
```

houssain-barouni (commented):
@maryarm A general question: What is the difference in usage between the following two requests?
```
POST /users/1234/authorizations/search
{}
```
and

```
POST /authorizations/search
{
    "ownerKey":  1234
}
```

maybe for the first one I expect that we return 404 when a user with key 1234 cannot be found and 403 if the caller cannot read the user with that key, is that correct?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/AuthorizationQueryController.java
```diff
@@ -41,13 +43,21 @@ public ResponseEntity<AuthorizationSearchResponse> searchAuthorizations(
...
-      path = "/v2/users/{id}/authorizations/search",
+      path = "/v2/users/{userKey}/authorizations/search",
       produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
       consumes = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<AuthorizationSearchResponse> searchUserAuthorizations(
-      @PathVariable("id") final long id,
+      @PathVariable("userKey") final long userKey,
       @RequestBody(required = false) final AuthorizationSearchQueryRequest query) {
-    return SearchQueryRequestMapper.toAuthorizationQuery(query)
+    var finalQuery = query;
```
danielkelemen, line 52: 🔧 You could use the query from the parameter without the `final` keyword.

maryarm:
> @maryarm A general question: What is the difference in usage between the following two requests?
> 
> ```
> POST /users/1234/authorizations/search
> {}
> ```
> 
> and
> 
> ```
> POST /authorizations/search
> {
>     "ownerKey":  1234
> }
> ```
> 
> maybe for the first one I expect that we return 404 when a user with key 1234 cannot be found and 403 if the caller cannot read the user with that key, is that correct?

Very good point, what you suggest makes sense but it's not clear from spec and what I understand is that these two endpoints can be used interchangeably (we had the same condition for /users/search and get a user).

christinaausley (approved):
No grammatical adjustments on my end 👍

pepopowitz (approved):
👍 spec changes look good to me

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/query/TypedSearchQueryTransformer.java
```diff
@@ -51,8 +51,8 @@ public SearchQueryRequest toSearchQueryRequest(
 
     final var sort = query.sort();
     final var sorting = toSearchSortOptions(sort, reverse);
-    if (sorting != null && !sorting.isEmpty()) {
-      builder.sort(sorting);
+    if (!sorting.isEmpty()) {
+      builder.sort(query.retainValidSortings(sorting));
```
houssain-barouni, line 55: @maryarm the invalid sorting fields are already reported in https://github.com/camunda/camunda/blob/778d3f04b151c083ffde088126a1e94aafc56602/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java#L884-L899
do you expect to have other flow that can bring any invalid sort fields?

maryarm: Maybe I can reuse this method, Although my case is not reporting to the rest caller, the problem is that there is a logic that adds default sorting with the field "key" and as Authorization doesn't have the search is throwing exception

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2466,6 +2466,72 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
 
+  /users/{userKey}/authorizations/search:
+    post:
+      tags:
+        - Authorization
+      summary: "Query a user's authorizations (alpha)"
```
tmetzke, line 2473: 🔧 To be consistent, this should just be "Query user authorizations (alpha)"
```suggestion
      summary: "Query user authorizations (alpha)"
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2466,6 +2466,72 @@ paths:
...
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "404":
+          description: "Not found"
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
tmetzke, line 2527: 🔧 I think this was mentioned before: those response seem unlikely for a search endpoint. If they are not, all other search endpoints should add them for consistency. Do we have any solid indication that those response codes will ever be seen by users for this search endpoint?

tmetzke: I have adjusted this for other search endpoints with https://github.com/camunda/camunda/pull/24190 to streamline the endpoints.

tmetzke:
The things mentioned above regarding the OpenAPI spec can of course simply go into a follow-up PR but please address them somehow, @maryarm. Otherwise, our OpenAPI spec becomes inconsistent for users quite fast.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2523,6 +2523,57 @@ paths:
...
+        - Authorization
+      summary: "Query user authorizations (alpha)"
+      description: |
+        Search for user authorizations based on given criteria.
+
+        :::note
+        This endpoint is an alpha feature and not enabled on Camunda clusters out of the box.
+        See the [Camunda 8 REST API overview](/apis-tools/camunda-api-rest/camunda-api-rest-overview.md#query-api)
+        for further details.
+        :::
```
tmetzke, line 2538: 🚀 Thanks for adjusting the spec directly 🦖 

🔧 Can we adjust this description as well for consistency?
```suggestion
        Note that this endpoint is an alpha feature and not enabled on Camunda clusters out of the box.
        The Camunda 8 API (REST) Overview page provides further details.

        :::note
        This endpoint is an [alpha feature](/components/early-access/alpha/alpha-features.md) and may be subject to change
        in future releases.
        :::
```

tmetzke:
> > @maryarm A general question: What is the difference in usage between the following two requests?
> > ```
> > POST /users/1234/authorizations/search
> > {}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```
> > POST /authorizations/search
> > {
> >     "ownerKey":  1234
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > maybe for the first one I expect that we return 404 when a user with key 1234 cannot be found and 403 if the caller cannot read the user with that key, is that correct?
> 
> Very good point, what you suggest makes sense but it's not clear from spec and what I understand is that these two endpoints can be used interchangeably (we had the same condition for /users/search and get a user).

Good questions indeed. I would argue that the search endpoints simply don't return results that you are not allowed to see (tenant-wise and permission-wise). The search endpoints will not return 404s, 401s, or 403s but rather an empty result because nothing you can access matches your query.

If you use GET endpoints, you'll get those status codes potentially.

houssain-barouni:
> > > @maryarm A general question: What is the difference in usage between the following two requests?
> > > ```
> > > POST /users/1234/authorizations/search
> > > {}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and
> > > ```
> > > POST /authorizations/search
> > > {
> > >     "ownerKey":  1234
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > maybe for the first one I expect that we return 404 when a user with key 1234 cannot be found and 403 if the caller cannot read the user with that key, is that correct?
> > 
> > 
> > Very good point, what you suggest makes sense but it's not clear from spec and what I understand is that these two endpoints can be used interchangeably (we had the same condition for /users/search and get a user).
> 
> Good questions indeed. I would argue that the search endpoints simply don't return results that you are not allowed to see (tenant-wise and permission-wise). The search endpoints will not return 404s, 401s, or 403s but rather an empty result because nothing you can access matches your query.
> 
> If you use GET endpoints, you'll get those status codes potentially.

Thanks for the clarification @tmetzke .
I would say it depends. In the case of the first endpoint, `POST /users/1234/authorizations/search`, the user key is part of the path-context, meaning we’re querying something within the `/users/1234` resource. So, if the user with that key doesn’t exist, it seems more logical to return a 404 status to indicate that the user resource itself is missing.

tmetzke:
> Thanks for the clarification @tmetzke . I would say it depends. In the case of the first endpoint, `POST /users/1234/authorizations/search`, the user key is part of the path-context, meaning we’re querying something within the `/users/1234` resource. So, if the user with that key doesn’t exist, it seems more logical to return a 404 status to indicate that the user resource itself is missing.

Oh yeah, you're absolutely right, I totally missed that 🙈 
That makes by far the most sense for all "subresource" search endpoints 👍 
Let's add it here, sorry for the noise 😅 
I will check all other search endpoints again regarding this as a follow-up 🔍

maryarm:
> > Thanks for the clarification @tmetzke . I would say it depends. In the case of the first endpoint, `POST /users/1234/authorizations/search`, the user key is part of the path-context, meaning we’re querying something within the `/users/1234` resource. So, if the user with that key doesn’t exist, it seems more logical to return a 404 status to indicate that the user resource itself is missing.
> 
> Oh yeah, you're absolutely right, I totally missed that 🙈 That makes by far the most sense for all "subresource" search endpoints 👍 Let's add it here, sorry for the noise 😅 I will check all other search endpoints again regarding this as a follow-up 🔍

Thanks for the interesting discussion, based on our latest agreement with @tmetzke we are not sure if this is worth the effort as we need to do extra querying to verify the user before we do the actual search query: 
"We could say that we don't do this for now, it could be an improvement in future iterations if people would like this. It's simply more detailed status codes for such cases, the general endpoint functionality doesn't change. But it would also be fine to use the general search endpoint semantics only and not provide those detailed status codes to save on additional queries to Elasticsearch."

PR #24009, opened 2024-10-24
pg-fixes-completion-variables

panagiotisgts:
## Description
- Implement the UserTaskCompletionVariableHandler to capture variables during user task completion in order to set correct relationship in the `tasklist-task` index.
- Include `position` field in TaskVariableEntity
- [bugfix] Use `formId` instead of `deployedResourceName` for FormHandler

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FormHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FormHandler.java
@@ -55,7 +55,7 @@ public class FormHandler implements ExportHandler<FormEntity, Form> {
     entity
         .setVersion((long) value.getVersion())
         .setKey(value.getFormKey())
-        .setBpmnId(value.getResourceName())
+        .setBpmnId(value.getFormId())
         .setSchema(new String(value.getResource(), StandardCharsets.UTF_8))
         .setTenantId(value.getTenantId())
         .setIsDeleted(record.getIntent().name().equals(FormIntent.DELETED.name()));
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.handlers;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.camunda.exporter.store.BatchRequest;
+import io.camunda.webapps.schema.descriptors.tasklist.template.TaskTemplate;
+import io.camunda.webapps.schema.entities.tasklist.TaskJoinRelationship;
+import io.camunda.webapps.schema.entities.tasklist.TaskJoinRelationship.TaskJoinRelationshipType;
+import io.camunda.webapps.schema.entities.tasklist.TaskVariableEntity;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.intent.UserTaskIntent;
+import io.camunda.zeebe.protocol.record.value.UserTaskRecordValue;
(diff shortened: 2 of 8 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FormHandler.java
```diff
@@ -55,7 +55,7 @@ public void updateEntity(final Record<Form> record, final FormEntity entity) {
     entity
         .setVersion((long) value.getVersion())
         .setKey(value.getFormKey())
-        .setBpmnId(value.getResourceName())
+        .setBpmnId(value.getFormId())
```
Zelldon, line 58: ❌ ❓ This seems a bit weird to me. Is this expected? Why is the property called bpmnId but we fill the formId? 🤔 Can we rename the property then?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  public Class<TaskVariableEntity> getEntityType() {
+    return TaskVariableEntity.class;
+  }
+
+  @Override
+  public boolean handlesRecord(final Record<UserTaskRecordValue> record) {
+    return UserTaskIntent.COMPLETED.name().equals(record.getIntent().name())
+        && record.getValue().getVariables() != null
+        && !record.getValue().getVariables().isEmpty();
```
Zelldon, line 59: 👍🏼

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
```diff
@@ -0,0 +1,133 @@
...
+    try {
+      variableStringValue = MAPPER.writeValueAsString(variableValue);
+    } catch (final JsonProcessingException e) {
+      LOGGER.error(
+          String.format(
+              "Failed to parse variable '%s' value as string '%s'", variableName, variableValue),
+          e);
+    }
+
+    if (variableStringValue.length() > variableSizeThreshold) {
```
Zelldon, line 105: ❓ Was the threshold inclusive or exclusive before?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandlerTest.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  @Test
+  void shouldHandleRecord() {
+    // given
+    final Record<UserTaskRecordValue> variableRecord =
+        factory.generateRecord(ValueType.USER_TASK, r -> r.withIntent(UserTaskIntent.COMPLETED));
+
+    // when - then
+    assertThat(underTest.handlesRecord(variableRecord)).isTrue();
```
Zelldon, line 55: 🤔 Does this generator also generate a record with a variable?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandlerTest.java
```diff
@@ -0,0 +1,221 @@
...
+    assertThat(variableEntity.getIsTruncated()).isFalse();
+    assertThat(variableEntity.getFullValue()).isNull();
+    assertThat(variableEntity.getJoin()).isNotNull();
+    assertThat(variableEntity.getJoin().getParent()).isEqualTo(scopeKey);
+    assertThat(variableEntity.getJoin().getName())
+        .isEqualTo(TaskJoinRelationshipType.TASK_VARIABLE.getType());
+  }
+
+  @Test
+  void shouldUpdateEntityFromRecordWithFullValue() {
```
Zelldon, line 188: 👍🏼

ChrisKujawa (commented):
Looks good to me in general. There is one issue we should first clarify, then I think we are good to go.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandlerTest.java
panagiotisgts: Not sure, maybe I should specify it since it could cause a flaky test if it does not contain one, good point :+1:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FormHandler.java
panagiotisgts: This is how it was back on Tasklist's original Index and Entity. I maintained the same notation across while doing the ports.

ChrisKujawa (approved):

PR #24010, opened 2024-10-24
test: filter for FlowNodeInstance with incident

### LINKED ISSUE #23978: Flaky `io.camunda.it.client.ProcessInstanceAndFlowNodeInstanceQueryTest shouldQueryFlowNodeInstanceByIncident`
**Summary**

- How often does the test fail? Often
- Does it block your work? No
- Do we suspect that it is a real failure? No clue

**Failures**

<pre>
org.opentest4j.AssertionFailedError: 

expected: 1
 but was: 20
	at io.camunda.it.client.ProcessInstanceAndFlowNodeInstanceQueryTest.shouldQueryFlowNodeInstanceByIncident(ProcessInstanceAndFlowNodeInstanceQueryTest.java:676)
	at java.base/java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:194)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
</pre>
</details>

**Logs**

Couldn't get any because the artifacts aren't stored. See https://camunda.slack.com/archives/C071KP5BTHB/p1729758811450069

Example CI run: https://github.com/camunda/camunda/actions/runs/11495201327

mustafadagher:
## Description

* Filter for `FlowNodeInstance` that has `incident = true` instead of one that has an incident key. The test seems to fail because the flow node instance can have an incident key but `incident = false` as those fields are apparently decoupled. The test case then uses the `hasIncident` value for querying with that value being `false`. Thus, we find all 20 flow node instances.
* We now throw an exception in case of no flowNode with `hasIncident = true` is found.

## Related issues

closes #23978

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/ProcessInstanceAndFlowNodeInstanceQueryTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/ProcessInstanceAndFlowNodeInstanceQueryTest.java
@@ -89,9 +89,9 @@ public class ProcessInstanceAndFlowNodeInstanceQueryTest {
     flowNodeInstance = allFlowNodeInstances.getFirst();
     flowNodeInstanceWithIncident =
         allFlowNodeInstances.stream()
-            .filter(f -> f.getIncidentKey() != null)
+            .filter(FlowNodeInstance::getIncident)
             .findFirst()
-            .orElse(null);
+            .orElseThrow();
   }
 
   @AfterAll


tmetzke:
* ⚠️ The flakiness might still prevail but fail now in the `beforeAll` if it can happen that the flow node instance can have an incident key but `hasIncident` is still `false`.
  * @ralfpuchert, would be great to get your take on this scenario

ralfpuchert:
I think we should add a "wait for incident in flow node instance method" in `QueryTest` as precondition for getting the flow node instance keys. 
Then it can be used as:
```java
    waitUntilFlowNodeInstanceHasIncident(zeebeClient, 1);
    // store flow node instances for querying
    final var allFlowNodeInstances =
        zeebeClient.newFlownodeInstanceQuery().page(p -> p.limit(100)).send().join().items();
    flowNodeInstance = allFlowNodeInstances.getFirst();
    flowNodeInstanceWithIncident =
        allFlowNodeInstances.stream()
            .filter(FlowNodeInstance::getIncident)
            .findFirst()
            .orElseThrow();
```

tmetzke:
Superseded by https://github.com/camunda/camunda/pull/24149

Issue #22271, opened 2024-09-12
Include Optimize in the single JAR with the rest of the stack


buccarel:
### Context

Currently, Optimize is present within the monorepo but it's not inheriting from the zeebe parent POM (i.e., parent/pom.xml) and it is not being packaged as part of the single JAR with the rest of the stack. The objective of this task is to establish such parenting relationship for Optimize and ensure that Optimize is packages as part of the single application correctly.

Once this task is completed, it shall unblock the following tasks:
* https://github.com/camunda/camunda/issues/22294
* https://github.com/camunda/camunda/issues/22347
* https://github.com/camunda/camunda/issues/22351

### Acceptance Criteria

[ ] The `optimize` POM inherits from `parent/pom.xml`
[ ] Make sure that the Optimize distribution includes the dependencies versions defined in the Zeebe parent
[ ] `dist/` is the new distribution directory, instead of `optimize-distro`. Update: We keep `optimize-distro` for this release and will remove it with https://github.com/camunda/camunda/issues/25621
[ ] It is possible to launch Optimize with the rest of the stack
[ ] It is possible to launch Optimize as a standalone application
[ ] It's possible to launch the rest of the stack without Optimize
[ ] All of the TODOs in Optimize that were waiting for the completion of this task are resolved

### Links

* Part of https://github.com/camunda/product-hub/issues/2289

```[tasklist]
### Pull Requests
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [ ] The change is implemented as described on all target environments/versions/modes
- [ ] The documentation changes are as expected

### Completion

- [ ] All Review stages are successfully completed
- [ ] All associated PRs are merged to the main branch(es) and stable branches
- [ ] The correct version labels are applied to the issue

PR #24012, opened 2024-10-24
build: fix checkstyle errors in Optimize

### LINKED ISSUE #22267: Fix checkstyle errors in compliance with the rules of the monorepo
### Context

Currently, Optimize is present within the monorepo but it's not inheriting from the zeebe parent POM (i.e., `parent/pom.xml`). Soon, this change will need to happen and one of its implications is that Optimize will need to comply with the code style rules configured in such parent POM.

### Acceptance Criteria

[ ] with Optimize using the same configuration as the other modules, the command `mvn -f optimize/pom.xml -q checkstyle:check` does not fail.

### Testing Notes

* The parent change to `parent/pom.xml` is not in the scope of this task. Nevertheless, such temporary change could be a quick way to use the checkstyle command with the appropriate configuration.

### Links

* part of https://github.com/camunda/product-hub/issues/2289

```[tasklist]
### Pull Requests
```

## Task Lifecycle

#### Engineering Review

#### QA Review

### Completion

buccarel:
## Description

The context for this task is defined in https://github.com/camunda/camunda/issues/22267

With this PR:

* since Optimize is not yet inheriting from zeebe-parent, in order to perform the checkstyle validation, a new pipeline is being introduced, named `Optimize Checkstyle`, that temporarily inherits from zeebe-parent, before running `mvn checkstyle:check`. The inheritance only happens in the runner and does not land into the actual repo. Once we inherit from zeebe-parent, we can remove this extra pipeline.

* the inner classes Fields are many, and they have many consumers. Even though they have checkstyle errors, I decided to suppress the errors for those, to save time. Please let me know if we want this to be addressed as well.

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/22267


.github/workflows/optimize-checkstyle-check.yml
```diff
@@ -0,0 +1,70 @@
+# TODO: Remove this pipeline once Optimize inherits from zeebe-parent,
```
RomanJRW, line 1: what do you think about having this as part of the acceptance criteria of https://github.com/camunda/camunda/issues/22271?

optimize/backend/src/it/java/io/camunda/optimize/test/it/extension/db/DatabaseTestService.java
```diff
@@ -62,7 +62,7 @@
 public abstract class DatabaseTestService {
 
   protected static final ObjectMapper OBJECT_MAPPER = createObjectMapper();
-  private static final Logger log = org.slf4j.LoggerFactory.getLogger(DatabaseTestService.class);
+  private static final Logger LOG = org.slf4j.LoggerFactory.getLogger(DatabaseTestService.class);
```
RomanJRW, line 65: NABD but why do we have the fully qualified method here rather than an import?

RomanJRW (approved):
I am going to approve this because the changes are largely just formatting (I have not reviewed every single file here). A couple of small comments, but I think passing CI will be the biggest indicator of health here

.github/workflows/optimize-checkstyle-check.yml
buccarel: that's a very good idea.
I'll add it. We have more than one TODO like this.
Thanks

optimize/backend/src/it/java/io/camunda/optimize/test/it/extension/db/DatabaseTestService.java
buccarel: I don't think there's a reason. Maybe it got generated by some IDE, this way.
I'm not changing it here simply because that would trigger all of the pipelines again, but possibly I can sneak the change in, in the next PR

RomanJRW: I think it's NABD, and the same comment probably applies to many more files. If it's easy, then feel free to do it. Equally, feel free to ignore it 👍

PR #24015, opened 2024-10-24
Check permissions on user creation

### LINKED ISSUE #22595: Verify permissions in `UserCreateProcessor`
<h2>Description</h2>

- **Resource type**: `USER`
- **Possible permissions**:
    -  `*`

remcowesterhoud:
## Description

Adds permission checks to the `UserCreateProcessor` and adds an IT to verify it works as expected.

For this I had to make a slight modification to the `AuthCheckBehavior`. We need to exempt command written by Zeebe from the auth checks. If we don't we wouldn't be permitted to create the default user, as no authorization would be set.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22595

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
@@ -57,6 +57,11 @@ public final class AuthorizationCheckBehavior {
       return true;
     }
 
+    if (!request.getCommand().hasRequestMetadata()) {
+      // The command is written by Zeebe internally. Internal Zeebe commands are always authorized
+      return true;
+    }
+
     final var userKey =
         (Long) request.getCommand().getAuthorizations().get(Authorization.AUTHORIZED_USER_KEY);
     if (userKey == null) {
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/EngineProcessors.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/EngineProcessors.java
@@ -227,7 +227,8 @@ public final class EngineProcessors {
         processingState,
         writers,
         commandDistributionBehavior,
-        config);
+        config,
+        authCheckBehavior);
 
     ClockProcessors.addClockProcessors(
         typedRecordProcessors,
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
@@ -8,6 +8,8 @@
 package io.camunda.zeebe.engine.processing.user;
 
 import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior;
+import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.AuthorizationRequest;
 import io.camunda.zeebe.engine.processing.streamprocessor.DistributedTypedRecordProcessor;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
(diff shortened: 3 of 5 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -57,6 +57,11 @@ public boolean isAuthorized(final AuthorizationRequest request) {
       return true;
     }
 
+    if (!request.getCommand().hasRequestMetadata()) {
+      // The command is written by Zeebe internally. Internal Zeebe commands are always authorized
+      return true;
+    }
```
remcowesterhoud, line 63: @korthout I'm 99% sure this is safe. I need to identify whether a command is written by Zeebe, or by a user. Checking the request mestadata seems safe to me. Do you know any reasons cases where this wouldn't be true?

korthout: At this time, all commands that have request metadata are user commands. However, we did once discuss the idea of responding to inter-partition commands rather than sending an ACK command. There may be reasons in the future to change this.

We could consider using the `sourceRecordPosition` for this. Only commands that are the result of processing have a source record position set. But there are some exceptions like commands written by checkers that in addition to user commands also have no source record position. So, it may be hard to use.

Another option to consider is the RECORDS_BY_TYPE of [CommandApiRequestReader](https://github.com/camunda/camunda/blob/13d83cfd8b5d4fd2f5106ac1d54b963a8b031cee/zeebe/broker/src/main/java/io/camunda/zeebe/broker/transport/commandapi/CommandApiRequestReader.java#L53). This is filled with command types that can be sent through the broker's command api.

remcowesterhoud: > At this time, all commands that have request metadata are user commands. However, we did once discuss the idea of responding to inter-partition commands rather than sending an ACK command. There may be reasons in the future to change this.

I think that should be fine because we don't verify authorizations on distributed commands. I don't see issues there.

> We could consider using the sourceRecordPosition for this. Only commands that are the result of processing have a source record position set. But there are some exceptions like commands written by checkers that in addition to user commands also have no source record position. So, it may be hard to use.

Yes what you described is exactly why it's not possible to use as it is now.

> Another option to consider is the RECORDS_BY_TYPE of [CommandApiRequestReader](https://github.com/camunda/camunda/blob/13d83cfd8b5d4fd2f5106ac1d54b963a8b031cee/zeebe/broker/src/main/java/io/camunda/zeebe/broker/transport/commandapi/CommandApiRequestReader.java#L53). This is filled with command types that can be sent through the broker's command api.

I don't see how this would work. There we only store the value types, not the intents. So while we may say you need to check messages, we only want to do that for the `PUBLISH` intent, as the `EXPIRE` is internal. This becomes more complex with signals, because even if you say you need to check broadcasts, that's not entirely true as signals are broadcasted internally as well.

----------

My conclusion is I think the current approach is fine, even if we would change distirbution to this mechanism. If we later reach a state where we cannot use the request metadata here we can always change it.

korthout: Okay. One last idea to consider is a strict mapping of supported intents as user commands. But I'm not sure yet what that should look like. It's something you could consider exploring still.

remcowesterhoud: I feel like a mapping still doesn't solve the problem where we have a single intent used both internally and externally such as signal broadcasting. You'd always have to check for something else in combination.

nicpuppa (approved):

cmur2:
Hi, apparently attempting to merge this PR leads to build failures in the merge queue affecting all jobs and making the failure rate sky-rocketing today:

image

Since the failure happened on every attempt it seems to be reproducable, please don't try to merge again without changes. Any idea why that happens?

remcowesterhoud:
Seems like something changed on main that causes compilation failure. I'll rebase and fix

cmur2:
Thanks, it seems like rebasing to main did it and there was no deeper problem.

PR #24019, opened 2024-10-24
Exporter test framework

### LINKED ISSUE #23886: Test framework for Camunda exporter
**Description**
The camunda exporter should support both Elasticsearch(ELS) and Opensearch(OS), to avoid repetitive tests due to the similarity of OS and ELS clients write a [TestTemplateInvocationContextProvider](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.html) which will due all scaffolding work to support test templates.

EuroLew:
## Description

Part of a trio of PRs to implement a camunda exporter testing framework. PR 3/3

Create the `CamundaExporterITInvocationProvider` which provides a framework for running test templates against both elasticsearch(ELS) and opensearch(OS).

This also involves refactoring the `CamundaExporterIT` tests to use this framework.

Success Critieria:
- [x] Invocation provider which starts both ELS and OS and can verify the engine state. (PR 2/3)
- [x] Client adapter for both ELS and OS such that test templates can query documents and indices. (PR 1/3)
- [x] Refactor `CamundaExporterIT` so tests use the invocation provider.  

## Related issues

closes #23886

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
@@ -27,10 +27,10 @@ import io.camunda.exporter.schema.SchemaTestUtil;
 import io.camunda.exporter.utils.CamundaExporterITInvocationProvider;
 import io.camunda.exporter.utils.SearchClientAdapter;
 import io.camunda.exporter.utils.TestSupport;
+import io.camunda.security.entity.Permission;
 import io.camunda.webapps.schema.descriptors.IndexDescriptor;
 import io.camunda.webapps.schema.descriptors.IndexTemplateDescriptor;
 import io.camunda.webapps.schema.entities.usermanagement.AuthorizationEntity;
-import io.camunda.webapps.schema.entities.usermanagement.Permission;
 import io.camunda.webapps.schema.entities.usermanagement.UserEntity;
 import io.camunda.zeebe.exporter.api.ExporterException;
 import io.camunda.zeebe.exporter.api.context.Context;
@@ -349,7 +349,7 @@ final class CamundaExporterIT {
         .map(
             permissionValue ->
                 new Permission(
-                    permissionValue.getPermissionType().name(), permissionValue.getResourceIds()))
+                    permissionValue.getPermissionType(), permissionValue.getResourceIds()))
         .collect(Collectors.toList());
   }
 
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
@@ -8,7 +8,9 @@
 package io.camunda.exporter;
 
 import static io.camunda.exporter.schema.SchemaTestUtil.mappingsMatch;
+import static io.camunda.exporter.utils.CamundaExporterITInvocationProvider.*;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.CALLS_REAL_METHODS;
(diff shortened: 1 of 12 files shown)


mustafadagher (commented):
This LGTM (after we adopt the comments from the previous PRs).
I have just one observation about the last commit.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -349,7 +349,7 @@ private List<Permission> extractPermissions(final AuthorizationRecordValue recor
         .map(
             permissionValue ->
                 new Permission(
-                    permissionValue.getPermissionType().name(), permissionValue.getResourceIds()))
+                    permissionValue.getPermissionType(), permissionValue.getResourceIds()))
```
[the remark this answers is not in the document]
EuroLew: yes I think I messed up a rebase somewhere

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
```diff
@@ -51,13 +46,41 @@
...
   @BeforeEach
-  public void refresh() throws IOException {
+  public void refresh(final TestInfo testInfo) throws IOException {
+    if (testInfo.getDisplayName().contains("elasticsearch")) {
+      elsClient.indices().delete(req -> req.index("*"));
+      elsClient.indices().deleteIndexTemplate(req -> req.name("*"));
+    } else if (testInfo.getDisplayName().contains("opensearch")) {
+      osClient.indices().delete(req -> req.index(config.getIndex().getPrefix() + "*"));
+      osClient.indices().deleteIndexTemplate(req -> req.name("*"));
+    }
```
mustafadagher, line 81: 💭 This does the job indeed, good thinking! But I'm also afraid of relying on the test method display name, it feels like that's easily breakable.
I don't have a better idea right now, I'm just thinking out loud.. 

💭 Maybe the only other idea is to add the `connectionType` as a parameter and use `testInfo.getTestMethod() -> parameters` to make this decision

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
```diff
@@ -0,0 +1,168 @@
...
+
+  @Override
+  public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(
+      final ExtensionContext extensionContext) {
+
+    try {
+      elsClient.indices().delete(req -> req.index("*"));
+      elsClient.indices().deleteIndexTemplate(req -> req.name("*"));
+      osClient.indices().delete(req -> req.index("*"));
+      osClient.indices().deleteIndexTemplate(req -> req.name("*"));
```
mustafadagher, line 101: 🔧 this is not consistent with the change introduced in SchemaManagerIt

mustafadagher (commented):
Great job @EuroLew ! 🚀  Almost there!! please check my last comments

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
```diff
@@ -0,0 +1,170 @@
...
+  private final List<AutoCloseable> closeables = new ArrayList<>();
+
+  public CamundaExporterITInvocationProvider() {
+    final var osConfig = getConfigWithConnectionDetails(OPENSEARCH);
+    osClient = new OpensearchConnector(osConfig.getConnect()).createClient();
+    osClientAdapter = new SearchClientAdapter(osClient);
+
+    final var elsConfig = getConfigWithConnectionDetails(ELASTICSEARCH);
+    elsClient = new ElasticsearchConnector(elsConfig.getConnect()).createClient();
+    elsClientAdapter = new SearchClientAdapter(elsClient);
```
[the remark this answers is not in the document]
mustafadagher: nvm! seems like you've already fixed this

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
EuroLew: I don't like adding a parameter we don't use for tests, I'll just use named arguments insteada

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
EuroLew: there's not an easy way to know which testcase we are in unless we add a parameter that is not used for tests. This fundamentally accomplishes the same thing. For each test that modifies els we get 1 els wipe instead of 2.

mustafadagher: in general, I don't think the `InvocationContextProvider` is the right/best place to reset/refresh the clients. This is better suited in the `beforeEach`.

in the `beforeEach` (similarly to the `SchemaManagerIT`) you can access the `TestInfo`, which has knowledge in this case about `config` parameter. You can use that to know which connection type is configured, and clean the indices accordingly

EuroLew: I don't think it can be done in the `beforeEach` for the specific test clases as the `TestInfo` doesn't have the parameter instance, just information about it. 

I think it's fine to do it in the invocation providers `AfterEach` block

mustafadagher: will the `AfterEach` run if an exception was thrown from one method run?

mustafadagher (approved):
Thanks a lot Euro! Such a great job! 🚀 

❓ There's still an open discussion point if `reseting indices is safe at AfteEach`, but I'm approving already, we can decide to change it or leave it as is.

PR #24020, opened 2024-10-24
Camunda exporter invocation provider

EuroLew:
## Description

Part of a trio of PRs to implement a camunda exporter testing framework. PR 2/3

Invocation provider which tests can extend to run a test template against elasticsearch and opensearch.

Success Criteria:
- [x] Supports test templates running against and ELS and OS instance by providing configuration to connect to the ELS and OS containers.

## Related issues

relates to #23886

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.utils;
+
+import static io.camunda.exporter.config.ConnectionTypes.ELASTICSEARCH;
+import static io.camunda.exporter.config.ConnectionTypes.OPENSEARCH;
+import static java.util.Arrays.asList;
+
+import io.camunda.exporter.config.ConnectionTypes;
+import io.camunda.exporter.config.ExporterConfiguration;
+import io.camunda.search.connect.es.ElasticsearchConnector;
+import io.camunda.search.connect.os.OpensearchConnector;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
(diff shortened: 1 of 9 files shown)


mustafadagher:
should this PR be based on main? or on the branch from PR 1? there's repeated code between the two PRs?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
```diff
@@ -0,0 +1,161 @@
...
+
+  private ExporterConfiguration getConfigWithConnectionDetails(
+      final ConnectionTypes connectionType) {
+    final var config = new ExporterConfiguration();
+    config.getIndex().setPrefix(CONFIG_PREFIX);
+    config.getBulk().setSize(1); // force flushing on the first record
+    switch (connectionType) {
+      case ELASTICSEARCH -> config.getConnect().setUrl(elsContainer.getHttpHostAddress());
+      case OPENSEARCH -> config.getConnect().setUrl(osContainer.getHttpHostAddress());
+      default -> throw new IllegalArgumentException("Unknown connection type: " + connectionType);
```
mustafadagher, line 51: 🔧 connectionType is an enum, it can never hold an `unsupported` case. by adding a default, we are actually removing a compile time insurance that people won't add a new connection type without supporting configuring it.
So I suggest to remove the default here.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
```diff
@@ -0,0 +1,161 @@
...
+    final var elsClientAdapter = new SearchClientAdapter(elsClient);
+
+    try {
+      elsClient.indices().delete(req -> req.index("*"));
+      elsClient.indices().deleteIndexTemplate(req -> req.name("*"));
+      osClient.indices().delete(req -> req.index("*"));
+      osClient.indices().deleteIndexTemplate(req -> req.name("*"));
+    } catch (final Exception e) {
+      throw new RuntimeException("Failed to reset container data", e);
+    }
```
mustafadagher, line 97: 🔧 same as in previous PR, is there a way to only reset data for the client under test not both? if there's a simple way to check and do it, it would be great.

mustafadagher (commented):
Again, Great approach in general. 🚀  Just minor change requests.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
EuroLew: The clean up here occurs before providing the invocation context to a test template, so it works as you want e.g.
1. clean up ols and es
2. run els test case
3. run os test case
4. next test template 
5. clean up ols and es
6. etc

EuroLew:
Review comments implemented in #24019

PR #24021, opened 2024-10-24
Search client adapter

EuroLew:
## Description

Part of a trio of PRs to implement a camunda exporter testing framework. PR 1/3

Create a search client adapter such that common actions for tests which include 
- Verifying an index state
- Verifying an index template state
- verifying a policy
- verifying a document state

can be done using one interface.

Note, the implementation of the `SearchClientAdapter` copies from `SchemaTestUtil` but the code in `SchemaTestUtil` cannot be removed as the `CamundaExporterIT` tests rely on it.  The repeated code in `SchemaTestUtil` will be removed in PR 3/3 when the `CamundaExporterIT` is refactored to use the `SearchClientAdapter`

Success Criteria:
- [x] Create search client adapter for ELS and OS.
- [x] Refactor `SchemaManagerIT` to use the adapter to simplify testing code.

## Related issues

relates to #23886

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
@@ -7,11 +7,6 @@
  */
 package io.camunda.exporter.schema;
 
-import static io.camunda.exporter.schema.SchemaTestUtil.elsPolicyToNode;
-import static io.camunda.exporter.schema.SchemaTestUtil.getElsIndexAsNode;
-import static io.camunda.exporter.schema.SchemaTestUtil.getElsIndexTemplateAsNode;
-import static io.camunda.exporter.schema.SchemaTestUtil.getOpensearchIndexAsNode;
-import static io.camunda.exporter.schema.SchemaTestUtil.getOpensearchIndexTemplateAsNode;
 import static io.camunda.exporter.schema.SchemaTestUtil.mappingsMatch;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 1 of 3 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/SearchClientAdapter.java
```diff
@@ -0,0 +1,165 @@
...
+    final var policyAsMap =
+        SchemaResourceSerializer.serialize(
+            co.elastic.clients.json.jackson.JacksonJsonpGenerator::new,
+            (gen) -> lifecyclePolicy.serialize(gen, ELS_JSON_MAPPER));
+
+    return MAPPER.valueToTree(policyAsMap);
+  }
+
+  public JsonNode getIndexAsNode(final String indexName) throws IOException {
+    switch (elsClient != null ? "elsClient" : "osClient") {
```
mustafadagher, line 94: ❌ in this switch and the ones below: are we ever going to the `default` case?
we never check if `osClient` is null, so if `elsClient` is null, we assume that the `osClient` is not. And we will never reach the default even if we have no client instantiated.

I suggest we use a simple `if, else`, as in
```java
    if (elsClient != null) {
      // DO something with elsClient
    } else if (osClient != null) {
      // DO something with osClient
    }
    throw new IllegalStateException("Must instantiate client in SearchClientAdapter");

File: zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java [hunk 2]
```diff
@@ -51,13 +45,38 @@
...
+    config.getConnect().setUrl(OPENSEARCH_CONTAINER.getHttpHostAddress());
+    osClient = new OpensearchConnector(config.getConnect()).createClient();
+  }
+
   @BeforeEach
   public void refresh() throws IOException {
+    elsClient.indices().delete(req -> req.index("*"));
+    elsClient.indices().deleteIndexTemplate(req -> req.name("*"));
+    osClient.indices().delete(req -> req.index(config.getIndex().getPrefix() + "*"));
+    osClient.indices().deleteIndexTemplate(req -> req.name("*"));
```
mustafadagher, line 77: 🔧 This will lead to double the time in the config phase of each test method. could we think of a way to only execute the clean-up for the client currently used for test? That would be nice to have if possible

mustafadagher (changes requested):
Great job @EuroLew! I like it a lot. 🚀 
I just left two comments with enhancements, one if more important than the other.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/SearchClientAdapter.java
mustafadagher: you can also `requireNonNull` for the injected client in both constructors, and remove the `throw new IllegalStateException` from all the methods

EuroLew:
Review comments implemented in #24019

Issue #16276, opened 2024-02-08
[ZPA] Improve operational quality of services 


korthout:
## Introduction

In an effort to improve the operations quality of product teams, opportunities for growth were raised during the [Platform Operations Quality Assessment](https://docs.google.com/spreadsheets/d/1SFt0O6u0cqt6_9izfRlJdCPF_dCYGe_AeNOTj9vxeow/edit#gid=1885598468) of the Zeebe Process Automation team (ZPA). This issue provides an overview of these issues.

## Task Breakdown

This is a work in progress. At this time, it's just a copy of the results from the assessment. We'll need to transform this into actionable tasks and possibly break them up further.

```[tasklist]
### Tasks
- [ ] Several resources that exist in the Zeebe repo, are covering parts of the ZDP, not the ZPA team, which need to be clarified or adjusted.
- [ ] We should share more knowledge about how to add metrics and how to consume them, build dashboards, etc.
- [ ] We should discuss more and try to understand what is bad for our users (like when it takes long to trigger a timer), and based on this write metrics and alerts.
- [ ] It was raised that we currently have two teams sharing one medic, which makes it sometimes hard for the people in charge (especially new joiners). Might be worth investigating whether this can be split off.
- [ ] Changes to incident handbooks or processes, seem to be sometimes missed or not communicated well enough, we should improve that.
- [ ] We should think about whether we need more configurations that we can change on runtime, to avoid downtime and make it more flexible for the users.
- [ ] We need clarity on dependency update PRs that are not auto-mergeable.
- [ ] Get better hands-on experience with the service we provide to understand whether resources need to be adjusted (might be an opportunity to save costs or even improve availability).
- [ ] A common pattern/strategy or format on how to write good user-facing error messages. If they follow always the same pattern they provide a better UX and are also better to understand, etc.
```

### Out of scope

None yet.

## Discussions and Decisions

...

## Links

- Product Hub (internal): Not tracked by product hub
- [Operations Quality Assessment](https://docs.google.com/spreadsheets/d/1SFt0O6u0cqt6_9izfRlJdCPF_dCYGe_AeNOTj9vxeow/edit#gid=1885598468)

korthout:
- https://github.com/camunda/zeebe/issues/16579

megglos:
> A common pattern/strategy or format on how to write good user-facing error messages. If they follow always the same pattern they provide a better UX and are also better to understand, etc.

[...]

PR #24033, opened 2024-10-24
fix: write response for COMPLETE user task command with correct request metadata

### LINKED ISSUE #23925: Respond to `COMPLETE` user task command after task listener execution
## Description:
Currently, when `client.newUserTaskCompleteCommand(userTaskKey).send()` is invoked, it results in a timeout error, but at the same time user task completes successfully. The issue occurs because when finalizing the `COMPLETE` user task command after all task listener jobs are completed, the engine uses the `requestId` and `requestStreamId` from the `COMPLETE_TASK_LISTENER` command instead of the original `COMPLETE` command.

To resolve this issue, we need to:
1. Persist the `requestId` and `requestStreamId` from the original `COMPLETE` user task command.
2. Use these fields after task listener execution to finalize the original `COMPLETE` user task command.
3. Ensure that the client receives response after all task listeners are processed and user task completed.

This will allow proper tracking of user task completion ensure the client receives successfull response when all task listeners are completed.

ce-dmelnych:
## Description
When the `client.newUserTaskCompleteCommand(userTaskKey).send().join()` method is invoked, it results in a timeout error despite the user task completing successfully. This occurs because the engine uses the `requestId` and `requestStreamId` from the `COMPLETE_TASK_LISTENER` command instead of the original `COMPLETE` command when finalizing the user task.

### Fix:
This PR implements the following changes:
1. Persists the `requestId` and `requestStreamId` from the original `COMPLETE` user task command from `UserTaskProcessor`:
> Note: Generally, persistence logic should be handled via `*Applier` classes. However, since this involves
> request-related data - in the case of command reconstruction, new request values will be produced anyway,
> so the invocation of persistence logic for request metadata from `UserTaskProcessor` is acceptable.
> This approach has precedent in `ProcessInstanceCreationCreateWithResultProcessor` also.

2. Uses the original `COMPLETE` command request metadata after all task listener jobs are completed to send the correct response.
   This ensures that the request metadata is available to finalize the original command after processing
   all task listeners, allowing the engine to respond correctly to the request.

These changes prevent the timeout error and ensure that the client receives the appropriate success response when all task listeners have finished executing and the user task is completed.

## Checklist

- [ ] for CI changes:

## Related issues

closes #23925

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/UserTaskProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/UserTaskProcessor.java
@@ -26,9 +26,10 @@ import io.camunda.zeebe.engine.state.immutable.ElementInstanceState;
 import io.camunda.zeebe.engine.state.immutable.EventScopeInstanceState;
 import io.camunda.zeebe.engine.state.immutable.ProcessState;
 import io.camunda.zeebe.engine.state.immutable.ProcessingState;
-import io.camunda.zeebe.engine.state.immutable.UserTaskState;
 import io.camunda.zeebe.engine.state.immutable.UserTaskState.LifecycleState;
 import io.camunda.zeebe.engine.state.instance.ElementInstance;
+import io.camunda.zeebe.engine.state.instance.UserTaskRecordRequestMetadata;
+import io.camunda.zeebe.engine.state.mutable.MutableUserTaskState;
 import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeTaskListenerEventType;
 import io.camunda.zeebe.protocol.impl.record.value.usertask.UserTaskRecord;
 import io.camunda.zeebe.protocol.record.RejectionType;
@@ -41,7 +42,7 @@ public class UserTaskProcessor implements TypedRecordProcessor<UserTaskRecord> {
 
   private final UserTaskCommandProcessors commandProcessors;
   private final ProcessState processState;
-  private final UserTaskState userTaskState;
+  private final MutableUserTaskState userTaskState;
   private final ElementInstanceState elementInstanceState;
   private final EventScopeInstanceState eventScopeInstanceState;
 
@@ -55,6 +56,7 @@ public class UserTaskProcessor implements TypedRecordProcessor<UserTaskRecord> {
 
   public UserTaskProcessor(
       final ProcessingState state,
+      final MutableUserTaskState userTaskState,
       final KeyGenerator keyGenerator,
       final BpmnBehaviors bpmnBehaviors,
       final Writers writers,
@@ -63,7 +65,7 @@ public class UserTaskProcessor implements TypedRecordProcessor<UserTaskRecord> {
         new UserTaskCommandProcessors(
             state, keyGenerator, bpmnBehaviors, writers, authCheckBehavior);
     this.processState = state.getProcessState();
-    this.userTaskState = state.getUserTaskState();
+    this.userTaskState = userTaskState;
     this.elementInstanceState = state.getElementInstanceState();
     this.eventScopeInstanceState = state.getEventScopeInstanceState();
 
@@ -127,6 +129,20 @@ public class UserTaskProcessor implements TypedRecordProcessor<UserTaskRecord> {
     final var eventType = mapIntentToEventType(intent);
 
     if (userTaskElement.hasTaskListeners(eventType)) {
+      /*
+       * Store the original command's metadata (requestId, requestStreamId) before creating
+       * the first task listener job. This metadata will be used later to finalize the original
+       * command after all task listeners have been processed, ensuring that the engine can respond
+       * appropriately to the original command request.
+       *
+       * Note:
+       * Typically, persistence logic should be handled via `*Applier` classes. However, since
+       * this involves request-related data, so in the case of command reconstruction,
+       * we will have new request values anyway, so persisting these data here is acceptable.
+       * A similar approach has been used in `ProcessInstanceCreationCreateWithResultProcessor`.
+       */
+      storeUserTaskRecordRequestMetadata(command);
+
       final var listener = userTaskElement.getTaskListeners(eventType).getFirst();
       final var userTaskElementInstance = getUserTaskElementInstance(persistedRecord);
       final var context = buildContext(userTaskElementInstance);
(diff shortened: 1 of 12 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/UserTaskProcessor.java
```diff
@@ -127,6 +129,20 @@ private void handleCommandProcessing(
...
+       * command after all task listeners have been processed, ensuring that the engine can respond
+       * appropriately to the original command request.
+       *
+       * Note:
+       * Typically, persistence logic should be handled via `*Applier` classes. However, since
+       * this involves request-related data, so in the case of command reconstruction,
+       * we will have new request values anyway, so persisting these data here is acceptable.
+       * A similar approach has been used in `ProcessInstanceCreationCreateWithResultProcessor`.
+       */
+      storeUserTaskRecordRequestMetadata(command);
```
korthout, line 144: 👍 I like that you store it here. This ensures it's stored for all user task commands, rather than just for the complete command.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskCompleteProcessor.java
```diff
@@ -80,10 +84,29 @@ public void onFinalizeCommand(
...
+            responseWriter.writeResponse(
+                userTaskKey,
+                UserTaskIntent.COMPLETED,
+                userTaskRecord,
+                ValueType.USER_TASK,
+                metadata.getRequestId(),
+                metadata.getRequestStreamId()),
+        () ->
+            responseWriter.writeEventOnCommand(
+                userTaskKey, UserTaskIntent.COMPLETED, userTaskRecord, command));
```
korthout, line 109: 🔧 It's probably never the case, but we could consider the case where the command doesn't have request metadata and it also wasn't stored in the state.

```suggestion
    if (command.requestId != null && command.requestStreamId() != null) {
      responseWriter.writeEventOnCommand(
          userTaskKey, UserTaskIntent.COMPLETED, userTaskRecord, command));
    } else {
      // If the record request metadata is present, it indicates that "complete" task listeners were
      // configured, and that the normal flow of the "COMPLETE" command was interrupted by the
      // "COMPLETE_TASK_LISTENER" command. In this case, we need to use the `requestId` and
      // `requestStreamId` from the persisted metadata, which refers to the original "COMPLETE"
      // command to write the response back.
      recordRequestMetadata.ifPresent(
        metadata ->
            responseWriter.writeResponse(
                userTaskKey,
                UserTaskIntent.COMPLETED,
                userTaskRecord,
                ValueType.USER_TASK,
                metadata.getRequestId(),
                metadata.getRequestStreamId()));
    }
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/immutable/UserTaskState.java
```diff
@@ -21,6 +23,8 @@ public interface UserTaskState {
 
   UserTaskIntermediateStateValue getIntermediateState(final long userTaskKey);
 
+  Optional<UserTaskRecordRequestMetadata> getRecordRequestMetadata(final long userTaskKey);
```
korthout, line 26: 🔧 As this returns an optional, we could consider naming this `findRecordRequestMetadata` to clarify that it could optionally return a value or not.

korthout (approved):
Thanks @ce-dmelnych 🚀 

👍 I like the changes. 

💭 The commit that stores the metadata from the processor rather than an event should IMO describe why this is acceptable. I know it's currently described in a code comment which is also appropriate, but it could get out-of-date. This is not the case with commit messages as they describe the changes in the context of that moment in time rather than the code itself.

🔧 Please have a look at my comments, I've made a few suggestions.

ana-vinogradova-camunda (approved):
LGTM 🚀 
Very clean and easy to understand changes, thanks @ce-dmelnych!

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskCompleteProcessor.java
ce-dmelnych: Thanks for the suggestion, @korthout! 🚀

I found that we can use the `command.hasRequestMetadata()` method to check if the current command contains request metadata. So, I’ve improved the previous implementation to only query request metadata from `UserTaskState` if it’s not present in the current command.

❓ Would it make sense to add `warn` logging when request metadata is missing in both: the current command and `UserTaskState`?
I noticed that logging isn’t widely used across command processors, but adding it here could improve visibility in such "unexpected" cases. I’d be glad to hear your thoughts on this!

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/immutable/UserTaskState.java
ce-dmelnych: Yeah, I've changed the method to return `Optional` just before committing the first commit 🙈.
Thanks Nico for highlighting this! I’ve renamed it to `findRecordRequestMetadata` 👍.

ce-dmelnych:
> 💭 The commit that stores the metadata from the processor rather than an event should IMO describe why this is acceptable. I know it's currently described in a code comment which is also appropriate, but it could get out-of-date. This is not the case with commit messages as they describe the changes in the context of that moment in time rather than the code itself.

Thank you, @korthout, for the thoughtful review and recommendations. I’ve updated the body of the first commit to explain why it’s acceptable in a such case to store request metadata directly from `UserTaskProcessor`.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskCompleteProcessor.java
korthout: You're right that logging isn't widely used, but this is something [we'd like to improve](https://github.com/camunda/camunda/issues/16276). 

However, too many logs is also bad, so we need to consider whether they are useful. The way I think about it is: would it be helpful in this context when that case were to occur? Here, if we don't have any request info available, we can't respond. But that's expected I guess. Perhaps I'm overlooking something though. Curious to your thought 🙇

ce-dmelnych: 🤔 Hmm… what I imagine is that having this log (e.g., `Request metadata was not present in the current command nor in user task state for user task with key {}. The client will not receive a response.`) could help clarify cases where the client didn’t receive a response - not due to a timeout issue (such as a long-running listener exceeded the Tasklist timeout value of 10-15 seconds (or whatever is configured)) but specifically because request metadata was unexpectedly missing.

Additionally, since this is a "_should not happen_" case, we normally wouldn’t expect to see it in the logs. So, its presence could serve as an indicator that something went wrong.

Or maybe I’m just overthinking it! 😄

ce-dmelnych: Anyway, we can always add the logging later if it turns out to be helpful!
For now, I’m merging it as is 😉

PR #24037, opened 2024-10-24
feat: rework tree path

### LINKED ISSUE #24061: When exporting the incident populate the tree path fields
### Description

Within #16915 we started receiving in exported Zeebe records 3 new fields, that describe position of the incident and corresponding process and flow node instances in two hierarchies: "inter tree path" when parent process has call activity and child process has an incident and "intra tree path" when process instance scope contain some nested scopes (subprocess, multi-tenancy) and incident is fired in the nested scope.
We need to work through these fields and convert them to `treePath` values for different Operate entities so that all related Operate features continue working.

### TODO

- [ ] fill in "inter" `treePath` in process instance in `list-view` index: #24037
- [ ] fill in "inter" `treePath` in `incident` entity: #24511
- [ ] fill in "intra" `treePath` in flow node instances in `list-view` and `flow-node-instance` indices #24527 

### Additional information

* Related Zeebe changes with description of the new incident fields: https://github.com/camunda/camunda/issues/16915 and https://github.com/camunda/camunda/pull/20293
* document with description of Operate fields: https://docs.google.com/document/d/1Nh8Jln5_aWIS1O5NnC-SGRblijizDTSujG7zIfc-GOY/edit

### LINKED ISSUE #24063: Deal with changed format of `flowNodeId` in `treePath`
### Description

Before 8.7 `treePath` for process instances and incidents looked like this:
```
PI_100/FN_call1/FNI_101/PI_102/FN_call2/FNI_103/PI_104/FN_servicetask/FNI_105
```
where the value prefixed with `FN_` is the `elementId` (or `flowNodeId`) of the corresponding flow node (can be either call activity or any node with an incident).

[Now](https://github.com/camunda/camunda/pull/20293) we get the field `callingElementPath` in exported incident record which contains the encoded values of `elementId`. "The Element Ids are encoded by lexicographically sorting all calling activity elements in one deployment resource, and using the index of each call activity Id in this sorted list as the respective encoded value."

To resolve `flowNodeId` from encoded value, we need to store call activity ids in process cache and use it in corresponding export handlers.

### TODO

* add list of call activity ids to process cache: they must be sorted lexicographically for ease of use
* resolve element id from process cache in export handlers which store tree paths for incident and process instance

sdorokhova:
* add list of `callActivityIds` to `ProcessCache`:
  * we don't store `callActivityIds` in `process` index
  * instead we read them from BPMN XML when loading data in cache
* build tree path for process instance on every created incident. Tree path makes use of `callActivityIds` from `ProcessCache`

related to #24061 and #24063


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
```diff
@@ -0,0 +1,108 @@
...
+      }
+      if (keysWithinOnePI.get(0).equals(processInstanceKey)) {
+        break;
+      }
+      treePath
+          .appendFlowNode(String.valueOf(callingElementPath.get(i)))
+          .appendFlowNodeInstance(String.valueOf(keysWithinOnePI.get(1)));
+    }
+    entity.setTreePath(treePath.toString());
+  }
```
Zelldon: 🔧 Can we explain this logic a bit with a comment? Like draw the general format and logic to build this, so we don't need to look up extra places to understand this part of the code. This would quite beneficial

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
```diff
@@ -0,0 +1,108 @@
...
+        treePath.startTreePath(keysWithinOnePI.get(0));
+      } else {
+        treePath.appendProcessInstance(keysWithinOnePI.get(0));
+      }
+      if (keysWithinOnePI.get(0).equals(processInstanceKey)) {
+        break;
+      }
+      treePath
+          .appendFlowNode(String.valueOf(callingElementPath.get(i)))
+          .appendFlowNodeInstance(String.valueOf(keysWithinOnePI.get(1)));
```
Zelldon: ❓ Has this always two instances?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
```diff
@@ -257,50 +240,6 @@ private void incrementFinishedCount() {
     */
   }
 
-  /// TODO - because it depends on listViewStore and flowNodeStore
-  private String getTreePathForCalledProcess(final ProcessInstanceRecordValue recordValue) {
-    // DEFAULT for now
```
Zelldon, line 262: 💪🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/IncidentHandlerTest.java
```diff
@@ -52,7 +52,7 @@ void testGetEntityType() {
   void shouldHandleRecord() {
     // given
     final Record<IncidentRecordValue> incidentRecord =
-        factory.generateRecord(ValueType.INCIDENT, r -> r.withIntent(ProcessIntent.CREATED));
```
Zelldon, line 55: 🤔 Wait why this didn't fail before?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
```diff
@@ -130,24 +130,7 @@ public void updateEntity(
...
-    piEntity.setTreePath(treePath);
-    //
-    //    if (piEntity.getTreePath() == null) {
-    //      final String treePath =
-    //          new TreePath()
-    //
-    // .startTreePath(ConversionUtils.toStringOrNull(recordValue.getProcessInstanceKey()))
-    //              .toString();
-    //      piEntity.setTreePath(treePath);
-    //    }
```
Zelldon, line 150: @sdorokhova what I thought we could do is to always set a basic treePath, when we are on a flat PI

Like we have PI and flow scope and key. I guess we could at least set for the ones where flow scope is either -1 or same as PI key? 🤔

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandlerTest.java
```diff
@@ -0,0 +1,175 @@
...
+        .isEqualTo("PI_111/FN_3/FNI_123/PI_222");
+    underTest.updateEntity(incidentRecord, processInstanceForListViewEntity2);
+
+    // when called process 3rd level
+    final ProcessInstanceForListViewEntity processInstanceForListViewEntity3 =
+        new ProcessInstanceForListViewEntity().setId(String.valueOf(pi3Key));
+    underTest.updateEntity(incidentRecord, processInstanceForListViewEntity3);
+    // then
+    assertThat(processInstanceForListViewEntity3.getTreePath())
+        .isEqualTo("PI_111/FN_3/FNI_123/PI_222/FN_6/FNI_234/PI_333");
```
Zelldon: 🤔 didn't we wanted to somehow extract the call activity element id via these index from the bpmn?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/ElasticSearchProcessCacheLoader.java
```diff
@@ -21,10 +25,13 @@ public class ElasticSearchProcessCacheLoader implements CacheLoader<Long, Cached
   private final ElasticsearchClient client;
   private final String processIndexName;
 
+  private final XMLUtil xmlUtil;
+
   public ElasticSearchProcessCacheLoader(
-      final ElasticsearchClient client, final String processIndexName) {
+      final ElasticsearchClient client, final String processIndexName, final XMLUtil xmlUtil) {
```
Zelldon, line 31: 🔧 Could we either introduce a functional interface or use a BiFunction to not depend on the XMlUtil directly?

Like define a interact where you pass in the XML bytes and the processId and get back the List of Call activities ? 

I think this makes it easier to test, and reduces unnecessary dependencies.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/XMLUtil.java
```diff
@@ -80,6 +81,15 @@ public Optional<ProcessEntity> extractDiagramData(
...
                       .add(new ProcessFlowNodeEntity(x.getId(), x.getName())));
+      // collect call activity ids
+      final Collection<CallActivity> callActivities =
+          modelInstance.getModelElementsByType(CallActivity.class);
+      processEntity.setCallActivityIds(
+          callActivities.stream()
+              .map(CallActivity::getId)
+              .filter(id -> processChildrenIds.contains(id))
+              .sorted()
+              .toList());
```
Zelldon, line 92: ❓ Do we have a separate UT for this class? If not we should add some.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
```diff
@@ -66,25 +72,42 @@ public ProcessInstanceForListViewEntity createNewEntity(final String id) {
...
+    final List<List<Long>> elementInstancePath = value.getElementInstancePath();
+    final List<Integer> callingElementPath = value.getCallingElementPath();
+    final List<Long> processDefinitionPath = value.getProcessDefinitionPath();
 
     final Long processInstanceKey = Long.valueOf(entity.getId());
 
+    // example of how the tree path is build when current instance is on the third level of calling
+    // hierarchy:
+    // PI_<parentProcessInstanceKey>/FN_<parentCallActivityId>/FNI_<parentCallActivityInstanceKey>/
+    // PI_<secondLevelProcessInstanceKey>/FN_<secondLevelCallActivityId>/FNI_<secondLevelCallActivityInstanceKey>PI_<currentProcessInstanceKey>
```
Zelldon, line 85: ❓ Is the flow scope key (of the current scope) also then part of the treePath at the end? Or only the PI?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
```diff
@@ -66,25 +72,42 @@ public ProcessInstanceForListViewEntity createNewEntity(final String id) {
...
     for (int i = 0; i < elementInstancePath.size(); i++) {
       final List<Long> keysWithinOnePI = elementInstancePath.get(i);
-      if (treePath.isEmpty()) {
-        treePath.startTreePath(keysWithinOnePI.get(0));
-      } else {
-        treePath.appendProcessInstance(keysWithinOnePI.get(0));
-      }
+      treePath.appendProcessInstance(keysWithinOnePI.get(0));
       if (keysWithinOnePI.get(0).equals(processInstanceKey)) {
+        // we stop building the tree path when we reach current processInstanceKey
```
Zelldon, line 91: 🔧 We reached the leaf of the TreePath

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
```diff
@@ -66,25 +72,42 @@ public ProcessInstanceForListViewEntity createNewEntity(final String id) {
...
       }
-      treePath
-          .appendFlowNode(String.valueOf(callingElementPath.get(i)))
-          .appendFlowNodeInstance(String.valueOf(keysWithinOnePI.get(1)));
+      // get call activity id from processCache
+      final Optional<CachedProcessEntity> cachedProcess =
+          processCache.get(processDefinitionPath.get(i));
+      if (cachedProcess.isPresent()
+          && cachedProcess.get().callElementIds() != null
+          && callingElementPath.get(i) != null) {
```
Zelldon, line 99: 🔧 Lets do the gets once

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
```diff
@@ -66,25 +72,42 @@ public ProcessInstanceForListViewEntity createNewEntity(final String id) {
...
+        treePath.appendFlowNode(
+            cachedProcess.get().callElementIds().get(callingElementPath.get(i)));
+      } else {
+        LOGGER.debug(
+            "No process found in cache. TreePath won't contain proper callActivityId. processInstanceKey: {}, processDefinitionKey: {}, incidentKey: {}",
+            processInstanceKey,
+            processDefinitionPath.get(i),
+            record.getKey());
+        treePath.appendFlowNode(String.valueOf(callingElementPath.get(i)));
```
Zelldon, line 108: ❓ This is the ordinal I guess?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandlerTest.java
```diff
@@ -25,14 +28,17 @@
...
 import java.util.Map;
+import java.util.Optional;
 import org.junit.jupiter.api.Test;
 
 public class ListViewProcessInstanceFromIncidentHandlerTest {
 
   private final ProtocolFactory factory = new ProtocolFactory();
   private final String indexName = "test-list-view";
+
+  private final ProcessCache processCache = mock(ProcessCache.class);
```
Zelldon, line 39: ❓ Could we also use a simple fake implementation, based on map? Might make the test easier, as we don't need to mock and write what the mock needs to do, etc. (lots of coupling with the implementation otherwise)

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandlerTest.java
```diff
@@ -170,6 +198,21 @@ public void shouldUpdateEntityFromRecord() {
     underTest.updateEntity(incidentRecord, processInstanceForListViewEntity3);
     // then
     assertThat(processInstanceForListViewEntity3.getTreePath())
-        .isEqualTo("PI_111/FN_3/FNI_123/PI_222/FN_6/FNI_234/PI_333");
+        .isEqualTo("PI_111/FN_callActivity1/FNI_123/PI_222/FN_callActivity2/FNI_234/PI_333");
+  }
+
+  @Test
+  public void shouldUpdateEntityFromRecordWithoutCallActivity() {
```
Zelldon, line 205: 👍🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandlerTest.java
```diff
@@ -170,6 +198,21 @@ public void shouldUpdateEntityFromRecord() {
...
+    final Long pi3Key = 333L;
+    final Record<IncidentRecordValue> incidentRecord =
+        createIncidentRecord(List.of(List.of(pi3Key, 345L)), List.of(), List.of(777L));
+
+    // when
+    final ProcessInstanceForListViewEntity processInstanceForListViewEntity =
+        new ProcessInstanceForListViewEntity().setId(String.valueOf(pi3Key));
+    underTest.updateEntity(incidentRecord, processInstanceForListViewEntity);
+    // then
+    assertThat(processInstanceForListViewEntity.getTreePath()).isEqualTo("PI_333");
```
Zelldon, line 216: 🔧 This we could have always, even without incidents right?

ChrisKujawa (approved):
Great stuff! Thanks @sdorokhova I have smaller comments, we might want to follow-up, but nothing blocking I think.

I guess makes sense that @mustafadagher also takes a look at he built the other side, from Zeebe.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/IncidentHandlerTest.java
mustafadagher: because `IncidentHandler.handlesRecord()` uses a string comparison not Intent enum comparison.
`return !intentStr.equals(IncidentIntent.RESOLVED.toString());`

I've changed this to compare enums on every handler I've moved, but not everyone does that.

🔧 @sdorokhova can we change that in handler?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandlerTest.java
mustafadagher: it's done in another PR. 
https://github.com/camunda/camunda/pull/24511

qa/integration-tests/src/test/resources/process/single-task.bpmn
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<bpmn:definitions xmlns:bpmn="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI" xmlns:di="http://www.omg.org/spec/DD/20100524/DI" xmlns:dc="http://www.omg.org/spec/DD/20100524/DC" xmlns:zeebe="http://camunda.org/schema/zeebe/1.0" id="Definitions_1771k9d" targetNamespace="http://bpmn.io/schema/bpmn" exporter="Zeebe Modeler" exporterVersion="0.5.0">
+  <bpmn:process id="process" isExecutable="true">
+    <bpmn:startEvent id="StartEvent_1">
+      <bpmn:outgoing>SequenceFlow_04ev4jl</bpmn:outgoing>
+    </bpmn:startEvent>
```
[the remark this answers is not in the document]
sdorokhova: Accidentally pushed. Will delete it.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
sdorokhova: For process instance entity no. For incident entity it is, but this is addressed in another Handler and PR.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
sdorokhova: Yes. We discussed with @mustafadagher , that here I will:
* increase the log level till warning
* will leave the ordinal instead of `callActivityId` here, simply not to loose this information
* will check the code from Operate side, what will be the consequences of having ordinal instead of proper `callActivityId`. I already checked but can't say for sure. Post importer may be affected and also getting statistics for list view diagram badges (`/api/process-instances/statistics` Internal API endpoint). Will add it as todo item in my next task.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
sdorokhova: @Zelldon I decided to set default tree path value for ELEMENT_ACTIVATING and BpmnElementType=PROCESS as this is the first record that comes for process instance before any incident record

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
sdorokhova: @Zelldon sorry, what do you mean here?

mustafadagher:
The commit message here https://github.com/camunda/camunda/pull/24037/commits/fd20781b52484b46a029346b0ed26ab2ce511b83 is wrong. I just resolved a conflict in the browser here.

Issue #23591, opened 2024-10-15
Unify C8Run scripts between windows and linux


jessesimpson36:
**Describe the issue:**

C8Run has platform-specific parts of it, and as result, we ended up with 2 different projects,  we have a
* bash script representing how the project works on linux + mac
* golang script representing how the project works on windows

**Historical justification**

C8Run began by adapting the camunda 7 run distribution scripts, which features a bash script for linux+mac and a **batch** script for windows.  For C8Run, I started with the same premise. I successfully created a bash script, however, I found that I was exceeding the capabilities of the batch scripting language, and made the decision to switch to Golang for the windows version.  I also had this idea that I was being consistent with the scripts in the camunda repo  (download camunda-zeebe .tar.gz file and untar it, and there's a bash and batch script for `./bin/camunda`), which means that other Camunda developers would be able to contribute to it easier. However,  it turned out that those scripts were created by bots, and bash + batch are not languages  many camunda developers is familiar with.

**What are our hopes for unifying these scripts?**

There's some logic that is identical between the windows and linux versions. I'd prefer if these bits of logic were not duplicated.  I also want the packaging process to be identical for both, and would like to be able to do things like cross-compile a package, such as a windows build but on a linux machine.

Furthermore, there is some functionality that is implemented for Linux, that is not implemented for Windows. If this was one script, the hope would be that there would not be any left-out functionality.

hisImminence:
Done!

PR #24040, opened 2024-10-25
fix: unify c8run scripts

jessesimpson36:
## Description

#23591 

c8run currently exists as a bash script for linux+mac and a golang script for windows. However, there is a lot of redundant logic when implemented across the two languages, so I instead want to move to an all-golang script that will handle the specific parts of OS that needs to be different from eachother in the form of an interface with minimal functions.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


c8run/internal/archive/archive.go
```diff
@@ -0,0 +1,289 @@
...
+				header.Name = path + "/"
+				if err := tw.WriteHeader(header); err != nil {
+					return err
+				}
+				return nil
+
+			}
+		})
+		if err != nil {
+			return err
```
sjiekak, line 69: you might be interested to wrap the error and provide more information about which file had an error

c8run/internal/archive/archive.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	tr := tar.NewReader(gz)
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			panic(err)
```
sjiekak, line 103: why panic ?

c8run/internal/archive/archive.go
```diff
@@ -0,0 +1,289 @@
...
+			_, err = os.Stat(parent)
+			if errors.Is(err, os.ErrNotExist) {
+				if err := os.MkdirAll(parent, 0755); err != nil {
+					panic(err)
+				}
+			}
+		}
+		file, err := os.OpenFile(
+			absFileName,
+			os.O_RDWR|os.O_CREATE|os.O_TRUNC,
```
sjiekak, line 125: can be a constant

c8run/internal/archive/archive.go
```diff
@@ -0,0 +1,289 @@
...
+		if finfo.Mode().IsDir() {
+			if err := os.MkdirAll(absFileName, 0755); err != nil {
+				panic(err)
+			}
+			continue
+		} else {
+			parent := filepath.Dir(absFileName)
+			_, err = os.Stat(parent)
+			if errors.Is(err, os.ErrNotExist) {
+				if err := os.MkdirAll(parent, 0755); err != nil {
```
sjiekak, line 118: `0755` can be a constant, you could also use os flags as you did below

c8run/internal/archive/archive.go
```diff
@@ -0,0 +1,289 @@
...
+			panic(err)
+		}
+		if cpErr != nil {
+			panic(cpErr)
+		}
+		if n != finfo.Size() {
+			return fmt.Errorf("wrote %d, want %d", n, finfo.Size())
+		}
+	}
+	return nil
```
sjiekak, line 143: watch out, juggling errors.... you may remove `cpErr` and keep using `err`. The closeError reports the wrong error as well.

c8run/internal/archive/archive.go
```diff
@@ -0,0 +1,289 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(tw, file)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
sjiekak, line 175: return err

c8run/internal/unix/stub.go
```diff
@@ -0,0 +1,33 @@
+//go:build !(linux || darwin)
+
+package unix
+
+import (
+	"os"
+	"os/exec"
+)
+
+func (w *UnixC8Run) OpenBrowser(name string) {
```
sjiekak, line 10: this is probably implementing an interface, right ? Be nice, and use

```
var _ ThisInterface = &UnixC8Run{}
```

c8run/package.go
```diff
@@ -0,0 +1,153 @@
...
+		os.Remove(filepath.Join("log", "camunda.log"))
+	}
+	_, err = os.Stat(filepath.Join("log", "connectors.log"))
+	if errors.Is(err, os.ErrNotExist) {
+		os.Remove(filepath.Join("log", "connectors.log"))
+	}
+	_, err = os.Stat(filepath.Join("log", "elasticsearch.log"))
+	if errors.Is(err, os.ErrNotExist) {
+		os.Remove(filepath.Join("log", "elasticsearch.log"))
+	}
```
sjiekak, line 27: 3 remarks:
- style: use for loop
- correctness : seems like the files are being deleted only when they don't exist
- approach : prefer something simpler which allows to clean undesired files  using a single command. It would be nice if the files needing cleaning would be under a specific parent that cou can simply remove

c8run/package.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func downloadAndExtract(filePath, url, extractDir string, extractFunc func(string, string) error) error {
+	_, err := os.Stat(filePath)
+	if errors.Is(err, os.ErrNotExist) {
+		err = archive.DownloadFile(filePath, url)
```
sjiekak, line 33: "download if not exist" can be part of download archive

c8run/internal/unix/unix.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"syscall"
+)
+
+func (w *UnixC8Run) OpenBrowser(name string) {
+	operateUrl := "http://localhost:8080/operate/login"
```
sjiekak, line 15: :warning:  why operate url ???

c8run/internal/windows/windows.go
```diff
@@ -0,0 +1,66 @@
...
+	openBrowserCmd := exec.Command("cmd", "/C", openBrowserCmdString)
+	openBrowserCmd.SysProcAttr = &syscall.SysProcAttr{
+		// CreationFlags: 0x08000000 | 0x00000200, // CREATE_NO_WINDOW, CREATE_NEW_PROCESS_GROUP : https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags
+	}
+	fmt.Println(name + " has successfully been started.")
+	openBrowserCmd.Run()
+}
+
+func (w *WindowsC8Run) GetProcessTree(commandPid int) []*os.Process {
+	return process_tree(int(commandPid))
```
sjiekak, line 25: not part of the PR but... the method `process_tree` may return `[]*os.Process{foo, bar, nil}`, which is what a caller might expect.

```
processList := make([]*os.Process, len(tree))
```

should be

```
processList := make([]*os.Process, 0, len(tree))
```

c8run/package.go
```diff
@@ -0,0 +1,153 @@
...
+	camundaFilePath := "camunda-zeebe-" + camundaVersion + ".zip"
+	camundaUrl := "https://github.com/camunda/camunda/releases/download/" + camundaVersion + "/" + camundaFilePath
+	connectorsFilePath := "connector-runtime-bundle-" + camundaVersion + "-with-dependencies.jar"
+	connectorsUrl := "https://repo1.maven.org/maven2/io/camunda/connector/connector-runtime-bundle/" + camundaVersion + "/" + connectorsFilePath
+
+	Clean(camundaVersion, elasticsearchVersion)
+
+	err := downloadAndExtract(elasticsearchFilePath, elasticsearchUrl, "elasticsearch-"+elasticsearchVersion, archive.UnzipSource)
+	if err != nil {
+		panic(err)
```
sjiekak, line 62: package functions should return an error and let main panic... otherwise this is difficult to test

sjiekak (commented):
few things can be improved. Can the code be formatted too ?

c8run/internal/archive/archive.go
```diff
@@ -0,0 +1,289 @@
...
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func DownloadFile(filepath string, url string) error {
+	out, err := os.Create(filepath)
+	if err != nil {
+		return err
```
hamza-m-masood, line 19: You could include context in error messages to make them more informative. For example `fmt.Errorf("...")`

c8run/package.go
jessesimpson36: https://github.com/camunda/camunda/pull/24040/commits/4aa3968f155af3b001d67e383aedef54f6eb22af

the above  commit addresses this comment

c8run/internal/windows/windows.go
jessesimpson36: what is the purpose of the zero in your second argument? what information does that convey?

c8run/internal/archive/archive.go
jessesimpson36: I also addressed this in https://github.com/camunda/camunda/pull/24040/commits/4aa3968f155af3b001d67e383aedef54f6eb22af

c8run/internal/archive/archive.go
jessesimpson36: I've removed my panics to just main and stubs.
https://github.com/camunda/camunda/pull/24040/commits/4aa3968f155af3b001d67e383aedef54f6eb22af

c8run/internal/archive/archive.go
jessesimpson36: I like the idea, but in implementation, my first attempt at this resulted in an error message like 

"Failed to create file named file.txt, Original Error: file.txt already exists"

Which isn't any more descriptive than "file.txt already exists".

What sort of context do people usually add to their error messages? do they append filenames+line numbers ?

jessesimpson36:
I had to rebase on main since Vault was reporting errors on the credentials. I assume because a later commit on main changes them

jessesimpson36:
> few things can be improved. Can the code be formatted too ?

I formatted each golang file with `go fmt` in this commit

https://github.com/camunda/camunda/pull/24040/commits/82e99dea81e94d389901cd15516a9417409df716

c8run/internal/archive/archive.go
jessesimpson36: Open flags are not the same as file mode permissions. But I did separate these out into constants in this commit:

https://github.com/camunda/camunda/pull/24040/commits/ba17a0fe96c2dda5b11eeb0834a0e1bfe262396f

c8run/internal/archive/archive.go
jessesimpson36: https://github.com/camunda/camunda/pull/24040/commits/ba17a0fe96c2dda5b11eeb0834a0e1bfe262396f

you right.

c8run/internal/windows/windows.go
```diff
@@ -0,0 +1,67 @@
...
+	fmt.Println(name + " has successfully been started.")
+	openBrowserCmd.Run()
+	return nil
+}
+
+func (w *WindowsC8Run) GetProcessTree(commandPid int) []*os.Process {
+	return process_tree(int(commandPid))
+}
+
+func (w *WindowsC8Run) GetVersionCmd(javaBinaryPath string) *exec.Cmd {
```
Sijoma, line 29: Jumping on the review bandwagon! 

VersionCmd would be enough from the naming perspective and more in line with Go style guides.

Prefixing methods and functions is not idiomatic in Go. 
https://go.dev/doc/effective_go#Getters

I believe you would only do that when it's actually a GET request. 
https://google.github.io/styleguide/go/decisions.html#getters

c8run/main.go
```diff
@@ -1,23 +1,20 @@
 package main
 
 import (
+	"c8run/internal/unix"
+	"c8run/internal/windows"
 	"errors"
+        pkgErrors "github.com/pkg/errors"
```
Sijoma, line 7: This package is archived since 2021.

c8run/internal/unix/unix.go
```diff
@@ -0,0 +1,64 @@
...
+	mainClass := "io.camunda.connector.runtime.app.ConnectorRuntimeApplication"
+	springConfigLocation := "--spring.config.location=" + parentDir + "/connectors-application.properties"
+	connectorsCmd := exec.Command(javaBinary, "-cp", classPath, mainClass, springConfigLocation)
+	connectorsCmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	return connectorsCmd
+}
+
+func (w *UnixC8Run) GetCamundaCmd(camundaVersion string, parentDir string, extraArgs string) *exec.Cmd {
+	camundaCmdString := parentDir + "/camunda-zeebe-" + camundaVersion + "/bin/camunda"
+	fmt.Println(camundaCmdString)
```
Sijoma, line 60: I would leave the printing of the command outside of this function to have a cleaner separation of concerns.

c8run/main.go
```diff
@@ -71,24 +68,21 @@ func queryCamundaHealth(name string, url string) {
 		fmt.Println("Error: " + name + " did not start!")
 		os.Exit(1)
```
Sijoma, line 69: Not in this PR. However, I would return the error:

"fmt.Errorf("Application %s did not start!", name)"

The caller can then handle what they want to do with that. Otherwise it is surprising that the function does not return an error in this case but exits.

You also already handle the case the same (your main.go exits on error and prints it)

c8run/internal/archive/archive.go
Sijoma: Usually its "DownloadFile: error on file creation: <error coming from os.Create>" but it depends a bit what you want to expose to the consumer of your function. 

The text usually helps to identify the code path that failed by searching for the string. Some logging libraries also support stack traces. 

Go Blog about wrapping errors: https://go.dev/blog/go1.13-errors#errors-and-package-apis

c8run/internal/windows/windows.go
jessesimpson36: Thanks for jumping in. More eyes is good on such a new project.
I fixed the naming of the functions in ea381f58309

c8run/internal/unix/unix.go
jessesimpson36: Good point. I added 377cc498872 to move the print statements into the main function (or remove them in cases where it wasn't useful.

c8run/main.go
jessesimpson36: that's a good point. I returned an error in 967cbfb2200 instead of printing+exiting

jessesimpson36: I'm not concerned about the weird function name. I'd wait for someone to complain about it before changing that.

c8run/package.go
jessesimpson36: I addressed style + correctness in 3e472c772df .  I will think on "approach" later.

c8run/main.go
jessesimpson36: Do you have recommendations for a replacement for this library? I kinda introduced the pkg/errors module as result of another review comment.

jessesimpson36: I removed the dependency on pkg/errors and used a combination of stdlib errors package and runtime/debug to produce stacktraces.

c8run/internal/archive/archive.go
jessesimpson36: I included function name  prefixes to each of the errors in d82b4686346 as you say.  Also included better log messages for some of the functions.

c8run/internal/archive/archive.go
jessesimpson36: I added this context for all errors

c8run/package.go
jessesimpson36: I've pushed that logic into DownloadFile.

c8run/package.go
jessesimpson36: I included `./c8run clean` which will delete these files.

c8run/go.mod
```diff
@@ -0,0 +1,3 @@
+module c8run
```
Sijoma, line 1:
```suggestion
module github.com/camunda/camunda/c8run
```

This would be the typical naming for a go module. 
`module github.com/org-name/repo-name/path/to/module-dir`

Then you can also get `go install` to work - (not sure if it also would work with the other name).  

So you can do the following:
```
go install github.com/camunda/camunda/c8run (installs c8run to your go path)
c8run (run camunda!!! 🏃 )
```

jessesimpson36: I think I did this in https://github.com/camunda/camunda/pull/24040/commits/1b49a0ea11314c6e68f18770c3f98146a8f420dd but it's hard to tell because I don't think I can reference a branch name in a `go install` command.

I'm also unsure how versioning works in golang. is it all just git tags? if so then well, I don't control the git tags applied in the camunda monorepo. it's worth a try though.

c8run/go.mod
```diff
@@ -1,3 +1,4 @@
-module c8run
+module github.com/camunda/camunda/c8run
 
+replace github.com/camunda/camunda/c8run => ./
```
Sijoma: Not sure why the replace directive is needed? Usually that would only be required if you "replace" a public module with an external one.

jessesimpson36: Already removed it in a later commit. I realized it wasn't necessary.

I thought what would happen is if you import github.com/camunda/camunda/c8run then it would  do a git clone on main and use that for your library, and then it'd be a mismatch of which versions of the library you're using.

jessesimpson36: https://github.com/camunda/camunda/pull/24040/commits/ad6c501685ea273d36a8d96562c78be0afc236be

Sijoma (approved):
LGTM after very thorough review! 🚀

PR #24048, opened 2024-10-25
Add Mapping to Role on Role.ADD_ENTITY command

### LINKED ISSUE #23843: Add `Mapping` to `Role` on `Role.ADD_ENTITY` command
<h2>Description</h2>

- In the `RoleAddEntityProcessor` extend the `isEntityPresent` by checking the `MappingState` if a `Mapping` with the provided `entityKey` exists.
- Extend the `PersistedMapping` with an extra property:
    - `ArrrayProperty<LongValue> roleKeys`.
- In the `RoleEntityAddedApplier` add a case if the `entityType` == `MAPPING`. Inside this case add the role to the persisted mapping similar to how we already do for users.

nicpuppa:
## Description

This PR adds `roleKeys` property into the `PersistedMapping` object and allow the `RoleEntityAddedApplier` to add the role key into the mapping state if the EntityType is MAPPING

## Related issues

closes #23843

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleEntityAddedApplier.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleEntityAddedApplier.java
@@ -8,6 +8,8 @@
 package io.camunda.zeebe.engine.state.appliers;
 
 import io.camunda.zeebe.engine.state.TypedEventApplier;
+import io.camunda.zeebe.engine.state.mutable.MutableMappingState;
+import io.camunda.zeebe.engine.state.mutable.MutableProcessingState;
 import io.camunda.zeebe.engine.state.mutable.MutableRoleState;
 import io.camunda.zeebe.engine.state.mutable.MutableUserState;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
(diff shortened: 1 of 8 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleEntityAddedApplier.java
```diff
@@ -31,11 +30,14 @@ public RoleEntityAddedApplier(final MutableProcessingState state) {
...
-    if (EntityType.MAPPING.equals(value.getEntityType())) {
-      mappingState.addRole(value.getEntityKey(), value.getRoleKey());
+    switch (value.getEntityType()) {
+      case USER -> userState.addRole(value.getEntityKey(), value.getRoleKey());
+      case MAPPING -> mappingState.addRole(value.getEntityKey(), value.getRoleKey());
+      default ->
+          throw new IllegalStateException(
+              "Unknown or unsupported entity type: '"
+                  + value.getEntityType()
+                  + "'. Please contact support for clarification.");
```
remcowesterhoud, line 40: 🔧 Could you change this so that we follow our [error message guidelines](https://github.com/camunda/camunda/wiki/Error-Guidelines)?

Eg. `Expected to add entity '<key>' to role '<key>', but entities of type '<type>' cannot be added to roles`

remcowesterhoud (approved):
Thanks @nicpuppa, 1 small comment. Please take a look 👀

PR #24069, opened 2024-10-25
fix: exclude task-template from being populated and created by tasklist when CamundaExporter is enabled

panagiotisgts:
## Description
- Change `tasklist-task` version to 8.5.0
- When `ZEEBE_BROKER_EXPORTERS_CAMUNDA_CLASSNAME=io.camunda.exporter.CamundaExporter`
- - Exclude `tasklist-task` from Tasklist's SchemaManager
- - Exclude UserTask and Form record from being handled by Tasklist's 8.7 Importer

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/dist/src/main/resources/application-tasklist.properties
+++ b/dist/src/main/resources/application-tasklist.properties
@@ -14,6 +14,9 @@ graphql.schema-strategy=annotations
 graphql.annotations.base-package=io.camunda.tasklist
 graphql.annotations.always-prettify=false
 graphql.annotations.input-prefix=
+
+# Temporary workaround for CamundaExporter
+camunda.tasklist.exporterClassName=${zeebe.broker.exporters.camunda.className:${zeebe.broker.exporters.camundaExporter.className:${ZEEBE_BROKER_EXPORTERS_CAMUNDA_CLASSNAME:${ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_CLASSNAME}:}}}
 #---
 spring.config.activate.on-profile=identity-auth
 spring.security.oauth2.resourceserver.jwt.issuer-uri=${camunda.identity.issuer:${camunda.tasklist.identity.issuerUrl:${CAMUNDA_TASKLIST_IDENTITY_ISSUER_URL:}}}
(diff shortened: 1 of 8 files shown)


ChrisKujawa (approved):
For the moment it makes sense I guess. In general, we should get rid of the 8.7 importer. What do you think when or how we can delete it? Operate has no 8.7 Importer I believe? 🤔

panagiotisgts:
> In general, we should get rid of the 8.7 importer. What do you think when or how we can delete it? Operate has no 8.7 Importer I believe?

Operate also has an 8.7 Importer for the time being. When Tasklist FE migrates to V2 APIs most of Tasklist's services and business logic will be removed, along with the 8.7 importer.

dist/src/main/resources/application-tasklist.properties
```diff
@@ -14,6 +14,9 @@ graphql.schema-strategy=annotations
 graphql.annotations.base-package=io.camunda.tasklist
 graphql.annotations.always-prettify=false
 graphql.annotations.input-prefix=
+
+# Temporary workaround for CamundaExporter
+camunda.tasklist.exporterClassName=${zeebe.broker.exporters.camunda.className:${zeebe.broker.exporters.camundaExporter.className:${ZEEBE_BROKER_EXPORTERS_CAMUNDA_CLASSNAME:${ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_CLASSNAME}:}}}
```
Zelldon, line 19: TIL; Did you tested this somehow? :)

ChrisKujawa (changes requested):
I realized your planned change will not work for SaaS (what we try to fix). As the importer is running in separate container/pod.

Lets just fix the version aa first increment. I try to get the importer disabled in SaaS for now.

panagiotisgts:
Why wouldn't it? If the config property is present the behavior should be the same. Am I missing something?

Zelldon:
Each deployment has its own config map. That is the current state at least

panagiotisgts:
Closing this as is no longer needed

PR #24070, opened 2024-10-25
feat: [RDBMS] merge executionQueue items if possible

stoerti:
## Description

Introduced a QueueItemMerger interface to the ExecutionQueue. This can be used to search for existing compatible queueItems and merge the new queueItem into the existing one.

With this the number of flushes and/or SQL statements can be reduced.

closes #24051


db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/ExecutionQueue.java
```diff
@@ -53,56 +55,100 @@ public void registerPostFlushListener(final PostFlushListener listener) {
...
+      final var session =
+          sessionFactory.openSession(
+              ExecutorType.BATCH, TransactionIsolationLevel.READ_UNCOMMITTED);
+
+      var flushedElements = 0;
+      try {
+        while (!queue.isEmpty()) {
+          final var entry = queue.peek();
+          LOG.trace("[RDBMS ExecutionQueue, Partition {}] Executing entry: {}", partitionId, entry);
+          session.update(entry.statementId(), entry.parameter());
```
p-wunderlich, line 86: Hm I know, we also had it before, but is it correct to always do an update? Although we also have inserts? (Don't know what difference it makes for mybatis ... )

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/FlowNodeInstanceWriter.java
```diff
@@ -36,13 +38,47 @@ public void create(final FlowNodeInstanceDbModel flowNode) {
...
+
+    public EndFlowNodeToInsertMerger(final EndFlowNodeDto dto) {
+      this.dto = dto;
+    }
+
+    @Override
+    public boolean canBeMerged(final QueueItem queueItem) {
+      return queueItem.contextType() == ContextType.FLOW_NODE
+          && queueItem.id().equals(dto.flowNodeInstanceKey())
+          && queueItem.parameter() instanceof FlowNodeInstanceDbModel;
```
p-wunderlich, line 69: Should always be a FlowNodeInstanceDbModel, or? Since instanceof is always a little bit expansive when it comes to performance, we could maybe remove it?

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/FlowNodeInstanceWriter.java
```diff
@@ -36,13 +38,47 @@ public void create(final FlowNodeInstanceDbModel flowNode) {
...
+    }
+
+    @Override
+    public QueueItem merge(final QueueItem originalItem) {
+      final var newParameter =
+          FlowNodeInstanceDbModelBuilder.of((FlowNodeInstanceDbModel) originalItem.parameter())
```
p-wunderlich, line 75: Hm, for UserTask i created a "toBuilder" function on the Model itself which creates the builder .. :D (At the end, same behaviour, but we should be consistent of how we create the builder. => Needs to be discussed.

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/FlowNodeInstanceWriter.java
p-wunderlich: same for the other merger

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/ProcessInstanceWriter.java
```diff
@@ -33,12 +34,50 @@ public void create(final ProcessInstanceDbModel processInstance) {
...
   }
 
-  public void update(final ProcessInstanceDbModel processInstance) {
-    executionQueue.executeInQueue(
-        new QueueItem(
-            ContextType.PROCESS_INSTANCE,
-            processInstance.processInstanceKey(),
-            "io.camunda.db.rdbms.sql.ProcessInstanceMapper.update",
-            processInstance));
+  public void end(
```
p-wunderlich, line 37: I'm somehow not happy with the function name "end" :D

zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/ProcessInstanceExportHandler.java
```diff
@@ -38,33 +38,15 @@ public void export(final Record<ProcessInstanceRecordValue> record) {
...
-              DateUtil.toOffsetDateTime(Instant.ofEpochMilli(record.getTimestamp())),
-              value.getTenantId(),
-              value.getParentProcessInstanceKey(),
-              value.getParentElementInstanceKey(),
-              null,
-              value.getVersion()));
+      processInstanceWriter.end(
+          value.getProcessInstanceKey(),
+          ProcessInstanceState.CANCELED,
+          DateUtil.toOffsetDateTime(Instant.ofEpochMilli(record.getTimestamp())));
```
p-wunderlich, line 49: I created in the DateUtil class (on the variables PR) a function which also does the ofEpochMilli stuff, since we always use it like this.

db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/ExecutionQueue.java
```diff
@@ -53,56 +55,100 @@ public void registerPostFlushListener(final PostFlushListener listener) {
...
+  public boolean tryMergeWithExistingQueueItem(final QueueItemMerger... combiners) {
+    synchronized (queue) {
+      int index = queue.size() - 1;
+      for (final Iterator<QueueItem> it = queue.descendingIterator(); it.hasNext(); ) {
+        final QueueItem item = it.next();
+
+        for (final QueueItemMerger merger : combiners) {
+          if (merger.canBeMerged(item)) {
+            LOG.debug("Merging new item with item {}, {}", item.contextType(), item.id());
+            final var newItem = merger.merge(item);
```
p-wunderlich, line 134: since newItem is just used once, this could also be inlined in the queue.set function.

db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/ExecutionQueue.java
stoerti: Funny thing: MyBatis forwards all insert() calls to update() 

`  @Override
  public int insert(String statement, Object parameter) {
    return update(statement, parameter);
  }
`

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/FlowNodeInstanceWriter.java
stoerti: Currently yes. But in future maybe we also have other parameter objects (e.g. when adding incidents). Then we could either include the statement type to the condition or check for the instance type of the parameter. I've for now chosen the instance type to avoid magic string constants.

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/ProcessInstanceWriter.java
stoerti: 💭  "Finish"? I couldn't use "complete" because it's already a state name (and there is also cancel/terminate).

zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/ProcessInstanceExportHandler.java
stoerti: ❓Where? I can't find it in the PR

p-wunderlich: A is in the userTask branch. 

  public static OffsetDateTime toOffsetDateTime(final Long timestamp) {
    return OffsetDateTime.ofInstant(Instant.ofEpochMilli(timestamp), ZoneOffset.UTC);
  }

romansmirnov (commented):
Thanks for raising this PR. To me, merging the statements as much as possible makes sense. However, I didn't go into all the details in this PR because I would like to align on the approach first.

I want to discuss an alternative approach - similar to what is already done in the ES/OS-based `camunda-exporter`. There the entities to be flushed, are cached:

https://github.com/camunda/camunda/blob/622f2cd835abe17412b0523941b3ef11d3819c53/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java#L26

Basically, all state changes are happening in-memory first, and then the flush happens. In terms of RDBMS, maybe we could take a similar approach with the difference that in RDBMS no upserts will happen.

Based on that, I think of that approach:
* When collecting the batch to flush, the entities to store/update are cached.
* Each cached entity has a persistent state assigned like `transient` and `persistent`:
  * When a process instance create event is processed, then a process instance entity is being created, and its persistent state is `transient`, meaning, it is not stored yet in the DB.
  * When a process instance complete event is processed, then a process instance entity being created, and its persistent state is `persistent`, meaning, the actual entity is already stored in DB, now this entity gets updated in the DB.
  * When the process instance create and complete event is processed during the same batch, then the entity is in the persistent state `transient`and the complete event will only update the process instance state and end time on the cached object.
* When flushing
  * When the persistent state is `transient`, an `insert` statement is executed.
  * When the persistent state is `persistent`, and `update` statement is executed.

The benefits of this approach are:
* We use the same/similar approach as with the ES/OS-based `camunda-exporter`on how to "merge" the SQL statements to avoid unnecessary update statements.
* I believe how to update an entity becomes easier, because it is not necessary to create dedicated model classes and update statements for different types of events, and additionally some merge logic.

Let's have a follow-up next week on this topic.

p-wunderlich (approved):

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/FlowNodeInstanceWriter.java
```diff
@@ -36,13 +35,26 @@ public void create(final FlowNodeInstanceDbModel flowNode) {
...
+    }
+  }
+
+  private boolean mergeToQueue(
+      final long key,
+      final Function<FlowNodeInstanceDbModelBuilder, FlowNodeInstanceDbModelBuilder>
+          mergeFunction) {
+    return executionQueue.tryMergeWithExistingQueueItem(
+        new UpsertMerger<>(
+            ContextType.FLOW_NODE, key, FlowNodeInstanceDbModel.class, mergeFunction));
```
danielkelemen, line 58: 💭 A bit late but: Could these methods be moved to the `ExecutionQueue`? I think it would be nicer if the EQ would decide if it merges or just adds the item, instead of the writers. It also seems a bit repetitive this way.

Issue #5478, opened 2020-10-05
Find a good value for max append per follower


Zelldon:
**Description**

The [MAX_APPENDS](https://github.com/zeebe-io/zeebe/blob/develop/atomix/cluster/src/main/java/io/atomix/raft/cluster/impl/RaftMemberContext.java#L34) is used to limit how many appends are send to one follower at once. Currently it is static and set to two. It is likely that this configure can help to improve the throughput. We should evaluate and benchmark different values to get an better understand how it impacts the performance and what would be a good value.

Blocked by #5477

npepinpe:
As with the other one, we can revise the priority if we see that this could help us reach our current performance goal sooner.

megglos:
relates to https://github.com/camunda/product-hub/issues/642
https://github.com/camunda/product-hub/issues/643

megglos:
@falko do you have any input in this?

entangled90:
I ran some benchmarks with different values for MAX_APPENDS in both the default scenario (with 150 PI/s) and the realistic scenario.

# Preconditions
- I waited for the cluster to be balanced (each pod was leader of 1 partition)
- I waited for the nodes to enter into a steady state
- I checked that the in the panel "Inflight request count" the number of requests reached the configured value

# Results 
## Default Scenario
In this scenario, the payload is fixed 32KB (same as the MAX_BATCH_SIZE): all batches will contain only one record.

- [MAX_APPEND=2 (default value)](https://grafana.dev.zeebe.io/d/zeebe-dashboard/zeebe?orgId=1&var-DS_PROMETHEUS=prometheus&var-cluster=All&var-namespace=cs-max-appends&var-pod=All&var-partition=All&from=1728314454000&to=1728315795000)
![image](https://github.com/user-attachments/assets/8656c24e-5773-41e0-b9e6-9fea4e1f638a)

- [MAX_APPEND=4](https://grafana.dev.zeebe.io/d/zeebe-dashboard/zeebe?orgId=1&var-DS_PROMETHEUS=prometheus&var-cluster=All&var-namespace=cs-max-appends&var-pod=All&var-partition=All&from=1728316554000&to=1728317715000)
![image](https://github.com/user-attachments/assets/00ea24af-4f94-40ee-a42c-61d8f571ada2)

[...]

PR #24073, opened 2024-10-25
feat: set default MAX_APPEND_PER_FOLLOWER to 6

entangled90:
## Description
Default value for MAX_APPEND_PER_FOLLOWER is set now to 6

## Related issues
relates #5478


npepinpe (approved):

lenaschoenburg:
@entangled90 It'd be good to update the config templates too https://github.com/camunda/camunda/blob/77747f40471573d3ae9363ca5d00fd1b6c6726fd/dist/src/main/config/broker.standalone.yaml.template#L1071

entangled90:
Thanks for spotting it! modified `broker.yaml.template` as well

PR #24082, opened 2024-10-25
Removing recursion from process cancellation logic

### LINKED ISSUE #21991: StackOverflow during process cancelation
**Describe the bug**

This issue originated from a the following [support case](https://jira.camunda.com/browse/SUPPORT-23461?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D) where the client deployed a recursive model where P1 calls upon P2 which calls P3 which in turn calls again P1, this created a huge nest of PIs that started wondering performance.

Upon trying to cancel the PI which also in turn also deletes their children, the client had a stackOverflow Exception due to the large number of nested PIs. This issue is quite similar to [StackOverflowError during process instance modification · Issue #20307 · camunda/camunda · GitHub](https://github.com/camunda/camunda/issues/20307).

Work around:

To address this they would need to ban all the PIs that are calling each other before this renders their cluster unusable.
https://github.com/camunda-cloud/runbook/blob/main/zeebe/troubleshooting-guide.md#endless-recursion-loop-in-process-instances

- Forward to the affected Leader broker managment port 9600, e.g. kubectl port-forward zeebe-0 9600
- Pause processing via `curl -XPOST localhost:9600/actuator/partitions/pauseProcessing`
- Follow this [guide](https://github.com/camunda-cloud/runbook/blob/main/zeebe/zeebe-investigate-data.md#investigate-state) to find all process instances
- Follow this [guide](https://github.com/camunda-cloud/runbook/blob/main/zeebe/zeebe-investigate-data.md#force-banning-all-existing-instances) to ban all instances.
- Resume processing `curl -XPOST localhost:9600/actuator/partitions/resumeProcessing`.

**To Reproduce**

-->Deploy a process model as described above, let it grow (we don't know what is the threshold of nested children where this happens), and try to ban the parent process instance. 

**Expected behavior**

It should ban the PI and all its children without problems.

**Log/Stacktrace**

image

**Environment:**
- OS: 
- Zeebe Version: 
- Configuration: 

**Support Cases:**
https://jira.camunda.com/browse/SUPPORT-23461

ana-vinogradova-camunda:
## Description

There was an issue where cancelling would run into a `StackOverflowError` because of the recursive nature of the terminating logic. 

In this PR the recursive logic is removed and there is a test for the solution that verifies `StackOverflowError` is not present anymore, however, I am not committing it as it requires to adjust stack size.

## Checklist

- [ ] for CI changes:

## Related issues

closes #21991

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCancelProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCancelProcessor.java
@@ -26,6 +26,7 @@ import io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent;
 import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
 import io.camunda.zeebe.protocol.record.value.PermissionType;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
+import java.util.Optional;
 
 public final class ProcessInstanceCancelProcessor
     implements TypedRecordProcessor<ProcessInstanceRecord> {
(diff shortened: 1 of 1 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCancelProcessor.java
```diff
@@ -135,17 +136,24 @@ private boolean validateCommand(
...
-
       final var parentProcessInstanceKey = instance.getValue().getParentProcessInstanceKey();
       if (parentProcessInstanceKey > 0) {
-
-        return getRootProcessInstanceKey(parentProcessInstanceKey);
+        return Optional.of(parentProcessInstanceKey);
       }
     }
-    return instanceKey;
+    return Optional.empty();
```
remcowesterhoud, line 157: 🔧 Personally I find this easier to read. But your solution is definitely not wrong, so I'll leave it up to you if you want to change it.

```java
  private long getRootProcessInstanceKey(long instanceKey) {
    var instance = elementInstanceState.getInstance(instanceKey);
    while (instance != null && instance.getValue().hasParentProcess()) {
      instanceKey = instance.getValue().getParentProcessInstanceKey();
      instance = elementInstanceState.getInstance(instanceKey);
    }
    
    return instanceKey;
  }
```

remcowesterhoud (approved):
Thanks @ana-vinogradova-camunda 🚀

remcowesterhoud:
@ana-vinogradova-camunda has a holiday today, and since the code freeze is tomorrow I'll merge it as-is. The solution works and can always be refactored later if we deem it necessary.

PR #24086, opened 2024-10-25
feat: Implement basic document upload

vsgoulart:
## Description

This PR implements document uploads on Tasklist using the new `filepicker` component from `form-js`

I couldn't add a test because `DataTransfer` is not mocked on `jsdom`. I'll try to add an E2E tests Monday.

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/team-hto/issues/662

### DIFF
--- a/tasklist/client/src/modules/components/FormJSRenderer/index.tsx
+++ b/tasklist/client/src/modules/components/FormJSRenderer/index.tsx
@@ -18,7 +18,10 @@ import '@bpmn-io/form-js-viewer/dist/assets/form-js-base.css';
 import '@bpmn-io/form-js-carbon-styles/src/carbon-styles.scss';
 
 type Props = {
-  handleSubmit: (variables: Variable[], files?: File[]) => Promise<void>;
+  handleSubmit: (
+    variables: Variable[],
+    files?: Map<string, File[]>,
+  ) => Promise<void>;
   schema: string;
   data?: Record<string, unknown>;
   readOnly?: boolean;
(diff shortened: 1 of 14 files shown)


tasklist/client/src/modules/mutations/useUploadDocuments.tsx
```diff
@@ -0,0 +1,70 @@
...
+          metadataMapping.push([
+            key,
+            await Promise.all(metadata.map(({response}) => response!.json())),
+          ]);
+        }
+
+        return new Map(metadataMapping);
+      }
+
+      throw new Error('Failed to upload all files');
```
vsgoulart, line 64: All this logic is necessary because they didn't manage to finish the endpoint for batch uploads. This should be implemented on the coming weeks and then I'll be able to remove this.

tasklist/client/src/modules/components/FormJSRenderer/index.tsx
```diff
@@ -88,6 +93,70 @@ function useScrollToError(manager: FormManager) {
...
+    if (metadata === undefined) {
+      return;
+    }
+
+    result = set(result, filepickerPath, metadata);
+  });
+
+  return result;
+}
+
```
vsgoulart, line 159: This is necessary because on `form-js` we can have inputs inside groups and lists which makes it hard to find the path of filepickers in the schema. This is also not trivial to do on the `form-js` so for simplicity sake we decided to keep it on Tasklist now, we might change that in the future.

We need the paths because after uploading the documents we need to replace the file key from the file store with the metadata by document upload endpoint.

### Example
We submit a form with with a filepicker nested inside a group and dynamiclist which outputs the data like this:
```json
{
  "group": {
    "list" [ 
      {
        "filepicker_input": "file::key"
      }
    ]
  }
}
```

Then after a successful upload we will enrich this data with the metadata returned by the endpoint like so:
```json
{
  "group": {
    "list" [ 
      {
        "filepicker_input": [/** metadata file0 */, /** metadata file1 */]
      }
    ]
  }
}
```

pedesen (approved):
LGTM :+1:

PR #24111, opened 2024-10-28
Add ArchUnit test to ensure processors doing authorization checks

### LINKED ISSUE #22896: Verify we are checking permissions in all processors
<h2>Description</h2>

We need to ensure when someone introduces a new processor, they will consider authorizations. For this we should introduce an ArchUnit that:

1. Finds all processors
2. Verifies they are implementing an `Authorizable` interface
    - **or** the processor is explicitly listed as "does not require authoirzation checks" 
3. Fail if there is a processor that doesn't match the above checks

Not all processors are required to implement authorization checks. Because of this the test should keep track of a collection of processors that are allowed to not implement `Authorizable`. New processor must be explicitly added to this list. This forces a developer to think about authorizations and can determine if they should be added or not.

remcowesterhoud:
## Description

This test has 2 goals. The first is ensuring the authorization checks are not removed from a processor by accident. The second goal is ensuring devs are triggered to think about authorizations when they create a new processor. If they add a new processor and don't perform auth checks, this test will fail. To circumvent this the processors needs to be explicitly excluded from authorization checks.

I had to add he `UserDeleteProcessor` auth checks to make it succeed. This was missed and didn't have an issue. The ITs for it will be implemented in https://github.com/camunda/camunda/issues/21694

## Checklist

- [ ] for CI changes:

## Related issues

closes #22896

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationArchTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationArchTest.java
@@ -19,47 +19,18 @@ import com.tngtech.archunit.lang.ArchCondition;
 import com.tngtech.archunit.lang.ArchRule;
 import com.tngtech.archunit.lang.ConditionEvents;
 import com.tngtech.archunit.lang.conditions.ArchConditions;
-import io.camunda.zeebe.engine.processing.bpmn.BpmnStreamProcessor;
-import io.camunda.zeebe.engine.processing.deployment.distribute.DeploymentDistributeProcessor;
-import io.camunda.zeebe.engine.processing.deployment.distribute.DeploymentDistributionCompleteProcessor;
-import io.camunda.zeebe.engine.processing.distribution.CommandDistributionAcknowledgeProcessor;
-import io.camunda.zeebe.engine.processing.distribution.CommandDistributionContinueProcessor;
-import io.camunda.zeebe.engine.processing.distribution.CommandDistributionFinishProcessor;
+import io.camunda.zeebe.engine.processing.ExcludeAuthorizationCheck;
 import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior;
 import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.AuthorizationRequest;
 import io.camunda.zeebe.engine.processing.job.DefaultJobCommandPreconditionGuard;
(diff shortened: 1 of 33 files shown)


remcowesterhoud:
@koevskinikola Let me know if anything's unclear. It was my first time working with ArchUnit and I can imagine it'd be the same for you 😄

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationArchTest.java
```diff
@@ -0,0 +1,183 @@
...
+    packages = "io.camunda.zeebe.engine.processing..",
+    importOptions = ImportOption.DoNotIncludeTests.class)
+public class AuthorizationArchTest {
+
+  @ArchTest()
+  public static final ArchRule PROCESSOR_CHECKS_AUTHORIZATION =
+      classes()
+          .that(thatProcessCommands())
+          .and(thatAreNotExcludedFromAuthorizationChecks())
+          .should(checkAuthorization());
```
koevskinikola, line 74: 👍 Very readable, very demure 👏

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationArchTest.java
```diff
@@ -0,0 +1,183 @@
...
+    processors.add(ProcessMessageSubscriptionCreateProcessor.class);
+    processors.add(ProcessMessageSubscriptionDeleteProcessor.class);
+    processors.add(ProcessInstanceBatchActivateProcessor.class);
+    processors.add(ProcessInstanceBatchTerminateProcessor.class);
+    processors.add(ProcessInstanceCreationCreateWithResultProcessor.class);
+    processors.add(TimerCancelProcessor.class);
+    processors.add(TimerTriggerProcessor.class);
+    processors.add(UserTaskProcessor.class);
+    return processors;
+  }
```
koevskinikola, line 142: 🔧 Did you consider creating an annotation interface like `@ExcludeAuthorizationCheck` that we can use to annotate processors that shouldn't perform this check?

This will be more readable to devs, and we won't need to update this class whenever we add a new processor. We will just need to add the following to this test:

```
.and()
.areNotAnnotatedWith(ExcludeAuthorizationCheck.class)
```

koevskinikola (approved):
👏 Nice job @remcowesterhoud, the code looks great! 

I have a suggestion below that might be worth discussing before the code is merged.

Issue #23987, opened 2024-10-24
Add Operation feature and define filter operations for `ProcessInstanceFilter`


danielkelemen:
## Description

In order to support advanced search queries for `/process-instances` we need to prepare the internal structure to support different operations and execute them properly.

## Acceptance criteria

- Introduce a new `Operation` class that supports the required operation types.
- Extend the required `ProcessInstanceFilter` properties with operations.
- Make sure underlying structures work with the new structure.
  - `SearchQueryRequestMapper`, `ProcessInstanceFilterTransformer`, RDBMS mappers etc.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/24097
```

PR #24113, opened 2024-10-28
fix: update ProcessInstanceMapper.xml properties

danielkelemen:
## Description

- Fixes the `ProcessInstanceMapper.xml` mybatis properties after rename.

## Related issues

https://github.com/camunda/camunda/issues/23987

### DIFF
--- a/db/rdbms/src/main/resources/mapper/ProcessDefinitionMapper.xml
+++ b/db/rdbms/src/main/resources/mapper/ProcessDefinitionMapper.xml
@@ -25,7 +25,8 @@
     NAME,
     TENANT_ID,
     VERSION_TAG,
-    VERSION
+    VERSION,
+    FORM_ID
     FROM PROCESS_DEFINITION
     <include refid="io.camunda.db.rdbms.sql.ProcessDefinitionMapper.searchFilter"/>
     <if test="sort != null and sort.orderings != null and !sort.orderings.isEmpty()">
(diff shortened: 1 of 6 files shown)


danielkelemen:
@stoerti, Do you know a quick fix for the current RDBMS CI issue?
`Caused by: java.net.BindException: Address already in use`

db/rdbms/src/main/resources/mapper/ProcessDefinitionMapper.xml
```diff
@@ -25,7 +25,8 @@
     NAME,
     TENANT_ID,
     VERSION_TAG,
-    VERSION
+    VERSION,
+    NULL as FORM_ID
```
stoerti, line 29: When quick-fixing things in RDBMS like this, can you create a follow-up issue with label `component/rdbms` and assign it to Patrick or me? Otherwise we will almost never find out that there is a new field we don't support :)

danielkelemen: Yes, if it is committed with this PR then I will create a follow-up so it's not forgotten.
I might drop that commit. I just wanted to see a green RDBMS job but it looks like it has other problems.
It doesn't really belong to this PR.

stoerti:
I've fixed the failing pipeline and also added a fix for the failing test because of the recently added formId inProcessDefinitionEntity

stoerti (approved):

PR #24124, opened 2024-10-28
Add Archiver configuration to Camunda Exporter.

### LINKED ISSUE #24078: Add Archiver configuration to Camunda Exporter configuration
**Description**

The goal is to add a new sub-configuration item in the `ExporterConfiguration` of the Camunda Exporter. The new class should contain the following properties:

- `rolloverEnabled`: controls if we run the archiver or not (failsafe)
- `rolloverDateFormat`: this format will be used to create timed indices. It must correspond to the `rolloverInterval` parameter.
- `elsRolloverDateFormat`: Rollover date format for ELS
- `rolloverInterval`: Interval description for "date histogram" aggregation, which is used to group finished instances.
- `rolloverBatchSize`: Maximum size of items to process in an archiving batch
- `waitPeriodBeforeArchiving`
- `delayBetweenRuns`: In case archiver runs without delays, two subsequent runs may try to process the same process entities (because of Elasticsearch refresh behaviour). In general, it's fine, but there are two side effects: 1. We do the job, that is not needed anymore -> wasting CPU time 2. Metrics will become incorrect, as it's not possible to distinguish such (duplicated) calls from normal ones.

**Open questions**

> [!Note]
> These should be answered before any PR is accepted.

- Assumption is ILM or retention settings were already migrated to the exporter, but this needs to be double checked.
- Do we want to keep the `rollover` prefix for the config?
- How easy should the migration from the old config to the new one be (impact on property names)?

rodrigo-lourenco-lopes:
## Description

This PR adds the Archiver configuration to the Camunda Exporter. These properties were extracted from ArchiverProperties.java .

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/24078

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ConfigValidator.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ConfigValidator.java
@@ -8,6 +8,8 @@
 package io.camunda.exporter.config;
 
 import io.camunda.zeebe.exporter.api.ExporterException;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
@@ -21,8 +23,13 @@ public final class ConfigValidator {
    */
   private static final String PATTERN_MIN_AGE_FORMAT = "^[0-9]+[dhms]$";
 
+  private static final String PATTERN_DATE_INTERVAL_FORMAT = "^(?:[1-9]\\d*)([smhdwMy])$";
+
   private static final Predicate<String> CHECKER_MIN_AGE =
       Pattern.compile(PATTERN_MIN_AGE_FORMAT).asPredicate();
+
+  private static final Predicate<String> CHECK_DATE_INTERVAL =
+      Pattern.compile(PATTERN_DATE_INTERVAL_FORMAT).asPredicate();
 
   private ConfigValidator() {}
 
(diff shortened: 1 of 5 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ConfigValidator.java
npepinpe, line 1: :x: When wrapping exceptions, please add the original exception as a cause (i.e. second argument). This allows us to keep the full stack trace of where the error occurred.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
```diff
@@ -226,4 +237,100 @@ public String toString() {
       return "BulkConfiguration{" + "delay=" + delay + ", size=" + size + '}';
     }
   }
+
+  public static class ArchiverConfiguration {
+    private boolean rolloverEnabled = true;
+    private String rolloverDateFormat = "yyyy-MM-dd";
+    private String elsRolloverDateFormat = "date";
+    private String rolloverInterval;
```
npepinpe, line 245: :question: Why drop the default value here? I can see in the original properties it's "1d"?

npepinpe (approved):
:rocket: Only minor comments, don't think we need another review round (but of course you can request one :slightly_smiling_face:)

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
npepinpe, line 1: :x: Please migrate the tests to `ConfigValidatorTest`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
rodrigo-lourenco-lopes: I had accidentally copied the "no value" from task list instead of operator I think.

PR #24129, opened 2024-10-28
Opensearch IndexMappingProperty bug fix

EuroLew:
## Description

The type definition of a property must contain all fields of that property such as as `index` or any other defined, the `ElasticsearchEngineClient` fixed this by implementing the `propertyToMap` to serialize a `Property` into a `Map`.
However this was forgotten to be done for Opensearch(OS) this would cause schema validation errors for OS.

- [x] OS supports full index mapping property serialization of all fields.
- [x] Modify tests which build `IndexMappingProperty` instances to catch this error case.


mustafadagher (approved):
LGTM!

PR #24136, opened 2024-10-28
feat: source event position is updated during batch processing

### LINKED ISSUE #18379: Add record field that references the command that produced the record
**Description**

In the past, the `sourceRecordPosition` of a record was a pointer from a record to the position of the record whose processing caused it to be appended, i.e. the position of the source of the record.

This was a very useful property, as it allows root causing why a record was appended, i.e. the processing of the command at position x resulted in this record.

With the introduction of [command batch processing](https://github.com/camunda/zeebe/issues/11416), this field now references the record position processed at the start of the command batch processing. This means that all records in a batch refer to the same source record position. It no longer points exactly to the record whose processing caused the record to be appended.

This makes the root-causing much harder because it's possible for a single command to produce several follow-up commands. After processing just a few commands that produced several commands, it is no longer possible to know which command is the source of a specific record.

filipecampos:
## Description
We're updating the value of `currentSourceIndex` in `BufferedProcessingResultBuilder` to point to the actual originating command in the batch of records. This value will then be set in the `sourceIndex` field of the new `RecordBatchEntry` upon its creation, during the batch processing.

## Checklist

- [x] unit test

## Related issues

closes #18379

### DIFF
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/BufferedProcessingResultBuilder.java
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/BufferedProcessingResultBuilder.java
@@ -41,6 +41,7 @@ final class BufferedProcessingResultBuilder implements ProcessingResultBuilder {
   private final RecordBatch mutableRecordBatch;
   private ProcessingResponseImpl processingResponse;
   private final long operationReference;
+  private int currentSourceIndex;
 
   BufferedProcessingResultBuilder(final RecordBatchSizePredicate predicate) {
     this(predicate, operationReferenceNullValue());
@@ -48,14 +49,25 @@ final class BufferedProcessingResultBuilder implements ProcessingResultBuilder {
 
   BufferedProcessingResultBuilder(
       final RecordBatchSizePredicate predicate, final long operationReference) {
+    this(predicate, operationReference, -1);
+  }
+
+  BufferedProcessingResultBuilder(
+      final RecordBatchSizePredicate predicate,
+      final long operationReference,
+      final int currentSourceIndex) {
     mutableRecordBatch = new RecordBatch(predicate);
     this.operationReference = operationReference;
+    this.currentSourceIndex = currentSourceIndex;
+  }
+
+  void setCurrentSourceIndex(final int currentSourceIndex) {
+    this.currentSourceIndex = currentSourceIndex;
   }
 
   @Override
   public Either<RuntimeException, ProcessingResultBuilder> appendRecordReturnEither(
       final long key, final RecordValue value, final RecordMetadata metadata) {
-
     if (operationReference != operationReferenceNullValue()) {
       metadata.operationReference(operationReference);
     }
@@ -69,7 +81,8 @@ final class BufferedProcessingResultBuilder implements ProcessingResultBuilder {
     if (value instanceof final UnifiedRecordValue unifiedRecordValue) {
       final var metadataWithValueType = metadata.valueType(valueType);
       final var either =
-          mutableRecordBatch.appendRecord(key, metadataWithValueType, -1, unifiedRecordValue);
+          mutableRecordBatch.appendRecord(
+              key, metadataWithValueType, currentSourceIndex, unifiedRecordValue);
       if (either.isLeft()) {
         return Either.left(either.getLeft());
       }
@@ -102,7 +115,7 @@ final class BufferedProcessingResultBuilder implements ProcessingResultBuilder {
             .rejectionReason(rejectionReason)
             .valueType(valueType)
             .operationReference(operationReference);
-    final var entry = RecordBatchEntry.createEntry(key, metadata, -1, value);
+    final var entry = RecordBatchEntry.createEntry(key, metadata, currentSourceIndex, value);
     processingResponse = new ProcessingResponseImpl(entry, requestId, requestStreamId);
     return this;
   }
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
@@ -149,6 +149,7 @@ public final class ProcessingStateMachine {
   private long writtenPosition = StreamProcessor.UNSET_POSITION;
   private long lastSuccessfulProcessedRecordPosition = StreamProcessor.UNSET_POSITION;
   private long lastWrittenPosition = StreamProcessor.UNSET_POSITION;
+  private int lastProcessedCommandIndex = -1;
   private int onErrorRetries;
   // Used for processing duration metrics
   private Histogram.Timer processingTimer;
@@ -294,7 +295,6 @@ public final class ProcessingStateMachine {
         processingMetrics.observeCommandCount(processedCommandsCount);
       }
 
-      finalizeCommandProcessing();
       writeRecords();
     } catch (final RecoverableException recoverableException) {
       // recoverable
@@ -340,8 +340,8 @@ public final class ProcessingStateMachine {
    *
    * <p>Should be called after processing or error handling is done.
    */
-  private void finalizeCommandProcessing() {
-    lastProcessedPositionState.markAsProcessed(typedCommand.getPosition());
+  private void finalizeCommandProcessing(final long commandPosition) {
+    lastProcessedPositionState.markAsProcessed(commandPosition);
     processedCommandsCount = 0;
   }
 
@@ -356,7 +356,6 @@ public final class ProcessingStateMachine {
     final var processingResultBuilder =
         new BufferedProcessingResultBuilder(
             logStreamWriter::canWriteEvents, initialCommand.getOperationReference());
-    final var currentSourceIndex = -1;
     var lastProcessingResultSize = 0;
 
     // It might be that we reached the batch size limit during processing a command.
@@ -374,6 +373,7 @@ public final class ProcessingStateMachine {
     while (!pendingCommands.isEmpty() && processedCommandsCount < currentProcessingBatchLimit) {
       final var indexedCommand = pendingCommands.removeFirst();
       final var command = indexedCommand.command();
+      lastProcessedCommandIndex = indexedCommand.index;
 
       currentProcessor =
           recordProcessors.stream()
@@ -388,7 +388,7 @@ public final class ProcessingStateMachine {
           collectBatchProcessingStepResult(
               currentProcessingResult,
               lastProcessingResultSize,
-              // +1 since we already need include the current command in the calculation
+              // +1 since we need to include the current command in the calculation
               pendingCommands.size() + processedCommandsCount + 1,
               currentProcessingBatchLimit);
 
@@ -575,7 +575,6 @@ public final class ProcessingStateMachine {
     pendingWrites = currentProcessingResult.getRecordBatch().entries();
     pendingResponses = currentProcessingResult.getProcessingResponse().stream().toList();
 
-    finalizeCommandProcessing();
     writeRecords();
   }
 
@@ -592,10 +591,6 @@ public final class ProcessingStateMachine {
                   processingException, typedCommand, processingResultBuilder);
           pendingWrites = currentProcessingResult.getRecordBatch().entries();
           pendingResponses = currentProcessingResult.getProcessingResponse().stream().toList();
-          // we need to mark the command as processed, even if the processing failed
-          // otherwise we might replay the events, which have been written during
-          // #onProcessingError again on restart
-          finalizeCommandProcessing();
         });
   }
 
@@ -648,6 +643,10 @@ public final class ProcessingStateMachine {
             // We write various type of records. The positions are always increasing and
             // incremented by 1 for one record (even in a batch), so we can count the amount
             // of written records via the lastWritten and now written position.
+            final var lastProcessedCommandPosition =
+                writtenPosition - pendingWrites.size() + lastProcessedCommandIndex + 1;
+            finalizeCommandProcessing(lastProcessedCommandPosition);
+
             final var amount = writtenPosition - lastWrittenPosition;
             metrics.recordsWritten(amount);
             updateState();
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
@@ -38,6 +38,7 @@ import io.camunda.zeebe.stream.api.records.ExceededBatchRecordSizeException;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
 import io.camunda.zeebe.stream.api.scheduling.ScheduledCommandCache;
 import io.camunda.zeebe.stream.api.state.MutableLastProcessedPositionState;
+import io.camunda.zeebe.stream.impl.ProcessingStateMachine.BatchProcessingStepResult.IndexedCommand;
 import io.camunda.zeebe.stream.impl.metrics.ProcessingMetrics;
 import io.camunda.zeebe.stream.impl.metrics.StreamProcessorMetrics;
 import io.camunda.zeebe.stream.impl.records.RecordValues;
@@ -55,6 +56,7 @@ import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.function.BooleanSupplier;
+import org.agrona.collections.MutableInteger;
 import org.slf4j.Logger;
 
 /**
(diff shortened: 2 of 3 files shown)


zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
```diff
@@ -366,6 +367,10 @@ private void batchProcessing(final TypedRecord<?> initialCommand) {
 
       final var command = pendingCommands.removeFirst();
 
+      if (!command.equals(initialCommand)) {
+        currentSourceIndex = command.getKey();
```
lenaschoenburg, line 371: :x: This is not correct because it mixes two different things: we are interested in the _index_ of this command within the batch. This is different to the record _key_ which, for these purposes, is an arbitrary number unrelated to the index.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
```diff
@@ -427,8 +435,8 @@ private BatchProcessingStepResult collectBatchProcessingStepResult(
                         context.getPartitionId(),
                         entry.recordValue(),
                         entry.recordMetadata()));
-                toWriteEntry = LogAppendEntry.ofProcessed(entry);
               }
+              toWriteEntry = LogAppendEntry.ofProcessed(entry, lastProcessedCommandIndex);
```
lenaschoenburg, line 439: :x: This might work right now but I think it's conceptually wrong. Here, we get an `entry` that might already have some source index set and wrap it to mark it as processed. With your changes, we now overwrite the source index too. But in principle, the engine is free to set its own source index and we should not overwrite it just like that.

I would suggest sticking to what I suggested initially: making the `currentSourceIndex` a property of the `BufferedProcessingResultBuilder`. We update it during batch processing so that the result builder can use this index instead of `-1` if users of that result builder don't provide their own source index.

Does that make sense so far?

lenaschoenburg (changes requested):

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/StreamProcessorTest.java
```diff
@@ -425,12 +425,12 @@ public void shouldProcessFollowUpEventsAndCommands() {
...
       assertThat(followup.getSourceEventPosition())
           .as("Followup records should point to source event")
-          .isEqualTo(firstRecordPosition);
+          .isEqualTo(followup.getKey() == 3 ? 3 : firstRecordPosition);
       final var followupMetadata = new RecordMetadata();
       followup.readMetadata(followupMetadata);
       assertThat(followupMetadata.getOperationReference())
           .as("Followup records should contain the same operation reference of the initial command")
-          .isEqualTo(operationReference);
+          .isNotEqualTo(operationReference);
```
filipecampos, line 433: This makes the test case pass. Do you think it's enough or we should deal with the changes differently?

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
filipecampos: Reverted.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
filipecampos: Reverted the changes to `LogAppendEntry`.

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/StreamProcessorTest.java
```diff
@@ -481,6 +481,77 @@ public void shouldSetSourcePointerForFollowUpRecords() {
...
+
+    final var logStreamReader = streamPlatform.getLogStream().newLogStreamReader();
+    logStreamReader.seekToFirstEvent();
+    final var firstRecord = logStreamReader.next();
+    assertThat(firstRecord.getSourceEventPosition()).isEqualTo(-1);
+
+    while (logStreamReader.hasNext()) {
+      final var record = logStreamReader.next();
+      assertThat(record.getSourceEventPosition()).isEqualTo(record.getKey() < 6 ? 1 : 3);
+    }
```
lenaschoenburg, line 552: :wrench: That's a bit too smart for my liking, I'd rather see the expected source event positions spelled out and checked for each record.

:wrench: This doesn't assert that we see the expected number of records, so the test could pass if there are records missing. This is solved too if we just read the 5 expected records and then assert for each of them that they have the correct source event position.

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/StreamProcessorTest.java
```diff
@@ -481,6 +481,77 @@ public void shouldSetSourcePointerForFollowUpRecords() {
     assertThat(logStreamReader.next().getSourceEventPosition()).isEqualTo(firstRecordPosition);
   }
 
+  @Test
+  public void shouldSetSourcePointerCorrectlyForFollowUpRecords() {
```
lenaschoenburg, line 485: :wrench: Naming wise this is very close to the existing `shouldSetSourcePointerForFollowUpRecords`. I'd suggest to rename it to make it clearer that this is about batch processing. Perhaps something like `shouldTrackSourcePointerDuringBatchProcessing`?

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
```diff
@@ -55,18 +56,19 @@
...
- * +------------------+            +--------------------+
- * |                  |            |                    |      exception
- * | tryToReadNextRecord() |----------->|  processCommand()  |------------------+
- * |                  |            |                    |                  v
- * +------------------+            +--------------------+            +---------------+
+ * +------------------------+            +--------------------+
+ * |                        |            |                    |      exception
+ * |  tryToReadNextRecord() |----------->|  processCommand()  |------------------+
+ * |                        |            |                    |                  v
+ * +------------------------+            +--------------------+      +---------------+
```
lenaschoenburg, line 71: :heart: Thanks for the cleanup!

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/BufferedProcessingResultBuilder.java
```diff
@@ -41,21 +41,33 @@ final class BufferedProcessingResultBuilder implements ProcessingResultBuilder {
...
+  BufferedProcessingResultBuilder(
+      final RecordBatchSizePredicate predicate,
+      final long operationReference,
+      final int currentSourceIndex) {
     mutableRecordBatch = new RecordBatch(predicate);
     this.operationReference = operationReference;
+    this.currentSourceIndex = currentSourceIndex;
+  }
+
+  public void setCurrentSourceIndex(final int currentSourceIndex) {
```
lenaschoenburg, line 64: :wrench: This could be package-private, no? Because we don't want to give everyone the ability to do this, pretty much only the ProcessingStateMachine is allowed.

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/StreamProcessorTest.java
lenaschoenburg: :x: No, this change is unexpected and would break the `operationReference` feature if I understand correctly. Can you check why the `operationReference` is different now?

lenaschoenburg (changes requested):
Thanks for cleaning up the PR @filipecampos and finishing the tests. Please take a look at my comments. It mostly looks good except the unexpected change to `operationReference`.

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/StreamProcessorTest.java
filipecampos: Thanks for the help finding the issue. That change has been reverted.

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/StreamProcessorTest.java
filipecampos: I unrolled the loop then.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/BufferedProcessingResultBuilder.java
filipecampos: Agreed. Removed the public modifier.

lenaschoenburg (commented):
Thanks, you changes look good!

:thought_balloon: There is now a failing test that prevents the merge: `shouldRestoreStateAtEachStepInExecution`. I spent a bit looking into it and it's an interesting case. The test fails because after replaying, the `ReplayStateMachine` now reports a `lastSourceEventPosition` that points at the last processed command in a batch, while the `ProcessingStateMachine` reports `lastSuccessfulProcessedRecordPosition` which always points at the command that started a batch, not at a command within a batch. So the await for "Until all events have been replayed" in `stopAndRestartEngineAndCompareStates` does not complete. Removing that await makes the test pass but to be honest I'm a bit concerned that this semantic difference could cause other problems. I'll have to think about it.

lenaschoenburg:
So I'd treat this test failure as a weakness in the `ProcessingStateMachine`: The `lastSuccessfulProcessedRecordPosition` is not accurate because it is not updated for commands that were processed within a batch. Here is a patch that fixes this and makes the test pass: 


```
Index: zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
===================================================================
diff --git a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
--- a/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java	(revision Staged)
+++ b/zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java	(date 1730997904883)
@@ -202,7 +202,6 @@
     processingFilter =
         new MetadataEventFilter(
                 recordMetadata -> recordMetadata.getRecordType() == RecordType.COMMAND)
-            .and(record -> !record.shouldSkipProcessing())
             .and(context.processingFilter());
     clock = context.getClock();
   }
@@ -243,7 +242,14 @@
       currentRecord = logStreamReader.next();
 
       if (processingFilter.applies(currentRecord)) {
-        processCommand(currentRecord);
+        if (currentRecord.shouldSkipProcessing()) {
+          // we already processed this command, just update the last processed position but
+          // otherwise skip it.
+          lastSuccessfulProcessedRecordPosition = currentRecord.getPosition();
+          skipRecord();
+        } else {
+          processCommand(currentRecord);
+        }
       } else {
         skipRecord();
       }
```

I'd suggest applying this patch and checking if that breaks any other tests. Please let me know if you have any questions.

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/StreamProcessorTest.java
```diff
@@ -481,6 +481,97 @@ public void shouldSetSourcePointerForFollowUpRecords() {
...
+    verify(defaultRecordProcessor, TIMEOUT.times(2)).process(any(), any());
+
+    final var logStreamReader = streamPlatform.getLogStream().newLogStreamReader();
+    logStreamReader.seekToFirstEvent();
+    final var firstRecord = logStreamReader.next();
+    assertThat(firstRecord.getSourceEventPosition()).isEqualTo(-1);
+
+    await("should write follow up events")
+        .untilAsserted(() -> assertThat(logStreamReader.hasNext()).isTrue());
+    assertThat(logStreamReader.hasNext()).isTrue();
```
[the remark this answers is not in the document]
filipecampos: Well caught. Thanks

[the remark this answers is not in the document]
lenaschoenburg: This is a repeating pattern. I see that you removed the first instance but not the following

lenaschoenburg (approved):
Let's go :rocket: 

:thought_balloon: I think the second change to adjust the `ProcessingStateMachine` could have been it's own commit, explaining why that change was necessary.

:thought_balloon: We should announce that change to the other team, I'll follow up with you on that.

lenaschoenburg:
Oh nevermind, the latest change to the `ProcessingStateMachine` causes a problem, as seen in the `shouldNotCreateDefaultUserOnRestart` test :disappointed:

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
```diff
@@ -634,6 +643,13 @@ private void writeRecords() {
             // We write various type of records. The positions are always increasing and
             // incremented by 1 for one record (even in a batch), so we can count the amount
             // of written records via the lastWritten and now written position.
```
lenaschoenburg, line 645: :x: This comment is at the wrong position now, it explains the `amount` calculation we use for metrics.

:wrench: This made me realize, the way we calculate `amount` is not really correct because it doesn't account for records from other sources. Instead, we can just say `metrics.recordsWritten(pendingWrites.size())`

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
```diff
@@ -578,10 +591,6 @@ private void errorHandlingInTransaction(final Throwable processingException) thr
                   processingException, typedCommand, processingResultBuilder);
           pendingWrites = currentProcessingResult.getRecordBatch().entries();
           pendingResponses = currentProcessingResult.getProcessingResponse().stream().toList();
-          // we need to mark the command as processed, even if the processing failed
-          // otherwise we might replay the events, which have been written during
-          // #onProcessingError again on restart
-          finalizeCommandProcessing();
```
lenaschoenburg, line 584: :thought_balloon: This is the only change I'm unsure about. `errorHandlingInTransaction` is called when `updateState` fails to commit the transaction. Afterwards, we just call `updateState` again. I'm afraid there is a preexisting bug here because `errorHandlingInTransaction` produces a new `currentProcessingResult`. But in this code path, nothing ever tries to write this: we don't call `writeRecords`. Because we moved the `finalizeCommandProcessing` call to `writeRecords` we have no made this preexisting bug worse, not only do we not write the results from handling the error, we also don't mark the command as processed.

lenaschoenburg: Actually it doesn't matter for this change because `updateState` is called after we've written and thus called `finalizeCommandProcessing` already. So yes, there might be an old bug where `updateState` is not writing new processing results from error handling in case the transaction can't be committed but that's a different problem.

lenaschoenburg (approved):
:wrench: One last suggestion, I'd rename the PR to something like "feat: source event position is updated during batch processing" or similar. The issue number is in the title is not necessary.

:1st_place_medal: Thanks for your endurance, this was way more complicated than I anticipated :rocket:

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
lenaschoenburg: I've discussed this with @Zelldon and we both came to the conclusion that this is an old bug (that might never be triggered though). I reported it here: https://github.com/camunda/camunda/issues/24675

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
filipecampos: Placed the comment in the right place again and changed `amount` and also changed the argument for `metrics.recordsWritten`.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ProcessingStateMachine.java
filipecampos: OK. Leaving this as is then.

Issue #23383, opened 2024-10-11
Incident recovery for Execution Listeners does not work when retries are exhausted


kristinkomschow:
Part of #21974 epic.

### Describe the bug

Operate currently does not allow for resolving an incident in an Execution Listener that has run out of retries. The operation will always fail, due to the [code that updates the number of retries](https://github.com/camunda/camunda/blob/668c60f199598ee3f2758e238f682ac91f028b05/operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation/ResolveIncidentHandler.java#L44-L46) being skipped.

This happens because Execution Listener incidents have an own error type "[EXECUTION_LISTENER_NO_RETRIES](https://github.com/camunda/camunda/blob/36b799887aff7aaa11abc0f5fb57265d52e291c6/zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/ErrorType.java#L25)".

Needs backport to 8.6

Workarounds:

* Change the job retries via API
* Change the error type in the incident's Elasticsearch document from `UNKNOWN` to `JOB_NO_RETRIES`

### To Reproduce

1. Fail an Execution Listener job with no retries left
2. Attempt to resolve the incident in the operate incident view
3. The retry will always fail again

### Current behavior

Incident resolution for Execution Listener jobs with no retries always fails due to no retries left.

### Expected behavior

Operate updates the amount of retries before attempting the incident resolution.

### Environment

- OS: [e.g. MacOS]
- Browser: [e.g. chrome, safari]
- Operate Version: [e.g. 8.1]

### Additional context

Report of the bug in slack: https://camunda.slack.com/archives/C06CNGWTBAS/p1728461807862369
Discussion about bug details and how we want to fix it: https://camunda.slack.com/archives/C06CNGWTBAS/p1728461829921679

Pull requests:

* Main: https://github.com/camunda/camunda/pull/24140
* 8.6: https://github.com/camunda/camunda/pull/24162

---------------------------------------------------------------------------------------------

### Acceptance Criteria

- [ ] Incident resolution for Execution Listeners with no more retries updates the number of retries before attempting a retry
- [ ] The solution contains an end-to-end test that checks this behavior

### Definition of Ready - Checklist

- [ ] The bug has been reproduced by the assignee in the environment compatible with the provided one; otherwise, the issue is closed with a comment
- [ ] The issue has a meaningful title, description, and testable acceptance criteria
[...]

PR #24127, opened 2024-10-28
feat: added Task Listener no retries error type

kristinkomschow:
## Description

Added error types described for task listeners in the specification.
`EXTRACT_VALUE_ERROR` already existed beforehand, so it didn't need to be added.

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes https://github.com/camunda/camunda/issues/23945

PR #24140, opened 2024-10-28
fix: retry job on execution listener incident

ThorbenLindhauer:
- when Operate resolves an incident that was raised in an execution listener, it should update retries

related to #23383

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/ErrorType.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/ErrorType.java
@@ -14,7 +14,8 @@ public enum ErrorType {
   UNSPECIFIED("Unspecified"),
   UNKNOWN("Unknown"),
   IO_MAPPING_ERROR("I/O mapping error"),
-  JOB_NO_RETRIES("No more retries left"),
+  JOB_NO_RETRIES("No more retries left", true),
+  EXECUTION_LISTENER_NO_RETRIES("Execution Listener no more retries left", true),
   CONDITION_ERROR("Condition error"),
   EXTRACT_VALUE_ERROR("Extract value error"),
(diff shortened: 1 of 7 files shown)


kristinkomschow (approved):
I really like this implementation. Thank you!

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/ErrorType.java
```diff
@@ -14,7 +14,8 @@ public enum ErrorType {
   UNSPECIFIED("Unspecified"),
   UNKNOWN("Unknown"),
   IO_MAPPING_ERROR("I/O mapping error"),
-  JOB_NO_RETRIES("No more retries left"),
+  JOB_NO_RETRIES("No more retries left", true),
+  EXECUTION_LISTENER_NO_RETRIES("No more retries left", true),
```
ThorbenLindhauer, line 18: @kristinkomschow one thought about this line, after I saw https://github.com/camunda/camunda/pull/24127: Should we have a different message here, e.g. `Execution Listener no more retries left`? I'm not sure what the significance of it is. Do we display it in the frontend?

kristinkomschow: In the figma prototype there was no distinction between the message for Listeners or normal Jobs, but I find it logical to add "Execution Listener" to the message if we have a separate error type for it.
And yes, it gets displayed in the frontend as "Incident Type".

ThorbenLindhauer: I adjusted the wording both here and on the maintenance branch. I also rebased because there were concurrent changes in `ResolveIncidentHandler`. I will let the CI run and then merge it tomorrow.

PR #24144, opened 2024-10-29
feat: apply search operation in VariableFilter

### LINKED ISSUE #23989: Define filter operations for `VariableFilter`
## Description

In order to support advanced search queries for `/variables` we need to prepare the internal structure to support different operations and execute them properly.

## Acceptance criteria

- Extend the required `VariableFilter` properties with operations.
- Make sure underlying structures work with the new structure.
  - `SearchQueryRequestMapper`, `*FilterTransformer`, RDBMS mappers etc.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/24144
```

danielkelemen:
## Description

- Applies search operations in VariableFilter.
- Refactor filter methods to contain at least one parameter and varargs.
- Added common class FilterUtil.

## Related issues

closes https://github.com/camunda/camunda/issues/23989

### DIFF
--- a/search/search-domain/src/main/java/io/camunda/search/filter/ProcessInstanceFilter.java
+++ b/search/search-domain/src/main/java/io/camunda/search/filter/ProcessInstanceFilter.java
@@ -9,13 +9,12 @@ package io.camunda.search.filter;
 
 import static io.camunda.util.CollectionUtil.*;
 
+import io.camunda.util.FilterUtil;
 import io.camunda.util.ObjectBuilder;
 import java.time.OffsetDateTime;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.stream.Collectors;
 
 public record ProcessInstanceFilter(
     List<Operation<Long>> processInstanceKeyOperations,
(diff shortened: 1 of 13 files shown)


search/search-domain/src/main/java/io/camunda/search/filter/ProcessInstanceFilter.java
```diff
@@ -51,37 +50,34 @@ public static final class Builder implements ObjectBuilder<ProcessInstanceFilter
...
-    }
-
     public Builder processInstanceKeyOperations(final List<Operation<Long>> operations) {
       processInstanceKeyOperations = addValuesToList(processInstanceKeyOperations, operations);
       return this;
     }
 
-    public Builder processInstanceKeys(final Long... values) {
-      return processInstanceKeyOperations(mapToOperationEq(values));
+    public Builder processInstanceKeys(final Long value, final Long... values) {
```
tmetzke, line 58: ❓ What is the benefit of using this method signature compared to before?

danielkelemen: This way it needs to be called with at least one parameter. It can't be used like `processInstanceKeys()`, which wouldn't make sense.

tmetzke: I see, yeah, that makes sense, compile-time clarity on how you can use the API. Sure 👍

tmetzke (approved):

PR #24155, opened 2024-10-29
Add MappingCreateProcessor

### LINKED ISSUE #23839: Process `Mapping.CREATE` command
<h2>Description</h2>

image

1. Create new class `MappingCreateProcessor` and implement the `DistributedTypedRecordProcessor<MappingRecord>` interface
2. Register the processor in the `EngineProcessor` (preferably via a new class similar to the `UserProcessors`)
3. In the `processNewCommand method:
    - Check if the user is authorized to create Mappings
        - Resource type: `MAPPING_RULE`
        - Permission type: `CREATE`
     - Check if the Mapping already exists. If yes, reject the command with `ALREADY_EXISTS`. If no continue.
     - Write a `Mapping.CREATED` event and respond to the client
    - Write a `Mapping.CREATED` event
5. Unit test the `Mapping.CREATE` command
6. Unit test the `Mapping.CREATE` command distribution.

nicpuppa:
## Description

This PR adds the `MappingCreateProcessor` 

## Related issues

closes #23839

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/CreateMappingMultiPartitionTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/CreateMappingMultiPartitionTest.java
@@ -17,6 +17,7 @@ import io.camunda.zeebe.protocol.record.RecordType;
 import io.camunda.zeebe.protocol.record.ValueType;
 import io.camunda.zeebe.protocol.record.intent.CommandDistributionIntent;
 import io.camunda.zeebe.protocol.record.intent.MappingIntent;
+import io.camunda.zeebe.protocol.record.intent.RoleIntent;
 import io.camunda.zeebe.protocol.record.value.CommandDistributionRecordValue;
 import io.camunda.zeebe.test.util.record.RecordingExporter;
 import io.camunda.zeebe.test.util.record.RecordingExporterTestWatcher;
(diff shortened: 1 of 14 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/CreateMappingMultiPartitionTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Test
+  public void distributionShouldNotOvertakeOtherCommandsInSameQueue() {
+    // given the user creation distribution is intercepted
+    for (int partitionId = 2; partitionId <= PARTITION_COUNT; partitionId++) {
+      interceptMappingCreateForPartition(partitionId);
+    }
+    final var claimName = UUID.randomUUID().toString();
+    final var claimValue = UUID.randomUUID().toString();
+    engine.mapping().newMapping(claimName).withClaimValue(claimValue).create();
```
remcowesterhoud, line 108: ❌ You're not testing if commands don't overtake each other here. You're only sending one command, so all that's being tested is the redistribution of that command. To test that stuff doesn't overtake each other you'll have to send another distributed command on the same queue, with a different value type or intent. Then you can guarantee it's ordered.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/MappingTest.java
```diff
@@ -0,0 +1,70 @@
...
+    // when
+    final var duplicatedMappingRecord =
+        engine
+            .mapping()
+            .newMapping(claimName)
+            .withClaimValue(claimValue)
+            .expectRejection()
+            .create();
+
+    final var createdRole = mappingRecord.getValue();
```
remcowesterhoud, line 57: ❌ It's a mapping, not a role 😄 But really it's fine to remove this create assertion. That's what the other test case is for. We can limit this case to the rejection.

remcowesterhoud (dismissed):
Nice work, just some commens in the tests.

koevskinikola (approved):
LGTM now 👍

PR #24160, opened 2024-10-29
feat: add ui for user's authorizations

### LINKED ISSUE #23794: add UI to show list of authorization of a user

maryarm:
## Description

Add ui to show authorizations in UserDetail page

## Related issues

closes #23794


identity/client/src/utility/api/users/index.ts
```diff
@@ -1,10 +1,11 @@
...
+import { ApiDefinition, apiDelete, apiPost, apiPut } from "../request";
 import { SearchResponse } from "src/utility/api";
+import { EntityData } from "src/components/entityList/EntityList";
 
 export const USERS_ENDPOINT = "/users";
 
-export type User = {
-  id: string;
+export type User = EntityData & {
+  key: number;
```
pedesen, line 8: This makes `id` an optional property of User (because it's optional in the EntityData type). And it makes it necessary to use the non-null assertion operator (`!`) a lot on `user.id!`. 

Is this intentional or can `id` EntityData be made non-optional?

pedesen, line 8: Also I don't fully understand the difference between id, key and username. Maybe you can elaborate a bit more on this.

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <>
+      <EntityList
+        title={t("Authorizations assigned to user")}
+        data={authorizations == null ? [] : authorizations.items}
+        headers={[{ header: t("ResourceType"), key: "resourceType" }]}
+        onEntityClick={showAuthorizationDetails}
+        addEntityLabel="Assign authorization"
+        onAddEntity={() => {}}
```
pedesen, line 56: In my opinion it would be a bit cleaner to make `onAddEntity` optional on the EntityList component, because then you don't need to pass an empty function. What do you think?

identity/client/src/pages/users/detail/index.tsx
```diff
@@ -21,24 +21,25 @@ import Flex from "src/components/layout/Flex";
...
   } = useApi(getUserDetails, {
-    id,
+    username: id,
   });
   const [editUser, editUserModal] = useEntityModal(EditModal, reload);
   const [deleteUser, deleteUserModal] = useEntityModal(DeleteModal, () =>
     navigate("..", { replace: true }),
   );
-
+  const user = userSearchResults != null ? userSearchResults.items[0] : null;
```
pedesen, line 42: It's a good practice to use the strict equality operator (`!==`) here, because it checks if userSearchResults is of type null and does not do any further conversion of it.

```suggestion
  const user = userSearchResults !== null ? userSearchResults.items[0] : null;
```

identity/client/src/utility/api/users/index.ts
maryarm: username is the name chosen by user and we use it for login and also in search api. id and key can be used interchangeably for now (only their type is different) but may be later (when implementation of all entities is done) we can revise and make them one.

maryarm: I think id is optional because when entity is not created yet (stored in backend) the value is null.

identity/client/src/pages/users/detail/authorization/List.tsx
maryarm: makes. sense but as in in next tasks, I have to add functions to manage onAddEntity, I thinks it's not urgently required.

identity/client/src/utility/api/users/index.ts
pedesen: Okay, but in the places you use `user.id!` it is guaranteed that it's not null? Otherwise there might be unexpected behavior which can't be caught by typescript.

maryarm: you are right and as I understand from the flow it's available in this place as it loads a detail of an existing user.

pedesen (approved):

PR #24166, opened 2024-10-29
feat: apply search operation in UserTaskFilter

### LINKED ISSUE #23990: Define filter operations for `UserTaskFilter`
## Description

In order to support advanced search queries for `/user-tasks` we need to prepare the internal structure to support different operations and execute them properly.

## Acceptance criteria

- Extend the required `UserTaskFilter ` properties with operations.
- Make sure underlying structures work with the new structure.
  - `SearchQueryRequestMapper`, `*FilterTransformer`, RDBMS mappers etc.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/24166
```

danielkelemen:
## Description

- Applies search operations in UserTaskFilter.

## Related issues

closes https://github.com/camunda/camunda/issues/23990

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
@@ -11,15 +11,20 @@ import static io.camunda.search.clients.query.SearchQueryBuilders.and;
 import static io.camunda.search.clients.query.SearchQueryBuilders.exists;
 import static io.camunda.search.clients.query.SearchQueryBuilders.hasChildQuery;
 import static io.camunda.search.clients.query.SearchQueryBuilders.hasParentQuery;
+import static io.camunda.search.clients.query.SearchQueryBuilders.intOperations;
 import static io.camunda.search.clients.query.SearchQueryBuilders.longTerms;
 import static io.camunda.search.clients.query.SearchQueryBuilders.not;
 import static io.camunda.search.clients.query.SearchQueryBuilders.or;
+import static io.camunda.search.clients.query.SearchQueryBuilders.stringOperations;
 import static io.camunda.search.clients.query.SearchQueryBuilders.stringTerms;
+import static java.util.Optional.ofNullable;
 
 import io.camunda.search.clients.query.SearchQuery;
 import io.camunda.search.clients.transformers.ServiceTransformers;
+import io.camunda.search.filter.Operation;
 import io.camunda.search.filter.UserTaskFilter;
 import io.camunda.search.filter.VariableValueFilter;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
```diff
@@ -81,22 +83,11 @@ public SearchQuery toSearchQuery(final UserTaskFilter filter) {
               not(hasChildQuery("taskVariable", taskVarNameQuery)));
 
       // Combine taskVariable, processVariable, and subprocessVariable queries with OR logic
-      variablesQuery = or(taskVariableQuery, processVariableCondition);
+      queries.add(or(taskVariableQuery, processVariableCondition));
     }
+    ofNullable(exists("flowNodeInstanceId")).ifPresent(queries::add); // Default to task
```
tmetzke, line 88: 🔧 I think you can add this search condition directly, right? Without the nullable part of it.
```suggestion
    queries.add(exists("flowNodeInstanceId")); // Default to task
```

ralfpuchert (commented):
Do we have tests for the changes?

search/search-domain/src/main/java/io/camunda/search/filter/UserTaskFilter.java
```diff
@@ -8,23 +8,26 @@
...
 import static io.camunda.util.CollectionUtil.collectValuesAsList;
 
+import io.camunda.util.FilterUtil;
 import io.camunda.util.ObjectBuilder;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
-public final record UserTaskFilter(
+public record UserTaskFilter(
```
tmetzke, line 20: 💭 Same as in https://github.com/camunda/camunda/pull/24172, we might as well add operation awareness to all fields (except `boolean`s) already.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
```diff
@@ -119,16 +110,20 @@ private SearchQuery getUserTaskKeysQuery(final List<Long> userTaskKeys) {
...
-  private SearchQuery getCandidateGroupsQuery(final List<String> candidateGroups) {
-    return stringTerms("candidateGroups", candidateGroups);
+  private List<SearchQuery> getAssigneesQuery(final List<Operation<String>> assignees) {
+    return stringOperations("assignee", assignees);
   }
 
-  private SearchQuery getAssigneesQuery(final List<String> assignee) {
-    return stringTerms("assignee", assignee);
+  private List<SearchQuery> getPrioritiesQuery(final List<Operation<Integer>> priorities) {
+    return intOperations("priority", priorities);
```
tmetzke, line 126: ❓ This is a new filter option that is untested, right? The other fields are already covered by existing tests and don't change so the existing tests should verify that we don't introduce any regressions with this, right?

tmetzke (approved):
👍 Generally looks fine
🔧 There some minor issues to cover, e.g. adding operations to all fields and adding transformer tests for the priority filter you added.

Pre-approving already since structurally this is fine and could also be followed up on.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
danielkelemen: Yes, it should be tested too. I will add tests.

search/search-domain/src/main/java/io/camunda/search/filter/UserTaskFilter.java
danielkelemen: I'll add them in a follow-up ticket.

PR #24168, opened 2024-10-29
Operate's OperationHandlers propagate `operationreference` with zeebe commands

### LINKED ISSUE #24028: Update all operate's OperationHandlers to send the `operationReference` along with the underlying zeebe commands
Update all operate's [OperationHandlers](https://github.com/camunda/camunda/tree/main/operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation) to send the `operationReference` along with the underlying zeebe commands.

Below are related decisions we made and documented earlier on a [parent issue comment](https://github.com/camunda/camunda/issues/18372#issuecomment-2435662446).

**❓ Where to get the operation reference when sending it to zeebe client?**
We can rely on [OperationEntity](https://github.com/camunda/camunda/blob/main/operate/schema/src/main/java/io/camunda/operate/entities/OperationEntity.java) id right now and change [generateId()](https://github.com/camunda/camunda/blob/a60cdd2579ff109fa19632536eb64ad1f97ec2a1/operate/schema/src/main/java/io/camunda/operate/entities/OperationEntity.java#L215-L217) to generate a long value instead of a UUID.

**❓ How to generate this operation reference and ensure its uniqueness?**
We can rely on uuid.getMostSignificantBits() for now. 100% uniqueness is not a great deal for operate, specially when it comes to operation, and the rare chance of collision using uuid.getMostSignificantBits() can/shall be tolerated.

**❓ Now that we are changing the id to become a long instead a UUID, what happens if an operation gets schedule before updating to the new version, and then parsing the wrong format Id fails?**
We can ignore the operations we fail to parse their Ids. It's a rare case and will happen only for a temporary period during migration

mustafadagher:
## Description
Updated all operate's [OperationHandlers](https://github.com/camunda/camunda/tree/main/operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation) to send the `operationReference` along with the underlying zeebe commands

- Set `OperationEntity` and `BatchOperationEntity` Ids to `UUID.mostSignificantBits`
- Refactor: call `generateId()` in `OperationEntity` and `BatchOperationEntity` constructors. Instead of generating the Id as a separate expression after each instantiation
- Propagate `OperationReference` with all Zeebe client calls
- Handles parsing `operation.id` to `operationReference` gracefully, and ignores setting it if `Invalid Format`

## Related issues

closes #24028

### DIFF
--- a/operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation/AbstractOperationHandler.java
+++ b/operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation/AbstractOperationHandler.java
@@ -150,7 +150,9 @@ public abstract class AbstractOperationHandler implements OperationHandler {
       final long operationReference = Long.parseLong(id);
       command.operationReference(operationReference);
     } catch (final NumberFormatException e) {
-      // ignore
+      LOGGER.debug(
+          "The operation reference provided is not a number: {}. Ignoring propagating it to zeebe commands.",
+          id);
     }
     return command;
   }
--- a/operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation/AbstractOperationHandler.java
+++ b/operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation/AbstractOperationHandler.java
@@ -15,6 +15,7 @@ import io.camunda.operate.webapp.writer.BatchOperationWriter;
 import io.camunda.webapps.schema.entities.operation.OperationEntity;
 import io.camunda.webapps.schema.entities.operation.OperationState;
 import io.camunda.zeebe.client.ZeebeClient;
+import io.camunda.zeebe.client.api.command.CommandWithOperationReferenceStep;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import java.util.Arrays;
@@ -141,5 +142,16 @@ public abstract class AbstractOperationHandler implements OperationHandler {
       LOGGER.debug("Operation {} was sent to Zeebe", operation.getId());
     }
     recordCommandMetric(operation);
+  }
+
+  protected static <T extends CommandWithOperationReferenceStep<T>> T withOperationReference(
+      final T command, final String id) {
+    try {
+      final long operationReference = Long.parseLong(id);
+      command.operationReference(operationReference);
+    } catch (final NumberFormatException e) {
+      // ignore
+    }
+    return command;
   }
 }
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operation/BatchOperationEntity.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operation/BatchOperationEntity.java
@@ -112,7 +112,9 @@ public class BatchOperationEntity extends AbstractExporterEntity<BatchOperationE
   }
 
   public void generateId() {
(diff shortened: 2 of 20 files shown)


webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operation/BatchOperationEntity.java
```diff
@@ -108,7 +108,8 @@ public BatchOperationEntity setSortValues(final Object[] sortValues) {
   }
 
   public void generateId() {
-    setId(UUID.randomUUID().toString());
+    final long operationReference = UUID.randomUUID().getMostSignificantBits();
```
sdorokhova: Why do we change also `BatchOperationEntity` id? It's not necessary for our purpose, or?

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operation/BatchOperationEntity.java
```diff
@@ -26,6 +26,10 @@ public class BatchOperationEntity extends AbstractExporterEntity<BatchOperationE
 
   @JsonIgnore private Object[] sortValues;
 
+  public BatchOperationEntity() {
+    generateId();
```
sdorokhova, line 30: I feel bad about putting id generation in constructor, as default constructor is also used by Jackson when mapping json data to an object. It must then call setter to set `id` from json, but still I don't feel good about this. (Just a concern, please decide yourself)

operate/webapp/src/main/java/io/camunda/operate/webapp/zeebe/operation/AbstractOperationHandler.java
```diff
@@ -142,4 +143,15 @@ protected void markAsSent(final OperationEntity operation, final Long zeebeComma
...
     recordCommandMetric(operation);
   }
+
+  protected static <T extends CommandWithOperationReferenceStep<T>> T withOperationReference(
+      final T command, final String id) {
+    try {
+      final long operationReference = Long.parseLong(id);
+      command.operationReference(operationReference);
+    } catch (final NumberFormatException e) {
+      // ignore
```
sdorokhova, line 153: Could we put some logging here please? May be DEBUG level.

sdorokhova (approved):
Looks good to me!
Added some minor comments in code.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operation/BatchOperationEntity.java
sdorokhova: Just to give some more context here: `batchOperationId` is smth visible to the users in UI and also smth that will be stored in `list-view` index for the purpose of filtering process instances by this value. It feels safe, but I would prefer to keep changes smaller not to touch broader functionality. What do you think?

mustafadagher: I can revert this change. I just thought they were very similar and it felt natural to change both to be alike. but will leave this alone

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operation/BatchOperationEntity.java
mustafadagher: hmmm good point. I'll try to think of a different way to do it.

Issue #24164, opened 2024-10-29
Define filter operations for `DecisionInstanceFilter`


danielkelemen:
## Description

In order to support advanced search queries for `/decision-instances` we need to prepare the internal structure to support different operations and execute them properly.

## Acceptance criteria

- Extend the required `DecisionInstanceFilter` properties with operations.
- Make sure underlying structures work with the new structure.
  - `SearchQueryRequestMapper`, `*FilterTransformer`, RDBMS mappers etc.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/24172
```

PR #24172, opened 2024-10-29
feat: apply search operation in DecisionInstanceFilter

danielkelemen:
## Description

- Applies search operations in DecisionInstanceFilter.

## Related issues

closes https://github.com/camunda/camunda/issues/24164

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/DecisionQueryTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/DecisionQueryTest.java
@@ -28,6 +28,8 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.time.Duration;
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -565,6 +567,29 @@ class DecisionQueryTest {
 
     // then
     assertThat(result.items().size()).isEqualTo(3);
+  }
+
+  @Test
+  public void shouldRetrieveDecisionInstanceByEvaluationDate() {
+    // given
+    final var allResult =
+        zeebeClient.newDecisionInstanceQuery().page(p -> p.limit(1)).send().join();
+    final var di = allResult.items().getFirst();
+
+    // when
+    final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSXX");
+    final var offsetDateTime = OffsetDateTime.parse(di.getEvaluationDate(), formatter);
+    final var result =
+        zeebeClient
+            .newDecisionInstanceQuery()
+            .filter(f -> f.evaluationDate(offsetDateTime.toString()))
+            .send()
+            .join();
+
+    // then
+    assertThat(result.items()).hasSize(1);
+    assertThat(result.items().getFirst().getDecisionInstanceKey())
+        .isEqualTo(di.getDecisionInstanceKey());
   }
 
   @Test
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
@@ -87,7 +87,7 @@ public final class DecisionInstanceFilterTransformer
     final var stringOperations =
         decisionDefinitionKeyOperations.stream()
             .map(op -> new Operation<>(op.operator(), String.valueOf(op.value())))
-            .collect(Collectors.toList());
+            .toList();
     return stringOperations("decisionDefinitionId", stringOperations);
   }
 
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
@@ -23,7 +23,6 @@ import io.camunda.search.filter.Operation;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 
 public final class DecisionInstanceFilterTransformer
     implements FilterTransformer<DecisionInstanceFilter> {
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
@@ -8,59 +8,48 @@
 package io.camunda.search.clients.transformers.filter;
 
 import static io.camunda.search.clients.query.SearchQueryBuilders.and;
+import static io.camunda.search.clients.query.SearchQueryBuilders.dateTimeOperations;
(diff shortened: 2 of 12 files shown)


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
```diff
@@ -96,12 +81,13 @@ private SearchQuery getProcessInstanceKeysQuery(final List<Long> processInstance
...
-        "decisionDefinitionId",
-        decisionDefinitionKeys != null
-            ? decisionDefinitionKeys.stream().map(String::valueOf).toList()
-            : null);
+  private List<SearchQuery> getDecisionDefinitionKeysQuery(
+      final List<Operation<Long>> decisionDefinitionKeyOperations) {
+    final var stringOperations =
+        decisionDefinitionKeyOperations.stream()
+            .map(op -> new Operation<>(op.operator(), String.valueOf(op.value())))
+            .collect(Collectors.toList());
```
tmetzke, line 89: 🙃 You can abbreviate this to `.toList()` (since Java 9 or 11 or something)
```suggestion
            .toList();
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -182,6 +182,10 @@ private static DecisionInstanceFilter toDecisionInstanceFilter(
           .map(s -> convertEnum(s, DecisionInstanceState.class))
           .ifPresent(builder::states);
       ofNullable(filter.getEvaluationFailure()).ifPresent(builder::evaluationFailures);
+      ofNullable(filter.getEvaluationDate())
+          .map(SearchQueryRequestMapper::toOffsetDateTime)
+          .map(f -> List.of(Operation.gte(f), Operation.lt(f)))
+          .ifPresent(builder::evaluationDateOperations);
```
tmetzke, line 188: 🔧 Let's maybe treat this as an equals for now? Is that even possible? 🤔

tmetzke (approved):
👍 Looks good to me.
🔧 It's an intermediate state for the evaluation date filter field, so it's not super important how we treat it for now. I made a non-blocking suggestion there.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
danielkelemen: It should be possible, I'm changing it. This field is currently not tested and will change with next iteration, I'll also add tests there.

search/search-domain/src/main/java/io/camunda/search/filter/DecisionInstanceFilter.java
```diff
@@ -8,23 +8,26 @@
...
 import io.camunda.search.entities.DecisionInstanceEntity.DecisionDefinitionType;
 import io.camunda.search.entities.DecisionInstanceEntity.DecisionInstanceState;
+import io.camunda.util.FilterUtil;
 import io.camunda.util.ObjectBuilder;
+import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
 public record DecisionInstanceFilter(
```
tmetzke, line 23: 💭 I was just wondering: should we update all attributes to be "operation-ready" already? That makes it a lot easier in the future to simply add Advanced Search capabilities to all attributes (can even be done step by step for chunks of attributes in the REST request mapper).

tmetzke: That is also what we did for the process instance filter.

danielkelemen: We could do it, it's not a huge effort anymore as we have all the search operation query builder methods ready.
Are there any fields where operations don't make sense? Or can I add it for _all_ of them?

tmetzke: I think only `boolean` fields in general don't get advanced search, I think.

qa/integration-tests/src/test/java/io/camunda/it/client/DecisionQueryTest.java
```diff
@@ -567,6 +569,29 @@ public void shouldRetrieveDecisionInstanceByStateAndType() {
...
+        zeebeClient.newDecisionInstanceQuery().page(p -> p.limit(1)).send().join();
+    final var di = allResult.items().getFirst();
+
+    // when
+    final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSXX");
+    final var offsetDateTime = OffsetDateTime.parse(di.getEvaluationDate(), formatter);
+    final var result =
+        zeebeClient
+            .newDecisionInstanceQuery()
+            .filter(f -> f.evaluationDate(offsetDateTime.toString()))
```
danielkelemen, line 585: 💬 @tmetzke, I added a QA test for the new field `evaluationDate` and noticed that our REST api uses the `OffsetDateTime` to parse dates which accepts offsets with colon only (`2024-06-05T08:29:15.027+00:00`) but the API returns results without colon (`2024-06-05T08:29:15.027+0000`).

Looks like with colon is correct:
[OpenAPI `date-time`](https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings) uses the [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) and that  defines the offset with colon (or just `Z`):
```
time-numoffset  = ("+" / "-") time-hour ":" time-minute
```

It would be probably nicer to return dates in the same format as we accept or accept dates with offset without colon, too.

tmetzke: Interesting find, I agree that it should be consistent. Can you create a follow-up issue so we can collect and find out where we return dates without offset and why?

tmetzke: The `di.getEvaluationDate()` would ideally yield an `OffsetDateTime` and not a String so we can work with the date type in the controllers as mappers. The REST response model should ideally also require an OffsetDateTime object anyway.

The offset date time without colon might just be the String we receive from ElasticSearch? We should maybe accommodate for that in the transformers so the REST search layer receives a date type and doesn't have to care about the different datetime strings from different database backends.

danielkelemen: Yes, I think it just comes from ES right now.
Created a follow-up ticket: https://github.com/camunda/camunda/issues/24218

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionInstanceFilterTransformer.java
```diff
@@ -8,59 +8,48 @@
...
-import io.camunda.search.clients.transformers.filter.DateValueFilterTransformer.DateFieldFilter;
 import io.camunda.search.entities.DecisionInstanceEntity.DecisionDefinitionType;
 import io.camunda.search.entities.DecisionInstanceEntity.DecisionInstanceState;
-import io.camunda.search.filter.DateValueFilter;
 import io.camunda.search.filter.DecisionInstanceFilter;
+import io.camunda.search.filter.Operation;
+import java.time.OffsetDateTime;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
```
tmetzke, line 26: 🔧 Needs to be removed after `toList()` is used directly.
```suggestion
import java.util.stream.Collectors;
```

PR #24180, opened 2024-10-29
Update process instance entity with properties from ProcessCache

### LINKED ISSUE #24179: Update and use ProcessCache in CamundaExporter
- [ ] Update cache when exporting PROCESS record
- [ ] Use process cache to update `ProcessInstanceForListViewEntity` when exporting PROCESS_INSTANCE record.

deepthidevaki:
## Description

* `ProcessCache` is instantiated per exporter (per partition)
* Process cache is updated when exporting PROCESS record
* Process name and versionTag in the process instance entitity is updated from the values from the cache. One special case here is what if process do not exist in the backend and thus there is no entry in the cache for that process. This should not happen normally. The only case when this happens is the process was explicitly deleted from the backend or the backend suffered a data loss. In this case, we do not want to block exporting. So instead of throwing an exception, we can deliberately choose a behavior. In this PR, we chose the `bpmnProcessId` if the name is not found. This is similar to what was done in the Operate Importer. Alternative is to not set the `processName`. However, this might have impact on other components, especially search api. 

## Related issues

closes #24179

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
@@ -9,6 +9,8 @@ package io.camunda.exporter;
 
 import static java.util.Map.entry;
 
+import io.camunda.exporter.adapters.ClientAdapter;
+import io.camunda.exporter.cache.ProcessCacheImpl;
 import io.camunda.exporter.config.ConnectionTypes;
 import io.camunda.exporter.config.ExporterConfiguration;
 import io.camunda.exporter.handlers.AuthorizationHandler;
@@ -81,7 +83,7 @@ public class DefaultExporterResourceProvider implements ExporterResourceProvider
   private Set<ExportHandler> exportHandlers;
 
   @Override
-  public void init(final ExporterConfiguration configuration) {
+  public void init(final ExporterConfiguration configuration, final ClientAdapter clientAdapter) {
     final var globalPrefix = configuration.getIndex().getPrefix();
     final var isElasticsearch =
         ConnectionTypes.from(configuration.getConnect().getType())
@@ -130,6 +132,12 @@ public class DefaultExporterResourceProvider implements ExporterResourceProvider
             AuthorizationIndex.class,
             new AuthorizationIndex(globalPrefix, isElasticsearch));
 
+    final var processCache =
+        new ProcessCacheImpl(
+            10000,
+            clientAdapter.getProcessCacheLoader(
+                indexDescriptorsMap.get(ProcessIndex.class).getFullQualifiedName()));
+
     exportHandlers =
         Set.of(
             new UserHandler(indexDescriptorsMap.get(UserIndex.class).getFullQualifiedName()),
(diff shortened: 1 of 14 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
```diff
@@ -16,7 +17,7 @@
 
 public interface ExporterResourceProvider {
 
-  void init(ExporterConfiguration configuration);
+  void init(ExporterConfiguration configuration, final ClientAdapter clientAdapter);
```
Zelldon, line 20: ❓ We actually just pass in here the client Adapter because at the end we want a cache loader. We had to add the cache loader to the ClientAdapter even. 

Couldn't we instead just pass in here the CacheLoader? And when calling the method we give a representive instantiation?

ChrisKujawa (commented):
Thanks @deepthidevaki 👍🏼 

I have two smaller things we should look at other than that looks good 👍🏼

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
```diff
@@ -123,6 +125,12 @@ public void init(final ExporterConfiguration configuration) {
             AuthorizationIndex.class,
             new AuthorizationIndex(globalPrefix, isElasticsearch));
 
+    final var processCache =
+        new ProcessCacheImpl(
+            10000,
```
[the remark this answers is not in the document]
deepthidevaki: There is a follow up issue for configurations.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
deepthidevaki: It was not possible because the cache loader needs the process index name, which is set only in the provider.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
```diff
@@ -320,16 +325,24 @@ private void completeOperation(final Record<ProcessInstanceRecordValue> record)
...
+            .get(processDefinitionKey)
+            .map(CachedProcessEntity::name)
+            // If the cache does not contain the process definition then the process has been
+            // deleted from the backend. This is a special case which can happen only if there was a
+            // data loss in the backend. In that case, inorder to not block the exporter, we can
+            // return the bpmnProcessId as the process name.
+            .orElse(null);
+    if (processName == null || processName.isEmpty()) {
+      return bpmnProcessId;
+    }
```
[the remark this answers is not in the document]
deepthidevaki: Was not sure about this because in the original Operate code a blank processName was replaced by `bpmnProcessId` if it is either null or empty. 

With your suggestion, it can return a null value for processName. @sdorokhova What is the expected behavior here?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
Zelldon: So what you actually need is a factory for the CacheLoader.

deepthidevaki: Refactored this to use a factory.

ChrisKujawa (approved):
🚀

PR #24184, opened 2024-10-29
feat: backup only segment containing data not present in snapshots 

### LINKED ISSUE #23082: Partition backup fails because of concurrent log compaction
**Describe the bug**

Backups in some partitions fails because the segment file is deleted before uploading it to the backup store. 

```
Caused by: java.io.UncheckedIOException: java.nio.file.NoSuchFileException: /xxx/camunda/zeebebrokers-test/data2/raft-partition/partitions/3/raft-partition-partition-3-1.log
```

Example log from an affected cluster. (log is in reverse order of timeline)
```
2024-10-02 06:39:17.529 DEBUG  [zb-fs-workers-0] be.snapshots.impl.FileBasedSnapshotStore: [Broker-0] [SnapshotStore-3] Deleting previous snapshot 62884-14-620635-565330

2024-10-02 06:39:17.526  INFO  [zb-fs-workers-0] be.snapshots.impl.FileBasedSnapshotStore: [Broker-0] [SnapshotStore-3] Committed new snapshot 69074-15-620653-620654

2024-10-02 06:35:37.997  WARN  [zb-actors-0] da.zeebe.backup.management.BackupService: [Broker-0] [BackupService-3] Failed to take backup 20241002063536 at position 620653

2024-10-02 06:35:37.996 DEBUG  [zb-actors-0] e.backup.management.InProgressBackupImpl: [Broker-0] [BackupService-3] Released reservation for snapshot 62884-14-620635-565330

2024-10-02 06:35:37.994 DEBUG  [zb-actors-0] io.camunda.zeebe.backup.s3.S3BackupStore: [Broker-0] [BackupService-3] Reading manifest object of BackupIdentifierImpl[nodeId=2, partitionId=3, checkpointId=20241002063536]

2024-10-02 06:35:37.994  INFO  [zb-actors-0] io.camunda.zeebe.backup.s3.S3BackupStore: [Broker-0] [BackupService-3] Marking BackupIdentifierImpl[nodeId=2, partitionId=3, checkpointId=20241002063536] as failed


Caused by: java.io.UncheckedIOException: java.nio.file.NoSuchFileException: /xxxxx/camunda/zeebebrokers-test/data2/raft-partition/partitions/3/raft-partition-partition-3-1.log


2024-10-02 06:35:36.556 DEBUG  [zb-actors-1] e.backup.management.InProgressBackupImpl: [Broker-0] [BackupService-3] Reserved snapshot 62884-14-620635-565330

2024-10-02 06:35:36.554 DEBUG  [zb-actors-1] e.backup.management.InProgressBackupImpl: [Broker-0] [BackupService-3] Attempting to reserve snapshot 62884-14-620635-565330

2024-10-02 06:35:36.554 DEBUG  [raft-server-2-3] munda.zeebe.journal.file.SegmentsManager: [Broker-2] [raft-server-3] raft-partition-partition-3 - Deleting log up from 1 up to 1 (removing 1 segments)


2024-10-02 06:35:36.459  INFO  [zb-fs-workers-1] be.snapshots.impl.FileBasedSnapshotStore: [Broker-0] [SnapshotStore-3] Committed new snapshot 62884-14-620635-565330

2024-10-02 06:19:16.014  INFO  [zb-actors-1] io.camunda.zeebe.logstreams.snapshot    : [Broker-0] [SnapshotDirector-3] Finished taking temporary snapshot, need to wait until last written event position 620652 is committed, current commit position is 0. After that snapshot will be committed.

2024-10-02 06:19:15.493 DEBUG  [zb-fs-workers-0] io.camunda.zeebe.logstreams.snapshot    : [Broker-0] [SnapshotStore-3] Taking temporary snapshot into /ANR-ATE-TPQ/LOANS-CM/camunda/zeebebrokers-test/data2/raft-partition/partitions/3/snapshots/62884-14-620635-565330.

```

Root cause:
This is a race condition between taking backup and taking snapshot. To avoid this case, we are already "reserving" the snapshots before taking the backup. This way we ensure that the associated files are not deleted before backup is completed. However what happens here is that

1. Snapshot is taken
2. Log compaction is scheduled
3. Backup is triggered
4. Backup manager collects segment files
5. Log is compacted - delete some of the segment files collected by BackupManager
6. Reserve snapshot
7. Start saving backup
6. ==> Fail to save segment file, because the collected file is concurrently deleted. 

BackupManager is finding the segment files before before the log compaction occurs. This should be prevented. Either backup manager can lock the segment files from compaction, or it should collect only the segment files that won;t be compacted based on the reserved snapshot. 

**To Reproduce**

To reproduce, backup and snapshot must run around the same time. The customer is encountering this problem frequently, even though we have not observed this before. So not sure how easy it is to reproduce. 

**Expected behavior**

Backup should be successful even if snapshot and compaction is run concurrently. 

**Environment:**
- OS: 
- Zeebe Version: 
- Configuration: 

Related to https://jira.camunda.com/browse/SUPPORT-23674

entangled90:
## Description
Previously, all segments present in the directory would be marked
eligible for backup without coordinator with the LogCompaction.

With this change, the BackupManager will check with the Journal which
segments contain data that is not present in the selected snapshot,
limiting the possibility that these files will be canceled while a
backup is in place.

However, nothing stops the LogCompaction to delete these files if the
backup is particularly slow. To avoid such issue another synchronization
mechanism should be added to the LogCompaction.

## Related issues

closes #23082

### DIFF
--- a/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/InProgressBackupImpl.java
+++ b/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/InProgressBackupImpl.java
@@ -24,13 +24,14 @@ import io.camunda.zeebe.util.VersionUtil;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Predicate;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -50,7 +51,7 @@ final class InProgressBackupImpl implements InProgressBackup {
 
   private final Path segmentsDirectory;
 
-  private final Predicate<Path> isSegmentsFile;
+  private final JournalInfoProvider journalInfoProvider;
 
   // Snapshot related data
   private boolean hasSnapshot = true;
@@ -67,14 +68,14 @@ final class InProgressBackupImpl implements InProgressBackup {
       final int numberOfPartitions,
       final ConcurrencyControl concurrencyControl,
       final Path segmentsDirectory,
-      final Predicate<Path> isSegmentsFile) {
+      final JournalInfoProvider journalInfoProvider) {
     this.snapshotStore = snapshotStore;
     this.backupId = backupId;
     this.checkpointPosition = checkpointPosition;
     this.numberOfPartitions = numberOfPartitions;
     this.concurrencyControl = concurrencyControl;
     this.segmentsDirectory = segmentsDirectory;
-    this.isSegmentsFile = isSegmentsFile;
+    this.journalInfoProvider = journalInfoProvider;
   }
 
   @Override
@@ -104,6 +105,7 @@ final class InProgressBackupImpl implements InProgressBackup {
               } else if (snapshots.isEmpty()) {
                 // no snapshot is taken until now, so return successfully
                 hasSnapshot = false;
+                availableValidSnapshots = Collections.emptySet();
                 result.complete(null);
               } else {
                 final var eitherSnapshots = findValidSnapshot(snapshots);
@@ -176,26 +178,37 @@ final class InProgressBackupImpl implements InProgressBackup {
   @Override
   public ActorFuture<Void> findSegmentFiles() {
     final ActorFuture<Void> filesCollected = concurrencyControl.createFuture();
-    try (final var stream = Files.list(segmentsDirectory)) {
-      final var fileSet =
-          stream
-              .filter(isSegmentsFile)
-              .collect(
-                  Collectors.toMap(
-                      path -> segmentsDirectory.relativize(path).toString(), path -> path));
-
-      if (fileSet.isEmpty()) {
-        filesCollected.completeExceptionally(
-            new IllegalStateException("Segments must not be empty"));
-        return filesCollected;
+    try {
+      long maxIndex = 0L;
+      if (availableValidSnapshots != null) {
+        maxIndex =
+            availableValidSnapshots.stream()
+                .mapToLong(PersistedSnapshot::getIndex)
+                .max()
+                .orElse(0L);
       }
-
-      segmentsFileSet = new NamedFileSetImpl(fileSet);
-      filesCollected.complete(null);
-    } catch (final IOException e) {
+      final var segments = journalInfoProvider.getTailSegments(maxIndex);
+      segments.whenComplete(
+          (journalMetadata, throwable) -> {
+            if (throwable != null) {
+              filesCollected.completeExceptionally(throwable);
+            } else if (journalMetadata.isEmpty()) {
+              filesCollected.completeExceptionally(
+                  new IllegalStateException("Segments must not be empty"));
+            } else {
+              final Map<String, Path> map =
+                  journalMetadata.stream()
+                      .collect(
+                          Collectors.toMap(
+                              path -> segmentsDirectory.relativize(path).toString(),
+                              Function.identity()));
+              segmentsFileSet = new NamedFileSetImpl(map);
+              filesCollected.complete(null);
+            }
+          });
+    } catch (final Exception e) {
       filesCollected.completeExceptionally(e);
     }
-
     return filesCollected;
   }
 
--- a/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/JournalInfoProvider.java
+++ b/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/JournalInfoProvider.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.backup.management;
+
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Get the some information from the journal from an outside thread. All APIs are thread safe and
+ * can be safely used from a different thread then the one used by Raft.
+ */
+public interface JournalInfoProvider {
+
+  /**
+   * Get the segment files that are after or equal to an index.
+   *
+   * <p>As an example, it can Useful for getting all the segment files >= commit_index or
+   * checkpoint_index.
+   *
+   * @param index the index in the log to select the segments
+   * @return a future containing a collection of segment files
+   */
+  CompletableFuture<Collection<Path>> getTailSegments(long index);
+}
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
@@ -30,7 +30,7 @@ import io.camunda.zeebe.scheduler.SchedulingHints;
 import io.camunda.zeebe.snapshots.PersistedSnapshot;
 import io.camunda.zeebe.snapshots.impl.FileBasedSnapshotStore;
 import io.camunda.zeebe.snapshots.impl.FileBasedSnapshotStoreImpl;
-import io.camunda.zeebe.test.util.AutoCloseableRule;
+import io.camunda.zeebe.test.DynamicAutoCloseable;
 import io.camunda.zeebe.util.FileUtil;
 import io.camunda.zeebe.util.buffer.DirectBufferWriter;
 import java.io.IOException;
@@ -56,12 +56,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @ExtendWith(MockitoExtension.class)
-public class ConcurrentBackupCompactionTest {
+public class ConcurrentBackupCompactionTest extends DynamicAutoCloseable {
 
   private static final Logger LOG = LoggerFactory.getLogger(ConcurrentBackupCompactionTest.class);
   private static final String SNAPSHOT_FILE_NAME = "file1";
   @TempDir Path dataDirectory;
-  private final AutoCloseableRule closeRule = new AutoCloseableRule();
   private ActorScheduler actorScheduler;
   private SegmentedJournal journal;
   private FileBasedSnapshotStore snapshotStore;
@@ -76,12 +75,11 @@ public class ConcurrentBackupCompactionTest {
 
   @BeforeEach
   void setUp() {
-    actorScheduler = closeRule.manage(ActorScheduler.newActorScheduler().build());
+    actorScheduler = manage(ActorScheduler.newActorScheduler().build());
     actorScheduler.start();
-    backupStore = closeRule.manage(new InMemoryMockBackupStore());
+    backupStore = manage(new InMemoryMockBackupStore());
     snapshotStore =
-        closeRule.manage(
-            new FileBasedSnapshotStore(0, partitionId, dataDirectory, snapshotPath -> Map.of()));
+        manage(new FileBasedSnapshotStore(0, partitionId, dataDirectory, snapshotPath -> Map.of()));
     actorScheduler.submitActor(snapshotStore, SchedulingHints.IO_BOUND);
 
     final var partitionMetadata =
@@ -90,7 +88,7 @@ public class ConcurrentBackupCompactionTest {
     final var raftPartition = new RaftPartition(partitionMetadata, null, dataDirectory.toFile());
 
     journal =
-        closeRule.manage(
+        manage(
             SegmentedJournal.builder()
                 .withDirectory(dataDirectory.toFile())
                 .withName(raftPartition.name())
@@ -109,7 +107,7 @@ public class ConcurrentBackupCompactionTest {
                 });
 
     backupService =
-        closeRule.manage(
+        manage(
             new BackupService(
                 nodeId,
                 partitionId,
@@ -125,7 +123,7 @@ public class ConcurrentBackupCompactionTest {
 
   @AfterEach
   public void tearDown() {
-    closeRule.after();
+    close();
   }
 
   @Test
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.broker.backup;
+
+import static java.nio.file.StandardOpenOption.CREATE_NEW;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import io.atomix.cluster.MemberId;
+import io.atomix.primitive.partition.PartitionId;
+import io.atomix.primitive.partition.PartitionMetadata;
+import io.atomix.raft.impl.LogCompactor;
+import io.atomix.raft.metrics.RaftServiceMetrics;
+import io.atomix.raft.partition.RaftPartition;
+import io.atomix.raft.storage.log.RaftLog;
+import io.atomix.utils.concurrent.ThreadContext;
+import io.camunda.zeebe.backup.management.BackupService;
+import io.camunda.zeebe.broker.utils.InlineThreadContext;
+import io.camunda.zeebe.journal.JournalMetaStore;
+import io.camunda.zeebe.journal.file.SegmentedJournal;
+import io.camunda.zeebe.scheduler.ActorScheduler;
+import io.camunda.zeebe.scheduler.SchedulingHints;
+import io.camunda.zeebe.snapshots.PersistedSnapshot;
+import io.camunda.zeebe.snapshots.impl.FileBasedSnapshotStore;
+import io.camunda.zeebe.snapshots.impl.FileBasedSnapshotStoreImpl;
+import io.camunda.zeebe.test.util.AutoCloseableRule;
+import io.camunda.zeebe.util.FileUtil;
+import io.camunda.zeebe.util.buffer.DirectBufferWriter;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.time.Duration;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import org.agrona.concurrent.UnsafeBuffer;
+import org.awaitility.Awaitility;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.stubbing.Answer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ExtendWith(MockitoExtension.class)
+public class ConcurrentBackupCompactionTest {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ConcurrentBackupCompactionTest.class);
+  private static final String SNAPSHOT_FILE_NAME = "file1";
+  @TempDir Path dataDirectory;
+  private final AutoCloseableRule closeRule = new AutoCloseableRule();
+  private ActorScheduler actorScheduler;
+  private SegmentedJournal journal;
+  private FileBasedSnapshotStore snapshotStore;
+  private InMemoryMockBackupStore backupStore;
+  private BackupService backupService;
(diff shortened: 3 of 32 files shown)


zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
```diff
@@ -0,0 +1,14 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.journal;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface JournalInfoProvider {
+  CompletableFuture<Segments> getTailSegments(long index);
```
npepinpe, line 13: :question: Why pass a specific type instead of `Map<Long, Path>`? Do you expect the `Segments` type to grow in the future, or perform some kind of functionality?

zeebe/journal/src/main/java/io/camunda/zeebe/journal/file/SegmentedJournal.java
```diff
@@ -187,6 +190,13 @@ public boolean isOpen() {
     return open;
   }
 
+  @Override
+  public SortedMap<Long, Path> getTailSegments(final long index) {
+    // lazy view on the map
```
npepinpe, line 195: :question: I've never used this, but if I understand correctly, this means the transform function is applied lazily, i.e. only when the element is accessed, correct?

Considering this is a view, and the underlying map is mutable, should we not eat the cost of generating an immutable copy (anyway the map should not be very large)? I think it would make it a lot easier to reason about, i.e. the tail segments at calling time.

zeebe/snapshot/src/main/java/io/camunda/zeebe/snapshots/TransientSnapshot.java
```diff
@@ -16,12 +16,12 @@ public interface TransientSnapshot extends PersistableSnapshot {
...
    * Takes a snapshot on the given path. This can be persisted later via calling {@link
-   * PersistableSnapshot#persist()}. Based on the implementation this could mean that this is writen
-   * before on a temporary folder and then moved to the valid snapshot directory.
+   * PersistableSnapshot#persist()}. Based on the implementation this could mean that this is
+   * written before on a temporary folder and then moved to the valid snapshot directory.
    *
    * @param takeSnapshot the predicate which should take the snapshot and should return true on
    *     success
-   * @return true on success, false otherwise
+   * @return a future reflecting the result of the operation
```
npepinpe, line 24: :+1: Good catch

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
```diff
@@ -0,0 +1,14 @@
...
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.journal;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface JournalInfoProvider {
```
npepinpe, line 12: :clown_face: Not sure about the name, but I also don't have a better idea :smile: I'm terrible at naming anyway.

zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/InProgressBackupImpl.java
```diff
@@ -176,26 +179,37 @@ public ActorFuture<Void> findSnapshotFiles() {
...
+      segments.whenComplete(
+          (journalMetadata, throwable) -> {
+            if (throwable != null) {
+              filesCollected.completeExceptionally(throwable);
+            } else if (journalMetadata.segments().isEmpty()) {
+              filesCollected.completeExceptionally(
+                  new IllegalStateException("Segments must not be empty"));
+            } else {
+              final Map<String, Path> map =
+                  journalMetadata.segments().values().stream()
```
npepinpe, line 201: :wrench: Considering the main consumer of the API ends up looking at only the values, do we need to return a map there? Can't we just return a `Collection<Path>`? That would also remove a bunch of other refactoring done around the signature there.

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
```diff
@@ -0,0 +1,232 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.stubbing.Answer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ExtendWith(MockitoExtension.class)
+public class ConcurrentBackupCompactionTest {
```
npepinpe, line 63: :x: We try and write all new tests using Junit 5. Some tests are still with Junit 4 _if_ they use things (e.g. rules) which are not yet usable for Junit 5. In this case, we have an `AutoCloseResources` extension, and there is a `@TmpDir` annotation for Junit 5. There's also a Mockito extension. So I don't see any reason not to write the test as Junit 5, unless I missed something.

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
npepinpe, line 12: :wrench: If the only implementation of the interface is the `RaftServer`, this is perhaps better moved to the module holding that class, as it doesn't really have anything to do with journal. That likely means removing the dependency on the `journal` module from the `backup` module?

zeebe/backup/pom.xml
```diff
@@ -70,6 +75,18 @@
...
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-atomix-cluster</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.agrona</groupId>
+      <artifactId>agrona</artifactId>
+      <scope>provided</scope>
```
npepinpe, line 87: :wrench: I assume this was automatically added? Not sure it makes sense to have provided here.

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
```diff
@@ -0,0 +1,232 @@
...
+  static void afterAll() throws ExecutionException, InterruptedException {
+    actorScheduler.stop().get();
+  }
+
+  @BeforeEach
+  void setUp() {
+    // methods use by the compactor to schedule callbacks
+    doAnswer(
+            inv -> {
+              ((Runnable) inv.getArgument(0)).run();
```
npepinpe, line 97: :question: So call backs are invoked immediately from the caller thread, inline? Why not just use a dummy `ThreadContext` which does that? We have a `DeterministicSingleThreadContext` for example, but you could also just create one:

```java
private static final class InlineThreadContext implements ThreadContext {

  @Override
  public Scheduled schedule(
      final Duration initialDelay, final Duration interval, final Runnable callback) {
    callback.run();
    return new Scheduled() {
      @Override
      public void cancel() {}

      @Override
      public boolean isDone() {
        return true;
      };
    };
  }

  @Override
  public void execute(final Runnable command) {
    command.run();
  }
}
```

npepinpe: I'm not a big fan of mocks like this, I find it harder to understand what's going on than with stubs.

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
npepinpe, line 1: :thought_balloon: In the `journal` module, we have a `TestJournal` (and some sub classes) utility to create segments of predictable size and control compaction. I'm wondering if we shouldn't make some of these utilities available to other modules... :thinking:

npepinpe (changes requested):
:+1: Good stuff, the general approach is quite good.

I would like to go over the main test `ConcurrentBackupCompactionTest` together, as there's a lot there, and I want to be sure we get it right. The rest I think can be done async, and likely wouldn't need another review pass.

:thought_balloon: I didn't understand your comment though about requiring more in the log compactor. The LogCompactor will never delete segments below its known compaction bound, so while this is frozen by the snapshot store, it should not be possible for it to delete the segments until the reservation is released. What _could_ happen is the log being truncated or reset, but our understanding was that this was not possible at the moment on the leader (though who knows, this may change, so we may still want some stronger requirements by reserving segments as well).

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
entangled90: I agree it's not a good name, hard to find a proper one as you say though.

zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/InProgressBackupImpl.java
entangled90: The reasoning was that if we need to then filter the paths from [snapshot_index..commit_index] in the future it may be helpful to have a SortedMap, but the size of the map is probably negligible so it does not matter much.

zeebe/backup/pom.xml
entangled90: Unforunately it's really used in "source" code only in the test scope. 

However, it's also used in the compile scope by the journal module, so If I set `scope=test` then the compile task does not succeed. 

However If I don't put the `scope=test` the lint task does not succeed because it should be put with `scope=test`.

I hope I explained the situation, is there a better solution?

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
entangled90: Interesting, I didn't know about that. Unfortunately in this test I tried to make the segment size just big enough to hold one record, but I'm not very happy about that, as it's fixed and over time it may become too small (e.g. by adding fields to the SBE schema for example).

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
entangled90: I'll move it to `DeterministingSingleThreadContext`. :+1:

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
entangled90: In hindsight I put it here because I though the journal would implement it, but later realized that journal does not use future at all and so I just implemented in raft module.
I'll move it. This means that `zeebe-backup` will depend on `zeebe-atomix-cluster` which currently does not.

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
entangled90: Sorry was looking for `AutoCloseResources` but I only found `AutoCloseableRule` so I thought it would be the right one. Will move to what you suggested.

zeebe/journal/src/main/java/io/camunda/zeebe/journal/file/SegmentedJournal.java
entangled90: Yes, even though the underlying map is concurrent, it's better to have an immutable copy at a specific point in time.

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
entangled90: When I first wrote I was tempted to add `committed_index` for example, so I thought it could be used in the future. As of now it's potentially overcomplicating things, but we were already discussing about addding an optimization related to the commit index. 
The name `Segments` is probably misleading though.

zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/InProgressBackupImpl.java
entangled90: I think it's probably better to do as you suggested and decide in the future if the api needs to be changed.

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
entangled90: I think it's better for now to just return a `Collection<Path>` as you suggest.

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
npepinpe: Hm, the other option is putting it in a module that both `zeebe-atomix-cluster` and `zeebe-backup` share as dependency, and where it would also make sense. But I honestly don't know of any. 

A third option is having it as an interface purely in `zeebe-backup`, as a way to specify its dependencies, and have the place where we wire it together provide an adapter. This fully decouples both modules, and makes the integration module responsible. I think that's also fine, imo, esp. since the interface is (so far) a special purpose case for the backups :shrug:

zeebe/backup/pom.xml
npepinpe: Ah, maven :sweat_smile: 

No, it was just unexpected to have `provided` here, that's all. That said, can we set it to test if we remove the dependency on the `journal` module?

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
entangled90: `DeterministingSingleThreadContext` is defined in zeebe-atomix-cluster, I'll just use the stub you wrote.

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
npepinpe: It's unfortunately a helper only available in the journal :thinking: I'm not sure how much it uses internals either, I don't remember, so I don't know if it's easy to make it available to external modules =/

But if we decouple it from the journal anyway, maybe we don't need it?

zeebe/backup/pom.xml
entangled90: Yes probably, but we cannot depend on `zeebe-atomix-cluster` as well. Is there a way to ignore this error when linting? this is more of a linter issue rather than maven per se.

entangled90: Fixes by configuring the plugin to ignore false positives

zeebe/journal/src/main/java/io/camunda/zeebe/journal/JournalInfoProvider.java
entangled90: I went for the third option as it's the cleanest as you suggested.

zeebe/backup/src/test/java/io/camunda/zeebe/backup/management/ConcurrentBackupCompactionTest.java
entangled90: Now the test has been moved to the broker module, as it's more or less an integration test. Should we keep it like it is now?

zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/JournalInfoProvider.java
```diff
@@ -0,0 +1,16 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.backup.management;
+
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.concurrent.CompletableFuture;
+
+public interface JournalInfoProvider {
```
npepinpe, line 14: :x: Please add Javadocs describing the contract of this API.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
```diff
@@ -0,0 +1,216 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.stubbing.Answer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ExtendWith(MockitoExtension.class)
+public class ConcurrentBackupCompactionTest {
```
npepinpe, line 61: :x: In order for `@AutoCLoseResource` to work, we need to annotate the class with `@AutoCloseResources`

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/InMemoryMockBackupStore.java
```diff
@@ -0,0 +1,127 @@
...
+        id,
+        (k, status) ->
+            new BackupStatusImpl(
+                k,
+                status.descriptor(),
+                BackupStatusCode.FAILED,
+                Optional.of(failureReason),
+                status.created(),
+                Optional.of(Instant.now())));
+    return null;
```
npepinpe, line 88: :question: Does it make sense to return null? If we don't care about this implementation, it's better to throw a `new UnsupportedOperationException("Not yet implemented; implemented it when required")`. It's generally how we introduce new stubs progressively, by only implementing the subset we care about and throwing elsewhere with a clear message, so if someone ever needs this new use case, they can implement it properly.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+  @BeforeAll
+  static void beforeAll() {
+    actorScheduler = ActorScheduler.newActorScheduler().build();
+    actorScheduler.start();
+  }
+
+  @AfterAll
+  static void afterAll() throws ExecutionException, InterruptedException {
+    actorScheduler.stop().get();
```
npepinpe, line 86: :wrench: I think `ActorScheduler` is `AutoCloseable`, so you could also just annotate it with `@AutoCloseResource` and remove the `afterAll` method entirely.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
```diff
@@ -0,0 +1,216 @@
...
+  private SegmentedJournal journal;
+  @AutoCloseResource private FileBasedSnapshotStore snapshotStore;
+  @AutoCloseResource private InMemoryMockBackupStore backupStore;
+  @AutoCloseResource private BackupService backupService;
+  private final int nodeId = 1;
+  private final int partitionId = 1;
+  private LogCompactor logCompactor;
+  private final ThreadContext threadContext = new InlineThreadContext();
+  @Mock private RaftLog raftLog;
+  @Mock private RaftServiceMetrics raftMetrics;
```
npepinpe, line 76: :upside_down_face: Since it's so cheap to create (e.g. `new RaftServiceMetrics("1")`), I would not bother mocking it but instead actually return an object.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
```diff
@@ -0,0 +1,216 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.stubbing.Answer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ExtendWith(MockitoExtension.class)
+public class ConcurrentBackupCompactionTest {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ConcurrentBackupCompactionTest.class);
+  private static ActorScheduler actorScheduler;
```
npepinpe, line 64: :question: Any reason to make it static? It's fairly cheap to create (all things considered), and anything shared between tests may create flakiness (e.g. oops, I didn't close a previous actor, it's still running in the next test).

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
```diff
@@ -0,0 +1,216 @@
...
+    // because the BackupStore it's blocked)
+    final var backupResultFut = backupService.takeBackup(backupIdx, backupIdx);
+
+    Awaitility.await("snapshot is reserved")
+        .until(
+            () ->
+                snapshotStore.getLatestSnapshot().map(PersistedSnapshot::isReserved).orElse(false));
+
+    // backup service is blocked
+    //    final var blockOnBackup = blockActor(backupService);
```
npepinpe, line 155: :x: Can be removed

npepinpe (approved):
:+1: Please address all comments before merging, but I don't think we need another review

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
entangled90: It was intended to be static, probably added by mistake and never noticed it afterwards

zeebe/broker/src/test/java/io/camunda/zeebe/broker/backup/ConcurrentBackupCompactionTest.java
entangled90: Unfortunately `@AutoCloseResource` does not guarantee an order in which the resources are closed and the test fails to stop correctly, probably because the `ActorScheduler` is stopped before the last actor.

entangled90: I created a class `DynamicAutoCloseable` to extract the logic from `AutoCloseableRule` to be used without direct support from  Junit 4/5

Issue #20579, opened 2024-07-29
[EPIC] Handle C8 REST Query API availability


tmetzke:
## Introduction

The Camunda 8 REST API features a partially introduced Query API with 8.6 that is (planned to be) completed with 8.7.
This part of the REST API has to be disabled by default with 8.6 and enabled with 8.7.
Before it is completed, its availability has to be configurable, i.e. it can be enabled via a configuration property.
When it is completed, its availability is not configurable any more.

## Concept

With 8.6, a new configuration property exists to enable the C8 REST Query API (e.g. search for user tasks and process instances). With the property being enabled, the respective REST controller beans are created in the REST gateway. Otherwise, the beans are not created and search paths are not accessible.

Since the Java client for Camunda 8 also contains parts of the Query API, the related commands are marked as `@ExperimentalApi` until the endpoints are enabled by default.

The REST API explorers in the public documentation as well as the OpenAPI specification of the C8 API declare the Query API endpoints with 8.6 already. The endpoints are however clearly marked as experimental and disabled by default.

### Before Query API is completed (8.6)

```[tasklist]
#### Gateway API
- [x] camunda/camunda#20580
- [x] camunda/camunda#22224
```

```[tasklist]
#### Java client
- [x] camunda/camunda#20596
```

```[tasklist]
#### Testing
- [x] camunda/camunda#20645
```

```[tasklist]
#### Documentation
- [x] camunda/camunda-docs#4097
- [x] camunda/camunda#20619
- [x] camunda/camunda-docs#4121
```

```[tasklist]
#### Console
- [x] ~Make Query API an activatable alpha feature~
```

### After Query API is completed (planned for 8.7)

```[tasklist]
#### Gateway API
- [ ] https://github.com/camunda/camunda/issues/25993
```

```[tasklist]
#### Java client
- [ ] https://github.com/camunda/camunda/issues/26009
```

```[tasklist]
#### Testing
- [x] Adjust automated tests for REST API configuration properties
```

```[tasklist]
#### Documentation
- [ ] https://github.com/camunda/camunda-docs/issues/4761
- [ ] https://github.com/camunda/camunda/issues/26022
```

## Discussions and Decisions

* Instead of marking Query API as "experimental", we'll make it an [alpha feature](https://docs.camunda.io/docs/reference/alpha-features/).

## Links

- Product Hub (internal): https://github.com/camunda/product-hub/issues/2243

PR #24190, opened 2024-10-29
chore: bring changes in docs upstream to OpenAPI spec

pepopowitz:
## Description

Pulls multiple changes discussed in https://github.com/camunda/camunda-docs/pull/4534 into the OpenAPI spec.

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -758,7 +758,7 @@ paths:
       responses:
         "200":
           description: >
-            The user task varialbes search successful response.
+            The user task variables search successful response.
           content:
             application/json:
               schema:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -60,12 +60,19 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/TopologyResponse"
+        "500":
+          description: >
+            An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
   /license:
     get:
       tags:
         - License
       operationId: getLicense
-      summary: Get status of Camunda license
+      summary: Get license status
       description: Obtains the status of the current Camunda license
       responses:
         "200":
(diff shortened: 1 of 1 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -887,7 +887,7 @@ paths:
         To change the time, the clock must be pinned again with a new timestamp.
 
         :::note
-        This endpoint is an [alpha feature](/reference/alpha-features.md) and may be subject to change
+        This endpoint is an [alpha feature](/components/early-access/alpha/alpha-features.md) and may be subject to change
```
pepopowitz, line 890: 💬 Having a docs link in our OpenAPI spec feels....weird. No tool outside of the docs knows what these paths are.

What do we think about removing these links in the source spec, and injecting them as part of the docs generation process (by looking for the text "This endpoint is an alpha feature")?

tmetzke: Yep, I totally agree, I had the same issue with this. But it's a general issue with writing OpenAPI descriptions that also look as expected in our docs, e.g. adding notes and warnings for endpoints.

For alpha features, we can of course do something like that, yeah. As an endpoint developer, if you want to have it generated for your endpoint, you can add an `(alpha)` to the end of the path summary.

tmetzke: It's not gonna be possible to add more details to the alpha note then for an endpoint developer but we can manage, I think.

For example, the document endpoints provide more details on why they are alpha features still (they use an in-memory store currently, thus they're not production-ready). But this can be a "regular" part of the endpoint description. We can still generate the alpha note at the beginning of the endpoint description automatically.

For the query endpoints, this will be a bit harder because they actually want to link back to the C8 REST API overview docs where the general explanation sits. We don't have to worry too much here, though, this will be removed with 8.7 anyway soon with https://github.com/camunda/camunda/issues/20579 🙂

tmetzke: I added a proposal how this could look like with https://github.com/camunda/camunda/pull/24190/commits/0849e8745da86b2d9a0860b5fab641acfa01e2ae.

We can move to generate that alpha note at some point and remove them from the spec itself when we're ready.

tmetzke:
I streamlined the spec a bit again. We have become inconsistent in naming, wording, response codes, and descriptions a bit lately.

tmetzke (approved):

PR #24196, opened 2024-10-30
Fix identity-sdk concurrency issue

### LINKED ISSUE #23853: Number of threads created in Gateway for blocking identity calls causing "too many concurrent streams" error
**Describe the bug**

After upgrading from Camunda 8.5.2 to Camunda 8.6.2, a customer started to see frequent 500 errors coming from Identity. The error message is `java.io.IOException: too many concurrent streams` caused by Identity SDK's http client. The error message is originated from Zeebe gateway.

This is due to switching to use virtual threads instead of a configured number of threads in a thread pool for Zeebe Gateway (see https://github.com/camunda/camunda/issues/18697). That means now we allow creating unlimited number of threads per blocking identity call. Since Zeebe gateway doesn't limit number of threads created for execution, identity SDK can now create unlimited number of streams but Identity SDK's HttpClient itself fails to handle it.

**Expected behavior**

Identity SDK's Http client shouldn't fail with `too many concurrent streams`.

**Log/Stacktrace**

 <p>

```
Exception:
java.util.concurrent.CompletionException
Message:
java.io.IOException: too many concurrent streams
Stacktrace:
java.util.concurrent.CompletableFuture.encodeRelay
java.util.concurrent.CompletableFuture.uniComposeStage
java.util.concurrent.CompletableFuture.thenCompose
jdk.internal.net.http.MultiExchange.responseAsyncImpl
jdk.internal.net.http.MultiExchange.lambda$responseAsync0$2
com.dynatrace.agent.introspection.threading.completable.FunctionWrapper.apply(FunctionWrapper.java:22)
java.util.concurrent.CompletableFuture$UniCompose.tryFire
java.util.concurrent.CompletableFuture.postComplete
java.util.concurrent.CompletableFuture$AsyncSupply.run
jdk.internal.net.http.HttpClientImpl$DelegatingExecutor.execute
java.util.concurrent.CompletableFuture.completeAsync
jdk.internal.net.http.MultiExchange.responseAsync
jdk.internal.net.http.HttpClientImpl.sendAsync
jdk.internal.net.http.HttpClientImpl.send
jdk.internal.net.http.HttpClientFacade.send
io.camunda.identity.sdk.impl.rest.RestClient.send(RestClient.java:119)
io.camunda.identity.sdk.impl.rest.RestClient.request(RestClient.java:106)
io.camunda.identity.sdk.impl.TenantsImpl.forToken(TenantsImpl.java:38)
jdk.internal.reflect.DirectMethodHandleAccessor.invoke
java.lang.reflect.Method.invoke
io.camunda.identity.sdk.annotation.AnnotationProcessor.lambda$apply$0(AnnotationProcessor.java:33)
jdk.proxy2.$Proxy134.forToken
io.camunda.zeebe.gateway.interceptors.impl.IdentityInterceptor.interceptCall(IdentityInterceptor.java:99)
io.grpc.ServerInterceptors$InterceptCallHandler.startCall(ServerInterceptors.java:269)
io.grpc.internal.ServerImpl$ServerTransportListenerImpl.startWrappedCall(ServerImpl.java:701)
io.grpc.internal.ServerImpl$ServerTransportListenerImpl.access$2200(ServerImpl.java:408)
io.grpc.internal.ServerImpl$ServerTransportListenerImpl$1HandleServerCall.runInternal(ServerImpl.java:613)
io.grpc.internal.ServerImpl$ServerTransportListenerImpl$1HandleServerCall.runInContext(ServerImpl.java:603)
io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133)
java.util.concurrent.ThreadPerTaskExecutor$TaskRunner.run
java.lang.VirtualThread.run
```

</p>
</details>

Related support case: [SUPPORT-24047](https://jira.camunda.com/browse/SUPPORT-24047)

koevskinikola:
## Description

Fixes an issue where too many tenant requeststo Identity cause the service to fail with an `IOException` due to too many concurrent requests.

## Checklist

- [ ] for CI changes:

## Related issues

closes #23853

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/shared/security/IdentityAuthenticationManager.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/security/IdentityAuthenticationManager.java
@@ -11,9 +11,11 @@ import io.camunda.identity.sdk.Identity;
 import io.camunda.identity.sdk.authentication.AccessToken;
 import io.camunda.identity.sdk.tenants.dto.Tenant;
 import io.camunda.zeebe.gateway.impl.configuration.MultiTenancyCfg;
+import io.camunda.zeebe.gateway.impl.identity.IdentityTenantService;
 import io.camunda.zeebe.protocol.record.value.TenantOwned;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.BadCredentialsException;
@@ -25,6 +27,7 @@ import org.springframework.stereotype.Component;
 public final class IdentityAuthenticationManager implements AuthenticationManager {
 
   private final Identity identity;
+  private final IdentityTenantService tenantService;
   private final MultiTenancyCfg multiTenancy;
 
   @Autowired
@@ -32,6 +35,7 @@ public final class IdentityAuthenticationManager implements AuthenticationManage
       final Identity identity, final MultiTenancyCfg multiTenancy) {
     this.identity = identity;
     this.multiTenancy = multiTenancy;
+    tenantService = new IdentityTenantService(identity);
   }
 
   @Override
@@ -61,8 +65,8 @@ public final class IdentityAuthenticationManager implements AuthenticationManage
     }
 
     try {
-      return identity.tenants().forToken(token).stream().map(Tenant::getTenantId).toList();
-    } catch (final RuntimeException e) {
+      return tenantService.getTenantsForToken(token).stream().map(Tenant::getTenantId).toList();
+    } catch (final RuntimeException | ExecutionException e) {
       throw new InternalAuthenticationServiceException(
           "Expected Identity to provide authorized tenants, see cause for details", e);
     }
--- a/dist/src/main/java/io/camunda/zeebe/shared/security/IdentityAuthenticationManager.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/security/IdentityAuthenticationManager.java
@@ -10,6 +10,7 @@ package io.camunda.zeebe.shared.security;
 import io.camunda.identity.sdk.Identity;
 import io.camunda.identity.sdk.authentication.AccessToken;
 import io.camunda.identity.sdk.tenants.dto.Tenant;
+import io.camunda.zeebe.gateway.impl.configuration.ExperimentalCfg;
 import io.camunda.zeebe.gateway.impl.configuration.MultiTenancyCfg;
 import io.camunda.zeebe.gateway.impl.identity.IdentityTenantService;
 import io.camunda.zeebe.protocol.record.value.TenantOwned;
@@ -32,10 +33,12 @@ public final class IdentityAuthenticationManager implements AuthenticationManage
 
   @Autowired
   public IdentityAuthenticationManager(
-      final Identity identity, final MultiTenancyCfg multiTenancy) {
+      final Identity identity,
+      final MultiTenancyCfg multiTenancy,
+      final ExperimentalCfg experimentalCfg) {
     this.identity = identity;
     this.multiTenancy = multiTenancy;
-    tenantService = new IdentityTenantService(identity);
+    tenantService = new IdentityTenantService(identity, experimentalCfg.getIdentityRequest());
   }
 
   @Override
--- a/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
+++ b/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
@@ -11,9 +11,10 @@ import io.camunda.identity.sdk.Identity;
 import io.camunda.identity.sdk.IdentityConfiguration;
 import io.camunda.identity.sdk.authentication.exception.TokenVerificationException;
 import io.camunda.identity.sdk.tenants.dto.Tenant;
+import io.camunda.zeebe.gateway.cmd.ConcurrentRequestException;
 import io.camunda.zeebe.gateway.impl.configuration.GatewayCfg;
 import io.camunda.zeebe.gateway.impl.configuration.IdentityCfg;
-import io.camunda.zeebe.gateway.impl.configuration.IdentityRequestCfg;
+import io.camunda.zeebe.gateway.impl.configuration.IdentityServiceCfg;
 import io.camunda.zeebe.gateway.impl.configuration.MultiTenancyCfg;
 import io.camunda.zeebe.gateway.impl.identity.IdentityTenantService;
 import io.camunda.zeebe.gateway.interceptors.InterceptorUtil;
@@ -25,7 +26,6 @@ import io.grpc.ServerInterceptor;
 import io.grpc.Status;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.RejectedExecutionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -56,10 +56,10 @@ public final class IdentityInterceptor implements ServerInterceptor {
   public IdentityInterceptor(
       final Identity identity,
       final MultiTenancyCfg multiTenancy,
-      final IdentityRequestCfg identityRequestCfg) {
+      final IdentityServiceCfg identityServiceCfg) {
     this.identity = identity;
     this.multiTenancy = multiTenancy;
-    tenantService = new IdentityTenantService(identity, identityRequestCfg);
+    tenantService = new IdentityTenantService(identity, identityServiceCfg);
   }
 
   private static Identity createIdentity(final IdentityCfg config) {
@@ -117,8 +117,8 @@ public final class IdentityInterceptor implements ServerInterceptor {
       final var context = InterceptorUtil.setAuthorizedTenants(authorizedTenants);
       return Contexts.interceptCall(context, call, headers, next);
 
-    } catch (final RejectedExecutionException ree) {
-      return denyTenantCallAndLog(call, Status.UNAVAILABLE, ree);
+    } catch (final ConcurrentRequestException cre) {
+      return denyTenantCallAndLog(call, Status.UNAVAILABLE, cre);
     } catch (final RuntimeException | ExecutionException e) {
       return denyTenantCallAndLog(call, Status.UNAUTHENTICATED, e);
     }
--- a/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
+++ b/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
@@ -13,6 +13,7 @@ import io.camunda.identity.sdk.authentication.exception.TokenVerificationExcepti
 import io.camunda.identity.sdk.tenants.dto.Tenant;
 import io.camunda.zeebe.gateway.impl.configuration.IdentityCfg;
 import io.camunda.zeebe.gateway.impl.configuration.MultiTenancyCfg;
+import io.camunda.zeebe.gateway.impl.identity.IdentityTenantService;
 import io.camunda.zeebe.gateway.interceptors.InterceptorUtil;
 import io.grpc.Contexts;
 import io.grpc.Metadata;
--- a/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
+++ b/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
@@ -11,7 +11,9 @@ import io.camunda.identity.sdk.Identity;
 import io.camunda.identity.sdk.IdentityConfiguration;
 import io.camunda.identity.sdk.authentication.exception.TokenVerificationException;
 import io.camunda.identity.sdk.tenants.dto.Tenant;
+import io.camunda.zeebe.gateway.impl.configuration.GatewayCfg;
 import io.camunda.zeebe.gateway.impl.configuration.IdentityCfg;
+import io.camunda.zeebe.gateway.impl.configuration.IdentityRequestCfg;
 import io.camunda.zeebe.gateway.impl.configuration.MultiTenancyCfg;
 import io.camunda.zeebe.gateway.impl.identity.IdentityTenantService;
 import io.camunda.zeebe.gateway.interceptors.InterceptorUtil;
@@ -36,19 +38,28 @@ public final class IdentityInterceptor implements ServerInterceptor {
   private final IdentityTenantService tenantService;
   private final MultiTenancyCfg multiTenancy;
 
-  public IdentityInterceptor(final IdentityCfg config, final MultiTenancyCfg multiTenancy) {
-    this(createIdentity(config), multiTenancy);
+  public IdentityInterceptor(final IdentityCfg config, final GatewayCfg gatewayCfg) {
+    this(
+        createIdentity(config),
+        gatewayCfg.getMultiTenancy(),
+        gatewayCfg.getExperimental().getIdentityRequest());
   }
 
   public IdentityInterceptor(
-      final IdentityConfiguration configuration, final MultiTenancyCfg multiTenancy) {
-    this(new Identity(configuration), multiTenancy);
+      final IdentityConfiguration configuration, final GatewayCfg gatewayCfg) {
+    this(
+        new Identity(configuration),
+        gatewayCfg.getMultiTenancy(),
+        gatewayCfg.getExperimental().getIdentityRequest());
   }
 
-  public IdentityInterceptor(final Identity identity, final MultiTenancyCfg multiTenancy) {
+  public IdentityInterceptor(
+      final Identity identity,
+      final MultiTenancyCfg multiTenancy,
+      final IdentityRequestCfg identityRequestCfg) {
     this.identity = identity;
     this.multiTenancy = multiTenancy;
-    tenantService = new IdentityTenantService(identity);
+    tenantService = new IdentityTenantService(identity, identityRequestCfg);
   }
 
   private static Identity createIdentity(final IdentityCfg config) {
--- a/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
+++ b/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
@@ -21,6 +21,8 @@ import io.grpc.ServerCallHandler;
 import io.grpc.ServerInterceptor;
 import io.grpc.Status;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.RejectedExecutionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -30,6 +32,7 @@ public final class IdentityInterceptor implements ServerInterceptor {
   private static final Metadata.Key<String> AUTH_KEY =
       Metadata.Key.of("Authorization", Metadata.ASCII_STRING_MARSHALLER);
   private final Identity identity;
+  private final IdentityTenantService tenantService;
   private final MultiTenancyCfg multiTenancy;
 
   public IdentityInterceptor(final IdentityCfg config, final MultiTenancyCfg multiTenancy) {
@@ -44,6 +47,7 @@ public final class IdentityInterceptor implements ServerInterceptor {
   public IdentityInterceptor(final Identity identity, final MultiTenancyCfg multiTenancy) {
     this.identity = identity;
     this.multiTenancy = multiTenancy;
+    tenantService = new IdentityTenantService(identity);
   }
 
   private static Identity createIdentity(final IdentityCfg config) {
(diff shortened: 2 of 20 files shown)


koevskinikola:
ℹ️ Implementation notes:
- I decided to implement the fix starting with `8.6` as with `8.7` Identity will be restructured and merged into Zeebe. So adding the fix into `8.7-SNAPSHOT` will create more clean-up work once we migrate to the new Identity architecture.
- I added an `ExperimentalCfg` to the gateway for the configuration properties used by the `tenant` cache and `semaphore`. This might be useful for any future work, especially now that we have a lot more teams using the Gateway.
- I could reproduce the issue with the IT test, but wasn't able to easily `assert` it. The `IOException` gets lost somewhere in the "gRPC translation" to the client. However, with debugging I could consistently reproduce the occurrence in the gRPC `IdentityInterceptor`. I'm open to any improvements to the code there.
- I named the `identity-sdk` wrapper class `IdentityTenantService`. It contains both a `LoadingCache` and a `semaphore`. I'm open to suggestions in regards to the naming and configuration defaults. I derived the defaults from the minimal amount of concurrent requests that triggered the bug in the above IT test.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
```diff
@@ -0,0 +1,80 @@
...
+public class IdentityTenantService {
+
+  private static final TimeUnit TIME_UNIT = TimeUnit.MILLISECONDS;
+
+  private final LoadingCache<String, List<Tenant>> tenantCache;
+  private final Semaphore semaphore;
+  private final Identity identity;
+
+  private final boolean isCachingEnabled;
+  private final long semaphoreTimeout;
```
npepinpe, line 31: :wrench: I would keep calling it by its logical name, `tenantRequestTimeout`.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
```diff
@@ -0,0 +1,80 @@
...
+            .maximumSize(config.getTenantCacheSize())
+            .build(
+                new CacheLoader<>() {
+                  @Override
+                  public List<Tenant> load(final String token) {
+                    return getTenantsForTokenThrottled(token);
+                  }
+                });
+    ree =
+        new RejectedExecutionException(
```
npepinpe, line 51: :question: So the stack trace will always show this right? We don't know where exactly it was thrown. I guess it's OK; not ideal, but it's definitely cheaper all things considered :thinking:

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
```diff
@@ -0,0 +1,80 @@
...
+            String.format(
```
npepinpe, line 52: :wrench: The message only explains what happens. I would update it to match our normal style, and also to indicate why this could have happened, and what a user can do, i.e. enable a self-service approach for a user. Something like:

```
Expected to fetch tenants from Identity within %d%s, but there are too many concurrent requests; either increase the tenant request capacity, or scale Identity to complete requests faster.
```

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private List<Tenant> getTenantsForTokenThrottled(final String token) {
+    try {
+      if (!semaphore.tryAcquire(semaphoreTimeout, TIME_UNIT)) {
+        throw ree;
+      }
+      return getTenantsForTokenInternal(token);
+    } catch (final InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw ree;
```
npepinpe, line 71: :question: I'm not sure this is the right thing to do here. If the thread was interrupted, it doesn't mean that you timed out, it just means the caller thread is interrupted/likely to be shutdown. In most cases, I suspect throwing here doesn't really make a difference. 

:wrench: At any rate, I would recommend throwing a new exception, if only to keep the cause so we have a proper stack trace chain.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  private List<Tenant> getTenantsForTokenThrottled(final String token) {
+    try {
+      if (!semaphore.tryAcquire(semaphoreTimeout, TIME_UNIT)) {
+        throw ree;
+      }
+      return getTenantsForTokenInternal(token);
+    } catch (final InterruptedException e) {
+      Thread.currentThread().interrupt();
```
npepinpe, line 70: :+1: :rocket:

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/configuration/IdentityRequestCfg.java
```diff
@@ -0,0 +1,97 @@
...
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.impl.configuration;
+
+import java.util.Objects;
+
+public class IdentityRequestCfg {
```
npepinpe, line 12: :upside_down_face: imo this is more of a `IdentityServiceCfg`, but honestly this is very much a nit :melting_face:

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
```diff
@@ -30,6 +32,7 @@ public final class IdentityInterceptor implements ServerInterceptor {
   private static final Metadata.Key<String> AUTH_KEY =
       Metadata.Key.of("Authorization", Metadata.ASCII_STRING_MARSHALLER);
   private final Identity identity;
```
remcowesterhoud, line 37: 💭 If we also put the token verification in the wrapper we don't need his field here anymore and it all goes through the wrapper. That way we don't differentiate between sdk calls, which could cause confusion.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
```diff
@@ -97,24 +113,32 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
...
-          "Denying call {} as the authorized tenants could not be retrieved from Identity. Error message: {}",
-          methodDescriptor.getFullMethodName(),
-          e.getMessage());
-      return deny(
-          call,
-          Status.UNAUTHENTICATED
-              .augmentDescription(
-                  "Expected Identity to provide authorized tenants, see cause for details")
-              .withCause(e));
+    } catch (final RejectedExecutionException ree) {
```
npepinpe, line 120: :wrench: I would propose using a specific exception type if only to avoid a possible `RejectedExcutionException` coming from the Identity SDK itself, which also uses some stuff internally via the HttpClient.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/multitenancy/MultiTenancyOverIdentityHighRequestCountIT.java
```diff
@@ -0,0 +1,450 @@
...
+                    .processDefinitionKey(processDefinitionKey)
+                    .tenantId(TENANT_A)
+                    .send());
+      }
+
+      // when
+      final List<Future<ZeebeFuture<ProcessInstanceEvent>>> results =
+          executorService.invokeAll(piList);
+
+      Thread.sleep(10000);
```
npepinpe, line 272: :fearful: This looks like flakiness waiting to happen. Can we come up with a different approach?

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/multitenancy/MultiTenancyOverIdentityHighRequestCountIT.java
```diff
@@ -0,0 +1,450 @@
...
+    process =
+        Bpmn.createExecutableProcess(processId)
+            .startEvent()
+            .serviceTask("task", b -> b.zeebeJobType("type"))
+            .endEvent()
+            .done();
+  }
+
+  // Regression test: https://github.com/camunda/camunda/issues/23853
+  @Test
```
npepinpe, line 236:
```suggestion
  @RegressionTest("https://github.com/camunda/camunda/issues/23853")
```

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/multitenancy/MultiTenancyOverIdentityHighRequestCountIT.java
npepinpe, line 1: :thought_balloon: A lot of the code here is reused from `MultiTenancyOverIdentityIT`. Normally I would propose to not duplicate, esp. things like the `PostgresHelper`, but I guess if we're getting rid of much of this later, maybe this is the right approach for now.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityTenantService.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private final RejectedExecutionException ree;
+
+  private final LoadingCache<String, List<Tenant>> tenantCache;
+  private final Semaphore semaphore;
+  private final Identity identity;
+
+  private final int semaphoreCapacity = 300;
+  private final long semaphoreTimeout = 1;
+  private final long cacheSize = 1000;
```
remcowesterhoud: ❓ How did you decide on the size? It feels like 1000 is a lot, but then again maybe it isn't. I've got no clue 😅

npepinpe (approved):
:rocket: Great stuff!

Generally I think we could probably unit test the `IdentityTenantService`, since the only new tests is very broad, and likely difficult to diagnose when things go wrong. Still pre-approving because all-in-all I don't think we need another review (but feel free to ask for one). Do address all comments before merging.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityTenantService.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class IdentityTenantService {
+
+  private final RejectedExecutionException ree;
+
+  private final LoadingCache<String, List<Tenant>> tenantCache;
+  private final Semaphore semaphore;
+  private final Identity identity;
+
+  private final int semaphoreCapacity = 300;
```
remcowesterhoud: ❓ Again, how did you decide on this? Do we have some numbers from which we found a sensible limit?

dist/src/main/java/io/camunda/zeebe/shared/security/IdentityAuthenticationManager.java
```diff
@@ -25,13 +27,15 @@
 public final class IdentityAuthenticationManager implements AuthenticationManager {
 
   private final Identity identity;
```
remcowesterhoud, line 30: 💭 Same comment here regarding wrapping all the identity sdk call into the service

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -142,18 +143,22 @@ yield createProblemDetail(
             HttpStatus.SERVICE_UNAVAILABLE, pnfeMsg, pnfe.getClass().getName());
       case final PartitionInactiveException pie:
         final var pieMsg =
-            "Expected to handle gRPC request, but the target partition is currently inactive";
+            "Expected to handle REST API request, but the target partition is currently inactive";
```
remcowesterhoud, line 146: 👍 Interesting that this wasn't caught before 😅

remcowesterhoud (approved):
Thanks @koevskinikola, no blockers from my side but a few comments 👀

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/configuration/IdentityRequestCfg.java
koevskinikola: Sounds better to me, I'll refactor it.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityTenantService.java
koevskinikola: In the IT test, the bug wasn't triggered if less than 300 concurrent requests were sent to the gateway, and between 300-500 the test/bug occurrence became flaky. So I think 300 is a nice limit to prevent entering the "flaky zone".

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityTenantService.java
koevskinikola: That's a good point, I didn't have a good gauge on what would be a good value here. The cache size is directly correlated to the number of users, i.e. Identity JWT tokens, and that number may vary significantly per Zeebe user.

I agree, 1000 seems like a very large number for JWT tokens. I'll try to poll the Identity team for some metrics so I can derive some sensible number.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
koevskinikola: The problem is that this code is gRPC-specific. My goal with the wrapper was to make it reusable for gRPC and the REST API. I would need to play around with lambdas (maybe even generics) to achieve that. 

It will probably be a nicer solution than what is currently provided, but we need to deliver a fix for this by next week, and I don't want to risk introducing a regression.

@remcowesterhoud would it be acceptable if I create a `kind/toil` issue for this?

dist/src/main/java/io/camunda/zeebe/shared/security/IdentityAuthenticationManager.java
koevskinikola: See my answer [here](https://github.com/camunda/camunda/pull/24196#discussion_r1824102403).

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityTenantService.java
npepinpe: Well, we can also see it from a memory point of view. Can we estimate how much 1000 entries would take? If it's still low, than that's fine I think.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
koevskinikola: @npepinpe considering your comment [here](https://github.com/camunda/camunda/pull/24196#discussion_r1822944741), I'll only throw it if the semaphore lock can't be acquired (on line 67). Does that make sense?

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
npepinpe: Would it make sense to have a specific type thrown in this service for all cases where we couldn't grab the lease? Or would we handle them differently?

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityTenantService.java
koevskinikola: I'm not sure it is. We're storing a list of tenants, so if there is a large number of tenants (let's say 200-300?) mapped to 1000 JWT tokens, is this still a low memory footprint?

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
remcowesterhoud: Maybe I'm missing something, but I thought the `IdentityAuthenticationManager` was the rest part? They both make a call to `identity.authentication().verifyToken(tokenValue);` which seems like we could move it to the wrapper.

Anyway it was only a thought, feel free to ignore it.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/multitenancy/MultiTenancyOverIdentityHighRequestCountIT.java
koevskinikola: Yup, in the interest of time, I copied most of it. In turn, I think the `MultiTenancyOverIdentityIT` code is copied from `GatewayAuthenticationIdentityIT`.

Should I create a follow-up `kind/toil` issue for this cleanup?

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityInterceptor.java
koevskinikola: Ah okay, I misunderstood. I thought you wanted to wrap the token extraction from the request as well.

During the solution discussion with @npepinpe, we decided to avoid adding other `identity-sdk` requests to this wrapper since no problems have been reported about them. We can expand the wrapper if additional issues are reported.

In the end, this fix is only a band-aid. The best solution would be to make Identity more scalable and we are currently making Identity more scalable by integrating its data model in Zeebe. After we're finished with the re-architecture, this fix won't be needed.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/identity/IdentityTenantService.java
koevskinikola: UPDATE: I also created a new Zeebe `ConcurrentRequestException`, so it's something unique that can be mapped to this feature.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/multitenancy/MultiTenancyOverIdentityHighRequestCountIT.java
koevskinikola: I'll remove it, the `ExecutorService#invokeAll(...)` method actually blocks until all tasks are completed. I missed that.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/multitenancy/MultiTenancyOverIdentityHighRequestCountIT.java
koevskinikola: FYI: https://github.com/camunda/camunda/issues/24254

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/IdentityTenantService.java
koevskinikola: Considering the average size of a JWT string (cache key) is ~1KB, and a single `tenantId` with max size (32 UTF-8 chars) is ~56bytes, plus the overhead of the Cache objects (let's say ~32 bytes if it's similar to a `Map.Entry`), a single entry in the cache would be ~1,1KB.

- Let's make an assumption that an installation has ~50 tenants (that's still quite a large number). If they are all assigned to a single JWT token, that's 50x56bytes=2,8KB.
- For 1000 entries where all JWT tokens are assigned to all tenants: 1000x(1,1KB+2,8KB+32bytes)=**3,9 MB**

I'll drop the default size to 200 entries. It's still a large number of JWT tokens, but the memory footprint should be ~780 KB then.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/multitenancy/MultiTenancyOverIdentityHighRequestCountIT.java
```diff
@@ -0,0 +1,449 @@
...
+                  future -> {
+                    try {
+                      return future.get();
+                    } catch (final InterruptedException | ExecutionException e) {
+                      throw new RuntimeException(e);
+                    }
+                  })
+              .map(ZeebeFuture::toCompletableFuture)
+              .filter(CompletableFuture::isCompletedExceptionally)
+              .map(CompletableFuture::exceptionNow)
```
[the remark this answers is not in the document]
koevskinikola: @remcowesterhoud I tried adding an additional assertion out, but it seems like a high load is not always reached in Zeebe. I think that's acceptable, as the issue is in Identity not being able to handle high load, and the wrapper is handling that.

Issue #21567, opened 2024-08-26
JobRecord elementId gets overwritten when an error is thrown during job execution


kristinkomschow:
**Describe the bug**

When an error occurs during job execution, the `elementId` (which usually contains the flowNodeId) of the job record gets overwritten with `NO_CATCH_EVENT_FOUND`:
https://github.com/camunda/camunda/blob/4392c73cb62369f24d2dcee5968bbd2469af0ea6/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobThrowErrorProcessor.java#L150

This is problematic for Operate when retrieving executed jobs for a given flow node (e.g. to show Execution Listener jobs for flow nodes).

Context why this was introduced: https://github.com/camunda/camunda/blame/4392c73cb62369f24d2dcee5968bbd2469af0ea6/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobThrowErrorProcessor.java#L42-L48

**To Reproduce**

**Expected behavior**

The elementId of JobRecords still contains the flowNodeId after an error is thrown.

**Log/Stacktrace**

<details><summary>Full Stacktrace</summary>
 <p>

```
<STACKTRACE>
```

</p>
</details>

**Environment:**
- OS: 
- Zeebe Version: 8.6.0-alpha5
- Configuration:

mustafadagher:
ZPA Triage:
- The impact is that consumers of exporters see a wrong element id in the error thrown event
- That creates a subsequent performance impact for operate right now

korthout:
>by introducing a new field to the JobRecord,

Alternatively, perhaps we can simply introduce a new JobIntent for this case.

ThorbenLindhauer:
@mustafadagher should this ticket be closed?

mustafadagher:
> [@mustafadagher](https://github.com/mustafadagher) should this ticket be closed?

HI @ThorbenLindhauer, actually no.
I've resolved the symptom just for operate by setting the correct value in the incident event. But the JobRecord still element Id still gets overwritten until the incident is resolved, it gets reset again.

So if we want to resolve this at the root, we still need this ticket (see my [comment](https://github.com/camunda/camunda/issues/21567#issuecomment-2447120559) above)

ThorbenLindhauer:
[...]

PR #24206, opened 2024-10-30
fix: set the correct elementId in Incident record when Job throw with no catch event

mustafadagher:
## Description

Right now, we use `NO_CATCH_EVENT_FOUND` as a marker element ID to indicate that a given catch event could not be found
We set the same in the incident record too, and this creates an issue for operate

In this PR, when raising an incident, if the job element Id is set to `NO_CATCH_EVENT_FOUND`, the element instance state is queried to provide the correct elementId and set it in the IncidentRecord
Also as a side benefit, we can remove this look-up in the IncidentResolvedApplier

See [related slack discussion](https://camunda.slack.com/archives/CSQ2E3BT4/p1730127004045349).

## Related issues

relates to #21567


ChrisKujawa (approved):
Great stuff 🚀 Thanks for fixing 👍🏼

korthout:
@mustafadagher Please backport this fix 🙇

PR #24213, opened 2024-10-30
Add process cache metrics

### LINKED ISSUE #24208: Add metrics for monitoring process cache performance

deepthidevaki:
## Description

Adds metrics for ProcessCache in CamundaExporter. Following metrics has been added:

* Cache hit count
* Cache misses count
* Cache eviction count
* Duration for loading cache entry from the backend
* Time spent trying to load cache entry but failed

The metrics is added via a concrete implementation of `StatsCounter` which will be used by the Caffeine implementation to directly record the statistics. This provides a clean way to record the metrics. Alternative is to use the default implementation for statistics and periodically query it and update micrometer metrics.

## Related issues

closes #24208

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -22,6 +22,7 @@ import static io.camunda.zeebe.protocol.record.ValueType.VARIABLE;
 
 import co.elastic.clients.util.VisibleForTesting;
 import io.camunda.exporter.adapters.ClientAdapter;
+import io.camunda.exporter.cache.ExporterCacheMetrics;
 import io.camunda.exporter.config.ConfigValidator;
 import io.camunda.exporter.config.ExporterConfiguration;
 import io.camunda.exporter.exceptions.PersistenceException;
@@ -56,6 +57,7 @@ public class CamundaExporter implements Exporter {
   private CamundaExporterMetrics metrics;
   private Logger logger;
   private BackgroundTaskManager taskManager;
+  private ExporterCacheMetrics exporterCacheMetrics;
 
   public CamundaExporter() {
     this(new DefaultExporterResourceProvider());
@@ -74,7 +76,8 @@ public class CamundaExporter implements Exporter {
     context.setFilter(new CamundaExporterRecordFilter());
     metrics = new CamundaExporterMetrics(context.getMeterRegistry());
     clientAdapter = ClientAdapter.of(configuration);
-    provider.init(configuration, clientAdapter.getExporterEntityCacheProvider());
+    provider.init(
+        configuration, clientAdapter.getExporterEntityCacheProvider(), context.getMeterRegistry());
 
     taskManager =
         BackgroundTaskManager.create(
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
@@ -9,6 +9,7 @@ package io.camunda.exporter;
 
 import static java.util.Map.entry;
 
+import io.camunda.exporter.cache.ExporterCacheMetrics;
 import io.camunda.exporter.cache.ExporterEntityCacheImpl;
 import io.camunda.exporter.cache.ExporterEntityCacheProvider;
 import io.camunda.exporter.config.ConnectionTypes;
@@ -81,6 +82,7 @@ import io.camunda.webapps.schema.descriptors.usermanagement.index.MappingIndex;
 import io.camunda.webapps.schema.descriptors.usermanagement.index.RoleIndex;
 import io.camunda.webapps.schema.descriptors.usermanagement.index.TenantIndex;
 import io.camunda.webapps.schema.descriptors.usermanagement.index.UserIndex;
+import io.micrometer.core.instrument.MeterRegistry;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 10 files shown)


deepthidevaki:
The visualization for these metrics in Grafana should be created in a separate PR. But just to test that the metrics are recorded and reported correctly, I have run a benchmark and added some basic visualization for the metrics which you can view it [here](https://grafana.dev.zeebe.io/d/temp-zeebe-process-cache/temp-zeebe-process-cache?orgId=1&var-DS_PROMETHEUS=prometheus&var-cluster=All&var-namespace=dd-process-cache-metrics&var-pod=All&var-partition=All&from=1730462639627&to=1730463494682)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/ProcessCacheMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private static final String NAMESPACE = "zeebe.camunda.exporter.processcache";
+
+  private final Counter hitCount;
+  private final Counter missCount;
+  private final Timer loadSuccessDuration;
+  private final Timer loadFailureDuration;
+  private final Counter evictionCount;
+
+  public ProcessCacheMetrics(final MeterRegistry meterRegistry) {
```
Zelldon, line 28: :x: We should add labels for the partition

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/ProcessCacheMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  public void recordEviction(final int weight, final RemovalCause cause) {
+    evictionCount.increment();
+  }
+
+  @Override
+  public CacheStats snapshot() {
+    // not implemented, as we don't need it
+    return null;
```
Zelldon, line 86: ❓ Can we return null?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/ProcessCacheMetrics.java
```diff
@@ -0,0 +1,92 @@
...
+    missCount =
+        Counter.builder(meterName("misses"))
+            .description("Number of cache misses")
+            .register(meterRegistry);
+
+    evictionCount =
+        Counter.builder(meterName("evictions"))
+            .description("Number of cache evictions")
+            .register(meterRegistry);
+
```
Zelldon, line 43: Can we also handle via a label, type

ChrisKujawa (approved):
Thanks @deepthidevaki 👍🏼 

As discussed I will take over and do some smaller changes, where I see necessary

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/ProcessCacheMetrics.java
Zelldon: This is actually done already in the context https://github.com/camunda/camunda/blob/main/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/context/ExporterContext.java#L49-L54

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/ProcessCacheMetrics.java
Zelldon: Should be fine as snapshot is a delegation from the Cache to retrieve a snapshot of metrics when user calls Cache # get statistics

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/ProcessCacheMetrics.java
Zelldon: It looked like we had to define tags directly with values. So I first thought this might not feasible, BUT actually it is just to define the type (it seems to be).

See https://stackoverflow.com/a/49171202/2165134

Zelldon: What I think we can do is: 

```

access = Counter.builder(meterName("access"))
            .description("Number of cache access by type")
            .tag("type", "")
            .register(meterRegistry);
```

And then later:

```
  @Override
  public void recordHits(final int count) {
    Metrics.counter(meterName("access"), "type", "hit").increment(count);
  }
```

Zelldon: I guess one issue with this is again using static registry 🤔 instead of the actual live object?

Zelldon: Thanks to @lenaschoenburg I realized we can use the actual registry we have at the hand to do the same.

Zelldon:
For posterity, I tried to also clean a bit the interfaces here https://github.com/camunda/camunda/tree/ck-process-cache (timeboxed), but run into issues like we need the index name from process index, which we don't have in the CamundaExporter only in descriptors, etc.

I will keep it for now (I think the ClientAdapter interface feels a bit polluted, but there is also not much benefit right now)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -56,6 +57,7 @@ public class CamundaExporter implements Exporter {
   private CamundaExporterMetrics metrics;
   private Logger logger;
   private BackgroundTaskManager taskManager;
+  private ExporterCacheMetrics exporterCacheMetrics;
```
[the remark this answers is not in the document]
Zelldon: Huch might be happen due to resolving merge conflicts.

Thanks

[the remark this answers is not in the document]
Zelldon: @mustafadagher fixed here https://github.com/camunda/camunda/pull/24347/commits/3748e3205ef4d5c64262f4d161ed9c94f8813b13

Issue #22536, opened 2024-09-20
Replace zbctl usage in `zeebe-testbench` workflow with equivalent REST API call


megglos:
Replace usage of zbctl in https://github.com/camunda/camunda/blob/main/.github/workflows/zeebe-testbench.yaml#L84
with:
```
      - name: Start Test
        shell: bash
        run: |
          token=$(curl --request POST ${ZEEBE_AUTHORIZATION_SERVER_URL} \
            --header 'Content-Type: application/x-www-form-urlencoded' \
            --data-urlencode "grant_type=client_credentials" \
            --data-urlencode "audience=zeebe.camunda.io" \
            --data-urlencode "client_id=${ZEEBE_CLIENT_ID}" \
            --data-urlencode "client_secret=${ZEEBE_CLIENT_SECRET}" \
            | jq -j ".access_token")

          variables=$(echo "${{ inputs.variables }}" | envsubst)

          curl -L -X POST "${ZEEBE_ADDRESS}/v2/process-instances" \
            -H "Authorization: Bearer ${token}" \
            -H "Content-Type: application/json" \
            -H "Accept: application/json" \
            --data-raw "{
              \"bpmnProcessId\": \"${{ inputs.processId }}\",
              \"variables\": $variables
            }"
        env:
          ZEEBE_CLIENT_SECRET: ${{ steps.secrets.outputs.TESTBENCH_PROD_CLIENT_SECRET }}
          ZEEBE_ADDRESS: ${{ steps.secrets.outputs.TESTBENCH_PROD_REST_ADDRESS }}
          ZEEBE_AUTHORIZATION_SERVER_URL: 'https://login.cloud.camunda.io/oauth/token'
          ZEEBE_CLIENT_ID: 'Jg9caDRuAWHchvM7JiaVlndL-qVFfp~0'
```
Once the testbench cluster is updated to 8.6.0.

PR #24215, opened 2024-10-30
Fix Daily QA workflow

nicpuppa:
## Description

Due to the deprecation of zbctl in favour of REST the daily QA workflow on main is not running since last week, due to an issue parsing the variables.

related to: #22536


tmetzke (approved):
🦖 Looks good to me 👍 
❓ Do we have a test run that can verify this works as expected now?

PR #24237, opened 2024-10-30
Exporter handler tests

### LINKED ISSUE #23888: Camunda exporter handler tests
**Description**

Handlers in the camunda exporter supports a number of records, ensure that when records are exported, corresponding entities are written to the correct index.

```[tasklist]
### Tasks
- [x] IT for UserHandler
- [x] IT for AuthorizationHandler
- [x] IT for DecisionHandler
- [x] IT for ListViewProcessInstanceFromProcessInstanceHandler
- [x] IT for ListViewFlowNodeFromIncidentHandler
- [x] IT for ListViewFlowNodeFromJobHandler
- [x] IT for ListViewFlowNodeFromProcessInstanceHandler
- [x] IT for ListViewVariableFromVariableHandler
- [x] IT for VariableHandler
- [x] IT for DecisionRequirementsHandler
- [x] IT for PostImporterQueueFromIncidentHandler
- [x] IT for FlowNodeInstanceIncidentHandler
- [x] IT for FlowNodeInstanceProcessInstanceHandler
- [x] IT for IncidentHandler
- [x] IT for SequenceFlowHandler
- [x] IT for DecisionEvaluationHandler
- [x] IT for ProcessHandler
- [x] IT for MetricFromProcessInstanceHandler
- [x] IT for FormHandler
- [x] IT for EventFromIncidentHandler
- [x] IT from EventFromJobHandler
- [x] IT from EventFromProcessInstanceHandler
- [x] IT from EventFromProcessMessageSubscriptionHandler
```

EuroLew:
## Description

Exporter handler integration tests which end at verifying the contents of Opensearch/Elasticsearch

## Related issues

closes #23888

### DIFF
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/listview/FlowNodeInstanceForListViewEntity.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/listview/FlowNodeInstanceForListViewEntity.java
@@ -197,18 +197,14 @@ public class FlowNodeInstanceForListViewEntity
         activityId,
         activityState,
         activityType,
-        incidentKeys,
         errorMessage,
         incident,
         jobFailedWithRetriesLeft,
         tenantId,
-        pendingIncident,
         position,
         positionIncident,
         positionJob,
-        joinRelation,
-        startTime,
-        endTime);
+        joinRelation);
   }
 
   @Override
@@ -225,19 +221,15 @@ public class FlowNodeInstanceForListViewEntity
     final FlowNodeInstanceForListViewEntity that = (FlowNodeInstanceForListViewEntity) o;
     return incident == that.incident
         && jobFailedWithRetriesLeft == that.jobFailedWithRetriesLeft
-        && pendingIncident == that.pendingIncident
         && Objects.equals(processInstanceKey, that.processInstanceKey)
         && Objects.equals(activityId, that.activityId)
         && activityState == that.activityState
         && activityType == that.activityType
-        && Objects.equals(incidentKeys, that.incidentKeys)
         && Objects.equals(errorMessage, that.errorMessage)
         && Objects.equals(tenantId, that.tenantId)
         && Objects.equals(position, that.position)
         && Objects.equals(positionIncident, that.positionIncident)
         && Objects.equals(positionJob, that.positionJob)
-        && Objects.equals(joinRelation, that.joinRelation)
-        && Objects.equals(startTime, that.startTime)
-        && Objects.equals(endTime, that.endTime);
+        && Objects.equals(joinRelation, that.joinRelation);
   }
 }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
@@ -15,6 +15,7 @@ import io.camunda.zeebe.protocol.record.ValueType;
 import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
 import io.camunda.zeebe.protocol.record.value.IncidentRecordValue;
 import java.time.OffsetDateTime;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 
 public class PostImporterQueueFromIncidentHandler
@@ -69,7 +70,7 @@ public class PostImporterQueueFromIncidentHandler
         .setIntent(intent)
         .setKey(record.getKey())
         .setPosition(record.getPosition())
-        .setCreationTime(OffsetDateTime.now())
+        .setCreationTime(OffsetDateTime.now().truncatedTo(ChronoUnit.MILLIS))
         .setPartitionId(record.getPartitionId())
         .setProcessInstanceKey(recordValue.getProcessInstanceKey());
   }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
@@ -15,7 +15,6 @@ import io.camunda.zeebe.protocol.record.ValueType;
 import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
 import io.camunda.zeebe.protocol.record.value.IncidentRecordValue;
 import java.time.OffsetDateTime;
-import java.time.temporal.ChronoUnit;
 import java.util.List;
 
 public class PostImporterQueueFromIncidentHandler
@@ -70,7 +69,7 @@ public class PostImporterQueueFromIncidentHandler
         .setIntent(intent)
         .setKey(record.getKey())
         .setPosition(record.getPosition())
-        .setCreationTime(OffsetDateTime.now().truncatedTo(ChronoUnit.MILLIS))
+        .setCreationTime(OffsetDateTime.now())
         .setPartitionId(record.getPartitionId())
         .setProcessInstanceKey(recordValue.getProcessInstanceKey());
   }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
@@ -79,4 +79,9 @@ public class PostImporterQueueFromIncidentHandler
   public void flush(final PostImporterQueueEntity entity, final BatchRequest batchRequest) {
     batchRequest.add(indexName, entity);
   }
+
+  @Override
+  public String getIndexName() {
+    return indexName;
+  }
 }
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
@@ -353,14 +353,11 @@ final class CamundaExporterIT {
     private <S extends ExporterEntity<S>, T extends RecordValue> Record<T> recordGenerator(
         final ExportHandler<S, T> handler, final Supplier<Record<T>> createRecord) {
       // Sometimes the factory generates record with intents that are not supported by the handler.
-      for (int i = 0; i < 500; i++) {
-        final var record = createRecord.get();
-        if (handler.handlesRecord(record)) {
-          return record;
-        }
-      }
-      throw new IllegalArgumentException(
-          "Could not generate handled record for " + handler.getClass().getSimpleName());
+      var record = createRecord.get();
+      do {
+        record = createRecord.get();
+      } while (!handler.handlesRecord(record));
+      return record;
     }
 
     private <S extends ExporterEntity<S>, T extends RecordValue>
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
@@ -474,7 +474,7 @@ final class CamundaExporterIT {
         final SearchClientAdapter clientAdapter)
         throws IOException {
 
-      final var currentTime = OffsetDateTime.now();
+      final var currentTime = OffsetDateTime.now().truncatedTo(ChronoUnit.MILLIS);
       final var mocked = mockStatic(OffsetDateTime.class);
       mocked.when(OffsetDateTime::now).thenReturn(currentTime);
 
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
@@ -440,7 +440,7 @@ final class CamundaExporterIT {
     }
 
     private <S extends ExporterEntity<S>, T extends RecordValue>
-        Supplier<Record<T>> getHandlerRecord(final ExportHandler<S, T> handler) {
+        Supplier<Record<T>> getRecordWhichCanBeHandled(final ExportHandler<S, T> handler) {
 
       final Map<Class<?>, Supplier<Record<T>>> customRecordGenerators =
           Map.of(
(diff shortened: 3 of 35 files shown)


webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/listview/FlowNodeInstanceForListViewEntity.java
```diff
@@ -197,18 +197,14 @@ public int hashCode() {
...
         jobFailedWithRetriesLeft,
         tenantId,
-        pendingIncident,
         position,
         positionIncident,
         positionJob,
-        joinRelation,
-        startTime,
-        endTime);
+        joinRelation);
```
Zelldon, line 207: ❓ Why we deleted these properties?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
```diff
@@ -69,7 +70,7 @@ public void updateEntity(
         .setIntent(intent)
         .setKey(record.getKey())
         .setPosition(record.getPosition())
-        .setCreationTime(OffsetDateTime.now())
+        .setCreationTime(OffsetDateTime.now().truncatedTo(ChronoUnit.MILLIS))
```
Zelldon, line 73: ❓ Not sure whether we can simply do that? What was the reason behind this?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+        throws IOException {
+      final var provider = new DefaultExporterResourceProvider();
+      provider.init(config);
+
+      final var exporter =
+          createExporter(
+              new HashSet<>(provider.getIndexDescriptors()),
+              new HashSet<>(provider.getIndexTemplateDescriptors()),
+              config);
+      for (final var handler : provider.getExportHandlers()) {
```
Zelldon, line 353: 🔧 Lets assert that we have at least one handler.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+        final ExporterConfiguration config, final SearchClientAdapter clientAdapter)
+        throws IOException {
+      final var provider = new DefaultExporterResourceProvider();
+      provider.init(config);
+
+      final var exporter =
+          createExporter(
+              new HashSet<>(provider.getIndexDescriptors()),
+              new HashSet<>(provider.getIndexTemplateDescriptors()),
+              config);
```
Zelldon, line 352: 🔧 I guess we could also create the exporter with the provider directly. And call the init/configure and open method here to make more clear that this happens as part of the test as well.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+      for (final var handler : provider.getExportHandlers()) {
+        exportTest(exporter, handler, clientAdapter);
```
Zelldon, line 354: 💭 To make the test a bit understandable we might can concentrate on one handler at a time in the test. We could provide the different handlers via a parameter.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+      }
+    }
+
+    private <S extends ExporterEntity<S>, T extends RecordValue> Record<T> recordGenerator(
+        final ExportHandler<S, T> handler, final Supplier<Record<T>> createRecord) {
+      for (int i = 0; i < 500; i++) {
```
Zelldon, line 360: ❓ What is the reason behind this? Having here a loop?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+
+      final Supplier<Record<T>> defaultRecordGen =
+          () ->
+              recordGenerator(
+                  handler,
+                  () ->
+                      factory.generateRecord(
+                          handler.getHandledValueType(),
+                          r -> r.withTimestamp(System.currentTimeMillis())));
+      return () -> customRecordGenerators.getOrDefault(handler.getClass(), defaultRecordGen).get();
```
Zelldon, line 457: ❓ Does this make sense to recreate the suppliers everytime again?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+        if (handler.handlesRecord(record)) {
+          return record;
+        }
+      }
+      throw new IllegalArgumentException(
+          "Could not generate handled record for " + handler.getClass().getSimpleName());
+    }
+
+    private <S extends ExporterEntity<S>, T extends RecordValue>
+        Supplier<Record<T>> getHandlerRecord(final ExportHandler<S, T> handler) {
```
Zelldon, line 371: ❓ ❌ TBH I'm not fully sure what we are doing here. Or trying to achive here. 

Is this logic because of a missing mapping between handler and record?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+      handler.updateEntity(record, expectedEntity);
+      exporter.export(record);
+
+      mocked.close();
+
+      final var responseEntity =
+          clientAdapter.get(
+              expectedEntity.getId(), handler.getIndexName(), handler.getEntityType());
+
+      if (!responseEntity.equals(expectedEntity)) {
```
Zelldon, line 482: ❓ ❌ Hm is it really expected that they are not equal? And especially that we don't do any assert at all? 🤔

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,158 @@ private ExporterResourceProvider mockResourceProvider(
...
+      final var entityId = handler.generateIds(record).getFirst();
+      final var expectedEntity = handler.createNewEntity(entityId);
+      handler.updateEntity(record, expectedEntity);
+      exporter.export(record);
+
+      mocked.close();
+
+      final var responseEntity =
+          clientAdapter.get(
+              expectedEntity.getId(), handler.getIndexName(), handler.getEntityType());
```
Zelldon, line 480: 👍🏼 I like that we have a way to verify that the actual and expected state with the entity and handler vs what the exporter does.

ChrisKujawa (changes requested):
Hey @EuroLew 

first of all great stuff! 

First, I was a bit afraid that we went in the wrong direction and only tested the handlers and build extra logic to sent data to ES (in our tests). But later I realized that we indeed test and verify our exporter (incl. schema creation 👍🏼) and what the exporter does when consuming the record. This is great 👍🏼 

I think we can improve a little on the test itself to make it a bit more understandable and maintainable. 

What I was thinking whether we could maybe get the handler and exporter as parameter in the test method (maybe even the corresponding record), so we concentrate then per test only on one handler. Makes the test more clearer and straight forward to understand (plus we can see individual test cases running and even failing).

We have then separate the logic of creating handlers, exporter and  corresponding record. 

I guess we need to take a closer look at the record creation I guess? It seems to be quite complicated right now? Maybe we can find a different way.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/listview/FlowNodeInstanceForListViewEntity.java
EuroLew: Yes sorry I talked to @sdorokhova about whether this is ok, since these fields are `@JsonIgnore` they do not appear in elasticsearch and therefore they are not relevant to the equals and hashcode, keeping it in breaks my tests as a record generates an entity with this fields but when we retrieve that same entity from ELS it does not have these fields

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
EuroLew: Elasticsearch actually trancates the time, there you send the expectedEntity which is what is flushed to ELS, you see the document on ELS, and then you see the result of the get in responseEntity (which shows the document in ELS it is as we expect) 
image
image

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
EuroLew: I should add a comment, but basically sometimes records are generated from the given `ValueType` but the handler cannot handle the record because it generates with the wrong intents

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
EuroLew: oops my bad this was something that was there from WIP code.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
EuroLew: Basically the `factory` used to generate records is not perfect, all it knows is the field types and it generates a random value for that. For some handlers when they require certain states i the record for example `ListViewFlowNodeFromProcessInstanceHandler` requires in `handlesRecord` that the record is not a process event and its intent is one of 5 intents, maybe over enough random generations the factory randomly generated record would create a record that could be handled but this would result in a flaky test.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
EuroLew: No I'll move them into functions.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/listview/FlowNodeInstanceForListViewEntity.java
Zelldon: Wondering why we had them 🤔

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/PostImporterQueueFromIncidentHandler.java
Zelldon: But isn't this something we have to do everywhere then? 🤔

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
Zelldon: That sounds like a good candidate fora flaky test then

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -382,4 +334,163 @@ private ExporterResourceProvider mockResourceProvider(
...
+  class ExportHandlerTests {
+
+    @TestTemplate
+    void allHandlerTestsWithInvocationProvider(
+        final CamundaExporter exporter,
+        final SearchClientAdapter clientAdapter,
+        final ExportHandler<?, ?> handler)
+        throws IOException {
+
+      exportTest(exporter, handler, clientAdapter);
```
Zelldon, line 349: 🔧 Lets inline this.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterHandlerITInvocationProvider.java
```diff
@@ -0,0 +1,133 @@
...
+    final var elsExporter = new CamundaExporter();
+    elsExporter.configure(
+        new ExporterTestContext()
+            .setConfiguration(
+                new ExporterTestConfiguration<>(ELASTICSEARCH.toString(), elsConfig)));
+    elsExporter.open(new ExporterTestController());
+
+    // doesn't matter which config we use as handlers are unaware of who they export to
+    final var provider = new DefaultExporterResourceProvider();
+    provider.init(osConfig);
```
Zelldon, line 66: ❓ I think that is not true, as they check whether it is ES or OS right?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
Zelldon: @EuroLew I think I understand now the issue better. The problem is that some handlers only support specific intents. Obviously we dont want to couple this knowledge of which intent is handled by which handler in our IT.

But I think it would be safer if we use here a while loop instead of a for. Otherwise there exist the chance we dont generate a fiting record.

What I think we could do is:

```
do {
// generate record
} while(!handler.handles(record))
```

Does this make sense?

Zelldon: To even make this more performant (elegant?) we can move this one layer up in the set up code. Do it once for all handlers we have (generating a record + passing to the handler to get the expected entity) and then the test method only retrieves

- CamundaExporter
- ExpectedEntity
- ClientAdapter

Wdyt?

ChrisKujawa (approved):
I think we are already pretty good, just smaller comments and ideas. Happy to discuss them. :)

I guess nothing blocking so pre-approving :) Good work 👍🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
EuroLew: I'm ok with passing the record it to stop it being generated twice per handler, but I think the record should still be passed in so there's a logical progression to the test in that 
1. generate the expected entity

PR #24241, opened 2024-10-30
Fix Operate FE E2E tests locally

### LINKED ISSUE #24266: Operate FE tests fail locally
### Description

The Operate frontend tests are failing locally. This is related to a previous problem that was partially solved as part of [a previous PR](https://github.com/camunda/camunda/pull/23309), where they were initially failing on the CI.

The problem ultimately stems from [a change made in June 2024](https://github.com/camunda/camunda/pull/18604) that added `/operate` to the base path of the application, which in turn makes it necessary to use relative paths in some places, like the E2E tests.

Furthermore, there seems to be a distinction between CI and local environments as some tests pass in one but not in the other in different scenarios regarding paths being relativized or not, which is evidenced by the [change on the previous PR to base path configuration in Playwright's configuration file](https://github.com/camunda/camunda/blob/1e1256a4da4fc5be5e2c6835bc1287a4c3883215/operate/client/playwright.config.ts#L28-L30).

With this issue, the goal is to stabilize the E2E tests, working on both the CI and local environments for development, by making the paths that are used in the `goto` functions in the E2E tests relative.

### Additional Information

### Acceptance Criteria

vitorwtavares:
## Description

- Create helper function to add dot to relativize path locally for e2e tests
- Use helper function in all `goto` function uses to avoid tests breaking locally

## Additional Information

closes #24266 

## Checklist

- [ ] for CI changes:

### DIFF
--- a/operate/client/e2e-playwright/pages/utils/relativizePath.ts
+++ b/operate/client/e2e-playwright/pages/utils/relativizePath.ts
@@ -1,0 +1,15 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+
+const IS_LOCAL = !Boolean(process.env.CI);
+
+const relativizePath = (path: string) => {
+  return IS_LOCAL ? `.${path}` : path;
+};
+
+export {relativizePath};
--- a/operate/client/e2e-playwright/pages/utils/relativizePath.ts
+++ b/operate/client/e2e-playwright/pages/utils/relativizePath.ts
@@ -6,10 +6,8 @@
  * except in compliance with the Camunda License 1.0.
  */
 
-const IS_LOCAL = !Boolean(process.env.CI);
-
 const relativizePath = (path: string) => {
-  return IS_LOCAL ? `.${path}` : path;
+  return `.${path}`;
 };
 
 export {relativizePath};
--- a/operate/client/e2e-playwright/pages/DecisionInstance.ts
+++ b/operate/client/e2e-playwright/pages/DecisionInstance.ts
@@ -8,6 +8,7 @@
 
 import {Locator, Page} from '@playwright/test';
 import {Paths} from 'modules/Routes';
+import {relativizePath} from './utils/relativizePath';
 
 export class DecisionInstance {
   private page: Page;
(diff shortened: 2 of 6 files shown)


operate/client/e2e-playwright/pages/utils/relativizePath.ts
```diff
@@ -0,0 +1,15 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+
+const IS_LOCAL = !Boolean(process.env.CI);
+
+const relativizePath = (path: string) => {
+  return IS_LOCAL ? `.${path}` : path;
+};
```
pedesen, line 13: I like the idea of the relativizePath function, but I don't understand the point of introducing a new switch for `process.env.CI` here, while we already have a switch in [playwright.config.ts](https://github.com/camunda/camunda/blob/main/operate/client/playwright.config.ts#L29)? :thinking:  In my opinion this introduces more points of failure/confusion.

I think it should rather be adjusted in the playwright config and relativizePath should always return a relative path (relative to `/` or to `/operate/`, depending on the environment).

```suggestion
const relativizePath = (path: string) => {
  return `.${path}`;
};
```

pedesen (changes requested):

operate/client/e2e-playwright/pages/utils/relativizePath.ts
vitorwtavares: @pedesen I'm not sure I understand your proposal here, so I'll just try to break down what I'm thinking here and if needed we can sync on this on Monday.

I believe the `relativizePath` function will always have to be used in a per-instance basis, meaning it will have to be called inside a `goto` function in a test file or imported helper function (like `navigateToDecisions`, for example). As I see it, it's mostly a way to avoid having dots added to some paths with no additional context, making it difficult to maintain. Given that, in those test and helper files we don't have access to the base path nor do we have access to the information if this is a CI or local environment (outside of importing it like it's done in the `relativizePath` file currently).

Adding a dot to the end of the basePath does not work in turning the path into a relative one, so I don't really see a way to centralize this in the playwright config file and avoid the per-instance using of the helper function or manual addition of a dot.

If you just mean exporting using the base path generated from the `getBasePath` function from the playwright config file instead of using `process.env.CI`, I personally don't see any difference between them apart from importing the CI flag from env being more legible, since the `getBasePath` will always be based on that environment variable either way, and both of them are getting that data from the same source.

I might be missing something, and it might be easier to explain on a quick call, so if I am, we can schedule that for next week and I'd love to understand this better.

pedesen: For transparency: we clarified in a call and came to the conclusion that my suggested changes work and are bit cleaner than the original solution.

pedesen (approved):

PR #24256, opened 2024-10-31
Move OperationTemplate and BatchOperationTemplate to webapps-schema

### LINKED ISSUE #24255: Move OperationTemplate to Webapps-schema

mustafadagher:
## Description
The PR provides the following:
- Move OperationTemplate and BatchOperationTemplate to webapps-schema
- Configure spring beans for OperationTemplate and BatchOperationTemplate
- add OperationTemplate as an option in DefaultExporterResourceProvider

**PS: while reviewing, you can safely ignore the first commit, it's just an automatic refactoring (move classes) done by the IDE and the change in imports resulting from that.**

**Any changes made by me was done in the subsequent commits.**
## Related issues

closes #24255


sdorokhova (approved):
LGTM

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/operate/template/BatchOperationTemplate.java
```diff
@@ -32,21 +26,8 @@ public class BatchOperationTemplate extends OperateTemplateDescriptor implements
...
-    indexPrefix = properties.getIndexPrefix(DatabaseInfo.getCurrent());
-    isElasticsearch = DatabaseInfo.isElasticsearch();
-  }
-
-  @Override
-  public String getIndexPrefix() {
-    return properties.getIndexPrefix();
+  public BatchOperationTemplate(final String indexPrefix, final boolean isElasticsearch) {
+    super(indexPrefix, isElasticsearch);
   }
```
rodrigo-lourenco-lopes, line 32: 🔧 I think here we also need to add the getVersion() correct? here will be "8.1.0" since the original was an extension of [AbstractIndexDescriptor](https://github.com/camunda/camunda/blob/b88820c5cec26171efc9a93e761705c15dab1dec/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java).

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
```diff
@@ -84,25 +88,28 @@ public void init(final ExporterConfiguration configuration) {
...
-            new IncidentTemplate(globalPrefix, isElasticsearch),
-            SequenceFlowTemplate.class,
-            new SequenceFlowTemplate(globalPrefix, isElasticsearch),
-            DecisionInstanceTemplate.class,
-            new DecisionInstanceTemplate(globalPrefix, isElasticsearch),
-            EventTemplate.class,
-            new EventTemplate(globalPrefix, isElasticsearch),
-            TaskTemplate.class,
-            new TaskTemplate(globalPrefix, isElasticsearch));
+        Map.ofEntries(
```
mustafadagher, line 91: `Map.of()` only accepts a max of 10 entries, that's why I had to change this.

rodrigo-lourenco-lopes (approved):
Looks good 🚀 do you think that the change I suggested makes sense? Since we do get the versions in the other templates.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/operate/template/BatchOperationTemplate.java
mustafadagher: I'm happy to add it, but could you explain to me where did the `8.1.0` value come from?

rodrigo-lourenco-lopes: The original `BatchOperationTemplate` extents several classes until we have the `AbstractIndexDescriptor` which implements the getVersion( ) that returns "8.1.0".

rodrigo-lourenco-lopes: Just so that this template can return the same version as the one in operate.

ChrisKujawa (approved):

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/operate/template/BatchOperationTemplate.java
mustafadagher: This does extend the same classes, no?
And I see here that it has `1.0.0` not `8.1.0`, that's why I'm confused

https://github.com/camunda/camunda/blob/b88820c5cec26171efc9a93e761705c15dab1dec/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java#L49-L52

Zelldon: That is currently running in our benchmarks. Just do bring in some data maybe this helps in the discussion

Issue #24275, opened 2024-10-31
Add carbon breadcrumbs to reports/dashboards


OmranAbazid:
Engineering DRI: @OmranAbazid

### Context

See https://github.com/camunda/camunda-optimize/issues/6864

### Acceptance Criteria

For reports, there is a breadcrumb to back to, 
- collections home page
- Parent collection
- Parent dashboard

For dashboards, there is a breadcrumb to back to, 
- collections home page
- Parent collection

For instant dashboards and adoption dashboard reports, there is a breadcrumb to back to dashboards page 

### Testing Notes
- The description above should be sufficient

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/24276
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [x] The change is implemented as described on all target environments/versions/modes
- [x] The documentation changes are as expected

### Completion

- [ ] All Review stages are successfully completed
- [ ] All associated PRs are merged to the main branch(es) and stable branches
- [ ] The correct version labels are applied to the issue

PR #24276, opened 2024-10-31
feat: add carbon breadcrumbs

OmranAbazid:
## Description

See related issue

## Checklist

- [ ] for CI changes:

## Related issues

related to #24275

### DIFF
--- a/optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.scss
+++ b/optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.scss
@@ -1,0 +1,3 @@
+.Breadcrumbs {
+  margin: 0 40px;
+}
--- a/optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.test.tsx
+++ b/optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.test.tsx
@@ -1,0 +1,126 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+
+import {runLastEffect} from '__mocks__/react';
+import {Link, useLocation} from 'react-router-dom';
+import {shallow} from 'enzyme';
+import {BreadcrumbItem, BreadcrumbSkeleton} from '@carbon/react';
+
+import Breadcrumbs from './Breadcrumbs';
+import {getEntityId, loadEntitiesNames} from './service';
+
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useLocation: jest.fn(),
+}));
+
+jest.mock('hooks', () => ({
+  useErrorHandling: () => ({
+    mightFail: jest.fn().mockImplementation((data, cb) => cb(data)),
+  }),
+}));
+
+jest.mock('./service', () => ({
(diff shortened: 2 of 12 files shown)


optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,83 @@
...
+import {loadEntitiesNames, EntityNamesResponse, getEntityId} from './service';
+
+import './Breadcrumbs.scss';
+
+export default function Breadcrumbs() {
+  const {mightFail} = useErrorHandling();
+  const location = useLocation();
+  const [entityNames, setEntityNames] = useState<EntityNamesResponse>();
+  const collection = getEntityId('collection', location.pathname);
+  const dashboard = getEntityId('dashboard', location.pathname);
```
vsgoulart, line 26: Why can't we use something like `useMatches` from `react-router-dom`?

optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const mockUseLocation = useLocation as jest.Mock;
+const mockLoadEntitiesNames = loadEntitiesNames as jest.Mock;
+const mockGetEntityId = getEntityId as jest.Mock;
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+it('should render BreadcrumbSkeleton while entityNames is not set', () => {
```
vsgoulart, line 41: I would recommend against mentioning internal implementations details like `BreadcrumbSkeleton` and `entityNames`. It should be something like `it should render a skeleton` and then the test shows the logic necessary for that to happen so the tests are more resilient to change.

I know it's probably done like this everywhere in the Optimize codebase but we could start improving it little by little.

optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.scss
```diff
@@ -0,0 +1,3 @@
+.Breadcrumbs {
+  margin: 0 40px;
```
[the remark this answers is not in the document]
OmranAbazid: I could not do this because the breadcrumbs needs to be alighted with the rest of the layout which does not have carbon spacing, but thanks for raising the comment. I created https://github.com/camunda/camunda/issues/24716
to address this for the whole layout.

optimize/client/src/modules/components/index.js
```diff
@@ -51,3 +51,4 @@ export {InstanceViewTable} from './InstanceViewTable';
 export {TenantInfo} from './TenantInfo';
 export {TargetSelection} from './TargetSelection';
 export {Loading} from './Loading';
+export {Breadcrumbs} from './Breadcrumbs';
```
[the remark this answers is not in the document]
OmranAbazid: Thanks for sharing. I did not know about this problem.
It seems that it is not a problem for esbuild as long as we do not use "*" in imports.
For development, it will slow it down, as vite currently loads everything in barrel files. Vite is however considering improving this: https://github.com/vitejs/vite/issues/8237

I won't create an issue for this for now since we already have a lot of existing migrations we are doing. I am happy to revisit this in the future after we merge the apps. It might be good to align with other webapps on this anyway.

optimize/client/src/modules/components/Breadcrumbs/Breadcrumbs.tsx
OmranAbazid: We are still using react-router v5 which does not have `useMatches`. We can however use `useRouteMatch` but i find using it makes the solution more complicated.

vsgoulart (approved):

PR #24280, opened 2024-10-31
feat: add multi-document upload endpoint

### LINKED ISSUE #22577: [Doc Handling] Support uploading multiple files via the API
Currently the document API only supports single file upload. We would like to extend the API by providing a method to upload multiple files at the same time.

Primary use case: uploading files from Tasklist

For simplicity, we can omit the metadata for such requests.

Example request body:

```
Content-Type: multipart/form-data; boundary=---
---
Content-Disposition: form-data; name="file1"; filename="a.txt"
Content-Type: text/plain

Content of a.txt.
---
Content-Disposition: form-data; name="file2"; filename="a.html"
Content-Type: text/html

<!DOCTYPE html><title>Content of a.html.</title>

---
```

Open question:
- Should we rely on part names (e.g. file1, file2)? If so, we can also allow passing the metadata (e.g. metadata1, metadata2)
  - An alternative approach would be to ignore the part names, expect all parts to be files and not support metadata

chillleader:
## Description

This PR will introduce the batch document creation capability. This change is needed to support multiple file upload from the Tasklist form, as the form frontend communicates directly with the document API.

**Suggested batch document endpoint format:**
* `POST /documents/batch`
* Body: multipart with array of `files` and array of `metadata`
* Response:
  * All files uploaded successfully => 200
  * Some files failed to upload => 207 
  * Response body: Object with a list of uploaded and failed documents
    * `createdDocuments`: List of document references
    * `failedDocuments`: List of custom detail objects
    * The server doesn't do any compensational deletions in case of a partial failure
  * Bad request => 400, response body: normal ProblemDetail

* Java client implementation: uses a similar document builder pattern as the single document creation format (shares the same interface).

**Notes:**
* Custom document IDs not supported
* Filename is used to identify the error objects for each individual file in the multi-status response
  * The server will validate that each document has a file name in the upload request
  * ❓ what can we do with file name clashes?
    * Option 1: the controller will enforce distinct names for each file in the batch
    * Option 2: add document IDs to the request (e.g. as another request part array). Enforce that the client always provides document IDs for each document and use it as the identifier
  * I went for option 1 for now, but happy to discuss

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/22577


chillleader:
@tmetzke could you please take a quick look at this initial sketch whenever you have a moment? I wanted to get a second pair of eyes on this before proceeding with the implementation, as I haven't found any other batch operations in the API.

I'm mostly concerned about how the client would identify which error in the multi-status response corresponds to every individual file in the request. I feel like adding document IDs to the request as a required request part would allow us to solve this in a more-deterministic way, however, just having file names as an identifier would also work for the common use cases and would result in a simpler format.

tmetzke:
Thanks, @chillleader 👍 

I agree that file names should be good enough here.
What benefit would document IDs bring by themselves? Users would have to provide them with each file and they would have to be unique, wouldn't they? The same can be done with file names already in multipart form data requests that include binary data. That is a standardized format already and known to REST API users.

We actually have another endpoint that does something similar: [Deploy resources](https://github.com/camunda/camunda/blob/05704d4c1cc4f1cc63c8135f790252bc748ce003/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L2674-L2715).

Regarding document metadata, how do you ensure the metadata you are looking at matches the file entry? Potentially, users could upload 5 files but only 4 metadata items. One idea to solve this: make the metadata part of the form entries using custom headers as described in the [OpenAPI guide](https://swagger.io/docs/specification/v3_0/describing-request-body/multipart-requests/#specifying-custom-headers).

qa/integration-tests/src/test/java/io/camunda/it/client/CreateDocumentBatchTest.java
```diff
@@ -0,0 +1,311 @@
...
+    // given
+    zeebeClient = testStandaloneCamunda.newClientBuilder().build();
+    final var documentContent1 = new ByteArrayInputStream("test one".getBytes());
+
+    // when
+    final var response =
+        zeebeClient
+            .newCreateDocumentBatchCommand()
+            .addDocument()
+            .fileName("file1.txt")
```
mo-okocha, line 104: Filename is supplied here, so is the test is redundant. Can we remove the filename to validate this

service/src/main/java/io/camunda/service/DocumentServices.java
```diff
@@ -53,18 +57,44 @@ public CompletableFuture<DocumentReferenceResponse> createDocument(
...
+    final List<CompletableFuture<Void>> futures =
+        requests.stream()
+            .map(
+                request -> {
+                  final var storeRequest =
+                      new DocumentCreationRequest(
+                          request.documentId, request.contentInputStream, request.metadata);
+                  return getDocumentStore(request.storeId)
+                      .thenCompose(
+                          storeRecord -> storeRecord.instance().createDocument(storeRequest))
```
mo-okocha, line 83: Since we are doing batch uploads, it would be better to update the `AwsDocumentStore` and `GcpDocumentStore` to increase the executor threadpool to avoid bottlenecks. They both use `Executors.newSingleThreadExecutor()` at the moment which will make our batch uploads sequential

chillleader: Good point! I added a way to customize the executor service when creating the store. We also use a common thread pool now for all stores.

mo-okocha (approved):
Everything looks good overall. There are a few failing tests that still need to be addressed, but I'm going ahead and approving it.

Issue #24218, opened 2024-10-30
Consistent date-time format in request & result


danielkelemen:
## Description

REST API `date-time` format should be consistent both in what the APIs accept and what they return.

Currently, it looks like we only accept ISO-8601 ([RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6)) but return a slightly different format without colon in the offset part (`2024-06-05T08:29:15.027+0000`). This wouldn't be accepted in a request as we use [OffsetDateTime](https://docs.oracle.com/javase/8/docs/api/index.html?java/time/OffsetDateTime.html) and that expects a colon!

## Acceptance criteria

- Accept and return dates in the same date time format.
- This format should be the [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) specified one that the [OpenAPI date-time](https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings) also uses.
  - Accepts `2024-06-05T08:29:15.027+00:00`, `2024-06-05T08:29:15.027Z`

```[tasklist]
### Tasks

```

tmetzke:
## Assumption

We use the date string we receive from the data layer (Elasticsearch, OpenSearch, RDBMS) and return it as is.

## Proposal

* For date-time types, we work with `OffsetDateTime` in the search layer, not `String`
* The date-time String we fetch from the data layer is parsed into an `OffsetDateTime` so we don't have to care about different formats from the data layer and ship a consistent value in the APIs.

PR #24289, opened 2024-11-01
feat: Extend License for isCommercial and expiresAt

### LINKED ISSUE #24092: C8 License - Extend License Endpoint for isCommercial and expiresAt
## Description

As part of the [Licence expiration for non-commercial use](https://github.com/camunda/product-hub/issues/2529) project, two new fields need to be added to the `/license` endpoint. These two fields will determine whether or not a license is for commercial or non-commerical use, and for determining the expiration date of a license

The two fields: `isCommercial` and `expiresAt` 

- `isCommercial`
  - This will be a boolean field
  - Will be `false` when a license contains a `non-commerical=true` property 
  - Will be `false` otherwise
- `expiresAt`
  - This will be a `string` field
  - The value will be a unix timestamp representing when a license will become expired
  - Will be `null` in the case where the `expiryDate` field is not present on a license

An example payload for the endpoint would look something like this

```
GET /license
{
   "validLicense": true,
   "environment": "self-managed",
   "isCommercial": true,
   "expiresAt": "1730055658847"
}
```

matthewBearCamunda:
## Description

Adds functionality for `non-commerical` licenses.

- Upgrades the `camunda-license-checker` library
- `non-commerical` and `expiresAt` are added to the `/license` endpoint

## Checklist

- [ ] for CI changes:

## Related issues

closes #24092


matthewBearCamunda:
@tmetzke - If you could review this PR please. You reviewed the [previous license API PR](https://github.com/camunda/camunda/pull/21156), figured you'd be a good one of this too

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3960,6 +3960,13 @@ components:
         licenseType:
           description: Will return the license type property of the Camunda license
           type: string
+        isCommercial:
+          description: Will be false when a license contains a non-commerical=true property
+          type: boolean
+        expiresAt:
+          description: The time in UTC when the Camunda license expires
```
tmetzke, line 3967: 🔧 Tying this to UTC by definition is not needed and makes it a breaking change if we ever return anything else. The offset date time that will be returned here contains offset information. This will be sufficient for any consuming application already.

service/src/main/java/io/camunda/service/license/CamundaLicense.java
```diff
@@ -37,6 +46,14 @@ public synchronized LicenseType getLicenseType() {
...
   }
 
+  public synchronized boolean isCommercial() {
+    return isCommercial;
+  }
+
+  public synchronized String expiresAt() {
+    return DATE_TIME_FORMATTER.format(validUntil);
+  }
+
```
tmetzke, line 56: 🔧 Let the consumers format this information as needed. The license shouldn't care about this.

service/src/main/java/io/camunda/service/ManagementServices.java
```diff
@@ -25,4 +25,12 @@ public boolean isCamundaLicenseValid() {
...
     return license.getLicenseType();
   }
+
+  public boolean isCommercialCamundaLicense() {
+    return license.isCommercial();
+  }
+
+  public String getCamundaLicenseExpirationDate() {
+    return license.expiresAt();
+  }
```
tmetzke, line 35: 🔧 Let the consumers returning data care about the format, not the service layer. That way, consumers have the opportunity to provide dates consistently. Otherwise, they can either only ship formatted date strings as received (bad UX) or parse and reformat (horrible implementation).

tmetzke (commented):
👍 Looks really good already.
🔧 I would leave the date formatting to the API, ultimately. It should be able to decide which format to ship.
ℹ️ I want to make date formatting in the API more consistent with https://github.com/camunda/camunda/issues/24218.

tmetzke (approved):
👍 Thanks for the swift adjustments here. Looks good to me 🦖 
🙃 I'm thinking whether `OffsetDateTime` or `Instant` is more appropriate here. I think it's the former but I'd be interested in your opinion here.

matthewBearCamunda:
@tmetzke - I went with `Instant` as the docs label it as the simplest representation of a moment in time, and it can be easily converted into `OffsetDateTime` and `ZonedDateTime`

Ultimately, I'm happy using either as long as we're consistent with picking one or the other

tmetzke:
> @tmetzke - I went with `Instant` as the docs label it as the simplest representation of a moment in time, and it can be easily converted into `OffsetDateTime` and `ZonedDateTime`
> 
> Ultimately, I'm happy using either as long as we're consistent with picking one or the other

I understand, makes sense to me in general, thanks, @matthewBearCamunda 👍 
I'd vote for `OffsetDateTime` to be as consistent as possible in the API and services. We can directly create this from the `java.util.Date` the license checker returns. The implicit "this is based on UTC" notion the license checker result contains is then made explicit in our code/platform.

PR #24294, opened 2024-11-01
feat: enforce schema property descriptions in c8 rest spec

pepopowitz:
## Description

1. Adds a vacuum rule that requires a `description` on every property in a schema.
2. Adds a `description` to all properties that fail this new rule.
3. Modifies a few search query schemas to use `allOf` inheritance on their `filter` property, to accommodate [this issue from the docs](https://github.com/camunda/camunda-docs/issues/4610#issuecomment-2483454870). 

## What's not included

I considered also adding a rule for operations that define request/response properties inline instead of through a schema. I decided against this, because there are not very many operations that define properties this way -- and [only one inline property right now does not have a `description`](https://github.com/camunda/camunda/blob/f0a7064d6a1370c83618b2749377e2970d99c4b6/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L2303). Let me know if you think I made the wrong choice, and should include a rule for this.

## I learned some things about schema inheritance, and I'm not sure this PR ended up in the right place. I think we are forced to choose the least bad option.

**Edit:** we concluded that we would prefer `allOf` inheritance over direct `$ref`, especially because of https://github.com/camunda/camunda-docs/issues/4610#issuecomment-2483454870. 

See [this thread](https://github.com/camunda/camunda/pull/24294#discussion_r1835123062) for more details. 

Basically, I ended up converting a handful of `allOf` inherited schemas to direct `$ref` inheritance. During some investigation about specific properties I came to better understand the difference between these two approaches, and I felt this better portrayed the schema. 

However, the conversion _does_ propagate a pattern that _could_ result in a non-obvious crash for developers, if they introduce direct `$ref` inheritance of a schema that lacks a `description`. 

On the other hand, with `allOf` inheritance, if a property is lacking a `description` in the parent schema, there is no crash, but _vacuum still fails with a red state but no description why_. 

In addition, the docs render the inherited description of a property that uses direct `$ref` inheritance. They do not render an explicit description on an `allOf` inherited property.

So it's tradeoffs. Here's a table defining the approaches, consolidating the pros/cons that I just described.

| Inheritance approach | Pros | Cons |
| -- | -- | -- | 
| direct `$ref` | a more accurate portrayal of the schema | crashes vacuum if we miss a description on an inherited schema |
| | displays the inherited schema's description in the docs | |
| `allOf` | does not crash vacuum if a property is missing a description | fails validation with no error if a property is missing a description  |
|  |  | does not display the property description in the docs | 

One other thought is that turning the `component-description` to `error` severity might result in fewer crashes. This doesn't prevent the crash _directly_; if an inherited schema lacks a property, the crash happens rather than the `component-description` violation being emitted. It _would_ require us to fill in more schema-level descriptions, though, which would give developers better examples to copy-paste when they add a schema.

I don't know....what do you think???


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2865,6 +2849,7 @@ components:
       type: object
       properties:
         items:
+          description: sjh
```
pepopowitz, line 2852: Each of these is a fake description placeholder, which I will fill in as part of this PR.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3662,55 +3657,13 @@ components:
         tenantId:
           description: The tenant ID of the incident.
           type: string
-    OperationItem:
```
pepopowitz, line 3665: Not actually used in the spec, so I deleted them. (It's a long story why I found the ones that aren't used.....coming in another comment.)

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4386,46 +4363,34 @@ components:
...
+          description: sjh
           type: array
           items:
             type: object
         lastSortValues:
+          description: sjh
           type: array
           items:
             type: object
-    VariableValueFilterRequest:
```
pepopowitz, line 4406: Not actually referenced anywhere in the spec.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4862,7 +4831,14 @@ components:
...
         metadata:
-          $ref: "#/components/schemas/DocumentMetadata"
+          description: sjh
+          # sjhsjhsjh -- this is different from the source!
+          #   it needs to use allOf to consume DocumentMetadata.
+          #   I'm not sure why, because the oas spec seems to say applying the $ref without allOf is fine...
+          #   but we did allOf in every other schema for some reason, so this is the only one that failed.
+          type: object
+          allOf:
+            - $ref: "#/components/schemas/DocumentMetadata"
```
pepopowitz, line 4841: oh boy, this was an adventure. 

* I was getting weird crashes from vacuum with my new rule added, on these properties that pointed directly at a `#ref`. 
   * The crash message was `panic: interface conversion: *base.ObjectReference is not base.Foundational: missing method AddEdge`
* To track it down I removed the entire spec and added it back in a little at a time until I identified the culprit...that's how I found the objects that weren't used (mentioned in a previous comment).
* I don't know _why_ my rule was crashing on these properties. The rule finds all property nodes, and then checks for the existence of a `description` property, none of which seems to relate to the `$ref`. 
* I don't think that the direct `$ref` (the left side of this diff) is a violation of the OpenAPI spec. 
* But weirdly, we do this `allOf` inheritance instead of direct `$ref` for most properties already anyway, and the `allOf` inheritance works fine with the rule....
* Since there are only a few cases where we are using direct `$ref` instead of `allOf` inheritance, it seems reasonable to just change those few cases.
* My big concern is that if someone attempts to add a direct `$ref`, the output from vacuum is not helpful at all.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4912,22 +4886,32 @@ components:
...
+          description: sjh
           type: string
     DeploymentMetadata:
       type: object
       properties:
         processDefinition:
-          $ref: "#/components/schemas/DeploymentProcess"
+          description: sjh
+          allOf:
+            - $ref: "#/components/schemas/DeploymentProcess"
```
pepopowitz, line 4902: Same adventure as described above.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
pepopowitz: ok, I figured out why this was happening, or at least, how this example was different from others.

1. If a property inherits another schema via `allOf`, then a `description` is required on the property, or else the new rule will fail (clearly) for this property. 
2. If a property inherits another schema via a direct `$ref`, then no `description` is required on the _property_. A `description` _is_ required on the inherited schema, though, and if it doesn't exist, the vacuum rule will crash. 

I actually think that direct `$ref` inheritance is more accurate in describing the spec, so I've changed many instances of `allOf` inheritance to direct `$ref` inheritance. _However..._ this spreads a pattern in the spec that _could_ result in a vacuum crash, which I don't feel great about. I didn't consider this until after I'd already started converting to direct`$ref`s though, and I wanted to be consistent with the changes I was making, so that's where I ended up. We should discuss this tradeoff though, and choose a preference.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -337,7 +337,6 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
-
```
pepopowitz, line 340: Sorry for this noise. I was using empty lines to track my progress. I can put these back if you'd prefer.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3093,9 +3080,9 @@ components:
       type: object
       properties:
         filter:
-          allOf:
-            - $ref: "#/components/schemas/ProcessDefinitionFilterRequest"
+          $ref: "#/components/schemas/ProcessDefinitionFilterRequest"
```
pepopowitz, line 3083: This is an example of my inheritance conversion. 

The left side would have required a `description` on the `filter` property; that description would never be emitted in the docs. The lack of it also caused vacuum to fail validation...with no message pointing to this line. 

The right side requires a description on the inherited `ProcessDefinitionFilterRequest` schema (and if it's not there, vacuum crashes 😅 😬 ). The description in the inherited schema does display in the docs.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4347,37 +4308,45 @@ components:
...
+          description: The maximum number of items to return in one request.
           type: integer
           format: int32
         searchAfter:
+          description: Items to search after. Correlates to the `lastSortValues` property of a previous search response.
           type: array
           items:
             type: object
         searchBefore:
+          description: Items to search before. Correlates to the `firstSortValues` property of a previous search response.
```
pepopowitz, line 4338: Let me know if I got these descriptions incorrect. I took somewhat-educated guesses.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4386,49 +4355,34 @@ components:
...
+          description: Total items matching the criteria.
           type: integer
           format: int64
         firstSortValues:
+          description: The first items in the results. Correlates to the `searchBefore` field of an ensuing request.
           type: array
           items:
             type: object
         lastSortValues:
+          description: The last items in the results.  Correlates to the `searchAfter` field of an ensuing request.
```
pepopowitz, line 4373: Let me know if these descriptions are inaccurate, too.

pepopowitz:
It also looks like I have some merge conflicts, which makes sense because I started this branch like two weeks ago 😬. I'll sort those out next week.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: So, this vacuum behavior makes a lot of sense in general. If you use a `$ref`, all sibling attributes are ignored because a ref is meant to be a drop-in that will be expanded in this place and "overrides" whatever was defined as siblings to it. Thus, the description must sit in the `$ref` as well.

If you go for an `allOf`, you must provide the description at the property itself. The type of the property described by an `allOf` is basically an anonymous, custom type that needs a description itself.

So, I would go about it this way:
* If there is a dedicated description at the property, referencing one `$ref` in an `allOf`, keep it as is because that property wants to define a dedicated description, overriding that of the `$ref`.
* If there is no description at an `allOf` property using one `$ref`, make it a direct `$ref` and use the description from that reference.

Let me know if that makes sense or if I overlooked anything.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: This makes a lot of sense and this is how it should be 👍 
The crashing vacuum is an issue, of course 😅

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: As far as I can see, this is absolutely correct 👍

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke:
```suggestion
          description: The sort values of the first item in the result set. Use this in the `searchBefore` field of an ensuing request.
          type: array
          items:
            type: object
        lastSortValues:
          description: The sort values of the last item in the result set. Use this in the `searchAfter` field of an ensuing request.
```

.github/vacuum-ruleset.yaml
```diff
@@ -602,3 +602,14 @@ rules:
...
+        category:
+            description: Documentation is really important, in OpenAPI, just about everything can and should have a description. This set of rules checks for absent descriptions, poor quality descriptions (copy/paste), or short descriptions.
+            id: descriptions
+            name: Descriptions
+        description: Schema properties must have a description.
+        given: $.components.schemas[*].properties[*]
+        then:
+            field: description
+            function: truthy
+        severity: error
```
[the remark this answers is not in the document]
pepopowitz: Yes, this is checking something different than each of those rules. The component rule checks only for a description on the schema object itself; the parameter rule checks only for descriptions on operation parameters but not the schemas they implement; the operation rule checks only the description of the operation itself.

I was surprised to not find a rule that did this already, but I couldn't see anything in [the docs](https://quobix.com/vacuum/rules/descriptions/).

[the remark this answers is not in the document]
pepopowitz: I did also take a look at [one of the existing rules](https://github.com/daveshanley/vacuum/blob/main/functions/openapi/operation_descriptions.go) to see if it might make sense to try to recreate it but pointed at schema properties instead, and it felt a little more complicated than I'm comfortable with, having zero golang experience. 

If you think this is worth pursuing though, let me know, and I can try to make it happen.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
pepopowitz: What you're saying makes sense, and I _think_ we're in alignment in terms of when to use which. 

I also _think_ this means I shouldn't revert any of my changes, because I don't see any where it seems more semantic to describe the schema property instead of the inherited component. I may have misinterpreted.

pepopowitz:
> I'm a few steps removed from the schema work. Could you let me know what you're looking for with my review here?

@akeller I'm looking for (1) your awareness aware of the work that's happening here, and (2) your general approval that this seems like a good thing to automate.

akeller (commented):
This is my handwavy approval and acknowledgment. 

Having a rule for this does make sense, but I can see on the PR you are working on the details of exactly what that means. Ideally it doesn't crash or fail silently, but if we have to accept either unintuitive behavior, let's make sure it's documented so API authors can save themselves.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
pepopowitz: Heads up that I found [an issue in how the docs generator handles these two different approaches](https://github.com/camunda/camunda-docs/issues/4610). I think I'm changing my opinion, to always use `allOf`, because it doesn't wreck the docs right now 😅. 

I'm going to come back and update in this PR.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4180,37 +4157,45 @@ components:
...
         page:
+          description: Pagination criteria.
           allOf:
             - $ref: "#/components/schemas/SearchQueryPageRequest"
           type: object
     SearchQueryPageRequest:
       type: object
       properties:
         from:
+          description: The index to start searching from.
```
pepopowitz, line 4174:
```suggestion
          description: The index of items to start searching from.
```

A little more clarity, inspired by https://camunda.slack.com/archives/C06UKS51QV9/p1731939210511069.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
pepopowitz: PR is updated to revert these changes!

pepopowitz:
Ok @tmetzke, sorry for the churn on this. This is now cleaned up, with noise removed from the diff, and ready for a final review.

pepopowitz:
@tmetzke I'm not comfortable merging this without one more review/approval from you. No pressure, I just want to make sure it didn't slip through the cracks.

tmetzke:
> @tmetzke I'm not comfortable merging this without one more review/approval from you. No pressure, I just want to make sure it didn't slip through the cracks.

Sorry, I couldn't look at this again so far due to internal events. I'll have a look next Wednesday, hope that's fine 🙈

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: 🚀

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: All good 👍

.github/vacuum-ruleset.yaml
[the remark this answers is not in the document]
tmetzke: OK, I see. Good to have your additional rule, no need to go for adding it any differently 👍

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4590,49 +4599,38 @@ components:
       type: object
       properties:
         page:
+          description: Pagination information about the search results.
           allOf:
             - $ref: "#/components/schemas/SearchQueryPageResponse"
-          type: object
```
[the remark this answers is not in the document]
tmetzke: Shouldn't make any difference, I think

tmetzke (approved):
👍 Looks good to me, thanks for pursuing this, @pepopowitz 🚀 
💭 Let's see if we get stuck with this anywhere in the near future. We can document this in the [REST developer guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md).

tmetzke:
The spec items you removed are referenced in the Java code apparently, this is why everything fails in CI.
I'll have a quick look if this is really needed there. If not, I'll clean up the code so this passes. If it's more complex, let's leave the spec items in there although they shouldn't be used if they are not referenced.

tmetzke:
I removed some unused code (at least I think so) that referenced the spec classes. Let's see.

pepopowitz:
Ah, I was wondering if so many checks were usually that red 😅 

Thanks for fixing!

tmetzke:
By the way, regarding the `$ref` and `allOf` topic, this might be interesting as well: https://redocly.com/docs/resources/all-of#siblings-to-ref-s

PR #24310, opened 2024-11-04
Migrate BatchOperations archive job to Camunda Exporter

### LINKED ISSUE #24093: Migrate Operate archiving jobs
**Description**

Migrate the Operate Archiver jobs, namely `ProcessInstancesArchiverJob` and `BatchOperationArchiverJob`. The issue is to de-Springify them - injecting via constructor instead of using autowiring - and then construct them or feed them to the archiver (to be decided during kickoff/implementation).

If the repository is not yet migrated, the jobs can already be tested using a test implementation of the `ArchiverRepository`, though no E2E tests can be added yet.

rodrigo-lourenco-lopes:
## Description

This PR migrates the BatchOperationArchiverJob to the Camunda Exporter with no archiving logic yet implemented. Therfore the E2E will be added later. 

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/24093

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/metrics/CamundaExporterMetrics.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/metrics/CamundaExporterMetrics.java
@@ -23,6 +23,7 @@ public class CamundaExporterMetrics {
   private final AtomicInteger bulkMemorySize = new AtomicInteger(0);
   private final Timer flushLatency;
   private final Counter processInstancesArchived;
+  private final Counter batchOperationsArchived;
   private Timer.Sample flushLatencyMeasurement;
 
   public CamundaExporterMetrics(final MeterRegistry meterRegistry) {
@@ -35,6 +36,7 @@ public class CamundaExporterMetrics {
             .publishPercentileHistogram()
             .register(meterRegistry);
     processInstancesArchived = meterRegistry.counter(meterName("archived.process.instances"));
+    batchOperationsArchived = meterRegistry.counter(meterName("archived.batch.operations"));
   }
 
   public ResourceSample measureFlushDuration() {
@@ -79,6 +81,11 @@ public class CamundaExporterMetrics {
 
   public void recordProcessInstancesArchived(final int count) {
     processInstancesArchived.increment(count);
+  }
+
+  // counts the number of operations in the batch.
+  public void operationsArchived(final int count) {
+    batchOperationsArchived.increment(count);
   }
 
   private String meterName(final String name) {
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/metrics/CamundaExporterMetrics.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/metrics/CamundaExporterMetrics.java
@@ -83,8 +83,7 @@ public class CamundaExporterMetrics {
     processInstancesArchived.increment(count);
   }
 
-  // counts the number of operations in the batch.
-  public void operationsArchived(final int count) {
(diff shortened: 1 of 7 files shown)


rodrigo-lourenco-lopes:
@npepinpe because our `BatchOperationTemplate extends OperateTemplateDescriptor` we get our destination index name being generated as: `operate-batch-operation-(...)`. Since the template is no longer exported through the archiver should we still have the prefix `operate`? 
image

npepinpe:
Good catch, I think the prefix shouldn't be there, but I'm not yet super familiar with the new structure. Can you check with @mustafadagher ?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/metrics/CamundaExporterMetrics.java
```diff
@@ -81,6 +83,11 @@ public void recordProcessInstancesArchived(final int count) {
     processInstancesArchived.increment(count);
   }
 
+  // counts the number of operations in the batch.
```
npepinpe, line 86: :x: Comments should be for why, not what/how. One exception is if the how is quite complex (for good reasons). But in this case, I don't think this comment is useful. 

:wrench: Since we have both Operations and BatchOperations in Operate, I would propose we rename this method `batchOperationsArchived`.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/TestRepository.java
npepinpe, line 1: :+1: Good initiative to extract it

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/metrics/CamundaExporterMetrics.java
```diff
@@ -35,6 +36,7 @@ public CamundaExporterMetrics(final MeterRegistry meterRegistry) {
             .publishPercentileHistogram()
             .register(meterRegistry);
     processInstancesArchived = meterRegistry.counter(meterName("archived.process.instances"));
+    batchOperationsArchived = meterRegistry.counter(meterName("archived.batch.operations"));
```
npepinpe, line 39: :question: This is a new metric, right? IIRC there was no existing metric for the count of batch operations archived?

npepinpe (approved):
:+1: Good stuff! Please address all comments, but no need for another review

mustafadagher:
> Good catch, I think the prefix shouldn't be there, but I'm not yet super familiar with the new structure. Can you check with @mustafadagher ?

@rodrigo-lourenco-lopes @npepinpe 
If you see `AbstractIndexDescriptor`, you will notice that `getFullQualifiedName` adds the `componentName` to the index name. `OperateTemplateDescriptor` sets the `ComponentName` to `operate`. So it will probably be there.
https://github.com/camunda/camunda/blob/3f78d0a6adf4602e31e5523b514fb9cde4b094a5/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java#L28-L31

rodrigo-lourenco-lopes:
@mustafadagher does this mean we should keep the index name as is? or should this template not extend `OperateTemplateDescriptor `?

mustafadagher:
> @mustafadagher does this mean we should keep the index name as is? or should this template not extend `OperateTemplateDescriptor `?

Are you setting the `index name` anywhere in this PR? I can't spot that out.

I think `BatchOperationTemplate` is/was only used by operate, no? or was there a similar one used in other apps too that you need to take care of?

If it's the first. then you don't need to do anything. the index name will still be the same if you're relying on `getFullQualifiedName`

PR #24319, opened 2024-11-04
Do not create archiver if rollover is not enabled

### LINKED ISSUE #24290: Do not create archiver if rollover is not enabled
# Description

In the `CamundaExporter`, we shouldn't create any archiver (leave it null) if `ArchiverConfiguration#isRolloverEnabled` is false.

Not exactly sure how we would test this, but considering the logic is quite simple, we can omit a test there I guess.

rodrigo-lourenco-lopes:
## Description

This PR checks the archiver configuration `IsRolloverEnabled` to determine if we should create the archiver. 

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/24290

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -92,14 +92,17 @@ public class CamundaExporter implements Exporter {
             provider.getIndexDescriptors(),
             provider.getIndexTemplateDescriptors(),
             configuration);
-    archiver =
-        Archiver.create(
-            context.getPartitionId(),
-            context.getConfiguration().getId().toLowerCase(),
-            configuration.getArchiver(),
-            provider,
-            metrics,
-            logger);
+
+    if (configuration.getArchiver().isRolloverEnabled()) {
+      archiver =
+          Archiver.create(
+              context.getPartitionId(),
+              context.getConfiguration().getId().toLowerCase(),
+              configuration.getArchiver(),
+              provider,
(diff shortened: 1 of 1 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -166,6 +168,29 @@ final var record = factory.generateRecord(ValueType.USER);
     verify(controllerSpy).updateLastExportedRecordPosition(record2.getPosition());
   }
 
+  @TestTemplate
+  void shouldNotCreateArchiverIfRolloverIsDisabled(
```
npepinpe, line 172: :x: For an integration test, you would verify via the actual result, e.g. nothing is getting archived. But we don't have an easy way of asserting the "absence" of something.

:wrench: This would fit better a unit test, as we don't need the whole set up to really guarantee this, unless we want to actually assert things against the DB. I understand there's no existing unit test class for this, so I can see where you're coming from. Plus, I don't really know how we'd write a unit test here. Normally, you would pass in the pre-built resource (e.g. Archiver in this case), and then have the exporter do stuff with it, and then check. But in this case, we don't want to create it. Maybe one option is to check if we created the background threads? That's _kind_ of breaking the abstraction, but not too much...

Considering we will have post importer and other things in the archiver soon, and rollover only applies to archiver, I would propose leaving the test out for now, but keeping the production code in, since it's very simple. We'll reassess after about testing once we know how we'll split the post-importer and archiver jobs.

npepinpe (approved):

PR #23864, opened 2024-10-22
feat: exporters support role create/update events

matthewBearCamunda:
## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes https://github.com/camunda/camunda/issues/23185

PR #24324, opened 2024-11-04
feat: camunda exporters support tenant create update events

### LINKED ISSUE #23135: Camunda exporters support tenant create events

Oleksiivanov:
## Description

- Added TenantIndex to descriptors map
- Created camunda-tenant.json for tenant indexing
- Implemented TenantEntity with fields matching TenantRecord
- Added TenantHandler for tenant export handling
- Added test to validate tenant export

## Checklist

- [ ] for CI changes:

## Related issues

closes #23135

### DIFF
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/TenantIndex.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/TenantIndex.java
@@ -16,8 +16,7 @@ public class TenantIndex extends UserManagementIndexDescriptor {
   public static final String TENANT_KEY = "tenantKey";
   public static final String TENANT_ID = "tenantId";
   public static final String NAME = "name";
-  public static final String ENTITY_KEY = "entityKey";
-  public static final String ENTITY_TYPE = "entityType";
+  public static final String ASSIGNED_MEMBER_KEYS = "assignedMemberKeys";
 
   public TenantIndex(final String indexPrefix, final boolean isElasticsearch) {
     super(indexPrefix, isElasticsearch);
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/TenantIndex.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/TenantIndex.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.webapps.schema.descriptors.usermanagement.index;
+
+import io.camunda.webapps.schema.descriptors.usermanagement.UserManagementIndexDescriptor;
+
+public class TenantIndex extends UserManagementIndexDescriptor {
+  public static final String INDEX_NAME = "tenant";
+  public static final String INDEX_VERSION = "8.7.0";
+
+  public static final String TENANT_KEY = "tenantKey";
+  public static final String TENANT_ID = "tenantId";
+  public static final String NAME = "name";
+  public static final String ENTITY_KEY = "entityKey";
+  public static final String ENTITY_TYPE = "entityType";
+
+  public TenantIndex(final String indexPrefix, final boolean isElasticsearch) {
+    super(indexPrefix, isElasticsearch);
+  }
+
+  @Override
+  public String getVersion() {
+    return INDEX_VERSION;
+  }
+
+  @Override
+  public String getIndexName() {
+    return INDEX_NAME;
+  }
+}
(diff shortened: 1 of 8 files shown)


webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/TenantIndex.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class TenantIndex extends UserManagementIndexDescriptor {
+  public static final String INDEX_NAME = "tenants";
+  public static final String INDEX_VERSION = "8.7.0";
+
+  public static final String TENANT_KEY = "tenantKey";
+  public static final String TENANT_ID = "tenantId";
+  public static final String NAME = "name";
+  public static final String ENTITY_KEY = "entityKey";
+  public static final String ENTITY_TYPE = "entityType";
```
Ben-Sheppard, line 20: 🔧 Sorry @Oleksiivanov this is my fault, in the ES indexes we will want a field that contains a list of the assigned entities, for example [this is the same in roles](https://github.com/camunda/camunda/pull/23864/files#diff-b1529f330014dfe26200f492ceae14fc9c3999d18607958db41fa2d2e6356ca1R18), the record fields contain these two you have here as a way to provide members to be assigned to the tenant.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/TenantEntity.java
```diff
@@ -0,0 +1,78 @@
...
+public class TenantEntity extends AbstractExporterEntity<TenantEntity> {
+
+  public static final String DEFAULT_TENANT_IDENTIFIER = "<default>";
+
+  private String id;
+  private Long tenantKey;
+  private String tenantId;
+  private String name;
+  private Long entityKey;
+  private String entityType;
```
Ben-Sheppard, line 21: 🔧 This will change based on my previous comment

Ben-Sheppard (commented):
Just a comment around the fields in use otherwise the code looks as expected!

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/TenantEntity.java
```diff
@@ -0,0 +1,78 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.webapps.schema.entities.usermanagement;
+
+import io.camunda.webapps.schema.entities.AbstractExporterEntity;
+
+public class TenantEntity extends AbstractExporterEntity<TenantEntity> {
+
+  public static final String DEFAULT_TENANT_IDENTIFIER = "<default>";
```
emulbreh, line 14: When is this used vs `TenantOwned.DEFAULT_TENANT_IDENTIFIER`?

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/TenantIndex.java
emulbreh, line 20: What do we use these static fields for? I've also noticed them in the other index descriptors, but they almost all seem to be unused.

webapps-schema/src/main/resources/schema/elasticsearch/create/index/identity-tenants.json
```diff
@@ -0,0 +1,25 @@
...
+      },
+      "name": {
+        "type": "keyword"
+      },
+      "entityKey": {
+        "type": "long"
+      },
+      "entityType": {
+        "type": "keyword"
+      }
```
emulbreh, line 22: What does the entityKey+entityType reference point to? I'm a bit confused because I would either expect these documents to describe a named tenant entity or a "tenant annotation" on other entities, which I wouldn't expect to contain a denormalized tenant name.

emulbreh (commented):
This looks good to me, as far as I understand it. I've left a few questions, but those are mainly for my understanding.

webapps-schema/src/main/resources/schema/elasticsearch/create/index/identity-tenants.json
Ben-Sheppard: In this index definition these fields aren't really relevant, in the Zeebe engine there is a TenantRecord, the fields exist there as a way for the broker to send commands to add (or remove) an entity (member) from a tenant. The field stored in ES/OS would actually be a collection of all the entities (members) assigned to this specific tenant.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/TenantEntity.java
Ben-Sheppard: Good question, as we already have the protocol as a dependency we could just use `TenantOwned.DEFAULT_TENANT_IDENTIFIER` it would remain consistent as that enum is also used elsewhere i.e. the RequestMapper class.

In fact the parent class also suggests this too :) https://github.com/camunda/camunda/blob/main/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/AbstractExporterEntity.java#L20

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/TenantIndex.java
Ben-Sheppard: Very good question, I introduced them initially in other index classes but that was more for alignment (and at a time when the new Exporter code was new to me :D ) - I agree actually though that for our implementations here, these fields seem irrelevant (in the older index classes i.e. DecisionIndex the static variables are used elsewhere in the code.

Ben-Sheppard: Its possible that these can be deleted so I'd leave them as they are for now :)

Ben-Sheppard: Do nothing for now and we can add in a clean up PR if necessary

Ben-Sheppard (approved):
Thank you for addressing my comments! Lets get this in and continue with progress :D

Issue #24262, opened 2024-10-31
Operate loses connection to AWS Opensearch


sdorokhova:
### Describe the bug

When running Operate with AWS Opensearch, it starts well but after a while starts logging exceptions in logs and Operate UI stops responding. (Exception in comment)
```
2024-10-31 13:05:23.419 ERROR 7 --- [ecords_reader_2] i.c.o.s.o.c.s.ZeebeRichOpenSearchClient  : Failed to search index: [zeebe-record-decision]! Reason: java.lang.Throwable: Acquire operation took longer than the configured maximum time. This indicates that a request cannot get a connection from the pool within the specified maximum time. This can be due to high request rate.
Consider taking any of the following actions to mitigate the issue: increase max connections, increase acquire timeout, or slowing the request rate.
Increasing the max connections can increase client throughput (unless the network interface is already fully utilized), but can eventually start to hit operation system limitations on the number of file descriptors used by the process. If you already are fully utilizing your network interface or cannot further increase your connection count, increasing the acquire timeout gives extra time for requests to acquire a connection before timing out. If the connections doesn't free up, the subsequent requests will still timeout.
If the above mechanisms are not able to fix the issue, try smoothing out your requests so that large traffic bursts cannot overload the client, being more efficient with the number of times you need to call AWS, or by increasing the number of hosts sending requests.

java.lang.RuntimeException: java.lang.Throwable: Acquire operation took longer than the configured maximum time. This indicates that a request cannot get a connection from the pool within the specified maximum time. This can be due to high request rate.
Consider taking any of the following actions to mitigate the issue: increase max connections, increase acquire timeout, or slowing the request rate.
Increasing the max connections can increase client throughput (unless the network interface is already fully utilized), but can eventually start to hit operation system limitations on the number of file descriptors used by the process. If you already are fully utilizing your network interface or cannot further increase your connection count, increasing the acquire timeout gives extra time for requests to acquire a connection before timing out. If the connections doesn't free up, the subsequent requests will still timeout.
If the above mechanisms are not able to fix the issue, try smoothing out your requests so that large traffic bursts cannot overload the client, being more efficient with the number of times you need to call AWS, or by increasing the number of hosts sending requests.
        at org.opensearch.client.transport.aws.AwsSdk2Transport.performRequest(AwsSdk2Transport.java:204) ~[opensearch-java-2.5.0.jar!/:?]
[...]

PR #24334, opened 2024-11-04
fix: Switch back to Apache client for AWS Opensearch requests

ThorbenLindhauer:
## Checklist

- [ ] for CI changes:

## Related issues

related to #24262

### DIFF
--- a/operate/common/pom.xml
+++ b/operate/common/pom.xml
@@ -126,7 +126,7 @@
 
     <dependency>
       <groupId>software.amazon.awssdk</groupId>
-      <artifactId>netty-nio-client</artifactId>
+      <artifactId>apache-client</artifactId>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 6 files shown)


houssain-barouni (commented):
Hi @ThorbenLindhauer ,
it seems that in Tasklist and Operate we have two Opensearch client beans: sync and async.
I wonder if we can use `ApacheHttpClient` for the sync client, and use it in the archiver. And, keep the async client using `NettyNioAsyncHttpClient` and use it to clear scroll that works only with it ?

ThorbenLindhauer:
@houssain-barouni interesting idea, I hadn't considered that.
I'm a bit reluctant though to do that, because this would give us also overall more dependencies, more CVEs, more bugs that we are potentially affected by. No super specific reason, but I think it's just good practice to only have one Http client library towards AWS Opensearch. I also didn't check if we have this problematic pattern with the archiver in other places where we might still use the async client.

houssain-barouni (approved):
Hoping the ClearScroll issue gets solved soon with the apache client
It seems there is some progress about that https://github.com/opensearch-project/opensearch-java/pull/1256

kristinkomschow (approved):

PR #21631, opened 2024-08-27
feat: Optimize - log license information on startup and add license i…

matthewBearCamunda:
Backport of https://github.com/camunda/camunda/pull/21199

PR #24338, opened 2024-11-04
feat: Extend License for isCommercial and expiresAt - Optimize

### LINKED ISSUE #24339: C8 License - Optimize- Extend License Endpoint for isCommercial and expiresAt
### Context

As part of the [Licence expiration for non-commercial use](https://github.com/camunda/product-hub/issues/2529) project, two new fields need to be added to the `/ui-configuration` endpoint. These two fields will determine whether or not a license is for commercial or non-commerical use, and for determining the expiration date of a license

The two fields: `isCommercial` and `expiresAt` 

- `isCommercial`
  - This will be a boolean field
  - Will be `false` when a license contains a `non-commerical=true` property 
  - Will be `true` otherwise
- `expiresAt`
  - This will be a `string` field
  - It will be formatted as `date-time` as per the [OpenAPI docs](https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings)
  - Will be `null` in the case where the `expiryDate` field is not present on a license

This ticket is for the Optimize specific implementation of https://github.com/camunda/camunda/issues/24092

### Links

Parent epic: https://github.com/camunda/product-hub/issues/2550

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/24338
```

## Task Lifecycle

#### Engineering Review

#### QA Review

### Completion

matthewBearCamunda:
## Description

Add `isCommerical` and `expiresAt` to the `/ui-configuration` endpoint. These are two new fields from the license checker, and they need to be exposed so that they can be read by the front end

Also, be able to handle an unlimited license. These licenses functionally do not have an expiration date

## Checklist

- [ ] for CI changes:

## Related issues

closes #24339

Previous license work for Optimize was done with this PR: https://github.com/camunda/camunda/pull/21631

### DIFF
--- a/optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
+++ b/optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
@@ -84,7 +84,9 @@ public class CamundaLicense {
         isValid = false;
       } else {
         isCommercial = licenseKey.isCommercial();
-        validUntil = licenseKey.getValidUntil().toInstant().atOffset(ZoneOffset.UTC);
+        if (licenseKey.getValidUntil() != null) {
+          validUntil = licenseKey.getValidUntil().toInstant().atOffset(ZoneOffset.UTC);
+        }
         isValid = true;
       }
 
--- a/optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
+++ b/optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
@@ -29,7 +29,7 @@ public class CamundaLicense {
   private boolean isValid;
   private LicenseType licenseType;
   private boolean isCommercial;
-  private OffsetDateTime validUntil;
+  private OffsetDateTime expiresAt;
   private boolean isInitialized;
 
   public CamundaLicense(final String license) {
@@ -49,7 +49,7 @@ public class CamundaLicense {
   }
 
   public synchronized OffsetDateTime expiresAt() {
-    return validUntil;
+    return expiresAt;
   }
 
   public synchronized void initializeWithLicense(final String license) {
@@ -85,7 +85,7 @@ public class CamundaLicense {
       } else {
         isCommercial = licenseKey.isCommercial();
         if (licenseKey.getValidUntil() != null) {
-          validUntil = licenseKey.getValidUntil().toInstant().atOffset(ZoneOffset.UTC);
+          expiresAt = licenseKey.getValidUntil().toInstant().atOffset(ZoneOffset.UTC);
         }
         isValid = true;
       }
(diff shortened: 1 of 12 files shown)


optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
```diff
@@ -40,6 +44,14 @@ public synchronized LicenseType getLicenseType() {
     return licenseType;
   }
 
+  public synchronized boolean isCommercial() {
+    return isCommercial;
+  }
+
+  public synchronized OffsetDateTime expiresAt() {
+    return validUntil;
+  }
```
RomanJRW, line 53: is there a reason we mix "valid until" and "expires at" rather than just using one term?

optimize/backend/src/main/java/io/camunda/optimize/service/CamundaLicenseService.java
```diff
@@ -28,4 +29,12 @@ public boolean isCamundaLicenseValid() {
   public LicenseType getCamundaLicenseType() {
     return license.getLicenseType();
   }
+
+  public boolean isCommercialCamundaLicense() {
+    return license.isCommercial();
+  }
+
+  public OffsetDateTime getCamundaLicenseExpirationDate() {
```
RomanJRW, line 37: Maybe `getCamundaLicenseExpiresAt` or `getCamundaLicenseExpirationDateTime` would be better here? It's not a date, and the former suggesting even keeps it more generic than any implementation

optimize/backend/src/test/java/io/camunda/optimize/service/uiconfiguration/UIConfigurationServiceTest.java
```diff
@@ -170,6 +178,25 @@ public void testCamundaLicenseInformationIsReturned() {
     // then
     assertThat(configurationResponse.getLicenseType()).isEqualTo("saas");
     assertThat(configurationResponse.isValidLicense()).isEqualTo(true);
+    assertThat(configurationResponse.isCommercial()).isEqualTo(false);
+    assertThat(configurationResponse.getExpiresAt()).isEqualTo("2024-10-29T15:14:13Z");
```
RomanJRW, line 182: maybe a constant for this string would make sense

RomanJRW (commented):
Not sure I have enough context to be the approver here, especially as it touches classes I am not super familiar with. A few comments any from me

optimize/backend/src/main/java/io/camunda/optimize/license/CamundaLicense.java
matthewBearCamunda: `validUntil` is the existing term for expiration from the license checker, `expiresAt` is what the C3 Header expects and therefore what the API & backend uses

I'll update the variable to `expiresAt`

optimize/backend/src/main/java/io/camunda/optimize/service/CamundaLicenseService.java
matthewBearCamunda: I changed the monorepo implementation as well

RomanJRW (approved):
Pre-approving, but you have a legit checkstyle violation to fix

PR #24340, opened 2024-11-04
fix: remove docs links from c8 REST spec

pepopowitz:
## Description

https://github.com/camunda/camunda-docs/pull/4541 removes the need to include docs links for alpha notes in our C8 REST spec. The links will be injected when the docs are built, and we can keep this spec free of links that don't make sense to anyone but the docs. 

This PR removes those links.

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -497,7 +497,7 @@ paths:
         The Camunda 8 API (REST) Overview page provides further details.
 
         :::note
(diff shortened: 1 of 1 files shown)


tmetzke:
I think we should remove the whole `:::note` and automatically add it in the `camunda-docs`. Other OpenAPI spec viewers/editors cannot interpret that note, I guess?

pepopowitz:
> I think we should remove the whole `:::note` and automatically add it in the `camunda-docs`. Other OpenAPI spec viewers/editors cannot interpret that note, I guess?

Yeah, that makes sense! I had tunnel-vision and didn't even see those bits. 

I'll update in this PR.

pepopowitz:
all set for re-review, @tmetzke.

tmetzke (approved):

Issue #24094, opened 2024-10-26
Implement ES/OS archiver repository


npepinpe:
**Description**

Migrate the implementations of the `ArchiverRepository` from the Operate archiver. Again we need to de-Springify it, and more importantly, migrate from using the old `RestHighLevelClient` to use the new `ElasticsearchAsyncClient` (or OpenSearch equivalent). Likely this issue can be done in 2 PRs: first doing the ES implementation, and second the OS one.

> [!Note]
> It could be more interesting to implement the required methods in the search client directly as our top level abstraction layer, but for the sake of time, we'll keep the refactoring to a minimum and leave this as a follow up.

> [!Note]
> Looking through the repository code, it seems we have some sequences of operations we would like to be atomic, but cannot guarantee (i.e. lack of transactions). We will ignore this for now.

- The new connectors in the `search-clients` module should provide a factory method for the asynchronous ES/OS client.
- The `moveDocuments` and `getDestinationIndexName` should likely be moved from the old `Archiver` to the new repository
- Any metrics should be recreated in the new Camunda Exporter metrics class, and migrated to Micrometer.

PR #24341, opened 2024-11-04
Implement Elasticsearch archiver repository

npepinpe:
## Description

> [!Note]
> This is only tested manually at the moment. I will add test. This will make the PR bigger however, which is why I think we could do a first pass already, reduce the overall burden.

This PR adds an Elasticsearch implementation of the `ArchiverRepository`. 

Creation of the repository was moved from the resource provider to the `Archiver` itself. This is because the resource provider doesn't really know anything about the ES connection in itself. I thought about putting in the `ClientAdapter`, but since the repository needs to know about the `Archiver`'s executor, it broke all kinds of abstraction. So in the end, the `Archiver` will build it itself, and thus needs the complete `ExporterConfiguration` at this point. It's a little bit more coupling than I'd like, but acceptable.

You'll note in the `Archiver` that we create a custom `ProcessInstanceDependant` based on the `TaskTemplate`. This is because, as opposed to other templates, that one has two kinds of documents which are linked to process instances by different fields. I've asked [here](https://camunda.slack.com/archives/C06F0GLJNFM/p1730750256298019) if we can change that, but for now, it means we need this special adapter.

Other than that, I ported the pre-existing ES repository code to the new ES asynchronous client. This was surprisingly straightforward. While porting, I kept the same logic, but changed one thing in the asynchronous search: instead of keyed results for the date histogram, I went with an array. I mention it because there was an explicit comment in the old repository specifying we want a `keyed` result, but then we were just grabbing the first bucket (and specifying we only want _one_ bucket). It seemed more natural to use an array for this.

## Checklist

- [ ] for CI changes:

## Related issues

related to #24094

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
@@ -77,7 +77,7 @@ public final class ElasticsearchRepository implements ArchiverRepository {
     this.config = config;
     this.retention = retention;
     this.processInstanceIndex = processInstanceIndex;
-    this.batchOperationIndex = batchOperationTemplate;
+    this.batchOperationIndex = batchOperationIndex;
     this.client = client;
     this.executor = executor;
     this.metrics = metrics;
@@ -204,10 +204,7 @@ public final class ElasticsearchRepository implements ArchiverRepository {
         QueryBuilders.bool(q -> q.must(endDateQ, isProcessInstanceQ, partitionQ));
 
     return createSearchRequest(
-        processInstanceIndex.getFullQualifiedName(),
-        combinedQuery,
-        aggregation,
-        ListViewTemplate.END_DATE);
+        processInstanceIndex, combinedQuery, aggregation, ListViewTemplate.END_DATE);
   }
 
   private ArchiveBatch createArchiveBatch(final SubmitResponse<?> search) {
@@ -276,10 +273,7 @@ public final class ElasticsearchRepository implements ArchiverRepository {
                     .lte(JsonData.of(config.getArchivingTimePoint())));
 
     return createSearchRequest(
-        batchOperationIndex.getFullQualifiedName(),
-        endDateQ,
-        aggregation,
-        BatchOperationTemplate.END_DATE);
+        batchOperationIndex, endDateQ, aggregation, BatchOperationTemplate.END_DATE);
   }
 
   private SubmitRequest createSearchRequest(
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
@@ -185,7 +185,7 @@ public final class ElasticsearchRepository implements ArchiverRepository {
         AggregationBuilders.dateHistogram()
             .field(ListViewTemplate.END_DATE)
             .calendarInterval(rolloverInterval)
-            .format(config.getRolloverDateFormat())
+            .format(config.getElsRolloverDateFormat())
             .keyed(false) // get result as an array (not a map)
             .build();
     final var sortAggregation =
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
@@ -54,8 +54,8 @@ public final class ElasticsearchRepository implements ArchiverRepository {
   private final int partitionId;
   private final ArchiverConfiguration config;
   private final RetentionConfiguration retention;
-  private final ListViewTemplate processInstanceTemplate;
-  private final BatchOperationTemplate batchOperationTemplate;
+  private final String processInstanceIndex;
+  private final String batchOperationIndex;
   private final ElasticsearchAsyncClient client;
   private final Executor executor;
   private final CamundaExporterMetrics metrics;
@@ -67,8 +67,8 @@ public final class ElasticsearchRepository implements ArchiverRepository {
       final int partitionId,
       final ArchiverConfiguration config,
       final RetentionConfiguration retention,
-      final ListViewTemplate processInstanceTemplate,
-      final BatchOperationTemplate batchOperationTemplate,
+      final String processInstanceIndex,
+      final String batchOperationIndex,
       @WillCloseWhenClosed final ElasticsearchAsyncClient client,
       final Executor executor,
       final CamundaExporterMetrics metrics,
@@ -76,8 +76,8 @@ public final class ElasticsearchRepository implements ArchiverRepository {
     this.partitionId = partitionId;
     this.config = config;
     this.retention = retention;
-    this.processInstanceTemplate = processInstanceTemplate;
-    this.batchOperationTemplate = batchOperationTemplate;
+    this.processInstanceIndex = processInstanceIndex;
+    this.batchOperationIndex = batchOperationTemplate;
     this.client = client;
     this.executor = executor;
     this.metrics = metrics;
@@ -204,7 +204,7 @@ public final class ElasticsearchRepository implements ArchiverRepository {
         QueryBuilders.bool(q -> q.must(endDateQ, isProcessInstanceQ, partitionQ));
 
     return createSearchRequest(
-        processInstanceTemplate.getFullQualifiedName(),
+        processInstanceIndex.getFullQualifiedName(),
         combinedQuery,
         aggregation,
         ListViewTemplate.END_DATE);
@@ -276,7 +276,7 @@ public final class ElasticsearchRepository implements ArchiverRepository {
                     .lte(JsonData.of(config.getArchivingTimePoint())));
 
     return createSearchRequest(
-        batchOperationTemplate.getFullQualifiedName(),
+        batchOperationIndex.getFullQualifiedName(),
         endDateQ,
         aggregation,
         BatchOperationTemplate.END_DATE);
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
@@ -125,6 +125,7 @@ public final class ElasticsearchRepository implements ArchiverRepository {
             .settings(
                 settings ->
                     settings.lifecycle(lifecycle -> lifecycle.name(retention.getPolicyName())))
+            .index(destinationIndexName)
             .allowNoIndices(true)
             .ignoreUnavailable(true)
             .build();
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
@@ -18,6 +18,7 @@ import co.elastic.clients.elasticsearch._types.aggregations.Aggregation;
 import co.elastic.clients.elasticsearch._types.aggregations.AggregationBuilders;
 import co.elastic.clients.elasticsearch._types.aggregations.CalendarInterval;
 import co.elastic.clients.elasticsearch._types.aggregations.DateHistogramBucket;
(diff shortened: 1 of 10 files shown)


npepinpe:
@rodrigo-lourenco-lopes - as I mention at the top of the PR description, there's no tests yet. I still thought it's worth having a first pass, as a way to split what could be a bit large (mostly because the tests will add probably another 200-300 lines of code).

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
```diff
@@ -0,0 +1,278 @@
...
+    // TODO: verify if I close the transport that any pending futures are cancelled
+    client._transport().close();
+  }
+
+  private Aggregation createFinishedInstancesAggregation() {
+    final var dateAggregation =
+        AggregationBuilders.dateHistogram()
+            .field(ListViewTemplate.END_DATE)
+            .calendarInterval(rolloverInterval)
+            .format(config.getRolloverDateFormat())
```
rodrigo-lourenco-lopes, line 188: ❓  In the original implementation here we use `getElsRolloverDateFormat`

rodrigo-lourenco-lopes (approved):
The PR looks good 🚀 Although I have less confidence in my review of the ESrepository as I am not super familiar with implementations with ES. I based my review looking at the previous implementation.

There is only one comment that should be addressed, and we also need to implement here the batchOperations now that these PR was merged.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
npepinpe: Oh, good catch!

rodrigo-lourenco-lopes:
Not sure if in scope of this specific PR, but now that we implement the batch operations as well, should we allocate already 2 threads for this?
According to the [docs](https://docs.google.com/document/d/184JfguLkLOp03zArBq5IX236CO9yFZY4hCh3L8Hv0-U/edit?tab=t.0) here we should identify if we are running in the partition 1 and allocate two threads here. Should each job have exclusively one thread here? @npepinpe

npepinpe:
Yeah I think so. Can you open a follow issue to do this? We can detect if this is partition one, and only then we would create the batch operations job, and also set the number of background threads to 2.

npepinpe:
I added integration test for the repository. Nothing yet for the archiver itself, I would look into that after the OS repository is done, as we will write parameterized tests most likely.

npepinpe:
Whew, this turned out much bigger than expected, apologies :(

rodrigo-lourenco-lopes:
If there is not much new in the tests I think its fine if we don't split them, as I already took a look at most of it 🤔

npepinpe:
I'm referring to the OS one, so we can take a more iterative approach there.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
```diff
@@ -125,6 +125,7 @@ public CompletableFuture<Void> setIndexLifeCycle(final String destinationIndexNa
             .settings(
                 settings ->
                     settings.lifecycle(lifecycle -> lifecycle.name(retention.getPolicyName())))
+            .index(destinationIndexName)
```
rodrigo-lourenco-lopes, line 128: ❓ we apply the retention period each time that we archive the indexes, so it is not expected that if we change the configuration for the retention period on the exporter configuration to affect already existing archives correct?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
```diff
@@ -0,0 +1,255 @@
...
+import org.slf4j.LoggerFactory;
+import org.testcontainers.elasticsearch.ElasticsearchContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@SuppressWarnings("resource")
+@Testcontainers
+@AutoCloseResources
+final class ElasticsearchRepositoryIT {
+  private static final ObjectMapper MAPPER = new ObjectMapper();
```
rodrigo-lourenco-lopes, line 54: This mapper is not being used, probably loose form the previous iteration.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
```diff
@@ -0,0 +1,255 @@
...
+    final var sourceIndexName = UUID.randomUUID().toString();
+    final var destIndexName = UUID.randomUUID().toString();
+    final var repository = createRepository();
+    final var documents =
+        List.of(
+            new TestDocument("1", "2024-01-01"),
+            new TestDocument("2", "2024-01-04"),
+            new TestDocument("3", "2024-01-02"));
+    documents.forEach(doc -> index(sourceIndexName, doc));
+    testClient.indices().refresh(r -> r.index(sourceIndexName));
```
rodrigo-lourenco-lopes, line 177: ❓ I am missing some knowledge of ES, why do we need to refresh here?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
```diff
@@ -203,7 +204,75 @@ void shouldReindexDocuments() throws IOException {
...
+    // aggregation will fail
+    createBatchOperationIndex();
+    documents.forEach(doc -> index(batchOperationIndex, doc));
+    testClient.indices().refresh(r -> r.index(batchOperationIndex));
+    config.setRolloverBatchSize(3);
+
+    // when
+    final var result = repository.getBatchOperationsNextBatch();
+
+    // then - we expect only the first two documents created two days ago to be returned
```
rodrigo-lourenco-lopes, line 266: here is hours as well, and same for the test above :)

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
```diff
@@ -203,7 +204,75 @@ void shouldReindexDocuments() throws IOException {
...
+    // aggregation will fail
+    createProcessInstanceIndex();
+    documents.forEach(doc -> index(processInstanceIndex, doc));
+    testClient.indices().refresh(r -> r.index(processInstanceIndex));
+    config.setRolloverBatchSize(3);
+
+    // when
+    final var result = repository.getProcessInstancesNextBatch();
+
+    // then - we expect only the first two documents created two days ago to be returned
```
rodrigo-lourenco-lopes, line 235: I guess you meant to say we only expect one document

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
```diff
@@ -203,7 +204,75 @@ void shouldReindexDocuments() throws IOException {
...
+        DateTimeFormatter.ofPattern("yyyy-MM-dd").withZone(ZoneId.systemDefault());
+    assertThat(result).succeedsWithin(Duration.ofSeconds(30));
+    final var batch = result.join();
+    assertThat(batch.ids()).containsExactly("1");
+    assertThat(batch.finishDate()).isEqualTo(dateFormatter.format(now.minus(Duration.ofHours(2))));
+  }
+
+  @Test
+  void shouldGetBatchOperationsNextBatch() throws IOException {
+    // given - 3 documents, two of which were created over an hour ago, one of which was created
```
rodrigo-lourenco-lopes, line 246: here is also 2 hours

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
```diff
@@ -203,7 +204,75 @@ void shouldReindexDocuments() throws IOException {
         .containsExactlyInAnyOrder("1", "2", "3");
   }
 
-  private void index(final String index, final TestDocument document) {
+  @Test
+  void shouldGetProcessInstancesNextBatch() throws IOException {
+    // given - 5 documents, two of which were created over an hour ago, one of which was created,
```
rodrigo-lourenco-lopes, line 209: 4 documents right?
❓ Also I am not sure if I understood the test, are we not supposed to get the documents 1 and 2 here? since both are process instances and were created two hours ago? But I checked locally and its passing

rodrigo-lourenco-lopes (approved):
Thanks for adding the tests, much better coverage now 🚀 Just had some questions on things I did not understood.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/ElasticsearchRepository.java
npepinpe: Yes, it is, but there is a separate issue to tackle this :+1:

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
npepinpe: When you index documents in ES, the request returns when the server has accepted it (so persisted somewhere). However, because there's some caching happening in ES (I'm also not an expert, so I can't go into details confidently :sweat_smile:), it may take a bit for writes to become visible to clients. Refresh forces that to happen in a synchronous way, so we can guarantee the documents are visible afterwards.

Without it, the test would become flaky, basically.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
npepinpe: Oops! That was from the previous iteration :see_no_evil:

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
npepinpe: Yeah, I copied this from the batch operations test :see_no_evil:

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
npepinpe: Yeah, 4, initially I had planned 5 but realized it's not required.

So we have 4 docs:

```java
new TestProcessInstance(
  "1", twoHoursAgo, ListViewTemplate.PROCESS_INSTANCE_JOIN_RELATION, 1);
new TestProcessInstance(
  "2", twoHoursAgo, ListViewTemplate.PROCESS_INSTANCE_JOIN_RELATION, 2);
new TestProcessInstance("3", twoHoursAgo, ListViewTemplate.ACTIVITIES_JOIN_RELATION, 1);
new TestProcessInstance(
  "4", now.toString(), ListViewTemplate.PROCESS_INSTANCE_JOIN_RELATION, 1);
```

We're archiving _only for partition 1_. So we only have "1", "3", and "4" for partition 1. Then we only select those from the `PROCESS_INSTANCE_JOIN_RELATION` - so that's only "1", and "4". Finally, "4" was created right now, so we only expect "1" which was created over 1h ago (our archiving time point).

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
rodrigo-lourenco-lopes: oh I see, thanks for the explanation 🙌

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/ElasticsearchRepositoryIT.java
rodrigo-lourenco-lopes: ah right the partitions here make difference! Thanks 🙌

PR #24342, opened 2024-11-04
Add group update processor

### LINKED ISSUE #23546: Process `Group.UPDATE` command
<h2>Description</h2>

image

1. Create new class `GroupUpdateProcessor` and implement `DistributedTypedRecordProcessor<GroupRecord>`
2. Register the processor in the `GroupProcessors`
    - Check if the group already exists. If no, reject the command with `NOT_FOUND`. If yes continue.
    - Check if the user is authorized to update groups:
        - Resource type: `GROUP`
        - Permission type: `UPDATE`
        - ResourceId: group name
    - Write a `Group.UPDATED` event and respond to the client
    - Write a `Group.UPDATED` event
5. Unit test the Group.UPDATED command
6. Unit test the Group.UPDATED command distribution

koevskinikola:
## Description

Add a processor for group update commands
and pass it to the appropriate engine
classes

## Checklist

- [ ] for CI changes:

## Related issues

closes #23546 x

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/GroupUpdateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/GroupUpdateProcessor.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.identity;
+
+import static io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.UNAUTHORIZED_ERROR_MESSAGE;
+
+import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
+import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.AuthorizationRequest;
+import io.camunda.zeebe.engine.processing.streamprocessor.DistributedTypedRecordProcessor;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedRejectionWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedResponseWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
+import io.camunda.zeebe.engine.state.distribution.DistributionQueue;
+import io.camunda.zeebe.engine.state.immutable.GroupState;
+import io.camunda.zeebe.protocol.impl.record.value.group.GroupRecord;
+import io.camunda.zeebe.protocol.record.RejectionType;
+import io.camunda.zeebe.protocol.record.intent.GroupIntent;
+import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
+import io.camunda.zeebe.protocol.record.value.PermissionType;
+import io.camunda.zeebe.stream.api.records.TypedRecord;
+import io.camunda.zeebe.stream.api.state.KeyGenerator;
+
+public class GroupUpdateProcessor implements DistributedTypedRecordProcessor<GroupRecord> {
+
+  private final GroupState groupState;
+  private final KeyGenerator keyGenerator;
+  private final AuthorizationCheckBehavior authCheckBehavior;
+  private final StateWriter stateWriter;
+  private final TypedRejectionWriter rejectionWriter;
+  private final TypedResponseWriter responseWriter;
+  private final CommandDistributionBehavior commandDistributionBehavior;
+
+  public GroupUpdateProcessor(
+      final GroupState groupState,
+      final KeyGenerator keyGenerator,
+      final AuthorizationCheckBehavior authCheckBehavior,
+      final Writers writers,
+      final CommandDistributionBehavior commandDistributionBehavior) {
+    this.groupState = groupState;
+    this.keyGenerator = keyGenerator;
+    this.authCheckBehavior = authCheckBehavior;
+    this.commandDistributionBehavior = commandDistributionBehavior;
+    stateWriter = writers.state();
+    responseWriter = writers.response();
+    rejectionWriter = writers.rejection();
+  }
+
+  @Override
+  public void processNewCommand(final TypedRecord<GroupRecord> command) {
+    final var record = command.getValue();
+    final var persistedRecord = groupState.get(record.getGroupKey());
+    if (persistedRecord.isEmpty()) {
+      final var errorMessage =
+          "Expected to update group with key '%s', but a group with this key does not exist."
+              .formatted(record.getGroupKey());
+      rejectionWriter.appendRejection(command, RejectionType.NOT_FOUND, errorMessage);
+      responseWriter.writeRejectionOnCommand(command, RejectionType.NOT_FOUND, errorMessage);
+      return;
+    }
+
+    final var groupName = record.getName();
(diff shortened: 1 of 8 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/group/CreateGroupMultiPartitionTest.java
```diff
@@ -100,16 +99,9 @@ public void shouldDistributeInIdentityQueue() {
   public void distributionShouldNotOvertakeOtherCommandsInSameQueue() {
     // given the user creation distribution is intercepted
     for (int partitionId = 2; partitionId <= PARTITION_COUNT; partitionId++) {
-      interceptUserCreateForPartition(partitionId);
+      interceptGroupCommandForPartition(partitionId, GroupIntent.CREATE);
```
nicpuppa, line 102: ❌ You're not testing if commands don't overtake each other here. You're only sending one command, so all that's being tested is the redistribution of that command. To test that stuff doesn't overtake each other you'll have to send another distributed command on the same queue, with a different value type or intent. Then you can guarantee it's ordered.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/group/GroupTest.java
```diff
@@ -49,4 +49,38 @@ public void shouldNotDuplicate() {
...
+        engine.group().updateGroup(groupKey).withName(updatedName).update();
+
+    final var updatedGroup = updatedGroupRecord.getValue();
+    Assertions.assertThat(updatedGroup)
+        .isNotNull()
+        .hasFieldOrPropertyWithValue("name", updatedName);
+  }
+
+  @Test
+  public void shouldRejectUpdatedIfNoGroupExists() {
```
nicpuppa, line 72: ❌ I'm missing the test where the processor reject the command because you are trying to update a group with the same name

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/GroupUpdateProcessor.java
```diff
@@ -0,0 +1,106 @@
...
+    if (!authCheckBehavior.isAuthorized(authorizationRequest)) {
+      final var errorMessage =
+          UNAUTHORIZED_ERROR_MESSAGE.formatted(
+              authorizationRequest.getPermissionType(), authorizationRequest.getResourceType());
+      rejectionWriter.appendRejection(command, RejectionType.UNAUTHORIZED, errorMessage);
+      responseWriter.writeRejectionOnCommand(command, RejectionType.UNAUTHORIZED, errorMessage);
+      return;
+    }
+
+    final var groupKey = groupState.getGroupKeyByName(groupName);
```
nicpuppa, line 80: ❓ Is this step necessary ? Is it not enough to say that if the provided group name is the same of the persisted record the processor should reject the command ?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/group/UpdateGroupMultiPartitionTest.java
```diff
@@ -0,0 +1,140 @@
...
+  public void shouldDistributeInIdentityQueue() {
+    // when
+    final var name = UUID.randomUUID().toString();
+    final var groupKey = engine.group().newGroup(name).create().getKey();
+    engine.group().updateGroup(groupKey).withName(name + "-updated").update();
+
+    // then
+    assertThat(
+            RecordingExporter.commandDistributionRecords()
+                .limitByCount(r -> r.getIntent().equals(CommandDistributionIntent.FINISHED), 1)
```
nicpuppa, line 98: ❌ The limit should be 2

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/group/UpdateGroupMultiPartitionTest.java
```diff
@@ -0,0 +1,140 @@
...
+                .withIntent(CommandDistributionIntent.ENQUEUED))
+        .extracting(r -> r.getValue().getQueueId())
+        .containsOnly(DistributionQueue.IDENTITY.getQueueId());
+  }
+
+  @Test
+  public void distributionShouldNotOvertakeOtherCommandsInSameQueue() {
+    // given the user creation distribution is intercepted
```
nicpuppa, line 106: 🔧 This is not related to user creation

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/group/UpdateGroupMultiPartitionTest.java
```diff
@@ -0,0 +1,140 @@
...
+    for (int partitionId = 2; partitionId <= PARTITION_COUNT; partitionId++) {
+      interceptGroupCommandForPartition(partitionId, GroupIntent.UPDATE);
```
nicpuppa, line 108: ❌ You should intercept the Group creation so you can verify that the distribution do not overtake other commands in the queue

nicpuppa (changes requested):
Hey @koevskinikola I left some comments, please have a look

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/GroupUpdateProcessor.java
koevskinikola: That's a very good point, thanks!

I'll update this processor and the`RoleUpdateProcessor` to avoid this extra state access.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/group/CreateGroupMultiPartitionTest.java
koevskinikola: @nicpuppa I'm creating another group on line 108, so two `GroupIntent.CREATE` commands are sent.

Or did I misunderstand something?

koevskinikola:
@nicpuppa I implemented your feedback except [this one](https://github.com/camunda/camunda/pull/24342#discussion_r1829003673), let me know how it looks now.

nicpuppa (approved):
Thanks @koevskinikola

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/GroupUpdateProcessor.java
remcowesterhoud: @nicpuppa @koevskinikola This was necessary. If the group name changed we must check the state that there is not already another group with this name. Checking it with the provided record is not the same thing.

nicpuppa: What do you mean ? We check the provided with the record in the state

remcowesterhoud: > What do you mean ? We check the provided with the record in the state

We get the Group from the state using the key provided in the command. Then we check if the name in our state is the same as the name of the provided command.

What the check should do is make sure there are no other Groups (with different keys) that have the same name. The name should be unique.

remcowesterhoud: So basically:

```java
if (groupState.getByName(name).isPresent()) {
   // reject because a group with this name already exists.
}
```

nicpuppa: Ok got it, make sense 👍

remcowesterhoud: @nicpuppa Will you create a fix for it?

nicpuppa: yes

PR #24353, opened 2024-11-05
fix: commit last transaction after running migrations

### LINKED ISSUE #24352: Version update is not persisted in state because transaction is not committed
After running migrations, we persist the new broker version to the state so that it can be checked for compatibility on the next update.

In one release benchmark, I've noticed repeated logs that claim a patch upgrade occurs, long after that update was initially done: `PatchUpgrade[from=8.5.7, to=8.5.8]` 
It seems as if the version update was not persisted in the state.

**Hypothesis:** A migration task opened an iterator, which starts a transaction. Then the `upsert` of the version finds that there is already an open transaction and just appends an operation to it. The transaction is never committed afterwards, such that the update is lost.

**Proposed fix:** Commit the transaction at the end of the `DbMigrationStep`.

**Severity:** Quite high because a lost update might prevent further updates with no straigh-forward way to recover. Additionally, modifications done by migration tasks might also be lost in the same way.

lenaschoenburg:
Running migrations might end with an open transaction that contains changes that are not committed yet. To avoid losing these changes, the `MigrationTransitionStep` will do one last commit.

closes https://github.com/camunda/camunda/issues/24352


npepinpe (approved):
I'm surprised we never noticed :cold_sweat:

PR #24356, opened 2024-11-05
Check authorization on Job activation

### LINKED ISSUE #22890: Verify permissions in `JobBatchActivateProcessor`
<h2>Description</h2>

Authorization for Job activation will work similar to multi-tenancy. We will only activate jobs that the user is permitted for. Any other Jobs we ignore. This is similar to the way it works in C7. This decision was made [here](https://camunda.slack.com/archives/C06UYJMMETZ/p1730203727608179).

- **Permission type**: `UPDATE`
- **Possible permissions**:
    -  `*`
    - `bpmnProcessId`

remcowesterhoud:
## Description

Users must be authorized to activate Jobs. This authorization is done at a bpmnProcessId level. Since job activation happens in batches, there could be multiple jobs with different process ids. Users are not necessarily authorized for
all of these jobs. The decision was made to ignore jobs that the user lacks authorizations for. We do this by getting a set of process ids that the user is authorized for and comparing these with the process id of the job itself. If the user is not authorized we will skip this job and continue with the next one.
There is one exception to the rule. If the user has a wildcard permission (*) it is authorized to activate all jobs.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22890

### DIFF
--- a/service/src/main/java/io/camunda/service/JobServices.java
+++ b/service/src/main/java/io/camunda/service/JobServices.java
@@ -53,6 +53,7 @@ public final class JobServices<T> extends ApiServices<JobServices<T>> {
             .setTimeout(request.timeout())
             .setWorker(request.worker())
             .setVariables(request.fetchVariable());
+    brokerRequest.setAuthorization(authentication.token());
     activateJobsHandler.activateJobs(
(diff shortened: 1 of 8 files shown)


nicpuppa (approved):
👏

PR #24362, opened 2024-11-05
Centralize creation of default search DB containers

npepinpe:
## Description

We've copied the same piece of code to create a default ES/OS containers in many places (mostly ES). This was all based on the `TestSupport` file originally added to the `zeebe-elasticsearch-exporter` module.

I've extracted it to a share utility in `zeebe-test-util`, and replaced its usages (as much as I could find).

The container is still modifiable when required, but the new support class provides a nice base, as in most cases we never deviate from the default.

## Related issues

related to #24094


npepinpe:
Seems like the last test failures are flakes.

rodrigo-lourenco-lopes (approved):
Great refactoring, nice to see so many removed lines 🙌

PR #24378, opened 2024-11-05
feature: when a follower fails to flush respond with error to leader

### LINKED ISSUE #14867: Retry complete follower append on flush error

entangled90:
## Description 
When a follower fails to flush, reply to the leader with a `succeeded=false` reply, so the leader will reset the follower. Once the follower is able to flush again, its logs will converge with the leader's log.
The new test would not pass without the fix, because the exception thrown when flushing would mark the follower "INACTIVE" and it would be unable to resume processing.

A custom RaftElectionConfig is used to avoid nodes configured as faulty to become leader.

## Related issues 
closes #14867

### DIFF
--- a/zeebe/atomix/cluster/src/test/java/io/atomix/raft/RaftFlushErrorTest.java
+++ b/zeebe/atomix/cluster/src/test/java/io/atomix/raft/RaftFlushErrorTest.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.atomix.raft;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.atomix.cluster.MemberId;
+import io.atomix.raft.RaftRule.Configurator;
+import io.atomix.raft.RaftServer.Builder;
+import io.atomix.raft.RaftServer.Role;
+import io.atomix.raft.partition.RaftElectionConfig;
+import io.atomix.raft.storage.RaftStorage;
+import io.atomix.raft.storage.log.RaftLogFlusher;
+import io.camunda.zeebe.journal.Journal;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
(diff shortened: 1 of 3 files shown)


zeebe/atomix/cluster/src/test/java/io/atomix/raft/RaftFlushErrorTest.java
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RunWith(Parameterized.class)
```
npepinpe, line 37: :question: It's unclear to me why we are parameterizing this test. What are we covering that wouldn't covered by running it for a single run?

zeebe/atomix/cluster/src/test/java/io/atomix/raft/RaftFlushErrorTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+  private static RaftLogFlusher.Factory faultyFlusher(
+      final Supplier<Boolean> faultyWhen, final Runnable notifyFaultyFlush) {
+    return (ignored) ->
+        new RaftLogFlusher() {
+          @Override
+          public void flush(final Journal journal) {
+            if (faultyWhen.get()) {
+              notifyFaultyFlush.run();
+              throw new RuntimeException(new IOException("Failed sync"));
```
npepinpe, line 126: :x: Shouldn't we flush when not faulty?

npepinpe (approved):
:tada: Thanks! Looks pretty good, but please address all comments before merging.

- :question: Just to clarify, if the follower fails to flush, it already did append the entries. So when the leader retries, it would deterministically retry and the follower would be able to determine that it can "re-append" (I'm not sure if we actually write it again, but I think we check), this is safe since the checksums match for each index, right? So it would succeed on a second run.
  - If the entries had changed between runs, then the log would be correctly truncated (as they would not be committed), so it should also be fine. Can you verify this?

zeebe/atomix/cluster/src/test/java/io/atomix/raft/RaftFlushErrorTest.java
entangled90: I took a similar approach as the other Raft tests, just to test with clusters of different sizes.  I can also fix the number of members to a fixed value.

entangled90: I checked the code coverage and there are no difference between a parametrized test and 3 node configuration. I will remove the parametrized

PR #24384, opened 2024-11-05
Migrate escalation boundary events

### LINKED ISSUE #24374: Migrate escalation boundary events
## Description
It should be possible to migrate tasks with escalation boundary events. Escalation catch event doesn't have a subscription record but its **static** escalation code stored in the catch event element itself. When an escalation is thrown, the escalation catch event is accessed through the process instance itself.

Given above, there is no need (or no way) to introduce a migration intent for escalation catch event. Consider the following example, where an active service task has an escalation boundary event attached. When migrating the process instance the escalation boundary event will be migrated to the target process **regardless of mapping instruction is provided for the escalation boundary element or not**. After the instance is migrated, **the escalation code will be updated** as in the target process since it is static value.

Screenshot 2024-11-05 at 14 14 37

Please note that subprocesses still need to be mapped as they are still active elements of the process. 
Two things needs to be implemented:
- Add `BpmnEventType.ESCALATION` to supported boundary element types
- Verify a task can be migrated with a escalation boundary event
- Verify escalation boundary event is created in the target when it doesn't exist in target source
- Verify escalation boundary event is deleted from source when it doesn't exist in the target

berkaycanbc:
## Description

See #24374 for detailed description.

## Related issues

closes #24374


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateEscalationEventTest.java
```diff
@@ -0,0 +1,540 @@
...
+import org.junit.Test;
+import org.junit.rules.TestWatcher;
+
+public class MigrateEscalationEventTest {
+
+  private static final String ESCALATION_CODE_NUMBER = "404";
+  private static final String ESCALATION_CODE_NUMBER_2 = "405";
+  private static final String ESCALATION_CODE_NUMBER_3 = "406";
+  private static final String ESCALATION_CODE_NUMBER_4 = "407";
+  @Rule public final EngineRule engine = EngineRule.singlePartition();
```
korthout, line 30: 🔧 We can reuse the engine between the test cases to speed up these tests.

```suggestion
  @ClassRule public static final EngineRule ENGINE = EngineRule.singlePartition();
```

korthout (approved):
Thanks @berkaycanbc 🚀 

👍 LGTM 

🔧 I have one suggestion to speed up the test cases

🔧 Please consider adding this to https://github.com/camunda/camunda/issues/22180 as it suffers from the same problem

PR #24386, opened 2024-11-05
Migrate escalation event subprocesses

### LINKED ISSUE #24385: Migrate escalation event subprocesses
## Description

(Since escalation event subprocess migration is expected to work the same as error event subprocess migration, I didn't change error start events to escalation ones in screenshots for saving time on documenting the issue.)

### Migrate active escalation event subprocess

Screenshot 2024-09-10 at 16 04 08

Mappings should be provided between event subprocesses (sub1 -> sub2) and active elements inside event subprocesses (B -> B). 

### Migrate escalation event subprocess start event subscription

Screenshot 2024-09-10 at 16 02 05

To migrate the start event subscription of a escalation event subprocess a mapping needs to be provided for active elements (A-> A) and between escalation start events (start1 -> start2). This case currently doesn't provide much advantage since escalation code will not change during the migration, but will be useful when we introduce the feature for re-evaluating escalation codes.

### Subscribe to escalation event subprocess start event subscription

Screenshot 2024-09-10 at 16 00 04

In this case, a mapping should be provided for the active element (A -> A). The escalation event subscription will be created after the migration for escalation start event (start2).

### Unsubscribe from escalation event subprocess start event subscription

Screenshot 2024-09-10 at 15 58 13

In this case, a mapping should be provided for the active element (A -> A). The escalation event subscription will be deleted after the migration for escalation start event (start1).

berkaycanbc:
## Description

See #24385 for detailed description.

## Related issues

closes #24385

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -228,13 +228,21 @@ public class ProcessInstanceMigrationMigrateProcessor
         sourceProcessDefinition,
         elementInstanceRecord,
         EnumSet.of(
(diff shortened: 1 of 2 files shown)


korthout (approved):
Thanks @berkaycanbc 💯 

LGTM 👍

PR #24388, opened 2024-11-05
Batchoperation jobs only in partition 1

### LINKED ISSUE #24383: BatchOperations is only created in parition 1
BatchOperations are not run per partition and should only be done in partition 1. This means that in the `Archiver` we only should `createBatchOperationJob` if we are running on partition 1.  See [here](https://docs.google.com/document/d/184JfguLkLOp03zArBq5IX236CO9yFZY4hCh3L8Hv0-U/edit?tab=t.0).

Additionally, when calling `defaultExecutor` for partition 1 we should allocate 2 threads to the executor preform both both the `ProcessInstanceJob` and the `BatchOperationJob`. We already pass the partition id to `Archiver.create()` so the changes should not be to big.

rodrigo-lourenco-lopes:
## Description

This PR makes it so that we only run the `BatchOperations` in partition 1. Additionally, we allocate 2 background threads if in partition 1 so that we can have one available for both `ProcessInstanceJob` and `BatchOperationJob`.

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/24383

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/Archiver.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/Archiver.java
@@ -7,6 +7,8 @@
  */
 package io.camunda.exporter.archiver;
 
+import static io.camunda.zeebe.protocol.Protocol.START_PARTITION_ID;
+
 import io.camunda.exporter.ExporterResourceProvider;
 import io.camunda.exporter.archiver.ArchiverRepository.NoopArchiverRepository;
 import io.camunda.exporter.config.ConnectionTypes;
@@ -87,16 +89,20 @@ public final class Archiver implements CloseableSilently {
             .name("exporter-" + exporterId + "-p" + partitionId + "-bg-", 0)
             .uncaughtExceptionHandler(FatalErrorHandler.uncaughtExceptionHandler(logger))
             .factory();
-    final var executor = defaultExecutor(threadFactory);
+    final var executor = defaultExecutor(threadFactory, partitionId);
     final var repository =
         createRepository(config, resourceProvider, partitionId, executor, metrics, logger);
     final var archiver = new Archiver(partitionId, repository, logger, executor);
     final var processInstanceJob =
         createProcessInstanceJob(metrics, logger, resourceProvider, repository, executor);
-    final var batchOperationJob =
-        createBatchOperationJob(metrics, logger, resourceProvider, repository, executor);
+    if (partitionId == START_PARTITION_ID) {
+      final var batchOperationJob =
+          createBatchOperationJob(metrics, logger, resourceProvider, repository, executor);
+      archiver.start(config.getArchiver(), processInstanceJob, batchOperationJob);
+    } else {
+      archiver.start(config.getArchiver(), processInstanceJob);
+    }
 
-    archiver.start(config.getArchiver(), processInstanceJob, batchOperationJob);
     return archiver;
   }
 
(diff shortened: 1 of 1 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/Archiver.java
```diff
@@ -25,6 +32,7 @@
 import org.slf4j.Logger;
 
 public final class Archiver implements CloseableSilently {
+  private static final int BATCH_OPERATION_PARTITION_ID = 1;
```
npepinpe, line 35: :x: We should use `Protocol.START_PARTITION_ID` to be safe. I don't think we will change in the future, but better safe than sorry, and it's a very small change.

npepinpe (commented):
:wrench: I didn't see any test for this. I'm not sure if it's easy to do so (e.g. a unit test in `ArchiverTest`), but ideally we should try. It's possible we would need to change the code a bit to make it easier to test :thinking: wdyt?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/Archiver.java
```diff
@@ -142,9 +148,11 @@ private static BatchOperationArchiverJob createBatchOperationJob(
...
   }
 
-  private static ScheduledThreadPoolExecutor defaultExecutor(final ThreadFactory threadFactory) {
-    // TODO: set size to 2 in case we need to do batch operations
-    final var executor = new ScheduledThreadPoolExecutor(1, threadFactory);
+  private static ScheduledThreadPoolExecutor defaultExecutor(
+      final ThreadFactory threadFactory, final int partitionId) {
+    // if we are present on partition 1, we need to have 2 threads to handle both jobs
+    final var executor =
+        new ScheduledThreadPoolExecutor(partitionId == START_PARTITION_ID ? 2 : 1, threadFactory);
```
npepinpe, line 155: :wrench: We now have the check for `partitionId == START_PARTITION_ID` duplicated. What if we ever change the logic? How about:

- We create the jobs in a list/array. 
- The batch operation job is added only if the `partitionId == START_PARTITION_ID`.
- We pass the number of jobs to the default executor method, and we create `Math.max(jobs.size(), BG_THREAD_LIMIT)` (where `BG_THREAD_LIMIT` is some sane value - we can add a TODO about making it configurable in the future).

That way the logic is more or less in one place, and when we add more jobs, we will automatically "parallelize" them.

npepinpe (approved):
Regarding the testing comment, that's fair. Please address my last comment before merging :+1:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/Archiver.java
rodrigo-lourenco-lopes: Does not seem super easy to do since we have to instantiate the executor before we create the jobs which also needs to know the number of jobs 🤔 

Later, we are going to add the post archiver, which will run in all partitions, so it will not add to logic complexity from here. However, for additional jobs, I am not sure if a different logic is used. 

Do you think it is ok to merge as is now, and after the post archiver jobs decide on the limits of the number of threads and if we should use different checks here?

npepinpe: Let's do that.

Issue #23180, opened 2024-10-08
Implement Authorization checks for GET Queries


houssain-barouni:
### Description

Implement authorization checks for GET queries.

### Requirements:

1. Authorization Flow:
* Fetch the entity from the data source first:
  *  authorization checks should not be applied on the search query that gets the entity.
  * tenantId filter (deduced from user's authenticated tenantIds) should be applied to guarantee data isolation.
* Perform the authorization check after the entity is fetched to determine whether the user is allowed to access it.
  * If the entity is found but the user is not authorized to access it, return an unauthorized response.
  * This distinction is essential to avoid exposing unauthorized entities while correctly handling cases where the entity doesn’t exist.
  
2. Permission Evaluation:
* Similar to search queries, use the user’s context to check the relevant permissions, such as, userKey, groupKeys and roleKeys
* Permission checks should be performed based on the resource type (e.g., Process Definition, Decision Definition) and required permission type (e.g., READ).

|status| Endpoint           | Permission | checked attribute   |
|-|----------------|---------------|---------------|
|https://github.com/camunda/camunda/pull/24413| /process-definitions/{processDefinitionKey}  | PROCESS_DEFINITION.READ  | processDefinition.processDefinitionId |
|https://github.com/camunda/camunda/pull/24413| /process-definitions/{key}/xml  | PROCESS_DEFINITION.READ  | processDefinition.processDefinitionId |
|https://github.com/camunda/camunda/pull/24413| /process-definitions/{processDefinitionKey}/form  | PROCESS_DEFINITION.READ  | processDefinition.processDefinitionId |
|https://github.com/camunda/camunda/pull/24567| /process-instances/{processInstanceKey}  | PROCESS_DEFINITION.READ_INSTANCE  | processInstance.processDefinitionId |
|https://github.com/camunda/camunda/pull/24567| /flownode-instances/{flowNodeInstanceKey}  | PROCESS_DEFINITION.READ_INSTANCE  | flowNodeInstance.processDefinitionId |
|https://github.com/camunda/camunda/pull/24567| /variables/{variableKey}  | PROCESS_DEFINITION.READ_INSTANCE  | variable.processDefinitionId |
|https://github.com/camunda/camunda/pull/24567| /incidents/{incidentKey}  | PROCESS_DEFINITION.READ_INSTANCE  | incident.processDefinitionId |
[...]

PR #24400, opened 2024-11-05
refactor: create security-services modules with AuthorizationChecker

houssain-barouni:
## Description

`camunda-security-services` module providing [`AuthorizationChecker`](https://github.com/camunda/camunda/pull/24400/files#diff-02f314d2c0da150bb8d8ab36731d7209368e274e002a10b7460ce4788e99d435) service with minimal dependencies:
- `camunda-security-core`
- `camunda-search-domain`
- `camunda-search-client`

This allows `AuthorizationChecker` to enable shared and reusable permission checks across modules that need to validate authorizations against data in ES/OS (and later RDBMS). 
In this pull request, `AuthorizationChecker` is used in `camunda-service` to apply the authorization checks on GET endpoints, and within `camunda-search-client-query-transfomers` to append authorization filters to search queries. 
It can be also used in `operate-webapp` to apply the permission checks for v1 endpoints (it can be autowired as spring is still used in Operate)

## Checklist

- [ ] for CI changes:

## Related issues

relates to https://github.com/camunda/camunda/issues/23180

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -324,6 +324,11 @@
 
     <dependency>
       <groupId>io.camunda</groupId>
+      <artifactId>camunda-security-services</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.camunda</groupId>
       <artifactId>document-store</artifactId>
     </dependency>
 
--- a/dist/src/main/java/io/camunda/application/commons/service/ServiceSecurityConfiguration.java
+++ b/dist/src/main/java/io/camunda/application/commons/service/ServiceSecurityConfiguration.java
@@ -8,7 +8,9 @@
 package io.camunda.application.commons.service;
 
 import io.camunda.application.commons.service.ServiceSecurityConfiguration.ServiceSecurityProperties;
+import io.camunda.search.clients.AuthorizationSearchClient;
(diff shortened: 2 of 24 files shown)


tmetzke (approved):
👍 This is really helpful for better integration of permission checks 🚀 
⚠️ I am a bit concerned about the dependencies between the search and security modules, to be honest. We don't have any cyclic dependencies but some submodules of `search` depend on some of `security` and vice-versa. This might become quite hard to understand and maintain pretty quickly.

houssain-barouni:
> ⚠️ I am a bit concerned about the dependencies between the search and security modules, to be honest. We don't have any cyclic dependencies but some submodules of `search` depend on some of `security` and vice-versa. This might become quite hard to understand and maintain pretty quickly.

@tmetzke You’re absolutely right. Previously, to avoid this issue, we had to rely on a low-level search client to retrieve authorization data needed within the search module. 
This led to duplication of authorization logic across multiple places. I think this approach provides a good balance. The security module depends only on `AuthorizationSearchClient`, stripped of security aspects with `this.authorizationSearchClient = authorizationSearchClient.withSecurityContext(SecurityContext.withoutAuthentication());`, which effectively avoids the circular dependency implicitly.

PR #24401, opened 2024-11-05
feat: support user delete events

Ben-Sheppard:
## Description

This PR achieves two tasks:
1. The existing user handler is refined to handle creates and updates
2. A new handler is introduced that processes user delete commands

Some notes:
* I think in the future the `CamundaExporterIT` will have issues with these new handlers as the current tests revert to a default record generator which selects an intent at random related to the value type. When we have handlers that only support certain intents then the records may not be picked up due to a mismatch in intent values and therefore the tests will fail (because no record has been exported).
* I tried to add in an IT that also considers intents (which worked fine for user create) however failed when it came to the user delete side because there is some additional setup required there. Its likely that in those circumstances we want the tests to be specific and the handlers not included in the current generic tests.


EuroLew:
Hey I made the `CamundaExporterIT` handler tests less generic and more handler specific so you should have no problem writing tests with more specific requirements, #24405.

Ben-Sheppard:
> Hey I made the `CamundaExporterIT` handler tests less generic and more handler specific so you should have no problem writing tests with more specific requirements, #24405.

Ah thank you @EuroLew - in that case I will likely follow up with another PR for the IT (or rebase this one and add it depending on whats merged first :D )

Oleksiivanov (approved):

PR #24405, opened 2024-11-06
Flaky test fix for CamundaExporterIT

EuroLew:
## Description

Fixes flaky tests for handlers, and added tests for operation handlers, a more generic approach was abandoned as per #24401 there are simply too many unique handlers requiring scenarios other than 
1. record
2. create entity
3. flush entity 

## Related issues

closes #


mustafadagher (commented):
Thanks Euro! 
I don't see a problem in this specific PR, But I have some thoughts about the overall Integration Test Approach:

💭 While here we are trying to test individual handlers, in my view the tests should test that the `Exporter` can export certain `RecordValues` and produce a certain `State/Document` in the end.

Each record type of specific intent or state could be handled by a few handlers and result in a specific documents. We should setup the tests this way..

for example, a process instance record, should be handled by those three handlers and produce these documents.

I don't care about the handlers then, I care about the end result. i.e. If I handle this record, I expect this end state.

We might end up with many tests similar in convention to something like `ShouldHandleTerminatedProcessInstanceRecord()` and that tests updating an `operation entity, persisting an event entity, upserting a ListViewFlowNode and FlowNodeInstance,.. etc.

So again, a record type oriented approach, rather than a handler oriented approach, and verifying the end result of exporting.

Let me know if my idea is unclear, and we shall have a call to talk about it if you wish.

mustafadagher (approved):
I just had a call with @EuroLew, where we discussed my [previous comment](https://github.com/camunda/camunda/pull/24405#pullrequestreview-2417839329).

Basically, my thoughts were that testing handlers in isolation (from one another) is not really a true `Integration Test` for exporting. It doesn't also ensure that every handler is configured properly in the resource provider.
Testing `Record ValueType/Intent` is more testing the full lifecycle or a real-life exporter setup.

We did not come to an agreement on view points. And since my comment was only a recommendation and an invitation for discussion. I see no problem in merging the changes here.

Zelldon:
> I don't care about the handlers then, I care about the end result. i.e. If I handle this record, I expect this end state.

I agree with this.

Zelldon:
But to be fair i think the previous requirement described were more based on handlers, especially because it was or will be hard to determine the expected result. (At least this was my assumption)

But i agree, actually it is an implementation detail that should be hidden. Meaning we should concentrate on the behavior.

PR #24408, opened 2024-11-06
feat: support updating roles via RoleServices

### LINKED ISSUE #24402: Camunda services support updating a role
### Description
Role updates should be supported in the `camunda-service` module. An example of the implementation of updating via the service module can be found [here](https://github.com/camunda/camunda/blob/main/service/src/main/java/io/camunda/service/UserServices.java#L62-L70)

### Acceptance Criteria
- [ ] A `BrokerRoleUpdateRequest` exists hint: [BrokerUserUpdateRequest](https://github.com/camunda/camunda/blob/main/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserUpdateRequest.java)
  - [ ] The field(s) that can be updated for a role are [`name`]
- [ ] A method exists with signature `updateRole(final RoleDTO roleDTO)`
- [ ] The `RoleServiceTest` class should contain unit tests for the `updateRole` method - hint: an example of working with the broker commands in tests can be found [here](https://github.com/camunda/camunda/blob/main/service/src/test/java/io/camunda/service/UserServiceTest.java#L66-L81)

emulbreh:
## Description

The PR adds support for updating roles via `RoleServices::updateRole`.

closes #24402

### DIFF
--- a/service/pom.xml
+++ b/service/pom.xml
@@ -83,6 +83,14 @@
     </dependency>
 
     <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-gateway</artifactId>
+      <classifier>tests</classifier>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-api</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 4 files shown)


service/pom.xml
```diff
@@ -82,6 +82,14 @@
...
     </dependency>
 
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-gateway</artifactId>
+      <classifier>tests</classifier>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+
```
Ben-Sheppard, line 92: ❓ I'm wondering what has caused this dependency addition,

emulbreh: That's for `StubbedBrokerClient`

Ben-Sheppard: Ah I see, nice addition :)

Ben-Sheppard (approved):
LGTM thank you for contributing so early!

Issue #22626, opened 2024-09-24
Remove permissions from Authorization


remcowesterhoud:
<h2>Description</h2>

- Add new Intents to the `AuthorizationIntent`
    - `REMOVE_PERMISSIONS`
    - `PERMISSIONS_REMOVED`
 - Create `AuthorizationRemovePermissionsProcessor` in this:
     - Verify if the owner exists
     - Verify the user is authorized to remove permissions
     - If it does, write a `PERMISSIONS_REMOVED` event and distribute the command on the Identity queue. Else, reject the command
 - Create `PermissionsRemovedApplier` in this:
     - Remove the permissions from the `Authorization` in the `AuthorizationState`

PR #24411, opened 2024-11-06
feat: add delete permissions from authorizations

### LINKED ISSUE #24365: create command and processor for an authorization deletetion

maryarm:
## Description

Camunda service allows to update permissions of an authorization assigned to an entity. The add permission has been implemented before, in this task the deletion is handled. 

## Related issues

closes #24365


remcowesterhoud:
🤔 I was working on the exact same thing. Do we have a duplicate task? https://github.com/camunda/camunda/issues/22626

remcowesterhoud:
@maryarm I think you edited my comment 😅

maryarm:
> FWIW there's a key difference between the solution I've got and the solution you've got here. I kept a clear separation between adding and removing permissions. It seems you've discarded that idea and went with a single update.

My first approach was like what you suggested here (I've created the processor and applier and ...) but then I've seen they are almost the same and the only difference is in how to handle the storage change. With this approach I didn't need to change the controller/service to map the patch request to a different broker request based on action and so on.

But I that aside,  in my idea the important point here is, how we have to proceed? I was on track to start the export based on this change so now I'm waiting the see what will be the best approach to proceed.

remcowesterhoud:
@maryarm I will keep them separated because there will be differences between the two (validations for example). I'll delegate the duplication to a behavior class.

remcowesterhoud:
@maryarm Are you okay with closing this in favor of https://github.com/camunda/camunda/pull/24490?

PR #24413, opened 2024-11-06
feat: apply authorization check on GET process definition endpoints

houssain-barouni:
## Description

Adding PROCESS_DEFINITION.READ permission check on the following endpoints

```
GET /process-definitions/{processDefinitionKey}
GET /process-definitions/{processDefinitionKey}/xml
GET /process-definitions/{processDefinitionKey}/form
```

## Checklist

- [ ] for CI changes:

## Related issues

relates to #23180


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -90,8 +90,9 @@ public static ProblemDetail mapErrorToProblem(
     }
     return switch (error) {
       case final NotFoundException nfe:
-        yield createProblemDetail(
-            HttpStatus.NOT_FOUND, nfe.getMessage(), RejectionType.NOT_FOUND.name());
+        yield createProblemDetail(HttpStatus.NOT_FOUND, nfe.getMessage(), nfe.getClass().getName());
```
houssain-barouni, line 93: I have changed the Problem.title of `NotFoundException` so that it is consistent with other Exceptions below

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/ProcessDefinitionQueryControllerTest.java
```diff
@@ -190,6 +192,34 @@ public void shouldReturnProcessDefinitionForValidKey() {
     verify(processDefinitionServices).getByKey(23L);
   }
 
+  @Test
+  public void shouldReturn403ForForbiddenProcessDefinitionKey() {
```
tmetzke, line 196: 🔧 Can we add test cases for the form and XML endpoints as well to ensure they use the same permission checks? That makes it simpler to understand later which endpoints use which permissions by just looking at the test cases.

tmetzke (approved):
👍 This is awesome, great to see that the required code for this is really straightforward, good preparation work in the previous commits 🚀 
🔧 I believe we can add two more test cases for the other endpoints to use the test cases as documentation and ensure we don't accidentally remove the permission checks from them in the future by any refactoring.

Pre-approving since the code functionally is fine already 🦖

PR #24418, opened 2024-11-06
feat: add role endpoints

emulbreh:
## Description

This PR adds endpoints for retrieving, creating and updating roles. It's part of #22389.

- [x] Create role - `POST /v2/roles`
- [x] Search roles - `POST /v2/roles/search` 
- [x] Get a single role - `GET /v2/roles/{key}`
- [x] Update a role - `PATCH /v2/roles/{key}`

### Acceptance Criteria
- [ ] Unit tests

### DIFF
--- a/service/src/main/java/io/camunda/service/RoleServices.java
+++ b/service/src/main/java/io/camunda/service/RoleServices.java
@@ -21,7 +21,7 @@ import io.camunda.zeebe.broker.client.api.BrokerClient;
 import io.camunda.zeebe.gateway.impl.broker.request.BrokerRoleUpdateRequest;
 import io.camunda.zeebe.gateway.impl.broker.request.role.BrokerRoleCreateRequest;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
-import java.util.Set;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
 public class RoleServices extends SearchQueryService<RoleServices, RoleQuery, RoleEntity> {
@@ -60,15 +60,16 @@ public class RoleServices extends SearchQueryService<RoleServices, RoleQuery, Ro
     return sendBrokerRequest(new BrokerRoleUpdateRequest(roleKey).setName(name));
   }
 
-  public RoleEntity getByRoleKey(final Long roleKey) {
-    final SearchQueryResult<RoleEntity> result =
-        search(SearchQueryBuilders.roleSearchQuery().filter(f -> f.roleKey(roleKey)).build());
-    if (result.total() < 1) {
-      throw new NotFoundException(String.format("Role with roleKey %d not found", roleKey));
-    } else {
-      return result.items().stream().findFirst().orElseThrow();
-    }
+  public RoleEntity getRole(final Long roleKey) {
+    return findRole(roleKey)
+        .orElseThrow(
+            () -> new NotFoundException("Role with roleKey %d not found".formatted(roleKey)));
   }
 
-  public record RoleDTO(long roleKey, String name, Set<Long> assignedMemberKeys) {}
+  public Optional<RoleEntity> findRole(final Long roleKey) {
+    return search(SearchQueryBuilders.roleSearchQuery().filter(f -> f.roleKey(roleKey)).build())
+        .items()
+        .stream()
+        .findFirst();
+  }
 }
--- a/service/src/main/java/io/camunda/service/RoleServices.java
+++ b/service/src/main/java/io/camunda/service/RoleServices.java
@@ -52,13 +52,12 @@ public class RoleServices extends SearchQueryService<RoleServices, RoleQuery, Ro
         brokerClient, securityContextProvider, roleSearchClient, authentication);
   }
 
-  public CompletableFuture<RoleRecord> createRole(final RoleDTO request) {
-    return sendBrokerRequest(new BrokerRoleCreateRequest().setName(request.name()));
+  public CompletableFuture<RoleRecord> createRole(final String name) {
+    return sendBrokerRequest(new BrokerRoleCreateRequest().setName(name));
   }
 
-  public CompletableFuture<RoleRecord> updateRole(final RoleDTO request) {
-    return sendBrokerRequest(
-        new BrokerRoleUpdateRequest(request.roleKey()).setName(request.name()));
+  public CompletableFuture<RoleRecord> updateRole(final long roleKey, final String name) {
+    return sendBrokerRequest(new BrokerRoleUpdateRequest(roleKey).setName(name));
   }
 
   public RoleEntity getByRoleKey(final Long roleKey) {
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1902,6 +1902,8 @@ paths:
         - Role
       operationId: createRole
       summary: Create role
+      description: |
+        Create a new role
       requestBody:
         content:
           application/json:
@@ -1924,13 +1926,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
         "401":
-          description: |
-            The request to create a role was unauthorized.
-            More details are provided in the response body.
-          content:
-            application/problem+json:
-              schema:
-                $ref: "#/components/schemas/ProblemDetail"
+          $ref: "#/components/responses/Unauthorized"
         "403":
           description: |
             The request to create a role was denied.
@@ -1940,13 +1936,41 @@ paths:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
         "500":
-          description: An internal error occurred while processing the request.
+          $ref: "#/components/responses/InternalServerError"
+
+  /roles/{roleKey}:
+    get:
+      tags:
+        - Role
+      operationId: getRole
+      summary: Get role
+      description: |
+        Get a role by its key
+      parameters:
+        - name: roleKey
+          in: path
+          required: true
+          description: The role key.
+          schema:
+            type: integer
+            format: int64
+      responses:
+        "200":
+          description: The role is successfully returned.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/RoleItem"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "404":
+          description: The role with the given key was not found.
           content:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
-
-  /roles/{roleKey}:
+        "500":
+          $ref: "#/components/responses/InternalServerError"
     patch:
       tags:
         - Role
@@ -1977,6 +2001,8 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+        "401":
+          $ref: "#/components/responses/Unauthorized"
         "404":
           description: The role with the roleKey is not found.
           content:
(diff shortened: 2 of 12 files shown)


service/src/main/java/io/camunda/service/RoleServices.java
```diff
@@ -52,24 +53,25 @@ public RoleServices withAuthentication(final Authentication authentication) {
...
-    }
+  public RoleEntity getRole(final Long roleKey) {
+    return findRole(roleKey)
+        .orElseThrow(
+            () -> new NotFoundException("Role with roleKey %d not found".formatted(roleKey)));
   }
 
-  public record RoleDTO(long roleKey, String name, Set<Long> assignedMemberKeys) {}
+  @VisibleForTesting
+  public Optional<RoleEntity> findRole(final Long roleKey) {
```
tmetzke, line 71: 🔧 As far as I'm aware, `VisibleForTesting` is not meant for public members. The Javadoc of this annotation says
> Indicates that the visibility was strengthened purely for testing purposes.
> NOTE: this should not be used on public members, but is meant only to highlight package-private or protected fields, types, etc., which aren't private in order to allow comprehensive tests. In general, you should try to avoid this, but at times there is no other way.

service/src/main/java/io/camunda/service/RoleServices.java
```diff
@@ -52,24 +53,25 @@ public RoleServices withAuthentication(final Authentication authentication) {
...
+    return search(SearchQueryBuilders.roleSearchQuery().filter(f -> f.roleKey(roleKey)).build())
+        .items()
+        .stream()
+        .findFirst();
```
tmetzke, line 75: ❓ Should we protect against scenarios where we find multiple entries with this key? Be it for issues on the data layer or refactorings that mess this up later (e.g. renaming fields in the data layer and forgetting to adjust this in the search layer). The `UserTaskServices` do something like this. In the end, the contract of this method should be to find the one role with this key and to indicate when this wasn't possible (i.e. not found or too many found).

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
         "500":
           $ref: "#/components/responses/InternalServerError"
 
+  /roles:
+    post:
+      tags:
+        - Role
+      operationId: createRole
+      summary: Create role
```
tmetzke, line 1945: 🔧 Let's provide a `description` for this endpoint as well.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "401":
+          description: |
+            The request to create a role was unauthorized.
+            More details are provided in the response body.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
tmetzke, line 1974: 🔧 You just recently introduced reusable responses for this, right? 🙂 The description will be _slightly_ more generic then but I think that is fine.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
...
+        "500":
+          description: An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
tmetzke, line 1988: 🔧 You just recently introduced reusable responses for this, right? 🙂

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/RoleItem"
+        "401":
+          description: The request lacks valid authentication credentials.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
tmetzke, line 2018: 🔧 You just recently introduced reusable responses for this, right? 🙂 The description will be slightly more generic then but I think that is fine.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
...
+        "500":
+          description: An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
tmetzke, line 2030: 🔧 You just recently introduced reusable responses for this, right? 🙂

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
...
+        "500":
+          description: >
+            An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
tmetzke, line 2073: 🔧 You just recently introduced reusable responses for this, right? 🙂

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
...
+        "500":
+          description: An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
tmetzke, line 2109: 🔧 You just recently introduced reusable responses for this, right? 🙂

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3711,6 +3882,62 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/UserResponse"
+    RoleCreateRequest:
+      type: "object"
+      properties:
+        name:
+          type: "string"
```
tmetzke, line 3889: 🔧 Let's provide a `description` for this parameter as well.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3711,6 +3882,62 @@ components:
...
+      type: object
+      properties:
+        changeset:
+          $ref: "#/components/schemas/RoleChangeset"
+      required:
+        - changeset
+    RoleChangeset:
+      type: object
+      properties:
+        name:
```
tmetzke, line 3907: 🔧 Let's provide a `description` for this parameter as well. This should explain what this does, e.g. changing the name of the role. It appears trivial but at first sight I had to think about it for a second, if this changes only the name of the role itself (so, just a "cosmetic" change) or has any further implications.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -243,6 +245,16 @@ public static Either<ProblemDetail, UpdateJobRequest> toJobUpdateRequest(
                     updateRequest.getChangeset().getTimeout())));
   }
 
+  public static Either<ProblemDetail, UpdateRoleRequest> toRoleUpdateRequest(
```
tmetzke, line 248: 🔧 The `Either` is not needed if there is no case that leads to a `Left`. We can just return the `UpdateRoleRequest` if we don't do any validation.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -243,6 +245,16 @@ public static Either<ProblemDetail, UpdateJobRequest> toJobUpdateRequest(
...
+      final RoleUpdateRequest roleUpdateRequest, final long roleKey) {
+    return Either.right(new UpdateRoleRequest(roleKey, roleUpdateRequest.getChangeset().getName()));
+  }
+
+  public static Either<ProblemDetail, CreateRoleRequest> toRoleCreateRequest(
```
tmetzke, line 253: 🔧 Same as above, no `Either` needed.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1936,6 +1937,176 @@ paths:
...
+        - Role
+      operationId: createRole
+      summary: Create role
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/RoleCreateRequest"
+      responses:
+        "202":
```
tmetzke, line 1952: 🔧 Shouldn't this be a `201 Created`? I guess we have a similar inconsistency for other endpoints that "create" resources. While a 200 is understandable maybe, a 202 usually means the action was just accepted but not completed yet. This isn't really the case here, is it?

```suggestion
        "201":
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/ResponseMapper.java
```diff
@@ -323,6 +325,11 @@ public static ResponseEntity<Object> toUserCreateResponse(final UserRecord userR
     return new ResponseEntity<>(response, HttpStatus.ACCEPTED);
   }
 
+  public static ResponseEntity<Object> toRoleCreateResponse(final RoleRecord roleRecord) {
+    final var response = new RoleCreateResponse().roleKey(roleRecord.getRoleKey());
+    return new ResponseEntity<>(response, HttpStatus.ACCEPTED);
```
tmetzke, line 330: 🔧 If we change the OpenAPI spec to 201, we should change it here as well.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/RoleControllerTest.java
```diff
@@ -0,0 +1,89 @@
...
+import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
+import java.util.concurrent.CompletableFuture;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.http.MediaType;
+
+@WebMvcTest(RoleController.class)
+public class RoleControllerTest extends RestControllerTest {
```
tmetzke, line 30: 🔧 Let's test some edge cases and ensure the controller reacts accordingly, e.g. providing no name when creating a role, receiving an exception from the service layer like NOT_FOUND. The `UserControllerTest` provides some examples for this.

tmetzke (commented):
👍 The base functionality looks good to me 🚀 
🔧 There are some smaller bits here and there that can be improved but nothing major, good work 🦖

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
emulbreh: I was following `createUser`, which does return 202. As far as I understand, the API is only eventually consistent (because writes need to be exported to ES/OS), so 202 would make sense. 201 also works for me though, or even 200 if we make it return the full entity.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
emulbreh: Right, I forgot to update this PR afterwards!

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
emulbreh: I chose Either for consistency with the other mapping functions, but I now see that there's already exceptions. I'll change it.

emulbreh: I actually ended up adding validation, so it's back to Either now.

pepopowitz (approved):
The spec looks good from my perspective!

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Yeah, I saw the create user endpoint as well. The eventual consistency does not count here, if you ask me. The command is processed in the engine and comes back with a result that is also written to the log at that point. The eventual consistency is only for secondary storage, that should not influence the API response.
So, a 201 makes most sense here in my regard.

If you're up for it, you can also adjust the user endpoint, actually.

tmetzke: Either here or in a follow-up to not bloat this PR

emulbreh: I'll make a follow-up PR.

service/src/main/java/io/camunda/service/RoleServices.java
emulbreh: I agree, there's a trade-off though: with a similar argument, we could check that the returned results actually have the requested key (this is not completely unthinkable given that we're querying ES). But the additional code is unlikely to pay off and hurts readability. So we need to draw the line somewhere. 

I don't want to turn this into an argument though, if you prefer to check for multiple results that works for me.

tmetzke: Get your point and agree, we have to draw the line somewhere 👍
I am fine with keeping things as they are here specifically 👍

PR #24420, opened 2024-11-06
feat: add partial OS repository with delete documents support

npepinpe:
## Description

This PR adds the beginning of an `OpenSearchRepository` implementation, right now only support `deleteDocuments`. We should implement each method as a separate PR, and finally one PR to integrate it in the `Archiver` itself.

Right now the `OpenSearchRepository` extends `NoopArchiverRepository` and overrides the functionality it implements, and we should keep that going until it implements everything, then swap to just implementing `ArchiverRepository`.

## Related issues

related to #24094

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/OpenSearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/OpenSearchRepository.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.archiver;
+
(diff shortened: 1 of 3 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/OpenSearchRepository.java
```diff
@@ -0,0 +1,125 @@
...
+            .index(sourceIndexName)
+            .slices(AUTO_SLICES)
+            .conflicts(Conflicts.Proceed)
+            .query(q -> q.terms(termsQuery))
+            .build();
+
+    final var timer = Timer.start();
+    return sendRequestAsync(() -> client.deleteByQuery(request))
+        .whenCompleteAsync((ignored, error) -> metrics.measureArchiverDelete(timer), executor)
+        .thenApplyAsync(DeleteByQueryResponse::total, executor)
```
rodrigo-lourenco-lopes, line 92: This is the total of records processed correct? Are we using this in the future for metrics? In the documentation I also saw another response field which is delete -> "Total number of documents deleted.". I am not sure what is exactly the difference between the two in this case. 🤔

rodrigo-lourenco-lopes (approved):
Looks good! 🚀 🚀

PR #24432, opened 2024-11-06
refactor: harmonize Problem response creation in Query controllers

houssain-barouni:
## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionRequirementsQueryController.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionRequirementsQueryController.java
@@ -7,7 +7,6 @@
  */
 package io.camunda.zeebe.gateway.rest.controller;
 
-import static io.camunda.zeebe.gateway.rest.Loggers.REST_LOGGER;
 import static io.camunda.zeebe.gateway.rest.RestErrorMapper.mapErrorToResponse;
 
 import io.camunda.search.query.DecisionRequirementsQuery;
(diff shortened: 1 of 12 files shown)


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionRequirementsQueryController.java
```diff
@@ -84,11 +72,9 @@ public ResponseEntity<DecisionRequirementsItem> getByKey(
           .body(
               SearchQueryResponseMapper.toDecisionRequirements(
                   decisionRequirementsServices.getByKey(decisionRequirementsKey)));
-    } catch (final Exception exc) {
-      REST_LOGGER.warn("An exception occurred in get Form by key.", exc);
-      final var problemDetail =
-          RestErrorMapper.mapErrorToProblem(exc, RestErrorMapper.DEFAULT_REJECTION_MAPPER);
-      return RestErrorMapper.mapProblemToResponse(problemDetail);
+    } catch (final Exception e) {
+      REST_LOGGER.debug("An exception occurred in getDecisionRequirements.", e);
```
tmetzke, line 76: 💭 The stack trace will already tell us where the exception occurred, right? Could we simply add the logging to the error mapping method so it's handled for all callers?

tmetzke (approved):
So, all tests are simply passing with that change. Either, the previous code was really not needed, or we don't test exceptional cases well enough 😄
I have to admit I suspect the latter of those options but hey, #yolo 🤣 
I really like the change, by the way 👍

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/DecisionRequirementsQueryController.java
houssain-barouni: Good idea 👍 
I just figured out that most of the exceptions are logged in https://github.com/camunda/camunda/blob/befb8e130097dbe154ebc7458108e73f5cf45d8a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java#L86-L175

I will add some debug/trace logs for the missing ones

PR #24433, opened 2024-11-06
build: replace Jetty with Tomcat in Optimize

### LINKED ISSUE #24278: Replace Jetty with Tomcat in Optimize 8
### Context

Currently, Optimize runs on Jetty.
In order to integrate Optimize in the Single Application, it should use Tomcat, instead.

### Acceptance Criteria

[ ] Optimize runs on Tomcat instead of Jetty

### Links

* Part of https://github.com/camunda/product-hub/issues/2289

### Breakdown

```[tasklist]
### Pull Requests
```

## Task Lifecycle

#### Engineering Review

#### QA Review

### Completion

buccarel:
## Description

Currently, the Optimize backend is running on the web server Jetty. In order to facilitate the move towards the single-jar and single-application projects, we need to replace Jetty with Tomcat, because:
* every application is using Tomcat
* the only way to make Tomcat and Jetty coexist in a multi-module application would be to use ad-hoc profile to mutually exclude the server that is not in use, and this is not what we want

This PR removes Jetty and its dependencies and adapts the configuration to work using Tomcat instead.

**Important Notes:**

* custom error handler removed and not replaced with anything: I tried to understand what cases its code would solve but I couldn't find any. After consulting with Josh, I decided to remove it, as it doesn't seem to impact anything. If I am wrong about it, please feel free to let me know and it will be quick to reintroduce it.

* there are some slight differences between Jetty and Tomcat related to the order some things are computed. Because of some small differences in behavior, we are installing the filters on the pattern `/*` and then we are checking whether or not we apply the filter, directly within the filter

* the pipeline `e2e-cloud-test` has been disabled: it's correct functioning is impacted by Tomcat, as the new webserver doesn't allow for the auth cookie to be passed from our oauth server (https) to the local instance of the application (http) that is used to run the tests. A new issue has been created to fix this problem eventually.

# Testing

Before merging this PR, an extensive QA/testing should be performed, to make sure the application works as expected, identically to when Jetty was the web server. Some of the important things we want to make sure are the following:
* regression tests on the UI
* migrations shall work as expected
* static files are served with the /external prefix, with the cluster ID and with the contextPath as expected
* external APIs shall work as expected
  * with /external/api/... URLs
  * with /api/external/URLs
* non-external, publicly accessible APIs shall work as expected
* the sharing functionality shall work as expected
* the user-facing server customizations shall work as expected
* **(done @buccarel)** the Gzip compression shall work as expected
* **(done @buccarel)** HTTP2 works as expected
* the application works as expected when clusterId is present (cloud environment)
* all of the things mentioned above shall work as expected even with a custom context path, as well as in SaaS

## Checklist

- [ ] for CI changes:

## Related issues

* closes https://github.com/camunda/camunda/issues/24278
* part of https://github.com/camunda/product-hub/issues/2289


optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,220 @@
...
+  private static final String LOGIN_ENDPOINT = "/login";
+  private static final String METRICS_ENDPOINT = "/metrics";
+  private static final String URL_BASE = "/#";
+
+  public static final String EXTERNAL_SUB_PATH = "/external";
+  public static final String ALLOWED_URL_EXTENSION =
+      String.join(
+          "|",
+          new String[] {
+            "/",
```
buccarel, line 66: I had to add this, otherwise Tomcat would not validate the home page, resulting into an infinite loop of redirections to the home page.

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,220 @@
...
+        LOG.debug("Registering bean externalResourcesHandler...");
+        URL webappURL = getClass().getClassLoader().getResource("webapp");
+        if (webappURL == null) {
+          LOG.debug("Static content directory 'webapp' not found. No bean will be registered.");
+          return;
+        }
+
+        String webappPath = webappURL.toExternalForm().replaceFirst("file:", "");
+        ServletRegistration.Dynamic webappServlet =
+            servletContext.addServlet("external-home", ExternalHomeServlet.class);
```
buccarel, line 119: To do this, I had to implement a custom servlet that serves the static files. I didn't find a more immediate or automatic way to do it with Tomcat. If you know how to do it better, please feel free to let me know

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ExternalHomeServlet.java
```diff
@@ -0,0 +1,42 @@
+/*
```
buccarel, line 1: This is the new servlet I mentioned in my own comment above. It serves the static files under the webapp directory (the built client, basically)

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ResponseHeadersFilter.java
```diff
@@ -0,0 +1,44 @@
+package io.camunda.optimize.tomcat;
```
buccarel, line 1: I am always chaining this filter.
I don't have a lot of context about these headers. If you think this only applies to a subset of the cases, please feel free to let me know.

optimize/backend/src/main/java/io/camunda/optimize/rest/security/ccsm/CCSMSecurityConfigurerAdapter.java
```diff
@@ -130,7 +130,8 @@ protected SecurityFilterChain configureWebSecurity(final HttpSecurity http) {
                       // public share related resources (API)
                       .requestMatchers(
                           new AntPathRequestMatcher(
-                              createApiPath(EXTERNAL_SUB_PATH + DEEP_SUB_PATH_ANY)))
+                              createApiPath(EXTERNAL_SUB_PATH + DEEP_SUB_PATH_ANY)),
+                          new AntPathRequestMatcher(EXTERNAL_SUB_PATH + "/api/**"))
```
buccarel, line 134: I had to add this to allow `/external/api/anything` URLs.

I don't know why it was working fine without this rule before. My guess is that the URL transformation from `/external/api/*` to `/api/external/*` was happening before getting here, with Jetty.

I am doing that transformation with a new filter now. See TomcatConfig.java.

RomanJRW (commented):
Please do not consider this a "proper" review. I have not tested things and really just wanted to get a feel for the approach. Generally, I think this looks reasonable and in some ways is simpler than the quite convoluted Jetty config. I have left some comments/thoughts on different areas to consider.

Most important are:
- Testing: This is a comprehensive change and comprehensive QA should come with it, clearly
- Config: Let's be sure that all of the various ways in which we allowed customers to configure their jetty server are still configurable now. Or if not the case, we would need to be clear in our documentation that we are deleting some config option

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
buccarel: I really don't know. It's been a puzzle

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,220 @@
...
+          });
+
+  private static final String HTTP11_NIO_PROTOCOL = "org.apache.coyote.http11.Http11Nio2Protocol";
+
+  @Autowired private ConfigurationService configurationService;
+  @Autowired private Environment environment;
+
+  private void setSendServerVersion(Connector connector, boolean toggle) {
+    connector.setXpoweredBy(toggle);
+  }
```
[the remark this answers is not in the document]
buccarel: Yeah in general I am planning to move all of the private methods below.
I went for the wrapping method because the new connector's name is not as self-explainatory as the old method we had.

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,220 @@
...
+
+  @Bean
+  public FilterRegistrationBean<ExternalApiRewriteFilter> externalApiRewriteFilter() {
+    FilterRegistrationBean<ExternalApiRewriteFilter> filterRegistrationBean =
+        new FilterRegistrationBean<>();
+    filterRegistrationBean.setFilter(new ExternalApiRewriteFilter());
+    filterRegistrationBean.addUrlPatterns("/*");
```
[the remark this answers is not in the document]
buccarel: When I tried, it didn't work.
Maybe I'm doing it wrong, but it's possible that the pattern matching is being applied slightly differently, so I went for the safest option.

Maybe I'll try again

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,220 @@
...
+    registrationBean.setFilter(responseHeadersFilter);
+    return registrationBean;
+  }
+
+  public int getPort(final String portType) {
+    final String portProperty = environment.getProperty(portType);
+    if (portProperty == null) {
+      return portType.equals(EnvironmentPropertiesConstants.HTTPS_PORT_KEY)
+          ? configurationService.getContainerHttpsPort()
+          : configurationService.getContainerHttpPort().orElse(8090);
```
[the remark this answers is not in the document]
buccarel: I agree with you on this.
I'll change it.
Do we need to change documentation about this?

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ExternalApiRewriteFilter.java
```diff
@@ -0,0 +1,43 @@
...
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import org.slf4j.Logger;
+
+public class ExternalApiRewriteFilter implements Filter {
+
+  private static final String EXTERNAL_API_PREFIX = "/external/api";
+  private static final String API_EXTERNAL_PREDIX = "/api/external";
```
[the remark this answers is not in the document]
buccarel: Ooops.
Fixed.

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ExternalApiRewriteFilter.java
```diff
@@ -0,0 +1,43 @@
...
+      org.slf4j.LoggerFactory.getLogger(ExternalApiRewriteFilter.class);
+
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+      throws IOException, ServletException {
+    HttpServletRequest httpRequest = (HttpServletRequest) request;
+    HttpServletResponse httpResponse = (HttpServletResponse) response;
+
+    String requestURI = httpRequest.getRequestURI();
+
+    if (requestURI.contains(EXTERNAL_API_PREFIX)) {
```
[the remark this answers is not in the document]
buccarel: In the previous case, we had a regular expression matching.

I got a bit defensive with these because in a separate case, that didn't work as expected anymore (didn't investigate why). I am trying to keep things simpler, when more complex code is not necessary. Let me know if you want me to restore the previous pattern matching.

In this case, we can be more specific and I will do it, but basically it's the case where the URL starts with `/external/api/`. I can change contains with startsWith.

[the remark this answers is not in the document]
buccarel: replaced contains in favor of startsWith

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ResponseHeadersFilter.java
```diff
@@ -0,0 +1,44 @@
...
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class ResponseHeadersFilter implements Filter {
+
+  private ConfigurationService configurationService;
+
+  public ResponseHeadersFilter(ConfigurationService configurationService) {
+    if(configurationService == null) {
```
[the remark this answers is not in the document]
buccarel: Yep, I didn't apply spotless yet.
When the PR will be final, all of these things will be addressed, so that the PR will be green.

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java
```diff
@@ -120,6 +120,7 @@ public class ConfigurationService {
   private Integer containerHttpsPort;
   private Integer actuatorPort;
   private Boolean containerHttp2Enabled;
+  private Boolean containerHttpsEnabled;
```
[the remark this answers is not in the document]
buccarel: No this was an experiment of mine, to check the behavior.
I forgot to remove this, when I restored the code. Sorry

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ExternalApiRewriteFilter.java
```diff
@@ -0,0 +1,43 @@
...
+      String rewrittenURI = requestURI.replaceFirst(EXTERNAL_API_PREFIX, API_EXTERNAL_PREDIX);
+      RequestDispatcher dispatcher = request.getRequestDispatcher(rewrittenURI);
+      dispatcher.forward(httpRequest, httpResponse);
```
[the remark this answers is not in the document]
buccarel: No, I am doing `chain.doFilter` only in the `else` branch.
The reason is that after doing `dispatcher.forward`, a new request with a new URI will take the place of the current one. The chain filter will be processed against that

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ExternalApiRewriteFilter.java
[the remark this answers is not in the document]
RomanJRW: Happy with startsWith, as long as this works with the context path too as discussed 👍

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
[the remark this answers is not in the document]
RomanJRW: I don't think so

optimize/backend/src/main/java/io/camunda/optimize/tomcat/URLRedirectFilter.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @Override
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+      throws IOException, ServletException {
+    HttpServletRequest httpRequest = (HttpServletRequest) request;
+    HttpServletResponse httpResponse = (HttpServletResponse) response;
+
+    String requestPath = httpRequest.getRequestURI();
+    if (redirectPattern.matcher(requestPath).matches()) {
```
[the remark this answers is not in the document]
buccarel: As we discussed via PM, for the moment let's keep the TomcatConfig simpler and match the URLs inside the filters, for technical reasons tied to differences between Jetty and Tomcat.

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,273 @@
...
+        webappServlet.setInitParameter("resourceBase", webappPath);
+        webappServlet.addMapping("/external/*");
+        webappServlet.addMapping("/*");
+        webappServlet.setLoadOnStartup(1);
+      }
+    };
+  }
+
+  @Bean
+  FilterRegistrationBean<ExternalApiRewriteFilter> externalApiRewriter() {
```
buccarel, line 146: Should I move all of the filters in the file `FilterBeansConfig` maybe?

RomanJRW: move them where? To the TomcatConfig?

buccarel: in the file FilterBeansConfig.java, maybe

optimize/backend/pom.xml
```diff
@@ -161,21 +127,28 @@
...
+    <dependency>
+      <groupId>org.apache.tomcat.embed</groupId>
+      <artifactId>tomcat-embed-core</artifactId>
+      <version>10.1.30</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.tomcat</groupId>
+      <artifactId>tomcat-annotations-api</artifactId>
+      <version>10.1.30</version>
```
RomanJRW, line 148: NABD but we could bind these versions to a property

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,273 @@
...
+      @Override
+      public void customize(final TomcatServletWebServerFactory factory) {
+        final Optional<String> contextPath = getContextPath();
+        if (contextPath.isPresent()) {
+          factory.setContextPath(contextPath.get());
+        }
+
+        // NOTE: With the current implementation, we are always installing 2 connectors,
+        //   one for HTTP, one for HTTPs. The latter can be HTTP/1.1 or HTTP/2 depending
+        //   on the configuration.
```
RomanJRW, line 104: This feels clearer than before, thanks 👍

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,273 @@
...
+
+  private static final String HTTP11_NIO_PROTOCOL = "org.apache.coyote.http11.Http11Nio2Protocol";
+
+  @Autowired private ConfigurationService configurationService;
+
+  @Autowired private Environment environment;
+
+  @Bean
+  WebServerFactoryCustomizer<TomcatServletWebServerFactory> tomcatFactoryCustomizer() {
+    LOG.debug("Setting up connectors...");
```
RomanJRW, line 93:
```suggestion
    LOG.debug("Setting up Server connectors...");
```
Nitpicky, but feels appropriate given that connectors within Camunda means a very different thing

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,273 @@
...
+    };
+  }
+
+  @Bean
+  ServletContextInitializer externalResourcesServlet() {
+    LOG.debug("Registering servlet 'externalResourcesServlet'...");
+    return new ServletContextInitializer() {
+      @Override
+      public void onStartup(final ServletContext servletContext) throws ServletException {
+        LOG.debug("Registering bean externalResourcesHandler...");
```
RomanJRW, line 127: this doesn't look right

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  private void setMaxHeaderSize(final Connector connector) {
+    // NOTE: In Tomcat, the request and response header size are both controlled
+    // by a single property called maxHeaderSize.
+    final int maxHeaderSize =
+        Math.max(
+            configurationService.getMaxResponseHeaderSizeInBytes(),
+            configurationService.getMaxRequestHeaderSizeInBytes());
+    connector.setProperty("maxHeaderSize", String.valueOf(maxHeaderSize));
```
RomanJRW, line 246: probably a good one for @matthewBearCamunda to test.

I am not sure I like the approach of taking the max value for both here. I think a better solution would be to either just pick one and we delete the other property (documenting that we are doing so). Alternatively, we introduce a new property to better capture this behaviour.

In any case, you can create a separate issue to revisit this in a follow-up if you like

optimize/backend/src/main/java/io/camunda/optimize/tomcat/URLRedirectFilter.java
```diff
@@ -0,0 +1,72 @@
...
+      throws IOException, ServletException {
+    final HttpServletRequest httpRequest = (HttpServletRequest) request;
+    final HttpServletResponse httpResponse = (HttpServletResponse) response;
+    final String requestPath = httpRequest.getRequestURI();
+    final String contextPath = httpRequest.getContextPath();
+
+    /* Handle missing trailing slash to home */
+    if (contextPath.equals(requestPath)) {
+      httpResponse.sendRedirect(redirectPath);
+      return;
```
RomanJRW, line 45: does the redirect path contain the context path?

optimize/backend/src/main/java/io/camunda/optimize/tomcat/URLRedirectFilter.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /* Validate requests to the home page */
+    if ((contextPath + "/").equals(requestPath)) {
+      chain.doFilter(request, response);
+      return;
+    }
+
+    /* Validate requests to the static resources */
+    String staticRequestPath = requestPath;
+    if (!"".equals(contextPath)) {
```
RomanJRW, line 56:
```suggestion
    if (!contextPath.isEmpty()) {
```

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
buccarel: Yeah, I forgot to comment here.

For the moment, I opted not to change the configuration and just use it the best way I could, but my preference would be to change the documentation and use a variable that corresponds to what Tomcat uses, that is, one single variable that controls the min and the max size.

I decided not to do it now just because it implies a doc change. I am going to file a task to do it later, thanks.

buccarel: Done:
https://github.com/camunda/camunda/issues/24629

optimize/backend/src/main/java/io/camunda/optimize/tomcat/URLRedirectFilter.java
buccarel: Yes, in TomcatConfig, we compute it as `getContextPath().orElse("") + "/#"`

matthewBearCamunda (commented):
Just a couple of things I found. I focused on the `TomcatConfig` class as thats the one that most closely relates to the work with the jetty header

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,271 @@
...
+    connector.setProperty("compression", "on");
+    connector.setProperty("compressionMinSize", "23");
+    connector.setProperty("compressionNoCompressionMethods", ""); // all methods
+    connector.setProperty("useSendfile", "false");
+    connector.setProperty("compressableMimeType", String.join(",", COMPRESSED_MIME_TYPES));
+  }
+
+  private void setMaxHeaderSize(final Connector connector) {
+    // NOTE: In Tomcat, the request and response header size are both controlled
+    // by a single property called maxHeaderSize.
```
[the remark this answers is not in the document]
buccarel: Oh I think you're right!
I don't know how I didn't see this when I checked. Thanks!

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
buccarel: I think this goes away thanks to Matt's comment below

matthewBearCamunda: @buccarel - I'd still want us to manually test out the header size change

optimize/backend/src/main/java/io/camunda/optimize/tomcat/SaasRequestAdjustmentFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    final HttpServletRequest httpRequest = (HttpServletRequest) request;
+    String requestURI = httpRequest.getRequestURI();
+
+    /* Strip cluster ID */
+    if (!clusterId.isEmpty() && requestURI.startsWith("/" + clusterId)) {
+      requestURI = requestURI.substring(("/" + clusterId).length());
+      forward = true;
+    }
+
+    /* transform /external/api -> /api/external */
```
buccarel, line 45: the reason why I'm applying 2 transformation in the same filter (instead of having a common externalApiRewriter + the cluster ID filter) is because after the first transformation+forward, the forwarded request doesn't go through the filter chain again

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,260 @@
...
+
+  private void enableGzipSupport(final Connector connector) {
+    connector.setProperty("compression", "on");
+    connector.setProperty("compressionMinSize", "23");
+    connector.setProperty("compressionNoCompressionMethods", ""); // all methods
+    connector.setProperty("useSendfile", "false");
+    connector.setProperty("compressableMimeType", String.join(",", COMPRESSED_MIME_TYPES));
+  }
+
+  private void configureHttpConnector(final Connector connector) {
```
matthewBearCamunda, line 226: Connector settings in this method and `configureHttpsConnector()` are doing the same thing in some cases. Could you use a `private` helper method for the ones the two methods have in common?

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
```diff
@@ -105,6 +111,21 @@ public AuthenticationCookieFilter authenticationCookieFilter(final HttpSecurity
             .build());
   }
 
+  @Bean
+  @Order(-100)
```
matthewBearCamunda, line 115: What is this annotation for? Could you add a comment explaining what `-100` is doing here? It's also used in the method itself

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
```diff
@@ -141,7 +164,8 @@ protected SecurityFilterChain configureWebSecurity(final HttpSecurity http) {
                       // public share related resources (API)
                       .requestMatchers(
                           new AntPathRequestMatcher(
-                              createApiPath(EXTERNAL_SUB_PATH + DEEP_SUB_PATH_ANY)))
+                              createApiPath(EXTERNAL_SUB_PATH + DEEP_SUB_PATH_ANY)),
+                          new AntPathRequestMatcher(EXTERNAL_SUB_PATH + "/api/**"))
```
matthewBearCamunda, line 168: Could you use `OptimizeResourceConstants.REST_API_PATH` here?

optimize/backend/src/main/java/io/camunda/optimize/tomcat/CcsmRequestAdjustmentFilter.java
```diff
@@ -0,0 +1,46 @@
...
+import jakarta.servlet.Filter;
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.RequestDispatcher;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class CcsmRequestAdjustmentFilter implements Filter {
```
matthewBearCamunda, line 19: What is this filter needed for? Why modify the URI that is being used?

optimize/backend/src/main/java/io/camunda/optimize/tomcat/SaasRequestAdjustmentFilter.java
```diff
@@ -0,0 +1,61 @@
...
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.RequestDispatcher;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class SaasRequestAdjustmentFilter implements Filter {
```
matthewBearCamunda, line 20: Some code in this filter looks similar to the code in `CcsmRequestAdjustmentFilter`, could you refactor the common code so that nothing is duplicated?

optimize/backend/src/main/java/io/camunda/optimize/FilterBeansConfig.java
```diff
@@ -31,6 +32,9 @@
  */
 @Configuration
 public class FilterBeansConfig {
+
+  @Autowired private ConfigurationService configurationService;
```
matthewBearCamunda, line 36: Does `ConfigurationService` need to be added here? I don't see where it is used

buccarel: Oh, it's a change I did then I reverted and I forgot to remove this.
Nice catch, thanks!

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
buccarel: I'm going to use a constant in both cases with value -100 to set them and I will also add comments.

@Order specifies the order of loading, the other specifies the order in the filter chain. I'm using a high priority value with both, so that this filter happens before anything. It's been working in my case, but if I am wrong with something, please feel free to let me know.

optimize/backend/src/main/java/io/camunda/optimize/tomcat/SaasRequestAdjustmentFilter.java
buccarel: When I tried the refactoring, I saw that the duplicate part is this:
```
if (requestURI.startsWith("/external/api/")) {
      final String rewrittenURI = requestURI.replaceFirst("/external/api/", "/api/external/");
```
everything else has small differences (like the `forward = true` that is only done in 1 class) that would cost more overhead with the refactoring.

Although I don't have a strong opinion about this: if it's a blocker for you, I'll try to refactor it, otherwise I'd prefer leaving it as it is now, for simplicity

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
matthewBearCamunda: Is the goal here to load this bean first? You can use `Ordered.HIGHEST_PRECEDENCE` instead of a magic number here

optimize/backend/src/main/java/io/camunda/optimize/tomcat/CcsmRequestAdjustmentFilter.java
RomanJRW: @buccarel - thanks for the comprehensive write up here 👍 

Can you create issue(s) for the follow-up tasks you mention here? They could be linked to the epic as follow-up tasks if appropriate, but it would be good to ensure they don't get missed/forgotten

.github/workflows/optimize-e2e-test-cloud.out
```diff
@@ -0,0 +1,106 @@
...
+      - ".github/actions/**"
+      - ".github/workflows/optimize-*"
+      - "bom/*"
+      - "parent/*"
+      - "pom.xml"
+      - "optimize/**"
+      - "optimize.Dockerfile"
+  workflow_dispatch:
+
+jobs:
```
RomanJRW, line 32: rather than commenting out the file, maybe a line like this with a comment would be sufficient:
```
# Some comment explaining why we disabled it
if: false
```

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,257 @@
...
+import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory;
+import org.springframework.boot.web.server.WebServerFactoryCustomizer;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
+import org.springframework.boot.web.servlet.ServletContextInitializer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
+
+@Configuration
+public class TomcatConfig {
```
RomanJRW, line 47: NABD but could call this `OptimizeTomcatConfig` just for specificity, given tomcat is used broadly

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
```diff
@@ -105,6 +112,21 @@ public AuthenticationCookieFilter authenticationCookieFilter(final HttpSecurity
...
+        new FilterRegistrationBean<>();
+    registration.setOrder(Ordered.HIGHEST_PRECEDENCE); /* position in the chain */
+    registration.addUrlPatterns("/*");
+    registration.setFilter(saasRequestAdjustmentFilter);
+    return registration;
+  }
+
   @Bean
   @Order(1)
   protected SecurityFilterChain configurePublicApi(final HttpSecurity http) {
```
RomanJRW, line 132: out of interest, did you test the public API?

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
```diff
@@ -125,14 +147,16 @@ protected SecurityFilterChain configureWebSecurity(final HttpSecurity http) {
...
                   httpRequests
+                      // ready endpoint is public for infra
                       .requestMatchers(new AntPathRequestMatcher(createApiPath(READYZ_PATH)))
                       .permitAll()
                       // public share resources
                       .requestMatchers(
                           new AntPathRequestMatcher(EXTERNAL_SUB_PATH + "/"),
+                          new AntPathRequestMatcher("/index*"),
                           new AntPathRequestMatcher(EXTERNAL_SUB_PATH + "/index*"),
+                          new AntPathRequestMatcher(STATIC_RESOURCE_PATH + "/**"),
```
RomanJRW, line 159: why was this not needed before?

optimize/backend/src/main/java/io/camunda/optimize/tomcat/CcsmRequestAdjustmentFilter.java
RomanJRW, line 19: Maybe go with CCSM capitalized just for consistency with other beans in Optimize

optimize/backend/src/main/java/io/camunda/optimize/tomcat/CcsmRequestAdjustmentFilter.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public void doFilter(
+      final ServletRequest request, final ServletResponse response, final FilterChain chain)
+      throws IOException, ServletException {
+    final HttpServletRequest httpRequest = (HttpServletRequest) request;
+
+    String requestURI = httpRequest.getRequestURI();
+    final String originalRequestURI = requestURI;
+
+    /* strip contextPath */
```
RomanJRW, line 30: nice that you added comments to the filters, I think this was the reason it was so difficult to understand before

optimize/backend/src/main/java/io/camunda/optimize/tomcat/SaasRequestAdjustmentFilter.java
```diff
@@ -0,0 +1,75 @@
...
```
RomanJRW, line 20:
```suggestion
public class CCSaaSRequestAdjustmentFilter implements Filter {
```
Again, just for consistency

RomanJRW (approved):
Note: I would also like a review from @matthewBearCamunda 

Generally I think things look ok to me, great job! The real test for its correctness will come with QA. Please ensure that we test things comprehensively, including:
- The public API
- Shared endpoints
- UI functionality
- Authentication flow

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
buccarel: I did some local tests in self-managed as well as some tests against the cluster.
I listed the API  as one of the things to QA in the description

RomanJRW: and to confirm, you mean the public API and not the shared report/dashboard functionality?

buccarel: To my knowledge, the shared functionality uses the API with the `/external/api/` prefix.
I tried to execute:
* /external/api/ui-configuration
* /api/external/ui-configuration
* /api/ui-configuration (this should be public)

and everything worked, in self-managed and against the cluster. I did some other tests with the readyz endpoint, but not more extended than this. So I pretty much made sure that the requests are forwarded to the correct handler, but a more exhaustive test set should be done, to make sure all of the APIs work.

If we're good with time, an integration test could be good

optimize/backend/src/main/java/io/camunda/optimize/tomcat/CcsmRequestAdjustmentFilter.java
buccarel: Done: https://github.com/camunda/camunda/issues/24910

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/security/ResponseHeadersConfiguration.java
```diff
@@ -112,4 +117,35 @@ public String toString() {
...
+      headers.put("X-XSS-Protection", getXsssProtection());
+    }
+
+    if (getXContentTypeOptions()) {
+      headers.put("X-Content-Type-Options", "nosniff");
+    }
+
+    if (getContentSecurityPolicy() != null && getContentSecurityPolicy().length() > 0) {
+      headers.put("Content-Security-Policy", getContentSecurityPolicy());
+    }
```
matthewBearCamunda, line 147: Instead of using a string for the header properties, you can use existing contants

For example, `HttpHeaders.X_XSS_PROTECTION` instead of `"X-XSS-Protection"`

.github/workflows/optimize-e2e-test-cloud.out
```diff
@@ -0,0 +1,106 @@
...
+          vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
+          vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
+      - name: Setup Chrome
+        shell: bash
+        run: |
+          # remove existing chrome
+          sudo apt-get remove -y google-chrome-stable
+          # install chrome 126
+          wget https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_126.0.6478.182-1_amd64.deb
+          sudo apt-get update && sudo apt-get install -y ./google-chrome-stable_126.0.6478.182-1_amd64.deb
```
matthewBearCamunda: I know this is a copy/paste from the other workflow, and this is not a blocker, but is this going to be easily maintainable? Chrome 126 will _eventually_ go out of support. Is there a better way to get a good version of Chrome rather than hard-coding it?

optimize/backend/pom.xml
```diff
@@ -12,12 +14,11 @@
...
-    <jersey.version>3.1.3</jersey.version>
     <spring.security.version>6.3.4</spring.security.version>
-    <nimbus.jose.jwt.version>9.41.2</nimbus.jose.jwt.version>
-    <it.test.excludedGroups />
-    <it.test.includedGroups />
-    <test.forkCount />
+    <it.test.excludedGroups/>
+    <it.test.includedGroups/>
+    <test.forkCount/>
+    <version.tomcat>10.1.30</version.tomcat>
```
matthewBearCamunda, line 21: Why use `10.1.30` when `10.1.33` is the latest version?

optimize/backend/src/main/java/io/camunda/optimize/TomcatConfig.java
```diff
@@ -0,0 +1,257 @@
...
+    LOG.debug("Registering filter 'urlRedirector'...");
+
+    final String contextPath = getContextPath().orElse("");
+
+    // This regex includes all the URL suffixes that we allow.
+    // The list of suffixes is stored in ALLOWED_URL_EXTENSION, and the regex does not
+    // handle the home page URL: that is handled explicitly from within the URLRedirectFilter.
+    final String regex = "^(?!" + "(" + contextPath + ALLOWED_URL_EXTENSION + ")).+";
+
+    final URLRedirectFilter filter = new URLRedirectFilter(regex, contextPath + "/#");
```
matthewBearCamunda: Can use `URL_BASE` instead of `/#` here

optimize/backend/pom.xml
buccarel: if I remember well, I simply followed the suggestion of the IDE. I'll try with the latest version

optimize/backend/src/main/java/io/camunda/optimize/tomcat/MaxRequestSizeFilter.java
```diff
@@ -89,7 +89,7 @@ public long getMaxContentLength() {
   public void writeErrorResponse(
       final HttpServletResponse httpResponse, final String errorMessage, final int statusCode) {
     httpResponse.reset();
-    httpResponse.setContentType(MimeTypes.Type.APPLICATION_JSON_UTF_8.toString());
+    httpResponse.setContentType(APPLICATION_JSON_UTF_8);
```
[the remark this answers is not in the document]
buccarel: The problem is that class is in a jetty dependency:
https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/http/MimeTypes.Type.html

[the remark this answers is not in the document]
matthewBearCamunda: Are there any similar constants in Tomcat? It'd be better to use something that exists rather than adding something new here

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ResponseHeadersFilter.java
```diff
@@ -0,0 +1,46 @@
...
+    this.configurationService = configurationService;
+  }
+
+  @Override
+  public void doFilter(
+      final ServletRequest request, final ServletResponse response, final FilterChain chain)
+      throws IOException, ServletException {
+    final HttpServletResponse httpResponse = (HttpServletResponse) response;
+
+    final var headers =
```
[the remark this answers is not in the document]
RomanJRW: I think in the mono repo we can use `var` if needed. Not to say this can't be changed, just that there is less strictness around this than in the Optimize repo

optimize/backend/src/main/java/io/camunda/optimize/tomcat/MaxRequestSizeFilter.java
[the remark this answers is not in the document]
buccarel: I found `MediaType.JSON_UTF_8.type()` from Guava. We have that library included, I'll use that :D

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ResponseHeadersFilter.java
```diff
@@ -0,0 +1,46 @@
...
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class ResponseHeadersFilter implements Filter {
+
+  private final ConfigurationService configurationService;
+
+  public ResponseHeadersFilter(final ConfigurationService configurationService) {
+    if (configurationService == null) {
```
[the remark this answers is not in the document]
buccarel: I think you're right, yeah.
Thanks

optimize/backend/src/main/java/io/camunda/optimize/tomcat/ResponseHeadersFilter.java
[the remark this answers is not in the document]
buccarel: I don't like using "var" very much, I was trying to be quick here but then I forgot to change it.
Thanks for noticing

.github/workflows/optimize-e2e-test-cloud.out
buccarel: I think we can create an issue for this and resolve it in another PR, or even fix it when we decide to upgrade chrome?

optimize/backend/src/main/java/io/camunda/optimize/rest/security/cloud/CCSaaSSecurityConfigurerAdapter.java
RomanJRW: These are non-authenticated parts of the API. There are also parts of our API [here](https://docs.camunda.io/optimize/next/apis-tools/optimize-api/overview/), like external variable ingestion.

I am ok with this being pushed to QA to validate, I just want to ensure that it is not missed

matthewBearCamunda (approved):
🎉

PR #24436, opened 2024-11-06
Migrate compensation boundary event subscriptions

### LINKED ISSUE #24377: Migrate compensation boundary event subscriptions
## Description
It should be possible to migrate compensation boundary event subscriptions. Mapping instructions should be provided for the following:
- Elements that have a compensation boundary event attached
- Subprocesses that are enclosing the compensation boundary event (will be clarified in later sections)

Note: Please ignore arrows between boundary events in the images below.

### Migrate compensation event subscription
Consider the following case where the instance waiting on service task `B`. To migrate service  task `B` to `D` and preserve the compensation event subscription after the migration, following mapping instructions should be provided:
- `B -> D`
- `A -> C`

Screenshot 2024-11-05 at 14 38 26

### Migrate compensation event subscription enclosed by subprocess
Consider the following case where the instance waiting on service task `B`. To migrate service  task `B` to `D` and preserve the compensation event subscription after the migration, following mapping instructions should be provided:
- `B -> D`
- `A -> C`
- `subProcess1 -> subProcess2`

Screenshot 2024-11-08 at 12 35 15

### Delete compensation event subscription
Consider the same case above where the instance waiting on service task `B`. But this time we **do not** need to preserve the compensation subscription and only provide mapping between active elements `B -> D`. After the migration the compensation subscription for `boundary1` will be **deleted**

Screenshot 2024-11-05 at 14 46 12

### Create compensation event subscription
It is important to note that new compensation boundary events contained in the target process definition only take effect for activity instances that are not started or not finished yet. We will keep the behaviour aligned with C7 as well (see [C7 docs](https://docs.camunda.org/manual/7.22/user-guide/process-engine/process-instance-migration/#adding-compensation-events)). For example, consider the following example where the instance waiting on service task `B`. If the process instance is migrated (by providing mapping instruction between `B` -> `D`), then triggering compensation throw event afterwards is **not** going to compensate `C`

Screenshot 2024-11-05 at 15 00 31

However, for the following case, a compensation subscription will be created:

### Active elements with compensation boundary event attached

Active elements with compensation boundary event attached can be migrated as usual since compensation event subscriptions are created while completing the element. Consider the following example where the instance waiting on service task `A`. When the process is migrated by providing a mapping instruction between `A -> B`, the compensation subscription for service task `B` will **not** be created during migration. But on completion of the element `B` a compensation subscription will be created.

Screenshot 2024-11-05 at 15 09 37

We will introduce a new `CompensationSubscriptionIntent:MIGRATED` intent to represent the subscription migration.

```[tasklist]
### Tasks
- [ ] Verify the migration works for nested subprocesses with compensation events
```

berkaycanbc:
## Description

Check related issue #24377 and commit messages for details.

Please note that there is no validation for compensation subscription mappings. Those rejection cases will be implemented in a separate PR.

## Related issues

closes #24377

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
@@ -174,6 +174,20 @@ public class ProcessInstanceMigrationCatchEventBehaviour {
           recordCopy.setProcessDefinitionKey(targetProcessDefinition.getKey());
           recordCopy.setCompensableActivityId(targetActivityId);
 
+          // set the compensation handler id if subscription belongs to an activity with a boundary
+          targetProcessDefinition
+              .getProcess()
+              .getElementById(targetActivityId, ExecutableActivity.class)
+              .getBoundaryEvents()
+              .stream()
+              .filter(event -> event.getEventType() == BpmnEventType.COMPENSATION)
+              .findFirst()
+              .ifPresent(
+                  boundary ->
+                      recordCopy.setCompensationHandlerId(
+                          BufferUtil.bufferAsString(
+                              boundary.getCompensation().getCompensationHandler().getId())));
+
           stateWriter.appendFollowUpEvent(
               subscription.getKey(), CompensationSubscriptionIntent.MIGRATED, recordCopy);
         });
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
@@ -40,7 +40,6 @@ import io.camunda.zeebe.protocol.record.intent.ProcessMessageSubscriptionIntent;
 import io.camunda.zeebe.protocol.record.intent.SignalSubscriptionIntent;
 import io.camunda.zeebe.protocol.record.intent.TimerIntent;
 import io.camunda.zeebe.protocol.record.value.BpmnElementType;
-import io.camunda.zeebe.protocol.record.value.BpmnEventType;
 import io.camunda.zeebe.util.buffer.BufferUtil;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -175,18 +174,13 @@ public class ProcessInstanceMigrationCatchEventBehaviour {
           recordCopy.setCompensableActivityId(targetActivityId);
 
           // set the compensation handler id if subscription belongs to an activity with a boundary
-          targetProcessDefinition
-              .getProcess()
-              .getElementById(targetActivityId, ExecutableActivity.class)
-              .getBoundaryEvents()
-              .stream()
-              .filter(event -> event.getEventType() == BpmnEventType.COMPENSATION)
-              .findFirst()
-              .ifPresent(
-                  boundary ->
-                      recordCopy.setCompensationHandlerId(
-                          BufferUtil.bufferAsString(
-                              boundary.getCompensation().getCompensationHandler().getId())));
+          final ExecutableActivity targetElement =
+              targetProcessDefinition
+                  .getProcess()
+                  .getElementById(targetActivityId, ExecutableActivity.class);
+          compensationSubscriptionBehaviour
+              .getCompensationHandlerId(targetElement)
+              .ifPresent(recordCopy::setCompensationHandlerId);
 
           stateWriter.appendFollowUpEvent(
               subscription.getKey(), CompensationSubscriptionIntent.MIGRATED, recordCopy);
@@ -209,8 +203,8 @@ public class ProcessInstanceMigrationCatchEventBehaviour {
                   .getElementById(compensableActivityId, ExecutableActivity.class);
 
           final boolean shouldBeMigrated =
-              shouldCompensationBeMigrated(
-                  sourceElementIdToTargetElementId, compensableActivity, compensableActivityId);
+              sourceElementIdToTargetElementId.containsKey(
+                  BufferUtil.bufferAsString(compensableActivity.getId()));
 
           if (shouldBeMigrated) {
             // We will migrate this mapped catch event, so we don't want to unsubscribe from it.
@@ -227,32 +221,6 @@ public class ProcessInstanceMigrationCatchEventBehaviour {
         });
 
     return compensationSubscriptionsToMigrate;
-  }
-
-  private static boolean shouldCompensationBeMigrated(
-      final Map<String, String> sourceElementIdToTargetElementId,
-      final ExecutableActivity compensableActivity,
-      final String compensableActivityId) {
-    if (compensableActivity.getElementType() == BpmnElementType.SUB_PROCESS) {
-      return sourceElementIdToTargetElementId.containsKey(
-          BufferUtil.bufferAsString(compensableActivity.getId()));
-    } else {
-      // compensable activity is a task
-      final var compensationBoundaryEvent =
-          compensableActivity.getEvents().stream()
-              .filter(event -> event.getEventType() == BpmnEventType.COMPENSATION)
-              .findFirst()
-              .orElseThrow(
-                  () ->
-                      new IllegalStateException(
-                          String.format(
-                              "Expected to find a compensation event for activity '%s' but not found.",
-                              compensableActivityId)));
-      return sourceElementIdToTargetElementId.containsKey(
-              BufferUtil.bufferAsString(compensableActivity.getId()))
-          && sourceElementIdToTargetElementId.containsKey(
-              BufferUtil.bufferAsString(compensationBoundaryEvent.getId()));
-    }
   }
 
   private void migrateSignalEvents(
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
@@ -10,12 +10,15 @@ package io.camunda.zeebe.engine.processing.processinstance;
 import static io.camunda.zeebe.engine.processing.processinstance.ProcessInstanceMigrationPreconditions.requireNoPendingMsgSubMigrationDistribution;
 
 import io.camunda.zeebe.engine.processing.bpmn.BpmnElementContextImpl;
+import io.camunda.zeebe.engine.processing.bpmn.behavior.BpmnCompensationSubscriptionBehaviour;
 import io.camunda.zeebe.engine.processing.common.CatchEventBehavior;
+import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableActivity;
 import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableCatchEventSupplier;
 import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
 import io.camunda.zeebe.engine.processing.processinstance.ProcessInstanceMigrationPreconditions.ProcessInstanceMigrationPreconditionFailedException;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedCommandWriter;
+import io.camunda.zeebe.engine.state.compensation.CompensationSubscription;
 import io.camunda.zeebe.engine.state.deployment.DeployedProcess;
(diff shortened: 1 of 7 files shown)


berkaycanbc:
@nicpuppa I undrafted the PR. I decided to implement mapping instruction validations in a separate PR. 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+                .withIntent(CompensationSubscriptionIntent.MIGRATED)
+                .withProcessInstanceKey(processInstanceKey)
+                .getFirst()
+                .getValue())
+        .describedAs("Expect that the process definition is updated")
+        .hasProcessDefinitionKey(targetProcessDefinitionKey)
+        .describedAs("Expect that the compensable activity is updated")
+        .hasCompensableActivityId("C")
+        .describedAs("Expect that the compensation handler id is unchanged")
+        .hasCompensationHandlerId("undoA");
```
nicpuppa, line 227: ❓ Same here, I'm expecting this Id changed

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+
+    assertThat(
+            RecordingExporter.compensationSubscriptionRecords()
+                .withIntent(CompensationSubscriptionIntent.TRIGGERED)
+                .withProcessInstanceKey(processInstanceKey)
+                .limit(2))
+        .extracting(Record::getValue)
+        .extracting(CompensationSubscriptionRecordValue::getCompensableActivityId)
+        .hasSize(2)
+        .containsExactly("subProcess2", "C");
```
nicpuppa, line 239: ❓ Same comment on the other test case applies here

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+                .getFirst()
+                .getValue())
+        .describedAs("Expect that the process definition is updated")
+        .hasProcessDefinitionKey(sourceProcessDefinitionKey)
+        .describedAs("Expect that the compensable activity is updated")
+        .hasCompensableActivityId("A")
+        .describedAs("Expect that the compensation handler id is unchanged")
+        .hasCompensationHandlerId("undoA")
+        .describedAs("Expect that the compensation boundary event in the source is deleted")
+        .isNotNull();
```
nicpuppa, line 318: 🔧 I'll put this at the beginning of the chain

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+    // then
+    Assertions.assertThat(
+            RecordingExporter.processInstanceRecords(ProcessInstanceIntent.ELEMENT_MIGRATED)
+                .withProcessInstanceKey(processInstanceKey)
+                .withElementType(BpmnElementType.PROCESS)
+                .getFirst())
+        .describedAs("Expect that the process instance is migrated")
+        .isNotNull();
+
+    engine.job().ofInstance(processInstanceKey).withType("A").complete();
```
nicpuppa, line 379: ❓ Should the type of the job be `B` ?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+                .getFirst()
+                .getValue())
+        .describedAs("Expect that the process definition is updated")
+        .hasProcessDefinitionKey(targetProcessDefinitionKey)
+        .describedAs("Expect that the compensable activity is updated")
+        .hasCompensableActivityId("B")
+        .describedAs("Expect that the compensation handler id is unchanged")
+        .hasCompensationHandlerId("undoB")
+        .describedAs("Expect that the compensation boundary event in the source is created")
+        .isNotNull();
```
nicpuppa, line 394: 🔧 I'll move this at the beginning of the chain

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+            RecordingExporter.compensationSubscriptionRecords()
+                .withIntent(CompensationSubscriptionIntent.CREATED)
+                .withProcessInstanceKey(processInstanceKey)
+                .getFirst()
+                .getValue())
+        .describedAs("Expect that the process definition is updated")
+        .hasProcessDefinitionKey(targetProcessDefinitionKey)
+        .describedAs("Expect that the compensable activity is updated")
+        .hasCompensableActivityId("B")
+        .describedAs("Expect that the compensation handler id is unchanged")
```
nicpuppa, line 391: 🔧 The description is not fully correct, what do you mean with this phrase ?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+                .getFirst()
+                .getValue())
+        .describedAs("Expect that the compensation boundary event in the target is completed")
+        .hasProcessDefinitionKey(targetProcessDefinitionKey)
+        .describedAs("Expect that the compensable activity is updated")
+        .hasCompensableActivityId("B")
+        .describedAs("Expect that the compensation handler id is unchanged")
+        .hasCompensationHandlerId("undoB")
+        .describedAs("Expect that the compensation boundary event in the source is created")
+        .isNotNull();
```
nicpuppa, line 411: 🔧 I'll move this at the beginning of the chain

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
```diff
@@ -125,6 +144,103 @@ public void handleCatchEvents(
...
+    compensationSubscriptionsToMigrate.forEach(
+        subscription -> {
+          final var sourceActivityId = subscription.getRecord().getCompensableActivityId();
+          final var targetActivityId = sourceElementIdToTargetElementId.get(sourceActivityId);
+
+          final var compensationSubscriptionRecord = subscription.getRecord();
+          final var recordCopy = new CompensationSubscriptionRecord();
+          recordCopy.wrap(compensationSubscriptionRecord);
+          recordCopy.setProcessDefinitionKey(targetProcessDefinition.getKey());
+          recordCopy.setCompensableActivityId(targetActivityId);
```
nicpuppa, line 175: ❓ I think we are missing some fields to set here, like the `processInstanceKey` or `compensationHandlerId` or `compensableActivityInstanceKey`

nicpuppa (commented):
Hey @berkaycanbc I had a quick review and left some comments, please have a look 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
```diff
@@ -0,0 +1,520 @@
...
+                .withIntent(CompensationSubscriptionIntent.MIGRATED)
+                .withProcessInstanceKey(processInstanceKey)
+                .getFirst()
+                .getValue())
+        .describedAs("Expect that the process definition is updated")
+        .hasProcessDefinitionKey(targetProcessDefinitionKey)
+        .describedAs("Expect that the compensable activity is updated")
+        .hasCompensableActivityId("C")
+        .describedAs("Expect that the compensation handler id is unchanged")
+        .hasCompensationHandlerId("undoA");
```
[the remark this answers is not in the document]
berkaycanbc: Yes, we should. 🚀 Updated it, please take a look.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
berkaycanbc: In other migration tests as well, we first make sure the overall migration was successful. However, I think, we can completely remove the assertion since the `migrate()` already awaits MIGRATED event to be written. Updated.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
berkaycanbc: During migration, we do not evaluate expressions yet. It is expected to be implemented in the later phases of the epic. Therefore, job type doesn't change. See [docs](https://docs.camunda.io/docs/components/concepts/process-instance-migration/#jobs-expressions-and-input-mappings).

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
berkaycanbc: It was all wrong due to copy paste from above methods. Updated them.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
berkaycanbc: It is unnecessary, completely removed it.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
berkaycanbc: It is unnecessary, completely removed it.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
berkaycanbc: `processInstanceKey` and `compensableActivityInstanceKey` should stay the same since they are related to the process **instance** (not the process definition) and process instance related data does not change during the migration. We only update process definition related data (ids, definition key etc.). However, `compensationHandlerId` is part of the definition and needs to be updated. I already send the change for that one.

Please let me know if you need more clarification around process instance migration internals.

berkaycanbc:
@nicpuppa thanks for the review. 🚀 I sent updates, please take look.

Also, I sent the last commit to remove a need for mapping compensation boundary events themselves as they are not adding any value to the internal implementation as well as to users.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCompensationSubscriptionTest.java
nicpuppa: Ok sorry for missed this 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationCatchEventBehaviour.java
```diff
@@ -174,6 +174,20 @@ private void migrateCompensationEvents(
           recordCopy.setProcessDefinitionKey(targetProcessDefinition.getKey());
           recordCopy.setCompensableActivityId(targetActivityId);
 
+          // set the compensation handler id if subscription belongs to an activity with a boundary
+          targetProcessDefinition
```
nicpuppa, line 178: 🔧 There is a method in the CompensationSubscriptionBehaviour that return the handlerId given an `ExecutableActivity`. You can reuse it

nicpuppa (approved):
Thanks for the changes and the clarifications, I just left a suggestion please have a look 👀

PR #24458, opened 2024-11-07
feat: add getProcessInstancesNextBatch to OS repository

npepinpe:
## Description

This PR implements `getProcessInstancesNextBatch` for the OS repository. The main difference is that with the OS client, we don't submit an asynchronous search task, but instead it's just the normal search API via the asynchronous client, so I had to adapt our code a little from the ES repository.

Other than that, it's pretty much the same.

## Related issues

related to #24094


rodrigo-lourenco-lopes (approved):
Looks good to me 👍

PR #24463, opened 2024-11-07
process migration service

### LINKED ISSUE #23814: Implement tasklist data migration
Goal:
When updating from 8.6 to 8.7 data from existing `tasklist-process` index must be migrated to `operate-process` index. This is done by a migration app which is a separate application that will be deployed along with the core C8 components. Currently, there is already a migration app for identity `io.camunda.identity.migration.MigrationRunner`. Our goal is to add the tasklist migration logic to the same app, so that we have only one migration app. 

- [ ] Move migration app out of identity to a common module
- [ ] Allow MigrationRunner to run identity and tasklist data migration in parallel. 
    * Tasklist migration could be just a no-op at this point
    * MigrationRunner should run for ever until the application is stopped
- [ ] Add tasklist data migration logic
    * Support ElasticSearch
    * Migration of on process from `tasklist-process` is as follows
        * Migrate
            * tasklist-process.formId -> operate-process.formId
            * Set operate-process.isPublic from the process definition XML (see [ref](https://docs.google.com/document/d/1Y12848yrn3lcSFjHrgIFuhMJXrZxRxbR-Oec43uchoA/edit#bookmark=id.hqotxwa4weo))
        * Delete the process from `tasklist-process`
            * This is done so that if the migration app is restarted, it doesn't migrate already migrated processes again. 
- [ ] Allow configuring the source and target index names/prefix
        * This should be configurable because the users might have configured index prefix in the existing deployments. 
- [ ] Support OpenSearch
- [ ] Add tests 
    - [ ]  verify that existing data is migrated
    - [ ]  verify that newly imported data is also migrated

panagiotisgts:
## Description

Draft PR for the implementation of the Migration service. The migration process is now solely based on the `operate-process` and `operate-migration-repository` indices to do updates in-place. The steps of the migration process are as follows:

- Lookup `operate-migration-repository` index for an 8.7.0-1 record containing a Process Definition ID to continue where it left off, else start from the begining
- Fetch a batch of records from `operate-process` with ID lexicographical sorted greater than the one retrieved
- Extract data from the BPMN XML regarding Forms in _Start Events_
- Update the respective records with the extracted data
- Write in `operate-migration-repository` the last successful migrated Process Definition ID
- Repeat until no record from `operate-process` index is left

ToDos:
- [x] Changes regarding embedded forms
- [x] Identity migration service
- [x] 8.6 Importer finished signal
- [ ] Artifact and docker builds

> Since the migration involves only processes, all references to Tasklist have been removed. The service will now be split in the `process-migration` and next to come `identity-migration` modules.

@maryarm Tagging you for visibility for your side of migration implementation. 
## Checklist

- [ ] for CI changes:

## Related issues

closes #23814


migration/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+  </parent>
+
+  <artifactId>migration</artifactId>
+  <packaging>pom</packaging>
+  <name>Migration Parent</name>
+
+  <modules>
+    <module>migration-api</module>
+    <module>tasklist-migration</module>
+    <module>webapp</module>
```
Zelldon, line 25: ❓ Do we need a webapp?

migration/tasklist-migration/src/main/java/io/camunda/migration/tasklist/TasklistMigrationProperties.java
```diff
@@ -0,0 +1,64 @@
...
+    return batchSize;
+  }
+
+  public void setBatchSize(final int batchSize) {
+    this.batchSize = batchSize;
+  }
+
+  public static class IndexConfig {
+    private String sourcePrefix;
+    private String targetPrefix;
```
Zelldon, line 46: 💭 I think one prefix should be enough. We expect users to migrate to one global prefix (either none) or one, but including always component names.

migration/tasklist-migration/src/main/java/io/camunda/migration/tasklist/util/IndexUtil.java
```diff
@@ -0,0 +1,43 @@
...
+                && !properties.getIndex().getTargetPrefix().isEmpty()
+            ? properties.getIndex().getTargetPrefix()
+            : "";
+    /* Currently issue appending prefix twice
+      : OPERATE_PREFIX;
+    */
+    return new io.camunda.webapps.schema.descriptors.operate.index.ProcessIndex(indexPrefix, true)
+        .getFullQualifiedName();
+  }
+}
```
Zelldon, line 43: ❌  See above. I think we should expect one global prefix, that is either set or not. Maybe we can reuse the logic we have in our other web apps-schema module

migration/tasklist-migration/src/main/java/io/camunda/migration/tasklist/util/XMLUtil.java
```diff
@@ -0,0 +1,132 @@
...
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.parsers.SAXParserFactory;
+import org.camunda.bpm.model.xml.ModelException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+import org.xml.sax.helpers.DefaultHandler;
+
+public class XMLUtil {
```
Zelldon, line 28: ❌ I feel we duplicating this class quite often. Can we reuse the other classes?

migration/tasklist-migration/src/test/java/io/camunda/migration/tasklist/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,245 @@
...
+            .setVersion(1)
+            .setBpmnXml(TestData.xmlWithoutForm())
+            .setBpmnProcessId("formProcess"),
+        true);
+    awaitRecordsArePresent();
+    // then
+    migrator.run();
+
+    // verify
+    assertTargetRecordsAreMigrated();
```
Zelldon, line 124: 🔧 I would inline this method. It makes it easier to follow - to directly see what is asserted in the test.

migration/tasklist-migration/src/test/java/io/camunda/migration/tasklist/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,245 @@
...
+            .setBpmnXml(TestData.xmlWithForm())
+            .setBpmnProcessId("formProcess"),
+        true);
+    createEntity(
+        new ProcessEntity()
+            .setId("2")
+            .setVersion(1)
+            .setBpmnXml(TestData.xmlWithoutForm())
+            .setBpmnProcessId("formProcess"),
+        false);
```
Zelldon, line 171: 🔧 Reading this it is not clear to me what these flags mean. 

I would propose that we have two methods one `createEntity` and one `createEntityWithDuplicate`

migration/tasklist-migration/src/test/java/io/camunda/migration/tasklist/IndexUtilTest.java
```diff
@@ -0,0 +1,62 @@
...
+        "%s-%s-%s_"
+            .formatted(
+                "operate",
+                io.camunda.webapps.schema.descriptors.operate.index.ProcessIndex.INDEX_NAME,
+                "8.3.0"),
+        targetIndexName);
+  }
+
+  @Test
+  public void sourceAndTargetIndexWithPrefix() {
```
Zelldon, line 40: ❌ This should be adjusted when following the advise from above.

migration/tasklist-migration/src/test/java/io/camunda/migration/tasklist/TestData.java
```diff
@@ -0,0 +1,234 @@
...
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migration.tasklist;
+
+public interface TestData {
+
+  static String tasklistIndex() {
```
Zelldon, line 12: ❓ Could we use some existing schema? Feels brittle to specify own schema here.

migration/webapp/src/main/java/io/camunda/migration/webapp/RuntimeContainer.java
```diff
@@ -0,0 +1,48 @@
...
+  public void start() throws InterruptedException {
+    final CountDownLatch latch = new CountDownLatch(migrators.size());
+    migrators.forEach(
+        m ->
+            executor.submit(
+                () -> {
+                  try {
+                    m.run();
+                  } catch (final MigrationException ex) {
+                    LOG.error(ex.getMessage());
```
Zelldon, line 41:
```suggestion
                    LOG.error(ex.getMessage(), ex);
```

Zelldon: Can we retry some exceptions?

migration/webapp/src/main/java/io/camunda/migration/webapp/MigrationApp.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.CommandLineRunner;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.WebApplicationType;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Profile;
+
+@SpringBootApplication(scanBasePackages = {"io.camunda.migration"})
+@Profile("!test")
+public class MigrationApp implements CommandLineRunner {
```
Zelldon, line 21: ❓ Actually I thought Identity Team already set up a skeleton app. I guess we should check with them.

ChrisKujawa (changes requested):
Thanks @panagiotisgts as discussed let's go the route of not deleting records/data. 

We can keep track of where we left of in the import-position (we will discuss this in detail today again).

With that we keep data (useful if something goes wrong), allows manual intervention and healing. The user can drop the index, if everything went well.

migration/tasklist-migration/src/main/java/io/camunda/migration/tasklist/util/IndexUtil.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public static String getTargetIndexName(final TasklistMigrationProperties properties) {
+    final String indexPrefix =
+        properties.getIndex() != null
+                && properties.getIndex().getTargetPrefix() != null
+                && !properties.getIndex().getTargetPrefix().isEmpty()
+            ? properties.getIndex().getTargetPrefix()
+            : "";
+    /* Currently issue appending prefix twice
+      : OPERATE_PREFIX;
```
Zelldon: :x:This is fixed

migration/tasklist-migration/src/main/java/io/camunda/migration/tasklist/util/IndexUtil.java
```diff
@@ -0,0 +1,44 @@
...
+public class IndexUtil {
+  private static final String INDEX_FORMAT = "%s-%s-%s_";
+  private static final String TASKLIST_PREFIX = "tasklist";
+  private static final String OPERATE_PREFIX = "operate";
+
+  /** Returns the source index name based on the configuration. */
+  public static String getSourceIndexName(final TasklistMigrationProperties properties) {
+    return properties.getIndex() != null
+            && properties.getIndex().getSourcePrefix() != null
+            && !properties.getIndex().getSourcePrefix().isEmpty()
```
Zelldon: 🔧 I think we discussed this, we don;t need to provide multiple index prefixes.

migration/webapp/src/main/java/io/camunda/migration/webapp/RuntimeContainer.java
```diff
@@ -0,0 +1,39 @@
...
+import io.camunda.migration.api.Migrator;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadPoolExecutor;
+import org.springframework.stereotype.Component;
+
+@Component
+public class RuntimeContainer {
+
```
Zelldon: 🔧 Would here be the place where we set the profile?

migration/tasklist-migration/src/main/java/io/camunda/migration/tasklist/adapter/os/OpensearchAdapter.java
```diff
@@ -0,0 +1,118 @@
...
+    entities.forEach(e -> migrateEntity(e, bulkRequest));
+
+    try {
+      final BulkResponse response = client.bulk(bulkRequest.build());
+      final List<String> idsToBeDeleted = getMigratedIds(response);
+
+      if (idsToBeDeleted.isEmpty()) {
+        return !response.errors();
+      }
+      final BulkRequest deleteRequest = buildDeleteRequest(idsToBeDeleted);
```
Zelldon: 🤔 I thought we dont want to delete anymore?

Zelldon:
@panagiotisgts did you made changes on this PR after our discussion or do you have a separate PR?

Zelldon:
Looks like Github was tricking me 🤔 I saw old code somehow

migration/process-migration/src/test/java/io/camunda/migration/process/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,345 @@
...
+    awaitRecordsArePresent(ProcessEntity.class, processIndex.getFullQualifiedName());
+
+    // then
+    migrator.migrateBatch(List.of(entityToBeMigrated));
+    awaitRecordsArePresent(ProcessorStep.class, migrationRepositoryIndex.getFullQualifiedName());
+    esClient.indices().refresh();
+    // verify
+    assertProcessorStepContentIsStored("1");
+    final var processorRecords =
+        readRecords(ProcessorStep.class, migrationRepositoryIndex.getFullQualifiedName());
```
Zelldon, line 117:
```suggestion
    // given
    properties.setBatchSize(1);
    final ProcessEntity entityToBeMigrated = TestData.processEntityWithPublicFormId(1L);
    final ProcessEntity entityNotToBeMigrated = TestData.processEntityWithPublicFormId(2L);
    writeProcessToIndex(entityToBeMigrated);
    writeProcessToIndex(entityNotToBeMigrated);
    awaitRecordsArePresent(ProcessEntity.class, processIndex.getFullQualifiedName());

    // when
    migrator.migrateBatch(List.of(entityToBeMigrated));
    awaitRecordsArePresent(ProcessorStep.class, migrationRepositoryIndex.getFullQualifiedName());
    esClient.indices().refresh();

    // then
    assertProcessorStepContentIsStored("1");
    final var processorRecords =
        readRecords(ProcessorStep.class, migrationRepositoryIndex.getFullQualifiedName());
```

migration/process-migration/src/test/java/io/camunda/migration/process/ProcessModelUtilTest.java
```diff
@@ -0,0 +1,82 @@
...
+            entity.getBpmnXml().getBytes(), entity.getBpmnProcessId());
+    assertThat(startEvent).isPresent();
+    final var isPublic = ProcessModelUtil.extractIsPublic(startEvent.get());
+
+    assertThat(isPublic).isPresent();
+    assertThat(isPublic.get()).isTrue();
+  }
+
+  @Test
+  void extractedDataShouldBeEmpty() {
```
Zelldon, line 66: ❌ lets use always the same nameing pattern for consistency, start with `should..`

migration/process-migration/src/test/java/io/camunda/migration/process/ElasticsearchMigrationRunnerIT.java
Zelldon: The sections are always:

* Given
* When
* Then

:)

migration/process-migration/src/test/java/io/camunda/migration/process/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,345 @@
...
+    assertThat(records.stream().filter(r -> r.getKey() == 3).findFirst().get().getFormKey())
+        .isEqualTo("camunda-forms:bpmn:testForm");
+    assertThat(records.stream().filter(r -> r.getKey() == 3).findFirst().get().getFormId())
+        .isNull();
+    assertThat(records.stream().filter(r -> r.getKey() == 3).findFirst().get().getIsFormEmbedded())
+        .isTrue();
+  }
+
+  @Test
+  public void migrationShouldCompleteWithMultipleRounds() throws IOException {
```
Zelldon, line 200:
```suggestion
  public void shouldCompleteMigrationWithMultipleRounds() throws IOException {
```

migration/process-migration/src/test/java/io/camunda/migration/process/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,345 @@
...
+    assertProcessorStepContentIsStored("9");
+    assertThat(records.size()).isEqualTo(20);
+    assertThat(records.stream().noneMatch(r -> r.getIsPublic().equals(Boolean.FALSE))).isTrue();
+    assertThat(records.stream().noneMatch(r -> r.getFormId() == null)).isTrue();
+    assertThat(records.stream().allMatch(r -> r.getFormKey() == null)).isTrue();
+    assertThat(records.stream().noneMatch(ProcessEntity::getIsFormEmbedded)).isTrue();
+  }
+
+  @Test
+  public void migrationShouldPickUpFromStoredId() throws IOException {
```
Zelldon, line 223:
```suggestion
  public void shouldMigrationPickUpFromStoredId() throws IOException {
```

migration/process-migration/src/test/java/io/camunda/migration/process/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,345 @@
...
+        .isTrue();
+    assertThat(
+            records.stream()
+                .filter(r -> r.getKey() > 5)
+                .allMatch(r -> r.getIsFormEmbedded().equals(Boolean.FALSE)))
+        .isTrue();
+  }
+
+  @Test
+  public void migrationShouldDoNothingWhenFinalStepIsPresent() throws IOException {
```
Zelldon, line 272: Same here

migration/process-migration/src/test/java/io/camunda/migration/process/TestData.java
```diff
@@ -0,0 +1,184 @@
...
+      </bpmndi:BPMNEdge>
+    </bpmndi:BPMNPlane>
+  </bpmndi:BPMNDiagram>
+</bpmn:definitions>
+  """;
+  }
+
+  static String bpmnProcessWithFormIdReference() {
+    return """
+<?xml version="1.0" encoding="UTF-8"?>
```
Zelldon, line 93: ❓ Why not creating these as resource files?

migration/process-migration/src/test/java/io/camunda/migration/process/TestData.java
```diff
@@ -0,0 +1,184 @@
...
+        .setId(String.valueOf(id))
+        .setKey(id)
+        .setVersion(1)
+        .setBpmnXml(bpmnProcessWithoutForm())
+        .setBpmnProcessId("formProcess");
+  }
+
+  static String bpmnProcessWithFormKeyReference() {
+    return """
+<?xml version="1.0" encoding="UTF-8"?>
```
Zelldon, line 43: BTW for simple processes you can also use the BPMN model api

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/Adapter.java
```diff
@@ -0,0 +1,88 @@
...
+          Thread.sleep(backoff);
+        } catch (final InterruptedException ie) {
+          Thread.currentThread().interrupt();
+          throw new MigrationException("Retry interrupted", ie);
+        }
+        backoff *= 2;
+      }
+    }
+    throw new MigrationException("Operation failed");
+  }
```
Zelldon, line 87: ❌  I think we have in our util module several retry strategies maybe you can use some of these

migration/migration-app/src/main/java/io/camunda/migration/app/RuntimeContainer.java
```diff
@@ -0,0 +1,49 @@
...
+  public void start(final ApplicationArguments args) throws InterruptedException {
+    final CountDownLatch latch = new CountDownLatch(migrators.size());
+    migrators.forEach(
+        m ->
+            executor.submit(
+                () -> {
+                  try {
+                    m.run(args);
+                  } catch (final MigrationException ex) {
+                    LOG.error(ex.getMessage());
```
Zelldon, line 42: 🤔 I guess this would mean that if one migratory fails, it will not retry right? as we wait for the others to complete

ChrisKujawa (commented):
In general looks already quite good 👍🏼 I have some comments.

What we still need is I think to integrate it in the single application, and then a standalone application (like we have now). Meaning I guess integrating in dist?

I think what we can do is to split this PR up, add first the business logic with tests, add application, add it to the dist. Order we might need to adjust not sure here.

dist/src/main/java/io/camunda/application/MigrationApplication.java
```diff
@@ -0,0 +1,129 @@
...
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Profile;
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.Profiles;
+
+@SpringBootApplication
+@Profile({"migration", "identity-migration | process-migration"})
```
Zelldon, line 39: ❓ Just for my understanding (call me spring noob): Does this mean 

A) `migration` AND one of `identity-migration | process-migration` has to be specified 
or 
B) any of the profiles will enable this

dist/src/main/java/io/camunda/application/MigrationApplication.java
```diff
@@ -0,0 +1,129 @@
...
+public class MigrationApplication implements ApplicationRunner {
```
Zelldon, line 40: 🔧 I think we have normally the `Standalone..` naming pattern.

dist/src/main/java/io/camunda/application/MigrationApplication.java
Zelldon: If this is the Standalone application I guess we shouldn't have the profiles here.

migration/process-migration/src/test/resources/migration/process/process-without-form-reference.bpmn
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Zelldon, line 1: 👍🏼

migration/process-migration/src/test/java/io/camunda/migration/process/ElasticsearchMigrationRunnerIT.java
```diff
@@ -269,20 +272,21 @@ public void migrationShouldPickUpFromStoredId() throws IOException {
   }
 
   @Test
-  public void migrationShouldDoNothingWhenFinalStepIsPresent() throws IOException {
-    // when
+  public void shouldNotMigrateWhenFinalStepIsPresent() throws IOException {
+    // given
```
Zelldon, line 275: 👍🏼

dist/src/main/java/io/camunda/application/MigrationApplication.java
Zelldon: I think we should do it similar to what we have with the other Standalone apps for example look at the Broker

* We have[ a StandaloneBroker](https://github.com/camunda/camunda/blob/main/dist/src/main/java/io/camunda/application/StandaloneBroker.java#L15-L18) - with implementing the main etc
* We have[ a Broker module configuration - that is enabled by profile](https://github.com/camunda/camunda/blob/main/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java#L36-L43)


Similar we have for gateway, operate,tasklist, etc.

Does this make sense @panagiotisgts ?

panagiotisgts: - It should be one of the three, forgot to remove the `|` operator 
- - The `migration` profile, activates the other two as well
- - If none of these are present it will not run

Yeah, it makes sense, will fix this one.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/ConditionalOnRestGatewayEnabled.java
```diff
@@ -23,7 +22,6 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Documented
-@ConditionalOnWebApplication
```
panagiotisgts, line 26: @Zelldon @nicpuppa Not sure whether this is 100% correct, but these configurations cannot be enabled when running non-webapp spring configurations. In case of the migration service, when running as a Standalone app, we do not need the whole web env setup.

dist/src/main/java/io/camunda/application/commons/migration/MigrationsModuleConfiguration.java
```diff
@@ -0,0 +1,86 @@
...
+  @PostConstruct
+  public void migrate() {
+    try (final ScheduledExecutorService scheduledExecutorService =
+        Executors.newScheduledThreadPool(1)) {
+      scheduledExecutorService.schedule(
+          () -> {
+            performMigration();
+            LOG.info("Migration finished, shutting down");
+          },
+          0,
```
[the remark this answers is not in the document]
panagiotisgts: When the migration profile is run embedded on other configurations, we should run it in parallel and not block until it's completed. Since this is run on `@PostCostrunct` to be automatically run, it would otherwise block until it was finished.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/ConditionalOnRestGatewayEnabled.java
panagiotisgts: It's required to enable the _CamundaServices_ [configurations](https://github.com/camunda/camunda/blob/5709d730877b5e35cf8b3a0764f18ce1c4239416/dist/src/main/java/io/camunda/application/commons/service/CamundaServicesConfiguration.java#L150) required for Identity migration side at this moment at least. If they follow another strategy we can skip enabling these configuration beans.

dist/src/main/java/io/camunda/application/commons/migration/MigrationsModuleConfiguration.java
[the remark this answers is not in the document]
Zelldon: I see, could we maybe reuse the same Executor we create then also in the other method? Or simply do Thread.run? 🤔

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/ConditionalOnRestGatewayEnabled.java
Zelldon: I see thanks, but I guess removing this might break also some parts 🤔 Or did it work?

Wondering how this works for the Standalone gateway 🤔

panagiotisgts: It did work, and seems like it does not break anything :angel:  but not 100% sure that's why I asked for @nicpuppa 's input here.

AFAIK the properties inside the `@ConditionalOnProperty` are _AND_ bound and also the link between the two annotations is also _AND_.

nicpuppa: I'm not sure why I got pinged here @panagiotisgts. I think you mistakenly tag me instead @npepinpe (?)

panagiotisgts: > I'm not sure why I got pinged here ...

Not once but twice... 

>Hm this would mean we cant run a gateway alone? As it is not a webapp? 🤔

It's run with the _Servlet_ application type.

npepinpe: So the issue is that what, the migration app is not a web app (doesn't start a server), but we need some of the configurations annotated by `@ConditionalOnRestGatewayEnabled` in the migration app itself? It sounds like these should be pulled out with different conditions then, as they must now be shared?

I think I'm missing context here. I added this before we had these kind of requirements, so if you see a need to change things go ahead. Or did you have a question on how it works?

panagiotisgts:
Closing since it will be split into smaller chunks.

PR #24473, opened 2024-11-07
feat: Update authorization implementation for Operate (internal api)

### LINKED ISSUE #23352: Update authorization implementation for Operate internal endpoints
## Overview

Operate uses `PermissionService` to check permissions on `process definitions` and `decision definitions`. `PermissionService` uses Identity SDK for retrieving authorizations.
In current implementation all authorizations are retrieved and then filtered by application logic.

From 8.7 on the Identity SDK is not available anymore. 
`PermissionService` needs to retrieve authorizations from `AuthorizationService`.

Retrieving a list of `AuthorizationEntity` can be done for example:
```java
 final List<AuthorizationEntity> authorizationEntities =
        authorizationServices
            .search(
                AuthorizationQuery.of(
                    q ->
                        q.filter(
                            f ->
                                f.resourceKey(processDefinitionId)
                                    .resourceType(RESOURCE_TYPE_PROCESS_DEFINITION))))
            .items();
```

### Notes from PMs:

- The Operate V1 Public API should work as it did for customers, including permissions. We don’t want to introduce any breaking changes with the V1 API. If no permission was used in an endpoint, we must not introduce it.
- For Operate internal API - old permissions should be mapped to new permissions so that nothing changes for customers.
- As a user, if I had “write” permissions per process definition, I should still have “write” permissions per process definition with 8.7
- As a user, if I had “read” permissions per process definition, I should still have “read” permissions per process definition with 8.7
- As a user, if I had “read+write” permissions per process definition, I should still have “read+write” permissions per process definition with 8.7

### Components for permissions:

- [Permission](https://github.com/camunda/camunda/blob/a1ed2210f98726ce8a1e8efac9b39d68e0471d86/security/security-core/src/main/java/io/camunda/security/entity/Permission.java#L13-L13)
- [PermissionType](https://github.com/camunda/camunda/blob/a1ed2210f98726ce8a1e8efac9b39d68e0471d86/security/security-core/src/main/java/io/camunda/zeebe/protocol/record/value/PermissionType.java#L10-L10)
- [AuthorizationEntity](https://github.com/camunda/camunda/blob/a1ed2210f98726ce8a1e8efac9b39d68e0471d86/search/search-domain/src/main/java/io/camunda/search/entities/AuthorizationEntity.java#L16-L16)
- [AuthorizationServices](https://github.com/camunda/camunda/blob/a1ed2210f98726ce8a1e8efac9b39d68e0471d86/service/src/main/java/io/camunda/service/AuthorizationServices.java#L32-L32)

## Proposal

- `PermissionService` implements its methods in terms of components mentioned above
- No usage of Identity SDK (Identity object as main entry) anymore
- The permissions shouldn't be stored in session or some other cache-like objects. They need to be retrieved in every call.
- The retrieval should be specific as possible. Use the query filter.
- Replace `IdentityPermission` with `PermissionType` 
- v1 API checks also permissions

## Tasks

Adjust types and mapping of types:
- [ ] Adjust permission types or map them to the ones that are used in Operate
     - [ ] IdentityPermission -> PermissionType 

The following queries needs to be rewritten:
- [ ] Retrieve process definition permissions by process definition id (bpmnProcessId)      
- [ ] Retrieve decision definition permissions by decision id
- [ ] Retrieve process definitions keys by permission 
- [ ] Retrieve decision definitions keys by permission 
- [ ] Has permission for process definition id
- [ ] Has permission for decision id

Add missing checks:
- [ ] Add permissions check in v1 API

ghost:
## Description

feat: Update authorization implementation for Operate (internal api)

- updated `PermissionsService` to use the new authorization implementation
- this PR only covers the internal API, since public API v1 does not use permissions

Closes https://github.com/camunda/camunda/issues/23352


operate/webapp/pom.xml
```diff
@@ -76,6 +76,19 @@
...
+      <artifactId>camunda-security-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>camunda-security-services</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-gateway-rest</artifactId>
+    </dependency>
```
houssain-barouni, line 90: You added this dependency to use `TenantAttributeHolder` in operate-webapp, 
can we rather move it to a shared dependency like `camunda-security-services` ?

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,287 @@
...
+   */
+  public Set<String> getResourcePermissions(
+      final String resourceId, final AuthorizationResourceType resourceType) {
+    final Set<String> permissions = new HashSet<>();
+    if (permissionsEnabled() && isAuthenticated()) {
+      for (PermissionType permissionType : PermissionType.values()) {
+        if (isAuthorized(resourceId, resourceType, permissionType)) {
+          permissions.add(permissionType.name());
+        }
+      }
```
houssain-barouni, line 87: ⚠️ This loop will currently trigger 14 queries to database
Why not adding a new method in AuthorizationChecker that does the same in one query?

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,287 @@
...
+      final TenantsService tenantsService) {
+    this.operateProperties = operateProperties;
+    this.authorizationChecker = authorizationChecker;
+    this.tenantsService = tenantsService;
+  }
+
+  @PostConstruct
+  public void logCreated() {
+    LOGGER.debug("PermissionsService bean created.");
+  }
```
houssain-barouni, line 50: This log is not needed anymore as `PermissionsService` bean is not conditional anymore

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+    return null;
+  }
+
+  private boolean isAuthenticated() {
+    return (getAuthenticatedUserKey() != null);
+  }
+
+  public boolean permissionsEnabled() {
+    return operateProperties.getIdentity().isResourcePermissionsEnabled();
```
houssain-barouni, line 222: we should use the new common properties cf https://github.com/camunda/camunda/blob/main/security/security-core/src/main/java/io/camunda/security/configuration/SecurityConfiguration.java
You can autowire a `SecurityConfiguration` bean

houssain-barouni (changes requested):
This needs few adjustments.
Maybe we should have an integration test that uses the new Camunda exporter?

operate/webapp/pom.xml
mihail-ca: Why would a security module depend on a rest module ? IMHO rest modules should be top level, nothing depend on them, except other rest modules, like Operate webapp.

houssain-barouni: No. I did not say we should make security module depend on rest module, I said we just need to move `TenantAttributeHolder ` to a shared security module
Maybe to `io.camunda:camunda-authentication` which already used by Operate webapp and also uses Spring

mihail-ca:
> This needs few adjustments. Maybe we should have an integration test that uses the new Camunda exporter?

Do you have an example of such test ? Thank you.

nathansandi:
> > This needs few adjustments. Maybe we should have an integration test that uses the new Camunda exporter?
> 
> Do you have an example of such test ? Thank you.

You can use test standalone camunda class, .withCamundaExporter() method, so you can initialise directly the exporters. User Tasks Controller IT is already using the exporters.

mihail-ca:
> This needs few adjustments. Maybe we should have an integration test that uses the new Camunda exporter?

@houssain-barouni I added the class `OperatePermissionsIT.java` for IT tests, but the tests are not functional, because, as discussed internally, this depends on changes that Identity needs to do, so I will add more tests here once Identity team finishes. FYI @tmetzke

houssain-barouni:
> > This needs few adjustments. Maybe we should have an integration test that uses the new Camunda exporter?
> 
> @houssain-barouni I added the class `OperatePermissionsIT.java` for IT tests, but the tests are not functional, because, as discussed internally, this depends on changes that Identity needs to do, so I will add more tests here once Identity team finishes. FYI @tmetzke

if we have dependencies with identity that are not covered yet (I presume it is related to the spring profiles that we need to use)  we can do it later in another pull request.
Other than that , is the pull request ready to be reviewed again ? @mihail-ca

mihail-ca:
> > > This needs few adjustments. Maybe we should have an integration test that uses the new Camunda exporter?
> > 
> > 
> > @houssain-barouni I added the class `OperatePermissionsIT.java` for IT tests, but the tests are not functional, because, as discussed internally, this depends on changes that Identity needs to do, so I will add more tests here once Identity team finishes. FYI @tmetzke
> 
> if we have dependencies with identity that are not covered yet (I presume it is related to the spring profiles that we need to use) we can do it later in another pull request. Other than that , is the pull request ready to be reviewed again ? @mihail-ca

@houssain-barouni yes, it is ready for re-review. I will update the IT tests as needed, once Identity part is done.

operate/webapp/src/main/java/io/camunda/operate/webapp/rest/ClientConfig.java
```diff
@@ -45,7 +47,7 @@ public String asJson() {
     canLogout = profileService.currentProfileCanLogout();
     isLoginDelegated = profileService.isLoginDelegated();
     tasklistUrl = operateProperties.getTasklistUrl();
-    resourcePermissionsEnabled = operateProperties.getIdentity().isResourcePermissionsEnabled();
```
houssain-barouni, line 48: ❓ do we have a confirmation that the new property should be used?

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsConfigurer.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package io.camunda.operate.webapp.security.permission;
+
+import io.camunda.security.configuration.SecurityConfiguration;
+import io.camunda.security.impl.AuthorizationChecker;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class PermissionsConfigurer {
```
houssain-barouni, line 16:
```suggestion
public class SecurityConfigurer {
```
As this config class also creates `TenantsService` bean

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,284 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+public class PermissionsService {
+
+  public static final String RESOURCE_KEY_ALL = "*";
+  public static final String RESOURCE_TYPE_PROCESS_DEFINITION = "process-definition";
+  public static final String RESOURCE_TYPE_DECISION_DEFINITION = "decision-definition";
```
houssain-barouni, line 29: 🔧  These constants are not used anymore. We can delete them

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,284 @@
...
+    return hasPermissionForResource(
+        decisionId, AuthorizationResourceType.DECISION_DEFINITION, identityPermission);
+  }
+
+  /**
+   * hasPermissionForResource
+   *
+   * @return true if the user has the given permission for the resource
+   */
+  public boolean hasPermissionForResource(
```
houssain-barouni, line 113: 🔧 this method is only used inside this class, you can make it private

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,284 @@
...
+      final IdentityPermission identityPermission) {
+    if (!permissionsEnabled()) {
+      return true;
+    }
+    if (!isAuthorized()) {
+      return false;
+    }
+    if (identityPermission == null) {
+      throw new IllegalStateException("Identity permission cannot be null");
+    }
```
houssain-barouni, line 125: 🔧 no need for this nullity check, as we completely control the way the method is invoked

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,284 @@
...
+        AuthorizationResourceType.DECISION_DEFINITION, identityPermission);
+  }
+
+  /**
+   * getResourcesWithPermission
+   *
+   * @return resources for which the user has the given permission; the result matches either all
+   *     resources, or a list of resourceIds
+   */
+  public ResourcesAllowed getResourcesWithPermission(
```
houssain-barouni, line 159: same here, this can be private

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  private SecurityContext getSecurityContext(Authorization authorization) {
+    return new SecurityContext(getAuthentication(), authorization);
+  }
+
+  private io.camunda.security.auth.Authentication getAuthentication() {
+    final Long authenticatedUserKey = getAuthenticatedUserKey();
+    final List<String> authorizedTenants = tenantsService.tenantIds();
+    // groups and roles will come later
```
houssain-barouni, line 215: can you create a GH issue for this, so that we don't forget ?

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  /**
+   * @return true if permissions checks are enabled
+   */
+  public boolean permissionsEnabled() {
+    return securityConfiguration.getAuthorizations().isEnabled();
+  }
+
+  private boolean isAuthorized() {
```
houssain-barouni, line 185: `isAuthenticated()` ?
I expect that the call is not authenticated, is detected beforehand. But maybe we miss the OIDC cases handing, isn't?

houssain-barouni (commented):
Globally looks good 👍 
Still some minor comments and questions

operate/webapp/src/main/java/io/camunda/operate/webapp/rest/ClientConfig.java
mihail-ca: Yes, it was confirmed here: https://camunda.slack.com/archives/C06UKS51QV9/p1732288942822749?thread_ts=1732099939.173189&cid=C06UKS51QV9

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsConfigurer.java
mihail-ca: Isn't tenant also used to control permissions after all ?

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
mihail-ca: This method is actually subject to change, after we have the implementation from Identity. I am not sure about how OIDC is supposed to work though.

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
mihail-ca: Added here: https://github.com/camunda/camunda/issues/25282

houssain-barouni (approved):
LGTM! 👍 
@mihail-ca please next time make sure that after a code review, the requested changes are added as a new commit. Otherwise it makes it difficult to do a second review.

Issue #14863, opened 2023-10-26
Exporter templates may get overridden by previous version during rolling update


megglos:
**Describe the bug**

If a zeebe update contains changes to the elastic/opensearch exporter templates, these can temporarily get overridden by the previous version during a rolling update if a pod with the previous version restarts in the process.
If this happens and index requests from a pod with the new version are sent to elastic/opensearch, the index request might fail due to a schema mismatch.

**To Reproduce**
Porbably very hard to time 😅
1. Start a rolling update
2. kill a pod that still runs the previous version
3. when it restarts it will override the templates with the previous state
4. Any index request from a pod with the new version will now fail if the last restarted pod was the previous version

**Expected behavior**

Schema templates should either be versioned, so older templates cannot override newer or they should use matchers that include the version of the schema, so they can't conflict with each other and an index is only created if a template with a matching version is present.

**Log/Stacktrace**

<details><summary>Full Stacktrace</summary>
 <p>

```
2023-10-25 13:42:13.924 [Broker-1] [zb-fs-workers-2] [Exporter-2] WARN 
      io.camunda.zeebe.broker.exporter.elasticsearch - Unexpected exception occurred on periodically flushing bulk, will retry later.
io.camunda.zeebe.exporter.ElasticsearchExporterException: Failed to flush bulk request: [Failed to flush 89 item(s) of bulk request [type: strict_dynamic_mapping_exception, reason: mapping set to strict, dynamic introduction of [sequence] within [_doc] is not allowed]]
	at io.camunda.zeebe.exporter.ElasticsearchClient.throwCollectedBulkError(ElasticsearchClient.java:183) ~[zeebe-elasticsearch-exporter-8.2.13.jar:8.2.13]
	at io.camunda.zeebe.exporter.ElasticsearchClient.exportBulk(ElasticsearchClient.java:166) ~[zeebe-elasticsearch-exporter-8.2.13.jar:8.2.13]
	at io.camunda.zeebe.exporter.ElasticsearchClient.flush(ElasticsearchClient.java:107) ~[zeebe-elasticsearch-exporter-8.2.13.jar:8.2.13]
	at io.camunda.zeebe.exporter.ElasticsearchExporter.flush(ElasticsearchExporter.java:166) ~[zeebe-elasticsearch-exporter-8.2.13.jar:8.2.13]
[...]

PR #24483, opened 2024-11-07
ES/OS Exporter writes to correct versioned indices

### LINKED ISSUE #24316: ES/OS exporter are writing to wrong versioned Zeebe indices
**Describe the bug**

Imagine the following scenario:

We have the following log:

```
         Exporter pos
             ||
             \/                               
[ R 8.6 | R 8.6 | R 8.6 | R 8.6 | R 8.7 | R 8.7 | R 8.7 | R 8.7 | R 8.7 | R 8.7]
```

When doing an update to 8.7, the Engine will create new records with the new version. For simplicity, we just show the version above. 

Let's say the exporter is currently in a position where the Record was written with version 8.6.

When the ES/OS exporter continues now with a record in version 8.6 it will write them into a Zeebe index named 8.7, because of this https://github.com/camunda/camunda/blob/main/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java#L63

This means it will be imported by Operate as a record for version 8.7 https://github.com/camunda/camunda/blob/main/operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportJob.java#L151

### Why is it a problem:

  * it is unexpected when the Broker writes a record with version 8.6 and it is exported as 8.7
  * changing the record format could lead to further problems (luckily we were mostly backwards compatible)
  * **Important:** With the Camunda Exporter project we expect that we get rid of the Importer, and need to make sure that all old data is correctly imported ( Note: we have differentiation between 8.7 and 8.6 Importer modules. We will have no 8.7 Importer component).

**To Reproduce**

Run an update, and see that the exporter writes records directly to new indices.

**Expected behavior**

Records are written into the corresponding indices (where records belong to).

EuroLew:
## Description

If the log has records which were generated in an older version of camunda and the brokers (and thus exporters) update, the old records should write to the old versioned zeebe indices instead of the new ones.
e.g.
- log has 8.6 records
- exporter updates to 8.7
- as exporter runs the 8.6 records get written to the 8.6 zeebe indices. (Currently it writes to the 8.7 zeebe indice)

## Related issues

closes #24316

### DIFF
--- a/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
@@ -164,15 +164,7 @@ class ElasticsearchClient implements AutoCloseable {
    * @return true if request was acknowledged
    */
   public boolean putIndexTemplate(final ValueType valueType) {
-    final String templateName =
-        indexRouter.indexPrefixForValueType(valueType, VersionUtil.getVersionLowerCase());
-    final Template template =
-        templateReader.readIndexTemplate(
-            valueType,
-            indexRouter.searchPatternForValueType(valueType),
-            indexRouter.aliasNameForValueType(valueType));
-
-    return putIndexTemplate(templateName, template);
+    return putIndexTemplate(valueType, VersionUtil.getVersionLowerCase());
   }
 
   public boolean putIndexTemplate(final ValueType valueType, final String version) {
--- a/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
@@ -26,6 +26,7 @@ import io.camunda.zeebe.exporter.dto.PutIndexTemplateResponse;
 import io.camunda.zeebe.exporter.dto.Template;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.util.VersionUtil;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer.Sample;
 import java.io.IOException;
@@ -163,7 +164,8 @@ class ElasticsearchClient implements AutoCloseable {
    * @return true if request was acknowledged
    */
   public boolean putIndexTemplate(final ValueType valueType) {
-    final String templateName = indexRouter.indexPrefixForValueType(valueType);
+    final String templateName =
+        indexRouter.indexPrefixForValueType(valueType, VersionUtil.getVersionLowerCase());
     final Template template =
         templateReader.readIndexTemplate(
             valueType,
--- a/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
@@ -175,6 +175,17 @@ class ElasticsearchClient implements AutoCloseable {
     return putIndexTemplate(templateName, template);
   }
 
+  public boolean putIndexTemplate(final ValueType valueType, final String version) {
+    final String templateName = indexRouter.indexPrefixForValueType(valueType, version);
+    final Template template =
+        templateReader.readIndexTemplate(
+            valueType,
+            indexRouter.searchPatternForValueType(valueType, version),
+            indexRouter.aliasNameForValueType(valueType));
+
+    return putIndexTemplate(templateName, template);
+  }
+
   /**
    * Creates or updates the component template on the target Elasticsearch. The template is read
    * from {@link TemplateReader#readComponentTemplate()}.
--- a/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporter.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporter.java
@@ -23,8 +23,8 @@ import java.io.IOException;
 import java.time.Duration;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import org.slf4j.Logger;
@@ -59,7 +59,7 @@ public class ElasticsearchExporter implements Exporter {
   private MeterRegistry registry;
 
   private long lastPosition = -1;
-  private Map<String, Boolean> indexTemplatesCreated;
+  private Set<String> indexTemplatesCreated;
 
   @Override
   public void configure(final Context context) {
@@ -72,7 +72,7 @@ public class ElasticsearchExporter implements Exporter {
     pluginRepository.load(configuration.getInterceptorPlugins());
 
     context.setFilter(new ElasticsearchRecordFilter(configuration));
-    indexTemplatesCreated = new HashMap<>();
+    indexTemplatesCreated = new HashSet<>();
     registry = context.getMeterRegistry();
   }
 
@@ -120,7 +120,7 @@ public class ElasticsearchExporter implements Exporter {
 
   @Override
   public void export(final Record<?> record) {
-    if (!indexTemplatesCreated.containsKey(record.getBrokerVersion())) {
+    if (!indexTemplatesCreated.contains(record.getBrokerVersion())) {
       createIndexTemplates(record.getBrokerVersion());
 
       updateRetentionPolicyForExistingIndices();
@@ -357,7 +357,7 @@ public class ElasticsearchExporter implements Exporter {
       }
     }
 
-    indexTemplatesCreated.put(version, true);
+    indexTemplatesCreated.add(version);
   }
 
   private void createIndexLifecycleManagementPolicy() {
@@ -370,14 +370,6 @@ public class ElasticsearchExporter implements Exporter {
   private void createComponentTemplate() {
     if (!client.putComponentTemplate()) {
       log.warn("Failed to acknowledge the creation or update of the component template");
-    }
-  }
-
-  private void createValueIndexTemplate(final ValueType valueType) {
-    if (!client.putIndexTemplate(valueType)) {
-      log.warn(
-          "Failed to acknowledge the creation or update of the index template for value type {}",
-          valueType);
     }
   }
 
--- a/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporter.java
+++ b/zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporter.java
@@ -23,6 +23,8 @@ import java.io.IOException;
 import java.time.Duration;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import org.slf4j.Logger;
@@ -57,7 +59,7 @@ public class ElasticsearchExporter implements Exporter {
   private MeterRegistry registry;
 
   private long lastPosition = -1;
-  private boolean indexTemplatesCreated;
+  private Map<String, Boolean> indexTemplatesCreated;
 
   @Override
   public void configure(final Context context) {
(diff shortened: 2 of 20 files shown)


zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java
```diff
@@ -61,6 +62,10 @@ String aliasNameForValueType(final ValueType valueType) {
   /** Returns the index for this value type, minus the current date. */
   String indexPrefixForValueType(final ValueType valueType) {
```
Zelldon, line 63: I think we shouldn't use this method anymore, as we always need to make sure that we write into the right version or?

EuroLew: I looked at the other usages and I agree 
- tests 
- `RecordIndexRouter.searchPatternForValueType[81]` we want this to return the versionUtil version
- `ElasticsearchClient.putIndexTemplate[165]` we want this to return the versionUtil version as we are created the current exporter (8.7.0) indices. 

All this can be refactored such that the (ValueType, Version) is the only function that is used but just wanted to make sure it's the correct logic as a draft.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java
```diff
@@ -40,7 +41,18 @@ final class RecordIndexRouter {
...
+
+    // if the record version is less than the current version use that as the record should be
+    // written to old versioned index.
+    if (recordVersion.isPresent() && recordVersion.get().compareTo(indexVersion) < 0) {
+      indexVersion = recordVersion.get();
+    }
+
+    return (indexPrefixForValueType(record.getValueType(), indexVersion.toString().toLowerCase())
+            + INDEX_DELIMITER)
         + formatter.format(timestamp);
```
Zelldon, line 56: ❌ I think we should "simply" always use the Record version, no?

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
```diff
@@ -163,7 +164,8 @@ public boolean shouldFlush() {
    * @return true if request was acknowledged
    */
   public boolean putIndexTemplate(final ValueType valueType) {
-    final String templateName = indexRouter.indexPrefixForValueType(valueType);
+    final String templateName =
+        indexRouter.indexPrefixForValueType(valueType, VersionUtil.getVersionLowerCase());
```
Zelldon, line 168: ❓ Here we can use the VersionUtil, because we create the new template right? For the new version of the Exporter?

ChrisKujawa (changes requested):
I think we should use always the record version, but happy to discuss this in more depth. I might oversee something.

@remcowesterhoud could you also have a short look, whether the changes are fine, as this falls in the ZPA (just to cross-check).

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
remcowesterhoud: Yes, this should be fine.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java
remcowesterhoud: I agree. The version is set on the record. If it's older we should write it to the older index. If it's not older, it should be the same as the version that comes out of `VersionUtil` anyway. So the version comparison is redundant.

remcowesterhoud (commented):
I can think of one edge-case where this could go wrong:

1. Broker is updated to 8.6.0
2. ES/OS is down
3. Broker is processing and tries to export records with version 8.6.0, which fails due to ES/OS being down
4. Broker is updated to 8.7.0 and continues processing
5. ES/OS comes back up
6. Exporting starts. The exporter tries to export the old records to the 8.6.0 index. This template doesn't exist, as they couldn't get created due to ES/OS being down.

When this happens a workaround would be to manually create the missing index template(s)

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/RecordIndexRouter.java
EuroLew: True the record version is what we should write to but there is a lot of tests which will fail as a result of not checking if the recordVersion is a valid semantic version as the factory which generates records will fill it with a random value. `

I agree though I'll work through the tests

image

EuroLew:
> I can think of one edge-case where this could go wrong:
> 
> 1. Broker is updated to 8.6.0
> 2. ES/OS is down
> 3. Broker is processing and tries to export records with version 8.6.0, which fails due to ES/OS being down
> 4. Broker is updated to 8.7.0 and continues processing
> 5. ES/OS comes back up
> 6. Exporting starts. The exporter tries to export the old records to the 8.6.0 index. This template doesn't exist, as they couldn't get created due to ES/OS being down.
> 
> When this happens a workaround would be to manually create the missing index template(s)

Adds a fix for the edge case https://github.com/camunda/camunda/pull/24483/commits/6e1f06a75ed38dad561c877b2325e2949c6dca34 where due to ES/OS being down templates are not created.

For each record version we create templates for that version if not created.

This is just a WIP commit to gauge interest in the solution

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
```diff
@@ -175,6 +175,17 @@ public boolean putIndexTemplate(final ValueType valueType) {
...
+  public boolean putIndexTemplate(final ValueType valueType, final String version) {
+    final String templateName = indexRouter.indexPrefixForValueType(valueType, version);
+    final Template template =
+        templateReader.readIndexTemplate(
+            valueType,
+            indexRouter.searchPatternForValueType(valueType, version),
+            indexRouter.aliasNameForValueType(valueType));
+
+    return putIndexTemplate(templateName, template);
+  }
```
remcowesterhoud, line 187: 🔧 You can reduce some duplication here by changing the method above this to just call this one with the `VersionUtil.getVersionLowerCase()` as the version

remcowesterhoud (commented):
@EuroLew I looked at your WIP commit. I like the solution and I believe it'll work 👍 

I think with this we may have solved https://github.com/camunda/camunda/issues/14863 as well 🤔 

We also need to do this for the OS exporter.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
EuroLew: yep was just doing it quick and dirty to see if it was worth doing

remcowesterhoud (approved):
Thanks @EuroLew, looks good to me 🚀 

Would you mind closing the issue I mentioned in previous comment when this PR is merged. I am 99% sure this is resolved with versioned exporting 🙂

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporter.java
```diff
@@ -245,117 +247,117 @@ private void createIndexTemplates() {
...
       }
 
       if (index.authorization) {
-        createValueIndexTemplate(ValueType.AUTHORIZATION);
+        createValueIndexTemplate(ValueType.AUTHORIZATION, version);
       }
     }
 
-    indexTemplatesCreated = true;
+    indexTemplatesCreated.add(version);
```
Zelldon, line 360: ❓ So we always try to create the indices on restart?

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
```diff
@@ -163,11 +164,15 @@ public boolean shouldFlush() {
    * @return true if request was acknowledged
    */
   public boolean putIndexTemplate(final ValueType valueType) {
-    final String templateName = indexRouter.indexPrefixForValueType(valueType);
+    return putIndexTemplate(valueType, VersionUtil.getVersionLowerCase());
```
Zelldon, line 167: Do we ever want to use this? With Broker version?

zeebe/exporters/elasticsearch-exporter/src/test/java/io/camunda/zeebe/exporter/ElasticsearchExporterIT.java
```diff
@@ -325,11 +355,94 @@ final var record = factory.generateRecord(ValueType.JOB);
...
+          .setPortBindings(List.of(currentPort + ":9200"));
+      CONTAINER.start();
+      Awaitility.await().until(CONTAINER::isRunning);
+
+      // when
+      final var record2 = factory.generateRecord(r -> r.withBrokerVersion("8.7.0"));
+      try (final var mockVersion =
+          Mockito.mockStatic(VersionUtil.class, Mockito.CALLS_REAL_METHODS)) {
+        mockVersion.when(VersionUtil::getVersionLowerCase).thenReturn("8.7.0");
+        configureExporter(false);
```
Zelldon, line 411: ❓ What does this do?

zeebe/exporters/elasticsearch-exporter/src/test/java/io/camunda/zeebe/exporter/RecordIndexRouterTest.java
```diff
@@ -142,7 +149,8 @@ void shouldReturnSearchPatternForValueType() {
     final var valueType = ValueType.PROCESS;
 
     // when
-    final var prefix = router.searchPatternForValueType(valueType);
+    final var prefix =
+        router.searchPatternForValueType(valueType, VersionUtil.getVersionLowerCase());
```
Zelldon, line 153: 🔧 I think it would be beneficial to also have a test not using the VersionUtil, to verify that we actual use the parameter not the static

zeebe/exporters/opensearch-exporter/src/test/java/io/camunda/zeebe/exporter/opensearch/OpensearchExporterIT.java
```diff
@@ -413,14 +441,18 @@ void shouldNotTimeoutWhenUpdatingLifecyclePolicyForExistingIndices() {
       // using 490 here as we will export one more record after (1 main shard, 1 replica)
       final int limit = 490;
       for (int i = 0; i < limit; i++) {
-        final var record = factory.generateRecord(ValueType.JOB);
+        final var record =
+            factory.generateRecord(
+                ValueType.JOB, r -> r.withBrokerVersion(VersionUtil.getVersionLowerCase()));
```
Zelldon, line 446: 🔧 Might make sense to make this the default - that the broker version is set correctly?

ChrisKujawa (approved):
Thanks. I think the solution likely works in our cases, but strictly speaking it is not correct, as we actually create the schema 8.7 just with a 8.6 version in the name. 

Just as feedback: Personally, I would have made a cut on this PR to scope it only to the one bug (and fix) we had planned before. 

And go with the part @remcowesterhoud mentioned:
> When this happens a workaround would be to manually create the missing index template(s)

Or look at it _separately_ when we have time for it (which we don't really have right now afaik).

I will any way approve, to not block us further and make progress here.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporter.java
EuroLew: Yes we always try to create on restart, this mimics the behaviour of the old `indexTemplatesCreated `logic where it was a boolean.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchClient.java
EuroLew: It's only used for tests right now and no the default implementation is always the VersionUtil version

zeebe/exporters/elasticsearch-exporter/src/test/java/io/camunda/zeebe/exporter/ElasticsearchExporterIT.java
EuroLew: oops yea it doesn't need to be configured twice as the port doesn't change I'll move it out.

PR #24497, opened 2024-11-07
ci: push testbench images to SaaS registry

lenaschoenburg:
In SaaS we can't use images from gcr.io/zeebe/ anymore so we'll push them to the blessed SaaS registry instead.

The starter/worker images can remain on gcr.io because they are running outside of SaaS.

### DIFF
--- a/.github/workflows/zeebe-testbench.yaml
+++ b/.github/workflows/zeebe-testbench.yaml
@@ -75,7 +75,7 @@ jobs:
       - uses: ./.github/actions/build-platform-docker
         id: build-zeebe-docker
         with:
-          repository: europe-docker.pkg.dev/camunda-saas-registry/zeebe/zeebe
+          repository: europe-docker.pkg.dev/camunda-saas-registry/zeebe-io/zeebe
           version: ${{ steps.image-tag.outputs.image-tag }}
           push: true
           distball: ${{ steps.build-zeebe.outputs.distball }}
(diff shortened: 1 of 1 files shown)


lenaschoenburg:
AFAICT we have no way of testing this without merging to main. Since QA is broken anyway, there's no risk in merging this without testing before.

.github/workflows/zeebe-testbench.yaml
```diff
@@ -62,7 +75,7 @@ jobs:
       - uses: ./.github/actions/build-platform-docker
         id: build-zeebe-docker
         with:
-          repository: gcr.io/zeebe-io/zeebe
+          repository: europe-docker.pkg.dev/camunda-saas-registry/zeebe/zeebe
```
than-pet, line 78:
```suggestion
          repository: europe-docker.pkg.dev/camunda-saas-registry/zeebe-io/zeebe
```

than-pet (changes requested):

.github/workflows/zeebe-testbench.yaml
lenaschoenburg: Are you sure? I can access europe-docker.pkg.dev/camunda-saas-registry/zeebe/zeebe but not europe-docker.pkg.dev/camunda-saas-registry/zeebe-io/zeebe

than-pet: Yes, you can try again.
It will work now.

than-pet (approved):

PR #24498, opened 2024-11-07
fix: handle flush error when leader tries to commit

### LINKED ISSUE #15463: Safely handle flush exceptions in Raft
> @deepthidevaki @oleschoenburg - could we implement minimal fault tolerance to flush error with the following:
> 
> 1. Follower/passive role truncates the log and retries the full append on flush error
> 2. Leader truncates the log to commit index and steps down on flush error
> 
> In both cases, truncating should invalidate the buffers/page cache. Since we only update the commit index on successful flush, then truncating to the last known commit index should be safe. 
> 
> This can cause a lot of disruption, but it will at least avoid inconsistencies. WDYT?
> 
> I would be happy to implement at least safety (even if performance/availability is terrible) first, and leave performance to the NFS feature request in the future.

_Originally posted by @npepinpe in https://github.com/camunda/zeebe/issues/14843#issuecomment-1783143543_

entangled90:
## Description
All flush() commands in the journal module now can throw a checked exception FlushException, in order to make sure that all possible exceptions thrown when flushing are handled.
A new class `CheckedJournalException` has been created, since `JournalException` extends `RuntimeException`, thus it would not be a checked exception.

Using `Either<Throwable,Void>` was also taken into consideration as a first approach, but I was afraid that it was likely that a left either value could be easily discarded as a return value by mistake, while this is not possible with checked exception.
This also means that using checked exception "pollutes" a lot the code so the idea is to use them only in critical parts of the code when we must ensure that exceptions are handled properly. 

When the leader fails to flush, it resets its commitIndex to the previous value and transitions to the Follower role.

## Related issues
closes #15463

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
@@ -26,6 +26,7 @@ import io.atomix.cluster.MemberId;
 import io.atomix.raft.ElectionTimer;
 import io.atomix.raft.RaftApplicationEntryCommittedPositionListener;
 import io.atomix.raft.RaftCommitListener;
+import io.atomix.raft.RaftException.CommitFailedException;
 import io.atomix.raft.RaftRoleChangeListener;
 import io.atomix.raft.RaftServer;
 import io.atomix.raft.RaftServer.Role;
@@ -73,8 +74,10 @@ import io.atomix.raft.zeebe.EntryValidator;
 import io.atomix.utils.concurrent.ThreadContext;
 import io.atomix.utils.logging.ContextualLoggerFactory;
 import io.atomix.utils.logging.LoggerContext;
+import io.camunda.zeebe.journal.CheckedJournalException.FlushException;
 import io.camunda.zeebe.snapshots.PersistedSnapshot;
 import io.camunda.zeebe.snapshots.ReceivableSnapshotStore;
+import io.camunda.zeebe.util.CheckedRunnable;
 import io.camunda.zeebe.util.exception.UnrecoverableException;
 import io.camunda.zeebe.util.health.FailureListener;
 import io.camunda.zeebe.util.health.HealthMonitorable;
@@ -546,12 +549,24 @@ public class RaftContext implements AutoCloseable, HealthMonitorable {
     checkArgument(commitIndex >= 0, "commitIndex must be positive");
     final long previousCommitIndex = this.commitIndex;
     if (commitIndex > previousCommitIndex) {
-      this.commitIndex = commitIndex;
-      raftLog.setCommitIndex(Math.min(commitIndex, raftLog.getLastIndex()));
       if (isLeader()) {
         // leader counts itself in quorum, so in order to commit the leader must persist
-        raftLog.flush();
+        try {
+          raftLog.flush();
+        } catch (final FlushException e) {
+          if (log.isWarnEnabled()) {
+            log.warn(
+                "Failed to flush commit at index %s, resetting journal to %s and stepping down"
+                    .formatted(commitIndex, previousCommitIndex),
+                e);
+          }
+          transition(Role.FOLLOWER);
+          throw new CommitFailedException(
+              "Failed to commit index %s because of a flush error: %s", commitIndex, e);
+        }
       }
+      raftLog.setCommitIndex(Math.min(commitIndex, raftLog.getLastIndex()));
+      this.commitIndex = commitIndex;
       meta.storeCommitIndex(commitIndex);
       final var clusterConfig = cluster.getConfiguration();
       if (clusterConfig != null) {
@@ -638,7 +653,7 @@ public class RaftContext implements AutoCloseable, HealthMonitorable {
       return CompletableFuture.completedFuture(null);
     }
 
-    return CompletableFuture.runAsync(raftLog::forceFlush, threadContext);
+    return CompletableFuture.runAsync(CheckedRunnable.toUnchecked(this::flushLog), threadContext);
   }
 
   /**
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/LeaderAppender.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/LeaderAppender.java
@@ -19,6 +19,7 @@ package io.atomix.raft.roles;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import io.atomix.raft.RaftException;
+import io.atomix.raft.RaftException.CommitFailedException;
 import io.atomix.raft.RaftException.NoLeader;
 import io.atomix.raft.RaftServer;
 import io.atomix.raft.cluster.RaftMember;
@@ -560,9 +561,13 @@ final class LeaderAppender {
     // commit.
     // The updated commit index will be sent to passive/reserve members on heartbeats.
     if (raft.getCluster().isSingleMemberCluster()) {
-      raft.setCommitIndex(index);
-      completeCommits(index);
-      return CompletableFuture.completedFuture(index);
+      try {
+        raft.setCommitIndex(index);
+        completeCommits(index);
+        return CompletableFuture.completedFuture(index);
+      } catch (final CommitFailedException e) {
+        return CompletableFuture.failedFuture(e);
+      }
     }
 
     if (!open) {
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java
@@ -46,6 +46,8 @@ import io.atomix.raft.snapshot.impl.SnapshotChunkImpl;
 import io.atomix.raft.storage.log.IndexedRaftLogEntry;
 import io.atomix.raft.storage.log.RaftLogReader;
 import io.atomix.raft.storage.system.Configuration;
+import io.camunda.zeebe.journal.CheckedJournalException;
+import io.camunda.zeebe.journal.CheckedJournalException.FlushException;
 import io.camunda.zeebe.journal.JournalException;
 import io.camunda.zeebe.journal.JournalException.InvalidChecksum;
 import io.camunda.zeebe.journal.JournalException.InvalidIndex;
@@ -54,7 +56,11 @@ import io.camunda.zeebe.snapshots.ReceivedSnapshot;
 import io.camunda.zeebe.snapshots.SnapshotException.SnapshotAlreadyExistsException;
 import io.camunda.zeebe.snapshots.impl.SnapshotChunkId;
 import io.camunda.zeebe.util.Either;
+import io.camunda.zeebe.util.ExponentialBackoffRetryDelay;
+import io.camunda.zeebe.util.SubClassOf;
 import io.camunda.zeebe.util.logging.ThrottledLogger;
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.List;
@@ -65,6 +71,13 @@ import org.agrona.concurrent.UnsafeBuffer;
 
 /** Passive state. */
 public class PassiveRole extends InactiveRole {
+
+  private static final SubClassOf<Exception> RETRYABLE_EXCEPTIONS =
+      SubClassOf.make(
+          CheckedJournalException.class,
+          JournalException.class,
+          UncheckedIOException.class,
+          IOException.class);
 
   private final ThrottledLogger throttledLogger = new ThrottledLogger(log, Duration.ofSeconds(5));
   private final SnapshotReplicationMetrics snapshotReplicationMetrics;
@@ -84,15 +97,16 @@ public class PassiveRole extends InactiveRole {
 
   @Override
   public CompletableFuture<RaftRole> start() {
-
-    return super.start().thenRun(this::truncateUncommittedEntries).thenApply(v -> this);
+    return super.start()
+        .thenCompose((ignored) -> truncateUncommittedEntriesWithRetries())
+        .thenApply(v -> this);
   }
 
   @Override
   public CompletableFuture<Void> stop() {
     abortPendingSnapshots();
 
-    // as a safe guard, we clean up any orphaned pending snapshots
+    // as a safeguard, we clean up any orphaned pending snapshots
     try {
       raft.getPersistedSnapshotStore().purgePendingSnapshots().join();
     } catch (final Exception e) {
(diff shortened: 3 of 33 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
```diff
@@ -546,12 +548,25 @@ public long setCommitIndex(final long commitIndex) {
     checkArgument(commitIndex >= 0, "commitIndex must be positive");
     final long previousCommitIndex = this.commitIndex;
     if (commitIndex > previousCommitIndex) {
-      this.commitIndex = commitIndex;
       raftLog.setCommitIndex(Math.min(commitIndex, raftLog.getLastIndex()));
       if (isLeader()) {
         // leader counts itself in quorum, so in order to commit the leader must persist
-        raftLog.flush();
+        try {
```
entangled90, line 554: This is the change in the leader's behaviour

zeebe/atomix/cluster/src/test/java/io/atomix/raft/FaultyFlusherConfigurator.java
```diff
@@ -0,0 +1,76 @@
...
+  private RaftLogFlusher.Factory faultyFlusher(
+      final Supplier<Boolean> faultyWhen, final Runnable notifyFaultyFlush) {
+    return (ignored) ->
+        new RaftLogFlusher() {
+          @Override
+          public void flush(final Journal journal) throws FlushException {
+            if (faultyWhen.get()) {
+              notifyFaultyFlush.run();
+              if (withDataLoss) {
+                journal.deleteAfter(journal.getLastIndex() - 1);
```
entangled90, line 44: Do you think this addition is worth?

zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java
```diff
@@ -106,7 +106,11 @@ public CompletableFuture<Void> stop() {
   /** Truncates uncommitted entries from the log. */
   private void truncateUncommittedEntries() {
     if (role() == RaftServer.Role.PASSIVE && raft.getLog().getLastIndex() > raft.getCommitIndex()) {
-      raft.getLog().deleteAfter(raft.getCommitIndex());
+      try {
+        raft.getLog().deleteAfter(raft.getCommitIndex());
+      } catch (final FlushException e) {
+        throw new RuntimeException(e);
```
npepinpe, line 112: :wrench: Not a big fan of throwing generic runtime exceptions. If we're just rethrowing, why not sneaky throw via `LangUtil.rethrowUnchecked`? 

:x: But also, why are we rethrowing here? From what I can see, this will likely cause us to go inactive. What we would want is likely to just retry forever, no? Or maybe with a set number of time, before going inactive. Otherwise one blip and you end up inactive, I think. Note that I'm not 100% sure, so I could be wrong - but looking at the code, it looks like that's what will happen.

zeebe/journal/src/main/java/io/camunda/zeebe/journal/file/FlushableSegment.java
```diff
@@ -22,7 +24,8 @@ interface FlushableSegment {
    * <p>If the method returns true, then it is guaranteed that the modified pages for this segment *
    * have been flushed to disk (according to the underlying file system).
    *
+   * @throws FlushException if for any reason the flush failed
    * @return true if the segment flushed successfully, false otherwise
    */
-  boolean flush();
+  boolean flush() throws FlushException;
```
npepinpe, line 30: :question: I wonder if the boolean return value makes sense anymore, since we now throw if the flush failed :shrug: 

:wrench: It probably doesn't, and we should change that. Can gladly be a follow up PR/issue.

zeebe/journal/src/main/java/io/camunda/zeebe/journal/file/Segment.java
```diff
@@ -121,14 +122,14 @@ public long lastIndex() {
...
-  public boolean flush() {
+  public boolean flush() throws FlushException {
     final long lastIndex = lastIndex();
 
     try (final var ignored = metrics.observeSegmentFlush()) {
       buffer.force();
     } catch (final UncheckedIOException e) {
       if (isOpen()) {
-        throw e;
+        throw new FlushException(e.getCause());
```
npepinpe, line 132: :question: Any advantage in using the cause directly instead of the exception itself? I guess we don't lose too much info?

zeebe/atomix/cluster/src/test/java/io/atomix/raft/FaultyFlusherConfigurator.java
npepinpe: Hm, I'm not sure what's the additional test coverage with this case. We already have data loss tests. So here specifically it would be testing data loss on flush error, right? I don't think it's harmful, but I'm not seeing the big advantage. Can you elaborate on it?

zeebe/atomix/cluster/src/test/java/io/atomix/raft/roles/RaftLeaderFlushErrorTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    // then
+    assertThatThrownBy(() -> commitListener.awaitCommit(Duration.ofSeconds(2)))
+        .isExactlyInstanceOf(TimeoutException.class);
+
+    // when
+    isFaulty.set(false);
+    LOG.info("Leader flusher is not faulty anymore");
+
+    assertThatThrownBy(() -> commitListener.awaitCommit(Duration.ofSeconds(2)))
```
npepinpe, line 97: :question: Why await again that the commit fails? Since we passed the assertion above, the future is already committed, so this will immediately pass still. Seems redundant.

zeebe/atomix/cluster/src/test/java/io/atomix/raft/roles/RaftLeaderFlushErrorTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    // flusher fails to flush
+    isFaulty.set(true);
+
+    LOG.info("Leader flusher set to faulty");
+    // when
+    final var commitListener = raftRule.appendEntryAsync();
+
+    // then
+    assertThatThrownBy(() -> commitListener.awaitCommit(Duration.ofSeconds(2)))
```
npepinpe, line 90: :x: :thinking: It's not ideal to just wait out until something doesn't happen. But then again, it's always hard to test the absence of something.

Maybe we can rethink the approach here. Maybe we can instead assert one or more of the following:

Leader steps down. Why not await a new leader? Since it had quorum, and it truncates entries on becoming a follower, then the other node (who gave quorum) should be elected, right? :thinking: 

That would also allow us to test that the old leader became follower, as expected.

npepinpe (changes requested):
:rocket: This was more complex than I expected (sort of - I guess it makes sense :smile:)

zeebe/atomix/cluster/src/main/java/io/atomix/raft/impl/RaftContext.java
```diff
@@ -546,12 +549,25 @@ public long setCommitIndex(final long commitIndex) {
...
+          raftLog.flush();
+        } catch (final FlushException e) {
+          if (log.isWarnEnabled()) {
+            log.warn(
+                "Failed to flush commit at index %s, resetting journal to %s and stepping down"
+                    .formatted(commitIndex, previousCommitIndex),
+                e);
+          }
+          raftLog.setCommitIndex(previousCommitIndex);
+          transition(Role.FOLLOWER);
```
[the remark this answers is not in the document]
npepinpe: Oh, so we truncate on becoming follower, right? So it should be fine. As long as we just keep trying to become follower if this doesn't work :smile:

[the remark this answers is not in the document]
npepinpe: Now that I think about it...truncate could fail (I/O error). So we should make sure we retry instead of just going inactive. But that can be a follow up PR/issue in order to keep this small, and this is still a big improvement on the current behavior :+1:

zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/LeaderAppender.java
```diff
@@ -560,9 +561,13 @@ public CompletableFuture<Long> appendEntries(final long index) {
...
     if (raft.getCluster().isSingleMemberCluster()) {
-      raft.setCommitIndex(index);
-      completeCommits(index);
-      return CompletableFuture.completedFuture(index);
+      try {
+        raft.setCommitIndex(index);
+        completeCommits(index);
+        return CompletableFuture.completedFuture(index);
+      } catch (final CommitFailedException e) {
+        return CompletableFuture.failedFuture(e);
```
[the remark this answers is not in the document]
entangled90: In this case, we transition to follower, which will fail the client's request. Given that it's a single node cluster we should be reelected again. The client can retry after that. Is that ok or should we implement retries here?

zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java
```diff
@@ -104,10 +108,21 @@ public CompletableFuture<Void> stop() {
...
+            try {
+              raft.getLog().deleteAfter(raft.getCommitIndex());
+            } catch (final FlushException e) {
+              LangUtil.rethrowUnchecked(e);
+            }
+          }
+        };
+    final var retryStrategy =
+        new ExponentialBackoffRetryDelay(Duration.ofSeconds(1), Duration.ofMillis(50));
+    return raft.getThreadContext().retryUntilSuccessful(runnable, retryStrategy);
```
npepinpe, line 125: :thought_balloon: This sounds good in general - I'm wondering if I'm not forgetting a case where this could fail in an unrecoverable way, in which case we would want to go inactive/dead. :thinking: 

Maybe we can double check with Lena in case I'm forgetting anything.

zeebe/atomix/utils/src/test/java/io/atomix/utils/concurrent/SingleThreadContextTest.java
```diff
@@ -49,4 +54,26 @@ public void shouldInvokeHandlerOnException() throws InterruptedException {
...
+    final var result =
+        threadContext.retryUntilSuccessful(
+            () -> {
+              if (!success.get()) {
+                throw new RuntimeException("Expected");
+              }
+            },
+            delayStrategy);
+
+    Thread.sleep(250);
```
npepinpe, line 72: :wrench: I would advise against thread sleep, as this can easily be flaky. Typically we use things like CyclicBarrier or CountDownLatch to control asynchronous execution. Another option would be using an `AtomicInteger` in the retry runnable, use awaitility to wait until it's been called a couple times, and then flip the switch, and then assert eventually it completes :+1:

zeebe/util/src/main/java/io/camunda/zeebe/util/CheckedRunnable.java
```diff
@@ -7,10 +7,22 @@
...
+import org.agrona.LangUtil;
+
 /** A simple extension of runnable which allows for exceptions to be thrown. */
 @FunctionalInterface
 // ignore generic exception warning here as we want to allow for any exception to be thrown
 @SuppressWarnings("java:S112")
 public interface CheckedRunnable {
   void run() throws Exception;
+
+  static java.lang.Runnable toRunnable(final CheckedRunnable runnable) {
```
npepinpe, line 19:
```suggestion
  static java.lang.Runnable toUnchecked(final CheckedRunnable runnable) {
```

zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/LeaderAppender.java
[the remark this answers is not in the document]
npepinpe: No, I think that's fine :+1:

npepinpe (approved):
Generally looks good, there's only one thing which I'd like to get Lena's opinion on, just to be sure I'm not forgetting anything.

zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java
npepinpe: Maybe we want to really _only_ retry IOExceptions forever, but not retry anything else? :thinking:

zeebe/atomix/utils/src/main/java/io/atomix/utils/concurrent/Scheduler.java
```diff
@@ -77,6 +81,38 @@ default Scheduled schedule(
...
+            }
+          }
+        };
+
+    runnable.run();
+    return result;
+  }
+
+  default CompletableFuture<Void> retryUntilSuccessful(
+      final Runnable runnable, final RetryDelayStrategy retryDelayStrategy) {
```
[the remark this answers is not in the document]
entangled90: I'll add one :+1:

zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java
entangled90: Yes I would say so as well

zeebe/atomix/cluster/src/test/java/io/atomix/raft/roles/RaftLeaderFlushErrorTest.java
npepinpe: Normally we should elect the node with the longest log. But I now remember that we also truncate the uncommitted entries on becoming candidate, so I guess the other follower will also do the same, and then we will default to the priority instead.

zeebe/atomix/cluster/src/main/java/io/atomix/raft/roles/PassiveRole.java
npepinpe: So should we mark https://github.com/camunda/camunda/issues/24634 done as well?

zeebe/util/src/main/java/io/camunda/zeebe/util/SubClassOf.java
```diff
@@ -0,0 +1,43 @@
...
+ *   IllegalStateException.class
+ *   );
+ * subclassOf.test(new Exception("Error"));
+ *
+ * }</pre>
+ *
+ * @param classes
+ * @param <A>
+ */
+public record SubClassOf<A>(Class<? extends A>[] classes) implements Predicate<A> {
```
npepinpe, line 27: :+1: Nice

zeebe/util/src/test/java/io/camunda/zeebe/util/SubClassOfTest.java
```diff
@@ -0,0 +1,37 @@
...
+class SubClassOfTest {
+  @Test
+  public void shouldReturnCorrectSubclass() {
+    final var subclass =
+        SubClassOf.make(IllegalArgumentException.class, IllegalStateException.class);
+
+    assertTrue(subclass.test(new IllegalArgumentException("")));
+    assertTrue(subclass.test(new IllegalStateException("")));
+    assertTrue(subclass.test(new MyIllegalArgumentException()));
+    assertFalse(subclass.test(new RuntimeException("")));
```
npepinpe, line 23: :x: Please use AssertJ and not the built-in assertions. They are more flexible, and just more consistent with our overall style.

zeebe/util/src/test/java/io/camunda/zeebe/util/SubClassOfTest.java
```diff
@@ -0,0 +1,37 @@
...
+  public void shouldNotReturnTrueIfEmpty() {
+    final var subclass = SubClassOf.make();
+    final var objects = new Object[] {"", 1, 1.0, new Object[] {}};
+
+    for (final var obj : objects) {
+      assertFalse(subclass.test(obj));
+    }
+  }
+
+  public static final class MyIllegalArgumentException extends IllegalArgumentException {}
```
npepinpe, line 36: :wrench: Can be private - just as a habit, good to always make things with the lowest required visibility

PR #24499, opened 2024-11-07
OS repository implement setIndexLifeclePolicy

### LINKED ISSUE #24094: Implement ES/OS archiver repository

rodrigo-lourenco-lopes:
## Description

This PR implements setIndexLifeclePolicy in the OS repository along with tests. 

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/24094

### DIFF
--- a/zeebe/exporters/camunda-exporter/pom.xml
+++ b/zeebe/exporters/camunda-exporter/pom.xml
@@ -121,6 +121,11 @@
     </dependency>
 
     <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>org.testcontainers</groupId>
       <artifactId>junit-jupiter</artifactId>
       <scope>test</scope>
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/OpenSearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/OpenSearchRepository.java
@@ -7,6 +7,7 @@
  */
 package io.camunda.exporter.archiver;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import io.camunda.exporter.archiver.ArchiverRepository.NoopArchiverRepository;
 import io.camunda.exporter.config.ExporterConfiguration.ArchiverConfiguration;
 import io.camunda.exporter.config.ExporterConfiguration.RetentionConfiguration;
@@ -41,6 +42,8 @@ import org.opensearch.client.opensearch.core.SearchRequest;
 import org.opensearch.client.opensearch.core.SearchResponse;
 import org.opensearch.client.opensearch.core.reindex.Source;
 import org.opensearch.client.opensearch.core.search.Hit;
+import org.opensearch.client.opensearch.generic.OpenSearchGenericClient;
+import org.opensearch.client.opensearch.generic.Requests;
 import org.slf4j.Logger;
 
 public final class OpenSearchRepository extends NoopArchiverRepository {
@@ -59,7 +62,7 @@ public final class OpenSearchRepository extends NoopArchiverRepository {
   private final Executor executor;
   private final CamundaExporterMetrics metrics;
   private final Logger logger;
-
+  private final OpenSearchGenericClient genericClient;
   private final CalendarInterval rolloverInterval;
 
   public OpenSearchRepository(
@@ -82,6 +85,7 @@ public final class OpenSearchRepository extends NoopArchiverRepository {
     this.metrics = metrics;
     this.logger = logger;
 
+    genericClient = new OpenSearchGenericClient(client._transport(), client._transportOptions());
     rolloverInterval = mapCalendarInterval(config.getRolloverInterval());
   }
 
@@ -110,6 +114,37 @@ public final class OpenSearchRepository extends NoopArchiverRepository {
   }
 
   @Override
+  public CompletableFuture<Void> setIndexLifeCycle(final String destinationIndexName) {
+    if (!retention.isEnabled()) {
+      return CompletableFuture.completedFuture(null);
+    }
+
+    final AddPolicyRequestBody value = new AddPolicyRequestBody(retention.getPolicyName());
+    final var request =
+        Requests.builder().method("POST").endpoint("_plugins/_ism/add/" + destinationIndexName);
+
+    return sendRequestAsync(
+            () ->
+                genericClient.executeAsync(
+                    request.json(value, genericClient._transport().jsonpMapper()).build()))
+        .thenComposeAsync(
+            response -> {
+              if (response.getStatus() >= 400) {
+                return CompletableFuture.failedFuture(
+                    new ExporterException(
+                        "Failed to set index lifecycle policy for index: "
+                            + destinationIndexName
+                            + ".\n"
+                            + "Status: "
+                            + response.getStatus()
+                            + ", Reason: "
+                            + response.getReason()));
+              }
+              return CompletableFuture.completedFuture(null);
+            });
+  }
+
+  @Override
   public CompletableFuture<Void> deleteDocuments(
       final String sourceIndexName,
       final String idFieldName,
@@ -128,17 +163,6 @@ public final class OpenSearchRepository extends NoopArchiverRepository {
         .whenCompleteAsync((ignored, error) -> metrics.measureArchiverDelete(timer), executor)
         .thenApplyAsync(DeleteByQueryResponse::total, executor)
         .thenApplyAsync(ok -> null, executor);
-  }
-
-  private SearchRequest createFinishedBatchOperationsSearchRequest(final Aggregation aggregation) {
-    final var endDateQ =
-        QueryBuilders.range()
-            .field(BatchOperationTemplate.END_DATE)
-            .lte(JsonData.of(config.getArchivingTimePoint()))
-            .build();
-
-    return createSearchRequest(
-        batchOperationIndex, endDateQ.toQuery(), aggregation, BatchOperationTemplate.END_DATE);
   }
 
   @Override
@@ -165,6 +189,17 @@ public final class OpenSearchRepository extends NoopArchiverRepository {
     return sendRequestAsync(() -> client.reindex(request))
         .whenCompleteAsync((ignored, error) -> metrics.measureArchiverReindex(timer), executor)
         .thenApplyAsync(ignored -> null, executor);
+  }
+
+  private SearchRequest createFinishedBatchOperationsSearchRequest(final Aggregation aggregation) {
+    final var endDateQ =
+        QueryBuilders.range()
+            .field(BatchOperationTemplate.END_DATE)
+            .lte(JsonData.of(config.getArchivingTimePoint()))
+            .build();
+
+    return createSearchRequest(
+        batchOperationIndex, endDateQ.toQuery(), aggregation, BatchOperationTemplate.END_DATE);
   }
 
   private ArchiveBatch createArchiveBatch(final SearchResponse<?> search) {
@@ -282,6 +317,8 @@ public final class OpenSearchRepository extends NoopArchiverRepository {
         .size(0)
         .build();
   }
+
+  private record AddPolicyRequestBody(@JsonProperty("policy_id") String policyId) {}
 
   @FunctionalInterface
   private interface RequestSender<T> {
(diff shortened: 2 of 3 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
```diff
@@ -81,6 +87,81 @@ void shouldDeleteDocuments() throws IOException {
...
+    final var result = repository.setIndexLifeCycle(indexName);
+
+    // then
+    assertThat(result).succeedsWithin(Duration.ofSeconds(30));
+
+    // Takes a while for the policy to be applied
+    Awaitility.await()
+        .atMost(3, TimeUnit.SECONDS)
+        .pollInterval(1, TimeUnit.SECONDS)
+        .until(() -> indexContainsRetentionPolicy(indexName, genericClient));
```
npepinpe, line 115: :x: I would change the assertion. What happens now when it fails? You just get that you expected something to be true, but it was false, right? It doesn't give you a lot of information on failure, which will make it difficult when you want to diagnose problems.

Instead, we can try and get some of that information in the failure message.

Let's change the signature of `indexContainsRetentionPolicy`, and instead have a method which just returns the policy of the index. For that, we first need to be able to parse the body. Since the sample response from the API is:

```json
{
  "index_1": {
    "index.plugins.index_state_management.policy_id": "policy_1"
  }
}
```

Introducing a type to parse that would be difficult, because the top level field names are basically the index name - so they're dynamic! So it's hard to define a POJO for deserialization. But since the structure is simple enough, we can just use the basic JSON API.

We add a method which fetches the policy for a given index. I saw you used the asynchronous client, but it's much easier and less flaky in a test to just use the synchronous variant:

> [!Note]
> We'd have created a static `ObjectMapper MAPPER` property in the class before.

```java
  private String fetchPolicyForIndex(
      final String indexName, final OpenSearchGenericClient genericClient) {
    final var request =
        Requests.builder().method("get").endpoint("_plugins/_ism/explain/" + indexName).build();
    try {
      final var response = genericClient.execute(request);
      final var jsonString = response.getBody().orElseThrow().bodyAsString();
      final var json = MAPPER.readTree(jsonString);
      final var index = json.get(indexName);
      if (index == null) {
        throw new AssertionError(
            "Failed to explain non-existent index '%s'; see response: %s"
                .formatted(indexName, jsonString));
      }

      return index.findPath("index.plugins.index_state_management.policy_id").asText();
      throw new AssertionError("Failed to fetch policy for index " + indexName, e);
    }
  }
```

Then we can use the `untilAsserted` variant of `Awaitility`, which will print out the last failed assertion on failure, giving us some more information about why the test failed.

```java
    Awaitility.await("until the policy has been visibly applied")
        .untilAsserted(
            () ->
                assertThat(fetchPolicyForIndex(indexName, genericClient))
                    .isEqualTo(retention.getPolicyName()));
```

> [!Note]
> We also don't really need to override defaults of Awaitility - let's do that only if we see a need to after it's already running for a while.

It's really important to have good error messages on test failures, otherwise it's easy to lose time trying to understand what went wrong.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
```diff
@@ -21,19 +22,24 @@
...
 import org.opensearch.client.opensearch.core.search.Hit;
+import org.opensearch.client.opensearch.generic.OpenSearchGenericClient;
+import org.opensearch.client.opensearch.generic.Requests;
 import org.opensearch.client.transport.rest_client.RestClientTransport;
 import org.opensearch.testcontainers.OpensearchContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.shaded.org.awaitility.Awaitility;
```
npepinpe, line 42: :x: Be careful not to import shaded dependencies, but the actual dependency. It's a very easy mistake to make with an IDE :sweat_smile: 

> [!Note]
> In case you never came across that, shaded dependencies are dependencies which have been fully merged into another one - think putting a JAR into another JAR - and then all their classes are relocated under a nested package. So instead of `org.awaitility.Awaitility`, you get `org.testcontainers.shaded.org.awaitility.Awaitility`.

It's better to use the real dependency, so we can properly keep it up to date and detect CVEs.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
```diff
@@ -81,6 +87,81 @@ void shouldDeleteDocuments() throws IOException {
...
+  @Test
+  void shouldSetIndexLifeCycle() throws IOException {
+    // given
+    final var indexName = UUID.randomUUID().toString();
+    final var repository = createRepository();
+    testClient.indices().create(r -> r.index(indexName));
+
+    final OpensearchEngineClient engineClient = new OpensearchEngineClient(testClient);
+    final OpenSearchGenericClient genericClient =
+        new OpenSearchGenericClient(testClient._transport(), testClient._transportOptions());
```
npepinpe, line 99: :wrench: Both can be `final var`, because we're constructed the object directly, so the type is obvious from the right-hand-side of the declaration. It's better not to use var only if the type is not really obvious.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
```diff
@@ -81,6 +87,81 @@ void shouldDeleteDocuments() throws IOException {
...
+
+    retention.setEnabled(true);
+    retention.setPolicyName("operate_delete_archived_indices");
+
+    // when
+    createLifeCyclePolicy(engineClient);
```
npepinpe, line 105: :wrench: Considering the `engineClient` is only used for this method call, why not just create it in there, and remove the argument? Since `testClient` is available anyway to the whole class.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
```diff
@@ -151,6 +240,8 @@ private RestClientTransport createRestClient() {
     return new RestClientTransport(restClient, new JacksonJsonpMapper());
   }
 
+  private record TestBatchOperation(String id, String endDate) implements TDocument {}
```
npepinpe, line 243: :x: I guess this is a left over from another PR?

npepinpe (changes requested):
:+1: That was more complex than expected, no fault to you, OpenSearch has a really weird API :s

Please address all comments - I don't think we'll need another review, but I'm not approving yet because it's a lot of changes.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
rodrigo-lourenco-lopes: It is also used in indexContainsRetentionPolicy() but I will see if it is possible to refactor after rewriting the test 🤔

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
```diff
@@ -81,6 +87,81 @@ void shouldDeleteDocuments() throws IOException {
...
+    // when
+    createLifeCyclePolicy(engineClient);
+    final var result = repository.setIndexLifeCycle(indexName);
+
+    // then
+    assertThat(result).succeedsWithin(Duration.ofSeconds(30));
+
+    // we query the index lifecycle policy a few times to ensure it's not set
+    // since it takes a while for the policy to be applied.
+    final AtomicInteger attemptCounter = new AtomicInteger(0);
```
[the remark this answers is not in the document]
rodrigo-lourenco-lopes: This is indeed a much cleaner solution, thanks 🙌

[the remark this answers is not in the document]
npepinpe: Or as we discussed, it might be even better to make a unit test out of it, no? Then asserting that `repository.setIndexLifeCycle` succeeds, because without a valid database, it would always fail.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
rodrigo-lourenco-lopes: Good catch 🙌

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/archiver/OpenSearchRepository.java
```diff
@@ -146,6 +189,74 @@ private <T> CompletableFuture<T> sendRequestAsync(final RequestSender<T> sender)
...
+        .requestCache(false)
+        .allowNoIndices(true)
+        .ignoreUnavailable(true)
+        .source(source -> source.fetch(false))
+        .query(query -> query.constantScore(q -> q.filter(filterQuery)))
+        .aggregations(DATES_AGG, aggregation)
+        .sort(sort -> sort.field(field -> field.field(sortField).order(SortOrder.Asc)))
+        .size(0)
+        .build();
+  }
```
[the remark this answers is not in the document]
rodrigo-lourenco-lopes: Yes just need to rebase 👍

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
```diff
@@ -81,6 +87,81 @@ void shouldDeleteDocuments() throws IOException {
...
+        .pollInterval(1, TimeUnit.SECONDS)
+        .until(
+            () -> {
+              attemptCounter.incrementAndGet();
+              return indexContainsRetentionPolicy(indexName, genericClient)
+                  || attemptCounter.get() >= 3;
+            });
+  }
+
+  boolean indexContainsRetentionPolicy(
```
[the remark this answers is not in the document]
rodrigo-lourenco-lopes: Good catch 👍

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
rodrigo-lourenco-lopes: Good point, the previous error message was not useful at all, thanks for the suggestion 🙌

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/archiver/OpenSearchRepositoryIT.java
rodrigo-lourenco-lopes: Solved with the rabase as well

npepinpe (approved):

zeebe/exporters/camunda-exporter/pom.xml
```diff
@@ -210,6 +210,13 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+      <version>2.18.0</version>
+      <scope>test</scope>
```
npepinpe, line 217: :x: Isn't it used in production code as well?

PR #24503, opened 2024-11-07
feat: multi-store config for document module

### LINKED ISSUE #24508: [Doc Handling] Multiple store support
**Description**

We need to introduce environment-based configuration of the document store module that would allow the users to:
- Plug in new store implementations
- Override and configure store IDs
- Pass custom properties to the document stores

chillleader:
## Description

This PR will introduce the basic configuration for the document module that would allow defining document store properties as environment variables.

**Goals:**
* Custom document stores can be plugged into Camunda without modifying the document module
* Users can associate IDs with document store implementations arbitrarily
  * e.g. this should be possible: there is only one implementation for the GCP document store, but logically there are multiple store instances of the GCP store with different IDs, each connecting to its own bucket
* If no document store is configured, an instance of the in-memory document store is used (in the future can be replaced with local file system store)

**Example configuration:**
```
DOCUMENT_DEFAULT_STORE_ID=gcp-first

DOCUMENT_STORE_GCP-FIRST_CLASS=io.camunda.document.store.gcp.GcpDocumentStoreProvider
DOCUMENT_STORE_GCP-FIRST_BUCKET=first-bucket-name

DOCUMENT_STORE_GCP-SECOND_CLASS=io.camunda.document.store.gcp.GcpDocumentStoreProvider
DOCUMENT_STORE_GCP-SECOND_BUCKET=second-bucket-name
```

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/24508

### DIFF
--- a/document/store/src/main/java/io/camunda/document/store/EnvironmentConfigurationLoader.java
+++ b/document/store/src/main/java/io/camunda/document/store/EnvironmentConfigurationLoader.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.store;
+
+import io.camunda.document.api.DocumentStoreConfiguration;
+import io.camunda.document.api.DocumentStoreConfiguration.DocumentStoreConfigurationRecord;
+import io.camunda.document.api.DocumentStoreProvider;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * DOCUMENT_HASH_VERIFICATION_ENABLED=true
+ * DOCUMENT_STORE_GCP_CLASS=io.camunda.document.store.gcp.GcpDocumentStoreProvider
+ * DOCUMENT_STORE_GCP_BUCKET=my-bucket
+ * DOCUMENT_STORE_INMEMORY_CLASS=io.camunda.document.store.inmemory.InMemoryDocumentStoreProvider
+ */
+public class EnvironmentConfigurationLoader implements DocumentStoreConfigurationLoader {
+
+  private static final String CONFIGURATION_PREFIX = "DOCUMENT_";
+  private static final String STORE_PREFIX = "DOCUMENT_STORE_";
+  private static final String DEFAULT_DOCUMENT_STORE_ID = "DEFAULT_STORE_ID";
+
+  private static final Pattern DOCUMENT_STORE_PROPERTY_PATTERN =
+      Pattern.compile("^" + STORE_PREFIX + "(?<id>[^_]+)_(?<property>.+)$");
+  private static final String STORE_CLASS = "CLASS";
+
+  private final Map<String, String> environmentVariableOverrides = new HashMap<>();
+
+  @Override
+  public DocumentStoreConfiguration loadConfiguration() {
+    return new DocumentStoreConfiguration(
+        getRootLevelProperty(DEFAULT_DOCUMENT_STORE_ID).map(String::toLowerCase).orElse(null),
+        loadDocumentStoreProperties());
+  }
+
+  // For testing purposes
+  void addEnvironmentVariableOverride(final String key, final String value) {
+    environmentVariableOverrides.put(key, value);
+  }
+
+  private List<DocumentStoreConfigurationRecord> loadDocumentStoreProperties() {
+    final var envVars = new HashSet<>(environmentVariableOverrides.keySet());
+    envVars.addAll(System.getenv().keySet());
+    final var matches =
+        envVars.stream().map(DOCUMENT_STORE_PROPERTY_PATTERN::matcher).filter(Matcher::matches);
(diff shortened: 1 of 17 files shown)


document/store/src/main/java/io/camunda/document/store/EnvironmentConfigurationLoader.java
```diff
@@ -0,0 +1,103 @@
...
+        getRootLevelProperty(DEFAULT_DOCUMENT_STORE_ID).map(String::toLowerCase).orElse(null),
+        loadDocumentStoreProperties());
+  }
+
+  // For testing purposes
+  void addEnvironmentVariableOverride(final String key, final String value) {
+    environmentVariableOverrides.put(key, value);
+  }
+
+  private List<DocumentStoreConfigurationRecord> loadDocumentStoreProperties() {
```
nicpuppa, line 51: This method look great, but it's hard to read. I'd suggest to split it into separate methods.

```suggestion
  private List<DocumentStoreConfigurationRecord> loadDocumentStoreProperties() {
    final var storeProperties = extractStorePropertiesFromEnvVars();
    return storeProperties.entrySet().stream()
        .map(this::createConfigurationRecord)
        .toList();
  }

  private Map<String, Map<String, String>> extractStorePropertiesFromEnvVars() {
    Set<String> envVars = new HashSet<>(environmentVariableOverrides.keySet());
    envVars.addAll(System.getEnv().keySet());
    final Map<String, Map<String, String>> storeProperties = new HashMap<>();
    envVars.stream()
        .map(DOCUMENT_STORE_PROPERTY_PATTERN::matcher)
        .filter(Matcher::matches)
        .forEach(matcher -> {
          final var storeId = matcher.group("id");
          final var storeProperty = matcher.group("property");
          final var value = getEnvVariable(matcher.group()).orElse(null);
          storeProperties.computeIfAbsent(storeId, k -> new HashMap<>()).put(storeProperty, value);
        });
    return storeProperties;
  }

  private DocumentStoreConfigurationRecord createConfigurationRecord(Map.Entry<String, Map<String, String>> entry) {
    final var storeId = entry.getKey();
    final var properties = entry.getValue();
    final var storeClassName = properties.remove(STORE_CLASS);

    if (storeClassName == null) {
      throw new IllegalArgumentException("Store class not defined for store: " + storeId);
    }

    final Class<? extends DocumentStoreProvider> storeClass = resolveStoreClass(storeClassName);
    return new DocumentStoreConfigurationRecord(storeId.toLowerCase(), storeClass, properties);
  }

  private Class<? extends DocumentStoreProvider> resolveStoreClass(final String storeClassName) {
    try {
      return (Class<? extends DocumentStoreProvider>) Class.forName(storeClassName);
    } catch (ClassNotFoundException e) {
      throw new IllegalArgumentException("Store class not found: " + storeClassName, e);
    }
  }
```

document/store/src/main/java/io/camunda/document/store/EnvironmentConfigurationLoader.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Override
+  public DocumentStoreConfiguration loadConfiguration() {
+    return new DocumentStoreConfiguration(
+        getRootLevelProperty(DEFAULT_DOCUMENT_STORE_ID).map(String::toLowerCase).orElse(null),
+        loadDocumentStoreProperties());
+  }
+
+  // For testing purposes
+  void addEnvironmentVariableOverride(final String key, final String value) {
```
nicpuppa, line 47: 🔧 I'd prefer to remove this method from the code. You can inject the map in the constructor instead, or you can use the system properties.

If you choose the second approach you need to modify the `getEnvVariable()` with a fallback on the system properties, like:

```
private Optional<String> getEnvVariable(final String name) {
  return Optional.ofNullable(environmentVariableOverrides.get(name))
      .or(() -> Optional.ofNullable(System.getenv(name)))
      .or(() -> Optional.ofNullable(System.getProperty(name)));
}
```

nicpuppa (commented):
Hey @chillleader I don't have enough context to approve this PR, but I left some suggestions. Please have a look

document/store/src/main/java/io/camunda/document/store/EnvironmentConfigurationLoader.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * DOCUMENT_HASH_VERIFICATION_ENABLED=true
+ * DOCUMENT_STORE_GCP_CLASS=io.camunda.document.store.gcp.GcpDocumentStoreProvider
+ * DOCUMENT_STORE_GCP_BUCKET=my-bucket
+ * DOCUMENT_STORE_INMEMORY_CLASS=io.camunda.document.store.inmemory.InMemoryDocumentStoreProvider
+ */
+public class EnvironmentConfigurationLoader implements DocumentStoreConfigurationLoader {
```
chillleader, line 27: todo: double check if there is a similar existing approach in other Zeebe components. Potentially we will need to configure this using application properties.

document/store/src/main/java/io/camunda/document/store/inmemory/InMemoryDocumentStore.java
```diff
@@ -5,7 +5,7 @@
  * Licensed under the Camunda License 1.0. You may not use this file
  * except in compliance with the Camunda License 1.0.
  */
-package io.camunda.document.store;
+package io.camunda.document.store.inmemory;
```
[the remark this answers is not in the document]
chillleader: Added some tests, thanks for the hint :)

sbuettner (approved):

chillleader:
Added a follow up to consider properties config: https://github.com/camunda/camunda/issues/24865

PR #24511, opened 2024-11-07
impl: compose `treePath` for incident entity

### LINKED ISSUE #24061: When exporting the incident populate the tree path fields

sdorokhova:
## Description

Implement `treePath` field for incident entity.

## Related issues

closes #24061

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/IncidentHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/IncidentHandler.java
@@ -9,8 +9,12 @@ package io.camunda.exporter.handlers;
 
 import static io.camunda.webapps.schema.descriptors.operate.template.IncidentTemplate.*;
 
+import io.camunda.exporter.cache.ExporterEntityCache;
+import io.camunda.exporter.cache.process.CachedProcessEntity;
 import io.camunda.exporter.store.BatchRequest;
 import io.camunda.exporter.utils.ExporterUtil;
+import io.camunda.exporter.utils.ProcessCacheUtil;
+import io.camunda.webapps.operate.TreePath;
 import io.camunda.webapps.schema.entities.operate.ErrorType;
 import io.camunda.webapps.schema.entities.operate.IncidentEntity;
 import io.camunda.webapps.schema.entities.operate.IncidentState;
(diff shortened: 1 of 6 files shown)


mustafadagher (approved):
LGTM! 🚀 

🔧  There's only one failing check `commitlint`, I think because `impl:` is not an accepted prefix. But it's not a mandatory check as well.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/IncidentHandler.java
```diff
@@ -127,6 +135,51 @@ public String getIndexName() {
...
+                processCache, processDefinitionPath.get(i), callingElementPath.get(i));
+        if (callActivityId.isPresent()) {
+          treePath.appendFlowNode(callActivityId.get());
+        } else {
+          LOGGER.debug(
+              "No process found in cache. TreePath won't contain proper callActivityId. processInstanceKey: {}, processDefinitionKey: {}, incidentKey: {}",
+              processInstanceKey,
+              processDefinitionPath.get(i),
+              record.getKey());
+          treePath.appendFlowNode(String.valueOf(callingElementPath.get(i)));
```
mustafadagher, line 174: ❓ how are we making any use of this `callingElementIndex` later on if we didn't find it in the cache now?

sdorokhova: As agreed, I increase the logging level to warning and will be checking how Operate code will deal with such cases, when instead of `callActivityId` we have it's callActivityIndex.

PR #24512, opened 2024-11-07
fix: allow empty operate index prefix

megglos:
## Description

Since a [static component part](https://github.com/camunda/camunda/pull/23865) was introduced to operate index names, all Operate indices were created as `operate-operate-*` by the still present importer instead of just `operate-*` as the default prefix was still set to `operate`.

Setting the prefix to an empty value was not possible as this caused the importer to fail starting as it resulted in an [unsupported template name starting with `_`](https://github.com/camunda/camunda/pull/24512/files#diff-fb8758c4356e435bc7a059595c2042f7b165d7df76e85e4ccc2b57fa2afcd675R283-R286).

See https://camunda.slack.com/archives/C06F0GLJNFM/p1730977896526089?thread_ts=1730410639.126879&cid=C06F0GLJNFM

## Checklist

- [ ] for CI changes:

## Related issues

relates #23972

### DIFF
--- a/operate/common/src/main/java/io/camunda/operate/property/OperateElasticsearchProperties.java
+++ b/operate/common/src/main/java/io/camunda/operate/property/OperateElasticsearchProperties.java
@@ -11,7 +11,7 @@ import java.util.Map;
 
 public class OperateElasticsearchProperties extends ElasticsearchProperties {
 
-  public static final String DEFAULT_INDEX_PREFIX = "operate";
+  public static final String DEFAULT_INDEX_PREFIX = "";
   private static final int DEFAULT_NUMBER_OF_SHARDS = 1;
   private static final int DEFAULT_NUMBER_OF_REPLICAS = 0;
   private static final String DEFAULT_REFRESH_INTERVAL = "1s";
@@ -27,7 +27,7 @@ public class OperateElasticsearchProperties extends ElasticsearchProperties {
     return indexPrefix;
   }
 
-  public void setIndexPrefix(String indexPrefix) {
+  public void setIndexPrefix(final String indexPrefix) {
     this.indexPrefix = indexPrefix;
   }
 
@@ -55,7 +55,7 @@ public class OperateElasticsearchProperties extends ElasticsearchProperties {
     return refreshInterval;
   }
 
-  public void setRefreshInterval(String refreshInterval) {
+  public void setRefreshInterval(final String refreshInterval) {
     this.refreshInterval = refreshInterval;
   }
 
@@ -63,7 +63,7 @@ public class OperateElasticsearchProperties extends ElasticsearchProperties {
     return numberOfShardsForIndices;
   }
 
-  public void setNumberOfShardsForIndices(Map<String, Integer> numberOfShardsForIndices) {
+  public void setNumberOfShardsForIndices(final Map<String, Integer> numberOfShardsForIndices) {
     this.numberOfShardsForIndices = numberOfShardsForIndices;
   }
 
@@ -71,7 +71,7 @@ public class OperateElasticsearchProperties extends ElasticsearchProperties {
     return numberOfReplicasForIndices;
   }
 
-  public void setNumberOfReplicasForIndices(Map<String, Integer> numberOfReplicasForIndices) {
+  public void setNumberOfReplicasForIndices(final Map<String, Integer> numberOfReplicasForIndices) {
     this.numberOfReplicasForIndices = numberOfReplicasForIndices;
   }
 }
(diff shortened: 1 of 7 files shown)


operate/schema/src/main/java/io/camunda/operate/schema/IndexSchemaValidator.java
```diff
@@ -35,22 +35,10 @@ public class IndexSchemaValidator {
...
-
-  private Set<String> getAllIndexNamesForIndex(final String index) {
-    final String indexPattern = String.format("%s-%s*", getIndexPrefix(), index);
+  private Set<String> getAllIndexNamesForIndex(final IndexDescriptor index) {
+    final String indexPattern = index.getAllVersionsIndexNameRegexPattern();
     LOGGER.debug("Getting all indices for {}", indexPattern);
-    final Set<String> indexNames = schemaManager.getIndexNames(indexPattern);
-    // since we have indices with similar names, we need to additionally filter index names
-    // e.g. task and task-variable
-    final String patternWithVersion = String.format("%s-%s-\\d.*", getIndexPrefix(), index);
```
megglos, line 46: this regex is already present when using `index.getAllVersionsIndexNameRegexPattern();`

operate/common/src/main/java/io/camunda/operate/property/OperateElasticsearchProperties.java
```diff
@@ -11,7 +11,7 @@
 
 public class OperateElasticsearchProperties extends ElasticsearchProperties {
 
-  public static final String DEFAULT_INDEX_PREFIX = "operate";
+  public static final String DEFAULT_INDEX_PREFIX = null;
```
Zelldon, line 14: ❓️maybe we set it to ""?

operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/IndexOldSchemaValidatorIT.java
```diff
@@ -260,6 +261,7 @@ private String getFullQualifiedIndexName(final IndexDescriptor index, final Stri
   }
 
   private String getFullQualifiedIndexName(final String indexNamePart, final String version) {
-    return String.format("%s-%s-%s_", operatePrefix, indexNamePart, version);
+    return String.format(
```
Zelldon, line 264: This fails when the prefix is empty or null right?

operate/schema/src/main/java/io/camunda/operate/schema/IndexSchemaValidator.java
```diff
@@ -166,6 +159,11 @@ public Map<IndexDescriptor, Set<IndexMappingProperty>> validateIndexMappings() {
     return newFields;
   }
 
+  private String getOperateIndexPattern() {
+    final String prefix = schemaManager.getIndexPrefix();
+    return (prefix != null && !prefix.isBlank() ? prefix + "-" : "") + OPERATE + "*";
```
Zelldon, line 164: "".equals(prefix) might be easier 🤔

operate/schema/src/main/java/io/camunda/operate/schema/elasticsearch/ElasticsearchSchemaManager.java
```diff
@@ -280,8 +282,10 @@ public IndexMapping getExpectedIndexFields(final IndexDescriptor indexDescriptor
   }
 
   private String settingsTemplateName() {
-    final OperateElasticsearchProperties elsConfig = operateProperties.getElasticsearch();
-    return String.format("%s_template", elsConfig.getIndexPrefix());
+    final String indexPrefix = getIndexPrefix();
```
Zelldon, line 285: I guess you checked that it is named like this?

ChrisKujawa (approved):
Thanks for fixing!  Small comments nothing blocking, i havent tried it. Just reviewing via phone :)

operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/IndexOldSchemaValidatorIT.java
megglos: yes, this test class is always using a prefix though

operate/schema/src/main/java/io/camunda/operate/schema/elasticsearch/ElasticsearchSchemaManager.java
megglos: not sure what you mean, it was named "operate_template" previously, now with the default empty prefix it will be "operate_template" still.

operate/schema/src/main/java/io/camunda/operate/schema/IndexSchemaValidator.java
megglos: yes, unless it contains whitespaces or is null :D I just took what we already use from [AbstractIndexDescriptor.java#L57](https://github.com/camunda/camunda/blob/main/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/AbstractIndexDescriptor.java#L57) 

I will actually expose and reuse that function.

operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/IndexOldSchemaValidatorIT.java
megglos: will use the same formatter here too

ghost (approved):

Issue #24297, opened 2024-11-04
Migrate post-importer to archiver


npepinpe:
# Description

The goal is to migrate the existing post-importer as a new background task to be executed inside the exporter, e.g. alongside the archiver jobs.

The post importer task should reschedule itself like the archiver jobs. For the first iteration, we can keep the hard coded delay between runs (3s), instead of making it configurable (like with the archiver task), as this is the current state of the post-importer.

To do this, we will refactor the existing archiver such that it becomes a `BackgroundTaskManager`, and the `ReschedulingTask` to accept more generic tasks. Every background task will now return as result how much work items they did, and pass to the rescheduling task what is the minimum expected amount of work to know whether or not to increase the backoff delay. For the post importer, we will keep this as 1, so we keep the same behavior as before.

```[tasklist]
### First iteration
- [ ] https://github.com/camunda/camunda/pull/24921
- [ ] https://github.com/camunda/camunda/pull/24922
- [ ] https://github.com/camunda/camunda/issues/24929
- [ ] https://github.com/camunda/camunda/pull/25067
- [ ] https://github.com/camunda/camunda/issues/24930
- [ ] https://github.com/camunda/camunda/issues/24931
- [ ] https://github.com/camunda/camunda/issues/24990
```

Zelldon:
@npepinpe should we remove this issue from https://github.com/camunda/camunda/issues/23721 and make it a separate epci? So we can close the other one already

npepinpe:
Yes, I will do that monday, had meetings all day. We still will not close this unt we have tests though (imo)

npepinpe:
Moved for posterity, previous description:

```
- It fetches all new incident updates that have been imported, but not yet processed by the post-importer (up to a maximum) (see `Fetching the incident updates`)
- With the results, it then searches for the related process instances (see `Searching for related process instances`)
- Then it updates the related instances, as well as anything on the tree path, in order to mark incidents as created or resolved. (see `Updating related process instances`).

As you can see, this is quite coupled with the importing process. So there's a couple things we need to do:

1. We need to update the complete tree path for related process instances when an incident is created.
2. We need to update the complete tree path for related process instances when an incident is resolved.

At the moment, the exporter is not handling RESOLVED records, only creation. And since we don't have an importer or post-importer queue anymore, we need a new way to track what we've processed, and what we haven't, in order to avoid doing too much work.

[...]

PR #24513, opened 2024-11-07
refactor: prepare for different types of background tasks

npepinpe:
## Description

This PR refactors the existing archiver in preparation to add different kinds of background tasks, such as the post-importer, or the one which will update the batch operation count.

## Related issues

related to #24297

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
@@ -87,7 +87,7 @@ public final class BackgroundTaskManager implements CloseableSilently {
       final String exporterId,
       final ExporterConfiguration config,
       final ExporterResourceProvider resourceProvider,
-      final io.camunda.exporter.metrics.CamundaExporterMetrics metrics,
+      final CamundaExporterMetrics metrics,
       final Logger logger) {
     final var threadFactory =
         Thread.ofPlatform()
@@ -112,7 +112,7 @@ public final class BackgroundTaskManager implements CloseableSilently {
   }
 
   private static ProcessInstancesArchiverJob createProcessInstanceJob(
-      final io.camunda.exporter.metrics.CamundaExporterMetrics metrics,
+      final CamundaExporterMetrics metrics,
       final Logger logger,
       final ExporterResourceProvider resourceProvider,
       final ArchiverRepository repository,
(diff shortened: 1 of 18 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
```diff
@@ -101,7 +106,7 @@ public static Archiver create(
   }
 
   private static ProcessInstancesArchiverJob createProcessInstanceJob(
-      final CamundaExporterMetrics metrics,
+      final io.camunda.exporter.metrics.CamundaExporterMetrics metrics,
```
rodrigo-lourenco-lopes, line 109: The prefix can be removed it still points to the same place. This probably happened automatically 🤔

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
```diff
@@ -128,7 +133,7 @@ private static ProcessInstancesArchiverJob createProcessInstanceJob(
   }
 
   private static BatchOperationArchiverJob createBatchOperationJob(
-      final CamundaExporterMetrics metrics,
+      final io.camunda.exporter.metrics.CamundaExporterMetrics metrics,
```
rodrigo-lourenco-lopes, line 136: same here

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
```diff
@@ -41,41 +41,43 @@ public ReschedulingTask(
...
-        .exceptionallyAsync(this::onArchivingError, executor)
-        .thenAcceptAsync(this::rescheduleJob, executor);
+    result
+        .thenApplyAsync(this::onWorkPerformed, executor)
+        .exceptionallyAsync(this::onError, executor)
+        .thenAcceptAsync(this::reschedule, executor);
   }
 
-  private long onBatchArchived(final int count) {
+  private long onWorkPerformed(final int count) {
```
rodrigo-lourenco-lopes, line 60: ❓ I dont know exactly the other jobs that we going to be used here besides the post archiver, but are we generalizing here so that in the future we can add another kind of background tasks?

rodrigo-lourenco-lopes (approved):
Thanks for the changes 🚀 
I just had one question below and one change.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
npepinpe: Yes, because the next kind is the `post-importer`, not `post-archiver` ;) So it doesn't archive anything, but modifies things after import.

rodrigo-lourenco-lopes: Ah I see, totally makes sense then!

PR #24519, opened 2024-11-08
fix: switch client credential cache key to clientId

### LINKED ISSUE #20471: `OAuthCredentialsCache` in Zeebe Java client does not work correctly in multi-tenancy setup
**Describe the bug**
The `OAuthCredentialsCache` in the Zeebe Java client does not work well with multi-tenancy and the Camunda Identity-based client authorization / token validation.
If you use the Java client to deploy to tenant 1 first and then try to deploy to tenant 2 (using different client credentials that are only valid for tenant 2), the first deployment request succeeds, but the second one will fail with a `PERMISSION_DENIED` error status. This is because the Zeebe Java client tries to re-use the cached OAuth token from the first request that is only authorized for tenant 1.
The underlying problem seems to be that the OAuth token audience is used as the cache key. The audience is usually the constant value `zeebe-api`, so that the cache will not be tenant-aware.

**To Reproduce**
The easiest way to reproduce the problem is to run Camunda 8 Self-Managed with [Docker Compose](https://github.com/camunda/camunda-platform) and deploy from Web Modeler.
1. Enable multi-tenancy (`MULTI_TENANCY_ENABLED=true`)
1. Enable client authorization (`ZEEBE_AUTHENTICATION_MODE=identity`)
1. Start Camunda 8 (including Web Modeler)
1. Create two tenants in Identity: `tenant-1` and `tenant-2`
1. Create two M2M applications in Identity: `app-tenant-1` and `app-tenant-2`
    - Assign `app-tenant-1` to `tenant-1` and assign permissions for the Zeebe API
    - Assign `app-tenant-2` to `tenant-2` and assign permissions for the Zeebe API
1. Create a new BPMN diagram in Web Modeler
1. Deploy to `tenant-1` with the client credentials for `app-tenant-1` -> deployment succeeds
1. Try to deploy to `tenant-2` with the client credentials for `app-tenant-2` -> deployment fails with `Expected to handle gRPC request DeployResource with tenant identifier 'tenant-2', but tenant is not authorized to perform this request`

deploy

**Expected behavior**
It should be possible to deploy to different tenants using different client credentials.

**Additional Context**
- related to https://jira.camunda.com/browse/SUPPORT-22768
- also related to https://jira.camunda.com/browse/SUPPORT-23752

**Environment:**
- Camunda / Zeebe Version: 8.5.2

megglos:
## Description

The previous key used was the audience, which does conflict when switching clients or even clusters, as the audience is not changing in such cases and invalid credentials are thus used, causing intermittent 401s or 403s.

## Checklist

- [ ] for CI changes:

## Related issues

closes #20471

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCache.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCache.java
@@ -57,11 +57,11 @@ public final class OAuthCredentialsCache {
   private static final ReentrantReadWriteLock.WriteLock WRITE_LOCK = READ_WRITE_LOCK.writeLock();
 
   private final File cacheFile;
-  private final AtomicReference<Map<String, OAuthCachedCredentials>> audiences;
+  private final AtomicReference<Map<String, OAuthCachedCredentials>> credentialsByClientId;
 
   public OAuthCredentialsCache(final File cacheFile) {
     this.cacheFile = cacheFile;
-    audiences = new AtomicReference<>(new HashMap<>());
+    credentialsByClientId = new AtomicReference<>(new HashMap<>());
   }
 
   public OAuthCredentialsCache readCache() throws IOException {
@@ -72,7 +72,7 @@ public final class OAuthCredentialsCache {
       }
 
       final Map<String, OAuthCachedCredentials> cache = MAPPER.readValue(cacheFile, TYPE_REFERENCE);
-      audiences.set(cache);
+      credentialsByClientId.set(cache);
     } finally {
       READ_LOCK.unlock();
     }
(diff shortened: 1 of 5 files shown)


megglos:
As this changes the key value used, I guess we will have some old entries orphaned in the cache file, which never get removed. But I don't see a real impact by that and would not invest into a general cleanup, wdyt @remcowesterhoud ?

remcowesterhoud (approved):
LGTM 👍 

> As this changes the key value used, I guess we will have some old entries orphaned in the cache file, which never get removed. But I don't see a real impact by that and would not invest into a general cleanup, wdyt @remcowesterhoud ?

We shouldn't worry about this. The entries aren't orphaned for long as we don't write individual entries to the cachefile. When we write something to this file, [the entire cache is written again](https://github.com/camunda/camunda/blob/main/clients/java/src/main/java/io/camunda/zeebe/client/impl/oauth/OAuthCredentialsCache.java#L83-L98). So if a client is updated to this version, it'll be restarted, the in-memory cache is empty and the key won't exist in the cache file. At that point we will write to the cache and all orphans are removed.

remcowesterhoud:
@megglos Do you still have time to take care of the backports before your FTO?

megglos:
yes I'm taking care of them, thanks for the ping!

PR #24522, opened 2024-11-08
ci: remove operate ci-test-summary

kellervater:
## Description

This PR can only be merged when the required "Operate CI test summary" setting is removed from GitHub.

## Checklist

- [ ] for CI changes:

## Related issues

relates to https://github.com/camunda/team-infrastructure/issues/676


kellervater:
I think the e2e test failures aren't related. But I can investigate

cmur2 (changes requested):
Left a couple of comments, nothing major

> I think the e2e test failures aren't related. But I can investigate

I don't see how your PR would cause those so no investigation needed

Sidenote: You didn't copy the logic of https://github.com/camunda/camunda/blob/75bb7b2e7b12adc65318ab59417010c25b934e56/.github/workflows/operate-ci-build-reusable.yml#L78 which I agree with since it appears unused to me, or did you find out what it is there for? If not, maybe you could clean it up in one go as a pathfinder? :)

cmur2:
> someone forgot to actually create the slack-post step.

File git history might show that this step got removed at some point but the secret got overlooked 🤔

.github/actions/paths-filter/action.yml
```diff
@@ -29,6 +32,14 @@ outputs:
         steps.filter-common.outputs.github-actions-change == 'true' ||
         steps.filter-identity.outputs.identity-frontend-change == 'true'
       }}
+  operate-backend-changes:
+    description: Output whether Operate backend tests should be run based on GitHub event and files changed
+    value: >-
+      ${{
+        github.event_name == 'push' ||
+        steps.filter-common.outputs.github-actions-change == 'true' ||
```
cmur2, line 40: Sorry for not spotting earlier: We should also check for `steps.filter-common.outputs.maven-change == 'true'` since this output is meant to influence (java) backend builds using Maven

.github/actions/paths-filter/action.yml
```diff
@@ -1,7 +1,10 @@
 ---
 name: Common paths filter for camunda/camunda
 
-description: Common filters to detect and group changes against a base branch
+description: |
+  Common filters to detect and group changes against a base branch
+  It leverages the `dorny/paths-filter` action to filter paths
+  Said third-party action uses picomatch patterns to filter paths
```
cmur2, line 7: Thanks!

cmur2:
After adding the `steps.filter-common.outputs.github-actions-change == 'true'` the PR will be fine from my side.

We should also let the Operate team know and give a change to review it since they might spot something we didn't + it can affect them.

kellervater:
@cmur2 
Addressed remaining remarks.

Also [asked operate](https://camunda.slack.com/archives/C071KP5BTHB/p1731501203172549) to review the PR.

cmur2 (approved):

.github/workflows/ci.yml
```diff
@@ -406,6 +407,43 @@ jobs:
...
+          vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
+          vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
+      - name: Build Maven
+        run: |
+          ./mvnw clean deploy -B -T1C -P -docker,skipFrontendBuild -DskipTests -D skipOptimize -DskipChecks -DaltStagingDirectory=${{ github.workspace }}/staging-${{ env.BRANCH_NAME }} -DskipRemoteStaging=true -Dmaven.deploy.skip=true
+      - name: Build Docker image
+        uses: ./.github/actions/build-platform-docker
+        with:
+          repository: 'registry.camunda.cloud/team-operate/camunda-operate'
+          dockerfile: operate.Dockerfile
```
ThorbenLindhauer, line 436: In the operate-merge-ci workflow we didn't build a Docker image into the `team-operate` repository, see https://github.com/camunda/camunda/blob/c2f1e689080bcdc8955413216ec5a56aabefdc40/.github/workflows/operate-merge-ci.yaml#L21. I think we only have that for PR builds via https://github.com/camunda/camunda/blob/c2f1e689080bcdc8955413216ec5a56aabefdc40/.github/workflows/operate-ci.yml#L43 so that a Docker images of a PR branch gets built for testing.

Overall I wonder if we need this job at all in the unified CI. Apparently we don't need the Docker image part (we still need the snapshot build for main branch that is further down below). And my impression is that the Maven build is also covered by other jobs, like the java-checks one. I'm not fully certain about that as I struggle a bit with the overall complexity, but maybe we could discuss that?

We could need the job later when we add more of the actual Operate CI into the unified CI, i.e. the tests that get run via https://github.com/camunda/camunda/blob/c2f1e689080bcdc8955413216ec5a56aabefdc40/.github/workflows/operate-ci.yml.

cmur2: I think the Operate merge CI still builds the image locally (and thus verifies as part of the merge queue that it is buildable), it just does not [push](https://github.com/camunda/camunda/blob/c2f1e689080bcdc8955413216ec5a56aabefdc40/.github/workflows/operate-merge-ci.yaml#L21) it.

That is actually the same behavior as this PR since the [push input](https://github.com/camunda/camunda/blob/b50c5d56eea76f8d19b2814d35ffede66c42226f/.github/actions/build-platform-docker/action.yml#L26) is `false` by default, meaning build but no push. @kellervater explicitly passing `push: false` is redundant in https://github.com/camunda/camunda/pull/24522/commits/dea03610de5c81cb89594b3cd00ef844b62353de

cmur2: > But it seems to cause confusion when left out.

Then lets add a comment, I'd prefer to not overwrite default parameters.

ThorbenLindhauer: Thanks Christian, I agree with you reasoning about the Docker image. Let's keep it then.

cmur2:
@kellervater can you adjust https://github.com/camunda/camunda/pull/24522/commits/dea03610de5c81cb89594b3cd00ef844b62353de to not overwrite the `push` parameter default value and instead add an explanatory comment on what it is supposed to do please? Then we should be OK to merge (after changing GH settings to remove the required status check)

kellervater:
> @kellervater can you adjust https://github.com/camunda/camunda/commit/dea03610de5c81cb89594b3cd00ef844b62353de to not overwrite the push parameter default value and instead add an explanatory comment on what it is supposed to do please? Then we should be OK to merge (after changing GH settings to remove the required status check)

both done! Also squashed all commits to avoid a mess in the history.

PR #24414, opened 2024-11-06
feat: camunda services support creating a tenant

Oleksiivanov:
## Description

Support tenant creation in camunda-services and implement basic getter for getByTenantKey

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes #23137

PR #24526, opened 2024-11-08
feat: camunda services support updating a tenant

Oleksiivanov:
## Description

added updateTenant method to TenantService
added BrokerTenantUpdateRequest
added tests for new method

Note: 
can be merged only after #24414 

## Checklist

- [ ] for CI changes:

## Related issues

closes #23138

### DIFF
--- a/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantUpdateRequest.java
+++ b/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantUpdateRequest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.impl.broker.request.tenant;
+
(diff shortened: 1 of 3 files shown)


zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantUpdateRequest.java
```diff
@@ -0,0 +1,47 @@
...
+public class BrokerTenantUpdateRequest extends BrokerExecuteCommand<TenantRecord> {
+  private final TenantRecord tenantDto = new TenantRecord();
+
+  public BrokerTenantUpdateRequest(final long key) {
+    super(ValueType.TENANT, TenantIntent.UPDATE);
+    setPartitionId(Protocol.DEPLOYMENT_PARTITION);
+    request.setKey(key);
+  }
+
+  public BrokerTenantUpdateRequest setTenantId(final String tenantId) {
```
Ben-Sheppard, line 26: ❌ - the `tenantId` field should be treated as immutable (sorry I think that I didn't include that in the issue). Allowing the update of this field would mean that we would also have to handle migration of all the data within that tenant.

Only the name should be updated IMO :)

Ben-Sheppard: Yep, just checked and I added it there, I've removed it from the issue for clarity

Ben-Sheppard (changes requested):
One change otherwise looks good to me :)

Ben-Sheppard (approved):
Thank you for addressing my comment :)

PR #24533, opened 2024-11-08
feat: camunda services support deleting a tenant

### LINKED ISSUE #23139: Camunda services support deleting a tenant

Oleksiivanov:
## Description

added deleteTenant method to TenantService
added BrokerTenantDeleteRequest
added tests for new method

Note:
can be merged only after https://github.com/camunda/camunda/pull/24526

## Checklist

- [ ] for CI changes:

## Related issues

closes #23139

### DIFF
--- a/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantDeleteRequest.java
+++ b/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantDeleteRequest.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.impl.broker.request.tenant;
+
+import io.camunda.zeebe.broker.client.api.dto.BrokerExecuteCommand;
+import io.camunda.zeebe.protocol.Protocol;
+import io.camunda.zeebe.protocol.impl.record.value.tenant.TenantRecord;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.intent.TenantIntent;
+import org.agrona.DirectBuffer;
+
+public class BrokerTenantDeleteRequest extends BrokerExecuteCommand<TenantRecord> {
+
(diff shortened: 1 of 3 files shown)


zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantDeleteRequest.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private final TenantRecord requestDto = new TenantRecord();
+
+  public BrokerTenantDeleteRequest() {
+    super(ValueType.TENANT, TenantIntent.DELETE);
+    setPartitionId(Protocol.DEPLOYMENT_PARTITION);
+  }
+
+  public BrokerTenantDeleteRequest setTenantKey(final long tenantKey) {
+    request.setKey(tenantKey);
```
Ben-Sheppard, line 27: 🔧 I think this should this be `requestDto.setTenantKey()` otherwise the processor won't have anything to delete as the record will be empty

Ben-Sheppard (changes requested):

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantDeleteRequest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.agrona.DirectBuffer;
+
+public class BrokerTenantDeleteRequest extends BrokerExecuteCommand<TenantRecord> {
+
+  private final TenantRecord requestDto = new TenantRecord();
+
+  public BrokerTenantDeleteRequest() {
+    super(ValueType.TENANT, TenantIntent.DELETE);
+    setPartitionId(Protocol.DEPLOYMENT_PARTITION);
+  }
```
Ben-Sheppard, line 24: In other broker requests where we are performing an action against a resource that is in the system we set the `request` key to the key of the resource (for example in this class it would be `tenantKey`).

Can you confirm if we need to also do that here or is it ok to leave the request key empty?

Ben-Sheppard: Ah so I mean as an example [here](https://github.com/camunda/camunda/blob/main/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUpdateJobRequest.java#L25) the `request.setKey()` is used with the key of the entity we have, which I think is different to `setTenantKey` as that sets it on the record sent through. (sorry I realise that I miss spoke in my original message, I meant request, not record :)

Ben-Sheppard (approved):
Thank you for addressing my comments, let get this merged :)

PR #24539, opened 2024-11-08
Remove permission REST api and Java client

### LINKED ISSUE #22627: Support removing permissions in the Java Client
<h2>Description</h2>

Send an Authorization `REMOVE_PERMISSION` command via the Java Client. This also includes integration testing of the feature.

remcowesterhoud:
## Description

This PR adds support for permission removal through the REST api and the Java client. The biggest part of this is refactoring and testing. The endpoint for adding a permissions could be reused, so the actual implementation was very straightforward.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22627

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/RemovePermissionsCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/RemovePermissionsCommandImpl.java
@@ -17,12 +17,12 @@ package io.camunda.zeebe.client.impl.command;
 
 import io.camunda.zeebe.client.api.JsonMapper;
 import io.camunda.zeebe.client.api.ZeebeFuture;
-import io.camunda.zeebe.client.api.command.AddPermissionsCommandStep1;
-import io.camunda.zeebe.client.api.command.AddPermissionsCommandStep1.AddPermissionsCommandStep2;
-import io.camunda.zeebe.client.api.command.AddPermissionsCommandStep1.AddPermissionsCommandStep3;
-import io.camunda.zeebe.client.api.command.AddPermissionsCommandStep1.AddPermissionsCommandStep4;
 import io.camunda.zeebe.client.api.command.FinalCommandStep;
-import io.camunda.zeebe.client.api.response.AddPermissionsResponse;
+import io.camunda.zeebe.client.api.command.RemovePermissionsCommandStep1;
+import io.camunda.zeebe.client.api.command.RemovePermissionsCommandStep1.RemovePermissionsCommandStep2;
+import io.camunda.zeebe.client.api.command.RemovePermissionsCommandStep1.RemovePermissionsCommandStep3;
+import io.camunda.zeebe.client.api.command.RemovePermissionsCommandStep1.RemovePermissionsCommandStep4;
+import io.camunda.zeebe.client.api.response.RemovePermissionsResponse;
 import io.camunda.zeebe.client.impl.http.HttpClient;
 import io.camunda.zeebe.client.impl.http.HttpZeebeFuture;
 import io.camunda.zeebe.client.protocol.rest.AuthorizationPatchRequest.ActionEnum;
@@ -33,11 +33,11 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.apache.hc.client5.http.config.RequestConfig;
 
-public class AddPermissionsCommandImpl
-    implements AddPermissionsCommandStep1,
-        AddPermissionsCommandStep2,
-        AddPermissionsCommandStep3,
-        AddPermissionsCommandStep4 {
+public class RemovePermissionsCommandImpl
+    implements RemovePermissionsCommandStep1,
+        RemovePermissionsCommandStep2,
+        RemovePermissionsCommandStep3,
+        RemovePermissionsCommandStep4 {
 
   private final PatchAuthorizationCommand delegate;
   private final HttpClient httpClient;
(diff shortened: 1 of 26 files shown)


remcowesterhoud:
Can be merged after #24490

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AddPermissionsCommandImpl.java
```diff
@@ -41,49 +39,42 @@ public class AddPermissionsCommandImpl
...
 
   public AddPermissionsCommandImpl(
       final long ownerKey, final HttpClient httpClient, final JsonMapper jsonMapper) {
     this.httpClient = httpClient;
     this.jsonMapper = jsonMapper;
-    httpRequestConfig = httpClient.newRequestConfig();
-    request = new AuthorizationPatchRequest().action(ActionEnum.ADD);
     path = "/authorizations/" + ownerKey;
+    httpRequestConfig = httpClient.newRequestConfig();
+    delegate = new PatchAuthorizationCommand(ActionEnum.ADD);
```
nicpuppa, line 54: ❌ The default request timeout should be set in the constructor as well. Please have a look at other commands, like `SetVariables` or `DeployResources`

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/RemovePermissionsCommandImpl.java
```diff
@@ -0,0 +1,95 @@
...
+  private final String path;
+  private final RequestConfig.Builder httpRequestConfig;
+
+  public RemovePermissionsCommandImpl(
+      final long ownerKey, final HttpClient httpClient, final JsonMapper jsonMapper) {
+    this.httpClient = httpClient;
+    this.jsonMapper = jsonMapper;
+    path = "/authorizations/" + ownerKey;
+    httpRequestConfig = httpClient.newRequestConfig();
+    delegate = new PatchAuthorizationCommand(ActionEnum.REMOVE);
```
nicpuppa, line 54: ❌ The default request timeout should be set in the constructor as well. Please have a look at other commands, like SetVariables or DeployResources

zeebe/protocol/revapi.json
```diff
@@ -90,6 +90,30 @@
...
+          "code": "java.method.removed",
+          "old": "method io.camunda.zeebe.protocol.record.value.PermissionAction io.camunda.zeebe.protocol.record.value.ImmutableAuthorizationRecordValue::getAction()",
+          "justification": "Action is not required. The record was released with 8.6 without any way to write it. Because of this we can safely remove it."
+        },
+        {
+          "ignore": true,
+          "code": "java.method.removed",
+          "old": "method io.camunda.zeebe.protocol.record.value.ImmutableAuthorizationRecordValue io.camunda.zeebe.protocol.record.value.ImmutableAuthorizationRecordValue::withAction(io.camunda.zeebe.protocol.record.value.PermissionAction)",
+          "justification": "Action is not required. The record was released with 8.6 without any way to write it. Because of this we can safely remove it."
+        },
```
nicpuppa, line 116: ❓ Should this lines be added into the `ignored-changes.json` file ? According to https://github.com/camunda/camunda/wiki/Breaking-Changes

nicpuppa (commented):
Hey @remcowesterhoud great work, and nice commits split. I left some comments, please have a look

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AddPermissionsCommandImpl.java
remcowesterhoud: @nicpuppa Are you sure this is required? From what I can see I have a request config here which is created on line 53:

```java
httpRequestConfig = httpClient.newRequestConfig();
```

If we look in the implementation of this method:

```java
   * Creates a new request configuration builder with the default values. The builder can be used to
   * customize the request configuration for a specific request.
   *
   * @return a new request configuration builder
   */
  public RequestConfig.Builder newRequestConfig() {
    return RequestConfig.copy(defaultRequestConfig);
  }
```

This default request config is built like:

```java
  private Builder defaultClientRequestConfigBuilder() {
    return RequestConfig.custom()
        .setResponseTimeout(Timeout.of(config.getDefaultRequestTimeout()))
        // TODO: determine if the existing (gRPC) property makes sense for the HTTP client
        .setConnectionKeepAlive(TimeValue.of(config.getKeepAlive()))
        // hard cancellation may cause other requests to fail as it will kill the connection; can be
        // enabled when using HTTP/2
        .setHardCancellationEnabled(false);
  }
```

So the default request timeout is already set on this, no?

remcowesterhoud: There's a bunch of other commands where we also don' set this. Correlate message and Clock pin for example

zeebe/protocol/revapi.json
remcowesterhoud: Done

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AddPermissionsCommandImpl.java
remcowesterhoud: The config is passed along to the request. It's already on the request.

nicpuppa: Sorry, I didn't saw your comment above. I wasn't aware that we are setting the default while building the config (but it makes total sense). All good then, we should probably get rid of the duplicates in each commands implementation

nicpuppa (approved):

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AddPermissionsCommandImpl.java
remcowesterhoud: >I wasn't aware that we are setting the default while building the config

Definitely don't git blame it 😉

nicpuppa: Yeah saw that, but git blame is laying. All the files in the java client were moved when we wanted to replace the Zeebe Client with the Camunda Client, this is why almost all the files in the java client are under my name 😄

PR #24546, opened 2024-11-08
refactor: reuse common error responses

emulbreh:
## Description

This PR uses references to share the definition of common responses in the OpenAPI description.

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -61,12 +61,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/TopologyResponse"
         "500":
-          description: >
-            An internal error occurred while processing the request.
-          content:
-            application/problem+json:
-              schema:
-                $ref: "#/components/schemas/ProblemDetail"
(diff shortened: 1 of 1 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2083,13 +1845,7 @@ paths:
...
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
-        "500":
-          description: >
-            An internal error occurred while processing the request.
-          content:
-            application/problem+json:
-              schema:
-                $ref: "#/components/schemas/ProblemDetail"
+        "500": {$ref: "#/components/responses/InternalServerError"}
```
tmetzke, line 1848: 🔧 For consistency in the spec, can we use the "verbose" notation? If we want to move to the condensed format, we should do that holistically.
```suggestion
        "500":
          $ref: "#/components/responses/InternalServerError"
```

tmetzke (changes requested):
👍 This is awesome, I love it 💯 
🔧 Let's consistently use the "verbose" notation in the spec if possible.

tmetzke (approved):
🚀 Great stuff!

PR #24563, opened 2024-11-08
test: refactors some ITs to unit tests

npepinpe:
## Description

Migrates some of the repository tests that were previously ITs to be simply unit tests. This cuts down how long the tests take, and also makes them much narrower scope, so if something goes wrong, it's much faster to diagnose and fix.


rodrigo-lourenco-lopes (approved):
Thanks for extracting these 🙌

PR #24568, opened 2024-11-08
Migrate compensation boundary event subscriptions only by boundary event mapping

### LINKED ISSUE #24487: Migrate compensation boundary event subscriptions only by boundary event mapping
## Description

In the previous implementation, while migrating compensation boundary event subscriptions user required to provide mapping instructions for the activity and for any enclosing subprocesses. With this issue, we can further improve on the situation and only require a mapping between compensation boundary events. Providing mapping instructions only between compensation boundary events is enough to migrate them.

### Migrate compensation event subscription
Consider the following case where the instance waiting on service task `B`. To migrate service  task `B` to `D` and preserve the compensation event subscription after the migration, following mapping instructions should be provided:
- `B -> D`
- `boundary1 -> boundary2`

(please ignore invalid dots between activities)
Screenshot 2024-11-05 at 14 38 26

### Migrate compensation event subscription enclosed by subprocess
Consider the following case where the instance waiting on service task `B`. To migrate service  task `B` to `D` and preserve the compensation event subscription after the migration, following mapping instructions should be provided:
- `B -> D`
- `boundary1 -> boundary2`

(please ignore invalid dots between activities)
Screenshot 2024-11-08 at 12 35 15

## Validation Cases

While providing mapping instructions we need to make sure they are correctly mapped. When a mapping instruction for a task with a compensation boundary event attached is provided, we need to make sure the following:
* Target task must have a compensation boundary event attached
  * In our first example above, we need to make sure `C` has a compensation boundary event attached. 
* Source and target task should be in the same flow scope
  * In our second example above, we need to make sure `A`&`C` and  `subProcess1`&`subProcess2` is at the same level of flow scopes.

berkaycanbc:
## Description

Check related issue description and commit messages for details.

## Related issues

closes #24487

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnCompensationSubscriptionBehaviour.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnCompensationSubscriptionBehaviour.java
@@ -462,6 +462,12 @@ public class BpmnCompensationSubscriptionBehaviour {
             });
   }
 
+  public List<CompensationSubscription> getSubscriptionsByProcessInstanceKey(
+      final BpmnElementContext context) {
+    return compensationSubscriptionState.findSubscriptionsByProcessInstanceKey(
+        context.getTenantId(), context.getProcessInstanceKey());
+  }
+
   private void appendCompensationSubscriptionDeleteEvent(
       final CompensationSubscription subscription) {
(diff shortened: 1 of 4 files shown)


nicpuppa (approved):
The implementation looks good 👍

berkaycanbc:
@nicpuppa I've spotted a bug related to multi instance handlers. I've sent the fix. Please take another look. 👀

nicpuppa (approved):

PR #24570, opened 2024-11-08
Support v2 CRUD operations for users and roles from identity client

emulbreh:
## Description

This PR is a collection of changes to make viewing and managing users and roles work in the identity client.


zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/UserControllerTest.java
```diff
@@ -310,6 +312,35 @@ void deleteUserShouldReturnNoContent() {
     verify(userServices, times(1)).deleteUser(key);
   }
 
+  @Test
+  void updateUserShouldReturnNoContent() {
```
Ben-Sheppard, line 316: 🔧 Can we also add a test in here to check the 409 conflict? :)

Ben-Sheppard (commented):
Couple of comments but on the whole seems reasonable and I'd be keen to merge and refine later

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/UserControllerTest.java
emulbreh: The change here is just documenting the behaviour of the existing create endpoint, which already has a test here: https://github.com/camunda/camunda/blob/8025a375dfafb7a0b84ba4f8853e13acffb55089/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/UserControllerTest.java#L82 
It's not quite right because we actually throw a different exception, but I can fix that in a follow up. The unit tests are a bit too shallow for my taste, which makes it hard to prevent this kind of divergence.

Ben-Sheppard: Thats fine with me :)

Ben-Sheppard (approved):
Looks like there is a unit test failure, I'm going to approve early though so you are free to merge once this has been resolved

PR #24576, opened 2024-11-09
Enable OpenSearch support for Archiver

### LINKED ISSUE #24094: Implement ES/OS archiver repository

npepinpe:
## Description

This PR wires the OpenSearch repository for the archiver, allowing it pick the right implementation between ES/OS.

It also adds integration tests for `ArchiverRepository#moveDocuments` for both the ES and OS implementations.

closes #24094

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepository.java
@@ -184,7 +184,6 @@ public final class ElasticsearchRepository implements ArchiverRepository {
 
   @Override
   public void close() throws Exception {
-    // TODO: verify if I close the transport that any pending futures are cancelled
     client._transport().close();
   }
 
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepositoryIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepositoryIT.java
@@ -203,6 +203,46 @@ final class ElasticsearchRepositoryIT {
   }
 
   @Test
+  void shouldMoveDocuments() throws IOException {
+    // given
+    final var sourceIndexName = UUID.randomUUID().toString();
+    final var destIndexName = UUID.randomUUID().toString();
+    final var repository = createRepository();
+    final var documents =
+        List.of(new TestDocument("1"), new TestDocument("2"), new TestDocument("3"));
+    documents.forEach(doc -> index(sourceIndexName, doc));
+    testClient.indices().refresh(r -> r.index(sourceIndexName));
+    testClient.indices().create(r -> r.index(destIndexName));
+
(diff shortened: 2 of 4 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepository.java
```diff
@@ -184,7 +184,6 @@ public CompletableFuture<Void> reindexDocuments(
 
   @Override
   public void close() throws Exception {
-    // TODO: verify if I close the transport that any pending futures are cancelled
```
rodrigo-lourenco-lopes, line 187: ❓ You mentioned that you did a small test for this correct?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepositoryIT.java
```diff
@@ -202,6 +202,46 @@ void shouldReindexDocuments() throws IOException {
...
+    final var sourceIndexName = UUID.randomUUID().toString();
+    final var destIndexName = UUID.randomUUID().toString();
+    final var repository = createRepository();
+    final var documents =
+        List.of(new TestDocument("1"), new TestDocument("2"), new TestDocument("3"));
+    documents.forEach(doc -> index(sourceIndexName, doc));
+    testClient.indices().refresh(r -> r.index(sourceIndexName));
+    testClient.indices().create(r -> r.index(destIndexName));
+
+    // when - delete the first two documents
```
rodrigo-lourenco-lopes, line 217: here is move instead of delete, and the same for the OS test.

rodrigo-lourenco-lopes (approved):
Looks good 💯 just a comment to change

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepository.java
npepinpe: Yeah, just a quick and dirty unit test. I don't think it's worth keeping, but it was just to confirm for myself that this behaves as expected.

PR #24592, opened 2024-11-11
fix: allow empty tasklist index prefix

### LINKED ISSUE #23973: Optional Prefix in Tasklist Backend
We want to support an optional global prefix, introduced with https://github.com/camunda/camunda/issues/23638, config with https://github.com/camunda/camunda/issues/23974

We need to make sure that the component name is not overwritten by the prefix (if set). Component name needs to be part of the actual name. The prefix is optional and needs to be prepended.

megglos:
## Description

This adds empty prefix support to the tasklist index classes. Although most get removed going forward, this will allow empty prefix usage already with 8.7.0-SNAPSHOT and also make it the default while Tasklist still creates indices.

Note, I didn't migrate the Tasklist template/index classes to be based on the abstract classes from the common webapps schema module. Assuming this is not relevant as needed template/index setup will be done by the exporter going forward anyway.

This is a similar change as done to Operate with #24512.

## Checklist

- [ ] for CI changes:

## Related issues

closes #23973

### DIFF
--- a/tasklist/els-schema/src/main/java/io/camunda/tasklist/schema/indices/AbstractIndexDescriptor.java
+++ b/tasklist/els-schema/src/main/java/io/camunda/tasklist/schema/indices/AbstractIndexDescriptor.java
@@ -8,6 +8,10 @@
 package io.camunda.tasklist.schema.indices;
 
 import static io.camunda.tasklist.property.TasklistProperties.ELASTIC_SEARCH;
+import static io.camunda.webapps.schema.descriptors.AbstractIndexDescriptor.ALL_VERSIONS_INDEX_NAME_PATTERN;
+import static io.camunda.webapps.schema.descriptors.AbstractIndexDescriptor.FULL_QUALIFIED_INDEX_NAME_PATTERN;
+import static io.camunda.webapps.schema.descriptors.AbstractIndexDescriptor.formatIndexPrefix;
+import static io.camunda.webapps.schema.descriptors.ComponentNames.TASK_LIST;
 
 import io.camunda.tasklist.property.TasklistProperties;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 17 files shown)


tasklist/els-schema/src/main/java/io/camunda/tasklist/schema/indices/AbstractIndexDescriptor.java
```diff
@@ -47,6 +45,31 @@ public String getSchemaClasspathFilename() {
...
   @Override
   public String getAllVersionsIndexNameRegexPattern() {
-    return String.format("%s-%s-\\d.*", tasklistProperties.getIndexPrefix(), getIndexName());
+    return formatAllVersionsIndexNameRegexPattern(getIndexPrefix(), getIndexName());
+  }
+
+  public static String formatFullQualifiedIndexName(
+      final String indexPrefix, final String indexName, final String version) {
+    return String.format(
+        "%s%s-%s-%s_", formatIndexPrefix(indexPrefix), TASK_LIST, indexName, version);
```
Zelldon, line 54: 🔧 Should we maybe make this format `"%s%s-%s-%s_"` global accessible (static const) like the formatIndexPreifx, and reuse this everywhere?

tasklist/els-schema/src/main/java/io/camunda/tasklist/schema/indices/AbstractIndexDescriptor.java
```diff
@@ -47,6 +45,31 @@ public String getSchemaClasspathFilename() {
...
+    return formatPrefixAndComponent(indexPrefix) + "*";
+  }
+
+  public static String formatPrefixAndComponent(final String indexPrefix) {
+    return formatIndexPrefix(indexPrefix) + TASK_LIST;
+  }
+
+  public static String formatAllVersionsIndexNameRegexPattern(
+      final String prefix, final String index) {
+    return String.format("%s%s-%s-\\d.*", formatIndexPrefix(prefix), TASK_LIST, index);
```
Zelldon, line 67: 🔧 Here the same for the all version pattern

ChrisKujawa (approved):
Thanks 🚀 💪🏼 

Just one thought about moving the pattern to a common place and make it reusable for everyone, so we not always define it again.

PR #24597, opened 2024-11-11
ci: fix how we create testbench generations

lenaschoenburg:
This fixes how we authenticate and changes the image used when generating the generation to use a registry alias instead of the direct link.


than-pet (approved):
lgtm

Issue #23724, opened 2024-10-18
C8 API Variable value as free-form object


ralfpuchert:
### Description

Currently the value of a variable is a JSON object encoded and stored as String value.

Example: 
```json
{
 "name": "person"
 "value" : "{\"name\":  \"Homer\"}"
}
```

The value should be an JSON object:

```json
{
 "name": "person"
 "value" : {"name":  "Homer"}
}
```

It is called `Free-Form Object` in [OpenAPI](https://swagger.io/docs/specification/v3_0/data-models/data-types/#objects)

The value can be used in search filter and for result items:

VariableFilter:
```
    VariableFilterRequest:
      description: Variable filter request.
      type: object
      properties:
        variableKey:
          description: The key for this variable.
          type: integer
          format: int64
       ...
        value:
          description: The value of the variable.
          type: object
```

VariableItem:
```yaml
    VariableItem:
      description: Variable search response item.
      type: object
      properties:
        variableKey:
        ...
        value:
          description: Value of this variable.
          type: object
```

### TODO

- [ ] How handle this internally in API. Storage in datastore is currently as string. Can this be stored as an object?
- [ ] How to implement this and where. In service layer or conversion in controller?
- [ ] Implement as result item
- [ ] Implement as filter

### Additional Information

Raised in [Slack thread](https://camunda.slack.com/archives/C06UKS51QV9/p1728977700713929)

tmetzke:
@aleksander-dytko, this would be an improvement from the V1 variable search.
As far as I can see, this is straightforward to implement but still needs some time to implement.
I would like to put this into our last C8 API 8.7 increment (Nov 4 - Nov 24) to potentially make it to 8.7-alpha2.
It doesn't have the highest priority but it can also be tackled by anyone in the team actually, so an ideal candidate for a nice-to-have.

We can also deliver it with 8.8+ but the filter API for the variable, user task, and process instance search would need to be adjusted from `string` to `object` for the variable `value` attribute. 
This is a generalization (moving from a more specific to a more generic type), so it's backward-compatible regarding user input. However, it's not 100 percent compatible regarding generated REST API clients of all programming languages. But we can document it in a migration guide which should still be OK.

Overall, I'd like to treat this as a nice-to-have for 8.7 but we should be able to deliver it with 8.8+ as well.
@jonathanlukas, let me know if any of the above reasoning is wrong or if I overlooked anything.

tmetzke:
[...]

PR #24604, opened 2024-11-11
feat: add advanced rest string filters

### LINKED ISSUE #23993: Support `String` advanced REST filters
## Description

In order to fully support advanced search queries for `String` we need to incorporate these into our OpenAPI specification accordingly.

## Acceptance criteria

- Introduce `String` advanced filter in the OpenAPI specification.
  - Filters support:
    - The implicit (`{"key": "key"}`) and explicit (`{"key": {"$eq": "key"}}`) filter scenarios.
    - Required operations defined in the parent epic. (We can add more)
  - Create `JsonDeserializer` to support implicit filters.
- Define these filters for the required properties in `ProcessInstanceFilter`, `VariableFilter` and `UserTaskFilter`.
- Make sure underlying structures work with the new structure. (`SearchQueryRequestMapper`)
- Accommodate changes in the Java client.
- Tests are defined for the new features. (client/java, qa, gateway-rest)

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/24604
```

danielkelemen:
## Description

- Defines String advanced REST search filter in OpenAPI specification.
- Replaces required properties with advanced filters.
- Adds tests for these properties.

## Related issues

closes https://github.com/camunda/camunda/issues/23993

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/ProcessInstanceFilter.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/ProcessInstanceFilter.java
@@ -18,6 +18,8 @@ package io.camunda.zeebe.client.api.search.filter;
 import io.camunda.zeebe.client.api.search.query.TypedSearchQueryRequest.SearchRequestFilter;
 import io.camunda.zeebe.client.protocol.rest.IntegerFilterProperty;
 import io.camunda.zeebe.client.protocol.rest.LongFilterProperty;
+import io.camunda.zeebe.client.protocol.rest.ProcessInstanceStateFilterProperty;
+import io.camunda.zeebe.client.protocol.rest.StringFilterProperty;
 
 public interface ProcessInstanceFilter extends SearchRequestFilter {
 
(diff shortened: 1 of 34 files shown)


tmetzke (approved):
👍 Looks good to me, nice work 🦖 
❓ The only thing I am wondering about here is if we really need to move away from object-based variable value tests in RDBMS or if we can keep them. But as far as I can see, this is not possible with the new variable filter defined in _search-domain_?

danielkelemen:
> ❓ The only thing I am wondering about here is if we really need to move away from object-based variable value tests in RDBMS or if we can keep them. But as far as I can see, this is not possible with the new variable filter defined in _search-domain_?

@tmetzke, Yes, the `VariableFilter` value expects string operations now so that needs this.
I tried a bit to have objects on API level, too but ran into a few questions and then just went for String. I don't think it would be a huge effort to support objects though but probably needs a bit more time.

tmetzke:
> > ❓ The only thing I am wondering about here is if we really need to move away from object-based variable value tests in RDBMS or if we can keep them. But as far as I can see, this is not possible with the new variable filter defined in _search-domain_?
> 
> @tmetzke, Yes, the `VariableFilter` value expects string operations now so that needs this. I tried a bit to have objects on API level, too but ran into a few questions and then just went for String. I don't think it would be a huge effort to support objects though but probably needs a bit more time.

Fair enough, thanks for the details. Makes sense like it is for now. We can follow-up with https://github.com/camunda/camunda/issues/23724

christinaausley (approved):

PR #24626, opened 2024-11-12
Mapping created record are exported

### LINKED ISSUE #24453: Camunda exporters support mapping created events
### Description 
To be able to access the mapping information in the components we need to export the data to the downstream storage configured by the customer. Once exported, the service/search classes are able to query the data to then use in further code paths, or in the case of Identity, shown in the UI.

An example of an exporter handler can be found [here](https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserCreatedUpdatedHandler.java)

Examples of the index descriptors can be found [here](https://github.com/camunda/camunda/blob/main/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/UserIndex.java)

### Acceptance Criteria
- [x] A new `MappingIndex` class exists in the `usermanagement` package [here](https://github.com/camunda/camunda/tree/main/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index) 
- [x] The `MappingIndex` class is added to the descriptors map [here](https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java#L60)
- [x] A `camunda-mapping.json` exists [here](https://github.com/camunda/camunda/tree/main/webapps-schema/src/main/resources/schema/elasticsearch/create/index) and [here](https://github.com/camunda/camunda/tree/main/webapps-schema/src/main/resources/schema/opensearch/create/index)
- [x] A `MappingEntity` exists [here](https://github.com/camunda/camunda/tree/main/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement) with fields that match the core fields of `MappingRecord` in Zeebe 
- [x] A `MappingCreatedHandler` exists [here](https://github.com/camunda/camunda/tree/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers)
- [ ] A test is added to [here](https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java) checking the export of a tenant
- [ ] A unit test is added to check behaviours - hint: https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserCreatedUpdatedHandlerTest.java

nicpuppa:
## Description

This PR adds the support to export mapping created records

## Related issues

closes #24453

### DIFF
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/MappingIndex.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/MappingIndex.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.webapps.schema.descriptors.usermanagement.index;
+
+import io.camunda.webapps.schema.descriptors.usermanagement.UserManagementIndexDescriptor;
+
+public class MappingIndex extends UserManagementIndexDescriptor {
+  public static final String INDEX_NAME = "mappings";
+  public static final String INDEX_VERSION = "8.7.0";
+
+  public static final String MAPPING_KEY = "mappingKey";
+  public static final String CLAIM_NAME = "claimName";
+  public static final String CLAIM_VALUE = "claimValue";
+
+  public MappingIndex(final String indexPrefix, final boolean isElasticsearch) {
+    super(indexPrefix, isElasticsearch);
+  }
+
+  @Override
+  public String getVersion() {
+    return INDEX_VERSION;
+  }
+
+  @Override
+  public String getIndexName() {
+    return INDEX_NAME;
+  }
+}
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/MappingIndex.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/MappingIndex.java
@@ -13,10 +13,6 @@ public class MappingIndex extends UserManagementIndexDescriptor {
   public static final String INDEX_NAME = "mappings";
   public static final String INDEX_VERSION = "8.7.0";
 
-  public static final String MAPPING_KEY = "mappingKey";
-  public static final String CLAIM_NAME = "claimName";
-  public static final String CLAIM_VALUE = "claimValue";
-
   public MappingIndex(final String indexPrefix, final boolean isElasticsearch) {
     super(indexPrefix, isElasticsearch);
   }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MappingCreatedHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MappingCreatedHandler.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.handlers;
+
+import io.camunda.exporter.exceptions.PersistenceException;
+import io.camunda.exporter.store.BatchRequest;
+import io.camunda.webapps.schema.entities.usermanagement.MappingEntity;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.value.MappingRecordValue;
+import java.util.List;
+
+public class MappingCreatedHandler implements ExportHandler<MappingEntity, MappingRecordValue> {
+  private final String indexName;
+
+  public MappingCreatedHandler(final String indexName) {
+    this.indexName = indexName;
+  }
+
+  @Override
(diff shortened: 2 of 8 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MappingCreatedHandler.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  @Override
+  public Class<MappingEntity> getEntityType() {
+    return MappingEntity.class;
+  }
+
+  @Override
+  public boolean handlesRecord(final Record<MappingRecordValue> record) {
+    return getHandledValueType().equals(record.getValueType());
```
Ben-Sheppard, line 37: 🔧 As this handler is targeted to handle create we should also include the intent type in this check for example the user handler checks for Updated and Created ([REF](https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserCreatedUpdatedHandler.java#L40-L44))

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/MappingIndex.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.camunda.webapps.schema.descriptors.usermanagement.UserManagementIndexDescriptor;
+
+public class MappingIndex extends UserManagementIndexDescriptor {
+  public static final String INDEX_NAME = "mappings";
+  public static final String INDEX_VERSION = "8.7.0";
+
+  public static final String MAPPING_KEY = "mappingKey";
+  public static final String CLAIM_NAME = "claimName";
+  public static final String CLAIM_VALUE = "claimValue";
```
Ben-Sheppard, line 18: 💭 I think these can be removed, I used them in other handlers to follow a pattern but I'm not sure, beyond historical exporter reasons, if they are used for anything without our context here. Maybe @Zelldon could clarify?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/CamundaExporterHandlerIT.java
```diff
@@ -377,6 +378,15 @@ void shouldExportUsingOperationFromVariableDocumentHandler(
     testForOperationHandlers(handler, config, clientAdapter);
   }
 
+  @TestTemplate
+  void shouldExportUsingMappingCreatedHandler(
+      final ExporterConfiguration config, final SearchClientAdapter clientAdapter)
+      throws IOException {
+    final var handler = getHandler(config, MappingCreatedHandler.class);
+    basicAssertWhereHandlerCreatesDefaultEntity(
+        handler, config, clientAdapter, defaultRecordGenerator(handler));
```
Ben-Sheppard, line 387: 💭 I believe if you set specific intents in the handler then there is a different method you need to use here to generate the record. I believe it is [this method](https://github.com/camunda/camunda/blob/main/zeebe/protocol-test-util/src/main/java/io/camunda/zeebe/test/broker/protocol/ProtocolFactory.java#L186).

The `defaultRecordGenerator` method here uses the [base method](https://github.com/camunda/camunda/blob/main/zeebe/protocol-test-util/src/main/java/io/camunda/zeebe/test/broker/protocol/ProtocolFactory.java#L168) which selects a random intent (which is likely not to be one that your handler is configured to support)

Ben-Sheppard (commented):
Couple of comments from my side otherwise looks good :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MappingCreatedHandler.java
Ben-Sheppard: I don't believe we can, when we handle the adding of the record to the batch we don't know what the intent is at that point (we are dealing with an entity), so at that point we wouldn't know whether to add/delete (at least that was true the last time I worked with the code)

nicpuppa: If it works for `UserCreatedUpdatedHandler` can work for `MappingCreatedDeletedHandler` right ?

Ben-Sheppard: Created + updated can use the same `add` batch method method in the `flush` method, deleted needs to use the `delete` batch method on a batch request so the record is removed from the indexes.

nicpuppa: You're right, I missed it 💪

Ben-Sheppard (approved):
One question around support for intent handling but other than that it looks fine to me so I will approve early :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MappingCreatedHandler.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public Class<MappingEntity> getEntityType() {
+    return MappingEntity.class;
+  }
+
+  @Override
+  public boolean handlesRecord(final Record<MappingRecordValue> record) {
+    return getHandledValueType().equals(record.getValueType())
+        && MappingIntent.CREATED.equals(record.getIntent());
```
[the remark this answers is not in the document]
nicpuppa: There is no UPDATE for mapping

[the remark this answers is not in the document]
Ben-Sheppard: 🤦🏻 sorry I overlooked that (not sure how, I wrote the issues 😂 ) - as you were this PR looks fine :D

PR #24643, opened 2024-11-12
Improve unauthorized error messages

### LINKED ISSUE #24611: Improve unauthorized error messages
<h2>Description</h2>

Our current generic error message when someone is unauthorized is:

```
Unauthorized to perform operation 'CREATE_INSTANCE' on resource 'PROCESS_DEFINITION'
```

Whilst this seems fine at first, it's not descriptive enough for when a user requires a more specific permission. For example, when I create a process instance I need permissions to create a process instance with bpmn process id `foo`. If I'm lacking permissions for the `foo` process id this error message is not descriptive enough. Instead, we should mention `foo` in the message.

```
Unauthorized to perform operation 'CREATE_INSTANCE' on resource 'PROCESS_DEFINITION' with identifier 'foo'
```

remcowesterhoud:
## Description

The current unauthorized messages are a bit vague for users. It will let them know that they are lacking authorizations to perform action X on resource Y. However, it will not give any details regarding the resource. For example, we don't tell what the process id is if a user is unauthorized to complete a job. This is useful information for them to know, as this is the specific resource id they must provide permissions for if they run into this error.
This PR adds these details where necessary.

## Checklist

- [ ] for CI changes:

## Related issues

closes #24611

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
@@ -27,6 +27,8 @@ public final class AuthorizationCheckBehavior {
 
   public static final String UNAUTHORIZED_ERROR_MESSAGE =
       "Unauthorized to perform operation '%s' on resource '%s'";
+  public static final String UNAUTHORIZED_ERROR_MESSAGE_WITH_RESOURCE =
+      UNAUTHORIZED_ERROR_MESSAGE + " with %s";
   public static final String WILDCARD_PERMISSION = "*";
   private final AuthorizationState authorizationState;
   private final UserState userState;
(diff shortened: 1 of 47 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -27,6 +27,8 @@ public final class AuthorizationCheckBehavior {
 
   public static final String UNAUTHORIZED_ERROR_MESSAGE =
       "Unauthorized to perform operation '%s' on resource '%s'";
+  public static final String UNAUTHORIZED_ERROR_MESSAGE_WITH_RESOURCE =
```
nicpuppa, line 30: ❓ Should we create static variables for each elements ? Like UNAUTHORIZED_ROLE_ERROR_MESSAGE, UNAUTHORIZED_GROUP_ERROR_MESSAGE, instead of write every time "group name '%s'" or "role name '%s'"

nicpuppa (approved):
Overall looks great, just left a small question/suggestion. Please have a look 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
remcowesterhoud: I'm gonna leave it as is. There's already some exceptions to the rule, for example on resource deletion. I'd prefer not having all these constants for this and put the responsibility in the processors. They know what resource id they are checking, so they know what to put here.

PR #24645, opened 2024-11-12
fix: revert common datetime format and improve deserializer

### LINKED ISSUE #24218: Consistent date-time format in request & result

tmetzke:
## Description

The previously defined date time format is used for writing data and reading it. This must not change as the exported data changes compared to previous versions otherwise.

The date deserializer is made more robust so it can read the defined format but also those produced by writers that don't use the `ConnectConfiguration` but a plain `OffsetDatetime::toString` directly, using a colon-based offset format, potentially using 'Z'.

## Checklist

- [ ] for CI changes:

## Related issues

closes #24218

### DIFF
--- a/search/search-client-connect/src/main/java/io/camunda/search/connect/jackson/CustomOffsetDateTimeDeserializer.java
+++ b/search/search-client-connect/src/main/java/io/camunda/search/connect/jackson/CustomOffsetDateTimeDeserializer.java
@@ -19,9 +19,12 @@ import java.time.format.DateTimeParseException;
 
 public final class CustomOffsetDateTimeDeserializer extends JsonDeserializer<OffsetDateTime> {
 
-  private DateTimeFormatter formatter;
+  private static final DateTimeFormatter OFFSET_DATE_TIME_FORMATTER_NO_COLON =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSXXXX");
 
-  public CustomOffsetDateTimeDeserializer(DateTimeFormatter formatter) {
+  private final DateTimeFormatter formatter;
+
+  public CustomOffsetDateTimeDeserializer(final DateTimeFormatter formatter) {
     this.formatter = formatter;
   }
 
@@ -31,14 +34,26 @@ public final class CustomOffsetDateTimeDeserializer extends JsonDeserializer<Off
 
     OffsetDateTime parsedDate;
     try {
+      // parse with the defined formatter, covering ~90% of all data reads
       parsedDate = OffsetDateTime.parse(parser.getText(), formatter);
-    } catch (final DateTimeParseException exception) {
-      parsedDate =
-          ZonedDateTime.parse(
-                  parser.getText(),
-                  DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss")
-                      .withZone(ZoneId.systemDefault()))
-              .toOffsetDateTime();
+    } catch (final DateTimeParseException dtpe1) {
+      try {
+        // try with the default colon-based formatter, full seconds offset and 'Z' allowed
+        // this can be introduced by data writers that don't use the ConnectConfiguration
+        parsedDate = OffsetDateTime.parse(parser.getText());
+      } catch (final DateTimeParseException dtpe2) {
+        try {
+          // try with no colon offset, full seconds offset allowed, e.g. '...-083015'
+          parsedDate = OffsetDateTime.parse(parser.getText(), OFFSET_DATE_TIME_FORMATTER_NO_COLON);
+        } catch (final DateTimeParseException dtpe3) {
+          parsedDate =
+              ZonedDateTime.parse(
+                      parser.getText(),
+                      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss")
+                          .withZone(ZoneId.systemDefault()))
+                  .toOffsetDateTime();
+        }
+      }
     }
     return parsedDate;
   }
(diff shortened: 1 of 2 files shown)


houssain-barouni (approved):

search/search-client-connect/src/main/java/io/camunda/search/connect/jackson/CustomOffsetDateTimeDeserializer.java
```diff
@@ -31,14 +34,26 @@ public OffsetDateTime deserialize(final JsonParser parser, final Deserialization
...
-          ZonedDateTime.parse(
-                  parser.getText(),
-                  DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss")
-                      .withZone(ZoneId.systemDefault()))
-              .toOffsetDateTime();
+    } catch (final DateTimeParseException dtpe1) {
+      try {
+        // try with the default colon-based formatter, full seconds offset and 'Z' allowed
+        // this can be introduced by data writers that don't use the ConnectConfiguration
+        parsedDate = OffsetDateTime.parse(parser.getText());
```
houssain-barouni, line 43: @tmetzke I am wondering if this is already covered by `yyyy-MM-dd'T'HH:mm:ss.SSSXXXX` ?

search/search-client-connect/src/main/java/io/camunda/search/connect/jackson/CustomOffsetDateTimeDeserializer.java
```diff
@@ -19,9 +19,12 @@
 
 public final class CustomOffsetDateTimeDeserializer extends JsonDeserializer<OffsetDateTime> {
 
-  private DateTimeFormatter formatter;
+  private static final DateTimeFormatter OFFSET_DATE_TIME_FORMATTER_NO_COLON =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSXXXX");
```
houssain-barouni, line 23: @tmetzke is this formatter equal to `DateTimeFormatter.ISO_OFFSET_DATE_TIME` ?

search/search-client-connect/src/main/java/io/camunda/search/connect/jackson/CustomOffsetDateTimeDeserializer.java
tmetzke: No, the `XXXX` (4x) offset pattern is only for "no colon, full second details allowed". It cannot parse colon-based offset. See also [DateTimeFormatter docs](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/format/DateTimeFormatter.html):
> Offset X and x: This formats the offset based on the number of pattern letters. One letter outputs just the hour, such as '+01', unless the minute is non-zero in which case the minute is also output, such as '+0130'. Two letters outputs the hour and minute, without a colon, such as '+0130'. Three letters outputs the hour and minute, with a colon, such as '+01:30'. **Four letters outputs the hour and minute and optional second, without a colon, such as '+013015'**. Five letters outputs the hour and minute and optional second, with a colon, such as '+01:30:15'. Six or more letters throws IllegalArgumentException. Pattern letter 'X' (upper case) will output 'Z' when the offset to be output would be zero, whereas pattern letter 'x' (lower case) will output '+00', '+0000', or '+00:00'.

The parsing works with similar semantics, allowing **up to** seconds-detail in offsets, but less is also fine. But colons it cannot detect.

search/search-client-connect/src/main/java/io/camunda/search/connect/jackson/CustomOffsetDateTimeDeserializer.java
tmetzke: No, the closest equivalent to that formatter would be using `XXXXX` (5x), which can parse colon-based offsets with up to seconds-detail precision. But even that is not close enough, as the `ISO_OFFSET_DATE_TIME` also allows milliseconds and seconds to be left out in the date String.

tmetzke: The `ISO_OFFSET_DATE_TIME` is actually used in the `OffsetDateTime.parse(String)` method 🙂

search/search-client-connect/src/main/java/io/camunda/search/connect/jackson/CustomOffsetDateTimeDeserializer.java
danielkelemen: Yes, OffsetDateTime uses `XXXXX` (5). `yyyy-MM-dd'T'HH:mm:ss.SSS[XXXXX][XXXX]` could probably parse both.

tmetzke: I tried with that notation in local tests but this doesn't work for me. Maybe I am doing something wrong though 😄

danielkelemen: Yeah, I didn't test it too much. It fixed the IT tests locally for me.

danielkelemen: Yes, these changes work too!

Issue #24095, opened 2024-10-26
Migrate existing E2E archiver tests to the new exporter


npepinpe:
**Description**

Once all of the building blocks are in place for the new exporter's background archiver, we should migrate the existing E2E tests.

The first step is to list the existing tests here to get an idea of the coverage. After migrating them to the new exporter (or to our Camunda QA module), we should then figure out if we need increase the test coverage, or if it's sufficient.

rodrigo-lourenco-lopes:
- tasklist/`ArchiverIT`
- operate/`ArchiverZeebeIT`
- tasklist/`OneNodeArchiverIT`
- operate/`OneNodeArchiverZeebeIT`
- tasklist/`ReindexOpenSearchIT`
- tasklist/`ReindexElasticSearchIT`
- taslist/`OnlyArchiverIT`
- operate/`OnlyArchiverIT`

The last two are only testing bean configurations so they don’t need to be moved.

In the cases of `ArchiverIT` and `ArchiverZeebeIT`, or `OneNodeArchiverIT` and `OneNodeArchiverZeebeIT` these greatly overlap in the tests that perform, and can be merged after migrating.

We can also extend the tests to add some missing cases such as changing the lifecycle, and running the same tests in ES and OS.

npepinpe:
Can you explain how it's related to the archiver? I don't see it.

rodrigo-lourenco-lopes:
We can probably as a first step migrate the tests in ArchiverIT and ArchiverZeebeIT using just the operate client, this way we dont have to migrate the tasklist client.

`ArchiverZeebeIT`

- `testArchivingProcessInstances`
  - The tests deploys a process, starts instances instances every day the past 3 days and takes a day to complete the. After this it asserts the metrics for the finished instances and that the corresponding batches are archived in the corresponding order and witht the correct number of jobs. After this we make sure the instances are in the correct index and the metrics for the archiver show the correct count or PIs archived, and operations.
  - Some of these tests can probably be divided into smaller/simpler tests?
- `testArchivingBatchOperations`
  - Creates batch operations and makes sure they are archived in the correct index.
- `testArchivingDecisionInstances`
  - Not sure if I understand this test, we check if are archiving decision instances, but we deploy 2 and only check for the existence of one.
- `testArchivingOnlyOneHourOldData`
  - Checks that only archives after the waiting period, this tests would be actually much more readable without the bean instanciations.
- `testArchivingBigInstance`
[...]

PR #24666, opened 2024-11-13
refactor: ensure task manager is retry-able and test-able

npepinpe:
## Description

By passing the tasks to the manager at construction time, and not immediately starting it, we can improve how it will be handled in the exporter's open phase if that gets retried, and make sure it is retried in a safe way (e.g. do not resubmit the same task). That way, we ensure there is not the same task running twice which could cause concurrency issues.

Additionally, this will make it easier to unit and integration test, because we can now control when and how the tasks are executed, and have better introspection into the background task manager.

## Related issues

related to #24095

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
@@ -32,21 +32,21 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.Executor;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.WillCloseWhenClosed;
 import org.agrona.CloseHelper;
 import org.slf4j.Logger;
 
 public final class BackgroundTaskManager implements CloseableSilently {
-
   // TODO: consider making this configurable
   private static final int MAX_BACKGROUND_THREADS = 3;
+
   private final int partitionId;
   private final ArchiverRepository repository;
   private final Logger logger;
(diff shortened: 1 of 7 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
```diff
@@ -94,21 +117,19 @@ public static BackgroundTaskManager create(
...
-      final var batchOperationJob =
-          createBatchOperationJob(metrics, logger, resourceProvider, repository, executor);
-      archiver.start(config.getArchiver(), processInstanceJob, batchOperationJob);
-    } else {
-      archiver.start(config.getArchiver(), processInstanceJob);
+      tasks.add(createBatchOperationJob(metrics, logger, resourceProvider, repository, executor));
     }
 
-    return archiver;
+    executor.setCorePoolSize(Math.min(tasks.size(), MAX_BACKGROUND_THREADS));
```
rodrigo-lourenco-lopes, line 130: Nice work around to use a list of tasks 🙌

rodrigo-lourenco-lopes (approved):
Thanks for the changes and also the refactoring, looks better now 🚀

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
npepinpe: I updated it to actually just set the max size, and keep core size to 0. Threads time out after a minute, meaning if they have nothing to do for a minute, they will shut down.

So this means when we only have one job, it should pretty much keep the thread alive all the time. When we have two jobs, if there isn't much to do, the jobs will be interleaved, and we should still only use one thread :+1: If we have more work and more jobs, then we'll have more threads.

npepinpe: We could consider enforcing always having at minimum 1 thread anyway :shrug:

PR #24669, opened 2024-11-13
Change reported name of HealthMonitorable and refactor HealthMonitor

entangled90:
## Description

- Change the name of `ZeebePartitionHealth` and `RaftPartition`, so that the displayed name in Health functionalities is more consistent
-  refactor `HealthMonitor` to use the `HealthMonitorable`'s name instead of a separated string

Before:
```json
"health": {
    "id": "Partition-1",
    "name": "Partition-1",
    "status": "HEALTHY",
    "componentsState": "HEALTHY",
    "children": [
      {
        "id": "ZeebePartition-1",
        "name": "ZeebePartition-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "raft-partition-partition-1",
        "name": "raft-partition-partition-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "SnapshotDirector-1",
        "name": "SnapshotDirector-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "StreamProcessor-1",
        "name": "StreamProcessor-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "Exporter-1",
        "name": "Exporter-1",
        "status": "HEALTHY",
        "children": []
      }
    ]
  }
```

After:
```json
"health": {
    "id": "Partition-1",
    "name": "Partition-1",
    "status": "HEALTHY",
    "componentsState": "HEALTHY",
    "children": [
      {
        "id": "SnapshotDirector-1",
        "name": "SnapshotDirector-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "ZeebePartitionHealth-1",
        "name": "ZeebePartitionHealth-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "StreamProcessor-1",
        "name": "StreamProcessor-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "Exporter-1",
        "name": "Exporter-1",
        "status": "HEALTHY",
        "children": []
      },
      {
        "id": "RaftPartition-1",
        "name": "RaftPartition-1",
        "status": "HEALTHY",
        "children": []
      }
    ]
  }
```

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/ZeebePartition.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/ZeebePartition.java
@@ -43,7 +43,6 @@ public final class ZeebePartition extends Actor
         DiskSpaceUsageListener,
         SnapshotReplicationListener {
 
-  public static final String ACTOR_NAME_PREFIX = "Partition";
   private static final Logger LOG = Loggers.SYSTEM_LOGGER;
 
   private final StartupProcess<PartitionStartupContext> startupProcess;
@@ -83,10 +82,10 @@ public final class ZeebePartition extends Actor
 
     transition.setConcurrencyControl(actor);
 
-    actorName = buildActorName(transitionContext.getPartitionId());
+    final var partitionId = context.getPartitionId();
+    actorName = buildActorName("ZeebePartition", partitionId);
     transitionContext.setComponentHealthMonitor(
-        new CriticalComponentsHealthMonitor(
-            ZeebePartition.buildActorName(transitionContext.getPartitionId()), actor, LOG));
+        new CriticalComponentsHealthMonitor(componentName(partitionId), actor, LOG));
     zeebePartitionHealth = new ZeebePartitionHealth(transitionContext.getPartitionId(), transition);
     healthMetrics = new HealthMetrics(transitionContext.getPartitionId());
     healthMetrics.setUnhealthy();
@@ -94,8 +93,8 @@ public final class ZeebePartition extends Actor
     roleMetrics = new RoleMetrics(transitionContext.getPartitionId());
   }
 
-  public static String buildActorName(final int partitionId) {
-    return buildActorName(ACTOR_NAME_PREFIX, partitionId);
+  public static String componentName(final int partitionId) {
+    return String.format("Partition-%s", partitionId);
   }
 
   public PartitionAdminAccess getAdminAccess() {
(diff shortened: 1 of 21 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/impl/steps/ExporterDirectorPartitionTransitionStep.java
```diff
@@ -129,7 +129,7 @@ private ActorFuture<Void> openExporter(
     final ExporterDirector director =
         exporterDirectorBuilder.apply(exporterCtx, context.getExporterPhase());
 
-    context.getComponentHealthMonitor().registerComponent(director.getName(), director);
+    context.getComponentHealthMonitor().registerComponent(director);
```
entangled90, line 132: `LogStream` is not `HealthMonitorable`, so I guess this was not doing anything.

lenaschoenburg: I think you commented on the wrong line but not registering logstream is correct. I refactored the logstream a while ago to not be `HealthMonitorable` but forgot to remove the registration.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/ZeebePartition.java
```diff
@@ -82,17 +83,21 @@ public ZeebePartition(
 
     transition.setConcurrencyControl(actor);
 
-    actorName = buildActorName("ZeebePartition", transitionContext.getPartitionId());
+    actorName = buildActorName(transitionContext.getPartitionId());
     transitionContext.setComponentHealthMonitor(
         new CriticalComponentsHealthMonitor(
-            "Partition-" + transitionContext.getPartitionId(), actor, LOG));
+            ZeebePartition.buildActorName(transitionContext.getPartitionId()), actor, LOG));
```
lenaschoenburg, line 89: :upside_down_face: Might as well just use `actorName` directly here

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/ZeebePartition.java
```diff
@@ -43,6 +43,7 @@ public final class ZeebePartition extends Actor
         DiskSpaceUsageListener,
         SnapshotReplicationListener {
 
+  public static final String ACTOR_NAME_PREFIX = "Partition";
```
lenaschoenburg, line 46: :wrench: Hm, now we are mixing two things, the actor name and the name of the health monitoring component. I'd prefer if we keep the actor name as is and use a separate prefix + method to build the component name. That also avoids having to call `ZeebePartition#buildActorName` from `BrokerHealthCheckService`, instead we could call `ZeebePartition#componentName` (or a similarly named method).

lenaschoenburg (approved):
Generally looks good, thanks for the cleanup :+1: 

Please consider my comment regarding the actor naming, I'd rather not have that change.
No further review required though, feel free to merge after addressing comments.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/ZeebePartition.java
entangled90: I'll use componentName, see comment below

entangled90:
As discussed privately I rename the method `getName()` (which clashes with one in `Actor`) to `componentName` in order to have the possibility for the two of them to be different.

PR #24673, opened 2024-11-13
fix: cancel user tasks on instance modification

### LINKED ISSUE #24672: Zeebe user tasks not terminated on process instance modification
**Describe the bug**

When terminating a user task activity with implementation type `Zeebe user task` via process instance modification, the Zeebe user task is not canceled. As a result, it is not removed from downstream applications like Tasklist.

**To Reproduce**

1. Create an instance of a process with a Zeebe user task.
2. Arrive at the user task in the process instance so it is activated.
3. Using process instance modification, terminate the user task (it does not matter if the user task is re-activated as well through the modification or where tokens are added).

**Expected behavior**

* The user task is canceled. There is a `UserTask.CANCELED` event on the log stream.
* As a result, the user task is removed from Tasklist.

**Environment:**
- OS: all
- Zeebe Version: 8.5, 8.6, 8.7.0-alpha1
- Configuration: n/a

tmetzke:
## Description

When a Zeebe user task is terminated as part of a process instance modification, it must be canceled as well.

Previously, only the element instance was terminated.

## Related issues

closes #24672


mustafadagher (approved):
LGTM! 🚀

PR #24680, opened 2024-11-13
fix: throw correct error instead of InternalError

entangled90:
## Description
`InternalError` was thrown by mistake (which extends `VirtualMachineError` and could case the VM to terminate.
Correct error is now thrown


lenaschoenburg (approved):
Thanks! I'd suggest to backport it too, just to be safe.

PR #24695, opened 2024-11-13
Handle potential null aggregate

npepinpe:
## Description

There seems to be a race condition where sometimes the aggregation returned is null. This only happens when there is nothing in the indices (i.e. nothing to archive). I couldn't verify if this is actually expected (as it's not documented anywhere) - and since sometimes it's just an empty aggregation (instead of null), it's hard to say.

That said, it seems low impact, and worth just adding a check.

I also added a tiny commit, lowering the log level in `BatchOperationArchiverJob` (as it was logged often). Seemed overkill to open another PR for that :smile: 

## Checklist

- [ ] for CI changes:

## Related issues

closes #


npepinpe:
Alright, just needs to have #24666 merged first :crossed_fingers:

rodrigo-lourenco-lopes (approved):
Thanks for the additional check 🚀

PR #24704, opened 2024-11-14
feat: export zeebe components as nodes and edges to be visualized in grafana

### LINKED ISSUE #24296: Create metrics that can be used to create a tree/graph like dashboard in grafana
Create metrics or http endpoint such that they can be used to create a NodeGraph in graphana.
Prometheus is not an officially supportred data source for NodeGraph so it can be a bit tricky to implement.

~~Potentially explore other datasources (such as [Infinity](https://grafana.com/grafana/plugins/yesoreyeram-infinity-datasource/)) to see if it's worth the effort.~~

entangled90:
## Description
A new class `HealthTreeMetrics` has been added that exports the health status of each component with micrometer.

Each component is registered with an "actor-like" path (e.g.  `Broker-0/partition-1/StreamProcessor`).
Exported metrics as an example:
```
zeebe_broker_health_nodes{id="Broker-0",partition="none",path="Broker-0"} 1.0
zeebe_broker_health_nodes{id="Exporter-1",partition="1",path="Broker-0/Partition-1/Exporter-1"} 1.0
zeebe_broker_health_nodes{id="Partition-1",partition="1",path="Broker-0/Partition-1"} 1.0
zeebe_broker_health_nodes{id="Partition-1",partition="none",path="Broker-0/Partition-1"} 1.0
zeebe_broker_health_nodes{id="RaftPartition-1",partition="1",path="Broker-0/Partition-1/RaftPartition-1"} 1.0
zeebe_broker_health_nodes{id="SnapshotDirector-1",partition="1",path="Broker-0/Partition-1/SnapshotDirector-1"} 1.0
zeebe_broker_health_nodes{id="StreamProcessor-1",partition="1",path="Broker-0/Partition-1/StreamProcessor-1"} 1.0
zeebe_broker_health_nodes{id="ZeebePartitionHealth-1",partition="1",path="Broker-0/Partition-1/ZeebePartitionHealth-1"} 1.0
```

Due to [Prometheus fixed number of tags](https://docs.micrometer.io/micrometer/reference/implementations/prometheus.html#_limitation_on_same_name_with_different_set_of_tag_keys), even metrics that don't have a partition must have the "partition" label, which is filled with the value "none".

`Partition-1` is registered twice, once with `partition="none"` and once with `partition="1"` because it's being registered in the `BrokerHealthCheckComponent` and in itself.  This may look redudant but I think it's actually helpful in the dashboard when filtering via "partition=none" you can see the status of all partitions without their children.

Moreover, a callback to `meterRegistrationfailure` is called at startup so that a log line at warn  is logged if metrics are registerd with an invalid number of labels. 
> [!WARNING]
> I noticed that if we would throw an exception on meter registration,  one test would be failing because in tasklist one metric is register with either 0 labels or 2. That's why I decided to just log a line instead of throwing an exception for the moment.

An example of a grafana dashboard that I created with a "Stat timeline" visualization:
image

The visualization will be added to zeebe dashboard once this PR is in, in a followup PR
## Related issues

closes #24296


dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
```diff
@@ -137,6 +138,15 @@ private void cleanupWorkingDirectory() {
...
   }
 
+  private void prometheusRegistrySettings(final PrometheusMeterRegistry meterRegistry) {
+    meterRegistry
+        .config()
+        .onMeterRegistrationFailed(
+            (id, error) -> {
+              LOGGER.warn("Failed to register metric {}: {}", id, error);
+            });
+  }
```
npepinpe, line 148: :wrench: My understanding of Spring is this kind of code should go in a general configuration component. I assume here it's because you didn't want to apply it to the whole stack, right? Because a test fails? So if we fix the issue in tasklist, we can just apply this as a configuration component for everyone?

monitor/docker-compose.yml
```diff
@@ -36,4 +36,11 @@ services:
       - GF_SECURITY_ADMIN_PASSWORD=camunda
       - GF_USERS_ALLOW_SIGN_UP=false
 
+  zeebe:
+    image: camunda/zeebe:current-test
```
npepinpe, line 40:
```suggestion
    image: camunda/zeebe:SNAPSHOT
```

It's more portable this way in general

npepinpe: If you want an easy override, we do this in other places:

```
image: camunda/zeebe:${ZEEBE_VERSION:-SNAPSHOT}
```

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/monitoring/HealthTreeMetrics.java
```diff
@@ -0,0 +1,129 @@
...
+                    "id", component.componentName(), "path", pathOf(component.componentName()))))
+        .register(meterRegistry);
+  }
+
+  public String pathOf(final String from) {
+    final var pathElements = new LinkedList<String>();
+    pathElements.add(from);
+    var parent = relationships.get(from);
+    var i = 0;
+    while (parent != null && i++ < MAX_PATH_ITERATIONS) {
```
npepinpe, line 108: :thought_balloon: Nice foresight, but unless we have a bug, I doubt we'll reach that...though famous last words right? :smile:

zeebe/broker/src/test/java/io/camunda/zeebe/broker/system/monitoring/HealthTreeMetricsTest.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.stream.Stream;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class HealthTreeMetricsTest {
+  private MeterRegistry meterRegistry;
+  @AutoCloseResource private HealthTreeMetrics metrics;
```
npepinpe, line 31: :x: This will not work if you don't register the extension by adding `@AutoCloseResources` on the class itself.

npepinpe (approved):
:rocket: Looking forward to the final thing!

Some comments, but all minor; please address before merging.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/monitoring/HealthTreeMetrics.java
entangled90: Yes, since it's just to prevent something very bad because of a mistake in a "secondary" feature.

dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
entangled90: This part has been moved to #24780

PR #24710, opened 2024-11-14
Use multi-line string in `denied` description

korthout:
## Description

There are many ways to create multi-line strings in yaml files, but generally either `>` and `|` are used to specify the block style.

However, if `>` is just placed inside the string, then it becomes part of the string. That was not intentional, and is what this commit fixes.

## Related issues

Relates to https://github.com/camunda/camunda/pull/24637


tmetzke (approved):
⭐ Nice catch! Thanks, @korthout 🦖

PR #24714, opened 2024-11-14
feat: deployment state tracks whether all deployments are available or not

### LINKED ISSUE #24708: Deployment state tracks whether all deployments are available or not
Add a simple boolean flag to indicate whether all deployment records are available or not. It should default to `false` and offer a method to flip it to `true`.

lenaschoenburg:
Uses the `DEFAULT` column family to track whether all deployments have been stored or not. Defaults to `false` because that's the current state. Only allows flipping to `true`, not back to `false` again because that's not an anticipated usecase.

Depends on https://github.com/camunda/camunda/pull/24692
Closes #24708


remcowesterhoud (approved):
LGTM 👍

Issue #23492, opened 2024-10-14
Utilize a dedicated error type for Task Listener jobs when no retries left


ce-dmelnych:
## Description:

The [ErrorType.TASK_LISTENER_NO_RETRIES](https://github.com/camunda/camunda/blob/3015a30426d5243cdc631865c7cab4c87e31f7df/zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/ErrorType.java#L27)  error type was introduced in the `complete` Task Listener support PR https://github.com/camunda/camunda/pull/22576.

This task focuses on ensuring the new `TASK_LISTENER_NO_RETRIES` error type is used for incidents created when retries for Task Listener jobs are exhausted.

### The implementation should cover the following:
- Updating relevant engine components to use `TASK_LISTENER_NO_RETRIES` type for incidents created when Task Listener job retries are exhausted.
- Coordinating with the Operate team to ensure the new error type is recognized and handled correctly by Operate application.

### Dependencies:
Should be implemented after: https://github.com/camunda/camunda/issues/23946

PR #24717, opened 2024-11-14
feat: resolve incidents in task listeners

### LINKED ISSUE #23946: Resolving incidents with error type TASK_LISTENER_NO_RETRIES
Part of #21974 epic.

### Description
Resolving incidents with error type `TASK_LISTENER_NO_RETRIES` should update the retries and then retry the job.
The implementation should follow the same approach as #23383.

TODO

- [x] Add TASK_LISTENER_NO_RETRIES to [ErrorType enum](https://github.com/camunda/camunda/blob/main/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/ErrorType.java)
- [x] Expand tests to cover Task Listeners (see #23383)

### Acceptance Criteria

- [x] Failed Task Listeners with no retries left can be resolved via Operate
- [x] integration tests have been extended to also cover Task Listeners

#### Manual testing steps:
Note: Zeebe SNAPSHOT version required to support task listeners!

1. Run Zeebe, Operate and Tasklist
2. Deploy process with Task Listener of event type "complete" (e.g. [the one used for integration tests](https://github.com/camunda/camunda/blob/490565936577c887fc3df91a9f553ca1be087542/operate/qa/integration-tests/src/test/resources/task-listener.bpmn) ) and start an instance
3. Assign user task to yourself and complete it (it won't complete because the listener blocks it)
4. Fail the listener job that is created without retries left -> Incident should appear in Operate
5. Resolve incident via Operate UI (retry)
6. The retry count of the job should go up to 1 (from 0) and should complete normally

Hint: most of this is the same as for the implementation for Execution Listeners https://github.com/camunda/camunda/issues/23383#issuecomment-2459728833

- Versions to validate:
Operate: 8.7.0-alpha
Zeebe: 8.7.0-alpha

### :green_book: Link to the test case 

**PR**:
main: https://github.com/camunda/camunda/pull/24717

kristinkomschow:
## Description

Task Listeners will get a new error type "TASK_LISTENER_NO_RETRIES" that we need to be able to resolve like other "no retries" error types.

To support that, the following was added:
- `TASK_LISTENER_NO_RETRIES` was added as ErrorType equivalent to [EXECUTION_LISTENER_NO_RETRIES](https://github.com/camunda/camunda/pull/24140)
- Integration test to check for the error type being imported correctly (`IncidentZeebeRecordProcessorIT#shouldImportTaskListenerNoRetriesIncident()`)
- Integration Test to check the error type is being exported correctly (`IncidentHandlerTest#shouldHandleTaskListenerNoRetriesErrorType()`)
- Integration Test for resolving incidents in task listeners (`OperationZeebeIt#testResolveIncidentForTaskListener()`)

## Related issues

closes https://github.com/camunda/camunda/issues/23946

follow up to create integration tests: https://github.com/camunda/camunda/issues/25451


kristinkomschow:
Additional implementation notes on the `OperationZeebeIt#testResolveIncidentForTaskListener()` IT:
- This test creates an incident via the Zeebe API so that we get the actual error type that Zeebe creates for these jobs (which currently is the default `JOB_NO_RETRIES` error type) and resolves it via the Operate API call
- The setup for the Integration Tests was changed slightly for this:
      1. The SNAPSHOT version of Zeebe is now used again for the ITs since we are developing this feature in parallel with Zeebe and need the most recent Zeebe version to even get user task listener support
      2. The Zeebe HTTP port had to be exposed additionally because the user task commands sent to Zeebe do not work via gRCP and instead call the (v2) REST API
- The `userTaskKey` needed in the test is read from the Operate index because the API to query user tasks in Zeebe is experimental and didn't work properly when I tried it

sdorokhova (changes requested):
Production part looks good to me (`ErrorType` class) and exporter unit test.

BUT there is a problem with integration tests. We have the circular dependency here: from Camunda exporter dev side we plan to get rid of Operate importer 8.7 soon. Therefore we should not switch integration test run on SNAPSHOT.

My proposal is:
* to keep only production code and unit test changes in this PR
* test the changes manually with QA help
* create an issue to add integration tests later, those that will rely on exporter

kristinkomschow:
I removed the the commit that updated the zeebe version to the SNAPSHOT and reverted the one with the ITs that need the latest zeebe version (so I can still link to the commit in the new issue I'm creating for adding ITs). I'll create the issue for the ITs in a little bit and link it here.

kristinkomschow:
QA should test this after the related zeebe issue to actually use this error type is also merged (https://github.com/camunda/camunda/issues/23492) @ArezooAzar

kristinkomschow:
I created the issue for integration tests here: https://github.com/camunda/camunda/issues/25451

sdorokhova (approved):
LGTM!

PR #24730, opened 2024-11-14
feat: add advanced rest date-time filters

danielkelemen:
## Description

- Defines Date Time advanced REST search filter in OpenAPI specification.
- Replaces required properties with advanced filters.
- Adds tests for these properties.

## Related issues

closes https://github.com/camunda/camunda/issues/23994

### DIFF
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
@@ -11,6 +11,7 @@ import io.camunda.search.filter.Operation;
 import io.camunda.search.filter.Operator;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
@@ -34,6 +35,8 @@ public class AdvancedSearchFilterUtil {
     if (!tClass.isInstance(value)) {
       if (tClass == String.class) {
         return (T) value.toString();
+      } else if (tClass == OffsetDateTime.class) {
+        return (T) OffsetDateTime.parse((String) value);
       }
     }
 
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
@@ -12,6 +12,7 @@ import io.camunda.search.filter.Operator;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.time.OffsetDateTime;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Function;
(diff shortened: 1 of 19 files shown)


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
```diff
@@ -34,6 +35,8 @@ private static <T> T convertValue(final Class<T> tClass, final Object value) {
     if (!tClass.isInstance(value)) {
       if (tClass == String.class) {
         return (T) value.toString();
+      } else if (tClass == OffsetDateTime.class) {
+        return (T) OffsetDateTime.parse((String) value);
```
tmetzke, line 39: ❓ Just to clarify if I understand this correctly: this is trying to parse the input we get in the API for a date-time property, right? So, if I send value `"abc123"` to such an attribute, this will fail here? Is this tested in the `RestControllerTest` somehow? I can only see that we test with "valid" dates.

tmetzke (approved):
👍 Looks good to me in general 🚀 
❓ I left an understanding question and a potential testing improvement idea.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
danielkelemen: Yes, this would fail with invalid date string. Good idea to test this too and probably needs a proper error message.
With other properties it's not a problem because there Jackson already parses it into the correct value (or fails).

christinaausley (approved):

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/DecisionInstanceQueryControllerTest.java
```diff
@@ -380,6 +381,70 @@ void shouldSearchVariablesWithAdvancedFilter(
...
+    final var request = "{\"filter\": {\"evaluationDate\": []}";
+
+    // when / then
+    final var expectedResponse =
+        """
+            {
+              "type": "about:blank",
+              "title": "Bad Request",
+              "status": 400,
+              "detail": "Failed to read request",
```
tmetzke, line 428: ❓ Is that the default message from Jackson/Spring? That is quite generic and won't help the user much, I assume 🙈 Can we do any better? 😄

tmetzke: The location/field that cannot be read would help already, I assume.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtilTest.java
```diff
@@ -0,0 +1,143 @@
...
+                    .map(v -> AdvancedSearchFilterUtil.convertValue(pClass, v))
+                    .toList());
+        case EXISTS, NOT_EXISTS -> method.invoke(filter, operator.equals(Operator.EXISTS));
+        default -> method.invoke(filter, AdvancedSearchFilterUtil.convertValue(pClass, op.value()));
+      }
+    }
+    return filter;
+  }
+
+  private static Stream<Arguments> provideAdvancedLongFilterParameters() {
```
tmetzke, line 61: 🔧 Since there are other types in here as well, should we just name this `provideAdvancedFilterParameters`

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/DecisionInstanceQueryControllerTest.java
danielkelemen: Yes, that is just the default. I'll look into that, we could probably add a Spring exception handler for it.

danielkelemen: Changed it to `Request property [%s] cannot be parsed`.

PR #24735, opened 2024-11-14
fix: make role indexing and querying work

emulbreh:
We must use the standard key property name in the role index mapping to make the default sorting in FieldSortingTransformer work.

Also, we shouldn't crash when assignedMemberKeys is null.


Ben-Sheppard (approved):
Thank you for catching this @emulbreh :) - I've also spotted that this is an issue in the `identity-tenants` file too `tenantKey` used. But I can sort that in a follow up PR

PR #24741, opened 2024-11-14
Add job result corrections to protocol

### LINKED ISSUE #24712: Add `corrections` to `JobRecord`'s `result`
A task listener needs to be able to make corrections. These need to be passed along with the job completion to the engine. As a first start, we'll need to store these `corrections` in the Job `COMPLETE` command. 

To unblock future steps, this will allow the gateway to fill the corrections property of the command and the engine to process the corrections from the command.

While corrections will be a job concept, the usage will focus on task listeners. This means that the actual corrections that can be made will focus on user task terminology like `assignee`, `priority`, and `candidateGroups`. I checked the possibility to make the corrections a dynamic map that does not specify the keys it may contain, but this is not supported by our msg-pack implementation out of the box as it would have to rely on the `undeclaredProperties` of an `UnpackedObject` which are currently inaccessible. Instead, it makes more sense to hardcode the expected keys.

To allow only correcting some of the properties, and not all, [we'll need to align](https://github.com/camunda/camunda/issues/15530#issuecomment-1884437984) with the implementation used for user task updates, i.e. `changeset` and `changedAttributes`.

The corrections should be added to a result as follows:
```json
{
  "result": {
    "corrections": {
      "assignee": "string",
      "dueDate": "string",
      "followUpDate": "string",
      "candidateUsers": ["string"],
      "candidateGroups": ["string"],
      "priority": "int32"
    },
    "correctedAttributes": ["key of corrections"]
  }
}
```

korthout:
## Description

Introduces two new properties to `JobResult`:
- `correctedAttributes`: a list of strings referencing those properties of `corrections` that actually changed
- `corrections`: an object that can contains corrections, but may contain defaults. You should always refer to the `correctedAttributes` to know what was actually corrected.

This split was needed because we don't support writing `null` into values in our msgpack implementation.

Tests have been expanded to cover all cases.

An example of the new job result is:
```json
"result": {
  "denied": false,
  "correctedAttributes": [
    "assignee",
    "dueDate",
    "followUpDate",
    "candidateGroups",
    "candidateUsers",
    "priority"
  ],
  "corrections": {
    "assignee": "frodo",
    "dueDate": "today",
    "followUpDate": "tomorrow",
    "candidateGroups": ["fellowship", "eagles"],
    "candidateUsers": ["frodo", "sam", "gollum"],
    "priority": 1
  }
}
```

## Related issues

closes #24712

### DIFF
--- a/tasklist/importer-860/src/main/java/io/camunda/tasklist/zeebeimport/v860/record/value/JobRecordValueImpl.java
+++ b/tasklist/importer-860/src/main/java/io/camunda/tasklist/zeebeimport/v860/record/value/JobRecordValueImpl.java
@@ -200,7 +200,7 @@ public class JobRecordValueImpl extends RecordValueWithPayloadImpl implements Jo
   }
 
   @Override
-  public JobResultValueImpl getResult() {
+  public JobResultValue getResult() {
     // Not used by importer, dummy implementation for compiler
     return null;
   }
@@ -314,13 +314,5 @@ public class JobRecordValueImpl extends RecordValueWithPayloadImpl implements Jo
         + ", changedAttributes="
         + changedAttributes
         + '}';
-  }
-
-  public static class JobResultValueImpl implements JobResultValue {
-    @Override
-    public boolean isDenied() {
-      // Not used by importer, dummy implementation for compiler
-      return false;
-    }
   }
 }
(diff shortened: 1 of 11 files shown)


ce-dmelnych (approved):
LGTM! 👍🚀
Thanks, Nico, for the thorough explanations in the commit messages. They clearly explain the reasoning behind the decisions and make the changes easy to follow.
Great work!

korthout:
@ana-vinogradova-camunda Merging to make progress, but feel free to review after merging. Any requested changes we can make in a separate PR.

PR #24745, opened 2024-11-14
Implement setLifeCycleToAllIndexes in ES and OS.

rodrigo-lourenco-lopes:
## Description

This test adds the implementation of setLifeCycleToAllIndexes along the unit tests for OS and ES.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


rodrigo-lourenco-lopes:
> In this PR I am not sure about 2 details, if we should pass additionally the ConnectConfiguration to the repositories since now we also need the Index prefix in the repositories, or if its just better just to pass the entire exporter configuration, since now we are passing 3 configurations there, all from the exporter config.
The second question is that if in the both cases where we apply policies, setIndexLifeCycle() and setLifeCycleToAllIndexes() if we should make sure that the policy exists (and create if it does not)?

after discussing the topic, we decided that `setIndexLifeCycle()` and `setLifeCycleToAllIndexes()` should not check if policy exists as this gets created before synchronously.
As for the `ConnectConfiguration` there is not point passing the entire configuration if we are only using the index prefix, so we only pass this field.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/OpenSearchRepository.java
```diff
@@ -195,6 +193,45 @@ public CompletableFuture<Void> reindexDocuments(
...
+  private List<String> fetchIndexMatchingIndexes(final String indexWildCard) throws IOException {
+    final List<String> indices = new ArrayList<>();
+    client
+        .cat()
+        .indices()
+        .thenAccept(
+            response ->
+                response.valueBody().stream()
+                    .map(IndicesRecord::index)
+                    .filter(index -> index.matches(indexWildCard))
```
npepinpe, line 205: :wrench: `String#matches` will convert this to a compiled `Pattern` instance. Since our index never changes, we can already compile it ourselves once, and then just use to match.

```suggestion
  final var matcher = Pattern.compile(indexWildCard).asMatchPredicate();

  client
        .cat()
        .indices()
        .thenAccept(
            response ->
                response.valueBody().stream()
                    .map(IndicesRecord::index)
                    .filter(matcher)
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/OpenSearchRepository.java
```diff
@@ -195,6 +193,45 @@ public CompletableFuture<Void> reindexDocuments(
...
+                    .forEach(indices::add))
+        .join();
```
npepinpe, line 207: :x: Joining here will pin the background thread indefinitely. This applies as well to the ES repository version, which also joins the result of fetching the indices.

Instead, we should compose the operations:

```java
  @Override
  public CompletableFuture<Void> setLifeCycleToAllIndexes() {
    if (!retention.isEnabled()) {
      return CompletableFuture.completedFuture(null);
    }

    final var indexWildCard = "^" + indexPrefix + INDEX_WILDCARD;
    return fetchIndexMatchingIndexes(indexWildCard)
      .thenComposeAsync(indices -> setIndexLifeCycle(indices.toArray(String[]::new)), executor);
  }

  private CompletableFuture<List<String>> fetchIndexMatchingIndexes(final String indexWildCard) throws IOException {
    return client
        .cat()
        .indices()
        .thenAccept(
            response ->
                response.valueBody().stream()
                    .map(IndicesRecord::index)
                    .filter(index -> index.matches(indexWildCard))
                    .toList());

    return indices;
  }
```

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/archiver/OpenSearchRepositoryTest.java
```diff
@@ -65,6 +66,7 @@ private OpenSearchRepository createRepository() {
         1,
         new ArchiverConfiguration(),
         retention,
+        new ConnectConfiguration().getIndexPrefix(),
```
npepinpe, line 69: :wrench: You can just pass some "foo" or something here, doesn't really matter.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchRepository.java
```diff
@@ -187,6 +191,34 @@ public CompletableFuture<Void> reindexDocuments(
         .thenApplyAsync(ignored -> null, executor);
   }
 
+  private List<String> fetchMatchingIndexes(final String indexWildcard) {
+    final Pattern indexNamePattern = Pattern.compile(indexWildcard);
+    return client
+        .indices()
+        .get(new Builder().index(ALL_INDICES).build())
+        .join()
```
npepinpe, line 199: :x: Do not join to avoid pinning the thread. See the comment in the OS reposiotry class.

npepinpe (changes requested):

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
```diff
@@ -225,6 +225,7 @@ yield new ElasticsearchRepository(
             partitionId,
             config.getArchiver(),
             config.getRetention(),
+            config.getConnect(),
```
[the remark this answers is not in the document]
rodrigo-lourenco-lopes: Good catch!

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/OpenSearchRepository.java
rodrigo-lourenco-lopes: Good point all these are returning completable futures!

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/OpenSearchArchiverRepository.java
```diff
@@ -113,39 +118,29 @@ public CompletableFuture<ArchiveBatch> getBatchOperationsNextBatch() {
...
   public CompletableFuture<Void> setLifeCycleToAllIndexes() {
-    return CompletableFuture.completedFuture(null);
+    if (!retention.isEnabled()) {
+      return CompletableFuture.completedFuture(null);
+    }
+    final String indexWildCard = "^" + indexPrefix + INDEX_WILDCARD;
+
+    try {
+      return fetchIndexMatchingIndexes(indexWildCard)
+          .thenCompose(indices -> setIndexLifeCycle(indices.toArray(String[]::new)));
```
npepinpe, line 140: :x: Use `thenComposeAsync` and specify the executor. This ensures that we will skip the callback if the background task manager is closed.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchArchiverRepository.java
```diff
@@ -185,6 +188,33 @@ public CompletableFuture<Void> reindexDocuments(
         .thenApplyAsync(ignored -> null, executor);
   }
 
+  private CompletableFuture<List<String>> fetchMatchingIndexes(final String indexWildcard) {
+    final Pattern indexNamePattern = Pattern.compile(indexWildcard);
+    return client
+        .indices()
+        .get(new Builder().index(ALL_INDICES).build())
+        .thenApply(
```
npepinpe, line 196: :x: Use `thenApplyAsync` and specify the executor, that way we will not execute the callback if the background task manager is shut down.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchArchiverRepository.java
```diff
@@ -113,27 +121,22 @@ public CompletableFuture<ArchiveBatch> getBatchOperationsNextBatch() {
...
 
   @Override
   public CompletableFuture<Void> setLifeCycleToAllIndexes() {
-    return CompletableFuture.completedFuture(null);
+    if (!retention.isEnabled()) {
+      return CompletableFuture.completedFuture(null);
+    }
+    final String indexWildcard = "^" + indexPrefix + INDEX_WILDCARD;
+    return fetchMatchingIndexes(indexWildcard)
+        .thenCompose(this::setIndexLifeCycleToMatchingIndices);
```
npepinpe, line 139: :x: Use `thenComposeAsync` and specify the executor.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/OpenSearchArchiverRepository.java
```diff
@@ -195,6 +190,43 @@ public CompletableFuture<Void> reindexDocuments(
...
+  }
+
+  private CompletableFuture<Void> applyPolicyToIndex(final String index) {
+    final AddPolicyRequestBody value = new AddPolicyRequestBody(retention.getPolicyName());
+    final var request = Requests.builder().method("POST").endpoint("_plugins/_ism/add/" + index);
+    return sendRequestAsync(
+            () ->
+                genericClient.executeAsync(
+                    request.json(value, genericClient._transport().jsonpMapper()).build()))
+        .thenComposeAsync(
```
npepinpe, line 213: :x: Specify the executor as well to ensure we don't execute the callback after shutdown.

npepinpe (approved):
:rocket: 

Please address all comments before merging, but no need for another review

PR #24748, opened 2024-11-15
refactor: remove default implementation of Scheduler#schedule 

entangled90:
## Description
Default implementation of `schedule(Duration, Runnable)` (which schedule a runnable once) would call `schedule(Duration, Duration, Runnable)` (which schedules a runnable at a fixed rate).
In production files the only implementation of the interface is `SingleThreadContext` which overrides the default implementation of the interface, so there is no real behaviour change, but it still is a small cleanup in case this interface is implemented by other classes.


lenaschoenburg (approved):
Nice catch :+1:

PR #24749, opened 2024-11-15
fix: group and role name should be unique

nicpuppa:
## Description

When trying to update a group/role we should verify that the provided updated name is unique in the state, if there is another group/role with the same name a rejection is thrown

## Related issues

related to https://github.com/camunda/camunda/issues/23546


remcowesterhoud (approved):
Thanks @nicpuppa 🙇 Would you mind also doing it for Roles? I think Nikola mentioned on the other PR he changed it for Roles as well.

remcowesterhoud (approved):
Thank you! 🙇

PR #24750, opened 2024-11-15
refactor: adjust decision instance search domain classes

houssain-barouni:
## Description

As discussed in [Slack](https://camunda.slack.com/archives/C06UKS51QV9/p1731598147347449?thread_ts=1731062628.410319&cid=C06UKS51QV9)
In this pull request, I separate search domain model from database model
- `io.camunda.search.entities.*` entities are reserved for the search domain model and use the business terminology.
- `io.camunda.webapps.schema.entities.*` entities to deserialize the results from ES/OS
- transform data entity to business entity in the searchClients
- for sorting, create ES/OS transformers per entity

I applied the change only for `DecisionInstance` to have a "small" PR to review. This will be rolled out for other entities in other PRs

## Checklist

- [ ] for CI changes:

## Related issues

relates to https://github.com/camunda/camunda/issues/24911

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/ServiceTransformers.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/ServiceTransformers.java
@@ -9,6 +9,7 @@ package io.camunda.search.clients.transformers;
 
 import io.camunda.search.clients.core.SearchQueryRequest;
 import io.camunda.search.clients.query.SearchQuery;
+import io.camunda.search.clients.transformers.entity.DecisionInstanceEntityTransformer;
 import io.camunda.search.clients.transformers.filter.AuthorizationFilterTransformer;
 import io.camunda.search.clients.transformers.filter.ComparableValueFilterTransformer;
 import io.camunda.search.clients.transformers.filter.DateValueFilterTransformer;
@@ -25,12 +26,11 @@ import io.camunda.search.clients.transformers.filter.UserFilterTransformer;
 import io.camunda.search.clients.transformers.filter.UserTaskFilterTransformer;
 import io.camunda.search.clients.transformers.filter.VariableFilterTransformer;
 import io.camunda.search.clients.transformers.filter.VariableValueFilterTransformer;
-import io.camunda.search.clients.transformers.query.SearchQueryResultTransformer;
 import io.camunda.search.clients.transformers.query.TypedSearchQueryTransformer;
 import io.camunda.search.clients.transformers.result.DecisionInstanceResultConfigTransformer;
 import io.camunda.search.clients.transformers.result.DecisionRequirementsResultConfigTransformer;
 import io.camunda.search.clients.transformers.result.ProcessInstanceResultConfigTransformer;
-import io.camunda.search.clients.transformers.sort.FieldSortingTransformer;
+import io.camunda.search.clients.transformers.sort.DecisionInstanceFieldSortingTransformer;
 import io.camunda.search.filter.AuthorizationFilter;
 import io.camunda.search.filter.ComparableValueFilter;
 import io.camunda.search.filter.DateValueFilter;
@@ -56,7 +56,6 @@ import io.camunda.search.query.FormQuery;
 import io.camunda.search.query.IncidentQuery;
 import io.camunda.search.query.ProcessDefinitionQuery;
 import io.camunda.search.query.ProcessInstanceQuery;
-import io.camunda.search.query.SearchQueryResult;
 import io.camunda.search.query.TypedSearchQuery;
 import io.camunda.search.query.UserQuery;
 import io.camunda.search.query.UserTaskQuery;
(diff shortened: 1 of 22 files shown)


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/ServiceTransformers.java
```diff
@@ -56,16 +56,17 @@
...
 import io.camunda.search.query.UserQuery;
 import io.camunda.search.query.UserTaskQuery;
 import io.camunda.search.query.VariableQuery;
 import io.camunda.search.result.DecisionInstanceQueryResultConfig;
 import io.camunda.search.result.DecisionRequirementsQueryResultConfig;
 import io.camunda.search.result.ProcessInstanceQueryResultConfig;
+import io.camunda.search.sort.DecisionInstanceSort;
 import io.camunda.search.sort.FlowNodeInstanceSort;
 import io.camunda.search.sort.SortOption;
+import io.camunda.webapps.schema.entities.operate.dmn.DecisionInstanceEntity;
```
tmetzke, line 69: 🔧 Ideally, we don't have component-specific stuff in here as we want to move away from components, so anything containing `Operate` or such should be avoided.

❓ Are the transformers only used by ES/OS but not RDBMS? Just trying to understand on which level we're introducing those specific entities now.

tmetzke (commented):
At first sight, this looks like the improvement we're looking for, using the domain-specific names in the REST gateway and general search domain (or anything that is used by ES/OS and RDBMS alike).

I cannot judge all the classes involved in detail as I'm not that familiar with the code and modules involved. If this restricts ES/OS-specific mappings to classes that are not used by RDBMS, the solution is the right one in my view.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/ServiceTransformers.java
houssain-barouni: >Ideally, we don't have component-specific stuff in here as we want to move away from components, so anything containing Operate or such should be avoided.
This is the datamodel that is used by the new exporter `io.camunda:webapps-schema` :/ it seems they kept, for some reason, `operate` in the package names

>❓ Are the transformers only used by ES/OS but not RDBMS? Just trying to understand on which level we're introducing those specific entities now.
Yes these transformers are exclusive to ES/OS. RDBMS has other database model and uses other transformers (see `RdbmsSearchClient`)

tmetzke: > This is the datamodel that is used by the new exporter io.camunda:webapps-schema :/ it seems they kept, for some reason, operate in the package names

Understood, not our "fault" then, thanks for clarifying 👍 We should flag this with the exporter team though in my view. @houssain-barouni, can you ask them why they kept the component name?

houssain-barouni: it seems this will be cleaned soon https://camunda.slack.com/archives/C06F0GLJNFM/p1731667183101189?thread_ts=1731666811.834889&cid=C06F0GLJNFM

search/search-domain/src/main/java/io/camunda/search/entities/DecisionInstanceEntity.java
```diff
@@ -11,27 +11,27 @@
 import java.util.List;
 
 public record DecisionInstanceEntity(
-    long key,
-    String id,
+    Long decisionInstanceKey,
```
stoerti, line 14: 🔧  Since this behaves completely different to all other entities with a key and an ID (key is *not* the PK and not unique), should we document this here?

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/sort/DecisionInstanceFieldSortingTransformer.java
```diff
@@ -0,0 +1,50 @@
...
+  public String apply(final String domainField) {
+    return switch (domainField) {
+      case "decisionInstanceKey" -> KEY;
+      case "decisionInstanceId" -> ID;
+      case "state" -> STATE;
+      case "evaluationDate" -> EVALUATION_DATE;
+      case "evaluationFailure" -> EVALUATION_FAILURE;
+      case "processDefinitionKey" -> PROCESS_DEFINITION_KEY;
+      case "processInstanceKey" -> PROCESS_INSTANCE_KEY;
+      case "decisionDefinitionKey" -> DECISION_DEFINITION_ID; // yes, this is correct
```
stoerti, line 36: 💭 Will this be fixed somewhere?

stoerti (approved):
Looks good to me 👍

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/sort/DecisionInstanceFieldSortingTransformer.java
houssain-barouni: due to historical data, we need to stick with it, unless we do a data migration

PR #24751, opened 2024-11-15
feat: start deployment reconstruction on startup if needed

### LINKED ISSUE #24715: Start deployment reconstruction on startup if needed
Add a `StreamProcessorLifecycleAware` task that checks whether we've already reconstructed and stored all deployments or not. If not, it should write a new command that starts the reconstruction.

lenaschoenburg:
Adds a new startup task that checks whether all deployments are stored in the state and writes a new `Deployment/Reconstruct` command if not.

The new `Deployment/Reconstruct` command has a placeholder processor that writes the  `Deployment/ReconstructedAll` event to confirm the command. The applier updates the deployment state to mark all deployments as reconstructed now. This is fake as long as we don't actually reconstruct the deployments.

closes #24715

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
@@ -41,6 +41,11 @@ public final class DeploymentReconstructionStarter implements StreamProcessorLif
   @Override
   public TaskResult execute(final TaskResultBuilder taskResultBuilder) {
     LOG.debug("Not all deployments are stored, starting reconstruction");
+    // There's a chance that this command is written more than once, for example when the broker is
+    // restarted before the first command is processed.
+    // This is fine as long as the command is idempotent and does not contain any state. Do not
+    // start to add more state to the command or make the command non-idempotent unless you also
+    // find a way to prevent that this initial command is written more than once.
     taskResultBuilder.appendCommandRecord(DeploymentIntent.RECONSTRUCT, new DeploymentRecord());
     return taskResultBuilder.build();
   }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.deployment;
+
+import io.camunda.zeebe.engine.state.immutable.DeploymentState;
+import io.camunda.zeebe.stream.api.ReadonlyStreamProcessorContext;
+import io.camunda.zeebe.stream.api.StreamProcessorLifecycleAware;
+import io.camunda.zeebe.stream.api.scheduling.Task;
+import io.camunda.zeebe.stream.api.scheduling.TaskResult;
+import io.camunda.zeebe.stream.api.scheduling.TaskResultBuilder;
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DeploymentReconstructionStarter implements StreamProcessorLifecycleAware, Task {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DeploymentReconstructionStarter.class);
+  private final DeploymentState deploymentState;
+
+  public DeploymentReconstructionStarter(final DeploymentState deploymentState) {
+    this.deploymentState = deploymentState;
+  }
+
+  @Override
+  public void onRecovered(final ReadonlyStreamProcessorContext context) {
+    if (deploymentState.hasStoredAllDeployments()) {
+      LOG.trace("All deployments are already stored, skipping reconstruction");
+      return;
+    }
+
+    context.getScheduleService().runDelayed(Duration.ZERO, this);
+  }
+
+  @Override
+  public TaskResult execute(final TaskResultBuilder taskResultBuilder) {
+    LOG.debug("Not all deployments are stored, starting reconstruction");
+    return taskResultBuilder.build();
+  }
+}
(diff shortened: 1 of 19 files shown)


lenaschoenburg:
`shouldRestoreState` tests are failing because we always write the `Reconstruct` command now, which leads to new keys being generated after a simple restart. To make this pass, we already have to flip the flag to indicate that we are done reconstructing all deployments (even though that bit isn't implemented yet) to avoid redoing it again.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/DeploymentReconstructedAllApplier.java
```diff
@@ -0,0 +1,27 @@
...
+    implements TypedEventApplier<DeploymentIntent, DeploymentRecord> {
+  private final MutableDeploymentState deploymentState;
+
+  public DeploymentReconstructedAllApplier(final MutableDeploymentState deploymentState) {
+    this.deploymentState = deploymentState;
+  }
+
+  @Override
+  public void applyState(final long key, final DeploymentRecord value) {
+    deploymentState.markALlDeploymentsAsStored();
```
remcowesterhoud, line 25:
```suggestion
    deploymentState.markAllDeploymentsAsStored();
```

🤡

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
```diff
@@ -0,0 +1,47 @@
...
+
+    context.getScheduleService().runDelayed(Duration.ZERO, this);
+  }
+
+  @Override
+  public TaskResult execute(final TaskResultBuilder taskResultBuilder) {
+    LOG.debug("Not all deployments are stored, starting reconstruction");
+    taskResultBuilder.appendCommandRecord(DeploymentIntent.RECONSTRUCT, new DeploymentRecord());
+    return taskResultBuilder.build();
+  }
```
remcowesterhoud, line 46: ❓ So what is the plan here? Will you write all the reconstruct commands in this starter? That seems like a bad idea as if there's a lot of processes it could fill up the log significantly.

On the other hand you only have a `DeploymentRecord`. How will you use this to gradually reconstruct the deployment state?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
```diff
@@ -0,0 +1,47 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(DeploymentReconstructionStarter.class);
+  private final DeploymentState deploymentState;
+
+  public DeploymentReconstructionStarter(final DeploymentState deploymentState) {
+    this.deploymentState = deploymentState;
+  }
+
+  @Override
+  public void onRecovered(final ReadonlyStreamProcessorContext context) {
+    if (deploymentState.hasStoredAllDeployments()) {
```
remcowesterhoud, line 33: ❌ I don't think this check is safe? What happens if it already started the reconstruction and the broker restarts. The flag isn't set yet, as it'll get set once evrything is reconstructed. I think we need a flag to indicate if the reconstruction is started 🤔

remcowesterhoud (changes requested):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
lenaschoenburg: True, I did not think of this :+1:

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
lenaschoenburg: My plan was to do it incrementally, i.e. write the first `Reconstruct` command which writes the next one etc. That way we automatically benefit from batch processing too.
On `Deployment/Reconstructed` we can update the state based on the contained resources in the `DeploymentRecord`.

remcowesterhoud: So basically:
- Get the first resource key (whether it's a process/form/drd)
- Write the resoncstruct command for this key
- In that processor get the next resource key (whether it's a process/form/drd)
- Write a followup reconstruct command for this key

I guess you know you're done once you find a resource that contains a deployment key that isn't -1?

lenaschoenburg: > I guess you know you're done once you find a resource that contains a deployment key that isn't -1?

No, there can also be resources with a deployment key but that deployment key doesn't exist.

I think the sequence should be this: 
1. Write a Reconstruct command (no key necessary)
2. On processing, look for the next resource.
3. Find all resources with the same deployment key if it has one
4. Write a Reconstructed event with the rebuilt DeploymentRecord
5. Write the next Reconstruct command
6. If we can't find a resource without a matching deployment write the ReconstructedAll event and no Reconstruct command.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructionStarter.java
lenaschoenburg: We might still write the initial Reconstruct command, then restart before processing it and write it again. We could mark the first one as such and then reject the second one if we've already started, _but:_ we could also just ignore this? At worst, we'd write a redundant Reconstruct command. But that wouldn't cause any conflicts or inconsistencies because the Reconstruct command doesn't carry any state and ends up as a no-op if there's nothing to do?

lenaschoenburg: As discussed, this is difficult to avoid and there's no harm in writing a redundant Reconstruct command either. Let's leave it for now.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ReplayStateMachine.java
```diff
@@ -210,7 +210,7 @@ private boolean tryToReplayBatch(final Batch batch) throws Exception {
           batch.forEachRemaining(this::replayEvent);
 
           if (batchSourceEventPosition > snapshotPosition) {
-            lastProcessedPositionState.markAsProcessed(batchSourceEventPosition);
+            lastProcessedPositionState.markAsProcessed(lastReadRecordPosition);
```
remcowesterhoud, line 213: 🔧 I don't think this is supposed to be in here.

remcowesterhoud (approved):
Thanks @lenaschoenburg 🚀 

It would be great if we could somewhere in the code document that if a `Deployment.RECONSTRUCT` is written the command **must** be stateless.

zeebe/stream-platform/src/main/java/io/camunda/zeebe/stream/impl/ReplayStateMachine.java
lenaschoenburg: Oh wow, that should not have slipped in here :grimacing:

PR #24757, opened 2024-11-15
Fetch authorized tenants from the state

### LINKED ISSUE #24639: Authorized tenants are retrieved from state
### Description
In the existing code, the tenants that a particular command is authorized for are retrieved from the Identity service, in the case of GRPC this is via the `IdentityInterceptor`, in the case of rest it is via `IdentityAuthenticationManager`. 

With the new architecture tenant information will now be available directly within the engine, this means that instead of taking the information from elsewhere, we should just use the state in Zeebe.

### Acceptance criteria
- [x] Remove Identity interceptor from gateway-grpc
- [x] Remove usage of identity sdk
- [x] Remove from processors/states tenant check
- [x] Add tenant verification in AuthorizationCheckBehavior

nicpuppa:
## Description

Remove the tenant interceptor from the gateway (gRPC). Adjusted processors/states in order to retrieve tenants from the user state and not from the authorizations info in each command.

## Related issues

closes #24639


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -79,8 +87,23 @@ public boolean isAuthorized(final AuthorizationRequest request) {
...
+  private boolean isUserAuthorizedForTenant(
+      final AuthorizationRequest request, final long userKey) {
+    final var tenantId = request.tenantId;
+    if (tenantId.equals(TenantOwned.DEFAULT_TENANT_IDENTIFIER)) {
+      return true;
+    }
+
+    final var userOptional = userState.getUser(userKey);
+    if (userOptional.isEmpty()) {
+      return false;
```
nicpuppa, line 102: 💭 Not sure about this

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCancelProcessor.java
```diff
@@ -95,7 +94,8 @@ private boolean validateCommand(
         new AuthorizationRequest(
                 command,
                 AuthorizationResourceType.PROCESS_DEFINITION,
-                PermissionType.UPDATE_PROCESS_INSTANCE)
+                PermissionType.UPDATE_PROCESS_INSTANCE,
+                command.getValue().getTenantId())
```
nicpuppa, line 98: 💭 In this way we might break user space, as the rejection changes. But IMO is more appropriate to return UNATHORIZED instead of NOT_FOUND

This is valid also for other processors

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareCancelProcessInstanceTest.java
```diff
@@ -61,40 +60,6 @@ public void shouldCancelInstanceForDefaultTenant() {
         .hasIntent(ProcessInstanceIntent.ELEMENT_TERMINATED);
   }
 
-  @Test
-  public void shouldRejectCancelInstanceForUnauthorizedTenant() {
```
nicpuppa, line 65: 💭 I didn't found a way to test this with the new tenant check. If you have ideas please let me know. 
I want to add some IT test for this, same as the authorization ITs, but I need to wait until we have the tenant command on the client

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCancelProcessor.java
houssain-barouni: Tenants are rather used for data isolation, so it should behave as if the objects of a tenant do not exist for users from other tenants

nicpuppa: Yeah, it's a good point 👍 We can argue to use FORBIDDEN instead of NOT_FOUND tho

nicpuppa: At the moment we have different behaviour, in the REST API we return 401 for unauthorized tenants, in the gRPC is 403. In the processors we reject the command with NOT_FOUND

remcowesterhoud:
@nicpuppa Is it safe to say that this is not quite ready review considering we're pairing on the interceptor tomorrow and here you've removed it already?

nicpuppa:
Let's wait tomorrow's outcome @remcowesterhoud , but I think that the Identity interceptor needs to be removed

nicpuppa:
This PR needs to be reviewed after https://github.com/camunda/camunda/pull/25422 , some of the classes removed here might be still valid

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCancelProcessor.java
```diff
@@ -94,7 +94,8 @@ private boolean validateCommand(
...
```
remcowesterhoud, line 98: ❌ This is wrong. When cancelling a process instance users don't specify a tenant id on the command. Only a process instance key. What we should do here is use the tenant id that's on this process instance:

`elementInstance.getValue().getTenantId()`

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceModificationModifyProcessor.java
```diff
@@ -194,7 +194,8 @@ public void processRecord(final TypedRecord<ProcessInstanceModificationRecord> c
         new AuthorizationRequest(
                 command,
                 AuthorizationResourceType.PROCESS_DEFINITION,
-                PermissionType.UPDATE_PROCESS_INSTANCE)
+                PermissionType.UPDATE_PROCESS_INSTANCE,
+                value.getTenantId())
```
remcowesterhoud, line 198: ❌ This is wrong. When cancelling a process instance users don't specify a tenant id on the command. Only a process instance key. What we should do here is use the tenant id that's on this process instance:

`processInstance.getValue().getTenantId()`

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareCancelProcessInstanceTest.java
remcowesterhoud: The test still seems valid to me. What's the problem in running it? Is it that authorizations is not enabled on the engine? If so I'm confident that we can find a way to enable this.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareModifyProcessInstanceTest.java
```diff
@@ -72,49 +71,6 @@ public void shouldModifyInstanceForDefaultTenant() {
         .hasIntent(ProcessInstanceModificationIntent.MODIFIED);
   }
 
-  @Test
-  public void shouldRejectModifyInstanceForUnauthorizedTenant() {
```
remcowesterhoud, line 76: ❌ This test also seems relevant to me still. I wouldn't want to delete this.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/variable/VariableDocumentUpdateProcessor.java
```diff
@@ -67,7 +67,8 @@ public void processRecord(final TypedRecord<VariableDocumentRecord> record) {
         new AuthorizationRequest(
                 record,
                 AuthorizationResourceType.PROCESS_DEFINITION,
-                PermissionType.UPDATE_PROCESS_INSTANCE)
+                PermissionType.UPDATE_PROCESS_INSTANCE,
+                value.getTenantId())
```
remcowesterhoud, line 71: ❌ Tenant is not provided by users for this command. Here we should also make sure to get the tenant from the element instance.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareUpdateVariablesTest.java
```diff
@@ -63,41 +62,6 @@ public void shouldUpdateVariablesForDefaultTenant() {
         .hasIntent(VariableDocumentIntent.UPDATED);
   }
 
-  @Test
-  public void shouldRejectUpdateVariablesForUnauthorizedTenant() {
```
remcowesterhoud, line 67: ❌ Same as the other tests. This is still a valid case we should be covering.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationCheckBehaviorTest.java
```diff
@@ -158,6 +158,49 @@ public void shouldGetResourceIdentifiersWhenRoleHasPermissions() {
     assertThat(resourceIdentifiers).containsExactlyInAnyOrder(resourceId1, resourceId2);
   }
 
+  @Test
```
remcowesterhoud, line 161: ❌ Can we also add test cases in this class for the `getAuthorizedTenantIds` method?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/TenantClient.java
```diff
@@ -279,6 +279,11 @@ public TenantAddEntityClient withEntityType(final EntityType entityType) {
       return this;
     }
 
+    public TenantAddEntityClient withTenantId(final String tenantId) {
+      tenantRecord.setTenantId(tenantId);
+      return this;
+    }
```
remcowesterhoud, line 285: ❌ We shouldn't need this. When assigning an entity to a tenant it works with keys so tenant id shouldn't need to be set.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/EndpointManager.java
```diff
@@ -493,17 +490,11 @@ private <GrpcRequestT, BrokerResponseT> BrokerRequest<BrokerResponseT> mapToBrok
 
     final BrokerRequest<BrokerResponseT> brokerRequest = requestMapper.apply(grpcRequest);
 
-    final List<String> authorizedTenants =
-        multiTenancy.isEnabled()
```
remcowesterhoud, line 497: 🤡 The `multiTenancy` field is no longer used now and can be removed.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/EndpointManager.java
```diff
@@ -493,17 +490,11 @@ private <GrpcRequestT, BrokerResponseT> BrokerRequest<BrokerResponseT> mapToBrok
...
-
     final var authorizationToken =
         Authorization.jwtEncoder()
             .withIssuer(JwtAuthorizationBuilder.DEFAULT_ISSUER)
             .withAudience(JwtAuthorizationBuilder.DEFAULT_AUDIENCE)
-            .withSubject(JwtAuthorizationBuilder.DEFAULT_SUBJECT)
-            .withClaim(Authorization.AUTHORIZED_TENANTS, authorizedTenants);
+            .withSubject(JwtAuthorizationBuilder.DEFAULT_SUBJECT);
 
     // retrieve the user claims from the context and add them to the token
```
remcowesterhoud, line 499: ❓ One question I've got regarding this is how will this work with Basic auth? It seems the current `AuthenticationInterceptor` is only suited to deal with oauth. I'd expect we need to set the user key on the authorizations in that case. Similar to what we do in REST in the `RequestMapper`:

`token.withClaim(Authorization.AUTHORIZED_USER_KEY, authenticatedUserKey);`

remcowesterhoud (changes requested):
Great initial version @nicpuppa. Please don't be alarmed by all the comments, it's a large PR after all 😄 Let me know if anything is unclear.

❌ You forgot to change the `JobBatchCollector`. This class collects Jobs for activation with a user provided list of tenant ids. This needs to change to:

- Get all the authorized tenant ids from the state
- Check if all the provided tenant ids by the user are part of the list from the state
    - If not, reject the command
    - If yes, process it as it already does.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareCancelProcessInstanceTest.java
nicpuppa: Yes, that's the reason

zeebe/gateway-grpc/src/test/java/io/camunda/zeebe/gateway/impl/MultiTenancyDisabledTest.java
```diff
@@ -229,28 +200,6 @@ public void evaluateDecisionResponseHasTenantId() {
     assertThat(response.getTenantId()).isEqualTo(TenantOwned.DEFAULT_TENANT_IDENTIFIER);
   }
 
-  @Test
-  public void activateJobsRequestShouldContainDefaultTenantAsAuthorizedTenants() {
```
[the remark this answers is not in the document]
nicpuppa: I'll check

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareCancelProcessInstanceTest.java
remcowesterhoud: Happy to pair with you to get authorizations enabled in the engine rule. Just ping me 🙂

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -70,6 +72,11 @@ public boolean isAuthorized(final AuthorizationRequest request) {
       authorizedResourceIdentifiers =
           getUserAuthorizedResourceIdentifiers(
               userKey.get(), request.getResourceType(), request.getPermissionType());
+
+      // verify if the user is authorized for the tenant
+      if (!isUserAuthorizedForTenant(request, userKey.get())) {
+        return false;
```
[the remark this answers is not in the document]
nicpuppa: Yes, this applies also for the comment above about UNATHORIZED vs NOT_FOUND. I'll change the behaviour of the method to return the correct rejection error and message

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -178,6 +178,10 @@ private boolean hasRequiredPermission(
   }
 
   public List<String> getAuthorizedTenantIds(final TypedRecord<?> command) {
+    // todo: this is a temporary solution until we adjust all the tests to fetch the tenant from the state
+    if (command.getAuthorizations().get(Authorization.AUTHORIZED_TENANTS) != null) {
+      return (List<String>) command.getAuthorizations().get(Authorization.AUTHORIZED_TENANTS);
+    }
```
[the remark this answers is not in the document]
nicpuppa: I need to create a separate issue, I was waiting for feedback before doing that. I think the test refactor can be tackled later

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceCancelProcessor.java
nicpuppa: Good catch 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceModificationModifyProcessor.java
nicpuppa: Good catch 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/TenantClient.java
nicpuppa: What do you mean, we add the `tenantId` to the user

remcowesterhoud: What I would expect is that we use the key to get the tenant from the state. In the state we already have tenant id store. So basically, get the `tenantId` from the `persistedRecord`, instead of the `record` 🙂

remcowesterhoud: Looking at the API spec we won't be getting a tenant id in the command:

```yaml
  /tenants/{tenantKey}/users/{userKey}:
    put:
      tags:
        - Tenant
      operationId: assignUserToTenant
      summary: Assign a user to a tenant
      description: Assign a single user to a specified tenant.
      parameters:
        - name: tenantKey
          in: path
          required: true
          description: The unique identifier of the tenant.
          schema:
            type: integer
            format: int64
        - name: userKey
          in: path
          required: true
          description: The unique identifier of the user.
          schema:
            type: integer
            format: int64
```

nicpuppa: I'll fix the bug in the processor and remove the test helper method 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareModifyProcessInstanceTest.java
nicpuppa: Same problem with all the tenantAwareTest, the authorization is not enabled

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareUpdateVariablesTest.java
nicpuppa: Same problem with all the tenantAwareTest, the authorization is not enabled

zeebe/gateway-grpc/src/test/java/io/camunda/zeebe/gateway/impl/MultiTenancyDisabledTest.java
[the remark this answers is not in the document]
nicpuppa: We don't send the tenant in the token anymore. So this test is not valid

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/EndpointManager.java
remcowesterhoud: As much as I hate to say it, if we go that route I'd rather work with username for the REST gateway as well. That way we at least keep them aligned.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -50,36 +57,61 @@ public AuthorizationCheckBehavior(
...
   }
 
   private static Optional<Long> getUserKey(final AuthorizationRequest request) {
-    return Optional.ofNullable(
-        (Long) request.getCommand().getAuthorizations().get(Authorization.AUTHORIZED_USER_KEY));
+    return getUserKey(request.getCommand());
+  }
+
+  private boolean isUserAuthorizedForTenant(
+      final AuthorizationRequest request, final Optional<PersistedUser> userOptional) {
```
remcowesterhoud, line 103: 🔧 It's a bad practice to pass optionals as parameters. When you do this you will have always have to introduce some conditional flow in the method in case the optional is not set. Since you now pass it as a parameter to two methods it means both these methods have to deal with this potentially empty optional.

Instead what we should do is when we get this user check if we actually found it. If not we can just reject it with an unauhorized straight away. In this method and the other method that received this optional we can change it to receive the user. because we know we'll have found it at that point in time.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/EndpointManager.java
nicpuppa: I'm not sure what's the impact of changing the behaviour, it might be too much effort at the moment

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationCheckBehaviorTest.java
```diff
@@ -158,6 +158,68 @@ public void shouldGetResourceIdentifiersWhenRoleHasPermissions() {
...
+        new AuthorizationRequest(command, resourceType, permissionType, anotherTenantId)
+            .addResourceId(resourceId);
+    final var authorized = authorizationCheckBehavior.isAuthorized(request);
+
+    // then
+    assertThat(authorized.isRight()).isFalse();
+  }
+
+  @Test
+  public void shouldGetAuthorizedTenantIds() {
```
remcowesterhoud, line 205: ❌ Some scenarios aren't covered:

1. We expect an empty list if we can't find the user key on the command
2. We expect an empty list if we can't find the user in the state

remcowesterhoud (changes requested):
Thank for the changes @nicpuppa 🚀 I have a few more comments apart from the ones that are still open.

You also didn't address my comment about the `JobBatchCollector` yet 🙂

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
nicpuppa: I did that because the behaviour is not the same for the two methods, in the tenant check we return false (that means that the user is unauthorized basically). In the other one we return an empty set and then we continue with verify the permission. 
I can't reject the command if the user is not found, otherwise I'll change the behaviour of the method

remcowesterhoud: The empty set achieves the same thing. It means we have no authorized resource identifiers. That means when we do the comparison it will never have the required permissions and will always be unauthroized. Using the empty set was just a convenient way to deal with this at that time.

```java
  private boolean hasRequiredPermission(
      final Set<String> requiredResourceIdentifiers,
      final Set<String> authorizedResourceIdentifiers) {
    return authorizedResourceIdentifiers.stream().anyMatch(requiredResourceIdentifiers::contains);
  }
```

You can safely say that if there is no user, it is not authorized.

nicpuppa: I'll change it

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/EndpointManager.java
koevskinikola: > As much as I hate to say it, if we go that route I'd rather work with username for the REST gateway as well. That way we at least keep them aligned.

@remcowesterhoud I like the idea of keeping the gateways aligned. My only concern is that we will add an extra read from the Zeebe State, and we generally try to reduce State reads/lookups. 

If we replace the `userKey` with a `username` in the REST API gateway, that means an additional state lookup will be always performed with every request. 

If we don't align them, the "username workaround" will remain on gRPC only. With the deprecation of most of the gRPC calls, an additional state lookup will only be performed for job-related gRPC calls. It might be more optimal at the cost of alignment.

nicpuppa:
Just a recap on the open topics

- Add IT tests, this is actually blocked because some tenant commands are missing in the client
- Figure out how to enable the authentication in the engine for testing purpose (tenant aware tests), [see](https://github.com/camunda/camunda/pull/24757#discussion_r1850400287)
- Create a new issue to adjust all the tests to fetch tenant from the state, [see](https://github.com/camunda/camunda/pull/24757#discussion_r1869263182) 
- Create a new issue to add the username to the user claim provided by the gRPC gateway and retrieve the user in the processors with the username if the user_key is not available, [see](https://github.com/camunda/camunda/pull/24757#discussion_r1869440064)

cc @remcowesterhoud @koevskinikola @Ben-Sheppard 

Please add more if I miss something

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/multitenancy/TenantAwareCancelProcessInstanceTest.java
remcowesterhoud: @nicpuppa I made a quick example on what you need to change in the `EngineRule` to allow enabling authorizations. You can find it [here](https://github.com/camunda/camunda/commit/a4b5306ec6c1946d4e1593a713ec7949a2588a29).

With that you can enable it like:

```java
  @ClassRule
  public static final EngineRule ENGINE =
      EngineRule.singlePartition()
          .withSecurityConfig(config -> config.getAuthorizations().setEnabled(true));
```

remcowesterhoud: Note that this just a quick hack and doesn't allow you to set multiple configurations, as calling `withSecurirtyConfig` a second time would override it. But it should give you an idea.

nicpuppa:
@remcowesterhoud I re-added the tenant aware tests

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/EngineRule.java
```diff
@@ -171,6 +173,11 @@ public EngineRule withStreamProcessorMode(final StreamProcessorMode streamProces
     return this;
   }
 
+  public EngineRule withSecurityConfig(final Consumer<SecurityConfiguration> modifier) {
+    securityConfigModifier = modifier;
+    return this;
+  }
```
remcowesterhoud, line 179: 💭 Would've been nice if we could call the method multiple times without overriding the previous config. But I suppose it's good enough for now.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobBatchActivateProcessor.java
```diff
@@ -122,8 +130,8 @@ private void rejectCommand(final TypedRecord<JobBatchRecord> record) {
       rejectionType = RejectionType.INVALID_ARGUMENT;
       rejectionReason = String.format(format, "type", "present", "blank");
     } else {
-      throw new IllegalStateException(
-          "Expected to reject an invalid activate job batch command, but it appears to be valid");
+      rejectionType = RejectionType.NOT_FOUND;
+      rejectionReason = "No jobs found to activate";
```
remcowesterhoud, line 134: ❌ Before in gRPC we would return an error with the message
`"Expected to handle gRPC request ActivateJobs with tenant identifiers '%s', but tenant is not authorized to perform this request";`

This would have a `PERMISSION_DENIED_VALUE` error type.

We should try to keep this aligned as much as we can. So I'd propose if the user is not authorized for all the tenants provided in the request we return an `UNAUTHORIZED` rejection type with an error message saying that the user is not provided to activate jobs for tenant X.

remcowesterhoud: We also need to add a test case for this

remcowesterhoud (changes requested):
Almost there 😅

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/job/JobBatchActivateProcessor.java
```diff
@@ -130,8 +132,10 @@ private void rejectCommand(final TypedRecord<JobBatchRecord> record) {
       rejectionType = RejectionType.INVALID_ARGUMENT;
       rejectionReason = String.format(format, "type", "present", "blank");
     } else {
-      rejectionType = RejectionType.NOT_FOUND;
-      rejectionReason = "No jobs found to activate";
+      rejectionType = RejectionType.UNAUTHORIZED;
+      rejectionReason =
+          "User is not authorized to activate jobs for tenants '%s'"
+              .formatted(value.getTenantIds());
```
remcowesterhoud, line 138: ❌ I'm going to be annoying and refer to the [error message guidelines](https://github.com/camunda/camunda/wiki/Error-Guidelines).

```suggestion
      rejectionType = RejectionType.UNAUTHORIZED;
      rejectionReason =
          "Expected to activate job batch for tenants '%s', but user is not authorized. Authorized tenants are '%s'"
              .formatted(value.getTenantIds(), authorizedTenantIds);
```

It does mean we need to have access to the authorized tenants here. Honestly it's probably easier to do this check in a separate method returning an `Either<Rejection, JobBatchRecord>` and changing `isValid` to be the same. That way you can just chain the validations together.

remcowesterhoud (approved):
Thanks @nicpuppa, I appreciate your patience with all these comments. I have one more, sorry 🙈 

I will pre-approve the PR though.

nicpuppa:
🚀

PR #24760, opened 2024-11-15
feat: introduce advanced filter builders in client/java

### LINKED ISSUE #24534: [C8 API] Java Client uses builder pattern for advanced properties

danielkelemen:
## Description

- Add advanced rest filter property builder interfaces, implementations and callbacks.
- Update filter fields.
- Adjust tests.

## Related issues

closes https://github.com/camunda/camunda/issues/24534


clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/VariableFilter.java
```diff
@@ -29,6 +29,14 @@ public interface VariableFilter extends SearchRequestFilter {
    */
   VariableFilter variableKey(final Long key);
 
+  /**
+   * Filters variables by the specified key using {@link LongPropertyBuilderCallback}.
+   *
+   * @param callback the key {@link LongPropertyBuilderCallback} of the variable
+   * @return the updated filter
+   */
+  VariableFilter variableKey(final LongPropertyBuilderCallback callback);
```
tmetzke, line 38: 🔧 Instead of using a dedicated interface name here, we could also just define those as `Consumer<LongPropertyBuilder>`, right? That spares users one step in looking up what the interface does as `Consumer`s are a Java feature that we don't need to explain again. What do you think?

💭 We could even just name the builders `LongProperty` and such. They follow a builder pattern, which is fine, but it doesn't necessarily have to be encoded in the name again. In the end, as a user, you want to filter on, for example, a variable search, on property `variableKey`, and provide a `LongProperty`. You can do that by:
1. Provide a plain `long` value as a shortcut for `eq`.
2. Provide a complete `LongProperty` (e.g. a stored one you reuse or have built externally in an app and just hand it in).
3. Provide a `Consumer<LongProperty>` that let's you change the `LongProperty` that is handed into the consumer.

I hope that makes sense, maybe I am overlooking something 😄

tmetzke (approved):
👍 The current approach looks a lot nicer for the user space, nice job 🚀 
💭 I added some thoughts on how we could maybe make the code more readable and intuitive. I might also overlook something that makes this unusable, though 😄

clients/java/src/main/java/io/camunda/zeebe/client/api/search/filter/VariableFilter.java
danielkelemen: I was also thinking about this. I've changed the builder class names and refactored filters to just `Consumer`s. It is definitely cleaner this way.

PR #24761, opened 2024-11-15
Process distributed commands idempotently

### LINKED ISSUE #24739: Identity related processors do not process distributed commands idempotently
**Describe the bug**

Distributed command must be processed idempotently. Due to our redistribution mechanism it could receive the distributed command more than once. For example, if the acknowledgement between P2 and P1 gets lost, P2 will have processed the command. P1 didn't receive the acknowledgement and doesn't know this. It will redistribute it again. P2 should detect it already processed the command. It must reject the command and send acknowledgement again.

remcowesterhoud:
## Description

It turns out not all distributed commands were processed idempotently. This is an issue as it could result in an inifnite retry loop. This becomes a bigger issue when queues are involved, as the infinite retry causes the queue to never proceed.

This PR makes our processors that distribute idempotent. It also adds a test generic that verifies all processors which process distributed commands are idempotent.

When I was building I realized it required a migration. Any existing distributions are not queued. If we process those idempotently it could result in state differences between partitions. For this reason I added a migration task distributios of deployment creations and resource deletions. This ensures these existing distributions will now be ordered as well, meaning it is safe to make them idempotent.

## Checklist

- [ ] for CI changes:

## Related issues

closes #24739

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
@@ -30,6 +30,8 @@ import io.camunda.zeebe.stream.api.records.TypedRecord;
 import io.camunda.zeebe.stream.api.state.KeyGenerator;
 
 public class RoleAddEntityProcessor implements DistributedTypedRecordProcessor<RoleRecord> {
+  private static final String ENTITY_ALREADY_ASSIGNED_ERROR_MESSAGE =
+      "Expected to add entity with key '%s' to role with key '%s', but the entity is already assigned to this role.";
 
   private final RoleState roleState;
   private final UserState userState;
@@ -111,7 +113,19 @@ public class RoleAddEntityProcessor implements DistributedTypedRecordProcessor<R
 
   @Override
   public void processDistributedCommand(final TypedRecord<RoleRecord> command) {
-    stateWriter.appendFollowUpEvent(command.getKey(), RoleIntent.ENTITY_ADDED, command.getValue());
+    final var record = command.getValue();
+    roleState
+        .getEntityType(record.getRoleKey(), record.getEntityKey())
+        .ifPresentOrElse(
+            entityType -> {
+              final var errorMessage =
+                  ENTITY_ALREADY_ASSIGNED_ERROR_MESSAGE.formatted(
+                      record.getEntityKey(), record.getRoleKey());
+              rejectionWriter.appendRejection(command, RejectionType.INVALID_STATE, errorMessage);
+            },
+            () ->
+                stateWriter.appendFollowUpEvent(command.getKey(), RoleIntent.ENTITY_ADDED, record));
+
     commandDistributionBehavior.acknowledgeCommand(command);
   }
 
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
@@ -121,7 +121,7 @@ public class RoleAddEntityProcessor implements DistributedTypedRecordProcessor<R
               final var errorMessage =
                   ENTITY_ALREADY_ASSIGNED_ERROR_MESSAGE.formatted(
                       record.getEntityKey(), record.getRoleKey());
-              rejectionWriter.appendRejection(command, RejectionType.INVALID_STATE, errorMessage);
+              rejectionWriter.appendRejection(command, RejectionType.ALREADY_EXISTS, errorMessage);
             },
             () ->
                 stateWriter.appendFollowUpEvent(command.getKey(), RoleIntent.ENTITY_ADDED, record));
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionDeleteProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionDeleteProcessor.java
@@ -142,6 +142,15 @@ public class ResourceDeletionDeleteProcessor
       rejectionWriter.appendRejection(command, RejectionType.NOT_FOUND, exception.getMessage());
       responseWriter.writeRejectionOnCommand(
           command, RejectionType.NOT_FOUND, exception.getMessage());
+
+      // TODO verify this fix. There is a risk attached to it as deployment creation and resource
+      //  deletion were not distributed in guaranteed order before.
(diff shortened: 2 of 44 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/distribution/CommandDistributionIdempotencyTest.java
```diff
@@ -0,0 +1,621 @@
...
+    this.processor = processor;
+  }
+
+  @AfterClass
+  public static void afterClass() {
+    // TODO we need to add a scenario for this processor. I wasn't sure how to trigger it :)
+    DISTRIBUTING_PROCESSORS.remove(MessageSubscriptionMigrateProcessor.class);
+    if (!DISTRIBUTING_PROCESSORS.isEmpty()) {
+      fail("No test scenario found for processors: '%s'".formatted(DISTRIBUTING_PROCESSORS));
+    }
```
korthout, line 118: 😍

remcowesterhoud:
@lenaschoenburg would you mind checking the last commits where I've added the migration?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/migration/to_8_7/DbDistributionMigrationState8dot7.java
```diff
@@ -0,0 +1,144 @@
...
+          if (isDeploymentOrDeletion && persistedDistribution.getQueueId().isEmpty()) {
+            persistedDistribution.setQueueId(DistributionQueue.DEPLOYMENT.getQueueId());
+            commandDistributionRecordColumnFamily.update(
+                this.distributionKey, persistedDistribution);
+            queuedCommandDistributionColumnFamily.insert(queuedDistributionKey, DbNil.INSTANCE);
+
+            if (isFirstInQueue.get()) {
+              isFirstInQueue.set(false);
+            } else {
+              retriableDistributionColumnFamily.deleteExisting(distributionPartitionKey);
```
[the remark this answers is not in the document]
remcowesterhoud: The `retriableDistributionColumnFamily` is the CF we iterate over when we retry a distribution. Since we want these commands to be queued we only want to retry the first one. Only once this is ACK'd should we continue with the next command in the queue. Hence why we leave the first distribution we find in there, and remove the others from it.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/distribution/DbDistributionState.java
```diff
@@ -284,6 +284,38 @@ public void foreachRetriableDistribution(final PendingDistributionVisitor visito
...
+                  partitionId);
+              // we ignore this currently
+              return true;
+            }
+            lastDistributionKey.set(distributionKey);
+            lastCommandDistribution.set(commandDistributionRecord);
+          }
+
+          final var commandDistributionRecord = new CommandDistributionRecord();
+          commandDistributionRecord.wrap(lastCommandDistribution.get()).setPartitionId(partitionId);
```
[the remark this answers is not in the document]
remcowesterhoud: That's not entirely true. We create a new `CommandDistributionRecord` which has it's own implementation of a `wrap()` method. This wrap just uses getters and setters to set the properties from one to another. This doesn't work with buffers so it should be fine.

I just copy-pasted it from one of the other methods we have where we iterate through a column family 😄 

That said, there is a more efficient way to do this using the copyFrom method:

```java
          final var commandDistributionRecord = new CommandDistributionRecord();
          commandDistributionRecord.copyFrom(lastPendingDistribution.get());
          commandDistributionRecord.setPartitionId(partitionId);
          visitor.visit(distributionKey, commandDistributionRecord);
```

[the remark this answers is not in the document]
korthout: @remcowesterhoud Should we replace the `wrap` method's implementation with `copyFrom`?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/immutable/DistributionState.java
```diff
@@ -111,7 +122,7 @@ interface PendingDistributionVisitor {
      * @param distributionKey The key of the pending distribution
      * @param pendingDistribution The pending distribution itself as command distribution record
      */
-    void visit(final long distributionKey, final CommandDistributionRecord pendingDistribution);
+    boolean visit(final long distributionKey, final CommandDistributionRecord pendingDistribution);
```
berkaycanbc, line 125: 🔧 Would be useful to explain when to return false or true.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceDeletionDeleteProcessor.java
```diff
@@ -142,6 +142,15 @@ public ProcessingError tryHandleError(
       rejectionWriter.appendRejection(command, RejectionType.NOT_FOUND, exception.getMessage());
       responseWriter.writeRejectionOnCommand(
           command, RejectionType.NOT_FOUND, exception.getMessage());
+
+      // TODO verify this fix. There is a risk attached to it as deployment creation and resource
```
berkaycanbc, line 146: 🔧 We can remove this comment as we now have a migration to set queue id for deployment create and resource delete distributions in the state.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleAddEntityProcessor.java
```diff
@@ -111,7 +113,19 @@ final var record = command.getValue();
...
-    stateWriter.appendFollowUpEvent(command.getKey(), RoleIntent.ENTITY_ADDED, command.getValue());
+    final var record = command.getValue();
+    roleState
+        .getEntityType(record.getRoleKey(), record.getEntityKey())
+        .ifPresentOrElse(
+            entityType -> {
+              final var errorMessage =
+                  ENTITY_ALREADY_ASSIGNED_ERROR_MESSAGE.formatted(
+                      record.getEntityKey(), record.getRoleKey());
+              rejectionWriter.appendRejection(command, RejectionType.INVALID_STATE, errorMessage);
```
berkaycanbc, line 124: 🔧 Would be better to set rejection type as `ALREADY_EXISTS` to align with other distribution rejections. (e.g. in `TenantAddEntityProcessor`)

berkaycanbc (approved):
Thanks @remcowesterhoud very comprehensive PR 🚀 

I've asked couple of minor changes, feel free to apply the ones that makes sense. 👍 

I committed the test case to verify the idempotency of `MessageSubscription.MIGRATE` command distribution. ✅

PR #24780, opened 2024-11-15
feature: PrometheusMeterService will throw an exception when registering invalid metrics

entangled90:
## Description

Configure prometheus to throw an exception if a metric is not being registered correctly.
This required a small fix in taskslist importer as one metric was registered with 0 labels first and then with 2 labels afterwards.

### DIFF
--- a/dist/src/main/java/io/camunda/application/commons/metrics/MetricsConfiguration.java
+++ b/dist/src/main/java/io/camunda/application/commons/metrics/MetricsConfiguration.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.application.commons.metrics;
+
+import io.micrometer.prometheus.PrometheusMeterRegistry;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.stereotype.Component;
+
+@Component
+public class MetricsConfiguration implements BeanPostProcessor {
+
+  /** Configure PrometheusMeterRegistry if it's configured */
+  @Override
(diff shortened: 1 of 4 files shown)


tasklist/importer/src/main/java/io/camunda/tasklist/zeebeimport/es/RecordsReaderElasticSearch.java
houssain-barouni, line 1: We need to do the same in https://github.com/camunda/camunda/blob/main/tasklist/importer/src/main/java/io/camunda/tasklist/zeebeimport/os/RecordsReaderOpenSearch.java

dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
```diff
@@ -38,7 +38,8 @@
     basePackages = {
       "io.camunda.zeebe.broker",
       "io.camunda.zeebe.shared",
-      "io.camunda.authentication"
+      "io.camunda.authentication",
+      "io.camunda.application.commons.metrics"
```
houssain-barouni, line 42: why this package needs to be added in `BrokerModuleConfiguraiton` ? it is already scanned by [`CommonsModuleConfiguration`](https://github.com/camunda/camunda/blob/main/dist/src/main/java/io/camunda/application/commons/CommonsModuleConfiguration.java)

dist/src/main/java/io/camunda/application/commons/metrics/MetricsConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+package io.camunda.application.commons.metrics;
+
+import io.micrometer.prometheus.PrometheusMeterRegistry;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanPostProcessor;
+
+public class MetricsConfiguration implements BeanPostProcessor {
+
+  @Override
+  public Object postProcessAfterInitialization(final Object bean, final String beanName)
```
houssain-barouni, line 17: can you please add some java docs explaining why this is required ?

houssain-barouni (changes requested):
We also need to update the Opensearch implementation

houssain-barouni (approved):
LGTM!

PR #24782, opened 2024-11-15
[PoC] Identity ES/OS Group index data restructuring

koevskinikola:
## Description

A small PoC for an ES/OS data restructuring for the Identity Group index.

The goal is to introduce a join-relation in the Group index so that entity keys can be stored in the same index without nesting.

The PoC introduces the following:
- An `EntityJoinRelation` class (defines a 1:N join relationship) which contains the following properties:
  - `name` (String) - specifies if the entry is a parent/owner (1), or a child (N)
  - `parent` (Long) - if the entry is a child (N), specifies the parent/owner ID 
- An `IdentityJoinRelationshipType` enum. It specifies all the possible roles for the join relation. The enum can be expanded to support `role` and `tenant`:
  - `group`
  - `member`
- The `GroupEntity` is expanded to include a `join` property
- The existing Group handlers are adjusted to define the join-relation role of the entry they are adding to ES/OS
- An IT test for the new `GroupEntityRemovedHandler` is added.
   - Note: I'm facing some issues with the IT tests when a group keys need to be re-used for different handlers.
 
## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/EntityJoinRelation.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/EntityJoinRelation.java
@@ -30,6 +30,25 @@ public class EntityJoinRelation {
     return this;
   }
 
+  public static class EntityJoinRelationFactory {
+
+    private IdentityJoinRelationshipType parent;
+    private IdentityJoinRelationshipType child;
+
+    public EntityJoinRelationFactory(final IdentityJoinRelationshipType parent, final IdentityJoinRelationshipType child) {
+      this.parent = parent;
+      this.child = child;
+    }
+
+    public EntityJoinRelation createParent() {
+      return new EntityJoinRelation().setName(parent.getType());
+    }
+
+    public EntityJoinRelation createChild(final long parentKey) {
+      return new EntityJoinRelation().setName(child.getType()).setParent(parentKey);
+    }
+  }
+
   public enum IdentityJoinRelationshipType {
     GROUP("group"),
     MEMBER("member");
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/EntityJoinRelation.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/EntityJoinRelation.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.webapps.schema.entities.usermanagement;
+
+public class EntityJoinRelation {
+
+  private String name;
+  private Long parent;
+
+  public String getName() {
+    return name;
+  }
+
+  public EntityJoinRelation setName(final String name) {
+    this.name = name;
+    return this;
+  }
+
+  public Long getParent() {
+    return parent;
+  }
+
+  public EntityJoinRelation setParent(final Long parent) {
+    this.parent = parent;
+    return this;
+  }
+
+  public enum IdentityJoinRelationshipType {
+    GROUP("group"),
+    MEMBER("member");
+
+    private final String type;
+
+    IdentityJoinRelationshipType(final String type) {
+      this.type = type;
+    }
+
+    public String getType() {
+      return type;
+    }
+  }
+}
--- a/webapps-schema/src/main/resources/schema/elasticsearch/create/index/identity-groups.json
+++ b/webapps-schema/src/main/resources/schema/elasticsearch/create/index/identity-groups.json
@@ -11,8 +11,14 @@
       "name": {
         "type": "keyword"
       },
-      "assignedMemberKeys": {
+      "memberKey": {
         "type": "long"
+      },
+      "join": {
+        "type": "join",
+        "relations": {
+          "group": "member"
+        }
       }
     }
   }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/GroupEntityAddedHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/GroupEntityAddedHandler.java
@@ -9,8 +9,7 @@ package io.camunda.exporter.handlers;
 
 import io.camunda.exporter.exceptions.PersistenceException;
 import io.camunda.exporter.store.BatchRequest;
-import io.camunda.webapps.schema.entities.usermanagement.EntityJoinRelation;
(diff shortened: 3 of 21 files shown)


webapps-schema/src/main/resources/schema/elasticsearch/create/index/identity-groups.json
```diff
@@ -11,8 +11,14 @@
       "name": {
         "type": "keyword"
       },
-      "assignedMemberKeys": {
+      "entityKey": {
```
emulbreh, line 14:
```suggestion
      "memberKey": {
```
That would make it a bit clearer that it's a property of the child.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/GroupEntityAddedHandler.java
```diff
@@ -66,24 +55,17 @@ public GroupEntity createNewEntity(final String id) {
...
   @Override
   public void flush(final GroupEntity entity, final BatchRequest batchRequest)
       throws PersistenceException {
-
-    batchRequest.updateWithScript(
-        indexName,
-        entity.getId(),
-        SCRIPT_ADD_ENTITY,
-        Map.of("addKeys", entity.getAssignedMemberKeys()));
+    batchRequest.add(indexName, entity);
```
emulbreh, line 68:
```suggestion
    batchRequest.addWithRouting(indexName, entity, Long.toString(entity.getJoin().getParent());
```
According to [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html#:~:text=It%20is%20required%20to%20index%20the%20lineage%20of%20a%20parent%20in%20the%20same%20shard%20so%20you%20must%20always%20route%20child%20documents%20using%20their%20greater%20parent%20id.). We need the same change in GroupEntityRemovedHandler.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/EntityJoinRelation.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  public EntityJoinRelation setParent(final Long parent) {
+    this.parent = parent;
+    return this;
+  }
+
+  public enum IdentityJoinRelationshipType {
+    GROUP("group"),
+    MEMBER("member");
```
emulbreh, line 35: I think it'd be nice to move this so that we don't need to pack all types into one global enum.
For example, we could make it a nested enum in GroupIndex and either reference entries by name or parametrize EntityJoinRelation:
```java
class GroupIndex {
  …
  enum JoinName {
    GROUP,
    MEMBER
  }
  …
}

class EntityJoinRelation<E extends Enum<E>> {
  E getName() {…}
}
```
Another option (and I think I like this best) would be a factory for EntityJoinRelation and no enum at all:
```java
class GroupIndex {
  …
  public static final joinRelationFactory = new EntityJoinRelationFactory("group", "member");
  …
}

// and then:
EntityJoinRelation r = GroupIndex.joinRelationFactory.createParent();
EntityJoinRelation r = GroupIndex.joinRelationFactory.createChild(parentId);
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/GroupEntityRemovedHandler.java
```diff
@@ -0,0 +1,77 @@
...
+        new EntityJoinRelation()
+            .setName(IdentityJoinRelationshipType.MEMBER.getType())
+            .setParent(value.getGroupKey());
+    entity.setEntityKey(value.getEntityKey()).setJoin(joinRelation);
+  }
+
+  @Override
+  public void flush(final GroupEntity entity, final BatchRequest batchRequest)
+      throws PersistenceException {
+    batchRequest.delete(indexName, entity.getId());
```
emulbreh, line 70:
```suggestion
    batchRequest.deleteWithRouting(indexName, entity.getId());
```
We don't have that method yet, but it looks like it's easy to add.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/GroupEntityAddedHandler.java
```diff
@@ -55,7 +44,7 @@ public boolean handlesRecord(final Record<GroupRecordValue> record) {
 
   @Override
   public List<String> generateIds(final Record<GroupRecordValue> record) {
-    return List.of(String.valueOf(record.getKey()));
+    return List.of(String.valueOf(record.getValue().getEntityKey()));
```
emulbreh, line 47: Is this correct? The member key (e.g. a user key) can't be the id of the member document because it can be a related to multiple groups.

koevskinikola: That's an excellent point, we'll need a new unique value. I wonder if concatenating the group and entity keys into a single "child" key is acceptable.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/EntityJoinRelation.java
koevskinikola: @emulbreh with the current approach we can just expand the enum with `ROLE` and `TENANT` values and reuse the class in the Role- and Tenant-related handlers as well. 

❓ Why is it a problem to pack multiple values in a single enum?

emulbreh: > Why is it a problem to pack multiple values in a single enum?

It doesn't help to ensure that the right enum values are used for an index. 
We could leave it untyped and just work with the name strings, but then I'd remove the enum entirely and just keep constants in the handler or index descriptor.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/GroupEntityAddedHandler.java
emulbreh: How do we generate the key of a new group?

koevskinikola: We don't generate it in the exporter, we use the same key that is generated by Zeebe when the group is created. I think making a composite key, i.e. `groupKey-entityKey` would work here. We already do something similar [here](https://github.com/camunda/camunda/blob/78cb2c5deab1507b9dab8c5a5dbbafa81bd89a4a/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/listview/VariableForListViewEntity.java#L26-L28).

koevskinikola:
Closing this PoC PR in favor of a proper feature PR: https://github.com/camunda/camunda/pull/25303

@emulbreh please continue your review on the feature PR I linked in this comment.

PR #24838, opened 2024-11-19
Register `OrderedCommandDistributionMigration` migration task

### LINKED ISSUE #24836: `OrderedCommandDistributionMigration` is never executed
**Describe the bug**

This migration task is never registered in the `DbMigratorImpl`. As a result this migration is never executed.

This migration was part of the 8.6 release. It was never executed for clusters running on this version.

**Impact**

The migration is responsible for moving command distributions into a new ColumnFamily, the `retriableDistributionColumnFamily`. Since 8.6 this CF is used for the redistribution of commands in cases no acknowledgement was received. The only commands we distribute in this version are `Deployment.CREATE` and `ResourceDeletion.DELETE`.

If these commands are no longer retried we could reach a data inconsistency between the partitions. What a user could notice is:

1. A deployment is created on some partitions, but not on others. They would notice that sometimes a process instance creation fails with the message `Expected to find process definition with key '...', but none found` 
2. A deployment is deleted on some partitions, but not on others. They could notice that the process instance creation fails as expected on some partitions, but will succesfully be created on others.

The chance of this happening is fairly medium. There must've been pending distributions for this to occur. Users can easily recover from this by redeploying their resources.

**Expected behavior**

- The migration should run and the state should remain consistent amongst partitions.
- We should add a test to ensure all migration tasks are added and ran.

**Environment:**
- Zeebe Version: 8.6

remcowesterhoud:
## Description

This migration was part of the 8.6 release, but was never registered. As a result it never ran. This PR adds registers the task and adds a test to make sure we don't encounter this problem in the future.

## Checklist

- [ ] for CI changes:

## Related issues

closes #24836

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/migration/MigrationTaskRegistrationTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/migration/MigrationTaskRegistrationTest.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.state.migration;
+
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+
+import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.core.domain.JavaClass;
+import com.tngtech.archunit.core.importer.ImportOption;
+import com.tngtech.archunit.junit.AnalyzeClasses;
+import com.tngtech.archunit.junit.ArchTest;
+import com.tngtech.archunit.lang.ArchRule;
+
+@AnalyzeClasses(
+    packages = "io.camunda.zeebe.engine.state.migration",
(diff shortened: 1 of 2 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/migration/MigrationTaskRegistrationTest.java
```diff
@@ -0,0 +1,42 @@
...
+    importOptions = ImportOption.DoNotIncludeTests.class)
+public class MigrationTaskRegistrationTest {
+
+  @ArchTest()
+  public static final ArchRule MIGRATION_TASKS_MUST_BE_REGISTERED =
+      classes()
+          .that()
+          .implement(MigrationTask.class)
+          .should()
+          .beAssignableFrom(anyOfTheRegisteredMigrationTasks());
```
lenaschoenburg, line 30: :thought_balloon: This reads a bit weird, I would have expected something like `.isOneOf(registeredMigrationsTaskClasses())` but as long as it works...

lenaschoenburg (approved):
Sorry, forgot to send the approval. Looks good to me :+1:

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/migration/MigrationTaskRegistrationTest.java
remcowesterhoud: That would read nicer, but this is the ArchUnit API so not much I can do about it 🤷

PR #24841, opened 2024-11-19
refactor: adjust user management domain entities

houssain-barouni:
## Description

same as https://github.com/camunda/camunda/pull/24750
Applied to:
- Authorization
- Role
- Tenant
- User

## Checklist

- [ ] for CI changes:

## Related issues

relates to https://github.com/camunda/camunda/issues/24911

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -89,7 +89,8 @@ public class SearchClients
             transformers,
             new DocumentAuthorizationQueryStrategy(this),
             securityContext);
-    return executor.search(filter, AuthorizationEntity.class);
+    return executor.search(
+        filter, io.camunda.webapps.schema.entities.usermanagement.AuthorizationEntity.class);
   }
 
   @Override
@@ -100,7 +101,8 @@ public class SearchClients
             transformers,
             new DocumentAuthorizationQueryStrategy(this),
             securityContext);
-    return executor.findAll(filter, AuthorizationEntity.class);
+    return executor.findAll(
+        filter, io.camunda.webapps.schema.entities.usermanagement.AuthorizationEntity.class);
   }
 
   @Override
@@ -218,7 +220,8 @@ public class SearchClients
             transformers,
             new DocumentAuthorizationQueryStrategy(this),
             securityContext);
-    return executor.search(filter, RoleEntity.class);
+    return executor.search(
+        filter, io.camunda.webapps.schema.entities.usermanagement.RoleEntity.class);
   }
 
   @Override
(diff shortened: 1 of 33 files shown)


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
```diff
@@ -98,7 +99,8 @@ public List<AuthorizationEntity> findAllAuthorizations(final AuthorizationQuery
             transformers,
             new DocumentAuthorizationQueryStrategy(this),
             securityContext);
-    return executor.findAll(filter, AuthorizationEntity.class);
+    return executor.findAll(
+        filter, io.camunda.webapps.schema.entities.usermanagement.AuthorizationEntity.class);
```
emulbreh, line 103: It would be great if we could find a naming convention that didn't require qualification. That's not only going to help when reading and navigating the code, but also when we talk about it. A few options (for the webapps.schema entities):
* `FooExporterEntity`
* `ExportedFoo`
* `FooIndexEntity`

houssain-barouni: Yes, I agree with the proposal
I preferred not to do it in these pull requests to reduce the number of changed files, but we can do that afterward

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/AuthorizationFilterTransformer.java
```diff
@@ -22,13 +28,14 @@ public final class AuthorizationFilterTransformer
   @Override
   public SearchQuery toSearchQuery(final AuthorizationFilter filter) {
     return and(
-        longTerms("ownerKey", filter.ownerKeys()),
-        filter.ownerType() == null ? null : term("ownerType", filter.ownerType()),
-        stringTerms("permissions.resourceIds", filter.resourceKeys()),
-        filter.resourceType() == null ? null : term("resourceType", filter.resourceType()),
+        longTerms(OWNER_KEY, filter.ownerKeys()),
+        filter.ownerType() == null ? null : term(OWNER_TYPE, filter.ownerType()),
+        stringTerms("%s.%s".formatted(PERMISSIONS, PERMISSIONS_RESOURCEIDS), filter.resourceKeys()),
```
stoerti, line 33: ❓  A field named "resourceIds" is mapped to "resourceKeys". In general convention in the API a key is a Long value and an ID is a String value.

Shouldn't the resourceKey be a resourceId?

houssain-barouni: indeed, good catch!
It seems that it was updated everywhere-else except in the filter, I will update it

houssain-barouni: Fixed @stoerti

stoerti (approved):

Issue #23913, opened 2024-10-23
Importer need to communicate when importing is done


Zelldon:
## Description

With introducing the Camunda Exporter we will replace the previously deployed Importer applications.

The applications still need to run for a bit, to import left-over data from Zeebe indices before they can turned off. 

Importers should communicate either via metric or API (actuator) that they catched up and consumed all data from Zeebe indices.

Zelldon:
Affecting also documentation https://github.com/camunda/camunda-docs/issues/4493 and Operator/SM to understand when Importer can be shutdown (Migration app to understand for how long it needs to run).

Zelldon:
ℹ️ Discussed this topic today with @romansmirnov 

We might found a good way to detect when importing has reached the end. See related document


1. We need to fix first https://github.com/camunda/camunda/issues/24316 to make sure ES/OS exporter are writing to correct indices. This should make sure that exporters are writing only to 8.7 zeebe indices when all 8.6 records have been written
2. On the importer side we can detect when first data arrives in 8.7 indices
3. When we have imported everything in the 8.6 indices, and we see data in 8.7 indices we should be able to detect that importing is done
4. Importing done should be communicated via writing to the `import-position` index. We need to extend index with additional flag, here we need to distinguish by partition and value type as the importer imports per partition and value type concurrently
5. As soon as we are done for all of these value types and partitions we can mark importing as done

This index and values can be consumed then by migration and exporter (to start exporting, resolving concurrency problems).

Zelldon:
Q to answer: Do we read in sequence? Can we assume that if we read 8.7 record in importer, we have read ALL 8.6 for that value type and partition.

**Our hypothesis:**

 * we can remove or flag a record reader done - if we receive a 8.7 sub batch
 * If an import job is "cold" (are not seeing any data for a while) BUT we saw in different readers has a 8.7 record - we can expect that there will be no more data from 8.6

Needs to be verified.

Zelldon:
Regarding 4. to be specific: we plan to use ES to communicate with different applications importer/migration app (as it is  common/shared between)

PR #24849, opened 2024-11-19
Commincate when done opensearch operate

EuroLew:
## Description

Opensearch support for operate importer marking its importing being done via the import position index. Just porting over the work done in #24671

## Related issues

relates to #23913


operate/importer/src/main/java/io/camunda/operate/zeebeimport/RecordsReaderHolder.java
```diff
@@ -75,9 +79,32 @@ public boolean hasPartitionCompletedImporting(final int partitionId) {
     return partitionsCompletedImporting.contains(partitionId);
   }
 
-  public int incrementEmptyBatches(final int partitionId, final ImportValueType importValueType) {
+  public void incrementEmptyBatches(final int partitionId, final ImportValueType importValueType) {
     final var reader = getRecordsReader(partitionId, importValueType);
-    return countEmptyBatchesAfterImportingDone.merge(reader, 1, Integer::sum);
+    countEmptyBatchesAfterImportingDone.merge(reader, 1, Integer::sum);
```
mustafadagher, line 84: 💭 Just thinking out loud:
Could there ever be a race condition while incrementing this emptyBatches value?
I mean, could it happen that two different threads are ran against the same reader and they come to increment this value in the same time? 
if it's a possibility, then we should make the value here an AtomicInteger. But I'm not sure if this is really an issue to worry about.

operate/pom.xml
```diff
@@ -175,7 +175,7 @@
     <profile>
       <id>operateItOpensearch</id>
       <properties>
-        <includeITNames>**/api/v1/**/*IT.java</includeITNames>
+        <includeITNames>**/api/v1/**/*IT.java, **/OpensearchFinishedImportingIT.java</includeITNames>
```
mustafadagher, line 178: ❓ why do we need a special treatment for this?

operate/importer/src/main/java/io/camunda/operate/zeebeimport/ReadBatchErrorMessage.java
```diff
@@ -0,0 +1,20 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.operate.zeebeimport;
+
+import java.util.function.BiFunction;
+
+public final class ReadBatchErrorMessage {
+  public static final BiFunction<String, String, String> READ_BATCH_ERROR_MESSAGE =
```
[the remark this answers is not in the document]
Zelldon: Agree here, I think we don't need to overcomplicate things.

operate/importer/src/main/java/io/camunda/operate/zeebeimport/RecordsReaderHolder.java
```diff
@@ -75,9 +79,32 @@ public boolean hasPartitionCompletedImporting(final int partitionId) {
...
+
+  public boolean isRecordReaderCompletedImporting(
+      final int partitionId, final ImportValueType importValueType) {
+    if (hasPartitionCompletedImporting(partitionId)) {
+
+      final var reader = getRecordsReader(partitionId, importValueType);
+      return Objects.equals(
+          countEmptyBatchesAfterImportingDone.get(reader),
+          MINIMUM_EMPTY_BATCHES_FOR_COMPLETED_READER);
+    }
```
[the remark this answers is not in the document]
EuroLew: Good catch, this was a bug that would have cause some issues in conjunction with the bug I fixed in https://github.com/camunda/camunda/pull/24849/commits/01a2744621b8e07479605bfa02d94ed149da2161. If both bugs were left in. After a record reader is marked as completed, the next subsequent 8.7.0 record would cause it to be marked as not completed 👍

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
```diff
@@ -182,6 +172,47 @@ final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 1);
...
+    EXPORTER.export(record2);
+    esClient.indices().refresh(new RefreshRequest("*"), RequestOptions.DEFAULT);
+
+    // receives 8.7 record and marks partition as finished importing
+    zeebeImporter.performOneRoundOfImport();
+
+    // then
+    // require multiple checks to avoid race condition. If records are written to zeebe indices and
+    // before a refresh, the record reader pulls the import batch is empty so it then says that the
+    // record reader is done when it is not.
```
Zelldon, line 195: 🔧 
```suggestion
    // Require multiple checks to avoid race condition. 
    // Otherwise: If records are written to zeebe indices and before a refresh, the record reader pulls an empty import batch, then it might assume falsely
    // that it is done, while it is not.
```

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
```diff
@@ -182,6 +172,47 @@ final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 1);
...
+        .until(() -> isRecordReaderIsCompleted("1-process-instance"));
+
+    final var record3 = generateRecord(ValueType.PROCESS_INSTANCE, "8.7.0", 1);
+    EXPORTER.export(record3);
+    esClient.indices().refresh(new RefreshRequest("*"), RequestOptions.DEFAULT);
+
+    zeebeImporter.performOneRoundOfImport();
+
+    Awaitility.await()
+        .during(Duration.ofSeconds(10))
```
Zelldon, line 211: 🔧 Here we wait again for 10 seconds, I think we removed this the last time.

operate/qa/integration-tests/src/test/java/io/camunda/operate/opensearch/OpensearchFinishedImportingIT.java
```diff
@@ -0,0 +1,254 @@
...
+    final var newVersionProcessInstanceRecord =
+        generateRecord(ValueType.PROCESS_INSTANCE, "8.7.0", 1);
+    EXPORTER.export(newVersionProcessInstanceRecord);
+    final var newVersionDecisionEvalRecord =
+        generateRecord(ValueType.DECISION_EVALUATION, "8.7.0", 1);
+    EXPORTER.export(newVersionDecisionEvalRecord);
+
+    for (int i = 0; i <= RecordsReaderHolder.MINIMUM_EMPTY_BATCHES_FOR_COMPLETED_READER; i++) {
+      // simulate existing decision records left to process so it is not marked as completed
+      final var decisionRecord2 = generateRecord(ValueType.DECISION, "8.6.0", 1);
```
Zelldon, line 135: That is not possible right? As we exported already 8.7, might make sense to adjust the test to real world scenario.

operate/qa/integration-tests/src/test/java/io/camunda/operate/opensearch/OpensearchFinishedImportingIT.java
```diff
@@ -0,0 +1,254 @@
...
+      zeebeImporter.performOneRoundOfImport();
+    }
+
+    Awaitility.await()
+        .atMost(Duration.ofSeconds(30))
+        .until(
+            () ->
+                isRecordReaderIsCompleted("1-process-instance")
+                    && isRecordReaderIsCompleted("1-decision-evaluation")
+                    && !isRecordReaderIsCompleted("1-decision"));
```
Zelldon, line 148: 🤔 💭 This is a problem if we never mark our partition as fully complete

Zelldon: Or would it be complete if we now run into 5 empty batches? ❓

operate/qa/integration-tests/src/test/java/io/camunda/operate/opensearch/OpensearchFinishedImportingIT.java
```diff
@@ -0,0 +1,254 @@
...
+    final var partitionTwoRecord = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 2);
+    EXPORTER.export(record);
+    EXPORTER.export(partitionTwoRecord);
+    osClient.index().refresh("*");
+
+    zeebeImporter.performOneRoundOfImport();
+
+    // when
+    final var record2 = generateRecord(ValueType.PROCESS_INSTANCE, "8.7.0", 1);
+    final var partitionTwoRecord2 = generateRecord(ValueType.PROCESS_INSTANCE, "8.7.0", 2);
```
Zelldon, line 164: ❓ What happens if this record is not written?

operate/qa/integration-tests/src/test/java/io/camunda/operate/opensearch/OpensearchFinishedImportingIT.java
Zelldon: This would be possible for example on a different partition

ChrisKujawa (approved):
🚀 Thanks Euro

I have smaller comments, please have a look before merging. We might want to add one ore two more tests.

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
EuroLew: It's not ideal but we really want to make sure this is not set to false and if we assert that it is true we could get a race condition. I have tried the following 
image
where we fresh the indices and then check that it's true but this is a false positive where the test passes while the actual behaviour is wrong.

EuroLew: An alternative I can think of is a thread.sleep but it's basically the same issue, this comes does to ES behaviour where an operation is not immediately executed and we must wait for the refresh timer.

operate/qa/integration-tests/src/test/java/io/camunda/operate/opensearch/OpensearchFinishedImportingIT.java
EuroLew: I'm not quite sure what you mean, on line 133 we run `MINIMUM_EMPTY_BATCHES_FOR_COMPLETED_READER` number of rounds of importing, and for the process instance and decision evaluation record readers this will result in 5 empty batches which given the exported 8.7 record (which marked the partition completed)

operate/qa/integration-tests/src/test/java/io/camunda/operate/opensearch/OpensearchFinishedImportingIT.java
EuroLew: If an 8.7 record is never written to the zeebe indices then the partition is never marked as completed so importing will never "finish"

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
ChrisKujawa: :D I had this discussion in my life so often now.

My learning so far: We can't assert something that shouldn't happen. You can assert something existing or a behavior you can observe. Either we rethink the test or leave it as it is.

operate/qa/integration-tests/src/test/java/io/camunda/operate/opensearch/OpensearchFinishedImportingIT.java
ChrisKujawa: Ok but this means only for this specific partition right? 

Just to confirm: It is enough to have ONE record with version 8.7 to mark the partition complete right?

ChrisKujawa (approved):

PR #24852, opened 2024-11-19
feat: Ensure endpoints are backward compatible when zeebe is deployed standalone

### LINKED ISSUE #24850: Ensure endpoints are backward compatible when zeebe is deployed standalone
When zeebe, operate and tasklist are deployed separately there is no reason to break compatibility.

entangled90:
## Description
A new WebEndpoint controller (at `backups`) has been added that is enabled only when the profile is zeebe, but not operate or tasklist.
The new controller just delegate every call directly to the backup controller mounted at `/backup-runtime`

Operate & Tasklist controller have not yet been migrated to the new schema, so they are not part of this PR

## Related issues
closes #24850

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/shared/management/BackupEndpointStandalone.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/management/BackupEndpointStandalone.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.shared.management;
+
+import org.springframework.boot.actuate.endpoint.annotation.DeleteOperation;
+import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
+import org.springframework.boot.actuate.endpoint.annotation.Selector;
+import org.springframework.boot.actuate.endpoint.annotation.WriteOperation;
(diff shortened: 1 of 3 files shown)


dist/src/main/java/io/camunda/zeebe/shared/management/BackupEndpointStandalone.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.boot.actuate.endpoint.annotation.WriteOperation;
+import org.springframework.boot.actuate.endpoint.web.WebEndpointResponse;
+import org.springframework.boot.actuate.endpoint.web.annotation.WebEndpoint;
+import org.springframework.context.annotation.Profile;
+import org.springframework.lang.NonNull;
+import org.springframework.stereotype.Component;
+
+@Component
+@WebEndpoint(id = "backups")
+@Profile("broker & !operate & !tasklist")
```
entangled90, line 23: Is this expression correct? should I add "!optimize & !identity" as well?

dist/src/main/java/io/camunda/zeebe/shared/management/BackupEndpointStandalone.java
```diff
@@ -0,0 +1,50 @@
...
+public class BackupEndpointStandalone {
+  private final BackupEndpoint backupEndpoint;
+
+  @Autowired
```
entangled90, line 27: Not necessary, I'll remove it before merging.

npepinpe (approved):
Pre-approving, but I would use the `standalone` profile to detect when an application is standalone. Once that's changed, we can follow the same strategy for the others.

dist/src/main/java/io/camunda/zeebe/shared/management/BackupEndpointStandalone.java
entangled90: I think I just followed Intellij autocompletions :smile: 
I'll put it private

PR #24854, opened 2024-11-19
refactor: move shared backup classes into a new webapps-backup module

### LINKED ISSUE #24843: Refactor common backup logic from operate into a new webapps-backup module
Considering that we will unify the backup of operate & tasklist, it's a good first step to move interfaces & POJOs to a new webapp-backup module. 

operate-webapps will depend on this module.

The implementation of `BackupService` will stay in the operate-webapp module for the moment, as tasklist owns another implementation as well and they will need to be merged.

The behaviour must not be changed, it's purely an internal refactoring.

entangled90:
## Description
Extracted an interface from operate's BackupService and move into the new shared module.
All classes used by this interface have been moved to the shared module.
A new Metadata record has been created with no logic whatsoever in contrast to operate Metadata, which can be converted to the new record with a new method.

Eventually all the backup logic from operate and tasklist will be moved into `webapps-backup` module.
Controllers will be implemented in `dist` (as it's done with zeebe) and we will be able to directly use classes generated from openapi spec.

No behaviour change is expected from this PR.
## Related issues

closes #24843


npepinpe (approved):
:rocket:

PR #24859, opened 2024-11-19
Support multi handler flush

### LINKED ISSUE #24736: Potential lost information in Camunda exporter
### Description

Every record that is exported with Camunda exporter is processed by several [ExportHandlers](https://github.com/camunda/camunda/tree/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers). The entity is cached until [`flush` method is called](https://github.com/camunda/camunda/blob/e78dd201e46fbd3897e44b64bd78b7693916dd46/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java#L62). Flush from the latest called handler will be called. But in case the entity already exists in Elastic, only part of the fields will be updated, those updated in the last handler. We may lose some other field values, that were set in previous handlers.

### TODO

* we need to find a way to properly accumulate all the updates from all the called handlers, when performing flush

### Additional Information

### Acceptance Criteria

ChrisKujawa:
### Context

> Current state:
> - In the ExporterBatchWriter we pass records to all handlers that accept the value type and want to handle the given record
> - The results of each handler are stored in a cache (with the corresponding handler)
> - Results are identified by id and entity type
> - On flush, we call for each cached entity the corresponding handler to flush
> 
> **Expectations (previous):**
> 
> - The export handlers have a one-to-one mapping to an entity. 
> - This means per entity we have one specific handler, and vice-versa
> - It could happen that the same handler is called multiple times during a batch because we see multiple events related to the same entity (ACTIVIATED-COMPLETED, etc.)
> 
> Problem:
> 
> - We have implemented multiple handlers that reuse entity types, even if they set only a subset of properties. 
> - Consequence:
>     - As they use the same type, and in some cases even generate the same id only the last handler wins. Flush is only called on the last one, which consumed the record.
>     - Only partial data is updated.
> - On creation (missing document) this is not an issue because we use mostly upserts, which uses the complete entity object. Here all data is written.
>
> https://github.com/camunda/camunda/issues/24736#issuecomment-2485444919

We were approaching this by caching all handlers related to the entity.

> [!Important]
>
> We are aware that this will cause more upserts per entity, and this might have a potential performance impact. But, we have no actual data to prove that this is for sure the case.
>
> - As we want to fix this critical bug ASAP we **[accept](https://camunda.slack.com/archives/C06F0GLJNFM/p1732020688807309?thread_ts=1731426489.455309&cid=C06F0GLJNFM)** this risk. 
> - If we see performance issues/regressions we can reconsider the solution and solve it differently.
> - As an additional task, we will extend our metric support

### In this PR

* Added new tests (first failing) that verify multiple registered handlers on the BatchWriter are called and flushing entities
* Store more handlers per entity, before we only stored one
* Call all handlers for an entity to flush all related data

## Checklist

- [ ] for CI changes:

## Related issues

For context https://camunda.slack.com/archives/C06F0GLJNFM/p1732015209561459?thread_ts=1731426489.455309&cid=C06F0GLJNFM

closes https://github.com/camunda/camunda/issues/24736

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java
@@ -16,8 +16,10 @@ import io.camunda.zeebe.protocol.record.ValueType;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /** Caches exporter entities of different types and provide the method to flush them in a batch. */
 @SuppressWarnings({"rawtypes", "unchecked"})
@@ -43,20 +45,19 @@ public class ExporterBatchWriter {
       final Record<?> record, final ExportHandler handler, final String id) {
     final var cacheKey = new EntityIdAndEntityType(id, handler.getEntityType());
 
-    final ExporterEntity entity;
+    final EntityAndHandler entityAndHandler =
+        cachedEntities.computeIfAbsent(
+            cacheKey,
+            (k) -> {
+              final ExporterEntity entity = handler.createNewEntity(id);
+              return new EntityAndHandler(entity, new HashSet<>());
+            });
 
-    final boolean alreadyCached = cachedEntities.containsKey(cacheKey);
-    if (alreadyCached) {
-      entity = cachedEntities.get(cacheKey).entity();
-    } else {
-      entity = handler.createNewEntity(id);
-    }
-
+    final var entity = entityAndHandler.entity;
     handler.updateEntity(record, entity);
 
-    // always store the latest handler in the tuple, because that is the one
-    // taking care of flushing
-    cachedEntities.put(cacheKey, new EntityAndHandler(entity, handler));
+    // we store all handlers for an entity to make sure not to miss any flushes
+    entityAndHandler.handlers.add(handler);
   }
 
   public void flush(final BatchRequest batchRequest) throws PersistenceException {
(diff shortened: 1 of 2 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java
```diff
@@ -43,20 +45,19 @@ private void updateAndCacheEntity(
...
-      entity = cachedEntities.get(cacheKey).entity();
-    } else {
-      entity = handler.createNewEntity(id);
-    }
+    final EntityAndHandlers entityAndHandlers =
+        cachedEntities.computeIfAbsent(
+            cacheKey,
+            (k) -> {
+              final ExporterEntity entity = handler.createNewEntity(id);
+              return new EntityAndHandlers(entity, new HashSet<>());
```
mustafadagher, line 53: 💭 if we care about the order handlers are inserted in we should use `LinkedHashSet` instead of `HashSet`

mustafadagher (approved):
This looks great! 🚀 I've left only one thought below

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java
Zelldon: Do we really care? Right now i think it doesnt matter. But happy to be proven otherwise :)

mustafadagher: right now probably not. but it could happen that in the same batch, there is a handler that add a document, and another that updates it.
`update` operations would fail if the document did not exist

Zelldon: But i guess some deterministic behavior wouldnt hurt 👍

Zelldon: @mustafadagher I added a new test and the fix, via using the linked hash set.

PR #24866, opened 2024-11-19
feat: reconstruct deployment for process resources

lenaschoenburg:
Adds support for finding the first process resource and reconstructing a matching deployment for it.
If the process is already marked with a deployment key, we reuse that key and also attach all other processes with the same deployment key.
Otherwise, we create a new deployment with key equal to the key of the process resource.

relates to #24817

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
@@ -68,7 +68,8 @@ public class DeploymentReconstructProcessor implements TypedRecordProcessor<Depl
     } else {
       deploymentRecord = createNewDeploymentForResource(resource);
     }
-    stateWriter.appendFollowUpEvent(key, DeploymentIntent.RECONSTRUCTED, deploymentRecord);
+    stateWriter.appendFollowUpEvent(
+        deploymentRecord.getDeploymentKey(), DeploymentIntent.RECONSTRUCTED, deploymentRecord);
   }
 
   private Resource findNextResource() {
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
@@ -124,7 +124,7 @@ public class DeploymentReconstructProcessor implements TypedRecordProcessor<Depl
       case ProcessResource(final var process) -> {
         deploymentRecord.setDeploymentKey(process.getKey());
         deploymentRecord.setTenantId(process.getTenantId());
-        attachResourceToDeployment(deploymentRecord, resource);
+        attachResourceMetadataToDeployment(deploymentRecord, resource);
       }
     }
     return deploymentRecord;
@@ -139,20 +139,15 @@ public class DeploymentReconstructProcessor implements TypedRecordProcessor<Depl
     deploymentRecord.setDeploymentKey(deploymentKey);
     deploymentRecord.setTenantId(tenantId);
     for (final var resource : resources) {
-      attachResourceToDeployment(deploymentRecord, resource);
+      attachResourceMetadataToDeployment(deploymentRecord, resource);
     }
     return deploymentRecord;
   }
 
-  private void attachResourceToDeployment(
+  private void attachResourceMetadataToDeployment(
       final DeploymentRecord deploymentRecord, final Resource resource) {
     switch (resource) {
       case ProcessResource(final var process) -> {
-        deploymentRecord
-            .resources()
-            .add()
-            .setResourceName(process.getResourceName())
-            .setResource(process.getResource());
         deploymentRecord
             .processesMetadata()
             .add()
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
@@ -11,6 +11,10 @@ import io.camunda.zeebe.engine.processing.ExcludeAuthorizationCheck;
 import io.camunda.zeebe.engine.processing.streamprocessor.TypedRecordProcessor;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
+import io.camunda.zeebe.engine.state.immutable.DecisionState;
+import io.camunda.zeebe.engine.state.immutable.FormState;
+import io.camunda.zeebe.engine.state.immutable.ProcessState;
+import io.camunda.zeebe.engine.state.immutable.ProcessingState;
 import io.camunda.zeebe.protocol.impl.record.value.deployment.DeploymentRecord;
 import io.camunda.zeebe.protocol.record.intent.DeploymentIntent;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
@@ -18,11 +22,20 @@ import io.camunda.zeebe.stream.api.state.KeyGenerator;
 
 @ExcludeAuthorizationCheck
 public class DeploymentReconstructProcessor implements TypedRecordProcessor<DeploymentRecord> {
-  private final StateWriter stateWriter;
   private final KeyGenerator keyGenerator;
+  private final ProcessState processState;
+  private final FormState formState;
+  private final DecisionState decisionState;
+  private final StateWriter stateWriter;
 
-  public DeploymentReconstructProcessor(final KeyGenerator keyGenerator, final Writers writers) {
+  public DeploymentReconstructProcessor(
+      final KeyGenerator keyGenerator,
+      final ProcessingState processingState,
+      final Writers writers) {
     this.keyGenerator = keyGenerator;
+    processState = processingState.getProcessState();
+    formState = processingState.getFormState();
+    decisionState = processingState.getDecisionState();
     stateWriter = writers.state();
   }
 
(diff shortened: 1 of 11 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
```diff
@@ -42,7 +53,134 @@ public DeploymentReconstructProcessor(
...
+
+    final DeploymentRecord deploymentRecord;
+    if (resource.deploymentKey() != NO_DEPLOYMENT_KEY) {
+      final var allResourcesOfDeployment = findResourcesWithDeploymentKey(resource.deploymentKey());
+      deploymentRecord =
+          recreateDeploymentForResources(
+              resource.deploymentKey(), resource.tenantId(), allResourcesOfDeployment);
+    } else {
+      deploymentRecord = createNewDeploymentForResource(resource);
+    }
```
remcowesterhoud, line 70: 🙃 Personally, I would flip the statements so there's no negation needed. I always find it reads easier, but it's w/e.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
```diff
@@ -42,7 +53,134 @@ public DeploymentReconstructProcessor(
...
+
+    return null;
+  }
+
+  private Set<Resource> findResourcesWithDeploymentKey(final long deploymentKey) {
+    // Iterate through processState, formState, and decisionState to find resources that are marked
+    // with the given deployment key.
+    final var resources = new HashSet<Resource>();
+    processState.forEachProcess(
+        null,
```
remcowesterhoud, line 104: ❌ Instead of starting from the beginning we could start with the deployment key. It seems the deployment key is always generated as a first step when processing a new deployment. That means subsequent resource keys must have a higher value. By doing this we don't have to go through the entire state all the time. We start with the first process in the deployment, and we can stop as soon as we encounter a process that has a deployment key that's not equal to the provided key.

remcowesterhoud: We should add a test case to ensure this behavior remains 🙂

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
```diff
@@ -139,20 +139,15 @@ private DeploymentRecord recreateDeploymentForResources(
...
     deploymentRecord.setTenantId(tenantId);
     for (final var resource : resources) {
-      attachResourceToDeployment(deploymentRecord, resource);
+      attachResourceMetadataToDeployment(deploymentRecord, resource);
     }
     return deploymentRecord;
   }
 
-  private void attachResourceToDeployment(
+  private void attachResourceMetadataToDeployment(
```
remcowesterhoud, line 148: ❓ I assume when scaling you will iterate over the deployments, grab the binary from the state using the resource metadata available and add them to the record before distributing?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessorTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+  @BeforeEach
+  void setUp() {
+    final var keyGenerator = new DbKeyGenerator(1, zeebeDb, transactionContext);
+    final var eventAppliers = new EventAppliers().registerEventAppliers(state);
+    final var writers = new Writers(() -> resultBuilder, eventAppliers);
+
+    resultBuilder = new FakeProcessingResultBuilder<>();
+    processor = new DeploymentReconstructProcessor(keyGenerator, state, writers);
+  }
```
remcowesterhoud, line 50: ❓ What's the reason you deviated from the way we normally test our processor with the `EngineRule`?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
```diff
@@ -42,7 +53,134 @@ public DeploymentReconstructProcessor(
...
+
+  private Resource findNextResource() {
+    final var foundProcess = new MutableReference<PersistedProcess>();
+    processState.forEachProcess(
+        null, // TODO: Continue where we left off
+        process -> {
+          final var deploymentKey = process.getDeploymentKey();
+          if (deploymentKey != NO_DEPLOYMENT_KEY
+              && deploymentState.hasStoredDeploymentRecord(deploymentKey)) {
+            return true;
```
remcowesterhoud, line 83: 💭 I wonder if we can actually stop here if we found a deployment key that already has a stored deployment associated with it. I think if this is the case it must've been created on the latest version where we store this, so all processes after would also have been created on that same version.

remcowesterhoud (changes requested):
Sorry for taking so long to review 🙇 I have a few comments, please have a look!

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
lenaschoenburg: Exactly. No need to have the binary resource duplicated if we can also just look it up when needed.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessorTest.java
lenaschoenburg: At least two reasons:
1. The `Reconstruct` command is written internally so all of the ceremony in adding a "client" for it etc. is not needed long term.
2. In my tests I want to directly mutate the state to simulate specific scenarios. For example resources existing without their deployment. The `EngineRule` is not really set up for this.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
lenaschoenburg: Processes are stored by tenant first so there's no way to start iteration at a certain deployment key.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
lenaschoenburg: > I think if this is the case it must've been created on the latest version where we store this

Or we already recreated it. I think this should be possible but depends on how exactly we determine the `previousProcess`. So we should tackle this together in a later PR.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/deployment/ProcessStateTest.java
```diff
@@ -1041,6 +1048,45 @@ public void shouldAddProcessAfterOnlyVersionIsDeleted() {
...
+    processState.putProcess(process2.getKey(), process2);
+    processState.putProcess(process3.getKey(), process3);
+    final var visitor = mock(PersistedProcessVisitor.class);
+    when(visitor.visit(any())).thenReturn(true);
+
+    // when
+    processState.forEachProcess(null, visitor);
+
+    // then
+    verify(visitor, times(3)).visit(any());
```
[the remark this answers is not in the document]
lenaschoenburg: It's a bit ugly because we can't just verify based on the argument because iteration does not make a copy. But it makes sense to be stricter here :+1: 
I've added an intermediary visitor to copy out the key and verify only that now. 53aeabd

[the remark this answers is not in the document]
lenaschoenburg: Hm, the argument captor would need to either make a copy or only capture the key. What I have written here is basically the same.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessor.java
remcowesterhoud: Because in the meantime a new deployment for a tenant with a lower alphanumeric value occurred?

lenaschoenburg: I meant that the problem is fundamentally that you can only iterate through processes in tenant id order - so if we are searching for all processes with a given deployment key, where should we start searching?

remcowesterhoud: That's only because you pass the deployment key to this method. You have access to a full resource which contains a tenant id. So you do know what the previous tenant is and could continue from there.

lenaschoenburg: Ah right, because a deployment can't contain resources from multiple tenants? Yeah that would work then :+1:

remcowesterhoud: I strongly believe we should do this. Because depending on the amount of processes this could be a real bottleneck performance-wise 😄

lenaschoenburg: Done in 95fb89c

remcowesterhoud (approved):
Awesome, thanks @lenaschoenburg 🚀

PR #24870, opened 2024-11-19
feat: adopt invocation provider supporting AWS OS

igpetrov:
## Description

feat: adopt invocation provider supporting AWS OS

## Checklist

- [ ] for CI changes:

## Related issues

- https://github.com/camunda/team-data-engine-maintenance/issues/49

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/AWSOpenSearchDatabaseCallbackDelegate.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/AWSOpenSearchDatabaseCallbackDelegate.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.utils;
+
+import static io.camunda.exporter.config.ConnectionTypes.OPENSEARCH;
+
+import io.camunda.exporter.config.ConnectionTypes;
+import io.camunda.exporter.config.ExporterConfiguration;
+import io.camunda.search.connect.os.OpensearchConnector;
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.opensearch.client.opensearch.OpenSearchClient;
+
+public class AWSOpenSearchDatabaseCallbackDelegate implements SearchDatabaseCallbackDelegate {
+
+  protected static final String IT_OPENSEARCH_AWS_INSTANCE_URL_PROPERTY =
+      "camunda.it.opensearch.aws.url";
+
+  protected SearchClientAdapter awsOsClientAdapter;
+  private OpenSearchClient osClient;
+
+  @Override
+  public void afterAll(final ExtensionContext context) throws IOException {}
+
+  @Override
+  public void afterEach(final ExtensionContext context) throws IOException {
+    if (context.getDisplayName().equals(OPENSEARCH.getType())) {
+      osClient.indices().delete(req -> req.index("*"));
+      osClient.indices().deleteIndexTemplate(req -> req.name("*"));
+    }
+  }
+
+  @Override
+  public void beforeAll(final ExtensionContext context) throws IOException {
+    final var osConfig = getConfigWithConnectionDetails(OPENSEARCH);
+    osClient = new OpensearchConnector(osConfig.getConnect()).createClient();
+    awsOsClientAdapter = new SearchClientAdapter(osClient);
+  }
+
+  @Override
+  public ExporterConfiguration getConfigWithConnectionDetails(
+      final ConnectionTypes connectionType) {
+    final var config = new ExporterConfiguration();
+    config.getIndex().setPrefix(CamundaExporterITInvocationProvider.CONFIG_PREFIX);
+    config.getBulk().setSize(1); // force flushing on the first record
+
+    final var openSearchAwsInstanceUrl =
+        Optional.ofNullable(System.getProperty(IT_OPENSEARCH_AWS_INSTANCE_URL_PROPERTY))
+            .orElseThrow();
+    config.getConnect().setUrl(openSearchAwsInstanceUrl);
+
+    config.getConnect().setIndexPrefix(CamundaExporterITInvocationProvider.CONFIG_PREFIX);
+    config.getConnect().setType(OPENSEARCH.getType());
+    return config;
+  }
+
+  // - name: Configure AWS Credentials for China region audience
+  //      uses: aws-actions/configure-aws-credentials@v4
+  //      with:
+  //        audience: sts.amazonaws.com.cn
+  //        aws-region: us-east-3
+  //        role-to-assume: arn:aws-cn:iam::123456789100:role/my-github-actions-role
+
+  @Override
+  public Map<ConnectionTypes, SearchClientAdapter> contextAdapterRegistration() {
+    return Map.of(OPENSEARCH, awsOsClientAdapter);
+  }
+}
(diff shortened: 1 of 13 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/AWSOpenSearchDatabaseCallbackDelegate.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.opensearch.client.opensearch.OpenSearchClient;
+
+public class AWSOpenSearchDatabaseCallbackDelegate implements SearchDatabaseCallbackDelegate {
+
+  protected static final String AWS_OPENSEARCH_SYSTEM_PROPERTY_NAME =
+      "camunda.integ.test.opensearch.aws_instance_url";
```
Zelldon, line 24: 🔧 
```suggestion
      "camunda.it.opensearch.aws_instance_url";
```

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/ContainerizedSearchDatabaseCallbackDelegate.java
```diff
@@ -0,0 +1,104 @@
...
+    osClient = new OpensearchConnector(osConfig.getConnect()).createClient();
+    osClientAdapter = new SearchClientAdapter(osClient);
+
+    final var elsConfig = getConfigWithConnectionDetails(ELASTICSEARCH);
+    elsClient = new ElasticsearchConnector(elsConfig.getConnect()).createClient();
+    elsClientAdapter = new SearchClientAdapter(elsClient);
+
+    closeables.add(elsContainer);
+    closeables.add(osContainer);
+  }
```
Zelldon, line 83: I think an alternative with less changes would be:

```suggestion
    
    // pre-create os client config
    final var osConfig = getDefaultConfig(OPENSEARCH);
 
    final var shouldRunAgainstAWSOS =
        Boolean.parseBoolean(            Optional.ofNullable(System.getProperty(SHOULD_RUN_AGAINST_AWS)).orElse("false"));
    if (shouldRunAgainstAWSOS) {
       // test should run against AWS Opensearch (external dependency)
       final var openSearchAwsInstanceUrl =
       
 Optional.ofNullable(System.getProperty(AWS_OPENSEARCH_SYSTEM_PROPERTY_NAME)).orElseThrow();
      osConfig.getConnect().setUrl(openSearchAwsInstanceUrl);
    } else {
        elsContainer = TestSearchContainers.createDefeaultElasticsearchContainer();
        osContainer = TestSearchContainers.createDefaultOpensearchContainer();
    
        elsContainer.start();
        osContainer.start();
   
        osConfig.getConnect().setUrl(osContainer.getHttpHostAddress());
        
        final var elsConfig = getDefaultConfig(ELASTICSEARCH);
        elsConfig.getConnect().setUrl(elsContainer.getHttpHostAddress());
        elsClient = new ElasticsearchConnector(elsConfig.getConnect()).createClient();
        elsClientAdapter = new SearchClientAdapter(elsClient);
        
        closeables.add(elsContainer);
        closeables.add(osContainer);
    }
    
    osClient = new OpensearchConnector(osConfig.getConnect()).createClient();
    osClientAdapter = new SearchClientAdapter(osClient);
  }
  
  
    @Override
  public ExporterConfiguration getDefaultConfig(
      final ConnectionTypes connectionType) {
    final var config = new ExporterConfiguration();
    config.getIndex().setPrefix(CamundaExporterITInvocationProvider.CONFIG_PREFIX);
    config.getBulk().setSize(1); // force flushing on the first record
    config.getConnect().setType(connectionType.getType());
    return config;
  }
  
```

ChrisKujawa (commented):
Hey Igor thanks for this! 

Makes in general sense to me. I think we can make it a bit simpler maybe 🤔 

I also think we don't need two properties, probably one is enough to validate whether we want to run against AWS. Meaning if URL is set we know we want to run IT against AWS. WDYT?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/ContainerizedSearchDatabaseCallbackDelegate.java
igpetrov: Good suggestion. I was worried that if I leave AWS OS in previous implementation (`CamundaExporterITInvocationProvider`), we might still at some point have additional logic specifically related to AWS OS or docker containers where at which moment it may become painful to refactor; on the other hand we don't have this issue now so maybe it is not a bad idea at all.
I will rewrite as you suggest.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/AWSOpenSearchDatabaseCallbackDelegate.java
igpetrov: Done

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/CamundaExporterITInvocationProvider.java
```diff
@@ -42,58 +32,27 @@ public class CamundaExporterITInvocationProvider
...
-    if (connectionType == ELASTICSEARCH) {
-      config.getConnect().setUrl(elsContainer.getHttpHostAddress());
-
-    } else if (connectionType == OPENSEARCH) {
-      config.getConnect().setUrl(osContainer.getHttpHostAddress());
+  private final SearchDatabaseCallbackDelegate callbackDelegate;
+
+  public CamundaExporterITInvocationProvider() {
+    final var shouldRunAgainstAWSOS =
+        Optional.ofNullable(System.getProperty(IT_OPENSEARCH_AWS_INSTANCE_URL_PROPERTY))
```
npepinpe, line 39: :wrench: Let's document this in the class Javadoc so it's clear how this invocation provider is meant to be used, how it can be configured, etc.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/SearchDatabaseCallbackDelegate.java
```diff
@@ -0,0 +1,23 @@
...
+package io.camunda.exporter.utils;
+
+import io.camunda.exporter.config.ConnectionTypes;
+import io.camunda.exporter.config.ExporterConfiguration;
+import java.util.Map;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+
+public interface SearchDatabaseCallbackDelegate
```
npepinpe, line 17: :wrench: Similarly, if we are going with this approach, we should understand how this is meant to be implemented and used in the future. e.g. do I have to implement after all? before each? What should I implement for `contextAdapterRegistration`, and when is it called?

So I would strongly urge to document this interface (as we should in general document interface/abstract types for similar reasons)

npepinpe (approved):
:question: If I understood correctly, does it mean that if the property is on, I only run the AWS OpenSearch context? So the expected usage is that we will have a separate workflow/CI job which will run with that property, and our normal CI pipeline would run with the containerized, right?

igpetrov:
>❓ If I understood correctly, does it mean that if the property is on, I only run the AWS OpenSearch context? So the expected usage is that we will have a separate workflow/CI job which will run with that property, and our normal CI pipeline would run with the containerized, right?

@npepinpe that is correct

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/SearchDatabaseCallbackDelegate.java
igpetrov: Yes, absolutely agree. I will do that before the merge.

Issue #23808, opened 2024-10-22
Global prefix is used in search


Zelldon:

ChrisKujawa:
@tmetzke @nathansandi what is the status of the prefix in the search clients?

tmetzke:
> @tmetzke @nathansandi what is the status of the prefix in the search clients?

As linked in the issue description, there is an active open PR from @emulbreh to complete this topic.

PR #24898, opened 2024-11-20
feat: support global prefix to userTaskFilterTransformer

nathansandi:
## Description

- Base implementation for client support the globalPrefix
- Include global prefix over the UserTaskFilterTrasnform
- Ajust newInstance calls to support the new parameter

It will be possible apply integration tests once all the indexes is adapted 

## Related issues

- https://github.com/camunda/camunda/issues/23808


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
```diff
@@ -62,10 +62,12 @@ public class SearchClients
   private final SecurityContext securityContext;
 
   public SearchClients(
-      final DocumentBasedSearchClient searchClient, final boolean isCamundaExporterEnabled) {
+      final DocumentBasedSearchClient searchClient,
+      final boolean isCamundaExporterEnabled,
+      final String globalPrefix) {
```
Zelldon, line 67: 🔧 I guess we can also just call it prefix. :)

search/search-client-query-transformer/src/test/java/io/camunda/search/clients/transformers/sort/AbstractSortTransformerTest.java
```diff
@@ -16,7 +16,7 @@
 
 public class AbstractSortTransformerTest {
 
-  private final ServiceTransformers transformers = ServiceTransformers.newInstance(false);
+  private final ServiceTransformers transformers = ServiceTransformers.newInstance(false, null);
```
Zelldon, line 19:
```suggestion
  private final ServiceTransformers transformers = ServiceTransformers.newInstance(false, "");
```

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/ServiceTransformers.java
```diff
@@ -96,9 +96,10 @@ public final class ServiceTransformers {
 
   private ServiceTransformers() {}
 
-  public static ServiceTransformers newInstance(final boolean isCamundaExporterEnabled) {
+  public static ServiceTransformers newInstance(
+      final boolean isCamundaExporterEnabled, final String globalPrefix) {
```
Zelldon, line 100: ❓ Why we have this boolean `isCamundaExporterEnabled` ?

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
```diff
@@ -104,8 +109,17 @@ public SearchQuery toSearchQuery(final UserTaskFilter filter) {
...
+          String.format(
+              TaskTemplate.FULL_QUALIFIED_INDEX_NAME_PATTERN,
+              TaskTemplate.formatIndexPrefix(globalPrefix),
+              TaskTemplate.COMPONENT_NAME,
+              TaskTemplate.INDEX_NAME,
+              TaskTemplate.INDEX_VERSION);
+
+      return List.of(indexName);
     }
+    // TDB - The listview is used when the exporter is not enabled, this will be removed in 8.8
```
Zelldon, line 122: Ah now I see. Hm but the Camunda Exporter will be standard right?

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
```diff
@@ -104,8 +109,17 @@ public SearchQuery toSearchQuery(final UserTaskFilter filter) {
...
   public List<String> toIndices(final UserTaskFilter filter) {
     if (isCamundaExporterEnabled) {
-      return List.of("tasklist-task-8.5.0_");
+      final String indexName =
+          String.format(
+              TaskTemplate.FULL_QUALIFIED_INDEX_NAME_PATTERN,
+              TaskTemplate.formatIndexPrefix(globalPrefix),
+              TaskTemplate.COMPONENT_NAME,
+              TaskTemplate.INDEX_NAME,
+              TaskTemplate.INDEX_VERSION);
```
Zelldon, line 118: 🔧 ❓ Maybe we can introduce a static method on the `TaskTemplate` that allows us to simple use it without constructing the format on our own.

Example:

```
   final String indexName = TaskTemplate.getIndexNameWithPrefix(globalPrefix);
```

In TaskTemplate:

```
  public static String getIndexNameWithPrefix(final String prefix) {
    return new TaskTemplate(prefix, false).getFullQualifiedName();
   }
```

Just as an idea to reuse the `getFullQualifiedName` but we could also have the format in the static again.

ChrisKujawa (approved):
Looks good to me in general, there are just some smaller hints. :)

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java
nathansandi: In. 8.8 yes 👍 , This will be kept for 8.7

ChrisKujawa: Even for 8.7 we will not have an 8.7 importer. :)

PR #24899, opened 2024-11-20
Add group update service

### LINKED ISSUE #24879: Camunda services support updating a group
### Description
Group updates should be supported in the `camunda-service` module. An example of the implementation of updating via the service module can be found [here](https://github.com/camunda/camunda/blob/main/service/src/main/java/io/camunda/service/UserServices.java#L62-L70)

### Acceptance Criteria
- [x] A `BrokerGroupUpdateRequest` exists hint: [BrokerUserUpdateRequest](https://github.com/camunda/camunda/blob/main/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/BrokerUserUpdateRequest.java)
  - [x] The field(s) that can be updated for a group are [`name`]
- [x] A method exists with signature `updateGroup(final GroupDTO groupDTO)`
- [x] The `GroupServiceTest` class should contain unit tests for the `updateGroup` method - hint: an example of working with the broker commands in tests can be found [here](https://github.com/camunda/camunda/blob/main/service/src/test/java/io/camunda/service/UserServiceTest.java#L66-L81)

koevskinikola:
## Description

Add `updateGroup` method to the `GroupService` class.

## Checklist

- [ ] for CI changes:

## Related issues

closes #24879

### DIFF
--- a/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/group/BrokerGroupUpdateRequest.java
+++ b/zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/group/BrokerGroupUpdateRequest.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.impl.broker.request.group;
+
+import io.camunda.zeebe.broker.client.api.dto.BrokerExecuteCommand;
+import io.camunda.zeebe.protocol.Protocol;
+import io.camunda.zeebe.protocol.impl.record.value.group.GroupRecord;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.intent.GroupIntent;
+import org.agrona.DirectBuffer;
+
+public class BrokerGroupUpdateRequest extends BrokerExecuteCommand<GroupRecord> {
+
+  private final GroupRecord requestDto = new GroupRecord();
+
+  public BrokerGroupUpdateRequest(final long groupKey) {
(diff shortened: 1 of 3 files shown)


zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/group/BrokerGroupUpdateRequest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.agrona.DirectBuffer;
+
+public class BrokerGroupUpdateRequest extends BrokerExecuteCommand<GroupRecord> {
+
+  private final GroupRecord requestDto = new GroupRecord();
+
+  public BrokerGroupUpdateRequest(final long groupKey) {
+    super(ValueType.GROUP, GroupIntent.UPDATE);
+    setPartitionId(Protocol.DEPLOYMENT_PARTITION);
+    request.setKey(groupKey);
```
nicpuppa, line 24: 🔧 If I'm not wrong the key should be sets into the requestDto also, can you double check ? I think in the processor we are using the groupKey from the record itself

nicpuppa (commented):
Looks good, I left just a question can you verify it ?

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/group/BrokerGroupUpdateRequest.java
koevskinikola: That's a good point. Yes, we get it from the record value. I will set it to both places to be safe and consistent.

nicpuppa: Another point, we should set the partition after set the key in the request. Otherwise the partition might be override by the `setKey` method

koevskinikola:
✅ @nicpuppa I applied your review feedback. The branch is rebased on `[nk-24815-group-create-service](https://github.com/camunda/camunda/tree/nk-24815-group-create-service)`, so it can be merged as soon as that one is finished merging.

nicpuppa (approved):

PR #24900, opened 2024-11-20
process migration implementation

panagiotisgts:
## Description

Implementation of the migration for the `operate-process` index to include form related fields.
Takes into account the 8.6 Importer completed for the `operate-process` index, utilizing the `completed` flag over the `operate-import-position` index.

> Opening as draft for review, will rebase this on main once it's parent is merged.
## Checklist

- [ ] for CI changes:

## Related issues

Part of: https://github.com/camunda/camunda/issues/23814

Split PR from origin: https://github.com/camunda/camunda/pull/24463


migration/process-migration/src/main/java/io/camunda/migration/process/adapter/Adapter.java
```diff
@@ -0,0 +1,95 @@
...
+    step.setAppliedDate(OffsetDateTime.now(ZoneId.systemDefault()));
+    return step;
+  }
+
+  default <T> T doWithRetry(
+      final String message, final Callable<T> callable, final RetryPredicate<T> retryPredicate)
+      throws Exception {
+    final ExponentialBackoffRetryDelay retryDelay =
+        new ExponentialBackoffRetryDelay(getMaxDelay(), getMinDelay());
+    return RetryOperation.<T>newBuilder()
```
Zelldon, line 84: ❓ Looks like we are the only consumer of this class. Can we either move this class to this module or use something battle-tested from some util dependency like guava

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/MigrationRepositoryIndex.java
```diff
@@ -0,0 +1,33 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migration.process.adapter;
+
+import io.camunda.webapps.schema.descriptors.operate.OperateIndexDescriptor;
+
+/* Fork of Operate's MigrationRepositoryIndex descriptor for correct prefixing */
+public class MigrationRepositoryIndex extends OperateIndexDescriptor {
```
Zelldon, line 13: ❓ Should we move this to the webapps-schema?

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/es/ElasticsearchAdapter.java
```diff
@@ -0,0 +1,248 @@
...
+    entities.forEach(entity -> migrateEntity(entity, bulkRequest));
+    final BulkResponse response;
+    try {
+      response =
+          doWithRetry(
+              "Migrate entities %s".formatted(idList),
+              () -> client.bulk(bulkRequest.build()),
+              (res) -> res == null || res.errors() || res.items().isEmpty());
+    } catch (final Exception e) {
+      throw new MigrationException("Failed to migrate entities %s".formatted(idList), e);
```
Zelldon, line 77: ❓ As far as I understood we retry ES and IO exceptions

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/Adapter.java
```diff
@@ -0,0 +1,95 @@
...
+    if (entity.getFormId() != null) {
+      updateMap.put(ProcessIndex.FORM_ID, entity.getFormId());
+    }
+    if (entity.getFormKey() != null) {
+      updateMap.put(ProcessIndex.FORM_KEY, entity.getFormKey());
+    }
+    return updateMap;
+  }
+
+  default ProcessorStep upsertProcessorStep(final String processDefinitionKey) {
```
Zelldon, line 68: ❌ Lets call it differently. It is not really doing an upsert. It is creating an entity.

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/os/OpensearchAdapter.java
```diff
@@ -0,0 +1,246 @@
...
+
+  @Override
+  public Duration getMinDelay() {
+    return properties.getMinRetryDelay();
+  }
+
+  @Override
+  public Duration getMaxDelay() {
+    return properties.getMaxRetryDelay();
+  }
```
Zelldon, line 221: ❓ I think most of these methods we don't need here or? We could also simply use the properties in the Runner?

ChrisKujawa (commented):
Thanks Panos. I tried to review thoroughly as possible, but I might missed maybe still something, as this PR is again quite large.

I guess we could even split this (for the next time) further. For example MigrationRunner + Adapter, then different implementations for OS and ES.

What I'm missing is tests, UT and IT. But I think they were part of the other PR right, maybe you add them later?

Most of my comments are questions, because I was not fully sure. But some are also request for changes, especially where we seem to miss to run callbacks.

We might be able to even reduce some code duplication with the search clients, but I might also oversee something.

Let me know when you want to sync on this and discuss this review f2f :)

migration/process-migration/src/main/java/io/camunda/migration/process/MigrationRunner.java
```diff
@@ -7,12 +7,125 @@
...
+    adapter.writeLastMigratedEntity(lastMigratedProcessDefinitionKey);
+    return lastMigratedProcessDefinitionKey;
+  }
+
+  private void scheduleNextBatch(final ScheduledExecutorService scheduler) {
+    try {
+      if (countdownTask == null && isImporterFinished()) {
+        startCountdown(scheduler);
+      }
+      scheduler.schedule(() -> {}, adapter.getMinDelay().toSeconds(), TimeUnit.SECONDS).get();
```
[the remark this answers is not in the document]
panagiotisgts: Yes it's expected. It's a simple delay to add between the migration rounds.

migration/process-migration/src/main/java/io/camunda/migration/process/MigrationRunner.java
```diff
@@ -7,12 +7,125 @@
...
+    }
+  }
+
+  private void startCountdown(final ScheduledExecutorService scheduler) {
+    LOG.info(
+        "Importer finished, migration will keep running for {}",
+        properties.getPostImporterTimeout());
+    countdownTask =
+        scheduler.schedule(
+            () -> {}, properties.getPostImporterTimeout().getSeconds(), TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
panagiotisgts: Yes, this `countdownTask` is a timer that's started when the 8.6 importer signal is _received_ for the condition to continue the migration.
```
private boolean shouldContinue(final List<ProcessEntity> processes) {
    if (!processes.isEmpty()) {
      return true;
    }
    return countdownTask == null || !countdownTask.isDone();
  }

File: migration/process-migration/src/main/java/io/camunda/migration/process/MigrationRunner.java [hunk 8]
```diff
@@ -7,12 +7,125 @@
...
+  }
+
+  private boolean isImporterFinished() {
+    final Set<ImportPositionEntity> importPositions;
+    try {
+      importPositions = adapter.readImportPosition();
+      return !importPositions.isEmpty()
+          && importPositions.stream().allMatch(ImportPositionEntity::getCompleted);
```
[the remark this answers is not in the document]
panagiotisgts: Yes, they are being handled correctly. The adapter will always return a collection empty or without null items and the `getCompleted` is a primitive boolean so it would be _false_  if not set.

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/MigrationRepositoryIndex.java
panagiotisgts: Do we have to? I think not, for greenfield 8.7 deployments you won't have to do migration and for existing ones this index would be already present and it won't be used again after 8.7.

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/os/OpensearchAdapter.java
```diff
@@ -0,0 +1,246 @@
...
+    processIndex = new ProcessIndex(connectConfiguration.getIndexPrefix(), false);
+    importPositionIndex = new ImportPositionIndex(connectConfiguration.getIndexPrefix(), false);
+    client = new OpensearchConnector(connectConfiguration).createClient();
+  }
+
+  @Override
+  public String migrate(final List<ProcessEntity> entities) throws MigrationException {
+    final BulkRequest.Builder bulkRequest = new BulkRequest.Builder();
+    final var idList = entities.stream().map(ProcessEntity::getId).toList();
+    entities.forEach(e -> migrateEntity(e, bulkRequest));
```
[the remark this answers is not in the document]
panagiotisgts: There is a chance that an entity threw some error in ES regarding the update, that's why you'd have to parse the response and find out which was the last successful updated entity.

Unfortunately, ES/OS do not perform these updates in a transactional fashion so you can always have partial failures.

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/os/OpensearchAdapter.java
```diff
@@ -0,0 +1,246 @@
...
+      searchResponse =
+          doWithRetry(
+              "Fetching next process batch",
+              () -> client.search(request, ProcessEntity.class),
+              res -> res.timedOut() || Boolean.TRUE.equals(res.terminatedEarly()));
+    } catch (final Exception e) {
+      throw new MigrationException("Failed to fetch next processes batch", e);
+    }
+    return searchResponse.hits().hits().stream().map(Hit::source).toList();
+  }
```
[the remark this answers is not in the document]
panagiotisgts: Technically we could, however this would require additional dependencies to be loaded in the module. Also, generally speaking I believe that data migrations are better to be interacting with the database directly and not include intermediates. Reduce liability factors :smiling_face_with_tear:

migration/process-migration/src/main/java/io/camunda/migration/process/util/ProcessModelUtil.java
```diff
@@ -0,0 +1,87 @@
...
+import io.camunda.zeebe.model.bpmn.instance.StartEvent;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeFormDefinition;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeProperties;
+import java.io.ByteArrayInputStream;
+import java.util.Optional;
+import org.camunda.bpm.model.xml.instance.ModelElementInstance;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProcessModelUtil {
```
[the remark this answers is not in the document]
panagiotisgts: I think you missed the commit containing the tests :smile: .

migration/process-migration/src/test/java/io/camunda/migration/process/it/es/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,510 @@
...
+    assertThat(stepRecords.getFirst().getContent()).isEqualTo("2");
+  }
+
+  @Test
+  public void shouldMigrateDuringCountdown() throws IOException {
+    // given
+    properties.setImporterFinishedTimeout(Duration.ofSeconds(2));
+    properties.setMinRetryDelay(Duration.ofSeconds(3));
+    properties.setMaxRetryDelay(Duration.ofSeconds(3));
+    properties.setBatchSize(4);
```
ChrisKujawa, line 350: 🔧 Lets try to find smaller numbers for this or maybe a different way to control the behavior, like controllable clock or something 🤔

migration/process-migration/src/test/java/io/camunda/migration/process/it/es/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,510 @@
...
+                .until(
+                    () -> {
+                      migrator.run();
+                      return true;
+                    }));
+
+    final var records =
+        readRecords(ProcessorStep.class, migrationRepositoryIndex.getFullQualifiedName());
+    assertThat(records).isEmpty();
+  }
```
ChrisKujawa, line 401:
```suggestion
  @Test
  public void shouldRunIndefinitelyWhenANonCompletedImportPositionExists() throws IOException {
  final var latch = new CountDownLatch(1)
    writeImportPositionToIndex(
        TestData.notCompletedImportPosition(1), TestData.notCompletedImportPosition(2));
    esClient.indices().refresh();
    awaitRecordsArePresent(ImportPositionEntity.class, importPositionIndex.getFullQualifiedName());

    assertThrows(
        ConditionTimeoutException.class,
        () ->
            Awaitility.await()
                .atMost(Duration.ofSeconds(10))
                .until(
                    () -> {
                      migrator.run();
                      latch.countDown()
                      return true;
                    }));


    assertThaT(latch.get()).isEqual(1)
    TestData.completedImportPosition(1)
    latch.await();
    assertThaT(latch.get()).isEqual(0)
    
    final var records =        readRecords(ProcessorStep.class, migrationRepositoryIndex.getFullQualifiedName());
    assertThat(records).isEmpty();
  }
```

migration/process-migration/src/test/java/io/camunda/migration/process/it/es/ElasticsearchMigrationRunnerIT.java
```diff
@@ -0,0 +1,510 @@
...
+  @Test
+  @Order(Integer.MAX_VALUE)
+  public void shouldThrowException() {
+    ES_CONTAINER.close();
+    properties.setMaxRetries(2);
+    properties.setMinRetryDelay(Duration.ofSeconds(1));
+
+    final var ex = assertThrows(MigrationException.class, migrator::run);
+    assertThat(ex.getMessage()).isEqualTo("Failed to fetch last migrated process");
+  }
```
ChrisKujawa, line 443: 🔧 Let's try to create a config with wrong connection URL to not stop container

ChrisKujawa (approved):
Thank you very much! 🚀

ChrisKujawa:
In general we should add after we finalized the migration application add a E2E/QA test on root level that actually does an 8.6 to 8.7 update.

migration/process-migration/src/test/java/io/camunda/migration/process/it/AdapterTest.java
```diff
@@ -0,0 +1,278 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.testcontainers.OpensearchContainer;
+import org.testcontainers.elasticsearch.ElasticsearchContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+public abstract class AdapterTest {
```
ChrisKujawa, line 53: 🔧 I think we could also do it similar as we did in the exporter or root qa with the test providers instead of an abstract class, but these are just details. I think we are also fine to go ahead, with in mind that we want to get rid of it soon again.

migration/process-migration/src/test/java/io/camunda/migration/process/it/MigrationRunnerIT.java
```diff
@@ -494,17 +444,19 @@ private <T> List<T> readRecords(final Class<T> clazz, final String indexName) th
...
   }
 
+  @Override
   private void assertProcessorStepContentIsStored(final String processDefinitionId)
       throws IOException {
     final var records =
         readRecords(ProcessorStep.class, migrationRepositoryIndex.getFullQualifiedName());
     assertThat(records.size()).isEqualTo(1);
     assertThat(records.getFirst().getContent()).isEqualTo(String.valueOf(processDefinitionId));
-  }
```
ChrisKujawa: ❌ We should remove the out commented code :)

ChrisKujawa (approved):
Thanks @panagiotisgts 🚀

migration/process-migration/src/main/java/io/camunda/migration/process/util/AdapterRetryDecorator.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final double RETRY_DELAY_MULTIPLIER = 2;
+  private final ProcessMigrationProperties properties;
+
+  public AdapterRetryDecorator(final ProcessMigrationProperties properties) {
+    this.properties = properties;
+  }
+
+  public <T> T decorate(
+      final String message, final Callable<T> callable, final Predicate<T> retryPredicate)
+      throws Exception {
```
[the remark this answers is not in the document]
panagiotisgts: Unfortunately not because of the typed `Predicate<T>` that's being passed in order to have access to the Response of the respective request. It could work with a base builder and casting `<?>` to `<T>` and building on the same RetryConfig each time but could have side effects.

PR #24903, opened 2024-11-20
feat: add decision evaluation metric exporter handler

### LINKED ISSUE #22997: Implement writing usage metrics for decision instances in Camunda exporter
### Description

When working on Export handlers we lost piece of functionality related with usage metrics for decision instances that [we have in importer](https://github.com/camunda/camunda/blob/main/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/DecisionEvaluationZeebeRecordProcessor.java#L71).

### TODO

* Write metric records for every decision instance.

### Additional Information

Similar functionality for process instances: https://github.com/camunda/camunda/blob/feat-operate-exporter/zeebe/exporters/operate-exporter/src/main/java/io/camunda/zeebe/operate/exporter/handlers/MetricFromProcessInstanceHandler.java

panagiotisgts:
## Description

Export Metrics from DecisionEvaluation records
## Checklist

- [ ] for CI changes:

## Related issues

closes #22997

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
@@ -13,6 +13,7 @@ import io.camunda.exporter.store.BatchRequest;
 import io.camunda.webapps.schema.entities.MetricEntity;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.intent.DecisionEvaluationIntent;
 import io.camunda.zeebe.protocol.record.value.DecisionEvaluationRecordValue;
 import java.time.Instant;
 import java.time.OffsetDateTime;
@@ -45,7 +46,8 @@ public class MetricFromDecisionEvaluationHandler
 
   @Override
   public boolean handlesRecord(final Record<DecisionEvaluationRecordValue> record) {
-    return true;
+    return record.getIntent().name().equals(DecisionEvaluationIntent.EVALUATED.name())
+        || record.getIntent().name().equals(DecisionEvaluationIntent.FAILED.name());
   }
 
   @Override
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.handlers;
+
+import static io.camunda.exporter.utils.ExporterUtil.tenantOrDefault;
+
+import io.camunda.exporter.store.BatchRequest;
+import io.camunda.webapps.schema.entities.MetricEntity;
+import io.camunda.zeebe.protocol.record.Record;
+import io.camunda.zeebe.protocol.record.ValueType;
+import io.camunda.zeebe.protocol.record.value.DecisionEvaluationRecordValue;
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MetricFromDecisionEvaluationHandler
+    implements ExportHandler<MetricEntity, DecisionEvaluationRecordValue> {
+
+  protected static final String EVENT_DECISION_INSTANCE_EVALUATED =
+      "EVENT_DECISION_INSTANCE_EVALUATED";
+  private static final String ID_PATTERN = "%d-%d";
+
+  private final String indexName;
+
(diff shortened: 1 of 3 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
```diff
@@ -0,0 +1,90 @@
...
+      final Record<DecisionEvaluationRecordValue> record, final MetricEntity entity) {
+    final DecisionEvaluationRecordValue recordValue = record.getValue();
+    final OffsetDateTime timestamp =
+        OffsetDateTime.ofInstant(Instant.ofEpochMilli(record.getTimestamp()), ZoneOffset.UTC);
+    final String tenantId = tenantOrDefault(recordValue.getTenantId());
+    final String value = entity.getId();
+    entity
+        .setEvent(EVENT_DECISION_INSTANCE_EVALUATED)
+        .setValue(value)
+        .setId(null)
```
mustafadagher, line 76: ❓ is that intentional to remove the id?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  @Override
+  public Class<MetricEntity> getEntityType() {
+    return MetricEntity.class;
+  }
+
+  @Override
+  public boolean handlesRecord(final Record<DecisionEvaluationRecordValue> record) {
+    return true;
```
mustafadagher, line 48: ❓ do we really want to handle all intents here? or only `EVALUATED`?

panagiotisgts: Judging from the previous [implementation](https://github.com/camunda/camunda/blob/stable/8.6/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/DecisionEvaluationZeebeRecordProcessor.java) I think we want to consume both `EVALUATED` and `FAILED` intents.

mustafadagher: that's interesting since the event name is EVENT_DECISION_INSTANCE_`EVALUATED`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
sdorokhova: I think we use ids generated by Elastic/Opensearch here, as we never update the records.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
sdorokhova: I think from business side even when execution failed decision instance was still technically executed and we should count this. We can double check with PM of course, but since it was implemented like this from the beginning, I would rather say that metric name is not fully correct. 
But again we can't change it as it's a part of contract and the metric is being documented and used this way 🤷

mustafadagher: can we then at least add another tag to the metric to separate evaluated than failed?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
mustafadagher: does that mean that we should bring back the Id here and also in the other metric handlers?

panagiotisgts: It's not mandatory I believe to do so, ES/OS can generate the IDs since the only reason these records exist are to be counted i believe, to determine usage.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandler.java
panagiotisgts: We could theoretically split these events in two and also change the usage metrics query, but I'm not entirely sure where that expands to and what it could affect.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/MetricFromDecisionEvaluationHandlerTest.java
```diff
@@ -0,0 +1,149 @@
...
+    assertThat(underTest.getHandledValueType()).isEqualTo(ValueType.DECISION_EVALUATION);
+  }
+
+  @Test
+  void testGetEntityType() {
+    assertThat(underTest.getEntityType()).isEqualTo(MetricEntity.class);
+  }
+
+  @Test
+  void shouldHandleRecord() {
```
mustafadagher, line 51: 🔧 this doesn't verify that we can handle `FAILED` too.

mustafadagher (approved):
Based on the discussion on slack. All look good to me now! 🎉 

I have one minor suggestion in the tests below.

mustafadagher (approved):

PR #24921, opened 2024-11-20
Preparation work for post importer migration

npepinpe:
## Description

This PR includes some preparation work for the migration of the post importer:

- Renames the `ElasticsearchRepository` and `OpenSearchRepository` to include `Archiver` in the name, since we will soon have similar repository patterns but specifically for the incident update task.
- Extracts the construction of the `BackgroundTaskManager` to a factory, as we were starting to have a _lot_ of static methods in there to do so, and each had to carry the state around as arguments.
- Changes the signature of the repositories to return `CompletionStage` instead of `CompletableFuture`, to indicate only the repositories should be able to complete the results.

## Related issues

related to #24297

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -32,6 +32,7 @@ import io.camunda.exporter.schema.SchemaManager;
 import io.camunda.exporter.store.BatchRequest;
 import io.camunda.exporter.store.ExporterBatchWriter;
 import io.camunda.exporter.tasks.BackgroundTaskManager;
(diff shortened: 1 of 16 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -106,11 +108,7 @@ public void open(final Controller controller) {
     writer = createBatchWriter();
 
     scheduleDelayedFlush();
-
-    // // start archiver after the schema has been created to avoid transient errors
-    if (configuration.getArchiver().isRolloverEnabled()) {
-      taskManager.start();
-    }
+    taskManager.start();
```
entangled90, line 111: :question: does this change the behavior of the component at startup or is the configuration always true?

entangled90 (approved):
:rocket:

PR #24947, opened 2024-11-21
refactor: add webapp user fields to central user object to support auth transition

Ben-Sheppard:
## Description

This PR ensures that the fields on the `CamundaUser` central object can support the existing fields on the webapp specific user objects, this is important as it allows us to transition the webapp auth implementations to the `camunda-authentication` and reduce the impact on the webapp UIs.

### DIFF
--- a/authentication/src/main/java/io/camunda/authentication/service/BasicCamundaUserService.java
+++ b/authentication/src/main/java/io/camunda/authentication/service/BasicCamundaUserService.java
@@ -28,9 +28,6 @@ public class BasicCamundaUserService implements CamundaUserService {
 
   @Override
   public String getCurrentUsername() {
-    final var authenticatedUser =
-        (CamundaUser) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
-
-    return authenticatedUser.getUsername();
+    return getCurrentUser().getUsername();
   }
 }
--- a/authentication/src/main/java/io/camunda/authentication/service/BasicCamundaUserService.java
+++ b/authentication/src/main/java/io/camunda/authentication/service/BasicCamundaUserService.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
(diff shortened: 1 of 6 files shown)


security/security-core/src/main/java/io/camunda/security/entity/ClusterMetadata.java
```diff
@@ -0,0 +1,119 @@
...
+        + "uuid='"
+        + uuid
+        + '\''
+        + ", name='"
+        + name
+        + '\''
+        + ", urls="
+        + urls
+        + '\''
+        + '}';
```
emulbreh, line 93:
```suggestion
    return "ClusterMetadata{uuid='%s', name='%s', urls='%s'}".formatted(uuid, name, urls);
```

security/security-core/src/main/java/io/camunda/security/entity/ClusterMetadata.java
```diff
@@ -0,0 +1,119 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.security.entity;
+
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
```
emulbreh, line 15: Do we need both of these copyright notices?

authentication/src/main/java/io/camunda/authentication/service/BasicCamundaUserService.java
```diff
@@ -0,0 +1,36 @@
...
+  public String getUserToken() {
+    return null;
+  }
+
+  @Override
+  public String getCurrentUsername() {
+    final var authenticatedUser =
+        (CamundaUser) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+
+    return authenticatedUser.getUsername();
```
emulbreh, line 34:
```suggestion
    return getCurrentUser().getUsername();
```

emulbreh:
The code looks sensible to me, but I don't have enough context yet to understand what it actually achieves.

emulbreh (approved):

PR #24950, opened 2024-11-21
fix: enable the old backup endpoint for gateway when it is deployed standalone

### LINKED ISSUE #24850: Ensure endpoints are backward compatible when zeebe is deployed standalone

entangled90:
## Description
Added a new Profiles class that contain these expression for reusability and so they can be tested in a unit test

## Related issues

closes #24850

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/shared/management/BackupEndpointStandalone.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/management/BackupEndpointStandalone.java
@@ -7,19 +7,19 @@
  */
 package io.camunda.zeebe.shared.management;
 
+import io.camunda.zeebe.shared.profiles.ProfileStandaloneBrokerOrGateway;
 import org.springframework.boot.actuate.endpoint.annotation.DeleteOperation;
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.boot.actuate.endpoint.annotation.Selector;
 import org.springframework.boot.actuate.endpoint.annotation.WriteOperation;
 import org.springframework.boot.actuate.endpoint.web.WebEndpointResponse;
 import org.springframework.boot.actuate.endpoint.web.annotation.WebEndpoint;
-import org.springframework.context.annotation.Profile;
 import org.springframework.lang.NonNull;
 import org.springframework.stereotype.Component;
 
 @Component
 @WebEndpoint(id = "backups")
-@Profile("broker & standalone")
+@ProfileStandaloneBrokerOrGateway
 public final class BackupEndpointStandalone {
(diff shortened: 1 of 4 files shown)


dist/src/main/java/io/camunda/zeebe/shared/Profiles.java
```diff
@@ -0,0 +1,15 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.shared;
+
+public class Profiles {
+  @SuppressWarnings("unused")
+  public static final String BROKER_STANDALONE = "broker & standalone";
+
+  public static final String BROKER_OR_GATEWAY_STANDALONE = "(broker | gateway) & standalone";
```
npepinpe, line 14: :question: Couldn't we use meta annotations? For example:

```java
@Target({ElementType.METHOD})
@Documented
@Inherited
@Profile("broker & standalone")
public @interface ProfileStandaloneBroker {}
```

Then:

```java
@Component
@WebEndpoint(id = "backups")
@ProfileStandaloneBrokerProfile
public final class BackupEndpointStandalone {
```

I don't have a strong argument: it's clearer to me with the custom annotation, but it's likely a personal preference.

That said, Spring already provides a `Profiles` class, so I would at least rename it if we keep this approach.

npepinpe (commented):
:rocket: 

One thing, I need to check what happens with the standalone gateway now. I guess it's just the REST API, and no web apps frontend? Would I expect it to be able to take a webapp backup as a user? :thinking:

dist/src/main/java/io/camunda/zeebe/shared/Profiles.java
entangled90: I originally put them as simple Strings so they can be easily tested with a standalone unit test. However I split "BackupEndpointStandaloneTest" with both profiles so I guess we are covering both cases anyway :shrug: 

If you see no values in the `ProfileTest` we can proceed with the meta annotations :+1:

npepinpe: It's really just testing Spring - I understand the idea is to make sure if we ever change them something breaks (so encoding expectations/behavior), but it's likely still low value.

I didn't push it as a change request though because again, I think both approaches are fine - I don't have strong arguments for the meta annotation, just that it's easier to understand and maintain for me, but that's pretty weak :smile:

entangled90: Let's go with the meta annotation :+1:

npepinpe (approved):
:rocket:

PR #24952, opened 2024-11-21
Reiterate over Camunda Exporter IT (Handlers)

### LINKED ISSUE #24502: Reiterate over how the current Integration Tests are written, and consider if they need to be changed or improved
Refer to discussion points in https://github.com/camunda/camunda/pull/24405#pullrequestreview-2417839329 and https://github.com/camunda/camunda/pull/24405#pullrequestreview-2421432800 .

The changes to move from the current setup to the one I suggested should be fairly simple.

We already have a setup where we take a handler and we generate a record that fits this handler, then we also generate an expected entity, and we assert that the exporter exports an record similar to the expected.

Now we will still generate a record that has a specific value type and we export it, the changes will be:
- Rather than mocking the exporter provider, we use the real default one, so all handlers should be picked up and ran for this record
- Rather than expecting one entity as a result, it could be several, it could mean doing different updates to the same entity. 
 
Which is how we think of exporting in principal, we want to manipulate some documents based on some zeebe states, do we do it using handlers, or do we embed it directly in exporter export(), the IT shouldn't care.

ChrisKujawa:
### Current state:

1. We have unit tests for all our components   
    - Schema Manager
    - All handlers
    - BatchWriter
    - Config
    - etc.

7. We have integration tests for all 
    - Schema Manager 
    - Camunda Exporter
    - Handlers <- here we have some issues

:information_source: The list might not be fully complete

### Recap - current challenges:

- The IT for the exporter handler seem to be quite brittle and synthetic, as they have a specific setup and only consider one handler at once
- This is why this issue exists https://github.com/camunda/camunda/issues/24502
- Adding a test with multiple handlers seems to be problematic for others at least the current setup doesn't really support such https://camunda.slack.com/archives/C06UYJMMETZ/p1731938072246229
  - Necessary when you consider entity updates/deletions, that need creation first

### Discussion / Proposal:

If we think about the testing pyramid again and think about our modules/components we can see that we have already quite high coverage on the UT level.

testing-pyramind
integration-test-exporter

- We make sure that all handlers are unit-tested
- We even verify that the correct requests are executed on the client (like add/delete, etc.)
  - That the client does the right thing, is nothing we need to validate/verify in this component/module. We can expect this to be done elsewhere.

What I think we only need in our Camunda Exporter IT is one or two real integration test that:

1. Sets up the Camunda Exporter with all default handlers
2. Connects to ES/OS
3. Creates the schemas
4. Exporters one specific record, fex for variables
5. Then we verify that data is written on ES/OS (we can do it similarly as we did before with handler IT)

This would allow us to verify whether we are able to communicate with ES/OS, can create the schema, and can write/export into it. Everything else should be already tested on the UT level.

For some special cases, we can consider still adding some more IT, for example when we know that handlers are updating the same entity (like for the ListView). This might be something worth verifying as well. Or when we expect we can delete an entity after it was created.

### This PR 

  * Removes the CamundaHandlerIT as it tests handlers alone only. 
  * Replace this with one IT validating the described above 

In the next iteration - follow up to validate
    * like exporting a simple variable record, and exporting something that covers multiple updates like PI for list view
    *  Others can add more tests then based on their special needs (if they have any).

## Related issues

closes https://github.com/camunda/camunda/issues/24502


npepinpe:
If I understand correctly, focus here is mostly the handlers and how we test in the exporter module, right? That makes sense.

What we really need is a good set of E2E/QA/black box tests which are completely data store agnostic - for example, are incidents propagated along the tree path in the UI? If I deploy a process and create an instance, can I query it eventually? etc. These would also speed up development in the future if we add more data stores.

npepinpe (approved):

ChrisKujawa:
> If I understand correctly, focus here is mostly the handlers and how we test in the exporter module, right? That makes sense.
> 
> What we really need is a good set of E2E/QA/black box tests which are completely data store agnostic - for example, are incidents propagated along the tree path in the UI? If I deploy a process and create an instance, can I query it eventually? etc. These would also speed up development in the future if we add more data stores.

Totally agree 👍🏼

Issue #22392, opened 2024-09-17
Implement `mapping-rule` endpoints


Ben-Sheppard:
### Description

Mapping rule endpoints provide CRUD options for mapping rules.

* The endpoints should be implemented as described in [REST API Developer Guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)
* `{place-holder}` is the placeholder for parameters like `key`

### Endpoints

- [x] Create mapping rule - `POST /v2/mapping-rules` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.9ula0nupgc8f)
- [x] Search mapping rules - `POST /v2/mapping-rules/search` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.ac4uy4iju5se)
- [ ] Get a single mapping rule - `GET /v2/mapping-rules/{key}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.xeantwy19aio)
- [ ] Update a mapping rule - `PATCH /v2/mapping-rules/{key}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.d4udhz79t2ne)
- [x] Delete a mapping rule - `DELETE /v2/mapping-rules/{key}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.pzdfgevzrw6o)s

### Acceptance Criteria

- [ ] Unit tests
- [ ] Integration tests

PR #24960, opened 2024-11-21
Add create mapping endpoint

nicpuppa:
## Description

This PR adds the create mapping endpoint

## Related issues

related to #22392

### DIFF
--- a/service/src/main/java/io/camunda/service/MappingServices.java
+++ b/service/src/main/java/io/camunda/service/MappingServices.java
@@ -75,5 +75,5 @@ public class MappingServices
         .findFirst();
   }
 
-  public record MappingDTO(Long mappingKey, String claimName, String claimValue) {}
+  public record MappingDTO(String claimName, String claimValue) {}
 }
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -46,7 +46,7 @@ tags:
   - name: User
   - name: User task
   - name: Variable
-  - name: Mapping
+  - name: Mapping Rule
 
 paths:
   /topology:
@@ -2204,7 +2204,7 @@ paths:
   /mapping-rules:
     post:
       tags:
-        - Mapping
+        - Mapping Rule
       operationId: createMappingRule
       summary: Create mapping rule
       description: |
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -46,6 +46,7 @@ tags:
   - name: User
   - name: User task
   - name: Variable
+  - name: Mapping
 
 paths:
   /topology:
@@ -1985,7 +1986,6 @@ paths:
             More details are provided in the response body.
         "500":
           $ref: "#/components/responses/InternalServerError"
-
   /authorizations/search:
     post:
       tags:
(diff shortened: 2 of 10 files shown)


christinaausley (approved):
No grammatical adjustments on my end 👍

service/src/main/java/io/camunda/service/MappingServices.java
```diff
@@ -75,5 +75,5 @@ public Optional<MappingEntity> findMapping(final Long mappingKey) {
         .findFirst();
   }
 
-  public record MappingDTO(Long mappingKey, String claimName, String claimValue) {}
+  public record MappingDTO(String claimName, String claimValue) {}
```
emulbreh, line 78: We keep the key in the DTO for users, primarily for updates as far as I can tell. I think it's actually nicer to keep the key outside so it doesn't have to be nullable, but let's make it consistent across all entities.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2201,6 +2200,50 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+
+  /mapping-rules:
+    post:
+      tags:
+        - Mapping
```
emulbreh, line 2207: Let's make this "Mapping Rule" for clarity.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -325,6 +328,13 @@ public static Either<ProblemDetail, UserDTO> toUserDTO(
                 passwordEncoder.encode(request.getPassword())));
   }
 
+  public static Either<ProblemDetail, MappingDTO> toMappingRequest(
```
emulbreh, line 331:
```suggestion
  public static Either<ProblemDetail, MappingDTO> toMappingDTO(
```
for consistency with toUserDTO().

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/MappingControllerTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+  @MockBean private MappingServices mappingServices;
+
+  @BeforeEach
+  void setup() {
+    when(mappingServices.withAuthentication(any(Authentication.class))).thenReturn(mappingServices);
+  }
+
+  @Test
+  void createMappingShouldReturnAccepted() {
```
emulbreh, line 42:
```suggestion
  void createMappingShouldReturnCreated() {
```

service/src/main/java/io/camunda/service/MappingServices.java
emulbreh: This isn't from this PR, but I think it'd be good to standardize the name of these objects to either "mapping rule" or just "mapping". We don't have to fix it in this PR though.

emulbreh (approved):

service/src/main/java/io/camunda/service/MappingServices.java
nicpuppa: > We keep the key in the DTO for users, primarily for updates as far as I can tell. I think it's actually nicer to keep the key outside so it doesn't have to be nullable, but let's make it consistent across all entities.

It's not in the scope of this PR, so I'll create a follow-up for this. Same for the second comment

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
nicpuppa: We should create a follow-up also for this case, we have a mix of `DTO` and `request` object. Can you do this ?

PR #24961, opened 2024-11-21
feat: Trigger assignment listeners after UserTask creation with defined `assignee`

### LINKED ISSUE #24233: Start assignment when finalizing creation
The `assignment` task listener should trigger when a user task with [the `assignee` property](https://docs.camunda.io/docs/next/components/modeler/bpmn/user-tasks/#assignments) specified is created.

To support this on creation specifically, we'll need to do one of:
1. ❌ append an `ASSIGN` command at the end of finalizing the `CREATE` command
2. 🆗 append an `ASSIGNING` event and either create a task listener job for the `assignment` event type or immediately finalize it with the `ASSIGNED` event

> [!NOTE]
> Option 1 could produce a race condition if the user manages to append a command for this user task before the `CREATE` command is finalized, i.e. before the `ASSIGN` is appended. In that case, the user's command would be processed before the engine has assigned it, even though the user task had an `assignee` property defined. This would be possible from within a `create` task listener, as the user task key is known at that time. This increases the possibility of users trying this by accident. Therefore, I propose we instead implement this using option 2.

This concept raises other some non-blocking questions: 
- **Should the `assignee` be set in the `CREATED` event before the task has been `ASSIGNING`?**
- **Should the `assignee` be available to the `create` task listeners?**

ce-dmelnych:
## Description

This pull request introduces the functionality for triggering `assignment` listeners when a UserTask is created with a defined `assignee` property.

Key changes:
- When a UserTask is configured with a predefined `assignee`:
  - `ASSIGNING` event is emitted after UserTask creation.
  - If `assignment` listeners are defined, they are triggered between the `ASSIGNING` and `ASSIGNED` events.
  - Without `assignment` listeners, the `ASSIGNED` event is emitted immediately after `ASSIGNING`.

## Checklist

- [ ] for CI changes:


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateUserTaskTest.java
```diff
@@ -208,9 +208,9 @@ public void shouldWriteMigratedEventForUserTask() {
                     Map.entry("followup2", "PT10H")))
             .create();
 
-    // await user task creation
+    // await user task assignment
     final var userTask =
-        RecordingExporter.userTaskRecords(UserTaskIntent.CREATED)
+        RecordingExporter.userTaskRecords(UserTaskIntent.ASSIGNED)
```
ana-vinogradova-camunda, line 213: ❓ Not sure why MigrateUserTaskTest was affected?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
```diff
@@ -75,9 +77,13 @@ public void onCommand(
   public void onFinalizeCommand(
       final TypedRecord<UserTaskRecord> command, final UserTaskRecord userTaskRecord) {
     final long userTaskKey = command.getKey();
+    final String action =
+        Objects.requireNonNullElse(
+            StringUtils.firstNonEmpty(command.getValue().getAction(), userTaskRecord.getAction()),
+            StringUtils.EMPTY);
```
ana-vinogradova-camunda, line 83: 👍 I like that, very clear sequence.

ce-dmelnych:
I've attempted to clean up the commit history of this PR after accidentally updating it with a merge commit. Unfortunately, something went wrong during the process, which resulted in a strange state of this PR: "Processing update" message appeared with an infinite spinner next to PR name, blocked `Update branch` button.
image

So, I created a new PR (#25253) and abandoning this one.
Apologies for any inconvenience caused.

PR #24962, opened 2024-11-21
refactor: reference users with keys instead of usernames

emulbreh:
Using keys for all entity references will simplify how we implement common functionality.

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
@@ -87,7 +87,7 @@ public class UserCreateProcessor implements DistributedTypedRecordProcessor<User
     command.getValue().setUserKey(key);
 
     stateWriter.appendFollowUpEvent(key, UserIntent.CREATED, command.getValue());
-    addUserPermissions(key, username);
+    addUserPermissions(key);
     responseWriter.writeEventOnCommand(key, UserIntent.CREATED, command.getValue(), command);
 
     distributionBehavior
@@ -103,16 +103,16 @@ public class UserCreateProcessor implements DistributedTypedRecordProcessor<User
     distributionBehavior.acknowledgeCommand(command);
   }
 
-  private void addUserPermissions(final long key, final String username) {
+  private void addUserPermissions(final long key) {
     final var authorizationRecord =
         new AuthorizationRecord()
             .setOwnerKey(key)
             .setOwnerType(AuthorizationOwnerType.USER)
             .setResourceType(AuthorizationResourceType.USER)
             .addPermission(
-                new Permission().setPermissionType(PermissionType.READ).addResourceId(username))
+                new Permission().setPermissionType(PermissionType.READ).addResourceId(key))
             .addPermission(
-                new Permission().setPermissionType(PermissionType.UPDATE).addResourceId(username));
+                new Permission().setPermissionType(PermissionType.UPDATE).addResourceId(key));
     commandWriter.appendFollowUpCommand(
         key, AuthorizationIntent.ADD_PERMISSION, authorizationRecord);
   }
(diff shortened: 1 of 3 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
```diff
@@ -103,16 +103,16 @@ public void processDistributedCommand(final TypedRecord<UserRecord> command) {
...
         new AuthorizationRecord()
             .setOwnerKey(key)
             .setOwnerType(AuthorizationOwnerType.USER)
             .setResourceType(AuthorizationResourceType.USER)
             .addPermission(
-                new Permission().setPermissionType(PermissionType.READ).addResourceId(username))
+                new Permission().setPermissionType(PermissionType.READ).addResourceId(key))
             .addPermission(
-                new Permission().setPermissionType(PermissionType.UPDATE).addResourceId(username));
+                new Permission().setPermissionType(PermissionType.UPDATE).addResourceId(key));
```
remcowesterhoud, line 115: ❌ The resource identifier for a user is the username, not the key. The permission we describe here is that the user has read permissions for users with a specific username.

These resource identifiers are exposed to users. It's easier for them to use the username, instead of some arbitrary key. Changing this would impact the UX negatively.

remcowesterhoud (commented):
@emulbreh could you explain what you're trying to do here? I'm missing a lot of context. Do you have an issue this PR refers to? We don't want to change resource identifiers from the username to a key.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserCreateProcessor.java
emulbreh: I'm trying to understand the concept of resource identifiers vs resource keys and other identifiers and when we're using which. There's a thread in the project channel with more questions. The change here just seemed like low hanging fruit.

The UI can map keys to a display value. That's something we'd want to do for most entities eventually so that we can link to detail pages or show a tooltip inline.

Is there something I can read about the resource identifier concept?

remcowesterhoud: [This](https://docs.google.com/document/d/1ikxw9ARCv-MBwASnj7WhW5uHCndCRk2B35xi-cB2N0M/edit?tab=t.0#heading=h.m2zg2f7pb01i) document is a start. It somewhat explains the permissions concept in C8.

emulbreh: While this does list examples of how entities can be selected for resources-level permissions, it's not clear to me why we're using a username instead of a key or how we should be referencing roles. 
Are we using different properties for backwards compatibility?

remcowesterhoud: >it's not clear to me why we're using a username instead of a key

Technically it'd be possible to use keys as well. I wasn't part of the conversations when this document was initially created and accepted the use of usernames. We defined that usernames should be unique so in that sense it shouldn't matter if we use the key or a username. Add to that that this data is exposed to users and they need to provide this identifier through the API it was more convenient to use the username directly. We could blindly copy the input the user provided instead of having a mapping to find the key and use this instead.

This also aligns with the approach that was taken in C7, which hopefully means a smoother transition when migratiing to C8.

> Are we using different properties for backwards compatibility?

Not quite. In theory the way it works with the old Identity could remain the same and we could do all the mappings and stuff on the Zeebe side. Since the entire authorization concept is new withing Zeebe we don't have to worry about backwards compatibility at this level.

> how we should be referencing roles.

I'm not sure what you mean here.

remcowesterhoud: Things like this were discussed in the document I linked earlier.
If you think it could be improved I suggest you raise it in the #prj-identity-rearchitecture slack channel so the correct stakeholders can give input.

For now we shouldn't change this as the way it is currently matches with the design that's been created.

emulbreh: I proposed this in a [thread](https://camunda.slack.com/archives/C06UYJMMETZ/p1732184870062729) yesterday, but I guess that wasn't visible enough. Can you point me to the related discussion in the doc?

remcowesterhoud: I can't right now as I'm traveling, but it should be somewhere in the resolved comments

PR #24965, opened 2024-11-21
feat: apply authorization checks on roles

houssain-barouni:
## Description

Adding authorization checks for Roles search and get by key.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


service/src/main/java/io/camunda/service/RoleServices.java
```diff
@@ -62,16 +62,16 @@ public CompletableFuture<RoleRecord> updateRole(final long roleKey, final String
...
-        .items()
-        .stream()
-        .findFirst();
+    final var result =
+        roleSearchClient
+            .withSecurityContext(securityContextProvider.provideSecurityContext(authentication))
+            .searchRoles(roleSearchQuery(q -> q.filter(f -> f.roleKey(roleKey))));
+    final var roleEntity = getSingleResultOrThrow(result, roleKey, "Role");
+    final var authorization = Authorization.of(a -> a.role().read());
+    if (!securityContextProvider.isAuthorized(roleEntity.name(), authentication, authorization)) {
```
houssain-barouni, line 71: @Ben-Sheppard @emulbreh 
in the [doc](https://docs.google.com/document/d/1ikxw9ARCv-MBwASnj7WhW5uHCndCRk2B35xi-cB2N0M/edit?tab=t.0#heading=h.id62p5c0a9jc), it is mentionned that the authorization checks should be applied on the following role field
```
Resource Key:
Role Id operationsEngineer
Role Id admin
Role Id fullAccess
Role Id viewer
Role Id temporaryAccess
* (wildcard)
```
however, there is no `roleId` attribute for roles, is it actually the `name` ?

emulbreh: We're having a related discussion on https://github.com/camunda/camunda/pull/24962.
I probably don't have enough historical context, but my impression was that we'd ideally want to reference resources by key whenever possible. 

We shouldn't use the name as that's a user-editable label and we wouldn't want to update authorizations when it changes.

houssain-barouni: I see. I guess it should be an immutable id defined by the user on Object creation. Like we have for other objects like processDefinitionId ...

emulbreh: I believe `processDefinitionId` is the bpmn process id, not a synthetic identifier. It's also not a unique identifier as it's shared across multiple versions of the process. This is probably a special case. 
We do have a user-provided `tenantId` on tenant entities, which (as far as I understand) is primarily for backwards compatibility, so another special case.
Most other entities probably could (and IMO should) use their key as a resource identifier.

houssain-barouni: It is more convenient for an admin to grant authorizations based on predefined IDs. Otherwise, they would need to update all users authorizations whenever a new version of a process definition is deployed, as it generates a new key. From my understanding, this is the scope of the 8.7 release and will be compatible with the existing authorization system. I assume that implementing fine-grained permissions based on the generated keys will be done in future releases. 
@Ben-Sheppard please correct me if needed?

PR #24991, opened 2024-11-21
feat: include prefix support for the transformers

nathansandi:
## Description

For visibility
@emulbreh is working to unify the descriptors + tenant over this PR,  https://github.com/camunda/camunda/pull/25048/files#diff-7da04f0e010698982a4f835e1438845c9879f0acd372487c963a1135203b85a0 
Which will cover what was made on this PR 

I will leave the PR open for now (in case of any decision changes) but will close ASAP the @emulbreh  PR is merged. 

- Apply index prefix for all filter transformers
- Add function to build the complete filter name

The base implementation is merged located here:  
https://github.com/camunda/camunda/pull/24981 

## Related issues

related to https://github.com/camunda/camunda/issues/23808, superseded by https://github.com/camunda/camunda/pull/25048


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/DecisionDefinitionFilterTransformer.java
```diff
@@ -51,7 +58,8 @@ public SearchQuery toSearchQuery(final DecisionDefinitionFilter filter) {
 
   @Override
   public List<String> toIndices(final DecisionDefinitionFilter filter) {
-    return List.of("operate-decision-8.3.0_alias");
+    final String indexPrefix = DecisionIndex.getIndexNameWithPrefix(prefix);
```
emulbreh, line 61: Hi @nathansandi, Houssain just pointed me to this PR. I've actually addressed the same issue from a slightly different angle in https://github.com/camunda/camunda/pull/25048/commits/dfc35dc89f298f5effc108baf495317430ac3e5d. The main goal there is to return the IndexDescriptor from toIndices() directly, so that I have access to tenancy information on them. But it also means that I get the full index name here for free.
What do you think of this approach?

emulbreh: I notices that we sometimes return the alias and sometimes the index itself, but I believe this can be unified: we already create an alias for all indices and templates and so we can always query the alias.

I want to return the IndexDescriptor from the toIndices so I have access to configuration on them. This also gets rid of the hardcoded indices.

emulbreh: It's still work in progress, but I would look like this: https://github.com/camunda/camunda/pull/25048/files#diff-7da04f0e010698982a4f835e1438845c9879f0acd372487c963a1135203b85a0

nathansandi: Sounds good to me ;) 
I am going tag @ChrisKujawa for visibility, so as you are covering this on your PR.
Please, let me know if you need something from my side

I gonna leave this PR open for while, just in case any decision changes.

ChrisKujawa: @nathansandi sorry I had not time to look into this, do you still need my input?

emulbreh: @nathansandi Do you have time to review my PR? If the full set of changes is too much, I'm particular interested in feedback on the first commit, which passes index descriptors into filter transformers.

nathansandi: Hello @emulbreh  :) I have checked this morning. So far looks good to me.

PR #24993, opened 2024-11-21
feat: include form metadata in reconstructed deployment

lenaschoenburg:
This extends deployment reconstruction with support for Form metadata. I've followed the same pattern that we used for process metadata.

A new utility class is added to convert between the persisted resource and the metadata record.

depends on #24866
relates to  #24817

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessorTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessorTest.java
@@ -21,6 +21,7 @@ import io.camunda.zeebe.engine.util.stream.FakeProcessingResultBuilder;
 import io.camunda.zeebe.protocol.Protocol;
 import io.camunda.zeebe.protocol.impl.record.UnifiedRecordValue;
 import io.camunda.zeebe.protocol.impl.record.value.deployment.DeploymentRecord;
+import io.camunda.zeebe.protocol.impl.record.value.deployment.FormRecord;
 import io.camunda.zeebe.protocol.impl.record.value.deployment.ProcessRecord;
 import io.camunda.zeebe.protocol.record.intent.DeploymentIntent;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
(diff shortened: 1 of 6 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/deployment/FormStateTest.java
```diff
@@ -66,4 +75,52 @@ void shouldReturnEmptyIfNoFormIsDeployedForFormIdAndVersionTag() {
...
+    formState.storeFormInFormColumnFamily(form2);
+    formState.storeFormInFormColumnFamily(form3);
+
+    // when -- iterating through all forms
+    final var visitor = Mockito.mock(PersistedFormVisitor.class);
+    when(visitor.visit(any())).thenReturn(true);
+    formState.forEachForm(null, visitor);
+
+    // then -- visited all three forms
+    verify(visitor, times(3)).visit(any());
```
remcowesterhoud, line 96: 🔧 Same comments as in the processes PR. It'd be cleaner in these tests to actually verify the arguments using a captor to make sure you're not skipping the wrong one.

remcowesterhoud (commented):
I've reviewed it, but overall a lot of the comments would be the same as the other PR. Let's get this one merged and depending on the changes made there we can revisit this one.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessorTest.java
```diff
@@ -216,4 +217,126 @@ record -> {
...
+                  .satisfies(
+                      deploymentRecord -> {
+                        assertThat(deploymentRecord.getFormMetadata()).hasSize(1);
+                        assertThat(deploymentRecord.getDeploymentKey()).isEqualTo(deploymentKey);
+                      });
+            });
+  }
+
+  @Test
+  void shouldIncludeAllFormsOfDeployment() {
```
remcowesterhoud, line 296: ❓ Will you make a test in the end that combines different resources? So basically one that verifies if we have a process, that it can also find the forms and DRGs within the same deployment?

remcowesterhoud (approved):
Nice work, just some minor comments 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/DeploymentReconstructProcessorTest.java
lenaschoenburg: We should have a test for this at the end :+1:

PR #24997, opened 2024-11-21
Exporter ignores 8.6 records

### LINKED ISSUE #24907: Camunda exporter only exporter 8.7 records
**Description**

The exporter should only export 8.7 records.

Success Criteria:
- [ ] Export 8.7 records
- [ ] Do not export < 8.7.0 records
- [ ] 8.6 records are simply skipped and ack'ed

EuroLew:
## Description

Camunda exporter will not export 8.6 records but still update the last exported position like they did (I'm not totally sure if also updating the last exported position is correct someone can correct me).

## Related issues

closes #24907


ChrisKujawa (approved):
Thanks @EuroLew 🚀 

I think we might want to improve in the future how we can better unit tests the exporter, maybe with a separate abstraction layer so we can better inject dependencies.

PR #25038, opened 2024-11-22
Operate import metric

ChrisKujawa:
## Description

* Add new gauge to communicate that importing is done by partition and value type
     * This is useful for users to know when to stop the importer
     * This is useful for us to understand where importing stands, and which value type might not complete

## Checklist

- [ ] for CI changes:

## Related issues

related https://github.com/camunda/camunda/issues/23913

### DIFF
--- a/operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
+++ b/operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
@@ -11,6 +11,7 @@ import io.camunda.operate.Metrics;
 import io.camunda.operate.property.OperateProperties;
 import io.camunda.operate.store.ImportStore;
 import io.camunda.webapps.schema.entities.operate.ImportPositionEntity;
+import io.micrometer.core.instrument.Gauge;
 import jakarta.annotation.PostConstruct;
 import java.io.IOException;
 import java.time.OffsetDateTime;
@@ -42,6 +43,7 @@ public class ImportPositionHolder {
       new HashMap<>();
   private final Map<String, ImportPositionEntity> inflightImportPositions = new HashMap<>();
   private final Map<String, ImportPositionEntity> inflightPostImportPositions = new HashMap<>();
+  private final Map<String, Gauge> importPositionCompletedGauges = new HashMap<>();
 
   private ScheduledFuture<?> scheduledImportPositionUpdateTask;
   private final ReentrantLock inflightImportPositionLock = new ReentrantLock();
@@ -128,14 +130,16 @@ public class ImportPositionHolder {
           if (importPosition == null) {
             importPosition = lastProcessedPosition;
             LOGGER.error("SET import position {}; NOT EXIST BEFORE", importPosition);
-            metrics.registerGauge(
-                Metrics.GAUGE_NAME_IMPORT_POSITION_COMPLETED,
-                importPosition,
-                (pos) -> pos.getCompleted() ? 1.0 : 0.0,
-                Metrics.TAG_KEY_PARTITION,
-                Integer.toString(partition),
-                Metrics.TAG_KEY_IMPORT_POS_ALIAS,
-                aliasName);
+            importPositionCompletedGauges.put(
+                key,
+                metrics.registerGauge(
+                    Metrics.GAUGE_NAME_IMPORT_POSITION_COMPLETED,
+                    importPosition,
+                    (pos) -> pos.getCompleted() ? 1.0 : 0.0,
+                    Metrics.TAG_KEY_PARTITION,
+                    Integer.toString(partition),
+                    Metrics.TAG_KEY_IMPORT_POS_ALIAS,
+                    aliasName));
           } else {
             importPosition
                 .setPosition(lastProcessedPosition.getPosition())
@@ -145,7 +149,8 @@ public class ImportPositionHolder {
             LOGGER.error("SET import position {}; HAS EXIST BEFORE", importPosition);
           }
           inflightImportPositions.put(key, importPosition);
-        });
+        },
+        "record last loaded pos");
   }
 
   public void recordLatestPostImportedPosition(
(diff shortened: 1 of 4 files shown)


operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
```diff
@@ -172,6 +177,104 @@ final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 1);
...
+  public void shouldNotMarkedAsCompletedViaMetricsWhenImportingIsNotDone() throws IOException {
+    // given
+    final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 1);
+    final var partitionTwoRecord = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 2);
+    EXPORTER.export(record);
+    EXPORTER.export(partitionTwoRecord);
+    esClient.indices().refresh(new RefreshRequest("*"), RequestOptions.DEFAULT);
+
+    // when
+    for (int i = 0; i <= RecordsReaderHolder.MINIMUM_EMPTY_BATCHES_FOR_COMPLETED_READER; i++) {
```
EuroLew, line 190: An extra `performOneRoundOfImport` call is needed as the first import round will not result in an empty batch.

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
EuroLew, line 1: I would like an extra test where after one round of importing the gauges for all the record readers say they are not completed

EuroLew (approved):
Just an extra test otherwise LGTM

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
```diff
@@ -172,6 +177,104 @@ final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 1);
...
+    final Gauge partitionTwoImportStatus =
+        metrics
+            .get(Metrics.GAUGE_NAME_IMPORT_POSITION_COMPLETED)
+            .tags(
+                Metrics.TAG_KEY_PARTITION,
+                "2",
+                Metrics.TAG_KEY_IMPORT_POS_ALIAS,
+                "process-instance")
+            .gauge();
+    assertThat(partitionTwoImportStatus.value()).isEqualTo(0.0);
```
ChrisKujawa, line 220: @EuroLew you mean this test or?

EuroLew: For the communication of importer being finished (Assuming each record reader has an entry in the import postion index from processing 8.6 records which it should have? I don't think I can guarantee this assumption for prod though) A specific record reader does not need messages for it to be marked as completed e.g.

1. process instance record reader receives an 8.7 record marks partition 1 as completed
2. job record reader for partition 1 never receives a record but because it knows the partition is complete and it gets 5 empty batches we can mark it as complete

however for these metrics solutions we will only mark a record reader as 1.0 or 0.0 IF we receive a record for that record reader, given the PR description says this is intended to know where importing stands and whether it's done, this could be a problem as there could be uncompleted record readers that do not show up in metrics as 0.0 (if the corresponding record reader does not receive a record).

Or maybe this is not a big deal, if we have it configured somewhere or stored somewhere the amount of record readers we have we can compare how many 1.0 record readers we have

ChrisKujawa: >however for these metrics, we will only mark a record reader as 1.0 or 0.0 IF we receive a record for that record reader, given the PR description says this is intended to know where importing stands and whether it's done, this could be a problem as there could be uncompleted record readers that do not show up in metrics as 0.0 (if the corresponding record reader has never received a record).

That is not true 🤔 It will register the first time it sets the import position the gauge https://github.com/camunda/camunda/pull/25038/commits/ba3d5edb021505abfa0aca2bc743d9820bcf2d77#diff-c2f26dd940eb71359693407274f4e401317be38aab4d7e51162e7ecafa775b27R132 

Here it keeps the reference of the import position entity. This can either be when we import more records or when we complete with empty batches, because we take the same path on marking as done.

EuroLew: resolved

ChrisKujawa: I synced with @EuroLew we concluded that this should work, as the gauge is keeping track of the object reference and on marking reader complete we also run this code part

ChrisKujawa:
https://docs.micrometer.io/micrometer/reference/concepts/gauges.html#:~:text=If%20you%20see%20your%20gauge%20reporting%20for%20a%20few%20minutes%20and%20then%20disappearing%20or%20reporting%20NaN%2C%20it%20almost%20certainly%20suggests%20that%20the%20underlying%20object%20being%20gauged%20has%20been%20garbage%20collected.

operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
```diff
@@ -124,15 +129,28 @@ public void recordLatestLoadedPosition(final ImportPositionEntity lastProcessedP
...
+                    Integer.toString(partition),
+                    Metrics.TAG_KEY_IMPORT_POS_ALIAS,
+                    aliasName));
           } else {
             importPosition
                 .setPosition(lastProcessedPosition.getPosition())
                 .setSequence(lastProcessedPosition.getSequence())
                 .setIndexName(lastProcessedPosition.getIndexName())
                 .setCompleted(lastProcessedPosition.getCompleted());
+            LOGGER.error("SET import position {}; HAS EXIST BEFORE", importPosition);
```
ChrisKujawa, line 149: @EuroLew We had some debugging logs added because of the failed tests 😅 They are now on main. Can you remove them again please?

PR #25050, opened 2024-11-22
feat: support patch permissions of an authorizations

### LINKED ISSUE #20451: Camunda service supports updating (add/remove) permissions of an authorizations
### Description
Camunda service allows to update permissions of an authorization assigned to an entity. 

### Acceptance Criteria
- [x] #24365
- [ ] add UI for user's authorization permission update
- [ ] add UI for group's authorization permission update
- [ ] add UI for role's authorization permission update
- [x] #24419

maryarm:
## Description

Camunda service allows to update permissions of an authorization assigned to an entity.
As part of this PR, we also changed the way the authorizations' list is displayed and switched to an expandable data table to reflect the API response.

## Related issues

closes #20451

### DIFF
--- a/identity/client/src/pages/users/detail/authorization/List.tsx
+++ b/identity/client/src/pages/users/detail/authorization/List.tsx
@@ -47,6 +47,7 @@ import { useEntityModal } from "src/components/modal";
 import { Authorization } from "src/utility/api/authorizations";
 import { DataTableRenderProps } from "@carbon/react/lib/components/DataTable/DataTable";
 import styled from "styled-components";
+import { gray30, spacing02, spacing04, spacing05 } from "@carbon/elements";
 
 type AuthorizationsListProps = {
   user: User;
@@ -59,18 +60,18 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
   const ToolbarContent = styled.div`
     display: flex;
     justify-content: flex-end;
-    gap: 16px; /* Replace with the appropriate Carbon spacing token */
+    gap: ${spacing05};
   `;
 
   const StyledListItem = styled(ListItem)`
     display: flex;
     justify-content: space-between;
     align-items: center;
-    padding-left: 10px;
+    padding-left: ${spacing04};
   `;
   const StyledDivider = styled(Section)`
-    margin: 5px 0;
-    border-top: 1px solid #ccc;
+    margin: ${spacing02} 0;
+    border-top: 1px solid ${gray30};
   `;
 
   const {
--- a/identity/client/src/pages/users/detail/authorization/List.tsx
+++ b/identity/client/src/pages/users/detail/authorization/List.tsx
@@ -253,27 +253,6 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
                                 )}
                               </ul>
                             )}
-                            {/* Add button if not all permissions are present */}
-                            {/*{checkMissingPermissions(dataMap[row.id].permissions).length > 0 && (
-                                                    <div style={{ marginTop: '16px', textAlign: 'right' }}>
-                                                      <button
-                                                          type="button"
-                                                          onClick={() => onAddPermission(row)}
-                                                          style={{
-                                                            border: 'none',
-                                                            background: '#0f62fe',
-                                                            color: 'white',
-                                                            cursor: 'pointer',
-                                                            padding: '8px 16px',
-                                                            borderRadius: '4px',
-                                                          }}
-                                                          aria-label="Add permission"
-                                                      >
-                                                        <Add style={{ marginRight: '8px' }} />
-                                                        Add Permission
-                                                      </button>
-                                                    </div>
-                                                )}*/}
                           </TableExpandedRow>
                         )}
                       </React.Fragment>
--- a/identity/client/src/pages/users/detail/authorization/List.tsx
+++ b/identity/client/src/pages/users/detail/authorization/List.tsx
@@ -40,7 +40,7 @@ import {
 } from "@carbon/react";
 
 import { useEntityModal } from "src/components/modal";
-import { Authorization, Permission } from "src/utility/api/authorizations";
+import { Authorization } from "src/utility/api/authorizations";
 import { DataTableRenderProps } from "@carbon/react/lib/components/DataTable/DataTable";
 
 type AuthorizationsListProps = {
--- a/identity/client/src/pages/users/detail/authorization/List.tsx
+++ b/identity/client/src/pages/users/detail/authorization/List.tsx
@@ -92,7 +92,7 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
 
   const rows = rowsWithoutId.map((item, index) => ({
     ...item,
-    id: index,
+    id: String(index),
   }));
 
   // Create a lookup map for rows by ID
@@ -173,7 +173,10 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
                   <TableRow>
                     <TableExpandHeader />
                     {headers.map((header) => (
-                      <TableHeader {...getHeaderProps({ header })}>
+                      <TableHeader
+                        {...getHeaderProps({ header })}
+                        key={header.key}
+                      >
                         {header.header}
                       </TableHeader>
                     ))}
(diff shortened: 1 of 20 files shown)


identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -34,8 +63,49 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
-    console.log(authorization);
+  const headers = [{ header: t("ResourceType"), key: "resourceType" }];
+
+  const rows = authorizations == null ? [] : authorizations.items;
+  // Create a lookup map for rows by ID
+  const dataMap = rows.reduce(
+    (map, item) => {
+      if (typeof item.key === "string") {
+        map[item.key] = item;
+      }
```
vsgoulart, line 74: I added this if but this breaks the behavior, which was already broken.
There's no `key` property coming in the payload so the table is not properly rendered and it always render the last item in the list

https://github.com/user-attachments/assets/efd65925-cbac-462a-a888-82748d863e06

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -47,23 +112,150 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
+          }: DataTableRenderProps<Authorization, unknown[]>) => (
+            <div>
+              {/* Table Toolbar with Add Row Button */}
+              <TableToolbar>
+                <TableToolbarContent
+                  style={{
+                    display: "flex",
+                    justifyContent: "flex-end",
+                    gap: "16px",
+                  }}
```
vsgoulart, line 135: Please don't inline CSS. Identity uses `styled-components` so we should use that for styling.
Also the gap should use the Carbon [spacing tokens](https://carbondesignsystem.com/elements/spacing/overview/)

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -47,23 +112,150 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
+                      cursor: "pointer",
+                      padding: "8px 16px",
+                      borderRadius: "4px",
+                      display: "flex",
+                      alignItems: "center",
+                    }}
+                  >
+                    <Add style={{ marginRight: "8px" }} />
+                    Add Authorizations
+                  </button>
```
vsgoulart, line 160: Please don't use a custom button, that's why we use the Carbon design system. [We already have a button built-in](https://react.carbondesignsystem.com/?path=/docs/components-button--overview#button-rendericon)

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -47,23 +112,150 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
+                                  (permission, index) => (
+                                    <React.Fragment key={index}>
+                                      <li
+                                        style={{
+                                          display: "flex",
+                                          justifyContent: "space-between",
+                                          alignItems: "center",
+                                          padding: "10px 0",
+                                        }}
+                                      >
```
vsgoulart, line 206: [Use should use Carbon list component](https://react.carbondesignsystem.com/?path=/docs/components-unorderedlist--overview)

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -47,23 +112,150 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
+                                        <div>
+                                          <strong>Permission Type:</strong>{" "}
+                                          {permission.permissionType}
+                                          <br />
+                                          <strong>Resource IDs:</strong>{" "}
+                                          {permission.resourceIds.join(", ")}
```
vsgoulart, line 212: Only style with CSS. [Carbon has specific typesets for such cases](https://carbondesignsystem.com/elements/typography/type-sets/)

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -47,23 +112,150 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
+                                          }
+                                          style={{
+                                            border: "none",
+                                            background: "none",
+                                            cursor: "pointer",
+                                          }}
+                                          aria-label="Delete permission"
+                                        >
+                                          <Edit />
+                                        </button>
```
vsgoulart, line 228: We should use the Carbon button component here too

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -47,23 +112,150 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
+                                      {index <
+                                        dataMap[row.id].permissions.length -
+                                          1 && (
+                                        <hr
+                                          style={{
+                                            border: "none",
+                                            borderTop: "1px solid #ccc",
+                                            margin: "5px 0",
+                                          }}
+                                        />
```
vsgoulart, line 240: I believe this can be done with CSS.
We also should use the Carbon spacing variables and color variables

vsgoulart (changes requested):

maryarm:
> @maryarm I pushed a fix for some TS errors that broke the build I'll do the full review now

Hi @vsgoulart , thanks for the review. I tried to address most of your comments, please take a look.

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -47,23 +138,115 @@ const List: FC<AuthorizationsListProps> = ({ user, loadingUser }) => {
...
+                                            {permission.permissionType}
+                                            <br />
+                                            <strong>Resource IDs:</strong>{" "}
+                                            {permission.resourceIds.join(", ")}
+                                          </div>
+                                          <Button
+                                            kind="ghost"
+                                            size="small"
+                                            iconDescription={t(
+                                              "Edit permission",
```
vsgoulart, line 220: Don't we have to create translations keys for these strings?

identity/client/src/pages/users/detail/authorization/List.tsx
```diff
@@ -22,6 +56,23 @@ type AuthorizationsListProps = {
...
+  const StyledListItem = styled(ListItem)`
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    padding-left: 10px;
+  `;
+  const StyledDivider = styled(Section)`
+    margin: 5px 0;
+    border-top: 1px solid #ccc;
+  `;
```
vsgoulart, line 74: Please use the Carbon variables for spacing and colors like in [here](https://github.com/camunda/camunda/blob/6177a514b5774aec1263a4dfde4e2446fee14f66/identity/client/src/components/documentation/index.tsx#L9)

vsgoulart (commented):
There are errors in the console, we have to fix those before merging this:
image
image
image
image
image

identity/client/src/pages/users/detail/authorization/List.tsx
vitorwtavares: I checked the codebase and it seems there are a lot of uses of that translate function but no keys were ever set. I also think they have to be manually defined, from what I could find there's only a handful of keys defined in some json files in this folder `src/utility/localization/en`.

I believe this is not worth the effort right now, so I vote we leave this as is for now until we can discuss if we need translation and when do we need it by. Even if it was worth to do it, it should be a separate PR fixing all instances of this. What do you think @vsgoulart?

vitorwtavares:
@vsgoulart Context for your review:
-  I fixed all the things you had pointed out here aside from the translation one like pointed out in the other comment.
-  Also did some light refactoring to improve the project's structure and facilitate the usage of the correct Carbon variables (added `@carbon/elements` as well as 2 type packages to use the autocomplete for style imports).

They're separated in 2 different commits if you'd like to look at them separately.

identity/client/src/pages/users/detail/authorization/PatchModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+      <SelectedResources>
+        {selectedResources.map((r) => (
+          <Tag
+            key={r}
+            onClose={onUnselectResource(r)}
+            size="md"
+            type="blue"
+            filter
+          >
+            {r}
```
vsgoulart, line 250:
```suggestion
        {selectedResources.map((resource) => (
          <Tag
            key={resource}
            onClose={onUnselectResource(r)}
            size="md"
            type="blue"
            filter
          >
            {resource}
```

vsgoulart (approved):
Just one minor comment that I think we should fix before merging but the rest looks good to me

Issue #24869, opened 2024-11-19
Combine operate `BackupServiceImpl` and tasklists `BackupManagerElasticSearch` in a single implementation in `webapp-backup`


entangled90:
After the preparation work done in #24842  & #24843, the next step is to combine the two implementation and move it to `webapp-backup`.

So far I haven't spotted any major differences between the two implementation, except for [this line in tasklists' version](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/es/backup/es/BackupManagerElasticSearch.java).
This line was added to fix [this issue](https://github.com/camunda/tasklist/issues/4704).

PR #25053, opened 2024-11-22
refactor: Move BackupServiceImpl from operate to webapps-backup

entangled90:
## Description
Moved `BackupServiceImpl` into `webapps-backup`: 
- removed `Prio1,Prio2,Prio3,Prio4` from operate and migrated the classes to use the ones from `webapps-schema`
- Created new `@Component` to instantiate the service in webapps-backup (which does not use Spring): this code will be migrated to `dist` as well when the refactoring is complete. 

## Related issues
relates #24869


webapps-backup/src/main/java/io/camunda/webapps/backup/BackupServiceImpl.java
```diff
@@ -49,38 +38,39 @@ public class BackupServiceImpl implements BackupService {
...
+    this.repository = repository;
+    backupProps = operateProperties;
   }
 
   @Override
   public void deleteBackup(final Long backupId) {
     repository.validateRepositoryExists(getRepositoryName());
     final String repositoryName = getRepositoryName();
     final int count = getIndexPatternsOrdered().length;
-    final String version = getCurrentOperateVersion();
```
entangled90, line 80: will this version cause problems when we merge them?

npepinpe: Good point. And technically yes - I didn't think we needed to fix the issue with the version, but possibly we might, as it would also affect Tasklist :thinking:

npepinpe: Let's bring it up in the tech channel, and tag Marcos and Svetlana/Thorben

entangled90: I'm working on the merge now and I noticed that tasklist is doing the same actually in its implementations.

entangled90:
Looks like the SpringContext is not being constructed correctly, I'll check the ITs locally :+1:

operate/webapp/src/main/java/io/camunda/operate/webapp/backup/BackupConfig.java
```diff
@@ -24,4 +27,35 @@ public ThreadPoolTaskExecutor getTaskExecutor() {
...
+      @Override
+      public int snapshotTimeout() {
+        return operateProperties.getSnapshotTimeout();
+      }
+
+      @Override
+      public Long incompleteCheckTimeoutInSeconds() {
+        return operateProperties.getIncompleteCheckTimeoutInSeconds();
+      }
+    };
```
npepinpe, line 59: :question: Why an anonymous class? There's a `BackupRepositoryPropsImpl` class, no?

npepinpe (approved):
:rocket: 

- :question: I assume since you deleted the `Prio1Backup` (and others) and it builds, that there was no usages outside the backup service? Meaning the template descriptors are already implementing the new `Prio1Backup` and other interfaces, right?

entangled90:
Yes all the classes that were implementing `Prio{1,2,3,4}` in operate have been migrated to the interfaces already defined in `webapp-schema` module.

operate/webapp/src/main/java/io/camunda/operate/webapp/backup/BackupConfig.java
entangled90: Unfortunately `BackupRepositoryPropsImpl` take a reference of the properties when the class is instantiated, which means that when testing the mocks are setup after the class was already instantiated. 
This implementation delegates to the properties, so it work even mocking the properties afterwards.
I will probably remove the record if it's not in use anymore in the next PRs

PR #25064, opened 2024-11-22
feat: return name for flownode-instances & user-tasks

### LINKED ISSUE #24893: [C8 API] Return `flowNodeName` in FlownodeInstance & UserTask APIs
## Description

For feature parity with V1 we need to provide flow node names in the APIs below. The REST API can fetch XML and read this data for each items.

## Acceptance criteria

- The REST API will take care of resolving such attributes by itself.
- Flownode instances V2 `search` and `get` APIs return `flowNodeName` for entities.
  - https://docs.camunda.io/docs/next/apis-tools/operate-api/specifications/search-4/
- User tasks V2 `search` and `get` APIs return `name` (`flowNodeName`) for entities.
  - https://docs.camunda.io/docs/next/apis-tools/tasklist-api-rest/specifications/search-tasks/
- Tests.
- No caching is required yet.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/25064
```

danielkelemen:
## Description

- Adds `XmlUtil` in gateway-rest to hold methods to parse xml attributes
- Adds `name` field for user-tasks search & get APIs
- Adds `flowNodeName` field for flownode-instances search & get APIs
- Adds java client properties
- Tests for QA & gateway-rest

## Related issues

closes https://github.com/camunda/camunda/issues/24893

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
@@ -91,6 +91,16 @@ class UserTaskQueryTest {
   }
 
   @Test
+  public void shouldUseUserTaskElementIdIfNameNotSet() {
+    // when
+    final var result =
+        camundaClient.newUserTaskQuery().filter(f -> f.elementId("test-2")).send().join();
+    // then
+    assertThat(result.items()).hasSize(1);
+    assertThat(result.items().getFirst().getName()).isEqualTo("test-2");
+  }
+
+  @Test
   public void shouldRetrieveVariablesFromUserTask() {
     final UserTaskVariableFilterRequest variableValueFilter =
         new UserTaskVariableFilterRequest().name("task02").value("1");
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
@@ -81,6 +81,16 @@ class UserTaskQueryTest {
   }
 
   @Test
+  public void shouldHaveCorrectUserTaskName() {
+    // when
+    final var result =
+        camundaClient.newUserTaskQuery().filter(f -> f.elementId("form_process")).send().join();
+    // then
+    assertThat(result.items()).hasSize(1);
+    assertThat(result.items().getFirst().getName()).isEqualTo("Form");
+  }
+
+  @Test
   public void shouldRetrieveVariablesFromUserTask() {
     final UserTaskVariableFilterRequest variableValueFilter =
         new UserTaskVariableFilterRequest().name("task02").value("1");
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
@@ -70,7 +70,6 @@ import io.camunda.zeebe.gateway.protocol.rest.VariableItem;
 import io.camunda.zeebe.gateway.protocol.rest.VariableSearchQueryResponse;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -145,7 +144,7 @@ public final class SearchQueryResponseMapper {
 
   public static FlowNodeInstanceSearchQueryResponse toFlowNodeInstanceSearchQueryResponse(
       final SearchQueryResult<FlowNodeInstanceEntity> result,
-      final Map<Long, Map<Long, String>> nameMap) {
+      final Map<Long, Map<String, String>> nameMap) {
     final var page = toSearchQueryPageResponse(result);
     return new FlowNodeInstanceSearchQueryResponse()
         .page(page)
@@ -168,7 +167,7 @@ public final class SearchQueryResponseMapper {
 
   public static UserTaskSearchQueryResponse toUserTaskSearchQueryResponse(
       final SearchQueryResult<UserTaskEntity> result,
-      final HashMap<Long, Map<Long, String>> nameMap) {
+      final Map<Long, Map<String, String>> nameMap) {
     final var page = toSearchQueryPageResponse(result);
     return new UserTaskSearchQueryResponse()
         .page(page)
@@ -292,15 +291,13 @@ public final class SearchQueryResponseMapper {
   }
 
   private static List<FlowNodeInstanceItem> toFlowNodeInstance(
-      final List<FlowNodeInstanceEntity> instances, final Map<Long, Map<Long, String>> nameMap) {
+      final List<FlowNodeInstanceEntity> instances, final Map<Long, Map<String, String>> nameMap) {
     return instances.stream()
         .map(
             instance ->
                 toFlowNodeInstance(
                     instance,
-                    nameMap
-                        .get(instance.processDefinitionKey())
-                        .get(instance.flowNodeInstanceKey())))
+                    nameMap.get(instance.processDefinitionKey()).get(instance.flowNodeId())))
         .toList();
   }
 
@@ -346,11 +343,11 @@ public final class SearchQueryResponseMapper {
   }
 
   private static List<UserTaskItem> toUserTasks(
-      final List<UserTaskEntity> tasks, final HashMap<Long, Map<Long, String>> nameMap) {
+      final List<UserTaskEntity> tasks, final Map<Long, Map<String, String>> nameMap) {
     return tasks.stream()
         .map(
             (UserTaskEntity t) ->
-                toUserTask(t, nameMap.get(t.processDefinitionKey()).get(t.userTaskKey())))
+                toUserTask(t, nameMap.get(t.processDefinitionKey()).get(t.elementId())))
         .toList();
   }
 
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
@@ -144,7 +144,8 @@ public final class SearchQueryResponseMapper {
   }
 
   public static FlowNodeInstanceSearchQueryResponse toFlowNodeInstanceSearchQueryResponse(
-      final SearchQueryResult<FlowNodeInstanceEntity> result, final HashMap<Long, String> nameMap) {
+      final SearchQueryResult<FlowNodeInstanceEntity> result,
+      final Map<Long, Map<Long, String>> nameMap) {
     final var page = toSearchQueryPageResponse(result);
     return new FlowNodeInstanceSearchQueryResponse()
         .page(page)
@@ -166,7 +167,8 @@ public final class SearchQueryResponseMapper {
   }
 
   public static UserTaskSearchQueryResponse toUserTaskSearchQueryResponse(
-      final SearchQueryResult<UserTaskEntity> result, final Map<Long, String> nameMap) {
+      final SearchQueryResult<UserTaskEntity> result,
+      final HashMap<Long, Map<Long, String>> nameMap) {
     final var page = toSearchQueryPageResponse(result);
     return new UserTaskSearchQueryResponse()
         .page(page)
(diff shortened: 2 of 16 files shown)


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -0,0 +1,87 @@
...
+        userTask.elementId());
+  }
+
+  public String extractFlowNodeName(
+      final String xml,
+      final String processDefinitionId,
+      final String flowNodeType,
+      final String flowNodeId) {
+    try {
+      final var documentBuilder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
```
danielkelemen, line 70: 💭 Would be nicer to extract these factories. However, these classes don't seem to be thread-safe according to spec: [DocumentBuilderFactory](https://docs.oracle.com/cd/E17802_01/webservices/webservices/docs/1.5/api/javax/xml/parsers/DocumentBuilderFactory.html#:~:text=provided%20the%20instance%20of%20the%20factory%20isn%27t%20being%20used%20in%20more%20than%20one%20thread%20at%20a%20time), [DocumentBuilder](https://docs.oracle.com/cd/E17802_01/webservices/webservices/docs/1.5/api/javax/xml/parsers/DocumentBuilder.html#:~:text=is%20NOT%20guaranteed%20to%20behave%20as%20per%20the%20specification%20if%20it%20is%20used%20concurrently%20by%20two%20or%20more%20threads).


Define them as `ThreadLocal` to make it faster?

houssain-barouni (changes requested):
I suggested small improvements 🔧 
I think caching could save us CPU and network bandwidth in this implementation. But as always caching is not easy to manage

danielkelemen:
> I think caching could save us CPU and network bandwidth in this implementation. But as always caching is not easy to manage

@houssain-barouni: Yes, flow nodes will be cached in the next increment.

qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
```diff
@@ -80,6 +80,16 @@ public void shouldRetrieveTaskByTaskVariable() {
     assertThat(result.items().size()).isEqualTo(1);
   }
 
+  @Test
+  public void shouldHaveCorrectUserTaskName() {
```
tmetzke, line 84: 🔧 We should also test that the activity id is used if no activity name exists.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/FlowNodeInstanceQueryController.java
```diff
@@ -52,12 +58,12 @@ public ResponseEntity<FlowNodeInstanceSearchQueryResponse> searchFlownodeInstanc
...
-              SearchQueryResponseMapper.toFlowNodeInstance(
-                  flownodeInstanceServices
-                      .withAuthentication(RequestMapper.getAuthentication())
-                      .getByKey(flowNodeInstanceKey)));
+      final FlowNodeInstanceEntity flowNode =
+          flownodeInstanceServices
+              .withAuthentication(RequestMapper.getAuthentication())
+              .getByKey(flowNodeInstanceKey);
+      final String name = xmlUtil.getFlowNodeName(flowNode);
+      return ResponseEntity.ok().body(SearchQueryResponseMapper.toFlowNodeInstance(flowNode, name));
```
tmetzke, line 66: 💭 We could move the name resolution into the service. That way, the entity the service layer returns contains the name, and the REST controller consumes this DTO. That makes the REST controller code a lot simpler (it mostly stays the same). The service layer is responsible for fetching the data correctly; the REST controllers shouldn't need to care. Any other consumer of the service layer would receive a "complete" entity result as well this way. Plus, this should be the task of the search layer, ultimately. For example, the RDBMS mapper might decide to deliver those values directly, no need for fetching it manually. This would be hidden in the search transformers then.

tmetzke (commented):
👍 Looks really good already 🚀 
🔧 I added another suggestion that decouples the REST layer from the name resolution.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/FlowNodeInstanceQueryController.java
danielkelemen: This would make more sense. I'll give this a shot.

> For example, the RDBMS mapper might decide to deliver those values directly

I guess the service layer could check if the names are set already and then do nothing.

tmetzke: The service layer shouldn't need to care either. The search layer must take care of delivering the right entity.

danielkelemen: But if the name resolution happens in the service then how can RDBMS decide to deliver them directly? Should we define a search client method for resolving the name?

tmetzke: Ah, sorry, the name resolution should happen in the search layer, not the service layer 🙈 I started off with the idea of doing it in the service layer, then realized it should actually go into search 😄

danielkelemen: Yeah, that could work. I could define a search layer method with default `XmlUtil` implementation for now.

danielkelemen: But wait, we want to cache this in the next step and that's supposed to be in the REST API.
That can only work if the REST API calls the service for name resolution.

1. REST API get entity (service -> search)
2. REST API fetch & cache names (service -> search)

tmetzke: And the search layer can't cache this? We can also sync on this, might be easier.

danielkelemen: There will be a follow-up ticket for moving the cache and name resolving into the search clients.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java
```diff
@@ -332,9 +340,11 @@ public static DecisionRequirementsItem toDecisionRequirements(
   }
 
   private static List<UserTaskItem> toUserTasks(
-      final List<UserTaskEntity> tasks, final Map<Long, String> nameMap) {
+      final List<UserTaskEntity> tasks, final HashMap<Long, Map<Long, String>> nameMap) {
```
houssain-barouni, line 343: 🔧 `Map<Long, Map<Long, String>>`

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -37,9 +43,23 @@ public XmlUtil(final ProcessDefinitionServices processDefinitionServices) {
...
+        this::getFlowNodeName);
+  }
+
   public String getFlowNodeName(final FlowNodeInstanceEntity flowNode) {
     final var processDefinition =
         processDefinitionServices.getByKey(flowNode.processDefinitionKey());
+    return getFlowNodeName(processDefinition, flowNode);
+  }
+
+  protected String getFlowNodeName(
```
houssain-barouni, line 61: 🔧 `private`

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -51,16 +71,66 @@ public String getFlowNodeName(final FlowNodeInstanceEntity flowNode) {
...
+        this::getUserTaskName);
+  }
+
   public String getUserTaskName(final UserTaskEntity userTask) {
     final var processDefinition =
         processDefinitionServices.getByKey(userTask.processDefinitionKey());
+    return getUserTaskName(processDefinition, userTask);
+  }
+
+  protected String getUserTaskName(
```
houssain-barouni, line 88: 🔧 `private`

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -51,16 +71,66 @@ public String getFlowNodeName(final FlowNodeInstanceEntity flowNode) {
...
+      final ProcessDefinitionEntity processDefinition, final UserTaskEntity userTask) {
     return extractFlowNodeName(
         processDefinition.bpmnXml(),
         processDefinition.processDefinitionId(),
         "userTask",
         userTask.elementId());
   }
 
+  protected <T> HashMap<Long, Map<Long, String>> getNamesForEntities(
```
houssain-barouni, line 97: 🔧 `private <T> Map<Long, Map<Long, String>>`

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -51,16 +71,66 @@ public String getFlowNodeName(final FlowNodeInstanceEntity flowNode) {
         flowNode.flowNodeId());
   }
 
+  public HashMap<Long, Map<Long, String>> getUserTasksNames(final List<UserTaskEntity> userTasks) {
+    return getNamesForEntities(
+        userTasks,
+        UserTaskEntity::processDefinitionKey,
+        UserTaskEntity::userTaskKey,
```
houssain-barouni, line 78: I think it is better that you pass the task elementId as all tasks sharing the same elementId+processDefinitionKey have the same name.
Then you return `Map<Long, Map<String, String>>` where the key of the second map is elementId
Same thing for flowNodeInstances
This way you parse the xml once per elementId of flowNodeId

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -51,16 +71,66 @@ public String getFlowNodeName(final FlowNodeInstanceEntity flowNode) {
         flowNode.flowNodeId());
   }
 
+  public HashMap<Long, Map<Long, String>> getUserTasksNames(final List<UserTaskEntity> userTasks) {
```
houssain-barouni, line 74: 🔧 `Map<Long, Map<Long, String>>`

houssain-barouni (commented):
Thank you for the changes 👍 
Still some minor enhancements 🔧

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
danielkelemen: I think this one is used in tests. I could change it to package private.

houssain-barouni: It is better to only unit test public methods, and to not change the visibility of private methods for that. If you still need to do so, keep it with default modifier (package visibility) and annotate it with @VisibleForTesting

danielkelemen: Changed it to a public method.

houssain-barouni: you mean it is changed to private

houssain-barouni (approved):
LGTM! 👍

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
danielkelemen: Yes, the method is now private and the test uses a public method.

Issue #24929, opened 2024-11-20
Add blocking implementation of incident update task logic


npepinpe:
# Description

As a first step, we will port over the business logic of processing pending incident updates to the new exporter as a background task.

To avoid duplication, we will adopt the same repository pattern as the archiver, and push _all_ document store accesses to the repositories, leaving us with a single `IncidentUpdateTask` which is document store agnostic, and can be better unit tested.

> [!Note]
> Since its logic is quite simple, and the task is made of multiple steps, it's quite difficult still to unit test it - but it should be easier if we can stub the repositories at least.
>
> Still, if we find that it's quite hard in the first pass, we can wait for the second iteration to make it more testable.

The result should be a single task which can generate a bulk update of the right incidents, and follows the same logic as the current post import actions, without any of the document store specific parts. These will be added in a follow up issue.

PR #25067, opened 2024-11-22
Only set routing on list view updates and use the FNI's parent PI ID for it

npepinpe:
## Description

This PR improves the test coverage for the `IncidentUpdateTask`, which surfaced two bugs, which are both fixed here as well:

- The routing property of the document should be set _only_ for the list view indices updates.
- The routing property of an FNI document in the list view should be its parent process instance key, and not just the incident's process instance key (which could be of a child process instance not associated with the FNI). We use the tree path to find the right routing for the FNI.

[See this Slack thread for more](https://camunda.slack.com/archives/C06F0GLJNFM/p1732281258373749)

## Related issues

related to #24929

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/IncidentUpdateTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/IncidentUpdateTask.java
@@ -219,7 +219,6 @@ public final class IncidentUpdateTask implements BackgroundTask {
       final var processInstanceKey = incident.incident().getProcessInstanceKey();
       final var treePath = data.incidentTreePaths().get(incident.id());
       final var newState = batch.newIncidentStates().get(incident.incident().getKey());
-      final var routing = String.valueOf(processInstanceKey);
 
       if (!data.processInstanceTreePaths().containsKey(processInstanceKey)) {
         if (!ignoreMissingData) {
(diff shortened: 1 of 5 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateTaskTest.java
```diff
@@ -162,30 +164,53 @@ public CompletionStage<List<ActiveIncident>> getActiveIncidentsByTreePaths(
...
+   * <p>An instance of that parent process instance (1) has a call activity instance flow node (2)
+   * with a child process instance (3), with a task flow node (4).
+   *
+   * <p>There is a single incident update by default (5) which is being created.
    *
    * <p>The pending update has a position of 10.
    */
   @Nested
-  final class SingleIncidentTest {
+  final class IncidentTreePathTest {
```
npepinpe, line 178: :thought_balloon: I realized that it's just better coverage to test with a tree path, as it covers the same cases as the single task, and it's not really more expensive.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/IncidentUpdateTask.java
```diff
@@ -313,27 +304,31 @@ private void createFlowNodeInstanceUpdates(
...
     final boolean changedState;
     if (hasIncident) {
-      changedState = data.addFniIdsWithIncidentIds(fniId, incidentId);
+      changedState = data.addFniIdsWithIncidentIds(fniId, incident.id());
     } else {
-      changedState = data.removeIncidentIdByFniId(fniId, incidentId);
+      changedState = data.removeIncidentIdByFniId(fniId, incident.id());
     }
 
+    final var treePath = new TreePath(incident.incident().getTreePath());
```
entangled90, line 321: :question: why making a copy of the TreePath if only need to extract the process instance id? Is it because it is mutable?

entangled90 (approved):

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/IncidentUpdateTask.java
npepinpe: It's not a copy, it's wrapping the String to make it a `TreePath` type. Or do you mean why is `TreePath` copying it to `StringBuffer`? I guess because it's mutable, yeah.

entangled90: Ah sorry I thought it was a copy-constructor, but `getTreePath` is just a string. :+1:

entangled90 (changes requested):
We should add some tests for the TreePath :-D and remove binary search

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateTaskTest.java
entangled90: Unit tests done now :+1:

entangled90 (approved):

Issue #24930, opened 2024-11-20
Add ES incident update repository implementation


npepinpe:
# Description

Implement all the methods in `IncidentUpdateRepository` for Elasticsearch. This should include integration tests of each method against an actual ES instance itself.

I would recommend splitting this into multiple PR, one per method. Inherit from the `NoopIncidentUpdateRepository` initially, and then go from there. This can be easily parallelized among multiple people.

PR #25144, opened 2024-11-25
Implement getPendingIncidentsBatch for Elasticsearch

npepinpe:
## Description

This PR adds the implementation of `getPendingIncidentsBatch` for Elasticsearch.

## Related issues

related to #24930

PR #25146, opened 2024-11-25
Implement getIncidentDocuments for Elasticsearch

npepinpe:
## Description

This PR implements `getIncidentDocuments` for the `ElasticsearchIncidentUpdateRepository`.

## Related issues

related to #24930

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
@@ -84,6 +84,7 @@ public final class ElasticsearchIncidentUpdateRepository extends NoopIncidentUpd
         .query(q -> q.bool(b -> b.must(idQ, partitionQ)))
         .allowNoIndices(true)
         .ignoreUnavailable(true)
+        .sort(s -> s.field(f -> f.field(IncidentTemplate.KEY)))
         .build();
   }
 
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
@@ -16,10 +16,13 @@ import co.elastic.clients.elasticsearch.core.SearchResponse;
 import co.elastic.clients.elasticsearch.core.search.SourceFilter;
 import co.elastic.clients.json.JsonData;
 import io.camunda.exporter.tasks.incident.IncidentUpdateRepository.NoopIncidentUpdateRepository;
+import io.camunda.webapps.schema.descriptors.operate.template.IncidentTemplate;
 import io.camunda.webapps.schema.descriptors.operate.template.PostImporterQueueTemplate;
+import io.camunda.webapps.schema.entities.operate.IncidentEntity;
 import io.camunda.webapps.schema.entities.operate.IncidentState;
 import io.camunda.webapps.schema.entities.operate.post.PostImporterActionType;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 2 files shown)


npepinpe:
Blocked by #25144

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -62,6 +65,38 @@ public CompletionStage<PendingIncidentUpdateBatch> getPendingIncidentsBatch(
...
+    final var request = createIncidentDocumentsRequest(incidentIds);
+
+    return client
+        .search(request, IncidentEntity.class)
+        .thenApplyAsync(this::createIncidentDocuments, executor);
+  }
+
+  private SearchRequest createIncidentDocumentsRequest(final List<String> incidentIds) {
+    final var idQ = QueryBuilders.ids(i -> i.values(incidentIds));
+    final var partitionQ =
```
npepinpe, line 80: :thought_balloon: This may not be really necessary, because the partition ID is actually encoded in the key/ID - but this relies on the fact that ID == key, which I'm not super keen on. So I added the filter here.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -62,6 +65,38 @@ public CompletionStage<PendingIncidentUpdateBatch> getPendingIncidentsBatch(
...
+        QueryBuilders.term(t -> t.field(IncidentTemplate.PARTITION_ID).value(partitionId));
+    return new SearchRequest.Builder()
+        .index(incidentAlias)
+        .query(q -> q.bool(b -> b.must(idQ, partitionQ)))
+        .allowNoIndices(true)
+        .ignoreUnavailable(true)
+        .build();
```
npepinpe, line 87: :exclamation: I removed the sort by key from the original implementation. I tried to understand why it was there in the first place, and honestly it doesn't make much sense. I assume it's a historical leftover?

Sorting by key (since keys are monotonically generated) gives you _some_ indication of chronology - but I still don't see how it's relevant here in any way, or how it used to be before that.

entangled90: :thought_balloon:  Should we remove the sorting after we tested that this implementation is working as expected? This may help simplify testing it since we have done less changes. 
Removing the sorting afterwards is quite quick.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -84,24 +119,16 @@ private SearchRequest createPendingIncidentsBatchRequest(final int size, final Q
...
     final var typeQ =
-        QueryBuilders.term()
-            .field(PostImporterQueueTemplate.ACTION_TYPE)
-            .value(f -> f.anyValue(JsonData.of(PostImporterActionType.INCIDENT)))
-            .build()
-            ._toQuery();
+        QueryBuilders.term(
+            t ->
+                t.field(PostImporterQueueTemplate.ACTION_TYPE)
+                    .value(PostImporterActionType.INCIDENT.name()));
```
entangled90, line 128: :question: is there any difference in using `value` vs `anyValue`?

entangled90 (approved):
Look good. Personally I would keep the sorting for the moment but I think it's your call :+1:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
npepinpe: Added the sort back just in case, you're right, easy to remove it later. I don't know if it has any sort of performance implications on the ES side (probably - but then again, unclear if it's important).

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
npepinpe: AFAIK it's just a short cut, but I didn't dig too much into it. I mostly lifted it from the old code anyway, as I wasn't sure how the client will serialize an enum.

PR #25151, opened 2024-11-25
refactor: make tests easily extensible to other DBs

npepinpe:
## Description

Refactors the `ElasticsearchIncidentUpdateRepositoryIT` into a single `IncidentUpdateRepositoryIT` with per-store sub classes, which will later let us extend it to OS very easily.

I know there are different ways of doing this:

1. `TestTemplate`. This is what's used for the other exporter tests. I opted against it here specifically for a few reasons. You can't compose extensions easily in Junit, so when you end up using a custom `TestTemplateInvocationContextProvider` extension, you can't reuse `Testcontainers` or `AutoCloseResources` or any other extension, which means you end up having to manually manage the lifecycle of other things yourself. You also can't really say this test should _only_ be for this context (e.g. a only OS test). Behavior of the lifecycle hooks and methods are not as obvious with `TestTemplate` either.
2. `ParameterizedTest` with specific context passed as argument. The main downside is you don't have a split between set up/tear down and the test execution, or not as nice anyway. You also can't easily reuse the normal lifecycle hooks, or any other extension points.

Using interfaces (or abstract class if you need to share state) is a slightly easier way, as the main class behaves just as a normal test class, but the sub-classes will let you parameterize it. It also lets you define DB specific tests in the sub class if you need it. And you can reuse and make use of any extension as you normally would. It's of course not perfect, but I think it's good compromise.

## Related issues

related to #24930


npepinpe:
Blocked by #25146

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
```diff
@@ -53,59 +52,142 @@
...
-        PARTITION_ID,
-        postImporterQueueTemplate.getAlias(),
-        incidentTemplate.getAlias(),
-        listViewTemplate.getAlias(),
-        flowNodeInstanceTemplate.getAlias(),
-        client,
-        Runnable::run);
+  protected abstract IncidentUpdateRepository createRepository();
+
+  static final class ElasticsearchIT extends IncidentUpdateRepositoryIT {
```
entangled90, line 88: :question: Do we need to add `@AutoCloseResource` here as well since it's a static inner class?

entangled90 (approved):
I really like using just inheritance for extending these tests :fire: .
Please just check my comment as I'm not sure if it's necessary or not.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
npepinpe: The annotation is inherited, so no need.

npepinpe:
Force pushed, but actually just rebasing on the base branch.

PR #25154, opened 2024-11-25
Implement bulkUpdate for the ElasticsearchIncidentUpdateRepository

npepinpe:
## Description

Implements the `bulkUpdate` method for the `ElasticsearchIncidentUpdateRepository`. 

For tests, I opted against using our common search clients from the search modules, specifically because they abstract away where or how we find these entities, and here we want to specifically see that we've updated X field for document Y in index Z. Since it's not so much to implement a really basic abstract search, I figured that's a better compromise.

## Related issues

related to #24930

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
@@ -8,11 +8,17 @@
 package io.camunda.exporter.tasks.incident;
 
 import co.elastic.clients.elasticsearch.ElasticsearchAsyncClient;
+import co.elastic.clients.elasticsearch._types.ErrorCause;
+import co.elastic.clients.elasticsearch._types.Refresh;
 import co.elastic.clients.elasticsearch._types.SortOrder;
 import co.elastic.clients.elasticsearch._types.query_dsl.Query;
 import co.elastic.clients.elasticsearch._types.query_dsl.QueryBuilders;
+import co.elastic.clients.elasticsearch.core.BulkRequest;
 import co.elastic.clients.elasticsearch.core.SearchRequest;
 import co.elastic.clients.elasticsearch.core.SearchResponse;
+import co.elastic.clients.elasticsearch.core.bulk.BulkOperation;
+import co.elastic.clients.elasticsearch.core.bulk.BulkResponseItem;
+import co.elastic.clients.elasticsearch.core.bulk.UpdateOperation;
 import co.elastic.clients.elasticsearch.core.search.SourceFilter;
 import co.elastic.clients.json.JsonData;
 import io.camunda.exporter.tasks.incident.IncidentUpdateRepository.NoopIncidentUpdateRepository;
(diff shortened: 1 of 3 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
```diff
@@ -319,4 +348,136 @@ private void setupIncidentUpdates(
...
+      final var flowNodes =
+          search(
+              flowNodeInstanceTemplate.getFullQualifiedName(),
+              FlowNodeInstanceTemplate.KEY,
+              List.of("1"),
+              FlowNodeInstanceEntity.class);
+      assertThat(flowNodes)
+          .hasSize(1)
+          .extracting(FlowNodeInstanceEntity::isIncident)
+          .containsExactly(true);
```
entangled90, line 480: :thought_balloon: I think we should add a test case when there's more than 1 element for each collection in IncidentBulkUpdate

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -75,6 +87,40 @@ public CompletionStage<Map<String, IncidentDocument>> getIncidentDocuments(
...
   }
 
+  @Override
+  public CompletionStage<Integer> bulkUpdate(final IncidentBulkUpdate bulk) {
+    final var updates = bulk.stream().map(this::createUpdateOperation).toList();
+    final var request =
+        new BulkRequest.Builder()
+            .operations(updates)
+            .source(s -> s.fetch(false))
+            .refresh(Refresh.WaitFor)
```
entangled90, line 97: :exclamation: I trust your expertise in settings these parameters, as I don't have much experience in ES queries

entangled90 (approved):
:rocket: Just check my comment on the tests :+1:

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
npepinpe: I thought about it - technically it would pass if I did something strange like only updating a single element of the map. I thought maybe it's too simple logic tbh, but I can refactor tests to always have 2 entities :shrug:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
npepinpe: I discussed it with Svetlana already: https://camunda.slack.com/archives/C06F0GLJNFM/p1732038876206759

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
npepinpe: Rewrote tests to include multiple items.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -150,5 +196,20 @@ private PendingIncidentUpdateBatch createPendingIncidentBatch(
     return new PendingIncidentUpdateBatch(highestPosition, incidents);
   }
 
+  private Throwable collectBulkErrors(final List<BulkResponseItem> items) {
```
npepinpe, line 199: :thought_balloon: FYI, taken from the old `ElasticsearchExporter` code

PR #25202, opened 2024-11-25
feat: java client supports updating and deleting tenants

Oleksiivanov:
## Description

 - added to Java client UpdateTenantCommand and DeleteTenantCommand comands 
 - added tests
 - correct code related to update tenant
 

## Checklist

- [ ] for CI changes:

## Related issues

closes #25018 and #25019

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeleteTenantCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeleteTenantCommandImpl.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright © 2017 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.zeebe.client.impl.command;
+
+import io.camunda.zeebe.client.api.ZeebeFuture;
+import io.camunda.zeebe.client.api.command.DeleteTenantCommandStep1;
+import io.camunda.zeebe.client.api.command.FinalCommandStep;
+import io.camunda.zeebe.client.api.response.DeleteTenantResponse;
+import io.camunda.zeebe.client.impl.http.HttpClient;
+import io.camunda.zeebe.client.impl.http.HttpZeebeFuture;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import org.apache.hc.client5.http.config.RequestConfig;
+
+public final class DeleteTenantCommandImpl implements DeleteTenantCommandStep1 {
+  private long tenantKey;
+  private final HttpClient httpClient;
+  private final RequestConfig.Builder httpRequestConfig;
+
+  public DeleteTenantCommandImpl(final HttpClient httpClient) {
+    this.httpClient = httpClient;
+    httpRequestConfig = httpClient.newRequestConfig();
+  }
+
+  @Override
+  public DeleteTenantCommandStep1 tenantKey(final long tenantKey) {
+    this.tenantKey = tenantKey;
+    return this;
+  }
+
+  @Override
+  public FinalCommandStep<DeleteTenantResponse> requestTimeout(final Duration requestTimeout) {
+    httpRequestConfig.setResponseTimeout(requestTimeout.toMillis(), TimeUnit.MILLISECONDS);
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<DeleteTenantResponse> send() {
+    ArgumentUtil.ensureGreaterThan("tenantKey", tenantKey, 0);
+    final HttpZeebeFuture<DeleteTenantResponse> result = new HttpZeebeFuture<>();
+    httpClient.delete("/tenants/" + tenantKey, httpRequestConfig.build(), result);
+    return result;
+  }
+}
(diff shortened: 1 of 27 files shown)


clients/java/src/main/java/io/camunda/zeebe/client/impl/command/UpdateTenantCommandImpl.java
```diff
@@ -0,0 +1,76 @@
...
+      final HttpClient httpClient, final JsonMapper jsonMapper, final long tenantKey) {
+    request = new TenantUpdateRequest();
+    this.httpClient = httpClient;
+    this.jsonMapper = jsonMapper;
+    httpRequestConfig = httpClient.newRequestConfig();
+    this.tenantKey = tenantKey;
+  }
+
+  @Override
+  public UpdateTenantCommandStep1 name(final String name) {
```
emulbreh, line 49: Do you know why we're using this pattern to enforce an order for the builder calls?

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/UpdateTenantCommandImpl.java
```diff
@@ -0,0 +1,76 @@
...
+    httpRequestConfig.setResponseTimeout(requestTimeout.toMillis(), TimeUnit.MILLISECONDS);
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<UpdateTenantResponse> send() {
+    ArgumentUtil.ensureGreaterThan("tenantKey", tenantKey, 0);
+    ArgumentUtil.ensureNotNull("name", request.getName());
+    final HttpZeebeFuture<UpdateTenantResponse> result = new HttpZeebeFuture<>();
+    final UpdateTenantResponseImpl response = new UpdateTenantResponseImpl();
```
emulbreh, line 65: What is this response used for? The only reference I see is the response::setResponse below, but that doesn't have any side effects besides mutating the response instance.

emulbreh: I'm sorry, now I see it. The setter returns `this` and it's used as a transform. It'd be nicer to use a constructor for this and pass UpdateTenantResponseImpl::new (so the response can be immutable), but as this is the pattern used for the other commands this isn't really for this PR.

service/src/test/java/io/camunda/service/TenantServiceTest.java
```diff
@@ -142,8 +142,8 @@ public void shouldUpdateTenantName() {
     final BrokerTenantUpdateRequest request = stubbedBrokerClient.getSingleBrokerRequest();
     assertThat(request.getIntent()).isEqualTo(TenantIntent.UPDATE);
     assertThat(request.getValueType()).isEqualTo(ValueType.TENANT);
-    assertThat(request.getKey()).isEqualTo(tenantDTO.key());
```
emulbreh, line 145: This is still true, right?

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeleteTenantCommandImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  public FinalCommandStep<DeleteTenantResponse> requestTimeout(final Duration requestTimeout) {
+    httpRequestConfig.setResponseTimeout(requestTimeout.toMillis(), TimeUnit.MILLISECONDS);
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<DeleteTenantResponse> send() {
+    ArgumentUtil.ensureGreaterThan("tenantKey", tenantKey, 0);
```
emulbreh, line 52: Do we need this check? We don't usually seem to check this the key and this seems like the wrong layer to assume anything about the key encoding.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/broker/request/tenant/BrokerTenantUpdateRequest.java
```diff
@@ -20,7 +20,7 @@ public class BrokerTenantUpdateRequest extends BrokerExecuteCommand<TenantRecord
   public BrokerTenantUpdateRequest(final long key) {
     super(ValueType.TENANT, TenantIntent.UPDATE);
     setPartitionId(Protocol.DEPLOYMENT_PARTITION);
-    request.setKey(key);
```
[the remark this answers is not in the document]
emulbreh: Thanks for sharing that! I was wondering before why we some commands do include this line.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/tenant/TenantUpdateProcessor.java
```diff
@@ -111,11 +111,9 @@ private boolean tenantIdConflict(final TenantRecord record) {
...
       final TenantRecord existingTenant, final TenantRecord updateRecord) {
-    if (!updateRecord.getTenantId().isEmpty()) {
-      existingTenant.setTenantId(updateRecord.getTenantId());
-    }
-    if (!updateRecord.getName().isEmpty()) {
-      existingTenant.setName(updateRecord.getName());
+    final var updatedName = updateRecord.getName();
+    if (!updatedName.isEmpty()) {
+      existingTenant.setName(updatedName);
     }
```
emulbreh, line 117: Can we throw an exception here if the tenant id is non-empty?

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/UpdateTenantCommandImpl.java
emulbreh: Thank you, that was exactly my question. It does enforce the order, which makes the API less flexible, but since the commands are mutable it can't actually enforce validation. Anyway, this is unrelated to this PR, I'm just curious how this design came to be.

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/DeleteTenantCommandImpl.java
emulbreh: Checking for known invalid values in the client makes sense for me. The key is an opaque value though: we only know that it can't be negative today because we know it is generated, but there's no guarantee that will always be the case. 
In any case, the API will [switch to String keys](https://camunda.slack.com/archives/C06UKS51QV9/p1732305476336959?thread_ts=1731429420.606829&cid=C06UKS51QV9) in the near future (for unrelated reasons), but that will make it clearer that keys are opaque values and clients shouldn't parse or validate them.

emulbreh (approved):

PR #25219, opened 2024-11-25
refactor: ES implementation of wasProcessInstanceDeleted

npepinpe:
## Description

This PR implements `wasProcessInstanceDeleted` for the `ElasticsearchIncidentUpdateRepository`. This switches the old implementation from a search request (with a size 0, relying on the response metadata to detect existence) to a simple count request - which is functionally the same.

## Related issues

related to #24930

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
@@ -49,6 +49,10 @@ import java.util.stream.Collectors;
 public final class ElasticsearchIncidentUpdateRepository extends NoopIncidentUpdateRepository {
 
   private static final int RETRY_COUNT = 3;
+  private static final List<FieldValue> DELETED_OPERATION_STATES =
+      List.of(
+          FieldValue.of(OperationState.SENT.name()),
+          FieldValue.of(OperationState.COMPLETED.name()));
   private final int partitionId;
   private final String pendingUpdateAlias;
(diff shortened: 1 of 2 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -33,6 +33,7 @@
           <package name="" withSubpackages="true" static="false" />
         </value>
       </option>
+      <option name="ENABLE_JAVADOC_FORMATTING" value="false" />
```
entangled90, line 36: :question:  Did you add this voluntarily?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -129,6 +151,25 @@ public CompletionStage<List<String>> analyzeTreePath(final String treePath) {
             response -> response.tokens().stream().map(AnalyzeToken::token).toList(), executor);
   }
 
+  private Query createProcessInstanceDeletedQuery(final long processInstanceKey) {
+    final var stateValues =
```
entangled90, line 155: :thought_balloon:  Not really necessary or important, but it can be static (and not modifiable)

entangled90 (approved):

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
npepinpe: Good catch :+1:

Issue #25245, opened 2024-11-25
Delete tests already covered by handler unit tests


mustafadagher:
Based on the analysis done in the document  https://docs.google.com/document/d/12vhPUj6nNqpKmhP7L7hrNuLcRgnllKGBGHnyIneDaUs/edit?usp=sharing

ChrisKujawa:
@mustafadagher can we close this?

PR #25246, opened 2024-11-25
Delete tests already covered by handler unit tests

mustafadagher:
## Description
Delete tests already covered by handler unit tests from `EventZeebeRecordProcessorIT` and `ListViewZeebeRecordProcessorIT`

PS: the tests that are still left in those two classes fall under two categories: 
1. `shouldImportMessageSubscriptionMigratedEvent`: we don't have a handler currently that handles `ProcessMessageSubscriptionIntent.MIGRATED`. Should that be added?
2. Many tests that verify we don't override the event entity if the exported record position is smaller than the current Event entity position.  I can't see how was that guaranteed in the operate's `EventZeebeRecordProcessor` before. Do we need an IT to ensure this behavior? how and where is this guaranteed?

## Related issues

relates to #25245


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewFlowNodeFromIncidentHandlerTest.java
```diff
@@ -156,6 +156,8 @@ public void shouldUpdateEntityFromRecord() {
         .isEqualTo(incidentRecord.getValue().getTenantId());
     assertThat(flowNodeInstanceForListViewEntity.getJoinRelation().getParent())
         .isEqualTo(incidentRecord.getValue().getProcessInstanceKey());
+    assertThat(flowNodeInstanceForListViewEntity.getPositionIncident())
+        .isEqualTo(incidentRecord.getPosition());
```
ChrisKujawa, line 160: 👍🏼

operate/qa/integration-tests/src/test/java/io/camunda/operate/zeebeimport/processors/EventZeebeRecordProcessorIT.java
```diff
@@ -187,89 +187,6 @@ public void shouldNotOverrideEventProcessMessageFields()
...
-    assertThat(updatedEvent.getProcessInstanceKey()).isEqualTo(event.getProcessInstanceKey());
-    assertThat(updatedEvent.getFlowNodeInstanceKey()).isEqualTo(event.getFlowNodeInstanceKey());
-    assertThat(updatedEvent.getMetadata()).isNull();
-    // new values
-    assertThat(updatedEvent.getEventSourceType()).isEqualTo(EventSourceType.PROCESS_INSTANCE);
-    assertThat(updatedEvent.getDateTime()).isNotNull();
-    assertThat(updatedEvent.getEventType()).isEqualTo(EventType.ELEMENT_COMPLETED);
-    assertThat(updatedEvent.getBpmnProcessId()).isEqualTo(newBpmnProcessId);
-    assertThat(updatedEvent.getProcessDefinitionKey()).isEqualTo(newProcessDefinitionKey);
-    assertThat(updatedEvent.getPosition()).isEqualTo(newPosition);
```
ChrisKujawa, line 465: 🔧 This is not checked in our tests

ChrisKujawa (approved):
Thanks @mustafadagher can you check again whether we assert all properties in our handler unit tests? It looks to me we miss some?

mustafadagher:
@ChrisKujawa @sdorokhova I'd love to hear your input about the points I left in the PR description

>PS: the tests that are still left in those two classes fall under two categories:
>
>1. shouldImportMessageSubscriptionMigratedEvent: we don't have a handler currently that handles ProcessMessageSubscriptionIntent.MIGRATED. Should that be added?
>2. Many tests that verify we don't override the event entity if the exported record position is smaller than the current Event entity position. I can't see how was that guaranteed in the operate's EventZeebeRecordProcessor before. Do we need an IT to ensure this behavior? how and where is this guaranteed?

mustafadagher:
> @ChrisKujawa @sdorokhova I'd love to hear your input about the points I left in the PR description
> 
> > PS: the tests that are still left in those two classes fall under two categories:
> > 
> > 1. shouldImportMessageSubscriptionMigratedEvent: we don't have a handler currently that handles ProcessMessageSubscriptionIntent.MIGRATED. Should that be added?
> > 2. Many tests that verify we don't override the event entity if the exported record position is smaller than the current Event entity position. I can't see how was that guaranteed in the operate's EventZeebeRecordProcessor before. Do we need an IT to ensure this behavior? how and where is this guaranteed?

For point number 2:
I tried to create an CamundaExporter IT test case to test it, but the test fails since the exporter doesn't have this guarantee. But I looked around, and I found that this is guaranteed by the `ExporterContainer`:
https://github.com/camunda/camunda/blob/43f63a4cf86b3b24cec9eaffb514b76571455e56/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterContainer.java#L218-L232

And it's tested in `ExporterContainerTest.shouldNotExportWhenRecordPositionIsSmaller`

I'll delete those tests then from here.

mustafadagher:
I also added support for `ProcessMessageSubscriptionIntent.MIGRATED` event in `EventFromProcessMessageSubscriptionHandler`

sdorokhova (approved):
Message subscription migration fix looks good to me. Thank you for taking care of this!

PR #25254, opened 2024-11-26
feat: implement tenant migration for identity

### LINKED ISSUE #25132: Camunda support tenant migration by implementing TenantMigrationHandler

maryarm:
## Description

In this PR, the migration handler for tenants (without members) is implemented.

## Related issues

closes #25132

### DIFF
--- a/migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationRunner.java
+++ b/migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationRunner.java
@@ -9,27 +9,32 @@ package io.camunda.migration.identity;
 
 import static java.util.Arrays.asList;
 
+import com.google.common.util.concurrent.Uninterruptibles;
 import io.camunda.migration.api.Migrator;
-import io.camunda.service.AuthorizationServices;
-import io.camunda.service.UserServices;
+import io.camunda.migration.identity.config.IdentityMigrationProperties;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.boot.ApplicationArguments;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Component;
 
+@EnableConfigurationProperties(IdentityMigrationProperties.class)
 @Component("identity-migrator")
 public class MigrationRunner implements Migrator {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(MigrationRunner.class);
+
   private ApplicationArguments args;
-  private final UserServices userService;
-  private final AuthorizationServices authorizationServices;
+
   private final AuthorizationMigrationHandler authorizationMigrationHandler;
+  private final TenantMigrationHandler tenantMigrationHandler;
 
   public MigrationRunner(
-      final UserServices userService,
-      final AuthorizationServices authorizationServices,
-      final AuthorizationMigrationHandler authorizationMigrationHandler) {
-    this.userService = userService;
-    this.authorizationServices = authorizationServices;
+      final AuthorizationMigrationHandler authorizationMigrationHandler,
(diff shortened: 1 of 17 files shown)


maryarm:
Hey @panagiotisgts and @ChrisKujawa, here is the draft PR for identity tenant migration. that would be nice if you take a look please. 
Also I have a few questions regarding configs here. Identity Migration needs to know the zeebe gateway-address and related properties, and also it's required to have the url for management identity. With that said, here are my Questions:

1. Should Zeebe properties be defined specific to identity (`camunda.migration.identity`) or on a higher level?
2. Is application-migration.properties the correct place  for these properties?
3. I'm unclear about M2M token that controller is using to access identity. Can we setup it for migration as a profile property or env variable?
4. Another important property is how I get the organizationId in each cluster. (It's not important for tenant as they are not available in SaaS and not organization specific but it's required for other entities)

panagiotisgts:
@maryarm 
> Should Zeebe properties be defined specific to identity (camunda.migration.identity) or on a higher level?

As this migration will be run as part of the C8 app bundle, you can have access to properties already defined, such as Zeebe/DB connection config etc. So if you can reuse these it should be good so as not to introduce too many new configuration changes, since these will have to be ported to HELM charts as well. In the Module [config](https://github.com/camunda/camunda/blob/13df3402d6f3a0c473fa3ed4e6808c28a1526480/dist/src/main/java/io/camunda/application/commons/migration/IdentityMigrationModuleConfiguration.java) you can include other services that can contain necessary configuration classes.

> Is application-migration.properties the correct place for these properties?

:+1:  Yes any new property required for the migration app can be placed in this file.

migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationRunner.java
```diff
@@ -38,18 +42,37 @@ public void run() {
...
+      }
+    }
   }
 
   @Override
   public void acceptArguments(final ApplicationArguments args) {
     this.args = args;
   }
+
+  public static void main(final String[] args) {
```
panagiotisgts, line 75: :x: This should be run as part of the StandaloneMigration, which runs all migrations, or the StandaloneCamunda by enabling the `Profile.IDENTITY_MIGRATION`

migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationRunner.java
```diff
@@ -38,18 +42,37 @@ public void run() {
...
 
-    // TODO: place holder to logic
-    System.out.println("Migration Logic");
+    if ("migrate".equals(command)) {
+      migrate();
+    }
+  }
+
+  private void migrate() {
+    while (true) {
```
panagiotisgts, line 54: :question: Will there be any conditions that the migration should end gracefully and not end up in a loop here?

operate/qa/integration-tests/pom.xml
```diff
@@ -43,6 +43,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>operate-importer-8_7</artifactId>
```
panagiotisgts, line 48: AFAIK the 8.7 importers will be removed since the new CamundaExporter will be taking over for 8.7. Why is this and the respective `tasklist-importer` required?

panagiotisgts (commented):
Can't verify the actual migration logic, left some general comments.

operate/qa/integration-tests/pom.xml
maryarm: I'm also confused, it's not related to my changes but I just added it to make the pipleline green while checking for dependency:analayze

lenaschoenburg (approved):
I'm just approving because I already started reviewing https://github.com/camunda/camunda/pull/25801 which builds on top of this

Issue #24441, opened 2024-11-07
[EPIC]: Remove 8.7 importing


ChrisKujawa:
### Description 

We want to get rid of the 8.7 importing, as part of moving the data aggregation to the Camunda Exporter completely.

### This means:

 - Camunda Exporter will take care of 8.7 data
 - Importers will take care of importing the rest of the data from 8.6, which means we will keep 8.6 Importers

This is for multiple reasons necessary:

 * indices about authentication will only be supported by Camunda Exporter - soon BE's should be migrated to
 * Tasklist V1 will make use of harmonized indices - only available via Camunda Exporter

For context https://camunda.slack.com/archives/C06F0GLJNFM/p1730886411155929

### Pre-requisite:

 * Current IT relies on Importer
 * Integration tests need to be adjusted to make use of Camunda Exporter

 
```[tasklist]
### Tasks
- [x] Delete tests that verify specific logic of importing/exporting a record - Should be part of Exporter IT
- [ ] https://github.com/camunda/camunda/issues/24822
- [x] Delete tests that verify specific logic of importing/exporting a record - e.g data is trimmed on importing
- [x] Adjust/refactor tests that verify query/search API (we want to keep v1 APIs for now in tact) - make use of the exporter - https://github.com/camunda/camunda/pull/25188
- [x] Remove 8.7 importers https://github.com/camunda/camunda/pull/25259
- [ ] https://github.com/camunda/camunda/issues/23044
- [ ] https://github.com/camunda/camunda/pull/24620
- [ ] https://github.com/camunda/camunda/issues/25245
- [ ] https://github.com/camunda/camunda/pull/25188
- [x] Migrate Tasklist E2E to Camunda Exporter - after we fixed the user task search
- [ ] https://github.com/camunda/camunda/issues/23045
- [ ] https://github.com/camunda/camunda/issues/25798
- [x] Remove Takslist 8.7 Importer - after IT and E2E are migrated
- [ ] https://github.com/camunda/camunda/pull/26325
```

ChrisKujawa:
Moved https://github.com/camunda/camunda/issues/22897 out of the epic, as we will keep it for the alpha3, and will remove it in January (or make one SchemaManager to resolve race conditions with REST API, need to be defined)

Context https://camunda.slack.com/archives/C06F0GLJNFM/p1733989369486779

ChrisKujawa:
Everything done!

PR #25259, opened 2024-11-26
refactor: delete 8.7 importer

ChrisKujawa:
## Description

8.7 Importer is no longer used, and was replaced
by the Camunda Exporter

## Checklist

- [ ] for CI changes:

## Related issues

related #24441


.github/workflows/tasklist-e2e-tests.yml
```diff
@@ -51,9 +51,9 @@ jobs:
         image: camunda/zeebe:8.6.0-alpha3
```
mustafadagher, line 51: ❓ @panagiotisgts should this be changed to `camunda/zeebe:SNAPSHOT`

mustafadagher:
💭 I have a feeling that the e2e tests fail because the client/webapp uses an `8.6` version of the camunda8 nodejs sdk.

@sdorokhova do you have an idea about (how to fix) this?

mustafadagher (approved):

ChrisKujawa:
> 💭 I have a feeling that the e2e tests fail because the client/webapp uses an `8.6` version of the camunda8 nodejs sdk.
> 
> @sdorokhova do you have an idea about (how to fix) this?

@pedesen do you have input on this?

pedesen:
> 💭 I have a feeling that the e2e tests fail because the client/webapp uses an 8.6 version of the camunda8 nodejs sdk.

@mustafadagher @ChrisKujawa we're mostly using the [ZeebeGrpcClient](https://camunda.github.io/camunda-8-js-sdk/classes/index.Zeebe.ZeebeGrpcClient.html) from camunda8 nodejs sdk, which seems to be fine.

But we're also using [ZeebeRestClient](https://camunda.github.io/camunda-8-js-sdk/classes/index.Zeebe.ZeebeRestClient.html) to complete user tasks in the e2e tests, which is the reason why the processInstanceListeners tests are failing. I guess we need to find another way to achieve this in the tests? cc @vitorwtavares

ChrisKujawa:
> > 💭 I have a feeling that the e2e tests fail because the client/webapp uses an 8.6 version of the camunda8 nodejs sdk.
> 
> @mustafadagher @ChrisKujawa we're mostly using the [ZeebeGrpcClient](https://camunda.github.io/camunda-8-js-sdk/classes/index.Zeebe.ZeebeGrpcClient.html) from camunda8 nodejs sdk, which seems to be fine.
> 
> But we're also using [ZeebeRestClient](https://camunda.github.io/camunda-8-js-sdk/classes/index.Zeebe.ZeebeRestClient.html) to complete user tasks in the e2e tests, which is the reason why the processInstanceListeners tests are failing. I guess we need to find another way to achieve this in the tests? cc @vitorwtavares

Ah I see this was replaced right? 

Could we disable these tests for now?

mustafadagher:
> @mustafadagher @ChrisKujawa we're mostly using the [ZeebeGrpcClient](https://camunda.github.io/camunda-8-js-sdk/classes/index.Zeebe.ZeebeGrpcClient.html) from camunda8 nodejs sdk, which seems to be fine.
> 
> But we're also using [ZeebeRestClient](https://camunda.github.io/camunda-8-js-sdk/classes/index.Zeebe.ZeebeRestClient.html) to complete user tasks in the e2e tests, which is the reason why the processInstanceListeners tests are failing. I guess we need to find another way to achieve this in the tests? cc @vitorwtavares

@pedesen @vitorwtavares 
Is the issue related to the rest service front-end related? because backend-wise, the v1 rest api should be still working.

I thought it has something to do with the sdk version or such.. 

We'd appreciate any help, we want to fix and merge this ASAP.

pedesen:
We also need to replace the importer in the local dev/e2e setup: https://github.com/camunda/camunda/blob/main/operate/Makefile

Otherwise we're not able to run Operate or e2e tests locally.

pedesen:
> Is the issue related to the rest service front-end related? because backend-wise, the v1 rest api should be still working.

@mustafadagher Then something else might be the issue. We're looking into it.

operate/config/zeebe.cfg.yaml
```diff
@@ -1,8 +1,7 @@
-zeebe.broker.exporters.elasticsearch:
-  className: io.camunda.zeebe.exporter.ElasticsearchExporter
+zeebe.broker.exporters.camundaexporter:
```
ChrisKujawa, line 1: 👍🏼

mustafadagher:
Failing e2e are being investigated separately ([slack](https://camunda.slack.com/archives/C06F0GLJNFM/p1733471637652659))

PR #25298, opened 2024-11-26
Update transient message state as side-effect

### LINKED ISSUE #25166: Process instance no longer subscribed to message after unsuccessful process instance migration
**Describe the bug**

We've discovered a critical bug that can delete the `ProcessMessageSubscription` and `MessageSubscription` during [Process Instance Migration](https://docs.camunda.io/docs/next/components/concepts/process-instance-migration/), even though the migration is rejected. This should not have any affects and all changes are supposed to have been rolled back.

https://jira.camunda.com/browse/SUPPORT-24591

**To Reproduce**

- Create an instance of a process that subscribes to a message catch event
- Migrate the process instance to a process with the same message catch event, without mapping the message catch event
- Notice that the process instance migration is rejected
  - "Expected to migrate process instance '2251799813685253' but active element with id 'Activity_04uuugc' attempts to subscribe to a message it is already subscribed to with name 'PutOnHold'. Migrating active elements that subscribe to a message they are already subscribed to is not possible yet. Please provide a mapping instruction to message catch event with id 'Event_0qc5a9d' to migrate the respective message subscription."
- After at least 30 seconds, publish a message
- Notice that the message is not correlated

**To recover: retry the process instance migration after waiting at least 30 seconds. This will automatically recreate the deleted message subscription as [documented](https://docs.camunda.io/docs/next/components/concepts/process-instance-migration/#dealing-with-catch-events)**:
> Add catch events: if a catch event of the target process is not the target of a mapping instruction, then a new subscription is opened during migration.

**Expected behavior**

When the migration is rejected, the process instance should not have been changed.

**Log/Stacktrace**

 <p>

```
..
13:35:16 C PI_MIGRATION MIGRATE        - #27-> -1 K04 - ProcessInstanceMigrationRecord {"targetProcessDefinitionKey":2251799813685251,"mappingInstructions":[{"sourceElementId":"A","targetElementId":"A"}],"processInstanceKey":2251799813685252}
13:35:16 R PI_MIGRATION MIGRATE        - #28->#27 K04 - ProcessInstanceMigrationRecord {"targetProcessDefinitionKey":2251799813685251,"mappingInstructions":[{"sourceElementId":"A","targetElementId":"A"}],"processInstanceKey":2251799813685252} !INVALID_STATE (Expected to migrate process instance '2251799813685252' but active element with id 'A' attempts to subscribe to a message it is already subscribed to with name 'message'. Migrating active elements that subscribe to a message they are already subscribed to is not possible yet. Please provide a mapping instruction to message catch event with id 'boundary' to migrate the respective message subscription.)
13:35:46 C MSG_SUB      DELETE         - #29-> -1  -1 - "message" (inter.) @[K09] in <process ?[K04]> (no vars)
13:35:46 E MSG_SUB      DELETED        - #30->#29 K12 - "message" (inter.) correlationKey: key @[K09] in <process "process"[K04]> (no vars)
13:35:46 C PROC_MSG_SUB DELETE         - #31->#29  -1 - "message" (inter.) @[K09] in <process ?[K04]> (no vars) (tenant: <default>)
13:35:46 E PROC_MSG_SUB DELETED        - #32->#29 K10 - "message" (inter.) correlationKey: key @[K09] in <process "process"[K04]> (no vars) (tenant: <default>)
```

</p>
</details>

**Environment:**
- Zeebe Version: TBD, at least 8.6.x
- Slack channel: https://camunda.slack.com/archives/C081F9YLAF6

berkaycanbc:
## Description

This PR will enable the transient state in  `ProcessMessageSubscriptionState` to be updated as a side effect.

### Why do we need this change?

This resolves a critical bug, where a process instance is unsubscribed from a message catch event during instance migration even though the migration was rejected by throwing an exception.

When the exception is thrown, we expect all state changes to be rolled back. However, the transient state for pending process message subscriptions was not rolled back. This led to the `PendingSubscriptionChecker` to visit this pending subscription, even though the rest of the engine considered it to be rolled back.

The transient state was updated by the `DbProcessMessageSubscriptionState.updateToClosingState()` which in turn is called by the `ProcessMessageSubscriptionDeletingApplier`.

By moving the transient state update from the dbstate (applier) to a side-effect in the `CatchEventBehavior`, we're able to ensure that the state change only takes effect after the command was successfully processed (completely).

## Related issues

closes #25166


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/ProcessMessageSubscriptionCreateProcessor.java
```diff
@@ -56,6 +65,17 @@ public void processRecord(final TypedRecord<ProcessMessageSubscriptionRecord> co
...
 
+      // update transient state in a side-effect to ensure that these changes only take effect after
+      // the command has been successfully processed
+      sideEffectWriter.appendSideEffect(
+          () -> {
+            transientState.remove(
+                new PendingSubscription(
+                    subscriptionRecord.getElementInstanceKey(),
+                    subscriptionRecord.getMessageName(),
+                    subscriptionRecord.getTenantId()));
```
korthout, line 76: ❌ Accessing `subscriptionRecord` in the side effect is not thread safe

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/ProcessMessageSubscriptionCorrelateProcessor.java
```diff
@@ -109,6 +117,19 @@ final var record = command.getValue();
...
+        // update transient state in a side-effect to ensure that these changes only take effect
+        // after
+        // the command has been successfully processed
+        sideEffectWriter.appendSideEffect(
+            () -> {
+              transientState.remove(
+                  new PendingSubscription(
+                      subscriptionRecord.getElementInstanceKey(),
+                      subscriptionRecord.getMessageName(),
+                      subscriptionRecord.getTenantId()));
```
korthout, line 129: ❌ Also not threadsafe

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/BpmnProcessors.java
```diff
@@ -142,13 +145,14 @@ private static void addMessageStreamProcessors(
       final MutableProcessingState processingState,
       final Supplier<ScheduledTaskState> scheduledTaskState,
       final Writers writers,
-      final InstantSource clock) {
+      final InstantSource clock,
+      final TransientPendingSubscriptionState transientState) {
```
korthout, line 149: 🔧 There exist two different instances of `TransientPendingSubscriptionState`: one for message subscriptions and one for process message subscriptions. Let's be explicit about which is passed along here by continuing to use the same name: `transientProcessMessageSubscriptionState`.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/ProcessingDbState.java
```diff
@@ -75,15 +75,14 @@
 
 public class ProcessingDbState implements MutableProcessingState {
 
+  final TransientPendingSubscriptionState transientProcessMessageSubscriptionState;
```
korthout, line 78: 🔧 Let's make this private
🔧 Let's place this at the bottom of this state fields. All other state classes are more relevant to this class then this transient state.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/message/PendingProcessMessageSubscriptionStateTest.java
```diff
@@ -27,27 +29,33 @@ public final class PendingProcessMessageSubscriptionStateTest {
...
   }
 
   @Test
   public void shouldNoVisitSubscriptionBeforeTime() {
     // given
     final ProcessMessageSubscriptionRecord record1 = subscriptionRecordWithElementInstanceKey(1L);
     persistentState.put(1L, record1);
-    transientState.onSent(record1, 1_000L);
+    addToTransientSubscriptionState(record1);
+    pendingSubscriptionState.onSent(record1, 1_000L);
```
korthout, line 50: 🔧 We might consider combining the call to `persistentState.put` and `addToTransientSubscriptionState` into a single method, and similarly group together the other operations to clarify which operations we expect to happen together. But I'm alright if we forgo that now.

korthout (changes requested):
Thanks @berkaycanbc 

❌ There are a few changes required before we can merge this

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/message/ProcessMessageSubscriptionCorrelateProcessor.java
```diff
@@ -109,6 +117,19 @@ final var record = command.getValue();
         stateWriter.appendFollowUpEvent(
             subscription.getKey(), ProcessMessageSubscriptionIntent.CORRELATED, record);
 
+        // update transient state in a side-effect to ensure that these changes only take effect
+        // after
+        // the command has been successfully processed
+        sideEffectWriter.appendSideEffect(
```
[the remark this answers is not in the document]
berkaycanbc: I actually expected it to be more readable if I separate them per processor instead of updating couple of processors at once. 🤔  The correlate processor is appending `CORRELATED` which triggers `subscriptionState.updateToOpenedState()` and that method was already updated in the previous commit.

korthout (approved):
Thanks @berkaycanbc 🚀 

👍 LGTM

💭 Let's backport/forwardport this!

korthout:
@berkaycanbc Please also port this to `release-8.7.0-alpha2` and inform the release manager.

PR #25299, opened 2024-11-26
build: fix Optimize integrity tests

buccarel:
## Description

This fixes some recently introduced null-pointer errors that were preventing the ITs to complete successfully.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/backend/pom.xml
+++ b/optimize/backend/pom.xml
@@ -22,8 +22,6 @@
   </properties>
 
   <dependencies>
-    <!-- Logging dependencies -->
-
     <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>net.logstash.logback</groupId>
--- a/optimize/backend/pom.xml
+++ b/optimize/backend/pom.xml
@@ -22,6 +22,8 @@
   </properties>
 
   <dependencies>
+    <!-- Logging dependencies -->
+
     <!-- this is necessary. Dependency:analyze error suppressed -->
     <dependency>
       <groupId>net.logstash.logback</groupId>
--- a/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/security/ResponseHeadersConfiguration.java
+++ b/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/security/ResponseHeadersConfiguration.java
@@ -123,12 +123,18 @@ public class ResponseHeadersConfiguration {
     final Map<String, String> headers = new HashMap<>();
 
     final List<String> strictTransportSecurityHeaderValues = new ArrayList<>();
-    if (getHttpStrictTransportSecurityMaxAge() > 0) {
+    if (getHttpStrictTransportSecurityMaxAge() != null) {
       strictTransportSecurityHeaderValues.add("max-age=" + getHttpStrictTransportSecurityMaxAge());
     }
-    if (getHttpStrictTransportSecurityIncludeSubdomains()) {
+
+    Boolean includeSubDomains = getHttpStrictTransportSecurityIncludeSubdomains();
+    if(includeSubDomains == null) {
+      includeSubDomains = false;
+    }
+    if (includeSubDomains) {
       strictTransportSecurityHeaderValues.add("includeSubDomains");
     }
+
     if (!strictTransportSecurityHeaderValues.isEmpty()) {
       headers.put(
           "Strict-Transport-Security",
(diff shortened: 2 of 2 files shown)


optimize/backend/pom.xml
```diff
@@ -22,8 +22,6 @@
   </properties>
 
   <dependencies>
-    <!-- Logging dependencies -->
```
buccarel, line 25: I removed that just to trigger the ITs

oleksandr-kriuchenko-lohika (approved):

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/security/ResponseHeadersConfiguration.java
```diff
@@ -123,12 +123,18 @@ public Map<String, String> getHeadersWithValues() {
...
-    if (getHttpStrictTransportSecurityMaxAge() > 0) {
+    if (getHttpStrictTransportSecurityMaxAge() != null) {
       strictTransportSecurityHeaderValues.add("max-age=" + getHttpStrictTransportSecurityMaxAge());
     }
-    if (getHttpStrictTransportSecurityIncludeSubdomains()) {
+
+    Boolean includeSubDomains = getHttpStrictTransportSecurityIncludeSubdomains();
+    if (includeSubDomains == null) {
+      includeSubDomains = false;
+    }
```
matthewBearCamunda, line 133: Instead adding null checks, could you modify the `getHttpStrictTransportSecurityIncludeSubdomains()` method to return `boolean` instead of `Boolean`?

The only other place this method is used is in a `toString()` method

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/security/ResponseHeadersConfiguration.java
```diff
@@ -139,7 +145,11 @@ public Map<String, String> getHeadersWithValues() {
       headers.put(HttpHeaders.X_XSS_PROTECTION, getXsssProtection());
     }
 
-    if (getXContentTypeOptions()) {
+    Boolean xContentTypeOptions = getXContentTypeOptions();
+    if (xContentTypeOptions == null) {
+      xContentTypeOptions = false;
+    }
+    if (xContentTypeOptions) {
```
matthewBearCamunda, line 152: Same here. Modify `getXContentTypeOptions()` to return a `boolean` so we can avoid `null` checks altogether. The getter is also used in two places, here and `toString()`

matthewBearCamunda (changes requested):
Null checks can be avoided completely by modifying the getters

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/security/ResponseHeadersConfiguration.java
```diff
@@ -123,12 +128,14 @@ public Map<String, String> getHeadersWithValues() {
     final Map<String, String> headers = new HashMap<>();
 
     final List<String> strictTransportSecurityHeaderValues = new ArrayList<>();
-    if (getHttpStrictTransportSecurityMaxAge() > 0) {
+    if (getHttpStrictTransportSecurityMaxAge() != null) {
```
matthewBearCamunda, line 135: Checking for `> 0` and `!= null` are two different checks

The null safe check here would be something like `getHttpStrictTransportSecurityMaxAge() != null && getHttpStrictTransportSecurityMaxAge() > 0`

buccarel: But it's possible to have that header =0

oleksandr-kriuchenko-lohika: do we have to exclude 0?

buccarel: I don't think so. That was just a mistake of mine, instead of checking != null, I checked > 0

matthewBearCamunda: Then this change LGTM then, I was concerned about unintended functionality change

matthewBearCamunda (approved):

PR #25048, opened 2024-11-22
Apply tenant filtering on ES/OS reads

emulbreh:
## Description

This PR adds tenant filters to search queries for tenant-scoped entities. Tenant-scoped entities return their tenant id field from the new `IndexDescriptor::getTenantId` method. Filter transformers are refactored to hold a reference to queried IndexDescriptor instead of a hardcoded index name. This makes it possible to access the tenant id field when the SearchQuery is constructed. Since querying multiple tenant-scoped indices would require cooperation of the filter transformer and since filter transformers query at most one index today, we restrict the api accordingly: `List<String> toIndices(…)` becomes `IndexDescriptor toIndex(…)`. Finally, for simplicity, we always query the alias of the returned index descriptor. 

* [x] populate tenant request attribute from ES/OS when using auth-basic
* [x] ensure the default tenant always exists
* [x] ensure that SearchClientBasedQueryExecutor correctly restricts tenant-scoped entities

## Related issues

closes https://github.com/camunda/camunda/issues/25077, closes https://github.com/camunda/camunda/issues/23808, supersedes https://github.com/camunda/camunda/pull/24991

PR #25304, opened 2024-11-26
refactor: rename identity indices

ChrisKujawa:
## Description

With rearchitecting the Camunda platform we want to create a more holistic view of our platform, to show that we building one product. This is why we also want to remove more and more of the distinguishment and references of components.

With 8.7 we will introduce new indices related to user management (that use currently the component name `identity`). This should be replaced with `camunda`. 

This PR changes the corresponding indices.

## Related issues

Related doc about harmonized indices
https://docs.google.com/document/d/1EFZ19Gx8Nf559pP_Bg8ObFMfGYdlq20age8P_WiSBOY/edit?tab=t.0#heading=h.c447h0byekxu


romansmirnov (approved):

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/util/AuthorizationsUtil.java
```diff
@@ -116,7 +117,11 @@ public void awaitUserExistsInElasticsearch(final String username) {
                     }
                   }"""
                           .formatted(username)))
-              .uri(new URI("http://%s/identity-users-8.7.0_/_count/".formatted(elasticsearchUrl)))
+              .uri(
+                  new URI(
```
nicpuppa, line 121: @ChrisKujawa this change has an impact on the failing IT test. It might be the reason why the test is failing

remcowesterhoud: Ideally we will change this and just send a request to the API. This is legacy from before that was available 😄 Not saying you need to do this in this PR, changing the hardcoded index would work for now I suppose.

ChrisKujawa: @remcowesterhoud @nicpuppa I dont think this is related.

1. It will definitely use the right index, with this approach
2. When debugging it is not going threw this
3. When changing it back it also doesn't work

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/AuthorizationFilterTransformer.java
```diff
@@ -40,6 +40,6 @@ public SearchQuery toSearchQuery(final AuthorizationFilter filter) {
 
   @Override
   public List<String> toIndices(final AuthorizationFilter filter) {
-    return List.of("identity-authorizations-8.7.0_alias");
+    return List.of("camunda-authorization-8.7.0_alias");
```
ChrisKujawa, line 43: Just for posterity: I don't like and also do not accept that we have this hardcoded. But there is a PR in progress that will change this here https://github.com/camunda/camunda/pull/25048

PR #25310, opened 2024-11-26
[POC]: Add intra treePath cache to Exporter 

ChrisKujawa:
## Description

> [!Note]
>
> More details [here](https://docs.google.com/document/d/1PrWLli07nLDterUe8AMDAyjs_mICNx0kYQfgC5v3Fro/edit?tab=t.0)

The general looks already much more impacted

general

compared to the base.
base-general

We later realized that the benchmarks were quite fragile as the CPU throttling was kicking in (due to node restart, bad scheduling, etc.) it had a major effect on the benchmarks.

### Benchmarks results

It is not really clear how much we can get out of these, as the CPU throttling was quite high (which can be due to bad Kubernetes scheduling, that we for real more CPU usage, or just stumbled and have to catch up with jobs and messages, etc.)

cpu
gc
memory

#### Latency
overall-latency
processing-latency
exporter-latency

#### Cache 

We can see in the caching metrics that we request ~30-50 times per second the treePath, and this can take from 3 to 10 ms. Meaning per partition per second we have to do ~300ms for resolving the tree path.

cachemetrics

#### ES CPU

Interesting was that the ES CPU was slightly impacted, as we doing more requests against ES. That also impacts the ES performance, and potential also the writing.

ES-CPU

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java
@@ -180,6 +180,13 @@ public class DefaultExporterResourceProvider implements ExporterResourceProvider
                 indexDescriptorsMap.get(FormIndex.class).getFullQualifiedName()),
             new ExporterCacheMetrics("form", meterRegistry));
 
+    final var intraTreePathCache =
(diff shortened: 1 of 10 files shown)


ChrisKujawa:
supersed by https://github.com/camunda/camunda/pull/25501

PR #25189, opened 2024-11-25
Add Identity setup record and intents

remcowesterhoud:
## Description

This PR adds the `IdentitySetup` record and intents. These will be used to do an initial initialization when a broker is started. This initialization shall create an Admin role with all permissions and a demo user which will be assigned to this role.

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

Relates to #24594

PR #25340, opened 2024-11-27
fix: exclude IDENTITY_SETUP from exported value types in tests

Ben-Sheppard:
## Description

In [PR](https://github.com/camunda/camunda/pull/25189) a new record type `IDENTITY_SETUP` was added to support the initialization of Identity entities in the Zeebe engine, this record type is causing issues with some exporter tests i.e. https://github.com/camunda/camunda/actions/runs/12034254379/job/33550362077?pr=25220. 

In this PR I add the record type to the list of excluded value types in the test utils. Running the tests locally shows that a previously failing test now passes.


romansmirnov (changes requested):
@Ben-Sheppard, I guess the change must happen here as well https://github.com/camunda/camunda/blob/main/operate/qa/integration-tests/src/test/java/io/camunda/operate/util/TestSupport.java ?

romansmirnov (approved):

PR #25346, opened 2024-11-27
feat: add firstSortValues over the api responses

### LINKED ISSUE #24625: Support `firstSortValues` in search responses
## Description

The search query contains [firstSortValues](https://github.com/camunda/camunda/blob/4e96bfbfbad19031a2f146fb97be92976b3150f0/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L4231) that can be used in subsequent requests to `searchBefore` (page backwards).

This attribute is currently not filled when we [map a search response](https://github.com/camunda/camunda/blob/4e96bfbfbad19031a2f146fb97be92976b3150f0/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryResponseMapper.java#L172-L178).

## Acceptance criteria

* The `firstSortValues` contain the `sortValues` of the first item of a non-empty search response

nathansandi:
## Description

- Add the firstSortValues over the search response
- Adjust unit tests
- Have one integration test to validate the scenario

## Related issues

closes https://github.com/camunda/camunda/issues/24625

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
@@ -351,6 +351,12 @@ class UserTaskQueryTest {
         .isLessThan(result.items().get(1).getCreationDate());
     assertThat(result.items().get(1).getCreationDate())
         .isLessThan(result.items().get(2).getCreationDate());
+
+    // Assert First and Last Sort Value matches the first and last item
+    assertThat(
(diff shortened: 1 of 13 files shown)


tmetzke (approved):
👍 Nice, this is a lot less code than I initially expected would be necessary. Since the integration test also works, I would assume this is enough 😄 
🔧 I added two smaller points but nothing blocking. Pre-approving since I assume you'll have a look at them before merging 🙂

PR #25356, opened 2024-11-27
Add get group by user method to GroupService

### LINKED ISSUE #24884: Camunda services support retrieving groups assigned to a user
### Description
Retrieve all Groups that are assigned to a User in the `camunda-service` module.

### Acceptance Criteria
- [ ] A method exists with signature `retrieveGroupByUserKey(final Long userKey)`
- [ ] The `GroupServiceTest` class should contain unit tests for the `retrieveGroupByUserKey` method - hint: an example of working with the broker commands in tests can be found [here](ht tps://github.com/camunda/camunda/blob/main/service/src/test/java/io/camunda/service/UserServiceTest.java#L66-L81)

koevskinikola:
## Description

Add get group by user method to GroupService

## Related issues

closes #24884

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/GroupFilterTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/GroupFilterTransformer.java
@@ -8,6 +8,7 @@
 package io.camunda.search.clients.transformers.filter;
 
 import static io.camunda.search.clients.query.SearchQueryBuilders.and;
+import static io.camunda.search.clients.query.SearchQueryBuilders.hasChildQuery;
 import static io.camunda.search.clients.query.SearchQueryBuilders.term;
 
 import io.camunda.search.clients.query.SearchQuery;
@@ -23,8 +24,13 @@ public class GroupFilterTransformer implements FilterTransformer<GroupFilter> {
 
     return and(
         term(GroupIndex.JOIN, IdentityJoinRelationshipType.GROUP.getType()),
-        filter.groupKey() == null ? null : term("key", filter.groupKey()),
-        filter.name() == null ? null : term("name", filter.name()));
+        filter.groupKey() == null ? null : term(GroupIndex.KEY, filter.groupKey()),
+        filter.name() == null ? null : term(GroupIndex.NAME, filter.name()),
+        filter.memberKey() == null
+            ? null
+            : hasChildQuery(
+                IdentityJoinRelationshipType.MEMBER.getType(),
+                term(GroupIndex.MEMBER_KEY, filter.memberKey())));
   }
 
   @Override
--- a/service/src/test/java/io/camunda/service/GroupServiceTest.java
+++ b/service/src/test/java/io/camunda/service/GroupServiceTest.java
@@ -120,6 +120,21 @@ public class GroupServiceTest {
   }
 
   @Test
+  public void shoulReturnListOfGroupForGetByUserKey() {
+    // given
+    final var group1 = mock(GroupEntity.class);
+    final var group2 = mock(GroupEntity.class);
+    final var result = new SearchQueryResult<>(2, List.of(group1, group2), Arrays.array());
+    when(client.searchGroups(any())).thenReturn(result);
+
+    // when
+    final var searchQueryResult = services.getGroupsByUserKey(1L);
+
+    // then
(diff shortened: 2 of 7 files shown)


koevskinikola:
@Ben-Sheppard @emulbreh just one review is enough. I don't think for these changes a Zeebe (ZPA) team member review is required. 

I haven't worked with the search API that much, so I'm not completely certain if the filter by `memberKey` approach I added here is the most correct.

⚠️ The PR is based on #25330, so I'll merge it after that one.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/GroupFilterTransformer.java
```diff
@@ -21,7 +21,8 @@ public SearchQuery toSearchQuery(final GroupFilter filter) {
 
     return and(
         filter.groupKey() == null ? null : term("key", filter.groupKey()),
-        filter.name() == null ? null : term("name", filter.name()));
+        filter.name() == null ? null : term("name", filter.name()),
+        filter.memberKey() == null ? null : term("memberKey", filter.memberKey()));
```
emulbreh, line 25: We need to use a `has_child` query now. We also need to add a filter to exclude the child documents from the result. There's an example in my PR for role members: https://github.com/camunda/camunda/pull/24609/files#diff-06612ab14d30600a4a3a4745a5d6605ad0a28c4217223ee9d4bd529a856b5e8c

koevskinikola:
@emulbreh I've applied your feedback. 

Commit [1615680](https://github.com/camunda/camunda/pull/25356/commits/1615680f3b9e89abf5e6078d83d6e1abc8a9e703) makes the Group index properties consistent with the indexes of other Identity-related entities (ex. Roles, Tenants, Users, etc). Let me know if you would prefer to move it in a separate PR.

service/src/test/java/io/camunda/service/GroupServiceTest.java
```diff
@@ -119,6 +119,21 @@ public void shouldThrowExceptionIfGroupNotFoundByKey() {
     assertThat(services.findGroup(key)).isEmpty();
   }
 
+  @Test
+  public void shoulReturnListOfGroupForGetByUserKey() {
```
emulbreh, line 123:
```suggestion
  public void shoulReturnListOfGroupsForGetByUserKey() {
```

webapps-schema/src/main/resources/schema/opensearch/create/index/identity-groups.json
```diff
@@ -5,7 +5,7 @@
       "id": {
         "type": "keyword"
       },
-      "groupKey": {
+      "key": {
```
emulbreh, line 8: Do you know how to repair the local state after a change like this? I've struggled getting primary and secondary storage in sync when I've made similar changes.

emulbreh (approved):

webapps-schema/src/main/resources/schema/opensearch/create/index/identity-groups.json
koevskinikola: Do you mean for local testing?

emulbreh: Yes, for my local dev env.

koevskinikola: I assume the issue comes from IT tests, where ES is used. Have you tried clearing the Docker volumes, maybe even some stale test containers that might have been built with older Zeebe/Exporter snapshots?

emulbreh: The issue is with the test data I have in my local env already: I've created users, groups, tenants and roles via the API or even the UI and I'd like to keep as much of that as possible.

koevskinikola: I see. I usually don't keep local test data, but I understand how it is useful when developing the frontend. AFAIK, the (Zeebe) ES indexes were versioned, so migration with the former exporter/importer setup wasn't an issue between different Camunda 8 versions. However, in your local setup, you're probably always using a `SNAPSHOT` version, and that creates conflicts.

I would suggest reaching out to the Operate or Tasklist teams, they may have more experience with this type of local setup.

PR #25378, opened 2024-11-27
fix: close static ClientAdapter mock after each test

emulbreh:
The static mock is global within a thread and must be closed before it can be set up again.

I'm running into this when running the tests locally from IntelliJ.


npepinpe (approved):
:+1: Thanks, good catch! My bad for forgetting :pray:

PR #25384, opened 2024-11-27
feat: create rest process cache & adjust controllers to use it

### LINKED ISSUE #24894: [C8 API] Introduce cache for `flowNodeName`

danielkelemen:
## Description

- Add `ProcessCache` with in memory map to store process flow node id-names
- Create XmlUtil methods to retrieve flow node names
- Tests

## Related issues

closes https://github.com/camunda/camunda/issues/24894


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCacheItem.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+package io.camunda.zeebe.gateway.rest.cache;
+
+import java.util.Map;
+
+public record ProcessCacheItem(Map<String, String> flowNodes) {
+
+  public String getFlowNodeName(final String key) {
+    return flowNodes.getOrDefault(key, key);
+  }
```
tmetzke, line 16: 🔧 Let's clarify what kind of `key` this is to better understand what we are fetching here. Is this a flow node id?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCacheItems.java
```diff
@@ -0,0 +1,17 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.rest.cache;
+
+import java.util.Map;
+
+public record ProcessCacheItems(Map<Long, ProcessCacheItem> processCacheItems) {
+  public String getFlowNodeName(final Long processDefinitionKey, final String key) {
+    final var processCacheItem = processCacheItems.get(processDefinitionKey);
+    return processCacheItem != null ? processCacheItem.getFlowNodeName(key) : key;
```
tmetzke, line 15: ❓ Can we use the `getOrDefault` method of the `Map` interface here? If so, do we need this extra class?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -7,151 +7,116 @@
...
-    final var type = BpmnElementType.valueOf(flowNode.type().name());
-    if (type.getElementTypeName().isEmpty()) {
-      return flowNode.flowNodeId();
+  public void extractFlowNodeNames(
+      final Set<Long> processDefinitionKeys,
+      final BiConsumer<Long, ProcessFlowNode> flowNodeConsumer) {
+    final var keysList = new ArrayList<>(processDefinitionKeys);
+    final var result =
+        processDefinitionServices.search(
+            ProcessDefinitionQuery.of(q -> q.filter(f -> f.processDefinitionKeys(keysList))));
```
tmetzke, line 62: ❓ Could we potentially get a paged result and need to ensure we fetch all items? I believe @houssain-barouni has some experience with this in internal methods, i.e. he did this for fetching authorizations. This might only be available in the search client, so moving this there as proposed earlier might help 🙂

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -7,151 +7,116 @@
...
+    final var saxParserFactory = getSAXParserFactory();
+    final var is =
+        new ByteArrayInputStream(processDefinition.bpmnXml().getBytes(StandardCharsets.UTF_8));
+    final var handler = new BpmnXmlParserHandler(processDefinition, flowNodeConsumer);
+    try {
+      saxParserFactory.newSAXParser().parse(is, handler);
+    } catch (final ParserConfigurationException | SAXException | IOException e) {
+      LOG.warn("Unable to parse diagram: " + e.getMessage(), e);
+    }
   }
```
tmetzke, line 81: 💭 Are we doing something similar in the Camunda exporter? I believe we shouldn't implement this twice if we do. We could extract a common implementation of such an XML utility that receives a provider or function upon initialization that can physically fetch the data and a consumer that can transform the result. Those will differ for each consumer, i.e. REST API and exporter, but the underlying mechanism and caching should be the same.

Maybe this is not true but I assume there are some synergies we could leverage here.

tmetzke (commented):
👍 I like that we're getting a cache here 🚀 
🔧 I added some clarification questions, thoughts, and suggestions. In general, I was wondering: what does the implementation in the exporter look like and could we create a common implementation that is reused in both components?

zeebe/gateway-rest/pom.xml
```diff
@@ -241,6 +246,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot</artifactId>
+      <scope>test</scope>
+    </dependency>
+
```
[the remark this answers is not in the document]
danielkelemen: Yes, this is only needed due to the `TestConfig`.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public ProcessCacheItem load(final Long key) {
+      final var flowNodes = new ConcurrentHashMap<String, String>();
+      xmlUtil.extractFlowNodeNames(key, (pdId, node) -> flowNodes.put(node.id(), node.name()));
+      return new ProcessCacheItem(flowNodes);
+    }
+
+    @Override
+    public Map<Long, ProcessCacheItem> loadAll(final Set<? extends Long> keys) {
```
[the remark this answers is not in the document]
danielkelemen: It is but the overridden method is defined this way 🤷‍♂️
https://github.com/ben-manes/caffeine/blob/2542ca8ca22fe4c30b62ee569ee71436fc44a37e/caffeine/src/main/java/com/github/benmanes/caffeine/cache/CacheLoader.java#L90

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
```diff
@@ -0,0 +1,98 @@
...
+      final var processMap = new HashMap<Long, ProcessCacheItem>();
```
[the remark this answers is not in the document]
danielkelemen: No, this map is just returned to Caffeine to cache the key values, it won't be accessed directly anymore. Threads will access the items through Caffeine.

The flowNodes inside the `ProcessCacheItem` will be read and could even be updated by multiple threads. I don't think we will ever update them, we could even change that to `unmodifiableMap`.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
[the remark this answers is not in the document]
tmetzke: Ah, got it, I didn't see the `@Override`🙈 😅

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
danielkelemen: We already have similar stuff at multiple places (exporter, operate, tasklist and now rest).
I could check if we can extract some logic, although they all are slightly different.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/cache/ProcessCacheTestBase.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {ProcessCache.class, TestConfig.class})
+public class ProcessCacheTestBase {
```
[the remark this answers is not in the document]
danielkelemen: It doesn't need to be Spring, in fact originally I implemented this without Spring.
I guess it is nicer to have them in QA without extra `TestConfig`. I'll refactor it.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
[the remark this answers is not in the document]
tmetzke: I see, that makes sense, yes 👍 Yeah, I was also wondering about the flow nodes above, I also believe they can be an immutable map ➕

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
danielkelemen: Good point!
It looks like default size is 100 so we either need to specify higher max (`keysList.size()`) or query all pages.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCacheItems.java
danielkelemen: There are 2 maps, first returns the `ProcessCacheItem` the second the node name.
So, we could probably use the `getOrDefault` but it needs to return a `ProcessCacheItem`.
Maybe something like this:

```
processCacheItems.getOrDefault(ProcessCacheItem.empty()).getFlowNodeName(key);
```

> If so, do we need this extra class?

We don't, I just thought it was nicer this way.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/XmlUtil.java
```diff
@@ -48,21 +48,25 @@ private SAXParserFactory getSAXParserFactory() {
...
-      final BiConsumer<Long, ProcessFlowNode> flowNodeConsumer) {
+      final BiConsumer<Long, ProcessFlowNode> processDefinitionKeyFlowNodeConsumer) {
     final var keysList = new ArrayList<>(processDefinitionKeys);
     final var result =
         processDefinitionServices.search(
-            ProcessDefinitionQuery.of(q -> q.filter(f -> f.processDefinitionKeys(keysList))));
+            ProcessDefinitionQuery.of(
+                q ->
+                    q.filter(f -> f.processDefinitionKeys(keysList))
+                        .page(p -> p.size(processDefinitionKeys.size()))));
```
tmetzke, line 66: 💭 It probably is never happening, but to name it here: we might still be limited to a subset of all results we want to get. I think the `size` is just a "hint" on how many results we'd like to receive, it can be smaller if the database has an internal limit that is smaller (e.g. we request 20k process definition keys but ES will only deliver 10k max).

Is this realistic? I don't know, probably not in this case here. Just wanted to leave this here still so we don't blindly run into this 😅

tmetzke: This is why @houssain-barouni implemented `findAll` for [authorization search](https://github.com/camunda/camunda/blob/50d3a76829a966858a77cb2100ab87409d0d97d3/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java#L100).

danielkelemen: I think I will add the LOG for now. When we do the follow-up ticket with moving it into search clients, we can tackle this with a similar `findAll` approach.
What do you think @tmetzke?

tmetzke: Yep sounds pragmatic to me 👍

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCacheItems.java
danielkelemen: Dropped it and replaced with `getOrDefault`s.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
```diff
@@ -0,0 +1,105 @@
...
+  private final LoadingCache<Long, ProcessCacheItem> cache;
+  private final XmlUtil xmlUtil;
+
+  public ProcessCache(final GatewayRestConfiguration configuration, final XmlUtil xmlUtil) {
+    this.xmlUtil = xmlUtil;
+    final var cacheBuilder =
+        Caffeine.newBuilder().maximumSize(configuration.getProcessCache().getMaxSize());
+    final var ttlMillis = configuration.getProcessCache().getExpirationMillis();
+    if (ttlMillis != null && ttlMillis > 0) {
+      cacheBuilder.expireAfterAccess(ttlMillis, TimeUnit.MILLISECONDS);
```
houssain-barouni, line 46: if a Cached item is actively read (via active polling from the client) , it can remain for a long time in the cache. Should we rather use `expireAfterWrite` ? 
a TTL is typically expressed on last write

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCacheItem.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.Map;
+
+public record ProcessCacheItem(Map<String, String> flowNodeIdNameMap) {
+
+  public String getFlowNodeName(final String flowNodeId) {
+    return flowNodeIdNameMap.getOrDefault(flowNodeId, flowNodeId);
+  }
+
+  public static ProcessCacheItem empty() {
+    return new ProcessCacheItem(Collections.unmodifiableMap(new HashMap<>()));
```
houssain-barouni, line 21:
```suggestion
    return new ProcessCacheItem(Collections.emptyMap());
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCacheItem.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public record ProcessCacheItem(Map<String, String> flowNodeIdNameMap) {
+
+  public String getFlowNodeName(final String flowNodeId) {
+    return flowNodeIdNameMap.getOrDefault(flowNodeId, flowNodeId);
+  }
+
+  public static ProcessCacheItem empty() {
```
houssain-barouni, line 20: We can have a constant immutable EMPTY ProcessCacheItem, an make empty() returning it instead of creating new instances

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/config/GatewayRestConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+  public static class ProcessCacheConfiguration {
+    private static final int DEFAULT_CACHE_SIZE = 100;
+
+    /**
+     * Process cache max size. Default value: {@link ProcessCacheConfiguration#DEFAULT_CACHE_SIZE}.
+     */
+    private int maxSize = DEFAULT_CACHE_SIZE;
+
+    /** Process cache expiration milliseconds. Default value: {@code null}. */
+    private Long expirationMillis = null;
```
houssain-barouni, line 31: no caching by default?

houssain-barouni (commented):
Great job! 🚀 
I had a couple of questions and suggestions @danielkelemen

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
danielkelemen: But if it's being read so frequently then it also makes sense to keep it in the cache, doesn't it?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/config/GatewayRestConfiguration.java
danielkelemen: This only turns off the the time based eviction. Without this the cache only evicts entries when max size is reached.

We could probably find a meaningful default for this, too. I wasn't sure what could make sense and I guess it will also depend on the actual environment, so I left the default `null` (=no expiration).
Do you have any suggestion for default?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
danielkelemen: On a second though, I guess the problem is that it might remain in the cache for so long that it gets outdated. I'm not sure if that can happen with flow nodes but I get your point.
We can switch it to `expireAfterWrite`.

houssain-barouni: I guess it will not be changed as we use processDefinitionKey as key of the cache map, and any change comes with new version of the process and a new key
well it is up to you :) you can keep expireAfterAccess and rename the configuration as `expirationIdleTime` instead of expiration

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/config/GatewayRestConfiguration.java
houssain-barouni: I see, in this case, we can have up to 100 process definitions loaded in memory

danielkelemen: Yes, with all their flow nodes that have names.

nathansandi (commented):
Great job, minor questinos
one suggestion related to the caching strategy

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class ProcessCache {
+
+  private final LoadingCache<Long, ProcessCacheItem> cache;
+  private final XmlUtil xmlUtil;
+
+  public ProcessCache(final GatewayRestConfiguration configuration, final XmlUtil xmlUtil) {
+    this.xmlUtil = xmlUtil;
+    final var cacheBuilder =
+        Caffeine.newBuilder().maximumSize(configuration.getProcessCache().getMaxSize());
```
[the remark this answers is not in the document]
danielkelemen: Yes, [Caffeine](https://github.com/ben-manes/caffeine), it's really nice, we have a few already in C8.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
```diff
@@ -0,0 +1,105 @@
...
+    final var ttlMillis = configuration.getProcessCache().getExpirationMillis();
+    if (ttlMillis != null && ttlMillis > 0) {
+      cacheBuilder.expireAfterAccess(ttlMillis, TimeUnit.MILLISECONDS);
+    }
+    cache = cacheBuilder.build(new ProcessCacheLoader());
```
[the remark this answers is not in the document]
danielkelemen: Interesting feature, I didn't even know about this! ([Refresh docs](https://github.com/ben-manes/caffeine/wiki/Refresh))

I'm not sure if we need it honestly, I'd rather skip it for now.
But we could keep it in mind, we will probably add more stuff to the cache later and it could get more useful.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/cache/ProcessCache.java
danielkelemen: Added the `idle` in the name.

houssain-barouni (approved):
LGTM!

PR #25413, opened 2024-11-28
Exporter flushing when importing is done

### LINKED ISSUE #24908: Camunda exporter starts flushing when importing is done.
**Description**

The camunda exporter should wait until the tasklist and operate importers have finished importing all 8.6 records before it begins flushing.

Success Criteria: 
- [ ] The camunda exporter should read the import position index to know when importing is done.
- [ ] The exporter should not flush until importing is done

EuroLew:
## Description

The camunda exporter should wait until the tasklist and operate importers have finished importing all 8.6 records before it begins flushing.

Success Criteria: 
- [x] The camunda exporter should read the import position index to know when importing is done.
- [x] The exporter should not flush until importing is done

## Related issues

closes #24908


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
```diff
@@ -163,6 +165,14 @@ public void setVariableSizeThreshold(final Integer variableSizeThreshold) {
       this.variableSizeThreshold = variableSizeThreshold;
     }
 
+    public boolean isWaitForImporters() {
```
mustafadagher, line 168: 🔧  `shouldWaitForImporters`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -162,6 +165,14 @@ public void export(final Record<?> record) {
     }
 
     writer.addRecord(record);
+
+    if (configuration.getIndex().isWaitForImporters() && !importersCompleted) {
+      return;
```
mustafadagher, line 170: 🔧 should we log info or debug here for visibility?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -213,6 +224,21 @@ private void flushAndReschedule() {
...
+  }
+
+  private void checkImportersCompletedAndReschedule() {
+    try {
+      importersCompleted =
+          searchEngineClient.importersCompleted(partitionId, configuration.getIndex().getPrefix());
+    } catch (final Exception e) {
+      LOG.warn("Unexpected exception occurred checking importers completed, will retry later.", e);
+    }
+    scheduleImportersCompletedCheck();
```
mustafadagher, line 239: ❌ shouldn't we stop scheduling if the result was `true` once? 
in other words: we should keep checking only if it hasn't yet completed. or is my understanding wrong?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/SearchEngineClientUtils.java
```diff
@@ -62,4 +64,28 @@ public static <T> T mapToSettings(
...
+    if (recordReaderStatuses.isEmpty()) {
+      return true;
+    }
+
+    final var partitionCompleted =
+        recordReaderStatuses.stream().anyMatch(ImportPositionEntity::getCompleted);
+
+    // If all record readers have an import position document we can check their completed values
+
+    if (recordReaderStatuses.size() == allImportValueTypes.length) {
```
mustafadagher, line 81: 🔧 let's pass the `length` to the method if it's the only thing we will use

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/SearchClientAdapter.java
```diff
@@ -160,4 +160,12 @@ public String index(final String id, final String index, final Object document)
     }
     return "";
   }
+
+  public void refresh() throws IOException {
```
mustafadagher, line 164: 👍

mustafadagher (approved):
It looks good to me overall, I've just left a few comments and suggestions below.

Will pre-approve, but please have a look at the comments. It's also good to wait for @ChrisKujawa to give a second pair of eyes on this.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/ImportValueType.java
```diff
@@ -5,7 +5,7 @@
  * Licensed under the Camunda License 1.0. You may not use this file
  * except in compliance with the Camunda License 1.0.
  */
-package io.camunda.operate.zeebe;
+package io.camunda.webapps.schema.descriptors;
 
 public enum ImportValueType {
```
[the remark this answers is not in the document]
EuroLew: 1. The exporter doesn't have access to the importer and I moving is better than adding the dependency. 
2. Also some tests do use the enum for example `ElasticsearchEngineClientIT [351]` where we refer to the `getAliasTemplate` value of the enum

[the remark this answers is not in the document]
mustafadagher: In the exporter we already have a `RecordFilter` that contains a `VALUE_TYPES_2_EXPORT`, isn't that more or less the same?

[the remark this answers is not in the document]
ChrisKujawa: Or could we write a different value to the import position index? Like ALL and then mark this completed when all in the importer are completed? 🤔

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/ZeebeESConstants.java
```diff
@@ -5,7 +5,7 @@
  * Licensed under the Camunda License 1.0. You may not use this file
  * except in compliance with the Camunda License 1.0.
  */
-package io.camunda.operate.zeebe;
+package io.camunda.webapps.schema.descriptors;
```
ChrisKujawa, line 8: ❓ Why we need this?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -162,6 +165,17 @@ public void export(final Record<?> record) {
     }
 
     writer.addRecord(record);
+
```
ChrisKujawa, line 168: ❌ There is one issue with this. We can't cache forever and everything. We should have a limit of how much we put into our writer. If we reached the size we should fail, to loop in the exporter director.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -115,6 +120,7 @@ public void open(final Controller controller) {
     writer = createBatchWriter();
 
     scheduleDelayedFlush();
+    scheduleImportersCompletedCheck();
```
ChrisKujawa, line 123: ❌ We should also have a direct check, otherwise we have a delay of schedule interval.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -213,6 +227,24 @@ private void flushAndReschedule() {
     scheduleDelayedFlush();
   }
 
+  private void scheduleImportersCompletedCheck() {
+    controller.scheduleCancellableTask(
+        Duration.ofSeconds(10), this::checkImportersCompletedAndReschedule);
```
ChrisKujawa, line 232: 🔧 Do we want to make this interval configurable?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
```diff
@@ -131,6 +131,8 @@ public static final class IndexSettings {
 
     private Integer variableSizeThreshold = DEFAULT_VARIABLE_SIZE_THRESHOLD;
 
+    private boolean shouldWaitForImporters = false;
```
ChrisKujawa, line 134: ❌ Per default it must be true no?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
```diff
@@ -178,6 +184,34 @@ public void putIndexLifeCyclePolicy(final String policyName, final String deleti
...
+    }
+  }
+
+  private SearchRequest importPositionDocuments(final int partitionId, final String indexPrefix) {
+    final var importPositionIndexName =
+        new ImportPositionIndex(indexPrefix, true).getFullQualifiedName();
+    return new SearchRequest.Builder()
+        .index(importPositionIndexName)
+        .size(ImportValueType.values().length)
+        .query(q -> q.match(m -> m.field(ImportPositionIndex.PARTITION_ID).query(partitionId)))
```
ChrisKujawa, line 211: 🔧 ❓ We could already match here the completed equals true right?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
```diff
@@ -178,6 +184,34 @@ public void putIndexLifeCyclePolicy(final String policyName, final String deleti
...
+      final var docs = client.search(importPositionSearchRequest, ImportPositionEntity.class);
+      return allImportersCompleted(
+          docs.hits().hits().stream().map(Hit::source).toList(), ImportValueType.values().length);
+    } catch (final IOException e) {
+      final var errMsg =
+          String.format(
+              "Failed to search documents in the import position index for partition [%s]",
+              partitionId);
+      LOG.error(errMsg, e);
+      throw new ElasticsearchExporterException(errMsg, e);
```
ChrisKujawa, line 201: ❌ I think it is expectedto fail if this index doesn't exist? Then we should return true.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/SearchEngineClientUtils.java
```diff
@@ -62,4 +63,27 @@ public static <T> T mapToSettings(
           e);
     }
   }
+
+  public static boolean allImportersCompleted(
+      final List<ImportPositionEntity> recordReaderStatuses, final int totalValueTypesCount) {
+    // For a fresh install where there are no import position documents
+    if (recordReaderStatuses.isEmpty()) {
+      return true;
+    }
```
ChrisKujawa, line 72: ❌ Then the index is not created or?

ChrisKujawa: Or do we create the index now in the exporter?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/SearchEngineClientUtils.java
```diff
@@ -62,4 +63,27 @@ public static <T> T mapToSettings(
...
+
+    if (recordReaderStatuses.size() == totalValueTypesCount) {
+      return recordReaderStatuses.stream().allMatch(ImportPositionEntity::getCompleted);
+    }
+
+    // If some record readers are missing their import position document we can say their status
+    // is equal to that of the partition completion status.
+
+    return recordReaderStatuses.stream().allMatch(ImportPositionEntity::getCompleted)
+        && partitionCompleted;
```
ChrisKujawa, line 87: 🔧 This additional check is I think not necessary.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/ImportValueType.java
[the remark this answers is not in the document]
EuroLew: not quite sure what you mean? Are you talking about a new document in the import position index?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -489,4 +492,126 @@ private ExporterResourceProvider mockResourceProvider(
...
+
+      controller.runScheduledTasks(Duration.ofMinutes(1));
+
+      // then
+      final var record =
+          factory.generateRecord(
+              ValueType.AUTHORIZATION,
+              r -> r.withBrokerVersion("8.7.0").withTimestamp(System.currentTimeMillis()));
+
+      camundaExporter.export(record);
```
ChrisKujawa, line 541: ❓ I guess this is the when ?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -489,4 +492,126 @@ private ExporterResourceProvider mockResourceProvider(
...
+              .setAliasName(ImportValueType.DECISION.getAliasTemplate())
+              .setCompleted(false);
+
+      clientAdapter.index(
+          decisionEntity.getId(), importPositionIndex.getFullQualifiedName(), decisionEntity);
+
+      clientAdapter.refresh();
+
+      controller.runScheduledTasks(Duration.ofMinutes(1));
+
```
ChrisKujawa, line 534: 🔧 Can we compact this in a method or something?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -489,4 +492,126 @@ private ExporterResourceProvider mockResourceProvider(
...
+          factory.generateRecord(
+              ValueType.AUTHORIZATION,
+              r -> r.withBrokerVersion("8.7.0").withTimestamp(System.currentTimeMillis()));
+
+      camundaExporter.export(record);
+
+      // then
+      assertThat(controller.getPosition()).isEqualTo(record.getPosition());
+      verify(controller, times(1))
+          .updateLastExportedRecordPosition(eq(record.getPosition()), any());
```
ChrisKujawa, line 597: 🔧 we might be able to directly assert the position

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/ElasticsearchEngineClientIT.java
```diff
@@ -307,4 +314,85 @@ void shouldAccountForAllPropertyFieldsWhenGetMappings() {
...
+      elsClient.update(updateRequest, ImportPositionEntity.class);
+
+      elsClient.indices().refresh();
+
+      final var importersCompleted = elsEngineClient.importersCompleted(partitionId, indexPrefix);
+      assertThat(importersCompleted).isEqualTo(false);
+    }
+
+    @Test
+    void shouldReturnImportersCompletedForFreshInstall() {
```
ChrisKujawa, line 369: 👍🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/ElasticsearchEngineClientIT.java
```diff
@@ -307,4 +314,85 @@ void shouldAccountForAllPropertyFieldsWhenGetMappings() {
     assertThat(complexProperty.get("index")).isEqualTo(false);
     assertThat(complexProperty.get("eager_global_ordinals")).isEqualTo(true);
   }
+
+  @Nested
+  class ImportersCompleted {
```
ChrisKujawa, line 319: 👍🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/ElasticsearchEngineClientIT.java
```diff
@@ -307,4 +314,85 @@ void shouldAccountForAllPropertyFieldsWhenGetMappings() {
...
+      elsClient.bulk(createImportPositionDocuments(partitionId, importPositionIndex));
+      elsClient.indices().refresh();
+
+      // then
+      final var importersCompleted = elsEngineClient.importersCompleted(partitionId, indexPrefix);
+      assertThat(importersCompleted).isEqualTo(true);
+    }
+
+    @Test
+    void shouldReturnRecordReadersNotCompletedIfSomeReadersCompletedFieldIsFalse()
```
ChrisKujawa, line 344: 👍🏼

ChrisKujawa (changes requested):
Thanks @EuroLew I have some remarks we should take a look at. Please see below. :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
EuroLew: The index should always exist either from the importer having created it, or the exporter creates it during the `open` phase.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/ZeebeESConstants.java
EuroLew: The importers uses these enums to write into the import position index, and so for the exporter to read from the index to know which record reader is done the enums are easiest to use without introducing magic strings because for example.

I think I will take the approach where I avoid moving it and create my own import value type -> index name mapping

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
EuroLew: I didn't think there was an issue with making it configurable, but to avoid any possible issues with this field it's probably better to be not configurable

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
EuroLew: I don't think so, if we also match the completed field then we cannot differentiate between a genuinely empty index and an index where all the documents have completed=false, we could add a query for the amount of documents to differentiate but at that point that's an unnecessary extra query when we could just query all the documents with one request. There will be at most a dozen documents anyway.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
EuroLew: Not sure what you mean here I already directly assert the position, do you mean remove the verify call?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
EuroLew: I don't think there's any reason a user would want to configure this.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
EuroLew: I refactored the importers completed logic to also match the completed field

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/SearchEngineClientUtils.java
EuroLew: refactored importers completed logic so this check is not needed

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
ChrisKujawa: Lol nevermind. I thought we can do `eq(position` but we already do 🤦🏼 Sorry.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/ImportValueType.java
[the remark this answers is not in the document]
ChrisKujawa: Yes I was talking about a new document :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
```diff
@@ -131,6 +131,10 @@ public static final class IndexSettings {
 
     private Integer variableSizeThreshold = DEFAULT_VARIABLE_SIZE_THRESHOLD;
 
+    private boolean shouldWaitForImporters = true;
+
+    private Integer maxCachedRecordsBeforeErrorWhileWaitingForImporters = 1_000;
```
mustafadagher, line 136: ❓ can we re-use the `bulk.size` here instead of creating a new config value?

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/ImportValueTypes.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.webapps.schema.descriptors;
+
+public enum ImportValueTypes {
```
mustafadagher, line 10: 💭  shall we rename this to `ExportValueType` and reuse it in the `CamundaExporterRecordFilter` instead of the set defined there?

💭 I might be wrong, but I think across the camunda codebase, we use singular rather than plural enum names

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
EuroLew: Sure I don't see any harm just wanted to make sure the reasoning was clear

mustafadagher: I understand. My feeling is this is temporary, and it makes sense to stop at the configured bulk.size anyway. We can make it clear through documentation for the 8.7 version

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/ImportValueTypes.java
EuroLew: With regards to the first point I don't think so, specifically this refers to all the value types the operate and tasklist importers can handle and we only track if those ones are finished. For example the importers don't handle any `USER` type.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/config/ExporterConfiguration.java
EuroLew: Sure I don't see a problem

mustafadagher (approved):

ChrisKujawa (changes requested):
Thanks @EuroLew I think there is one issue left, regarding merging the value types for the different imports. I think we should take a look at this.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
```diff
@@ -186,12 +185,24 @@ public void putIndexLifeCyclePolicy(final String policyName, final String deleti
...
-    final var importPositionSearchRequest = importPositionDocuments(partitionId, indexPrefix);
+    final var completedImportPositionDocumentsRequest =
+        completedImportPositionDocuments(partitionId, indexPrefix);
 
     try {
-      final var docs = client.search(importPositionSearchRequest, ImportPositionEntity.class);
-      return allImportersCompleted(
-          docs.hits().hits().stream().map(Hit::source).toList(), ImportValueTypes.values().length);
+      // brand new install no need to wait for importers to complete
+      if (isImportPositionIndexEmpty(indexPrefix)) {
```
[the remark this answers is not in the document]
EuroLew: Refactored so only one request is made for all documents, therefore no need for a request checking the index is empty

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -246,16 +246,15 @@ private void scheduleImportersCompletedCheck() {
   }
 
   private void checkImportersCompletedAndReschedule() {
+    if (!importersCompleted) {
+      scheduleImportersCompletedCheck();
+    }
```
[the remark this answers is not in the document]
EuroLew: Yes, but this call was made before the importers completed being set for tests. If we put this if statement after checking if `importersCompleted` for a fresh install the importsCompleted = true and the job is never scheduled so we can't use the `testcontroller.runScheduledTasks` to manually retry this check.

This is not ideal but logically it doesn't change anything and at worst it's 1 extra check

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -205,6 +194,19 @@ public void export(final Record<?> record) {
...
   }
 
+  private void ensureCachedRecordsLessThanBulkSize() {
+    final var maxCachedRecords = configuration.getBulk().getSize();
+
+    if (writer.getBatchSize() > maxCachedRecords) {
+      final var errMsg =
+          String.format(
+              "Reached the max bulk size amount of cached records [%d] while waiting for importers to finish, restarting exporter",
+              maxCachedRecords);
```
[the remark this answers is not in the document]
EuroLew: Not sure what you mean by `error loop here for the last record`?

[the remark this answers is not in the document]
mustafadagher: I think Chris was hinting at how we handle an exception in `export(record)`. We don't restart the exporter in case of exception. but rather the `ExporterContainer` will return `false` for this specific record we failed at
https://github.com/camunda/camunda/blob/c3b85b4812a7001bf53504bb11789f1e667d9152/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterContainer.java#L218-L232

And in the `ExporterDirector` we run this method with retry (with a backoff strategy)
https://github.com/camunda/camunda/blob/c3b85b4812a7001bf53504bb11789f1e667d9152/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java#L654-L655

So in practice: we are (error) looping on exporting this last record that caused this exception

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
```diff
@@ -186,12 +185,24 @@ public void putIndexLifeCyclePolicy(final String policyName, final String deleti
...
+      }
+
+      final var totalCompletedRecordReaders =
+          client
+              .search(completedImportPositionDocumentsRequest, ImportPositionEntity.class)
+              .hits()
+              .hits()
+              .size();
+
+      return totalCompletedRecordReaders == ImportValueTypes.values().length;
```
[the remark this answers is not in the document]
EuroLew: Did not realise that operate and tasklist used two different import position indices, that definitely changes the solution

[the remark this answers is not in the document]
EuroLew: With the addition of https://github.com/camunda/camunda/pull/25616, the new solution where we check all statuses have `completed=true` should work.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -200,14 +200,14 @@ public void export(final Record<?> record) {
...
+  private void ensureCachedRecordsLessThanBulkSize(final Record<?> record) {
     final var maxCachedRecords = configuration.getBulk().getSize();
 
     if (writer.getBatchSize() > maxCachedRecords) {
       final var errMsg =
           String.format(
-              "Reached the max bulk size amount of cached records [%d] while waiting for importers to finish, restarting exporter",
-              maxCachedRecords);
+              "Reached the max bulk size amount of cached records [%d] while waiting for importers to finish, retrying export for [%s]",
+              maxCachedRecords, record.toString());
```
mustafadagher, line 210: 🔧 this is kind of a nitpicking: but I'm not sure if we should log the whole record. if we need to point to it we shall log the record position.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -200,14 +200,14 @@ public void export(final Record<?> record) {
...
       LOG.error(errMsg);
```
mustafadagher, line 211: 🔧 also, is this an error? or more of a warning?

mustafadagher:
`CamundaExporterIT.shouldCreateHarmonizedSchemaEagerlyOnOpen` is failing because we now have an extra index it's not aware of

ChrisKujawa (changes requested):
Hey Euro we are close, I found two more minor things one related to adding more and more records and one regarding tests. Please have a look. If something is unclear please let me know.

In the afternoon I might be able to support, but before 2 my day is quite packed.

ChrisKujawa (approved):
I think the changes are straightforward so I pre-approve, feel free to merge after fixing the mentioned things above.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -217,6 +248,29 @@ private void flushAndReschedule() {
     scheduleDelayedFlush();
   }
 
+  private void scheduleImportersCompletedCheck() {
+    controller.scheduleCancellableTask(
+        Duration.ofSeconds(10), this::checkImportersCompletedAndReschedule);
+  }
+
+  private void checkImportersCompletedAndReschedule() {
+    if (!importersCompleted) {
```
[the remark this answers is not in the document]
EuroLew: if it is completed it will simply not reschedule and the task will stop

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -489,4 +493,116 @@ private ExporterResourceProvider mockResourceProvider(
...
+      final var recordId = authHandler.generateIds(record).getFirst();
+
+      assertThat(
+              clientAdapter.get(recordId, authHandler.getIndexName(), authHandler.getEntityType()))
+          .isNull();
+    }
+
+    @TestTemplate
+    void shouldFlushIfImportersAreCompleted(
+        final ExporterConfiguration config, final SearchClientAdapter clientAdapter)
```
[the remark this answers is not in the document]
EuroLew: I added the test the failing when we reach the bulk size, not sure the second test is needed as exporting is idempotent, also I'm not sure why it would write twice? Each `writer.addRecord` call will only add one set of cached entities no matter how many times you call it for a record, and if a record has `export` called on it and succeeds the retry logic shouldn't kick in.

PR #25417, opened 2024-11-28
feat: expose /me endpoint including application access

### LINKED ISSUE #25416: Authorized applications are exposed on a /me endpoint to help webapps understand access
### Description
In the new approach to authorizations in 8.7+ application access is controlled via authorizations and not the old style of permissions. This means that the webapps need to understand application access in a different way.  

### Acceptance criteria
- [ ] A field is exposed on the user object returned via a `/me` endpoint that lists the authorized applications.

Ben-Sheppard:
## Description

As described in the linked issue, webapps need to use a new method of applying access to the UI, this PR tackles a couple of important parts:
- [Refactor] expose a refined user object to only show the fields that are relevant to the end users
- [Feat] implement a new AuthenticationController that exposes an `/authentication/me` endpoint to retrieve the logged in user
- [Feat] add a field to show the authorized applications 

## Related issues

closes https://github.com/camunda/camunda/issues/25416

### DIFF
--- a/authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
+++ b/authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
@@ -7,8 +7,8 @@
  */
 package io.camunda.authentication.entity;
 
-import io.camunda.search.entities.TenantEntity;
 import io.camunda.security.entity.ClusterMetadata;
+import io.camunda.service.TenantServices.TenantDTO;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -22,7 +22,7 @@ public class CamundaUser extends User {
   private final Long userKey;
   private final String displayName;
   private List<String> authorizedApplications = List.of();
-  private List<TenantEntity> tenants = List.of();
+  private List<TenantDTO> tenants = List.of();
   private List<String> groups = List.of();
   private String salesPlanType;
   private Map<ClusterMetadata.AppName, String> c8Links = new HashMap<>();
@@ -53,7 +53,7 @@ public class CamundaUser extends User {
       final String password,
       final List<String> roles,
       final List<String> authorizedApplications,
-      final List<TenantEntity> tenants,
+      final List<TenantDTO> tenants,
       final List<String> groups,
       final String salesPlanType,
       final Map<ClusterMetadata.AppName, String> c8Links,
@@ -91,11 +91,15 @@ public class CamundaUser extends User {
     return getAuthorities().stream().map(GrantedAuthority::getAuthority).toList();
   }
 
+  public List<String> getGroups() {
+    return groups;
+  }
+
   public List<String> getAuthorizedApplications() {
     return authorizedApplications;
   }
 
-  public List<TenantEntity> getTenants() {
+  public List<TenantDTO> getTenants() {
     return tenants;
   }
 
@@ -104,11 +108,18 @@ public class CamundaUser extends User {
   }
 
   public Map<ClusterMetadata.AppName, String> getC8Links() {
+    if (c8Links == null) {
+      return Collections.emptyMap();
+    }
     return c8Links;
   }
 
   public boolean canLogout() {
     return canLogout;
+  }
+
+  public boolean isApiUser() {
+    return apiUser;
   }
 
   private static List<SimpleGrantedAuthority> prepareAuthorities(final List<String> roles) {
@@ -122,7 +133,7 @@ public class CamundaUser extends User {
     private String password;
     private List<String> roles = List.of();
     private List<String> authorizedApplications = List.of();
-    private List<TenantEntity> tenants = List.of();
+    private List<TenantDTO> tenants = List.of();
     private List<String> groups = List.of();
     private String salesPlanType;
     private Map<ClusterMetadata.AppName, String> c8Links;
@@ -166,7 +177,7 @@ public class CamundaUser extends User {
       return this;
     }
 
-    public CamundaUserBuilder withTenants(final List<TenantEntity> tenants) {
+    public CamundaUserBuilder withTenants(final List<TenantDTO> tenants) {
       this.tenants = tenants;
       return this;
     }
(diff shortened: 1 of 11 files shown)


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/AuthenticationController.java
```diff
@@ -0,0 +1,33 @@
...
+package io.camunda.zeebe.gateway.rest.controller;
+
+import io.camunda.authentication.entity.CamundaUserDTO;
+import io.camunda.authentication.service.CamundaUserService;
+import org.springframework.context.annotation.Profile;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@Profile("auth-basic")
```
emulbreh, line 17: How do web apps get a description of an authenticated user?

authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
```diff
@@ -105,13 +108,20 @@ public String getSalesPlanType() {
   }
 
   public Map<ClusterMetadata.AppName, String> getC8Links() {
+    if (c8Links == null) {
+      return Collections.emptyMap();
```
emulbreh, line 112: Can we do this in the constructor or is there a case where we need to distinguish null and the empty map?

emulbreh: Yes, I'm sorry, that could have been clearer. It's unfortunate that we normalize in multiple places now, but if that is too annoying we could always make the constructors call `this()`.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4441,6 +4463,63 @@ components:
         email:
           description: The email of the user.
           type: "string"
+    CamundaUser:
+      type: "object"
+      properties:
+          userId:
+            description: The ID of the user.
```
emulbreh, line 4470: Does this have to be exposed in the API? Would this ever be something other than the userKey?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4441,6 +4463,63 @@ components:
...
+          authorizedApplications:
+            description: The applications the User is authorized to use
+            type: array
+            items:
+              type: "string"
+          tenants:
+            description: The tenants the User is a member of
+            type: array
+            items:
+              type: "object"
```
emulbreh, line 4488: Could this be a `TenantItem` ?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4441,6 +4463,63 @@ components:
...
+              properties:
+                tenantId:
+                  type: "string"
+                name:
+                  type: "string"
+          groups:
+            description: The groups assigned to the user
+            type: array
+            items:
+              type: "string"
```
emulbreh, line 4498: Are these keys or names? I believe we don't have an immutable identifier like the tenant id for roles and groups.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/AuthenticationController.java
emulbreh: From slack: 
> there will also be a OIDC user service

emulbreh (approved):

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
Ben-Sheppard: `userId` as far as I can make out from the existing webapps is actually the `username` for example `"userId": "demo",`. I think its something the respective webapps team can answer but for now I've gone with caution and added it for consistency

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
Ben-Sheppard: I didn't opt for that object because I didn't want to expose the assigned members field also, just the id + name

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
Ben-Sheppard: These are names :)

zeebe/gateway-rest/pom.xml
```diff
@@ -147,6 +147,7 @@
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-crypto</artifactId>
+      <scope>test</scope>
```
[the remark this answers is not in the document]
Ben-Sheppard: Ah ok, yea adding the scope makes it avoid the dependency issue for me :)

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/AuthenticationController.java
Ben-Sheppard: We will eventually use this pattern also for OIDC (and LDAP) but it will be in a following PR :)

qa/integration-tests/src/test/java/io/camunda/it/auth/BasicAuthIT.java
```diff
@@ -67,10 +76,22 @@ void setUp() throws JsonProcessingException {
...
+                        AuthorizationOwnerType.USER.name(),
+                        AuthorizationResourceType.APPLICATION.name(),
+                        List.of(new Permission(PermissionType.ACCESS, Set.of("*"))))),
+                null));
+
     content =
         objectMapper.writeValueAsString(
             new UserRequest()
-                .username("demo")
+                .username("demo-".concat(UUID.randomUUID().toString()))
```
emulbreh, line 94: Is this working around a test isolation issue?

emulbreh (approved):

qa/integration-tests/src/test/java/io/camunda/it/auth/BasicAuthIT.java
Ben-Sheppard: I actually think this is likely not necessary as its all mocked, initially I added this in as we were trying to create a `demo:demo` user which adds confusion with the at times default user creation. I was planning to leave it in because it doesn't hurt and also shows that this value is unique and non clashing etc

PR #25420, opened 2024-11-28
fix: Task Service Adapter to update User Tasks in a Single Application via Tasklist v1 API

### LINKED ISSUE #21818: C8 Single app - Cannot Assign/Complete Zeebe user tasks using Tasklist
### To Reproduce

1. Run Tasklist + Zeebe broker as C8 single app (same when Operate is also running in the single app)
2. Configure Tasklist to use Zeebe instance running in the same app
3. Deploy and start a process containing Zeebe user task
4. Assign the Zeebe user task using Tasklist UI
5. Internal error is returned (authentication error returned when Tasklist calls Zeebe)

### Additional context

This is discussed in https://camunda.slack.com/archives/C06HTSPD5AP/p1724680654652539

It is caused by:
- Tasklist applies security filters on path `/v*/*` see [here](https://github.com/camunda/camunda/blob/9524ebc02e639f335f73daf85b770b86067c4d0f/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/BaseWebConfigurer.java#L95).
- Tasklist calls `/v2/users-tasks` Zeebe REST endpoints to assign/complete user tasks
- When we run Tasklist alongside Zeebe, the Tasklist Spring Security package is (undesirably) applied to Zeebe's REST endpoints, in addition to the existing Zeebe authentication interceptors. While the call from Tasklist to Zeebe is authenticated, it is being rejected by Tasklist's Spring Security checks.

**Long term fix**: UI should directly invoke `/v2/users-tasks` endpoints instead of calling Tasklist /v1 endpoints.

romansmirnov:
## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/21818

### DIFF
--- a/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/service/TaskService.java
+++ b/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/service/TaskService.java
@@ -29,10 +29,6 @@ import io.camunda.tasklist.webapp.security.AssigneeMigrator;
 import io.camunda.tasklist.webapp.security.UserReader;
 import io.camunda.webapps.schema.entities.tasklist.TaskEntity;
 import io.camunda.webapps.schema.entities.tasklist.TaskEntity.TaskImplementation;
-import io.camunda.zeebe.client.ZeebeClient;
-import io.camunda.zeebe.client.api.command.ClientException;
-import io.camunda.zeebe.client.api.command.CompleteJobCommandStep1;
-import io.camunda.zeebe.client.api.response.AssignUserTaskResponse;
 import java.io.IOException;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
@@ -52,9 +48,7 @@ public class TaskService {
 
   @Autowired private UserReader userReader;
 
-  @Autowired
-  @Qualifier("tasklistZeebeClient")
-  private ZeebeClient zeebeClient;
+  @Autowired private TasklistServiceAdapter tasklistServiceAdapter;
 
   @Autowired private TaskStore taskStore;
   @Autowired private VariableService variableService;
@@ -173,16 +167,7 @@ public class TaskService {
     final String taskAssignee = determineTaskAssignee(assignee);
 
     if (taskBefore.getImplementation().equals(TaskImplementation.ZEEBE_USER_TASK)) {
-      try {
-        final AssignUserTaskResponse assigneeResponse =
-            zeebeClient
-                .newUserTaskAssignCommand(Long.parseLong(taskId))
-                .assignee(taskAssignee)
-                .send()
-                .join();
-      } catch (final ClientException exception) {
-        throw new TasklistRuntimeException(exception.getMessage());
-      }
+      tasklistServiceAdapter.assignUserTask(taskBefore, taskAssignee);
     }
 
     final TaskEntity claimedTask = taskStore.persistTaskClaim(taskBefore, taskAssignee);
@@ -210,24 +195,7 @@ public class TaskService {
 
       final TaskEntity task = taskStore.getTask(taskId);
       taskValidator.validateCanComplete(task);
-
-      try {
-        if (task.getImplementation().equals(TaskImplementation.JOB_WORKER)) {
-          // complete
-          CompleteJobCommandStep1 completeJobCommand =
-              zeebeClient.newCompleteCommand(Long.parseLong(taskId));
-          completeJobCommand = completeJobCommand.variables(variablesMap);
-          completeJobCommand.send().join();
-        } else {
-          zeebeClient
-              .newUserTaskCompleteCommand(Long.parseLong(taskId))
-              .variables(variablesMap)
-              .send()
-              .join();
-        }
-      } catch (final ClientException exception) {
-        throw new TasklistRuntimeException(exception.getMessage());
-      }
+      tasklistServiceAdapter.completeUserTask(task, variablesMap);
 
       // persist completion and variables
       final TaskEntity completedTaskEntity = taskStore.persistTaskCompletion(task);
@@ -285,10 +253,10 @@ public class TaskService {
     final TaskEntity taskEntity = taskStore.persistTaskUnclaim(taskBefore);
     if (taskBefore.getImplementation().equals(TaskImplementation.ZEEBE_USER_TASK)) {
       try {
-        zeebeClient.newUserTaskUnassignCommand(taskBefore.getKey()).send().join();
-      } catch (final ClientException exception) {
+        tasklistServiceAdapter.unassignUserTask(taskEntity);
+      } catch (final Exception exception) {
         taskStore.persistTaskClaim(taskBefore, taskBefore.getAssignee());
-        throw new TasklistRuntimeException(exception.getMessage());
+        throw exception;
       }
     }
     return TaskDTO.createFrom(taskEntity, objectMapper);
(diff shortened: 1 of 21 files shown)


tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/service/StandaloneTaskMutationAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public StandaloneTaskMutationAdapter(
+      final @Qualifier("tasklistZeebeClient") ZeebeClient zeebeClient) {
+    this.zeebeClient = zeebeClient;
+  }
+
+  @Override
+  public void assignUserTask(final TaskEntity task, final String assignee) {
+    final var taskKey = task.getKey();
+    zeebeClient.newUserTaskAssignCommand(taskKey).assignee(assignee).send().join();
```
houssain-barouni, line 31: @romansmirnov starting from 8.7, this Zeebe client will be always required to be authenticated with client credentials, otherwise it will be blocked by the new [WebSecurityConfig](https://github.com/camunda/camunda/blob/main/authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java). This is not the case in SaaS for example.

dist/src/main/java/io/camunda/tasklist/zeebe/ServiceTaskMutationAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+import io.camunda.tasklist.webapp.service.TaskMutationAdapter;
+import io.camunda.webapps.schema.entities.tasklist.TaskEntity;
+import io.camunda.zeebe.gateway.rest.RequestMapper;
+import java.util.List;
+import java.util.Map;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+@Component
+@Profile("!standalone")
```
houssain-barouni, line 23: I am wondering why this cannot be generalized to all profiles.
I think it just requires https://github.com/camunda/camunda/blob/main/dist/src/main/java/io/camunda/application/commons/broker/client/BrokerClientConfiguration.java which is always created as it is part of `io.camunda.application.commons` package

romansmirnov: In that iteration, I just wanted to differentiate between starting Tasklist standalone with `StandaloneTasklist` and as a single application with `StandaloneCamunda`. My thinking is that if a user continues deploying the old architecture with 8.7, i.e., they deploy Tasklist standalone, which gets started with `StandaloneTasklist`, then "nothing" should change in terms of how Tasklist communicates with the Gateway/Process Engine. If they configured Zeebe Client credentials and assigned specific Tenants to the Zeebe Client, then nothing should be changed here. Basically, I was thinking of keeping a level of "backward compatibility".

If running in a single application or moving to a new architecture, the communication can happen via the services (eventually via the broker clients).

That's basically the direction where I am coming from.

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/service/StandaloneTaskMutationAdapter.java
romansmirnov: But how is this different to 8.6?

With 8.6, users can also enable authentication in the Gateway (gRPC and REST), which requires the Zeebe Client to authenticate with client credentials.

With 8.7 by default, authentication will be disabled for *development* purposes, i.e., clients don't need to authenticate. (I wonder if it might be the other way around in the HELM Charts.) However, if authentication is enabled, the Zeebe Client must authenticate with client credentials.

The approach I am following is to differentiate between the old architecture and the new architecture. In the old architecture (in which Tasklist is started via StandaloneTasklist) the communication continues with the Gateway via gRPC/REST - nothing changes. The same applies to authentication. In the new architecture (in which Tasklist runs as part of the Single Application), it will communicate via the broker internally in the C8 cluster. Users or clients only need to authenticate using the v1 REST API to complete (or assign) a task, but Zeebe Client is not used in the new architecture therefore.

houssain-barouni: in 8.6, users can also enable authentication in the Gateway with `identity-auth` profile. This profile is currently used only in SM **and not in SaaS**. in SM, tasklist uses a client id and secret to call Zeebe. These are set by customers or generated by helm.

In 8.7, we will have two new profiles:
- `basic-auth` that I guess will be used for local deployments
- `oidc-auth` that will be used in SaaS and SM. 
- Both, `basic-auth` and `oidc-auth` require authentication.

So it means that in SaaS, starting from 8.7, we will need to define a client id and secret to be used in Zeebe authentication flows, and inject it in Tasklist webapp deployment.

Maybe this something I need to discuss with identity team and it is out of the scope of the initial problem reported in https://github.com/camunda/camunda/issues/21818. The issue will definitely be solved by the changes you made in this pull request 👍

dist/src/main/java/io/camunda/tasklist/zeebe/ServiceTaskMutationAdapter.java
houssain-barouni: It makes sense. I was wondering if in a next iteration we can still bypass the REST and gRPC calls even in standalone mode. For example, it can only be activated with some configuration.

dist/src/main/java/io/camunda/tasklist/zeebe/ServiceTaskMutationAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+          .completeUserTask(taskKey, variables, "");
+    }
+  }
+
+  private Authentication getAuthenticationWithAuthorizedTenant(final TaskEntity task) {
+    // if multi-tenancy is enabled, then Tasklist already does the tenant-check,
+    // meaning we can assume that the user/client is at least authorized
+    // for tenant assigned to the given task.
+    TenantAttributeHolder.withTenantIds(List.of(task.getTenantId()));
+    return RequestMapper.getAuthentication();
```
houssain-barouni, line 68: This will not work if we are outside a Request scope, for example when it is done in a job.
Today, we don't have this use-case though.

romansmirnov: As you said, the scope is tasklist or does tasklist async calls outside of a request?

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/service/TaskService.java
```diff
@@ -270,7 +248,7 @@ public TaskDTO unassignTask(final String taskId) {
     final TaskEntity taskEntity = taskStore.persistTaskUnclaim(taskBefore);
     if (taskBefore.getImplementation().equals(TaskImplementation.ZEEBE_USER_TASK)) {
       try {
-        zeebeClient.newUserTaskUnassignCommand(taskBefore.getKey()).send().join();
+        taskMutationAdapter.unassignUserTask(taskEntity);
       } catch (final ClientException exception) {
         taskStore.persistTaskClaim(taskBefore, taskBefore.getAssignee());
```
houssain-barouni, line 253: we may need to catch the runtime exceptions thrown by the broker to apply this logic 🤔

romansmirnov:
Closing in favor of https://github.com/camunda/camunda/pull/25799

PR #25422, opened 2024-11-28
Forward user claims to the broker

### LINKED ISSUE #23907: Token claims are loaded onto the authentication in gRPC classes
## Description
With the changed that arrived in https://github.com/camunda/camunda/issues/22732 we began loading the token claim data onto the authentication that is passed into the engine, this is in preparation for being able to match mapping rules (and therefore authorizations) at a command level.

The changes in https://github.com/camunda/camunda/issues/22732 however only applied to the REST API and not gRPC, to be complete and align on experience we should also apply the same changes in the relevant gRPC classes.

### Hints
- The JWT that is built in the [EndpointManager](https://github.com/camunda/camunda/blob/main/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/EndpointManager.java#L487-L508) should also include the claims from the JWT as we do in the [RequestMapper](https://github.com/camunda/camunda/blob/main/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java#L433-L436) class (in REST)

nicpuppa:
## Description

This PR adds a new interceptor to decode the token, get the user claims from the decoded token and set them into the gRPC context. When the request is forwarded to the broker the user claims are set into the Authorization metadata in a token format.
See commits messages for more details

❗ ❓  The token validation is NOT part of this PR. Since is not easy to implement we need to discuss more in details about it

## Related issues

closes #23907

### DIFF
--- a/zeebe/auth/src/main/java/io/camunda/zeebe/auth/JwtDecoder.java
+++ b/zeebe/auth/src/main/java/io/camunda/zeebe/auth/JwtDecoder.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.auth;
+
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.interfaces.Claim;
+import com.auth0.jwt.interfaces.DecodedJWT;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class JwtDecoder {
+
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+  private final String token;
+  private DecodedJWT decodedJwt;
+
+  public JwtDecoder(final String token) {
+    if (token == null || token.trim().isEmpty()) {
+      throw new IllegalArgumentException("Token cannot be null or empty");
+    }
+    this.token = token;
+  }
+
+  /**
+   * Explicitly decode the JWT. This is optional since claims access auto-decodes.
+   *
+   * @return the current JwtDecoder instance
+   * @throws RuntimeException if decoding fails
+   */
+  public JwtDecoder decode() {
+    if (decodedJwt == null) {
+      try {
+        decodedJwt = JWT.decode(token);
+      } catch (final Exception e) {
+        throw new RuntimeException("Failed to decode JWT: " + e.getMessage(), e);
+      }
+    }
+    return this;
+  }
+
+  /**
+   * Retrieves all claims as a Map<String, Object>. Automatically decodes the token if it hasn't
+   * been decoded yet.
+   *
+   * @return Map of claims
+   */
+  public Map<String, Object> getClaims() {
+    if (decodedJwt == null) {
+      decode();
+    }
+    return convertClaimsToMap();
+  }
+
+  /**
+   * Converts claims from DecodedJWT into a Map<String, Object>.
+   *
+   * @return Map of claims
+   */
+  private Map<String, Object> convertClaimsToMap() {
+    final Map<String, Claim> claims = decodedJwt.getClaims();
+    if (claims.isEmpty()) {
+      return Collections.emptyMap();
+    }
+
+    final Map<String, Object> claimsMap = new HashMap<>();
+    claims.forEach((key, claim) -> claimsMap.put(key, convertClaimValue(claim)));
+    return claimsMap;
+  }
+
+  /**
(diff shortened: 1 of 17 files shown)


zeebe/auth/src/main/java/io/camunda/zeebe/auth/JwtDecoder.java
```diff
@@ -0,0 +1,112 @@
...
+import com.auth0.jwt.JWT;
+import com.auth0.jwt.interfaces.Claim;
+import com.auth0.jwt.interfaces.DecodedJWT;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class JwtDecoder {
```
nicpuppa, line 19: 💭 This class should replace the identity SDK, once we decide how to verify the token. 
When we will refactor the auth type format we are going to remove the other 2 encode/decode jwt class

Ben-Sheppard:
> ❗ ❓ The token validation is NOT part of this PR. Since is not easy to implement we need to discuss more in details about it

@nicpuppa should we discuss this on Monday?

nicpuppa:
Good to me @Ben-Sheppard

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/Gateway.java
```diff
@@ -391,6 +392,7 @@ private ServerServiceDefinition applyInterceptors(final BindableService service)
       } else {
         interceptors.add(new IdentityInterceptor(identityCfg, gatewayCfg.getMultiTenancy()));
       }
+      interceptors.add(new AuthenticationInterceptor());
```
koevskinikola, line 395: ❓ We still support `AuthMode.NONE`. Do you think it makes sense to only add this interceptor if the `AuthMode` is `Identity`?

This way, the `AuthenticationInterceptor` will always be invoked, even when we want authentication to be disabled. And I think we might want this mode for local test setups.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/AuthenticationInterceptor.java
koevskinikola, line 1: 💭 I assume you created a completely new class instead of adding this code to the `IdentityInterceptor`, as we will be removing the code there completely soon.

koevskinikola (changes requested):
The code looks good @nicpuppa, nice job!

The only blocker is how we add the `AuthenticationInterceptor` in the `Gateway` class. Let's discuss this before we can merge this PR.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/AuthenticationInterceptor.java
nicpuppa: Yes, this class will be removed here https://github.com/camunda/camunda/pull/24757

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/Gateway.java
koevskinikola: Yes, so the `AuthenticationInterceptor` shouldn't be invoked at all in that case.

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/InterceptorUtil.java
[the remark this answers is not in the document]
koevskinikola: Sorry, I rambled in my comment above.

My point was that we don't have to add the new `USER_CLAIMS` property in the `InterceptorUtil` class.

The goal for the `InterceptorUtil` class was to make it easier for users to implement multi-tenancy when using Zeebe in a standalone setup. If a user still wants to implement something custom, does the `USER_CLAIMS` property help them in this, or in implementing some custom Authorization mechanism?

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/Gateway.java
koevskinikola: 👍 Yes it is, and I mixed up the if-blocks in the code there. Sorry for wasting your time 😅

koevskinikola (approved):
LGTM @nicpuppa 👏

zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/InterceptorUtil.java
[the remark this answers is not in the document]
nicpuppa: Should I remove the property from here then ?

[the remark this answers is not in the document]
koevskinikola: Yup, it would be safer to remove it.

PR #25426, opened 2024-11-28
I can activate an ad-hoc subprocess

### LINKED ISSUE #25270: I can execute a process instance with an ad-hoc subprocess
**Description**

I want to execute a process instance with an ad-hoc subprocess. When the process instance enters the ad-hoc subprocess, it should activate the subprocess and behave like a regular embedded subprocess, such as applying input mappings and invoking execution listeners. The ad-hoc subprocess should stay active until its scope is terminated.

Requires https://github.com/camunda/camunda/issues/25266.

saig0:
## Description

Adds a BPMN stream processor to execute ad-hoc subprocesses. When entering an ad-hoc subprocess, the process instance
- Applies input mappings
- Invokes start execution listeners

The ad-hoc subprocess can be terminated by a cancel command and interrupted by a boundary event. The completion is not implemented because the inner activities can't be activated yet.

## Related issues

closes #25270


ana-vinogradova-camunda (approved):
LGTM 🚀

PR #24992, opened 2024-11-21
Create common api to trigger webapps backups

entangled90:
## Description
- Moved operate BackupController in `dist/` module:
  - Refactored to use `WebEndpoint` similar to Zeebe's 
  - Refactored error handling logic in controller similar to Zeebe's `BackupController`
  - Added backward compatible `BackupControllerStandalone`
- All BackupService implementation only throw exception defined in webapps-module: `BackupServiceWrapper` is not needed anymore
- removed backup controller in tasklist so that the one in dist can be used 
- Removed BackupServiceElasticSearchTest as it contained the same tests of BackupControllerIT
- Added `snapshotTimeout` & `incompleteCheckTimeoutInSeconds` to `BackupProperties` in Tasklist with the same defaults as operate

As a follow-up, I will move BackupControllerIT into the `dist` module as it contains common tests for both operate & tasklist.

> Must be merged after #25212 
## Related issues

closes #24465

Issue #24465, opened 2024-11-07
Create common API to trigger webapps Backups


marcosgvieira:
Create common API to take backups of Optimize, Identity, Tasklist and Operate together. The implementation should be similar as what [Operate/Tasklist are current doing](https://github.com/camunda/camunda/blob/4cf854cb77ad186d2f9fd9ef2d2a8cc77996052c/operate/webapp/src/main/java/io/camunda/operate/webapp/management/BackupController.java#L38), but should include indices from all the components

### Acceptance Criteria
- [ ] POST /webapps should trigger a backup of all webapps indices together
- [ ] POST /webapps should return a list of all scheduled snapshots:
```
{
  "scheduledSnapshots": [
    "camunda_webapps_8.7.0_part_1_of_6",
    "camunda_webapps_8.7.0_part_2_of_6",
    "camunda_webapps_8.7.0_part_3_of_6",
    "camunda_webapps_8.7.0_part_4_of_6",
    "camunda_webapps_8.7.0_part_5_of_6",
    "camunda_webapps_8.7.0_part_6_of_6",
    "camunda_webapps_8.7.0_part_6_of_7"
  ]
}
```
- [ ] Below error codes should be returned (but not restricted only to this list)

Code | Description
-- | --
200 OK | Backup was successfully started, snapshots will be created asynchronously. List of snapshots is returned in the response body (see example above). This list must be persisted together with the backup id to be able to restore it later.
400 Bad Request | In case something is wrong with backupId, e.g. the same backup id already exists.
500 Server Error | All other errors, e.g. ES returned error response when attempting to create a snapshot.
502 Bad Gateway | Elasticsearch or OpenSearch is not accessible, the request can be retried when it is back.

npepinpe:
@marcosgvieira - I thought we were not harmonizing Optimize yet with the other web apps until they are also using the same indices. Is that not correct?

PR #25434, opened 2024-11-28
Merge BackupRepositoryProps based on OperateProperties & TasklistProperties

### LINKED ISSUE #25410: Merge BackupConfiguration from operate & tasklist
Operate and tasklist currently have separate backup configuration, but the backup service will be completely shared, so we cannot have two different configurations.
Considering that tasklist's backup config contains only the repository name we will do the following:
- if only one of them is provided, use that one
- if both are provided and they have the same repository one, then use operate one (which has more settings)
- if both are provided and the repository name is different, fail fast and interrupt the startup.

entangled90:
## Description
Create BackupRepositoryProps based on the information present in OperateProperties & TasklistProperties.
If both are present, their repositoryName configured must match otherwise we fail fast.
If only one is present, we use that configuration.
If none is present we can't start the application.

> Based on #24992 , wait for it before merging

## Related issues
closes #25410


webapps-backup/src/main/java/io/camunda/webapps/backup/repository/BackupRepositoryProps.java
```diff
@@ -21,7 +21,9 @@ public interface BackupRepositoryProps {
 
   Long incompleteCheckTimeoutInSeconds();
 
-  boolean includeGlobalState();
+  default boolean includeGlobalState() {
```
entangled90, line 24: This setting was true for tasklist & false for operate, I guess this should be true by default

entangled90:
It's probably better to merge this branch into #24465 as it should probably fix the Spring startup issues of that one

dist/src/main/java/io/camunda/webapps/config/BackupConfig.java
```diff
@@ -0,0 +1,92 @@
...
+        throw new IllegalArgumentException(
+            "Different repository name configured for operate & tasklist backups: if both are configured, they must point to the same repository");
+      }
+      return props(operateProperties.getVersion(), operateBackup);
+    } else if (operateProperties != null) {
+      return props(operateProperties.getVersion(), operateProperties.getBackup());
+    } else if (tasklistProperties != null) {
+      return props(tasklistProperties.getVersion(), tasklistProperties.getBackup());
+    } else {
+      return BackupRepositoryProps.EMPTY;
```
entangled90, line 47: If we throw an exception here, IT will not pass. The backup controller will check if the repository name is not null or not empty before taking any backup.

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/BackupRepositoryProps.java
```diff
@@ -29,5 +42,7 @@ static Long defaultIncompleteCheckTimeoutInSeconds() {
     return 5 * 60L;
   }
 
-  boolean includeGlobalState();
+  default boolean includeGlobalState() {
```
entangled90, line 45: this flag=true was needed by tasklist

lenaschoenburg: :question: Why make this a default though? I'd prefer no default implementation and making this explicit for both Tasklist and Operate when you return the anonymous types implementing `BackupRepositoryProps`.

:question: Your changes overall mean that now both Operate and Tasklist have this as `true` whereas before it was only true for Tasklist. Is this intended?

lenaschoenburg (approved):

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/BackupRepositoryProps.java
entangled90: Because i'm not even sure if it should be configurable. Now there is no operate vs tasklist anymore, just "webapps" indices. 
By default (in ES/OS) is true, but it was set as false in operate. So in the end we could even get rid of this setting

lenaschoenburg: Okay makes sense I guess :+1:

PR #25437, opened 2024-11-28
feat: automatically create the default tenant

emulbreh:
## Description

@remcowesterhoud I wrote this for https://github.com/camunda/camunda/pull/25048, but after https://github.com/camunda/camunda/pull/25190 it looks like it could be included in `IdentitySetupInitializeProcessor`. Is that right?


remcowesterhoud:
🤔 I suppose multi-tenancy can not be used if authentication is not enabled, right @Ben-Sheppard? If so, yes we can put it in the `IdentityInitializer`.

remcowesterhoud:
This will be tackled in https://github.com/camunda/camunda/issues/25610

PR #25441, opened 2024-11-28
pg-25290-task-template-refactor

### LINKED ISSUE #25290: Adjust schema/exporter relationships
Adjust schema relationships and exporter to be able to search user tasks by Variables.

Reference: https://miro.com/app/board/uXjVLBTPaek=/

**Acceptance Criteria**
- [ ] When having a process with process variables (not local variables) for each process variable we should have a record representing the process variable as a child of a ProcessInstance
- [ ] When having a process containing a user task with local variables, for each local variable we should have:
    - [ ] A record representing the variable as a child of UserTask
    - [ ] A record representing the variable as a child of ProcessInstance
- [ ] When having relationships, the _routing should always be the ProcessInstanceId so we ensure that all the related records are part of the same ES/OS shard.
- [ ] Ids of UserTasks should be the flowNodeInstanceKey, so it's possible to achieve relationships between UserTasks and Variables
- [ ] Implementation should be done only for ZeebeUserTasks (V2)

panagiotisgts:
## Description

Refactor the TaskTemplate and it's relevant handlers to the new structure proposal.

- Route everything to `processInstanceId`
- Generate two variable entities for task variables, one linked to Process and one to the User Task
- Task local variables suffixed with `-local` to be able to be persisted twice
## Checklist

- [ ] for CI changes:

## Related issues

closes #25290

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/exporter/UserTaskExporterIT.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/exporter/UserTaskExporterIT.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 import org.awaitility.Awaitility;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.TestTemplate;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -119,6 +120,8 @@ public class UserTaskExporterIT {
   @Nested
   class UserTaskVariableIT {
 
+    // TODO Re-enable
+    @Disabled("Should be re enabled when Search clients are aligned with schema changes")
     @TestTemplate
     void shouldExportUserTaskVariable(final TestStandaloneBroker testBroker) {
       final var client = testBroker.newClientBuilder().build();
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
@@ -114,7 +114,7 @@ public class UserTaskCompletionVariableHandler
 
     final TaskJoinRelationship joinRelationship = new TaskJoinRelationship();
     joinRelationship.setParent(entity.getScopeKey());
-    joinRelationship.setName(TaskJoinRelationshipType.TASK_VARIABLE.getType());
+    joinRelationship.setName(TaskJoinRelationshipType.LOCAL_VARIABLE.getType());
     entity.setJoin(joinRelationship);
   }
 
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
@@ -128,7 +128,11 @@ public class UserTaskCompletionVariableHandler
     updateFields.put(TaskTemplate.JOIN_FIELD_NAME, entity.getJoin());
 
     batchRequest.upsertWithRouting(
-        indexName, entity.getId(), entity, updateFields, String.valueOf(entity.getScopeKey()));
+        indexName,
+        entity.getId(),
+        entity,
+        updateFields,
+        String.valueOf(entity.getProcessInstanceId()));
   }
 
   @Override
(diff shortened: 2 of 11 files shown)


qa/integration-tests/src/test/java/io/camunda/it/exporter/UserTaskExporterIT.java
```diff
@@ -119,6 +120,8 @@ private void waitForTask(
   @Nested
   class UserTaskVariableIT {
 
+    // TODO Re-enable
+    @Disabled("Should be re enabled when Search clients are aligned with schema changes")
```
ChrisKujawa, line 124: ❌ I think we should create an issue for this, to keep track of it.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskVariableHandler.java
```diff
@@ -54,8 +53,18 @@ public boolean handlesRecord(final Record<VariableRecordValue> record) {
...
   public List<String> generateIds(final Record<VariableRecordValue> record) {
-    return List.of(
-        ID_PATTERN.formatted(record.getValue().getScopeKey(), record.getValue().getName()));
+    final String id =
+        ID_PATTERN.formatted(record.getValue().getScopeKey(), record.getValue().getName());
+
+    /* Process Variable */
+    if (record.getValue().getScopeKey() == record.getValue().getProcessInstanceKey()) {
+      return List.of(id);
+    }
```
ChrisKujawa, line 62: 🤔 Not sure whether this is correct.

If the scope is a sub-process, this will create a local variable. But we want only local variables for user tasks or? 🤔

ChrisKujawa (commented):
There is some unclearity to me regarding the local variable. I also asked @romansmirnov @marcosgvieira that one of them looks at it

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskVariableHandler.java
panagiotisgts: That is true, I've tested a sub-process scenario and it generates local variables as well, good catch! I'll have to adjust this.

romansmirnov: The problem is that at this moment, it is not known if the scope is a subprocess or not. So, while I would agree to avoid unnecessary duplications, for now, there is no such information (quickly) available to filter by. So, IMO, that's actually expected and okay.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
```diff
@@ -31,7 +31,7 @@ public class UserTaskCompletionVariableHandler
   private static final Logger LOGGER =
       LoggerFactory.getLogger(UserTaskCompletionVariableHandler.class);
 
-  private static final String ID_PATTERN = "%s-%s";
+  private static final String ID_PATTERN = "%s-%s" + TaskTemplate.LOCAL_VARIABLE_SUFFIX;
```
romansmirnov, line 34: ❓ I don't understand why this handler is required? Snapshotting of variables happens in v1, but we no longer will do this with v2. IMO, this handler can be removed. (If we come to that conclusion, we should do this in a separate PR and it does not block this PR.)

romansmirnov (approved):
❓ Is my understanding correct, that handling of old (8.6) User Tasks will be done in a separate PR?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
romansmirnov: After some consideration, this handler must be rewritten to support the Tasklist v1 API. I will do this separately. This is not necessary for the v2 API.

panagiotisgts:
> ❓ Is my understanding correct, that handling of old (8.6) User Tasks will be done in a separate PR?

Yes @romansmirnov, the scope of this PR is just to handle the required changes in the Exporter to enable all other issues  regarding this topic to be unblocked.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java
panagiotisgts: @romansmirnov I believe we still need this snapshot of User Task completion variables for Form inputs and variables submitted via the User Task. These variables should maintain the state that they had when they were initially submitted by the end-user.

PR #25446, opened 2024-11-29
fix: various fixes to enable searching for mappings

Ben-Sheppard:
## Description

As part of a separate piece of work I noticed that there were some things missing related to the mapping entity, this PR adds in the missing classes such that mappings can be retrieved from ES/OS.

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/MappingFilterTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/MappingFilterTransformer.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.search.clients.transformers.filter;
+
+import static io.camunda.search.clients.query.SearchQueryBuilders.and;
+import static io.camunda.search.clients.query.SearchQueryBuilders.stringTerms;
+import static io.camunda.search.clients.query.SearchQueryBuilders.term;
+import static io.camunda.webapps.schema.descriptors.usermanagement.index.MappingIndex.CLAIM_NAME;
+import static io.camunda.webapps.schema.descriptors.usermanagement.index.MappingIndex.CLAIM_VALUE;
+
+import io.camunda.search.clients.query.SearchQuery;
+import io.camunda.search.filter.MappingFilter;
+import java.util.List;
+
+public class MappingFilterTransformer implements FilterTransformer<MappingFilter> {
+  @Override
+  public SearchQuery toSearchQuery(final MappingFilter filter) {
+    return and(
+        stringTerms(CLAIM_NAME, filter.claimNames()),
+        filter.claimName() == null ? null : term(CLAIM_NAME, filter.claimName()),
+        filter.claimValue() == null ? null : term(CLAIM_VALUE, filter.claimValue()));
+  }
+
+  @Override
+  public List<String> toIndices(final MappingFilter filter) {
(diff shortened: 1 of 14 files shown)


search/search-domain/src/main/java/io/camunda/search/filter/MappingFilter.java
```diff
@@ -7,13 +7,18 @@
...
 package io.camunda.search.filter;
 
+import static io.camunda.util.CollectionUtil.addValuesToList;
+
 import io.camunda.util.ObjectBuilder;
+import java.util.List;
 
-public record MappingFilter(Long mappingKey, String claimName, String claimValue)
+public record MappingFilter(
+    Long mappingKey, String claimName, List<String> claimNames, String claimValue)
```
emulbreh, line 16: Can we replace the singular `claimName` filter with a single `claimNames` list filter? The API would be a bit simpler if there's only one way to express a filter.

search/search-client-query-transformer/src/test/java/io/camunda/search/clients/transformers/filter/MappingQueryTransformerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  public static Stream<Arguments> queryFilterParameters() {
+    return Stream.of(
+        Arguments.of(
+            (Function<Builder, ObjectBuilder<MappingFilter>>)
+                f -> f.claimNames(List.of("foo", "bar")),
+            "claimName",
+            List.of("foo", "bar")),
```
emulbreh, line 63: If we construct a full `SearchQuery` as the expected value here we can get rid of the instanceof checks and also make it a bit more obvious what we're actually testing. 
```
SearchQuery.of(q -> q.terms(t -> t.stringTerms(List.of("foo", "bar")))
```

search/search-domain/src/main/java/io/camunda/search/filter/MappingFilter.java
Ben-Sheppard: We could but for now I'd prefer to keep it this way, its a pattern that we use consistently in other filters where there could be one or multiple keys i.e. AuthorizationFilter.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/MappingFilterTransformer.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.camunda.search.clients.query.SearchQuery;
+import io.camunda.search.filter.MappingFilter;
+import java.util.List;
+
+public class MappingFilterTransformer implements FilterTransformer<MappingFilter> {
+  @Override
+  public SearchQuery toSearchQuery(final MappingFilter filter) {
+    return and(
+        stringTerms(CLAIM_NAME, filter.claimNames()),
```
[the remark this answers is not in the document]
Ben-Sheppard: The `stringTerms` method here caters for null, if `filter.claimNames()` is null then no filters will be added :)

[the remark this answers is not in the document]
emulbreh: Ah, that's interesting, maybe we could do the same for the single term filter so all these null checks can go away.

search/search-client-query-transformer/src/test/java/io/camunda/search/clients/transformers/filter/MappingQueryTransformerTest.java
Ben-Sheppard: Great suggestion, I've changed this now

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/MappingFilterTransformer.java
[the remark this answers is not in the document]
Ben-Sheppard: Yea that could be an idea, though I'd defer to another PR for that as it has wider considerations I think :)

emulbreh (approved):

Oleksiivanov (approved):

PR #25456, opened 2024-11-29
fix: make user role management ui work

emulbreh:
## Description

* replace ids with keys
* use actual api endpoints

This was split off from https://github.com/camunda/camunda/pull/24609

### DIFF
--- a/identity/client/src/pages/users/detail/role/AddModal.tsx
+++ b/identity/client/src/pages/users/detail/role/AddModal.tsx
@@ -27,11 +27,12 @@ const AddModal: FC<UseEntityModalCustomProps<User, { userRoles: Role[] }>> = ({
   const [showSelectRoleError, setShowSelectRoleError] = useState(false);
 
   const {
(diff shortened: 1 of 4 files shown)


Ben-Sheppard (approved):
Thank you for moving this to a separate PR, this looks reasonable to me.

One question is around tests, I'm not sure the client has tests at the moment? If it doesn't then don't worry about it in this PR but if it does are there any existing test classes that we can add based on these changes?

emulbreh:
> Thank you for moving this to a separate PR, this looks reasonable to me.
> 
> One question is around tests, I'm not sure the client has tests at the moment? If it doesn't then don't worry about it in this PR but if it does are there any existing test classes that we can add based on these changes?

There are no unit tests right now: https://github.com/camunda/camunda/blob/main/identity/client/package.json#L13
I believe it's not the right time to start writing any at the moment. Once we start combining the webapps we'll likely refactor most of the identity UI to use shared components.

PR #25463, opened 2024-11-29
CI: add custom GHA best practices rules & linter

### LINKED ISSUE #25306: CI: add tool to lint for GHA best practices
## Description

As observed e.g. in https://github.com/camunda/camunda/pull/24989 or https://github.com/camunda/camunda/issues/21766 there are GHA best practices (having `timeout-minutes` on every job, having CI Health instrumentation in `ci.yml`) that https://github.com/rhysd/actionlint doesn't detect nor warn about.

It would be nice to have a tool (best case an existing one) to check for these configurable & extensible set of best practices against our set of GHA pipelines in the monorepo. Those should be automatically run on any GHA file change.

## Hint

Could also consider watching for discouraged GHA secrets usage or following GHA caching strategy.

❌ Check out https://stelligent.github.io/config-lint/ (Update: project last maintained 4 years ago, fails on YAML keys with hyphens like `timeout-minutes`)
✔️ Check out https://www.conftest.dev/ which is also [used by Infra team to lint K8s YAML files](https://github.com/camunda/infra-core/blob/stage/.lint/conftest/best-practices/README.md)

cmur2:
## Description

This PR aims to introduce additional linting rules for GHA workflows on top of what `actionlint` has to solve https://github.com/camunda/camunda/issues/25306. The new rules are opinionated and specifically tailored for the C8 monorepo CI and its CI Health observability framework (thus no standard tool has those rules).

MUST rules will fail the build, SHOULD rules are warnings but allow merging.

The new rules introduced are:

* all workflows MUST specify `name`
    * otherwise other rules relying on the name property might not work correctly
* all workflows MUST specify `on.push.branches` when using the `on.push` trigger
    * to make sure they only run on a small selected set of branches, e.g. `main` and `stable/*` - for PRs we use `on.pull_request` since that allows change detection and has more valuable CI health metrics (using target branch)
* all Unified CI related workflow jobs MUST specify `timeout-minutes`
* all Unified CI related workflow jobs SHOULD specify `timeout-minutes` with value of at most 15 minutes
* all Unified CI related workflow jobs MUST send CI Health metrics

This PR extends the `ci.yml` `actionlint` job, codifies the above rules as [OPA Rego policies](https://www.openpolicyagent.org/docs/latest/policy-language/) (good experience from Infra team) and uses `conftest` to verify them automatically.

This new framework allows adding more rules over time, extending scope and tightening rules from SHOULD to MUST if appropriate.

## Testing

Install https://www.conftest.dev/ locally and run e.g. `conftest test --policy .github/conftest-*.rego .github/workflows/*.yml` to execute the same rules on the same files as the new GHA step will do.

## Checklist

- [x] for CI changes:

## Related issues

closes #25306

### DIFF
--- a/.github/actionlint-matcher.json
+++ b/.github/actionlint-matcher.json
@@ -1,17 +1,17 @@
 {
-    "problemMatcher": [
-      {
-        "owner": "actionlint",
-        "pattern": [
-          {
-            "regexp": "^(?:\\x1b\\[\\d+m)?(.+?)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*: (?:\\x1b\\[\\d+m)*(.+?)(?:\\x1b\\[\\d+m)* \\[(.+?)\\]$",
-            "file": 1,
-            "line": 2,
-            "column": 3,
-            "message": 4,
-            "code": 5
-          }
-        ]
-      }
-    ]
-  }
+  "problemMatcher": [
+    {
+      "owner": "actionlint",
+      "pattern": [
+        {
+          "regexp": "^(?:\\x1b\\[\\d+m)?(.+?)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*: (?:\\x1b\\[\\d+m)*(.+?)(?:\\x1b\\[\\d+m)* \\[(.+?)\\]$",
+          "file": 1,
+          "line": 2,
+          "column": 3,
+          "message": 4,
+          "code": 5
+        }
+      ]
+    }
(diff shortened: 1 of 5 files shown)


clementnero (approved):
This custom approach is really powerful for the monorepo use cases to warn or enforce common rules. LGTM 🚀 

A non-blocking comment: if you need to bring warnings to the user's attention, creating an annotation could be nice, e.g. https://github.com/camunda/infra-global-github-actions/blob/main/rerun-failed-run/action.yml#L105

cmur2:
> if you need to bring warnings to the user's attention, creating an annotation could be nice

Good point, I also considered it but then didn't find any existing [problem matchers](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) for `contest` so neglected the idea but your input reaffirms that this is important.

It shouldn't be too much effort to create such a matcher myself given the good docs and [what we use for actionlint](https://github.com/camunda/camunda/blob/main/.github/actionlint-matcher.json).

cmur2:
> It shouldn't be too much effort to create such a matcher myself given the good docs and [what we use for actionlint](https://github.com/camunda/camunda/blob/main/.github/actionlint-matcher.json).

@clementnero I found an even easier way since `conftest` supports `github` image that includes those annotations automatically:

image

Issue #25424, opened 2024-11-28
[EPIC] Return key attributes as string values in the C8 REST API by default


tmetzke:
## Introduction

We change the C8 REST API default to `string`-based key attributes for content type `application/json` as proposed in https://github.com/camunda/camunda/issues/24805. This is a **breaking change** in the V2 API.

## Concept

* We introduce dedicated content types and DTOs to use in requests/responses to accept/return `int64` number-based or `string`-based key attributes for clients that need this. Upgrading clients can set the number content type if they need the old behavior.
* The content types are reflected at least in the 8.6 V2 GA endpoints (aka migrated Zeebe gRPC API) in the OpenAPI spec and reference documentation.
* Internally, in the REST API, key attributes remain as `long` values.
* The REST gateway transforms `long` values into `string` values or vice versa for requests and responses by default. Ideally, this happens generically in a Jackson (de-)serializer to avoid unnecessary code duplication in the controllers.
* We detect supported pre-8.7 clients via agent header and return/accept number-based key attributes if `application/json` is provided as the content type. 8.7 clients define the content type they need explicitly.
* The 8.6 V2 API receives the new content types as well to allow fixing existing client applications that need `string´-based keys. We don't change the default there to avoid breaking changes in a patch.

## Task Breakdown

```[tasklist]
### Gateway API 8.6 & 8.7
- [x] Add `string` and `int64`-based DTOs
- [x] Transform key attributes based on content type
- [x] Detect pre-8.7 clients via agent header
- [ ] https://github.com/camunda/camunda/issues/25760
- [ ] https://github.com/camunda/camunda/issues/26063
- [ ] https://github.com/camunda/camunda/issues/26653
- [ ] https://github.com/camunda/camunda/issues/26761
- [ ] https://github.com/camunda/camunda/issues/26251
```

```[tasklist]
### Gateway API 8.7
- [x] Make `string` keys the default
- [x] Decide on support for new content types for new 8.7 endpoints
- [ ] https://github.com/camunda/camunda/issues/26062
- [ ] https://github.com/camunda/camunda/issues/26465
- [ ] https://github.com/camunda/camunda/issues/27204
```

```[tasklist]
### Java SDK 8.7
- [x] Use string-based content type in Java SDK
```

```[tasklist]
### Documentation
- [x] Document key attribute changes in update guide
- [x] Document different content types and their purpose in OpenAPI spec and reference docs
- [x] Adjust 8.6 public reference specification
[...]

PR #25464, opened 2024-11-29
feat: introduce custom REST DTOs with string key attributes

tmetzke:
## Description

* Introduces custom media types for select endpoints to showcase media type negotiation
  * Job activation - collect results with string keys in REST response observer
  * User task search - mixed request and response types (request with plain, receive with custom)
  * Tenant fetching - accept results with string keys only

Hints:
* The job activation adjustments are pretty pragmatic and POC-style. For production code, we should look into refactoring this into a more generalized approach that doesn't copy that many classes and code.
* Content types are allowed to be mixed, e.g. send a request with plain json and expect a response with a specific type (even the "other" format with keys as strings) back
* This does not make keys as strings the default yet for simplicity (a lot of tests have to be adjusted for this to reflect the new default behavior).

## Checklist

- [ ] for CI changes:

## Related issues

related to #25424

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -102,6 +102,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/JobActivationResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/JobActivationResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/JobActivationResponseStringKeys"
         "400":
           description: >
             The provided data is not valid.
@@ -444,6 +450,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/TenantItem"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/TenantItem"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/TenantItemStringKeys"
         "400":
           description: The provided data is not valid.
           content:
@@ -850,6 +862,12 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/UserTaskSearchQueryRequest"
+          application/vnd.camunda.api.keys.number+json:
+            schema:
+              $ref: "#/components/schemas/UserTaskSearchQueryRequest"
+          application/vnd.camunda.api.keys.string+json:
+            schema:
+              $ref: "#/components/schemas/UserTaskSearchQueryRequestStringKeys"
       responses:
         "200":
           description: >
@@ -858,6 +876,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/UserTaskSearchQueryResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/UserTaskSearchQueryResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/UserTaskSearchQueryResponseStringKeys"
         "400":
           description: >
             The user task search query failed.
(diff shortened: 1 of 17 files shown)


tmetzke:
@danielkelemen, FYI, this is the POC.
Job controller tests are expected to fail, I didn't give them toooo much attention, just made the basic concept work.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3133,6 +3157,25 @@ components:
             type: integer
             format: int64
 
+    TenantItemStringKeys:
```
emulbreh, line 3160: Is it really worth it to duplicate all these schemas? We could avoid a lot of extra code by only documenting strings and adding a comment and annotation on each key property. The comment would point users to relevant documentation, the annotation (`x-class-extra-annotation`) would be used to pick a serializer.

tmetzke: I'm all with you on avoiding duplication. I'm not a huge fan either. However, regarding API correctness and clarity, I believe the proposed alternative will harm us more than help us in the long run. Hiding this behind documentation and returning a type different than the specified will lead to confusion.

We can try to keep the duplication to a minimum, I've tried to depict this by introducing base types that cover all non-key attributes. Ultimately, I believe we should be transparent and clear with the different results the API can deliver.

emulbreh: Aren't we going to end up with tens of thousands of additional lines of code? I fear this is going to be hard to maintain and evolve.

tmetzke: We're going to end up with more code for sure. The extent of that depends on how well we (can) reuse code (OpenAPI using `allOf` and common types, Java code will definitely have some duplications for the DTOs). We can also try to come up with a more generic solution in Java while keeping the explicit media type mapping.

While I totally understand the concern of maintainability and evolvability, it doesn't trump the API correctness and clarity in my regard. @danielkelemen, this can also be a special point to investigate when building the "real" thing:
- How can we keep the code duplication to a minimum (OpenAPI and Java)?
- Is there a way to more generically build the string-based DTOs in Java so we don't have to duplicate mapper methods for the affected DTOs?

emulbreh: Going forward, string keys will be the correct (and eventually only) way to use the API. For anyone getting started with the v2 API after 8.7 the numeric keys just clutter the documentation. I see the support for numeric keys as backwards compatibility mode for early adopters of the v2 API. We can point out in the release notes that the API is only backwards compatible with the new header. Our OpenAPI description could use string keys only and we provide a mechanically translated version with numeric keys for early adopters who use generated API clients.

tmetzke: 💭 I gave this some more thought and I also played around with our [API reference docs](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/camunda-8-rest-api/) a bit.

I believe the biggest downside to the custom header and the two API documents is discoverability. It is hidden in another specification document of the one API we provide. The API should describe what it does and be complete.

This is the biggest plus of this proposal:
* There is clear documentation on the data the endpoints in our API can provide and there is exactly one documentation. Yes, this means the API contains additional information on types many users won't need but it's nicely handled through tabs in the right places and the correct default behavior. So the cluttering part I would consider negligible regarding the reference documentation. The OpenAPI spec will contain more data but I doubt many users will scan this document manually. Generated clients will receive new DTOs that can be used for number-based cases if needed and otherwise be ignored.
* We can initially scope to support the new content type only in the 8.6 V2 GA endpoints (migrated Zeebe gRPC API). Thus, the effort stays manageable for 8.7.0. We can advise users to use `Accept: application/vnd.camunda.api.keys.number+json, application/json` globally to receive number-based responses if available and otherwise accept plain JSON. This way, not all endpoints have to produce all three media types, for example. If there is enough user demand, we could still add support for the custom media types to all endpoints.
* We can document the custom media types as deprecated in 8.7 already and remove them with 8.9 so users have enough time to migrate. We can remove code overhead again with 8.9 then.

There is downsides to this, and I see them, I still think this is the clearest and most pragmatic way forward that is also discoverable for users. And easier to explain.

danielkelemen: What we discussed:
- API specification needs to contain these references in order for the documentation to show both content types with examples.
- Controllers should already use string keys and hopefully with 8.9 we will be able to drop the whole conversion and just simply return the DTOs.
- I'll try to see if any Jackson abstraction is possible to do the conversion so we don't need to duplicate the controller endpoints in code.

Issue #24665, opened 2024-11-13
Migrate existing Tasklist v1 API to the harmonized index schema


romansmirnov:
### Description

Ensure that existing (public) [Tasklist v1 API](https://docs.camunda.io/docs/next/apis-tools/tasklist-api-rest/tasklist-api-rest-overview/) works with the harmonized index schema.

```[tasklist]
### Exporter
- [ ] https://github.com/camunda/camunda/pull/24434
- [ ] https://github.com/camunda/camunda/pull/24504
- [ ] https://github.com/camunda/camunda/pull/24520
- [ ] https://github.com/camunda/camunda/pull/24631
- [ ] https://github.com/camunda/camunda/pull/25682
```

```[tasklist]
### Data Access
- [ ] https://github.com/camunda/camunda/pull/24753
- [ ] https://github.com/camunda/camunda/pull/24783
- [ ] https://github.com/camunda/camunda/pull/24835
- [ ] https://github.com/camunda/camunda/pull/24860
- [ ] https://github.com/camunda/camunda/pull/24872
- [ ] https://github.com/camunda/camunda/pull/25386
- [ ] https://github.com/camunda/camunda/pull/24914
- [ ] https://github.com/camunda/camunda/pull/25379
- [x] Delete Web Sessions (Handled with https://github.com/camunda/camunda/pull/25220)
- [ ] https://github.com/camunda/camunda/pull/25468
- [ ] https://github.com/camunda/camunda/pull/25458
- [ ] https://github.com/camunda/camunda/pull/25432
- [ ] https://github.com/camunda/camunda/pull/25461
- [ ] https://github.com/camunda/camunda/pull/25472
```

```[tasklist]
### Archiver
- [ ] https://github.com/camunda/camunda/pull/25471
- [ ] https://github.com/camunda/camunda/pull/25470
```

```[tasklist]
### Migration
- [ ] https://github.com/camunda/camunda/issues/26065
```

```[tasklist]
### Chore
- [ ] https://github.com/camunda/camunda/pull/25576
- [ ] https://github.com/camunda/camunda/pull/25578
- [ ] https://github.com/camunda/camunda/pull/25799
- [ ] https://github.com/camunda/camunda/pull/26157
- [ ] https://github.com/camunda/camunda/pull/26088
- [ ] https://github.com/camunda/camunda/issues/25985
- [ ] https://github.com/camunda/camunda/issues/28122
[...]

PR #25472, opened 2024-11-29
refactor: replace Tasklist's Process Entity and Index

romansmirnov:
## Description

* Ensure that the old Tasklist Importer (8.6.0) continues to import to the "old" Tasklist Process Index
* Ensure that the old Tasklist Importer (8.7.0) imports to the harmonized Process Index (so that Tasklist ITs continues to run)
* Separate between `operateProcessIndex` and `tasklistProcessIndex` beans to ensure that the old backend gets the expected beans autowired with the expected configurations

## Checklist

- [ ] for CI changes:

## Related issues

related to #24665


romansmirnov:
@panagiotisgts, after checking the `process-migration` runner, I think the (deprecated) `TasklistProcessIndex` - which is used by the 8.6 import processor - is actually not required. Could you confirm? If so, I could delete it.

panagiotisgts:
@romansmirnov Correct, from 8.7 Tasklist will be based on the `operate-process` index and importing will be handled by Operate's 8.6 process importer.

romansmirnov:
@panagiotisgts, thanks! Okay, then I will delete that `TasklistProcessIndex` - I somehow assumed it must continue to write in the old `tasklist-process` index to ensure the migration can happen. But I got it now. I will quickly adjust it.

panagiotisgts:
The initial approach was to use the `tasklist-process` index to read records from, but since the two indices contain the same records, we decided to do the migration on the `operate-process` index alone and get rid of the Tasklist one.

romansmirnov:
Deleted the `TasklistProcessIndex` and the entity in an additional commit. Locally, everything works. Let's see what the CI tells 😄

panagiotisgts (approved):
LGTM. I've just rerun the Operate E2E for sanity check, even though it has been flaky of late.

PR #25473, opened 2024-12-01
fix: various fixes to enable exporting and searching tenants

Ben-Sheppard:
## Description

I noticed as part of running the app locally that som errors came up, this PR adds in fixes to that and I can confirm that locally I can create/search for tenants. Notably however:

- We need a `key` field on the index otherwise the search classes fail as its expected there is a `key` field
- We need to declare `memberKey` in the index as its on the core entity we use and if this field is not on the index, the batch fails because the mappings are set to strict

### DIFF
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/sql/columns/TenantSearchColumn.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/sql/columns/TenantSearchColumn.java
@@ -11,7 +11,7 @@ import io.camunda.search.entities.TenantEntity;
 import java.util.function.Function;
 
 public enum TenantSearchColumn implements SearchColumn<TenantEntity> {
-  TENANT_KEY("tenantKey", TenantEntity::tenantKey),
+  TENANT_KEY("key", TenantEntity::key),
   TENANT_ID("tenantId", TenantEntity::tenantId),
   NAME("name", TenantEntity::name);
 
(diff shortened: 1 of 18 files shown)


emulbreh (approved):

Oleksiivanov (approved):

stoerti:
A bit too late now, but @houssain-barouni renamed all key properties in all the entities to ${EntityName}key (tenantKey, userKey, roleKey...) to be compliant with the respective filter objects and use the same naming.

Changing it back now to just "key" just for the Tenant feels a bit off now. Especially making changes in the API when the reason is the name of an index field in ES. SHouldn't there be an intermediate-layer for this technical abstraction?

houssain-barouni:
> A bit too late now, but @houssain-barouni renamed all key properties in all the entities to ${EntityName}key (tenantKey, userKey, roleKey...) to be compliant with the respective filter objects and use the same naming.
> 
> Changing it back now to just "key" just for the Tenant feels a bit off now. Especially making changes in the API when the reason is the name of an index field in ES. SHouldn't there be an intermediate-layer for this technical abstraction?

+1
> We need a key field on the index otherwise the search classes fail as its expected there is a key field

We can have the `key` field in the index model represented by https://github.com/camunda/camunda/blob/main/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/TenantEntity.java
Then, map it to `tenantKey` using the transformers. I think it was already the case before these changes 🤔

PR #25483, opened 2024-12-02
feat: move operate Spring backup module into dist

### LINKED ISSUE #24869: Combine operate `BackupServiceImpl` and tasklists `BackupManagerElasticSearch` in a single implementation in `webapp-backup`

entangled90:
## Description
Spring components for BackupService & ES moved to dist, they will be used by tasklist as well

- [X] Wait for PR #24992 
## Related issues

closes #24869


dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
```diff
@@ -60,6 +61,18 @@ && validRepoName(tasklistBackup.getRepositoryName())
...
   }
 
+  @Bean("backupThreadPoolExecutor")
+  public ThreadPoolTaskExecutor getTaskExecutor() {
+    final ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+    executor.setCorePoolSize(1);
+    executor.setMaxPoolSize(1);
+    executor.setThreadNamePrefix("webapps_backup_");
+    // TODO why just 6?
+    executor.setQueueCapacity(6);
```
entangled90, line 71: Should we leave 6? not sure why such a small number has been selected for a queue

dist/src/main/java/io/camunda/application/commons/backup/BackupComponent.java
```diff
@@ -19,11 +19,13 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.stereotype.Component;
 
 @Component
 @Configuration
+@Profile("operate | tasklist")
 public class BackupComponent {
```
npepinpe, line 29: :thought_balloon: A little confusing, since it's more like `HistoryBackupComponent` or something :shrug:

dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
npepinpe: It's a very good question, I have no idea. Let's ask in the `-tech` channel.

dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOperateBackupRepository.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+@ConditionalOnProperty(
+    prefix = "camunda.database",
+    name = "type",
+    havingValue = "elasticsearch",
+    matchIfMissing = true)
+@Configuration
+@Profile("operate")
```
npepinpe, line 28: :question: Why not for standalone? Is it because we use Operate's for both when not standalone? Shouldn't we name it `ElasticsearchBackupRepository` then?

npepinpe (approved):
No blocker, just some questions.

dist/src/main/java/io/camunda/application/commons/backup/BackupComponent.java
entangled90: I have also changed the annotation in my local branch already

dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOperateBackupRepository.java
entangled90: Yes that's the reason. I will rename it.

PR #25486, opened 2024-12-02
fix: export VARIABLE_DOCUMENT records

sdorokhova:
## Description

In order to mark variable update operations as completed, we need to export `VARIABLE_DOCUMENT` records.

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -21,6 +21,7 @@ import static io.camunda.zeebe.protocol.record.ValueType.TENANT;
 import static io.camunda.zeebe.protocol.record.ValueType.USER;
 import static io.camunda.zeebe.protocol.record.ValueType.USER_TASK;
 import static io.camunda.zeebe.protocol.record.ValueType.VARIABLE;
+import static io.camunda.zeebe.protocol.record.ValueType.VARIABLE_DOCUMENT;
 
 import io.camunda.exporter.adapters.ClientAdapter;
 import io.camunda.exporter.config.ConfigValidator;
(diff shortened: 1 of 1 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -240,6 +241,7 @@ private record CamundaExporterRecordFilter() implements RecordFilter {
             PROCESS_INSTANCE,
             ROLE,
             VARIABLE,
+            VARIABLE_DOCUMENT,
```
mustafadagher, line 244: 🔧 I think we need to add `PROCESS_MESSAGE_SUBSCRIPTION` as well, right?

sdorokhova: Yes, you're right!
I will still merge this one, as people are complaining on variable update operations and will create another PR.

sdorokhova: Ok, I can't merge as it's not approved. Let me add another type now)

ChrisKujawa: Are we missing test coverage, as we haven;t seen this before? Can we add a test for it? 🤔 Or will we cover this when we migrate the existing IT to the exporter?

mustafadagher: We don't have test coverage for it indeed. The filter is only added to the exporter context and then used in the exporter container and the exporter director. So it's one level up.

We might indeed want to cover it with tests anyway, but it's hard to think how right now.

mustafadagher (approved):

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
sdorokhova: VARIABLE_DOCUMENT and PROCESS_MESSAGE_SUBSCRIPTION will be covered by migrated IT tests!
But I also found GROUP and MAPPING types, that were missing. Not sure whether we have any IT tests for those 🤔

PR #25498, opened 2024-12-02
feat: add limit/from pagination tests

### LINKED ISSUE #23722: Add integration tests for from/limit paging method
### Description

Currently paging is tested for client api with searchBefore/searchAfter approach.
Integration tests for from/limit are needed for query endpoints

### TODO

- [x] Process definition
- [x] Process instance
- [x] Decision definition
- [x] Decision instance
- [x] Decision requirements
- [x] Flow node instance
- [x] Incident
- [x] Variable

nathansandi:
## Description

Add Integration tests for limit/from pagination

## Related issues

closes https://github.com/camunda/camunda/issues/23722

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/IncidentQueryTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/IncidentQueryTest.java
@@ -23,8 +23,10 @@ import io.camunda.zeebe.client.api.search.response.Incident;
 import io.camunda.zeebe.qa.util.junit.ZeebeIntegration;
 import io.camunda.zeebe.qa.util.junit.ZeebeIntegration.TestZeebe;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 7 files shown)


qa/integration-tests/src/test/java/io/camunda/it/client/IncidentQueryTest.java
```diff
@@ -406,20 +410,30 @@ void shouldSortByState() {
         .containsExactlyElementsOf(sortedDesc);
   }
 
+  // TO-DO: Verify the following test case: the getJobKey is returning null, which is not really
```
nathansandi, line 413: @ralfpuchert  is there any reason why job key comes null here? 
Before the integration test was working because was only one element, but after I add 2 new incidents, as the job key was null, it was throwing error here, so I adjust this test in order to have null check.

nathansandi: For visibility:

Update: Me and @ralfpuchert  discussed in a call, we can leave this test as is for now, it could be related (or not) to a delay on import. Ralf will check it separately, and it will be worked in a different PR/issue.

ralfpuchert (approved):
Thank you @nathansandi for flagging the issue with `jobKey`.
I will play around with the marked test case and decide if we need a follow up issue.

PR #25513, opened 2024-12-02
Implement getActiveIncidentsByTreePath for ES

npepinpe:
## Description

This PR implements the `getActiveIncidentsByTreePath` for the `ElasticsearchIncidentUpdateRepository`.

I refactored the scrolling logic to be a little more generic, since we want to scroll and deserialize now, as opposed to the previous implementation where we only cared about the index and ID.

I don't particularly like it - it's a little too verbose/cumbersome, but I will likely have to refactor it again for the next method (`getProcessInstances`) where we also scroll, so maybe we tackle that then.

## Related issues

related to #24930

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
@@ -574,7 +574,7 @@ abstract sealed class IncidentUpdateRepositoryIT {
       assertThat(terms)
           .succeedsWithin(REQUEST_TIMEOUT)
           .asInstanceOf(InstanceOfAssertFactories.list(String.class))
-          .containsExactly(
+          .containsExactlyInAnyOrder(
               "PI_1",
               "PI_1/FN_call",
               "PI_1/FN_call/FNI_2",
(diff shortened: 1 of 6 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
```diff
@@ -775,15 +776,15 @@ void shouldGetFlowNodes() throws PersistenceException {
       batchRequest.executeWithRefresh();
 
       // when
-      final var flowNodes = repository.getFlowNodesInListView(List.of("1", "2"));
+      final var flowNodes = repository.getFlowNodeInstances(List.of("1", "2"));
```
npepinpe, line 779: :clown_face: Looks like we made a mistake in a previous PR, and we were not testing the right thing - or rather, the production code was picking it up from the wrong index :smile:

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateRepositoryIT.java
```diff
@@ -798,13 +799,106 @@ void shouldNotGetFlowNodesWithOtherKeys() throws PersistenceException {
...
+          "8",
+          newIncident(8).setState(IncidentState.ACTIVE).setTreePath("PI_3/FNI_4"));
+      batchRequest.addWithId(
+          incidentTemplate.getFullQualifiedName(),
+          "9",
+          newIncident(9).setState(IncidentState.ACTIVE).setTreePath("PI_5/FNI_6"));
+      batchRequest.executeWithRefresh();
+
+      // when
+      final var incidents = repository.getActiveIncidentsByTreePaths(List.of("PI_1", "PI_3"));
```
entangled90, line 894: :question:  All the TreePaths used as arguments have only depth equal to one, so I would use something like this:
```List.of("PI_1", "PI_3/FNI_4", "PI_5/FNI_6/FN_4")```
The last TreePath also tests deeper TreePaths than the incident

entangled90 (approved):
:rocket:  I just have a minor suggestion for tests :+1:

PR #25515, opened 2024-12-02
Implement getProcessInstances for ES

### LINKED ISSUE #24930: Add ES incident update repository implementation

npepinpe:
## Description

Implements `getProcessInstances` for the `ElasticsearchIncidentUpdateRepository`. One difference from the original implementation here, is that I also check the join relation property - the original only did for the FNIs, but I figured it's risky to rely on the ID always being correct.

> [!Note]
> With the additional tests to the clear scroll stuff, and the changes to fix it, I'm now thinking a good follow up would be to extract that to a utility you can test - that way, we don't need to keep adding new tests for each scrolling method using it. The fact it scrolls is pretty deep implementation detail imo.

## Related issues

closes #24930

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
@@ -29,13 +29,13 @@ import co.elastic.clients.elasticsearch.core.search.SourceFilter;
 import co.elastic.clients.elasticsearch.indices.AnalyzeRequest;
 import co.elastic.clients.elasticsearch.indices.analyze.AnalyzeToken;
 import co.elastic.clients.json.JsonData;
-import io.camunda.exporter.tasks.incident.IncidentUpdateRepository.NoopIncidentUpdateRepository;
 import io.camunda.webapps.schema.descriptors.operate.template.IncidentTemplate;
 import io.camunda.webapps.schema.descriptors.operate.template.ListViewTemplate;
 import io.camunda.webapps.schema.descriptors.operate.template.OperationTemplate;
 import io.camunda.webapps.schema.descriptors.operate.template.PostImporterQueueTemplate;
 import io.camunda.webapps.schema.entities.operate.IncidentEntity;
 import io.camunda.webapps.schema.entities.operate.IncidentState;
+import io.camunda.webapps.schema.entities.operate.listview.ProcessInstanceForListViewEntity;
 import io.camunda.webapps.schema.entities.operate.post.PostImporterActionType;
 import io.camunda.webapps.schema.entities.operation.OperationState;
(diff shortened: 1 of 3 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -53,7 +53,7 @@
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 
-public final class ElasticsearchIncidentUpdateRepository extends NoopIncidentUpdateRepository {
+public final class ElasticsearchIncidentUpdateRepository implements IncidentUpdateRepository {
```
entangled90, line 56: :tada:

npepinpe:
Did my force push discard your review? I just rebased on the other branch, nothing else :sweat_smile:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/ElasticsearchIncidentUpdateRepository.java
```diff
@@ -142,6 +142,29 @@ public CompletionStage<Collection<Document>> getFlowNodeInstances(
...
+            .index(listViewAlias)
+            .source(s -> s.filter(f -> f.includes(ListViewTemplate.TREE_PATH)))
+            .query(q -> q.bool(b -> b.must(idQ, typeQ)));
+
+    return fetchUnboundedDocumentCollection(
+        request,
+        ProcessInstanceForListViewEntity.class,
+        hit ->
+            new ProcessInstanceDocument(
+                hit.id(), hit.index(), Long.parseLong(hit.id()), hit.source().getTreePath()));
```
entangled90, line 165: :thought_balloon:  do we need to handle potential errors when applying the transformer? 
 `Long.parseLong` may throw an exception in the transformer.

We have three options:
- throw an exception: the future may never be completed and the exception is bubbled up
- we fail the future
- we may also want to skip that document, log the exception and keep processing the rest of the documents 

I don't have enough context to know what's the better approach here. 
We don't necessarily need to discuss it in this PR.

npepinpe: I believe we do handle the future when the exception is thrown, but assuming it's deterministic, we would get stuck there.

As this was the previous behavior, I would assume this is ok. That said, it's on my improvement list to not use the ID but instead the key from the document (which is already a long).

npepinpe:
Is there anything you'd like to discuss before approval? :thinking:

entangled90 (approved):

PR #25518, opened 2024-12-02
Implement OpenSearch version of the incident update repository

### LINKED ISSUE #24931: Add OS incident update repository implementation
# Description

Implement all the methods in `IncidentUpdateRepository` for OpenSearch. This should include integration tests of each method against an actual OS instance itself.

I would recommend splitting this into multiple PR, one per method. Inherit from the `NoopIncidentUpdateRepository` initially, and then go from there. This can be easily parallelized among multiple people.

It would make sense to wait until #24930 is done, and just copy that, since most of the time the code is almost the same.

npepinpe:
## Description

This PR implements the OpenSearch variant of the incident update repository. It's largely copied from the ES one, with the following changes:

- Package names are all from OpenSearch. This leads to seemingly lots of code duplication, but really is just because the types are different.
- The OpenSearchAsyncClient declares _all_ of its methods as throwing an `IOException`. Looking at the code though, it seems artificial, since nothing it calls does so. But as it's a checked exception, we can't get rid of it, and I also don't want to mistakenly assume it's not actually throwing anything, so we wrap every call and returned a failed future.

> [!Note]
> I did not add tests for the last point - as the logic is pretty basic, and the compiler enforces us doing something about the exception. There is of course still risk of someone doing the wrong thing, but I think it's quite minimal.

> [!Note]
> There are also no clear scroll tests here, mostly because it felt so redundant. Maybe we can look at them tomorrow together.

## Related issues

closes #24931

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/OpenSearchIncidentUpdateRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/OpenSearchIncidentUpdateRepository.java
@@ -1,0 +1,516 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.tasks.incident;
+
+import io.camunda.webapps.schema.descriptors.operate.template.IncidentTemplate;
+import io.camunda.webapps.schema.descriptors.operate.template.ListViewTemplate;
+import io.camunda.webapps.schema.descriptors.operate.template.OperationTemplate;
+import io.camunda.webapps.schema.descriptors.operate.template.PostImporterQueueTemplate;
+import io.camunda.webapps.schema.entities.operate.IncidentEntity;
+import io.camunda.webapps.schema.entities.operate.IncidentState;
+import io.camunda.webapps.schema.entities.operate.listview.ProcessInstanceForListViewEntity;
+import io.camunda.webapps.schema.entities.operate.post.PostImporterActionType;
+import io.camunda.webapps.schema.entities.operation.OperationState;
+import io.camunda.webapps.schema.entities.operation.OperationType;
+import io.camunda.zeebe.exporter.api.ExporterException;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/OpenSearchIncidentUpdateRepository.java
```diff
@@ -0,0 +1,516 @@
...
+  public CompletionStage<PendingIncidentUpdateBatch> getPendingIncidentsBatch(
+      final long fromPosition, final int size) {
+    final var query = createPendingIncidentsBatchQuery(fromPosition);
+    final var request = createPendingIncidentsBatchRequest(size, query);
+
+    try {
+      return client
+          .search(request, PendingIncidentUpdate.class)
+          .thenApplyAsync(this::createPendingIncidentBatch, executor);
+    } catch (final IOException e) {
```
entangled90, line 106: :thought_balloon: the client also throws `OpenSearchException` (which is a `RuntimeException`), so it does not have to be catched. I expect that in similar scenarios we are not catching them as well in the ES implementation. 

Is it ok to not fail the future in this case?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/OpenSearchIncidentUpdateRepository.java
```diff
@@ -0,0 +1,516 @@
...
+  }
+
+  /**
+   * Variant of {@link #fetchUnboundedDocumentCollection(SearchRequest.Builder, Class, Function)} to
+   * use when you don't care about the source document, meaning you won't be using any
+   * deserialization functionality.
+   */
+  private <T> CompletionStage<Collection<T>> fetchUnboundedDocumentCollection(
+      final SearchRequest.Builder requestBuilder, final Function<Hit<Object>, T> transformer) {
+    return fetchUnboundedDocumentCollection(requestBuilder, Object.class, transformer);
```
entangled90, line 276: :thought_balloon: there is quite a bit of logic which is very similar (but again not the same because of the types) that could have been abstracted away, but I understand that it may not be worth doing that as well as long as it's tested :+1:

entangled90 (approved):

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/OpenSearchIncidentUpdateRepository.java
npepinpe: No, we should catch both. But honestly, looking quickly through their code, this feels more like future proofing or compatibility with the blocking client API.

npepinpe:
Oh, I see...my bad.

PR #25523, opened 2024-12-03
fix: don't use mutable keys for forms cache

### LINKED ISSUE #25504: Form cache is corrupt and causes inconsistent processing
The `DbFormState` keeps an in-memory cache of `PersistedForm` by tenant id and form id to allow efficient in-memory access.
Unfortunately, the cache key is wrapping over a `DirectBuffer`: https://github.com/camunda/camunda/blob/6d0c27479f6a24745c90f352b18b5b7f33280a65/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbFormState.java#L312

At least in one commonly used code path, `FormCreatedV2Applier` supplies the full record to `storeFormInFormColumnFamily` which means that the cache key wraps over parts of this record. Since records are recycled, the buffer and thus the cache key mutates. This breaks the cache implementation which assumes immutable keys. 

As a result, the form cache and all features that rely on this cache cannot be trusted.

To reproduce, apply the following patch that simulates the record recycling in tests:

```patch
Index: zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/appliers/AbstractFormCreatedApplierTest.java
<+>UTF-8
===================================================================
diff --git a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/appliers/AbstractFormCreatedApplierTest.java b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/appliers/AbstractFormCreatedApplierTest.java
@@ -35,6 +35,7 @@
   TypedEventApplier<FormIntent, FormRecord> formCreatedApplier;
 
   private MutableProcessingState processingState;
+  private final FormRecord formRecord = new FormRecord();
 
   @BeforeEach
   void setup() {
@@ -83,7 +84,8 @@
       final int version,
       final long deploymentKey,
       final String tenant) {
-    return new FormRecord()
+    formRecord.reset();
+    return formRecord
         .setFormKey(key)
         .setFormId(id)
         .setVersion(version)
```

and then run tests such as `shouldFindFormByFormIdAndDeploymentKey`.

lenaschoenburg:
This fixes a severe bug where we were using mutable `DirectBuffer`s as the key when caching forms. As a result, the cache was corrupt and could cause inconsistent processing.

To avoid this, we are now using immutable `String`s for form ids everywhere. This is not just safer but also easier to use as most call sites apart from tests were already handling them as `String` and we can now avoid wrapping them in a temporary buffer.

Fixes #25504

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
@@ -242,7 +242,7 @@ public final class BpmnUserTaskBehavior {
   private Either<Failure, PersistedForm> findLatestFormById(
       final String formId, final String tenantId, final long scopeKey) {
     return formState
-        .findLatestFormById(wrapString(formId), tenantId)
+        .findLatestFormById(formId, tenantId)
         .<Either<Failure, PersistedForm>>map(Either::right)
         .orElseGet(
             () ->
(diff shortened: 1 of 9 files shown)


lenaschoenburg:
@koevskinikola Please double check that the backports are sufficient.

npepinpe (approved):

npepinpe:
I assume you omitted 8.3 on purpose?

koevskinikola (approved):
LGTM @lenaschoenburg 👍 

🙃 We can also consider removing the `getFormIdBuffer` method if it's not being used.

PR #25524, opened 2024-12-03
fix: don't use mutable keys for resource cache

### LINKED ISSUE #25505: Resource cache is corrupt
Similar to https://github.com/camunda/camunda/issues/25504, `DbResourceState` also keeps a cache of resources by tenant id and resource id. The resource id is again just a mutable buffer that `ResourceCreatedApplier` takes from recycled records.

As a result, entries in `resourcesByTenantIdAndIdCache` do not have a stable key and lookups may return the wrong values.
Resources appear to be unused until now so hopefully this isn't causing inconsistent processing yet.

lenaschoenburg:
Replaces the use of `DirectBuffer` in the resource cache key.
Also refactors `ResourceState` to not make use of `DirectBuffer` at all which should improve safety long term and make the API easier to use.

Fixes #25505


npepinpe (approved):

koevskinikola (approved):

korthout:
No need for a backport to `release-8.7.0-alpha2` ([details](https://github.com/camunda/camunda/pull/25529#issuecomment-2514072272))

koevskinikola:
FYI @kzraitel we did some changes to the `DbResourceState` that might impact later tasks due to a bug in the code.

PR #25531, opened 2024-12-03
feat: export treepath from PROCESS_INSTANCE records

### LINKED ISSUE #25474: [EPIC]: Build treePath in Engine and add respecitive properties on record(s)
## Description

In the early stages of the Camunda Exporter Project, we assumed that the treePath was only really necessary for incidents, and for most other cases it was not useful and could be avoided.

We [realized](https://camunda.slack.com/archives/C05QBRGMA9F/p1732562391807879) that the treePath is used in many more places. 

Operate uses it in several places, and the effort to change these based on our previous decision was rather intensive. This is especially an issue looking at our current timeline (release end of Dez).

This is the reason why we ran some POC’s one for [building the treePath in the Exporter](https://github.com/camunda/camunda/pull/25310) and one [in the Engine](https://github.com/camunda/camunda/pull/25352). We have implemented the POCs and benchmarked them intensively at the RSG (Hamburg 2024).

### Details on the benchmarking and decision

Based on the benchmarks and results, and intensive discussions with stakeholders and peers we [decided](https://camunda.slack.com/archives/C05QBRGMA9F/p1732872462881639) to go in the direction of implementing the treePath in the engine.

### Details on TreePath

There is a difference between inter and intra TreePath. The inter treePath goes over process definitions, including call activities and is used for showing incidents in Operate.

The intra treePath is just inside a process instance, and will/can be used to have a possibility to resolve existing variable scopes and understand the hierarchy in the PI.

We need both versions of treePaths, to support all current features!

> [!Important]
>
> Maxim:
> 
>  * internet - global -> inter treePath - global over multiple process instances
>  * intranet - local -> intra treePath - local on current process instance
> 

### Breakdown

```[tasklist]
### Engine
- [x] Adjust TreePathBuilder to accept flowScopeKey and RecordValue - avoiding look up of ElementInstance (that might not exist in ACTIVATING case)
- [x] Add inter treePath properties (similar to IncidentRecord) to ProcessInstanceRecord - ignore JSON output
- [ ] Use new compressed array property - provided by Lena
- [x] Find a place to calculate and add treePath properties to ProcessInstanceRecordValue - Activating intent
- [x] Make sure that properties are cleared for follow up events like ACTIVATED
- [x] Write a test to verify properties are correctly set - without CallActivity, with CallActivity, with multiple Callactivities
- [ ] Handle edge case with deep nesting - create incident for example (early, as otherwise Incident also runs into this issue)
- [x] Write tests that verify that on other intents properties are not set
```

```[tasklist]
### Exporter
- [ ] Consume new properties for FlowNodeInstance index - intra tree path (FlowNodeInstanceFromProcessInstanceHandler)
- [ ] Consume new properties for ListViewPI index - inter tree path (ListViewProcessInstanceFromProcessInstanceHandler)
- [ ] Migrate root QA/IT tests to Camunda Exporter (allows to validate that treePath works as expected)
- [ ] Fix bug - exporter should consume all treePath from all call activities - right now it is only the first one
```

sdorokhova:
## Description

* Fill in tree path fields in exporter handlers from PROCESS_INSTANCE record:
  * in process instance entity
  * in flow node instance entity
* stop filling them from INCIDENT record (not needed any more)
* enable Camunda exporter in root IT test

## Related issues

related with #25474


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandler.java
```diff
@@ -90,9 +93,23 @@ public void updateEntity(
...
-      entity.setTreePath(recordValue.getProcessInstanceKey() + "/" + record.getKey());
-      entity.setLevel(1);
+      final ProcessInstanceRecordValue value = record.getValue();
+      if (value.getElementInstancePath() != null && !value.getElementInstancePath().isEmpty()) {
+        // Build the treePath in the format
+        // <processInstanceKey>/<flowNodeInstanceKey>/.../<flowNodeInstanceKey> where upper level
+        // flowNodeInstanceKeys are normally subprocess(es) or multi-instance body This is an intra
+        // tree path that shows position of flow node instance inside one process instance.
+        final List<String> treePathEntries =
+            value.getElementInstancePath().getFirst().stream().map(String::valueOf).toList();
```
ChrisKujawa, line 103: ❌ This might be not correct. I think it should be `getLast`, see https://github.com/camunda/camunda/commit/a05003cd1ba59ead7a376ca88771017ffdc0c2b5#diff-a07129f0655c6899a0d04214069f7a23d3a2f09b7573b6d8317099e0422d565cR1103-R1105

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandlerTest.java
```diff
@@ -1,209 +0,0 @@
...
-    assertThat(processInstanceForListViewEntity2.getTreePath())
-        .isEqualTo("PI_111/FN_callActivity1/FNI_123/PI_222");
-
-    // when called process 3rd level
-    final ProcessInstanceForListViewEntity processInstanceForListViewEntity3 =
-        new ProcessInstanceForListViewEntity().setId(String.valueOf(pi3Key));
-    underTest.updateEntity(incidentRecord, processInstanceForListViewEntity3);
-    // then
-    assertThat(processInstanceForListViewEntity3.getTreePath())
-        .isEqualTo("PI_111/FN_callActivity1/FNI_123/PI_222/FN_callActivity2/FNI_234/PI_333");
```
ChrisKujawa, line 192: Svetlana is working on getting this test part of the other handler UT.

This we should definitely something we should cover.

🔧 I think we should have some test with more nesting inside the same process instance.

ChrisKujawa (commented):
Thanks @sdorokhova I already had a pre-review. Maybe my input is helpful.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandler.java
sdorokhova: Caught this in test and fixed in the next commit!

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandlerTest.java
sdorokhova: > I think we should have some test with more nesting inside the same process instance.

You mean more scopes? Additional scopes like subprocesses won't go into inter tree path. But I added them to record data: https://github.com/camunda/camunda/pull/25531/commits/65f102fce9e8abca643e0c5308be0e7e40dfa0e5#diff-394591f11d1232612dc04e957c52abd3140961414df36bda9546351c809e4ba0R373

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandlerTest.java
```diff
@@ -149,17 +150,18 @@ public void shouldUpsertEntityOnFlush() {
...
-            .setTreePath(null)
             .setFlowNodeId("flowNode1")
             .setProcessDefinitionKey(222L)
             .setBpmnProcessId("bpmnId")
             .setTenantId("tenantId")
-            .setLevel(0)
             .setStartDate(OffsetDateTime.now())
             .setEndDate(OffsetDateTime.now())
+            .setTreePath("444/356/111")
+            .setLevel(1)
```
ChrisKujawa, line 164: 🤔 Shouldn't this then 2?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandlerTest.java
```diff
@@ -193,12 +229,18 @@ public void shouldUpdateEntityFromRecord() {
         ImmutableProcessInstanceRecordValue.builder()
             .from(factory.generateObject(ProcessInstanceRecordValue.class))
             .withBpmnElementType(BpmnElementType.SERVICE_TASK)
+            .withElementInstancePath(
+                List.of(
+                    List.of(111L, 777L, 999L),
+                    List.of(222L, 789L, 888L),
+                    List.of(333L, 444L, 555L, 666L)))
```
ChrisKujawa, line 236: 👍🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandlerTest.java
```diff
@@ -307,15 +343,98 @@ void shouldUpdateEntityFromRecord() {
...
+    final Long pi3Key = 333L;
+    final Record<ProcessInstanceRecordValue> processInstanceRecord =
+        createProcessInstanceRecord(
+            List.of(
+                List.of(pi1Key, subprocess1Key, callActivity1Key),
+                List.of(pi2Key, subprocess2Key, callActivity2Key),
+                List.of(pi3Key)),
+            List.of(callActivityIndex1, callActivityIndex2),
+            List.of(processDefinitionKey1, processDefinitionKey2, 777L),
+            pi3Key);
```
ChrisKujawa, line 378: 👍🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandlerTest.java
```diff
@@ -307,15 +343,98 @@ void shouldUpdateEntityFromRecord() {
...
+    processCache.put(
+        processDefinitionKey2, new CachedProcessEntity(null, null, List.of("0", callActivityId2)));
+
+    // when called process 3rd level
+    final ProcessInstanceForListViewEntity processInstanceForListViewEntity3 =
+        new ProcessInstanceForListViewEntity().setId(String.valueOf(pi3Key));
+    underTest.updateEntity(processInstanceRecord, processInstanceForListViewEntity3);
+    // then
+    assertThat(processInstanceForListViewEntity3.getTreePath())
+        .isEqualTo("PI_111/FN_callActivity1/FNI_123/PI_222/FN_callActivity2/FNI_234/PI_333");
```
ChrisKujawa, line 392: 🔧 Maybe we call the variable entity to make it a bit more simpler.

❓ Understanding question on the treePath. Does this mean we will not contain a subprocess element instance key in the inter treePath?

sdorokhova: Exactly, we won't.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
```diff
@@ -264,4 +282,47 @@ private String getProcessName(final Long processDefinitionKey, final String bpmn
...
+          "No elementInstancePath is provided for process instance id: {}. TreePath will be set to default value.",
+          processInstanceKey);
+      return new TreePath().startTreePath(processInstanceKey);
+    }
+    // example of how the tree path is built when current instance is on the third level of calling
+    // hierarchy:
+    // PI_<parentProcessInstanceKey>/FN_<parentCallActivityId>/FNI_<parentCallActivityInstanceKey>/
+    // PI_<secondLevelProcessInstanceKey>/FN_<secondLevelCallActivityId>/FNI_<secondLevelCallActivityInstanceKey>/PI_<currentProcessInstanceKey>
+    final TreePath treePath = new TreePath();
+    for (int i = 0; i < elementInstancePath.size(); i++) {
```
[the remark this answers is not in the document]
ChrisKujawa: I had a wrong assumption here

ChrisKujawa (approved):
@sdorokhova I will approve already. Please take a look at the open suggestions, comments. Afterwards, let's go ahead and merge the changes, as multiple PRs depend on it.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandler.java
```diff
@@ -264,4 +282,47 @@ private String getProcessName(final Long processDefinitionKey, final String bpmn
...
+      final long key,
+      final Long processInstanceKey,
+      final List<List<Long>> elementInstancePath,
+      final List<Long> processDefinitionPath,
+      final List<Integer> callingElementPath) {
+    if (elementInstancePath == null || elementInstancePath.isEmpty()) {
+      LOGGER.warn(
+          "No elementInstancePath is provided for process instance id: {}. TreePath will be set to default value.",
+          processInstanceKey);
+      return new TreePath().startTreePath(processInstanceKey);
```
ChrisKujawa, line 297:
```suggestion
    if (elementInstancePath == null || elementInstancePath.isEmpty()) {
      LOGGER.warn(
          "No elementInstancePath is provided for process instance key: {}. TreePath will be set to default value (PI key).",
          processInstanceKey);
      return new TreePath().startTreePath(processInstanceKey);
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromIncidentHandler.java
mustafadagher, line 1: ❓ @sdorokhova was there a reasoning behind deleting this handler?

PR #25533, opened 2024-12-03
fix: exclude child documents when querying tenants and groups

emulbreh:
## Description

ES returns child documents together with the parents by default, but those cannot be handled by the entity transformers.

Roles queries already include the same filter.


koevskinikola (approved):
LGTM 👍

Oleksiivanov (approved):

PR #25537, opened 2024-12-03
feat: store reconstructed deployment in state and update references

### LINKED ISSUE #24817: Reconstruct deployment for a single resource

lenaschoenburg:
This adds an applier for the Deployment/Reconstructed event that:
1. Stores the deployment with all its metadata in the deployment state
2. Updates persisted processes, lookup tables and in-memory caches to point to the new deployment key
2. Updates persisted forms, lookup tables and in-memory caches to point to the new deployment key
2. Updates persisted decisions, lookup tables and in-memory caches to point to the new deployment key

Depends on #25487 
Closes #24817


lenaschoenburg:
This is on hold for now, we'll come back to this later.

Issue #24476, opened 2024-11-07
Remove current implementation of Backups from Tasklist, Operate and Optimize


marcosgvieira:
As we are centralizing the backups of Tasklist, Identity, Operate and Optimize we should remove the current implementation of backups for those components.

### Acceptance Criteria
- [x] As part of 8.7 there is no Tasklist specific implementation of backups
- [x] As part of 8.7 there is no Operate specific implementation of backups
- [ ] As part of 8.7 there is no Optimize specific implemenation of backups (here we have the risk of Optimize not making to be part of single-application as part of 8.7, please double check in [#prj-optimize-as-part-of-single-application](https://camunda.slack.com/archives/C07MMF3MRPA)

PR #25538, opened 2024-12-03
feat: remove BackupManager and its implementations

entangled90:
# Description
All classes related to backups in tasklist are removed

 `ElasticsearchTasklistBackupRepository` and `OpensearchTasklistBackupRepository` are created to reuse the tasklistEs/Os clients when it's not deployed with operate.
If the client will be merged, then there is no reason to have this duplication.

Note that I used operate's `ElasticSearchCondition` in the default repositories and tasklist's `ElasticSearchCondition` in `ElasticSearchTasklistBackupRepository`.  The same for OS repositories

## Related issues

relates #24476


npepinpe (approved):
:+1: 

- :x: Please document why we still have different Tasklist/Operate repository Spring configuration classes, in the PR description and also in the code (to hint to future maintainers when we can streamline this).

PR #25547, opened 2024-12-03
fix: use the resource loading cache correctly

### LINKED ISSUE #25507: Resource cache is not filled correctly and stays empty
`DbResourceState` defines `resourcesByTenantIdAndIdCache` as a loading cache, with the intention presumably being that any cache lookup would automatically compute ("load") the value.

However, the only cache lookup is a `getIfPresent`  call which does _not_ load the cache.

In effect, the cache is only populated in `storeResourceInResourceColumnFamily`, i.e. when deploying resources. After a restart, the cache remains empty and useless.

A simple fix to swap the non-loading `getIfPresent` with a loading `get` will not work because the `CacheLoader` is defined recursively, where a load would first start with a cache lookup. This is rejected by the Guava cache with `java.lang.IllegalStateException: Recursive load`.

koevskinikola:
Break the loading chain, i.e. if a resource is
not found in the cache, look it up directly in
the resource_by_id column family. The current
implementation was creating a loading loop.

Remove eager loading of the cache with lazy
loading, i.e. we only cache a resource if it is
explicitly looked up.

Replace the getIfPresent cache method with a get
method since the former doesn't load the cache if
the resource is not present there.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25507

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
@@ -24,6 +24,7 @@ import io.camunda.zeebe.engine.state.mutable.MutableResourceState;
 import io.camunda.zeebe.protocol.ZbColumnFamilies;
 import io.camunda.zeebe.protocol.impl.record.value.deployment.ResourceRecord;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
 
 public class DbResourceState implements MutableResourceState {
 
@@ -129,7 +130,7 @@ public class DbResourceState implements MutableResourceState {
                 new CacheLoader<>() {
                   @Override
                   public PersistedResource load(final TenantIdAndResourceId key) {
-                    return findLatestResourceById(key.resourceId, key.tenantId).orElse(null);
+                    return getPersistedResourceById(key.resourceId, key.tenantId);
                   }
                 });
   }
@@ -140,9 +141,6 @@ public class DbResourceState implements MutableResourceState {
     dbResourceKey.wrapLong(record.getResourceKey());
     dbPersistedResource.wrap(record);
     resourcesByKey.upsert(tenantAwareResourceKey, dbPersistedResource);
-    resourcesByTenantIdAndIdCache.put(
-        new DbResourceState.TenantIdAndResourceId(record.getTenantId(), record.getResourceId()),
-        dbPersistedResource.copy());
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
```diff
@@ -291,6 +279,7 @@ public void clearCache() {
   private PersistedResource getPersistedResourceById(
       final String resourceId, final String tenantId) {
     dbResourceId.wrapString(resourceId);
+    tenantIdKey.wrapString(tenantId);
```
lenaschoenburg, line 282: :grimacing: That sort of bug is already a security issue, right? Such an easy mistake to make with pretty bad consequences.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
```diff
@@ -140,9 +141,6 @@ public void storeResourceInResourceColumnFamily(final ResourceRecord record) {
     dbResourceKey.wrapLong(record.getResourceKey());
     dbPersistedResource.wrap(record);
     resourcesByKey.upsert(tenantAwareResourceKey, dbPersistedResource);
-    resourcesByTenantIdAndIdCache.put(
-        new DbResourceState.TenantIdAndResourceId(record.getTenantId(), record.getResourceId()),
-        dbPersistedResource.copy());
```
lenaschoenburg, line 145: :thought_balloon: This could stay IMO, it avoids an unnecessary cache load for recently deployed resources.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
```diff
@@ -301,10 +290,21 @@ private PersistedResource getPersistedResourceById(
...
   }
 
   private record TenantIdAndResourceId(String tenantId, String resourceId) {}
+
+  /**
+   * This exception is thrown when a DbResourceState cache can't find a PersistedResource in the
+   * state for the given parameters. This must be a checked exception, because of the way the {@link
+   * LoadingCache} works.
+   */
+  private static final class ResourceNotFoundException extends Exception {}
```
lenaschoenburg, line 309: :question: This appears unused?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
koevskinikola: My reasoning was to avoid overfilling the cache, but we'll always have the latest resource versions on hand this way. 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
koevskinikola: That's because in the rush to fix this, I forgot to use this! 😦 

I'll add a commit to use it properly.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbResourceState.java
koevskinikola: We wrap it in the `getLatestResourceVersion` method, so we didn't have a bug, but it's always preferable to set it explicitly since we may refactor the code and miss it in the future.

lenaschoenburg (approved):
Sorry, forgot to submit my review :facepalm: 

Looks good to me now. Interesting to learn about how `LoadingCache` uses checked/unchecked exceptions. It looks a bit foreign compared to the rest of the engine code.

PR #25562, opened 2024-12-03
test: await processor to reach the end to account for internal processing

lenaschoenburg:
With deployment reconstruction always producing some records right after startup, we need to relax this test such that it gives the stream processor a couple of seconds to reach the end.


remcowesterhoud (approved):
Nice find! 😉

PR #25566, opened 2024-12-03
feat: add user menu

emulbreh:
## Description
This PR borrows the user menu from operate for identity.

image

### DIFF
--- a/identity/client/.env
+++ b/identity/client/.env
@@ -1,0 +1,1 @@
+VITE_APP_VERSION=$npm_package_version
--- a/identity/client/package.json
+++ b/identity/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "identity-frontend",
   "private": true,
-  "version": "0.0.0",
+  "version": "8.7.0-SNAPSHOT",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/identity/client/src/components/layout/AppHeader.tsx
+++ b/identity/client/src/components/layout/AppHeader.tsx
@@ -3,11 +3,19 @@ import { useGlobalRoutes } from "src/components/global/useGlobalRoutes";
 import { useNavigate } from "react-router";
 import { useApi } from "src/utility/api";
 import { checkLicense, License } from "src/utility/api/headers";
+import { getAuthentication } from "src/utility/api/authentication";
+import { useCallback } from "react";
+import { ArrowRight } from "@carbon/react/icons";
 
 const AppHeader = () => {
   const routes = useGlobalRoutes();
   const navigate = useNavigate();
   const { data: license } = useApi(checkLicense);
+  const { data: camundaUser } = useApi(getAuthentication);
+
+  const logout = useCallback(() => {
+    console.log("logout");
+  }, []);
 
   return (
     <C3Navigation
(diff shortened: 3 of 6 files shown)


identity/client/.env
```diff
@@ -0,0 +1 @@
+VITE_APP_VERSION=$npm_package_version
```
Ben-Sheppard, line 1: I haven't looked at the infra around this but is this something that we have to manually manage? Is this set somewhere on release?

identity/client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "identity-frontend",
   "private": true,
-  "version": "0.0.0",
+  "version": "8.7.0-SNAPSHOT",
```
Ben-Sheppard, line 4: Similar question to above, do we have to manually manage this version?

emulbreh: Yes, it's not great. It's how the other webapps work though (e.g. [tasklist](https://github.com/camunda/camunda/blob/fc38c934dbb43f2b67bb10b2fdfada3dcdee2315/tasklist/client/package.json#L3), [operate](https://github.com/camunda/camunda/blob/fc38c934dbb43f2b67bb10b2fdfada3dcdee2315/operate/client/package.json#L3)).

identity/client/.env
emulbreh: It comes from the package.json file (where it's hardcoded). I've taken the same approach as [operate](https://github.com/camunda/camunda/blob/fc38c934dbb43f2b67bb10b2fdfada3dcdee2315/operate/client/.env#L1) and [tasklist](https://github.com/camunda/camunda/blob/fc38c934dbb43f2b67bb10b2fdfada3dcdee2315/tasklist/client/.env#L1) here.

identity/client/package.json
Ben-Sheppard: Ah I see, so in the other components they include an execution in their POM to set this as part of the pipeline ([REF](https://github.com/camunda/camunda/blob/fc38c934dbb43f2b67bb10b2fdfada3dcdee2315/tasklist/client/pom.xml#L41-L50)), can we also add this so we benefit from the same structure/flow?

emulbreh: Oh, this is better!

identity/client/src/components/layout/AppHeader.tsx
```diff
@@ -3,11 +3,19 @@ import { useGlobalRoutes } from "src/components/global/useGlobalRoutes";
...
+import { ArrowRight } from "@carbon/react/icons";
 
 const AppHeader = () => {
   const routes = useGlobalRoutes();
   const navigate = useNavigate();
   const { data: license } = useApi(checkLicense);
+  const { data: camundaUser } = useApi(getAuthentication);
+
+  const logout = useCallback(() => {
+    console.log("logout");
```
Ben-Sheppard, line 17: 🔧 Not sure if you intentionally left this in :D

identity/client/src/components/layout/AppHeader.tsx
```diff
@@ -31,6 +39,56 @@ const AppHeader = () => {
...
       }}
+      userSideBar={{
+        version: import.meta.env.VITE_APP_VERSION,
+        ariaLabel: "Settings",
+        customElements: {
+          profile: {
+            label: "Profile",
+            user: {
+              name: camundaUser?.displayName ?? "",
+              email: "",
```
Ben-Sheppard, line 50: Is email always missing? can we also use `camundaUser?.email ?? ""` here?

Ben-Sheppard (commented):
Couple of small comments but otherwise it seems reasonable to me :D

identity/client/src/components/layout/AppHeader.tsx
emulbreh: Yes, although if you prefer I could also remove the whole callback. I had replaced this with an actual logout action in https://github.com/camunda/camunda/pull/26262

Ben-Sheppard: Ah I'm fine with either if we're going to address this proactively in a upcoming PR

Ben-Sheppard (approved):
Thank you for addressing my comments here, I'm keen to get through the PRs so I'm happy for this to be merged and we can continue to make progress.

I'll leave the `console.log` comment up to you, if you think we'll address it soon then its fine, if not maybe a comment instead?

emulbreh:
> I'll leave the `console.log` comment up to you, if you think we'll address it soon then its fine, if not maybe a comment instead?

I'll leave it in so the logout link does something. We'll fix it soon.

PR #25570, opened 2024-12-03
fix: make sortOrderEnum and add exception path on advisor

### LINKED ISSUE #20512: Unified REST API: Replace sortOrder Field with Enum in SearchQuerySortRequest
## Description
We need to update the `sortOrder` field in the `SearchQuerySortRequest` from a string type to an enum with possible values `asc` and `desc` in our OpenAPI specification. The default value should be `asc`. Additionally, we need to ensure that the API validation correctly returns a 400 Bad Request with an appropriate `ProblemDetail` object when an invalid value is provided.

## Acceptance Criteria
- The sortOrder field in the SearchQuerySortRequest is an enum with values `asc` and `desc`, and the default is `asc`.
- The API returns a 400 Bad Request with a ProblemDetail object when an invalid sort order value is provided.
- Unit tests (`ProcessInstanceControllerTest`) are passing/updated with the new validation logic.

## Implementation proposal
We can use an ExceptionHandler on `MethodArgumentNotValidException` to identify the invalid input field

```java
@RestControllerAdvice
public class GlobalExceptionHandler {

    @ExceptionHandler(MethodArgumentNotValidException.class)
    public ResponseEntity<ProblemDetail> handleValidationExceptions(MethodArgumentNotValidException ex) {
        for (FieldError error : ex.getBindingResult().getFieldErrors()) {
            if ("sortOrder".equals(error.getField())) {
                // return the problem Detail with the invalid sortOrder
            }
           ...
        }

```

nathansandi:
## Description

- Update `rest-apy.yaml`  on Sort to use `enum`
- Update Integration tests
- Update Unit tests
- Update respective Request Mappers

## Related issues

closes https://github.com/camunda/camunda/issues/20512


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5793,9 +5793,8 @@ components:
           description: The field to sort by.
           type: string
         order:
-          description: The order in which to sort the field.
-          type: string
-          default: asc
+          description: The sort order.
+          $ref: "#/components/schemas/SortOrderEnum"
```
tmetzke, line 5797: 🙃 I believe that siblings next to `$ref` are ignored in OpenAPI so you can remove the description here. @pepopowitz, that is the way it works, right? 😆 I am horribly confused by now 🙈

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -6184,6 +6183,12 @@ components:
         - FAILED
         - UNSPECIFIED
         - UNKNOWN
+    SortOrderEnum:
+      description: The Sort Order.
+      enum:
+        - asc
+        - desc
+      default: asc
```
tmetzke, line 6191: 🔧 I believe that going upper case here would be more common, right?
```suggestion
        - ASC
        - DESC
      default: ASC
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -994,15 +993,12 @@ Either<ProblemDetail, T> buildSearchQuery(
...
 
-  private static List<String> applySortOrder(
-      final String order, final SortOption.AbstractBuilder<?> builder) {
-    final List<String> validationErrors = new ArrayList<>();
-    switch (order.toLowerCase()) {
-      case "asc" -> builder.asc();
+  private static void applySortOrder(
+      final SortOrderEnum order, final SortOption.AbstractBuilder<?> builder) {
+    switch (order.getValue()) {
       case "desc" -> builder.desc();
```
tmetzke, line 999: 🔧 I guess you can do a proper switch on the enum here, right?
```suggestion
    switch (order) {
      case DESC -> builder.desc();
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
```diff
@@ -33,7 +33,8 @@ public class GlobalControllerExceptionHandler extends ResponseEntityExceptionHan
       "Required request body is missing";
   private static final String REQUEST_BODY_PARSE_EXCEPTION_MESSAGE =
       "Request property [%s] cannot be parsed";
-  private static final String INVALID_ENUM_VALUE_EXCEPTION_MESSAGE = "Invalid Enum value";
+  // Thrown by the OpenApi in case of an invalid enum value
+  private static final String INVALID_ENUM_VALUE_EXCEPTION_MESSAGE = "Unexpected value";
```
tmetzke, line 37: 💭 Two ideas:
1. Can we be more explicit here in what is unexpected? Is the enum class and/or the field (key in the JSON) known? This could help create a more useful message for users, e.g. "The value provided in '<field>' is unsupported".
2. If possible, but this is an absolute bonus: if we know the enum class (because it's in the error message), we can even fetch and return the possible values, e.g. "The value provided in '<field>' is unsupported, it must be any of [<list of enum values>]".

nathansandi: > If possible, but this is an absolute bonus: if we know the enum class (because it's in the error message), we can even fetch and return the possible values, e.g. "The value provided in '' is unsupported, it must be any of [<list of enum values>]".

I have used the `INVALID_ENUM_VALUE_EXCEPTION_MESSAGE ` from OpenApiSPEC to validate the message. It comes very generic and we have a lot of ENUMS. 
The OpenAPI Spec does not bring the possible Enum values over this case. I think this would be not 

> 1. Can we be more explicit here in what is unexpected? Is the enum class and/or the field (key in the JSON) known? This could help create a more useful message for users, e.g. "The value provided in '<field>' is unsupported".

I can change in the line down there, this constant is only use to validate the error caused by the Open API Spec, I can update the error message over here: 
`detail = Objects.requireNonNull(httpMessageNotReadableException.getRootCause()).getMessage();`
And update the message accordingly.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nathansandi: I followed the requirements over the issue https://github.com/camunda/camunda/issues/20512 
I can move it all to ASC and DESC with no problems

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nathansandi: This is true, when I use allOf I can remove the description ;) I was fixing this right now

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
nathansandi: To give a better example, this is the stack trace from the Open API Spec from ENUMS

`org.springframework.http.converter.HttpMessageNotReadableException: JSON parse error: Cannot construct instance of `io.camunda.zeebe.gateway.protocol.rest.SortOrderEnum`, problem: Unexpected value 'dsc'
`

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: Alternatively, you can keep the description and move the `$ref` into an `allOf`.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/VariablesQueryControllerTest.java
```diff
@@ -183,13 +179,13 @@ void shouldSearchVariableWithSorting() {
...
             new VariableQuery.Builder()
-                .sort(new VariableSort.Builder().name().desc().value().asc().build())
+                .sort(new VariableSort.Builder().name().desc().build())
                 .build());
   }
 
   @Test
   void shouldInvalidateVariableSearchQueryWithBadSortOrder() {
-    // given
+    // given`
```
tmetzke, line 188:
```suggestion
    // given
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/GlobalControllerExceptionHandler.java
tmetzke: I gave this a quick look and discovered the following:
* The `HttpMessageNotReadableException` we receive has a cause exception of `ValueInstantiationException`.
* The value instantiation exception can tell you the location of the issue in the source JSON, i.e., the field where it occurs.

With that, we should be able to give a better hint here where the problematic field is. We could even use the defined type information (`io.camunda.zeebe.gateway.protocol.rest.SortOrderEnum`) to fetch all possible values and return those to the user.

For example, with the following criteria we can check if we have an enum issue:
```
  private boolean isUnknownEnumError(final Exception ex) {
    return ex instanceof HttpMessageNotReadableException
        && ex.getCause() instanceof final ValueInstantiationException instantiationException
        && instantiationException.getType().isEnumType();
  }
```

The following code would get the details we need:
```
...
    } else if (isUnknownEnumError(ex)) {
      final var instantiationException = (ValueInstantiationException) ex.getCause();
      final var options =
          Arrays.stream(instantiationException.getType().getRawClass().getEnumConstants())
              .map(Objects::toString)
              .toList();
      final var field = instantiationException.getPath().getLast().getFieldName();
      detail =
          "%s for enum field '%s'. Use any of the following values: %s"
              .formatted(
                  ((HttpMessageNotReadableException) ex).getRootCause().getMessage(),
                  field,
                  options);
    } else {
...
```

This would lead to the following error message for the enum:
```
Unexpected value 'dsc' for enum field 'order'. Use any of the following values: [ASC, DESC]
```

nathansandi: Thanks for driving this @tmetzke , this works. 
Super interesting, I miss this kind of exception (to be honest pretty new for me), I was checking around the `InvalidArgumentException`

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/VariablesQueryControllerTest.java
```diff
@@ -157,11 +157,7 @@ void shouldSearchVariableWithSorting() {
                 "sort": [
                     {
                         "field": "name",
-                        "order": "desc"
-                    },
-                    {
-                        "field": "value",
-                        "order": "asc"
+                        "order": "DESC"
```
tmetzke, line 160: ❓ Why are we removing this from this test case?

tmetzke (commented):
👍 Looks good overall, thanks, @nathansandi, for driving this.
💭 We should check if users can also provide lowercase input. I think we have to enable `MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS` in the `JacksonConfig` in the REST gateway module. Let's write some test cases to verify this works with lowercase input as well, shall we?

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/VariablesQueryControllerTest.java
nathansandi: I have re-added this test, but with `asc` lower case in order to test the proeprty 

`<useEnumCaseInsensitive>true</useEnumCaseInsensitive>` from OpenAPI generator , from [here](https://github.com/OpenAPITools/openapi-generator/pull/15273/files)

(Use `MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS` does not work for OpenApi Generator on Jackson Mapper Config)

tmetzke: Works for me 👍 
Thanks for looking into this, this will make the API a lot nicer to use and less error-prone 🚀

tmetzke (approved):

Issue #24932, opened 2024-11-20
Add acceptance tests for the incident update task


npepinpe:
# Description

Add QA tests to make sure we are processing updates properly. See for example `IncidentPostImportZeebeImportIT` and `IncidentZeebeImportIT` - they are both quite tied to the importer, but can be used as inspiration for test coverage.

We should at least cover:

- Deploying a process with a single task and incident on the task, I expect the following:
  - I can query the incident and it is active
  - I can query the process instance and it has an incident
  - I can query the flow node instance of the task and it has an incident

- Deploying a process with a single task and incident, then resolving this incident, I expect the following:
  - There are no incidents query-able
  - I can query the process instance and there is no incident
  - I can query the flow node instance of the task and there is no incident

- Deploying a process (P1) with a call activity (P2), which has a call activity (P3), which has a task with an incident:
  - I can query the incident and it is active
  - I can query the flow node instance of the task and it has an incident
  - I can query the second call activity process instance (P3) and it has an incident
  - I can query the second call activity flow node instance and it has an incident
  - I can query the first call activity process instance (P2) and it has an incident
  - I can query the first call activity flow node instance and it has an incident
  - I can query the root process instance (P1) and it has an incident

- Deploying a process (P1) with a parallel gateway, and a call activity (P2), and a parallel gateway with two tasks (T1, T2), and both tasks have an incident; I then resolve the incident on T2:
  - I can query the remaining incident and it is active
  - I can query the flow node instance of T1 and it has an incident
  - I can query the call activity process instance (P2) and it has an incident
  - I can query the call activity flow node instance and it has an incident
  - I can query the root process instance (P1) and it has an incident

ChrisKujawa:
We have a lot of tests already in the client package also related to incidents, might be that we already cover several parts. We should validate this.

sdorokhova:
@ChrisKujawa what do you mean by "client package"? Could you clarify please?
Do you think these test should be root integration test in /qa/integration-tests module?

npepinpe:
I think Chris means the Operate and Tasklist E2E tests (i.e. what's run in those CI jobs), since it starts the application, and using the front end, checks that incidents show up, etc. That might already be enough of an acceptance test?

It's a little hard to debug personally 🤔

ChrisKujawa:
[...]

PR #25572, opened 2024-12-03
Create fully working incident update task (old post-importer)

npepinpe:
## Description

This PR integrates the new post-importer fully into the project. There is some duplication around the repositories:

- We create more than one ES/OS client
- This cascades into keeping track of them in the background task manager, so we can close them
- Some test duplication

In the future, one of the tasks we want is to abstract the client itself away to have a single repository implementation for the archiver, and a single implementation for the incident update task. At the same time, we only have to track one closeable resource in the task manager: the client. So there'll be a lot of stuff cut down later, but I think it's OK to accept this as is for now.

I added only basic E2E tests, as more will be covered as part of https://github.com/camunda/camunda/issues/25595. Here, I mostly wanted to add tests for the very basic functionality to make sure it's working as intended.

One thing: the extension used everywhere (`BrokerWithCamundaExporterITInvocationProvider`) is not actually wiring OS. I'll follow up on that with the rest of the project team.

## Related issues

related to #24932

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
@@ -40,7 +40,7 @@ public final class BackgroundTaskManager implements CloseableSilently {
     this.archiverRepository =
         Objects.requireNonNull(archiverRepository, "must specify an archiver repository");
     this.incidentRepository =
-        Objects.requireNonNull(incidentRepository, "must specify an archiver repository");
+        Objects.requireNonNull(incidentRepository, "must specify an incident repository");
     this.logger = Objects.requireNonNull(logger, "must specify a logger");
     this.executor = Objects.requireNonNull(executor, "must specify an executor");
     this.tasks = Objects.requireNonNull(tasks, "must specify tasks");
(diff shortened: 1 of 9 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
```diff
@@ -29,12 +31,16 @@ public final class BackgroundTaskManager implements CloseableSilently {
...
+      final @WillCloseWhenClosed IncidentUpdateRepository incidentRepository,
       final Logger logger,
       final @WillCloseWhenClosed ScheduledThreadPoolExecutor executor,
       final List<Runnable> tasks) {
     this.partitionId = partitionId;
-    this.repository = Objects.requireNonNull(repository, "must specify a repository");
+    this.archiverRepository =
+        Objects.requireNonNull(archiverRepository, "must specify an archiver repository");
+    this.incidentRepository =
+        Objects.requireNonNull(incidentRepository, "must specify an archiver repository");
```
entangled90, line 43: :x: "must specify an incident repository"

entangled90 (approved):
Great! just fix the log message before merging

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
npepinpe: :see_no_evil:

npepinpe, line 43:
```suggestion
        Objects.requireNonNull(incidentRepository, "must specify an incident repository");
```

PR #25594, opened 2024-12-04
feat: add rdbms exporter to exporter e2e tests #24202

### LINKED ISSUE #24202: Provide E2E test for RDBMS Exporter / Search Client
We need a simple E2E Test which starts the while application to make sure if really all beans are provided, and if a basic "Durchstieg" is working.

There is already a PoC Branch for this, and the Camunda Exporter Team is using the same approach already. 
So we have two options here:

- Write our own standalone E2E Test, based on the Poc Branch
- Extend the E2E Tests from Camunda Exporter Team to support also Rdbms Databases

p-wunderlich:
- Add rdbms exporter with h2
- Add possibility to configure per test class which exporters should get started

## Related issues

closes #24202

### DIFF
--- a/qa/integration-tests/pom.xml
+++ b/qa/integration-tests/pom.xml
@@ -20,6 +20,9 @@
   <packaging>jar</packaging>
 
   <name>Camunda QA Integration Tests</name>
+  <properties>
+    <version.maven-surefire-junit5-tree-reporter>1.4.0</version.maven-surefire-junit5-tree-reporter>
+  </properties>
 
   <dependencies>
     <dependency>
@@ -325,6 +328,20 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-failsafe-plugin</artifactId>
         <configuration>
+          <reportFormat>plain</reportFormat>
+          <consoleOutputReporter>
+            <disable>true</disable>
+          </consoleOutputReporter>
+          <statelessTestsetInfoReporter implementation="org.apache.maven.plugin.surefire.extensions.junit5.JUnit5StatelessTestsetInfoTreeReporter">
+            <printStacktraceOnError>true</printStacktraceOnError>
+            <printStacktraceOnFailure>true</printStacktraceOnFailure>
+            <printStdoutOnError>true</printStdoutOnError>
+            <printStdoutOnFailure>true</printStdoutOnFailure>
+            <printStdoutOnSuccess>false</printStdoutOnSuccess>
+            <printStderrOnError>true</printStderrOnError>
+            <printStderrOnFailure>true</printStderrOnFailure>
+            <printStderrOnSuccess>false</printStderrOnSuccess>
+          </statelessTestsetInfoReporter>
           <systemProperties>
             <property>
               <name>identity.docker.image.version</name>
@@ -333,6 +350,13 @@
           </systemProperties>
           <groups>!rdbms</groups>
         </configuration>
+        <dependencies>
+          <dependency>
+            <groupId>me.fabriciorby</groupId>
+            <artifactId>maven-surefire-junit5-tree-reporter</artifactId>
+            <version>${version.maven-surefire-junit5-tree-reporter}</version>
+          </dependency>
+        </dependencies>
       </plugin>
     </plugins>
   </build>
--- a/qa/integration-tests/src/test/java/io/camunda/it/exporter/IncidentExporterIT.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/exporter/IncidentExporterIT.java
@@ -9,7 +9,7 @@ package io.camunda.it.exporter;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import io.camunda.it.utils.BrokerWithCamundaExporterITInvocationProvider;
+import io.camunda.it.utils.BrokerITInvocationProvider;
 import io.camunda.search.entities.IncidentEntity.ErrorType;
 import io.camunda.zeebe.client.ZeebeClient;
 import io.camunda.zeebe.client.api.search.filter.IncidentFilter;
@@ -19,10 +19,14 @@ import java.util.UUID;
 import java.util.function.Consumer;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.TestTemplate;
-import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
-@ExtendWith(BrokerWithCamundaExporterITInvocationProvider.class)
 public class IncidentExporterIT {
+
+  // RDBMS doesn't support Incidents up to now
+  @RegisterExtension
+  static final BrokerITInvocationProvider PROVIDER =
+      new BrokerITInvocationProvider().withoutRdbmsExporter();
 
   @TestTemplate
   void shouldExportIncident(final TestStandaloneBroker testBroker) {
--- a/qa/integration-tests/src/test/java/io/camunda/it/utils/BrokerITInvocationProvider.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/utils/BrokerITInvocationProvider.java
@@ -45,70 +45,83 @@ import org.testcontainers.elasticsearch.ElasticsearchContainer;
  * The tests must take this into account, as the state in Zeebe and the exporter backend is not
  * reset between test cases.
  */
-public class BrokerWithCamundaExporterITInvocationProvider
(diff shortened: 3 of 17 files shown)


qa/integration-tests/src/test/java/io/camunda/it/utils/BrokerWithCamundaExporterITInvocationProvider.java
```diff
@@ -124,18 +159,49 @@ public void beforeAll(final ExtensionContext context) {
...
+                          .withAdditionalProperties(additionalProperties)
+                          .withAdditionalProfiles(additionalProfiles)
+                          .start();
+                  CLOSEABLES.add(testBroker);
+                  TEST_BROKERS.put(entry, testBroker);
+                  testBroker.awaitCompleteTopology();
+                  final var zeebeClientTestFactory =
+                      new ZeebeClientTestFactory().withUsers(registeredUsers);
+                  zeebeClientTestFactories.put(entry, zeebeClientTestFactory);
+                  CLOSEABLES.add(zeebeClientTestFactory);
```
stoerti, line 196: Duplicate code in each switch branch. Can this be generalized?

qa/integration-tests/src/test/java/io/camunda/it/utils/BrokerWithCamundaExporterITInvocationProvider.java
```diff
@@ -124,18 +159,49 @@ public void beforeAll(final ExtensionContext context) {
...
+                          .withProperty("camunda.database.type", "rdbms")
+                          .withProperty(
+                              "spring.datasource.url",
+                              "jdbc:h2:mem:testdb;DB_CLOSE_DELAY=-1;MODE=PostgreSQL")
+                          .withProperty("spring.datasource.driver-class-name", "org.h2.Driver")
+                          .withProperty("spring.datasource.username", "sa")
+                          .withProperty("spring.datasource.password", "")
+                          .withProperty("logging.level.io.camunda.db.rdbms", "DEBUG")
+                          .withProperty("logging.level.org.mybatis", "DEBUG")
+                          .withExporter("rdbms", cfg -> cfg.setClassName("-"))
```
stoerti, line 186: For CamundaExporter this config happens in TestStandaloneBroker#withCamundaExporter.

Don't know which location is the best but we should use the same schema for both

qa/integration-tests/src/test/java/io/camunda/it/utils/BrokerWithCamundaExporterITInvocationProvider.java
```diff
@@ -45,24 +45,54 @@
...
+  public BrokerWithCamundaExporterITInvocationProvider(
+      final Set<ExporterType> activeExporterTypes) {
+    this();
+    this.activeExporterTypes = new HashSet<>(activeExporterTypes);
+  }
+
+  /**
+   * Only the Camunda Exporter against ES/OS is active
+   */
+  public static BrokerWithCamundaExporterITInvocationProvider onlyCamundaExporter() {
```
stoerti, line 85: Rather use with/without builder pattern

qa/integration-tests/src/test/java/io/camunda/it/utils/BrokerITInvocationProvider.java
```diff
@@ -45,71 +45,84 @@
  * The tests must take this into account, as the state in Zeebe and the exporter backend is not
  * reset between test cases.
  */
-public class BrokerWithCamundaExporterITInvocationProvider
+public class BrokerITInvocationProvider
```
npepinpe, line 48: :thought_balloon: I guess we should find a better name at some point :smile:

zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/RdbmsExporter.java
```diff
@@ -205,7 +205,7 @@ private void initializeRdbmsPosition() {
 
   private void flushAndReschedule() {
     flushExecutionQueue();
-    controller.scheduleCancellableTask(Duration.ofSeconds(5), this::flushAndReschedule);
+    controller.scheduleCancellableTask(Duration.ofMillis(100), this::flushAndReschedule);
```
npepinpe, line 208: :question: Did you lower it for tests, e.g. so it's faster? The simpler way would be to make it configurable. Is flushing every 100ms a good idea in prod? :thinking: Admittedly I might be lacking some context here.

npepinpe (approved):

qa/integration-tests/src/test/java/io/camunda/it/utils/BrokerITInvocationProvider.java
p-wunderlich: Yes! Ideas? :D

qa/integration-tests/pom.xml
```diff
@@ -20,6 +20,9 @@
   <packaging>jar</packaging>
 
   <name>Camunda QA Integration Tests</name>
+  <properties>
+    <version.maven-surefire-junit5-tree-reporter>1.4.0</version.maven-surefire-junit5-tree-reporter>
```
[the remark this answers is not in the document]
p-wunderlich: Yes, that's exactly the idea. I wanted to test this here in this module first. :)

qa/integration-tests/src/test/java/io/camunda/it/exporter/IncidentExporterIT.java
```diff
@@ -19,11 +19,15 @@
...
-import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
-@ExtendWith(BrokerWithCamundaExporterITInvocationProvider.class)
 public class IncidentExporterIT {
 
+  // RDBMS doesn't support Incidents up to now
+  @RegisterExtension
+  static final BrokerITInvocationProvider PROVIDER =
+      new BrokerITInvocationProvider().withoutRdbmsExporter();
```
ChrisKujawa, line 29: ❓ Does it maybe make more sense only to enable it where we want to run it? I guess it would also be fewer changes? 🤔 Not sure about the compatibility yet.

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/cluster/TestStandaloneBroker.java
```diff
@@ -235,4 +235,16 @@ public TestStandaloneBroker withCamundaExporter(final String elasticSearchUrl) {
...
+
+  public TestStandaloneBroker withRdbmsExporter() {
+    withProperty("camunda.database.type", "rdbms");
+    withProperty("spring.datasource.url", "jdbc:h2:mem:testdb;DB_CLOSE_DELAY=-1;MODE=PostgreSQL");
+    withProperty("spring.datasource.driver-class-name", "org.h2.Driver");
+    withProperty("spring.datasource.username", "sa");
+    withProperty("spring.datasource.password", "");
+    withProperty("logging.level.io.camunda.db.rdbms", "DEBUG");
+    withProperty("logging.level.org.mybatis", "DEBUG");
+    withExporter("rdbms", cfg -> cfg.setClassName("-"));
```
ChrisKujawa, line 247: ❓ Is it on purpose that we are not setting the RDBMNS exporter class name?

ChrisKujawa (commented):
Thanks Patrick 👍🏼 

I think it would make sense to make the changes in the TestStandaloneCamunda to make use of the RDBMS exporter in the REST API tests, we have in the root.

For example: ProcessInstanceAndFlowNodeInstanceQueryTest

qa/integration-tests/src/test/java/io/camunda/it/exporter/IncidentExporterIT.java
p-wunderlich: My thoughts with this was that it should be default that we also test against the rdbms exporter. So it should be a manual step to deactivate the exporter. But I'm also fine with doing it the other way around, what do you mean?

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/cluster/TestStandaloneBroker.java
p-wunderlich: Yes, we don't need it because the exporter will be created as spring bean. (Was introduced with this feature: [#22446](https://github.com/camunda/camunda/issues/22446)

PR #25616, opened 2024-12-04
Default import position documents for record readers

EuroLew:
## Description

We want to make sure every record reader is completed without making assumptions if a specific record reader does not have a corresponding document in the import position index, this creates for each record reader a default import position document thereby simplifying the checking of importer completion, simply just verify all documents have completed=true.

## Related issues

relates to #24908


operate/importer/src/main/java/io/camunda/operate/zeebeimport/elasticsearch/ElasticsearchRecordsReader.java
```diff
@@ -144,6 +144,16 @@ private void postConstruct() {
...
         new BackoffIdleStrategy(operateProperties.getImporter().getReaderBackoff(), 1.2f, 10_000);
+
+    try {
+      final var latestPosition =
+          importPositionHolder.getLatestScheduledPosition(
+              importValueType.getAliasTemplate(), partitionId);
+
+      importPositionHolder.recordLatestLoadedPosition(latestPosition);
+    } catch (final IOException e) {
+      LOGGER.error("Failed to write initial import position index document", e);
```
ChrisKujawa, line 155: :x: Please add more context to the mesaage, like the value type for example

operate/importer/src/main/java/io/camunda/operate/zeebeimport/opensearch/OpensearchRecordsReader.java
```diff
@@ -144,6 +144,16 @@ private void postConstruct() {
...
+
+    try {
+      final var latestPosition =
+          importPositionHolder.getLatestScheduledPosition(
+              importValueType.getAliasTemplate(), partitionId);
+
+      importPositionHolder.recordLatestLoadedPosition(latestPosition);
+    } catch (final IOException e) {
+      LOGGER.error("Failed to write initial import position index document", e);
+    }
```
ChrisKujawa, line 156: Same here

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
```diff
@@ -304,6 +304,26 @@ final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 1);
         .until(() -> isRecordReaderIsCompleted("1-process-instance"));
   }
 
+  @Test
+  public void shouldWriteDefaultEmptyDefaultImportPositionDocumentsOnRecordReaderStart() {
```
ChrisKujawa, line 308: :x: I would like to see also an test where it doesnt resets the positions to zero again.

E.g. you have importer running it has set value x. Stop. Start again the same position should be persisted. 

Additionally, we could verify for 8.6 not all position document were existing but with 8.7 they are. We can do it in a follow up update test i guess.

ChrisKujawa (changes requested):
Thanks euro, see my comment below.

tasklist/importer/src/main/java/io/camunda/tasklist/zeebeimport/RecordsReaderAbstract.java
```diff
@@ -82,6 +83,20 @@ public RecordsReaderAbstract(
...
+
+      importPositionHolder.recordLatestLoadedPosition(latestPosition);
+    } catch (final IOException e) {
+      LOGGER.error(
+          "Failed to write initial import position index document for value type [{}] and partition [{}]",
+          importValueType,
+          partitionId,
+          e);
+    }
   }
```
ChrisKujawa, line 100: ❓ Why we have this here again? 🤔

ChrisKujawa: Ah this is tasklist 👍🏼

ChrisKujawa (approved):
After we have verified that we are not overriding existing positions or reset them I think we are good. Pre-approving. Please add this test first.

operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
```diff
@@ -333,6 +333,28 @@ final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.6.0", 1);
     assertImportPositionMatchesRecord(record, ImportValueType.PROCESS_INSTANCE, 1);
   }
 
+  @Test
+  public void shouldWriteDefaultEmptyDefaultImportPositionDocumentsOnRecordReaderStart() {
+    recordsReaderHolder.getAllRecordsReaders().stream()
+        .map(ElasticsearchRecordsReader.class::cast)
+        .forEach(ElasticsearchRecordsReader::postConstruct);
```
mustafadagher, line 340: ❌ should not be needed. if the beans are marked as components, spring will call the `@PostConstruct` annotated methods after initializing the bean.

If something is not working correctly, this needs to be investigated rather than worked around. there's a possibility that it won't work in production if it's not working in (a spring managed) IT.

EuroLew: The beans are only instantiated once for the test (which is what we expect the default behaviour to be) changing this would require deeper changes to how all operate ITs are run which we take a lot longer.

EuroLew: if you try it yourself, put a breakpoint in the post construct function and debug the entire `FinishedImportingIT` file you will see for each record reader is only constructed once

mustafadagher: would annotating the test method you need a clean context for with `@DirtiesContext(methodMode = MethodMode.BEFORE_METHOD)` help?

EuroLew: It also resets the operate properties and causes cascading issues because other components also rely on the properties to access the indices (mainly the prefix values). Will just move on for now

PR #25623, opened 2024-12-04
feat: create default tenant on startup

### LINKED ISSUE #25610: Create the `<default>` tenant on startup
<h2>Description</h2>

We should create the `<default>` tenant on startup. Zeebe's state always stores items by tenant where necessary. If multi-tenancy is disabled and no tenant is available we will store the data using the `<default>` tenant.
Now that Zeebe becomes responsible for tenant management we should create this tenant and export an event so other components know that this tenant exists.

The tenant record should be created in the `IdentitySetupInitializer`. This initializer sends a `IdentitySetup.INITIALIZE` command which is processed in the `IdentitySetupInitializeProcessor`. Inside this processor we need to write the `Tenant.CREATED` with the provided record.
The `IdentitySetupRecord` needs to be extended with an additional property to store the `TenantRecord` in here.

------

Open questions:
- Should the default user be assigned to the default tenant? If this is the case we can tackle it in a follow-up issue.

lenaschoenburg:
Adds the default tenant `<default>` to the identity setup.
The default user and role are not associated with that tenant in any way, that's pending clarification: https://github.com/camunda/camunda/issues/25610#issuecomment-2517344335

Closes #25610

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/IdentitySetupInitializer.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/IdentitySetupInitializer.java
@@ -10,13 +10,16 @@ package io.camunda.zeebe.engine.processing.user;
 import io.camunda.zeebe.engine.EngineConfiguration;
 import io.camunda.zeebe.engine.Loggers;
 import io.camunda.zeebe.engine.state.immutable.RoleState;
+import io.camunda.zeebe.engine.state.immutable.TenantState;
 import io.camunda.zeebe.engine.state.immutable.UserState;
 import io.camunda.zeebe.engine.state.mutable.MutableProcessingState;
 import io.camunda.zeebe.protocol.Protocol;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.IdentitySetupRecord;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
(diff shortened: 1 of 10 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/IdentitySetupInitializer.java
```diff
@@ -33,12 +36,15 @@ public final class IdentitySetupInitializer implements StreamProcessorLifecycleA
   public static final String DEFAULT_USER_PASSWORD = "demo";
   public static final String DEFAULT_USER_EMAIL = "demo@demo.com";
   public static final String DEFAULT_ROLE_NAME = "Admin";
+  public static final String DEFAULT_TENANT_ID = TenantOwned.DEFAULT_TENANT_IDENTIFIER;
+  public static final String DEFAULT_TENANT_NAME = "Default";
```
lenaschoenburg, line 40: :thought_balloon: As you can see I've used the full might of my creative powers on this one. Not sure if the name needs to be discussed somewhere.

remcowesterhoud: You actually guessed exactly what's used in the old Identity. So it's perfect 😄

remcowesterhoud (approved):
Thanks @lenaschoenburg , nothing to add from my side

PR #25634, opened 2024-12-05
feat: verify permissions of groups

### LINKED ISSUE #24403: Zeebe can verify permissions of Groups
<h2>Description</h2>

When we get a user's authorized resource identifiers from the state, we should also get the authorized resource identifiers from all the user's groups. If the user is assigned to a groups that means it has all the authorizations that belong to the groups.

lenaschoenburg:
Includes the user's group authorizations when checking for permissions.
Tests in `AuthorizationCheckBehaviorTest` are extended.

I've also included a minor performance improvement and a small refactoring to prefer functional composition but feel free to oppose them, they are optional and can be dropped.

Closes #24403

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
@@ -63,13 +63,14 @@ public final class AuthorizationCheckBehavior {
       return true;
     }
 
-    Set<String> authorizedResourceIdentifiers = new HashSet<>();
-
+    final Stream<String> authorizedResourceIdentifiers;
     final var userKey = getUserKey(request);
     if (userKey.isPresent()) {
       authorizedResourceIdentifiers =
           getUserAuthorizedResourceIdentifiers(
               userKey.get(), request.getResourceType(), request.getPermissionType());
+    } else {
+      authorizedResourceIdentifiers = Stream.empty();
     }
 
     // Check if authorizations contain a resource identifier that matches the required resource
@@ -93,7 +94,8 @@ public final class AuthorizationCheckBehavior {
     }
 
     return getUserAuthorizedResourceIdentifiers(
-        userKey.get(), request.getResourceType(), request.getPermissionType());
+            userKey.get(), request.getResourceType(), request.getPermissionType())
+        .collect(Collectors.toSet());
   }
 
   public Set<String> getAuthorizedResourceIdentifiers(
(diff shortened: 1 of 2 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -147,6 +154,19 @@ private Set<String> getRoleAuthorizedResourceIdentifiers(
...
+      final AuthorizationResourceType resourceType,
+      final PermissionType permissionType) {
+    return groupKeys.stream()
+        .flatMap(
+            groupKey ->
+                authorizationState
+                    .getResourceIdentifiers(groupKey, resourceType, permissionType)
+                    .stream())
+        .collect(Collectors.toSet());
+  }
```
remcowesterhoud: 💭 🔧  I suppose we don't even need a specific group one. This method is an exact dupe of the `getRoleAuthorizedResourceIdentifiers` method. We could combine them in a single generic method.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -63,13 +63,14 @@ public boolean isAuthorized(final AuthorizationRequest request) {
...
-
+    final Stream<String> authorizedResourceIdentifiers;
     final var userKey = getUserKey(request);
     if (userKey.isPresent()) {
       authorizedResourceIdentifiers =
           getUserAuthorizedResourceIdentifiers(
               userKey.get(), request.getResourceType(), request.getPermissionType());
+    } else {
+      authorizedResourceIdentifiers = Stream.empty();
     }
```
remcowesterhoud, line 74:
```suggestion
    Stream<String> authorizedResourceIdentifiers = Stream.empty();
    final var userKey = getUserKey(request);
    if (userKey.isPresent()) {
      authorizedResourceIdentifiers =
          getUserAuthorizedResourceIdentifiers(
              userKey.get(), request.getResourceType(), request.getPermissionType());
    }
```

🔧 We can get rid of the else block here

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -102,23 +104,27 @@ public Set<String> getAuthorizedResourceIdentifiers(
...
 
-  private Set<String> getUserAuthorizedResourceIdentifiers(
+  private Stream<String> getUserAuthorizedResourceIdentifiers(
       final long userKey,
       final AuthorizationResourceType resourceType,
       final PermissionType permissionType) {
     final var userOptional = userState.getUser(userKey);
     if (userOptional.isEmpty()) {
-      return new HashSet<>();
+      return Stream.of();
```
remcowesterhoud, line 126:
```suggestion
      return Stream.empty();
```

🤡

remcowesterhoud (approved):
Thanks, I like the refactorings 👍 Just a few minor comments.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
lenaschoenburg: That was actually a conscious choice because the var can be final if it is immediately initialized in both the if and the else branch.

remcowesterhoud: I figured it might've been. I have not strong preference for one over the other tbh

PR #25635, opened 2024-12-05
Properly copy Identity related records

### LINKED ISSUE #25620: Identity records are not copied properly
Identity related records such as:
- `UserRecord`
- `TenantRecord`

implement a `wrap` method such as this:

https://github.com/camunda/camunda/blob/1e4f427b9d6c1ad63b2753e56ec021f7b7cda1a4/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/user/UserRecord.java#L40-L47

This in itself would be fine but unfortunately it is used as if this would do a `copy`.

For example, `PersistedUser#copy` ends up eventually using `UserRecord#wrap`.

Additionally, `TenantRecord` defines it's own `copy` method that is broken in the same way.

remcowesterhoud:
## Description

Replaces the wrap methods with a proper copy. This means the internal buffers actually points to different parts of memory.
I've also found a "bug" in retrieving the tenant from the state. I've fixed this in the second commit.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25620

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
@@ -100,7 +100,7 @@ public class PersistedTenant extends UnpackedObject implements DbValue {
    *
    * @param tenantRecord the TenantRecord from which to copy the data
    */
-  public void copy(final TenantRecord tenantRecord) {
+  public void from(final TenantRecord tenantRecord) {
     tenantKeyProp.setValue(tenantRecord.getTenantKey());
     tenantIdProp.setValue(tenantRecord.getTenantId());
     nameProp.setValue(tenantRecord.getName());
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
@@ -101,6 +101,8 @@ public class PersistedTenant extends UnpackedObject implements DbValue {
    * @param tenantRecord the TenantRecord from which to copy the data
    */
   public void copy(final TenantRecord tenantRecord) {
-    copyFrom(tenantRecord);
+    tenantKeyProp.setValue(tenantRecord.getTenantKey());
+    tenantIdProp.setValue(tenantRecord.getTenantId());
+    nameProp.setValue(tenantRecord.getName());
   }
 }
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
@@ -27,6 +27,12 @@ public class PersistedTenant extends UnpackedObject implements DbValue {
     declareProperty(nameProp);
   }
 
+  public PersistedTenant copy() {
+    final var copy = new PersistedTenant();
+    copy.copyFrom(this);
+    return copy;
+  }
+
   /**
    * Gets the tenant key.
    *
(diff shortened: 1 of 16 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
```diff
@@ -94,9 +100,7 @@ public PersistedTenant setName(final String name) {
    *
    * @param tenantRecord the TenantRecord from which to copy the data
    */
-  public void wrap(final TenantRecord tenantRecord) {
-    tenantKeyProp.setValue(tenantRecord.getTenantKey());
-    tenantIdProp.setValue(tenantRecord.getTenantId());
-    nameProp.setValue(tenantRecord.getName());
+  public void copy(final TenantRecord tenantRecord) {
+    copyFrom(tenantRecord);
```
lenaschoenburg, line 104: :x: This looks wrong to me, you are copying a `TenantRecord` into a `PersistedTenant`. They are not the same, they have different properties. This is probably not _completely_ broken because they both share some property names but it's conceptually wrong.

lenaschoenburg (changes requested):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
remcowesterhoud: You are right, good catch. I've reverted this.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/tenant/PersistedTenant.java
```diff
@@ -94,7 +100,7 @@ public PersistedTenant setName(final String name) {
    *
    * @param tenantRecord the TenantRecord from which to copy the data
    */
-  public void wrap(final TenantRecord tenantRecord) {
+  public void copy(final TenantRecord tenantRecord) {
```
lenaschoenburg, line 103: :upside_down_face: I'd argue that's more of a `from` than a `copy`

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/MappingRecord.java
```diff
@@ -24,10 +24,8 @@ public MappingRecord() {
     declareProperty(mappingKeyProp).declareProperty(claimNameProp).declareProperty(claimValueProp);
   }
 
-  public void wrap(final MappingRecord record) {
-    mappingKeyProp.setValue(record.getMappingKey());
-    claimNameProp.setValue(record.getClaimName());
-    claimValueProp.setValue(record.getClaimValue());
+  public void copy(final MappingRecord record) {
+    copyFrom(record);
```
lenaschoenburg, line 28: :thought_balloon: This naming now conflicts with the _other_ `copy` methods that return a new copy of itself. Since _this_ `copy` is just an alias for `copyFrom` which has a clearer name, could we just delete this and similar methods and just use `copyFrom` directly?

lenaschoenburg (approved):
Looks good!

Using `PersistedTenant` directly makes sense and looks cleaner :+1:

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/MappingRecord.java
remcowesterhoud: I'll jsut remove it, it's not even used 🤡

lenaschoenburg: (Same in other records of course)

PR #25636, opened 2024-12-05
feat: log indices in history-backup at startup

entangled90:
Log what indices are part of backups at startup

### DIFF
--- a/dist/src/main/java/io/camunda/application/commons/backup/HistoryBackupComponent.java
+++ b/dist/src/main/java/io/camunda/application/commons/backup/HistoryBackupComponent.java
@@ -17,6 +17,8 @@ import io.camunda.webapps.schema.descriptors.backup.Prio2Backup;
 import io.camunda.webapps.schema.descriptors.backup.Prio3Backup;
 import io.camunda.webapps.schema.descriptors.backup.Prio4Backup;
 import java.util.List;
+import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


dist/src/main/java/io/camunda/application/commons/backup/HistoryBackupComponent.java
```diff
@@ -44,6 +47,10 @@ public HistoryBackupComponent(
       final List<Prio4Backup> prio4BackupIndices,
       final BackupRepositoryProps backupRepositoryProps,
       final BackupRepository backupRepository) {
+    LOG.info("Prio1BackupIndices are {}", prio1BackupIndices);
+    LOG.info("Prio2BackupTemplates are {}", prio2BackupTemplates);
+    LOG.info("Prio3BackupTemplates are {}", prio3BackupTemplates);
+    LOG.info("Prio4BackupIndices are {}", prio4BackupIndices);
```
npepinpe, line 53: :x: I don't think this is really [info logging level](https://github.com/camunda/camunda/wiki/Logging). 

> INFO: information about the system which is useful for the user (note: in the case of the broker, user here is the SRE deploying it), e.g. leader changes, a new node added to or removed from the membership, etc.

This is pretty much implementation details as far as the user is concerned, so this would be DEBUG imo

npepinpe (approved):
No need for another review unless you disagree :)

Issue #25644, opened 2024-12-05
Same-site auth cookie is not being attached to the home page request after successful authentication


buccarel:
Engineering DRI: @buccarel

### Describe the bug

When authenticating against Identity, Optimize redirects the user to the home page using a see-other (HTTP 303) automatic redirection. This causes the same-site auth cookie not to be attached to the navigation to the home page.

This behavior is described in the [official draft of the RFC 6265 bis, in the last paragraph of section 4.1.2.7](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-4.1.2.7):

> The "SameSite" attribute affects cookie creation as well as delivery. Cookies which assert "SameSite=Lax" or "SameSite=Strict" cannot be set in responses to cross-site subresource requests, or cross-site nested navigations. They can be set along with any top-level navigation, cross-site or otherwise.

In particular, the sentence "Cookies... cannot be set in responses to cross-site subresource requests, or cross-site nested navigations" extends the definition of same-site navigation, effectively limiting it to requests initiated by the users, such as typing the URL in the top bar and pressing enter or clicking a link with their mouse. As automatic redirections are detected as "nested navigation" mechanisms, they do not attach the cookie.

The absence of the auth cookie results in the home page looping back to the authentication system, triggering an infinite loop that is stopped by the browser after a certain number of retries.

### Expected behaviour

The home page should be loaded after a successful authentication against Identity.

### To Reproduce

The detailed steps on how to reproduce the bug can be found in the support ticket https://jira.camunda.com/browse/SUPPORT-24651

### Potential solution

There is more than one solution to the problem. Two effective ways to solve the problems are the following:
* instead of redirecting to the home page, directly serve the index.html file
* instead of redirecting to the home page via see-other, redirect to a web page that performs the javascript operation
```javascript
window.location = targetUrl;
```
where targetUrl is the desired home page. This solution would work because the modification of `window.location` is interpreted similarly as the user typing the URL in the address bar of a browser.

### Environment observed:

- Optimize mode(s): self-managed
- Optimize version(s): 8.5, 8.6
- Database(s) (Elasticsearch/OpenSearch): n/a

### Links

* https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-4.1.2.7
* https://jira.camunda.com/browse/SUPPORT-24651

### Breakdown

```[tasklist]
### Pull Requests
[...]

PR #25647, opened 2024-12-05
bugfix: double-redirection to home page to fix auth issue in optimize

buccarel:
## Description

Context:
https://github.com/camunda/camunda/issues/25644

## Checklist

- [ ] for CI changes:

## Related issues

closes #


.gitignore
```diff
@@ -39,7 +39,6 @@ faces-config.NavData
 .gradle
 .worktree
 work/
-public/
```
RomanJRW, line 42: This should not be removed?

RomanJRW (changes requested):
Your commit message should just be `fix:...` rather than `bugfix:...`

.gitignore
buccarel: I need too add the new html file in a directory that is called public. Maybe I can add the file bypassing the gitignore then?

RomanJRW: maybe it's ok if nothing else is getting committed by removing this line

RomanJRW (approved):

Issue #24657, opened 2024-11-12
Build foundational test suite for Optimize 8


RomanJRW:
### Context

In Optimize 7, the codebase had a large number of ITs that gave a high level of confidence during development of new or changing features. However, with the move to C8, these tests were never rewritten for Optimize 8. The reasons for this include:
- Optimize 8 could not perform data import from Zeebe in a performant way
- There was no easy way to mock the identity integration to handle Optimize authentication/authorization
- Not all functionality was supported in Optimize 8 (although that gap has closed over time)

Given that the Optimize importer is planned to be extracted, and that Identity changes are happening, we acknowledge that a rewrite of these tests is no longer possible and we need to revisit the testing structure in from scratch.

Therefore, we would like to in this issue:
- Remove all of the existing test infrastructure from Optimize 8 that is not used
- Delete all of the IT classes that are commented out or disabled
- Identity the most important parts of the codebase and create unit tests for these classes

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/24875
- [ ] https://github.com/camunda/camunda/pull/25485
- [ ] https://github.com/camunda/camunda/pull/25484
- [ ] https://github.com/camunda/camunda/pull/25536
- [ ] https://github.com/camunda/camunda/pull/25608
- [ ] https://github.com/camunda/camunda/pull/25651
- [ ] https://github.com/camunda/camunda/pull/25755
- [ ] https://github.com/camunda/camunda/pull/25832
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [ ] The change is implemented as described on all target environments/versions/modes
- [ ] The documentation changes are as expected

### Completion

- [ ] All Review stages are successfully completed
- [ ] All associated PRs are merged to the main branch(es) and stable branches
- [ ] The correct version labels are applied to the issue

RomanJRW:
@KarpovIvan is moving focus to another project, so we have agreed to shift focus away from this. We will close this as complete, while understanding that the endeavour for increased test coverage will continue as we make changes to the codebase

PR #25651, opened 2024-12-05
feat: ValidationHelperTest test

KarpovIvan:
## Description

task: [24657](https://github.com/camunda/camunda/issues/24657)

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
+++ b/optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
@@ -11,9 +11,12 @@ import io.camunda.optimize.dto.optimize.query.analysis.BranchAnalysisRequestDto;
 import io.camunda.optimize.dto.optimize.query.report.ReportDataDto;
 import io.camunda.optimize.dto.optimize.query.report.single.configuration.AggregationDto;
 import io.camunda.optimize.dto.optimize.query.report.single.configuration.AggregationType;
+import io.camunda.optimize.dto.optimize.query.report.single.filter.data.operator.MembershipFilterOperator;
+import io.camunda.optimize.dto.optimize.query.report.single.process.filter.AssigneeFilterDto;
 import io.camunda.optimize.dto.optimize.query.report.single.process.filter.FilterApplicationLevel;
 import io.camunda.optimize.dto.optimize.query.report.single.process.filter.InstanceStartDateFilterDto;
 import io.camunda.optimize.dto.optimize.query.report.single.process.filter.ProcessFilterDto;
+import io.camunda.optimize.dto.optimize.query.report.single.process.filter.data.IdentityLinkFilterDataDto;
 import io.camunda.optimize.service.exceptions.OptimizeValidationException;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class ValidationHelperTest {
+
+  @Test
+  void shouldThrowExceptionWhenDtoIsInvalid() {
```
RomanJRW, line 27: what about testing a valid DTO?

optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,113 @@
...
+    // given
+    final ProcessFilterDto<?> validFilter = new InstanceStartDateFilterDto();
+    validFilter.setFilterLevel(FilterApplicationLevel.INSTANCE);
+
+    // when
+    final Throwable thrown = Assertions.catchThrowable(() -> ValidationHelper.validateProcessFilters(List.of(validFilter)));
+
+    // then
+    Assertions.assertThat(thrown)
+        .isInstanceOf(NullPointerException.class);
```
RomanJRW, line 81: this is kind of not testing what the method really does. A more correct test would be to ensure that the parameter is invalid rather than null. In these cases, it would throw a validation exception rather than an NPE.

That said, I don't think that this class is worth so much of your time, so the baseline here is already an improvement even if coverage is less than 100%

RomanJRW (approved):
Tentatively pre-approving, but accepting that this is only a low baseline of coverage rather than complete coverage for the class and its methods. I think this is acceptable

optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
KarpovIvan: fixed

optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class ValidationHelperTest {
+
+  @Test
+  void shouldBeSuccessfulWhenDtoCorrect() {
```
RomanJRW, line 30:
```suggestion
  void shouldBeSuccessfullyValidatedWhenDtoHasValidFieldValues() {
```

optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
```diff
@@ -0,0 +1,137 @@
...
+    final ProcessFilterDto<?> validFilter = new InstanceStartDateFilterDto();
+    validFilter.setFilterLevel(FilterApplicationLevel.INSTANCE);
+
+    // when
+    final Throwable thrown =
+        Assertions.catchThrowable(
+            () -> ValidationHelper.validateProcessFilters(List.of(validFilter)));
+
+    // then
+    Assertions.assertThat(thrown).isInstanceOf(NullPointerException.class);
```
RomanJRW, line 105: this assertion does not match the test name at all. It's also testing incorrect behaviour

RomanJRW (changes requested):

optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
KarpovIvan: fixed

optimize/backend/src/test/java/io/camunda/optimize/service/util/ValidationHelperTest.java
KarpovIvan: fixed

RomanJRW (approved):
We can accept this as good enough to ensure progress and allow movement to a higher priority topic. This adds some missing tests, which is better than having no coverage

PR #25653, opened 2024-12-05
Simplify DbAuthorizationState

remcowesterhoud:
## Description

Before we stored the resource identifiers by owner key, resource type and permission type. This becomes quite complex to understand. The whole domain becomes a lot simpler if we work with permissions for an owner for a specific resource
 type.

 Doing this also future proofs us. We'll likely want to check multiple permission types for a resource in the
 not-so-distant future. By already storing it like this now we prevent having to do a nasty migration later.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -58,26 +51,20 @@ public DbAuthorizationState(
...
+    ownerKeyAndResourceType = new DbCompositeKey<>(ownerKey, resourceType);
 
-    resourceIdsByOwnerKeyResourceTypeAndPermissionColumnFamily =
+    permissionsColumnFamily =
         zeebeDb.createColumnFamily(
-            ZbColumnFamilies.RESOURCE_IDS_BY_OWNER_KEY_RESOURCE_TYPE_AND_PERMISSION,
-            transactionContext,
-            ownerKeyAndResourceTypeAndPermissionCompositeKey,
-            resourceIdentifiers);
+            ZbColumnFamilies.PERMISSIONS, transactionContext, ownerKeyAndResourceType, permissions);
```
koevskinikola, line 58: 🔧 Should we pass a separate `Permissions` instance here, to avoid the read/write buffer re-use bug?

koevskinikola (approved):
LGTM @remcowesterhoud 👍 

I only have a small note below, maybe it's not critical for the `DbAuthorizationState`.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
remcowesterhoud: Why would we run into that bug here? Isn't this how always do it? E.g. in [the DbProcessState](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/deployment/DbProcessState.java#L125-L130)

PR #25654, opened 2024-12-05
Add corrections to complete job REST endpoint

### LINKED ISSUE #25447: Add `corrections` to complete job REST endpoint
Add corrections to complete job REST endpoint.
```[tasklist]
### Tasks
- [x] Update rest-api.yaml
- [x] Add `corrections` to RequestMapper
- [x] Add relevant tests to `JobControllerTest`
- [x] Add relevant tests to Complete Job Test
- [x] Add support for `correctedAttributes`
```

ana-vinogradova-camunda:
## Description

Adding corrections to complete job REST endpoint:

- Updating rest-api.yaml
- Adding `corrections` to RequestMapper
- Adding relevant tests

## Checklist

- [ ] for CI changes:

## Related issues

closes #25447

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -5646,7 +5646,6 @@ components:
           format: int32
           description: The priority of the task.
           minimum: 0
-          default: 50
           maximum: 100
           nullable: true
     JobUpdateRequest:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -5609,6 +5609,46 @@ components:
             As a result, the completion request is rejected and the task remains active.
             Defaults to false.
           nullable: true
+        corrections:
+          $ref: "#/components/schemas/JobResultCorrections"
+    JobResultCorrections:
+      type: object
+      description: Attributes that were corrected by the worker. Others are considered not corrected.
+      nullable: true
+      properties:
+        assignee:
+          type: string
+          description: Assignee of the task.
+        dueDate:
+          type: string
+          format: date-time
+          description: The due date of the task.
+          nullable: true
+        followUpDate:
+          type: string
+          format: date-time
+          description: The follow-up date of the task.
+          nullable: true
+        candidateUsers:
+          type: array
+          description: The list of candidate users of the task.
+          items:
+            type: string
+          nullable: true
+        candidateGroups:
+          type: array
+          description: The list of candidate groups of the task.
+          items:
+            type: string
+          nullable: true
+        priority:
+          type: integer
+          format: int32
+          description: The priority of the task.
+          minimum: 0
+          default: 50
+          maximum: 100
+          nullable: true
     JobUpdateRequest:
       type: object
       properties:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -5611,12 +5611,6 @@ components:
           nullable: true
         corrections:
           $ref: "#/components/schemas/JobResultCorrections"
-        correctedAttributes:
-          type: array
-          description: List of the attributes that were corrected by the worker.
-          items:
-            type: string
-          nullable: true
     JobResultCorrections:
       type: object
       description: Attributes that were corrected by the worker. Others are considered not corrected.
@@ -5635,13 +5629,13 @@ components:
           format: date-time
           description: The follow-up date of the task.
           nullable: true
-        candidateUsers:
+        candidateUsersList:
           type: array
           description: The list of candidate users of the task.
           items:
             type: string
           nullable: true
-        candidateGroups:
+        candidateGroupsList:
           type: array
           description: The list of candidate groups of the task.
           items:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -5613,12 +5613,26 @@ components:
           $ref: "#/components/schemas/JobResultCorrections"
     JobResultCorrections:
       type: object
-      description: Attributes that were corrected by the worker. Others are considered not corrected.
+      description: |
+        JSON object with attributes that were corrected by the worker.
+
+        The following attributes can be corrected, additional attributes will be ignored:
+
+        * `assignee` - reset by providing an empty String
+        * `dueDate` - reset by providing an empty String
+        * `followUpDate` - reset by providing an empty String
+        * `candidateGroups` - reset by providing an empty list
+        * `candidateUsers` - reset by providing an empty list
+        * `priority` - minimum 0, maximum 100, default 50
+
+        Providing any of those attributes with a `null` value or omitting it preserves
+        the persisted attribute's value.
       nullable: true
       properties:
         assignee:
           type: string
           description: Assignee of the task.
+          nullable: true
         dueDate:
           type: string
           format: date-time
@@ -5629,13 +5643,13 @@ components:
           format: date-time
           description: The follow-up date of the task.
           nullable: true
-        candidateUsersList:
+        candidateUsers:
           type: array
           description: The list of candidate users of the task.
           items:
             type: string
           nullable: true
-        candidateGroupsList:
+        candidateGroups:
           type: array
           description: The list of candidate groups of the task.
           items:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -5611,6 +5611,12 @@ components:
           nullable: true
         corrections:
           $ref: "#/components/schemas/JobResultCorrections"
+        correctedAttributes:
+          type: array
+          description: List of the attributes that were corrected by the worker.
+          items:
+            type: string
+          nullable: true
     JobResultCorrections:
       type: object
       description: Attributes that were corrected by the worker. Others are considered not corrected.
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
@@ -738,11 +738,11 @@ public class RequestMapper {
       correctedAttributes.add(UserTaskRecord.FOLLOW_UP_DATE);
     }
     if (jobResultCorrections.getCandidateUsers() != null) {
-      corrections.setCandidateUsers(jobResultCorrections.getCandidateUsers());
+      corrections.setCandidateUsersList(jobResultCorrections.getCandidateUsers());
       correctedAttributes.add(UserTaskRecord.CANDIDATE_USERS);
     }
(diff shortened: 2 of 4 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5268,6 +5268,51 @@ components:
             As a result, the completion request is rejected and the task remains active.
             Defaults to false.
           nullable: true
+        corrections:
+          $ref: "#/components/schemas/JobResultCorrections"
+        correctedAttributes:
+          type: array
+          description: List of the attributes that were corrected by the worker.
```
ce-dmelnych, line 5275: As outlined in https://github.com/camunda/camunda/pull/25654/files#r1879850809 we might not need to expose `correctedAttributes` to consume it via the REST API.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5268,6 +5268,51 @@ components:
...
+          items:
+            type: string
+          nullable: true
```
ce-dmelnych, line 5278: As noted in https://github.com/camunda/camunda/pull/25654/files#r1879850809, exposing `correctedAttributes` in the REST API may not be necessary.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -725,8 +726,31 @@ private static JobResult getJobResultOrDefault(final JobCompletionRequest reques
...
+        getStringOrEmpty(request, r -> r.getResult().getCorrections().getDueDate()));
+    corrections.setFollowUpDate(
+        getStringOrEmpty(request, r -> r.getResult().getCorrections().getFollowUpDate()));
+    corrections.setCandidateGroups(
+        getStringListOrEmpty(request, r -> r.getResult().getCorrections().getCandidateGroups()));
+    corrections.setCandidateUsers(
+        getStringListOrEmpty(request, r -> r.getResult().getCorrections().getCandidateUsers()));
+    corrections.setPriority(
+        getIntOrDefault(request, r -> r.getResult().getCorrections().getPriority(), -1));
+    return jobResult.setCorrections(corrections);
```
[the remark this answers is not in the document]
ana-vinogradova-camunda: 👍 I agree with the above.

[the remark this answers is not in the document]
ana-vinogradova-camunda: 💭 @ce-dmelnych @korthout can it be the case that user explicitly sets the value of the corrected field, for example, `assignee` to be empty and to override the existing `assignee` value?

Alternatively: potentially the list of `correctedAttributes` can be calculated on a java client side. Ex.: when user sets some correction the client also records this attribute as corrected.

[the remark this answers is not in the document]
korthout: >can it be the case that user explicitly sets the value of the corrected field, for example, assignee to be empty and to override the existing assignee value?

@ana-vinogradova-camunda Yes. This should be aligned with the User Task update and the Job update. They both allow using `null` to mean unchanged and a value to mean changed to. For strings, an empty string `""` is allowed to intent unassign.

>Alternatively: potentially the list of correctedAttributes can be calculated on a java client side. Ex.: when user sets some correction the client also records this attribute as corrected.

The API doesn't need the `correctedAttributes`. This field only exists because the records in our protocol don't allow `null` values, but the API does. So we can calculate the `correctedAttributes` in the gateway, similar to example shared above by Dima.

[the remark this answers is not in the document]
ce-dmelnych: 💭 For such cases, the user can provide a default value (e.g for the `assignee` property -> empty string value `""`) to reset it to the default via corrections. We can highlight this approach in the documentation (Camunda/REST-API). 

Additionally, on the Zeebe Java client side, we could introduce a convenient `reset<PropertyName>()` method (e.g `resetAssignee()`) alongside the ordinary `assignee(final String assignee)` builder method. This method would handle the default (empty string) setting logic internally, making it easier for users to reset fields.

What do you think about this approach, @ana-vinogradova-camunda, @korthout?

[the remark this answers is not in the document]
korthout: >resetAssignee()) alongside the ordinary assignee(final String assignee) builder method. This method would handle the default (empty string) setting logic internally, making it easier for users to reset fields.

@ce-dmelnych I worry that `reset` would be interpreted as rollback to before the transition. We currently don't support this conveniently, as the listener doesn't have access to the previous value, nor is any api to specify data rollbacks. It would be an interesting addition, when we add such functionality.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/JobControllerTest.java
```diff
@@ -347,6 +350,130 @@ void shouldCompleteJobWithResultDeniedTrue() {
...
+        """
+          {
+            "result": {
+              "denied": false,
+              "corrections": {
+                "assignee": "Test",
+                "candidateUsersList": ["UserA", "UserB"],
+                "candidateGroupsList": ["GroupA", "GroupB"],
+                "priority": 20
+              }
```
ce-dmelnych, line 435: 🔧 To emphasize that only explicitly provided values are considered corrected, you could use `null` for some fields in the `corrections` JSON.
This would highlight how Zeebe-rest-gateway handles fields that are not updated.

For example, the test could be modified like this:

```java
  @Test
  void shouldCompleteJobWithResultWithCorrectionsPartiallySet() {
    // given
    when(jobServices.completeJob(anyLong(), any(), any()))
        .thenReturn(CompletableFuture.completedFuture(new JobRecord()));

    final var request =
        """
          {
            "result": {
              "denied": false,
              "corrections": {
                "assignee": null,
                "dueDate": "2025-05-23T01:02:03+01:00",
                "candidateGroupsList": ["GroupA", "GroupB"],
                "priority": null
              }
            }
          }
        """;

    // when/then
    webClient
        .post()
        .uri(JOBS_BASE_URL + "/1/completion")
        .accept(MediaType.APPLICATION_JSON)
        .contentType(MediaType.APPLICATION_JSON)
        .bodyValue(request)
        .exchange()
        .expectStatus()
        .isNoContent();

    final var jobResultArgumentCaptor = ArgumentCaptor.forClass(JobResult.class);
    Mockito.verify(jobServices)
        .completeJob(eq(1L), eq(Map.of()), jobResultArgumentCaptor.capture());

    assertThat(jobResultArgumentCaptor.getValue().getCorrections())
        .isEqualTo(
            new JobResultCorrections()
                .setDueDate("2025-05-23T01:02:03+01:00")
                .setCandidateGroups(List.of("GroupA", "GroupB"))
                // The remaining fields have their default non-null values,
                // as they weren't corrected
                .setAssignee("")
                .setFollowUpDate("")
                .setCandidateUsers(List.of())
                .setPriority(-1));

    assertThat(jobResultArgumentCaptor.getValue().getCorrectedAttributes())
        .containsExactly(UserTaskRecord.DUE_DATE, UserTaskRecord.CANDIDATE_GROUPS);
  }
```

This would make it clear which fields were corrected and which were not.
Let me know what do you think about it! 😃

ce-dmelnych (approved):
Great work on this PR, Ana! 🎉

> I've proposed a few suggestions for the test.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/JobControllerTest.java
```diff
@@ -347,6 +350,130 @@ void shouldCompleteJobWithResultDeniedTrue() {
...
+
+    Assertions.assertEquals(
+        List.of(
+            "assignee",
+            UserTaskRecord.DUE_DATE,
+            UserTaskRecord.FOLLOW_UP_DATE,
+            UserTaskRecord.CANDIDATE_USERS,
+            UserTaskRecord.CANDIDATE_GROUPS,
+            UserTaskRecord.PRIORITY),
+        jobResultArgumentCaptor.getValue().getCorrectedAttributes());
```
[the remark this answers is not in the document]
ana-vinogradova-camunda: I like this approach - easier to read and cleaner 👍

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/JobControllerTest.java
ana-vinogradova-camunda: I think it might even be a separate test.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5765,6 +5765,45 @@ components:
             As a result, the completion request is rejected and the task remains active.
             Defaults to false.
           nullable: true
+        corrections:
+          $ref: "#/components/schemas/JobResultCorrections"
+    JobResultCorrections:
+      type: object
+      description: Attributes that were corrected by the worker. Others are considered not corrected.
```
korthout, line 5772: 🔧 We could align the description to the [`UserTaskChangeset`](https://github.com/camunda/camunda/blob/9cb4ef558a5ed06391678a7af0a63af2cc01fb22/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L5519-L5536). This description is more explicit about the behavior: set to `null` to ignore, set to `""` to reset a string, etc. I think the biggest difference is that `assignee` is not allowed there.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5765,6 +5765,45 @@ components:
...
+          description: The list of candidate users of the task.
+          items:
+            type: string
+          nullable: true
+        candidateGroupsList:
+          type: array
+          description: The list of candidate groups of the task.
+          items:
+            type: string
+          nullable: true
```
korthout, line 5799: 🔧 With [`UserTaskChangeset`](https://github.com/camunda/camunda/blob/9cb4ef558a5ed06391678a7af0a63af2cc01fb22/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L5519-L5536) in mind, we should actually refer to `candidateUsers` and `candidateGroups` rather the `*List` variants. I had not considered this, when I [said here](https://github.com/camunda/camunda/pull/25663#issuecomment-2536082943) that we should use it in the API. I'd rather have consistency in the API first.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5765,6 +5765,45 @@ components:
...
+        corrections:
+          $ref: "#/components/schemas/JobResultCorrections"
+    JobResultCorrections:
+      type: object
+      description: Attributes that were corrected by the worker. Others are considered not corrected.
+      nullable: true
+      properties:
+        assignee:
+          type: string
+          description: Assignee of the task.
```
korthout, line 5777: 🔧 `assignee` should also be `nullable`

korthout (approved):
Great work @ana-vinogradova-camunda 🚀 

🔧 I have a few suggestions. Please have a look

Issue #23822, opened 2024-10-22
Generate SQL scripts from Liquibase


stoerti:
As customer DBA I want to execute the SQL migration scripts manually to have full control over DDL changes.

**Acceptance criteria**
- in maven build, liquibase generates SQL scripts for each supported RDBMS vendor and publishes them as artifact
- RDBMS liquibase migration can be deactivated via application.yaml property

PR #25659, opened 2024-12-05
feat: generate SQL scripts from liquibase #23822

stoerti:
## Checklist

- [ ] for CI changes:

## Related issues

closes #


db/rdbms-schema/src/main/java/io/camunda/db/rdbms/LiquibaseScriptGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+    final var targetDir = args[0] + "/liquibase";
+    final var databases = Set.of("h2", "mysql", "postgresql");
+
+    for (final var database : databases) {
+      generateLiquibaseScript(
+          database, "db/changelog/rdbms-support/changelog-master.xml", targetDir, "master.sql");
+
+      generateLiquibaseScript(
+          database,
+          "db/changelog/rdbms-support/8_8_0/8.8.0_master.xml",
```
p-wunderlich, line 32: Hm, I think we should discuss the chronology of the files here. When we think about further releases ...what do you think about something like that:

db/changelog/(rdbms-support*)/
- changesets/8_8_0_changsets.xml
- changesets/8_8_1_changsets.xml

* do we really need our own namespace here? rdbms-support is very generic, and we only need liquibase because of the rbdms_support. If we really have other modules sometimes, we should at least be more specifiv like rdbms_exporter.

db/rdbms-schema/src/main/java/io/camunda/db/rdbms/LiquibaseScriptGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+          targetDir,
+          "8.8.0_master.sql");
+    }
+  }
+
+  public static void generateLiquibaseScript(
```
p-wunderlich, line 38: Do we need tests for this?

db/rdbms-schema/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+            <goals>
+              <goal>java</goal>
+            </goals>
+            <phase>process-classes</phase>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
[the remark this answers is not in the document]
stoerti: I would think, that it's now simply part of the package build and also des not really consumes a lot of time (3.4s for the whole build).

Limiting it to a specific workflow would probably lead to more confusion as it saves time/power.

db/rdbms-schema/src/main/java/io/camunda/db/rdbms/LiquibaseScriptGenerator.java
stoerti: Agreed. Renamed it to rdbms-exporter. I could think of an RDBMS persistent session store (and who knows what's to come).

Future generated SQL scripts would then have a name schema like "8.8.0_to_8.8.1.sql" or similar. The master would still contain all.

db/rdbms-schema/src/main/java/io/camunda/db/rdbms/LiquibaseScriptGenerator.java
stoerti: Yes. Added a test for the Script generation part.

p-wunderlich (approved):

PR #25665, opened 2024-12-05
fix: ignore inherited permissions when adding or removing

### LINKED ISSUE #25652: Permissions should be able to be granted to an entity when the permission already exists in one of their groups/roles
Given: An Initial User member of the admin role which provide the initial user with all authorizations
When doing:

```
PATCH http://localhost:8080/v2/authorizations/2251799813685310 with the request body
{
    "action": "ADD",
    "resourceType": "PROCESS_DEFINITION",
    "permissions": [
        {
            "permissionType": "READ",
            "resourceIds": [
                "*"
            ]
        }
    ]
}
```

Then, the response is:
```
{
    "type": "about:blank",
    "title": "ALREADY_EXISTS",
    "status": 409,
    "detail": "Command 'ADD_PERMISSION' rejected with code 'ALREADY_EXISTS': Expected to add 'READ' permission for resource 'PROCESS_DEFINITION' and resource identifiers '[*]' for owner '2251799813685310', but this permission for resource identifiers '[*]' already exist. Existing resource ids are: '[*, Process_00tn8v3]'",
    "instance": "/v2/authorizations/2251799813685310"
}
```

But there does not exist a "direct" authorization that provides the initial user with the permission READ for the resource * . Only the role membership.

------

When we check for existing permissions we get all the authorized resource identifiers for the entity. We need to make sure to not include any extras that come from the role or group.

https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/PermissionsBehavior.java#L81-L86

We should make sure to add a unit test to verify this scenario.

lenaschoenburg:
We only want to reject a request to add a permission if the permission already exists directly on the entity, not if it is already inherited through an assigned role or group.

Similarly, we need to reject deletion of a permission if the the permission does not exist directly.

This exposes a new method in `AuthorizationCheckBehavior` to only get _direct_ authorizations without included inherited ones. This method is then used to check for duplicated permissions.

Depends on #25634 to avoid merge conflicts
Fixes #25652

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/PermissionsBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/PermissionsBehavior.java
@@ -79,7 +79,7 @@ public class PermissionsBehavior {
     for (final PermissionValue permission : record.getPermissions()) {
       final var addedResourceIds = permission.getResourceIds();
       final var currentResourceIds =
-          authCheckBehavior.getAuthorizedResourceIdentifiers(
+          authCheckBehavior.getAllAuthorizedResourceIdentifiers(
               record.getOwnerKey(),
               record.getOwnerType(),
               record.getResourceType(),
@@ -108,7 +108,7 @@ public class PermissionsBehavior {
       final AuthorizationRecord record) {
     for (final PermissionValue permission : record.getPermissions()) {
       final var currentResourceIdentifiers =
-          authCheckBehavior.getAuthorizedResourceIdentifiers(
+          authCheckBehavior.getAllAuthorizedResourceIdentifiers(
               record.getOwnerKey(),
               record.getOwnerType(),
               record.getResourceType(),
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/PermissionsBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/PermissionsBehavior.java
@@ -105,11 +105,8 @@ public class PermissionsBehavior {
       final AuthorizationRecord record) {
     for (final PermissionValue permission : record.getPermissions()) {
       final var currentResourceIdentifiers =
-          authCheckBehavior.getAllAuthorizedResourceIdentifiers(
-              record.getOwnerKey(),
-              record.getOwnerType(),
-              record.getResourceType(),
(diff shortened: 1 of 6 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/PermissionsBehavior.java
```diff
@@ -108,7 +105,7 @@ public Either<Rejection, AuthorizationRecord> permissionDoesNotExist(
       final AuthorizationRecord record) {
     for (final PermissionValue permission : record.getPermissions()) {
       final var currentResourceIdentifiers =
-          authCheckBehavior.getAuthorizedResourceIdentifiers(
+          authCheckBehavior.getAllAuthorizedResourceIdentifiers(
```
remcowesterhoud, line 108: 💭 This should actually become direct as well. When we remove permissions from an entity we should make sure it's actually assigned to this entity directly. Do you want to take it as part of this, or should I create a separate issue for it?

remcowesterhoud (approved):
Sweeping up all them bugs 🧹🐛🐞

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/PermissionsBehavior.java
lenaschoenburg: Ah okay, I thought about it and came to the opposite conclusion: that we would allow deleting an inherited permission. But makes sense that we don't support that and only allow deleting on the entity directly.

remcowesterhoud: It's about expectations I suppose. As a user if I want to remove permission X from user Y I wouldn't expect that user Z would also lose the same permission.

lenaschoenburg: Oh right, that could easily cause mistakes, right :+1:

lenaschoenburg: Fixed in b8db2fd

remcowesterhoud (approved):
You're the best, I'll approve it again just because 😄

PR #25677, opened 2024-12-05
feat: ElasticsearchBackupRepository uses new ElasticsearchClient instead of RestHighLevelClient

### LINKED ISSUE #25676: Migrate ElasticsearchBackupRepository to use ElasticSearchClient instead of RestHighLevelClient

entangled90:
## Description
Main differences:
- SnapshotState is now a string and not an enum (which is the state of a single part of the snapshot)
  - INCOMPATIBLE does not exist anymore in the [api](https://www.elastic.co/guide/en/elasticsearch/reference/current/create-snapshot-api.html#create-snapshot-api-request-body)
- The new snapshot client is sync and not async
- :red_circle: Some option in the snaphot creation are not available in the client (even though they can be set from REST apis) (see my comment)
- I added an executor to make each method return immediately as before

## Related issues
closes #25676

### DIFF
--- a/dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
+++ b/dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
@@ -65,10 +65,10 @@ public class BackupConfig {
   public ThreadPoolTaskExecutor getTaskExecutor() {
     final ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
     executor.setCorePoolSize(1);
-    executor.setMaxPoolSize(1);
+    executor.setMaxPoolSize(8);
+    executor.setKeepAliveSeconds(60);
     executor.setThreadNamePrefix("webapps_backup_");
-    // TODO why just 6?
-    executor.setQueueCapacity(6);
+    executor.setQueueCapacity(4096);
     executor.initialize();
     return executor;
   }
--- a/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
+++ b/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
@@ -241,7 +241,7 @@ public class ElasticsearchBackupRepository implements BackupRepository {
                   groupingBy(
                       si -> {
                         final Metadata metadata =
-                            Metadata.fromMetadataES(si.metadata(), esClient._jsonpMapper());
+                            MetadataMarshaller.fromMetadata(si.metadata(), esClient._jsonpMapper());
                         Long backupId = metadata.backupId();
                         // backward compatibility with v. 8.1
                         if (backupId == null) {
@@ -397,7 +397,7 @@ public class ElasticsearchBackupRepository implements BackupRepository {
     final GetBackupStateResponseDto response = new GetBackupStateResponseDto(backupId);
     final var firstSnapshot = snapshots.getFirst();
     final Metadata metadata =
-        Metadata.fromMetadataES(firstSnapshot.metadata(), esClient._jsonpMapper());
+        MetadataMarshaller.fromMetadata(firstSnapshot.metadata(), esClient._jsonpMapper());
     final Integer expectedSnapshotsCount = metadata.partCount();
 
     if (snapshots.size() == expectedSnapshotsCount
--- a/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
+++ b/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
@@ -7,15 +7,22 @@
  */
 package io.camunda.webapps.backup.repository.elasticsearch;
 
+import static io.camunda.webapps.backup.repository.elasticsearch.SnapshotState.*;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
-import static org.elasticsearch.snapshots.SnapshotState.FAILED;
-import static org.elasticsearch.snapshots.SnapshotState.INCOMPATIBLE;
-import static org.elasticsearch.snapshots.SnapshotState.IN_PROGRESS;
-import static org.elasticsearch.snapshots.SnapshotState.PARTIAL;
-import static org.elasticsearch.snapshots.SnapshotState.SUCCESS;
 
+import co.elastic.clients.elasticsearch.ElasticsearchClient;
+import co.elastic.clients.elasticsearch._types.ElasticsearchException;
+import co.elastic.clients.elasticsearch._types.ShardFailure;
+import co.elastic.clients.elasticsearch._types.SortOrder;
+import co.elastic.clients.elasticsearch.snapshot.CreateSnapshotRequest;
+import co.elastic.clients.elasticsearch.snapshot.CreateSnapshotResponse;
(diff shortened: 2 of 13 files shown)


webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
```diff
@@ -237,30 +249,43 @@ public void executeSnapshotting(
...
-            .featureStates(new String[] {"none"})
-            .waitForCompletion(true);
+        CreateSnapshotRequest.of(
+            b ->
+                b.repository(snapshotRequest.repositoryName())
+                    .snapshot(snapshotRequest.snapshotName())
+                    .indices(snapshotRequest.indices())
+                    // ignoreUnavailable = false - indices defined by their exact name MUST be
+                    // present
+                    // allowNoIndices = true - indices defined by wildcards, e.g. archived, MIGHT BE
```
entangled90, line 259: I think this should be mapped to "[expand_wildcards"](https://www.elastic.co/guide/en/elasticsearch/reference/current/create-snapshot-api.html#create-snapshot-api-request-body) field in the json, but it's not present in the request from ElasticsearchClient

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
```diff
@@ -231,75 +280,33 @@ public List<GetBackupStateResponseDto> getBackups(final String repositoryName) {
...
-    };
-  }
+  public void onDeleteFailure(final Exception e) {}
 
   private boolean isSnapshotMissingException(final Exception e) {
-    return e instanceof ElasticsearchStatusException
-        && ((ElasticsearchStatusException) e)
-            .getDetailedMessage()
-            .contains(SNAPSHOT_MISSING_EXCEPTION_TYPE);
+    return e instanceof ElasticsearchException
```
entangled90, line 288: This is slightly different compared to the previous api, but there is no other field so it should be ok

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
```diff
@@ -231,75 +280,33 @@ public List<GetBackupStateResponseDto> getBackups(final String repositoryName) {
...
+            b ->
+                b.repository(repositoryName)
+                    .snapshot(snapshotNameProvider.getSnapshotNamePrefix(backupId) + "*"));
+
+    final GetSnapshotResponse response;
     try {
-      response = esClient.snapshot().get(snapshotsStatusRequest, RequestOptions.DEFAULT);
-      return response.getSnapshots();
-    } catch (final IOException | TransportException ex) {
+      response = esClient.snapshot().get(snapshotsStatusRequest);
```
entangled90, line 307: the new `TransportException` extends `IOException` so it's not needed

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
```diff
@@ -335,7 +342,7 @@ public boolean isSnapshotFinishedWithinTimeout(
       final List<SnapshotInfo> snapshotInfos = findSnapshots(repositoryName, backupId);
       final SnapshotInfo currentSnapshot =
           snapshotInfos.stream()
-              .filter(x -> Objects.equals(x.snapshotId().getName(), snapshotName))
+              .filter(x -> Objects.equals(x.snapshot(), snapshotName))
```
entangled90, line 345: the snapshot name is not inside snapshotId anymore

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
```diff
@@ -344,7 +351,7 @@ public boolean isSnapshotFinishedWithinTimeout(
         // No need to continue
         return false;
       }
-      if (currentSnapshot.state() == IN_PROGRESS) {
+      if (Objects.equals(currentSnapshot.state(), IN_PROGRESS.name())) {
```
entangled90, line 354: additional checks for null as it's now nullable

dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
```diff
@@ -65,10 +65,10 @@ && validRepoName(tasklistBackup.getRepositoryName())
   public ThreadPoolTaskExecutor getTaskExecutor() {
     final ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
     executor.setCorePoolSize(1);
-    executor.setMaxPoolSize(1);
+    executor.setMaxPoolSize(8);
+    executor.setKeepAliveSeconds(60);
     executor.setThreadNamePrefix("webapps_backup_");
-    // TODO why just 6?
-    executor.setQueueCapacity(6);
+    executor.setQueueCapacity(4096);
```
npepinpe, line 71: :question: Why do we explicitly set the queue anyway?

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/SnapshotState.java
```diff
@@ -0,0 +1,25 @@
...
+  SUCCESS("SUCCESS");
+  private final String state;
+
+  SnapshotState(final String state) {
+    this.state = state;
+  }
+
+  @Override
+  public String toString() {
+    return state;
```
npepinpe, line 23: :question: Why overload `toString` if we always return the name, and not use `SnapshotState.FAILED.name()`?

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/opensearch/MetadataMarshaller.java
npepinpe, line 1: :clown_face: I guess the duplication is because of the types/packages? :(

npepinpe (approved):
:rocket:

dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
entangled90: by default it has no limit, so it's probably better to limit it

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/opensearch/MetadataMarshaller.java
entangled90: Yes, they are completely identical :laughing:

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/SnapshotState.java
entangled90: It's probably not needed :+1:

dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
npepinpe: Oh, I see Spring just puts it "unbounded" (sort of). Hm. I guess it all depends how we deal with rejected execution exceptions...

npepinpe: Considering it was already setting it, then hopefully we do anyway :shrug:

PR #25681, opened 2024-12-05
fix: parse unexpected REST responses in client

### LINKED ISSUE #25087: Zeebe Java Client Fails to Handle Error Responses Gracefully
## Description

The Zeebe Java client currently does not handle error responses properly. When the server sends an error response with a payload that doesn’t match the expected response model, the client attempts to deserialize it into the expected type. This causes a deserialization exception instead of providing meaningful feedback about the error.

**Example**
```
io.camunda.zeebe.client.api.command.ClientException: io.camunda.zeebe.client.api.command.ClientException: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "timestamp" (class io.camunda.zeebe.client.protocol.rest.UserCreateResponse), not marked as ignorable (one known property: "userKey"])
 at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain: io.camunda.zeebe.client.protocol.rest.UserCreateResponse["timestamp"])
```

This behavior makes troubleshooting issues difficult because the actual problem from the server is obscured by client-side deserialization errors.

The client should throw a predefined exception that contains the raw error message from the server for easier debugging.

tmetzke:
## Description

If an unexpected response is returned for REST calls, the Java client needs to handle this in a meaningful way. The client tries to parse "application/json" responses as the defined target type. If this fails, it returns a ProblemDetail instead, containing the original payload JSON as detail information.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25087

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
@@ -15,11 +15,13 @@
  */
 package io.camunda.zeebe.client.impl.http;
 
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.json.async.NonBlockingByteBufferJsonParser;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.util.TokenBuffer;
 import io.camunda.zeebe.client.protocol.rest.ProblemDetail;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
@@ -91,13 +93,23 @@ public interface TypedApiEntityConsumer<T> {
 
     @Override
     public ApiEntity<T> generateContent() throws IOException {
-      buffer.asParserOnFirstToken();
-
-      if (isResponse) {
-        return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), type));
+      try {
+        if (isResponse) {
+          return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), type));
+        }
+        return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), ProblemDetail.class));
+      } catch (final IOException ioe) {
+        // write the original JSON response into an error response
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final JsonGenerator generator = json.createGenerator(output);
+        buffer.serialize(generator);
+        generator.flush();
+        return ApiEntity.of(
+            new ProblemDetail()
+                .title("Unexpected server response")
+                .status(500)
+                .detail(output.toString(StandardCharsets.UTF_8.name())));
       }
-
-      return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), ProblemDetail.class));
     }
 
     @Override
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
@@ -100,15 +100,9 @@ public interface TypedApiEntityConsumer<T> {
         return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), ProblemDetail.class));
       } catch (final IOException ioe) {
         // write the original JSON response into an error response
-        final ByteArrayOutputStream output = new ByteArrayOutputStream();
-        final JsonGenerator generator = json.createGenerator(output);
-        buffer.serialize(generator);
-        generator.flush();
+        final String jsonString = getJsonString();
         return ApiEntity.of(
-            new ProblemDetail()
-                .title("Unexpected server response")
-                .status(500)
-                .detail(output.toString(StandardCharsets.UTF_8.name())));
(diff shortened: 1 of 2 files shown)


clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
```diff
@@ -91,10 +92,18 @@ public JsonApiEntityConsumer(
...
+        try {
+          return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), type));
+        } catch (final IOException ioe) {
+          return ApiEntity.of(
+              new ProblemDetail()
+                  .title("Unexpected server response")
+                  .status(500)
+                  .detail(
+                      json.writeValueAsString(
+                          json.readValue(buffer.asParserOnFirstToken(), Map.class))));
```
tmetzke, line 105: 💭 I'm not really happy with this. Overall, I'm wondering if we should go for a similar approach as the raw entity and simply collect the text. That way, we can create a very generic catch clause for the whole `generateContent` method (including trying to parse the problem detail response) that is less prone to run into response obfuscation again. Currently, this can still happen if the response cannot be parsed into a `Map` either.

tmetzke: Alternative I will pursue: Flush the original JSON collected in the `TokenBuffer` out into an output stream "raw"

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
```diff
@@ -91,13 +93,23 @@ public JsonApiEntityConsumer(
...
-        return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), type));
+      try {
+        if (isResponse) {
+          return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), type));
+        }
+        return ApiEntity.of(json.readValue(buffer.asParserOnFirstToken(), ProblemDetail.class));
+      } catch (final IOException ioe) {
+        // write the original JSON response into an error response
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final JsonGenerator generator = json.createGenerator(output);
```
houssain-barouni, line 104: `JsonGenerator` is closable so may need to be closed
```suggestion
    try (final ByteArrayOutputStream output = new ByteArrayOutputStream();
         final JsonGenerator generator = json.createGenerator(output)) {
```

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
```diff
@@ -91,13 +93,23 @@ public JsonApiEntityConsumer(
...
+        buffer.serialize(generator);
```
houssain-barouni, line 105: `serialize` itself throws an `IOException`, I guess we handle it upstreams as before?

houssain-barouni (commented):
Thanks for fixing this! 💪 
one little suggestion 🔧

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
nathansandi, line 104: 💭 Should we place this inside a try, to ensure proper resource cleanup? 

```
try (JsonGenerator generator = json.createGenerator(output)) {
      buffer.serialize(generator);
      generator.flush();
  }
```

nathansandi (approved):
Looks good to me, I have added a minor suggestion related to `ByteArrayOutputStream`  usage, once it can throw a new exception inside the exception block.

nathansandi (changes requested):
I approve by accident. But so far, minor suggestion changes.

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
tmetzke: Yep, let's do that

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
tmetzke: We'll handle it explicitly to avoid obfuscating the original issue. However, I don't see much we can do anymore except for saying "we received something we cannot decipher"

clients/java/src/main/java/io/camunda/zeebe/client/impl/http/TypedApiEntityConsumer.java
```diff
@@ -136,6 +142,17 @@ public void releaseResources() {
...
     }
+
+    private String getJsonString() {
+      try (final ByteArrayOutputStream output = new ByteArrayOutputStream();
+          final JsonGenerator generator = json.createGenerator(output)) {
+        buffer.serialize(generator);
+        generator.flush();
+        return output.toString(StandardCharsets.UTF_8.name());
+      } catch (final Exception ex) {
+        return "Original response cannot be constructed";
```
houssain-barouni, line 153: we should pass the exception to the logger, you will thank me when you are the medic 😄 
```suggestion
        LOGGER.error("Failed to serialize the error response", ex);
        return "Original response cannot be constructed";
```

tmetzke: Yeah, you're right 😆

houssain-barouni (approved):
LGTM 👍

nathansandi (approved):
LGTM :) 🥇

PR #24091, opened 2024-10-25
Use tasklist-task index on Operate

panagiotisgts:
## Description
Make use of the new `tasklist-task` index in Operate's BE services for Elasticsearch and Opensearch

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes #23891

PR #25682, opened 2024-12-05
refactor: export snapshot variable on user task completion

romansmirnov:
## Description

* Export Snapshot Task Variables on User Task completion.
* This is necessary to support the Tasklist v1 API.
* Refactors the UserTaskReader in Operate to retrieve (snapshot) Task Variables from the snapshot task variable index.

## Checklist

- [ ] for CI changes:

## Related issues

related to #24665

### DIFF
--- a/operate/webapp/src/main/java/io/camunda/operate/webapp/elasticsearch/reader/ElasticsearchUserTaskReader.java
+++ b/operate/webapp/src/main/java/io/camunda/operate/webapp/elasticsearch/reader/ElasticsearchUserTaskReader.java
@@ -14,22 +14,20 @@ import io.camunda.operate.conditions.ElasticsearchCondition;
 import io.camunda.operate.exceptions.OperateRuntimeException;
 import io.camunda.operate.util.ElasticsearchUtil;
 import io.camunda.operate.webapp.reader.UserTaskReader;
+import io.camunda.webapps.schema.descriptors.tasklist.template.SnapshotTaskVariableTemplate;
 import io.camunda.webapps.schema.descriptors.tasklist.template.TaskTemplate;
+import io.camunda.webapps.schema.entities.tasklist.SnapshotTaskVariableEntity;
 import io.camunda.webapps.schema.entities.tasklist.TaskEntity;
-import io.camunda.webapps.schema.entities.tasklist.TaskJoinRelationship.TaskJoinRelationshipType;
-import io.camunda.webapps.schema.entities.tasklist.TaskVariableEntity;
 import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
 import org.elasticsearch.action.search.SearchRequest;
-import org.elasticsearch.index.query.BoolQueryBuilder;
-import org.elasticsearch.index.query.ExistsQueryBuilder;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
-import org.elasticsearch.join.query.HasParentQueryBuilder;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.slf4j.Logger;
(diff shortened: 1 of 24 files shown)


panagiotisgts (changes requested):
@romansmirnov This change breaks the `tasklist-task` index in regards to Variables.

The `tasklist-task` index is also responsible for handling the Task Completion variables snapshotting since the V2 APIs will use this index to retrieve User Task related variables. 

I believe we should mirror the previous handler to also flush the variables in the `tasklist-task-variable` index to maintain V1 compatibility.

romansmirnov:
> This change breaks the tasklist-task index in regards to Variables.
> 
> The tasklist-task index is also responsible for handling the Task Completion variables snapshotting since the V2 APIs will use this index to retrieve User Task related variables.
> 
> I believe we should mirror the previous handler to also flush the variables in the tasklist-task-variable index to maintain V1 compatibility.

The snapshotting feature is deprecated and will work only with the v1 API. In the case of v2 API, we are planning to implement a general concept that works not only for user tasks.

panagiotisgts:
> The snapshotting feature is deprecated and will work only with the v1 API. In the case of v2 API, we are planning to implement a general concept that works not only for user tasks.

Is this somewhere documented? I haven't seen anything regarding this so this is still my assumption. This may have some impact on some existing work that has been done, for example in this [PR](https://github.com/camunda/camunda/pull/24091) the user task variables are fetched from the `tasklist-task` index, expecting that the snapshotted variables will be there as well.

romansmirnov:
It is documented in some docs - I am pretty sure that this is the case, but I can't find them immediately. I think @marcosgvieira should also have the context.

Looking at the linked PR, it seems that in the old Operate importer implementation, the "variable map" was [stored as a String](https://github.com/camunda/camunda/pull/24091/files#diff-2da222e938f8c711b575fea9a49faa397defbdd35ccf61410a90c53ab17ec81dL200) but not as actual variables.

However, I would suggest the following:
* Let's keep the snapshot variable out of the `tasklist-task` index. That way, we ensure that we can, at some point, replace the snapshot user task variables with a common concept. That way, any v2 search based on the `tasklist-task` index would keep working the same, even when we remove the snapshot user task variable feature.
* For now, the snapshotted variables can retrieved from the `tasklist-task-variable` index, also in the case of the old Operate backend. That way, the metadata will show them as previously in the UI. Once we have a common concept as an audit log of variable changes, then the current functionality/feature in the Operate UI can be replaced by that.
* I can rewrite that referenced part in the old Operate backend to retrieve these variables from the `tasklist-task-variable` index in the context of this PR. (I wonder why no tests failed; otherwise, I would have done this already.)

romansmirnov:
I've adjusted the UserTaskReader in the old Operate backend.

operate/webapp/src/main/java/io/camunda/operate/webapp/elasticsearch/reader/ElasticsearchUserTaskReader.java
```diff
@@ -83,26 +86,17 @@ public Optional<TaskEntity> getUserTaskByFlowNodeInstanceKey(final long flowNode
   }
 
   @Override
-  public List<TaskVariableEntity> getUserTaskVariables(final long flowNodeInstanceKey) {
-    LOGGER.debug("Get UserTask Completed Variables by flowNodeInstanceKey {}", flowNodeInstanceKey);
+  public List<SnapshotTaskVariableEntity> getUserTaskVariables(final long taskKey) {
+    LOGGER.debug("Get UserTask Completed Variables by flowNodeInstanceKey {}", taskKey);
```
panagiotisgts, line 90: Nitpick, just the log message rename `flowNodeInstanceKey` to `userTaskKey`

panagiotisgts (approved):
LGTM

Issue #21724, opened 2024-08-29
Integrate aggregation into Search API


ralfpuchert:
### Description

Operate defines endpoints which involve aggregations on database:
* https://github.com/camunda/camunda/issues/21716 
* https://github.com/camunda/camunda/issues/21720
* https://github.com/camunda/camunda/issues/21717
* https://github.com/camunda/camunda/issues/21719
* https://github.com/camunda/camunda/issues/21721

The current search API does not support aggregations in [SearchRequest](https://github.com/camunda/camunda/blob/e9afa1edb0e498d11e0907f09803368665ba337c/search/search-client/src/main/java/io/camunda/search/clients/core/SearchQueryRequest.java#L23-L23)
and [SearchResponse](https://github.com/camunda/camunda/blob/e9afa1edb0e498d11e0907f09803368665ba337c/search/search-client/src/main/java/io/camunda/search/clients/core/SearchQueryResponse.java#L16-L16)

In order to support aggregations an aggregation framework which integrates in `SearchRequest` and `SearchResponse` is needed. This includes `Aggregation` (request/specification) and `Aggregate` (response/result) components with corresponding `Transformer` and `Builder`. Also different types of aggregations and aggregates are needed.

The following aggregations are needed to implement Operate statistic endpoints:

* [CardinalityAggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-cardinality-aggregation.html) 
* [FiltersAggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-filter-aggregation.html)
* [TermsAggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html) 
* [TopHitsAggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-top-hits-aggregation.html)
* [ChildrenAggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-children-aggregation.html)

### Documentation

* [OpenSearch](https://opensearch.org/docs/latest/aggregations/)
* [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html)

### Tasks

- [ ] Create Aggregation and Aggregate components for C8 API
- [ ] Integrate them in SearchRequest and SearchResponse
[...]

PR #25690, opened 2024-12-06
feat: add usage metrics endpoint 

### LINKED ISSUE #23812: Implement `Usage Metric` REST API
## Description

We decided to migrate the former actuator endpoints from [Tasklist](https://docs.camunda.io/docs/next/self-managed/tasklist-deployment/usage-metrics/) and [Operate](https://docs.camunda.io/docs/next/self-managed/operate-deployment/usage-metrics/) to the C8 REST API. 

This will make access to the metrics API easier, as the user is not depending on the management port which switched to default to 9600 with the 8.6 release for Tasklist and Operate.

The usage metrics endpoint should provide the Usage metrics as defined [in the public documentation](https://docs.camunda.io/docs/next/reference/usage-metrics/#definition-of-metrics).

We want to consolidate the former dedicated usage metrics endpoints into one single usage metrics endpoint.
We accept that we may need to perform two queries to `operate-metric-1.0.0_` and `tasklist-metric-1.0.0_` to build the result, as both indices will stay (see [this (🔒 internal) discussion](https://docs.google.com/spreadsheets/d/1XuzUmht3hML32QOfWiz_5lNo9B5lakqDz1tOM16oAUI/edit?gid=0#gid=0&range=E12)).

The endpoint should be implemented as described in the [REST API Developer Guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)

### Endpoint

* Summary: Query usage metrics
* Description: Search for usage metrics with defined criteria, e.g. for a specific date range.
* Path: `POST /v2/usage-metrics/search`
* Request body: search request with the following filter attributes
  * `startTime` - type `string[date]` - required, no default - the start time of the date range to fetch the metrics for
  * `endTime` - type `string[date]` - required, no default - the end time of the date range to fetch the metrics for
* Response:
  * `200 OK` - type `application/json` - a JSON object representing the usage metrics with the following fields
    * `assignees` - type `long` - the amount of unique active users
    * `processInstances` - type `long` - the amount of created process instances
    * `decisionInstances` - type `long` - the amount of executed decision instances
  * example response
    ```json
    {
      "assignees": 15,
      "processInstances": 12345,
      "decisionInstances": 5354365
    }
    ```

### Acceptance Criteria

- [x] Service Layer for Metrics search
- [x] REST API endpoint incl. OpenAPI spec
- [x] Integration tests

### Excluded from Scope

- Java client support - not doing as this endpoint has no use-case within process automation

### Related resources

* PDP epic: https://github.com/camunda/product-hub/issues/2544

### Implementation notes

- Tasklist uses [TermsAggregation](https://github.com/camunda/camunda/blob/50d3a76829a966858a77cb2100ab87409d0d97d3/tasklist/els-schema/src/main/java/io/camunda/tasklist/store/elasticsearch/TaskMetricsStoreElasticSearch.java#L100-L100)
  - for counting assignees  
- Operate uses [TermsAggregation](https://github.com/camunda/camunda/blob/50d3a76829a966858a77cb2100ab87409d0d97d3/operate/schema/src/main/java/io/camunda/operate/store/elasticsearch/dao/Query.java#L50-L50) in a abstraction layer `Query` 
  - for counting process instances
  - for counting decision instances 
- C8 API needs aggregations (at least `TermsAggregation`)
  - [PR]() for adding aggregations
#### Current solution
Apply distinct count in service layer until aggregations are integrated in search module.

### Pull requests

- [OpenAPI definition, Rest controller, Service infrastructure](https://github.com/camunda/camunda/pull/25690)

ralfpuchert:
## Description

Current implementation is done in Service layer by using Java Stream API.
When aggregation is integrated in search modules the implementation should be switch to do
the distinct count in databases.

* Add OpenAPI definition.
* Add RestController and service infrastructure.
* Fix [findAll](https://github.com/camunda/camunda/pull/25690/commits/121654110f47d406f316120656b9188bbb696815) bug
* Add tests for RestController, Filter, Sort and Service.
* Add UsageMetricsSearchClient methods in RDBMSClient to make integration tests green.

## Checklist

- [ ] for CI changes:

## Related issues

closes #23812

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -262,7 +262,7 @@ public class SearchClients
                 new UsageMetricsFilter.Builder()
                     .startTime(query.filter().startTime())
                     .endTime(query.filter().endTime())
-                    .events(event)
+                    .event(event)
                     .build())
             .build();
     final List<UsageMetricsEntity> metrics =
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -262,7 +262,7 @@ public class SearchClients
                 new UsageMetricsFilter.Builder()
                     .startTime(query.filter().startTime())
                     .endTime(query.filter().endTime())
-                    .event(event)
+                    .events(event)
                     .build())
             .build();
     final List<UsageMetricsEntity> metrics =
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -254,6 +254,9 @@ public class SearchClients
     return distinctCountUsageMetricsFor("EVENT_DECISION_INSTANCE_EVALUATED", query);
   }
 
+  /*
+   * The distinct count is implemented here by using Java Stream API until aggregations are in place.
+   */
   private Long distinctCountUsageMetricsFor(final String event, final UsageMetricsQuery query) {
     final var filter =
         new UsageMetricsQuery.Builder()
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -238,52 +238,37 @@ public class SearchClients
   }
 
   @Override
-  public SearchQueryResult<UsageMetricsEntity> searchUsageMetrics(final UsageMetricsQuery query) {
-    final var executor =
-        new SearchClientBasedQueryExecutor(
-            searchClient,
-            transformers,
-            new DocumentAuthorizationQueryStrategy(this),
-            securityContext);
-    final var assigneesFilter =
-        new UsageMetricsQuery.Builder()
-            .filter(
-                new UsageMetricsFilter.Builder()
-                    .startTime(query.filter().startTime())
-                    .endTime(query.filter().endTime())
-                    .events("task_completed_by_assignee")
-                    .build())
-            .build();
-    final var assigneesResult =
-        executor.search(
-            assigneesFilter, io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
-    final var processInstancesFilter =
-        new UsageMetricsQuery.Builder()
-            .filter(
-                new UsageMetricsFilter.Builder()
-                    .startTime(query.filter().startTime())
-                    .endTime(query.filter().endTime())
-                    .events("task_completed_by_assignee")
-                    .build())
-            .build();
-    final var processInstancesResult =
-        executor.search(
-            processInstancesFilter,
-            io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
-    final var decisionInstancesFilter =
-        new UsageMetricsQuery.Builder()
-            .filter(
-                new UsageMetricsFilter.Builder()
-                    .startTime(query.filter().startTime())
-                    .endTime(query.filter().endTime())
-                    .events("EVENT_DECISION_INSTANCE_EVALUATED")
-                    .build())
-            .build();
-    final var decisionInstancesResult =
-        executor.search(
-            decisionInstancesFilter,
-            io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
+  public Long countAssignees(final UsageMetricsQuery query) {
+    return distinctCountUsageMetricsFor("task_completed_by_assigne", query);
+  }
 
-    return new SearchQueryResult<>(0, List.of(), null /* aggregate */);
+  @Override
+  public Long countProcessInstances(final UsageMetricsQuery query) {
+    return distinctCountUsageMetricsFor("EVENT_PROCESS_INSTANCE_STARTED", query);
+  }
+
+  @Override
+  public Long countDecisionInstances(final UsageMetricsQuery query) {
+    return distinctCountUsageMetricsFor("EVENT_DECISION_INSTANCE_EVALUATED", query);
+  }
+
+  private Long distinctCountUsageMetricsFor(final String event, final UsageMetricsQuery query) {
+    final var filter =
+        new UsageMetricsQuery.Builder()
+            .filter(
+                new UsageMetricsFilter.Builder()
+                    .startTime(query.filter().startTime())
+                    .endTime(query.filter().endTime())
+                    .events(event)
+                    .build())
+            .build();
+    final List<UsageMetricsEntity> metrics =
+        new SearchClientBasedQueryExecutor(
+                searchClient,
+                transformers,
+                new DocumentAuthorizationQueryStrategy(this),
+                securityContext)
+            .findAll(filter, io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
+    return metrics.stream().map(UsageMetricsEntity::value).distinct().count();
   }
 }
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -26,6 +26,7 @@ import io.camunda.search.entities.UsageMetricsEntity;
 import io.camunda.search.entities.UserEntity;
 import io.camunda.search.entities.UserTaskEntity;
 import io.camunda.search.entities.VariableEntity;
+import io.camunda.search.filter.UsageMetricsFilter;
 import io.camunda.search.query.AuthorizationQuery;
 import io.camunda.search.query.DecisionDefinitionQuery;
 import io.camunda.search.query.DecisionInstanceQuery;
@@ -237,8 +238,52 @@ public class SearchClients
   }
 
   @Override
-  public UsageMetricsEntity searchUsageMetrics(final UsageMetricsQuery query) {
-    // TODO: Search Operate and Tasklist
-    return new UsageMetricsEntity(1L, 2L, 3L);
+  public SearchQueryResult<UsageMetricsEntity> searchUsageMetrics(final UsageMetricsQuery query) {
+    final var executor =
+        new SearchClientBasedQueryExecutor(
+            searchClient,
+            transformers,
+            new DocumentAuthorizationQueryStrategy(this),
+            securityContext);
+    final var assigneesFilter =
+        new UsageMetricsQuery.Builder()
+            .filter(
+                new UsageMetricsFilter.Builder()
+                    .startTime(query.filter().startTime())
+                    .endTime(query.filter().endTime())
+                    .events("task_completed_by_assignee")
+                    .build())
+            .build();
+    final var assigneesResult =
+        executor.search(
+            assigneesFilter, io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
+    final var processInstancesFilter =
+        new UsageMetricsQuery.Builder()
+            .filter(
+                new UsageMetricsFilter.Builder()
+                    .startTime(query.filter().startTime())
+                    .endTime(query.filter().endTime())
+                    .events("task_completed_by_assignee")
+                    .build())
+            .build();
+    final var processInstancesResult =
+        executor.search(
+            processInstancesFilter,
+            io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
+    final var decisionInstancesFilter =
+        new UsageMetricsQuery.Builder()
+            .filter(
+                new UsageMetricsFilter.Builder()
+                    .startTime(query.filter().startTime())
+                    .endTime(query.filter().endTime())
+                    .events("EVENT_DECISION_INSTANCE_EVALUATED")
+                    .build())
+            .build();
+    final var decisionInstancesResult =
+        executor.search(
+            decisionInstancesFilter,
+            io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
+
+    return new SearchQueryResult<>(0, List.of(), null /* aggregate */);
   }
 }
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -27,6 +27,7 @@ import io.camunda.search.entities.UserEntity;
 import io.camunda.search.entities.UserTaskEntity;
 import io.camunda.search.entities.VariableEntity;
 import io.camunda.search.filter.UsageMetricsFilter;
+import io.camunda.search.page.SearchQueryPage;
 import io.camunda.search.query.AuthorizationQuery;
 import io.camunda.search.query.DecisionDefinitionQuery;
 import io.camunda.search.query.DecisionInstanceQuery;
@@ -48,7 +49,9 @@ import io.camunda.search.query.VariableQuery;
 import io.camunda.security.auth.SecurityContext;
 import io.camunda.webapps.schema.descriptors.IndexDescriptors;
 import io.camunda.zeebe.util.CloseableSilently;
+import java.util.HashSet;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class SearchClients
     implements AuthorizationSearchClient,
@@ -239,17 +242,17 @@ public class SearchClients
 
   @Override
   public Long countAssignees(final UsageMetricsQuery query) {
-    return distinctCountUsageMetricsFor("task_completed_by_assigne", query);
+    return distinctCountUsageMetricsFor2("task_completed_by_assignee", query);
   }
 
   @Override
   public Long countProcessInstances(final UsageMetricsQuery query) {
-    return distinctCountUsageMetricsFor("EVENT_PROCESS_INSTANCE_STARTED", query);
+    return distinctCountUsageMetricsFor2("EVENT_PROCESS_INSTANCE_STARTED", query);
   }
 
   @Override
   public Long countDecisionInstances(final UsageMetricsQuery query) {
-    return distinctCountUsageMetricsFor("EVENT_DECISION_INSTANCE_EVALUATED", query);
+    return distinctCountUsageMetricsFor2("EVENT_DECISION_INSTANCE_EVALUATED", query);
   }
 
   private Long distinctCountUsageMetricsFor(final String event, final UsageMetricsQuery query) {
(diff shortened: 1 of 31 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -49,6 +49,7 @@ tags:
   - name: User
   - name: User task
   - name: Variable
+  - name: Usage metrics
```
tmetzke, line 52: 🔧 Let's get this into its alphabetically correct spot in the list 🙂

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -410,7 +425,12 @@ private static ProcessDefinitionFilter toProcessDefinitionFilter(
   }
 
   private static OffsetDateTime toOffsetDateTime(final String text) {
-    return StringUtils.isEmpty(text) ? null : OffsetDateTime.parse(text);
+    try {
+      return StringUtils.isEmpty(text) ? null : OffsetDateTime.parse(text);
+    } catch (final Exception e) {
+      final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSX");
+      return OffsetDateTime.parse(text, formatter);
```
tmetzke, line 432: ❓ Why do we need to add a custom formatter here?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/UsageMetricsQueryController.java
```diff
@@ -0,0 +1,55 @@
...
+import io.camunda.zeebe.gateway.rest.SearchQueryResponseMapper;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@CamundaRestQueryController
+@RequestMapping("/v2/usage-metrics")
+public class UsageMetricsQueryController {
```
tmetzke, line 28:
```suggestion
@CamundaRestController
@RequestMapping("/v2/usage-metrics")
public class UsageMetricsController {
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
ralfpuchert: In [current Usage Metrics documentation](https://docs.camunda.io/docs/next/self-managed/operate-deployment/usage-metrics/) the format `yyyy-MM-dd'T'HH:mm:ss.SSSZZ`
is mentioned. This format is not supported in `OffsetDateTime::parse`

One way would be to adjust the documentation.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
```diff
@@ -294,4 +298,42 @@ public SearchQueryResult<VariableEntity> searchVariables(final VariableQuery fil
...
+    return distinctCountUsageMetricsFor("EVENT_PROCESS_INSTANCE_STARTED", query);
+  }
+
+  @Override
+  public Long countDecisionInstances(final UsageMetricsQuery query) {
+    return distinctCountUsageMetricsFor("EVENT_DECISION_INSTANCE_EVALUATED", query);
+  }
+
+  /*
+   * The distinct count is implemented here by using Java Stream API until aggregations are in place.
```
megglos, line 318: ❓ @tmetzke @ralfpuchert do we have aggregation support still planned for 8.7?

ralfpuchert: They are planned for 8.8 according to [backlog](https://github.com/orgs/camunda/projects/126/views/1?filterQuery=-repo%3A%22camunda%2Fproduct-hub%22+aggre) : [issue](https://github.com/camunda/camunda/issues/21724)

tmetzke: Well, I'd say it depends a bit. We might need "full" aggregation support for offering current V1 Operate internal endpoints in V2 (or at least partially). However, the aggregation we need here is basically an option in the search API to "count" and "count distinct". We could start with a really basic integration of this in the current search API so ES/OS and RDBMS can implement and map this operation accordingly.

When we later build a "proper" aggregation part of the search API, we can move this there if needed. But we should try to get "count" and "count distinct" in for 8.7 to make this a well-performing API in my view.

Disclaimer: this might be a lot more complex then I currently anticipate. However, I believe if we focus on those two options, we can get them in "pragmatically" without building a whole "concept" in the search API for now.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4730,6 +4766,35 @@ components:
...
+        - endTime
+    UsageMetricsSearchQueryResponse:
+      type: object
+      properties:
+        assignees:
+          description: The amount of unique active users.
+          type: integer
+          format: int64
+        processInstances:
+          description: The amount of created process instances.
```
tmetzke, line 4791:
```suggestion
          description: The amount of created root process instances.
```

search/search-client-rdbms/src/main/java/io/camunda/search/rdbms/RdbmsSearchClient.java
```diff
@@ -126,6 +129,24 @@ public RdbmsSearchClient withSecurityContext(final SecurityContext securityConte
...
+  public Long countProcessInstances(final UsageMetricsQuery query) {
+    throw new UnsupportedOperationException(
+        "UsageMetricsClient countProcessInstances not implemented yet.");
+  }
+
+  @Override
+  public Long countDecisionInstances(final UsageMetricsQuery query) {
+    throw new UnsupportedOperationException(
+        "UsageMetricsClient countDecisionInstances not implemented yet.");
+  }
```
tmetzke, line 148: @ralfpuchert, I believe we should at least add a follow-up issue for this so this is not forgotten. If the RDBMS project has an issue for this already, we need to add a reference to this PR so the RDBMS team is aware.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -90,6 +94,21 @@ public final class SearchQueryRequestMapper {
 
   private SearchQueryRequestMapper() {}
 
+  public static Either<ProblemDetail, UsageMetricsQuery> toUsageMetricsQuery(
+      final UsageMetricsSearchQueryRequest request) {
+    if (request == null || (request.getStartTime() == null && request.getEndTime() == null)) {
```
tmetzke, line 99: 🔧 I guess all of them need to be defined, right?
```suggestion
    if (request == null || request.getStartTime() == null || request.getEndTime() == null) {
```

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UsageMetricsFilterTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+    operateMetricIndex = metricIndex;
+  }
+
+  @Override
+  public SearchQuery toSearchQuery(final UsageMetricsFilter filter) {
+    this.filter = filter;
+    final var queries = new ArrayList<SearchQuery>();
+    queries.add(stringTerms("event", filter.events()));
+    queries.addAll(dateTimeOperations("eventTime", Operation.gte(filter.startTime()).values()));
+    queries.addAll(dateTimeOperations("eventTime", Operation.lte(filter.endTime()).values()));
```
tmetzke, line 40: 🔧 Looking at other transformers, e.g. the [UserTaskFilterTransformer](https://github.com/camunda/camunda/blob/bfbfcf6a9c4ac5959ee109d72c60801e9a3fa533/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UserTaskFilterTransformer.java), we could improve here in the following:
* Reuse the field names defined in the webapps schema classes instead of using "magic" strings.
* If an `Operation` is already defined (by the way, I think it technically could be a list of operations here already), we take the operator and value from that instead of hardcoding it here again. If we change the query in the REST API, it will still be `gte` and `lte` as defined here.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UsageMetricsQueryControllerTest.java
```diff
@@ -0,0 +1,109 @@
...
+        .contentType(MediaType.APPLICATION_JSON)
+        .bodyValue("{}")
+        .exchange()
+        .expectStatus()
+        .isBadRequest()
+        .expectHeader()
+        .contentType(MediaType.APPLICATION_PROBLEM_JSON_VALUE)
+        .expectBody()
+        .json(
+            "{\"type\":\"about:blank\",\"title\":\"INVALID_ARGUMENT\",\"status\":400,\"detail\":\"startTime and endTime must be specified.\",\"instance\":\"/v2/usage-metrics/search\"}");
```
tmetzke, line 107: 🔧 This is quite hard to read, how about we use multiline strings like the other test cases?

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UsageMetricsQueryControllerTest.java
```diff
@@ -0,0 +1,109 @@
...
+                .filter(
+                    new UsageMetricsFilter.Builder()
+                        .startTime(Operation.gte(startTime))
+                        .endTime(Operation.lte(endTime))
+                        .build())
+                .build());
+  }
+
+  @Test
+  void shouldThrowExceptionIfNoStartAndEndTimeAreGiven() {
```
tmetzke, line 93: 🔧 We should test here even leaving out one of them and also that leaving out both leads to a bad request response.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3347,7 +3348,42 @@ paths:
...
           $ref: "#/components/responses/InternalServerError"
-
+  /usage-metrics/search:
+    post:
+      tags:
+        - Usage metrics
+      operationId: searchUsageMetrics
+      summary: Query usage metrics
+      description: |
+        Search for usage metrics by given start date and end date
```
pepopowitz, line 3358:
```suggestion
        Search for usage metrics by given start date and end date.
```

:wrench: Descriptions should end in a period.

pepopowitz (approved):
👍 One very small nit in the spec, otherwise the spec changes look solid 💪

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
megglos: Effectively using a filter aggregation on event name and obtaining the doc count from it should be suffice and we could potentially do that with a single query as well having one filter aggregation for each process, decision and task events querying both indices at the same time? Wdyt @ralfpuchert ?

I would be in favor to give that a try still before the 8.7 release, I could take a look next week in between the holidays if you don't have the capacity this week.

megglos (approved):
Thanks for making such good progress on this topic!

I only did a high level review on the functionality not being too familiar with the actual architecture and adherence to existing concepts, I would leave that part to @tmetzke to provide feedback on.

Two things I'd like to reconsider as I don't see a need/use-case for them, thus we could reduce complexity there for easier maintenance:
- sorting support, actually querying the metric entities is just done for counting, records are not served on the API, I would argue we don't need any sorting support
- Operations on startTime and endTime - I don't see a use case here either for using different operations than gter and lte respecitvely. Thus we could reduce complexity here too

We can also do this in follow-ups though, it's not urgent so both are 🔧 .

For the API to be production ready I think we need to switch to aggregtations though as otherwise we likely will run into request timeouts etc. iterating through high cardinality metric records. Thus I would like us to include https://github.com/camunda/camunda/issues/26106 still in 8.7. I'm fine merging this as a first iteration so we have the functional API available.

search/search-domain/src/main/java/io/camunda/search/filter/UsageMetricsFilter.java
```diff
@@ -0,0 +1,51 @@
...
+import java.time.OffsetDateTime;
+import java.util.List;
+
+public record UsageMetricsFilter(
+    List<String> events, Operation<OffsetDateTime> startTime, Operation<OffsetDateTime> endTime)
+    implements FilterBase {
+
+  public static final class Builder implements ObjectBuilder<UsageMetricsFilter> {
+    private List<String> events;
+    private Operation<OffsetDateTime> startTime;
```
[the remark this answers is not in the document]
tmetzke: We can do that. However, the search API and the helper methods we have already allow to easily map date time operations, we can simply reuse this and thus allow this search to be flexible in this layer. The callers, e.g. the REST API, can decide how complex the query needs to be and what options are exposed.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
tmetzke: I agree, that sounds like a feasible next iteration to try here. If you have time to look into it, @megglos, we'll definitely appreciate it 😅

tmetzke:
I will follow up on the smaller improvement ideas here today and tomorrow before merging this.
@megglos, if you have time in the next weeks, we'll appreciate any attempts at solving https://github.com/camunda/camunda/issues/26106 👍

ralfpuchert:
Hello @tmetzke, @megglos I think we can avoid loading all documents into memory by

- For Operate's process instance we can just ask for total hits and query only event process instance started. There shouldn't be any duplicates.
- For Operate's decision instance we can just ask for total hits. There shouldn't be any duplicates.
- Maybe even add general `count` method in `ElasticsearchSearchClient` for the two points above.
- For Tasklist assignees we could add another [ElasticsearchSearchClient::findAll](https://github.com/camunda/camunda/blob/bfbfcf6a9c4ac5959ee109d72c60801e9a3fa533/search/search-client-elasticsearch/src/main/java/io/camunda/search/es/clients/ElasticsearchSearchClient.java#L81) method with a consumer parameter. The consumer can collect every assignee into a set during the scrolling.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
tmetzke: Reverting this into its former state

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UsageMetricsQueryControllerTest.java
tmetzke: Adding test cases for this

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/UsageMetricsQueryControllerTest.java
tmetzke: Adjusting this to use multiline strings

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
tmetzke: Adjusting this accordingly

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Adding this change

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4730,6 +4766,35 @@ components:
...
+      type: object
+      properties:
+        startTime:
+          description: The start date and time for usage metrics.
+          type: string
+          format: date-time
+        endTime:
+          description: The end date and time for usage metrics.
+          type: string
+          format: date-time
```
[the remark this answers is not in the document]
tmetzke: Including a change for this to clarify

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Adding this

search/search-domain/src/main/java/io/camunda/search/filter/UsageMetricsFilter.java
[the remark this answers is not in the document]
tmetzke: Looking into this a bit more closely, I agree with @megglos. I simplified this since the operation is hardwired by the query, this is not as flexible as other endpoints.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/UsageMetricsFilterTransformer.java
tmetzke: Simplifying this by hardwiring the operations and using simple values in the filter

tmetzke:
⚠️ I decided to turn this into a GET endpoint with query parameters. This is not a search endpoint because this doesn't yield a list of items as a result. It returns one aggregated result. Using this as a GET endpoint feels more fitting when using it and also when implementing the required changes.

pepopowitz (approved):
👍 spec changes look good, and it makes sense to me as a GET.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/sort/UsageMetricsFieldSortingTransformer.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.search.clients.transformers.sort;
+
+public class UsageMetricsFieldSortingTransformer implements FieldSortingTransformer {
```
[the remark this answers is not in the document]
tmetzke: I have not removed this for now. It is actually a bit more involved to remove it, as far as I can see. I suggest we leave it in for now.

search/search-client-rdbms/src/main/java/io/camunda/search/rdbms/RdbmsSearchClient.java
tmetzke: @megglos, are we simply leaving this as is for now? As discussed, this might never become a thing for RDBMS anyway if we want to move this into Zeebe directly. If it does become a requirement, we can still add it to the RDBMS scope later?

PR #25699, opened 2024-12-06
Migrate active parallel gateway

### LINKED ISSUE #22128: Migrate active parallel gateway
## Description

For the migration of process instances with parallel gateways we need to consider the case where a parallel gateway is currently active. For example, because there is an incident raised on it.

This allows a special use case: resolving an incident because of an invalid execution listener `type` or `retries` expression.

An incident could also exist on the active parallel gateway if the execution listener's job is failed without retries. However, this incident can already be resolved by updating the job's retries directly, or through process instance modification (re-activate the parallel gateway).

berkaycanbc:
## Description

In order to support migration for active parallel gateways we need to add parallel gateway to supported element types. No other change is needed for incident migration because it is already handled as a part of the incident migration implementation.

## Related issues

closes #22128

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
@@ -57,7 +57,8 @@ public final class ProcessInstanceMigrationPreconditions {
           BpmnElementType.BUSINESS_RULE_TASK,
           BpmnElementType.SCRIPT_TASK,
           BpmnElementType.SEND_TASK,
-          BpmnElementType.MULTI_INSTANCE_BODY);
+          BpmnElementType.MULTI_INSTANCE_BODY,
+          BpmnElementType.PARALLEL_GATEWAY);
   private static final Set<BpmnElementType> UNSUPPORTED_ELEMENT_TYPES =
       EnumSet.complementOf(SUPPORTED_ELEMENT_TYPES);
   private static final Set<BpmnEventType> SUPPORTED_INTERMEDIATE_CATCH_EVENT_TYPES =
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateParallelGatewayTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateParallelGatewayTest.java
@@ -35,10 +35,10 @@ public class MigrateParallelGatewayTest {
     final String sourceProcessId = helper.getBpmnProcessId();
     final String targetProcessId = helper.getBpmnProcessId() + "_target";
 
-    final String NON_EXISTING_EL_JOB_TYPE_EXP = "=source_execution_listener_job_type";
-    final String EXISTING_EL_JOB_TYPE_EXP = "=target_execution_listener_job_type";
-    final String EXISTING_EL_JOB_TYPE_VAR = "target_execution_listener_job_type";
-    final String EXISTING_EL_JOB_TYPE_VALUE = "elJobType";
+    final String nonExistingElJobTypeExp = "=source_execution_listener_job_type";
+    final String existingElJobTypeExp = "=target_execution_listener_job_type";
+    final String existingElJobTypeVar = "target_execution_listener_job_type";
+    final String existingElJobTypeValue = "elJobType";
 
     final var deployment =
         ENGINE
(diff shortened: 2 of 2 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateParallelGatewayTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Rule public final TestWatcher watcher = new RecordingExporterTestWatcher();
+  @Rule public final BrokerClassRuleHelper helper = new BrokerClassRuleHelper();
+
+  @Test
+  public void shouldMigrateParallelGatewayWithIncident() {
+    final String sourceProcessId = helper.getBpmnProcessId();
+    final String targetProcessId = helper.getBpmnProcessId() + "_target";
+
+    final String nonExistingElJobTypeExp = "=source_execution_listener_job_type";
```
korthout, line 38: 🔧 Alternatively to a feel expression that can't be evaluated (and perhaps more commonly encountered), you could raise an incident by failing the execution listener job without retries.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateParallelGatewayTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    // then
+    ENGINE.incident().ofInstance(processInstanceKey).withKey(incident.getKey()).resolve();
+    ENGINE.job().ofInstance(processInstanceKey).withType(existingElJobTypeValue).complete();
+
+    assertThat(
+            RecordingExporter.processInstanceRecords(ProcessInstanceIntent.ELEMENT_ACTIVATED)
+                .withProcessInstanceKey(processInstanceKey)
+                .withElementType(BpmnElementType.END_EVENT)
+                .limit(2))
```
korthout, line 102: 🔧 There's a way to limit this on all records of the process instance
```suggestion
    assertThat(
            RecordingExporter.records()
                .limitToProcessInstance(processInstanceKey)
                .processInstanceRecords()
                .withIntent(ProcessInstanceIntent.ELEMENT_ACTIVATED)
                .withProcessInstanceKey(processInstanceKey)
                .withElementType(BpmnElementType.END_EVENT))
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateParallelGatewayTest.java
```diff
@@ -0,0 +1,108 @@
...
+        .extracting(r -> r.getValue().getElementId())
+        .describedAs(
+            "Expected to successfully evaluate execution listener job type and resolve incident")
+        .contains("end2");
```
korthout, line 106: ❓ I don't understand the last assertion. Why only check that one of the two end events was reached? Why not both?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
```diff
@@ -57,7 +57,8 @@ public final class ProcessInstanceMigrationPreconditions {
           BpmnElementType.BUSINESS_RULE_TASK,
           BpmnElementType.SCRIPT_TASK,
           BpmnElementType.SEND_TASK,
-          BpmnElementType.MULTI_INSTANCE_BODY);
+          BpmnElementType.MULTI_INSTANCE_BODY,
+          BpmnElementType.PARALLEL_GATEWAY);
```
korthout, line 61: 💭 Initially, this change worried me as I thought it opens up the possibility to migrate process instances with already taken incoming sequence flows of joining parallel gateways. However, that's still blocked by the concurrent command check:

>Expected to migrate process instance '2251799813685253' but a concurrent command was executed on the process instance. Please retry the migration.

korthout (approved):
LGTM 👍 

🔧 I have a few small comments.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateParallelGatewayTest.java
berkaycanbc: I attached other end event to the gateway only for using the parallel gateway in its purpose (running flows in parallel). I expect validating that we are able to migrate the incident on the gateway and being able to move forward in the flow after resolving the incident should be sufficient to test the intended behaviour.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
berkaycanbc: Yes, it is. I also checked that validation logic beforehand.

PR #25701, opened 2024-12-06
feat: add local variables search

### LINKED ISSUE #25292: Implement Iteration 2 (localVariables)
Adjust Implementation of C8 REST APIs to achieve [Iteration 2](https://docs.google.com/document/d/1YtC5W3Ln93M-BTwzq1iC-Ne1Qlvf5C7zqeo0DK6pjKc/edit?tab=t.0)

**Acceptance Criteria**

- [x]  When having a processInstance containing a process variable, when searching by local variables in the request, even 
when using the process variable and name properly, the user task should NOT be returned as part of the user task search response
- [x] When having a process instance containing a user task with a local variable, when search by this local variable name and value properly as a "localVariable" the user task should be returned

nathansandi:
## Description

- Include local variables search gateway/client
- Include integration tests for local search
- Include integration tests for process instance variable
- Add unit tests to the local variable

closes https://github.com/camunda/camunda/issues/25292


qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/usertask/UserTaskIT.java
```diff
@@ -208,6 +209,43 @@ public void shouldFindUserTaskByVariableName(final CamundaRdbmsTestApplication t
     assertUserTaskEntity(searchResult.hits().getFirst(), randomizedUserTask);
   }
 
+  @TestTemplate
+  void shouldFindUserTaskByLocalVariableName(final CamundaRdbmsTestApplication testApplication) {
```
nathansandi, line 213: I am aware of this test is failing because this:

`testApplication.getRdbmsService();` is returning null, the test is working locally. I am investigating the reason

panagiotisgts (commented):
@nathansandi Shall we include a test case with the two filters in conjunction and ensure that we don't receive duplicate items? Or are these filtered afterwards to be unique?

panagiotisgts (approved):
LGTM

nathansandi:
> @nathansandi Shall we include a test case with the two filters in conjunction and ensure that we don't receive duplicate items? Or are these filtered afterwards to be unique?

Thats a great question. Since the query is ultimately transformed into a single unified query, there’s no possibility of duplicate values in the result, as we are only returning tasks in an AND query. I also will add an integration test in order to validate this point.

PR #25703, opened 2024-12-06
feat: initialize identity even when authorizations are disabled

### LINKED ISSUE #25700: [General][Engine] create default user independently from authorizations enabled
### Description
Currently, when the single application starts we [initialise Identity with a default user](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/IdentitySetupInitializer.java) however this is only run when authorizations are enabled. 

The creation of a user should be independent of authorizations being enabled.

### Acceptance criteria
- [ ] When the single application is started, the demo user is created (with role)

lenaschoenburg:
This changes the identity initialization to _always_ run, even when authorizations are disabled. This is so that Identity can always see the default user and role.

:thought_balloon: Initially the issue was just about creating the default user and role. We now also include the role's wildcard permissions and the default tenant.

closes #25700

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/IdentitySetupInitializeProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/IdentitySetupInitializeProcessor.java
@@ -13,6 +13,7 @@ import io.camunda.zeebe.engine.processing.ExcludeAuthorizationCheck;
 import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
 import io.camunda.zeebe.engine.processing.streamprocessor.DistributedTypedRecordProcessor;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedRejectionWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
 import io.camunda.zeebe.engine.state.authorization.PersistedMapping;
 import io.camunda.zeebe.engine.state.distribution.DistributionQueue;
@@ -29,6 +30,7 @@ import io.camunda.zeebe.protocol.impl.record.value.authorization.Permission;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
 import io.camunda.zeebe.protocol.impl.record.value.tenant.TenantRecord;
 import io.camunda.zeebe.protocol.impl.record.value.user.UserRecord;
+import io.camunda.zeebe.protocol.record.RejectionType;
 import io.camunda.zeebe.protocol.record.intent.AuthorizationIntent;
 import io.camunda.zeebe.protocol.record.intent.IdentitySetupIntent;
 import io.camunda.zeebe.protocol.record.intent.MappingIntent;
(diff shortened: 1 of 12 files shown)


remcowesterhoud:
@koevskinikola Do you have capacity to review this?

koevskinikola:
@remcowesterhoud not today. Is tomorrow (12.12.2024) okay?

remcowesterhoud:
I won't get around to this today either, so that's fine with me. @lenaschoenburg hope you're fine with it too 😄

koevskinikola:
@lenaschoenburg @remcowesterhoud FYI to avoid duplicate reviews, I'll start reviewing this PR now.

koevskinikola (approved):
LGTM @lenaschoenburg  🎉

lenaschoenburg:
Thanks for the review! I rebased to resolve conflicts with the new configurable default user creation.

lenaschoenburg:
In combination with https://github.com/camunda/camunda/commit/768a1980461f2d0edc6e90220be55ad5dea4de2f, this breaks the `ReplayStateTest` because a restart will again write `IdentitySetup/Initialize` which then generates a new key to write `IdentitySetup/Initialized`.
I'm not sure what to do about this, I guess we need to re-introduce a duplication check somewhere? cc @remcowesterhoud and @koevskinikola

remcowesterhoud:
> In combination with [768a198](https://github.com/camunda/camunda/commit/768a1980461f2d0edc6e90220be55ad5dea4de2f), this breaks the `ReplayStateTest` because a restart will again write `IdentitySetup/Initialize` which then generates a new key to write `IdentitySetup/Initialized`. I'm not sure what to do about this, I guess we need to re-introduce a duplication check somewhere? cc @remcowesterhoud and @koevskinikola

Can we somehow detect if the `auth-basic` or `auth-oidc` profile is set? If we can we could use this to determine whether this initializer should run.

lenaschoenburg:
> Can we somehow detect if the auth-basic or auth-oidc profile is set? If we can we could use this to determine whether this initializer should run.

I don't think so, we want to keep Spring out of the engine.

remcowesterhoud:
> > Can we somehow detect if the auth-basic or auth-oidc profile is set? If we can we could use this to determine whether this initializer should run.
> 
> I don't think so, we want to keep Spring out of the engine.

If we can set a property depending on the profile we could use it to set a config option which we could access in the engine. Idk if that's possible though, let me see.

lenaschoenburg:
Sure but is that really the right direction? I'd rather not disable this in tests, otherwise bugs such as https://github.com/camunda/camunda/pull/25752 become really difficult to spot before they get released.

lenaschoenburg:
Or reintroduce the checks that prevent the `IdentitySetupInitializeProcessor` generates new keys if it is not needed?

remcowesterhoud:
You'd still be writing the records. It's pretty annoying having do consider those records in every single test case we have. Personally I still feel disabling it depending on the profile is the way to go

lenaschoenburg:
> It's pretty annoying having do consider those records in every single test case we have

That's true. It's partially solved now because `EngineRule` awaits the setup and resets the `RecordingExporter`

lenaschoenburg:
@remcowesterhoud d18b41f6230702b99cc334a6ff123765a03f118c contains the necessary changes to reject the `Initialize` command when all of the default entities already exist. That way we don't always generate new keys.

remcowesterhoud (approved):
Seems like a decent solution to me without having to duplicate it all in the initializer 👍

PR #25728, opened 2024-12-06
fix: only create one CREATED event per form on distribution

### LINKED ISSUE #25727: Same-name form resources inside deployments overridden with unexpected content on deployment distribution
**Describe the bug**

When multiple form resources in the same deployment have the same name and different content including different `formId`s, their content can be mixed up with that of any of the other forms.

**To Reproduce**

1. Deploy multiple forms with the same names but different form IDs and content.
2. Link the forms to different user tasks.
3. All forms have the same form content or mixed-up content with other forms.

**Expected behavior**

Each form has the expected and defined content.

**Root cause**

Similar to what was identified in https://github.com/camunda/camunda/issues/19834, we check resources based on resource names to decide on CREATED events to write. This leads to too many events for each form, creating as many CREATED events for each form resource as there are other form resources with the same name.

**Environment:**
- OS: 
- Zeebe Version: 
- Configuration: 

**Links**
* Related to https://jira.camunda.com/browse/SUPPORT-24633

tmetzke:
## Description

Upon deployment CREATE distribution, we only create one CREATED event per form with the defined
resource content per partition. We use the checksum to identify the unqiue form resources instead
of relying on resource names as those can occur multiple times per deployment.

Otherwise, we are creating too many CREATED events for each form resource, one for each other form
with the same name and the content of that other form. This can lead to forms having the wrong
content in the end, depending on which event was processed last for a form.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25727

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentCreateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentCreateProcessor.java
@@ -267,7 +267,9 @@ public final class DeploymentCreateProcessor
         .forEach(
             metadata -> {
               for (final DeploymentResource resource : deploymentEvent.getResources()) {
-                if (resource.getResourceName().equals(metadata.getResourceName())) {
+                final var resourceChecksum =
+                    deploymentTransformer.getChecksum(resource.getResource());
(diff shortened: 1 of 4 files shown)


remcowesterhoud (approved):
Thanks for the fix @tmetzke 👍

PR #25742, opened 2024-12-08
feat: batch operation update finished count task

### LINKED ISSUE #24084: Rework how `BatchOperationEntity#operationsFinishedCount` is updated
### Description

### Solution

### TODO

sdorokhova:
PR contains implementation of the task for updating batch operations finished count and interface for repository.

related with #24084

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.tasks.batchoperations;
+
+import io.camunda.exporter.tasks.BackgroundTask;
+import io.camunda.exporter.tasks.batchoperations.BatchOperationUpdateRepository.DocumentUpdate;
+import io.camunda.webapps.schema.descriptors.operate.template.BatchOperationTemplate;
+import io.camunda.webapps.schema.entities.operation.BatchOperationEntity;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+
+public class BatchOperationUpdateTask implements BackgroundTask {
+
+  private final BatchOperationUpdateRepository batchOperationUpdateRepository;
+
+  private final Logger logger;
+
+  public BatchOperationUpdateTask(
+      final BatchOperationUpdateRepository batchOperationUpdateRepository, final Logger logger) {
+    this.batchOperationUpdateRepository = batchOperationUpdateRepository;
+    this.logger = logger;
+  }
+
+  @Override
+  public CompletionStage<Integer> execute() {
+
+    final var batchOperations = batchOperationUpdateRepository.getNotFinishedBatchOperations();
+
+    if (batchOperations.size() > 0) {
+
+      final var finishedOperationsCount =
+          batchOperationUpdateRepository.getFinishedOperationsCount(
+              batchOperations.stream()
+                  .map(BatchOperationEntity::getId)
+                  .collect(Collectors.toList()));
+
+      final var documentUpdates =
+          finishedOperationsCount.stream()
+              .map(
+                  d ->
+                      new DocumentUpdate(
+                          d.batchOperationId(),
+                          Map.of(
+                              BatchOperationTemplate.COMPLETED_OPERATIONS_COUNT,
+                              d.finishedOperationsCount())))
+              .toList();
+
+      if (documentUpdates.size() > 0) {
+        return CompletableFuture.completedFuture(
+            batchOperationUpdateRepository.bulkUpdate(documentUpdates));
+      } // else return 0
+    }
+
+    return CompletableFuture.completedFuture(0);
+  }
+}
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
@@ -19,6 +19,7 @@ import org.slf4j.Logger;
 
 public class BatchOperationUpdateTask implements BackgroundTask {
 
+  private static final int NO_UPDATES = 0;
   private final BatchOperationUpdateRepository batchOperationUpdateRepository;
 
   private final Logger logger;
@@ -36,14 +37,14 @@ public class BatchOperationUpdateTask implements BackgroundTask {
 
     if (batchOperations.size() > 0) {
 
-      final var finishedOperationsCount =
+      final var finishedSingleOperationsCount =
           batchOperationUpdateRepository.getFinishedOperationsCount(
               batchOperations.stream()
                   .map(BatchOperationEntity::getId)
                   .collect(Collectors.toList()));
 
       final var documentUpdates =
-          finishedOperationsCount.stream()
+          finishedSingleOperationsCount.stream()
               .map(
                   d ->
                       new DocumentUpdate(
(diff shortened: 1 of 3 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -0,0 +1,64 @@
...
+              .map(
+                  d ->
+                      new DocumentUpdate(
+                          d.batchOperationId(),
+                          Map.of(
+                              BatchOperationTemplate.COMPLETED_OPERATIONS_COUNT,
+                              d.finishedOperationsCount())))
+              .toList();
+
+      if (documentUpdates.size() > 0) {
```
EuroLew, line 56: use .isEmpty instead

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -0,0 +1,64 @@
...
+    this.batchOperationUpdateRepository = batchOperationUpdateRepository;
+    this.logger = logger;
+  }
+
+  @Override
+  public CompletionStage<Integer> execute() {
+
+    final var batchOperations = batchOperationUpdateRepository.getNotFinishedBatchOperations();
+
+    if (batchOperations.size() > 0) {
```
EuroLew, line 37: use !isEmpty instead. I would prefer a short circuit where if the `batchOperations` is empty we return `CompletedFuture.completeFuture` as this makes things easier to read with less indentation

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -0,0 +1,64 @@
...
+                              d.finishedOperationsCount())))
+              .toList();
+
+      if (documentUpdates.size() > 0) {
+        return CompletableFuture.completedFuture(
+            batchOperationUpdateRepository.bulkUpdate(documentUpdates));
+      } // else return 0
+    }
+
+    return CompletableFuture.completedFuture(0);
```
EuroLew, line 62: Instead of using a magic number maybe an enum e.g. `NO_UPDATES`, `UPDATES_APPLIED`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -0,0 +1,64 @@
...
+  private final Logger logger;
+
+  public BatchOperationUpdateTask(
+      final BatchOperationUpdateRepository batchOperationUpdateRepository, final Logger logger) {
+    this.batchOperationUpdateRepository = batchOperationUpdateRepository;
+    this.logger = logger;
+  }
+
+  @Override
+  public CompletionStage<Integer> execute() {
```
EuroLew, line 33: ❓  Do we have worries about any part of `execute` failing? I assume the `BatchOperationUpdateRepository` makes calls to something and is not entirely in memory, or will everything be caught and handled by the individual method themselves?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTaskTest.java
```diff
@@ -0,0 +1,93 @@
...
+    // then
+    assertThat(result).succeedsWithin(Duration.ZERO).isEqualTo(2);
+    assertThat(repository.documentUpdates).hasSize(2);
+    assertThat(repository.documentUpdates)
+        .contains(
+            new DocumentUpdate("1", Map.of(BatchOperationTemplate.COMPLETED_OPERATIONS_COUNT, 5L)),
+            new DocumentUpdate("2", Map.of(BatchOperationTemplate.COMPLETED_OPERATIONS_COUNT, 6L)));
+  }
+
+  private static final class TestRepository implements BatchOperationUpdateRepository {
```
EuroLew, line 68: 🔧  I would prefer use of mock objects instead of an actual implementation as it avoids coupling tests with the implementation logic of the `TestRepository`.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -0,0 +1,64 @@
...
+                      new DocumentUpdate(
+                          d.batchOperationId(),
+                          Map.of(
+                              BatchOperationTemplate.COMPLETED_OPERATIONS_COUNT,
+                              d.finishedOperationsCount())))
+              .toList();
+
+      if (documentUpdates.size() > 0) {
+        return CompletableFuture.completedFuture(
+            batchOperationUpdateRepository.bulkUpdate(documentUpdates));
```
EuroLew, line 58: ❓  I'm assuming `bulkUpdate` interacts with ES/OS in this case with the return type as the boxed Integer, can the return value be null, or it throws an exception? Or will this again be handled by the method itself.

EuroLew (commented):
Just some small changes and I have some questions with regards to the critical path of execution and how errors would be handled

npepinpe (commented):
Looks good in general, my main concern is that it's entirely blocking when it comes to the interaction with the DB. I would encourage to avoid this, as these are mostly I/O bound tasks, and blocking on them ends up pinning the threads for the other tasks, potentially for quite a while.

It should be very small to refactor and support asynchronous, but I understand it's very different when you're not used to it. Happy to pair on it :+1:

sdorokhova:
@npepinpe exactly, it was not obvious for me, how to make it at once, so I decided to go step by step and first implement all calls as blocking. I actually have a TODO item on the issue to make things async, I should have mentioned it in this PR also.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
sdorokhova: Did that in the next PR: https://github.com/camunda/camunda/pull/25743/files#diff-51b7f2ad9c4e91779d547c372e43518ec31630e171bdff563bf3932dab78fcd3R37

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTaskTest.java
sdorokhova: I believe @npepinpe used repository pattern in order to avoid mocks in tests :) So I just followed the same approach as in other tests like [IncidentUpdateTaskTest](https://github.com/camunda/camunda/blob/814007d7109693278ef6e152b0298af4c84cd304/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/incident/IncidentUpdateTaskTest.java#L84).
I also find it easier to assert method parameters this way.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -0,0 +1,64 @@
...
+public class BatchOperationUpdateTask implements BackgroundTask {
+
+  private final BatchOperationUpdateRepository batchOperationUpdateRepository;
+
+  private final Logger logger;
+
+  public BatchOperationUpdateTask(
+      final BatchOperationUpdateRepository batchOperationUpdateRepository, final Logger logger) {
+    this.batchOperationUpdateRepository = batchOperationUpdateRepository;
+    this.logger = logger;
```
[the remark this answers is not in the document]
sdorokhova: Added the trace logging.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
sdorokhova: That's a valid point! I just added another TODO item to the issue and I would like to take care of this after I make thigs asynchronous. The TODO item sounds like: "think through error processing. If some internal call (e.g. to ES/OS) fails, we must make sure that error is properly logged and task is retried"

EuroLew (approved):
LGTM

PR #25743, opened 2024-12-08
feat: implement `BatchOperationUpdateRepository` repository for Elasticsearch

### LINKED ISSUE #24084: Rework how `BatchOperationEntity#operationsFinishedCount` is updated

sdorokhova:
This PR contains implementation of `BatchOperationUpdateRepository` for Elasticsearch. I had to refactor the code a bit, so that I could reuse the code for Elastic scrolling. 

Next steps in separate PRs:
* convert logic to async run
* implement the same for Opensearch
* (separate task) refactor Repository classes to use abstracted data access layer

related with #24084

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
@@ -8,6 +8,7 @@
 package io.camunda.exporter.tasks;
 
 import io.camunda.exporter.tasks.archiver.ArchiverRepository;
+import io.camunda.exporter.tasks.batchoperations.BatchOperationUpdateRepository;
 import io.camunda.exporter.tasks.incident.IncidentUpdateRepository;
 import io.camunda.zeebe.util.CloseableSilently;
 import io.camunda.zeebe.util.VisibleForTesting;
@@ -22,6 +23,7 @@ public final class BackgroundTaskManager implements CloseableSilently {
   private final int partitionId;
   private final ArchiverRepository archiverRepository;
   private final IncidentUpdateRepository incidentRepository;
+  private final BatchOperationUpdateRepository batchOperationUpdateRepository;
   private final Logger logger;
   private final ScheduledThreadPoolExecutor executor;
   private final List<Runnable> tasks;
@@ -33,6 +35,7 @@ public final class BackgroundTaskManager implements CloseableSilently {
       final int partitionId,
       final @WillCloseWhenClosed ArchiverRepository archiverRepository,
       final @WillCloseWhenClosed IncidentUpdateRepository incidentRepository,
+      final @WillCloseWhenClosed BatchOperationUpdateRepository batchOperationUpdateRepository,
       final Logger logger,
       final @WillCloseWhenClosed ScheduledThreadPoolExecutor executor,
       final List<Runnable> tasks) {
@@ -41,6 +44,9 @@ public final class BackgroundTaskManager implements CloseableSilently {
         Objects.requireNonNull(archiverRepository, "must specify an archiver repository");
     this.incidentRepository =
         Objects.requireNonNull(incidentRepository, "must specify an incident repository");
+    this.batchOperationUpdateRepository =
+        Objects.requireNonNull(
+            batchOperationUpdateRepository, "must specify a batch operation update repository");
     this.logger = Objects.requireNonNull(logger, "must specify a logger");
     this.executor = Objects.requireNonNull(executor, "must specify an executor");
     this.tasks = Objects.requireNonNull(tasks, "must specify tasks");
@@ -59,7 +65,8 @@ public final class BackgroundTaskManager implements CloseableSilently {
     CloseHelper.closeAll(
         error -> logger.warn("Failed to close resource for partition {}", partitionId, error),
         archiverRepository,
-        incidentRepository);
+        incidentRepository,
+        batchOperationUpdateRepository);
   }
 
   public void start() {
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
@@ -44,9 +44,12 @@ public final class BackgroundTaskManager implements CloseableSilently {
         Objects.requireNonNull(archiverRepository, "must specify an archiver repository");
     this.incidentRepository =
         Objects.requireNonNull(incidentRepository, "must specify an incident repository");
-    this.batchOperationUpdateRepository =
-        Objects.requireNonNull(
-            batchOperationUpdateRepository, "must specify a batch operation update repository");
+    // TODO enable null check when Opensearch is implemented
+    this.batchOperationUpdateRepository = batchOperationUpdateRepository;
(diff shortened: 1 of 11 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -33,25 +31,16 @@ public BatchOperationUpdateTask(
...
-    if (batchOperations.size() > 0) {
+    if (!batchOperationIds.isEmpty()) {
 
-      final var finishedSingleOperationsCount =
-          batchOperationUpdateRepository.getFinishedOperationsCount(
-              batchOperations.stream()
-                  .map(BatchOperationEntity::getId)
-                  .collect(Collectors.toList()));
+      final List<OperationsAggData> finishedSingleOperationsCount =
+          batchOperationUpdateRepository.getFinishedOperationsCount((List) batchOperationIds);
```
mustafadagher, line 39: 🔧 `getFinishedOperationsCount()` can be altered to receive a `Collection` as a parameter too, to get rid of this unchecked cast.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -33,25 +31,16 @@ public BatchOperationUpdateTask(
   @Override
   public CompletionStage<Integer> execute() {
 
-    final var batchOperations = batchOperationUpdateRepository.getNotFinishedBatchOperations();
+    final var batchOperationIds = batchOperationUpdateRepository.getNotFinishedBatchOperations();
 
-    if (batchOperations.size() > 0) {
+    if (!batchOperationIds.isEmpty()) {
```
mustafadagher, line 36: 🔧 this can be inverted for better readability and less nesting

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/BatchOperationUpdateTask.java
```diff
@@ -33,25 +31,16 @@ public BatchOperationUpdateTask(
...
-                  d ->
-                      new DocumentUpdate(
-                          d.batchOperationId(),
-                          Map.of(
-                              BatchOperationTemplate.COMPLETED_OPERATIONS_COUNT,
-                              d.finishedOperationsCount())))
+              .map(d -> new DocumentUpdate(d.batchOperationId(), d.finishedOperationsCount()))
               .toList();
 
       if (documentUpdates.size() > 0) {
```
mustafadagher, line 46: 🔧 this can be inverted as well, similar to the one above

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
```diff
@@ -41,6 +44,12 @@ public final class BackgroundTaskManager implements CloseableSilently {
         Objects.requireNonNull(archiverRepository, "must specify an archiver repository");
     this.incidentRepository =
         Objects.requireNonNull(incidentRepository, "must specify an incident repository");
+    // TODO enable null check when Opensearch is implemented
```
mustafadagher, line 47: 🔧 we can re-enable the null check and yield a `noop` repository in case of opensearch for now, instead of null.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/ElasticsearchBatchOperationUpdateRepository.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  @Override
+  public Collection<String> getNotFinishedBatchOperations() {
+    final var request =
+        new SearchRequest.Builder()
+            .index(batchOperationIndex)
+            .query(q -> q.bool(b -> b.mustNot(m -> m.exists(e -> e.field(END_DATE)))));
+    return ElasticsearchUtil.fetchUnboundedDocumentCollection(
+            client, executor, logger, request, BatchOperationEntity.class, hit -> hit.id())
```
mustafadagher, line 69:
```suggestion
            client, executor, logger, request, BatchOperationEntity.class, Hit::id)
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/ElasticsearchBatchOperationUpdateRepository.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  @Override
+  public List<OperationsAggData> getFinishedOperationsCount(final List<String> batchOperationIds) {
+    if (batchOperationIds == null || batchOperationIds.isEmpty()) {
+      return List.of();
+    }
+    final var batchOperationIdsValues = batchOperationIds.stream().map(FieldValue::of).toList();
+    final var completedStatesValues =
+        List.of(COMPLETED, FAILED).stream().map(FieldValue::of).toList();
```
mustafadagher, line 81:
```suggestion
        Stream.of(COMPLETED, FAILED).map(FieldValue::of).toList();
```

mustafadagher (approved):
Thanks @sdorokhova , I went through it and it looks good to me! 🚀 

I only left some cosmetic comments below

mustafadagher:
I see now that many of my comments above are already applied in the next PR.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
sdorokhova: This is already addresses in [the follow-up PR](https://github.com/camunda/camunda/pull/26613/files#diff-0e84de386774d12d3a5674b174883975ff0c32414f6867e134eaf4d7f67722c8R47), won't fix here to avoid conflicts

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/ElasticsearchBatchOperationUpdateRepository.java
sdorokhova: This is addresses in [the follow-up PR](https://github.com/camunda/camunda/pull/26603/files#diff-cde0b7337a19c67ccd668de04bf06dc40b7c9d6de2a90613e489cdaa6a027112R72), won't fix here to avoid conflicts

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/batchoperations/ElasticsearchBatchOperationUpdateRepository.java
sdorokhova: This is addresses in [the follow-up PR](https://github.com/camunda/camunda/pull/26603/files#diff-cde0b7337a19c67ccd668de04bf06dc40b7c9d6de2a90613e489cdaa6a027112R82), won't fix here to avoid conflicts

PR #25749, opened 2024-12-09
feat: apply user roles in authorizations checks

### LINKED ISSUE #25751: Apply user roles in authorizations checks
### Description:
Currently, authorization checks do not consider user roles stored in the database. This feature aims to integrate user roles into the authorization logic.

For `auth-basic` implementation, we need to:
* Fetch user all roles from the database
* Feed user roles in the `CamundaUser` object fed in spring context.
* Add role keys in `ownerKey` filter applied to fetch the authorizations object.
* Make sure Operate V1/internal endpoints use these logic

houssain-barouni:
## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/25751


authentication/src/main/java/io/camunda/authentication/entity/CamundaUser.java
```diff
@@ -54,15 +48,17 @@ public CamundaUser(
       final String displayName,
       final String username,
       final String password,
-      final List<String> roles,
+      final List<? extends GrantedAuthority> authorities,
```
houssain-barouni, line 51: @Ben-Sheppard @maryarm Here I have separated between Roles and spring's authority concept. can you confirm that the spring's `GrantedAuthority` concept will not be used with `auth-basic` profile? and will maybe kept for the default `auth` profile, that is used by Operate and Tasklist webapps?

Ben-Sheppard: Separating these IMO is fine, the old `auth` profile will not be kept so this field likely won't be used

Ben-Sheppard (approved):
There are quite a few changes in here however I did checkout the code locally to test and I checked that with authorizations enabled my roles were used to apply authorizations (for example checking no read for roles and what the Identity UI does, then assigning the authorization to the role) and it worked as expected :)

Whilst testing I did see that there seems to be an issue with the default `Admin` role, its not found with this code but I believe that is due to the join relationship and id routing used. If I create a role and assign it to a user it works fine. I'll follow up separately with that issue.

authentication/src/main/java/io/camunda/authentication/service/BasicCamundaUserService.java
```diff
@@ -32,7 +33,7 @@ public CamundaUserDTO getCurrentUser() {
         authenticatedUser.getAuthorizedApplications(),
         authenticatedUser.getTenants(),
         authenticatedUser.getGroups(),
-        authenticatedUser.getRoles(),
+        authenticatedUser.getRoles().stream().map(RoleEntity::name).toList(),
```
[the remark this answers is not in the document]
houssain-barouni: Actually, the `CamundaUserDTO` is used `GET /authentication/me` that returns roles and groups as List of Strings. Probably it is required to be this way by the UI https://github.com/camunda/camunda/blob/e528830a7bb8e4b7749108431a1e51da913309ef/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L4962-L4966

PR #25752, opened 2024-12-09
fix: do not generate keys in scheduled task to setup identity

lenaschoenburg:
Scheduled tasks are not allowed to modify the state. The identity setup relied on this to generate the keys for the default entities.

This is now fixed by generating the keys on the deployment partition when processing the (now key-less) `IdentitySetup/Initialize` command.


remcowesterhoud (approved):
LGTM! I'll likely make some changes in a next PR because I need to change the initializer to support multiple configured users. But this is a good baseline 👍

PR #25762, opened 2024-12-09
Allow configuration of default user credentials

### LINKED ISSUE #22572: Default user credentials should be configurable
<h2>Description</h2>

Currently the credentials for the default user are hardcoded to `demo/demo`. These credentials should become configurable by users.

remcowesterhoud:
## Description

This PR gives users the option to configure the credentials, name and email of the default user that's created once the Broker is started. The properties are set under:

- `camunda.security.initialization.defaultuser.username`
- `camunda.security.initialization.defaultuser.password`
- `camunda.security.initialization.defaultuser.name`
- `camunda.security.initialization.defaultuser.email`

## Checklist

- [ ] for CI changes:

## Related issues

closes #22572

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/utils/ZeebeClientTestFactory.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/utils/ZeebeClientTestFactory.java
@@ -7,9 +7,6 @@
  */
 package io.camunda.it.utils;
 
-import static io.camunda.zeebe.engine.processing.user.IdentitySetupInitializer.DEFAULT_USER_PASSWORD;
-import static io.camunda.zeebe.engine.processing.user.IdentitySetupInitializer.DEFAULT_USER_USERNAME;
-
 import io.camunda.zeebe.client.CredentialsProvider;
 import io.camunda.zeebe.client.ZeebeClient;
 import io.camunda.zeebe.client.protocol.rest.PermissionTypeEnum;
@@ -34,6 +31,9 @@ import org.slf4j.LoggerFactory;
 public final class ZeebeClientTestFactory implements AutoCloseable {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ZeebeClientTestFactory.class);
+
+  private static final String DEFAULT_USER_USERNAME = "demo";
+  private static final String DEFAULT_USER_PASSWORD = "demo";
 
   private final Map<String, User> usersRegistry = new HashMap<>();
   private final Map<String, ZeebeClient> cachedClients = new ConcurrentHashMap<>();
--- a/qa/integration-tests/src/test/java/io/camunda/it/utils/ZeebeClientTestFactory.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/utils/ZeebeClientTestFactory.java
@@ -7,6 +7,7 @@
  */
 package io.camunda.it.utils;
 
+import io.camunda.security.configuration.InitializationConfiguration;
 import io.camunda.zeebe.client.CredentialsProvider;
 import io.camunda.zeebe.client.ZeebeClient;
 import io.camunda.zeebe.client.protocol.rest.PermissionTypeEnum;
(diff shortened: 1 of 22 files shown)


qa/integration-tests/src/test/java/io/camunda/it/utils/ZeebeClientTestFactory.java
```diff
@@ -31,6 +28,9 @@
 
 public final class ZeebeClientTestFactory implements AutoCloseable {
 
+  private static final String DEFAULT_USER_USERNAME = "demo";
+  private static final String DEFAULT_USER_PASSWORD = "demo";
```
nicpuppa, line 32: ❓ Can we move this variables in the configuration class, and use them everywhere ? Not sure if it make sense

nicpuppa (approved):
LGTM, just a comment. Please have a look 👍

remcowesterhoud:
@nicpuppa I changed the implementation so it now support configuration of multiple users. Would you mind having another look?

remcowesterhoud:
Regarding the failing RDBMS IT: https://camunda.slack.com/archives/C071KP5BTHB/p1733925438712159

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/IdentitySetupInitializer.java
```diff
@@ -31,26 +27,16 @@
...
-      final SecurityConfiguration securityConfig, final MutableProcessingState processingState) {
+  public IdentitySetupInitializer(final SecurityConfiguration securityConfig) {
     this.securityConfig = securityConfig;
     passwordEncoder = PasswordEncoderFactories.createDelegatingPasswordEncoder();
-    userState = processingState.getUserState();
-    roleState = processingState.getRoleState();
-    tenantState = processingState.getTenantState();
   }
 
   @Override
```
houssain-barouni, line 42: @remcowesterhoud , is this condition still valid?
```
    if (!securityConfig.getAuthorizations().isEnabled()) {
      // If authorization is disabled we don't need to setup identity.
      LOG.debug("Skipping identity setup as authorization is disabled");
      return;
    }
```

how can I call the API if I don't have some initial credentials to pass the authentication checks?

houssain-barouni: when I use `auth-basic` profile, authentication checks are [enabled](https://github.com/camunda/camunda/blob/c4a7a53e628d8c0b13ff3e94570ed983db738062/authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java#L100-L138), any calls to authenticated paths like `/v2/**` should be authenticated.

When I use an unauthenticated client I get:
```
DEBUG org.springframework.web.servlet.mvc.method.annotation.HttpEntityMethodProcessor - Writing [{timestamp=Wed Dec 11 16:08:56 CET 2024, status=401, error=Unauthorized, path=/v2/users}]
```

remcowesterhoud: @houssain-barouni I'm confused what your point is here? With the `auth-basic` profile authorization checks are enabled. That makes sense because you say you want to use basic auth. Since it enables authentications a default user is created. If you use an unauthenticated client of course you'll be unauthorized. You just enabled basic auth so you need to add that to the client?

houssain-barouni: @remcowesterhoud yes, there is no correlation between `auth-basic` spring profile and authorizations
`auth-basic` just means that the users are created an managed inside Camunda, and not by a third-party OIDC or LDAP.
we can use `auth-basic` without having authorizations enabled. auth stands for authentication.

remcowesterhoud: So there is no way to run Camunda without authentication anymore? That seems not very user friendly?

houssain-barouni: I think we may have another profile reserved for that @Ben-Sheppard .
Anyway, I think we can have a property `camunda.security.initialization.defaultuser.enabled` set to `false` by default and set to `true` for `auth-basic`. I think for the oidc case we don t need to create a default user (maybe you need to create default mapping roles...)

remcowesterhoud: > Anyway, I think we can have a property `camunda.security.initialization.defaultuser.enabled` set to false

The `defaultUser` property is gone. It's now just a list of users. If no users are configured it will create a default demo/demo one.

nicpuppa (approved):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/IdentitySetupInitializer.java
houssain-barouni: >The defaultUser property is gone. It's now just a list of users. If no users are configured it will create a default demo/demo one.

We need another condition then to disable the data initialization

Issue #25760, opened 2024-12-09
[C8 API] Add `Long` serializer for key conversion


danielkelemen:

danielkelemen:
### Notes

- > Check if we need an equivalent LongDeserializer, too.
  - Not necessary, Jackson right now is configured to automatically do the type conversion if needed. I adjusted a few tests to send the key as String and they work fine, see the PR ([Ref](https://github.com/camunda/camunda/blob/dd440c05712ff2c9eaeb9a3c32b5bc7d6a5eb06a/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/DecisionDefinitionControllerTest.java#L128)).
- The `LongSerializer`
  - Checks headers for exact match, so users can't combine headers. Examples:
    - `Accept: application/json,application/vnd.camunda.api.keys.string +json` -> Serializes to default (8.6 number, 8.7 string) due to multiple accepts.
    - `Accept: application/json,application/vnd.camunda.api.keys.STRING+json` -> Serializes to default (8.6 number, 8.7 string) due to uppercase.
    - `Accept: application/vnd.camunda.api.keys.string+json` -> Serializes to string. ✅

PR #25766, opened 2024-12-09
feat: add Long serializer for REST API keys conversion

danielkelemen:
## Description

- Backport jackson configs & default OM config
- Add Long serializer for REST API keys
- Introduce new media types for string and numeric keys & adapt V2 APIs.
- Tests

## Related issues

closes https://github.com/camunda/camunda/issues/25760


tmetzke (approved):
👍 That is really neat ⭐ 
💡 It was news to me that Jackson simply converts everything on deserialization on best effort (e.g. strings to longs if it can) when requests come in.
❓ Should we use the custom POST and GET annotations everywhere (plus, no PATCH variant exists)? I think it's only (but haven't checked all of them) endpoints that return 204 NO CONTENT or problem JSON on error don't have it yet. Functionally, it shouldn't create any issues for users then, I believe. Code-wise, it is a bit inconsistent then though but might just be OK. Let me know what you think.

tmetzke:
> > 💡 It was news to me that Jackson simply converts everything on deserialization on best effort (e.g. strings to longs if it can) when requests come in.
> 
> As long as it's not disabled! But we will notice if that happens with the tests.
> 
> > ❓ Should we use the custom POST and GET annotations everywhere (plus, no PATCH variant exists)? I think it's only (but haven't checked all of them) endpoints that return 204 NO CONTENT or problem JSON on error don't have it yet. Functionally, it shouldn't create any issues for users then, I believe. Code-wise, it is a bit inconsistent then though but might just be OK. Let me know what you think.
> 
> I think we could use the custom annotations everywhere. Default values can still be overridden, if needed.

Let's do that, then. It's gonna be a lot more convenient for users to simply use that as a default accept header and not fail with any endpoint.
Also, it's more consistent then in the code base.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/annotation/PostMappingStringKeys.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.core.annotation.AliasFor;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@RequestMapping(method = RequestMethod.POST)
+public @interface PostMappingStringKeys {
```
tmetzke, line 25: ❓ How about we use a generic name here like `CamundaRestPostMapping` or something similar? This would not make this so much about the string keys but simply create a default annotation that we can keep in the future. It simply provides all the defaults you usually need for an endpoint.

danielkelemen: Sounds good to me, I'll update it.

PR #25771, opened 2024-12-09
Support activating Inclusive Gateway via Process Instance Modification

### LINKED ISSUE #20587: A process instance is stuck after modifing to an inclusive gateway
**Describe the bug**
I have the following process with an inclusive gateway:

image

Given:

* The task `A` is active

When:
* Modify task `A` to `Gateway` through the `ModifyProcessInstance` RPC.

Then:
* The process instance is stuck at the inclusive gateway.

**To Reproduce**
1. Deploy the following process [inclusive-gateway-stuck.bpmn](https://github.com/user-attachments/files/16411829/inclusive-gateway-stuck.bpmn.txt)
2. Modify task `A` to `Gateway` via `Operate`

**Expected behavior**
The inclusive gateway should be activated correctly.

**Environment:**
- OS:  
- Zeebe Version:  `8.6.0-SNAPSHOT`
- Configuration:

### LINKED ISSUE #25099: Process modification on inclusive gateway makes active token disappear
**Describe the bug**

Process modification on inclusive gateway makes active token disappear

**To Reproduce**

1. Navigate to process instance in Operate UI
2. Select modify instance wrench
3. On current activity, select move instance
4. Select the inclusive gateway
5. Apply modifications
6. No active token appears but the process continues

**Expected behavior**

Active token appears on current process or user is not allowed to make this action.

**Environment:**
C8 SaaS 8.6

JIRA ticket for reference:

https://jira.camunda.com/browse/SUPPORT-24445

korthout:
## Description

Adds support for activating Inclusive Gateway via a Process Instance Modification.

Specifically, adds the ability to reference an Inclusive Gateway in the activate instructions of a Process Instance Modification command. Previously, such activation instructions had no effect because the `ACTIVATE_ELEMENT` command for the gateway would be rejected.

By incrementing the number of times each of its incoming sequence flows have been taken, we ensure it isn't rejected. These recorded values are decremented again after the gateway has reached the `ELEMENT_ACTIVATING` lifecycle state.

These changes ensure that any previously taken sequence flows are retained after the modification has finished.

## Related issues

closes #20587 
closes #25099

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceModifiedEventApplier.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceModifiedEventApplier.java
@@ -68,7 +68,8 @@ final class ProcessInstanceModifiedEventApplier
             instruction -> {
               final var element =
                   process.getElementById(instruction.getElementId(), ExecutableFlowNode.class);
-              if (!element.getElementType().equals(BpmnElementType.PARALLEL_GATEWAY)) {
+              if (!element.getElementType().equals(BpmnElementType.PARALLEL_GATEWAY)
+                  && !element.getElementType().equals(BpmnElementType.INCLUSIVE_GATEWAY)) {
                 return;
               }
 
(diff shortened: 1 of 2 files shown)


tmetzke (approved):
👍 This is great, I really love that it is a one-liner to enable this 🧑‍🚀 
🌟 Great that you tested the joining behavior as well to ensure this works as expected 🚀

PR #25772, opened 2024-12-09
fix: exception type from ES is read correctly

entangled90:
## Description
After migration to ElasticsearchClient the type of the exception is not read correctly.
As a result, if a repository is missing an INTERNAL_SERVER_ERROR is returned to the client instead of NOT_FOUND.

BackupControllerIT has been fixed by migrating it to ElasticsearchClient as well

## Related issues

closes #25679

### DIFF
--- a/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
+++ b/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
@@ -14,14 +14,13 @@ import static java.util.stream.Collectors.toList;
 
 import co.elastic.clients.elasticsearch.ElasticsearchClient;
 import co.elastic.clients.elasticsearch._types.ElasticsearchException;
-import co.elastic.clients.elasticsearch._types.ShardFailure;
 import co.elastic.clients.elasticsearch._types.SortOrder;
 import co.elastic.clients.elasticsearch.snapshot.CreateSnapshotRequest;
 import co.elastic.clients.elasticsearch.snapshot.CreateSnapshotResponse;
-import co.elastic.clients.elasticsearch.snapshot.DeleteSnapshotResponse;
 import co.elastic.clients.elasticsearch.snapshot.GetSnapshotRequest;
 import co.elastic.clients.elasticsearch.snapshot.GetSnapshotResponse;
 import co.elastic.clients.elasticsearch.snapshot.SnapshotInfo;
+import co.elastic.clients.elasticsearch.snapshot.SnapshotShardFailure;
 import co.elastic.clients.elasticsearch.snapshot.SnapshotSort;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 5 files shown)


webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
```diff
@@ -312,10 +314,9 @@ public List<SnapshotInfo> findSnapshots(final String repositoryName, final Long
               "Encountered an error connecting to Elasticsearch while searching for snapshots. Repository name: [%s].",
               repositoryName);
       throw new BackupRepositoryConnectionException(reason, ex);
-    } catch (final Exception e) {
+    } catch (final ElasticsearchException e) {
```
entangled90, line 317: Previously the exceptions reaised from the try block where swallowed by this Exception clause and not rethrown

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/MetadataMarshaller.java
```diff
@@ -27,7 +26,12 @@ public static Map<String, JsonData> asJson(
   public static Metadata fromMetadata(
       final Map<String, JsonData> metadata, final JsonpMapper jsonpMapper) {
     try {
-      final var backupId = metadata.get("backupId").to(Long.class, jsonpMapper);
+      Long backupId = null;
+      try {
+        backupId = metadata.get("backupId").to(Long.class, jsonpMapper);
```
npepinpe, line 31: :wrench: If the exception is thrown because `metadata.get("backupId")` returns null, then let's check for that instead of triggering the exception and catching it =/

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/opensearch/MetadataMarshaller.java
```diff
@@ -26,7 +26,12 @@ public static Map<String, JsonData> asJson(
   public static Metadata fromMetadata(
       final Map<String, JsonData> metadata, final JsonpMapper jsonpMapper) {
     try {
-      final var backupId = metadata.get("backupId").to(Long.class, jsonpMapper);
+      Long backupId = null;
+      try {
+        backupId = metadata.get("backupId").to(Long.class, jsonpMapper);
```
npepinpe, line 31: :wrench: Same as above.

npepinpe (approved):

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/MetadataMarshaller.java
entangled90: I think the exception can be thrown by the `to` method as well in the tests as the `JsonData` is not null, but the value it wraps is.

entangled90: I added the null check as you suggested, but kept the try/catch for the reason explained in the comment above

npepinpe: That's why I asked if it came from `metadata.get` :sweat_smile:

entangled90: I think it may come from both :+1:

PR #25778, opened 2024-12-09
Add REST API endpoint for searching group-assigned users

### LINKED ISSUE #22387: Implement `group` endpoints
### Description

Group endpoints provides CRUD options for groups combined with assigning of additional resources i.e. users or roles.

* The endpoints should be implemented as described in [REST API Developer Guide](https://github.com/camunda/camunda/blob/main/docs/rest-controller.md)
* `{place-holder}` is the placeholder for parameters like `key`

### Endpoints

- [x] Create group - `POST /v2/groups` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.qc7j844hp0au)
- [x] Search groups - `POST /v2/groups/search` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.3r5lq4sb0tyu)
- [x] Get a single group - `GET /v2/groups/{key}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.vsmunaey9342)
- [x] Update a group - `PATCH /v2/groups/{key}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.27d3ji9wuci5)
- [x] Assign a user to a group - `POST /v2/groups/{key}/users/{userKey}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.7d4q4ehhqx7d)
- [x] Search users in a group - `POST /v2/groups/{key}/users/search` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.1425quaqoc48)
- [x] Remove a user from a group - `DELETE /v2/groups/{key}/users/{userKey}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.f2cf41s0545d)
- [x] Delete a group - `DELETE /v2/groups/{key}` - [API specification](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit#heading=h.oa81iyqm7b8)

### Acceptance Criteria

- [ ] Unit tests
- [ ] Integration tests

koevskinikola:
## Description

Adds a REST API endpoint for querying users assigned to a group with a given group key.

## Checklist

- [ ] for CI changes:

## Related issues

closes #22387

### DIFF
--- a/service/src/main/java/io/camunda/service/GroupServices.java
+++ b/service/src/main/java/io/camunda/service/GroupServices.java
@@ -8,11 +8,14 @@
 package io.camunda.service;
 
 import io.camunda.search.clients.GroupSearchClient;
+import io.camunda.search.clients.UserSearchClient;
 import io.camunda.search.entities.GroupEntity;
+import io.camunda.search.entities.UserEntity;
 import io.camunda.search.exception.NotFoundException;
 import io.camunda.search.query.GroupQuery;
 import io.camunda.search.query.SearchQueryBuilders;
 import io.camunda.search.query.SearchQueryResult;
+import io.camunda.search.query.UserQuery;
 import io.camunda.security.auth.Authentication;
 import io.camunda.security.auth.Authorization;
 import io.camunda.service.search.core.SearchQueryService;
@@ -27,19 +30,24 @@ import io.camunda.zeebe.protocol.impl.record.value.group.GroupRecord;
 import io.camunda.zeebe.protocol.record.value.EntityType;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 public class GroupServices extends SearchQueryService<GroupServices, GroupQuery, GroupEntity> {
 
   private final GroupSearchClient groupSearchClient;
+  private final UserSearchClient userSearchClient;
 
   public GroupServices(
       final BrokerClient brokerClient,
       final SecurityContextProvider securityContextProvider,
       final GroupSearchClient groupSearchClient,
+      final UserSearchClient userSearchClient,
       final Authentication authentication) {
     super(brokerClient, securityContextProvider, authentication);
     this.groupSearchClient = groupSearchClient;
+    this.userSearchClient = userSearchClient;
   }
 
   @Override
@@ -54,7 +62,7 @@ public class GroupServices extends SearchQueryService<GroupServices, GroupQuery,
   @Override
   public GroupServices withAuthentication(final Authentication authentication) {
     return new GroupServices(
-        brokerClient, securityContextProvider, groupSearchClient, authentication);
+        brokerClient, securityContextProvider, groupSearchClient, userSearchClient, authentication);
   }
 
   public CompletableFuture<GroupRecord> createGroup(final String name) {
@@ -72,6 +80,23 @@ public class GroupServices extends SearchQueryService<GroupServices, GroupQuery,
         .items()
         .stream()
         .toList();
+  }
+
+  public Set<Long> getUserKeysByGroupKey(final long groupKey) {
+    return search(SearchQueryBuilders.groupSearchQuery().filter(f -> f.groupKey(groupKey)).build())
+        .items()
+        .stream()
+        .map(GroupEntity::assignedMemberKeys)
+        .flatMap(Set::stream)
+        .collect(Collectors.toSet());
+  }
+
+  public SearchQueryResult<UserEntity> getUsersByGroupKey(final UserQuery userQuery) {
+    return userSearchClient
+        .withSecurityContext(
+            securityContextProvider.provideSecurityContext(
+                authentication, Authorization.of(a -> a.group().read())))
+        .searchUsers(userQuery);
   }
 
   public Optional<GroupEntity> findGroup(final Long groupKey) {
--- a/db/rdbms/src/main/resources/mapper/UserMapper.xml
+++ b/db/rdbms/src/main/resources/mapper/UserMapper.xml
@@ -35,7 +35,11 @@
   <sql id="searchFilter">
     WHERE 1 = 1
     <!-- basic filters -->
-    <if test="filter.key != null">AND USER_KEY = #{filter.key}</if>
+    <if test="filter.keys != null and !filter.keys.isEmpty()">
+      <foreach collection="filter.keys" item="key" open=" AND USER_KEY IN (" separator=", " close=" ) ">
+        #{key}
+      </foreach>
+    </if>
     <if test="filter.username != null">AND USERNAME = #{filter.username}</if>
     <if test="filter.name != null">AND NAME = #{filter.name}</if>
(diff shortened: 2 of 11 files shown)


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/GroupQueryController.java
```diff
@@ -55,6 +65,26 @@ public ResponseEntity<GroupSearchQueryResponse> searchGroups(
...
   }
 
+  @PostMapping(
+      path = "/{groupKey}/users/search",
+      produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
+      consumes = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<UserSearchResponse> searchUsersInGroup(
+      @PathVariable final long groupKey,
+      @RequestBody(required = false) final UserSearchQueryRequest query) {
+    final var groupMemberKeys =
```
emulbreh, line 75: I think this belongs to a different layer. It should be possible to get users by group programmatically, so it should be handled within a method either in UserServices or GroupServices. But as the strategy is specific to the limitations of ES/OS it actually should be encapsulated by the search client implementation so that the RDBMS implementation can use a plain join. 
There was a bit of a [discussion](https://camunda.slack.com/archives/C06UYJMMETZ/p1733489819132029?thread_ts=1733475623.559369&cid=C06UYJMMETZ) (with no conclusion) on Friday.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/GroupQueryController.java
```diff
@@ -55,6 +65,26 @@ public ResponseEntity<GroupSearchQueryResponse> searchGroups(
...
+        groupServices
+            .search(
+                SearchQueryBuilders.groupSearchQuery().filter(f -> f.groupKey(groupKey)).build())
```
emulbreh, line 78: This is limited by the page size, but we need all results here.

koevskinikola: That's a good point, thanks!

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/GroupQueryController.java
houssain-barouni: I agree that this should be moved to the service layer , there is a similar example for searching variables under user task scope [here](https://github.com/camunda/camunda/blob/e8369a7cc2a6b84d7c3534e2150dc9f209d6fbba/service/src/main/java/io/camunda/service/UserTaskServices.java#L153).

I think functionally it makes sense to first fetch the group, I expect we get:
- a 404 when a group with the provided groupKey does not exist. 
- a 403 if the caller does not have READ permission on this group
This would fetch the group member keys for free for ES, maybe it will be more costly for RDBMS.

I have no quick idea how to build a flexible strategy to apply group member keys filter, but we can do it later.

koevskinikola:
@emulbreh @houssain-barouni I finally managed to get some time and apply your feedback. Let me know if it looks okay now.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/GroupQueryController.java
koevskinikola: >a 404 when a group with the provided groupKey does not exist.
a 403 if the caller does not have READ permission on this group

AFAIK, these are automatically mapped when an exception is thrown from the client response.

service/src/main/java/io/camunda/service/GroupServices.java
```diff
@@ -74,6 +82,23 @@ public List<GroupEntity> getGroupsByUserKey(final long userKey) {
         .toList();
   }
 
+  public Set<Long> getUserKeysByGroupKey(final long groupKey) {
+    return search(SearchQueryBuilders.groupSearchQuery().filter(f -> f.groupKey(groupKey)).build())
```
emulbreh, line 86: We need to use `findAll()` here so we get all groups.

service/src/main/java/io/camunda/service/GroupServices.java
```diff
@@ -74,6 +82,23 @@ public List<GroupEntity> getGroupsByUserKey(final long userKey) {
...
+        .items()
+        .stream()
+        .map(GroupEntity::assignedMemberKeys)
+        .flatMap(Set::stream)
+        .collect(Collectors.toSet());
+  }
+
+  public SearchQueryResult<UserEntity> getUsersByGroupKey(final UserQuery userQuery) {
```
emulbreh, line 94: Isn't this just returning all matching users?

koevskinikola: Yes, from the discussion [here](https://github.com/camunda/camunda/pull/25778#discussion_r1878402048), we wanted to move the search logic into the service layer instead of the REST controller. 

We need some way to retrieve the `UserEntity` instances for users assigned to a certain group. We can either do this through a call for the user search client in the `GroupService` or the `UserService`. However, I'm not sure we want to pass a `UserService` instance to the `GroupService`, so we need to re-implement this method.

Or did I misunderstand your question?

emulbreh: The actual filtering logic is in the controller and mapper still: the construction of the userQuery is the important bit. `getUsersByGroupKey` doesn't actually filter by group even though its name suggests that. 

I believe it'd be good to move the query construction here. Alternatively, we could pass a set of user ids and make this an overload of UserServices::search that takes an extra collection of user keys: `search(Set<Long> userKeys, UserQuery query)`. 

Whatever path we choose, let's move this to UserServices as it takes a UserQuery and returns UserEntities.

remcowesterhoud:
@nicpuppa correct me if I'm wrong since you're more focussed on the groups part. It seems to me like it's going to be tackled with https://github.com/camunda/camunda/issues/30010

nicpuppa:
@remcowesterhoud correct

PR #25801, opened 2024-12-10
feat: implement tenants' mapping rules' migration

### LINKED ISSUE #25165: Camunda supports TenantMappingRule by implementing TenantMappingRuleMigrationHandler
### Description
Within scope of this issue, we assume that all available tenants are migrated and there is no need to create a tenant as part of migration. 
This issue needs [tenant migration task](https://github.com/camunda/camunda/issues/25130) be completed.

maryarm:
## Description

Migrate tenant mapping rules from Management identity to new identity.

## Related issues

closes #25165

### DIFF
--- a/migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
+++ b/migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migration.identity;
+
+import io.camunda.migration.identity.dto.MigrationStatusUpdateRequest;
+import io.camunda.migration.identity.dto.Tenant;
+import io.camunda.migration.identity.dto.TenantMappingRule;
+import io.camunda.migration.identity.midentity.ManagementIdentityClient;
+import io.camunda.migration.identity.midentity.ManagementIdentityTransformer;
+import io.camunda.migration.identity.service.MappingService;
+import io.camunda.migration.identity.service.TenantService;
+import io.camunda.search.entities.TenantEntity;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class TenantMappingRuleMigrationHandler implements MigrationHandler {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(TenantMappingRuleMigrationHandler.class);
+  private final ManagementIdentityClient managementIdentityClient;
+  private final ManagementIdentityTransformer managementIdentityTransformer;
+  private final TenantService tenantService;
+  private final MappingService mappingService;
+
+  public TenantMappingRuleMigrationHandler(
+      final ManagementIdentityClient managementIdentityClient,
+      final ManagementIdentityTransformer managementIdentityTransformer,
+      final TenantService tenantService,
+      final MappingService mappingService) {
+    this.managementIdentityClient = managementIdentityClient;
+    this.managementIdentityTransformer = managementIdentityTransformer;
+    this.tenantService = tenantService;
+    this.mappingService = mappingService;
+  }
+
+  @Override
+  public void migrate() {
+    List<TenantMappingRule> tenantMappingRules;
+    do {
+      tenantMappingRules = managementIdentityClient.fetchTenantMappingRules(SIZE);
+      managementIdentityClient.updateMigrationStatus(
+          tenantMappingRules.stream().map(this::createTenantMappingRule).toList());
+    } while (!tenantMappingRules.isEmpty());
+  }
+
+  private MigrationStatusUpdateRequest createTenantMappingRule(
+      final TenantMappingRule tenantMappingRule) {
+    try {
+      final var mappingKey =
+          mappingService.findOrCreateMapping(
+              tenantMappingRule.getName(),
+              tenantMappingRule.getClaimName(),
+              tenantMappingRule.getClaimValue());
+      for (final Tenant mappingTenant : tenantMappingRule.getAppliedTenants()) {
+        final var tenant = tenantService.fetch(mappingTenant.tenantId(), mappingTenant.name());
+        assignMappingToTenant(tenant, mappingKey);
+      }
+      return managementIdentityTransformer.toMigrationStatusUpdateRequest(tenantMappingRule, null);
+    } catch (final Exception e) {
+      LOGGER.error("Error creating tenant mapping rule", e);
+      return managementIdentityTransformer.toMigrationStatusUpdateRequest(tenantMappingRule, e);
+    }
+  }
+
+  private void assignMappingToTenant(final TenantEntity tenant, final Long mappingKey) {
+    try {
+      tenantService.assignMappingToTenant(tenant.key(), mappingKey);
+    } catch (final Exception e) {
+      if (!isConflictError(e)) {
+        throw e;
+      }
+    }
+  }
+}
--- a/migration/identity-migration/src/main/java/io/camunda/migration/identity/dto/MappingRule.java
+++ b/migration/identity-migration/src/main/java/io/camunda/migration/identity/dto/MappingRule.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migration.identity.dto;
+
+public class MappingRule {
+
+  protected MappingRuleType type;
+  protected String name;
+  protected String claimName;
+  protected String claimValue;
+  protected Operator operator;
+
+  public MappingRule() {}
+
+  public MappingRule(
+      final MappingRuleType type,
+      final String name,
+      final String claimName,
+      final String claimValue,
+      final Operator operator) {
+    this.type = type;
+    this.name = name;
+    this.claimName = claimName;
+    this.claimValue = claimValue;
+    this.operator = operator;
+  }
+
+  public MappingRuleType getType() {
+    return type;
+  }
+
+  public void setType(final MappingRuleType type) {
+    this.type = type;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(final String name) {
+    this.name = name;
+  }
+
+  public String getClaimName() {
+    return claimName;
+  }
+
+  public void setClaimName(final String claimName) {
+    this.claimName = claimName;
+  }
+
+  public String getClaimValue() {
+    return claimValue;
+  }
+
+  public void setClaimValue(final String claimValue) {
+    this.claimValue = claimValue;
+  }
+
+  public Operator getOperator() {
+    return operator;
+  }
+
+  public void setOperator(final Operator operator) {
+    this.operator = operator;
+  }
+
+  public enum MappingRuleType {
+    ROLE,
+    TENANT,
+    GROUP;
+  }
+
+  public enum Operator {
+    CONTAINS,
+    EQUALS;
+  }
+}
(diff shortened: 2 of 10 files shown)


migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
```diff
@@ -0,0 +1,84 @@
...
+      final TenantMappingRule tenantMappingRule) {
+    try {
+      final var mappingKey =
+          mappingService.findOrCreateMapping(
+              tenantMappingRule.getName(),
+              tenantMappingRule.getClaimName(),
+              tenantMappingRule.getClaimValue());
+      for (final Tenant mappingTenant : tenantMappingRule.getAppliedTenants()) {
+        final var tenant =
+            tenantService.fetchOrCreateTenant(mappingTenant.tenantId(), mappingTenant.name());
```
lenaschoenburg, line 66: :question: Why is this needed? The tenant migration always runs first so all tenant should exist already. If it's just to fetch the tenant key, I'd try and make that explicit, i.e. split `fetchOrCreateTenant` in two, `fetch` and `create`.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
```diff
@@ -0,0 +1,84 @@
...
+              ? null
+              : tenantMappingRules.get(tenantMappingRules.size() - 1);
+    } while (lastTenantMappingRule != null);
+  }
+
+  private MigrationStatusUpdateRequest createTenantMappingRule(
+      final TenantMappingRule tenantMappingRule) {
+    try {
+      final var mappingKey =
+          mappingService.findOrCreateMapping(
```
lenaschoenburg, line 60: :question: What's the point in searching for the mapping first? As long as we handle conflicts (see my other comment) we should be able to always create it without searching first.

migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ZeebeConnector.java
```diff
@@ -0,0 +1,65 @@
...
+    final var properties = identityMigrationProperties.getZeebe();
+    return newZeebeClient(properties);
+  }
+
+  public ZeebeClient newZeebeClient(final ZeebeProperties zeebeProperties) {
+    final var gatewayAddress = getGatewayAddress(zeebeProperties);
+    final ZeebeClientBuilder builder =
+        ZeebeClient.newClientBuilder()
+            .gatewayAddress(gatewayAddress)
+            .restAddress(URI.create("http://0.0.0.0:8081"))
```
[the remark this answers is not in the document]
maryarm: Ah, yes, it leaked from my local changes.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
maryarm: yes, but I wanted to reuse it for TenantMigrationHandler when we create the tenant and also with this we are more flexible in ordering the handlers.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
```diff
@@ -0,0 +1,84 @@
...
+              tenantMappingRule.getName(),
+              tenantMappingRule.getClaimName(),
+              tenantMappingRule.getClaimValue());
```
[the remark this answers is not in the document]
maryarm: because here we need the mapping key for next step (assign mapping to tenant) to be available, so here conflict is a error and we need to retry. ( I have to add that in this case it's very unlikely to have a conflict because a search query is executed before and it's just for very rare case that the mapping is created but is not  published to elastic yet.)

migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationHandler.java
```diff
@@ -0,0 +1,17 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migration.identity;
+
+public interface MigrationHandler {
+
+  void migrate();
+
+  default boolean isConflictError(final Exception e) {
+    return e.getMessage().contains("Failed with code 409: 'Conflict'");
```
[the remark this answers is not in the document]
maryarm: I'll check again but as I remember the exception is wrapped in another exception class that don't expose the status and checking the message is safer.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
maryarm: I need the mappingKey for next step but in case of conflict I can get the mappingKey.

dist/src/main/resources/application.properties
```diff
@@ -6,7 +6,7 @@ logging.register-shutdown-hook=true
 # This sever is enabled even for a standalone broker without any embedded gateway, in which case
 # there will simply be no endpoints available.
 server.address=0.0.0.0
-server.port=8080
+server.port=8081
```
lenaschoenburg, line 9: :x: this looks a change from local testing

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
lenaschoenburg: Since the refresh interval of ES indices is 1 second, I think there's a good chance that we will first create all tenants and then try to create them again here because we can't find them in ES yet. That would lead to CONFLICT errors which we don't handle here so the migration app would just fail.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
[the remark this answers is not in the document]
lenaschoenburg: But the retry would only occur because the migration app crashes, right? I think we should try to handle this explicitly and not rely on the crash.

migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ZeebeProperties.java
```diff
@@ -0,0 +1,53 @@
...
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.migration.identity.config;
+
+public class ZeebeProperties {
+
+  private String brokerContactPoint;
```
lenaschoenburg, line 12: :wrench: Since this is an entirely new configuration (under `camunda.migration.identity.zeebe`) I think we can ignore the backwards compatibility stuff around `brokerContactPoint` and only use `gatewayAddress` directly. So `ZeebeConnector#getGatewayAddress` can be removed too.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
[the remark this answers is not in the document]
maryarm: Actually not, retry will happens automatically on next loop during fetch mapping rules. The endpoints in identity designed to return failed or not already migrated errors. To be precise it was designed like this and I applied a fix in next PR to actually achieve it.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
maryarm: Conflict here is managed as an error status and we don't throw it to the caller, so in next loop for fetching the tenants we have them and can try. By the way do you think it's a good idea to have a pause (sleep) for a few seconds when we first experienced this error? so we have less conflict error.

migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ZeebeProperties.java
maryarm: I'll fix it in the next PR that will be ready by today.

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
maryarm: In my new PR, I have separated the `create` and `fetch` methods and I hope it prevent congesting zeebe with a lot of conflict requests.

migration/identity-migration/src/main/java/io/camunda/migration/identity/service/MappingService.java
```diff
@@ -0,0 +1,62 @@
...
+      final ZeebeClient zeebeClient,
+      final Authentication.Builder servicesAuthenticationBuilder) {
+    this.mappingServices =
+        mappingServices.withAuthentication(servicesAuthenticationBuilder.build());
+    this.zeebeClient = zeebeClient;
+  }
+
+  public Long findOrCreateUserWithUsername(final String username) {
+    return findOrCreateMapping(username + "_mapping", USERNAME_CLAIM, username);
+  }
```
lenaschoenburg, line 35: :question: unused?

migration/identity-migration/src/main/java/io/camunda/migration/identity/dto/MappingRule.java
```diff
@@ -0,0 +1,83 @@
...
+  public enum MappingRuleType {
+    ROLE,
+    TENANT,
+    GROUP;
+  }
+
+  public enum Operator {
+    CONTAINS,
+    EQUALS;
+  }
```
lenaschoenburg, line 82: :question: I guess this is also not yet supported, right? Perhaps we should throw an exception when encountering `CONTAINS` until it is supported?

migration/identity-migration/src/test/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandlerTest.java
```diff
@@ -0,0 +1,170 @@
...
+  @Mock private TenantService tenantService;
+  @Mock private MappingService mappingService;
+
+  private final ManagementIdentityTransformer managementIdentityTransformer =
+      new ManagementIdentityTransformer();
+  private TenantMappingRuleMigrationHandler migrationHandler;
+
+  @BeforeEach
+  void setUp() throws Exception {
+    MockitoAnnotations.openMocks(this).close();
```
lenaschoenburg, line 53: :question: What is this doing?

migration/identity-migration/src/test/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandlerTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+    // when
+    migrationHandler.migrate();
+
+    // then
+    verify(managementIdentityClient, times(2)).fetchTenantMappingRules(anyInt());
+    verify(tenantService, never()).fetch(any(), any());
+    verify(mappingService, times(2)).findOrCreateMapping(any(), any(), any());
+    verify(managementIdentityClient, times(2))
+        .updateMigrationStatus(migrationStatusCaptor.capture());
```
lenaschoenburg, line 94: :wrench: No argument captor needed, you can just verify the arguments directly here

lenaschoenburg (approved):

migration/identity-migration/src/main/java/io/camunda/migration/identity/dto/MappingRule.java
maryarm: In new identity both EQUALS and CONTAINS are same

migration/identity-migration/src/main/java/io/camunda/migration/identity/service/MappingService.java
maryarm: It leaked from next PR. ;)

migration/identity-migration/src/test/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandlerTest.java
maryarm: It's for creating @Mock instances. If you know a better approach, let me know and I apply in next tests.

lenaschoenburg: Ah okay, I think you could also just annotate the test class with ` @ExtendWith(MockitoExtension. class)`

PR #25804, opened 2024-12-10
refactor: use full index name in logs

ChrisKujawa:
## Description

The error logs, did not reflect the actual name of the indices. This was quite confusing when reading and debugging.


EuroLew (approved):
LGTM

Issue #25795, opened 2024-12-10
gRPC gateway need to supports BASIC AUTH


nicpuppa:
See [discussion](https://camunda.slack.com/archives/C06UYJMMETZ/p1733503270899139)

PR #25815, opened 2024-12-10
Use username if userKey is not available to retrieve user data

### LINKED ISSUE #25715: Add username as claim in the gRPC gateway

nicpuppa:
## Description

Not all the command might contains the user key.
In the gRPC gateway we are not able to retrieve the user key information, we use the user's username to fetch the persisted record.

In the gRPC gateway we already decode the user claims and put them into the token that the gateway send to the broker,
within this user claim the is also the username.

I created a separate issue to support basic auth also in the gRPC gateway https://github.com/camunda/camunda/issues/25795

## Related issues

related to https://github.com/camunda/camunda/issues/25719


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -216,16 +214,15 @@ public List<String> getAuthorizedTenantIds(final TypedRecord<?> command) {
...
         .map(
-            key ->
-                userState
-                    .getUser(key)
-                    .map(PersistedUser::getTenantIdsList)
-                    .orElse(authorizedTenantIds))
+            user -> {
+              final List<String> tenantIds = user.getTenantIdsList();
+              // Use default tenant IDs if user's tenant list is empty
+              return tenantIds.isEmpty() ? authorizedTenantIds : tenantIds;
```
nicpuppa, line 224: 💭 I'm not sure about this. This bug came up while testing the resolve incident gRPC endpoint. The incident record by default has the default tenant, I'm assuming in this scenario if the user doesn't have any tenant it is ok to return the default one

nicpuppa:
Let's wait for some feedback about using ES in the gateway

koevskinikola:
@nicpuppa should we convert this to a draft PR until the spike about ES in the gateway is done, and we know more on how we proceed?

PR #25828, opened 2024-12-10
Fix flaky metric tests with gauge

EuroLew:
## Description

Fix the flaky `Finishedimporting.shouldMarkImporterCompletedViaMetricsAsWell` test

### DIFF
--- a/operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
+++ b/operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
@@ -18,7 +18,6 @@ import java.time.temporal.ChronoUnit;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.locks.ReentrantLock;
 import org.slf4j.Logger;
@@ -43,8 +42,7 @@ public class ImportPositionHolder {
       new HashMap<>();
   private final Map<String, ImportPositionEntity> inflightImportPositions = new HashMap<>();
   private final Map<String, ImportPositionEntity> inflightPostImportPositions = new HashMap<>();
-  private final ConcurrentHashMap<String, Boolean> mostRecentProcessedImportPositions =
-      new ConcurrentHashMap<>();
+  private final Map<String, Boolean> mostRecentProcessedImportPositions = new HashMap<>();
 
   private ScheduledFuture<?> scheduledImportPositionUpdateTask;
   private final ReentrantLock inflightImportPositionLock = new ReentrantLock();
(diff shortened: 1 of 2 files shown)


operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
```diff
@@ -127,18 +128,9 @@ public void recordLatestLoadedPosition(final ImportPositionEntity lastProcessedP
           // update only import fields (not post import)
           final String key = getKey(aliasName, partition);
           ImportPositionEntity importPosition = inflightImportPositions.get(key);
+          registerRecordReaderCompletedGauge(partition, aliasName);
```
ChrisKujawa, line 131: ❓ Why we register this multiple times?

operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
```diff
@@ -43,7 +43,8 @@ public class ImportPositionHolder {
       new HashMap<>();
   private final Map<String, ImportPositionEntity> inflightImportPositions = new HashMap<>();
   private final Map<String, ImportPositionEntity> inflightPostImportPositions = new HashMap<>();
-  private final Map<String, Gauge> importPositionCompletedGauges = new HashMap<>();
+  private final ConcurrentHashMap<String, Boolean> mostRecentProcessedImportPositions =
```
ChrisKujawa, line 46: ❓ Wondering about this, as this would mean that the other maps are also not safe?!

ChrisKujawa, line 46: ❓ Wondering about this, as this would mean that the other maps are also not safe?!

operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
ChrisKujawa: What i mean is that we register it now everytime we come into the method 🤔 before it was enough to just do it once the first time.

operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
ChrisKujawa: Was this now the issue or the other fix? 😃

ChrisKujawa (approved):
Thanks Euro.

Would be good if we can clarif what the actual fix was see my comments and maybe we can remove the concurrent hash map thing? 

If we did this i think we can merge it

operate/importer/src/main/java/io/camunda/operate/zeebeimport/ImportPositionHolder.java
EuroLew: This is an artifact from a prior solution I tried doesn't need to be this way I'll make it a normal map

ChrisKujawa:
@EuroLew I think we have the same issue on tasklist

PR #25832, opened 2024-12-10
feat: OutlierAnalysisServiceTest branch

KarpovIvan:
## Description

task: [24657](https://github.com/camunda/camunda/issues/24657)

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/backend/src/test/java/io/camunda/optimize/service/OutlierAnalysisServiceTest.java
+++ b/optimize/backend/src/test/java/io/camunda/optimize/service/OutlierAnalysisServiceTest.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.optimize.service;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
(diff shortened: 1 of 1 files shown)


optimize/backend/src/test/java/io/camunda/optimize/service/OutlierAnalysisServiceTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.junit.jupiter.api.Test;
+
+public class OutlierAnalysisServiceTest {
+  private final DataSourceDefinitionAuthorizationService definitionAuthorizationService =
+      mock(DataSourceDefinitionAuthorizationService.class);
+  private final DurationOutliersReader outliersReader = mock(DurationOutliersReader.class);
+  private final OutlierAnalysisService outlierAnalysisService =
+      new OutlierAnalysisService(definitionAuthorizationService, outliersReader);
+
+  @Test
```
RomanJRW, line 36: The selection of tests here seems somewhat random. As an alternative, I would suggest:

- One parameterized test checking for when authorization passes. This should be parameterized with the different public methods of this class. The assertions should check that the parameter is sent to the corresponding `outliersReader` method. This means that the forwarded request must also be parameterized
- Another parameterized test checking for when authorization is missing. In this case, the method under test can be a parameter, but the assertion is always the same regardless

RomanJRW (changes requested):

optimize/backend/src/test/java/io/camunda/optimize/service/OutlierAnalysisServiceTest.java
KarpovIvan: @RomanJRW parameterized test will be bad solution here, because we have four different methods  with for 3 different models and with generic types. will be more clear create private method with deduplicating solution

KarpovIvan: fixed

RomanJRW (approved):
This provides a base layer of test coverage for the class in question. We will proceed with merging and allow priority to shift to other topics

PR #25840, opened 2024-12-10
New E2E Operate PI Migration test

### LINKED ISSUE #25807: ProcessInstanceMigration Operation test must use Operate REST API instead of direct call to Zeebe
## Description

When migrating the Operate IT and E2E we realized that several tests related to Process Instance migration were failing. Mostly related to the Operate Operations coupled to that.

First, we expected to be a result of not yet having https://github.com/camunda/camunda/issues/24084 implemented. But later we realized there might be an issue with (race condition?) when writing Operate BatchOperations and Exporter updating such.

Related thread
https://camunda.slack.com/archives/C06F0GLJNFM/p1733826665980999?thread_ts=1733471637.652659&cid=C06F0GLJNFM

ChrisKujawa:
## Description

This PR introduces two new acceptance tests for operates batch operation engine:

* migrate process instances 
* cancel process instances

## Related issues

closes https://github.com/camunda/camunda/issues/25807

### DIFF
--- a/qa/acceptance-tests/src/test/java/io/camunda/it/util/TestHelper.java
+++ b/qa/acceptance-tests/src/test/java/io/camunda/it/util/TestHelper.java
@@ -391,9 +391,8 @@ public final class TestHelper {
                 assertThat(batch.getOperationsTotalCount()).isEqualTo(expectedItems);
               } catch (final ProblemException e) {
                 if (e.code() == 404) {
-                  // If the batch operation is not found, it means it has been completed or failed.
-                  // We can ignore this exception and let the test fail later when checking the
-                  // status.
+                  // In the case where it could not be found in the db, we can ignore it.
+                  // When we start this await, it will always not be there.
                   return;
                 }
 
--- a/qa/acceptance-tests/src/test/java/io/camunda/it/util/TestHelper.java
+++ b/qa/acceptance-tests/src/test/java/io/camunda/it/util/TestHelper.java
@@ -24,6 +24,7 @@ import io.camunda.client.api.search.filter.ElementInstanceFilter;
 import io.camunda.client.api.search.filter.IncidentFilter;
 import io.camunda.client.api.search.filter.ProcessDefinitionFilter;
 import io.camunda.client.api.search.filter.ProcessInstanceFilter;
+import io.camunda.client.api.search.response.ProcessInstance;
 import io.camunda.client.api.search.response.SearchResponse;
 import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
 import java.time.Duration;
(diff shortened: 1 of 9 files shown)


qa/integration-tests/src/test/java/io/camunda/it/operate/OperateProcessInstanceMigrationIT.java
```diff
@@ -0,0 +1,243 @@
...
+    final var authorizationsUtil =
+        new AuthorizationsUtil(
+            testStandaloneCamunda, testStandaloneCamunda.getElasticSearchHostAddress());
+    final var defaultClient = authorizationsUtil.getDefaultClient();
+    // create super user that can read all process definitions
+    final var superZeebeClient =
+        authorizationsUtil.createUserAndClient(
+            SUPER_USER, "password", new Permissions(BATCH, UPDATE, List.of("*")));
+    superOperateClient = testStandaloneCamunda.newOperateClient(SUPER_USER, "password");
+  }
```
ChrisKujawa, line 69: @houssain-barouni I saw that you introduce this recently for another test, I thought I need this as well as the endpoint I want to query is annotated 

```

  @PreAuthorize("hasPermission('write')")
```



but not sure anymore (I don't feel it helps :D ) Can you quickly confirm I'm on the right track? 🙏🏼

qa/acceptance-tests/src/test/java/io/camunda/it/util/TestHelper.java
```diff
@@ -379,14 +381,24 @@ public static void waitForBatchOperationWithCorrectTotalCount(
...
-              assertThat(batch).isNotNull();
-              assertThat(batch.getOperationsTotalCount()).isEqualTo(expectedItems);
+              try {
+                final var batch =
+                    camundaClient.newBatchOperationGetRequest(batchOperationKey).send().join();
+                assertThat(batch).isNotNull();
+                assertThat(batch.getOperationsTotalCount()).isEqualTo(expectedItems);
+              } catch (final ProblemException e) {
+                if (e.code() == 404) {
+                  // If the batch operation is not found, it means it has been completed or failed.
```
stoerti: ❓ Doesn't that just mean, that the operation is not yet available in the ES index?

🔧  Might this me more readable?
```
.ignoreExceptionsMatching(
    e ->
        e instanceof ProblemException
        && ((ProblemException) e).code() == 404) // Ignore 404 errors

```

stoerti (commented):
LGTM 👍 

We might add also one test where we execute a batch operation not supported by the Zeebe engine (e.g. `ADD_VARIABLE`). But before we need https://github.com/camunda/camunda/pull/36068 to fix enum issues there.

mustafadagher (approved):

qa/acceptance-tests/src/test/java/io/camunda/it/util/TestHelper.java
p-wunderlich: Fixed the comment. But the ignoreExceptionsMatching pattern doesn't work, I don't know why ... hi simply does NOT ignore the exception then ... So I keep the old impl there.

PR #25842, opened 2024-12-10
feat: leave assigned roles, groups and tenants before deleting entity

lenaschoenburg:
This adjust the deletion of users and mapping such that we first emit `ENTITY_REMOVED` events for all assigned roles, groups and tenants. This resolves two problems:

1. Properly cleaning up internal state. Before this, deleting a user left remnants in some column families
2. Exporting the implicit cleanup such that the CamundaExporter does not have to re-implement this logic

:thought_balloon: Authorizations behave similar to memberships here, in that they are implicitly deleted when the owning entity is deleted. We could also emit `Authorization/DELETED` events for all directly owned authorizations before deleting a user or mapping.

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/MappingDeleteProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/MappingDeleteProcessor.java
@@ -16,12 +16,18 @@ import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedRejectionWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedResponseWriter;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
+import io.camunda.zeebe.engine.state.authorization.PersistedMapping;
 import io.camunda.zeebe.engine.state.distribution.DistributionQueue;
 import io.camunda.zeebe.engine.state.immutable.MappingState;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.MappingRecord;
+import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
+import io.camunda.zeebe.protocol.impl.record.value.group.GroupRecord;
 import io.camunda.zeebe.protocol.record.RejectionType;
+import io.camunda.zeebe.protocol.record.intent.GroupIntent;
 import io.camunda.zeebe.protocol.record.intent.MappingIntent;
+import io.camunda.zeebe.protocol.record.intent.RoleIntent;
 import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
+import io.camunda.zeebe.protocol.record.value.EntityType;
 import io.camunda.zeebe.protocol.record.value.PermissionType;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
 import io.camunda.zeebe.stream.api.state.KeyGenerator;
@@ -77,7 +83,7 @@ public class MappingDeleteProcessor implements DistributedTypedRecordProcessor<M
       return;
     }
 
-    stateWriter.appendFollowUpEvent(mappingKey, MappingIntent.DELETED, record);
+    deleteMapping(persistedMapping.get());
     responseWriter.writeEventOnCommand(mappingKey, MappingIntent.DELETED, record, command);
 
     final long key = keyGenerator.nextKey();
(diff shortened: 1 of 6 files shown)


remcowesterhoud:
@nicpuppa Do you have capacity to review this?

nicpuppa:
> @nicpuppa Do you have capacity to review this?

Yes

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserDeleteProcessor.java
```diff
@@ -100,13 +112,37 @@ final var record = command.getValue();
...
               final var message = USER_DOES_NOT_EXIST_ERROR_MESSAGE.formatted(record.getUserKey());
               rejectionWriter.appendRejection(command, RejectionType.NOT_FOUND, message);
             });
 
     distributionBehavior.acknowledgeCommand(command);
   }
+
+  private void deleteUser(final PersistedUser user) {
+    final var userKey = user.getUserKey();
+    for (final var roleKey : user.getRoleKeysList()) {
```
nicpuppa, line 136: 🔧 You can rewrite this using the `forEach()` method

```
    user.getRoleKeysList()
        .forEach(
            roleKey ->
                    roleKey,
                    RoleIntent.ENTITY_REMOVED,
                    new RoleRecord()
                        .setRoleKey(roleKey)
                        .setEntityKey(userKey)
                        .setEntityType(EntityType.USER)));
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserDeleteProcessor.java
```diff
@@ -100,13 +112,37 @@ final var record = command.getValue();
...
+      stateWriter.appendFollowUpEvent(
+          roleKey,
+          RoleIntent.ENTITY_REMOVED,
+          new RoleRecord()
+              .setRoleKey(roleKey)
+              .setEntityKey(userKey)
+              .setEntityType(EntityType.USER));
+    }
+    for (final var groupKey : user.getGroupKeysList()) {
```
nicpuppa, line 145: 🔧 You can rewrite this using the `forEach()` method

```
    user.getGroupKeysList()
        .forEach(
            groupKey ->
                    groupKey,
                    GroupIntent.ENTITY_REMOVED,
                    new GroupRecord()
                        .setGroupKey(groupKey)
                        .setEntityKey(userKey)
                        .setEntityType(EntityType.USER)));
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/MappingDeleteProcessor.java
```diff
@@ -104,4 +108,28 @@ final var record = command.getValue();
...
+    }
+    for (final var groupKey : mapping.getGroupKeysList()) {
+      stateWriter.appendFollowUpEvent(
+          groupKey,
+          GroupIntent.ENTITY_REMOVED,
+          new GroupRecord()
+              .setGroupKey(groupKey)
+              .setEntityKey(mappingKey)
+              .setEntityType(EntityType.MAPPING));
+    }
```
nicpuppa, line 142: 🔧 For both the for-loop, the previous comments apply also here

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/MappingDeletedApplier.java
```diff
@@ -45,31 +35,8 @@ public void applyState(final long key, final MappingRecord value) {
               "Expected to delete mapping with key '%s', but a mapping with this key does not exist.",
               value.getMappingKey()));
     }
-    final var persistedMapping = mapping.get();
-    removeMappingFromRoleState(persistedMapping);
-    removeMappingFromGroupState(persistedMapping);
-    removeMappingFromTenantState(persistedMapping);
     // remove mapping from authorization state
     authorizationState.deleteOwnerTypeByKey(mappingKey);
```
nicpuppa, line 39: ❓ Can you double check if here we missed to call `deleteAuthorizationsByOwnerKeyPrefix()` ?

nicpuppa (approved):
The implementation looks good, I left some small suggestions. Can you check my question @lenaschoenburg ?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/MappingDeletedApplier.java
lenaschoenburg: Good catch, I think so :+1:

lenaschoenburg: Fixed in 5949c0d

PR #25845, opened 2024-12-10
Rename task listener event types

### LINKED ISSUE #25544: Use gerunds as User Task Listener Event Types in Zeebe
Zeebe will offer backward compatibility for the previous naming, to support models created by previously released versions of Desktop Modeler and Web Modeler (and in case Modeling doesn't make it until the code freeze). We'll need to:
- Deprecate previously used event names, and add new names rather than replacing them
- Requires a minor change to the model validator (allow both the deprecated and new name)
- Requires a minor change to the model transformer (use new name if deprecated name is used)

korthout:
## Description

Renames the task listener event types providing backward compatibility:
- from `create`, `assignment`, `update`, `complete`, `cancel`
- to `creating`, `assigning`, `updating`, `completing`, `canceling`

Also renames the job listener event types to align their naming although these do not require backward compatibility (released initially in `8.7.0-alpha1`).

## Related issues

closes #25544

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
@@ -93,7 +93,7 @@ public class TaskListenerTest {
     // then
     assertTaskListenerJobsCompletionSequence(
         processInstanceKey,
-        JobListenerEventType.COMPLETE,
+        JobListenerEventType.COMPLETING,
         LISTENER_TYPE,
         LISTENER_TYPE + "_2",
         LISTENER_TYPE + "_3");
@@ -141,7 +141,7 @@ public class TaskListenerTest {
     // then
     assertTaskListenerJobsCompletionSequence(
         processInstanceKey,
-        JobListenerEventType.ASSIGNMENT,
+        JobListenerEventType.ASSIGNING,
         LISTENER_TYPE,
         LISTENER_TYPE + "_2",
         LISTENER_TYPE + "_3");
@@ -310,7 +310,7 @@ public class TaskListenerTest {
 
     // then
     assertTaskListenerJobsCompletionSequence(
-        processInstanceKey, JobListenerEventType.ASSIGNMENT, LISTENER_TYPE, LISTENER_TYPE + "_2");
+        processInstanceKey, JobListenerEventType.ASSIGNING, LISTENER_TYPE, LISTENER_TYPE + "_2");
 
     // ensure that `COMPLETE_TASK_LISTENER` commands were triggered between
     // `CLAIMING` and `ASSIGNED` events
@@ -355,7 +355,7 @@ public class TaskListenerTest {
     // then: verify the task listener completion sequence for the assignment event
     assertTaskListenerJobsCompletionSequence(
         processInstanceKey,
-        JobListenerEventType.ASSIGNMENT,
+        JobListenerEventType.ASSIGNING,
         LISTENER_TYPE,
         LISTENER_TYPE + "_2",
         LISTENER_TYPE + "_3");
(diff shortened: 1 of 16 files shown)


ana-vinogradova-camunda (approved):
🚀! LGTM 👍 

Great test to verify the correct transformation of deprecated task listener event types 👍

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
```diff
@@ -36,15 +36,16 @@ public enum JobListenerEventType {
...
   /**
-   * Represents the `assignment` event for a task listener. This event type is used to indicate that
-   * the listener should be triggered when a user task is assigned, claimed (assigned to the current
-   * user), or unassigned.
+   * Represents the `assigning` event for a task listener. This event type is used to indicate that
+   * the listener should be triggered when a user task is assigning. It allows to execute custom
+   * logic before the task is assigned, to correct user task data, and to deny the assignment.
    */
-  ASSIGNMENT
+  ASSIGNING,
```
ce-dmelnych, line 50: ❓ What do you think about also renaming the [UserTaskIntent-s](https://github.com/camunda/camunda/blob/1df25b546f55b29f3f007b82390a180103e5c8de/zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/intent/UserTaskIntent.java#L73-L87) to align with the updated task listener event types?

For example:
```
ASSIGNMENT_DENIED -> ASSIGNING_DENIED  
COMPLETION_DENIED -> COMPLETING_DENIED
```

This would make the naming consistent in `UserTaskIntent` also and improve the readability of the records flow:

```java
C ASSIGN  
E ASSIGNING  
// If the assigning listener denies the operation  
C DENY_TASK_LISTENER  
// Then we would have
E ASSIGNING_DENIED
```

ce-dmelnych: And adjust event applier names:
```scala
UserTaskAssignmentDeniedApplier -> UserTaskAssigningDeniedApplier
UserTaskCompletionDeniedApplier -> UserTaskCompletingDeniedApplier
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
ce-dmelnych, line 1: ❓ I noticed that this test class still uses the deprecated `complete` and `assignment` listener type. Are we planning to handle the migration to the new listener types later?

ce-dmelnych (approved):
🎉 LGTM 👍

Thank you, Nico, for ensuring a backward-compatible approach. This should help avoid conflicts in other PRs after merging this one.

I’m tempted to start using the new listener event types in tests of non-merged PRs already! 😄

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
ce-dmelnych: > ASSIGNING and COMPLETING can be considered as not having been denied because those events have actually happened and were emitted.

Thanks @korthout for highlighting this point — in that case, `ASSIGNMENT_DENIED` and `COMPLETION_DENIED` would suit better.

At the same time, later we’ll might have `UPDATE_DENIED` (according to [**Options 4**](https://camunda.slack.com/archives/C0622NWC9LK/p1728996247956779)), which makes it clear that finding the perfect naming convention isn’t straightforward! :see_no_evil:😸

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
korthout: Great point! I've added two more commits 00d27ef 0b1bf79. Any deprecation warning for usage of these types should be considered for replacement.

zeebe/protocol/src/main/java/io/camunda/zeebe/protocol/record/value/JobListenerEventType.java
ana-vinogradova-camunda: 💭 `ASSIGNMENT_DENIED` and `COMPLETION_DENIED` would suit better here I think 👍

PR #25852, opened 2024-12-10
feat: add unit test that checks Prio$NBackup interfaces are all injected

### LINKED ISSUE #25848: Test all interfaces are injected correctly via @Bean
Add a unit test to verify that all `Prio$NBackup` interfaces are injected correctly with spring.
ArchUnit can be used to scan the classpath

entangled90:
## Description
Small refactor of the `HistoryBackupComponent` to better test the priorities bean annotation.
We check that:
- each index does not implement two Priority interfaces
- each index that implements BackupPriority is injected correctly
- 
## Checklist
- [x] Decide how to implement the injection for indices under `usermanagement`
- [x] Wait until #25687 is merged as the test will fail
## Related issues

closes #25848


npepinpe (commented):
:x: I don't think this is correct. I think we are planning to have a global prefix for 8.7, but you should double check with @ChrisKujawa on this.

entangled90:
I just noticed that indexPrefix is present `camunda-search-client-connet` in configuration `io.camunda.search.connect.configuration.ConnectConfiguration.java`, would be that the correct one? @ChrisKujawa

ChrisKujawa:
> ❌ I don't think this is correct. I think we are planning to have a global prefix for 8.7, but you should double check with @ChrisKujawa on this.

This is half correct I would say.

**What is planned for 8.7**:

1. We have on all components a prefix configurable
    1.1. Webapps, Exporter, search
3. The component names are part of the index names (so it will not override when prefix is set)
4. The prefix is per default empty
5. We have in the helm charts one value that allows to set the prefix everywhere

**What is planned for 8.8**

1. We plan to streamline the configurations
2. This means we will have one global prefix config, that will be used in all components, usages


@entangled90 @npepinpe does this help?!

entangled90:
Does that mean that for 8.7 each webapp get its own index? 
That means that we need to backup for example twice (operate & tasklist).

This is probably working correctly when all apps are running in the same instance, but it's probably broken when operate & tasklist are running standalone, as the other app indices are not injected.

Did I understand correctly?

ChrisKujawa:
> Does that mean that for 8.7 each webapp get its own index? That means that we need to backup for example twice (operate & tasklist).
> 
> This is probably working correctly when all apps are running in the same instance, but it's probably broken when operate & tasklist are running standalone, as the other app indices are not injected.

Sorry but I'm not sure whether I can follow this question 😅 

>operate & tasklist are running standalone, as the other app indices are not injected.

This is something we should still support, as for example in SaaS we deploy the apps separate.

entangled90:
>Does that mean that for 8.7 each webapp get its own index?
That means that we need to backup for example twice (operate & tasklist).

This part of my comment  is wrong as they are using separate indices.

> This is something we should still support, as for example in SaaS we deploy the apps separate.

This can be done if the prefix is the same :+1: 

We probably need to add this to the docs as well

entangled90:
So currently when apps are running standalone the indices from the other app are not injected (I think):
- `TasklistIndexTemplateDescriptorsConfigurator` uses `TasklistProperties`
- `IndexTemplateDescriptorsConfigurator` uses `OperateProperties` and `DatabaseInfo`

So I assume that when running standalone those classes are not available (to be verified).

If that is true, how do we want to solve this?
I propose the following:
- Don't use Spring `@Brean` as it's quite hard to understand what's happening
- Create 1 configuration class where we allocate all the indices from those configs, using the config from the other app when one of `{TasklistList | Operate}Properties` is missing? 

Downside is that every time a new index is added, it must be manually added, but we can create a test case for it (similar to what's done in this PR).

zeebe/qa/integration-tests/pom.xml
```diff
@@ -28,6 +28,11 @@
       <artifactId>zeebe-broker</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.camunda</groupId>
```
entangled90, line 32: Don't know why it has to be added, but then if you add it it report it as an error

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/usermanagement/index/RoleIndex.java
```diff
@@ -7,11 +7,12 @@
  */
 package io.camunda.webapps.schema.descriptors.usermanagement.index;
 
+import io.camunda.webapps.schema.descriptors.backup.Prio5Backup;
 import io.camunda.webapps.schema.descriptors.usermanagement.UserManagementIndexDescriptor;
 import io.camunda.webapps.schema.entities.usermanagement.EntityJoinRelation.EntityJoinRelationFactory;
 import io.camunda.webapps.schema.entities.usermanagement.EntityJoinRelation.IdentityJoinRelationshipType;
 
-public class RoleIndex extends UserManagementIndexDescriptor {
+public class RoleIndex extends UserManagementIndexDescriptor implements Prio5Backup {
```
entangled90, line 15: Removed by mistake today in main in [commit](https://github.com/camunda/camunda/commit/23404a94f23f85263cfe5b2a7df18f7d1ec521a6#diff-b1529f330014dfe26200f492ceae14fc9c3999d18607958db41fa2d2e6356ca1R13)

ChrisKujawa:
> We probably need to add this to the docs as well

ChrisKujawa:
> * Don't use Spring @Brean as it's quite hard to understand what's happening
> * Create 1 configuration class where we allocate all the indices from those configs, using the config from the other app when one of {TasklistList | Operate}Properties is missing?

I guess I lack a bit of context, but I always favor of not using spring if possible


> Downside is that every time a new index is added, it must be manually added, but we can create a test case for it (similar to what's done in this PR).

My expectation is there shouldn't be added much more indices, at least not right now.

entangled90:
> My expectation is there shouldn't be added much more indices, at least not right now.
Sure, anyway the test case in this PR will fail when an index that extends BackupPriority is created and it's not added to this list :+1:

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
npepinpe, line 1: 👍 Looks good. I would remove the comments highlighting what is for which app, as I'm sure this will diverge and things will get more overlap, but not a blocker.

dist/src/main/java/io/camunda/application/commons/backup/ConfigValidation.java
npepinpe, line 1: 🙃 imo seems overengineered just to test values match, but also not a big deal. I think that without making it generic it's probably just one liners? 🤷‍♂️ But also not a big deal.

dist/src/test/java/io/camunda/application/commons/backup/BackupPrioritiesTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+  static {
+    ALL_IMPLEMENTATIONS =
+        new ClassFileImporter()
+                .importPackages(
+                    // TODO ADD optimize
+                    "io.camunda.webapps.schema") // , "io.camunda.optimize.service.db.schema.index")
+                .that(implement(BackupPriority.class))
+                .stream()
+                .filter(clz -> !clz.getName().contains("Abstract"))
```
npepinpe, line 42: ❓️ Is there no helper in ArchUnit to filter out abstract classes, and not rely on naming conventions? Seems quite brittle.

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/backup/BackupPriorities.java
npepinpe, line 1: ❓️Do we ever access individual lists separately, or only in their concatenates form? If the latter, let's just have a single list but combine them on creation.

zeebe/qa/integration-tests/pom.xml
npepinpe: What is the error reported?

webapps-schema/src/main/java/io/camunda/webapps/schema/descriptors/backup/BackupPriorities.java
entangled90: We use the outer level of the list to split the indices in diferent parts of the snapshot, that's why they are kept separated. We could introduce a class to make it more clear.

I'm planning to simplify this part and extract it from the `BackupServiceImpl` to improve testability, so we can address it in a followup PR

zeebe/qa/integration-tests/pom.xml
entangled90: Yes, that's why I had to add it originally..

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
entangled90: I added it because it's easier to see if you forgot a class in a package, you can just over the list as it's sorted alphabetically.

npepinpe (approved):

dist/src/main/java/io/camunda/application/commons/backup/ConfigValidation.java
entangled90: could be, I extracted this code from the original specialized implementation as it's much easier to test.
I originally expected to be able to use it also for another class, but in that case it was slightly different.

dist/src/test/java/io/camunda/application/commons/backup/BackupPrioritiesTest.java
entangled90: So for example all optimize indices follow this pattern (different from the other webapps):
```java
public abstract class BusinessKeyIndex<TBuilder> extends DefaultIndexMappingCreator<TBuilder>
    implements Prio6Backup
``
This is the class we actually want to test against, but unfortunately it's abstract. 
The concrete implementations are  `BusinessKeyIndexES` and  `BusinessKeyIndexOS`.

zeebe/qa/integration-tests/pom.xml
entangled90: I retried and seems to run fine now, so I removed it

PR #25854, opened 2024-12-10
feat: java client supports assigned groups to tenants

Oleksiivanov:
## Description

- added AssignGroupToTenantCommand to zeebe client
- added UnassignGroupFromTenantCommand to zeebe client
- fixed tenant applier and processor for support GROUP entity
- added tenantKeys list to groupPersisted
- added tests

## Checklist

- [ ] for CI changes:

## Related issues

closes #25784

### DIFF
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AssignGroupToTenantCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AssignGroupToTenantCommandImpl.java
@@ -28,19 +28,15 @@ public final class AssignGroupToTenantCommandImpl implements AssignGroupToTenant
 
   private final HttpClient httpClient;
   private final long tenantKey;
-  private long groupKey;
+  private final long groupKey;
   private final RequestConfig.Builder httpRequestConfig;
 
-  public AssignGroupToTenantCommandImpl(final HttpClient httpClient, final long tenantKey) {
+  public AssignGroupToTenantCommandImpl(
+      final HttpClient httpClient, final long tenantKey, final long groupKey) {
     this.httpClient = httpClient;
     this.tenantKey = tenantKey;
-    httpRequestConfig = httpClient.newRequestConfig();
-  }
-
-  @Override
-  public AssignGroupToTenantCommandStep1 groupKey(final long groupKey) {
     this.groupKey = groupKey;
-    return this;
+    httpRequestConfig = httpClient.newRequestConfig();
   }
 
   @Override
--- a/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AssignGroupToTenantCommandImpl.java
+++ b/clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AssignGroupToTenantCommandImpl.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright © 2017 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.zeebe.client.impl.command;
+
+import io.camunda.zeebe.client.api.ZeebeFuture;
+import io.camunda.zeebe.client.api.command.AssignGroupToTenantCommandStep1;
+import io.camunda.zeebe.client.api.response.AssignGroupToTenantResponse;
+import io.camunda.zeebe.client.impl.http.HttpClient;
+import io.camunda.zeebe.client.impl.http.HttpZeebeFuture;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import org.apache.hc.client5.http.config.RequestConfig;
+
+public final class AssignGroupToTenantCommandImpl implements AssignGroupToTenantCommandStep1 {
+
+  private final HttpClient httpClient;
+  private final long tenantKey;
+  private long groupKey;
+  private final RequestConfig.Builder httpRequestConfig;
+
+  public AssignGroupToTenantCommandImpl(final HttpClient httpClient, final long tenantKey) {
+    this.httpClient = httpClient;
+    this.tenantKey = tenantKey;
+    httpRequestConfig = httpClient.newRequestConfig();
+  }
+
+  @Override
+  public AssignGroupToTenantCommandStep1 groupKey(final long groupKey) {
+    this.groupKey = groupKey;
+    return this;
+  }
+
+  @Override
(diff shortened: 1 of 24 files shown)


koevskinikola:
@Oleksiivanov I see this PR is still prefixed with `[WIP]`. Should I start reviewing it, or should we convert it to a `draft` until it's ready?

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/AssignGroupToTenantCommandImpl.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Override
+  public AssignGroupToTenantCommandStep1 requestTimeout(final Duration timeout) {
+    httpRequestConfig.setResponseTimeout(timeout.toMillis(), TimeUnit.MILLISECONDS);
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<AssignGroupToTenantResponse> send() {
+    final HttpZeebeFuture<AssignGroupToTenantResponse> result = new HttpZeebeFuture<>();
```
koevskinikola, line 54: 🔧 We should add an assertion that the `groupKey` is set and required here.

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/UnassignGroupFromTenantCommandImpl.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public FinalCommandStep<UnassignGroupFromTenantResponse> requestTimeout(
+      final Duration requestTimeout) {
+    httpRequestConfig.setResponseTimeout(requestTimeout.toMillis(), TimeUnit.MILLISECONDS);
+    return this;
+  }
+
+  @Override
+  public ZeebeFuture<UnassignGroupFromTenantResponse> send() {
+    final HttpZeebeFuture<UnassignGroupFromTenantResponse> result = new HttpZeebeFuture<>();
```
koevskinikola, line 56: 🔧 We should add an assertion that the `groupKey` is set and required here.

clients/java/src/test/java/io/camunda/zeebe/client/tenant/UnassignGroupFromTenantTest.java
koevskinikola, line 1: ❌ We should add an `AssignGroupToTenantTest` class as well.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/tenant/AddEntityTenantTest.java
```diff
@@ -27,19 +30,17 @@ public class AddEntityTenantTest {
...
-            .getValue()
-            .getUserKey();
+  public void shouldAddEntitiesToTenant() {
+    // We can't use JUnit 5 parameterized tests due to JUnit 4's @Rule incompatibility, and JUnit
+    // 4's parameterized tests are not suitable for this test class    shouldAddEntityToTenant(USER,
+    //  createUser());
+    shouldAddEntityToTenant(GROUP, createGroup());
+    shouldAddEntityToTenant(MAPPING, createMapping());
+    shouldAddEntityToTenant(USER, createUser());
+  }
```
koevskinikola, line 40: 🔧  We can also split this test into a separate JUnit 4 parameterized test class or split these 3 cases into 3 separate test methods here. This way we don't have any test isolation.

koevskinikola (changes requested):
@Oleksiivanov the changes look good 👍 , there's just some minor feedback that needs to be addressed before we merge the PR.

clients/java/src/main/java/io/camunda/zeebe/client/impl/command/UnassignGroupFromTenantCommandImpl.java
koevskinikola: 🔧 @Oleksiivanov it might also be good to just put the `tenantKey` and `groupKey` in the command constructor. They are required (path) parameters, so it makes sense. Plus we already have a similar usage pattern with the`DeleteResourceCommandImpl`, `UpdateUserTaskCommandImpl`, and `JobUpdateCommandImpl`.

koevskinikola (approved):
LGTM @Oleksiivanov 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/group/PersistedGroup.java
```diff
@@ -9,18 +9,25 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 public class PersistedGroup extends UnpackedObject implements DbValue {
 
   private final LongProperty groupKeyProp = new LongProperty("groupKey");
   private final StringProperty nameProp = new StringProperty("name");
+  private final ArrayProperty<LongValue> tenantKeysProp =
+      new ArrayProperty<>("tenantKeys", LongValue::new);
```
remcowesterhoud, line 26: ❌ This should be aligned with users and mappings. The property should be a list of tenant ids, not keys. The reasoning behind this is that we must perform our tenant checks on ids. If we store the keys here that means we have to an additional state lookup for each of the tenants that is assigned to a group, which has an impact on performance. By storing a list of ids, we only need to get the group from the state and can perform cecks accordingly without looking up the tenant themselves.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/group/DbGroupState.java
```diff
@@ -126,6 +126,28 @@ public void delete(final long groupKey) {
     groupColumnFamily.deleteExisting(this.groupKey);
   }
 
+  @Override
+  public void addTenant(final long groupKey, final long tenantKey) {
+    this.groupKey.wrapLong(groupKey);
+    final PersistedGroup persistedGroup = groupColumnFamily.get(this.groupKey);
+    if (persistedGroup != null) {
```
remcowesterhoud, line 133: 🔧 We should remove the condition. The processor has already checked that the group exists. If someone ends up trying to add a tenant to a group somewhere and the group doesn't exist, it's compeltely fair to encounter an exception. That means the caller hasn't been thorough with the request and didn't validate the request properly.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/group/DbGroupState.java
```diff
@@ -126,6 +126,28 @@ public void delete(final long groupKey) {
...
+      persistedGroup.addTenantKey(tenantKey);
+      groupColumnFamily.update(this.groupKey, persistedGroup);
+    }
+  }
+
+  @Override
+  public void removeTenant(final long groupKey, final long tenantKey) {
+    this.groupKey.wrapLong(groupKey);
+    final var persistedGroup = groupColumnFamily.get(this.groupKey);
+    if (persistedGroup != null) {
```
remcowesterhoud, line 143: 🔧 We should remove the condition. The processor has already checked that the group exists. If someone ends up trying to add a tenant to a group somewhere and the group doesn't exist, it's compeltely fair to encounter an exception. That means the caller hasn't been thorough with the request and didn't validate the request properly.

remcowesterhoud (changes requested):
@Oleksiivanov I did a very quick and dirty review and found some small-ish things that should be changed. Overall the rest looks good to me at first glance 👍

remcowesterhoud (approved):
Thanks @Oleksiivanov

Issue #24677, opened 2024-11-13
Deployment metadata should be available in runtime state


lenaschoenburg:
Zeebe has the concept of deployments which create a bundle of resources, currently processes, forms and decision requirements.

Deployments currently have a short life, they only exist to _create_ the resouces. Once they are created on all partitions, deployments only live a shadow life by being referenced by key in some resources. The deployment key is used for features such as [deployment binding](https://docs.camunda.io/docs/next/components/best-practices/modeling/choosing-the-resource-binding-type/#supported-binding-types).

For future use and specifically to support redistribution of deployments when scaling up partitions, it'd be useful to store deployments permanently in the state such that we are able to find deployments and all associated resources quickly by key.

To avoid a mixed state where new deployments exist but old ones do not, we want to reconstruct old deployments once. Instead of using a long-running migration which poses performance and liveness issues, we want to initiate reconstruction once on startup by writing a command that does this iteratively.

```[tasklist]
### Implementation
- [ ] https://github.com/camunda/camunda/issues/24683
- [ ] https://github.com/camunda/camunda/issues/24708
- [ ] https://github.com/camunda/camunda/issues/24715
- [ ] https://github.com/camunda/camunda/issues/24817
- [ ] https://github.com/camunda/camunda/issues/25493
```

PR #25891, opened 2024-12-11
fix: disable deployment reconstruction while it's unfinished

lenaschoenburg:
This disables deployment reconstruction by not writing the first `Reconstruct` command on startup. We will re-enable it when we have time to finish #24677


remcowesterhoud (approved):
👍 😞

PR #25898, opened 2024-12-11
Wrap the GroupRecord instead of making a full copy before insertion

remcowesterhoud:
## Description

Wrapping is more performant than copying. In this place we don't need a full copy as we immediately insert it in our ColumnFamily.

## Checklist

- [ ] for CI changes:

## Related issues

relates to https://github.com/camunda/camunda/pull/25635#pullrequestreview-2481712329

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/group/PersistedGroup.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/group/PersistedGroup.java
@@ -13,6 +13,7 @@ import io.camunda.zeebe.msgpack.property.ArrayProperty;
 import io.camunda.zeebe.msgpack.property.LongProperty;
 import io.camunda.zeebe.msgpack.property.StringProperty;
 import io.camunda.zeebe.msgpack.value.StringValue;
+import io.camunda.zeebe.protocol.impl.record.value.group.GroupRecord;
 import io.camunda.zeebe.util.buffer.BufferUtil;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


remcowesterhoud:
@korthout any chance I could get a review on this very small PR?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/group/PersistedGroup.java
```diff
@@ -23,6 +24,11 @@ public PersistedGroup() {
     declareProperty(groupKeyProp).declareProperty(nameProp);
   }
 
+  public void wrap(final GroupRecord group) {
+    groupKeyProp.setValue(group.getGroupKey());
+    nameProp.setValue(group.getName());
```
korthout, line 29: 🔧 The `getName` method makes a string from the buffer, you can use `getNameBuffer()` to directly set the name buffer

remcowesterhoud:
Since this PR the `tenantIdsProp` got added to the `PersistedGroup`. Since a `GroupRecord` doesn't contain a list of tenant ids it makess sense to reset this array here. The wrap method is only called on creation at this moment. The list of tenant ids should always be empty at that point in time.

korthout:
>he wrap method is only called on creation at this moment. The list of tenant ids should always be empty at that point in time.

@remcowesterhoud But shouldn't the caller of `wrap` just trust that works regardless of when it's called? Why not just set the tenants as well?

remcowesterhoud:
> >he wrap method is only called on creation at this moment. The list of tenant ids should always be empty at that point in time.
> 
> 
> 
> @remcowesterhoud But shouldn't the caller of `wrap` just trust that works regardless of when it's called? Why not just set the tenants as well?

Because there is no list of tenants in the group record we are wrapping. So there's nothing to set.

korthout (approved):
>Because there is no list of tenants in the group record we are wrapping. So there's nothing to set.

@remcowesterhoud Oh wow! I see now that these are completely different entities with different properties. A `copyFrom` only worked because they share some property names.

LGTM 👍

PR #25906, opened 2024-12-11
ci: run backend jobs of Operate/Tasklist/Optimize/Zeebe when `webapps**` change

cmur2:
## Description

This gap was identified in https://camunda.slack.com/archives/C06F0GLJNFM/p1733737990494609 for Operate initially but also applies to Tasklist and (later) Optimize.

Since webapps schemas are also used for Zeebe exporters, it probably makes sense to run Zeebe CI too when `webapps**` change?

See https://github.com/search?q=repo%3Acamunda%2Fcamunda+io.camunda.webapps&type=code for which imports get used where.

## Checklist

- [x] for CI changes:

## Related issues

None

### DIFF
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -48,6 +48,7 @@ outputs:
         github.event_name == 'push' ||
         steps.filter-common.outputs.github-actions-change == 'true' ||
         steps.filter-operate.outputs.operate-backend-change == 'true' ||
+        steps.filter-common.outputs.webapps-change == 'true' ||
         steps.filter-common.outputs.maven-change == 'true' ||
         steps.filter-common.outputs.dockerfile-change == 'true'
       }}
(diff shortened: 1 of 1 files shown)


buccarel (approved):

houssain-barouni (approved):
If I am not mistaken, `webapps` modules are not yet used by Optimize, but this should come soon I guess

cmur2:
> not yet used by Optimize, but this should come soon I guess

Agree, I didn't find any current usage but want to be prepared - Optimize is a "webapp" after all so I think it makes sense to add it already.

RomanJRW:
@buccarel - I saw you approve the PR already, but can you confirm the assumption from @cmur2 above about the Optimize webapp?

cmur2:
Please merge this PR as soon as all feedback is addressed and all affected teams approved to close this gap in the change detection ASAP.

ChrisKujawa:
Shouldn't it be the other way around? When Zeebe changes all other dependents should be build and tested? 

If we change something in webapp this wouldn't impact the lower layer like Zeebe right?

npepinpe (approved):

npepinpe:
The thing is, we do use the webapps schema module in the exporter. That's the reasoning here - we should at least run the exporter related tests in such cases.

We could possibly narrow it down to `webapps-schema` only, I guess.

buccarel:
> > not yet used by Optimize, but this should come soon I guess
> 
> Agree, I didn't find any current usage but want to be prepared - Optimize is a "webapp" after all so I think it makes sense to add it already.

correct, currently optimize is a webapp but is not following the code structure of others. This was agreed some time ago, and the reason was that it would still work. Eventually, we will follow the same approach as the other applications.

PR #25918, opened 2024-12-11
feat: Support resource in deployment create processor

### LINKED ISSUE #24772: Support RPA file deployment in `DeploymentCreateProcessor` class
We need to support RPA file deployments through the `DeploymentRecord` and the `DeploymentCreatedProcessor`:

- [ ] Include Resource metadata to `DeploymentRecord`
- [ ] Implement `RpaResourceTransformer` and include it in `DeploymentTransformer` ([example](https://github.com/camunda/camunda/blob/46163c715fd8dfe6ac591c9430a1e487920ba364/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/FormResourceTransformer.java#L31))
- [ ] Transform deployed RPA file in `DeploymentCreateProcessor` class ([forms example](https://github.com/camunda/camunda/blob/9390158dca8098722013d4f1add8e7229cdaf803/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/DeploymentCreateProcessor.java#L253-L267))

kzraitel:
## Checklist

- [ ] for CI changes:

## Related issues

closes #24772

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
@@ -35,11 +35,9 @@ import org.slf4j.Logger;
 public final class DeploymentTransformer {
 
   private static final Logger LOG = Loggers.PROCESS_PROCESSOR_LOGGER;
-
+  private static final DeploymentResourceTransformer UNKNOWN_RESOURCE =
+      new UnknownResourceTransformer();
   private final Map<String, DeploymentResourceTransformer> resourceTransformers;
-
-  private final ResourceTransformer defaultResourceTransformer;
-
   private final ChecksumGenerator checksumGenerator = new ChecksumGenerator();
   // internal changes during processing
   private RejectionType rejectionType;
@@ -72,8 +70,8 @@ public final class DeploymentTransformer {
         new FormResourceTransformer(
             keyGenerator, stateWriter, checksumGenerator, processingState.getFormState());
 
-    defaultResourceTransformer =
-        new ResourceTransformer(
+    final var resourceTransformer =
+        new RpaTransformer(
             keyGenerator, stateWriter, checksumGenerator, processingState.getResourceState());
 
     resourceTransformers =
@@ -81,7 +79,8 @@ public final class DeploymentTransformer {
             entry(".bpmn", bpmnResourceTransformer),
             entry(".xml", bpmnResourceTransformer),
             entry(".dmn", dmnResourceTransformer),
-            entry(".form", formResourceTransformer));
+            entry(".form", formResourceTransformer),
+            entry(".rpa", resourceTransformer));
   }
 
   public DirectBuffer getChecksum(final byte[] resource) {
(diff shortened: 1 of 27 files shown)


zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/deployment/DeploymentRecord.java
```diff
@@ -193,10 +215,22 @@ public boolean hasForms() {
...
+  public boolean hasResources() {
+    return getResources().stream()
+        .map(io.camunda.zeebe.protocol.record.value.deployment.DeploymentResource::getResourceName)
+        .anyMatch(
+            x ->
+                !x.endsWith(".form")
+                    && !x.endsWith(".xml")
+                    && !x.endsWith(".bpmn")
+                    && !x.endsWith(".dmn"));
+  }
```
koevskinikola, line 227: ❌  This method will allow any type of resource other than a Form, BPMN, DMN, or XML file. We also need something to check for robot script files before transforming them.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/ResourceTransformer.java
```diff
@@ -0,0 +1,195 @@
...
+                resourceMetadataRecord
+                    .setResourceKey(newResourceKey.getAsLong())
+                    .setVersion(INITIAL_VERSION)
+                    .setDeploymentKey(deploymentRecord.getDeploymentKey()));
+  }
+
+  private Either<Failure, Resource> parseResource(final DeploymentResource resource) {
+    try {
+      final var res = JSON_MAPPER.readValue(resource.getResource(), Resource.class);
+      return validateResource(res);
```
koevskinikola, line 152: ❌  This code assumes that a resource file has JSON content. That's only expected if the resource file is a robot (RPA) script, but it doesn't have to be. And the whole `ResourceTransformer` class will be applied to any resources that are not BPMN, DMN, Form, or XML according to the usage of [this method](https://github.com/camunda/camunda/pull/25918/files#diff-b647b3d6d8c4f102119b11dfcaec9673c86cd3955efe6226c121495e12e070f0R223-R226).

I think we should re-name this transformer into a `RobotScriptTransformer` or `RpaTransformer`. Persisting into the `ResourceState` still makes sense, but the transformation logic should remain specific for RPA scripts.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
```diff
@@ -36,11 +36,10 @@ public final class DeploymentTransformer {
 
   private static final Logger LOG = Loggers.PROCESS_PROCESSOR_LOGGER;
 
-  private static final DeploymentResourceTransformer UNKNOWN_RESOURCE =
-      new UnknownResourceTransformer();
-
```
koevskinikola, line 41: ❌ Let's restore this, we still haven't developed any concepts for a general-resource transformation mechanism.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
```diff
@@ -216,23 +219,6 @@ private static void handleUnexpectedError(
...
-      final var failureMessage =
-          String.format("%n'%s': unknown resource type", resource.getResourceName());
-      return Either.left(new Failure(failureMessage));
-    }
-
-    @Override
-    public void writeRecords(
-        final DeploymentResource resource, final DeploymentRecord deployment) {}
-  }
-
```
koevskinikola, line 235: ❌ Let's revert this, we may still encounter non-supported files.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/transform/DeploymentTransformer.java
```diff
@@ -73,6 +72,10 @@ public DeploymentTransformer(
         new FormResourceTransformer(
             keyGenerator, stateWriter, checksumGenerator, processingState.getFormState());
 
+    defaultResourceTransformer =
+        new ResourceTransformer(
+            keyGenerator, stateWriter, checksumGenerator, processingState.getResourceState());
```
koevskinikola, line 77: ❌ The newly-implemented `ResourceTransformer` is specific to JSON files, and we should refactor it to be specific for RPA/Robot scripts. We can't make that assumption that this is the default for all deployed (non-BPMN/DMN/Form/XML) files.

koevskinikola (changes requested):
Hey @kzraitel, the code you implemented generally looks good. However, you implemented a generic resource transformer, to transform any type of resource, instead of one for RPA resources.

Let's focus for now on expanding the deployment steps to support RPA/robot scripts only. 

I'd like to validate any "default" resource transformer ideas with my team before we implement them.

koevskinikola (approved):
Looks good now @kzraitel 👍

PR #25919, opened 2024-12-11
feat: implement tenants' users  migration

### LINKED ISSUE #25844: Implement TenantMemberMigrationHandler for the membertype == USER

maryarm:
## Description

In this PR, the logic for migrating users assigned to a tenant is implemented.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25844


migration/identity-migration/src/main/java/io/camunda/migration/identity/UserTenantsMigrationHandler.java
```diff
@@ -0,0 +1,74 @@
...
+      userTenants = managementIdentityClient.fetchUserTenants(SIZE);
+      managementIdentityClient.updateMigrationStatus(
+          userTenants.stream().map(this::createTenantUser).toList());
+    } while (!userTenants.isEmpty());
+  }
+
+  private MigrationStatusUpdateRequest createTenantUser(final UserTenants userTenants) {
+
+    try {
+      final var userKey = mappingService.findOrCreateUserWithUsername(userTenants.username());
```
lenaschoenburg, line 53: :question: This creates a mapping, not a user? Is this expected?

maryarm: Good question and yes, it should be a mapping. because we have the User entity only for local testing and in production we should use the Mapping instead of Users. As I expect the migration to be used in prod and not for local testing, we only support mapping.

lenaschoenburg (approved):

PR #25920, opened 2024-12-11
feat: verify permissions of mappings

### LINKED ISSUE #24176: Zeebe can verify permissions of Mappings
<h2>Description</h2>

The `AuthorizationCheckBehavior` currently only checks authorizations on a user level. We need to support Authorization checks on Mapping level as well.

lenaschoenburg:
The `AuthorizationCheckBehavior` is extended to also check request claims for tenants and authorizations of matching mappings.

I've included a refactoring to store tenant ids instead of keys in the `PersistedMapping`. This is more aligned with how users work and makes it easier to verify a request's tenant id against the claimed mapping.

Closes #24176

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
@@ -15,7 +15,6 @@ import io.camunda.zeebe.engine.state.immutable.MappingState;
 import io.camunda.zeebe.engine.state.immutable.ProcessingState;
 import io.camunda.zeebe.engine.state.immutable.TenantState;
 import io.camunda.zeebe.engine.state.immutable.UserState;
-import io.camunda.zeebe.engine.state.tenant.PersistedTenant;
 import io.camunda.zeebe.engine.state.user.PersistedUser;
 import io.camunda.zeebe.protocol.record.RejectionType;
 import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
@@ -124,10 +123,7 @@ public final class AuthorizationCheckBehavior {
     if (request.tenantId.equals(TenantOwned.DEFAULT_TENANT_IDENTIFIER)) {
       return true;
     }
-    return tenantState
-        .getTenantKeyById(request.getTenantId())
-        .filter(persistedTenantKey -> mapping.getTenantKeysList().contains(persistedTenantKey))
-        .isPresent();
+    return mapping.getTenantIdsList().contains(request.getTenantId());
   }
 
   public Set<String> getAllAuthorizedResourceIdentifiers(final AuthorizationRequest request) {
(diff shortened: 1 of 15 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -165,6 +182,27 @@ private Stream<String> getUserAuthorizedResourceIdentifiers(
         Stream.concat(roleAuthorizedResourceIdentifiers, groupAuthorizedResourceIdentifiers));
   }
 
+  private Stream<String> getMappingsAuthorizedResourceIdentifiers(
+      final AuthorizationRequest request) {
+    return extractUserTokenClaims(request.getCommand())
+        .<PersistedMapping>mapMulti(
+            (claim, stream) ->
+                mappingState.get(claim.claimName(), claim.claimValue()).ifPresent(stream))
```
nicpuppa, line 187: 💭 This is just a thought. In the token there might be multiple claims, we are going to try to fetch all of them from state

nicpuppa:
I'm going to review this PR now

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -207,6 +245,23 @@ private static Optional<Long> getUserKey(final TypedRecord<?> command) {
...
+        .filter(entry -> entry.getKey().startsWith(Authorization.USER_TOKEN_CLAIM_PREFIX))
+        .flatMap(
+            claimEntry -> {
+              final var claimName =
+                  claimEntry.getKey().substring(Authorization.USER_TOKEN_CLAIM_PREFIX.length());
+              final var claimValue = claimEntry.getValue();
+              if (claimValue instanceof final Collection<?> collection) {
+                return collection.stream()
+                    .map(value -> new UserTokenClaim(claimName, value.toString()));
+              } else {
```
nicpuppa, line 263: 🔧 This else statement can be removed

nicpuppa (changes requested):
The implementation looks good 💪 

❌ I'm missing some QA IT tests, you can have a look at https://github.com/camunda/camunda/tree/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/authorization

lenaschoenburg:
> ❌ I'm missing some QA IT tests, you can have a look at https://github.com/camunda/camunda/tree/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/authorization

@nicpuppa What tests specifically are you missing? The tests you mentioned all just check whether commands check for authorization as expected. Since all of these processors just call into the `AuthorizationCheckBehavior`, which is already properly tested, I don't see the value of duplicating all of them to also test with mappings now.

nicpuppa:
> @nicpuppa What tests specifically are you missing? The tests you mentioned all just check whether commands check for authorization as expected. Since all of these processors just call into the `AuthorizationCheckBehavior`, which is already properly tested, I don't see the value of duplicating all of them to also test with mappings now.

The tests are testing the processors behaviour with users, I think we should tests the same but with mappings.

lenaschoenburg:
There are over 4000 lines of code in that module. I don't think it makes sense to duplicate this now, just to test mappings. The intent of these tests, at least as far as I can tell, is to check whether the processors call into `AuthorizationCheckBehavior` as expected, nothing more. As far as I can see, there are also not tests that check whether authorizations inherited through groups and roles are checked properly either.

I can "smoke" tests that verify that mappings are checked for one or two commands just to be sure. But surely not for every single command in there, unless we want to start testing all permutations of commands, users, mappings, groups, roles, tenants and authorizations.

lenaschoenburg:
@nicpuppa Writing these tests is exceedingly difficult right now so I've created a follow up issue for this: https://github.com/camunda/camunda/issues/26015

Could you approve for now?

nicpuppa (approved):

PR #25928, opened 2024-12-11
feat: make rdbms exporter configurable via environment #23821

### LINKED ISSUE #23821: Make Exporter properties configurable
As operator I want to make the exporter properties configurable.

**Acceptance criteria**
- ExecutionQueue#queueFlushSize is configurable via application.yaml
- RdbmsExporter#flushInterval is configurable
  - when interval value == null, then flush always on each item (e.g. in ITests)

stoerti:
## Description

Goals:
- make the following rdbmsExporterParameters configurable:
  - queueFlushSize: controls the max size the exporters' executionQueue should have before flushing
  - flushInterval: An ISO-8861 interval after which the exporters' executionQueue will always be flushed even if size < queueFlushSize

Configuration can happen via yaml:
```yaml
zeebe:
  broker:
    exporters:
      rdbms:
        args:
          flushInterval: 5000
          maxQueueSize: 1000
```

Side goals:
- Whenever either queueFlushSize or flushInterval is ZERO, the exporter will flush after each processed record

Refactorings:
- since the RdbmsExporter contains a lot logic now, I added some UnitTests. To do so properly, I had to split the RdbmsExporter into two classes. The RdbmsExporterWrapper now does the configuration of the actual RdbmsExporter and forwards records and commands to it.

## Related issues

closes #23821

### DIFF
--- a/dist/src/main/resources/application-rdbmsH2.yaml
+++ b/dist/src/main/resources/application-rdbmsH2.yaml
@@ -84,6 +84,9 @@ zeebe:
     exporters:
       rdbms:
         className: io.camunda.exporter.rdbms.RdbmsExporter
+        args:
+          flushInterval: 500
+          maxQueueSize: 1000
       #elasticsearch:
       #  className: io.camunda.zeebe.exporter.ElasticsearchExporter
       #  args:
--- a/dist/src/main/resources/application-rdbmsPostgres.yaml
+++ b/dist/src/main/resources/application-rdbmsPostgres.yaml
@@ -85,6 +85,9 @@ zeebe:
       rdbms:
         # Classname doesn't matter, since this exporter is present via spring bean
         className: RdbmsExporter
+        args:
+          flushInterval: 500
+          maxQueueSize: 1000
       #elasticsearch:
       #  className: io.camunda.zeebe.exporter.ElasticsearchExporter
       #  args:
--- a/zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/RdbmsExporter.java
+++ b/zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/RdbmsExporter.java
@@ -7,66 +7,56 @@
  */
 package io.camunda.exporter.rdbms;
 
-import io.camunda.db.rdbms.RdbmsService;
 import io.camunda.db.rdbms.write.RdbmsWriter;
 import io.camunda.db.rdbms.write.domain.ExporterPositionModel;
-import io.camunda.zeebe.exporter.api.Exporter;
-import io.camunda.zeebe.exporter.api.context.Context;
 import io.camunda.zeebe.exporter.api.context.Controller;
 import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.ValueType;
 import io.camunda.zeebe.util.VisibleForTesting;
 import java.time.Duration;
 import java.time.LocalDateTime;
-import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import org.slf4j.Logger;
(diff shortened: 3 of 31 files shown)


dist/src/main/resources/application-rdbmsH2.yaml
```diff
@@ -84,6 +84,9 @@ zeebe:
     exporters:
       rdbms:
         className: io.camunda.exporter.rdbms.RdbmsExporter
+        args:
+          flushInterval: 5000
```
p-wunderlich, line 88: I think above all for h2 we should use a lower flushInterval. Since this will mainly be used for local development. (500?)

dist/src/main/resources/application-rdbmsPostgres.yaml
```diff
@@ -85,6 +85,9 @@ zeebe:
       rdbms:
         # Classname doesn't matter, since this exporter is present via spring bean
         className: RdbmsExporter
+        args:
+          flushInterval: 5000
```
p-wunderlich, line 89: Also vor normal databases we should use a lower flushInterval. Otherwise it is as slow as the current impl where Camunda also have a 5s timeout somewhere.

zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/RdbmsExporter.java
```diff
@@ -7,66 +7,56 @@
...
+      controller.scheduleCancellableTask(flushInterval, this::flushAndReschedule);
+    }
 
     initializeRdbmsPosition();
     lastPosition = controller.getLastExportedRecordPosition();
     if (exporterRdbmsPosition.lastExportedPosition() > -1
         && lastPosition <= exporterRdbmsPosition.lastExportedPosition()) {
-      // This is needed since the brokers last exported position is from it's last snapshot and kann
-      // be different than ours.
+      // This is needed since the brokers last exported position is from its last snapshot and can
```
p-wunderlich, line 58: "kann" be different. :D

zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/RdbmsExporterWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** https://docs.camunda.io/docs/next/components/zeebe/technical-concepts/process-lifecycles/ */
+public class RdbmsExporterWrapper implements Exporter {
+
+  /** The partition on which all process deployments are published */
+  public static final long PROCESS_DEFINITION_PARTITION = 1L;
+
+  private static final int DEFAULT_FLUSH_INTERVAL = 5000;
```
p-wunderlich, line 41: I vote for a smaller flush interval. :D

zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/RdbmsExporterWrapper.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  @Override
+  public void configure(final Context context) {
+    if (context.getConfiguration().getArguments() != null) {
+      flushInterval =
+          Duration.ofMillis(
+              (Integer)
+                  context
+                      .getConfiguration()
```
p-wunderlich, line 66: Arg this formatting ... can you find a better way to set those props?

p-wunderlich (commented):
Hm a ExporterFactory creates a ExporterWrapper which creates a Exporter ...

p-wunderlich (approved):

PR #25943, opened 2024-12-11
Add java client group update command

### LINKED ISSUE #25930: Java client supports group updating

koevskinikola:
## Description

Implement a Camunda Java client Group update command with JUnit tests and Zeebe IT tests for the Group update feature.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25930


koevskinikola:
@nicpuppa @emulbreh just an FYI for a fix I applied here with commit [2ff3d73](https://github.com/camunda/camunda/pull/25943/commits/2ff3d735b4b43719d98fe6fdd1db28b0939c6b80). The same can be done for Roles.

If the partition ID is explicitly set to 1 (`DEPLOYMENT_PARTITION`), and the key doesn't exist, we prevent the Gateway from detecting false keys and returning `PartitionNotFoundExceptions`. The explicit setting was actually causing a socket timeout because the Broker was taking too long to locate a group for a false key.

Why we don't need to explicitly set the partition ID during (non-deployment/creation) operations?
A group is created on partition 1 (`DEPLOYMENT_PARTITION`), and the partition ID is encoded in the group key. With update operations, the partition ID can already be decoded from the key.

koevskinikola:
@Ben-Sheppard do you have some time to review this PR? Most of the team is busy with other tasks.

Oleksiivanov (approved):

PR #25970, opened 2024-12-11
feat: Replace string with enum in FlownodeInstanceFilter

### LINKED ISSUE #25908: Replace `string` with `enum` in `FlownodeInstanceFilter`

ghost:
feat: Replace string with enum in FlownodeInstanceFilter

Closes https://github.com/camunda/camunda/issues/25908


tmetzke (approved):
👍 That will make it a lot easier for Java client users 🚀

houssain-barouni (changes requested):
@mihail-ca

If we follow the [PartitionInfo.role](https://github.com/camunda/camunda/blob/7e9b19bcd1abae8262706616b237796f241b3537/clients/java/src/main/java/io/camunda/zeebe/client/impl/response/PartitionInfoImpl.java#L30) example:
	•	The [PartitionBrokerRole](https://github.com/camunda/camunda/blob/main/clients/java/src/main/java/io/camunda/zeebe/client/api/response/PartitionBrokerRole.java) enum is defined within the client model.
	•	The protocol enum is mapped to this client enum [here](https://github.com/camunda/camunda/blob/7e9b19bcd1abae8262706616b237796f241b3537/clients/java/src/main/java/io/camunda/zeebe/client/impl/response/PartitionInfoImpl.java#L36-L47).

This approach abstracts the client model from the protocol classes. Similarly, we could define a single enum class for the FlowNodeInstance state exposed to the client user, instead of using two enums: `FlowNodeInstanceFilterRequest.StateEnum` and `FlowNodeInstanceItem.StateEnum`. This would simplify the client user’s experience by eliminating the need to map between these two enums.

tmetzke:
> @mihail-ca
> 
> If we follow the [PartitionInfo.role](https://github.com/camunda/camunda/blob/7e9b19bcd1abae8262706616b237796f241b3537/clients/java/src/main/java/io/camunda/zeebe/client/impl/response/PartitionInfoImpl.java#L30) example: • The [PartitionBrokerRole](https://github.com/camunda/camunda/blob/main/clients/java/src/main/java/io/camunda/zeebe/client/api/response/PartitionBrokerRole.java) enum is defined within the client model. • The protocol enum is mapped to this client enum [here](https://github.com/camunda/camunda/blob/7e9b19bcd1abae8262706616b237796f241b3537/clients/java/src/main/java/io/camunda/zeebe/client/impl/response/PartitionInfoImpl.java#L36-L47).
> 
> This approach abstracts the client model from the protocol classes. Similarly, we could define a single enum class for the FlowNodeInstance state exposed to the client user, instead of using two enums: `FlowNodeInstanceFilterRequest.StateEnum` and `FlowNodeInstanceItem.StateEnum`. This would simplify the client user’s experience by eliminating the need to map between these two enums.

Yeah, good point, @houssain-barouni. Not exposing the protocol class (here: the enum from the REST protocol defined in OpenAPI) is generally something we should look out for.

Different follow-up issue I think: we need to check this in the client in general. I think we started exposing the REST classes here and there in search filter commands.

mihail-ca:
Ok, I will rework this. Just to clarify, so I should create an enum in the client that wraps (and hides) the enums from OpenAPI specs ?

clients/java/src/main/java/io/camunda/zeebe/client/api/search/FlowNodeInstanceState.java
```diff
@@ -0,0 +1,34 @@
...
+  COMPLETED,
+  TERMINATED;
+
+  public static FlowNodeInstanceFilterRequest.StateEnum toProtocolState(
+      FlowNodeInstanceState value) {
+    return (value == null) ? null : FlowNodeInstanceFilterRequest.StateEnum.fromValue(value.name());
+  }
+
+  public static FlowNodeInstanceState fromProtocolState(FlowNodeInstanceItem.StateEnum value) {
+    return (value == null) ? null : FlowNodeInstanceState.valueOf(value.name());
```
houssain-barouni, line 32: If the client is using an older version than the C8 cluster, the value can default to `UNKNOWN_DEFAULT_OPEN_API ` when the API returns an enum value that is not recognized by the protocol enum used by the client. In this scenario, should we map it to a predefined `UNKNOWN` enum value, or should we throw an explicit exception to indicate that this value is not supported? WDYT @mihail-ca @tmetzke?

mihail-ca: Currently [an exception is thrown](https://github.com/camunda/camunda/pull/25970/files#diff-ea8dac8514d4a25ce4625e646b4908230db960edb40537a66cfb3a4232761ec6R148-R155), when the protocol value is UNKNOWN_DEFAULT_OPEN_API. But honestly I don't know how this UNKNOWN_DEFAULT_OPEN_API enum is supposed to work, and how we should handle it.

tmetzke: I generally like the idea of keeping the client forward-compatible by being more resilient and handling unknown values. However, if I am not mistaken, other parts of the client are not consistent with this either, e.g. the [PartitionInfo](https://github.com/camunda/camunda/blob/7e9b19bcd1abae8262706616b237796f241b3537/clients/java/src/main/java/io/camunda/zeebe/client/impl/response/PartitionInfoImpl.java#L30) you mentioned earlier in this PR.

If we want this, I'd propose to create a follow-up issue to make the client more resilient and forward-compatible because it should be consistent and there might be even more places where this is needed.

What do you think, @houssain-barouni?

tmetzke: I guess we could add test cases that prove that the client throws an exception when the server actually returns an enum value we cannot handle, for the partition info and the flow node type and state enums.

mihail-ca: @tmetzke I added tests with exception [here](https://github.com/camunda/camunda/pull/25970/files#diff-ea8dac8514d4a25ce4625e646b4908230db960edb40537a66cfb3a4232761ec6R148). Or do you have something else in mind ? Thank you.

tmetzke: That is quite "internal" as a test and it's valid. However, I was thinking more of a test case that fires a search request from the client, mocks the server response to an unknown enum, and asserts that the client throws an exception in that case. Using the OpenAPI-specific enum type is an implementation detail that might change with a new version of the OpenAPI generator, actually.

clients/java/src/main/java/io/camunda/zeebe/client/api/search/FlowNodeInstanceState.java
```diff
@@ -0,0 +1,34 @@
...
+import io.camunda.client.protocol.rest.FlowNodeInstanceItem;
+
+public enum FlowNodeInstanceState {
+  ACTIVE,
+  COMPLETED,
+  TERMINATED;
+
+  public static FlowNodeInstanceFilterRequest.StateEnum toProtocolState(
+      FlowNodeInstanceState value) {
+    return (value == null) ? null : FlowNodeInstanceFilterRequest.StateEnum.fromValue(value.name());
```
tmetzke, line 28: 🔧 I suggest we handle this consistently in the client regarding error messages in such cases. Looking at the [PartitionInfo handling](https://github.com/camunda/camunda/blob/main/clients/java/src/main/java/io/camunda/zeebe/client/impl/response/PartitionInfoImpl.java#L36-L47), the error message is more user-focused. Let's do this here as well, ideally by creating a utility method for enum conversion that always yields the same error message for consistency.

clients/java/src/main/java/io/camunda/zeebe/client/api/search/FlowNodeInstanceState.java
tmetzke: I decided to follow @houssain-barouni's suggestion here and handle such cases gracefully. I also adjusted a bit how we do this in the existing places by introducing a dedicated enum value for such cases to allow differentiating those cases. I also added debug logs for such cases in the client.

tmetzke:
Some my rebasing was messed up and there are some commits in here that are unrelated, I am trying to remove them from this PR now.

tmetzke:
> Somehow my rebasing was messed up and there are some commits in here that are unrelated, I am trying to remove them from this PR now.

Fixed, I think. Plus, all open issues have been addressed, as far as I can see.
@houssain-barouni, would you mind giving this a second look? Ready from my end.

clients/java/src/main/java/io/camunda/client/impl/response/PartitionInfoImpl.java
```diff
@@ -18,14 +18,18 @@
...
 import io.camunda.client.protocol.rest.Partition.RoleEnum;
 import io.camunda.zeebe.gateway.protocol.GatewayOuterClass.Partition;
-import java.util.Arrays;
 import java.util.Objects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class PartitionInfoImpl implements PartitionInfo {
 
+  private static final Logger LOG = LoggerFactory.getLogger(PartitionInfoImpl.class);
```
houssain-barouni, line 31: it seems that is not used

houssain-barouni (approved):
Looks great 🚀 !
one minor comment about an unused LOG attribute

clients/java/src/main/java/io/camunda/client/impl/response/PartitionInfoImpl.java
tmetzke: Oh, nice catch, yep, unused 👍

PR #25998, opened 2024-12-12
Migrate joining parallel gateway with at least one incoming sequence flow is taken

### LINKED ISSUE #25640: Migrate joining parallel gateway with at least one incoming sequence flow is taken
## Description

We aim to extend the capabilities of our migration tool to support the migration of parallel gateway, enabling the tool to work with a broader range of BPMN elements. First parallel gateway migration case is active parallel gateway due to an incident is explained in issue #22128. Second case to migrate a joining parallel gateway is when at least one incoming sequence flow of the parallel is taken.

In this case, users **HAVE TO** provide a mapping instruction for the parallel gateway. That way we make sure taken sequence flow count on the gateway will remain the same in the target. Let's consider example:

Screenshot 2024-12-05 at 17 06 56

In the example above, assume that task `A` is completed and sequence flow `1` is taken. Users **have to** provide mapping instructions for `B -> B` and `join -> join`.

berkaycanbc:
## Description

I used the existing `ELEMENT_MIGRATED`  event to apply the gateway migration. It also conforms with the migrating logic since we allow migrating the gateway to a new id. However, it was a little tricky to handle it in the applier. Because, we do not have the element instance created for the gateway yet. The element instance for the gateway is created only when all incoming sequence flows to the gateway are taken. For me, this way looks like the one that matches with the existing element migration implementation but if any, I am open to discuss alternatives to handle the gateway migration differently.

Please note that gateway mapping validations are not implemented in this PR to keep it smaller. They will be implemented in the next PR.

## Related issues

closes #25640

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -296,7 +296,8 @@ public class ProcessInstanceMigrationMigrateProcessor
         targetProcessDefinition,
         targetElementId,
         processInstanceKey);
-    requireNoConcurrentCommand(eventScopeInstanceState, elementInstance, processInstanceKey);
+    requireNoConcurrentCommand(
+        eventScopeInstanceState, elementInstance, sourceProcessDefinition, processInstanceKey);
 
     stateWriter.appendFollowUpEvent(
         elementInstance.getKey(),
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -13,6 +13,8 @@ import static io.camunda.zeebe.engine.state.immutable.IncidentState.MISSING_INCI
 
 import io.camunda.zeebe.engine.Loggers;
 import io.camunda.zeebe.engine.processing.bpmn.behavior.BpmnBehaviors;
+import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableFlowNode;
+import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableSequenceFlow;
 import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
 import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior;
 import io.camunda.zeebe.engine.processing.identity.AuthorizationCheckBehavior.AuthorizationRequest;
@@ -34,6 +36,7 @@ import io.camunda.zeebe.engine.state.instance.ElementInstance;
 import io.camunda.zeebe.engine.state.routing.RoutingInfo;
 import io.camunda.zeebe.msgpack.spec.MsgPackHelper;
 import io.camunda.zeebe.protocol.impl.record.value.processinstance.ProcessInstanceMigrationRecord;
+import io.camunda.zeebe.protocol.impl.record.value.processinstance.ProcessInstanceRecord;
 import io.camunda.zeebe.protocol.impl.record.value.variable.VariableRecord;
 import io.camunda.zeebe.protocol.record.RejectionType;
 import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
@@ -43,16 +46,20 @@ import io.camunda.zeebe.protocol.record.intent.ProcessInstanceMigrationIntent;
 import io.camunda.zeebe.protocol.record.intent.UserTaskIntent;
 import io.camunda.zeebe.protocol.record.intent.VariableIntent;
 import io.camunda.zeebe.protocol.record.value.AuthorizationResourceType;
+import io.camunda.zeebe.protocol.record.value.BpmnElementType;
 import io.camunda.zeebe.protocol.record.value.BpmnEventType;
 import io.camunda.zeebe.protocol.record.value.PermissionType;
 import io.camunda.zeebe.protocol.record.value.ProcessInstanceMigrationRecordValue.ProcessInstanceMigrationMappingInstructionValue;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
+import io.camunda.zeebe.stream.api.state.KeyGenerator;
 import io.camunda.zeebe.util.buffer.BufferUtil;
 import java.util.ArrayDeque;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
+import org.agrona.DirectBuffer;
 import org.agrona.concurrent.UnsafeBuffer;
 import org.slf4j.Logger;
 
@@ -75,6 +82,7 @@ public class ProcessInstanceMigrationMigrateProcessor
   private final EventScopeInstanceState eventScopeInstanceState;
   private final AuthorizationCheckBehavior authCheckBehavior;
   private final ProcessInstanceMigrationCatchEventBehaviour migrationCatchEventBehaviour;
+  private final KeyGenerator keyGenerator;
 
   public ProcessInstanceMigrationMigrateProcessor(
       final Writers writers,
@@ -83,7 +91,8 @@ public class ProcessInstanceMigrationMigrateProcessor
       final CommandDistributionBehavior commandDistributionBehavior,
       final int partitionId,
       final RoutingInfo routingInfo,
-      final AuthorizationCheckBehavior authCheckBehavior) {
+      final AuthorizationCheckBehavior authCheckBehavior,
+      final KeyGenerator keyGenerator) {
     stateWriter = writers.state();
     responseWriter = writers.response();
     rejectionWriter = writers.rejection();
@@ -95,6 +104,7 @@ public class ProcessInstanceMigrationMigrateProcessor
     incidentState = processingState.getIncidentState();
     eventScopeInstanceState = processingState.getEventScopeInstanceState();
     this.authCheckBehavior = authCheckBehavior;
+    this.keyGenerator = keyGenerator;
 
     migrationCatchEventBehaviour =
         new ProcessInstanceMigrationCatchEventBehaviour(
@@ -308,6 +318,32 @@ public class ProcessInstanceMigrationMigrateProcessor
             .setVersion(targetProcessDefinition.getVersion())
             .setElementId(targetElementId));
 
+    final List<DirectBuffer> activeSequenceFlowIds = elementInstance.getActiveSequenceFlowIds();
+    if (!activeSequenceFlowIds.isEmpty()) {
+      final Set<ExecutableFlowNode> mappedGateways =
+          getMappedGateways(
+              sourceProcessDefinition,
+              targetProcessDefinition,
+              sourceElementIdToTargetElementId,
+              elementInstance);
+
+      mappedGateways.forEach(
+          targetGateway -> {
+            final var gatewayInstanceRecord = new ProcessInstanceRecord();
+            gatewayInstanceRecord.wrap(elementInstanceRecord);
+            gatewayInstanceRecord
+                .setElementId(targetGateway.getId())
+                .setBpmnElementType(targetGateway.getElementType())
+                .setBpmnEventType(targetGateway.getEventType())
+                .setFlowScopeKey(elementInstance.getKey());
+
+            stateWriter.appendFollowUpEvent(
+                keyGenerator.nextKey(),
+                ProcessInstanceIntent.ELEMENT_MIGRATED,
+                gatewayInstanceRecord);
+          });
+    }
+
     if (elementInstance.getJobKey() > 0) {
       final var job = jobState.getJob(elementInstance.getJobKey());
       if (job == null) {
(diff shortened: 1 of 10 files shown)


berkaycanbc:
@korthout please take a look at my notes in the PR description as well.

korthout (commented):
Thanks @berkaycanbc 🏗️ 

👍 Great work on retrieving all the needed data in the processor

I have two main concerns:
1. ❌ if we remove the concurrency check for activation of parallel gateway, then we need to add test cases that shows that this functions correctly. My worry is that the values in the `ACTIVATE_ELEMENT` command are taken as truth, while the values may have changed due to the migration. Example:
   - complete a task, which takes a sequence flow, and appends an `ACTIVATE_ELEMENT` command for the gateway
   - execute a `MIGRATE` command, changing the id of the gateway
   - try to process the `ACTIVATE_COMMAND` with out-of-date information like `processDefinitionKey`, `elementId` and such.
2. ❓ this implements the migration by writing `ELEMENT_MIGRATED` event for the gateway. However, no instance for the gateway exists yet. This becomes clear when we have to generate a key for it. I cannot find a reasoning for taking this approach over writing events for the taken sequence flows themselves. For example, `ELEMENT_MIGRATED` or a new `SEQUENCE_FLOW_MIGRATED` event. Did you consider this option?

berkaycanbc:
> ❓ this implements the migration by writing ELEMENT_MIGRATED event for the gateway. However, no instance for the gateway exists yet. This becomes clear when we have to generate a key for it. I cannot find a reasoning for taking this approach over writing events for the taken sequence flows themselves. For example, ELEMENT_MIGRATED or a new SEQUENCE_FLOW_MIGRATED event. Did you consider this option?

I think, migrating sequence flows makes more sense since we do not have an instance for the gateway yet. The reasoning was to align with behaviour where we need to update gateway ids in the state. Updated. 👍

berkaycanbc:
❌ if we remove the concurrency check for activation of parallel gateway, then we need to add test cases that shows that this functions correctly. My worry is that the values in the ACTIVATE_ELEMENT command are taken as truth, while the values may have changed due to the migration. Example:
complete a task, which takes a sequence flow, and appends an ACTIVATE_ELEMENT command for the gateway
execute a MIGRATE command, changing the id of the gateway
try to process the ACTIVATE_COMMAND with out-of-date information like processDefinitionKey, elementId and such.

@korthout activation of the gateway 'join1' should be rejected because not (actually none) all incoming sequence flows are taken. But we have no way to escape from this issue, right? Users must use instance modification to fix the instance.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceElementMigratedApplier.java
```diff
@@ -48,4 +59,32 @@ public void applyState(final long elementInstanceKey, final ProcessInstanceRecor
...
+            IntStream.range(0, number)
+                .forEach(
+                    ignore -> {
+                      // decrement the number of taken sequence flows for the old gateway
+                      elementInstanceState.decrementNumberOfTakenSequenceFlows(
+                          flowScopeKey, sourceGatewayId, sequenceFlowId);
+                      // increment the number of taken sequence flows for the new gateway
+                      elementInstanceState.incrementNumberOfTakenSequenceFlows(
+                          flowScopeKey, targetGatewayId, sequenceFlowId);
+                    });
```
korthout, line 86: 👍 I like it!

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/DbElementInstanceState.java
```diff
@@ -292,6 +292,27 @@ public void deleteProcessInstanceKeyByDefinitionKey(
         processInstanceKeyByProcessDefinitionKey);
   }
 
+  @Override
+  public void decrementNumberOfTakenSequenceFlows(
```
korthout, line 296: 🔧 Let's move this method directly under the other `decrementNumberOfTakenSequenceFlows` so they stay close together.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateParallelGatewayTest.java
```diff
@@ -112,4 +115,292 @@ public void shouldMigrateParallelGatewayWithIncident() {
...
+    ENGINE.job().ofInstance(processInstanceKey).withType("type2").complete();
+
+    assertThat(
+            RecordingExporter.records()
+                .skipUntil(
+                    r ->
+                        r.getValue() instanceof ProcessInstanceRecord
+                            && ((ProcessInstanceRecord) r.getValue()).getElementId().equals("task1")
+                            && r.getIntent() == ProcessInstanceIntent.ELEMENT_COMPLETED)
+                .processInstanceRecords()
```
korthout, line 264: 🔧 Can be simplified
```suggestion
            RecordingExporter.processInstanceRecords()
                .skipUntil(
                    r ->
                        r.getIntent() == ProcessInstanceIntent.ELEMENT_COMPLETED
                            && r.getValue().getElementId().equals("task1"))
```

korthout (changes requested):
Nice work @berkaycanbc 👏 

❌ I have a few more change requests and some suggestions.

🚧 We're getting close though

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -307,6 +318,30 @@ private void tryMigrateElementInstance(
...
+            final var sequenceFlowRecord = new ProcessInstanceRecord();
+            sequenceFlowRecord.wrap(elementInstanceRecord);
+            sequenceFlowRecord
+                .setElementId(sequenceFlow.getId())
+                .setBpmnElementType(sequenceFlow.getElementType())
+                .setBpmnEventType(sequenceFlow.getEventType())
+                .setFlowScopeKey(elementInstance.getKey());
+
+            stateWriter.appendFollowUpEvent(
+                keyGenerator.nextKey(), ProcessInstanceIntent.ELEMENT_MIGRATED, sequenceFlowRecord);
```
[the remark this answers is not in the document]
korthout: >We do not use the key in the applier

Correct

>we do not store the sequence flow with key, right?

Correct

To ensure the exported records reflect what happens. The key of a record is how it refers to a unique entity. As there has previously been a `PROCESS_INSTANCE:SEQUENCE_FLOW_TAKEN` event for the sequence flow instance with a specific key, we should actually write the `PROCESS_INSTANCE:ELEMENT_MIGRATED` event for the same sequence flow instance by writing it with the same key.

However, that would require us to persist the key of a sequence flow instance for some time (e.g. until the flow scope instance completes/terminates). This adds more complexity for a small gain. I think we can move this to out-of-scope by opening an issue for it.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
```diff
@@ -852,7 +854,18 @@ public static void requireNoConcurrentCommand(
     // or
     // An active sequence flow indicates a concurrent command. It is created when taking a
     // sequence flow and writing an ACTIVATE command for the next element.
-    if (eventTrigger != null || elementInstance.getActiveSequenceFlows() > 0) {
+    // We allow migrating joining parallel gateway with at least one incoming sequence flow is taken
+    final long activeFlowsToNonGateway =
+        elementInstance.getActiveSequenceFlowIds().stream()
```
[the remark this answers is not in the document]
berkaycanbc: This one should be resolved with the safety check added while retrieving active sequence flows.

berkaycanbc:
@korthout thanks for the review! 🚀 I sent updates and added some replies, please take a look. 👀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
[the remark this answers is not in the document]
berkaycanbc: Okay, got it now. Created the issue https://github.com/camunda/camunda/issues/26212 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -320,6 +320,8 @@ private void tryMigrateElementInstance(
 
     final List<DirectBuffer> activeSequenceFlowIds = elementInstance.getActiveSequenceFlowIds();
     if (!activeSequenceFlowIds.isEmpty()) {
+      checkSequenceFlowIdsIntegrity(elementInstance, processInstanceKey);
+
```
korthout, line 324: ❌ I had another thought and concluded that we cannot use the `activeSequenceFlowIds` to determine which sequence flow instances should be migrated, because by definition it does not support process instances created before 8.6.

💡 Instead, we can use your newly created method `ElementInstanceState.visitTakenSequenceFlows` to visit them. This provides all the needed info reliably as it is build around the `numberOfTakenSequenceFlowsColumnFamily` which has been the state tracking this for the joining parallel gateway since "forever". This would also mean we don't need a safety check

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -491,6 +493,19 @@ private static Set<ExecutableSequenceFlow> getSequenceFlowsToMigrate(
         .collect(Collectors.toSet());
   }
 
+  private static void checkSequenceFlowIdsIntegrity(
+      final ElementInstance elementInstance, final long processInstanceKey) {
+    if (elementInstance.getActiveSequenceFlows() != elementInstance.getActiveSequenceFlowsCount()) {
+      throw new SafetyCheckFailedException(
```
korthout, line 497: ~❌ The safety check exception leads to a logged error (which we'll see in Google Error Reporting if it occurs on SaaS). However, this is an expected case, not an error. Let's handle it the same as the concurrent case.~

EDIT: no longer needed

berkaycanbc:
@korthout I sent changes based on your last comment, please take a look. 👀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
berkaycanbc: @korthout Oh, I missed that as well. Alright, we had to determine active sequence flows using `numberOfTakenSequenceFlowsColumnFamily` data. I updated the code accordingly.

Note: `numberOfTakenSequenceFlowsColumnFamily` is only used for parallel and inclusive gateway. However, in precondition for concurrent command we need to make sure there are no active sequence flows to elements except parallel and inclusive gateway. Therefore, the check should stay the same in the concurrency precondition.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -307,6 +318,31 @@ private void tryMigrateElementInstance(
...
+
+      sequenceFlows.forEach(
+          sequenceFlow -> {
+            final var sequenceFlowRecord = new ProcessInstanceRecord();
+            sequenceFlowRecord.copyFrom(elementInstanceRecord);
+            sequenceFlowRecord
+                .setElementId(sequenceFlow.getId())
+                .setBpmnElementType(sequenceFlow.getElementType())
+                .setBpmnEventType(sequenceFlow.getEventType())
+                .setFlowScopeKey(elementInstance.getKey());
```
korthout, line 339: 💭 There may be a need to adjust the tree paths on migration, but we also don't take care of this for the active element instance yet.

🔧 We should probably reset the tree paths until we implement something correctly for it, to avoid filling the record with incorrect data

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
korthout, line 859: ❌ We still cannot always trust the `.getActiveSequenceFlowIds` value.

Instead we should do something like this:

```
if (eventTrigger) {
  reject
} else if (elementInstance.activeSequenceFlows > 0) {
  // check if there are active sequence flows to a gateway
  // and whether there are fewer then the generally active sequence flows in this element instance
  if (state.getNumberOfTakenSequenceFlows() < elementInstance.activeSequenceFlows) {
    // there are sequence flows active that don't flow a parallel/inclusive gateway
    reject
  }
}
```

korthout (changes requested):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -307,6 +318,31 @@ private void tryMigrateElementInstance(
...
 
+    final int numberOfTakenSequenceFlows =
+        elementInstanceState.getNumberOfTakenSequenceFlows(elementInstance.getKey());
+    if (numberOfTakenSequenceFlows > 0) {
+      final Set<ExecutableSequenceFlow> sequenceFlows =
+          getSequenceFlowsToMigrate(
+              sourceProcessDefinition,
+              targetProcessDefinition,
+              sourceElementIdToTargetElementId,
+              elementInstance);
```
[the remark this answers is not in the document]
berkaycanbc: Oh yeah, it will simply be empty and we will do the db call anyways. 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
berkaycanbc: @korthout I am a bit off from the tree path topic. Could you please clarify what needs to be done here?

korthout: For now let's add this:
```suggestion
            sequenceFlowRecord
                .setElementId(sequenceFlow.getId())
                .setBpmnElementType(sequenceFlow.getElementType())
                .setBpmnEventType(sequenceFlow.getEventType())
                .setFlowScopeKey(elementInstance.getKey())
                .resetElementInstancePath()
                .resetCallingElementPath()
                .resetProcessDefinitionPath();
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationPreconditions.java
berkaycanbc: @korthout how do we know if the count from `elementInstance.activeSequenceFlows` is complete? E.g. some elements are activated after 8.6 and stored active sequence flow count but it can still be less than the `state.getNumberOfTakenSequenceFlows()` which is wrong. I think, retrieving `elementInstance.getActiveSequenceFlowIds()` and retaining the ones that are not part of `elementInstanceState.getTakenSequenceFlows()` is the more proper way, right?

korthout (approved):
Great work @berkaycanbc 👏 

LGTM 👍

PR #25999, opened 2024-12-12
Add UserService to gRPC gateway

### LINKED ISSUE #25982: Add User service to gRPC gateway
In order to retrieve user information (userKey) in the gRPC gateway we need to enable it to search uses from ES/OS by the username. Similar to what we do in the REST gateway.

nicpuppa:
## Description

This PR adds the UserServices to the gRPC gateway (Embedded and Standalone)

## Related issues

closes #25982

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
+++ b/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
@@ -11,6 +11,7 @@ import io.atomix.cluster.AtomixCluster;
 import io.camunda.application.commons.configuration.BrokerBasedConfiguration;
 import io.camunda.identity.sdk.IdentityConfiguration;
 import io.camunda.security.configuration.SecurityConfiguration;
+import io.camunda.service.UserServices;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
 import io.camunda.zeebe.broker.exporter.repo.ExporterDescriptor;
 import io.camunda.zeebe.broker.exporter.repo.ExporterRepository;
@@ -54,6 +55,7 @@ public class BrokerModuleConfiguration implements CloseableSilently {
   private final BrokerShutdownHelper shutdownHelper;
   private final MeterRegistry meterRegistry;
   private final SecurityConfiguration securityConfiguration;
+  private final UserServices userServices;
 
   private Broker broker;
 
@@ -67,7 +69,8 @@ public class BrokerModuleConfiguration implements CloseableSilently {
       final BrokerClient brokerClient,
       final BrokerShutdownHelper shutdownHelper,
       final PrometheusMeterRegistry meterRegistry,
-      final SecurityConfiguration securityConfiguration) {
+      final SecurityConfiguration securityConfiguration,
+      final UserServices userServices) {
     this.configuration = configuration;
     this.identityConfiguration = identityConfiguration;
     this.springBrokerBridge = springBrokerBridge;
@@ -77,6 +80,7 @@ public class BrokerModuleConfiguration implements CloseableSilently {
     this.shutdownHelper = shutdownHelper;
     this.meterRegistry = meterRegistry;
     this.securityConfiguration = securityConfiguration;
+    this.userServices = userServices;
   }
 
   @Bean
--- a/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
+++ b/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
@@ -105,7 +105,8 @@ public class BrokerModuleConfiguration implements CloseableSilently {
             cluster,
             brokerClient,
             meterRegistry,
-            securityConfiguration);
+            securityConfiguration,
+            userServices);
     springBrokerBridge.registerShutdownHelper(
         errorCode -> shutdownHelper.initiateShutdown(errorCode));
     broker =
--- a/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
+++ b/dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
@@ -70,7 +70,8 @@ public class BrokerModuleConfiguration implements CloseableSilently {
       final BrokerShutdownHelper shutdownHelper,
       final PrometheusMeterRegistry meterRegistry,
       final SecurityConfiguration securityConfiguration,
-      final UserServices userServices) {
+      // The UserServices class is not available if you want to start-up the Standalone Broker
+      @Autowired(required = false) final UserServices userServices) {
     this.configuration = configuration;
     this.identityConfiguration = identityConfiguration;
     this.springBrokerBridge = springBrokerBridge;
--- a/dist/src/main/java/io/camunda/zeebe/gateway/GatewayModuleConfiguration.java
+++ b/dist/src/main/java/io/camunda/zeebe/gateway/GatewayModuleConfiguration.java
@@ -106,7 +106,8 @@ public class GatewayModuleConfiguration implements CloseableSilently {
             securityConfiguration,
             brokerClient,
             actorScheduler,
-            jobStreamClient.streamer());
+            jobStreamClient.streamer(),
+            userServices);
     springGatewayBridge.registerGatewayStatusSupplier(gateway::getStatus);
     springGatewayBridge.registerClusterStateSupplier(
         () ->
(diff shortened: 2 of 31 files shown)


nicpuppa:
💭 I'm not sure how to properly test this, if you have ideas please let me know

nicpuppa:
@tmetzke and @houssain-barouni , adding you are reviewers because we discussed the topic together. 

I don't want a review from everyone, but I'll appreciate feedbacks 👍

dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
```diff
@@ -70,7 +71,8 @@ public BrokerModuleConfiguration(
       final BrokerShutdownHelper shutdownHelper,
       final PrometheusMeterRegistry meterRegistry,
       final SecurityConfiguration securityConfiguration,
-      final UserServices userServices) {
+      // The UserServices class is not available if you want to start-up the Standalone Broker
+      final Optional<UserServices> userServices) {
```
remcowesterhoud, line 75: 🔧 ❓ Could we solve this with an overloaded constructor instead of an optional parameter?

remcowesterhoud (approved):
Thanks for splitting this up nicely in commits, very easy to revew 👍 Just one small thing, otherwise all good 🚀

tmetzke (commented):
As discussed offline last week, I can now disable the REST API and no user service bean is created since this resides in the `CamundaServicesConfiguration` which relies on the REST API being enabled. I think this could become an issue.

nicpuppa:
> As discussed offline last week, I can now disable the REST API and no user service bean is created since this resides in the `CamundaServicesConfiguration` which relies on the REST API being enabled. I think this could become an issue.

I can make it Optional also in the Gateway, if you disable the REST API also the authentication needs to be disabled

tmetzke:
> > As discussed offline last week, I can now disable the REST API and no user service bean is created since this resides in the `CamundaServicesConfiguration` which relies on the REST API being enabled. I think this could become an issue.
> 
> I can make it Optional also in the Gateway, if you disable the REST API also the authentication needs to be disabled

Maybe I don't understand yet why this is added to the gRPC gateway in the first place. Don't we want to use this for the gRPC communication now as well? If I disable the REST API, the user service is still needed in the gRPC gateway, isn't it?

remcowesterhoud:
> > > As discussed offline last week, I can now disable the REST API and no user service bean is created since this resides in the `CamundaServicesConfiguration` which relies on the REST API being enabled. I think this could become an issue.
> > 
> > 
> > I can make it Optional also in the Gateway, if you disable the REST API also the authentication needs to be disabled
> 
> Maybe I don't understand yet why this is added to the gRPC gateway in the first place. Don't we want to use this for the gRPC communication now as well? If I disable the REST API, the user service is still needed in the gRPC gateway, isn't it?

@tmetzke We need to fetch the user from ES and verify the password when gRPC is enabled as well. Before this was handled by Identity, but alas Identity is no more.

Realistically, I don't think there's any scenario where authentication would be enabled and REST disabled? All the idenitty related endpoints don't exist in gRPC. How would anyone work with gRPC + auth on the latest version without REST?

tmetzke:
> > > > As discussed offline last week, I can now disable the REST API and no user service bean is created since this resides in the `CamundaServicesConfiguration` which relies on the REST API being enabled. I think this could become an issue.
> > > 
> > > 
> > > I can make it Optional also in the Gateway, if you disable the REST API also the authentication needs to be disabled
> > 
> > 
> > Maybe I don't understand yet why this is added to the gRPC gateway in the first place. Don't we want to use this for the gRPC communication now as well? If I disable the REST API, the user service is still needed in the gRPC gateway, isn't it?
> 
> @tmetzke We need to fetch the user from ES and verify the password when gRPC is enabled as well. Before this was handled by Identity, but alas Identity is no more.
> 
> Realistically, I don't think there's any scenario where authentication would be enabled and REST disabled? All the idenitty related endpoints don't exist in gRPC. How would anyone work with gRPC + auth on the latest version without REST?

I could simply run different applications with different purposes, right? So I could have
* A dedicated application running a REST API, available only internally in my company network
* A dedicated application using gRPC, accessible by a wider audience, not allowing REST API interaction

nicpuppa:
I think it make sense to make the bean optional, mainly because if I want to run the standalone gRPC gateway without authentication and with the REST API disabled the gateway will fail to startup because the bean is not there. Am I right ? @tmetzke @remcowesterhoud

remcowesterhoud:
> I think it make sense to make the bean optional, mainly because if I want to run the standalone gRPC gateway without authentication and with the REST API disabled the gateway will fail to startup because the bean is not there. Am I right ? @tmetzke @remcowesterhoud

I think what @tmetzke is arguing is that it should be possible to run the gRPC gateway with authentication enabled, and without REST enabled.

tmetzke:
> > I think it make sense to make the bean optional, mainly because if I want to run the standalone gRPC gateway without authentication and with the REST API disabled the gateway will fail to startup because the bean is not there. Am I right ? @tmetzke @remcowesterhoud
> 
> I think what @tmetzke is arguing is that it should be possible to run the gRPC gateway with authentication enabled, and without REST enabled.

Even worse so, I think you CAN already run this but it will fail because there will be no user service bean for the gRPC gateway in that case 🙈

remcowesterhoud:
> >  > ...
> > 
> > @tmetzke We need to fetch the user from ES and verify the password when gRPC is enabled as well. Before this was handled by Identity, but alas Identity is no more.
> > Realistically, I don't think there's any scenario where authentication would be enabled and REST disabled? All the idenitty related endpoints don't exist in gRPC. How would anyone work with gRPC + auth on the latest version without REST?
> 
> I could simply run different applications with different purposes, right? So I could have
> 
> * A dedicated application running a REST API, available only internally in my company network
> * A dedicated application using gRPC, accessible by a wider audience, not allowing REST API interaction

But this is about the gateway so that would mean multiple instances of Zeebe, both with different configurations, right? I don't think we have support for that 😅

nicpuppa:
> I think what @tmetzke is arguing is that it should be possible to run the gRPC gateway with authentication enabled, and without REST enabled.

I'd argue back that the REST api should be the default now on, and only enable/disable the gRPC gateway 🤣

tmetzke:
> > I could simply run different applications with different purposes, right? So I could have
> > 
> > * A dedicated application running a REST API, available only internally in my company network
> > * A dedicated application using gRPC, accessible by a wider audience, not allowing REST API interaction
> 
> But this is about the gateway so that would mean multiple instances of Zeebe, both with different configurations, right? I don't think we have support for that 😅

What, we don't? 😱 🤣

tmetzke:
> > > I could simply run different applications with different purposes, right? So I could have
> > > 
> > > * A dedicated application running a REST API, available only internally in my company network
> > > * A dedicated application using gRPC, accessible by a wider audience, not allowing REST API interaction
> > 
> > 
> > But this is about the gateway so that would mean multiple instances of Zeebe, both with different configurations, right? I don't think we have support for that 😅
> 
> What, we don't? 😱 🤣

It is just two Spring apps using the same ES instance in the end.

remcowesterhoud:
Another argument why I believe this is fine. [We don't support basic auth in gRPC at the moment](https://camunda.slack.com/archives/C06UYJMMETZ/p1733499273857869?thread_ts=1733497792.542889&cid=C06UYJMMETZ). So not supporting it if REST is enabled would not be a breaking change.

nicpuppa:
> Another argument why I believe this is fine. [We don't support basic auth in gRPC at the moment](https://camunda.slack.com/archives/C06UYJMMETZ/p1733499273857869?thread_ts=1733497792.542889&cid=C06UYJMMETZ). So not supporting it if REST is disabled would not be a breaking change.

We need to [support](https://github.com/camunda/camunda/issues/25795) BASIC AUTH in grpc as well

dist/src/main/java/io/camunda/zeebe/broker/BrokerModuleConfiguration.java
nicpuppa: We could, but is it more clear then using Optional ?

We have another option:

```
@Autowired
public BrokerModuleConfiguration(
    final BrokerBasedConfiguration configuration,
    final IdentityConfiguration identityConfiguration,
    final SpringBrokerBridge springBrokerBridge,
    final ActorScheduler actorScheduler,
    final AtomixCluster cluster,
    final BrokerClient brokerClient,
    final BrokerShutdownHelper shutdownHelper,
    final PrometheusMeterRegistry meterRegistry,
    final SecurityConfiguration securityConfiguration,
    @Autowired(required = false) final UserServices userServices) {
  this.configuration = configuration;
  this.identityConfiguration = identityConfiguration;
  this.springBrokerBridge = springBrokerBridge;
  this.actorScheduler = actorScheduler;
  this.cluster = cluster;
  this.brokerClient = brokerClient;
  this.shutdownHelper = shutdownHelper;
  this.meterRegistry = meterRegistry;
  this.securityConfiguration = securityConfiguration;
  this.userServices = userServices;
}
```

I don't have a strong preference

remcowesterhoud (approved):

PR #26016, opened 2024-12-12
feat: add cluster purge endpoint to cluster API #25961

### LINKED ISSUE #25961: Add a new cluster purge endpoint to the cluster API
## Description

Extend the existing OpenAPI spec under `cluster-api.yml` in the dist module, and add a new `POST /purge` endpoint.

The new endpoint should take a single optional parameter, `dryRun`, which should be a boolean value.

It should return 202 on success, or the _same_ error codes as the `PATCH /` endpoint, as the errors would be similar.

A basic implementation is added under the `ClusterEndpoint` controller, which sends a new `PurgeRequest` to a specific topic to the coordinator node.

Please add or extend tests as for the other endpoints; note that since the coordinator will not handle anything yet, we cannot test the success path, but that's OK. Focus on adding tests that we will keep in the future (e.g. 502 is returned if there is no coordinator).

### Acceptance criteria

> [!Note]
> As usual, you can split the following into multiple PRs, as long as the last one is the one closing this issue.

- [x] A new `POST /purge` endpoint is added to the `cluster-api.yml` API spec.
  - [x] It returns the same responses and status codes as the `PATCH /` endpoint.
  - [x] It accepts a single `dryRun` parameter, like all the other endpoints.
- [x] A new method is added to the controller `ClusterEndpoint`, mapping the `POST /purge` endpoint from the OpenAPI spec.
  - [x] The endpoint sends a purge request to the coordinator node under the `cluster-purge` topic.
  - [x] A new `PurgeRequest` is added as part of our protocol.
  - [x] There is a new encoding method for the `PurgeRequest`.
  - [x] The `PurgeRequest` only has the `dryRun` parameter.
  - [x] The `PurgeRequest` is sent over a specific topic (e.g. `cluster-purge`) via the communication service.

### LINKED ISSUE #25962: Allow 0 replicas when leaving partitions during purge
## Description

There is currently a constraint when scaling down the cluster, preventing us from having less than 1 replica per partition. For purging, we want to scale down to 0 - as such, we need to parameterize the `PartitionLeaveApplier` to allow this behavior when happening during a purge operation.

## Acceptance criteria

- [x] If part of a purge operation, `PartitionLeaveApplier` can allow 0 replicas for a partition.
- [x] If **not** part of a purge operation, `PartitionLeaveApplier` **cannot** allow less than 1 replica per partition.

### LINKED ISSUE #25963: Do not assume partition 1 is always present
## Description

In the `PartitionBootstrapApplier`, we always assume that partition 1 is available, and use its `DynamicPartitionConfig` to populate other partitions on bootstrap.

As part of the purge operation, we will now allow scaling a cluster down to 0 partition. This would make bootstrapping fail, so we need to have a fallback when partition 1 does not exist.

To properly fix this, we need to understand if we should use `DynamicPartitionConfig#uninitialized` or `DynamicPartitionConfig#init` as the fallback value, and use that if no partitions are available.

## Acceptance criteria

- [x] We _know_ which `DynamicPartitionConfig` to use as fallback
- [x] I can bootstrap a new partition even if there are no known partitions in the cluster.

### LINKED ISSUE #25964: Transform a purge request into scale down and restore cluster change operations
## Description

The goal here is to implement a handler for the `PurgeRequest` on the coordinator node (in `ClusterConfigurationRequestServer`), which transforms the request into a series of planned change operations that results in the data _in Zeebe only_ to be deleted.

To do so, we need to ensure we register a new handler in `ClusterConfigurationRequestServer` and implement a decoder for the `PurgeRequest` added in #25961.

To handle the request, we will implement a new `purge` method in the `ClusterConfigurationManagementApi`. The implementation will be largely delegated to a new `PurgeTransformer`.

This class will take as input the current topology, and produce a sequence of operations which is essentially every broker leaving all their partitions, followed by restoring the current topology as it was.

> [!Warning]
> When bootstrapping partitions, they _must_ be bootstrapped in order of ID ascending. Meaning partition 1 first, then 2, then 3, etc., contiguously.

Broken down, it would be:

- For every broker `B` and partition `P`, there will be a `PARTITION_LEAVE(B,P)` command.
- For every partition `P` _ordered by partition ID ascending_, we will pick a single primary broker `B` out of the brokers in this partition, and produce a `PARTITION_BOOTSTRAP(B,P,<priority>)`, where `<priority>` is the priority that node for this partition as existing in the current topology.
- For every partition `P` (order does not matter here), and for all non-primary brokers `B` out of the brokers in this partition, we will produce a `PARTITION_JOIN(B,P,<priority>)`, where `<priority>` is the priority that node for this partition as existing in the current topology.

This means the returned expected topology should be the _exact same_ as before.

## Acceptance criteria

- [x] The coordinator node has a new handler for the `cluster-purge` topic to handle incoming `PurgeRequest`.
- [x] The handler delegates logic to a new `PurgeTransformer`.
- [x] The `PurgerTransformer` produces a sequence of operations which consists of every broker leaving all of its partitions, and then the topology being restored in the right order.
- [x] For every partition, there is a single `PartitionBootstrapOperation`.
  - [x] These bootstrap operations are for a broker that was already part of this partition.
  - [x] The priority is the same as it was previously.
  - [x] The bootstrap operations are all after the leave operations.
  - [x] The bootstrap operations are ordered by partition ID ascending amongst themselves.
- [x] For every partition, and all other non-primary brokers part of it, there is a single `PartitionJoinOperation` with the same priority as before.
  - [x] All join operations are after the bootstrap operations.

srsp:
## Description

Extends the cluster API with a purge endpoint. Current implementation does not actually purge, but returns a change plan, although that change plan is missing at least the `DELETE_HISTORY` change. 

## Checklist

- [ ] for CI changes:

## Related issues

closes #25961 
closes #25962
closes #25963 
closes #25964

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
@@ -26,6 +26,7 @@ import io.camunda.zeebe.management.cluster.Error;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.endpoint.web.annotation.RestControllerEndpoint;
@@ -363,12 +364,15 @@ public class ClusterEndpoint {
   }
 
   @PostMapping(path = "/purge", produces = "application/json")
-  public ResponseEntity<?> purge(@RequestParam(defaultValue = "false") final boolean dryRun) {
+  public CompletableFuture<ResponseEntity<?>> purge(
+      @RequestParam(defaultValue = "false") final boolean dryRun) {
     try {
-      return ClusterApiUtils.mapOperationResponse(
-          requestSender.purge(new PurgeRequest(dryRun)).join());
+      return requestSender
+          .purge(new PurgeRequest(dryRun))
+          .thenApply(ClusterApiUtils::mapOperationResponse)
+          .exceptionally(ClusterApiUtils::mapError);
     } catch (final Exception error) {
-      return ClusterApiUtils.mapError(error);
+      return CompletableFuture.completedFuture(ClusterApiUtils.mapError(error));
     }
   }
 
--- a/dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
+++ b/dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
@@ -16,6 +16,7 @@ import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest
 import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.ForceRemoveBrokersRequest;
 import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.JoinPartitionRequest;
 import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.LeavePartitionRequest;
+import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.PurgeRequest;
 import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequest.RemoveMembersRequest;
 import io.camunda.zeebe.dynamic.config.api.ClusterConfigurationManagementRequestSender;
 import io.camunda.zeebe.management.cluster.ClusterConfigPatchRequest;
@@ -359,6 +360,16 @@ public class ClusterEndpoint {
           };
       case changes -> new ResponseEntity<>(HttpStatusCode.valueOf(404));
     };
+  }
+
+  @PostMapping(path = "/purge", produces = "application/json")
+  public ResponseEntity<?> purge(@RequestParam(defaultValue = "false") final boolean dryRun) {
+    try {
+      return ClusterApiUtils.mapOperationResponse(
+          requestSender.purge(new PurgeRequest(dryRun)).join());
+    } catch (final Exception error) {
+      return ClusterApiUtils.mapError(error);
+    }
   }
 
   public record PartitionAddRequest(int priority) {}
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationRequestTopics.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationRequestTopics.java
@@ -22,7 +22,7 @@ public enum ClusterConfigurationRequestTopics {
 
   SCALE_CLUSTER("topology-cluster-scale"),
   PATCH_CLUSTER("topology-cluster-patch"),
-  PURGE("topology-purge"),
+  PURGE("topology-cluster-purge"),
   FORCE_REMOVE_BROKERS("topology-broker-force-remove");
 
   private final String topic;
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationRequestTopics.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationRequestTopics.java
@@ -22,6 +22,7 @@ public enum ClusterConfigurationRequestTopics {
 
   SCALE_CLUSTER("topology-cluster-scale"),
   PATCH_CLUSTER("topology-cluster-patch"),
+  PURGE("topology-purge"),
   FORCE_REMOVE_BROKERS("topology-broker-force-remove");
 
   private final String topic;
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PartitionReassignRequestTransformer.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PartitionReassignRequestTransformer.java
@@ -19,6 +19,7 @@ import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionJoinOperation;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionLeaveOperation;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionReconfigurePriorityOperation;
+import io.camunda.zeebe.dynamic.config.state.ExportersConfig;
 import io.camunda.zeebe.dynamic.config.util.ConfigurationUtil;
 import io.camunda.zeebe.dynamic.config.util.RoundRobinPartitionDistributor;
 import io.camunda.zeebe.util.Either;
@@ -139,7 +140,8 @@ public class PartitionReassignRequestTransformer implements ConfigurationChangeR
     final var primary =
         newMetadata.getPrimary().orElse(newMetadata.members().stream().findAny().orElseThrow());
     operations.add(
-        new PartitionBootstrapOperation(primary, partitionId, newMetadata.getPriority(primary)));
+        new PartitionBootstrapOperation(
+            primary, partitionId, newMetadata.getPriority(primary), ExportersConfig.empty()));
 
     // Join each remaining members to the partition
     for (final MemberId member : newMetadata.members()) {
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PartitionReassignRequestTransformer.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PartitionReassignRequestTransformer.java
@@ -19,7 +19,6 @@ import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionJoinOperation;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionLeaveOperation;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionReconfigurePriorityOperation;
-import io.camunda.zeebe.dynamic.config.state.ExportersConfig;
 import io.camunda.zeebe.dynamic.config.util.ConfigurationUtil;
 import io.camunda.zeebe.dynamic.config.util.RoundRobinPartitionDistributor;
 import io.camunda.zeebe.util.Either;
@@ -140,8 +139,7 @@ public class PartitionReassignRequestTransformer implements ConfigurationChangeR
     final var primary =
         newMetadata.getPrimary().orElse(newMetadata.members().stream().findAny().orElseThrow());
     operations.add(
-        new PartitionBootstrapOperation(
-            primary, partitionId, newMetadata.getPriority(primary), ExportersConfig.empty()));
+        new PartitionBootstrapOperation(primary, partitionId, newMetadata.getPriority(primary)));
 
     // Join each remaining members to the partition
     for (final MemberId member : newMetadata.members()) {
@@ -170,7 +168,7 @@ public class PartitionReassignRequestTransformer implements ConfigurationChangeR
     final var membersToLeave =
         oldMetadata.members().stream()
             .filter(member -> !newMetadata.members().contains(member))
-            .map(oldMember -> new PartitionLeaveOperation(oldMember, partitionId))
+            .map(oldMember -> new PartitionLeaveOperation(oldMember, partitionId, false))
             .toList();
     final var membersToChangePriority =
         oldMetadata.members().stream()
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
@@ -11,6 +11,7 @@ import io.atomix.cluster.MemberId;
 import io.camunda.zeebe.dynamic.config.changes.ConfigurationChangeCoordinator.ConfigurationChangeRequest;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfiguration;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation;
+import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.DeleteHistoryOperation;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionBootstrapOperation;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionJoinOperation;
 import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionLeaveOperation;
@@ -54,7 +55,9 @@ public final class PurgeRequestTransformer implements ConfigurationChangeRequest
       }
     }
 
-    // TODO Delete history (only coordinator node)
+    clusterConfiguration
+        .members()
+        .forEach((key, value) -> operations.add(new DeleteHistoryOperation(key)));
 
     primaries.forEach(
         (partitionId, bootstrapOperation) -> {
--- a/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
+++ b/zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
@@ -31,7 +31,7 @@ public final class PurgeRequestTransformer implements ConfigurationChangeRequest
       final ClusterConfiguration clusterConfiguration) {
 
     final SortedMap<Integer, PartitionBootstrapOperation> primaries =
-        findBootstrapMembers(clusterConfiguration.members());
+        createBootstrapOperations(clusterConfiguration.members());
 
     final Map<Integer, List<PartitionJoinOperation>> followers =
         new TreeMap<>(Comparator.naturalOrder());
@@ -63,6 +63,7 @@ public final class PurgeRequestTransformer implements ConfigurationChangeRequest
         (partitionId, bootstrapOperation) -> {
           operations.add(bootstrapOperation);
         });
+
     followers.forEach(
         (partitionId, joinOperations) -> {
           operations.addAll(joinOperations);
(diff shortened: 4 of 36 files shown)


zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationManagementApiTest.java
```diff
@@ -163,6 +164,24 @@ void shouldRemoveMembers() {
...
+        initialTopology
+            .addMember(id1, MemberState.initializeAsActive(Map.of()))
+            .addMember(id2, MemberState.initializeAsActive(Map.of())));
+    final var request = new PurgeRequest(false);
+
+    // when
+    final var changeStatus = clientApi.purge(request).join().get();
+
+    // then
+    // TODO Why are the planned changes empty? Because the RecordingChangeCoordinator is used?
```
srsp, line 180: @npepinpe Do you have an idea here? I see planned changes when starting the StandaloneBroker and curling the API, but I do not see them here. Only difference seems to be the coordinator implementation.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Map;
+
+public final class PurgeRequestTransformer implements ConfigurationChangeRequest {
+
+  @Override
+  public Either<Exception, List<ClusterConfigurationChangeOperation>> operations(
+      final ClusterConfiguration clusterConfiguration) {
+
+    // TODO Maybe select new primary here to use down there
+    final Map<Integer, PartitionBootstrapOperation> primaries = new HashMap<>();
```
npepinpe, line 29: :x: When iterating a map, the order is non-deterministic. So we should use a `SortedMap` here.

```java
final SortedMap<Integer, PartitionBootstrapOperation> primaries = new TreeSet(Comparators.naturalOrder());
```

That way, the `forEach` will be properly ordered.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationRequestTopics.java
```diff
@@ -22,6 +22,7 @@ public enum ClusterConfigurationRequestTopics {
 
   SCALE_CLUSTER("topology-cluster-scale"),
   PATCH_CLUSTER("topology-cluster-patch"),
+  PURGE("topology-purge"),
```
npepinpe, line 25: :wrench: I don't really care about the topology in there (I guess we can remain consistent), but would like at least the `cluster`, to highlight this is specifically about purging the complete cluster.

dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
```diff
@@ -361,6 +362,16 @@ public ResponseEntity<?> removeSubResource(
     };
   }
 
+  @PostMapping(path = "/purge", produces = "application/json")
+  public ResponseEntity<?> purge(@RequestParam(defaultValue = "false") final boolean dryRun) {
+    try {
+      return ClusterApiUtils.mapOperationResponse(
+          requestSender.purge(new PurgeRequest(dryRun)).join());
```
npepinpe, line 369: :upside_down_face: I know we do this every where in this controller, but it always makes me twitch to see a `join` on a future :smile: 

I think the following should just work OOTB - if it doesn't, then ignore this comment. But otherwise, let's let Spring deal with the asynchronicity:

```java
  @PostMapping(path = "/purge", produces = "application/json")
  public CompletableFuture<ResponseEntity<?>> purge(
      @RequestParam(defaultValue = "false") final boolean dryRun) {
    try {
      requestSender
          .purge(new PurgeRequest(dryRun))
          .thenApply(ClusterApiUtils::mapOperationResponse)
          .exceptionally(ClusterApiUtils::mapError);
    } catch (final Exception error) {
      return ClusterApiUtils.mapError(error);
    }
  }
```

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/api/ClusterConfigurationManagementApiTest.java
npepinpe: I'm assuming this is why we're not asserting the changes? That was one of my comment, but I guess this is why :smile: 

I don't see why, I mean other tests have `plannedChanges` and they're not empty.

npepinpe: Alright, found it - there is no partitions in the current topology, so nothing to leave/restore.

When asserting the changes, make sure to use hard coded values (like the other tests), so we can detect granular changes in the future.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionBootstrapOperation;
+import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionJoinOperation;
+import io.camunda.zeebe.dynamic.config.state.ClusterConfigurationChangeOperation.PartitionChangeOperation.PartitionLeaveOperation;
+import io.camunda.zeebe.util.Either;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class PurgeRequestTransformer implements ConfigurationChangeRequest {
```
npepinpe, line 22: :upside_down_face:

```suggestion
public final class PurgeTransformer implements ConfigurationChangeRequest {
```

Seems like the others don't have `Request` in the name :shrug:

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/management/ClusterEndpointIT.java
```diff
@@ -75,6 +75,32 @@ void shouldRequestPartitionLeave() {
     }
   }
 
+  @Test
+  void shouldRequestClusterPurge() {
+    final ClusterActuator actuator;
+    try (final var cluster = createCluster(1)) {
```
npepinpe, line 81: :wrench: Let's increase the replication factor - it's a cheap way to get more test coverage imo.

npepinpe (changes requested):
:rocket: Fast and clean :sunglasses: 

Some notes:

- :x: Please add a test case with a replication factor higher than one. I want to make sure we test that: 1) partitions are bootstrapped in order, 2) all partitions are left first, and 3) partitions are rejoined by all replicas, and bootstrapped only once. With a replication factor of 1, it's hard to guarantee. Feel free to split those into multiple tests if that makes more sense. For example, I could imagine all these properties are tested in a `PurgeRequestTransformerTest` unit test, much like the other API tests.

npepinpe:
FYI, this PR closes:

- https://github.com/camunda/camunda/issues/25961
- https://github.com/camunda/camunda/issues/25962
- https://github.com/camunda/camunda/issues/25963
- https://github.com/camunda/camunda/issues/25964

Please make sure all acceptance criteria are met, or split this into multiple PRs if it becomes too big.

dist/src/main/java/io/camunda/zeebe/shared/management/ClusterEndpoint.java
srsp: Spring can handle it, mvn says it's fine, but my IDE complains, but I'll just ignore the whining:
image

Latest Intellij, did restart, maybe it will vanish eventually.

npepinpe: :grinning: The `cannot convert T to T` message just makes me think of

image

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
srsp: 2 don't, 8 have it, if I found all implementations. I just went with the majority (which might not always be right) 😀

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/management/ClusterEndpointIT.java
srsp: Actually I can't get this test to run properly in this PR, because of the leave-partition-failcheck, which is changed in a coming PR. So I rewrote this test to expect this exception, and I will provide the real test in the coming PR. So in this version we can at least ensure, that the endpoint is callable and not throwing random exceptions.

npepinpe: Then let's add a TODO and come back to it later.

srsp:
This PR only closes these issues:
 * [Add a new cluster purge endpoint to the cluster API #25961](https://github.com/camunda/camunda/issues/25961)
 * [Transform a purge request into scale down and restore cluster change operations #25964](https://github.com/camunda/camunda/issues/25964)

These issues are covered with the next one, so the size of this PR does not get out of hand:
 * [Allow 0 replicas when leaving partitions during purge #25962](https://github.com/camunda/camunda/issues/25962)
 * [Do not assume partition 1 is always present #25963](https://github.com/camunda/camunda/issues/25963)

npepinpe:
Then we may want to consider merging https://github.com/camunda/camunda/issues/25962 and https://github.com/camunda/camunda/issues/25963 first, to avoid a weird interim state here.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/management/ClusterEndpointIT.java
```diff
@@ -75,6 +75,34 @@ void shouldRequestPartitionLeave() {
...
+      // then
+      assertThat(response.getPlannedChanges().stream().map(Operation::getOperation))
+          .containsExactlyElementsOf(
+              List.of(
+                  OperationEnum.PARTITION_LEAVE,
+                  OperationEnum.PARTITION_LEAVE,
+                  OperationEnum.PARTITION_LEAVE,
+                  OperationEnum.PARTITION_LEAVE,
+                  OperationEnum.DELETE_HISTORY,
+                  OperationEnum.DELETE_HISTORY,
```
npepinpe, line 98: :question: Why are we deleting twice? Shouldn't we only delete the history once on the coordinator node, deleting for all partitions?

npepinpe (changes requested):
:rocket:

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+          followers
+              .computeIfAbsent(partitionId, key -> new ArrayList<>())
+              .add(
+                  new PartitionJoinOperation(
+                      memberId, partitionId, partitions.getValue().priority()));
+        }
+      }
+    }
+
+    clusterConfiguration
```
npepinpe, line 58: :x: We should call `DELETE_HISTORY` only on one node, there's no reason to run it on all nodes, is there?

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/changes/DeleteHistoryApplier.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.atomix.cluster.MemberId;
+import io.camunda.zeebe.dynamic.config.changes.ConfigurationChangeAppliers.ClusterOperationApplier;
+import io.camunda.zeebe.dynamic.config.state.ClusterConfiguration;
+import io.camunda.zeebe.scheduler.future.ActorFuture;
+import io.camunda.zeebe.scheduler.future.CompletableActorFuture;
+import io.camunda.zeebe.util.Either;
+import java.util.function.UnaryOperator;
+
+final class DeleteHistoryApplier implements ClusterOperationApplier {
```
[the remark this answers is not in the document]
srsp: Yes, exactly.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Override
+  public Either<Exception, List<ClusterConfigurationChangeOperation>> operations(
+      final ClusterConfiguration clusterConfiguration) {
+
+    final SortedMap<Integer, PartitionBootstrapOperation> primaries =
+        createBootstrapOperations(clusterConfiguration.members());
+
+    final Map<Integer, List<PartitionJoinOperation>> followers =
+        new TreeMap<>(Comparator.naturalOrder());
```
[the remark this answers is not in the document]
srsp: It does probably not matter in which order the `PartitionJoinOperations` are executed, but it makes testing easier. This is probably not the best reason for keeping it ordered, but what about determinism? Should a purge always result in the same operations in the same order?

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PurgeRequestTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+    followers.forEach(
+        (partitionId, joinOperations) -> {
+          operations.addAll(joinOperations);
+        });
+
+    return Either.right(operations);
+  }
+
+  /** This method creates the BootstrapOperations for all leaders for each partition. */
+  private SortedMap<Integer, PartitionBootstrapOperation> createBootstrapOperations(
```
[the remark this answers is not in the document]
srsp: Yes, we select the primaries beforehand, because we don't know whether the current member is a primary if we have not seen the other partition members. Of course we could add it as primary, but when we encounter the next member and notice that this is actually the primary, we'd have to delete it from the primaries map and put it in the followers map and add the current as primary. That seemed more complex than figuring out the primaries beforehand.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/api/PartitionReassignRequestTransformer.java
```diff
@@ -139,7 +140,8 @@ private List<ClusterConfigurationChangeOperation> addPartition(
     final var primary =
         newMetadata.getPrimary().orElse(newMetadata.members().stream().findAny().orElseThrow());
     operations.add(
-        new PartitionBootstrapOperation(primary, partitionId, newMetadata.getPriority(primary)));
+        new PartitionBootstrapOperation(
+            primary, partitionId, newMetadata.getPriority(primary), ExportersConfig.empty()));
```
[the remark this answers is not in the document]
srsp: I don't think we changed behavior here. The `ExportersConfig.empty()` will only be used, if there is no partition at all. If there is still a partition 1, the `PartitionBootstrapOperation` will use the `ExportersConfig` of this partition. 

So maybe the naming and meaning of this parameter is not clear enough. 

We could rename it to `fallbackExportersConfig`. 

Or we could create a new Operation similar to `PartitionBootstrapOperation` that is only used to create the magical partition 1.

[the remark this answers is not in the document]
npepinpe: As discussed, we want to make sure the topology should be exactly the same as before.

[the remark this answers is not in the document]
npepinpe: Not just the `ExportersConfig`, but the complete `DynamicPartitionConfig`.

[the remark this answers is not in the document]
srsp: Like we agreed on in our sync, we will change the param of the `PartitionBootstrapOperation` to `Optional<DynamicPartitionConfig>`.

zeebe/dynamic-config/src/main/java/io/camunda/zeebe/dynamic/config/changes/PartitionBootstrapApplier.java
```diff
@@ -80,14 +97,17 @@ public Either<Exception, UnaryOperator<MemberState>> initMemberState(
...
 
-    // Let's assume Partition 1 always exists
     partitionConfig =
-        currentClusterConfiguration.members().values().stream()
-            .flatMap(m -> m.partitions().entrySet().stream().filter(p -> p.getKey() == 1))
-            .findFirst()
-            .get()
-            .getValue()
-            .config();
+        (config.orElse(
```
npepinpe, line 101: :wrench: Let's extract the fallback behavior to a method to make it clearer what we're doing. Added as a follow up in the main issue.

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/api/ClusterPurgeRequestTransformerTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+    // then
+    assertThat(result)
+        .isRight()
+        .right()
+        .satisfies(
+            operations -> {
+              assertThat(operations).hasSize(13);
+              assertThat(operations)
+                  .containsExactly(
```
npepinpe, line 100: :wrench: I think there's a assert method to check if it contains a sequence, so you don't need to specify _all_ operations, but instead you can specify just the sequence you're interested in (in this case, the bootstrap).

zeebe/dynamic-config/src/test/java/io/camunda/zeebe/dynamic/config/api/ClusterPurgeRequestTransformerTest.java
```diff
@@ -0,0 +1,190 @@
...
+            operations -> {
+              assertThat(operations)
+                  .contains(
+                      new PartitionBootstrapOperation(
+                          id0, 0, 2, Optional.of(partitionConfigWithExporter)));
+            });
+  }
+
+  @Test
+  void shouldRecreatePartitionsWithDifferentConfig() {
```
npepinpe, line 152: :rocket: Nice

npepinpe (approved):
:wrench: I know I said I didn't care, but the more I thought about it, the more I think we should remove the `isClusterPurge` from the `PartitionLeaveApplier`, and instead pass a `minimumAllowedReplicas`, which defaults to 1 (but we can set it to 0 during purge). I think this would make the intent clearer. I also added it as a follow up.

npepinpe:
In the future, we should try and keep the PRs smaller. I think in this case, we could have split it by first tackling:

- https://github.com/camunda/camunda/issues/25962
- https://github.com/camunda/camunda/issues/25963

These could be done fairly independently I think, as they simply modify existing behavior, and you can assert it works via tests.

You could then tackle https://github.com/camunda/camunda/issues/25964 immediately after these two. Even though the transformer would be "unused" (except in tests), since we did a prototype, you should have a good idea of how it'll be used. You can then unit test it extensively - and even if it needs a little modification when you get to https://github.com/camunda/camunda/issues/25961, it would still be less in general.

Finally, https://github.com/camunda/camunda/issues/25961 could also be split further, but I think at this point it wouldn't be all that necessary. But just for the exercise, we could think of the steps we took in the prototype.

- First you do https://github.com/camunda/camunda/issues/25962 and https://github.com/camunda/camunda/issues/25963
- Then you introduce an endpoint which sends a purge request that is not handled. Only test some basic expected cases (e.g. return 502 when no coordinator is available). This is still valuable as it implements the DTOs, the encoders/decoders, etc.
- Then you implement the transformer in https://github.com/camunda/camunda/issues/25961.
- Then you put it all together and add more tests :)

We can try and slice it together next time :+1:

Issue #25861, opened 2024-12-10
🎯 Online cluster data purge


npepinpe:
## Description

This is a sub-epic in the context of the [complete cluster purge epic](https://github.com/camunda/product-hub/issues/2563).

The goal is to have a single endpoint which will trigger deletion of any live data in _all_ brokers in the cluster, as well as any historical data tied to any configured exporters which support purging.

Out of scope here (but tackled elsewhere) is:
- Integration in CPT (done [here](https://github.com/camunda/camunda/issues/19182))
- Integration in SaaS
- Integration in C8Run (other than through the regular REST API)

### High level solution

> [!Note]
> The solution design relies heavily on the existing cluster scaling API (aka dynamic cluster configuration module), as well as extending the exporter API. By leveraging these, we immediately benefit from crash resilience, fault tolerance, distributed consensus/coordination, and observability.

You can broadly split the epic into two parallel work streams which will need to be integrated together later.

#### Exporter purge API

At the moment, exporters are instantiated in two different contexts:

- Once, at startup (and without a partition), for validation. There, their lifecycle is: `Exporter#configure(Context)`, _always_ followed by `Exporter#close()`.
- On every leader transition (within a partition context), for exporting. There, their lifecycle is: `Exporter#configure(Context)`, `Exporter#open(Controller)`, followed by many `Exporter#export(Record<?>)`, until a final `Exporter#close()`.

We will now add a third context: during purge. In this case, like in the validation phase, there is no partition. The lifecycle of instances in the purge phase will be: `Exporter#configure(Context)`, `Exporter#purge()`, and _always_ followed by `Exporter#close()`. As purging may be called multiple times, this phase should be idempotent, and essentially a no-op when there is nothing to delete.

> [!Warning]
> It's imperative to _always_ call `Exporter#close()` to ensure we release any resources. 

> [!Note]
> We do not call `Exporter#open(Controller)` here as this is typically used to set up the remote data source, which we _do not want to do here_.

This will later be used and orchestrated by the coordinator node via the cluster API.

#### Cluster purge API

We will add a new endpoint to the [cluster API](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/operations/cluster-scaling/#api-reference), `POST /actuator/cluster/purge`.

We opt for an explicit endpoint and not `DELETE /` to avoid any potential mistakes. It should be crystal clear that the user wants to purge the data.

[...]

PR #26024, opened 2024-12-12
feat: add new optional purge method to the exporter API #25861

### LINKED ISSUE #26028: Add a new optional purge method to the exporter API
## Description

Add a new optional purge method to the `Exporter` API. This method will later be called as part of a cluster purge operation to delete all exported data.

## Acceptance criteria

- [x] A new method `Exporter#purge` is added, with a default no-op implementation
- [x] The method is documented via Javadoc
  - [x] It's documented when it will be called
  - [x] It's documented that it must successfully return _only_ when the data has been successfully deleted
  - [ ] It's documented that it may be retried, and must be idempotent

srsp:
## Description

Add a new optional purge method to the exporter API.

## Checklist

- [ ] for CI changes:

### DIFF
--- a/zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/Exporter.java
+++ b/zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/Exporter.java
@@ -73,4 +73,19 @@ public interface Exporter {
    * @param record the record to export
    */
   void export(Record<?> record);
+
+  /**
+   * Hook to perform a data purge.
+   *
+   * <p>When this method is called, the exporter <b>deletes all data</b> that has been exported so
+   * far. This method should be blocking and only return when all data has been deleted.
+   *
+   * <p>Currently there are two scenarios where this method can be called:
+   *
+   * <ol>
+   *   <li>Before a backup is restored</il>
+   *   <li>Between integration tests
+   * </ol>
+   */
+  default void purge() throws Exception {}
 }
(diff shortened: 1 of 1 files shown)


zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/Exporter.java
```diff
@@ -73,4 +73,19 @@ default void close() {}
...
+  /**
+   * Hook to perform a data purge.
+   *
+   * <p>When this method is called, the exporter <b>deletes all data</b> that has been exported so
+   * far. This method should be blocking and only return when all data has been deleted.
+   *
+   * <p>Currently there are two scenarios where this method can be called:
+   *
+   * <ol>
+   *   <li>Before a backup is restored</il>
```
npepinpe, line 86: :smile: Thanks, but since we won't have the restore functionality until 8.9, I would remove this to avoid confusing people :see_no_evil:

zeebe/exporter-api/src/main/java/io/camunda/zeebe/exporter/api/Exporter.java
```diff
@@ -73,4 +73,19 @@ default void close() {}
...
+   *   <li>Between integration tests
```
npepinpe, line 87: :x: That's true, but more importantly, it's whenever a cluster purge operation occurs (which is what the test infrastructure will call). So let's just specify the cluster purge operation for now.

npepinpe (approved):
:rocket: 

Please address all comments before merging, but no need for another review

PR #25799, opened 2024-12-10
refactor: authorization checks in Tasklist

romansmirnov:
## Description

* Replaces the `hasPermission()` checks with the new authorization concept, i.e., checking if a user/client is allowed to complete, assign, and unassign a task and creating an instance.
* Implements two modes:
  * New Architecture: Instead of communicating with the Gateway, it directly executes broker requests. In that case, the engine will perform the authorization checks.
  * Old Architecture (compatibility): This mode ensures that the old backend makes requests to the Gateway to ensure a running dev environment and old test infrastructure. However, the old backend does the authorization checks using the `AuthorizationChecker`.
* Provides integration tests.

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

related to #24665

PR #26039, opened 2024-12-12
feat: handle rejection or delay of zeebe tasks

### LINKED ISSUE #25419: Allow Tasklist to handle rejections
When dealing with UserTasksListeners it may be the case that the OOTB Tasklist tries to complete or assign a task meanwhile a listener is being executed. When this happens C8 REST API is going to reject the assignment/completion command and an error is going to be returned:

```
{
    "type": "about:blank",
    "title": "INVALID_STATE",
    "status": 409,
    "detail": "Command 'COMPLETE' rejected with code 'INVALID_STATE': Expected to complete user task with key '2251799813686157', but it is in state 'COMPLETING'",
    "instance": "/v2/user-tasks/2251799813686157/completion"
}
```

When this happens, Tasklist should display and handle this error gracefully as described here: https://docs.google.com/document/d/1BxvvkdDJurzblKcnjbP00sUE3FCv5rDM81IOG4OG_cM/edit?tab=t.0#heading=h.ak839ldyiwxs

Another scenario that can happen is the TaskListener taking longer than expected to be executed, causing a time-out. When this happens Tasklist should also communicate that the operation is taking longer than expected. [See the design](https://docs.google.com/document/d/1BxvvkdDJurzblKcnjbP00sUE3FCv5rDM81IOG4OG_cM/edit?tab=t.0).

As part of this task is also expected that the following metrics are created to track the usage:
- How many times a Task was rejected due to UserTask Listeners when assigning and completing
- How many times a Task timed-out due to UserTask Listeners when assigning and completing

**Acceptance Criteria**
- [ ] When assigning/completing a Zeebe User Task using Tasklist OOTB considering that this UserTask has a Listener, in case the listener is succesful, the task assignment/completion should executed with success.
- [ ] When assigning/completing a Zeebe User Task using Tasklist OOTB considering that this UserTask has a Listener, in case the listener takes longer than expected and this results in a time-out, an error explaining the scenario should be displayed in Tasklist OOTB
- [ ] When assigning/completing a Zeebe User Task using Tasklist OOTB considering that this UserTask has already changed the state to ASSIGNING/COMPLETING, the operation should be rejected and an error explaining the sceario should be dusplayed in Tasklist OOTB
- [ ] We have mixpanel metrics to count rejections and time-outs of tasks

tsedekey:
## Description

This PR introduces enhancements to handle Zeebe task event errors in Tasklist:

1. **Error Notification for Task Listener Conflicts:**
When a user triggers a Zeebe task event (assignment, unassignment, or completion) and a task listener is already processing the same event, Zeebe returns an error. This PR ensures that the error is displayed to the user as a notification.

2. **Timeout Handling:**
If a task listener takes too long to process an event, Zeebe issues a timeout error. Tasklist now displays this as an informational notification to inform the user of the delay.

3. **Metrics Tracking:**
Code has been added to track the frequency of task listener rejections and delayed timeout notifications for system monitoring and analysis.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25419


tasklist/client/src/Tasks/Task/Details/getTaskAssignmentChangeErrorMessage.tsx
```diff
@@ -6,7 +6,13 @@
  * except in compliance with the Camunda License 1.0.
  */
 
-function getTaskAssignmentChangeErrorMessage(code: string) {
+import {t} from 'i18next';
+
+function getTaskAssignmentChangeErrorMessage(
+  code?: string,
+  statusCode?: number,
+  isAssigning?: boolean,
```
vsgoulart, line 14: Why are all these params optional? It's not good to make params optional unless strictly necessary to avoid branching in the code

tasklist/client/src/Tasks/Task/Details/Header.tsx
```diff
@@ -139,15 +140,47 @@ const AssignButton: React.FC<{
...
+
+      setAssignmentStatus('off');
+      if (shouldDisplayNotification(errorMessage, statusCode)) {
+        if (statusCode === 409) {
+          tracking.track({
+            eventName: isAssigned
+              ? 'task-unassignment-rejected-notification'
+              : 'task-assignment-rejected-notification',
+          });
+        }
```
vsgoulart, line 173: Parsing these raw responses is not good solution, we did this in the past as a temporary solution  as you seen it became permanent because it was never prioritized.

I understand that the V1 API will be deprecated in the coming quarters so it doesn't make sense to have proper error codes for the entire API but we could at least some basic error codes for this specific endpoint would be a good middle.
Parsing Java error like on `shouldDisplayInfoNotification` is not very nice

tasklist/client/src/modules/mutations/useCompleteTask.tsx
```diff
@@ -25,7 +25,15 @@ function useCompleteTask() {
...
 
-      throw error ?? new Error('Could not complete task');
+      if (error !== null || error !== undefined) {
+        const errorMessage =
+          error.response instanceof Response
+            ? (await error.response.json())?.message
+            : undefined;
+        throw new Error(errorMessage);
+      } else {
+        new Error('Could not complete task');
```
vsgoulart, line 35: `error` is never `null` or `undefined` the nullish coalescing was left over because the `request` function was changed and this wasn't updated.

tasklist/client/src/modules/mutations/useCompleteTask.tsx
```diff
@@ -25,7 +25,15 @@ function useCompleteTask() {
         return response.json();
       }
 
-      throw error ?? new Error('Could not complete task');
+      if (error !== null || error !== undefined) {
+        const errorMessage =
+          error.response instanceof Response
```
vsgoulart, line 30: You don't need to use `instanceof`, that's why we have the `variant` inside the error object

vsgoulart (changes requested):

tasklist/client/src/Tasks/Task/Details/getTaskAssignmentChangeErrorMessage.tsx
tsedekey: These have been removed after revising the implementation approach.

tasklist/client/src/Tasks/Task/Details/Header.tsx
tsedekey: I fully understand and agree with the point raised. I was also hesitant to rely on raw responses, but I wasn’t sure how much effort was appropriate given the V1 API deprecation. I’ve now added specific error codes for this endpoint as a middle ground.

tasklist/client/src/modules/mutations/useCompleteTask.tsx
tsedekey: I misunderstood the purpose of the variant. I’ve removed the instanceof check accordingly.

tasklist/client/src/Tasks/Task/Details/Header.tsx
```diff
@@ -137,11 +138,50 @@ const AssignButton: React.FC<{
         tracking.track({eventName: 'task-assigned'});
       }
     } catch (error) {
-      const errorMessage = (error as Error).message ?? '';
+      const errorMessage = parseJSON((error as Error).message);
+      if (
+        errorMessage?.status === 504 &&
```
vsgoulart, line 143: We shouldn't return the status code in the body, it's already present in the response'
This logic should be encapsulated on the backend, we should be able to derive the meaning/message from the error only by the error code

tasklist/client/src/Tasks/Task/Variables/index.tsx
```diff
@@ -125,6 +126,15 @@ const Variables: React.FC<Props> = ({
 
           setSubmissionState('finished');
         } catch (error) {
+          const errorMessage = parseJSON((error as Error)?.message);
```
vsgoulart, line 129: Why is this necessary?

tasklist/client/src/Tasks/Task/getCompleteTaskErrorMessage.tsx
vsgoulart, line 1: If we're introducing error codes we should be able to remove this function or just replace it with one where we provide the error code and returns the message

tasklist/client/src/Tasks/Task/index.tsx
```diff
@@ -150,16 +151,44 @@ const Task: React.FC = observer(() => {
   }
 
   function handleSubmissionFailure(error: Error) {
-    const errorMessage = isRequestError(error)
-      ? (error?.networkError?.message ?? error.message)
-      : error.message;
+    const errorMessage = parseJSON(error?.message);
+    if (
+      errorMessage?.status === 504 &&
+      errorMessage?.error_code === 'TASK_PROCESSING_TIMEOUT'
```
vsgoulart, line 157: it should be `errorCode`, we use camel case on the APIs

nathansandi (approved):
Generally looks goor for me,
only minor suggested changes.

tasklist/webapp/pom.xml
```diff
@@ -59,11 +59,6 @@
       <artifactId>grpc-api</artifactId>
     </dependency>
 
-    <dependency>
```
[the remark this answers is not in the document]
tsedekey: I believe it was originally added by @romansmirnov in this [PR](https://github.com/camunda/camunda/pull/25799/files#diff-4382e09e9e375662a4f96a3f8085a600a7375343e9d95f4f6996a767416e98cd). The `io.camunda.zeebe.gateway.cmd.ClientException` class from that library was used incorrectly in the `CamundaClientBasedAdapter` class. Once the correct exception class was used, the library was no longer required.

vsgoulart (approved):

PR #26045, opened 2024-12-12
fix: add support remove auth permissions

Oleksiivanov:
## Description

in this PR added handler for remove auth permissions;

 - Renamed `AuthorizationHandler` to `AuthorizationPermissionAddedHandler` for clarity.
 - Added `AuthorizationPermissionRemovedHandler` using an Elasticsearch update script to remove resource IDs cleanly and handle empty permissions.
 - Added unit tests for both handlers and updated integration tests to verify add/remove permission flows.

**Why I remove permissions by using the script :** 

The script avoids complex changes in entity logic by directly modifying permissions at the document level in Elasticsearch. This approach prevents the new methods in the authorization state and adding logic to auth processor. It also avoids introducing new events like `update` for the add/remove operations, or adding update logic to a separate handler, which would complicate the existing logic or do logic not clear.

While removing a few resource IDs could technically be seen as an update, for the user, it behaves as a clear "remove permission" operation. Using the script keeps the code clean, focused, and easier to maintain without breaking the current flow of add and remove events.

## Checklist

- [ ] for CI changes:

### DIFF
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/AuthorizationReader.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/read/service/AuthorizationReader.java
@@ -14,8 +14,7 @@ import io.camunda.db.rdbms.write.domain.AuthorizationDbModel;
 import io.camunda.search.entities.AuthorizationEntity;
 import io.camunda.search.query.AuthorizationQuery;
 import io.camunda.search.query.SearchQueryResult;
-import io.camunda.security.entity.Permission;
-import java.util.Optional;
+import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -30,18 +29,15 @@ public class AuthorizationReader extends AbstractEntityReader<AuthorizationEntit
     this.authorizationMapper = authorizationMapper;
   }
 
-  public Optional<AuthorizationEntity> findOne(
+  public List<AuthorizationEntity> find(
       final long ownerKey, final String ownerType, final String resourceType) {
-    final var result =
-        search(
+    return search(
             AuthorizationQuery.of(
                 b ->
                     b.filter(
                         f ->
-                            f.ownerKeys(ownerKey)
-                                .ownerType(ownerType)
-                                .resourceType(resourceType))));
-    return Optional.ofNullable(result.items()).flatMap(items -> items.stream().findFirst());
+                            f.ownerKeys(ownerKey).ownerType(ownerType).resourceType(resourceType))))
+        .items();
   }
 
   public SearchQueryResult<AuthorizationEntity> search(final AuthorizationQuery query) {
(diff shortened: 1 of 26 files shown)


remcowesterhoud (commented):
In the standup @romansmirnov mentioned this wouldn't work as-is. I'll ignore the review request for now. Please re-request it if the PR is ready and you deem my review necessary 🙂

romansmirnov:
@Oleksiivanov, the separation between add and remove makes sense to me. However, the fix does not consider that the event contains only the added permissions. So, in the case of the intent `PERMISSION_ADDED`, it would simply override the entire document, with the list of permissions coming along with the event, meaning any other permission is lost.

Of course, this can also be solved by using a script when permissions are added. But I don't recommend doing so.

Instead, we could flatten the data structure of an authorization, for example

```
{
  "mappings": {
    "dynamic": "strict",
    "properties": {
      "id": {
        "type": "keyword"
      },
      "ownerKey": {
        "type": "long"
      },
      "ownerType": {
        "type": "keyword"
      },
      "resourceType": {
        "type": "keyword"
      },
      "permissionType": {
        "type": "keyword"
      },
      "resourceId": {
        "type": "keyword"
      }
    }
  }
}
```

The ID would be then a composite of `ownerKey + resourceType + permissionType + resourceId`. Meaning, for each permission and resource being part of the `PERMISSION_ADDED` intent becomes a single document, for example:

Given the following event:

```
{
  "id": "1233",
  "ownerKey": 456,
  "ownerType": "USER",
  "resourceType": "PROCESS_DEFINITION",
  "permissions": [
    {
      "type": "READ",
      "resourceIds": ["a", "b"]
    },
    {
      "type": "READ_PROCESS_INSTANCE",
      "resourceIds": ["a"]
    }
  ]
}
````

This would result in 3 documents:

```
[{
  "id": "456_PROCESS_DEFINITION_READ_A",
  "ownerKey": 456,
  "ownerType": "USER",
  "resourceType": "PROCESS_DEFINITION",
  "permissionType": "READ",
  "resourceId": "a"
},
{
  "id": "456_PROCESS_DEFINITION_READ_B",
  "ownerKey": 456,
  "ownerType": "USER",
  "resourceType": "PROCESS_DEFINITION",
  "permissionType": "READ",
  "resourceId": "b"
},
{
  "id": "456_PROCESS_DEFINITION_READ_PROCESS_INSTANCE_A",
  "ownerKey": 456,
  "ownerType": "USER",
  "resourceType": "PROCESS_DEFINITION",
  "permissionType": "READ_PROCESS_INSTANCE",
  "resourceId": "b"
}]
```

Alternatives would be instead of `PATCH` requests, the entire (and update) authorization object must be `PUT`'ed. That way, in the exporter the entire Authorization would be received and simply be updated. This would also require that the authorization has a unique id (or more likely a key generated by the engine).

@remcowesterhoud, do you have input?

remcowesterhoud:
> @remcowesterhoud, do you have input?

What you said makes perfect sense to me. It means we'll have a disprecancy between how we store data within Zeebe versus the in the exporters, but that's fine imo. These components shouldn't be too tightly coupled anyway.

> Alternatives would be instead of PATCH requests, the entire (and update) authorization object must be PUT'ed. That way, in the exporter the entire Authorization would be received and simply be updated. This would also require that the authorization has a unique id (or more likely a key generated by the engine).

Regarding this. I considered this initially but chose against it to keep the API simple to use. From a user's perspective if I want to add an authorization I don't care about all the other ones that already exists and would prefer not having to specify all these in the request.

megglos:
as agreed in a call with @Oleksiivanov we will close this for now, as the patch command is not relevant for the scope of https://github.com/camunda/camunda/issues/26817

PR #26053, opened 2024-12-13
feat: implement camunda scoped variables for OIDC auth

Ben-Sheppard:
## Description

As per the conversation in Slack, we currently only expose the default spring properties when configuring OIDC, these property variables can be awkward to work with.

This PR introduces a new set of camunda scoped properties (`camunda.security.auth.oidc.XYZ`) which map to the spring values. What this means is that I can either:

1. Use the Spring values as ENV variables
2. Use the Camunda values as properties
3. Use the Camunda values as ENV variables

I have tested this with env vars and it works as expected, I have not yet tested with a custom application properties file.

### DIFF
--- a/dist/src/main/resources/application-auth-oidc.properties
+++ b/dist/src/main/resources/application-auth-oidc.properties
@@ -1,11 +1,11 @@
-spring.security.oauth2.client.registration.oidcclient.client-id=oidcclient2
-spring.security.oauth2.client.registration.oidcclient.client-secret=secret2
-spring.security.oauth2.client.registration.oidcclient.authorization-grant-type=authorization_code
-spring.security.oauth2.client.registration.oidcclient.redirect-uri=http://localhost:8080/login/oauth2/code/oidcclient2
+spring.security.oauth2.client.registration.oidcclient.client-id=${camunda.security.authentication.oidc.client-id:}
+spring.security.oauth2.client.registration.oidcclient.client-secret=${camunda.security.authentication.oidc.client-secret:}
+spring.security.oauth2.client.registration.oidcclient.authorization-grant-type=${camunda.security.authentication.oidc.grant-type:authorization_code}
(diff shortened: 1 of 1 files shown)


dist/src/main/resources/application-auth-oidc.properties
```diff
@@ -1,8 +1,8 @@
...
+spring.security.oauth2.client.registration.oidcclient.client-id=${camunda.security.auth.oidc.client-id:}
+spring.security.oauth2.client.registration.oidcclient.client-secret=${camunda.security.auth.oidc.client-secret:}
+spring.security.oauth2.client.registration.oidcclient.authorization-grant-type=${camunda.security.auth.oidc.grant-type:authorization_code}
+spring.security.oauth2.client.registration.oidcclient.redirect-uri=${camunda.security.auth.oidc.redirect-uri:}
 spring.security.oauth2.client.registration.oidcclient.provider=oidcclient
-spring.security.oauth2.client.registration.oidcclient.scope=openid,profile
+spring.security.oauth2.client.registration.oidcclient.scope=${camunda.security.auth.oidc.scopes:openid,profile}
 
-spring.security.oauth2.client.provider.oidcclient.issuer-uri=http://auth.identity:9000
+spring.security.oauth2.client.provider.oidcclient.issuer-uri=${camunda.security.auth.oidc.issuer-uri:}
```
emulbreh, line 8: We do have `camunda.security.authentication.method` now. Should we use the longer namespace here too? Also, do these properties need to be reflected in the configuration classes?

emulbreh (approved):

dist/src/main/resources/application-auth-oidc.properties
emulbreh: It's [here](https://github.com/camunda/camunda/blob/main/security/security-core/src/main/java/io/camunda/security/configuration/AuthenticationConfiguration.java#L16), but I'm not sure if we actually set it anywhere.

Issue #26056, opened 2024-12-13
Opensearch schema manager false negative during index validation


EuroLew:
**Description**

The opensearch index validator will return a false negative when comparing a schema .json file and the index that was created with that .json file, this is due to certain fields not appearing in the mappings of a serialised return object. In the given example the `eager_global_ordinals` field is not returned as part of the `join` field. This issue seems to be unique to opensearch and not elastic

![image](https://github.com/user-attachments/assets/9dbd1660-9b74-4d6a-93a0-0b09b21e5fa0)
![image](https://github.com/user-attachments/assets/ccf5bf85-97f6-4e26-9648-d388a8544ecf)

Success Criteria:
- [ ] Indices created by a schema file will not trigger an index validation error which comparing against the schema file
- [ ] Add a test to catch the issue of schema files clashing against their created indices.
```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/issues/26085
```

EuroLew:
The issue is already fixed by opensearch on Nov 13 https://github.com/Xtansia/opensearch-java/blob/0e386f4b508b995bdfc14780ea1f2a955ad41596/java-client/src/generated/java/org/opensearch/client/opensearch/_types/mapping/JoinProperty.java, they added eager_global_ordinals to the join serialisation. The only issue is that they haven't added to a release which they appear to do every 2-3 weeks.

For now to fix the breaking issue the field will be removed from affected schemas and when an `opensearch-java` client version is released with the fix we can add the `eager_global_ordinals` option back with the updated client

ChrisKujawa:
@EuroLew can this be closed?

PR #26060, opened 2024-12-13
Fix opensearch breaking schemas

EuroLew:
## Description

If a schema is created in opensearch according to a json file, when comparing the returned mapping from the opensearch client to the json file there should not be a schema validation error.

The issue is already fixed by opensearch on Nov 13 https://github.com/Xtansia/opensearch-java/blob/0e386f4b508b995bdfc14780ea1f2a955ad41596/java-client/src/generated/java/org/opensearch/client/opensearch/_types/mapping/JoinProperty.java, they added eager_global_ordinals to the join serialisation. The only issue is that they haven't added to a release which they appear to do every 2-3 weeks.

For now to fix the breaking issue the field will be removed from affected schemas and when an opensearch-java client version is released with the fix we can add the eager_global_ordinals option back with the updated client

## Checklist

- [x] Indices created by a schema file will not trigger an index validation error which comparing against the schema file
- [x] Add a test to catch the issue of schema files clashing against their created indices.

## Related issues

relates to #26056

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
@@ -10,10 +10,14 @@ package io.camunda.exporter.schema;
 import static io.camunda.exporter.schema.SchemaTestUtil.mappingsMatch;
 import static io.camunda.exporter.utils.CamundaExporterITInvocationProvider.CONFIG_PREFIX;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import co.elastic.clients.elasticsearch._types.ElasticsearchException;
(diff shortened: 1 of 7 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
```diff
@@ -381,4 +386,23 @@ void shouldAlsoUpdateCorrespondingIndexWhenIndexTemplateUpdated(
 
     assertThat(mappingsMatch(updatedIndex.get("mappings"), newMappingsFile)).isTrue();
   }
+
+  @TestTemplate
```
ChrisKujawa, line 390: 🔧 Please mark it as regression test and link to https://github.com/camunda/camunda/issues/26056

ChrisKujawa (approved):
Thanks Euro great catch and thanks for the quick fix 🚀 

Can you add to the PR description a bit more details (the information you shared to me, regarding OS bug fix and release, etc. that we should bring it back) Furthermore, create an follow up issue for bringing this back, please.

ChrisKujawa:
@EuroLew somehow our test is failing 🤔

EuroLew:
> @EuroLew somehow our test is failing 🤔

fixed @RegressionTest is not compatible with @TestTemplate by adding @RegressionTestTemplate

PR #26061, opened 2024-12-13
Task Exporter backwards compatibility

### LINKED ISSUE #25293: Adjust exporter and REST API V1 and V2 to avoid migration
As part of exporter we need to support User Tasks records that were created before using the importer. The current problem is that because we are improving the User Task Search by Variables, we are changing the way the `ids` and `_routing` are stored in the UserTask indice. The consequence of that is potential failures when trying to update records (Assigning/completing user Tasks), as the shard may be different than expected.

To overcome this challenge we need to have a way of identifying what record has the "old" id and what record has the "new" id in operations that are updating the user tasks.

- [ ] As the first step of the exporter, store the highest sequence of the user task on `tasklist-import-position-8.2.0_` (if still not stored)
- [ ] When generating the id of the UserTask entity, check if the sequence is higher or lower than the sequence stored
   - [ ] If higher, the id of the entity should be the `ElementInstanceKey`
   - [ ] If lower/equals, the id of the entity should be the `UserTaskKey`
- [ ] When flushing the UserTask we should compare the id to decide the value of the routing
   - [ ] If the id is equals as the FlowNodeInstanceId it means that the routing should be the processInstanceKey
   - [ ] If the id is equals as the UserTaskKey, it means that the routing should be the UserTaskKey

**Acceptance Criteria**
- [ ] Assigning a Zeebe User Task using v2 where this task was created on version 8.6, should work as expected and reflect in ES/OS.
- [ ] Assigning a Zeebe User Task using v2 where this task was created on version 8.7, should work as expected and reflect in ES/OS.
- [ ] Assigning a Zeebe User Task using v1 where this task was created on version 8.6, should work as expected and reflect in ES/OS.
- [ ] Assigning a Zeebe User Task using v1 where this task was created on version 8.7, should work as expected and reflect in ES/OS.
- [ ] Completing a Zeebe User Task using v2 where this task was created on version 8.6, should work as expected and reflect in ES/OS.
- [ ] Completing a Zeebe User Task using v2 where this task was created on version 8.7, should work as expected and reflect in ES/OS.
- [ ] Completing a Zeebe User Task using v1 where this task was created on version 8.6, should work as expected and reflect in ES/OS.
- [ ] Completing a Zeebe User Task using v1 where this task was created on version 8.7, should work as expected and reflect in ES/OS.
- [ ] Assigning a Job User Task using v1 where this task was created on version 8.6, should work as expected and reflect in ES/OS.
- [ ] Assigning a Job User Task using v1 where this task was created on version 8.7, should work as expected and reflect in ES/OS.
- [ ] Completing a Job User Task using v1 where this task was created on version 8.6, should work as expected and reflect in ES/OS.
- [ ] Completing a Job User Task using v1 where this task was created on version 8.7, should work as expected and reflect in ES/OS.
- [ ] Updating a Zeebe User Task using v2 where this task as created on version 8.6, should work as expected and reflect in ES/OS.
- [ ] Updating a Zeebe User Task using v2 where this task as created on version 8.7, should work as expected and reflect in ES/OS.

panagiotisgts:
## Description

- [x] Opensearch test is disabled due to issue when migrating. Should be fixed by #26060

Handle User Task updates/assignment/unassignment/completion with correct routing and ID setting on the CamundaExporter for records prior to 8.7.

## Checklist

- [ ] for CI changes:

## Related issues

closes #25293


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
```diff
@@ -140,8 +154,14 @@ public void flush(final TaskEntity entity, final BatchRequest batchRequest) {
...
 
+    final boolean previousVersionRecord = refersToPreviousVersionRecord(entity.getKey());
+
     batchRequest.upsertWithRouting(
-        indexName, entity.getId(), entity, updateFields, entity.getProcessInstanceId());
+        indexName,
+        previousVersionRecord ? String.valueOf(entity.getKey()) : entity.getId(),
+        entity,
+        updateFields,
+        previousVersionRecord ? String.valueOf(entity.getKey()) : entity.getProcessInstanceId());
```
ChrisKujawa, line 164: 🔧 I would like to have some comments, explaining actually what we are doing here and why.

For example: Based on the key, and the first seen created user task key, we can determine that the key is either created before or after.

If Before it means it was stored before with the user task key as ID (in ES/OS), so we have to use the same to update it.

_Thinking of this, might make sense to split the handler to make the logic separate? (not sure here, just a thought)_

If after it means we have to use the id == flownodeinstancekey as the ID of the document.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskJobBasedHandler.java
```diff
@@ -129,8 +141,15 @@ public void flush(final TaskEntity entity, final BatchRequest batchRequest) {
...
+
+    final boolean previousVersionRecord = refersToPreviousVersionRecord(entity.getKey());
+
     batchRequest.upsertWithRouting(
-        indexName, taskEntityId, entity, updateFields, processInstanceKey);
+        indexName,
+        previousVersionRecord ? String.valueOf(entity.getKey()) : taskEntityId,
+        entity,
+        updateFields,
+        previousVersionRecord ? String.valueOf(entity.getKey()) : processInstanceKey);
```
ChrisKujawa, line 152: 🔧 Here the same. Maybe we can extract this logic even to some helper, with clear API/docs

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskHandlerTest.java
```diff
@@ -48,10 +53,29 @@ public class UserTaskHandlerTest {
...
+  @BeforeEach
+  void resetMetadata() {
+    exporterMetadata.setFirstUserTaskKey(TaskImplementation.ZEEBE_USER_TASK, -1);
+  }
+
+  private static Stream<Arguments> taskIdArguments() {
+    return Stream.of(
+        // First 8.7 UserTask key, Record Key, PI key, FNI key, expected id
+        Arguments.of(100, 110, 111, 211, 211), Arguments.of(100, 10, 111, 211, 10));
+  }
```
ChrisKujawa, line 73: 👍🏼 

I think if we extract the logic into a helper, we could test this even more narrowed. Just as a thought. With two tests that are even named what they actually verify/test

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskHandlerTest.java
```diff
@@ -512,15 +585,16 @@ void flushedEntityShouldContainUpdates() {
     verify(mockRequest, times(1))
         .upsertWithRouting(
             indexName,
-            taskEntity.getId(),
+            recordKey < firstSeenKey
+                ? String.valueOf(recordKey)
+                : String.valueOf(flowNodeInstanceKey),
```
ChrisKujawa, line 590: 🔧 This is a good sign of splitting a test to scope them for there different use cases

qa/integration-tests/src/test/java/io/camunda/it/migration/util/MigrationITInvocationProvider.java
```diff
@@ -0,0 +1,510 @@
...
+            .until(
+                () ->
+                    esClient
+                            .indices()
+                            .get(GetIndexRequest.of(req -> req.index("*")))
+                            .get("operate-process-8.3.0_")
+                        != null);
+
+        // Copy previous tasklist-process to operate-process, required for V2 APIs
+        esClient.reindex(
```
ChrisKujawa, line 450: 😀 ❓ Is this the migration? Maybe we can integration our migration app, here? Like enabling it on 8.7?

qa/integration-tests/src/test/java/io/camunda/it/migration/util/MigrationITInvocationProvider.java
```diff
@@ -0,0 +1,510 @@
...
+                  testBrokers.put(db, broker);
+                  zeebeClients.put(db, zeebeUtils.get(db).getZeebeClient());
+
+                  if (afterMigrationFunc != null) {
+                    afterMigrationFunc.accept(zeebeClients.get(db), tasklistUtils.get(db));
+                  }
+
+                  closeables.add(broker);
+                  break;
+
```
ChrisKujawa, line 164: 🔧 that is a lot of set up code. Can we reduce this, or make this more readable? Like introduce separate methods, also useful for reducing with the OS setup?

qa/integration-tests/src/test/java/io/camunda/it/migration/util/TasklistUtil.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.awaitility.Awaitility;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
+
+public class TasklistUtil {
```
ChrisKujawa, line 29: 🔧 Can we maybe call it differently? Util sounds so... general. Like TasklistMigrationHelper, TasklistMigrationContainer or TasklistContainerSupport or something

🔧 We could even think about split this to Tasklistclient and TasklistContainer.
Maybe we can also make this an extension even. I think would be useful for other tests as well like: StandaloneCamundaTest (id did something similar, for Operate, see OperateClientTest to access v1)

ChrisKujawa: I guess we can iterate over it 👍🏼 the last point is nothing we need to do it immediately but good to have in mind.

qa/integration-tests/src/test/java/io/camunda/it/migration/util/ZeebeUtil.java
```diff
@@ -0,0 +1,157 @@
...
+
+  public ZeebeClient getZeebeClient() {
+    return zeebeClient;
+  }
+
+  private void extractVolume() {
+    try {
+      volume.extract(zeebeDataPath);
+      // Need to remove the .topology.meta to be recreated by new broker
+      Files.delete(zeebeDataPath.resolve("usr/local/zeebe/data/.topology.meta"));
```
ChrisKujawa, line 97: ❓ Out of curiosity what happens if we don't do it?

ChrisKujawa (approved):
Thanks @panagiotisgts 🚀 great work. Awesome that we have now even migration tests 👍🏼 I have several comments, but they are not really blockers, please have a look. Adjust where you see need, for some we might want to create follow-ups. 

👍🏼 🙇🏼

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
```diff
@@ -68,6 +73,13 @@ public boolean handlesRecord(final Record<UserTaskRecordValue> record) {
 
   @Override
   public List<String> generateIds(final Record<UserTaskRecordValue> record) {
+    if (record.getIntent().equals(UserTaskIntent.CREATED)) {
+      exporterMetadata.setFirstUserTaskKey(TaskImplementation.ZEEBE_USER_TASK, record.getKey());
+    }
```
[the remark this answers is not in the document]
ChrisKujawa: I think it will not update: as `if (firstUserTaskKeys.get(implementation) == UNSET_POSITION) {`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
```diff
@@ -44,11 +45,15 @@ public class UserTaskHandler implements ExportHandler<TaskEntity, UserTaskRecord
           UserTaskIntent.UPDATED);
   private final String indexName;
   private final ExporterEntityCache<String, CachedFormEntity> formCache;
+  private final ExporterMetadata exporterMetadata;
 
   public UserTaskHandler(
-      final String indexName, final ExporterEntityCache<String, CachedFormEntity> formCache) {
+      final String indexName,
```
ChrisKujawa, line 51: @romansmirnov do you have time to review the user task handler change whether it is what you expect?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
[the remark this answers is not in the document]
panagiotisgts: This is being done inside the `setFirstUserTaskKey` method of the metadata. Since we only care for the first `CREATED` intent to capture the value.

qa/integration-tests/src/test/java/io/camunda/it/migration/util/MigrationITInvocationProvider.java
panagiotisgts: This is required because V2 APIs check against `operate-process` which, if you run only Tasklist 8.6 is not being populated. Spinning operate in the background would not require this _migration_ of processes.

qa/integration-tests/src/test/java/io/camunda/it/migration/util/ZeebeUtil.java
panagiotisgts: If you don't remove it, the new Broker tries to access the file, but the file permissions set by the previous one do not allow it. Don't know if there is a workaround for this though.

panagiotisgts:
Did a refactor of the Migration test setup to clean the code a bit. Also, V2 APIs no longer return Job Worker based User Tasks so the test itself has been adapted.

Test will fail due to issue fixed on pending PR: https://github.com/camunda/camunda/pull/26244

qa/integration-tests/src/test/java/io/camunda/it/migration/util/MigrationITInvocationProvider.java
```diff
@@ -330,109 +189,113 @@ public MigrationITInvocationProvider withDatabaseTypes(final DatabaseType... dat
...
+            ? Map.of("url", databaseExternalUrls.get(databaseType))
+            : Map.of("url", databaseExternalUrls.get(databaseType), "type", "opensearch");
+    return cfg -> {
+      cfg.setClassName("io.camunda.exporter.CamundaExporter");
+      cfg.setArgs(
+          Map.of(
+              "connect",
+              connect,
+              "bulk",
+              Map.of("size", 10, "delay", 1),
```
ChrisKujawa, line 248: 🔧 We could set this to one

ChrisKujawa:
I would like to get this merged, so we can mark this as complete and we can continue.

I think we can iterate over the tests and some other parts, as mentioned above, but it looks already good enough.

PR #26067, opened 2024-12-13
feat: do not flush while waiting for importer

### LINKED ISSUE #26047: Camunda Exporter flushs even when importers are not done
**Describe the bug**

It was detected by @romansmirnov and then observed in our benchmarks as well that even because importers are not done https://github.com/camunda/camunda/issues/26046 that flushing still happens periodically

See related slack https://camunda.slack.com/archives/C06F0GLJNFM/p1734024612929269

EuroLew:
## Checklist

- [x] While waiting for the importer(s) to complete flushing does not happen.

## Related issues

closes #26047

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -243,6 +243,10 @@ public class CamundaExporter implements Exporter {
   }
 
   private void flushAndReschedule() {
+    if (configuration.getIndex().shouldWaitForImporters() && !importersCompleted) {
+      scheduleDelayedFlush();
+      return;
+    }
     try {
       flush();
       updateLastExportedPosition(lastPosition);
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -125,7 +125,6 @@ public class CamundaExporter implements Exporter {
 
     writer = createBatchWriter();
 
-    scheduleDelayedFlush();
     checkImportersCompletedAndReschedule();
     controller.readMetadata().ifPresent(metadata::deserialize);
     taskManager.start();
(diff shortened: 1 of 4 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -532,6 +532,8 @@ final var record =
 
       camundaExporter.export(record);
 
+      controller.runScheduledTasks(Duration.ofMinutes(1));
+
```
ChrisKujawa, line 536: 🔧 Can we use the configured time?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -243,6 +243,10 @@ private void scheduleDelayedFlush() {
   }
 
   private void flushAndReschedule() {
+    if (configuration.getIndex().shouldWaitForImporters() && !importersCompleted) {
+      scheduleDelayedFlush();
+      return;
+    }
```
ChrisKujawa, line 249: ❌ 🤔 Not sure whether this is the right fix. As this will no longer schedule more flushes. 

Can we maybe not schedule the first flush, and only schedule it if importing is done?

ChrisKujawa (changes requested):

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -532,6 +535,8 @@ final var record =
 
       camundaExporter.export(record);
 
+      controller.runScheduledTasks(Duration.ofSeconds(config.getBulk().getDelay()));
```
ChrisKujawa, line 538: ❓ Why we schedule this so often?

EuroLew: Refactored the test so there's only one `runScheduledTasks` to try and trigger the scheduled flush.

ChrisKujawa (approved):

Issue #26080, opened 2024-12-13
Follow-up: Fix page.firstSortValues returned by the search API


houssain-barouni:
**Describe the bug**

Follow-up of https://github.com/camunda/camunda/pull/26082

Currently in the [search result transformer](https://github.com/camunda/camunda/blob/94e8ae400f328a1fb49ea0882242bd8494f40c53/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/query/SearchQueryResultTransformer.java#L33-L38), we get only the sortValues of the last item of the page.

We also need to get the sortValues of the first item to pass it into firstSortValues returned in the response, to be used to fetch the previous page

PR #26082, opened 2024-12-13
fix: search does not correctly return lastSortValues

### LINKED ISSUE #26030: Elasticsearch Issue on INT with SNAPSHOT generation
**Describe the bug**

On INT when using a SNAPSHOT generation cluster (cluster id = 865f4ea4-ac5a-4a7c-9998-fa00607e23ee), I noticed that connectors is unhealthy due to an issue with elasticsearch and zeebe gateway

**To Reproduce**

1. Create a cluster on INT with a generation 'Cross Component Test Suite'
2. Observe the error

**Expected behavior**

The connectors pod should be healthy with no errors.

**Log/Stacktrace**

https://console.cloud.google.com/logs/query;query=resource.labels.namespace_name%3D%22865f4ea4-ac5a-4a7c-9998-fa00607e23ee-zeebe%22%0Aresource.labels.container_name%3D%22connectors%22;cursorTimestamp=2024-12-12T16:33:13.639467231Z;duration=PT1H?project=camunda-saas-int&inv=1&invt=Abj8DA
 <p>

```
{
    type: about:blank
    title: co.elastic.clients.elasticsearch._types.ElasticsearchException
    status: 500
    detail: Unexpected error occurred during the request processing: [es/search] failed: [search_phase_execution_exception] all shards failed
    instance: /865f4ea4-ac5a-4a7c-9998-fa00607e23ee/v2/process-definitions/search
}
```

</p>
</details>

**Environment:**
- OS: 
- Zeebe Version: 
- Configuration:

houssain-barouni:
## Description

More details can be found in https://github.com/camunda/camunda/issues/26030#issuecomment-2539729696

The search does not correctly return `lastSortValues` and `firstSortValues`.
These fields, which contain a list, are required by the client to fetch the next and previous pages, respectively.

When sorting is applied in the search, these fields are expected to hold the sort values of the first and last items on the page. Since sorting by key is always appended to the query sent to Elasticsearch, the sortValues also include the key of the last item, besides the values of the sort fields provided by the client.

In this pull request, I just fix the issue reported in the bug ticket, which fixed with returning the right `lastSortValues`.

We still need a second fix (follow-up https://github.com/camunda/camunda/issues/26080 (still in draft)) as:
Currently in the [search result transformer](https://github.com/camunda/camunda/blob/94e8ae400f328a1fb49ea0882242bd8494f40c53/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/query/SearchQueryResultTransformer.java#L33-L38), we get only the sortValues of the last item of the page.

We also need to get the sortValues of the first item to pass it into `firstSortValues` returned in the response, to be used to fetch the previous page

## Checklist

- [ ] for CI changes:

## Related issues

closes #26030

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/ProcessDefinitionQueryTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/ProcessDefinitionQueryTest.java
@@ -88,6 +88,76 @@ public class ProcessDefinitionQueryTest {
   }
 
   @Test
+  void shouldPaginateWithSortingByProcessDefinitionKey() {
+    // given
+    final var resultAll =
+        zeebeClient
+            .newProcessDefinitionQuery()
+            .sort(s -> s.processDefinitionKey().desc())
+            .send()
+            .join();
+
+    // when
+    final var firstPage =
+        zeebeClient
+            .newProcessDefinitionQuery()
+            .sort(s -> s.processDefinitionKey().desc())
+            .page(p -> p.limit(1))
+            .send()
+            .join();
+    final var secondPage =
+        zeebeClient
+            .newProcessDefinitionQuery()
+            .sort(s -> s.processDefinitionKey().desc())
+            .page(p -> p.limit(1).searchAfter(firstPage.page().lastSortValues()))
+            .send()
+            .join();
+
+    // then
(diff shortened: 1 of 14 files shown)


nathansandi (approved):
LGTM

ralfpuchert (approved):

PR #26083, opened 2024-12-13
feat: Migrate Optimize to Spring MVC #25377

### LINKED ISSUE #25377: Assess compatibility between Jersey and MVC
Engineering DRI: @oleksandr-kriuchenko-lohika

### Context

Currently, Optimize is defining its endpoint by using Jersey. This is an implementation that allows the use of `@Path(...)` annotation to build the various endpoints.

The rest of the applications are using the standard MVC approach.

While nothing prevents these 2 technologies to coexist and work together, as we are going to merge Optimize with the rest of the applications as a single-application program, there is the risk that some endpoints have shared prefixes that might be handled by MVC and by Jersey at the same time.

We should understand if the Jersey configuration of Optimize is compatible with the MVC configuration of the other applications, and, if that's not the case, we should apply a strategy to resolve the conflicts (e.g., migrating from Jersey to MVC or making sure that the endpoints handled by Jersey are also not claimed by the standard MVC approach).

### Acceptance Criteria

[ ] Optimize endpoints can coexist without clashing with the rest of the endpoints, in the single-application context
[ ] The error mappers are translated in MVC and work as expected
[ ] All of the end-to-end tests pass

### Links

* Part of https://github.com/camunda/product-hub/issues/2289

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/26083
```

## Task Lifecycle

#### Engineering Review

#### QA Review

### Completion

oleksandr-kriuchenko-lohika:
## Description

Migrate Optimize to Spring MVC

## Checklist

- [ ] for CI changes:

## Related issues

closes #25377


optimize/backend/src/main/java/io/camunda/optimize/rest/DefinitionRestService.java
```diff
@@ -212,43 +197,34 @@ public List<TenantWithDefinitionsResponseDto> getDefinitionsGroupedByTenant(
...
-        responseBuilder.entity(decisionDef.getDmn10Xml());
-        break;
-      default:
-        throw new BadRequestException("Unknown DefinitionType:" + type);
+      addNoStoreCacheHeader(bodyBuilder);
     }
 
-    return responseBuilder.build();
+    final String xml =
+        switch (type) {
```
buccarel, line 222: Does't a switch without default branch go against the checkstyle rules?

buccarel (approved):

optimize/backend/src/main/java/io/camunda/optimize/rest/DefinitionRestService.java
matthewBearCamunda: A default here isn't necessary because the enum only has two values, and both are handled here. A default case would never be executed

optimize/backend/src/main/java/io/camunda/optimize/rest/security/ccsm/CCSMAuthenticationCookieFilter.java
```diff
@@ -123,6 +123,15 @@ protected Object getPreAuthenticatedCredentials(final HttpServletRequest request
         .orElseGet(() -> AuthCookieService.getAuthCookieToken(request).orElse(null));
   }
 
+  private Optional<String> validToken(final String token) {
```
matthewBearCamunda, line 126: ❓ Why add this method and its usages? This is adding token verification to places it did not exist previously

optimize/backend/src/main/java/io/camunda/optimize/service/security/AuthCookieService.java
```diff
@@ -148,17 +141,14 @@ public NewCookie createCookie(
...
-        .path(cookiePath)
-        .domain(null)
-        .version(1)
-        .comment(null)
-        .maxAge(-1)
-        .expiry(expiresAt)
-        .secure(isSecureScheme(requestScheme))
-        .httpOnly(true)
-        .build();
+    final int maxAge = (int) (expiresAt.toInstant().toEpochMilli() - System.currentTimeMillis());
```
matthewBearCamunda, line 144: ❓ The `maxAge` was previously hard coded to `-1` for this method. Why change to a calculated value?

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/rest/pagination/PaginationScrollableRequestDto.java
```diff
@@ -12,35 +12,26 @@
...
-
-  @QueryParam(QUERY_LIMIT_PARAM)
   @Min(0)
   @DefaultValue("1000")
   @Max(MAX_RESPONSE_SIZE_LIMIT)
   protected Integer limit;
 
-  @QueryParam(QUERY_SCROLL_ID_PARAM)
-  protected String scrollId;
+  protected String searchRequestId;
```
matthewBearCamunda, line 22: ❓ The BE changes for this DTO change look correct, but does the FE need changing as well?

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/rest/pagination/PaginationScrollableRequestDto.java
```diff
@@ -12,35 +12,26 @@
...
 
-  @QueryParam(QUERY_SCROLL_TIMEOUT_PARAM)
   @Min(60)
   @DefaultValue("120")
-  protected Integer scrollTimeout;
+  protected Integer paginationTimeout;
```
matthewBearCamunda, line 26: ❓ Same as above, does the FE need changing to use the new variable name?

matthewBearCamunda (changes requested):
The PR is quite big, and I'd like to do another full review before being able to approve. I found some minor style adjustments that need to be made. However, most of my comments are questions around the implementation & TODOs

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/rest/pagination/PaginationScrollableRequestDto.java
matthewBearCamunda: Solved my own question here. It was renamed here due to the removal of the @QueryParam annotation, and the ultimate DTO that is used will have the same name for fields

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/rest/pagination/PaginationScrollableRequestDto.java
matthewBearCamunda: same answer as above

optimize/client/public/static/redirect.html
```diff
@@ -1,29 +0,0 @@
-<!DOCTYPE html>
```
buccarel, line 1: why was this deleted?

buccarel (changes requested):
Left a question about the redirect

optimize/backend/src/main/java/io/camunda/optimize/rest/providers/ResponseTimezoneFilter.java
```diff
@@ -26,17 +24,16 @@
...
 
-  @Override
-  public void filter(
-      final ContainerRequestContext requestContext,
-      final ContainerResponseContext responseContext) {
-    ObjectWriterInjector.set(new DateMod(extractTimezone(requestContext)));
+  public void filter(final HttpServletRequest request) {
+    // TODO migrate to Spring MVC
+    // TODO check if this works
+    ObjectWriterInjector.set(new DateMod(extractTimezone(request)));
```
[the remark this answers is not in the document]
matthewBearCamunda: This is the JIRA ticket for when this filter was implemented

https://jira.camunda.com/browse/OPT-3832

optimize/backend/src/main/java/io/camunda/optimize/rest/converters/StringToDefinitionTypeConverter.java
```diff
@@ -0,0 +1,20 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.optimize.rest.converters;
+
+import io.camunda.optimize.dto.optimize.DefinitionType;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.stereotype.Component;
+
+@Component
+public class StringToDefinitionTypeConverter implements Converter<String, DefinitionType> {
```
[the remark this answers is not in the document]
oleksandr-kriuchenko-lohika: This is part of ~~christmas~~ spring mvc magic - it is to allow parsing request parameters (coming as strings) into DefinitionType enum. Please check https://www.baeldung.com/spring-enum-request-param for details on how this works

optimize/backend/src/main/java/io/camunda/optimize/rest/providers/NotAuthorizedExceptionMapper.java
```diff
@@ -39,8 +37,9 @@ public Response toResponse(final NotAuthorizedException notAuthorizedException)
     LOG.debug("Mapping NotAuthorizedException");
 
     return Response.status(Response.Status.UNAUTHORIZED)
-        .type(MediaType.APPLICATION_JSON_TYPE)
-        .cookie(cookieService.createDeleteOptimizeAuthNewCookie(true))
+        .type(MediaType.APPLICATION_JSON_VALUE)
+        // TODO migrate to Spring MVC
```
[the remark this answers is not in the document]
oleksandr-kriuchenko-lohika: yes, but probably as a separate PR for migrating all the exception mappers to MVC (both to keep this PR smaller and to unblock integration into single app)

optimize/backend/src/main/java/io/camunda/optimize/rest/providers/ResponseTimezoneFilter.java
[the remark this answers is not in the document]
oleksandr-kriuchenko-lohika: Yeah, I didn't manage to check it. Considering we aimed to run a regression testing for the MVC changes, I hoped we could test it as part of it.

optimize/backend/src/main/java/io/camunda/optimize/rest/providers/ResponseTimezoneFilter.java
```diff
@@ -26,17 +24,16 @@
 
 @Provider
 @Component
-public class ResponseTimezoneFilter implements ContainerResponseFilter {
```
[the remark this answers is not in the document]
oleksandr-kriuchenko-lohika: Oh, yes, good point - it is not a regular filter and most of all it will not work with this change. Not sure whether we should make it work first or go without it and fix it during the single app testing as suggested by @RomanJRW

optimize/backend/src/main/java/io/camunda/optimize/rest/security/ccsm/CCSMAuthenticationCookieFilter.java
oleksandr-kriuchenko-lohika: Yes, that's correct - I've ran into issue without it. The problem is that on token expiration we are removing it from cookies, but still pass it further down to this place for processing (there was no option to replace a token in the request with a refreshed one). So to prevent using the outdated token (which was causing further errors) I've added a token verification here. Thanks for asking this - I wanted to explain this change, but forgot. Maybe there still is some more elegant solution for this, but I didn't manage to find any in a reasonable time.

optimize/backend/src/main/java/io/camunda/optimize/service/security/AuthCookieService.java
oleksandr-kriuchenko-lohika: Good question - I could barely recall why ) The point is that previously there was this strange setup:
`        .maxAge(-1)
        .expiry(expiresAt)`
where we considered expiresAt, but only as a cookie expiration, while we were setting maxAge to -1.
Tomcat cookies do not support expiration setting (it is deprecated), but only maxAge. So I've got to translate the expiration into maxAge.

optimize/client/public/static/redirect.html
oleksandr-kriuchenko-lohika: To me it looked more like a workaround. I hope that as part of migration to MVC I've addressed the root cause of the problem (attempt to authorize access to static resources), so hopefully we no longer need it. Yet if there is a requirement to authorize access to static resources then we may want to get it back 🤞

optimize/backend/src/main/java/io/camunda/optimize/rest/providers/ResponseTimezoneFilter.java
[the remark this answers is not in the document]
oleksandr-kriuchenko-lohika: Adapted to MVC now.

optimize/backend/src/main/java/io/camunda/optimize/rest/providers/NotAuthorizedExceptionMapper.java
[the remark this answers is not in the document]
oleksandr-kriuchenko-lohika: Adapted to MVC now.

optimize/client/public/static/redirect.html
oleksandr-kriuchenko-lohika: Verified that sameSite=strict is really not compatible with OIDC flow (which uses a redirect to a different site for authentication) and returned this workaround back

buccarel:
The application seems to be working, even in spite of the failed end-to-end tests (usually only 1 fails, and the others succeed. In this case I see around 10 failing). Do you happen to know why they're failing? I tried those failures locally and they seem to work (I tried renaming a collection).

If we don't know why, we could QA this anyways and merge based on the result of the QA.

buccarel (approved):

matthewBearCamunda (approved):

PR #26086, opened 2024-12-13
ci: disable cache restoration using chatops command

### LINKED ISSUE #24303: CI: add ChatOps command to enable/disable GHA cache usage on a PR
## Description

For some debugging cases it can be relevant to test CI pipelines with having the GHA cache restore _disabled_ to get the true runtime of jobs assuming a cold cache. That scenario is important sometimes to avoid problems since GitHub clears caches every night and our pipelines might take too long with a cold cache.

This has to be done manually but could be automated (adding `REMOVEME` commits temporarily) via the same approach as in #22483.

Documentation should be added to https://github.com/camunda/camunda/wiki/CI-&-Automation#chatops

## Hints

* @cmur2 is available for questions, pairing and code reviews

Kerruba:
## Description

With this PR we want to introduce the ability for developers to temporarily test workflows without cache restoration via some chat commands, `/disable-cache` and `/enable-cache`.

- `/ci-disable-cache` adds a new `ci:no-cache` label from the PR and creating an empty commit to trigger workflows automatically

- `/ci-enable-cache` complement the first command, by removing the `ci:no-cache` label from the PR and creating an empty commit to trigger workflows automatically 

Our cache functions, `setup-maven-cache` on the monorepo and `setup-yarn-cache` in the `infra-global-github-actions` repository have been updated to verify if PR has the `ci:no-cache` label. In case the label is present, the restoration cache step is skipped.

- PR for `setup-yarn-cache`: https://github.com/camunda/infra-global-github-actions/pull/312

A poc repository with the behavior has been created to test the functionality. Check [this PR](https://github.com/Kerruba/poc-disable-cache-command/pull/5) for an example.

## Checklist

- [ ] for CI changes:

## Related issues

closes #24303


.github/actions/is-cache-disabled/action.yml
```diff
@@ -0,0 +1,27 @@
+---
```
cmur2, line 1: 🔧  As you mentioned we should re-use the implementation of https://github.com/camunda/infra-global-github-actions/pull/312 when we move forward with this approach.

.github/actions/is-cache-disabled/action.yml
```diff
@@ -0,0 +1,27 @@
...
+    description: Output whether the current PR has a label 'ci:no-cache' to disable cache restoration
+    value: ${{ steps.detect-cache-disabled.outputs.is-cache-disabled == 'true' }}
+
+runs:
+  using: composite
+  steps:
+
+  - id: detect-cache-disabled
+    env:
+      IS_CACHE_DISABLED: ${{ contains(github.event.pull_request.labels.*.name, 'ci:no-cache') }}
```
cmur2, line 23: ❓ Does this work as expected* when not on a PR?

*I'd expect that e.g. `on: push` to main or stable branches or `on: schedule` etc the cache restoration is **enabled** (should be the default) and this action should return `is-cache-disabled == false`.

.github/actions/setup-maven-cache/action.yaml
```diff
@@ -83,9 +90,10 @@ runs:
         key: ${{ env.MAVEN_CACHE_KEY }}
         restore-keys: |
           ${{ env.MAVEN_CACHE_KEY_PREFIX }}
+
     - name: Restore maven cache
-      # Restore cache (but don't save it) if we're not on main or stable/* branches
-      if: ${{ !(startsWith(github.ref_name, 'stable/') || github.ref_name == 'main') }}
+      # Restore cache (but don't save it) if we're not on main or stable/* branches and cache is not disabled
+      if: ${{ !(startsWith(github.ref_name, 'stable/') || github.ref_name == 'main') && steps.is-cache-disabled.outputs.is-cache-disabled == 'false' }}
```
cmur2, line 96: > and cache is not disabled

❓ Did you consider naming the helper action and variables to avoid the double-negation? So have an `is-cache-enabled` action and check for `steps.is-cache-disabled.outputs.is-cache-disabled == 'true'` to avoid mental complexity of double negation ("So this work when the cache **is not** disabled aka it **is** enabled...").

If possible I'd go for an `is-cache-enabled` action.

.github/workflows/chatops-command-dispatch.yml
```diff
@@ -89,6 +89,8 @@ jobs:
           issue-type: pull-request
           commands: |
             ci-problems
+            disable-cache
+            enable-cache
```
cmur2, line 93: 🔧 I'd prefix with `ci-` to make scope of the commands clear (sorry I didn't catch this earlier in your proposal).

.github/actions/setup-maven-cache/action.yaml
```diff
@@ -74,6 +74,13 @@ runs:
           echo "MAVEN_CACHE_KEY=${KEY_PREFIX}-full-${HASH_FILES}" >> $GITHUB_ENV
           echo "MAVEN_CACHE_KEY_PREFIX=${KEY_PREFIX}-full" >> $GITHUB_ENV
         fi
+
+      # Check if PR contains a label to disable the cache
+    - name: Check if cache is disabled
+      if: ${{ !(startsWith(github.ref_name, 'stable/') || github.ref_name == 'main') }}
```
cmur2, line 80: ❓ I assume this condition is to make sure the `is-cache-disabled` action is only called on PRs? I'd prefer this logic to be residing in the `is-cache-disabled` inside that action so it always gives the "right" result, also `on: scheduled` etc workflows.

.github/actions/is-cache-disabled/action.yml
Kerruba: I've updated my poc and created two new workflows, one with `on: push` events (Push workflow) and one with `on: workflow_dispatch` (Dispatch workflow) (I will assume it's the same for `on: schedule`).

For the same commit, the Push workflow restore cache correctly while the CI workflow doesn't. That's the expected behavior based on the 3 conditions I'm using to disable cache (is a pr event, the `ci:no-cache` label is present and the event is on a "feature-branch" - not main nor stable/*) 
- https://github.com/Kerruba/poc-disable-cache-command/pull/5/checks?check_run_id=34603694007

When cache is enabled, both Push and CI workflows restore cache correctly
- https://github.com/Kerruba/poc-disable-cache-command/pull/5/checks?check_run_id=34604814824

.github/actions/is-cache-disabled/action.yml
cmur2: I approved https://github.com/camunda/infra-global-github-actions/pull/312 so you are free to merge it and apply the refactoring

cmur2 (approved):
Looks good

PR #26095, opened 2024-12-15
fix(mvc-auth-commons): The transformed jar is now used by intelliJ to start the project

### LINKED ISSUE #17422: mvc-auth-commons requires transformation
### Description

We are currently using 1.11.0 of com.auth0:mvc-auth-commons dependency. This dependency uses legacy servlet types instead of jakarta, so we are using the maven transformer plugin to adjust the bytecode at build time:

https://github.com/camunda/zeebe/blob/d058cdbf88ce8a613747fd2c44c8f69abe2ca4fb/operate/mvc-auth-commons/pom.xml#L95

When operate is built and run in intellij, there are issues with resolving these transformed types (only in Intellij, maven builds from the command line work fine), requiring the user to manually add the transformed mvc-auth-commons jar as a global dependency every time they set up a new branch or update dependencies, which is impacting more users now after the move to the mono-repo. See related slack convo:

https://camunda.slack.com/archives/C06HTSPD5AP/p1712593059050499

We need to determine if there is a cleaner way to accomplish this that does not require manual steps for the users, or if updating to the 2.0 version of the library will fix this (and how much effort would be required to do that). May be a question for tasklist as well to see if they are on 1.0 and if so, how do they handle this.

### Additional Information

### Acceptance Criteria

mathias-vandaele:
## Description

The project compiled successfully with Maven but failed to start in IntelliJ, causing issues for newcomers attempting to run the project locally. The problem stemmed from the setup of two modules in the repository, both of which included the same submodule responsible for transforming javax.* dependencies into jakarta.* using a transformer plugin.

In the Maven configuration, the transformed JAR was correctly picked up because Maven resolves dependencies across modules as expected. However, IntelliJ relied on the POM of the submodule itself to determine allowed dependencies, which led it to incorrectly use the untransformed JAR instead of the transformed one. This discrepancy between Maven and IntelliJ caused runtime issues.

To address this, I:

- Extracted the submodules responsible for transformation into a single, unified module at the root of the project.
- Configured the unified transforming module to handle the javax.* to jakarta.* transformation for both original modules.
- Updated the sibling modules to depend on the transformed artifacts produced by this unified module.

By separating the transformation logic into an independent module, IntelliJ no longer misinterprets the dependency structure, allowing it to function correctly in local development.

The drawback of this approach is that the unified transforming module cannot be related to other modules in the project architecture. This is necessary to prevent IntelliJ from using the POM of the "transforming" module to infer dependencies, which would reintroduce the original issue. As a result, the unified module must remain isolated in terms of module relationships, even though its output can be used wherever needed.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


houssain-barouni (changes requested):
🚀 Thanks @mathias-vandaele for taking care of this
you need to declare the new mvc-auth-commons module in the root [pom.xml](https://github.com/camunda/camunda/blob/main/pom.xml)

npepinpe (approved):
If I could promote you for this PR I would :heart:

npepinpe:
@mathias-vandaele - what should I buy to bribe you into finishing/merging this PR asap?

mathias-vandaele:
> 🚀 Thanks @mathias-vandaele for taking care of this you need to declare the new mvc-auth-commons module in the root [pom.xml](https://github.com/camunda/camunda/blob/main/pom.xml)

Hey, as I mentioned here @houssain-barouni :

`The drawback of this approach is that the unified transforming module cannot be related to other modules in the project architecture. This is necessary to prevent IntelliJ from using the POM of the "transforming" module to infer dependencies, which would reintroduce the original issue. As a result, the unified module must remain isolated in terms of module relationships, even though its output can be used wherever needed.`

To prevent this, the transforming module must be isolated from the project’s dependency hierarchy. While its output artifact can still be consumed by other modules, it cannot be part of the root POM. If it’s included, IntelliJ will prioritize the module's POM over the final transformed JAR, breaking the build experience for developers.

lenaschoenburg:
@mathias-vandaele Can we help you getting this merged somehow? :heart_hands:

houssain-barouni:
>The drawback of this approach is that the unified transforming module cannot be related to other modules in the project architecture. This is necessary to prevent IntelliJ from using the POM of the "transforming" module to infer dependencies, which would reintroduce the original issue. As a result, the unified module must remain isolated in terms of module relationships, even though its output can be used wherever needed.

@mathias-vandaele does it mean that we should declare the module in the root POM? and that we need to revert https://github.com/camunda/camunda/pull/26095/commits/e7a6aef4ccf2f55fb617826f41ed8c9a9ed603ba ?

I am not sure with this that the module will be released with `mvn release` goal, unless we do something custom

mathias-vandaele:
Hello @houssain-barouni,

Adding it to the module list will always generate the same issue, but I understand that not having it there poses a bunch of issues as well.

this module is just of module of transformation, the best would be to create another repo containing this module and release the artefact, I don't think this will be updated often. 

In this case you can use `camunda-mvc-auth-commons` where you want as long as it is in the remote repository. IntelliJ will always infer the dependency tree otherwise, without really looking at what is installed inside `.m2`

lev-safro:
Hello @mathias-vandaele !

I went through this [part](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md#run-via-intellij) of the `CONTRIBUTING.md` and I was wondering if this part is still relevant:

> To address this, open the Project Structure dialog and add the following 2 libraries to the classpath of your JDK (<version> being the current snapshot version, e.g. 8.8.0):
> $HOME/.m2/repository/io/camunda/operate-mvc-auth-commons/<version>-SNAPSHOT/operate-mvc-auth-commons-<version>-SNAPSHOT.jar
> $HOME/.m2/repository/io/camunda/tasklist-mvc-auth-commons/<version>-SNAPSHOT/tasklist-mvc-auth-commons-<version>-SNAPSHOT.jar

I checked my `M2`:
` ll | grep mvc-auth-commons`

And there were no such jars in this dir
`/.m2/repository/io/camunda`

Without this step I was able to build and run project locally from IntelliJ.

Shall we update the documentation and close this PR?

Thank you!

lenaschoenburg:
@lev-safro I think this is only needed for older stable branches now, i.e. `stable/8.6` and `stable/8.7`. I think we could clarify that in the docs. The artifacts will only be available once you have `mvn install`ed at least once on these branches.

mathias-vandaele:
I will be closing this PR, it is not relevant anymore

PR #26108, opened 2024-12-16
[CPT] feat: Migrate to the Camunda Exporter

saig0:
## Description

Replace the Elasticsearch exporter with the Camunda exporter in the Camunda container for CPT.

Bug in the Operate API with the Camunda exporter: https://github.com/camunda/camunda/issues/26115.

## Related issues

This migration is required to use the API since the Operate importers will be removed with 8.7.


testing/camunda-process-test-java/src/test/java/io/camunda/process/test/api/CamundaProcessTestExtensionIT.java
```diff
@@ -56,7 +56,6 @@ void shouldCreateProcessInstance() {
     // then
     CamundaAssert.assertThat(processInstance)
         .isActive()
-        .hasCompletedElements("start")
```
berkaycanbc, line 59: 🔧 Let's mention affected tests in #26115 and they should be put back along with the fix.

berkaycanbc (approved):
LGTM 👍 

Added one comment to make sure we update the test along with the bug fix.

testing/camunda-process-test-java/src/test/java/io/camunda/process/test/api/CamundaProcessTestExtensionIT.java
berkaycanbc: Then, it needs to be put back when CPT starts to use v2?

PR #26119, opened 2024-12-16
fix: Default Admin role not showing up in Identity UI

### LINKED ISSUE #26098: Admin role missing from Users `Assigned roles` UI/is not found in service calls for a members roles
### Description
The `Admin` role created on startup is not shown in the list of roles assigned to a user, additionally as part of testing https://github.com/camunda/camunda/pull/25749 I noticed that the `Admin` role is not returned in the service layer when finding roles assigned to a member.

In comparison, when I create a new role and assign it to a user then this is both shown in the UI and also returned in the service layer.

### Acceptance criteria
- [ ] The Admin role is shown in the UI when viewing the roles assigned to a user
- [ ] The admin role is included in the authentication layer when finding roles for the authenticated principle

remcowesterhoud:
## Description

By making sure we write the event with the right key, it should enable the exporter to handle it the same way when we user is added to a role manually.

## Checklist

- [ ] for CI changes:

## Related issues

closes #26098


Ben-Sheppard (approved):
I tested this locally and it appears to work as expected! Thanks for fixing this so quick

PR #26125, opened 2024-12-16
[CPT] fix: Don't use deprecated/internal ReflectionUtils

saig0:
## Description

Replace calls to deprecated or internal methods of JUnit's ReflectionUtils. These methods may be not available in newer versions of the library.

The same changes were applied to ZPT: https://github.com/camunda/zeebe-process-test/pull/1280

## Related issues

Related to https://github.com/camunda/zeebe-process-test/issues/1278.


berkaycanbc (approved):
LGTM 🚀

PR #26130, opened 2024-12-16
feat: implement purge in RdbmsExporter #26129

### LINKED ISSUE #26129: Support purge in RdbmsExporter
To enable the CPT project to purge the database, the RdbmsExporter must support the Purge of the connected database.

stoerti:
## Description

Implements the purge in the RdbmsExporter

## Related issues

closes #26129

### DIFF
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/RdbmsPurger.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/RdbmsPurger.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.db.rdbms.write.service;
+
+import io.camunda.db.rdbms.sql.PurgeMapper;
+import java.util.List;
+
+public class RdbmsPurger {
+
+  private static final List<String> TABLE_NAMES =
+      List.of(
+          "EXPORTER_POSITION",
+          "VARIABLE",
+          "FLOW_NODE_INSTANCE",
+          "PROCESS_INSTANCE",
+          "PROCESS_DEFINITION",
+          "CANDIDATE_USER",
+          "CANDIDATE_GROUP",
+          "USER_TASK",
+          "INCIDENT",
+          "DECISION_INSTANCE_INPUT",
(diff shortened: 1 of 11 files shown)


npepinpe (commented):
Just a question since I didn't follow the exporter evolution much - am I right to assume that if I want to run multiple C8 on the same RDBMS, I would define a different database name for each essentially? Would dropping that DB and recreating the schema not be enough/safer? Or do we expect users to have other things on that DB?

stoerti:
> Would dropping that DB and recreating the schema not be enough/safer? Or do we expect users to have other things on that DB?
We have thought of that too. I would expect customers to have a specific DB for Camunda but would not totally rely on that (C7 also often ran in the same schema as the rest of the service).

A "DROP TABLE" followed by automatic Liquibase setup could be forbidden by the DB user running Camunda (this is why we have issue #23822).

A TRUNCATE should also be very fast since the RDBMS also simply deletes all data/indices without checks and we can skip the creation afterwards.

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/RdbmsPurger.java
```diff
@@ -0,0 +1,52 @@
...
+public class RdbmsPurger {
+
+  private static final List<String> TABLE_NAMES =
+      List.of(
+          "EXPORTER_POSITION",
+          "PROCESS_DEFINITION",
+          "PROCESS_INSTANCE",
+          "VARIABLE",
+          "FLOW_NODE_INSTANCE",
+          "USER_TASK",
```
p-wunderlich, line 22: Since we don't have foreign keys, it doesn't matter, but for me it feels uncomfortable to truncate first the main table and then the children tables. :D Can we first truncate Candidate stuff, and then user_task? :D (Same for others)

p-wunderlich (approved):

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/RdbmsPurger.java
```diff
@@ -0,0 +1,52 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.db.rdbms.write.service;
+
+import io.camunda.db.rdbms.sql.PurgeMapper;
+import java.util.List;
+
+public class RdbmsPurger {
+
+  private static final List<String> TABLE_NAMES =
```
npepinpe, line 15: @stoerti - A little late, sorry for that, but just a thought: this means whenever a new table is added/removed in the schema, we have to remember to update this list, correct? Seems easy to forget and break something unsuspectingly.

Is there anything we can do to improve this? For example:
- Pulling the table names from some centralized schema file (?), or
- Writing a test which breaks if there is a missing and/or extra table

PR #26143, opened 2024-12-16
deps: remove explicit dependency on reactor

npepinpe:
## Description

This PR removes an old explicit dependency on the reactor project, leftover from when we were using WebFlux instead of the servlet API.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -142,7 +142,6 @@
     <version.postgresql>42.7.4</version.postgresql>
     <version.h2>2.3.232</version.h2>
     <version.aws-signing>2.3.1</version.aws-signing>
-    <version.reactor-netty>1.1.23</version.reactor-netty>
     <version.tc-keycloak>3.5.1</version.tc-keycloak>
     <version.jakarta-activation>2.1.3</version.jakarta-activation>
     <version.jakarta.json-api>2.1.3</version.jakarta.json-api>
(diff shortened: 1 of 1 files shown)


npepinpe:
@tmetzke - super small thing, just realized we had this unused dep. I'm tagging you though because I saw we still use some WebFlux stuff in the REST API tests :thinking: We use the servlet API everywhere now, should we consider using a different thing? I'm not a Spring expert so I don't know if it makes sense either, just wanted to mention it

tmetzke:
> @tmetzke - super small thing, just realized we had this unused dep. I'm tagging you though because I saw we still use some WebFlux stuff in the REST API tests 🤔 We use the servlet API everywhere now, should we consider using a different thing? I'm not a Spring expert so I don't know if it makes sense either, just wanted to mention it

To clarify: 
1. We can remove this here at proposed, even if we use the WebFlux stuff in tests, right? I assume that is a different dependency?
2. If I'm not mistaken, the WebFlux stuff we use is the WebClient for test request execution? If so, we can look into that but we preferred that over a more basic client in the tests due to the async response nature (completable futures) in some endpoints. The WebClient handles this nicely without much fuzz, some more basic ones we tried required a lot more code to make it work and also manual differentiation between async and sync endpoints. Happy for suggestions here though, of course.

tmetzke (approved):

houssain-barouni:
We need to backport this to stable/8.7 as a possible fix for https://jira.camunda.com/browse/SUPPORT-26744

Issue #25992, opened 2024-12-12
Authorizations are applied to search queries when using OIDC


Ben-Sheppard:

tmetzke:
Leaving this to the identity stream and unassigning @nathansandi here.
@romansmirnov or @megglos will take care of assigning this accordingly.

npepinpe:
This is done right?

ThorbenLindhauer:
Yes, this topic is implemented (and the implementation specifics are different from the description here). General direction to look at for the tickets are https://github.com/camunda/camunda/issues/26799 and https://github.com/camunda/camunda/issues/26785.

Issue #25780, opened 2024-12-09
I can use the profile `auth-oidc` to enable OIDC based authentication


Ben-Sheppard:
### Description
OIDC is currently supported in the web applications and is how we offer support for connecting to external providers. This support should be present and functioning in the `camunda-authentication` module

### Tasks
- [x] A OIDC focused web security configuration exists
- [ ] I can configure the OIDC connection using Camunda abstracted properties i.e. `camunda.auth.oidc.issuer-url`
- [ ] An OIDC user details service exists extending the `CamundaUserDetails` class
  - [ ] hint: the basic authentication user details class offers suggestions on getting application access from services, here we need to use mapping rules
- [ ] When in a SAAS context, the organization of the deployment is checked against the organizations on the users token
  - [ ] The orgs are in the claim `https://camunda.com/orgs` for a user token and `https://camunda.com/clusterId` for a M2M token
  - [ ] The organization ID of the cluster is passed into the single application via configuration in the property `camunda.security.organizationId` (proposal) (https://github.com/camunda-cloud/camunda-operator/issues/3571)
- [ ] Accessing the `/v2/authentication/me` endpoint returns the expected data for OIDC

romansmirnov:
> The organization ID of the cluster is passed into the single application via configuration in the property camunda.environmnent.organizationId (proposal) 

I propose having this property as part of the security configuration unless the organization id must be used somewhere else as well.

Ben-Sheppard:
> > The organization ID of the cluster is passed into the single application via configuration in the property camunda.environmnent.organizationId (proposal)
> 
> I propose having this property as part of the security configuration unless the organization id must be used somewhere else as well.

I agree that makes sense @romansmirnov - updated

PR #26144, opened 2024-12-16
Map OIDC tokens to camunda principals for session authentication

emulbreh:
## Description

This PR is a PoC for OIDC session login. It should work for both self-managed and SaaS environments. It contains the following changes:

* a `CamundaUserService` implementation for OIDC
* a `OidcUserService` implementation 
* support for `camunda.security.*` scoped configuration for OIDC (https://github.com/camunda/camunda/pull/26053)
* organization claim checks when `camunda.security.organization-id` is provided
* support for a context path in identity (https://github.com/camunda/camunda/pull/26347)
* updated paths for login, logout and current user endpoints in operate and tasklist client

The following configuration is necessary:
```
SPRING_PROFILES_ACTIVE=auth-oidc,identity,tasklist,operate
CAMUNDA_SECURITY_ORGANIZATION-ID=
CAMUNDA_SECURITY_AUTH_OIDC_CLIENT-ID=
CAMUNDA_SECURITY_AUTH_OIDC_CLIENTSECRET=
CAMUNDA_SECURITY_AUTH_OIDC_ISSUER-URI=
CAMUNDA_SECURITY_AUTH_OIDC_REDIRECTURI=
```

### Related
* https://github.com/camunda/camunda/issues/25992
* https://github.com/camunda/camunda/issues/25780


authentication/src/main/java/io/camunda/authentication/CamundaOidcUserService.java
```diff
@@ -0,0 +1,74 @@
...
+    final Set<Long> mappingKeys =
+        mappings.stream().map(MappingEntity::mappingKey).collect(Collectors.toSet());
+    if (mappingKeys.isEmpty()) {
+      LOG.debug("No mappings found for these claims: {}", claims);
+    }
+
+    return new CamundaOidcUser(
+        oidcUser,
+        mappingKeys,
+        new AuthenticationContext(
```
nathansandi, line 66: 🤔  I think if the mappingKeys come as empty, there is an pottential error here. I have tested with my OIDC user

```
11:34:53.136 [] [http-nio-0.0.0.0-8080-exec-5] [] ERROR org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/].[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception
java.lang.NullPointerException: Expected a non-null query parameter for the hasChild query.
```

Would be the case to incldue a nullcheck over here?

authentication/src/main/java/io/camunda/authentication/filters/TenantRequestAttributeFilter.java
```diff
@@ -56,24 +55,17 @@ protected void doFilterInternal(
...
-            "cannot find tenants: unsupported principal type {}", principal.getClass().getName());
-        yield null;
-      }
-    };
+    if (!(principal instanceof final Authentication auth)) {
+      LOG.error(
+          "cannot find tenants: unsupported principal type: {}", principal.getClass().getName());
+      return null;
+    }
+    if (!(auth.getPrincipal() instanceof final CamundaPrincipal camundaPrincipal)) {
```
romansmirnov, line 63: ❓ When testing with a client (not user) and enabled tenancy, any request fails with 

```
ERROR io.camunda.authentication.filters.TenantRequestAttributeFilter - cannot find tenants: unsupported principal type: org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken
```

Will the client case be handled separately?

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
```diff
@@ -198,7 +198,9 @@ private List<Long> getAuthenticatedUserRoleKeys() {
     if (requestAuthentication != null) {
       final Object principal = requestAuthentication.getPrincipal();
       if (principal instanceof final CamundaUser authenticatedPrincipal) {
```
romansmirnov, line 200: ❓ What about the `CamundaOidcUser` and clients?

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
```diff
@@ -194,13 +200,29 @@ public SearchQueryResult<TenantEntity> searchTenants(final TenantQuery filter) {
...
+  }
+
+  @Override
+  public SearchQueryResult<GroupEntity> searchGroups(final GroupQuery query) {
+    return new SearchClientBasedQueryExecutor(
+            searchClient,
+            transformers,
+            new DocumentAuthorizationQueryStrategy(this),
+            securityContext)
+        .search(query, io.camunda.webapps.schema.entities.usermanagement.GroupEntity.class);
```
romansmirnov, line 215:
```suggestion
      return getSearchExecutor()
        .search(query, io.camunda.webapps.schema.entities.usermanagement.GroupEntity.class);
```

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
```diff
@@ -194,13 +200,29 @@ public SearchQueryResult<TenantEntity> searchTenants(final TenantQuery filter) {
...
+
+  @Override
+  public List<GroupEntity> findAllGroups(final GroupQuery query) {
     return new SearchClientBasedQueryExecutor(
             searchClient,
             transformers,
             new DocumentAuthorizationQueryStrategy(this),
             securityContext)
-        .search(filter, io.camunda.webapps.schema.entities.usermanagement.GroupEntity.class);
+        .findAll(query, io.camunda.webapps.schema.entities.usermanagement.GroupEntity.class);
```
romansmirnov, line 225:
```suggestion
      return getSearchExecutor()
        .findAll(query, io.camunda.webapps.schema.entities.usermanagement.GroupEntity.class);
```

authentication/src/main/java/io/camunda/authentication/service/BasicCamundaUserService.java
```diff
@@ -7,36 +7,46 @@
...
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Service;
 
 @Profile("auth-basic")
 @Service
 public class BasicCamundaUserService implements CamundaUserService {
+  private Optional<CamundaUser> getCurrentCamundaUser() {
+    return Optional.ofNullable(SecurityContextHolder.getContext().getAuthentication())
+        .map(Authentication::getPrincipal)
+        .map(principal -> principal instanceof final CamundaUser user ? user : null);
```
romansmirnov, line 26:
```suggestion
        .filter(CamundaUser.class::isInstance);
```

Probably a matter of taste.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -482,7 +481,9 @@ public static Authentication getAuthentication() {
           instanceof final CamundaUser authenticatedPrincipal) {
         authenticatedUserKey = authenticatedPrincipal.getUserKey();
         authenticatedRoleKeys.addAll(
-            authenticatedPrincipal.getRoles().stream().map(RoleEntity::roleKey).toList());
+            authenticatedPrincipal.getAuthenticationContext().roles().stream()
```
romansmirnov, line 484: ❓ What about the `CamundaOidcUser`?

romansmirnov:
While I am not tagged as a reviewer, I still had a look at the PR. Could you please clarify two questions?

1. In some cases, it checks for `auth instanceof CamundaUser`. Must those checks be adjusted to consider the oidc user?
2. The PR expects that there is a `CamundaPrincipal,` which is only present when a user logs in but not when a client is interacting with the API. Will this be handled separately?

operate/webapp/src/main/java/io/camunda/operate/webapp/security/permission/PermissionsService.java
emulbreh: These instanceof checks all had to be changed to check for CamundaPrincipal instead. Thanks for catching that.

emulbreh:
> While I am not tagged as a reviewer, I still had a look at the PR. Could you please clarify two questions?
> 
> 1. In some cases, it checks for `auth instanceof CamundaUser`. Must those checks be adjusted to consider the oidc user?
> 2. The PR expects that there is a `CamundaPrincipal,` which is only present when a user logs in but not when a client is interacting with the API. Will this be handled separately?

(1) was an oversight, (2) is out of scope for this PR. I'll update the description.

emulbreh:
#26378 will replace this PR.

PR #26145, opened 2024-12-16
feat: change REST OpenAPI specification int64 keys to string

### LINKED ISSUE #26062: [C8 API] Add OpenAPI specification `String` keys support (8.7)
## Description

In order for our API documentation page to display both responses depending on the header, we need to define response objects with String and Numeric keys in the OpenAPI specification.

## Acceptance criteria

- Add OpenAPI responses with String and Long (in64) keys for the required APIs.
- Add OpenAPI response headers linked to the correct response object.
- Change the default response to String keys!
  - Find a new name for these response items so users don't get confused.
- Adjust required code changes & tests.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/26145
```

danielkelemen:
## Description

- Change default keys to string & add new number keys response objects
- Adjust Java Client tests
- Adjust Gateway REST tests

## Related issues

closes https://github.com/camunda/camunda/issues/26062


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/MappingController.java
```diff
@@ -87,13 +87,13 @@ public ResponseEntity<MappingItem> getMapping(@PathVariable final long mappingKe
       path = "/search",
       produces = {MediaType.APPLICATION_JSON_VALUE, MediaType.APPLICATION_PROBLEM_JSON_VALUE},
       consumes = MediaType.APPLICATION_JSON_VALUE)
-  public ResponseEntity<MappingSearchQueryResponse> searchMappings(
+  public ResponseEntity<MappingSearchResponse> searchMappings(
```
danielkelemen, line 90: 💭 It looked like `MappingSearchQueryResponse` was defined but not used in OpenAPI spec, while `MappingSearchResponse` was defined as response but not used here in controller.
Same with `MappingItem` & `MappingResponse`.

clients/java/src/main/java/io/camunda/client/impl/response/ActivatedJobImpl.java
```diff
@@ -211,6 +211,14 @@ private static Long getOrEmpty(final Long value) {
     return value == null ? -1L : value;
   }
 
+  private static Long parseLongOrEmpty(final String value) {
+    try {
+      return Long.parseLong(value);
+    } catch (final NumberFormatException e) {
+      return -1L;
+    }
+  }
```
tmetzke, line 220:
```suggestion
  private static Long parseLongOrEmpty(final String value) {
    value == null ? -1L : Long.parseLong(value);
  }
```

clients/java/src/main/java/io/camunda/client/impl/util/ParseUtil.java
```diff
@@ -0,0 +1,27 @@
...
+package io.camunda.client.impl.util;
+
+public class ParseUtil {
+
+  public static Long parseLongOrNull(final String input) {
+    try {
+      return Long.parseLong(input);
+    } catch (final NumberFormatException e) {
+      return null;
+    }
```
tmetzke, line 25: 🔧 Let's not obfuscate parsing errors, I would suggest.
```suggestion
    input == null ? null : return Long.parseLong(input);
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3539,40 +3810,76 @@ components:
...
+    TenantUpdateResponseNumberKeys:
+      deprecated: true
       type: object
+      allOf:
+        - $ref: "#/components/schemas/TenantUpdateResponseBase"
       properties:
         tenantKey:
           type: integer
-          description: The unique system-generated internal tenant ID.
+          description: The unique system-generated internal tenant ID
```
tmetzke, line 3831:
```suggestion
          description: The unique system-generated internal tenant ID.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3653,11 +3942,23 @@ components:
         - $ref: "#/components/schemas/SearchQueryRequest"
       description: User task search query request.
       type: object
-    UserTaskSearchQueryResponse:
+    UserTaskSearchQueryResponseNumberKeys:
+      description: User task search query response. (Number keys)
```
tmetzke, line 3946: 🔧 The `(Number keys)` documentation is not really clear for users, I believe. We could say something like `with key attributes as numeric values`. Wdyt?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/usermanagement/MappingController.java
tmetzke: Interesting. Might be something the Identity team mixed up accidentally.

danielkelemen: Probably. I guess the `MappingSearchQueryResponse` & `MappingItem` names are more consistent with our objects.
I'll switch the names back.

tmetzke (approved):
👍 Thanks for pushing this through 🚀 
🔧 There are some smaller things to correct here regarding parsing, I think.
💭 In general, there are two things I want to make transparent so we make a conscious decision:
* We are breaking the existing DTOs from 8.6 with this OpenAPI spec change. They previously returned number keys, now they return string keys. That will mean quite a change for existing 8.6 clients (as we can see with our own Java client here). That could be quite an ask when migrating from 8.6 to 8.7. What would be the alternative? Introduce new DTOs for 8.7 but their naming will be very difficult (either `XyzStringKeys` or an alternative that is still speaking but uses a different wording, e.g. `XyzResult` instead of `XyzResponse`).
* We now use string keys in the Java client "on the wire". This is fine since we'll have to do it eventually when number keys are not supported anymore in the future. One alternative would be to stick to number keys for now but we'll have to adjust this in the future anyway.

❓ Should we add tests for number keys in the REST gateway like we did for 8.6? I couldn't see any but I might have also simply missed them 🙈

tmetzke:
> > ❓ Should we add tests for number keys in the REST gateway like we did for 8.6? I couldn't see any but I might have also simply missed them 🙈
> 
> @tmetzke, This PR doesn't contain the serializer yet. So while number keys are defined, they can't be requested yet. I still need to finish the `StringSerializer` changes for 8.7, that will contain those tests.

Oh, yeah, I see. That makes sense 😅 Too many things in parallel to look at, I guess 🙈

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: I changed to this: `description: Tenant search response item. Key attributes as numeric values.`

pepopowitz (approved):
👍 I pulled this updated API spec into my local docs instance, and confirmed that the new response types represent properly in the generated docs.

tmetzke:
> Introducing new names for string key objects and leaving number key object names as is could be helpful indeed. Customers could just use the number header temporarily without any other change.
I think I will merge this because I'm afraid of the conflicts if I wait. But I can create a follow-up for finding a name and updating objects later again.

Yep, I agree, absolutely valid. Plus, I am not sure if it's actually a big issue. For this to be a problem, users must have created custom clients based on DTOs generated from our OpenAPI spec. I am wondering if this is a real use case already and if so, how many users this actually concerns.

> Yes, but then it's okay to already update the client with this PR, I think.
The nicest would be to switch to strings in the client as well. I can also follow-up on that next year.

Yeah, moving to strings in the client will be quite involved. You'll change a LOT of methods and DTOs there. Plus, this will also break Java client usage which is a LOT worse. Users expect this to be as stable as possible.

tmetzke:
No matter which direction we go here, let's try to be SUPER clear in a migration guide to help users understand why we break this and what the effort for them is to migrate.

PR #26147, opened 2024-12-16
25914 extend rdbms modul user tasks by variables

### LINKED ISSUE #25914: Extending Exporter and SearchClient for searching for user tasks by variables
The RDBMS Camunda 8 REST API V2 should support querying for user tasks based on process- and local variables.

p-wunderlich:
## Description

* Use operations api for VariableValueFilter to be consistent with other filters
  * This refactoring is also needed for the rdbms module. The current implementation (where the values are e.g. in a field called ‘eq’) is not usable with MyBatis because these words are used by the internal expression language. And the new operation already supports the module.

* Implement searching user tasks for variables in rdbms module

closes #25914


db/rdbms/src/main/resources/mapper/Commons.xml
```diff
@@ -78,4 +78,88 @@
...
+        <!-- can be ignored, doesn't make sence for searching of single variables -->
+        1 = 1
+      </when>
+      <when test="operation.operator.name().equals('GREATER_THAN')">
+        <if test="operation.type.toString == 'LONG'">
+          LONG_VALUE &gt; #{operation.value}
+        </if>
+        <if test="operation.type.toString == 'DOUBLE'">
+          DOUBLE_VALUE &gt; #{operation.value}
+        </if>
```
stoerti, line 119: ❓ What happens if the GREATER_THAN is performed on a STRING? Is it simply ignored or do we run in an error?

stoerti (commented):
Looks good, but I had two questions

db/rdbms/src/main/resources/mapper/Commons.xml
p-wunderlich: Depends on the database. But at least when the UntypedOperation is used, that should not happen.

search/search-client-query-transformer/src/test/java/io/camunda/search/clients/transformers/filter/VariableValueFilterTest.java
```diff
@@ -39,19 +41,15 @@ public void shouldSetFilters() {
     final var filter =
         filterBuilder
             .name("name")
-            .eq("equals")
-            .gt("greaterThen")
-            .gte("greaterThenOrEqual")
-            .lt("lessThen")
-            .lte("lessThenOrEqual")
+            .valueOperation(UntypedOperation.of(eq("equals")))
+            .valueOperation(UntypedOperation.of(gt("greaterThen")))
```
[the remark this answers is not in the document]
p-wunderlich: This was not autogenerated! :D But yes, at least testing the builder.

stoerti (approved):

search/search-domain/src/main/java/io/camunda/search/filter/UntypedOperation.java
```diff
@@ -46,4 +48,20 @@ public Object value() {
...
+
+  private static void validateOrThrow(final Operator operator, final ValueTypeEnum type) {
+    switch (operator) {
+      case GREATER_THAN:
+      case GREATER_THAN_EQUALS:
+      case LOWER_THAN:
+      case LOWER_THAN_EQUALS:
+        if (type != ValueTypeEnum.LONG && type != ValueTypeEnum.DOUBLE) {
+          throw new IllegalArgumentException(
+              "Unsupported value type for number operator (<,<=,>,>=): " + type);
```
danielkelemen, line 60: ❓Should we validate the `LIKE` & `STRING` as well?

danielkelemen (approved):
👍 Looks good to me!

search/search-domain/src/main/java/io/camunda/search/filter/UntypedOperation.java
p-wunderlich: Yes, good idea! :)

Issue #26134, opened 2024-12-16
Rename all the classes and packages related to Zeebe


houssain-barouni:
Re-apply https://github.com/camunda/camunda/issues/19485
- Switch packages in the client from `io.camunda.zeebe` to `io.camunda`
- Rename `ZeebeClient` and `ZeebeClientBuilder` to `CamundaClient` and `CamundaClientBuilder`

PR #26153, opened 2024-12-16
refactor: rename client classes package

houssain-barouni:
## Description

Renaming `io.camunda.zeebe`package  to `io.camunda` for all java client classes

## Checklist

- [ ] for CI changes:

## Related issues

relates to #26134


clients/java/src/main/java/io/camunda/client/api/ExperimentalApi.java
```diff
@@ -13,36 +13,14 @@
...
- *       it to an existing API is considered API-breaking.
- *   <li>Removing this annotation from an API gives it stable status.
- * </ol>
- *
- * <p>Note: This annotation is intended only for Zeebe library code. Users should not attach this
- * annotation to their own code.
- *
- * <p>This annotation was originally copied from io.grpc.ExperimentalApi, licensed under the Apache
- * License, Version 2.0. Copyright 2015 The gRPC Authors. Changes have been made since.
- */
```
tmetzke, line 45: ❓ Is there a specific reason to remove this?

tmetzke:
@nicpuppa, do you still remember what you did regarding the Zeebe test container and the classes it returned (since they are from the old package)?

tmetzke (commented):
👍 Generally looks like going into the right direction 🚀 
❓ There is one removal I don't understand, let me know your thoughts on that one.
💭 The Zeebe test container issue is something @nicpuppa also ran into, if I remember correctly. It is a bit annoying and I am not 100% sure yet what we do about those. We should ideally not depend on this in our own code base as it's not released together with the platform and can, apparently, get out of sync.

clients/java/src/main/java/io/camunda/client/api/ExperimentalApi.java
houssain-barouni: no, this is strange, it could be intellij

houssain-barouni:
> 👍 Generally looks like going into the right direction 🚀 ❓ There is one removal I don't understand, let me know your thoughts on that one. 💭 The Zeebe test container issue is something @nicpuppa also ran into, if I remember correctly. It is a bit annoying and I am not 100% sure yet what we do about those. We should ideally not depend on this in our own code base as it's not released together with the platform and can, apparently, get out of sync.

I am solving the issue with Zeebe test container in [this commit](https://github.com/camunda/camunda/pull/26153/commits/11cc848acfcd28576143325bd66d1f44501f7ef7). The solution basically allows `zeebe-test-container` to internally use `zeebe-client-java` (specifically version 8.4.5), while explicitly banning, with `Maven Enforcer Plugin`, the direct usage of` zeebe-client-java `as a dependency in the project.

For example, this approach helped identify that Optimize was still using the banned dependency:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.5.0:enforce (forbid-banned-dependencies) on project zeebe-qa-integration-tests: 
[ERROR] Rule 0: org.apache.maven.enforcer.rules.dependency.BannedDependencies failed with message:
[ERROR] 
[ERROR] io.camunda:zeebe-client-java:jar:8.6.6 <--- banned via the exclude/include list
```

nicpuppa:
@tmetzke @houssain-barouni can you remember me what is the problem with Zeebe test container ?

houssain-barouni:
> @tmetzke @houssain-barouni can you remember me what is the problem with Zeebe test container ?

`zeebe-test-containe`r uses an older version of `zeebe-client-java`, which has different client package names compared to the updated version introduced in this PR.
1. **Compilation Errors:**  where ZeebeContainer is used to build the client, such as [this example](https://github.com/camunda/camunda/blob/58ae68767536a54e6d324cab4a36be59070737d7/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/clustering/network/AsymmetricNetworkPartitionIT.java#L100). The client class returned by `ZeebeContainer` resides in the package i`o.camunda.zeebe.*`, whereas the updated code expects the client class in` io.camunda.*`. I resolved this by directly building the client in the test using the gateway address.
2. Test Runtime Errors: Excluding the zeebe-client-java dependency from zeebe-test-container (to prevent its usage in test classes) results in runtime errors. This happens because `zeebe-test-container` relies on `zeebe-client-java` for its internal logic, such as executing the topology command to verify cluster readiness. Instead of excluding the dependency, I kept it but enforced a rule to ban its direct usage in the project.

clients/java/src/main/java/io/camunda/client/api/ExperimentalApi.java
houssain-barouni: docs are back!

houssain-barouni:
@buccarel can you check this commit related Optimize https://github.com/camunda/camunda/pull/26153/commits/d008b4084b10387550f82901aa6bb07550e07188
- `zeebe-java-client` dependency changed to `camunda-java-client`
- removed `optimize-run-java-checks` workflow as it does not build the full project (and required dependencies from the project), it is already covered by the common CI

buccarel (approved):
The optimize commit looks good.
I also reviewed everything else and, to me, the renames are sound everywhere
🚀

tmetzke (approved):
👍 Changes look good to me ⭐

PR #26160, opened 2024-12-16
feat: Document available sort fields in REST API v2

### LINKED ISSUE #24363: Document available sort fields in REST API v2
### Description

The fields available in the Open API specification do not match exactly the sort fields that can be used by the users in the API requests, as some fields may not be available for sorting. So additional documentation is needed to describe the possible sort fields. The sort fields that can be used are usually defined in the class `SearchQueryRequestMapper.java`, in methods like `applyProcessInstanceSortField()`

Related issues:
https://github.com/camunda/camunda/issues/24146
https://forum.camunda.io/t/camunda-8-api-unknown-sortby/56412

ghost:
feat: Document available sort fields in REST API v2

Closes https://github.com/camunda/camunda/issues/24363


pepopowitz (approved):
👍 The spec changes look good to me!

tmetzke:
@pepopowitz, the OpenAPI linting fails with the following:
```
panic: interface conversion: *base.ObjectReference is not base.Foundational: missing method AddEdge

goroutine 20719 [running]:
github.com/pb33f/doctor/model.(*DrDocument).LocateModel(0xc000b300e0, 0xc000765720)
	/home/runner/go/pkg/mod/github.com/pb33f/doctor@v0.0.14/model/walk_model.go:83 +0x19f
github.com/daveshanley/vacuum/functions/core.(*Truthy).RunRule(0x18b6790?, {0xc00400eb10?, 0x1?, 0x0?}, {0xc0029f8c90, 0xc000[24](https://github.com/camunda/camunda/actions/runs/12383003523/job/34564913438?pr=26160#step:4:25)95c0, {0x139af00, 0xc00028bf70}, {0x0, 0x0}, ...})
	/home/runner/work/vacuum/vacuum/functions/core/truthy.go:88 +0x558
github.com/daveshanley/vacuum/motor.buildResults({0xc0002495c0, 0xc000[26](https://github.com/camunda/camunda/actions/runs/12383003523/job/34564913438?pr=26160#step:4:27)6000, 0xc000355540, {0x18b0aa8, 0xc0003736f0}, 0xc0000110b0, 0xc000d17d58, 0xc0001d3088, 0xc0000e0580, 0x0, ...}, ...)
	/home/runner/work/vacuum/vacuum/motor/rule_applicator.go:852 +0x6cb
github.com/daveshanley/vacuum/motor.runRule({0xc0002495c0, 0xc000266000, 0xc000355540, {0x18b0aa8, 0xc0003736f0}, 0xc0000110b0, 0xc000d17d58, 0xc0001d3088, 0xc0000e0580, 0x0, ...}, ...)
	/home/runner/work/vacuum/vacuum/motor/rule_applicator.go:766 +0xbd6
created by github.com/daveshanley/vacuum/motor.ApplyRulesToRuleSet.func7 in goroutine 20734
	/home/runner/work/vacuum/vacuum/motor/rule_applicator.go:622 +0x2a8
Error: Process completed with exit code 2.
```

Is this due to the description check rule that we added?

pepopowitz:
> @pepopowitz, the OpenAPI linting fails with the following:
> 
> ```
> panic: interface conversion: *base.ObjectReference is not base.Foundational: missing method AddEdge
> 
> goroutine 20719 [running]:
> github.com/pb33f/doctor/model.(*DrDocument).LocateModel(0xc000b300e0, 0xc000765720)
> 	/home/runner/go/pkg/mod/github.com/pb33f/doctor@v0.0.14/model/walk_model.go:83 +0x19f
> github.com/daveshanley/vacuum/functions/core.(*Truthy).RunRule(0x18b6790?, {0xc00400eb10?, 0x1?, 0x0?}, {0xc0029f8c90, 0xc000[24](https://github.com/camunda/camunda/actions/runs/12383003523/job/34564913438?pr=26160#step:4:25)95c0, {0x139af00, 0xc00028bf70}, {0x0, 0x0}, ...})
> 	/home/runner/work/vacuum/vacuum/functions/core/truthy.go:88 +0x558
> github.com/daveshanley/vacuum/motor.buildResults({0xc0002495c0, 0xc000[26](https://github.com/camunda/camunda/actions/runs/12383003523/job/34564913438?pr=26160#step:4:27)6000, 0xc000355540, {0x18b0aa8, 0xc0003736f0}, 0xc0000110b0, 0xc000d17d58, 0xc0001d3088, 0xc0000e0580, 0x0, ...}, ...)
> 	/home/runner/work/vacuum/vacuum/motor/rule_applicator.go:852 +0x6cb
> github.com/daveshanley/vacuum/motor.runRule({0xc0002495c0, 0xc000266000, 0xc000355540, {0x18b0aa8, 0xc0003736f0}, 0xc0000110b0, 0xc000d17d58, 0xc0001d3088, 0xc0000e0580, 0x0, ...}, ...)
> 	/home/runner/work/vacuum/vacuum/motor/rule_applicator.go:766 +0xbd6
> created by github.com/daveshanley/vacuum/motor.ApplyRulesToRuleSet.func7 in goroutine 20734
> 	/home/runner/work/vacuum/vacuum/motor/rule_applicator.go:622 +0x2a8
> Error: Process completed with exit code 2.
> ```
> 
> Is this due to the description check rule that we added?

It may be, though that is different than any errors I saw when writing it. I'll take a look this morning!

tmetzke:
Thanks, @pepopowitz, for those insights 👍 
I would propose we go for `allOf` and duplicate descriptions for now, although we're mostly accommodating the tool we're using. The change is not "wrong" in that sense, so I think that is the best tradeoff right now. I'd also like to not remove the vacuum rule for now if possible.

clients/java/pom.xml
```diff
@@ -23,7 +23,7 @@
   </licenses>
 
   <properties>
-    <version.java>8</version.java>
+    <version.java>21</version.java>
```
houssain-barouni, line 26: The Java client must be built with Java 8 to ensure compatibility with customers still using Java 8.

houssain-barouni (dismissed):
java version using in client should be reset to 8

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4024,13 +4024,32 @@ components:
...
+      properties:
+        field:
+          description: The field to sort by.
+          type: string
+          enum:
+            - key
+            - name
+            - tenantId
+        order:
+          $ref: "#/components/schemas/SortOrderEnum"
```
tmetzke, line 4039: ❓ Do we need to add this `order` attribute everywhere? It should come from the `SearchQuerySortRequest` already, I would assume.

clients/java/src/main/java/io/camunda/client/impl/search/query/VariableQueryImpl.java
```diff
@@ -65,8 +66,8 @@ public FinalSearchQueryStep<Variable> requestTimeout(final Duration requestTimeo
   }
 
   @Override
-  public CamundaFuture<SearchQueryResponse<Variable>> send() {
-    final HttpCamundaFuture<SearchQueryResponse<Variable>> result = new HttpCamundaFuture<>();
+  public ZeebeFuture<SearchQueryResponse<Variable>> send() {
+    final HttpZeebeFuture<SearchQueryResponse<Variable>> result = new HttpZeebeFuture<>();
```
tmetzke, line 70: 🔧 We need to account for the latest renamings in the Java client to `CamundaXyz` classes in this PR.

tmetzke (commented):
I am still reviewing, just some first things that caught my eye:
* Do we need to add the `order` property to all request DTOs in the OpenAPI spec?
* We need to incorporate the Java client renamings in this PR.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
mihail-ca: @pepopowitz @tmetzke I tried like below, is this what you mean ? But now it does not show the Possible values anymore.

Before:

```
    TenantSearchQuerySortRequest:
      allOf:
        - $ref: "#/components/schemas/SearchQuerySortRequest"
      properties:
        field:
          description: The field to sort by.
          type: string
          enum:
            - key
            - name
            - tenantId
        order:
          $ref: "#/components/schemas/SortOrderEnum"
      required:
        - field
```

After:

```
    TenantSearchQuerySortRequest:
      allOf:
        - $ref: "#/components/schemas/SearchQuerySortRequest"
      properties:
        field:
          description: The field to sort by.
          type: string
          enum:
            - key
            - name
            - tenantId
        order:
          description: The order in which to sort the related field.
          allOf:
            - $ref: "#/components/schemas/SortOrderEnum"
      required:
        - field
```

image

mihail-ca: @pepopowitz the last changed mentioned in this comment is not merged to this PR, that's why it's working. After step 1, you should also update with the sample code I put in **After** section above.

pepopowitz: Ah, my apologies, I assumed the referenced changes were committed.

I confirmed the experience you're seeing. I had not previously noticed that this is the case for any field using this style of inheritance of an enum field. Thanks for raising it! It appears to be a [known bug](http://github.com/PaloAltoNetworks/docusaurus-openapi-docs/pull/985) in the API docs generator. 

The good news is that it's fixed in a more recent version of the generator. The even better news is that I'm actively working toward updating our docusaurus instance, so that we can take advantage of that fix. When https://github.com/camunda/camunda-docs/pull/4799 lands, which I expect to happen in the next couple weeks, this issue will be resolved. 

I've confirmed in my local branch for https://github.com/camunda/camunda-docs/pull/4799 that when it does land, the enum will render properly for your changes: 

image

Note that the exclamation at the end of the description is me ensuring the correct description was being used.

**Resolution:** I think we should move forward with your changes, and accept that the docs are temporarily missing the enum values, until I can land my docusaurus 3 upgrade. We treat these `vNext` docs as a work-in-progress; this is an acceptable temporary regression (in the docs only) from my perspective.

tmetzke: @mihail-ca, what happens if you do not override the `order` attribute at all? So the description would just look like this:

```
    TenantSearchQuerySortRequest:
      allOf:
        - $ref: "#/components/schemas/SearchQuerySortRequest"
      properties:
        field:
          description: The field to sort by.
          type: string
          enum:
            - key
            - name
            - tenantId
        required:
        - field
```

Ideally, the `order` attribute should be picked up as is from the `SearchQuerySortRequest` and the `field` attribute should be overridden. This might create an issue with either (or both) the validation and the docusaurus plugin. But OpenAPI-wise, this would be the MVP solution we should use, as far as I can tell.

tmetzke: OK, I tried that myself locally and it doesn't work with the current docusaurus. The `order` attribute will not be displayed then.
image

It works fine in Swagger UI though
image

tmetzke: I was wondering: do we need to reference the generic `SearchQuerySortRequest` at all? We don't really need it; it doesn't provide much value anymore. We could just do the following
```
    TenantSearchQuerySortRequest:
      type: object
      properties:
        field:
          description: The field to sort by.
          type: string
          enum:
            - key
            - name
            - tenantId
        order:
          $ref: "#/components/schemas/SortOrderEnum"
      required:
        - field
```

Locally, this leads to the following for me which looks good, as far as I can tell
image

tmetzke: We can either remove the `SearchQuerySortRequest` from the OpenAPI spec and create a simple `record` in Java instead or keep it in OpenAPI but we don't really need it there, it is an implementation detail in our controllers.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQuerySortRequest.java
```diff
@@ -0,0 +1,44 @@
...
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.rest;
+
+import io.camunda.zeebe.gateway.protocol.rest.SortOrderEnum;
+
+public class SearchQuerySortRequest {
```
tmetzke, line 12: 🔧 I guess we can make this a [Java record](https://www.baeldung.com/java-record-keyword) instead of a Java plain class, right?

tmetzke (approved):
🚀 Thanks, @mihail-ca, for your ongoing effort here. I know it wasn't as easy as we hoped for (mostly due to OpenAPI validation and documentation back and forth). I appreciate your persistence here 👍 

👍 The code looks good to me
🔧 I added some smaller suggestions, nothing blocking so we could merge it the way it is. Thus, pre-approving already.

tmetzke:
@mihail-ca, we have to rebase this since https://github.com/camunda/camunda/pull/26479 was merged and brings a couple of changes to the OpenAPI spec.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4024,13 +4024,33 @@ components:
...
+          description: The field to sort by.
+          type: string
+          enum:
+            - key
+            - name
+            - tenantId
+        order:
+          description: The order in which to sort the related field.
+          allOf:
+            - $ref: "#/components/schemas/SortOrderEnum"
```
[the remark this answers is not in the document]
pepopowitz: @tmetzke ideally yes, however [earlier in the PR](https://github.com/camunda/camunda/pull/26160#issuecomment-2551955135) you and I agreed to ask  switch these all to `allOf` inheritance, due to some strange crashing behavior in the vacuum linting. Have conditions changed since then that make you think the crashes should no longer happen?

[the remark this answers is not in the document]
tmetzke: Yeah, I think conditions changed but I am not 100% sure. We don't have any `allOf` in those objects anymore so I THINK we can move to simple `$ref`s. I will give this a try as I would prefer that for further simplification, let's see.

tmetzke:
I will take over from @mihail-ca here to finish this so he can focus on other duties.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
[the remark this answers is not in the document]
pepopowitz: Conditions definitely changed! I swapped them all to direct $ref locally, and the linting works no problem.

And I did then confirm that the docs are generated appropriately.

I support this change, go for it!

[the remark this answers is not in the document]
tmetzke: I adjusted this in the OpenAPI spec

tmetzke:
Rebased onto `main` and addressed the remaining open suggestions

PR #26161, opened 2024-12-16
Emphasize message buffering and distiguish between publish and correlate

ingorichtsmeier:
## Description

Clearly distinguish between publish and correlating a message and the feature of buffering.

## Related issues

No issue created so far. During a customer discussion, I checked the endpoint and found, that message buffering is not mentioned in the API description.

If an additional issue is required, please let me know and I will create one.

closes #


tmetzke (approved):
Love this 🚀 
Thanks a lot, @ingorichtsmeier ⭐

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2893,6 +2894,8 @@ paths:
       description: |
         Publishes a message and correlates it to a subscription.
         If correlation is successful it will return the first process instance key the message correlated with.
+        The message is not buffered.
+        Use the publish message endpoint to send messaged that can be buffered.
```
tmetzke, line 2898:
```suggestion
        Use the publish message endpoint to send messages that can be buffered. 
```

PR #26191, opened 2024-12-17
feat: add more incident E2E tests

stoerti:
## Description

I have started with incidents and extended/rewritten the existing E2E tests a bit.

My goal with the refactoring was, to separate the technical parts of the test from the actual test. The test itself should only contain small commands (given, when) and checks (then) on business level but no technical glue code like creating and sending a Zeebe command via the client and unpacking the answer.
- **TestSteps**: Parent class for all general purpose steps and checks to execute during various tests (deploy and start processes, read entities)
- **IncidentSteps**: Step-class inside the Incident test suite. This contains more specific steps used only in the actual test suite ("cause an incident", "resolve an incident").

It is influenced a bit by the test framework [JGiven](https://jgiven.org/), which I found useful in many projects because it separates the actual test from the glue code.


p-wunderlich (approved):
I like this JGiven Like style. It separates the technical "setup" code from the business code. Very sad that we don't use kotlin, than we even could have real sentences with blanks in between. :D

npepinpe (changes requested):
Thanks for the additional coverage! 

I'm a little confused by the removal of the `IncidentUpdateIT` class, as it doesn't seem at first glance that the new tests provide the same coverage. For example, nowhere do we test the state of the incident (PENDING, ACTIVE, RESOLVED, etc.). Just that it exists? It's possible I missed the assertion, I found following the steps a little difficult, because they end up being 3 nested levels with lambdas and different matchers. 

:x: That brings me to the JGiven style. I don't know if we gain much from that, the actual set up, and loads of the assertions are now hidden away from the test, and since the steps tend to nest others, the names get progressively more generic, and it's unclear at first glance from the test what is always happening.

I think having the steps just perform business logic makes sense, e.g. `deployProcess`, `startProcess`, etc. But I would prefer keeping the assertions in the test themselves, as this is central to the test - it tells me exactly what it is covering. To some extent, the `given` and `when` sections also do, but a bit less. 

So I find it harder now to really assess what coverage is provided by test as the assertions are abstracted away from the test declaration.

It could of course be that I'm not used to this style, and I'm just being "conservative" :smile: So I'd be happy to hear from others (e.g. @ChrisKujawa), but in the absence of consensus from them, I would ask that you stick with our existing style.

PR #26195, opened 2024-12-17
Return correct status code when a user has insufficient permissions

### LINKED ISSUE #26010: Missing permissions should return status code 403 Forbidden
When a request is rejected because of missing permissions, the response status code is a 401 Unauthorized. This is incorrect, a 401 should indicate missing or invalid authentication. Instead we should return status code 403 Forbidden to indicate that authentication worked but the entity does not have the required permissions.

The root cause is that `RejectionType.UNAUTHORIZED` is used too broadly. When we find valid authentication details in a request but not the required permission we should return a new `RejectionType.FORBIDDEN`. The `RestErrorMapper` needs to be extended to map this to a 403.

remcowesterhoud:
## Description

This PR changes the HTTP status codes and gRPC error codes from unauthorized to forbidden if there are insufficients permissions. It contains some refactoring to make the error messages more generic instead of handling them in every processor individually.

## Checklist

- [ ] for CI changes:

## Related issues

closes #26010

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
@@ -33,10 +33,10 @@ import java.util.stream.Stream;
 
 public final class AuthorizationCheckBehavior {
 
-  public static final String UNAUTHORIZED_ERROR_MESSAGE =
+  public static final String FORBIDDEN_ERROR_MESSAGE =
       "Unauthorized to perform operation '%s' on resource '%s'";
-  public static final String UNAUTHORIZED_ERROR_MESSAGE_WITH_RESOURCE =
-      UNAUTHORIZED_ERROR_MESSAGE + ", required resource identifiers are one of '%s'";
+  public static final String FORBIDDEN_ERROR_MESSAGE_WITH_RESOURCE =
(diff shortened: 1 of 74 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/AuthorizationCheckBehavior.java
```diff
@@ -349,22 +358,30 @@ public Set<String> getResourceIds() {
     public String getTenantId() {
       return tenantId;
     }
+
+    public String getForbiddenErrorMessage() {
+      return hasAddedResourceIds
```
lenaschoenburg, line 363: :upside_down_face: Could also check that `resourceIds` is not just the wildcard. Perhaps a bit less brittle than the boolean change flag.

zeebe/protocol/src/main/resources/protocol.xml
```diff
@@ -86,6 +86,7 @@
       <validValue name="PROCESSING_ERROR">4</validValue>
       <validValue name="EXCEEDED_BATCH_RECORD_SIZE">5</validValue>
       <validValue name="UNAUTHORIZED">6</validValue>
+      <validValue name="FORBIDDEN">7</validValue>
```
lenaschoenburg, line 89: :thought_balloon: In theory we could rename `UNAUTHORIZED` to `UNAUTHENTICATED`, right? Because that's what it actually means now.

lenaschoenburg (approved):

PR #26196, opened 2024-12-17
Add optimize indices to backups

### LINKED ISSUE #25597: Unify Optimize backup interfaces with webapp-backups
Add priorities from `webapps-backups` to optimize indices so that they can be backup with the same workflow as the other webapps

entangled90:
## Description
Optimize indices are added to the list of indices to be part of the backup.

## Related issues

closes #25597


dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
```diff
@@ -102,37 +122,25 @@ private <A> Function<Map<String, A>, String> differentConfigFor(final String fie
...
-                    Optional.ofNullable(tasklistProperties)
-                        .map(prop -> prop.getDatabase().equals(TasklistProperties.ELASTIC_SEARCH))),
-                skipEmptyOptional())
-            .get();
+    final var indexPrefix = getIndexPrefix();
+
+    final boolean isElasticsearch = getIsElasticsearch();
+
+    final var indexService =
+        new OptimizeIndexNameService(indexPrefix.isBlank() ? "optimize" : indexPrefix);
```
RomanJRW, line 130: I think the default case is handled already by Optimize config, and I _believe_ that an empty assignment would already default to `optimize`. If possible, would rather have the defaulting handled in one place

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
```diff
@@ -65,20 +73,23 @@
...
 
 @Configuration
-@ProfileOperateTasklist
+@ProfileWebApp
 public class BackupPriorityConfiguration {
 
   private static final Logger LOG = LoggerFactory.getLogger(BackupPriorityConfiguration.class);
   private static final String NO_CONFIG_ERROR_MESSAGE =
-      "Expected operate or tasklist to be configured, but none of them is.";
+      "Expected operate, tasklist or optimize to be configured, but none of them is.";
```
RomanJRW, line 81:
```suggestion
      "Expected operate, tasklist or optimize to be configured, but none of them are.";
```

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
```diff
@@ -102,37 +122,25 @@ private <A> Function<Map<String, A>, String> differentConfigFor(final String fie
...
             new ImportPositionIndex(indexPrefix, isElasticsearch),
             // TASKLIST
-            new TasklistImportPositionIndex(indexPrefix, isElasticsearch));
+            new TasklistImportPositionIndex(indexPrefix, isElasticsearch),
+            // OPTIMIZE
+            new OptimizePrio1Delegate<>(
+                isElasticsearch
+                    ? new PositionBasedImportIndexES()
+                    : new PositionBasedImportIndexOS(),
+                indexService));
```
RomanJRW, line 143: I think we would also need to include `TimestampBasedImportIndexES` and `TimestampBasedImportIndexOS` in Prio1

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
entangled90: Yes, that was one thing that I needed to double check with you, but I forgot to ask beforehand :+1:

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
```diff
@@ -192,4 +201,83 @@ public BackupPriorities backupPriorities() {
...
+        allMatch(
+            Optional::empty,
+            differentConfigFor("indexPrefix"),
+            Map.of(
+                "operate",
+                Optional.ofNullable(operateProperties).map(OperateProperties::getIndexPrefix),
+                "tasklist",
+                Optional.ofNullable(tasklistProperties).map(TasklistProperties::getIndexPrefix),
+                "optimize",
+                Optional.ofNullable(optimizePrefix)),
```
entangled90, line 258: This might not be right thing to do as the optimizePrefix has "optimize" as default value, in contrast with the other apps which have "" as default. The other application preprend the component name to all indices (it's not part of the config)

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
entangled90: So when apps different from optimize are deployed standalone (i.e. without optimize) the prefix will still be equal to "optimize" right? the service-config.yaml is  included as well for those apps?

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
```diff
@@ -65,20 +73,23 @@
...
   final OperateProperties operateProperties;
   final TasklistProperties tasklistProperties;
   final String[] profiles;
+  private final String optimizePrefix;
+  private final Boolean optimizeIsElasticSearch;
 
   public BackupPriorityConfiguration(
       @Autowired(required = false) final OperateProperties operateProperties,
       @Autowired(required = false) final TasklistProperties tasklistProperties,
+      @Autowired(required = false) final OptimizeIndexNameService optimizeIndexNameService,
```
entangled90, line 92: @RomanJRW  what should we do in case this is not autowired? Is it a possibility?

dist/src/main/java/io/camunda/webapps/profiles/ProfileWebApp.java
```diff
@@ -19,5 +19,5 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 @Inherited
-@Profile("(operate | tasklist | identity | optimize)")
+@Profile("(operate | tasklist | optimize)")
```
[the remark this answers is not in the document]
npepinpe: Can you elaborate on this? Is it just that the indices don't have the marker interface?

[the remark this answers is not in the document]
npepinpe: We will support it after right?

dist/src/main/java/io/camunda/application/commons/backup/DynamicIndicesConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+import io.camunda.webapps.backup.DynamicIndicesProvider;
+import io.camunda.webapps.profiles.ProfileWebApp;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ProfileWebApp
+public class DynamicIndicesConfiguration {
+
+  private static final String ELASTICSEARCH_TYPE = "elasticsearch";
```
npepinpe, line 21: :wrench: there is an enum for it in the exporter module. I wonder if we shouldn't move it into the search module so it's commonly available, over plain string constants spread across the codebase.

dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOptimizeBackupRepository.java
```diff
@@ -0,0 +1,55 @@
...
+ * different qualifiers for the elasticSearch clients between tasklist & operate When the ES client
+ * is shared between operate & tasklist then this class can be removed after adding tasklist to the
+ * valid profiles for {@link ElasticsearchBackupRepository}
+ *
+ * <p>Note that the condition used refers to tasklist ElasticSearchCondition
+ */
+@Conditional(ElasticSearchCondition.class)
+@Configuration
+@Profile("optimize & !operate")
+// only active if standalone, otherwise the operate one is used
```
npepinpe, line 34: :question: is that true without the `standalone` profile?

dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOptimizeBackupRepository.java
```diff
@@ -0,0 +1,55 @@
...
+public class ElasticsearchOptimizeBackupRepository {
+
+  private final ElasticsearchClient esClient;
+  private final BackupRepositoryProps backupRepositoryProps;
+  private final ThreadPoolTaskExecutor threadPoolTaskExecutor;
+
+  public ElasticsearchOptimizeBackupRepository(
```
npepinpe, line 41: :question: Don't you need the `@Autowired` annotation here? :thinking:

npepinpe (approved):

dist/src/main/java/io/camunda/webapps/profiles/ProfileWebApp.java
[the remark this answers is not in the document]
entangled90: Yes, but the BackupController cannot be instantiated as of now for identity as the dependencies are missing (just wiring the configs)

dist/src/main/java/io/camunda/application/commons/backup/BackupPriorityConfiguration.java
entangled90: I tried removing `required=false` from `@Autowired(required=false) OptimizeIndexNameService` but then the application is unable to start.
For this reason I added as a last resort `new OptimizeIndexNameProvider(defaultIndexPrefix)`

dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOptimizeBackupRepository.java
entangled90: I  thought it was not needed when there is just 1 constructor?

dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOptimizeBackupRepository.java
entangled90: Yes probably better to add standalone as well, since it may be that `tasklist, optimize` profiles are active which would lead to two `Bean` (even though I think it's quite unlikely)

dist/src/main/java/io/camunda/application/commons/backup/DynamicIndicesConfiguration.java
entangled90: Probably yes, we currently have many instances of such enums scattered around the code base.

entangled90: I'll do it in a separate PR

dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOptimizeBackupRepository.java
entangled90: I'll do it in a separate PR

PR #26202, opened 2024-12-17
refactor: identity migration uses camunda services directly

lenaschoenburg:
This is an internal refactoring to make sure we just use services from `camunda-services` directly, without any wrappers in between.

This will make authentication a bit easier to handle and removes a bit of indirection.

Relates to #25825


maryarm (approved):
Nice job.

PR #26211, opened 2024-12-17
[CPT] feat: Improve logging

saig0:
## Description

Extend the logging to print all Docker images on start-up (image name + version tag).

Adjust the logger for the containers to separate the output streams. Effect: Std-err to ERROR log level + remove
output type (STDOUT) from logging messages.

Add a new logger for the Camunda and the Connectors container. Compared to the existing logger, it forwards log entries with the correct log level instead of using the `INFO` level for all log messages. Enable JSON logging for containers ([Camunda](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/configuration/logging/#google-stackdriver-json-logging) + [Connectors](https://docs.camunda.io/docs/next/self-managed/connectors-deployment/connectors-configuration/#google-stackdriver-json-logging)).

**Review note**
I didn't add test cases for the logging because it is cumbersome to verify logging. Since it is just logging and doesn't impact CPT features, I think it is okay to have no tests. I verified the logging manually.

### DIFF
--- a/testing/camunda-process-test-example/src/test/resources/application.yml
+++ b/testing/camunda-process-test-example/src/test/resources/application.yml
@@ -8,4 +8,4 @@ logging:
     # Hide default logging from Testcontainers
     org:
       testcontainers: warn
-    tc: warn
+    tc: error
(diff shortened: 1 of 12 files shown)


berkaycanbc (approved):
Thanks @saig0, very useful changes! LGTM 🚀 

Just a note, unfortunately, I do not have capacity to test changes manually but changes look as expected.

PR #26218, opened 2024-12-17
Migrate active inclusive gateway

### LINKED ISSUE #25673: Migrate active inclusive gateway
## Description
For the migration of process instances with inclusive gateways we need to consider the case where an inclusive gateway is currently active. For example, because there is an incident raised on it.

This allows a special use case: resolving an incident because of an invalid execution listener `type` or `retries` expression.

An incident could also exist on the active inclusive gateway if the execution listener's job is failed without retries. However, this incident can already be resolved by updating the job's retries directly, or through process instance modification (re-activate the inclusive gateway).

berkaycanbc:
## Description

In order to support migration for active inclusive gateways we need to add inclusive gateway to supported element types. No other change is needed for incident migration because it is already handled as a part of the incident migration implementation.

The PR is the same as what is implemented in #25699 but for inclusive gateway.

## Related issues

closes #25673


korthout (approved):
👍 LGTM

🔧 Please remove the failing test `MigrateProcessInstanceUnsupportedElementsTest.shouldRejectMigrationForActiveInclusiveGateway` before merging.

💭 As that is the last test case in the unsupported elements class, we could consider to add tests for the remaining unsupported elements. Alternatively, some new issues to document them would be nice as well. For example, `START_EVENT`, `BOUNDARY_EVENT` can be active with an incident thanks to execution listeners and/or IO mappings, so there's reason to migrate them to resolve problems with them. We also have a new `AD_HOC_SUB_PROCESS` and haven't implemented migrations for active `INTERMEDIATE_THROW_EVENT` and `END_EVENT` yet either, which may be backed by a job, or have an incident due to IO mappings / execution listener. Lastly, `TASK` and `MANUAL_TASK` support execution listeners I think, so there's another case.

PR #26227, opened 2024-12-18
Support corrections for `assigning` task listeners

### LINKED ISSUE #26032: Allow corrections for the assigning listener
Similar to:
- https://github.com/camunda/camunda/issues/24889

```[tasklist]
### Tasks
- [x] Append the `CORRECTED` event when a task listener job is completed with corrections
- [x] Fill the `ASSIGNED` event with the latest intermediate user task data
- [x] Possibly change the `ASSIGNED` event applier to store the corrected data
```

korthout:
## Description

Adds support for correcting user task data in the `assigning` task listeners.

- Adds a bunch of test cases, building on existing correction tests for `completing` listeners
- Ensures that corrected user task data is persisted on `ASSIGNED` event
- Fixed a bug preventing to deny the `assigning` event when `CLAIMING`
- Cleaned up some redundant code

> [!NOTE]
> There are two solutions to persisting the data, which are b128494d33f279bca9351ac35f61a8da2ee03578 and 658c1fd39e253186b82e96764fc13681ccc91696. I'd like to discuss which we prefer. I've provided both to facilitate the discussion.
>
> Note that the alternative also requires 84f95c8, ands leads to some failing tests becaue we don't fully always set the changed attributes on every command/event correctly yet. It will require changes [more changes](https://github.com/camunda/camunda/compare/korthout-26032-corrections-on-assigning...korthout-26227-alternative), and even some more still. It may be better to look at that separately. But let me know what you think.

## Related issues

closes #26032

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/UserTaskProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/UserTaskProcessor.java
@@ -131,7 +131,7 @@ public class UserTaskProcessor implements TypedRecordProcessor<UserTaskRecord> {
     switch (lifecycleState) {
       case COMPLETING ->
           writeRejectionForCommand(command, persistedRecord, UserTaskIntent.COMPLETION_DENIED);
-      case ASSIGNING ->
+      case ASSIGNING, CLAIMING ->
           writeRejectionForCommand(command, persistedRecord, UserTaskIntent.ASSIGNMENT_DENIED);
       default ->
           throw new IllegalArgumentException(
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
@@ -21,8 +21,6 @@ import io.camunda.zeebe.protocol.record.intent.UserTaskIntent;
 import io.camunda.zeebe.stream.api.records.TypedRecord;
 import io.camunda.zeebe.util.Either;
 import java.util.List;
-import java.util.Objects;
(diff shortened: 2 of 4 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/UserTaskProcessor.java
```diff
@@ -131,7 +131,7 @@ private void processDenyTaskListener(final TypedRecord<UserTaskRecord> command)
     switch (lifecycleState) {
       case COMPLETING ->
           writeRejectionForCommand(command, persistedRecord, UserTaskIntent.COMPLETION_DENIED);
-      case ASSIGNING ->
+      case ASSIGNING, CLAIMING ->
```
ana-vinogradova-camunda, line 134: Great catch, thanks @korthout 👍

ana-vinogradova-camunda:
🚀 Great test refactoring and clean up! 👍 

Regarding persisting the data:
- To me the second approach looks like a better option in general for the reasons mention in the commit: we won't update any unexpected attributes that could be added in the future.
- However, the second approach requires quite a lot of changes and that might bring risks in the short term.
- As a result we might want to stick with the first options and potentially consider the second one as an improvement for the future.

ana-vinogradova-camunda (approved):
Great PR @korthout  🔥

ce-dmelnych (approved):
Great PR, Nico! 👏 
Thank you for the excellent test coverage, fixes, and simplifications - this really enhances the code quality and maintainability.

---
Regarding the note on the two solutions for persisting data:

Both options are clean and easy to understand. The alternative approach, which leverages `changedAttributes`, is indeed more robust and future-proof. However, considering the limited timeframe before the code freeze, the first option seems more suitable for the current release. We can explore and implement the alternative solution in the next iteration when we have more time to address the additional changes required.

Thanks again for your hard work and attention to detail! 🚀

PR #26234, opened 2024-12-18
[CPT] fix: Element assertion can handle BPMN elements with the same name

### LINKED ISSUE #25355: [CPT] hasCompletedElements() throws an IllegalStateException: Duplicate key null
## Description

I have a process test with an `hasCompletedElements()` assertion. When running the test case, it throws the following exception:

```
java.lang.IllegalStateException: Duplicate key null (attempted merging values COMPLETED and COMPLETED)
```

The test case:
```

### PR DESCRIPTION
**Pull request by saig0**
## Description

Fix a bug in the element assertions if two BPMN elements have the same name (or an empty name). 

Note that the element assertions will change soon when switching it element IDs instead ([ref](https://github.com/camunda/camunda/issues/23294)).

## Related issues

closes #25355

### DIFF
--- a/testing/camunda-process-test-java/src/test/java/io/camunda/process/test/api/ProcessInstanceAssertTest.java
+++ b/testing/camunda-process-test-java/src/test/java/io/camunda/process/test/api/ProcessInstanceAssertTest.java
@@ -499,6 +499,27 @@ public class ProcessInstanceAssertTest {
 
       verify(camundaDataSource).getFlowNodeInstancesByProcessInstanceKey(PROCESS_INSTANCE_KEY);
     }
+
+    @Test
+    void shouldFailWithSameElementName() throws IOException {
+      // given
+      final FlowNodeInstanceDto flowNodeInstanceA = newActiveFlowNodeInstance("A");
+      final FlowNodeInstanceDto flowNodeInstanceA2 = newActiveFlowNodeInstance("A");
+
+      when(camundaDataSource.getFlowNodeInstancesByProcessInstanceKey(PROCESS_INSTANCE_KEY))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: testing/camunda-process-test-java/src/test/java/io/camunda/process/test/api/ProcessInstanceAssertTest.java [hunk 1]
```diff
@@ -608,6 +666,22 @@ void shouldHasTerminatedElements() throws IOException {
       CamundaAssert.assertThat(processInstanceEvent).hasTerminatedElements("A", "B");
     }
 
+    @Test
+    void shouldHasTwoTerminatedElements() throws IOException {
```
berkaycanbc, line 670: 🔧 
```suggestion 
    void shouldHaveTwoTerminatedElements() throws IOException {
```

testing/camunda-process-test-java/src/test/java/io/camunda/process/test/api/ProcessInstanceAssertTest.java
```diff
@@ -520,6 +541,22 @@ void shouldHasCompletedElements() throws IOException {
       CamundaAssert.assertThat(processInstanceEvent).hasCompletedElements("A", "B");
     }
 
+    @Test
+    void shouldHasTwoCompletedElements() throws IOException {
```
berkaycanbc, line 545: 🔧 
```suggestion 
    void shouldHaveTwoCompletedElements() throws IOException {
```

berkaycanbc (approved):
LGTM 🚀 

🔧 Left two naming correction suggestion.

PR #26242, opened 2024-12-18
Add standalone SchemaManager to distro which creates ElasticSearch Schemas via isolated CommandLine app (#25922)

### LINKED ISSUE #25922: Add standalone SchemaManager to distro which creates ElasticSearch Schemas via isolated CommandLine app
**Describe the issue:**

We need an isolated Java application that can create and update the necessary Schemas on ElasticSearch. The Java app should take all necessary configurations, but just create/update schemas.

This is required in enterprise setups, where the team looking at Camunda cannot create Schemas and need to hand over a defined script to ElasticSearch admins.

**Actual behavior:**

Currently the Schema is automatically created/updated on startup of the Camunda platform. In enterprise scenarios, the user for ElasticSearch access does not have the elevated priviliges to create Schemas.

**Expected behavior:**

Have a seperate Java app for schema creation/update that is stripped down to exactly this logic - but using the same logic as the overall platform.

berndruecker:
### Description

This PR adds a new `StandaloneSchemaManager` which allows setting up the schema for Elastic, for example in cases where the user cannot give permissions to C8 to set up the schema directly, and the application must be used separately by a different team.

### Related issues

closes #25922

### DIFF
--- a/dist/src/main/java/io/camunda/application/StandaloneSchemaManager.java
+++ b/dist/src/main/java/io/camunda/application/StandaloneSchemaManager.java
@@ -24,13 +24,31 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.ConfigurableApplicationContext;
 
+/**
+ * Soley create or update Schema for ElasticSearch by running this standalone application.
+ *
+ * <p>Configure with {@link ConnectConfiguration} properties, prefixed by `camunda.database`, for
+ * example:
+ *
+ * <pre>
+ * camunda.database.type=elasticsearch
+ * camunda.database.url=
+ * camunda.database.security.selfSigned=
+ * camunda.database.security.enabled=
+ * camunda.database.security.certificatePath=
+ * camunda.database.username=
+ * camunda.database.password=
+ * </pre>
+ *
+ * All of those porperties can also be handed over via environment variables, e.g.
+ * `CAMUNDA_DATABASE_URL`
+ */
 @SpringBootConfiguration
 @EnableConfigurationProperties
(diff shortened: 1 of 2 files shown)


dist/src/main/java/io/camunda/application/StandaloneSchemaManager.java
```diff
@@ -0,0 +1,112 @@
...
+            .listeners(new ApplicationErrorListener())
+            .build(args);
+
+    final ConfigurableApplicationContext applicationContext =
+        standaloneSchemaManagerApplication.run(args);
+
+    final SchemaManagerConnectConfiguration connectConfiguration =
+        applicationContext.getBean(SchemaManagerConnectConfiguration.class);
+
+    if (!"elasticsearch".equalsIgnoreCase(connectConfiguration.getType())) {
```
npepinpe, line 84: :question: I assume this is just a constraint for time sake, as the customer is using specifically ES. But nothing should speak against support OS, right?

dist/src/main/java/io/camunda/application/StandaloneSchemaManager.java
```diff
@@ -0,0 +1,112 @@
...
+      LOG.error(
+          "Cannot creating schema for anything other than Elasticsearch with this script for now...");
+      System.exit(0);
```
npepinpe, line 87: :x: 0 exit code is used for success, so let's set a non-zero error code at least, e.g. `1`.

npepinpe (approved):
:+1: Thanks for that!

Only thing is the exit code on error, just change it, and then let's merge.

I'll have someone add a basic integration test on this afterwards, and I'll check if we think it's worth quickly supporting also OS (since it should be a really low hanging fruit, and product already figured we could use this for other things :+1:)

PR #26244, opened 2024-12-18
Tasklist opensearch schema validation fix

panagiotisgts:
## Description

Ported Opensearch schema parsing from CamundaExporter.

When on Opensearch, Tasklist schema validator registers falls positives as differences in regards to the `tasklist-task` index mappings, due to the order of elements returned from Opensearch. Before doings the diff of the existing mappings and the ones parsed from the JSON schema, elements are now ordered using `TreeMap` instead of `LinkedHashMap` and nested `Lists` are also sorted properly.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/tasklist/els-schema/src/main/java/io/camunda/tasklist/schema/IndexMapping.java
+++ b/tasklist/els-schema/src/main/java/io/camunda/tasklist/schema/IndexMapping.java
@@ -10,10 +10,14 @@ package io.camunda.tasklist.schema;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.tasklist.exceptions.TasklistRuntimeException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 
 public class IndexMapping {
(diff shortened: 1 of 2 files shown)


tasklist/els-schema/src/main/java/io/camunda/tasklist/schema/IndexMapping.java
```diff
@@ -59,7 +63,37 @@ public Map<String, Object> toMap() {
...
+                property -> convertToOrderedStructures(property.getTypeDefinition())));
+  }
+
+  /**
+   * On Opensearch Schema validation fails when the schema is created by the Exporter (primarility
+   * for tasklist-task index and the `join` field), due to the order of parsed index mapping
+   * properties. Using TreeMap and sorting the Lists present in the properties ensures the order of
+   * properties when comparing the existing mappings with the ones defined in the schema.
+   */
+  private Object convertToOrderedStructures(final Object value) {
```
ChrisKujawa, line 76: ❓ Stupid question but couldn't we use ObjectMapper and JsonObject to compare against?

panagiotisgts: The comparison of the indices is currently based on `com.google.common.collect.MapDifference` to perform a diff between HashMaps. However, for nested Maps the order of keys and for lists inside them need to be similar.

There could be other ways to achieve this fix as well. This is what came to my mind to unblock this issue. 

Also, changing the order of the entries in the JSON schema fixed this issue but could arise sooner or later. It's weird it's not happening on the CamundaExporter aswell because the diff is done with the exact same way :dizzy:

EuroLew (approved):
Not pretty especially the repeated code but we will be removing this sometime in the future so LGTM for now.

PR #26254, opened 2024-12-18
Document hash verification

### LINKED ISSUE #24284: [Doc Handling] Add a cryptographic hash of the file contents to the document reference
**Description**

The first iteration of Document Handling does not cover authorization. Although process-definition-based authorizaiton will be handled centrally for the whole Camunda, we can introduce an intermediary solution to prevent attackers from accessing documents from a foreign process by brute-forcing the document ID.

- The document reference should include a cryptographic hash of the document content (e.g. SHA-256)
- In document content get requests, the hash value should be passed along with the document ID
- The server has to check the hash value to ensure only those who have access to the original document reference object can download the file.

mo-okocha:
## Checklist

- [ ] for CI changes:

## Related issues

closes #24284

### DIFF
--- a/clients/java/src/main/java/io/camunda/client/impl/CamundaClientImpl.java
+++ b/clients/java/src/main/java/io/camunda/client/impl/CamundaClientImpl.java
@@ -803,19 +803,24 @@ public final class CamundaClientImpl implements CamundaClient {
 
   @Override
   public DocumentContentGetRequest newDocumentContentGetRequest(final String documentId) {
-    return new DocumentContentGetRequestImpl(httpClient, documentId, null, config);
+    return new DocumentContentGetRequestImpl(httpClient, documentId, null, null, config);
   }
 
   @Override
   public DocumentContentGetRequest newDocumentContentGetRequest(
       final DocumentReferenceResponse documentReference) {
     return new DocumentContentGetRequestImpl(
-        httpClient, documentReference.getDocumentId(), documentReference.getStoreId(), config);
+        httpClient,
+        documentReference.getDocumentId(),
+        documentReference.getStoreId(),
+        documentReference.getContentHash(),
+        config);
   }
 
   @Override
   public CreateDocumentLinkCommandStep1 newCreateDocumentLinkCommand(final String documentId) {
-    return new CreateDocumentLinkCommandImpl(documentId, null, jsonMapper, httpClient, config);
+    return new CreateDocumentLinkCommandImpl(
+        documentId, null, null, jsonMapper, httpClient, config);
   }
 
   @Override
@@ -824,6 +829,7 @@ public final class CamundaClientImpl implements CamundaClient {
     return new CreateDocumentLinkCommandImpl(
         documentReference.getDocumentId(),
         documentReference.getStoreId(),
+        documentReference.getContentHash(),
         jsonMapper,
         httpClient,
         config);
--- a/clients/java/src/main/java/io/camunda/client/impl/fetch/DocumentContentGetRequestImpl.java
+++ b/clients/java/src/main/java/io/camunda/client/impl/fetch/DocumentContentGetRequestImpl.java
@@ -35,17 +35,20 @@ public class DocumentContentGetRequestImpl implements DocumentContentGetRequest 
   private final RequestConfig.Builder httpRequestConfig;
   private final String documentId;
   private String storeId;
+  private String contentHash;
 
   public DocumentContentGetRequestImpl(
       final HttpClient httpClient,
       final String documentId,
       final String storeId,
+      final String contentHash,
       final CamundaClientConfiguration configuration) {
     ensureNotNull("documentId", documentId);
     this.httpClient = httpClient;
     httpRequestConfig = httpClient.newRequestConfig();
     this.documentId = documentId;
     this.storeId = storeId;
+    this.contentHash = contentHash;
     requestTimeout(configuration.getDefaultRequestTimeout());
   }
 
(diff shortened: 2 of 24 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
chillleader, line 1: Here, we also need to extend the `DocumentReference` schema to include the hash property

clients/java/src/main/java/io/camunda/client/impl/CamundaClientImpl.java
```diff
@@ -796,14 +797,21 @@ public CreateDocumentCommandStep1 newCreateDocumentCommand() {
...
-    return new DocumentContentGetRequestImpl(httpClient, documentId, null, config);
+    return new DocumentContentGetRequestImpl(httpClient, documentId, null, null, config);
   }
 
   @Override
   public DocumentContentGetRequest newDocumentContentGetRequest(
       final DocumentReferenceResponse documentReference) {
+    final String contentHash =
+        (String)
+            documentReference.getMetadata().getCustomProperties().get(CONTENT_HASH_METADATA_KEY);
```
chillleader, line 808: I think we should include the document hash into the root-level properties of the document reference response along with the document ID and store ID, not in custom properties of the metadata.

clients/java/src/main/java/io/camunda/client/impl/CamundaClientImpl.java
```diff
@@ -796,14 +797,21 @@ public CreateDocumentCommandStep1 newCreateDocumentCommand() {
 
   @Override
   public DocumentContentGetRequest newDocumentContentGetRequest(final String documentId) {
-    return new DocumentContentGetRequestImpl(httpClient, documentId, null, config);
+    return new DocumentContentGetRequestImpl(httpClient, documentId, null, null, config);
```
chillleader, line 800: `DocumentContentGetRequest` should be extended to allow specifying the content hash. If the user creates the command using this method (only with document ID), it should be possible to specify content hash later in the builder.

clients/java/src/main/java/io/camunda/client/impl/fetch/DocumentContentGetRequestImpl.java
```diff
@@ -63,6 +66,9 @@ public CamundaFuture<InputStream> send() {
     if (storeId != null) {
       queryParams.put("storeId", storeId);
     }
+    if (contentHash != null) {
+      queryParams.put("contentHash", contentHash);
+    }
```
chillleader, line 71: Should we make this parameter required if the verification is enabled by default in all document stores?

chillleader (dismissed):
Great job so far, thank you!

I think we need to adjust the solution to make hash a root-level property of the document reference and include it into the data schema, instead of passing it as a custom metadata entry.

Besides that, it would be great to add tests to cover this new funcitonality (at least some e2e test cases in the client implementation).

dist/src/main/java/io/camunda/application/StandaloneCamunda.java
```diff
@@ -28,8 +25,7 @@
 public class StandaloneCamunda {
 
   private static final String SPRING_PROFILES_ACTIVE_PROPERTY = ACTIVE_PROFILES_PROPERTY_NAME;
-  private static final String DEFAULT_CAMUNDA_PROFILES =
```
marcosgvieira, line 31: @mo-okocha let's not change StandaloneCamunda, I believe this changes were done for development tests, let's revert it.

dist/src/main/java/io/camunda/application/StandaloneGateway.java
```diff
@@ -22,7 +23,10 @@ public static void main(final String[] args) {
 
     final var standaloneGatewayApplication =
         MainSupport.createDefaultApplicationBuilder()
-            .sources(CommonsModuleConfiguration.class, GatewayModuleConfiguration.class)
+            .sources(
+                CommonsModuleConfiguration.class,
+                GatewayModuleConfiguration.class,
+                BrokerModuleConfiguration.class)
```
marcosgvieira, line 29: @mo-okocha do we need to change here, why, can you check?

document/store/src/main/java/io/camunda/document/store/aws/AwsDocumentStore.java
```diff
@@ -193,6 +209,33 @@ private Either<DocumentError, DocumentLink> linkDocumentInternal(
...
+      final HeadObjectResponse documentInfo = getDocumentInfo(documentId);
+      if (documentInfo == null) {
+        return Either.left(new DocumentNotFound(documentId));
+      }
+      if (!documentInfo.hasMetadata()) {
+        return Either.left(new DocumentError.InvalidInput("No metadata found for document"));
+      }
+
+      if (!documentInfo.metadata().containsKey(CONTENT_HASH_METADATA_KEY)) {
+        return Either.left(new DocumentError.InvalidInput("No content hash found for document"));
```
marcosgvieira, line 224: @mo-okocha do we have unit tests for this?

marcosgvieira (changes requested):
@mo-okocha good work, there are some changes in Gateway and CamundaStandalone that I think are related to development, please have a look.

Also there are missing unit tests, we should focus on implementing them.

document/store/src/main/java/io/camunda/document/store/aws/AwsDocumentStore.java
mo-okocha: Unit tests have been added for each branch in the method

marcosgvieira (approved):
lgtm!

PR #26255, opened 2024-12-18
Move authorization ITs to engine tests

remcowesterhoud:
## Description

The authoriation ITs are very slow. By moving them to the engine tests we can cut the CI time from 15-20 minutes to a few seconds. All we need to ensure is that we keep a single test case to test our end-to-end solution. I have chosen to keep the deployment authorization IT, but renamed it slightly so it's clear it's no about deployment. The goal if the IT is to test basic auth over the REST gateway.

This PR is pretty big. Most of it moving test classes around. The migration commit is where I actually change the test cases to use the engine rule. It's still a lot, but it's the same trick in every test case. You could probably skim through it fairly quick.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


koevskinikola (approved):
LGTM @remcowesterhoud 👍

PR #26256, opened 2024-12-18
feat: remove optimize's implementation of backups

### LINKED ISSUE #24476: Remove current implementation of Backups from Tasklist, Operate and Optimize

entangled90:
## Description
Removes all code related to backups in optimize, since the implementation is now common to all webapps
## Checklist

- [x] Wait for #26196 to be merged

## Related issues

closes #24476


optimize/backend/src/main/java/io/camunda/optimize/OptimizeTomcatConfig.java
```diff
@@ -77,7 +76,6 @@ public class OptimizeTomcatConfig {
             OptimizeResourceConstants.STATIC_RESOURCE_PATH,
             OptimizeResourceConstants.ACTUATOR_ENDPOINT,
             PanelNotificationConstants.SEND_NOTIFICATION_TO_ALL_ORG_USERS_ENDPOINT,
-            RestConstants.BACKUP_ENDPOINT,
```
entangled90, line 80: Should endpoints defined in actuator in the `dist` module be listed here or not?

grlimacan (changes requested):
Hi @entangled90 , the removals you did are ok, but I believe there are still quite a few places where things can be removed. Here's a list of what I have found so far:

- AbstractDatabaseSchemaTestClient: everything related to snapshots
- ElasticSearchConfiguration / OpenSearchConfiguration/service-config.yaml / ConfigurationServiceConstants / ConfigurationServiceTest: Quite a few configs about Backup present - note it might be a wise choice to keep those though, because removing there might be a breaking change for existing customers (needs to be tested). If that's the case, please document accordingly
- DatabaseBackup.java can be removed

entangled90:
- `AbstractDatabaseSchemaTestClient` it's used by `Upgrade{Opensearch/Elasticsearch}SchemaIT` for example, so I would not remove it
- `DatabaseBackup` is still used atm from the common `dist` module (as optimize is still able to take backup, but it's just using shared code)

optimize/backend/src/main/java/io/camunda/optimize/OptimizeTomcatConfig.java
entangled90: @grlimacan is removing this ok? should I add another rule for "backupHistory" (a new endpoint that was added) ?

grlimacan: I am actually not famililar with the new TomcatConfig, @buccarel is probably the one that can answer this question best

grlimacan:
> * AbstractDatabaseSchemaTestClient

@entangled90 , here I meant not to remove the whole file, but all the parts that are related to snapshots. I'm also fine with keeping this for the moment, as long as there is a follow up ticket for it

grlimacan (approved):
Pre-approving, necessary changes (if any) are not critical

entangled90:
In the `Elasticsearch/OpenSearchSchemaTestClient`  a backup/restore was performed before upgrading the schema. I removed that part as it didn't feel directly related to the upgrade and then removed all the snapshot related methods that were not in used.

grlimacan (approved):
Happy for you to merge when pipeline is green

optimize/backend/src/main/java/io/camunda/optimize/OptimizeTomcatConfig.java
buccarel: From what I'm understanding, the handler of such new endpoint is already implemented outside of the optimize module. This should mean that there's no need to declare the endpoint in optimize

PR #26260, opened 2024-12-18
feat: add StringSerializer to support keys as numbers

### LINKED ISSUE #26251: [C8 API] Add Long serializer for key conversion
## Description

In order to support keys as `String` and `Long` without too much code duplication, let's add a generic `String` serializer.
Port of https://github.com/camunda/camunda/issues/25760 in 8.7.

## Acceptance criteria

- Added in 8.6 & port to 8.7 later.
- Add `StringSerializer`
  - Serialize only REST `String` fields that end with `...Key` (`tenantKey`) and arrays ending `...Keys` (`assignedMemberKeys`).
  - Serialize based on the present of custom headers.
- Define media types in controllers.
- Tests.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/26260
```

danielkelemen:
## Description

- Port of https://github.com/camunda/camunda/pull/25766
- Add String serializer for REST API keys
- Introduce new media types for string and numeric keys & adapt V2 APIs.
- Tests

## Related issues

closes https://github.com/camunda/camunda/issues/26251

### DIFF
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/serializer/StringSerializer.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/serializer/StringSerializer.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.rest.serializer;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.camunda.zeebe.gateway.rest.RequestMapper;
+import java.io.IOException;
+import org.springframework.http.HttpHeaders;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+public class StringSerializer extends JsonSerializer<String> {
+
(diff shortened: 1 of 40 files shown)


zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/serializer/StringSerializer.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    final RequestAttributes requestAttributes = RequestContextHolder.getRequestAttributes();
+    if (requestAttributes instanceof final ServletRequestAttributes r) {
+      final var acceptHeader = r.getRequest().getHeader(HttpHeaders.ACCEPT);
+
+      if (acceptHeader != null && acceptHeader.equals(RequestMapper.MEDIA_TYPE_KEYS_NUMBER_VALUE)) {
+        writeNumber(value, gen);
+        return;
+      }
```
tmetzke, line 71: 💭 One issue we'll run into with this type of handling values is that we have fields that are ending in `key` and are always `string` values. They must not be turned into numbers. One example is the `correlationKey` in message endpoints. This is why it works with the LongSerializer in 8.6: we will write number values as is if needed, but we can also turn them into strings if required. It doesn't work the other way around.

danielkelemen: Ah okay very good point, that's not so good. I'll think about this then, but looks like we won't have such a clean solution as with longs in 8.6...

danielkelemen: So after reconsidering this, I think the best option is to:

- We need to keep using `Long` keys as long as we need to support both strings and longs because we can only register the deserializer for longs. (With 8.9 we could switch entirely.)
- I will look for a new name for string key objects. Ideally something with `*Response` or similar (TBD)
- I will keep the previous names for long key objects. This also means users won't face a big breaking change with 8.7.
- Default response still changes to string keys so there is a breaking change but users can quickly fix it by adding the header.

PR #26261, opened 2024-12-18
test: extract OpensearchIT, ElasticsearchIT from IncidentUpdateRepositoryIT

entangled90:
Extracted from tIncidentUpdateRepositoryIT because static inner classes are not run from maven (but are run from IDEs)


npepinpe (approved):
:pray:

PR #26274, opened 2024-12-19
feat: update corrected user task attributes for `ASSIGNED` and `COMPLETED` events

### LINKED ISSUE #26206: Communicate corrected data to Tasklist
## Description

Adjust the [Camunda-Exporter](https://github.com/camunda/camunda/blob/65ca1f602b25717be0383b2c1834eaf519ec54fb/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java#L80) to update the correctable user task data only for the `UserTaskEvents` that follow operations  that support task listeners:

**Supported events:**
- `ASSIGNED`
- `COMPLETED`

**Correctable Data:**
- `assignee`
- `candidateGroupsList` (might be without list)
- `candidateUsersList` (might be without list)
- `dueDate`
- `followUpDate`
- `priority`

> [!NOTE]
Task listeners are currently supported only for `assign (claim)` and `complete` operations, hence updates should be limited to `ASSIGNED` and `COMPLETED` events.

ce-dmelnych:
## Description

This PR introduces the ability to update corrected user task attributes for the `ASSIGNED` and `COMPLETED` events in the `UserTaskHandler` class.

The changes ensure that any modifications made through task listener jobs are accurately reflected in the Tasklist application.

> This improvement leverages a new `updateChangedAttributes` method to handle attribute updates consistently and cleanly.

## Checklist

- [ ] for CI changes:

## Related issues

closes #26206

### DIFF
--- a/zeebe/exporters/camunda-exporter/pom.xml
+++ b/zeebe/exporters/camunda-exporter/pom.xml
@@ -191,6 +191,12 @@
     </dependency>
 
     <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-protocol-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
--- a/zeebe/exporters/camunda-exporter/pom.xml
+++ b/zeebe/exporters/camunda-exporter/pom.xml
@@ -191,12 +191,6 @@
     </dependency>
 
     <dependency>
-      <groupId>io.camunda</groupId>
-      <artifactId>zeebe-protocol-impl</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
@@ -113,42 +113,6 @@ public class UserTaskHandler implements ExportHandler<TaskEntity, UserTaskRecord
     entity.setJoin(joinRelation);
   }
 
-  /**
-   * Applies changes to the user task fields based on the attributes in the {@link
-   * UserTaskRecordValue}.
-   *
-   * <p>This method can be used for updating fields either:
-   *
-   * <ul>
-   *   <li>As a result of user task corrections configured while completing task listener jobs.
-   *   <li>As a result of regular user task updates.
-   * </ul>
-   *
-   * @param record the record containing user task data and changed attributes
-   * @param entity the task entity to be updated
-   */
-  private void updateChangedAttributes(
-      final Record<UserTaskRecordValue> record, final TaskEntity entity) {
-    final var value = record.getValue();
-
-    for (final String attribute : value.getChangedAttributes()) {
-      entity.getChangedAttributes().add(attribute);
-
-      switch (attribute) {
-        case "assignee" -> entity.setAssignee(value.getAssignee());
-        case "candidateGroupsList" ->
-            entity.setCandidateGroups(value.getCandidateGroupsList().toArray(new String[0]));
-        case "candidateUsersList" ->
-            entity.setCandidateUsers(value.getCandidateUsersList().toArray(new String[0]));
-        case "dueDate" -> entity.setDueDate(ExporterUtil.toOffsetDateTime(value.getDueDate()));
-        case "followUpDate" ->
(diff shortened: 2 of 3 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskHandlerTest.java
```diff
@@ -267,6 +268,85 @@ void shouldUpdateEntityFromRecordForCompletedIntent() {
...
+        .describedAs(
+            "Expected not changed user task record fields to have `null` values in task entity")
+        .extracting(
+            TaskEntity::getAssignee,
+            TaskEntity::getDueDate,
+            TaskEntity::getFollowUpDate,
+            TaskEntity::getPriority,
+            TaskEntity::getCandidateGroups,
+            TaskEntity::getCandidateUsers)
+        .containsOnlyNulls();
```
ana-vinogradova-camunda, line 378: Great check here 👍

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
```diff
@@ -249,4 +210,75 @@ private void createTaskEntity(final TaskEntity entity, final Record<UserTaskReco
...
+      switch (attribute) {
+        case "assignee" -> entity.setAssignee(value.getAssignee());
+        case "candidateGroupsList" ->
+            entity.setCandidateGroups(value.getCandidateGroupsList().toArray(new String[0]));
+        case "candidateUsersList" ->
+            entity.setCandidateUsers(value.getCandidateUsersList().toArray(new String[0]));
+        case "dueDate" -> entity.setDueDate(ExporterUtil.toOffsetDateTime(value.getDueDate()));
+        case "followUpDate" ->
+            entity.setFollowUpDate(ExporterUtil.toOffsetDateTime(value.getFollowUpDate()));
+        case "priority" -> entity.setPriority(value.getPriority());
```
ce-dmelnych, line 255: 💭 We could use `UserTaskRecord.ASSIGNEE` and other constants here if we made `zeebe-protocol-impl` a compile-scope dependency. However, since `zeebe-protocol-impl` was not explicitly defined as a production dependency in the `camunda-exporter` pom file before, I opted to retain string literals as used previously.

❔I’d appreciate your thoughts on introducing `zeebe-protocol-impl` as a compile-scope dependency to improve the maintainability or if there are reasons to avoid making it a compile dependency.
@panagiotisgts @korthout @ana-vinogradova-camunda

korthout: Personally, I'd avoid it, as to not open the door for other unintended usages. But the owning team may have different ideas.

korthout (approved):
LGTM 👍 

💡 Perhaps @marcosgvieira or @tsedekey is willing to review as well

zeebe/exporters/camunda-exporter/pom.xml
```diff
@@ -190,6 +190,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-protocol-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
```
korthout, line 197: 💭 Personally, I'd leave this out, as it's only use is to access few strings that should probably be part of the zeebe-protocol itself. Those can simply be re-defined in the test as constants.

ce-dmelnych: Thanks for the feedback, Nico! I've opted to replace the references to `UserTaskRecord` constants with string literals, as we currently have only a few tests where these attribute names are used. 

Initially, I used the constants from `UserTaskRecord` because `zeebe-protocol-impl` was available as a transitive dependency via `zeebe-protocol-test-util`. However, `maven-dependency-plugin` failed the build requiring to define it explicitly, so I've removed the definition of `zeebe-protocol-impl` dependency from the `camunda-exporter` test scope.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
ce-dmelnych: 💬 Thanks, Nico! I agree with your point about avoiding `zeebe-protocol-impl` as a compile-scope dependency to prevent unintended usages. Given this, I've decided to stick with string literals for now to keep things simple and avoid introducing new dependencies in the production code.

If the owning team has a different preference or if these attribute constants become part of the `zeebe-protocol` itself in the future, we can revisit this decision.

panagiotisgts (approved):
LGTM as well. I would also, agree in not bringin `zeebe-protocol-impl` on the Exporter. We could have these changeset values defined somewhere else, in the future.

PR #26275, opened 2024-12-19
feat: add Database.Type enum in search-connector module

entangled90:
## Description
A new enum `Database.Type` enum is added in the common search-connect module.
- static string are added so they can be used in annotations
- code in dist/ and zeebe uses this new annotation

The outer class `Database` is used to scope the static string values  and for namespacing

## Related issues
Comments from PR #26196

### DIFF
--- a/dist/src/main/java/io/camunda/application/commons/backup/DynamicIndicesConfiguration.java
+++ b/dist/src/main/java/io/camunda/application/commons/backup/DynamicIndicesConfiguration.java
@@ -18,8 +18,6 @@ import org.springframework.context.annotation.Configuration;
 @ProfileWebApp
 public class DynamicIndicesConfiguration {
 
-  private static final String ELASTICSEARCH_TYPE = "elasticsearch";
-
   private final ConnectConfiguration connectConfiguration;
   private final DocumentBasedSearchClient searchClient;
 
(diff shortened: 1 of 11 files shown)


npepinpe (approved):
:wrench: There's a `ConnectionTypes` in the Camunda Exporter which can likely be replaced by this new enum.

search/search-client-connect/src/main/java/io/camunda/search/connect/configuration/Database.java
```diff
@@ -0,0 +1,47 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.search.connect.configuration;
+
+public final class Database {
+  public static final String ElasticSearchString = "elasticsearch";
+  public static final String RdbmsString = "rdbms";
+  public static final String OpenSearchString = "opensearch";
+
+  public enum Type {
```
npepinpe, line 15: :wrench: Why not just have a single top level type `DatabaseType`? Also would be great if I can get the enum value from a string, as this will be a common case.

entangled90:
I did not change `ConnectionTypes` because it does not have the value "rdbms", is it planned for it as well?

npepinpe:
Not in the exporter, but why can't the exporter also use this enum? :thinking: Even if it doesn't use all of the types.

entangled90:
Yes it could be used, but it can be easy to forgot to validate it everywhere :man_shrugging:

PR #26290, opened 2024-12-19
feat: support m2m authentication for management identity

### LINKED ISSUE #25823: Creating m2m token for Management Identity is supported

maryarm:
## Description

Provide m2m token for calling the management identity endpoint.

Also a bug fixed: when management identity endpoint is not available, it caused never ending retries on migration app, now it fixed by returning an empty array in this case.

## Related issues

closes #25823

### DIFF
--- a/migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ManagementIdentityConnector.java
+++ b/migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ManagementIdentityConnector.java
@@ -7,27 +7,104 @@
  */
 package io.camunda.migration.identity.config;
 
+import io.camunda.identity.sdk.Identity;
+import io.camunda.identity.sdk.IdentityConfiguration;
 import io.camunda.migration.identity.midentity.ManagementIdentityClient;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpRequest;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.client.ClientHttpRequestExecution;
+import org.springframework.http.client.ClientHttpRequestInterceptor;
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.web.client.DefaultResponseErrorHandler;
 
 @Configuration
 public class ManagementIdentityConnector {
 
-  @Bean // will be closed automatically
+  @Bean
+  public Identity identity(final IdentityMigrationProperties identityMigrationProperties) {
+    final var properties = identityMigrationProperties.getManagementIdentity();
+    final var configuration =
+        new IdentityConfiguration(
+            properties.getBaseUrl(),
+            properties.getIssuerBackendUrl(),
+            properties.getIssuerBackendUrl(),
+            properties.getClientId(),
+            properties.getClientSecret(),
+            properties.getAudience(),
+            properties.getIssuerType());
+    return new Identity(configuration);
+  }
+
+  @Bean
   public ManagementIdentityClient managementIdentityClient(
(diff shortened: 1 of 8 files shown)


migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ManagementIdentityConnector.java
```diff
@@ -7,27 +7,104 @@
...
+      }
+      super.handleError(response);
+    }
+
+    @Override
+    public byte[] getResponseBody(final ClientHttpResponse response) {
+      try {
+        if (response.getStatusCode() == HttpStatus.NOT_FOUND) {
+          return "[]".getBytes(StandardCharsets.UTF_8);
+        }
```
lenaschoenburg, line 103: :question: What's the scenario where this is necessary? If the connection to Identity is configured but some endpoints return 404, we probably don't want to silently ignore this, right?

qa/integration-tests/src/test/java/io/camunda/it/migration/IdentityDefaultEntitiesIT.java
```diff
@@ -49,8 +48,14 @@ void canMigrateOldDefaultEntities() {
...
-            "camunda.migration.identity.managementIdentity.m2mToken",
-            getIdentityAccessToken(KEYCLOAK))
+            "camunda.migration.identity.management-identity.issuer-backend-url",
+            IdentityMigrationTestUtil.externalKeycloakUrl(KEYCLOAK) + "/realms/camunda-platform/")
+        .withProperty("camunda.migration.identity.management-identity.issuer-type", "KEYCLOAK")
+        .withProperty("camunda.migration.identity.management-identity.client-id", "migration-app")
+        .withProperty("camunda.migration.identity.management-identity.client-secret", "secret")
+        .withProperty(
+            "camunda.migration.identity.management-identity.audience",
+            "camunda-identity-resource-server")
```
lenaschoenburg, line 58: :wrench: We should introduce constants in `IdentityMigrationTestUtil` for this.

migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ManagementIdentityConnector.java
```diff
@@ -7,27 +7,104 @@
...
+    public M2MTokenInterceptor(final Identity identity, final String audience) {
+      this.identity = identity;
+      this.audience = audience;
+    }
+
+    @Override
+    public ClientHttpResponse intercept(
+        final HttpRequest request, final byte[] body, final ClientHttpRequestExecution execution)
+        throws IOException {
+      final String token = identity.authentication().requestToken(audience).getAccessToken();
```
lenaschoenburg, line 76: :question: Shouldn't we cache this somehow? When a couple of SaaS clusters are updating at the same time, we don't want to spam the management identity with token requests for every single request we make, right?

migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ManagementIdentityConnector.java
maryarm: We can add a log here but then ignore, because it's possible that we run migration app with a version of identity that doesn't support migration and then migration stuck in never ending retries.

migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ManagementIdentityConnector.java
maryarm: identity sdk is managing the cache itself.

migration/identity-migration/src/main/java/io/camunda/migration/identity/config/ManagementIdentityConnector.java
lenaschoenburg: I think my problem with this is that it is a very blunt tool. Ignoring all 404 feels too dangerous, I'm afraid that we will miss actual bugs with this.

lenaschoenburg (approved):
Thanks, looks good now :tada:

PR #26291, opened 2024-12-19
Migrate joining inclusive gateway with at least one incoming sequence flow is taken

### LINKED ISSUE #25674: Migrate joining inclusive gateway with at least one incoming sequence flow is taken
## Description

First inclusive gateway migration case is active parallel gateway due to an incident is explained in issue #25673. Second case to migrate a joining inclusive gateway is when at least one incoming sequence flow of the gateway is taken.

In this case, users **HAVE TO** provide a mapping instruction for the inclusive gateway. That way we make sure taken sequence flow count on the gateway will remain the same in the target. Let's consider example:

Screenshot 2024-12-05 at 18 54 30

In the example above, assume that task `A` is completed and sequence flow `1` is taken. Users **have to** provide mapping instructions for `B -> B` and `join -> join`.

berkaycanbc:
## Description

Similar to migrating parallel gateways, migrating inclusive gateways with at least one incoming sequence flow is taken is enabled in this PR. See the actual implementation in #25998 

## Related issues

closes #25674


korthout (approved):
LGTM 💯

Issue #26162, opened 2024-12-16
Auto-enable Migration when configuration are set


ChrisKujawa:
## Description

**Target: StandaloneCamunda**

To make migration easy as possible and do not need to configure too much as a user, we should try to interpolice configurations based on the given.

This is mostly to make it easy for StandaloneCamunda, and helm charts.

For example:

1. When Tasklist profile is enabled (default on StandloneCamunda) we should auto-enable the process migration profile.
    2. We can make use of the `camunda.database` config, as it is inherited (validate with @panagiotisgts )
2. When (old) Identity connection configurations are set, then we should enable identity migration
    3.  this likely useful for Helm charts, as we could reuse values and simply set them on the new StandaloneApplication

**Note:** When deploying Migration application, we have to configure everything explicitly.

PR #26304, opened 2024-12-19
feat: auto enable migration when tasklist is enabled

ChrisKujawa:
## Description

When Tasklist profile is enabled (default on StandloneCamunda) we should auto-enable the process migration profile, as it makes just sense to migrate related data.

We can make use of the camunda.database config, as it is inherited

## Related issues

related to https://github.com/camunda/camunda/issues/26162


panagiotisgts:
:thinking: I think we need some more changes to support this. Since the migration operates on the `operate-process` index, this will mean that every time a new Process Definition is being created, next time Tasklist ( or the C8 bundle ) starts it will migrate this _new_ Process Definition. Even though it's already been parsed by the respective CamundaExporter Handler correctly.

panagiotisgts:
@ChrisKujawa An optimization to keep in mind. I don't think this could get us in trouble, we could later look this up again.

panagiotisgts (approved):

Issue #19883, opened 2024-06-28
Fix REST client connection handling


tmetzke:
**Describe the bug**

The following integration tests don't work as expected over REST:
* `shouldActivateJobsIfBatchIsTruncated` in the [LongPollingActivateJobsTest](https://github.com/camunda/camunda/blob/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/LongPollingActivateJobsTest.java)
* `shouldActivateJobForOpenRequest` in the [LongPollingActivateJobsTest](https://github.com/camunda/camunda/blob/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/LongPollingActivateJobsTest.java)

**To Reproduce**

1. Add the option `true` to the test cases in the `booleans` array in the `ValueSource` annotation, basically reverting [this commit](https://github.com/camunda/camunda/pull/19333/commits/9cbadfb14bf60fcd1b695ea83684b53360d9db4f).
2. Run the tests

**Expected behavior**

The tests pass.

**Log/Stacktrace**

* The tests fail with a timeout on the Camunda client 

**Environment:**
- OS: all
- Zeebe Version: 8.6.0-alpha3
- Configuration: long polling enabled in the gateway

**Workaround**

Switch to using the gRPC protocol for job activation using long polling.

**Additional context**

This limitation is documented in our docs with https://github.com/camunda/camunda-docs/issues/4375

tmetzke:
## Observations

Copied from https://github.com/camunda/camunda/pull/19333#issuecomment-2196947497

> @koevskinikola and I investigated the other failing IT (`shouldActivateJobForOpenRequest`):
> * If we remove `sendActivateRequestsAndClose` AFTER the test's relevant client (lines 117/118) is opened (i.e. removing it from line 120), the test passes.
> * The `sendActivateRequestsAndClose` before the test's relevant client is not affecting the result in any way.
> * It looks like the connection of our test client is interrupted if new requests come in.
> * It doesn't exactly explain why the other test case fails but could have the same root cause (long-running connections not working on the server side).

tmetzke:
## Observations round 2

After another round of manual tests using cURL locally, I strongly assume the issue resides on the client instead of the server.
All IT scenarios are successful using cURL manually. All IT scenarios also fail using the Java client in a "production" setup, i.e. the test setup is not the root cause.

### Setups

#### Setup A
1. Start the `StandaloneBroker`.
[...]

PR #26317, opened 2024-12-20
fix 19883 bug on REST connection pooling

entangled90:
## Description
- the HTTP client does not limit the amount of data it receives (we could still add a safety limit on top as a multiplier of the current capacity)
- The broker response size (in SBE) is used to bound the maximum amount of data sent to the rest client as it cannot be known before hand (and serializing it at that stage is wasteful)

> [!WARNING]
> shouldActivateJobForOpenRequest in the [LongPollingActivateJobsTest](https://github.com/camunda/camunda/blob/main/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/client/command/LongPollingActivateJobsTest.java)
It's still not passing, but I'm not sure how if it's related to the same issue?

## Related issues

relates #19883

### DIFF
--- a/clients/java/src/main/java/io/camunda/client/impl/http/ApiEntityConsumer.java
+++ b/clients/java/src/main/java/io/camunda/client/impl/http/ApiEntityConsumer.java
@@ -52,14 +52,14 @@ final class ApiEntityConsumer<T> extends AbstractBinAsyncEntityConsumer<ApiEntit
       Arrays.asList(ContentType.TEXT_XML);
   private final ObjectMapper json;
   private final Class<T> type;
-  private final int maxCapacity;
+  private final int chunkSize;
 
   private TypedApiEntityConsumer<T> entityConsumer;
 
-  ApiEntityConsumer(final ObjectMapper json, final Class<T> type, final int maxCapacity) {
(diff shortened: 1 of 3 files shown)


clients/java/src/main/java/io/camunda/client/impl/http/ApiEntityConsumer.java
```diff
@@ -83,7 +83,7 @@ protected ApiEntity<T> generateContent() throws IOException {
 
   @Override
   protected int capacityIncrement() {
-    return maxCapacity - entityConsumer.getBufferedBytes();
+    return chunkSize;
```
npepinpe, line 86: :question: So we still silently ignore when responses would be too large right? :(

entangled90: No, with this change the client is always able to receive more data, in chunks of `chunkSize`.

We should probably put a hard limit (like 4x chunkSize) and throw an exception if that is breached.

npepinpe: Ah, yeah, got it :+1:

npepinpe (approved):

PR #26329, opened 2024-12-20
feat: RPA files REST endpoint

kzraitel:
## Checklist

- [ ] for CI changes:

## Related issues

closes #


kzraitel:
Merge after https://github.com/camunda/camunda/issues/24772

koevskinikola:
Hey @kzraitel, I changed the PR base from `main` to the `24772-Support-Resource-in-DeploymentCreateProcessor` branch so that I could filter out the relevant changes. The base will be reverted to `main` once #25918 is merged. I can also put it back to `main` if you prefer.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -7390,6 +7390,8 @@ components:
           $ref: "#/components/schemas/DeploymentDecisionRequirements"
         form:
           $ref: "#/components/schemas/DeploymentForm"
+        resource:
+          $ref: "#/components/schemas/DeploymentRpa"
```
koevskinikola, line 7394: ❌ We should generalize this part as well. To ensure that we can easily expand support for other types of resources, we should name this schema (and the `DeploymentRpaNumberKeys`) to `DeploymentResource` and `DeploymentResourceNumberKeys` appropriately.

We will prevent the deployment of other types of resources with the code implemented [here](https://github.com/camunda/camunda/pull/25918/files#diff-1ba96a098bc7b39fc9ef5c0aea87dfd2184a73a79b0f7574f01067d27cbd71bdR83), and through documentation on what types of resources are supported.

If we name this schema `DeploymentRpa` we will have to add new entities for each new resource type we add support for.

koevskinikola (changes requested):
The code looks good @kzraitel. However, it would be good to generalize the implementation here as well, so that responses provide generic "resource" metadata, even though we currently only support RPA resources for it.

This makes the endpoint future-proof for further resource type expansions.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -7662,6 +7666,44 @@ components:
...
+          description: |
+            The RPA ID, as parsed during deployment, together with the version forms a
+            unique identifier for a specific form.
+        version:
+          type: integer
+          format: int32
+          description: The assigned RPA version.
+        resourceName:
+          type: string
+          description: The resource name from which this RPA was parsed.
```
koevskinikola, line 7684: ❌ We still have mentions of RPA in this section. We should generalize it to `resource`.

```suggestion
            The resource ID, as parsed during deployment, together with the version forms a
            unique identifier for a specific form.
        version:
          type: integer
          format: int32
          description: The assigned resource version.
        resourceName:
          type: string
          description: The resource name from which this resource was parsed.
```

koevskinikola (approved):
Hey @kzraitel, the code looks good. 👍  

I only have some minor feedback related to the REST API docs, but I'm pre-approving this PR, so you can merge it as soon as the REST API docs are fixed.

PR #26331, opened 2024-12-20
fix: use owner key and resource type as primary key for authorizations

### LINKED ISSUE #26324: Camunda exporter does not create or update authorizations correctly
The `AuthorizationHandler` in the Camunda exporter uses the record key to create documents. This is incorrect, as the Engine does not use meaningful record keys for authorizations and instead relies on owner key + resource type to uniquely identify authorizations.

As a result, authorizations created with the same key (as done in `IdentitySetupInitializeProcessor`) get overwritten which completely breaks the resource based authorization feature.

lenaschoenburg:
This fixes a bug where the authorization record key was used for the document id of exported authorizations. This is incorrect as the record key can be arbitrary. Instead, we now use a composite of owner key and resource type which uniquely identifies the authorization. The engine uses the same composite key internally. This is also similar to the RDBMS exporter which uses owner key + owner type + resource type. Since owner key is unique across all owner types, we can safely omit this in the Camunda exporter.

Closes #26324

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/AuthorizationHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/AuthorizationHandler.java
@@ -18,8 +18,9 @@ import io.camunda.zeebe.protocol.record.value.AuthorizationRecordValue.Permissio
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


lenaschoenburg:
This is currently broken as the `AuthorizationHandler` just overwrites permissions instead of adding/removing them. This is also discussed here: https://github.com/camunda/camunda/pull/26045#issuecomment-2557761089

PR #26337, opened 2024-12-20
feat: classify some indices as non required that can be skipped if missing

### LINKED ISSUE #26120: Webapps backup fails if one application is not deployed
The backup of the webapps can succeed only if all webapps all indices have been created. 
The CamundaExporter will create all indices for tasklist & operate, but not for optimize.

We should be able to support a deployment where optimize is not deployed
However, the active profiles are not known cluster-wide currently.

When trying to perform a backup, if it fails because an optimize index is missing, we should skip all optimize indices.

entangled90:
## Description

Because optimize may not be deployed in all situations, its indices may not be present in ES/OS. If they are missing, the backup is retried without these indices.

The error type used is "index_not_found_exception:, I tested locally with opensearch by creating a snapshot of a non-existent index (named `missing`):
```
➜  curl localhost:9200/_snapshot/myrepo/mysnapshot -X PUT -d '{"indices": ["missing"]}' -H 'Content-Type:application/json'

{
  "error": {
    "root_cause": [
      {
        "type": "index_not_found_exception",
        "reason": "no such index [missing]",
        "index": "missing",
        "resource.id": "missing",
        "resource.type": "index_or_alias",
        "index_uuid": "_na_"
      }
    ],
    "type": "index_not_found_exception",
    "reason": "no such index [missing]",
    "index": "missing",
    "resource.id": "missing",
    "resource.type": "index_or_alias",
    "index_uuid": "_na_"
  },
  "status": 404
}

```

## Related issues

closes #26120

### DIFF
--- a/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
+++ b/webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
@@ -53,6 +53,7 @@ import org.slf4j.LoggerFactory;
 public class ElasticsearchBackupRepository implements BackupRepository {
   public static final String SNAPSHOT_MISSING_EXCEPTION_TYPE = "snapshot_missing_exception";
   private static final String REPOSITORY_MISSING_EXCEPTION_TYPE = "repository_missing_exception";
+  private static final String INDEX_MISSING_EXCEPTION_TYPE = "index_not_found_exception";
   private static final Logger LOGGER = LoggerFactory.getLogger(ElasticsearchBackupRepository.class);
   private final ElasticsearchClient esClient;
   private final BackupRepositoryProps backupProps;
@@ -240,6 +241,7 @@ public class ElasticsearchBackupRepository implements BackupRepository {
   @Override
   public void executeSnapshotting(
       final BackupService.SnapshotRequest snapshotRequest,
+      final boolean onlyRequired,
       final Runnable onSuccess,
       final Runnable onFailure) {
     final var request =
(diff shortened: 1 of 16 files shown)


entangled90:
Should we add a config to enable this feature? So for example, we disable it when we release in Saas as we know optimize is released and we avoid not doing the backup for optimize in case of bugs related to optimize indices

npepinpe:
I'm not super keen on this method of detection. I would prefer something more explicit honestly at start up, but :shrug: I understand it's complicated if we support the standalone deployments.

This is slowly making me think we should move to a simpler, dual-supported topology: either all in one node, or webapp/gateway and brokers (so only 2 deployments), if only to reduce maintenance effort. But out of scope for now.

webapps-backup/src/main/java/io/camunda/webapps/backup/repository/elasticsearch/ElasticsearchBackupRepository.java
```diff
@@ -269,7 +271,15 @@ public void executeSnapshotting(
             final var response = esClient.snapshot().create(request);
             listener.onResponse(response);
           } catch (final Exception e) {
-            listener.onFailure(e);
+            if (isErrorType(e, INDEX_MISSING_EXCEPTION_TYPE) && !onlyRequired) {
+              // Retry taking only the required indices
+              LOGGER.info(
```
npepinpe, line 276: :wrench: Is this useful information for the user? It means I will be logging this _every time_ if I never deploy Optimize, right? I would downgrade as debug (and I guess this highlights the limitations of this approach).

npepinpe (approved):
As I said, not my favorite option - can we add a follow up to eventually find a better option? Like being able to tell programmatically if Optimize is deployed in this cluster, so I would know what indices to expect.

PR #26347, opened 2024-12-20
feat: support context path in identity client

emulbreh:
## Description

Support a non-empty context path (`SERVER_SERVLET_CONTEXT_PATH`) in identity client. This is necessary for SaaS.

### DIFF
--- a/identity/client/src/App.tsx
+++ b/identity/client/src/App.tsx
@@ -1,11 +1,11 @@
 import { FC, StrictMode } from "react";
 import { BrowserRouter } from "react-router-dom";
-import { baseUrl } from "./configuration";
+import { getBaseUrl } from "./configuration";
 import AppRoot from "./components/global/AppRoot";
 import GlobalRoutes from "src/components/global/GlobalRoutes";
 
 const App: FC = () => (
-  <BrowserRouter basename={baseUrl}>
+  <BrowserRouter basename={getBaseUrl()}>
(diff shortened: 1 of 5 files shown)


Ben-Sheppard (approved):
As I'm more comfortable with Java I pulled the code and ran some tests locally, it works as I'd expect with one exception. 

I spotted something but I'm not convinced is down to this change, if I start the app with a context path and perform an action (i.e. create a group), the group is created and exported. If I then stop the app, remove the context path, start the app, refresh the page, create a group, the group call is successful and Zeebe thinks the group exists but it doesn't seem to be exported.

As I don't think that is related just to this change (seems strange that support for context path in Identity would impact that) I'm going to approve and look to raise a separate bug.

emulbreh:
> As I'm more comfortable with Java I pulled the code and ran some tests locally, it works as I'd expect with one exception.
> 
> I spotted something but I'm not convinced is down to this change, if I start the app with a context path and perform an action (i.e. create a group), the group is created and exported. If I then stop the app, remove the context path, start the app, refresh the page, create a group, the group call is successful and Zeebe thinks the group exists but it doesn't seem to be exported.
> 
> As I don't think that is related just to this change (seems strange that support for context path in Identity would impact that) I'm going to approve and look to raise a separate bug.

That's weird, but I agree that this frontend-only change cannot have this effect on the exporter.

PR #26358, opened 2024-12-20
deps: bump tomcat out of band to resolve a CVE

npepinpe:
## Description

This PR bumps Tomcat to resolve a CVE: CVE-2024-50379. While it does not affect C8, I've opened the PR in case not fixing it increases support load. I seriously doubt we would ever start using JSP, but who knows.

Preference would be getting #25062 merged, but it's blocked by a breaking Apache HTTP update (see https://github.com/camunda/camunda/pull/22241).

So if those are merged before, then we can close this. But in case they aren't, we can go ahead and merge this one.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -1031,6 +1031,26 @@
         <version>${version.spring-boot}</version>
       </dependency>
 
+      <!-- TODO: remove when https://github.com/camunda/camunda/pull/25062 is merged -->
+      <dependency>
+        <groupId>org.apache.tomcat.embed</groupId>
+        <artifactId>tomcat-embed-core</artifactId>
+        <version>10.1.34</version>
+      </dependency>
+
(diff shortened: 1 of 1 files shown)


houssain-barouni:
Hi @npepinpe , spring boot 3.3.7 (released last week) depends on tomcat `10.1.34` 
So it is better to bump spring boot version

npepinpe:
We have a pending update for 3.4.1, but it's blocked by a different update (on Apache HTTP). 

I suppose Renovate will not open an update for 3.3.7 if 3.4.1 is already opened, so we could bump to 3.3.7 for now :+1:

npepinpe:
Closing, @houssain-barouni will update the monorepo to a fixed version 3.3.7, and we'll still hopefully merge the 3.4.x eventually :smile:

Issue #26396, opened 2024-12-23
Unknown channel option 'SO_SNDBUF'


npepinpe:
**Description**

It seems we introduced new send and receive buffer parameters, and applied them not only to our Netty transport, but also to gRPC's Netty transport.

This should be reverted, as:

1. gRPC's Netty channels don't understand the `SO_SNDBUF` options
2. We should let gRPC properly configure its own channels

We can keep these options for our own transport.

filipecampos:
Bottomline, I agree with making the warning disappear. I don't have any strong feelings regarding the option, but perhaps we can use this opportunity to simply add two new buffer size parameters just for the gRPC channels and use the default values for now.

entangled90:
I don't see many childOptions set by default in grpc's NettyServer, see [NettyServer](https://github.com/grpc/grpc-java/blob/735b3f3fe6c6b9d8fc6019e8c89f81c065d9b0d6/netty/src/main/java/io/grpc/netty/NettyServer.java#L218), so I'm not sure which values these parameter will have. I see that the initial control flow window is set to 1MB.

It's hard to justify changing these defaults without the support of a benchmark, so for the moment we can just remove them.
If in the future we want to add properties for them we will just add them later.

PR #26403, opened 2024-12-24
refactor: apply SO_SNDBUF to child channels

npepinpe:
## Description

This PR applies the `SO_SNDBUF` channel configuration as a child option for gRPC channels. It's invalid to apply it to the main server channel, since it's used _only_ to receive requests. Whenever it receives one, it will spawn an ephemeral so-called "child" channel to handle it, and only there will it send anything.

Previously, this resulted in a warning at startup, saying we were applying an unknown option, resulting in the configuration option not being used at all.

> [!Note]
> I'm not marking this as closing the issue, as I'd like to challenge if we really want to overwrite gRPC defaults.

## Related issues

related to #26396


filipecampos (approved):
I agree with the changes due to the need to remove the warning. 
We could either remove the options entirely for gRPC or define new exclusive parameters to set different values from the existing ones.

Issue #24521, opened 2024-11-08
Zeebe 8.5 leaks backpressure in-flight spots


lenaschoenburg:
**Describe the bug**

We collected a heap dump from a leader that appears to be fully healthy, but rejects 100% of all requests. The heap dump shows that there are no free in-flight spot, even though processing has reached the end of the log without any errors.
As a workaround, the broker can be restarted.

**To Reproduce**

Not clear but there were many back-to-back leader changes before.

**Expected behavior**

In-flight spots should be freed up reliably and not get leaked

**Internal Resources**
Heap dump: https://drive.google.com/file/d/1vonp6zfFjz9U2Vzwt6QosfjheQpXtvJS/view
Discussion: https://camunda.slack.com/archives/C05DH1F5TAR/p1730727482749399

**Environment:**
- Zeebe Version: 8.5

entangled90:
Just some notes from looking at the heap dump:
 The `CommandRateLimiter.inFlight` counter is equal to 110, while the limit is 100.
 The number of listeners in the map `responseListeners` is equal to 110 (so the same as inFlights, no leak here)
 There is only partition 2 in this dump (is this correct?)
 There are some `Limiter` (create in function `acquire`) that were created even if the inFlight count was > 100, see the value of `currentInFlight` from screenshot 
![image](https://github.com/user-attachments/assets/2946aa9a-4647-48d0-a0a0-ca5793e3f4f5)
This might suggest that there is some concurrency issue on that and in fact the `acquire` method is not thread-safe as it checks the `inFlight` value without a CAS loop. However, `tryAcquire` is called only from `CommandApiRequestHandler` (which is an actor)

filipecampos:
The value of the `CommandRateLimiter.inFlight` counter can be higher than the `CommandRateLimiter.limit` in valid situations. For instance, this can occur when the intent is a whitelisted command because we always accept it, disregarding these values.
However, I agree with the concurrency issue you've identified and confirmed its occurrence with a unit test using multiple concurrent invocations to `CommandRateLimiter.tryAcquire` using `ProcessInstanceCreationIntent.CREATE`, which is not a whitelisted command, as the intent since the `inFlight` counter reached 101 and the limit was 100.
I'll create a patch to fix this but I do not think this is the exclusive root cause as the `inFlight` counter should be able to decrease anyway with the conclusion of the intents whether they're processed successfully, dropped or ignored.

npepinpe:
[...]

PR #26408, opened 2024-12-24
Fixes Zeebe 8.5 Limiters inFlight concurrent access

### LINKED ISSUE #26104: Backpressure at 100% after backup & restore
**Describe the bug**

We performed a backup & restore procedure on multiple clusters of different versions for an Elasticsearch migration. The procedure and the migration were successful. However, some clusters reported a backpressure of 100% after the restore. As a result, the processing stopped completely.   

We could not find the root cause of the backpressure in the error reporting or the metrics dashboard. All clusters and partitions reported a healthy state. The clusters didn't recover from the backpressure. All inflight requests were stuck. A restart of all brokers didn't change the situation. 

Finally, we **resolved the issue by disabling the backpressure** (via the environment variable `ZEEBE_BROKER_BACKPRESSURE_ENABLED`). The clusters continued processing as normal. We enabled the backpressure again and the clusters were still processing. Like a hiccup, everything was back to normal. 

:lock: Incident: [INC-2077](https://app.incident.io/camunda/incidents/2077)
:lock: [Incident Slack channel](https://camunda.slack.com/archives/C083SNG2TA7)

**To Reproduce**

We don't know. The issue occurred after a successful backup and restore procedure. The Zeebe version was not updated. 

**Hypothesis** (about the command limiters not being released):
- `CommandApiServiceImpl#getOnProcessedListener` is registered to the stream processor as a processing listener.
    - This listener completes the limiter via the record's request metadata.
    - If the metadata is missing, this is not called, and the limiter is not released.
    - If the metadata is incorrect, the limiter cannot be released.
- We clear the limiter whenever a partition transitions. However, the metric is not clear, which can lead to confusion.
    - If clearing does not work, then we would have old limiters which would not be released again.

Could be related to:
- https://github.com/camunda/camunda/issues/24521
- https://github.com/camunda/camunda/issues/20126

**Expected behavior**

A cluster continues processing after a backup and restore procedure.  

**Log/Stacktrace**

There were no warnings or errors related to the issue. Only warnings about failed job activation attempts. 

**Observations**:
- The inflight requests were not cleaned
- Nothing is written from the CommandAPI to the Journal
- The Sequencer is empty, likely because we can write more
- The request ID 258604061744431105 has an open limiter in the command API limiter of partition 3 in the heap dump.
    - This ID is a snowflake ID - when you extract the timestamp (the first 41 bits) by dropping the last 22 bits (the node and ID bits), adjusting for the 2023 offset, we get a timestamp of 2024-12-14 14h40 GMT (so 15h40 CEST)
    - Looking at Grafana, this is when the node last restarted.
    - If we look at the log, we can find the position command with this ID (and it is not duplicated), and it is 2771846695.
    - This position is < than the last processed position, meaning we probably processed it.
    - I say probably because I wonder if the restore happened while the node was offline or not 

<details>
  <summary>Screenshots</summary>

image
image
image
image
image
image

</details>

:lock: We stored data of an affected partition leader in GDrive [inc-migrate-legacy-es-prod](https://drive.google.com/drive/u/0/folders/1hNRUfcdYoBZEUc0ZbkA9RM7Ip01GLKdt). 

**Environment:**
- OS: 
- Zeebe Version: 8.3.16, 8.4.1, 8.4.13, 5.8.9
- Configuration: SaaS

filipecampos:
## Description
By analyzing the heap dump in #24521, we noticed that the value of `AbstractLimiter.inFlight` was higher than `AbstractLimiter.limit` in some cases. Although this can be valid, even if `inFlight` equals `limit` and the received intents are whitelisted commands, the intent is accepted to be processed and the `inFlight` counter incremented, possibly being assigned a value above `limit`.
We also noticed that the number of `inFlight` requests for some partitions of the analyzed clusters in #26104 when back-pressure was at 100%, could be above the `limit`.
By looking at the code in `CommandRateLimiter.acquire` and `AppendLimiter.acquire`, although `inFlight` is an `AtomicInteger`, we found that the comparison to the value of `limit` was performed without any synchronization mechanism and `inFlight` would be incremented independently afterwards in the call to `AbstractLimiter.createListener`.

With the introduced unit tests for both `CommandRateLimiter` and `AppendLimiter`, we were able to concurrently increment `inFlight` up to 101 with a `limit` of 100.

The changes to these 2 limiters enforces this condition, that `inFlight` < `limit` always, except for the whitelisted commands in the case of `CommandRateLimiter`.
The decision to use a `synchronized` section is a quick fix and also avoids the need to extend or reimplement the `createListener` method, which is inherited from `AbstractListener`, twice.

## Related issues

partially closes #26104 and #24521

### DIFF
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiterTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiterTest.java
@@ -9,6 +9,7 @@ package io.camunda.zeebe.broker.transport.backpressure;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.netflix.concurrency.limits.Limit;
 import com.netflix.concurrency.limits.limit.SettableLimit;
 import io.camunda.zeebe.protocol.record.intent.CommandDistributionIntent;
 import io.camunda.zeebe.protocol.record.intent.DeploymentDistributionIntent;
@@ -17,6 +18,12 @@ import io.camunda.zeebe.protocol.record.intent.Intent;
 import io.camunda.zeebe.protocol.record.intent.JobIntent;
 import io.camunda.zeebe.protocol.record.intent.ProcessInstanceCreationIntent;
 import io.camunda.zeebe.protocol.record.intent.ProcessInstanceIntent;
+import java.util.Collection;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


zeebe/broker/src/test/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiterTest.java
```diff
@@ -109,4 +116,77 @@ private static Stream<Arguments> provideWhitelistedIntents() {
         Arguments.of(DeploymentDistributionIntent.COMPLETE),
         Arguments.of(CommandDistributionIntent.ACKNOWLEDGE));
   }
+
+  @Test
+  void shouldNotAllowInFlightHigherThanLimitWithNormalCommands() throws InterruptedException {
+    // given
+    final int numThreads = 3000;
+    final int poolSize = 300;
```
entangled90, line 124: :thought_balloon:  How fast does it take to run this test? we can probably use a smaller limit and less threads (like 20/30)

zeebe/broker/src/test/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiterTest.java
```diff
@@ -109,4 +116,77 @@ private static Stream<Arguments> provideWhitelistedIntents() {
...
+
+    final Collection<Callable<Object>> tasks =
+        IntStream.range(0, numThreads)
+            .mapToObj(
+                i ->
+                    (Callable<Object>)
+                        () -> {
+                          final int sleepTime = ThreadLocalRandom.current().nextInt(limit);
+                          myRateLimiter.tryAcquire(0, i, context);
+                          Thread.sleep(sleepTime);
```
entangled90, line 138: :question:  any reason why the sleep time is the same as the concurrency limit?

entangled90 (approved):

zeebe/broker/src/test/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiterTest.java
filipecampos: Just used it as the max for the random time to sleep.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiterTest.java
filipecampos: Around 0.6s in my machine. Tested with different values and those numbers were the minimum that reproduced the concurrency issue reaching the specified limit of 100.

Issue #24328, opened 2024-11-04
[EPIC] Camunda Exporter potential improvements


Zelldon:
## Description

> [!Important]
>
> This is an issue to collect all possible enhancements that haven't yet implemented but are considered as further improvements (fex to improve performance, etc.).

Szik:
part of docs describing [restoring backpus](https://docs.camunda.io/docs/self-managed/operational-guides/backup-restore/zeebe-backup-and-restore/#restore) leaves me with some questions/suggestions for improvement
- description should name the applications name directly, right now, the reader must guess that the name is equal to the caption
- description of process for restoring is too generic, e.g.,missing explanation of what **zeebe-distribution-X.Y.Z.tar.gz** exactly is or contains would be good as it is mentioned nowhere in section
- when running slef-managed via helm charts, the application **restore** is already part of the zeebe application and one could refer to that or provide a link showing more information about the tar.gz and where to download
- a cross reference to [cold backups](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/operations/backups/#cold-backups) is useful as the documention left me with the open question as to why i can not create such "hot backups" locally/requires cloud storage (i know that this is in progress for upcoming minors)

npepinpe:
@ChrisKujawa - here's a new follow up, which I would put in the should haves, but I'll let you have last word on that: https://github.com/camunda/camunda/issues/28303

ChrisKujawa:
I see all these metric stuff as part of my work doing benchmarks

RomanJRW:
All follow-up issues are now merged, so closing this issue.

Note that there are still further identified pieces of work that need implementing, but they will be done so as independent issues outside the scope of the Exporter 8.8 delivery

PR #26411, opened 2024-12-24
fix: fixing tasklist startup for StandaloneCamunda and StandaloneTask…

marcosgvieira:
## Description

Tasklist wasn't starting properly using StandaloneCamunda with tasklist, broker profiles (what is a valid scenario) - unless when using operate together (as the classes were being loaded from there).
I applied a similar fix for Optimize, making it the "last" option when no other beans are loaded.

This can bring some configuration issues and should be revisited (example: when having different configurations for ES or OS in Optimize, Tasklist, Operate - what I consider not usual but still possible - the backups are going to take the first one)

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOptimizeBackupRepository.java
+++ b/dist/src/main/java/io/camunda/application/commons/backup/ElasticsearchOptimizeBackupRepository.java
@@ -32,7 +32,10 @@ import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 @Conditional(ElasticSearchCondition.class)
 @Configuration
 @Profile("optimize")
-@ConditionalOnMissingBean({ElasticsearchBackupRepository.class, ElasticsearchTasklistBackupRepository.class})
+@ConditionalOnMissingBean({
+  ElasticsearchBackupRepository.class,
+  ElasticsearchTasklistBackupRepository.class
+})
 // only active if standalone, otherwise the operate one is used
 public class ElasticsearchOptimizeBackupRepository {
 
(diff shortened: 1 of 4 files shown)


npepinpe (approved):
:rocket: 

- :wrench: I think generally the conditions could be simplified to just look for a `BackupRepository.class` bean.

And the long term fix should be really having _only_ one ES/OS/RDBMS client, and thus only one `BackupRepository` bean anyway.

Approving anyway as it unblocks us immediately - could you please create a follow up issue for the exporter team to revisit this later? And add it to the `Backups` tasklist here: https://github.com/camunda/camunda/issues/24328

> [!Note] This clearly highlights the issue with the fragmented configuration, and the many profiles and their combinations. Next year I will push for us to go to essentially two profiles: broker (I/O bound, stateful), and gateway/API/web apps (whatever you wanna call it :smile:). Anyway I'll go more into it later, this is out of scope :smile:

marcosgvieira:
> 🚀
> 
> * 🔧 I think generally the conditions could be simplified to just look for a `BackupRepository.class` bean.
> 
> And the long term fix should be really having _only_ one ES/OS/RDBMS client, and thus only one `BackupRepository` bean anyway.
> 
> Approving anyway as it unblocks us immediately - could you please create a follow up issue for the exporter team to revisit this later? And add it to the `Backups` tasklist here: #24328
> 
> > [!Note] This clearly highlights the issue with the fragmented configuration, and the many profiles and their combinations. Next year I will push for us to go to essentially two profiles: broker (I/O bound, stateful), and gateway/API/web apps (whatever you wanna call it 😄). Anyway I'll go more into it later, this is out of scope 😄

I agree, raised https://github.com/camunda/camunda/issues/26415 as a follow-up

PR #26433, opened 2024-12-30
fix: optimize's zeebe compatibility pipeline

buccarel:
## Description

The Zeebe Compatibility pipeline has been failing.
This PR wants to fix it.

A successful run of the pipeline can be seen here:
https://github.com/camunda/camunda/actions/runs/12547169349

The e2e test that is failing is only 1 and it's expected to fail even in `main`.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/backend/src/main/java/io/camunda/optimize/dto/zeebe/process/ZeebeProcessInstanceDataDto.java
+++ b/optimize/backend/src/main/java/io/camunda/optimize/dto/zeebe/process/ZeebeProcessInstanceDataDto.java
@@ -27,6 +27,9 @@ public class ZeebeProcessInstanceDataDto implements ProcessInstanceRecordValue {
   private String elementId;
   private long processInstanceKey;
   private String tenantId;
+  private BpmnEventType bpmnEventType;
+  private List<List<Long>> elementInstancePath;
+  private List<Long> processDefinitionPath;
 
   public ZeebeProcessInstanceDataDto() {}
 
@@ -91,17 +94,17 @@ public class ZeebeProcessInstanceDataDto implements ProcessInstanceRecordValue {
 
   @Override
   public BpmnEventType getBpmnEventType() {
-    throw new UnsupportedOperationException("Operation not supported");
+    return bpmnEventType;
   }
 
   @Override
   public List<List<Long>> getElementInstancePath() {
-    throw new UnsupportedOperationException("Operation not supported");
+    return elementInstancePath;
   }
 
   @Override
   public List<Long> getProcessDefinitionPath() {
-    throw new UnsupportedOperationException("Operation not supported");
+    return processDefinitionPath;
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


optimize/backend/src/main/java/io/camunda/optimize/dto/zeebe/process/ZeebeProcessInstanceDataDto.java
```diff
@@ -28,11 +29,16 @@ public class ZeebeProcessInstanceDataDto implements ProcessInstanceRecordValue {
   private long processInstanceKey;
   private String tenantId;
 
-  public ZeebeProcessInstanceDataDto() {}
+  private BpmnEventType bpmnEventType;
+  private List<List<Long>> elementInstancePath;
+  private List<Long> processDefinitionPath;
+  private List<Integer> callingElementPath;
 
-  @Override
```
buccarel, line 33: This is now required to be present.
I think it's not of any harm

optimize/backend/src/main/java/io/camunda/optimize/dto/zeebe/process/ZeebeProcessInstanceDataDto.java
```diff
@@ -91,22 +97,22 @@ public long getParentElementInstanceKey() {
 
   @Override
   public BpmnEventType getBpmnEventType() {
-    throw new UnsupportedOperationException("Operation not supported");
+    return bpmnEventType;
```
matthewBearCamunda, line 100: ❓ This method was not part of the single jar work like the other 3 changed methods are. Does this one need to be changed? It will always be returning `BpmnEventType.UNSPECIFIED` as the setter is never used

matthewBearCamunda (changes requested):
One question around `getBpmnEventType()`

optimize/backend/src/main/java/io/camunda/optimize/dto/zeebe/process/ZeebeProcessInstanceDataDto.java
buccarel: I got the list of methods that needed to be implemented (i.e., not to throw exception) by looking at the logs. This one was one of the error cause.

No setter is being used, i.e., even the other fields I just added are not being set by anything optimize-related. I am returning the variable instead of directly the value just for consistency. No strong reason.

I got inspired by a similar commit done in tasklist:
https://github.com/camunda/camunda/commit/c0651f26fefc5c8b26ba13b6677ab5d6ea8036ef#diff-eb77b015387bc1565eb155b69e8d7763a64c4637094f34dc4e1be287d3c29c0bR34

All I am doing is not throw exception, but all of these things are not used by Optimize (that's why initially we implemented them as exceptions, as per Josh's suggestion)

matthewBearCamunda: Thanks for the explanation! This PR is good to go then! 🎉

matthewBearCamunda (approved):
LGTM @buccarel

PR #26436, opened 2024-12-30
build: compile as Java 8 module due to transitive inclusion

npepinpe:
## Description

The new `camunda-security-core` module is pulled in the `zeebe-protocol` module, which is fixed to Java version 8 (as it may be included in older modules directly). 

> [!Note]
> I consider it out of scope here to start a debate about `zeebe-protocol` being on Java 8, but we can do that elsewhere.

To make sure anyone using `zeebe-protocol` on Java 8 (e.g. `zeebe-process-test`), we also need to make sure this module is compiled and released with the same version.

### DIFF
--- a/security/security-core/pom.xml
+++ b/security/security-core/pom.xml
@@ -14,9 +14,13 @@
     <version>8.7.0-SNAPSHOT</version>
   </parent>
 
-  <groupId>io.camunda</groupId>
   <artifactId>camunda-security-core</artifactId>
   <name>Camunda Security Core</name>
+
+  <properties>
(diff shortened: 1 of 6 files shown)


npepinpe:
I kind of wish there was a way to flag this, but I'm not aware of one :thinking:

megglos (approved):
❤️ 

What do you mean by flagging this?

npepinpe:
OK, so this breaks a bunch of stuff :smile: 

I'll migrate the records to classes, I thought it would be more straight forward :smile: 

By flagging, I mean the original PR would not have been merged because a check would fail since `zeebe-protocol` depends on a module which could be compiled in a higher version.

PR #26468, opened 2025-01-02
feat: manage 404 error for identity endpoints

### LINKED ISSUE #26323: Migration app should handle management identity when migration flag is disabled

maryarm:
## Description

Handle 404 errors in identity client and throws an NotImplementedException to prevent infinite retries.

## Related issues

closes #26323


migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationHandler.java
```diff
@@ -10,20 +10,35 @@
...
+
+  public void migrate() {
+    logger.info("Migrating started.");
+    List<T> batch;
+    do {
+      batch = fetchBatch();
+      process(batch);
+    } while (!batch.isEmpty());
+    logger.info("Migrating finished.");
+  }
```
lenaschoenburg, line 39: :thought_balloon: Good idea to unify this logic here :+1: Perhaps we could even extend it further (not in this PR) to handle retriable errors uniformly (e.g. conflicts). Because then we have a single place where we can also control retries & and backoff.

migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationHandler.java
```diff
@@ -10,20 +10,35 @@
...
 import io.camunda.zeebe.protocol.record.RejectionType;
+import java.util.List;
 import java.util.concurrent.CompletionException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-public interface MigrationHandler {
+public abstract class MigrationHandler<T> {
+  static final int SIZE = 100;
+  protected final Logger logger = LoggerFactory.getLogger(getClass());
```
lenaschoenburg, line 20: :wrench: I'd rather have dedicated loggers for each handler, that makes it easier to figure out where logs come from.

lenaschoenburg (changes requested):

dist/src/main/java/io/camunda/application/commons/migration/MigrationsRunner.java
```diff
@@ -33,14 +35,23 @@ public MigrationsRunner(final List<Migrator> migrators) {
...
+          result.get();
+        } catch (final ExecutionException e) {
+          LOG.error("Migrator failed", e.getCause());
+          failures.add(e.getCause());
+        }
       }
     }
+
+    if (!failures.isEmpty()) {
+      throw new Exception("#" + failures.stream() + " migrations failed");
```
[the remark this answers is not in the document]
maryarm: Good catch, I meant .size but turned to .stream. I changed it as you suggest.

migration/identity-migration/src/main/java/io/camunda/migration/identity/MigrationHandler.java
maryarm: because the logger is created with getClass, each of MigrationHandler has an specific logger.

dist/src/main/java/io/camunda/application/commons/migration/MigrationsRunner.java
```diff
@@ -33,14 +34,23 @@ public MigrationsRunner(final List<Migrator> migrators) {
   @Override
   public void run(final ApplicationArguments args) throws Exception {
     LOG.info("Starting {} migration tasks", migrators.size());
+    final Exception migrationsExceptions = new Exception();
```
lenaschoenburg, line 37: :upside_down_face: Provide a message here explaining that the migration failed

lenaschoenburg (approved):

PR #26479, opened 2025-01-02
refactor: change OpenAPI string & long key object names

### LINKED ISSUE #26465: [C8 API] Refactor `String` & `Long` keys objects
## Description

Let's try to minimize the breaking changes and confusions for users who already use our V2 APIs by keeping the `Long` keys object names and find new names for `String` keys objects.

## Acceptance criteria

- Revert `Long` key objects names.
- Find a new name for `String` key objects.
- Adjust gateway, client etc.

```[tasklist]
### Tasks
```

danielkelemen:
## Description

- Revert Long key objects
- Rename String key objects to *Result
- Adjust mapper classes
- Update Java Client to String key objects

## Related issues

closes https://github.com/camunda/camunda/issues/26465

### DIFF
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/serializer/LongSerializer.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/serializer/LongSerializer.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway.rest.serializer;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.camunda.zeebe.gateway.rest.RequestMapper;
+import java.io.IOException;
+import org.springframework.http.HttpHeaders;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+public class LongSerializer extends JsonSerializer<Long> {
+
+  public static final String PACKAGE_GATEWAY_REST = "io.camunda.zeebe.gateway.protocol.rest";
+  public static final String KEY = "Key";
+  public static final String KEYS = "Keys";
+
+  private void writeDefault(final Long value, final JsonGenerator gen) throws IOException {
+    writeString(value, gen);
+  }
+
+  private void writeString(final Long value, final JsonGenerator gen) throws IOException {
+    gen.writeString(String.valueOf(value));
+  }
+
+  private void writeNumber(final Long value, final JsonGenerator gen) throws IOException {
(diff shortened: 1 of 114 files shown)


danielkelemen:
TODOs

- [ ] TBD, if we want have reverted Java Client (https://github.com/camunda/camunda/pull/26479/commits/f6234bf2191370b743eb01568e55a92c6fb65f7f) or keep code changes and switch to String key objects.
- [ ] Cherry pick Long keys tests from other PR

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/serializer/LongSerializer.java
```diff
@@ -0,0 +1,80 @@
...
+      final var acceptHeader = r.getRequest().getHeader(HttpHeaders.ACCEPT);
+
+      if (acceptHeader != null && acceptHeader.equals(RequestMapper.MEDIA_TYPE_KEYS_NUMBER_VALUE)) {
+        writeNumber(value, gen);
+        return;
+      }
+      if (acceptHeader != null && acceptHeader.equals(RequestMapper.MEDIA_TYPE_KEYS_STRING_VALUE)) {
+        writeString(value, gen);
+        return;
+      }
```
tmetzke, line 75: 💭 Let's add that for the Java client (using the [agent header](https://github.com/camunda/camunda/blob/afc170997d6c200aa692f8adc7d318783650ffa4/clients/java/src/main/java/io/camunda/client/impl/http/HttpClientFactory.java#L141)) (if no accept headers are defined) we return Longs by default. This allows the Java clients from 8.5 and 8.6 to still receive Long values for keys. We probably need to do some String `startsWith` detection on the header including the version (8.6 and 8.5). Otherwise, we'll serve numbers also for the 8.7+ clients which we changed in this PR here to expect String values.

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/serializer/LongSerializerTest.java
```diff
@@ -0,0 +1,129 @@
...
+import io.camunda.zeebe.util.Either.Left;
+import jakarta.servlet.http.HttpServletRequest;
+import java.io.IOException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.http.HttpHeaders;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+class LongSerializerTest {
```
tmetzke, line 30: 🔧 When we added detection of the Java client, let's also test this here in the following scenarios at least:
1. The accept header is set (e.g. to String) and there is a Java client 8.6 or 8.5 agent header => string keys
2. No accept header is set and there is a Java client 8.6 or 8.5 agent header => number keys
3. No accept header is set and there is a Java client 8.7 header => string keys

tmetzke (commented):
👍 Looks really good to me, very nice job 🚀 
💭 I think we need to handle the older Java clients in the deserializer (if there is no follow-up issue for this already).

tmetzke (approved):
Approving, since the agent header lookup can go into a follow-up

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/serializer/LongSerializer.java
danielkelemen: Good idea, I'll follow up on this.

pepopowitz (approved):
👍 No concerns from the docs side re: spec changes!

PR #26493, opened 2025-01-03
feat: enable webapp functionality only if webappEnabled property is set

### LINKED ISSUE #26489: Backup controller must no be instantiated when webapps are disabled in properties
The backup functionality should not be enabled when profile matches `operate | tasklist` and the property `camunda.(operate|tasklist).webappEnabled` is set to false.
This allows running importers as the backup code is preventing them to startup

entangled90:
## Description
To allow running importers, the backup functionality should not be enabled if webappEnabled is set to false for operate or tasklist.

This problem was discovered when releasing 8.7.0-rc3

## Related issues

closes #26489

### DIFF
--- a/dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
+++ b/dist/src/main/java/io/camunda/application/commons/backup/BackupConfig.java
@@ -9,6 +9,7 @@ package io.camunda.application.commons.backup;
 
 import static io.camunda.application.commons.backup.ConfigValidation.*;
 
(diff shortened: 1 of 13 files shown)


entangled90:
Not sure if there is a proper way to test the importers in the CI

npepinpe (approved):
Honestly I lack context about this webappEnabled flag, and what it's expected to control. If this unblocks alpha3, then I trust you and we go ahead. But I would like to revisit generally how backups are taken for 8.7 here.

Issue #26296, opened 2024-12-19
POC: Identity Key/IDs alignment


remcowesterhoud:
<h1>Description</h1>

**Document**: https://docs.google.com/document/d/1-XeTGEnRPzsSuFdDJblwoLuSVNWvzT0bRclnLhEJe0A/edit?tab=t.0#heading=h.tdo3oan2anbm

<h2>Basic auth</h2>

**TODO**, may not be relevant for a POC

<h2>OIDC</h2>

<h3>Old situation</h3>

We would get a user token like:
```json
{
    "username": "remco",
    "groups": ["foo", "bar"]
}
```

**API**
To give permissions a user has to create a Mapping using a POST request:
```
POST /mapping-rules
{
    "claimName": "username",
    "claimValue": "remco",
    "name": "Username"
}
```

**State**
This will be stored in Zeebe internally in a ColumnFamily:
```java
private final ColumnFamily<DbCompositeKey<DbString, DbString>, PersistedMapping>  mappingColumnFamily
```

<h3>Proposal</h3>

⚠️ **Naming is subject to change**

We want to make our API as simple as possible for our users. Therefore, we want to allow users to configure some commonly used claim names, meaning they don't have to specify this everytime they create a new Mapping.

Examples of commonly used names are:
- username
- group name
- client id

**Config**
Users can configure the commonly used mapping using the configuration options provided by us, for example:
```yaml
camunda.security.authentication.oidc.username=username
camunda.security.authentication.oidc.groupname=groupname
camunda.security.authentication.oidc.clientid=clientid
```

**API**
To create a mapping users no longer need to specify the claim name. Because of the configuration the value is sufficient:
```
POST /simple-mapping
{
    "claimValue": "remco",
}
```

❓❗ How do we know which of the configuration options this value belongs to?

**State**
We can't reuse the existing ColumnFamilies as there is extra data in there that wouldn't be required for these simple mappings, nor can we reuse the user ColumnFamilies for this same reason. We'd need to introduce a new ColumnFamily. The key of this ColumnFamily would be the claim value. The value is to be determined. It can either be a `SimpleMapping` object or just a key might be sufficient.

```java
private final ColumnFamily<DbCompositeKey<DbString, DbString>, SimpleMapping> simpleMappingColumnFamily
```

❓ What should the value of the ColumnFamily be?
❓ How to deal with duplicate values? E.g. the username could be the same as the group name.

**Auth check**
Request comes in with a user token:

```json
{
    "username": "remco",
[...]

Issue #26502, opened 2025-01-03
[REST Gateway] OIDC implementation of multi-tenancy integration


houssain-barouni:
In https://github.com/camunda/camunda/issues/22842, the multi-tenancy integration was implemented for basic authentication.
We need to do the same for OIDC authentication:
- fetch authenticated tenants using mapping rules
- update `identity-auth` and `sso-auth` implementations in Operate and Tasklist to cover V1 endpoints

npepinpe:
Is this for 8.8, or older? Because this is tracked in other issues for 8.8.

PR #26506, opened 2025-01-03
[POC] Simple Mapping Rules

nicpuppa:
## Related issues

related to #26296


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationCheckBehaviorTest.java
```diff
@@ -751,6 +751,34 @@ public void shouldGetDefaultAuthorizedTenantForMapping() {
         .isEqualTo(TenantOwned.DEFAULT_TENANT_IDENTIFIER);
   }
 
+  @Test
+  public void shouldUseSimpleMapping() {
```
nicpuppa, line 755: ❌ This test doesn't work right now, because I probably need to adjust other part. 

It gives an overview on how simple mapping should works

remcowesterhoud:
❓ Does a simple mapping need to be assigned to roles/groups/tenants within Zeebe, or is this information we get from the user's token as well? If we do need to do this within Zeebe we must create some entity in order to keep track of assignments, or change our datastructure a bit.

💭 This means we can no longer check if a user actually exists in the case of OIDC. Since we get a token with the username on it that seems like a safe assumption. The external IDP already verified the user exists.

❓ How do we remove the permissions from a user again, eg if the user gets deleted. Is this a manual action that users need to do through the Identity UI? If so it might be good UX to provide a "clear" endpoint for this that deletes all the permissions. It's probably a common action when someone leaves the company.

remcowesterhoud:
@romansmirnov The POC looks promising, but we have a few open questions that are unclear to us. Could you answer these? Please look at the comment above this one 🙂

Ben-Sheppard:
> 8\. The identity frontend needs to be changed to work with the new API. (⚠️ @nicpuppa and I are not frontend devs and considered this out of scope for us)

This is fine, we have @vitorwtavares as a FE resource and are aware that there needs to be changes to the Identity client so this is not a concern for me.

> OIDC authentication needs to be implemented in the gateways. This is not done yet and will be considered out of scope for the epic we will create as a result of this POC. 

I think I'm missing something here but I'm a little confused by this part, what exactly is missing for OIDC in the gateways?

> @megglos @Ben-Sheppard Is support for "simple mappings" in scope for you?

Looking at the scope doc it is in scope for 8.7, advanced mappings are considered a stretch goal now.

remcowesterhoud:
>I think I'm missing something here but I'm a little confused by this part, what exactly is missing for OIDC in the gateways?

@Ben-Sheppard Sorry, I could've wirtten this down better. It's about #25992 and #26502

Ben-Sheppard:
Ah thanks for clarification there @remcowesterhoud - all good then they are known. I would see those likely actually a part of the stream that @emulbreh and I are looking at so that is also fine from my perspective

PR #26508, opened 2025-01-03
fix: Add missing database env var

saig0:
## Description

Add the environment variable `CAMUNDA_DATABASE_URL` to fix the startup of the Camunda container.


berkaycanbc (approved):
🚀

PR #26511, opened 2025-01-03
fix: match historic/rollover indices correctly

### LINKED ISSUE #25810: Index name: task. Not supported index changes are introduced. Data migration is required.
**Describe the bug**

https://console.cloud.google.com/errors/detail/CI6Ujunh37jOAQ;service=zeebe;time=P7D;locations=global?inv=1&invt=AbjvZQ&project=camunda-saas-int-chaos

**To Reproduce**

**Expected behavior**

**Log/Stacktrace**

**Environment:**
- OS: 
- Zeebe Version: 8.7.0-SNAPSHOT
- Configuration:

npepinpe:
## Description

This PR matches the rollover indices correctly, ensuring that lifecycle policies are _only_ applied to the rollover indices and nothing else.

This fix improves the test coverage by testing when we have no global prefix, and when there is one defined, and adding cases such as when there is another index with a different prefix, or when there is no historic suffix.

> [!Note]
> When there is no global prefix, we cannot properly test the case where the DB would be shared, meaning you have one app with a prefix, and one without, and we should apply the policy only to the one without. This is because we match names with a regex, and since index names may contain multiple separators `-`, we can't differentiate between `prefix-name-version_` and `name-with-separator-version_` (i.e. no prefix).

## Related issues

closes #25810

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchArchiverRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchArchiverRepository.java
@@ -51,7 +51,7 @@ public final class ElasticsearchArchiverRepository implements ArchiverRepository
   private static final String INSTANCES_AGG = "instancesAgg";
   private static final String DATES_SORTED_AGG = "datesSortedAgg";
   private static final String ALL_INDICES = "*";
-  private static final String INDEX_WILDCARD = ".+\\d+\\.\\d+\\.\\d+_.+$";
+  private static final String INDEX_WILDCARD = ".+-\\d+\\.\\d+\\.\\d+_.+$";
 
   private static final Time REINDEX_SCROLL_TIMEOUT = Time.of(t -> t.time("30s"));
   private static final Slices AUTO_SLICES =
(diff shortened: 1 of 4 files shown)


npepinpe:
Added both @megglos and @rodrigo-lourenco-lopes as reviewers - I assume @megglos is more focused on the release so will review this sooner, in which case I'd still like @rodrigo-lourenco-lopes to review this to learn.

npepinpe:
FYI, Operate E2E tests also fail in other PRs and on main with similar failures

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchArchiverRepository.java
```diff
@@ -50,7 +51,7 @@ public final class ElasticsearchArchiverRepository implements ArchiverRepository
   private static final String INSTANCES_AGG = "instancesAgg";
   private static final String DATES_SORTED_AGG = "datesSortedAgg";
   private static final String ALL_INDICES = "*";
-  private static final String INDEX_WILDCARD = "-.*-\\d+\\.\\d+\\.\\d+_.+$";
+  private static final String INDEX_WILDCARD = ".+-\\d+\\.\\d+\\.\\d+_.+$";
```
megglos, line 54: ❓ `.+` is matching the component name, right?

megglos (approved):
LGTM! Thanks for fixing this!

❓ One follow-up question as I didn't dig deeper, do we also support disabling ILM after it was enabled once? From what I currently see the `SchemaManager` creates the policy and we apply it via the archiver. If disabled it would just stay effective though?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/archiver/ElasticsearchArchiverRepository.java
npepinpe: Exactly, we always have a component name at least.

npepinpe:
> ❓ One follow-up question as I didn't dig deeper, do we also support disabling ILM after it was enabled once? From what I currently see the `SchemaManager` creates the policy and we apply it via the archiver. If disabled it would just stay effective though?

Yes, this is a limitation with pre 8.7.0 as well, but we're just maintaining the status quo for now (i.e. feature parity). There are other potentially breaking configuration changes (e.g. changing the rollover suffix for example) which are already problems pre 8.7 as well.

PR #26517, opened 2025-01-03
feat: update operation entity only if operation with the specified operation reference exist

### LINKED ISSUE #26093: New OperationReference on Client commands are not validated and can break system
## Description

We have recently added the support of OperationReferences on the CamundaClient commands. This was especially done to support Operate's BatchOperations/Operations. 

The problem is that there is no validation on the client side. An client/user can pass in anything, this might not match with any OperationReference.

Recently we have added support to the CamundaExporter to consume such, and update Operations in ES/OS. This can cause issues, as if the operations don't exist the update will fail.

The exporter will loop on this record, as it can't be consumed correctly.

## Expectation

 1. We do validation, if we can especially if we expect certain format.
 2. We do not fail on Exporting when something like this Operation reference is not valid. In Operate, previously in the importer we used an updateWithScript, might be a solution.

mustafadagher:
## Description

Currently if a wrong operation reference is specified the update operation will fail and the exporter will retry it

~~With this change, we are using an update script the first checks if the document exists then update.
Otherwise, it will set the batch operation as `noop`, meaning no operation will be performed~~

### Current Implementation
- Provided an API for A `BatchRequest` to keep track of different error handling behavior per index.
- Allowed `ExporterHandler`s to set their own error handling behavior
- Thrown Errors by default for all handlers
- Swallowed errors for Operation Index.

### Improvments:
- [ ] Is it worth it to make the `key` in the BatchRequest error handlers map a combination of `operation type + index name` or `index + document Id` (i.e `update-custom-prefox-operation-8.4.3` instead of `custom-prefox-operation-8.4.3`)?

## Related issues

closes #26093

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ElasticsearchBatchRequest.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ElasticsearchBatchRequest.java
@@ -11,14 +11,17 @@ import co.elastic.clients.elasticsearch.ElasticsearchClient;
 import co.elastic.clients.elasticsearch._types.ElasticsearchException;
 import co.elastic.clients.elasticsearch._types.Refresh;
 import co.elastic.clients.elasticsearch.core.BulkRequest;
+import co.elastic.clients.elasticsearch.core.BulkRequest.Builder;
 import co.elastic.clients.elasticsearch.core.BulkResponse;
 import co.elastic.clients.elasticsearch.core.bulk.BulkResponseItem;
 import io.camunda.exporter.exceptions.PersistenceException;
 import io.camunda.exporter.utils.ElasticsearchScriptBuilder;
 import io.camunda.webapps.schema.entities.ExporterEntity;
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


EuroLew (approved):
LGTM

mustafadagher:
I wrote some IT to verify that this really works as expected, and while it does for Elasticsearch, it fails for Opensearch.

I'm looking into how to fix it for opensearch too, and will commit both the fix and the test (hopefully later today)

mustafadagher:
> @mustafadagher what is the status here? Can we get it running for OS as well?

@ChrisKujawa no, actually I discovered that we log only one error if the bulk response had multiple ones, so it was not also working for ES, it just logged a different error related to date format.

I'll push the changes I have locally, and Let's align if this is actually needed, otherwise I can give the time to something more important.

mustafadagher:
After a lot of trials and errors since last week, I think the only viable solution if we don't want an error to be thrown if no document with the provided id existed, is to use upsert and provide a default document (operation) to insert in case of a missing document.

I wouldn't want to go with this solution if you ask my opinion, and I have a feeling that failing with `missing document` in case of a wrong/non-existent operation reference is an acceptable case.

Let me know how you feel..

mustafadagher:
I changed the design of the solution completely. Please re-read the PR description.

mustafadagher:
On a second thought right now, having the error handling related with the index name alone is error prone.
We have multiple handlers for the same index in some cases. And one could override the other.

Adding the entity id ro the key is not enough as well 
I will think of something else

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ElasticsearchBatchRequest.java
```diff
@@ -230,6 +236,11 @@ public void executeWithRefresh() throws PersistenceException {
     execute(true);
   }
 
+  @Override
+  public void onError(final String index, final Consumer<String> errorHandler) {
+    errorHandlers.put(index, errorHandler);
+  }
+
   private void execute(final boolean shouldRefresh) throws PersistenceException {
```
ChrisKujawa, line 244: If you ask me. I would accept here an error handler, and call it in the execute method if an error appears.

Instead of the indirection, where the RecordHandler adds themself as error handlers, then calling all of them here. I would do some more logic in the batch writer.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ExporterBatchWriter.java
```diff
@@ -73,6 +73,7 @@ public void flush(final BatchRequest batchRequest) throws PersistenceException {
       final ExporterEntity entity = entityAndHandler.entity();
       for (final var handler : entityAndHandler.handlers()) {
         handler.flush(entity, batchRequest);
+        handler.onError(batchRequest);
       }
     }
     batchRequest.execute();
```
ChrisKujawa, line 79: We pass in here a lambda or function - then based on the parameters like id, index, etc. we can find the handler and call it respectively.

PR #26536, opened 2025-01-04
fix: allow missing indices when archiving for now

npepinpe:
## Description

This PR allows bulk updates of index settings (specifically setting the ILM policy for them) even if some of the indices are missing. 

This could be optimized by removing indices if we know nothing was reindex instead, and failing, but we can look into that for a proper fix. Right now the goal is to unblock alpha3 :)

> [!Note]
> No OS changes required because OS does not allow bulk updates anyway, and it also doesn't care if the index doesn't exist, as you create a ISM plugin entry instead of updating the index settings.


megglos (approved):
🚀

megglos:
@npepinpe for the failing module tests see https://github.com/camunda/camunda/pull/26538

PR #26565, opened 2025-01-06
fix: $exist filter on query builder for dates

### LINKED ISSUE #26566: Bug: $exist filter is not working for Dates
To reproduce: 

- On Camunda Client filter any date to check if exists

Expected behaviour: 
- Return the documents that respective date exists

Current Behaviour:
- Exception on Date Time Parse

**Reason from the bug:** 
On 
search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java 
Is parsing (formatting) date independent of the type of input.

But when exists is used, it is sending Boolean instead of Offtimeset

nathansandi:
## Description

- Previously the date was formatting the Date independent of Filter Option Output,
- If I send exist(true) it has Parse Date Error, once it is a Boolean

Action: 
I just moved the format inside the conditions in order to format only when it is necessary 
- Implement integration tests

## Related issues

closes #26566


tmetzke (approved):
👍 Nice catch, @nathansandi 🚀 
⭐ Great that we have a test case for this as well now 👍

PR #26597, opened 2025-01-07
Feature/26041 add new cluster change executor api

### LINKED ISSUE #26041: Add a new `ClusterChangeExecutor` API to delete history data
## Description

In order to delete the history as a cluster operation, we need to provide it with an executor, similar to the `ClusterMembershipChangeExecutor` interface.

We will introduce a new `ClusterChangeExecutor` (as we're not changing the membership, but more broadly the cluster), with a single method: `ActorFuture<Void> deleteHistory();`.

This method will start an asynchronous deletion of the history storage, returning a future indicating success/failure.

When the future is completed successfully, all history data will have been purged. When a future is completed exceptionally, then not all (but none or some) may have been purged.

This operation should be idempotent, as it may be retried multiple times until successful.

### Implementation

We'll add the implementation of this executor in the `zeebe-broker` module, under the same package as `DynamicClusterConfigurationService`.

It should take in a `ConcurrencyControl` (on which it'll be executed, but which can be replaced during testing for determinism) and the `ExporterRepository`, and orchestrate the purge of the history by instantiating every exporter, configuring them, calling `Exporter#purge`, and closing them.

No need to be optimal here, we can handle the asynchronicity like this:

```java
public ActorFuture<Void> purgeExporters() {
  final ActorFuture<Void> result = concurrencyControl.createFuture();
  concurrencyControl.run(() -> {
    try {
      exporterRepository.getExporters().forEach(this:::purgeExporter);
      result.complete(null);
    } catch (final Exception e) {
      result.completeExceptionally(e);
    }
  });

  return result;
}

private void purgeExporter(final String id, final ExporterDescriptor descriptor) {
  // instantiate, configure, purge, close, etc.
}
```

### Registration

As this is our first non-partition specific executor, it will have to be registered in a different place as the other existing executors. To do that, we'll rename `ClusterConfigurationService#registerChangeExecutors` to `ClusterConfigurationService#registerPartitionChangeExecutors`. 

Since the new executor is a single instance per broker, we can add it as a final field to the `DynamicClusterConfigurationService`, and pass it via the constructor. 

> [!Note]
> It's only in that class though to be handed over again to the `ClusterConfigurationManagerService`, which will further pass it down to the `ConfigurationChangeAppliersImpl`, who will then pass it down later to the `DeleteHistoryApplier` (to be done in a follow up).

It will be created in `ClusterConfigurationManagerStep`. Both the `ConcurrencyControl` and the `ExporterRepository` can be obtained from the `BrokerStartupContext`.

> [!Note]
> You could argue we should introduce a new actor here instead of the Broker's, since if this operation takes a long time, we will be blocking shutdown. We can discuss that, or iterate on it in a second pass.

## Acceptance criteria

- [ ] A new `ClusterChangeExecutor` is introduced with a single asynchronous `deleteHistory` method.
  - [ ] The interface is documented via Javadoc.
- [ ] An implementation of the executor is added in the broker module next to the `DynamicClusterConfigurationService`.
  - [ ] The implementation is instantiating, configuring, purging, and closing all exporters when called.
  - [ ] The implementation is asynchronous using the actor scheduler (via `ConcurrencyControl`).
- [ ] The implementation is assigned to a final field in the `DynamicClusterConfigurationService`.
- [ ] The implementation is unit tested (e.g. via test exporter implementations).

macoun:
## Description

Added new interface `ClusterChangeExecutor` and its implementation.
Note that Executor::open() will not be called during deleteHistory.

## Related issues

closes #26041


srsp (approved):

zeebe/broker/src/main/java/io/camunda/zeebe/broker/partitioning/topology/ClusterChangeExecutorImpl.java
```diff
@@ -0,0 +1,75 @@
...
+            1,
+            meterRegistry,
+            StreamClock.system());
+
+    try {
+      exporter.configure(exporterContext);
+      exporter.purge();
+      exporter.close();
+    } catch (final Exception e) {
+      throw new RuntimeException(e);
```
npepinpe, line 72: :wrench: As we return nothing, we can be sneaky and use `LangUtil.rethrowUnchecked(e)` here instead of wrapping the exception. This allows us to return the original exception unchecked, avoiding polluting the stack trace unnecessarily. 

Additionally, if we're not adding any context here, then we bother with the rethrowing? We can just modify the method signature to allow throwing, since our caller anyway has a try/catch around it. It does mean dropping the `forEach()` with lambda and going for a plain `for/loop`, but that's fine.

If we do want to add some context, **which I would recommend for diagnosis**, then I would prefer seeing an exception type. If the goal is rapid diagnosis, having specific inspection types goes a long way to pinpoint where the error comes from, something like:

```java
public final class ClusterChangeExecutorImpl {
  // more ...

  private void purgeExporter(final String id, final ExporterDescriptor descriptor) {
    // more...

    try {
      // stuff
    } catch (final Exception e) {
      throw new ExporterPurgeException(
        "Failed to purge C8 data from exporter {} of type {}; operation will be retried.".formatted(
          id, 
          exporter.getClass()));
    }
  }
}

public final class ExporterPurgeException extends RuntimeException {
  public ExporterPurgeException(final String message, final Throwable cause) {
    super(message, cause);
  }
}
```

OK this was long :smile: To sum up, my strong suggestion :wrench: would be to add diagnosis information here, which would justify the re-throwing behavior :)

zeebe/broker/src/main/java/io/camunda/zeebe/broker/partitioning/topology/DynamicClusterConfigurationService.java
```diff
@@ -27,14 +28,19 @@ public class DynamicClusterConfigurationService implements ClusterConfigurationS
...
+  }
 
   @Override
   public PartitionDistribution getPartitionDistribution() {
     return partitionDistribution;
   }
 
   @Override
-  public void registerChangeExecutors(
+  public void registerPartitionChangeExecutors(
```
npepinpe, line 43: :+1:

zeebe/broker/src/test/java/io/camunda/zeebe/broker/partitioning/topology/ClusterChangeExecutorImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+      repository.validateAndAddExporterDescriptor("test-2", AuditExporter.class, null);
+    } catch (final ExporterLoadException e) {
+      Assertions.fail(e);
+    }
+
+    // when
+    final var executor =
+        new ClusterChangeExecutorImpl(
+            new TestConcurrencyControl(), repository, new SimpleMeterRegistry());
+    executor.deleteHistory().join();
```
npepinpe, line 39: :wrench: You can omit the join here, and just let AssertJ run its assertions on the future, e.g.

```java
    final var executor =
        new ClusterChangeExecutorImpl(
            new TestConcurrencyControl(), repository, new SimpleMeterRegistry());
    final Future<Void> result = executor.deleteHistory();

    // then
    Assertions.assertThat(result).succeedsWithin(Duration.ofSeconds(5));
    Assertions.assertThat(AuditExporter.AUDITS)
        .containsSubsequence("configure-test-1", "purge-test-1", "close-test-1");
    Assertions.assertThat(AuditExporter.AUDITS)
        .containsSubsequence("configure-test-2", "purge-test-2", "close-test-2");
```

zeebe/broker/src/test/java/io/camunda/zeebe/broker/partitioning/topology/ClusterChangeExecutorImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+    final ExporterRepository repository = new ExporterRepository();
+    try {
+      repository.validateAndAddExporterDescriptor("test-1", AuditExporter.class, null);
+      repository.validateAndAddExporterDescriptor("test-2", AuditExporter.class, null);
+    } catch (final ExporterLoadException e) {
+      Assertions.fail(e);
+    }
+
+    // when
+    final var executor =
```
npepinpe, line 36: :upside_down_face: Nit pick, but this is part of `given` :smile:

zeebe/broker/src/test/java/io/camunda/zeebe/broker/partitioning/topology/ClusterChangeExecutorImplTest.java
```diff
@@ -0,0 +1,78 @@
...
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import java.util.ArrayList;
+import java.util.List;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class ClusterChangeExecutorImplTest {
+
+  @Test
+  public void shouldRunPurgeForEveryExporter() {
```
npepinpe, line 25: :upside_down_face: Can be package private

zeebe/broker/src/test/java/io/camunda/zeebe/broker/partitioning/topology/ClusterChangeExecutorImplTest.java
npepinpe, line 1: :wrench: We should add a test checking that we _don't_ call `open` or `export`, as we will explicitly document that we _do not call open_.

npepinpe (commented):
:tada: 

I noted some suggestions and nit picks, nothing major or that would have blocked it. For the nit picks (the :upside_down_face:), you can take them or leave them, as you want :) For the suggestions (:wrench:), please either implement them, add them as follow ups in the main epic (remember we have a task list for those), or challenge me if you disagree, of course :)

PR #26603, opened 2025-01-07
feat: rework batch operation update to be run asynchronously

### LINKED ISSUE #24084: Rework how `BatchOperationEntity#operationsFinishedCount` is updated

sdorokhova:
## Description

* Reworked batch operation update task to run steps asynchronously
* small fix: when reading data from indices we're failing now, when shards/indices are unavailable. We're keeping `allowNoIndices(true)` as we don't want to get failures on first schema creation.

## Related issues

related with #24084


mustafadagher (approved):
LGTM! 🚀

PR #26606, opened 2025-01-07
New auth type for authenticated user and user claims

### LINKED ISSUE #23093: Refactor use of JWT for authenticated user and user token claims to a new auth type
### Description
With the merge of https://github.com/camunda/camunda/pull/22459 and https://github.com/camunda/camunda/pull/22853 we load some auth data onto the existing crafted JWT that is used later on in the command processing. The use of this JWT can be quite expensive and as we have the data in an already processed format we should leverage this.

### Acceptance criteria
- [ ]  A new `AuthDataFormat` exists for user token data/authenticated user key 
- [ ] `toDecodedMap` is updated to process and return the new `AuthDataFormat`
- [ ] A test exists in `AuthInfoTest` to check the correct data is returned

nicpuppa:
## Description

Remove usage of internal JWT token between gateway and broker in favour of a Map data type

## Related issues

closes #23093


zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/encoding/AuthInfo.java
```diff
@@ -73,25 +97,19 @@ public DirectBuffer toDirectBuffer() {
...
-        final String jwtToken = getAuthData();
-        return Authorization.jwtDecoder(jwtToken)
-            .withClaim(Authorization.AUTHORIZED_TENANTS)
-            .decode();
-      }
-      default -> {
-        return Map.of(
-            Authorization.AUTHORIZED_TENANTS, List.of(TenantOwned.DEFAULT_TENANT_IDENTIFIER));
-      }
+    if (claimsProp.hasValue()) {
```
nicpuppa, line 100: ❓ This method is still needed as we need to ensure that older records are still correctly processed, am I right ?

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/RecordBatchTest.java
```diff
@@ -159,7 +159,7 @@ void shouldOnlyAddUntilMaxBatchSizeIsReached() {
     assertThat(either.isLeft()).isTrue();
     assertThat(either.getLeft())
         .hasMessageContaining("Can't append entry")
-        .hasMessageContaining("[ currentBatchEntryCount: 1, currentBatchSize: 398]");
+        .hasMessageContaining("[ currentBatchEntryCount: 1, currentBatchSize: 408]");
```
nicpuppa, line 162: This value is increased because of the new property

zeebe/stream-platform/src/test/java/io/camunda/zeebe/stream/impl/RecordBatchTest.java
```diff
@@ -189,7 +189,7 @@ void shouldReturnTrueIfRecordSizeDoesntReachSizelimit() {
   @Test
   void shouldOnlyReturnTrueUntilMaxBatchSizeIsReached() {
     // given
-    final var recordBatch = new RecordBatch((count, size) -> size < 399);
+    final var recordBatch = new RecordBatch((count, size) -> size < 409);
```
nicpuppa, line 192: The new size is 408

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/encoding/AuthInfo.java
```diff
@@ -73,25 +97,19 @@ public DirectBuffer toDirectBuffer() {
...
     }
+    return Map.of();
   }
 
   @Override
   public String toString() {
-    String data = getAuthData();
-    data = data.isEmpty() ? "" : "." + data;
-    return formatProp.getValue().toString() + data;
+    return "AuthInfo{" + "authData=" + BufferUtil.bufferAsString(claimsProp.getValue()) + '}';
```
lenaschoenburg, line 112: :x: I don't think a `DocumentValue` is something we can/should interpret as a `String`. I'd suggest to just use the decoded map.

:wrench: The property name for the claims should be `claims`, not `authData`.

:wrench: I'd suggest to keep the old properties (format and authData) here, juuuust in case this goes wrong at some point.

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/encoding/AuthInfo.java
lenaschoenburg: Yes, we need the backwards compatibility :+1: 

:x: Using `hasValue` this way should not work as far as I can tell. `hasValue` always returns true when the property has a default value, which `DocumentProperty` always does.
I'd suggest that we instead use the `format` property to switch between decoding the JWT or just using the claims document.

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/encoding/AuthInfo.java
```diff
@@ -53,15 +56,36 @@ public AuthInfo setAuthData(final String authData) {
...
+
+  @JsonIgnore
+  public DirectBuffer getClaimsBuffer() {
+    return claimsProp.getValue();
+  }
+
   public void wrap(final AuthInfo authInfo) {
     formatProp.setValue(authInfo.getFormat());
     authDataProp.setValue(authInfo.getAuthData());
+    claimsProp.setValue(authInfo.getClaimsBuffer());
```
lenaschoenburg, line 81: :thought_balloon: Please double-check whether this is truly a `wrap`, i.e. we want to wrap over the same mutable byte buffers, or actually meant as a copy. `getAuthData` always copies, so it's possible that we need to do the same for the claims document.

lenaschoenburg (changes requested):

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/encoding/AuthInfo.java
nicpuppa: As far as I can see, the `wrap` method is used as a copy method.

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/encoding/AuthInfo.java
```diff
@@ -73,25 +90,25 @@ public DirectBuffer toDirectBuffer() {
...
-        final String jwtToken = getAuthData();
-        return Authorization.jwtDecoder(jwtToken)
-            .withClaim(Authorization.AUTHORIZED_TENANTS)
-            .decode();
-      }
-      default -> {
-        return Map.of(
-            Authorization.AUTHORIZED_TENANTS, List.of(TenantOwned.DEFAULT_TENANT_IDENTIFIER));
-      }
+    if (Objects.requireNonNull(getFormat()) == AuthDataFormat.JWT) {
```
lenaschoenburg, line 93: :upside_down_face: I don't think the `requireNonNull` is doing much here, a normal NPE will show you the same (or actually better?) information.

lenaschoenburg (approved):
Really nice cleanup, I'm very happy that we don't have to decode JWT in the engine anymore :tada:

PR #26609, opened 2025-01-07
fix: specify process instance index name on ProcessInstanceArchiverJobOpenSearch

### LINKED ISSUE #26576: Tasklist archiver doesn't handle process instances when using Opensearch
### Describe the bug

The Tasklist archiver is not functional when using Opensearch. This applies to both standalone (i.e. non-AWS) Opensearch and AWS Opensearch, but for different root causes (see below).

### To Reproduce

1. Configure Tasklist against Opensearch
    1. Variant a: Regular Opensearch Docker container
    2. Variant b: AWS Opensearch
2. Create process instances and complete them, so that they can be archived

### Actual behavior

* Process instances (and related data) are not archived (i.e. moved from runtime to dated indexes)

### Expected behavior

* Process instances are archived successfully

### Environment

- Tested with standalone Opensearch 2.17.1 (and the corresponding AWS version)

#### Standalone Opensearch

* The search query doesn't use a source: https://github.com/camunda/camunda/blob/ec157d1bbe13e086ec04dab3bec5da49653443a8/tasklist/archiver/src/main/java/io/camunda/tasklist/archiver/os/ProcessInstanceArchiverJobOpenSearch.java#L165
* Accordingly, the query always returns 0 results and nothing is archived
* On a side note, when scrolling the response, in this situation always the same scroll id `FGluY2x1ZGVfY29udGV4dF91dWlkDnF1ZXJ5VGhlbkZldGNoAA==` is returned. See this ES ticket that explains this: https://github.com/elastic/elasticsearch/issues/85709. Note that we are deleting the scrolling behavior with https://github.com/camunda/camunda/issues/26571

#### AWS Opensearch

* In contrast to standalone OS, AWS Opensearch seems to apply the request to all indexes in the cluster (also ignoring the index prefix of course, so we are breaking our schema isolation)
* So the query returns any documents that match, for example flow node instances
* For any such documents, we then attempt to delete variables, flow nodes, and process instances (via https://github.com/camunda/camunda/blob/stable/8.6/tasklist/archiver/src/main/java/io/camunda/tasklist/archiver/os/ProcessInstanceArchiverJobOpenSearch.java#L84-L100), which doesn't delete any data
  * There is one edge case where it does delete data: If you have multiple installations with different index prefix on the same OS instance, then a flow node instance key of one schema may be a process instance key in another schema; in that case, a process instance may be deleted incorrectly
* Accordingly, the archiver is busy spinning, always finding data but not deleting anything
* Note: we did not get to the bottom of why the behavior is different for standalone and AWS Opensearch

### Acceptance Criteria

* Fix the behavior for both standalone and AWS Opensearch
* Make sure we have test coverage for archiver behavior against Opensearch

panagiotisgts:
## Description

Specify the `process-instance` fully qualified index name for rocessInstanceArchiver Job on Opensearch.
## Checklist

- [ ] for CI changes:

## Related issues

closes #26576

### DIFF
--- a/tasklist/qa/integration-tests/src/test/java/io/camunda/tasklist/it/ArchiverIT.java
+++ b/tasklist/qa/integration-tests/src/test/java/io/camunda/tasklist/it/ArchiverIT.java
@@ -40,6 +40,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.UUID;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.BeanFactory;
@@ -47,7 +48,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 
 public class ArchiverIT extends TasklistZeebeIntegrationTest {
-
   @Autowired private BeanFactory beanFactory;
   @Autowired private ArchiverUtil archiverUtil;
   @Autowired private TaskTemplate taskTemplate;
@@ -236,8 +236,7 @@ public class ArchiverIT extends TasklistZeebeIntegrationTest {
 
   @Test
   public void shouldArchiveDocumentsOnlyFromDeclaredProcessInstanceIndex() throws IOException {
-    final String piCustomPrefix = "pi-custom-";
-    databaseTestExtension.createIndex(piCustomPrefix + processInstanceIndex.getIndexName());
+    final String piCustomPrefix = UUID.randomUUID().toString().substring(0, 10);
     final Instant currentTime = pinZeebeTime();
 
     // having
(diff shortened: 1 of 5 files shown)


ThorbenLindhauer:
@panagiotisgts I don't have time for a code review I'm afraid

ThorbenLindhauer:
In general, I'd suggest writing a test though

panagiotisgts:
The functionality of the Archiver itself is tested in [ArchiverIT](https://github.com/camunda/camunda/blob/stable/8.6/tasklist/qa/integration-tests/src/test/java/io/camunda/tasklist/it/ArchiverIT.java)

panagiotisgts:
As mentioned in my [comment](https://github.com/camunda/camunda/issues/26576#issuecomment-2575220068) on the bug report issue, Tasklsit does not archive Process related data as Operate does, we simply delete them since they have no value for it's domain after a PI is completed.

Unless I understood something wrong from your initial bug report, the only bug that could occur from this missing Index name specification is what you correctly described here,

> There is one edge case where it does delete data: If you have multiple installations with different index prefix on the same OS instance, then a flow node instance key of one schema may be a process instance key in another schema; in that case, a process instance may be deleted incorrectly

or any other potential match for this case.

This is no longer viable since the index name is now specified on the query. The existing ArchiverIT correctly covers the deletion of Process related data, the fact that it was already passing is due to the fact that we create and drop the indices for every test case, hence the behavior you described above could not be reproduced on the existing test. Writing test cases for correct index access would require it to be done for every ES/OS query we have, but this is supposed to be covered by the SDK itself.

ThorbenLindhauer:
Thanks for elaborating. I overall agree on what is covered and how it's working. But why not write a test that reproduces the edge case and thus makes sure that the code in place will not be removed again?

tasklist/qa/integration-tests/src/test/java/io/camunda/tasklist/it/ArchiverIT.java
```diff
@@ -238,6 +234,55 @@ public void shouldDeleteProcessInstanceRelatedData() throws ArchiverException, I
...
+    final List<String> allIds = new ArrayList<>();
+    allIds.addAll(ids1);
+    allIds.addAll(ids2);
+
+    final List<String> customIndexIds =
+        noSqlHelper.getIdsFromIndex(
+            ProcessInstanceIndex.ID, piCustomPrefix + processInstanceIndex.getIndexName(), allIds);
+    assertThat(customIndexIds).containsExactlyInAnyOrderElementsOf(allIds);
+    assertThat(noSqlHelper.getProcessInstances(allIds)).isEmpty();
+    databaseTestExtension.deleteIndex(piCustomPrefix + processInstanceIndex.getIndexName());
```
houssain-barouni, line 283: maybe we need to put this in a finally block to make sure that the index is deleted even after failure
the failsafe is configured to retry 2 times for flaky tests

tasklist/qa/integration-tests/src/test/java/io/camunda/tasklist/it/ArchiverIT.java
```diff
@@ -238,6 +234,55 @@ public void shouldDeleteProcessInstanceRelatedData() throws ArchiverException, I
     assertProcessInstancesExist(ids3);
   }
 
+  @Test
+  public void shouldArchiveDocumentsOnlyFromDeclaredProcessInstanceIndex() throws IOException {
+    final String piCustomPrefix = "pi-custom-";
```
houssain-barouni, line 239: I would recommend using a random value, the failsafe is configured to retry 2 times the flaky tests

houssain-barouni (commented):
We just need to make sure that the new test is re-triable

houssain-barouni (approved):
👍

PR #26610, opened 2025-01-07
refactor: don't track authorizations by resource id

lenaschoenburg:
This data was not needed and the entire column family was write-only in production code.

We can simply remove it and mark the column family as deprected until we have a need for it again.


remcowesterhoud (approved):
Thanks for cleaning this up @lenaschoenburg 🙇

PR #26613, opened 2025-01-07
feat: Opensearch implementation for batch operation update

### LINKED ISSUE #24084: Rework how `BatchOperationEntity#operationsFinishedCount` is updated

sdorokhova:
## Description

* implement batch operation update task for Opensearch
* create `OpensearchUtil` class for reusable code
* refactor [ElasticsearchBatchOperationUpdateRepositoryIT.java](https://github.com/camunda/camunda/compare/sd-batch-operations-finished-count-3...sd-batch-operations-finished-count-4?expand=1#diff-a012648004449d5dd1e3165862bba31be979636acaad87ea41ac9ae1004ac615) to extract Opensearch and Elasticsearch static inner classes, similar to commit 718960d7f1333797fce5fe10b08a3fc973079473

## Related issues

closes #24084

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/OpensearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/OpensearchRepository.java
@@ -21,36 +21,38 @@ import org.opensearch.client.opensearch.OpenSearchAsyncClient;
 import org.opensearch.client.opensearch._types.ErrorCause;
 import org.opensearch.client.opensearch._types.Time;
 import org.opensearch.client.opensearch.core.ClearScrollRequest;
-import org.opensearch.client.opensearch.core.SearchRequest;
+import org.opensearch.client.opensearch.core.SearchRequest.Builder;
 import org.opensearch.client.opensearch.core.bulk.BulkResponseItem;
 import org.opensearch.client.opensearch.core.search.Hit;
 import org.slf4j.Logger;
 
 public class OpensearchRepository {
 
-  private final Time SCROLL_KEEP_ALIVE = Time.of(t -> t.time("1m"));
-  private final int SCROLL_PAGE_SIZE = 100;
+  private static final Time SCROLL_KEEP_ALIVE = Time.of(t -> t.time("1m"));
+  private static final int SCROLL_PAGE_SIZE = 100;
+  protected final OpenSearchAsyncClient client;
+  protected final Executor executor;
+  protected final Logger logger;
+
+  public OpensearchRepository(
+      final OpenSearchAsyncClient client, final Executor executor, final Logger logger) {
+    this.client = client;
+    this.executor = executor;
+    this.logger = logger;
+  }
 
   /**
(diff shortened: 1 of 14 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/OpensearchUtil.java
```diff
@@ -0,0 +1,185 @@
...
+import org.opensearch.client.opensearch.OpenSearchAsyncClient;
+import org.opensearch.client.opensearch._types.ErrorCause;
+import org.opensearch.client.opensearch._types.Time;
+import org.opensearch.client.opensearch.core.ClearScrollRequest;
+import org.opensearch.client.opensearch.core.SearchRequest;
+import org.opensearch.client.opensearch.core.bulk.BulkResponseItem;
+import org.opensearch.client.opensearch.core.search.Hit;
+import org.slf4j.Logger;
+
+public class OpensearchUtil {
```
mustafadagher, line 29: 🔧 I have a minor bitter feeling with the naming `Util` here (and also for the ES one). 😄 
Looking at the code, it seems like a search `repository` to me, that doesn't need to contain static methods.

but you can ignore this comment for now haha

mustafadagher (approved):
LGTM!

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/OpensearchUtil.java
sdorokhova: Good point! I converted them to abstract repository classes and made other repositories to extend from them.

entangled90 (approved):

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/OpensearchRepository.java
```diff
@@ -0,0 +1,184 @@
...
+      final Executor executor,
+      final Logger logger,
+      final SearchRequest.Builder requestBuilder,
+      final Function<Hit<Object>, T> transformer) {
+    return fetchUnboundedDocumentCollection(
+        client, executor, logger, requestBuilder, Object.class, transformer);
+  }
+
+  public <TDocument, TResult> CompletionStage<Collection<TResult>> fetchUnboundedDocumentCollection(
+      final OpenSearchAsyncClient client,
```
entangled90, line 50: :wrench: given the class is called "*Repository" I would pass the dependencies (OpenSearchAsyncClient, Executor, Logger)  in the constructor

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/OpensearchUtil.java
mustafadagher: Thanks @sdorokhova ! Looks better! 

🔧 is it worth it to move the `client, executor, logger` to be protected fields of the abstract repositories, instead of passing them around in methods?

PR #26617, opened 2025-01-07
feat: added test to verify if truncate is executed on all tables #26129

stoerti:
## Description

Verifies, that the list of truncated tables is complete and exhaustive.

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/PurgerCompletenessIT.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/PurgerCompletenessIT.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.it.rdbms.db;
+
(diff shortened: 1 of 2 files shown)


qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/PurgerCompletenessIT.java
```diff
@@ -0,0 +1,72 @@
...
+        .containsExactlyInAnyOrderElementsOf(tablesToBeTruncated);
+    assertThat(tablesToBeTruncated)
+        .describedAs("should only truncate Camunda tables")
+        .containsExactlyInAnyOrderElementsOf(actuallyTruncated);
+  }
+
+  private List<String> getAllCamundaTableNames() {
+    return jdbcTemplate
+        .queryForList(
+            "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'PUBLIC' AND TABLE_NAME NOT IN ('DATABASECHANGELOG', 'DATABASECHANGELOGLOCK')")
```
p-wunderlich, line 67: This just works as long as we really want to purge ALL Camunda tables. But what if we e.g. have additional "technical" tables which we don't want to purge?

stoerti: We then could either include them in the exclusion list of this statement OR introduce a general prefix for all SearchClient tables.

I decided to postpone this for now as a future decision when we actually need it.

p-wunderlich (approved):

Issue #26421, opened 2024-12-28
Replace @AutoCloseResources extension with new built-in Junit AutoClose extension


npepinpe:
**Description**

Junit 5 recently introduced a new `@AutoClose` extension which supersedes our custom `@AutoCloseResources`, and works mostly the same.

You need to:

- Remove all references to `@AutoCloseResources`
- Replace all `@AutoCloseResource` with Junit's `@AutoClose`

Then remove the actual `@AutoCloseResources` annotation and the extension `AutoCloseResourceExtension`.

rinujohn:
Hello, Can I make this change. 
Regards,
Rinu George

eppdot:
Hey @npepinpe I started looking into this as an easy starter task, but realized that the `AutoClose` [Annotation](https://junit.org/junit5/docs/5.11.1/api/org.junit.jupiter.api/org/junit/jupiter/api/AutoClose.html) is only available in junit/jupiter 5.11.

But (at least in my local version) 5.10 is used in zeebe broker (where I started to play around). Do you see any risk updating and could you guide me with the best strategy to do it (😅 did not work with maven/java for some time). 
![Screenshot 2025-01-08 at 08 47 08](https://github.com/user-attachments/assets/e9d7ae17-c21b-418a-9d2c-27baf52c04ef)

npepinpe:
I don't foresee any issues using junit 5.11

eppdot:
👍 alright. Then I will just go ahead. 

Thats what was also mentioned in the changelog, but I can't really judge if its relevant

> Deprecations and Breaking Changes
> As mentioned in the Bug Fixes section above, field and method search algorithms now adhere to standard Java semantics regarding whether a given field or method is visible or overridden according to the rules of the Java language. The changes in the search algorithms may, however, result in breaking changes for some use cases. In light of that, it is possible to revert to the previous "legacy semantics". See the new [Field and Method Search Semantics](https://junit.org/junit5/docs/5.11.0/user-guide/index.html#extensions-supported-utilities-search-semantics) section of the User Guide for details.

npepinpe:
If CI is green (at least the Java parts), then all good 👍

eppdot:
Hmm, seems something is not right with my setup as the `parent/pom.xml` already states the newest version

![Screenshot 2025-01-08 at 09 01 50](https://github.com/user-attachments/assets/1ca6ab5e-e1fa-4d33-acf5-67b25afc6e7d)

[...]

PR #26641, opened 2025-01-08
Resolve junit 5 maven dependency convergence

korthout:
## Description

Maven dependencies take precedence by ordering. As the junit-bom was defined below spring-boot-starter-test, spring-boot-starter-test defined what junit version is used in our own maven project.

This caused the effective pom to use 5.10.5 rather than 5.11.4.

This commit moves the declaration of the junit-bom to ensure our declaration takes precedence.

## Related issues

NA

[Discussion in slack](https://camunda.slack.com/archives/C037RS2JHB8/p1736327541102319)

Unblocks #26421

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -935,6 +935,14 @@
 
       <!-- Third party dependencies -->
       <dependency>
+        <groupId>org.junit</groupId>
+        <artifactId>junit-bom</artifactId>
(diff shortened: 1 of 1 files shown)


lenaschoenburg (approved):
Thanks!

korthout:
Oh wow. I forgot to do that. Thanks @entangled90 🚀

PR #26647, opened 2025-01-08
fix: improve performance of range queries

### LINKED ISSUE #24195: [C8 API] Operation search query builders create one range query
## Description

Operations search query builders return range query for each range operation (gt, lt etc..). This is not optimal, it should add all range operations inside one range query instead.
Ref: https://github.com/camunda/camunda/blob/a1ed2210f98726ce8a1e8efac9b39d68e0471d86/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/query/SearchQueryBuilders.java#L354-L357

## Acceptance criteria

- `SearchQueryBuilders` class *Operations methods create one range query for range operations.
  - They will still create multiple queries (term, range) in total.

```[tasklist]
### Tasks

```

nathansandi:
## Description

Created one range query for range operations for all applicable data types.

## Related issues

closes https://github.com/camunda/camunda/issues/24195


danielkelemen (approved):
👍 Looks good!

PR #26657, opened 2025-01-08
pg-26655-replace-xmlutil

### LINKED ISSUE #26655: Replace or improve XMLUtil and add test coverage

panagiotisgts:
## Description

Extract a common utility ProcessModelReader and use it to replace XMLUtil 

Affected modules:
- Zeebe Utils
- Process Migration
- Camunda Exporter
- Gateway Rest

cc: @danielkelemen Adding you as reviewer since you were the author of the `XmlUtil` version for `ProcessCache` on gateway-rest and part of the `c8-api-team` , would like your feedback regarding the changes there.

ps: There is still one usage of a _XmlUtil_ on the[ RDBMS Exporter](https://github.com/camunda/camunda/blob/f829eb96bf488076459a951a3fd9ebc94b8c0303/zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/handlers/ProcessExportHandler.java#L52), based on Operate's XmlUtil, I'm not sure this is fully implemented though. I'm looking into it.
 
## Checklist

- [ ] for CI changes:

## Related issues

closes #26655


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/process/ElasticSearchProcessCacheLoader.java
```diff
@@ -57,8 +55,11 @@ public CachedProcessEntity load(final Long processDefinitionKey) throws IOExcept
 
   private List<String> extractCallActivityIdsFromDiagram(final ProcessEntity processEntity) {
     final String bpmnXml = processEntity.getBpmnXml();
-    final Optional<ProcessEntity> diagramData =
-        xmlUtil.extractDiagramData(bpmnXml.getBytes(), processEntity.getBpmnProcessId());
-    return diagramData.isPresent() ? diagramData.get().getCallActivityIds() : new ArrayList<>();
+    return ProcessModelReader.of(
```
sdorokhova, line 58: ❓ couple of questions here. 
1. Do you think we could move sorting to `ProcessModelReader`? I think we use call activities only from exporter now and they are always sorted, right? In case you want to keep `ProcessModelReader` more generic, may be we can have small utility in `camunda-exporter` module to convert collection of call activity flow nodes to sorted list of names?
Otherwise I think we need to extend `ProcessHandlerTest` to check that newly cached data is properly sorted (basically testing [this line](https://github.com/camunda/camunda/pull/26657/files#diff-e5bd0c492bc220bd63a5fd58db0b4f16d18b254a4da9efe46ab30ecfed5c5757R116)).
2. It looks like Elastic and Opensearch implementations of `extractCallActivityIdsFromDiagram` method are exactly the same (they we the same before your change also). Should we may be clean it up and move upper in the class hierarchy?

sdorokhova (approved):
Besides my tiny comment on call activities sorting, LGTM!

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/cache/process/ElasticSearchProcessCacheLoader.java
panagiotisgts: @sdorokhova Applied some changes in this [commit](https://github.com/camunda/camunda/pull/26657/commits/962d902542cd5e9b1227707a675101aae7719673). The order of commits got mixed-up by some wrong order of push/pull rebases I did.

danielkelemen (approved):
👍 The XmlUtil changes look good to me!
I like that we have less `XmlUtil`s now (although there are still a few more).

PR #26676, opened 2025-01-08
test: verify that zeebe can authenticate and authorize OIDC requests

### LINKED ISSUE #26015: Integration test for checking mapping permissions
We should have a basic integration test to verify that requests can be authorized through mappings, similar to other tests in `io.camunda.zeebe.it.authorization` which all test with users.

In https://github.com/camunda/camunda/pull/25920 we couldn't yet write any integration tests because we do not have a default mapping yet which we need to bootstrap the permissions.

The only workaround for this would have been to first start with basic auth, use the default user to create initial mappings, and then restart with oauth.

lenaschoenburg:
This copies two existing integration tests for basic authentication and adapts them for OAuth2/OIDC.

In these tests, we provide a Keycloak IDP and set it up with two clients and matching service accounts. We create Camunda clients for each of the Keycloak clients and set up a default mapping for one of them. The other client is not mapped by default.

Closes #26015

### DIFF
--- a/zeebe/qa/integration-tests/pom.xml
+++ b/zeebe/qa/integration-tests/pom.xml
@@ -282,6 +282,18 @@
     </dependency>
 
     <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-admin-client</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
       <scope>test</scope>
--- a/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/authorization/OidcAuthOverGrpcIT.java
+++ b/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/authorization/OidcAuthOverGrpcIT.java
@@ -1,0 +1,246 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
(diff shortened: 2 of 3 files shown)


remcowesterhoud (approved):
🚀🚀🚀

lenaschoenburg:
The changes from https://github.com/camunda/camunda/pull/26606 break this now:

 ```
java.lang.RuntimeException: Failed to serialize object '{user_token_scope=email profile, user_token_azp=zeebe, user_token_clientAddress=10.88.0.7, user_token_typ=Bearer, authorized_tenants=[<default>], user_token_acr=1, user_token_iss=http://localhost:40269/realms/camunda, user_token_clientHost=10.88.0.7, user_token_email_verified=false, user_token_client_id=zeebe, user_token_exp=2025-01-09T11:23:52Z, user_token_iat=2025-01-09T11:18:52Z, user_token_sub=bf78ee6a-b3df-45cd-be1b-63abd000fd7f, user_token_preferred_username=zeebe-service-account, user_token_jti=bd871804-21ec-4c96-be5e-e19254b14990}' to MessagePack
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type `java.time.Instant` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: java.util.HashMap["user_token_exp"])
```

It seems like Spring converts some claims to `Instant` which we then can't serialize as messagepack. I'm trying to find out what types we have to expect in claims because I don't want to add support for `Instant` only to find out later in production that some claims have a different unsupported type too.

lenaschoenburg:
Looks like this is coming from `MappedJwtClaimSetConverter` and its default behavior to map `exp`, `iat` and `nbf` claims to `Instant`. Essentially we have to keep the messagepack conversion in sync with the defaults of `MappedJwtClaimSetConverter`. For now, that's just adding support for `Instant` but I'm not sure how to best prevent any surprises in the future, in case `MappedJwtClaimSetConverter` gets new default mappings.

lenaschoenburg:
Argh, I just wanted to change target branch because this depends on https://github.com/camunda/camunda/pull/26719 but auto-merge was still enabled :facepalm:

PR #26698, opened 2025-01-09
feat: remove authorized tenants from REST broker requests

### LINKED ISSUE #26694: [REST Gateway] authorized tenants should not be send to the broker

nicpuppa:
## Description

Removed the authorized tenants id from the REST request sent to the broker.

## Related issues

closes #26694


remcowesterhoud (approved):
This change looks good, but isn't there a bunch of tests that need to be changed as well?

nicpuppa:
@remcowesterhoud there is a separate issue to adjust the tests https://github.com/camunda/camunda/issues/25718. I tried to do both of the issue in the same PR but was too much work to do for the tests.

PR #26706, opened 2025-01-09
feat: add LocalStorage document store implementation

### LINKED ISSUE #24846: [Doc Handling] Local file system document store implementation
**Description**

As part of the Document Handling capability in Camunda, we need to introduce the local file system document store implementation.

**Key requirements**

- Documents must be stored as files on the local disk of the system where Camunda is running
- This document store is not production-ready and is mainly intended for local development, therefore some limitations are allowed
- The local document store implementation must be enabled by default on self-managed installations
- The root directory path must be configurable via environment variables

**Supported operations:** create, get, delete documents, document hash verification

**Unsupported operations:** create document link

**Further limitations to be considered:**
- Should we support resource-based authorization? Metadata can be stored in a separate file, allowing us to access process definition key & process instance key. It can also be fine to skip this in the first iteration.

mo-okocha:
## Checklist

- [ ] for CI changes:

## Related issues

closes #24846

### DIFF
--- a/document/store/src/main/java/io/camunda/document/store/DocumentHashProcessor.java
+++ b/document/store/src/main/java/io/camunda/document/store/DocumentHashProcessor.java
@@ -12,10 +12,15 @@ import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.HexFormat;
+import org.apache.commons.codec.digest.MessageDigestAlgorithms;
 
 public class DocumentHashProcessor {
 
-  public static HashResult hash(final InputStream stream, final String algorithm) {
+  public static HashResult hash(final InputStream stream) {
+    return hash(stream, MessageDigestAlgorithms.SHA_256);
+  }
+
+  private static HashResult hash(final InputStream stream, final String algorithm) {
     final MessageDigest md;
     try {
       md = MessageDigest.getInstance(algorithm);
--- a/document/store/src/main/java/io/camunda/document/store/DocumentHashProcessor.java
+++ b/document/store/src/main/java/io/camunda/document/store/DocumentHashProcessor.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.store;
+
+import java.io.InputStream;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
+
+public class DocumentHashProcessor {
+
+  public static HashResult hash(final InputStream stream, final String algorithm) {
+    final MessageDigest md;
+    try {
+      md = MessageDigest.getInstance(algorithm);
+    } catch (final NoSuchAlgorithmException e) {
+      throw new RuntimeException(e);
+    }
+    final DigestInputStream digestStream = new DigestInputStream(stream, md);
+    final String contentHash = HexFormat.of().formatHex(md.digest());
+
+    return new HashResult(digestStream, contentHash);
+  }
+
+  public record HashResult(DigestInputStream inputStream, String contentHash) {}
+}
--- a/document/store/src/main/java/io/camunda/document/store/localstorage/FileHandler.java
+++ b/document/store/src/main/java/io/camunda/document/store/localstorage/FileHandler.java
@@ -12,7 +12,7 @@ import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
-public class FileHandler {
+class FileHandler {
 
   public InputStream getInputStream(final Path documentPath) throws IOException {
     return Files.newInputStream(documentPath);
--- a/document/store/src/main/java/io/camunda/document/store/localstorage/FileHandler.java
+++ b/document/store/src/main/java/io/camunda/document/store/localstorage/FileHandler.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.store.localstorage;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class FileHandler {
+
+  public InputStream getInputStream(final Path documentPath) throws IOException {
+    return Files.newInputStream(documentPath);
+  }
+
+  public void createFile(final InputStream stream, final Path path) throws IOException {
+    Files.copy(stream, path);
+  }
+
+  public boolean fileExists(final Path path) {
+    return Files.exists(path);
+  }
+
+  public void delete(final Path path) throws IOException {
+    Files.deleteIfExists(path);
+  }
+}
--- a/document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
+++ b/document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
@@ -1,0 +1,168 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.store.localstorage;
+
+import static org.apache.commons.codec.digest.MessageDigestAlgorithms.SHA_256;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.camunda.document.api.DocumentContent;
+import io.camunda.document.api.DocumentCreationRequest;
+import io.camunda.document.api.DocumentError;
+import io.camunda.document.api.DocumentError.DocumentAlreadyExists;
+import io.camunda.document.api.DocumentError.DocumentHashMismatch;
+import io.camunda.document.api.DocumentError.DocumentNotFound;
+import io.camunda.document.api.DocumentError.OperationNotSupported;
+import io.camunda.document.api.DocumentError.UnknownDocumentError;
+import io.camunda.document.api.DocumentLink;
+import io.camunda.document.api.DocumentMetadataModel;
+import io.camunda.document.api.DocumentReference;
+import io.camunda.document.api.DocumentStore;
+import io.camunda.document.store.DocumentHashProcessor;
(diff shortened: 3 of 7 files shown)


document/store/src/main/java/io/camunda/document/store/localstorage/FileHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.document.store.localstorage;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class FileHandler {
```
georgios-goulos, line 15: If `FileHandler` is only used within the package `io.camunda.document.store.localstorage` it might be worth making it package private

document/store/src/main/java/io/camunda/document/store/DocumentHashProcessor.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.HexFormat;
+
+public class DocumentHashProcessor {
+
+  public static HashResult hash(final InputStream stream, final String algorithm) {
+    final MessageDigest md;
+    try {
+      md = MessageDigest.getInstance(algorithm);
+    } catch (final NoSuchAlgorithmException e) {
+      throw new RuntimeException(e);
```
georgios-goulos, line 23: How will this `RuntimeException` be handled?

I understand that right now we're only using `SHA-256` as an algorithm so this should never really be thrown, but if we give this method the flexibility of receiving any hashing alg as a `String` it might be thrown in the future.

Since we're only using this method with `SHA-256` perhaps it's worth to change the signature to 
```
public static HashResult hash(final InputStream stream)
```
And hardcode `SHA-256` here? If there's a need to use another algorithm in the future we can change this to
```
  public static HashResult hash(final InputStream stream) {
    return hash(stream, SHA_256);
  }

public static HashResult hash(final InputStream stream, final String algorithm) {
...
}
```

Happy to discuss this of course

document/store/src/main/java/io/camunda/document/store/DocumentHashProcessor.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package io.camunda.document.store;
+
+import java.io.InputStream;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
+
+public class DocumentHashProcessor {
```
georgios-goulos, line 16: Are there unit tests for this class?

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  private Either<DocumentError, DocumentReference> createDocumentInternal(
+      final DocumentCreationRequest request) {
+    final String documentId = getDocumentId(request);
+    final Path documentFilePath = storagePath.resolve(documentId);
+    final Path documentMetaDataFilePath = storagePath.resolve(documentId + METADATA_SUFFIX);
+
+    if (fileHandler.fileExists(documentFilePath)
+        || fileHandler.fileExists(documentMetaDataFilePath)) {
```
georgios-goulos, line 94: This worries me a bit: what if only the metadata file exists but the original file does not exist (e.g. it was deleted before but something went wrong with deleting the metadata file)? This condition would suggest the file already exists.

It is probably an edge case, but I would think that if either of the two files is missing we consider the file as not existing anymore. Any chance this has been discussed before?

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
```diff
@@ -0,0 +1,168 @@
...
+    } catch (final IOException e) {
+      return Either.left(new UnknownDocumentError(e));
+    }
+    return Either.right(
+        new DocumentReference(request.documentId(), hashResult.contentHash(), request.metadata()));
+  }
+
+  private Either<DocumentError, DocumentContent> getDocumentInternal(final String documentId) {
+    final Path documentPath = storagePath.resolve(documentId);
+    final Path documentMetadataPath = storagePath.resolve(documentId + METADATA_SUFFIX);
```
georgios-goulos, line 114: nitpick: Since we're doing `documentId + METADATA_SUFFIX` in at least 2 places in this file it might be preferable to have a private method generate this.

```
private String getMetadataFilename(final String documentId) {
  return documentId + METADATA_SUFFIX;
}
```

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
```diff
@@ -0,0 +1,168 @@
...
+      if (!result.contentHash().equals(contentHash)) {
+        return Either.left(new DocumentHashMismatch(documentId, contentHash));
+      }
+    } catch (final IOException e) {
+      return Either.left(new UnknownDocumentError(e));
+    }
+    return Either.right(null);
+  }
+
+  private static String getDocumentId(final DocumentCreationRequest request) {
```
georgios-goulos, line 165: nitpick: Does this need to be `static`?

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStoreProvider.java
```diff
@@ -0,0 +1,53 @@
...
+public class LocalStorageDocumentStoreProvider implements DocumentStoreProvider {
+
+  private static final String STORAGE_PATH = "PATH";
+
+  @Override
+  public DocumentStore createDocumentStore(
+      final DocumentStoreConfigurationRecord configuration, final ExecutorService executor) {
+    final Path storagePath = getStoragePath(configuration);
+
+    LOG.info("Storage path created at {}", STORAGE_PATH);
```
georgios-goulos, line 28: I believe this would only output the value of `STORAGE_PATH` which is `PATH`.

Is this supposed to be logging `storagePath`?

Similar for the other log in this file and perhaps the generated message of `IllegalArgumentException`

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStoreProvider.java
```diff
@@ -0,0 +1,53 @@
...
+    return new LocalStorageDocumentStore(
+        storagePath, new FileHandler(), new ObjectMapper(), executor);
+  }
+
+  private Path getStoragePath(final DocumentStoreConfigurationRecord configuration) {
+
+    final String pathString = configuration.properties().get(STORAGE_PATH);
+    if (pathString == null) {
+      LOG.warn("Local Storage Document store {} property is not set", STORAGE_PATH);
+      return null;
```
georgios-goulos, line 39: Do we want to allow the application to proceed when `storagePath == null`?

I assume "yes" because document handling can be optional, but I thought I'd ask the question.

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
```diff
@@ -0,0 +1,168 @@
...
+  @Override
+  public CompletableFuture<Either<DocumentError, DocumentReference>> createDocument(
+      final DocumentCreationRequest request) {
+    return CompletableFuture.supplyAsync(() -> createDocumentInternal(request), executor);
+  }
+
+  @Override
+  public CompletableFuture<Either<DocumentError, DocumentContent>> getDocument(
+      final String documentId) {
+    return CompletableFuture.supplyAsync(() -> getDocumentInternal(documentId));
```
georgios-goulos, line 64: This doesn't use the injected `executor`. Is this intentional?

Same for `supplyAsync` in some of the other methods in this class

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+
+  private Either<DocumentError, Void> deleteDocumentInternal(final String documentId) {
+    final Path documentPath = storagePath.resolve(documentId);
+    final Path documentMetadataPath = storagePath.resolve(documentId + METADATA_SUFFIX);
+
+    try {
+      fileHandler.delete(documentPath);
+      fileHandler.delete(documentMetadataPath);
```
georgios-goulos, line 139: I see [GcpDocumentStore](https://github.com/camunda/camunda/blob/main/document/store/src/main/java/io/camunda/document/store/gcp/GcpDocumentStore.java#L185) returns a
```
return Either.left(new DocumentError.DocumentNotFound(documentId));
```
if the document does not exist, although [AwsDocumentStore](https://github.com/camunda/camunda/blob/main/document/store/src/main/java/io/camunda/document/store/aws/AwsDocumentStore.java) doesn't do that. This is a bit inconsistent. Any idea what's the correct behaviour? 

Might be a question for Product though

document/store/src/test/java/io/camunda/document/store/localstorage/LocalStorageDocumentStoreTest.java
```diff
@@ -0,0 +1,227 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class LocalStorageDocumentStoreTest {
+
+  private Path STORAGE_PATH;
+  private LocalStorageDocumentStore documentStore;
+  private final FileHandler fileHandler = new FileHandler();
+  private final ObjectMapper mapper = new ObjectMapper();
```
georgios-goulos, line 35: I don't think this `mapper ` is used

georgios-goulos (changes requested):

document/store/src/main/java/io/camunda/document/store/localstorage/FileHandler.java
mo-okocha: Updated

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
mo-okocha: I see your point on this. I added some changes to the logic to ensure no inconsistent documents are removed before the check. Have a look and let me know if the changes are good

document/store/src/main/java/io/camunda/document/store/DocumentHashProcessor.java
mo-okocha: No unit tests, although this is covered from the `LocalStorageDocumentStoreTest`

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
mo-okocha: Done!

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStoreProvider.java
mo-okocha: Yes, my mistake. Fixed

document/store/src/main/java/io/camunda/document/store/localstorage/LocalStorageDocumentStore.java
georgios-goulos: I cannot see the change but it's a small thing anyway. I will resolve this

georgios-goulos (approved):

Issue #24243, opened 2024-10-31
Camunda Exporter fails with:  Failed retrieving mappings from index/index templates with pattern [*]


ChrisKujawa:
## Description

Reported on INT with our RC for alpha1

https://console.cloud.google.com/errors/detail/CI7_od-oyLDzSA;service=zeebe;time=P7D;locations=global?project=camunda-saas-int

```
io.camunda.exporter.exceptions.ElasticsearchExporterException: Failed retrieving mappings from index/index templates with pattern [*]

at io.camunda.exporter.schema.elasticsearch.ElasticsearchEngineClient.getMappings ( io/camunda.exporter.schema.elasticsearch/ElasticsearchEngineClient.java:131 )
at io.camunda.exporter.schema.SchemaManager.validateIndexTemplates ( io/camunda.exporter.schema/SchemaManager.java:161 )
at io.camunda.exporter.schema.SchemaManager.startup ( io/camunda.exporter.schema/SchemaManager.java:47 )
at io.camunda.exporter.CamundaExporter.open ( io/camunda.exporter/CamundaExporter.java:102 )
at io.camunda.zeebe.broker.exporter.stream.ExporterContainer.lambda$openExporter$0 ( io/camunda.zeebe.broker.exporter.stream/ExporterContainer.java:118 )
at io.camunda.zeebe.util.jar.ThreadContextUtil.runCheckedWithClassLoader ( io/camunda.zeebe.util.jar/ThreadContextUtil.java:59 )
at io.camunda.zeebe.util.jar.ThreadContextUtil.runWithClassLoader ( io/camunda.zeebe.util.jar/ThreadContextUtil.java:35 )
at io.camunda.zeebe.broker.exporter.stream.ExporterContainer.openExporter ( io/camunda.zeebe.broker.exporter.stream/ExporterContainer.java:117 )
at io.camunda.zeebe.broker.exporter.stream.ExporterDirector.lambda$startActiveExportingMode$16 ( io/camunda.zeebe.broker.exporter.stream/ExporterDirector.java:532 )
at io.camunda.zeebe.scheduler.retry.BackOffRetryStrategy.run ( io/camunda.zeebe.scheduler.retry/BackOffRetryStrategy.java:51 )
at io.camunda.zeebe.scheduler.ActorJob.invoke ( io/camunda.zeebe.scheduler/ActorJob.java:85 )
[...]

PR #26707, opened 2025-01-09
Prefix migration changes

EuroLew:
## Description

When migrating the prefix the https://github.com/elastic/elasticsearch-java/issues/696 `Property 'type' was not found` bug was encountered this was fixed locally here https://github.com/camunda/camunda/commit/8ea3885067638df255acab3e1c2e93f293c25410 where the `type:custom` was manually added.

For legacy index templates from 8.6 this default type field does not exist so it causes issues with deserialisation, this bug was fixed in a later version of the client which this PR updates it to. This version change is not propagated widely as there are some breaking API usages in the client in other modules.

Also add the ability to pass in a prefix when running the standalone schema manager to make prefix migration simpler on the user 

## Related issues

relates to  https://github.com/camunda/camunda-docs/issues/4498

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -30,6 +30,7 @@
     <dependency>
       <groupId>co.elastic.clients</groupId>
       <artifactId>elasticsearch-java</artifactId>
+      <version>8.14.2</version>
     </dependency>
 
     <dependency>
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -15,6 +15,16 @@
 
   <name>Zeebe Distribution</name>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>co.elastic.clients</groupId>
+        <artifactId>elasticsearch-java</artifactId>
+        <version>8.14.2</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
 
     <dependency>
@@ -30,7 +40,6 @@
     <dependency>
       <groupId>co.elastic.clients</groupId>
       <artifactId>elasticsearch-java</artifactId>
-      <version>8.14.2</version>
     </dependency>
 
     <dependency>
--- a/dist/src/main/java/io/camunda/application/StandaloneSchemaManager.java
+++ b/dist/src/main/java/io/camunda/application/StandaloneSchemaManager.java
@@ -91,6 +91,7 @@ public class StandaloneSchemaManager {
 
     final ExporterConfiguration exporterConfig = new ExporterConfiguration();
     exporterConfig.setConnect(connectConfiguration);
+    exporterConfig.getIndex().setPrefix(connectConfiguration.getIndexPrefix());
 
     final IndexDescriptors indexDescriptors =
         new IndexDescriptors(connectConfiguration.getIndexPrefix(), IS_ELASTICSEARCH);
--- a/zeebe/exporters/camunda-exporter/pom.xml
+++ b/zeebe/exporters/camunda-exporter/pom.xml
@@ -53,6 +53,12 @@
     <dependency>
       <groupId>io.camunda</groupId>
       <artifactId>camunda-search-client-connect</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>co.elastic.clients</groupId>
+          <artifactId>elasticsearch-java</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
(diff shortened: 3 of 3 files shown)


dist/pom.xml
```diff
@@ -15,6 +15,16 @@
 
   <name>Zeebe Distribution</name>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>co.elastic.clients</groupId>
+        <artifactId>elasticsearch-java</artifactId>
+        <version>8.14.2</version>
```
ChrisKujawa, line 23: ❓ From what I can see from our dependencies https://github.com/camunda/camunda/blob/a2cd92fe25a728b72b4143e93022cd77a8146ac3/parent/pom.xml#L56 is that this is a minor update, can we be sure that this works with different server version, etc.?

dist/src/main/java/io/camunda/application/StandaloneSchemaManager.java
```diff
@@ -91,6 +91,7 @@ public static void main(final String[] args) {
 
     final ExporterConfiguration exporterConfig = new ExporterConfiguration();
     exporterConfig.setConnect(connectConfiguration);
+    exporterConfig.getIndex().setPrefix(connectConfiguration.getIndexPrefix());
```
ChrisKujawa, line 94: Can we separate this into a different PR? I feel like this is completely something different to the purpose of this PR?

ChrisKujawa (commented):
> For legacy index templates from 8.6 this default type field does not exist so it causes issues with deserialisation, this bug was fixed in a later version of the client which this PR updates it to. This version change is not propagated widely as there are some breaking API usages in the client in other modules.

What would it mean if we migrate the complete version to 8.14? Could we maybe check whether we can migrate fully to 8.14? We would need to check with Product. @npepinpe do you have any thoughts?

EuroLew:
Also as a note updating the java client to be able to de-serialise the type field should fix this bug https://github.com/camunda/camunda/issues/24243

npepinpe:
We typically hold off on bumping Elastic versions, as they aren't always compatible with all server versions. I'm not an expert here, so I can't easily evaluate this.

@ThorbenLindhauer do we have a documented policy on when we update ES versions? It seems valuable in this instance as it fixes some bugs, but as it comes with breaking change in its client, I'm a little reluctant. Can we maybe confirm it's fully backwards compatible with 8.x servers? :thinking:

EuroLew:
> We typically hold off on bumping Elastic versions, as they aren't always compatible with all server versions. I'm not an expert here, so I can't easily evaluate this.
> 
> @ThorbenLindhauer do we have a documented policy on when we update ES versions? It seems valuable in this instance as it fixes some bugs, but as it comes with breaking change in its client, I'm a little reluctant. Can we maybe confirm it's fully backwards compatible with 8.x servers? 🤔

Ok if it's possibly breaking I could go with the alternate solution which is to manually add the `type: custom` field to the normalizers

ThorbenLindhauer:
> @ThorbenLindhauer do we have a documented policy on when we update ES versions? It seems valuable in this instance as it fixes some bugs, but as it comes with breaking change in its client, I'm a little reluctant. Can we maybe confirm it's fully backwards compatible with 8.x servers? 🤔

We don't have a policy for all know. We could either check ES docs if they make any statement about client compatibility (ideally they would, many RDBMSs have it). Otherwise we could adjust the versions in our CI builds to the lowest supported and run them for the PR.

ChrisKujawa:
@EuroLew can you take a look at it or did you found a different solution?

EuroLew:
> @EuroLew can you take a look at it or did you found a different solution?

For some reason it's working now, maybe it was an issue on my end

panagiotisgts:
@EuroLew I run into this issue on latest `main` when migrating from stable/8.6 to 8.7 for the `operate-list-view` lowercase analyzer.
```
Error deserializing co.elastic.clients.elasticsearch._types.analysis.Normalizer: Property 'type' not found 
(JSON path: index_templates[75].index_template.template.settings.index.analysis.normalizer.case_insensitive) 
(line no=1, column no=162968, offset=-1)
```

EuroLew:
> @EuroLew I run into this issue on latest `main` when migrating from stable/8.6 to 8.7 for the `operate-list-view` lowercase analyzer.
> 
> ```
> Error deserializing co.elastic.clients.elasticsearch._types.analysis.Normalizer: Property 'type' not found 
> (JSON path: index_templates[75].index_template.template.settings.index.analysis.normalizer.case_insensitive) 
> (line no=1, column no=162968, offset=-1)
> ```

sigh, for now you can get unblocked one of two ways
1. update the ES client same as this PR To 8.14.2 locally 
2. manually update the ES indices by adding the `type: custom` field via the API, I'm pretty sure ES doesn't support individual field addition you will need to get the settings, manually update the settings and then copy and paste the new settings in a PUT.

This will be fixed in the java migration app (most likely with solution 2).

PR #26713, opened 2025-01-09
feat: map OIDC tokens to Camunda principals

Ben-Sheppard:
_**Note**: the commits in this PR have been broken out from the larger POC branch for OIDC support [here](https://github.com/camunda/camunda/pull/26378)._

## Description
This PR is designed to accept an authentication object produced by the OIDC mode and map it to the internal Camunda principal used. During this mapping process we use the token provided to find:

* Mappings which are matching by the token claims
* With the mapping keys:
  * The roles assigned
  * The application assigned
  * The tenants assigned
  * The groups assigned


emulbreh (approved):

authentication/src/main/java/io/camunda/authentication/CamundaOidcUserService.java
```diff
@@ -0,0 +1,81 @@
...
+        mappingKeys,
+        new AuthenticationContext(
+            roleServices.getRolesByMemberKeys(mappingKeys),
+            authorizationServices.getAuthorizedApplications(mappingKeys),
+            tenantServices.getTenantsByMemberKeys(mappingKeys).stream()
+                .map(TenantDTO::fromEntity)
+                .toList(),
+            groupServices.getGroupsByMemberKeys(mappingKeys).stream()
+                .map(GroupEntity::name)
+                .toList()));
```
Ben-Sheppard, line 79: 💭 Roman raised to me once before that we effectively perform the calls multiple times, once here and then also again down in the engine. I think for now this is fine but we should look to optimize this and only retrieve them whenever we really need to (either by the auth checker, or the engine). Here is the thread https://camunda.slack.com/archives/C06UYJMMETZ/p1734351732320729?thread_ts=1734343187.803619

authentication/src/main/java/io/camunda/authentication/entity/CamundaOidcUser.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  public String getEmail() {
+    return user.getEmail();
+  }
+
+  @Override
+  public String getDisplayName() {
+    return user.getClaim("nickname");
```
Ben-Sheppard, line 39: ❓ @emulbreh - is this always `nickname`? Should it be configurable?

authentication/src/main/java/io/camunda/authentication/entity/CamundaOidcUser.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  public Collection<? extends GrantedAuthority> getAuthorities() {
+    return user.getAuthorities();
+  }
+
+  @Override
+  public String getName() {
+    return user.getClaim("name");
```
Ben-Sheppard, line 74: ❓ @emulbreh - same here I think, can we always assume this is `name`? Should it also be configurable?

authentication/src/main/java/io/camunda/authentication/filters/TenantRequestAttributeFilter.java
```diff
@@ -56,24 +55,17 @@ protected void doFilterInternal(
...
-    };
+    if (!(principal instanceof final Authentication auth)) {
+      LOG.error(
+          "cannot find tenants: unsupported principal type: {}", principal.getClass().getName());
+      return null;
+    }
+    if (!(auth.getPrincipal() instanceof final CamundaPrincipal camundaPrincipal)) {
+      LOG.error("cannot find tenants: unsupported principal type: {}", auth.getClass().getName());
+      return null;
+    }
```
Ben-Sheppard, line 66: 💭 I wonder if here we can instead extract this to a single helper method and have one if guard?

authentication/src/main/java/io/camunda/authentication/service/OidcCamundaUserService.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.context.annotation.Profile;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Service;
+
+@Profile("auth-oidc")
+@Service
+public class OidcCamundaUserService implements CamundaUserService {
+  private static final String SALES_PLAN_TYPE = "";
+  private static final Map<AppName, String> C8_LINKS = Map.of();
```
Ben-Sheppard, line 26: 🔧 We should add a TODO here this needs to be set for SaaS purposes.

authentication/src/main/java/io/camunda/authentication/service/OidcCamundaUserService.java
```diff
@@ -0,0 +1,61 @@
...
+                  SALES_PLAN_TYPE,
+                  C8_LINKS,
+                  true);
+            })
+        .orElse(null);
+  }
+
+  @Override
+  public String getUserToken() {
+    return "";
```
Ben-Sheppard, line 59: 💭 Does this need to be an empty string or can we thrown an exception instead?

authentication/src/main/java/io/camunda/authentication/entity/CamundaOidcUser.java
emulbreh: I don't know what this should be, but hardcoding the claim name definitely isn't right. There's probably some overlap with the "simple mapping" configuration which also needs configuration for well-known claims.

authentication/src/main/java/io/camunda/authentication/filters/TenantRequestAttributeFilter.java
emulbreh: The error message is different because we're checking the type of different objects. A helper method would probably have to have the same structure in its body.

authentication/src/main/java/io/camunda/authentication/service/OidcCamundaUserService.java
emulbreh: I believe this method is unused. You added it in https://github.com/camunda/camunda/pull/24947/files#diff-cb31ab0a8e3d005cede786bc73cb750cad552a4e4096c9d894592b2eac80b22dR15.
I thought we would eventually need it to implement UserReader.getUserToken(), but that's only used for /v1/internal/users/token and maybe that can be removed too.

authentication/src/main/java/io/camunda/authentication/CamundaOidcUserService.java
emulbreh: Are there actually many requests that don't need most of the authentication context?

Ben-Sheppard: Yea exactly, so the decision would be that any call that results in an engine command (create, update, delete) will check the auth context directly in the engine, searches would need the context.

authentication/src/main/java/io/camunda/authentication/filters/TenantRequestAttributeFilter.java
Ben-Sheppard: True, we don't really gain much but it would just lift it and reduce this overall methods size

authentication/src/main/java/io/camunda/authentication/service/OidcCamundaUserService.java
Ben-Sheppard: Ah yes, this was for backwards compatibility. I think we can leave it as this for now

authentication/src/main/java/io/camunda/authentication/entity/CamundaOidcUser.java
Ben-Sheppard: For now I've reverted it to `user.getPreferredUsername()`

authentication/src/main/java/io/camunda/authentication/CamundaOidcUserService.java
Ben-Sheppard: Raised https://github.com/camunda/camunda/issues/26739

PR #26719, opened 2025-01-09
fix: drop unsupported user claims

### LINKED ISSUE #26712: Filter user token claims before forwarding
## Description
Currently, when we make a Zeebe broker request, we forward all the claims from the users token so that Zeebe can apply mapping rules. For example, in a test with Keycloak as OIDC provider, the token contains the following claims:

- user_token_scope: email profile
- user_token_azp: zeebe
- user_token_clientAddress: 10.88.0.13
- user_token_typ: Bearer
- user_token_acr: 1
- user_token_iss: http://localhost:42309/realms/camunda
- user_token_clientHost: 10.88.0.13
- user_token_email_verified: false
- user_token_client_id: zeebe
- user_token_exp: 2025-01-09T12:08:25Z
- user_token_iat: 2025-01-09T12:03:25Z
- user_token_sub: d1c8468e-cd9e-4485-995f-04dc70d0b7e6
- user_token_preferred_username: zeebe-service-account
- user_token_jti: 59e949d1-c5dc-47f5-876a-e61fc319c345

This causes two problems:
- That's a lot of claims! For Zeebe this means that we have to check all 14 claims for matching mapping rules. Additionally, the serialized command is becoming quite large. Most of the claims don't even make sense for mapping rules (how would you map an expiration date?).
- The claim type depends on Spring behavior that appears difficult to control. For example, we [just realized](https://github.com/camunda/camunda/pull/26676#issuecomment-2579915269) that Spring automatically converts exp , iat and nbf to Instant which we currently can't serialize. Of course we could add support for Instant but who's to say that Spring won't change default behavior and start mapping some claims to types we don't support?

## Proposed Solution

Ideally, we'd be defensive and require some sort of configuration about which claims are forwarded.
Since this is difficult to achieve, for now we can simply filter out [well-known claims](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1) that cause problem and are not useful for mapping rules anyway:
- `exp` (Expiration Time).
- `nbf` (Not Before)
- `iat` (Issued At)

lenaschoenburg:
This fixes #26712 by dropping the `exp`, `iat`, and `nbf` claims as well as any other claims with values that are not `String`, `Boolean` or `Number`.

### DIFF
--- a/zeebe/auth/pom.xml
+++ b/zeebe/auth/pom.xml
@@ -40,6 +40,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-engine</artifactId>
       <scope>test</scope>
--- a/zeebe/auth/src/main/java/io/camunda/zeebe/auth/ClaimTransformer.java
+++ b/zeebe/auth/src/main/java/io/camunda/zeebe/auth/ClaimTransformer.java
@@ -7,6 +7,7 @@
  */
 package io.camunda.zeebe.auth;
 
+import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import org.slf4j.Logger;
@@ -36,7 +37,8 @@ public final class ClaimTransformer {
    * </ul>
    *
    * Additionally, claims with unsupported value types will be dropped and a warning will be logged.
-   * Supported value types are {@link String}, {@link Boolean}, and {@link Number}.
+   * Supported value types are {@link String}, {@link Boolean}, and {@link Number}. We also pass
+   * through {@link Collection} values without checking the type of the elements.
    *
    * @param claims A collection of user claims to add to.
    * @param key The key of the claim, without prefix, as it appears in the token. If the key is not
@@ -53,7 +55,10 @@ public final class ClaimTransformer {
       return;
     }
 
-    if (value instanceof String || value instanceof Boolean || value instanceof Number) {
+    if (value instanceof String
+        || value instanceof Boolean
+        || value instanceof Number
+        || value instanceof Collection<?>) {
       claims.put(Authorization.USER_TOKEN_CLAIM_PREFIX + key, value);
     } else {
       LOG.debug("Dropping claim '{}' with unsupported value type '{}'", key, value.getClass());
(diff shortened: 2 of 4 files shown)


lenaschoenburg:
This breaks `tokenContainsExtraClaimsWithOidcAuth` where we use a claim that has a `List<String>` value. I'm not sure whether we actually want this or not @nicpuppa.

nicpuppa (changes requested):
❌ I think this transformer is not needed. We already transform the claims when we decode the token, we should change the logic there instead of adding this new class. You can have a look at the `JwtDecoder` class

lenaschoenburg:
`JwtDecoder` is used in two places at the moment:
1. In the old gRPC `AuthenticationInterceptor` to actually decode a token.
2. In `AuthInfo` if the old `JWT` format is used.

Notably, it's missing from the REST API. So yes, we could implement this logic in `JwtDecoder` but that would only help us with gRPC unless we start injecting our `JwtDecoder` into Spring.

nicpuppa:
Yes, you are right @lenaschoenburg. I thought about refactor the JwtDecoder a bit so you can re-use the logic in both gRPC and REST scenario. So for gRPC the claims will be removed while decoding the token, and for REST in the request mapper

lenaschoenburg:
But the REST request mapper doesn't need to decode JWTs at all, why make it use the `JwtDecoder` there? Filtering claims and adding a prefix is a different concept than decoding JWTs, hence the separate class.

nicpuppa:
I agree with you, my idea was to have a unique class (like an helper or something) that offer you all the methods related to token/claims you need. But all good, I don't want to make things more complex.

To answer this:

> This breaks `tokenContainsExtraClaimsWithOidcAuth` where we use a claim that has a `List<String>` value. I'm not sure whether we actually want this or not @nicpuppa.

I think we should be more flexible and allow users to provide multiple values per single claim. This came up also with @remcowesterhoud a couple of days ago in a discussion related to the Identity Id/key POC

remcowesterhoud:
> > This breaks `tokenContainsExtraClaimsWithOidcAuth` where we use a claim that has a `List<String>` value. I'm not sure whether we actually want this or not @nicpuppa.
> 
> I think we should be more flexible and allow users to provide multiple values per single claim. This came up also with @remcowesterhoud a couple of days ago in a discussion related to the Identity Id/key POC

Yes, we need this. For example there could be claim containing all the groups/roles a users belongs to. That's a very valid case we need to support.

nicpuppa (approved):
🚀

PR #26720, opened 2025-01-09
test: verify that zeebe can authenticate and authorize OIDC requests

### LINKED ISSUE #26712: Filter user token claims before forwarding

lenaschoenburg:
Re-opened https://github.com/camunda/camunda/pull/26676 because that one got merged accidentally :upside_down_face: 

Based on https://github.com/camunda/camunda/pull/26719 to fix https://github.com/camunda/camunda/issues/26712.

### DIFF
--- a/zeebe/qa/integration-tests/pom.xml
+++ b/zeebe/qa/integration-tests/pom.xml
@@ -277,6 +277,18 @@
     </dependency>
 
     <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.keycloak</groupId>
+      <artifactId>keycloak-admin-client</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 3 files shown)


remcowesterhoud (approved):
Assuming nothing changed 😄

PR #26722, opened 2025-01-09
Replace usage of internal Junit 5 helper removed in Junit 5.11

npepinpe:
## Description

It seems with the problem we had around having 2 JUnit versions at the same time, we missed that one helper we use a lot `ReflectionUtils.makeAccessible` was removed in Junit 5.11, which breaks many of our extensions' lifecycle.

This PR replaces usage with our own helper which emulates the old behavior.

### DIFF
--- a/zeebe/util/src/main/java/io/camunda/zeebe/util/ReflectUtil.java
+++ b/zeebe/util/src/main/java/io/camunda/zeebe/util/ReflectUtil.java
@@ -9,6 +9,8 @@ package io.camunda.zeebe.util;
 
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Member;
+import java.lang.reflect.Modifier;
 import java.util.stream.Stream;
 
 public final class ReflectUtil {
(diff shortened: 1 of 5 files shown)


npepinpe:
@entangled90 I don't have much time on this, so if there's any issues or anything to change, go ahead and take over the PR and apply them :)

zeebe/util/src/main/java/io/camunda/zeebe/util/ReflectUtil.java
```diff
@@ -28,6 +29,22 @@ public static <T> T newInstance(final Class<T> clazz) {
...
+   * object. This replaces the old Junit 5 `ReflectUtils.makeAccessible` which was removed from
+   * their platform.
+   *
+   * @param field the field to make accessible
+   * @param instance the instance on which we check accessibility
+   * @return the field, accessible
+   * @param <T> the type of the field
+   * @param <U> the type of the instance, typically just {@code Object}
+   */
+  public static <T extends AccessibleObject, U> T makeAccessible(final T field, final U instance) {
```
entangled90, line 43: I removed the 
```java
if (!field.canAccess(object))
```
because it will throw if field is not declared inside object, so it will fail in most cases.

JUnit original code would check for modifiers, but it does really make a difference, considering that it's used only in test classes

entangled90 (approved):

zeebe/util/src/main/java/io/camunda/zeebe/util/ReflectUtil.java
npepinpe: Let's do the same as them then :+1: Keep it identical.

As for `canAccess`, why would it fail in most cases? The expectation is you pass it an instance where the field is declared, no? 

Anyway, we could also omit the check, and just use `trySetAccessible` - it's a no-op when already true. If it returns false, we can throw the exception ourselves, or handle it in a way the test fails (throwing an `AssertionError` maybe?).

entangled90: Yes, for the field it would be ok, but the issue is for the method, which is defined in the `Closeable`, not in the test instance

PR #26724, opened 2025-01-09
feat: add OpenAPI String key objects

### LINKED ISSUE #26063: [C8 API] Add OpenAPI specification `String` keys support (8.6)
## Description

In order for our API documentation page to display both responses depending on the header, we need to define response objects with String and Numeric keys in the OpenAPI specification.

## Acceptance criteria

- Add OpenAPI responses with String and Long (in64) keys for the required APIs.
- Add OpenAPI response headers linked to the correct response object.
- Default response stays Long!
- No code changes required.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/26724
```

danielkelemen:
## Description

- Define OpenAPI string key objects.

## Related issues

closes https://github.com/camunda/camunda/issues/26063

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -70,6 +70,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/JobActivationResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/JobActivationResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/JobActivationResult"
         "400":
           description: >
             The provided data is not valid.
@@ -563,6 +569,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/UserTaskSearchQueryResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/UserTaskSearchQueryResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/UserTaskSearchQueryResult"
         "400":
           description: >
             The user task search query failed.
@@ -675,6 +687,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/CreateProcessInstanceResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/CreateProcessInstanceResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/CreateProcessInstanceResult"
         "400":
           description: The provided data is not valid.
         "500":
@@ -710,6 +728,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ProcessInstanceSearchQueryResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/ProcessInstanceSearchQueryResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/ProcessInstanceSearchQueryResult"
         "400":
           description: >
             The Process Instance Search Query failed.
@@ -889,6 +913,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/FlowNodeInstanceSearchQueryResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/FlowNodeInstanceSearchQueryResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/FlowNodeInstanceSearchQueryResult"
         "400":
           description: >
             The Flow node instance Search Query failed.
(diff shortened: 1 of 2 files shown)


danielkelemen:
Alternatively, we could ignore revapi with a single entry but it might be too vague:
```
{
  "ignore": true,
  "code": "java.annotation.attributeValueChanged",
  "annotationType": "com.fasterxml.jackson.annotation.JsonPropertyOrder",
  "attribute": "value",
  "justification": "Attribute order changes are not breaking changes."
}
```

clients/java/revapi.json
```diff
@@ -66,6 +66,134 @@
           "code": "java.class.removed",
           "old": "class io.camunda.zeebe.client.protocol.rest.UserTaskUpdateRequestChangeset",
           "justification": "The classes/methods are only internal and are not user-facing, and as such does not need to maintain any backwards compatibility guarantees."
+        },
```
tmetzke, line 69: I think this actually needs to go into an `ignored-changes.json` file instead. This one is for persistent changes over multiple releases. In that sense, some of the configurations above shouldn't be in here, I might have actually introduced them in the past 😅

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1150,6 +1204,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/MessagePublicationResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/MessagePublicationResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/MessagePublicationResponseResult"
```
tmetzke, line 1212: ❌ In 8.7, those are called differently (if I am not mistaken). That would lead to a breaking change from 8.6 to 8.7 since the DTO changes for the media type (same for all other response DTOs here). Maybe I am overlooking something though 😅 
```suggestion
                $ref: "#/components/schemas/MessagePublicationResult"
```

danielkelemen: No, you're right. Nice catch!

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2250,41 +2434,83 @@ components:
...
+        processInstanceKey:
+          type: string
+          description: The process instance key associated to this incident.
+        flowNodeInstanceKey:
+          type: string
+          description: The flow node instance key associated to this incident.
+        jobKey:
+          type: string
+          description: The job key, if exists, associated with this incident.
+    OperationItem:
```
tmetzke, line 2493: ❓ Do we even use the `OperationItem` DTO at all? I think we removed this in 8.7 since it's unused.

tmetzke (changes requested):
Two major things to look into:
1. The naming is off with 8.7, I think. That would become an issue when upgrading to 8.7 from 8.6 later.
2. The location for the Revapi ignores is not ideal, if I didn't miss anything.

clients/java/revapi.json
danielkelemen: I'll move them.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: It's used in the `ProcessInstanceItem`. But Git shows it as if I had added it but it was already there previously. Too many changes in one file, I guess.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: Fixed new `Result` names.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -563,6 +569,12 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/UserTaskSearchQueryResponse"
+            application/vnd.camunda.api.keys.number+json:
+              schema:
+                $ref: "#/components/schemas/UserTaskSearchQueryResponse"
+            application/vnd.camunda.api.keys.string+json:
+              schema:
+                $ref: "#/components/schemas/UserTaskSearchQueryResult"
```
tmetzke, line 577: 💭 Just something I noticed but is a follow-up: we don't do this media type handling for the search requests, only for the responses. The requests can contain filters on key attributes as well, though. Should we introduce this there as well, at least on OpenAPI-level? I don't think we have to adjust the REST controllers in any way. But for users it needs to be clear they can also send keys as Strings, right?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Ah yeah, I see. We actually removed this in 8.7 but on an alpha endpoint so I think we're fine.

tmetzke (approved):
👍 Looks good to me 🚀

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: Yes, I guess those need adjustment as well.
I'll create a follow-up ticket for it.

pepopowitz, line 577: 💭 If I'm understanding correctly, we are distinguishing between the two types of keys by naming schemas as a "Result" or a "Response".

Is there a reason to not be more explicit about those different types, like including "number" or "string" in the name? It seems likely that (a) someone new to the project would take longer to intuit why some schemas were named Result and some Response, and (b) even someone with experience on the project would forget the distinction and need to inspect the spec to figure it out.

pepopowitz (approved):
👍 I copied this spec into my local branch with a docusaurus upgrade, and everything looks as I'd expect it to.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Mostly to keep the DTOs names clean. We will remove the number-based types with (I think) the October '25 release. So, with 8.8, we'll only have the new string-based DTOs. There is nothing "special" about them anymore then, they have keys and those are strings.
Changing the existing number-based DTO names also doesn't work well since this would be a regression from 8.6 to 8.7 in the spec which is at least not great. The impact could be OK-ish but is not worth the issues it might cause.

Hope that makes sense in any way 😅

PR #26737, opened 2025-01-10
fix: @TestZeebe resources are closed even when constructed in a @BeforeEach

entangled90:
## Description
When a resource was constructed in a `@BeforeEach` callback, the `app` field in `ClusterResource` or `ApplicationResource` would be null as it was not yet instantiated.
The value of the field is get on every access to ensure this does not happen.

This problem is mostly apparent when some `static` field is initialized only once, because every test case gets a fresh instance of the class. (I found out this from ` io.camunda.operate.conditions.DatabaseInfo.getCurrent()`

### DIFF
--- a/zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/junit/ZeebeIntegrationExtension.java
+++ b/zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/junit/ZeebeIntegrationExtension.java
@@ -167,13 +167,13 @@ final class ZeebeIntegrationExtension
     // starting one fails
     resources.forEach(resource -> store.put(resource, resource));
     for (final var resource : resources) {
-      final var directory = createManagedDirectory(store, resource.cluster.name());
+      final var directory = createManagedDirectory(store, resource.cluster().name());
       manageCluster(directory, resource);
     }
   }
 
   private void manageCluster(final Path directory, final ClusterResource resource) {
-    final var cluster = resource.cluster;
+    final var cluster = resource.cluster();
 
     // assign a working directory for each broker that gets deleted with the extension lifecycle,
     // and not when the broker is shutdown. this allows to introspect or move the data around even
@@ -198,7 +198,7 @@ final class ZeebeIntegrationExtension
     // assign a working directory to the broker that gets deleted with the extension lifecycle,
     // and not when the broker is shutdown. this allows to introspect or move the data around even
     // after stopping a broker
-    if (resource.app instanceof final TestStandaloneBroker broker) {
+    if (resource.app() instanceof final TestStandaloneBroker broker) {
       final var directory = createManagedDirectory(store, "broker-" + broker.nodeId().id());
       setWorkingDirectory(directory, broker.nodeId(), broker);
     }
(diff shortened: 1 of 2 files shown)


zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/junit/ZeebeIntegrationExtension.java
```diff
@@ -250,97 +250,102 @@ private Path createManagedDirectory(final Store store, final String prefix) {
...
     return extensionContext.getStore(Namespace.create(ZeebeIntegrationExtension.class));
   }
 
-  private record ClusterResource(TestCluster cluster, TestZeebe annotation)
+  private record ClusterResource(Object testInstance, Field field, TestZeebe annotation)
       implements TestZeebeResource, CloseableResource {
 
+    public TestCluster cluster() {
+      try {
+        return (TestCluster) field.get(testInstance);
```
npepinpe, line 273: :thought_balloon: Can we/should we memoize this value? It doesn't hurt too much anyway since this isn't exactly in the hot path, but just wondering.

zeebe/qa/util/src/test/java/io/camunda/zeebe/qa/util/cluster/ZeebeIntegrationTest.java
```diff
@@ -0,0 +1,39 @@
...
+  public static void checkAllClustersAreClosed() {
+    ASSERTIONS.forEach(c -> assertThat(c.isStarted()).isFalse());
+  }
+
+  @Test
+  public void shouldCloseCluster() {
+    testCluster = new TestStandaloneBroker();
+    testCluster.start();
+
+    ASSERTIONS.add(testCluster);
```
npepinpe, line 37: :wrench: Let's assert it is started here to ensure the invariant.

npepinpe, line 37: :x: Please comment what this is testing and why we're doing it this way, as it's not immediately obvious.

npepinpe (approved):
:rocket: Please address all comments before merging :pray:

zeebe/qa/util/src/main/java/io/camunda/zeebe/qa/util/junit/ZeebeIntegrationExtension.java
entangled90: Considering the performance are not an issue, I assumed getting the field from the class each  time can avoid potential bugs (such as memoizing null)

npepinpe:
Quick check, should we backport this?

entangled90:
Good point.
From a quick check branch `stable/8.6` is using the same junit version, so it's probably better to backport it.

npepinpe:
Do you know if we have automated minor updates for <= 8.5? If so, we might update junit and break it again, so we should backport there. If not, then it's likely OK. I think we anyway only have it for versions 8.5, maybe 8.4? :thinking:

entangled90:
Also this [fix](https://github.com/camunda/camunda/commit/1ef6ccc1206918d74afbfd804ea1da9570e17758) was not backported, so probably these branches don't really have junit 5.11, but 5.10 (coming from spring)

npepinpe:
I don't know if we have `@TestZeebe` on 8.3 tbh

entangled90:
Yes it exists

PR #26463, opened 2025-01-02
refactor: moved object storage testcontainers to camunda/qa for testing backups

entangled90:
## Description
Util classes used by zeebe/qa/integration-tests that can be reused by common backup tests are moved to the common qa/util module

## Related issues

relates #25935

PR #26740, opened 2025-01-10
Revert "refactor: moved object storage testcontainers to camunda/qa for testing backups"

entangled90:
Reverts camunda/camunda#26463


npepinpe:
We can move them to `zeebe-test-util` so they're available everywhere. My point was that we should avoid having a circular dependency between the root and the sub-modules, as typically the root should depend on the sub modules, not the other way around (if that makes sense). If we need a shared container module, we look into that instead, but `zeebe-test-util` should be OK for now.

entangled90:
Should I close this Revert PR and open a new PR with the suggested changes?

npepinpe:
Open the new PR first, then we can close this one once that's accepted.

npepinpe:
Shall we close this in favor of the other one?

entangled90:
Closing as classes have been moved in #26923

PR #26754, opened 2025-01-10
fix: update get form schema descriptions

### LINKED ISSUE #26746: Update the descriptions for Get Form
**Description**

- The description of schema can be improved to The form content
- In general, the descriptions we have in the Tasklist V1 API are a lot more elaborate, let’s reuse them in V2 more

**Acceptance Criteria**

- Have the descriptions update on OpenApi Scpec reviewed by a tech writter + API team

nathansandi:
## Description

Update the attribute descriptions for the Get Form endpoint. 

## Related issues

closes https://github.com/camunda/camunda/issues/26746

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -7989,10 +7989,10 @@ components:
           description: The BPMN ID of the form.
           type: string
         schema:
-          description: The schema of the form.
+          description: The form content.
           type: object
         version:
-          description: The version of the form.
+          description: The version of the the deployed form.
           type: integer
           format: int64
     FormItem:


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -7989,10 +7989,10 @@ components:
           description: The BPMN ID of the form.
```
tmetzke, line 7989: ❓ Since we're at it right now: what is this information? The id I provide as a user in the form JSON and that I use in a user task to link to the form?

tmetzke (approved):
👍 Looks good to me 🍪 
❓ I added an understanding question, maybe you know the answer. If so, we could make that more understandable for users while we're at it.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nathansandi: Thanks for the question @tmetzke 

The ID refers to the Form ID present on the modeler, (similar how we do in tasks and process)
image

We can be more specific over this field specifying what refers considering the modeler on the attribute description.
What do you think?

tmetzke: OK, I see, thank you, @nathansandi 👍 
With this in mind, the name `bpmnId` doesn't make any sense for this property, if you ask me. It's actually the `formId` (we have to properly differentiate it from the `formKey` property then by description, e.g. `The user-provided identifier of the form` while the `formKey` rather is `The internal and unique system identifier of the form`).

tmetzke: @pepopowitz, FYI, regarding the glossary idea, I believe it could make sense to provide a similar wording for all `id` versus `key` attributes in the API. IDs are usually user-provided identifiers then, whereas keys usually are system-generated (I think the only exception currently is the `correlationKey` used in messaging endpoints and in the future maybe `businessKey`).

tmetzke: @nathansandi, can you create a follow-up to rename the `bpmnId` attribute here to `formId` and to adjust the description of this attribute and the `formKey`?

nathansandi: Hello @tmetzke 
Follow up issue created:
https://github.com/camunda/camunda/issues/26911

PR #26776, opened 2025-01-13
fix: operate status show

liuyq1209:
## Description

During one polling, both "fetchProcessInstance" and "fetchProcessInstanceDetailStatistics" were initiated simultaneously. However, the "processInstance" was directly updated in the success callback function of "fetchProcessInstance".

```
handleFetchSuccess = (processInstance: ProcessInstanceEntity | null) => {
    this.state.processInstance = processInstance;
    this.state.status = 'fetched';
  };
```

However, at this time, this part will be triggered and thus "stopPolling" will be executed. then, 

```
 this.completedFlowNodesDisposer = when(
      () =>
        processInstanceDetailsStore.state.processInstance !== null &&
        !processInstanceDetailsStore.isRunning &&
      () => {
        this.stopPolling();
      },
    );
```

Well, then there's a pretty high chance that after the response of the last "fetchProcessInstanceDetailStatistics" comes back, the polling will have already been stopped before it can be dealt with in time. As a result, the final state won't be updated on the graph.

so, i add this line

```
this.completedFlowNodesDisposer = when(
      () =>
        processInstanceDetailsStore.state.processInstance !== null &&
        !processInstanceDetailsStore.isRunning &&
        // when there is no active flow node, then cancel the polling;
        this.state.statistics.filter((v) => v.active).length === 0,
      () => {
        this.stopPolling();
      },
    );
```

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/operate/client/src/modules/stores/processInstanceDetailsStatistics.ts
+++ b/operate/client/src/modules/stores/processInstanceDetailsStatistics.ts
@@ -15,15 +15,15 @@ import {
   observable,
   computed,
 } from 'mobx';
-import {processInstanceDetailsStore} from 'modules/stores/processInstanceDetails';
-import {NetworkReconnectionHandler} from './networkReconnectionHandler';
+import { processInstanceDetailsStore } from 'modules/stores/processInstanceDetails';
+import { NetworkReconnectionHandler } from './networkReconnectionHandler';
 import {
   fetchProcessInstanceDetailStatistics,
   ProcessInstanceDetailStatisticsDto,
 } from 'modules/api/processInstances/fetchProcessInstanceDetailStatistics';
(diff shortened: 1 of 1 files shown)


vitorwtavares (changes requested):
Hello @liuyq1209. 

I'm not sure I see how the line you added will make a difference, as in the previous line we check if the process instance is still active, and if so, polling will not be stopped, so wouldn't that be a redundant check?

It would be great if you could explain a bit better what exactly is the problem that this cause, provide us with an example and steps to reproducing it so we can check it out.

After that, if we determine the validity of this change and you'd still like to go forward with merging it, I'd ask that you please put the PR in a mergeable state - meaning up to date with the main branch and passing all CI checks (tests passing, linting issues resolved and unnecessary changes removed).

PR #26798, opened 2025-01-13
feat: add variable name filter for the user task variable query

### LINKED ISSUE #26745: Query user task variable: add filter by variableName
**Description**

Add filter by **variableName** on Query User Task Variables

Acceptance Criteria:
- Filter with advanced capabilities Variables by **variableName**

nathansandi:
## Description

Add filtering capabilities on Search User Task Variable endpoint to the follow attributes:

- name

## Related issues

closes https://github.com/camunda/camunda/issues/26745

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -4094,7 +4094,7 @@ components:
       type: object
       properties:
         filter:
-          description: The user task search filters.
+          description: The user task variable search filters.
           allOf:
             - $ref: "#/components/schemas/VariableUserTaskFilterRequest"
     UserTaskSearchQueryResponse:
@@ -4201,7 +4201,7 @@ components:
           type: string
           description: The value of the variable.
     VariableUserTaskFilterRequest:
-      description: Variable
+      description: The user task variable search filters.
       type: object
       properties:
         name:
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -4092,6 +4092,11 @@ components:
         - $ref: "#/components/schemas/SearchQueryRequest"
       description: User task search query request.
       type: object
+      properties:
+        filter:
+          description: The user task search filters.
+          allOf:
+            - $ref: "#/components/schemas/VariableUserTaskFilterRequest"
     UserTaskSearchQueryResponse:
       description: User task search query response. Key attributes as numeric values.
       deprecated: true
@@ -4195,6 +4200,14 @@ components:
         value:
           type: string
           description: The value of the variable.
+    VariableUserTaskFilterRequest:
+      description: Variable
+      type: object
+      properties:
+        name:
+          description: Name of the variable.
+          allOf:
+            - $ref: "#/components/schemas/StringFilterProperty"
     UserTaskItemBase:
       description: Base properties for UserTaskItem.
       type: object
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
@@ -348,6 +348,8 @@ public final class SearchQueryRequestMapper {
     if (request == null) {
       return Either.right(SearchQueryBuilders.variableSearchQuery().build());
     }
+
+    final var filter = toUserTaskVariableFilter(request.getFilter());
     final var page = toSearchQueryPage(request.getPage());
     final var sort =
         toSearchQuerySort(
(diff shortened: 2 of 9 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4092,6 +4092,11 @@ components:
         - $ref: "#/components/schemas/SearchQueryRequest"
       description: User task search query request.
       type: object
+      properties:
+        filter:
+          description: The user task search filters.
```
tmetzke, line 4097:
```suggestion
          description: The user task variable search filters.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4092,6 +4092,11 @@ components:
...
+          allOf:
+            - $ref: "#/components/schemas/VariableUserTaskFilterRequest"
```
tmetzke, line 4099:
```suggestion
            - $ref: "#/components/schemas/UserTaskVariableFilterRequest"
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4195,6 +4200,14 @@ components:
         value:
           type: string
           description: The value of the variable.
+    VariableUserTaskFilterRequest:
```
tmetzke, line 4203:
```suggestion
    UserTaskVariableFilterRequest:
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4195,6 +4200,14 @@ components:
...
+      description: Variable
```
tmetzke, line 4204:
```suggestion
      description: The user task variable search filters.
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -348,15 +348,30 @@ public static Either<ProblemDetail, VariableQuery> toUserTaskVariableQuery(
...
             SortOptionBuilders::variable,
             SearchQueryRequestMapper::applyVariableSortField);
 
-    return buildSearchQuery(
-        FilterBuilders.variable().build(), sort, page, SearchQueryBuilders::variableSearchQuery);
+    return buildSearchQuery(filter, sort, page, SearchQueryBuilders::variableSearchQuery);
+  }
+
+  private static VariableFilter toUserTaskVariableFilter(
+      final VariableUserTaskFilterRequest filter) {
```
tmetzke, line 364:
```suggestion
      final UserTaskVariableFilterRequest filter) {
```

tmetzke (approved):
👍 The provided code looks good.
🔧 I added some smaller naming and description suggestions.
❓ General question: do we need to add something in the RDBMS component as well so we can filter user task variables accordingly by name?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
nathansandi: `UserTaskVariableFilterRequest` already exists.
Do you have maybe another idea? I had the same thoughts over here

tmetzke: Oh, I wasn't aware, thanks for that hint 😄 
In that case, just leave it as is, your naming is fine then 👍

nathansandi:
Hey @tmetzke  related to this question: 
> ❓ General question: do we need to add something in the RDBMS component as well so we can filter user task variables accordingly by name?

Related to this one :) I am on it. I have asked to the RDBMS team if they have it in their plate, otherwise I will create an issue and work on that. 
It looks like the operation is not supported by RDBMS yet.

tmetzke:
Thanks, @nathansandi, yeah, that makes sense. One last question then: does any of our newly added integration tests cover RDBMS testing? Otherwise, I'd ask to at least at that so we can ensure this works end-2-end in the future still (so no refactorings introduce a regression here).

nathansandi:
> Thanks, @nathansandi, yeah, that makes sense. One last question then: does any of our newly added integration tests cover RDBMS testing? Otherwise, I'd ask to at least at that so we can ensure this works end-2-end in the future still (so no refactorings introduce a regression here).

We have integration tests running over the queries, for example:
- https://github.com/camunda/camunda/blob/ce7f968afb0b4bb56782d16cc23a6ac23d662fd6/qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/usertask/UserTaskIT.java 
- https://github.com/camunda/camunda/blob/40e74a9b969dce4a91e17ea31246fa94eacf606e/qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/decisionrequirements/DecisionRequirementsIT.java
Which runs the tests over the entity, but not proper over the client. 

Where in CI is named as RDBMS IT Tests

For validate the queries, this test makes sense and looks enough. But so far, the best practice would be setup the client [module](https://github.com/camunda/camunda/blob/c16aba7aeef7fd3eb9c514a782fe4ab8c59544a4/qa/integration-tests/src/test/java/io/camunda/it/client) to run  with the RDMS profile, not just ES and OS. Which I believe this should be in a separate issue, once it involves the CI setup. 

Do you know if  there any reason why those tests was written over a different module?

tmetzke:
OK, good points, @nathansandi 👍 
I'd say we're good enough with this setup so far, let's merge this.
The unification of QA tests will come soon, this is an ongoing topic. It will ensure that all QA by default runs on all supported data layers for secondary storage (ES/OS/RDBMS), you'll only write the test cases once.

PR #26811, opened 2025-01-13
Work with username instead of user key in the Gateway (REST+GRPC)

### LINKED ISSUE #26778: Replace the user key on commands with a username
## Description

- Find the usages of the [AUTHORIZED_USER_KEY](https://github.com/camunda/camunda/blob/17e76cbee7bf9c87287c42b8478e814cd8761563/zeebe/auth/src/main/java/io/camunda/zeebe/auth/Authorization.java#L14)
- Rename the property to `AUTHORIZED_USERNAME` and make sure we set the username as value, instead of the user key.
    - In the `AuthorizedCheckBehavior` we need to find the user key by username and find the user with the key we found. This is a temporary step until we've modified the state.

remcowesterhoud:
## Description

We will move from working with a user key to working with a username throughout the stack. In the gateways we currently find the authenticated user and put the user key on the command. With this PR we will change this and put the username on the command instead.

## Checklist

- [ ] for CI changes:

## Related issues

closes #26778

### DIFF
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
@@ -549,6 +549,7 @@ public class RequestMapper {
                 .toList());
         if (authenticatedPrincipal instanceof final CamundaUser user) {
           authenticatedUserKey = user.getUserKey();
+          claims.put(Authorization.AUTHORIZED_USERNAME, user.getUsername());
         }
       }
 
@@ -558,10 +559,6 @@ public class RequestMapper {
             .forEach(
                 (key, value) -> claims.put(Authorization.USER_TOKEN_CLAIM_PREFIX + key, value));
       }
-    }
-
-    if (authenticatedUserKey != null) {
-      claims.put(Authorization.AUTHORIZED_USER_KEY, authenticatedUserKey);
     }
 
     return new Builder()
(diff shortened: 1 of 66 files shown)


remcowesterhoud:
Tests are now fixed, ready for review 😄

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -549,6 +549,7 @@ public static Authentication getAuthentication() {
                 .toList());
         if (authenticatedPrincipal instanceof final CamundaUser user) {
           authenticatedUserKey = user.getUserKey();
+          claims.put(Authorization.AUTHORIZED_USERNAME, user.getUsername());
```
koevskinikola, line 552: ❓ Do we need to add a `null` check to this line, like we had before? Or is it guaranteed that the `username` is always set in the `CamundaUser`?

koevskinikola (approved):
Looks good @remcowesterhoud. 👍 

I only have a small point of discussion/feedback below, but it's not critical for a re-review, so I'm pre-approving this PR.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
remcowesterhoud: I'll always be set. We fetch the user from ES and create this `CamundaUser` [here](https://github.com/camunda/camunda/blob/main/authentication/src/main/java/io/camunda/authentication/CamundaUserDetailsService.java#L66-L75). And username is a required field.

remcowesterhoud: Regarding the null-check before. We had it on the `auhtenticatedUserKey`. We set that only in 1 place. So I moved putting this claim there. We only need it for basic auth and with basic auth we will always have the user.

PR #26812, opened 2025-01-13
fix: in NettyMessaging service heartbeats are sent as req/reply based on role

### LINKED ISSUE #26602: java.lang.ClassCastException: class ProtocolRequest cannot be cast to class ProtocolReply
**Describe the bug**

In SaaS INT environment a `ClassCastException` causes cluster to fail on startup.

```
java.lang.ClassCastException: class io.atomix.cluster.messaging.impl.ProtocolRequest cannot be cast to class io.atomix.cluster.messaging.impl.ProtocolReply
```

In some way, `ProtocolRequest` object is passed to `AbstractClientConnection.dispatch(final ProtocolReply message)` whereas it has to be a `ProtocolReply` object:

https://github.com/camunda/camunda/blob/d49e93a9265b7498572db228f0da1541594de1fc/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/AbstractClientConnection.java#L37-L38

It is triggered by `channelRead0` method with a wrong type:

https://github.com/camunda/camunda/blob/d49e93a9265b7498572db228f0da1541594de1fc/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/NettyMessagingService.java#L1247-L1256

**To Reproduce**
I am not sure how `ProtocolRequest` gets passed to wrong connection type. It needs to be passed to server side connection instead of client connection.

**Expected behavior**

Either `ProtocolRequest` always passed to server side connection or we handle such case gracefully.

**Log/Stacktrace**

 <p>

```
java.lang.ClassCastException: class io.atomix.cluster.messaging.impl.ProtocolRequest cannot be cast to class io.atomix.cluster.messaging.impl.ProtocolReply (io.atomix.cluster.messaging.impl.ProtocolRequest and io.atomix.cluster.messaging.impl.ProtocolReply are in unnamed module of loader 'app')
	at io.atomix.cluster.messaging.impl.AbstractClientConnection.dispatch(AbstractClientConnection.java:29) ~[zeebe-atomix-cluster-8.6.0.jar:8.6.0]
	at io.atomix.cluster.messaging.impl.NettyMessagingService$MessageDispatcher.channelRead0(NettyMessagingService.java:1175) ~[zeebe-atomix-cluster-8.6.0.jar:8.6.0]
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:346) ~[netty-codec-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:318) ~[netty-codec-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1357) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:440) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:868) ~[netty-transport-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.epoll.AbstractEpollStreamChannel$EpollStreamUnsafe.epollInReady(AbstractEpollStreamChannel.java:799) ~[netty-transport-classes-epoll-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:501) ~[netty-transport-classes-epoll-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:399) ~[netty-transport-classes-epoll-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[netty-common-4.1.113.Final.jar:4.1.113.Final]
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[netty-common-4.1.113.Final.jar:4.1.113.Final]
	at java.base/java.lang.T
```

</p>
</details>

**Environment:**
- Zeebe Version: `8.6.0`
- Environment: `SaaS INT`

entangled90:
# Description 

Heartbeats are now initiated only from the client, which sends a `ProtocolRequest` message with subject `"internal-heartbeat"` whenever the channel has been idle for too much.
The server, upon receive such message, will send a `ProtocolReply` response with the same id as the request.
The client is then able to diagnose missed heartbeats and close the channel if he does not receive a reply in time.

If the server has received at least one heartbeat, but it did not receive an heartbeat for more than the specified timeout, it will close the connection.
This will avoid closing the connection when a client does not support heartbeats (i.e. a previous version)

## Related issues
closes #26602

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
@@ -1,0 +1,130 @@
+/* * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under * one or more contributor license agreements. See the NOTICE file distributed * with this work for additional information regarding copyright ownership. * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.atomix.cluster.messaging.impl;
+
+import io.atomix.cluster.messaging.impl.ProtocolReply.Status;
+import io.atomix.utils.net.Address;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.timeout.IdleState;
+import io.netty.handler.timeout.IdleStateEvent;
+import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicLong;
+import org.agrona.collections.LongHashSet;
+import org.slf4j.Logger;
+
+abstract sealed class HeartbeatHandler extends ChannelDuplexHandler {
+  // DO NOT CHANGE: changing the subject is a backward INCOMPATIBLE change.
+  static final String HEARTBEAT_SUBJECT = "internal-heartbeat";
+  static final byte[] HEARTBEAT_PAYLOAD = new byte[0];
+  protected final Logger log;
+
+  protected HeartbeatHandler(final Logger log) {
+    this.log = log;
+  }
+
+  static final class Server extends HeartbeatHandler {
+    boolean receivedHeartbeat = false;
+
+    Server(final Logger log) {
+      super(log);
+    }
+
+    @Override
+    public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
+      if (msg instanceof final ProtocolRequest request
+          && request.subject().equals(HEARTBEAT_SUBJECT)) {
+        receivedHeartbeat = true;
+        ctx.writeAndFlush(createHeartbeat(request.id()));
+        if (!Arrays.equals(request.payload(), HEARTBEAT_PAYLOAD)) {
+          log.warn(
+              "Received unexpected heartbeat payload from {}, perhaps the message subject {} is accidentally reused",
+              request.sender(),
+              HEARTBEAT_SUBJECT);
+        }
+      }
+
+      // Pass the message to the next handler, even though it's a heartbeat.
+      ctx.fireChannelRead(msg);
+    }
+
+    @Override
+    public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
+      if (!(evt instanceof final IdleStateEvent idleStateEvent)) {
+        return;
+      }
+      if (idleStateEvent.state() == IdleState.READER_IDLE) {
+        // don't close the connection if we did not receive any heartbeat
+        if (receivedHeartbeat) {
+          log.warn("Connection {} timed out on the server, closing channel", ctx.channel());
+          ctx.close();
+        }
+      }
+    }
+
+    private ProtocolMessage createHeartbeat(final long id) {
+      return new ProtocolReply(id, HEARTBEAT_PAYLOAD, Status.OK);
+    }
+  }
+
+  static final class Client extends HeartbeatHandler {
+    private final LongHashSet outstandingHeartbeats = new LongHashSet();
+    private final AtomicLong messageIdGenerator;
+    private final Address advertisedAddress;
+
+    Client(final Logger log, final AtomicLong messageIdGenerator, final Address advertisedAddress) {
+      super(log);
+      this.messageIdGenerator = messageIdGenerator;
+      this.advertisedAddress = advertisedAddress;
+    }
+
+    @Override
+    public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
+      if (msg instanceof final ProtocolReply reply && outstandingHeartbeats.contains(reply.id())) {
+        // remove all heartbeats sent before the last update
+        outstandingHeartbeats.removeIfLong(id -> id <= reply.id());
+        if (reply.status() != Status.OK) {
+          log.warn("Received a Heartbeat response with status {}", reply.status());
+        } else if (!Arrays.equals(reply.payload(), HEARTBEAT_PAYLOAD)) {
+          log.warn(
+              "Received unexpected heartbeat payload, perhaps the message subject {} is accidentally reused: {}",
+              HEARTBEAT_SUBJECT,
+              reply);
+        }
+      }
+
+      // Pass the message to the next handler, even though it's a heartbeat.
+      ctx.fireChannelRead(msg);
+    }
+
+    @Override
+    public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
+      if (!(evt instanceof final IdleStateEvent idleStateEvent)) {
+        return;
+      }
+      switch (idleStateEvent.state()) {
+        case READER_IDLE -> {
+          // don't close the connection if we did not send any heartbeat
+          if (!outstandingHeartbeats.isEmpty()) {
+            log.warn("Connection {} timed out on the client, closing channel", ctx.channel());
+            outstandingHeartbeats.clear();
+            ctx.close();
+          }
+        }
+        case WRITER_IDLE -> {
+          ctx.writeAndFlush(createHeartbeat());
+        }
+        default -> {}
+      }
+    }
+
+    private ProtocolRequest createHeartbeat() {
+      return new ProtocolRequest(
+          messageIdGenerator.incrementAndGet(),
+          advertisedAddress,
+          HEARTBEAT_SUBJECT,
+          HEARTBEAT_PAYLOAD);
+    }
+  }
+}
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
@@ -47,23 +47,27 @@ abstract sealed class HeartbeatHandler extends ChannelDuplexHandler {
   static final String HEARTBEAT_SUBJECT = "internal-heartbeat";
   static final byte[] HEARTBEAT_PAYLOAD = new byte[0];
   protected final Logger log;
+  protected final boolean forwardHeartbeats;
 
-  protected HeartbeatHandler(final Logger log) {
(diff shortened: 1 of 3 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -0,0 +1,130 @@
...
+      }
+
+      // Pass the message to the next handler, even though it's a heartbeat.
+      ctx.fireChannelRead(msg);
+    }
+
+    @Override
+    public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
+      if (!(evt instanceof final IdleStateEvent idleStateEvent)) {
+        return;
```
npepinpe, line 104: :question: Should we not forward it to the next handler? :thinking:

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -0,0 +1,130 @@
...
```
npepinpe, line 55: :question: Why don't we forward it to the next handler? Or does this work differently than others?

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -0,0 +1,130 @@
...
+      }
+      switch (idleStateEvent.state()) {
+        case READER_IDLE -> {
+          // don't close the connection if we did not send any heartbeat
```
npepinpe, line 108:
```suggestion
          // don't close the connection if we haven't sent any heartbeat yet
```

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -0,0 +1,130 @@
...
+          if (!outstandingHeartbeats.isEmpty()) {
+            log.warn("Connection {} timed out on the client, closing channel", ctx.channel());
```
npepinpe, line 110: :wrench: `"Connection {} timed out on the client after not receiving a heartbeat response from the server in {} ({} heartbeats pending)"`, ctx.channel(), timeout, outstandingHeartbeats.size())"`

For timeout, ideally I would know what's the oldest heartbeat I sent. But it's OK if that's particularly annoying info to get, we can drop that.

npepinpe (changes requested):
:x: It would be good to have metrics for this, mostly around missed heartbeats, so we can use them for diagnosis (with labels for the client/server addresses). Did we initially omit them because we expect it would have too high cardinality, like the others?

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -0,0 +1,130 @@
...
+      ctx.fireChannelRead(msg);
+    }
+
+    @Override
+    public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
+      if (!(evt instanceof final IdleStateEvent idleStateEvent)) {
+        return;
+      }
+      if (idleStateEvent.state() == IdleState.READER_IDLE) {
+        // don't close the connection if we did not receive any heartbeat
```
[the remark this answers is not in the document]
entangled90: Yes :smile:

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -0,0 +1,130 @@
...
+    Client(final Logger log, final AtomicLong messageIdGenerator, final Address advertisedAddress) {
+      super(log);
+      this.messageIdGenerator = messageIdGenerator;
+      this.advertisedAddress = advertisedAddress;
+    }
+
+    @Override
+    public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
+      if (msg instanceof final ProtocolReply reply && outstandingHeartbeats.contains(reply.id())) {
+        // remove all heartbeats sent before the last update
```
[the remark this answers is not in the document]
entangled90: I added this in order to keep the size `outstandingHeartbeats` small and to avoid it grow.
Because of that, I prematurely delete stale heartbeats whenever a new heartbeat is received. 
I'll add a comment :+1:

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
entangled90: Potentially we could, but I don't think there's any other handler that consumes it at the moment. 
The cardinality of these events is not that much as they are triggered only when the channel is Idle.

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
entangled90: Same as above

entangled90:
We probably need to check on the client as well if we ever received any heartbeat response from the server and close the connection only if the server supports them.

Another thing feels a bit wrong to me: 
In the client we are not creating a timer (with hearbeat timeout) for each heartbeat request that we are sending, but we only react when no data has been received from the channel.
This means that we could keep receiving data from the server, but no heartbeat responses and we would not close the connection.

npepinpe:
We should piggyback heartbeats on other messages for sure. So using correlation IDs in the protocol messages to synchronize between client and server, instead of expecting a specific heartbeat message. Does that make sense? We can quickly sync on it if I find time, but I'm pretty sure Lena would understand what I mean and can explain it if not :smile:

entangled90:
That's a very good point. I'll do as you suggested

entangled90:
One last improvement we could add is to send the heartbeat timeout from the client to the server as part of the handshake. That will avoid false positives if server and client have different heartbeat timeouts configured.
However, it requires a change to the binary protocol, so it would be better to do it in another PR.

entangled90:
:thought_balloon: I noticed that for every message receive whose subject has no handlers,  we log at debug. This will create too many logs, so I'm thinking to add a parameter to the constructor so that we can still test it, but in production it will be disabled

entangled90:
I'll create a new issue for adding a new protocol version so that the timeout is sent from the client to the server

npepinpe (approved):
:rocket: Before we release this, we should improve it so the heartbeat feature is part of the V3 protocol and clients/servers can negotiate it, I think it's a better approach than the current detection.

Additionally, we should try and piggyback off existing message by adding a logical clock to them, instead of sending additional messages. Tentatively, I think this would require another protocol change, so better to do before 8.7, but if you think not, we could put this in the backlog.

npepinpe (approved):

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Override
+    public void userEventTriggered(final ChannelHandlerContext ctx, final Object evt) {
+      if (!(evt instanceof final IdleStateEvent idleStateEvent)) {
+        return;
+      }
+      if (idleStateEvent.state() == IdleState.READER_IDLE) {
+        // don't close the connection if we did not receive any heartbeat
+        if (receivedHeartbeat) {
+          log.warn("Connection {} timed out on the server, closing channel", ctx.channel());
```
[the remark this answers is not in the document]
entangled90: sorry missed it :-D

PR #26742, opened 2025-01-10
ci: add backport label when touching the deprecated java client in 8.7

houssain-barouni:
## Description

As described in https://github.com/camunda/camunda/pull/26689 to mitigate double maintenance of Camunda and Zeebe client in main branch and in the future stable/8.7

`backport main` label is added to PRs touching zeebe client classes in stable/8.7, this ensure that we don't forget to update the Zeebe client classes in main branch (and future stable/8.8)

The scenario that can happen:
1- a developer does a fix in Camunda client classes
2- developer backports the fix to stable/8.7 => Zeebe client classes are changed in stable/8.7
3- the label `backport main` is automatically added to the backport PR, and ensure that the Zeebe client classes are also updated in main branch

I also add `deprecated-client` label (to be set with Red color) in a PR when a it updates files in from Zeebe client namespace

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

relates to https://github.com/camunda/camunda/issues/26814

PR #24781, opened 2024-11-15
fix: handle RawApiEntityConsumer buffer size correctly

chillleader:
## Description

Handle the buffer size of the RawApiEntityConsumer correctly - grow the buffer according to the number of bytes received, but only up to the max message size.

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

closes #24779

PR #26813, opened 2025-01-13
 Bring back Zeebe client classes to the camunda-client-java module

houssain-barouni:
## Description

The following steps were done:
- bring Zeebe client API classes (packages `io.camunda.zeebe.client` and `io.camunda.zeebe.client.api` from `stable/8.6`: https://github.com/camunda/camunda/pull/26813/commits/b2ecf8a9d902d4234243fc2873993084569b2e0e
- copy implementation classes from `main` to ensure that the implemetation is compatible with 8.7 server side: https://github.com/camunda/camunda/pull/26813/commits/7c03c5c8ca27fb95ab79ba89c8f30dd2058f7655
- `JobChangeset` and `ProblemDetail` protocol classes are exposed in the API, as the protocol package has changed, I added a wrapper classes keeping the previous package https://github.com/camunda/camunda/pull/26813/commits/582cade06bd37876bffeda78ea95cd657df9f482 and https://github.com/camunda/camunda/pull/26813/commits/075ac5fb71ffe48bdec9f85841f084bb409092a9
- The datamodel of some experimental APIs has changed:  https://github.com/camunda/camunda/pull/26813/commits/0aeacdc8c2f328ca64f0dc5d8ad9183ba899214d
- Make zeebe client send `user-agent` http header with `zeebe-client-java` value: https://github.com/camunda/camunda/pull/26813/commits/5366a7fa5423a3a29991b84e8756761bd3192e02
- bringing zeebe client unit tests from `stable/8.6` : https://github.com/camunda/camunda/pull/26813/commits/871da3c6a977f13cb2c1ce50c87db3240be7d244

## Checklist

- [ ] for CI changes:

## Related issues

relates to https://github.com/camunda/camunda/issues/26814


tmetzke:
> * copy implementation classes from `main` to ensure that the implemetation is compatible with 8.7 server side: [7c03c5c](https://github.com/camunda/camunda/commit/7c03c5c8ca27fb95ab79ba89c8f30dd2058f7655)

That is interesting. I see the following issues with this:
* If we need this, the 8.6 client might in fact not be forward-compatible, i.e., I cannot use the 8.6 client with an 8.7 cluster in all cases, is that correct?
* This could make forward "backporting" via the automation from `stable/8.6` complicated as the implementation classes are not in sync anymore, right?
* Do we have a changeset of what really needs to change in the 8.6 implementation classes so they work in the 8.7 codebase?

tmetzke:
Thanks, @houssain-barouni, I see. I did a diff with `stable/8.6` for this branch, ignoring the added classes for the new Camunda client.

## Current API adjustments from 8.6 to 8.7 in the Zeebe client
image

* 🟢 In `AddPermissionsCommandStep1` we expose REST DTOs for which the package changed. We need to clean this up in 8.7. It's part of an API that should be marked as experimental in 8.6 (we forgot that there and should patch that). So this is OK I think.
* 🟡 In the `PartitionBrokerHealth` and `PartitionBrokerRole` we added the `UNKNOWN_ENUM_VALUE` which is used in the client implementation. We can add this but we don't have to. It's an improvement to enum handling which we implemented with 8.7. We won't backport this to 8.6 so the question is: should we have this in Zeebe client 8.7?
* 🟢 In `ProcessInstanceFilter` and `UserTaskFilter` we expose REST DTOs for filtering variables. We need to fix this in Camunda client 8.7 so we don't expose DTOs in any filter API. For 8.7 Zeebe client we can move to the new REST DTOs or fix this as well, both works since it's experimental in 8.6.
* 🟢 In `ProcessInstance` we remove `getOperations` and `getCallHierarchy`. The class is only used in experimental API in 8.6 (process instance search), so changing this should be OK.
* 🟡 In `SearchRequestBuilders`, we now throw `UnsupportedOperationException`s for all methods. I am not sure yet why we have to do this in 8.7, @houssain-barouni. Can you shed some light on this? Strictly speaking, those methods are not marked as experimental in 8.6 and they are in the API package.

## Current IMPL adjustments from 8.6 to 8.7 in Zeebe client
* _TBD (will add this later)_

houssain-barouni:
> * 🟢 In `AddPermissionsCommandStep1` we expose REST DTOs for which the package changed. We need to clean this up in 8.7. It's part of an API that should be marked as experimental in 8.6 (we forgot that there and should patch that). So this is OK I think.

👍 we can handle that in https://github.com/camunda/camunda/issues/26851

> * 🟡 In the `PartitionBrokerHealth` and `PartitionBrokerRole` we added the `UNKNOWN_ENUM_VALUE` which is used in the client implementation. We can add this but we don't have to. It's an improvement to enum handling which we implemented with 8.7. We won't backport this to 8.6 so the question is: should we have this in Zeebe client 8.7?

In 8.6, unknown value was triggering RuntimeExcpetions https://github.com/camunda/camunda/blob/d2bf292663999de4c0031fbc1f902a9185319849/clients/java/src/main/java/io/camunda/zeebe/client/impl/response/PartitionInfoImpl.java#L43-L46
I think we can keep, it is a non-breaking improvement that we add in 8.7

> * 🟢 In `ProcessInstanceFilter` and `UserTaskFilter` we expose REST DTOs for filtering variables. We need to fix thi
s in Camunda client 8.7 so we don't expose DTOs in any filter API. For 8.7 Zeebe client we can move to the new REST DTOs or fix this as well, both works since it's experimental in 8.6.

👍 https://github.com/camunda/camunda/issues/26851

> * 🟢 In `ProcessInstance` we remove `getOperations` and `getCallHierarchy`. The class is only used in experimental API in 8.6 (process instance search), so changing this should be OK.

👍 

> * 🟡 In `SearchRequestBuilders`, we now throw `UnsupportedOperationException`s for all methods. I am not sure yet why we have to do this in 8.7, @houssain-barouni. Can you shed some light on this? Strictly speaking, those methods are not marked as experimental in 8.6 and they are in the API package.

I removed all the Filter implementations because the filter interfaces, such as ProcessInstanceFilter, diverged between versions 8.6 and 8.7. This change is acceptable since these builders are intended to create objects for invoking the experimental search APIs. It’s known that they won’t work with ZeebeClient. While the build will not fail, a runtime exception will prompt users to switch to CamundaClient.

tmetzke:
> 👍 we can handle that in #26851

Thanks a lot for creating that one 👍 

> I think we can keep it. It is a non-breaking improvement that we added in 8.7

Yeah, it's fine, let's keep the improvement.

> I removed all the Filter implementations because the filter interfaces, such as ProcessInstanceFilter, diverged between versions 8.6 and 8.7. This change is acceptable since these builders are intended to create objects for invoking the experimental search APIs. It’s known that they won’t work with ZeebeClient. While the build will not fail, a runtime exception will prompt users to switch to CamundaClient.

Did they all diverge that much really? In the end, I'm generally fine with this. It says to "move to Camunda client" for search queries which generally makes sense since it was experimental only in 8.6.

OK, so we're good on the API side (tackling the REST DTOs in the follow-up) 👍 

I'll have another look at the IMPL classes as I think there might be some changes that are going too far a bit or are at least debatable if we want to be able to forward "backport" from 8.6 later.

tmetzke:
_continuation of https://github.com/camunda/camunda/pull/26813#issuecomment-2589462939_

## Current IMPL adjustments from 8.6 to 8.7 in Zeebe client
image
image

* 🟡 The classes in the `command` and `response` packages all move to the new String-based key attribute DTOs. Looking at https://github.com/camunda/camunda/pull/26853, we'll return long keys for Zeebe clients. This would allow to keep the command classes mostly as they are in 8.6, using the same DTOs (but adjusting the package, removing "zeebe" from the imports). That would keep the command classes closer to what we have in 8.6 and should make any potential backports from there to 8.7 easier to understand. I would vote for copying the 8.6 classes and adjusting the packages. Anything that needs to be fixed afterward to make the tests (also just copied from 8.6 and left unchanged as much as possible) pass can be tracked/fixed individually and might actually uncover any regressions we introduced in 8.7.
* 🟢/🟡 The whole `search` package is removed in 8.7. The changes diverged a lot since we developed this quite a lot since 8.6. It was experimental in 8.6 and we hint at using the Camunda client. I think this is fine for 8.7 in the Zeebe classes.
* 🟡 The `http` package received some development in 8.7, not all of that is reflected in 8.6 as a patch. If we incorporate the new 8.7 code in Zeebe classes, backporting from 8.6 to 8.7 might be confusing in the future. If we want any of the improvements from 8.7 in 8.6, let's backport them and also adjust the Zeebe classes in 8.7.
  * 🟡 Same for the `oauth` and also `response` packages, I think.
* 🟡 The `Loggers` should use the 8.6 code instead.
* 🟡 The `ZeebeClientImpl` introduces unsupported operation exceptions, but the text mentions "ZeebeClient" twice.

Overall, I was wondering if we not maybe rather copy the 8.6 code here and adjust where needed (e.g. package names of REST DTOs). If the migration would take too much effort (e.g. introducing old REST DTOs classes manually), we can think about removing the implementation, e.g. like we did for the search parts.

houssain-barouni:
> * 🟡 The classes in the `command` and `response` packages all move to the new String-based key attribute DTOs. Looking at [feat: read java client header in LongSerializer #26853](https://github.com/camunda/camunda/pull/26853), we'll return long keys for Zeebe clients. This would allow to keep the command classes mostly as they are in 8.6, using the same DTOs (but adjusting the package, removing "zeebe" from the imports). That would keep the command classes closer to what we have in 8.6 and should make any potential backports from there to 8.7 easier to understand. I would vote for copying the 8.6 classes and adjusting the packages. Anything that needs to be fixed afterward to make the tests (also just copied from 8.6 and left unchanged as much as possible) pass can be tracked/fixed individually and might actually uncover any regressions we introduced in 8.7.
> * 🟢/🟡 The whole `search` package is removed in 8.7. The changes diverged a lot since we developed this quite a lot since 8.6. It was experimental in 8.6 and we hint at using the Camunda client. I think this is fine for 8.7 in the Zeebe classes.
> * 🟡 The `http` package received some development in 8.7, not all of that is reflected in 8.6 as a patch. If we incorporate the new 8.7 code in Zeebe classes, backporting from 8.6 to 8.7 might be confusing in the future. If we want any of the improvements from 8.7 in 8.6, let's backport them and also adjust the Zeebe classes in 8.7.
>   
>   * 🟡 Same for the `oauth` and also `response` packages, I think.
> * 🟡 The `Loggers` should use the 8.6 code instead.
> * 🟡 The `ZeebeClientImpl` introduces unsupported operation exceptions, but the text mentions "ZeebeClient" twice.
> 
> Overall, I was wondering if we not maybe rather copy the 8.6 code here and adjust where needed (e.g. package names of REST DTOs). If the migration would take too much effort (e.g. introducing old REST DTOs classes manually), we can think about removing the implementation, e.g. like we did for the search parts.

Thank you for identifying the differences. 🚀 
My reasoning for bringing the implementation classes from `main` instead of `stable/8.6` is coming from concerns about potential “implementation details” that might cause issues when using the 8.6 client with an 8.7 cluster. While such issues might not arise, my concern is that detecting them could be challenging due to the lack of sufficient automated tests for checking the forward compatibility of the client.
However, since the client code is now duplicated, the same concern applies to the `main` branch code, as `ZeebeClient` isn’t tested against the 8.7 cluster either.

houssain-barouni:
@tmetzke I copied 8.6 implementation with small adaptions https://github.com/camunda/camunda/pull/26813/commits/0b16e2825e02ec2c7b45ceb5c5d33117f7a69db7

tmetzke:
> Thank you for identifying the differences. 🚀 My reasoning for bringing the implementation classes from `main` instead of `stable/8.6` is coming from the potential "implementation details" that might cause issues when using the 8.6 client with an 8.7 cluster. While such issues might not arise, my concern is that detecting them could be challenging due to the lack of sufficient automated tests for checking the forward compatibility of the client. However, since the client code is now duplicated, the same concern applies to the `main` branch code, as `ZeebeClient` isn’t tested against the 8.7 cluster either.

That is totally reasonable, I had the same impulse in the beginning, to be honest 😅 
You're right about the testing part, of course. I was wondering if it makes sense to invest in testing the 8.7 Zeebe client against an 8.7 cluster now with QA tests. I am a bit reluctant though as this also simply creates duplicated code. But we ship the Zeebe client with 8.7 now and actually need to be sure it works with an 8.7 cluster then.

Regarding the changes here themselves, I compared the branch again with `stable/8.6` and I think this looks good now, really good job, @houssain-barouni 👍

There are now some fixes in 8.7 Camunda client that could be backported to 8.6, e.g. the client part of https://github.com/camunda/camunda/pull/26317. With that, we can also test the forward backporting to `main` after we merged https://github.com/camunda/camunda/pull/26742.

tmetzke (approved):

houssain-barouni:
> There are now some fixes in 8.7 Camunda client that could be backported to 8.6, e.g. the client part of #26317. With that, we can also test the forward backporting to `main` after we merged #26742.

@tmetzke I need first to backport https://github.com/camunda/camunda/pull/24781, so of the code to backport depends on it

tmetzke:
> @tmetzke I need first to backport #24781

Ah nice, I was looking for that one! Perfect, yes, let's do that first 👍

PR #26816, opened 2025-01-13
ci: add daily build to Unified CI to re-populate GHA caches

### LINKED ISSUE #25284: CI: re-populate GHA caches with `main` builds after nightly purge
## Description

As [observed the GitHub purges the GHA cache around 4 AM UTC](https://github.com/camunda/camunda/wiki/CI-&-Automation#github-actions-cache) each day. As per our [caching strategy](https://github.com/camunda/camunda/wiki/CI-&-Automation#caching-strategy) only `main` (default branch) and `stable/*` branch builds write to the GHA cache.

This means all PR builds after 4 AM UTC don't have a GHA cache available until the 1st `main` branch build of that day → this can lead to longer build runtimes, decreased developer experience and higher cost e.g. for Renovate PRs.

## Goal

Implement a mechanism to re-populate the GHA caches with a `main` branch build after the ~4 AM UTC nightly purge for the [Unified CI](https://github.com/camunda/camunda/wiki/CI-&-Automation#unified-ci) [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml). Calculate with some buffer and consider using GHA `scheduled` trigger for 6 AM UTC. Verify that the re-population works at least once after merging your changes to `main`.

## Hints

* @cmur2 is available for questions, pairing and code reviews

kalmarz:
## Description

Add a daily full build to the Unified CI to re-populate GHA caches in the morning.

## Checklist

- [x] for CI changes:

## Related issues

closes #25284

### DIFF
--- a/.github/actions/paths-filter/action.yml
+++ b/.github/actions/paths-filter/action.yml
@@ -1,6 +1,12 @@
 ---
 name: Common paths filter for camunda/camunda
 
+inputs:
+  trigger_source:
+    required: false
+    default: "unknown"
+    description: "Source of the workflow trigger"
+
 description: |
   Common filters to detect and group changes against a base branch
   It leverages the `dorny/paths-filter` action to filter paths
@@ -12,6 +18,7 @@ outputs:
     value: >-
       ${{
         github.event_name == 'push' ||
+        (github.event_name == 'schedule' && inputs.trigger_source == 'Unified CI') ||
         steps.filter-common.outputs.github-actions-change == 'true'
       }}
   maven-spotless-linter:
@@ -19,6 +26,7 @@ outputs:
     value: >-
       ${{
         github.event_name == 'push' ||
+        (github.event_name == 'schedule' && inputs.trigger_source == 'Unified CI') ||
         steps.filter-common.outputs.github-actions-change == 'true' ||
         steps.filter-common.outputs.java-code-change == 'true' ||
         steps.filter-common.outputs.maven-change == 'true' ||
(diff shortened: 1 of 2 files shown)


.github/actions/paths-filter/action.yml
```diff
@@ -1,6 +1,12 @@
 ---
 name: Common paths filter for camunda/camunda
 
+inputs:
+  trigger_source:
+    required: false
+    default: "unknown"
+    description: "Source of the workflow trigger"
```
cmur2, line 8: What is your motivation for adding this input? I see it is used to decide to only run `on: schedule` when triggered by the Unified CI but what does it protect against?

I know the Tasklist CI is using the `paths-filter` action but I don't think they run on schedule (and even if they do, there might be no downside)?

kalmarz: Indeed. I found a few scheduled workflows in the repo already. None of them uses the path filter at the moment but it might change in the future. Even new workflows could use it. The downside I can think of is the longer run, so 💵 eventually. So this is just a safety belt.

cmur2: Considering the use case of the paths-filter action (it should tell if certain parts of the CI need to be run based on what changed) I think it makes sense to always trigger on scheduled workflows, not just for the Unified CI.

I tried to come up with an example where a workflow is run on schedule & uses change detection and then does _not_ want to run their GHA job when triggered on schedule which is rather unlikely - the opposite (as for the Unified CI) is more likely for future workflows.

So I think it is fine to drop the `trigger_source` source which also makes the code easier to follow.

cmur2 (approved):

PR #26823, opened 2025-01-13
fix: make backoff configurable

sdorokhova:
## Description

In `BackgroundTask` mechanism we use idle backoff with hard-coded 1 minute maximum delay between runs. For runtime this is OK, but too big for tests. 
In this PR I make this maximum delay configurable, so that we can pass smaller value in E2E tests.

## Related issues

related with #24084


EuroLew (approved):
LGTM

PR #26824, opened 2025-01-13
fix: prio6 indices were not considered when opensearch is configured

entangled90:
## Description
For some reason, the prio6 indices were not part of the backup on opensearch probably because of a bad merge/formatting.


npepinpe (approved):
:rocket: wups :smile:

PR #26827, opened 2025-01-13
fix: Fixing issue with Zeebe compatibility tests

grlimacan:
## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/.github/workflows/optimize-zeebe-compatibility.yml
+++ b/.github/workflows/optimize-zeebe-compatibility.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        zeebe-version: [8.3.0, 8.4.0, 8.5.0, 8.6.0, latest, snapshot]
+        zeebe-version: [8.3.0, 8.4.0, 8.5.0, 8.6.0, snapshot]
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       # Parses the pom within less than second and exposes every value as an output of the step
--- a/optimize/backend/src/it/java/io/camunda/optimize/AbstractCCSMIT.java
+++ b/optimize/backend/src/it/java/io/camunda/optimize/AbstractCCSMIT.java
@@ -92,16 +92,9 @@ public abstract class AbstractCCSMIT extends AbstractIT {
   protected static boolean isZeebeVersion87OrLater() {
     final Pattern zeebeVersionPattern = Pattern.compile("8.([7-9]|\\d{2,})");
     return zeebeVersionPattern
-        .matcher(IntegrationTestConfigurationUtil.getZeebeDockerVersion())
-        .matches()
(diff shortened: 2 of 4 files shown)


optimize/backend/src/it/java/io/camunda/optimize/AbstractCCSMIT.java
```diff
@@ -100,14 +93,22 @@ protected static boolean isZeebeVersion87OrLater() {
     final Pattern zeebeVersionPattern = Pattern.compile("8.([7-9]|\\d{2,})");
     return zeebeVersionPattern
         .matcher(IntegrationTestConfigurationUtil.getZeebeDockerVersion())
-        .matches();
+        .matches()
+        || isZeebeVersionSnapshot();
   }
 
+  // After the 8.7.0 release, isZeebeVersion87OrLater() needs to call this method instead of
+  // isZeebeVersionSnapshot(), since latest is until then an 8.6.x release
```
RomanJRW, line 101: So does this become a change then that needs making each time a new minor release is made? Did I understand this right?

RomanJRW (commented):
One comment but it's quite an important one I think. I am wondering what value we even get from testing `latest` and whether we can simply just not do that. Instead, we can depend on the `.0` release to be correct enough, and that we don't break our Semver for patches after that

optimize/backend/src/it/java/io/camunda/optimize/AbstractCCSMIT.java
grlimacan: Not each time. This concrete method just needs to be changed once, since we only have this situation while `latest` is not a 8.7.0 version. This issue will however re-occur if we need to make checks for e.g. `is 880 or later`, because then at that time latest will be `8.7.x`. We could consider whether we actually have any value in testing `latest`, since that is sort of covered by testing the `8.6.0`. Then we don't need to make this differentiaton (see screenshot). What do you think @RomanJRW, shall I just remove `latest` from the pipeline and also the entire logic for it in the IT?
image

RomanJRW: Yeah I would say it's easier and cleaner to remove latest testing completely. We don't get much value from it, like you say 👍

grlimacan:
@RomanJRW please re-review, here's an example run: https://github.com/camunda/camunda/actions/runs/12765176859

.github/workflows/optimize-zeebe-compatibility.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        zeebe-version: [8.3.0, 8.4.0, 8.5.0, 8.6.0, latest, snapshot]
```
RomanJRW, line 18: we would also need to modify the `get-zeebe-version-from-tag.sh` script right?

RomanJRW (approved):
Generally LGTM, just one more place needs changes

PR #26832, opened 2025-01-13
feat: change default Optimize's default indexPrefix from "optimize" to ""

### LINKED ISSUE #26825: change default optimize prefix to "" from "optimize"

entangled90:
## Description
- [X] Default changed in `service-config.yaml`
- [X] Default changed in `OptimizeIndexNameService` and added a static `COMPONENT_NAME` (which equals `"optimize"`

## Related issues
closes #26825

### DIFF
--- a/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/db/DatabaseConstants.java
+++ b/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/db/DatabaseConstants.java
@@ -10,24 +10,15 @@ package io.camunda.optimize.service.db;
 public final class DatabaseConstants {
 
   public static final String AGGREGATION_CONTAINS_NULL = "isNull";
-
   // Note: we cap listings to 1000 as a generous practical limit, no paging
   public static final int LIST_FETCH_LIMIT = 1000;
-
   public static final int MAX_RESPONSE_SIZE_LIMIT = 10_000;
-
   public static final int NUMBER_OF_DATA_POINTS_FOR_AUTOMATIC_INTERVAL_SELECTION = 80;
-
   public static final int NUMBER_OF_RETRIES_ON_CONFLICT = 5;
-
   public static final int IGNORE_ABOVE_CHAR_LIMIT = 7000;
-
   public static final int MAX_GRAM = 10;
-
   public static final int DEFAULT_SHARD_NUMBER = 1;
-
   public static final String OPTIMIZE_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
-
   public static final String NUMBER_OF_SHARDS_SETTING = "number_of_shards";
   public static final String SORT_NULLS_FIRST = "_first";
   public static final String SORT_NULLS_LAST = "_last";
@@ -36,15 +27,12 @@ public final class DatabaseConstants {
   public static final String NGRAM_TOKENIZER = "ngram_tokenizer";
   public static final String IS_PRESENT_FILTER = "is_present_filter";
   public static final String IS_PRESENT_ANALYZER = "is_present_analyzer";
-
   public static final String DECISION_DEFINITION_INDEX_NAME = "decision-definition";
   public static final String DECISION_INSTANCE_MULTI_ALIAS = "decision-instance";
   public static final String DECISION_INSTANCE_INDEX_PREFIX = "decision-instance-";
-
   public static final String PROCESS_DEFINITION_INDEX_NAME = "process-definition";
   public static final String PROCESS_INSTANCE_MULTI_ALIAS = "process-instance";
   public static final String PROCESS_INSTANCE_INDEX_PREFIX = "process-instance-";
-
   public static final String SINGLE_PROCESS_REPORT_INDEX_NAME = "single-process-report";
   public static final String SINGLE_DECISION_REPORT_INDEX_NAME = "single-decision-report";
   public static final String COMBINED_REPORT_INDEX_NAME = "combined-report";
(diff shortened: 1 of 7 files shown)


optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/db/DatabaseConstants.java
```diff
@@ -10,24 +10,15 @@
 public final class DatabaseConstants {
 
   public static final String AGGREGATION_CONTAINS_NULL = "isNull";
```
entangled90, line 12: Unfortunately editing this file ran the formatter which created  all these whitespaces changes

optimize/util/optimize-commons/src/main/resources/service-config.yaml
```diff
@@ -386,7 +386,7 @@ opensearch:
     index:
       # the prefix prepended to all Optimize index and alias names
       # NOTE: Changing this after Optimize was already run before, will create new empty indexes
-      prefix: ${CAMUNDA_OPTIMIZE_OPENSEARCH_SETTINGS_INDEX_PREFIX:optimize}
+      prefix: ${CAMUNDA_OPTIMIZE_OPENSEARCH_SETTINGS_INDEX_PREFIX:}
```
RomanJRW, line 389: I think this gets parsed to null if I'm not mistaken, so an isEmpty call would NPE

entangled90: Good point! Let me check, I assumed that with the colon it would interpret it as ""

entangled90: I added a test in `ConfigurationServiceTest` and it get parsed as ""

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/db/schema/OptimizeIndexNameService.java
```diff
@@ -114,20 +119,33 @@ public static String getOptimizeIndexOrTemplateNameForAliasAndVersion(
...
+      return COMPONENT_NAME;
+    } else {
+      return String.format("%s-%s", indexPrefix, COMPONENT_NAME);
     }
-    return original.toLowerCase(Locale.ENGLISH);
   }
 
   public String getIndexPrefix() {
-    return indexPrefix;
+    return completeIndexPrefix;
```
entangled90, line 143: the complete prefix is then "{customPrefix}-optimize" so that when it's used for searching all optimize indices, it will not show indices from other apps as well (for example when indexPrefix is "")

entangled90:
There are quite a few tests that are failing, but I'm not sure if it's because they're flaky. In the meantime we can start taking a look at the changes.

buccarel (approved):

optimize/util/optimize-commons/src/main/resources/service-config.yaml
```diff
@@ -501,7 +501,7 @@ es:
     index:
       # the prefix prepended to all Optimize index and alias names
       # NOTE: Changing this after Optimize was already run before, will create new empty indexes
-      prefix: ${CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_PREFIX:optimize}
+      prefix: ${CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_PREFIX:}
```
RomanJRW, line 504: as above, my understanding was that this setting would go away and be replaced by the global prefix? Or instead is the intention just to reuse this config option for the global prefix value?

optimize/util/optimize-commons/src/test/java/io/camunda/optimize/service/db/schema/OptimizeIndexNameServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @ParameterizedTest
+  @ValueSource(strings = {"", "my-index"})
+  public void shouldReturnIndexWithPrefixAndComponent(final String indexPrefix) {
+    final var nameService = new OptimizeIndexNameService(indexPrefix);
+
+    assertThat(
+            nameService.getOptimizeIndexOrTemplateNameForAliasAndVersionWithPrefix(
+                "position-index", "8.3"))
+        .isEqualTo(withHyphen(indexPrefix) + "optimize-position-index_v8.3");
```
RomanJRW, line 29: not a big deal, but the Optimize index versions are just integers (unrelated to the actual project version), so lets reflect that in this test. Concretely, just use `v8` rather than `v8.3`

optimize/util/optimize-commons/src/test/java/io/camunda/optimize/service/db/schema/OptimizeIndexNameServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import io.camunda.optimize.service.db.es.schema.index.index.PositionBasedImportIndexES;
+import org.junit.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class OptimizeIndexNameServiceTest {
+
+  @ParameterizedTest
+  @ValueSource(strings = {"", "my-index"})
```
RomanJRW, line 22: what do you think about including a prefix that also contains the component name? i.e. `"my-optimize-index"`.

Same suggestion for all tests below

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/db/schema/OptimizeIndexNameService.java
```diff
@@ -103,6 +106,7 @@ private void setIndexPrefix(
     } else {
       indexPrefix = configurationService.getElasticSearchConfiguration().getIndexPrefix();
     }
+    setIndexPrefix(indexPrefix);
```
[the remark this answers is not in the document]
entangled90: It's just because I would move to use the global prefix in another PR. I would just focus on changing the default in this PR. What do you think?

optimize/util/optimize-commons/src/main/resources/service-config.yaml
entangled90: Focus of this PR is just about changing the default value, we will remove this setting in a follow up PR

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/db/schema/OptimizeIndexNameService.java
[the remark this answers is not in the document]
RomanJRW: That sounds perfectly fine to me 👍

RomanJRW (approved):
Thanks for all your work on this @entangled90! 💪 

I think we can approve this, but we should make sure we have quite heavy manual testing to accommodate the lack of good automated testing in Optimize ✅

PR #26834, opened 2025-01-13
Improve CommandRateLimiter logging to detect persistent back pressure

### LINKED ISSUE #26807: Improve `CommandRateLimiter` logging to detect persistent back-pressure
**Description**

This is a follow-up issue for #26104  since we were not able to reproduce the reported issue.

As we analyzed the data, we found that the value of the Limiter's `inFlight` counter could be higher than the value of `limit`. This can happen in expected situations, i.e. when processing a whitelisted command, but we also suspected that was also a possibility with commands that aren't whitelisted and were able to reproduce this with some unit tests.
Therefore, #26408 fixes this possibility of concurrent updates to the `inFlight` counter after comparing it with the value of `limit` before creating a new `Limiter.Listener`.

### Possible improvements
To better determine the root cause for the persistent back-pressure issue we need to improve the logging and there are multiple options which are not exclusive.

Logging of the values of fields like `inFlight`, `currentInflight`, `limit`, `rejectedCounter` , `droppedCounter`, `ignoredCounter`  and the current number of `Listeners` in `responseListeners`.
Currently, we have a `trace` statement in `CommandApiRequestHandler` when a call to `limiter.tryAcquire` returns `false` which occurs when the command isn't whitelisted and `inFlight >= limit`, but since the `trace` log level is usually disabled, it should not help and even if it was enabled it would fill the logs in the event of persistent back-pressure.

There are different approaches to trigger this logging but we have to ponder the resources consumption of each:
1. After some period of time, which is not ideal as requires a scheduled worker to handle this.
2. After a number of received commands (100 * `limit?), which is better than **1** as we rely on a _modulo_ comparison to determine when to log the information.
3. When Whitelisted commands are accepted and `inFlight` is equal to `limit`. This corresponds to the saturation point of the limiter but can flood the log in periods with heavy load.
4. When there is a role transition. Since the problem seems associated with periods with various role transitions, this can help depict the initial and final state of the limiters when the broker transitions between different roles. For instance, we can log the final state of a `RequestLimiter` of a partition in `removePartition` and its initial state in `addPartition` in `CommandApiRequestHandler`.
5. Store the value of `startTime`, assigned during the listener creation, for the last accepted command on a field of `CommandRateLimiter` and when a command is rejected, which happens when `inFlight` is greater or equal to `limit`, verify if a sufficiently large period of time, which can then be parameterized according to nature of the workload, has elapsed, trigger the logging messages warning that the partition is not processing any commands.

### Hypothesis for `inFlight` not decrementing as expected

- Listeners for commands are somehow "lost", having been withdrawn from `responseListeners` but the appropriate callback (`onSuccess`, `onIgnore` or `onDropped`) was never invoked, causing the `inFlight` value to be higher than it should. However this seems very unlikely as the analyzed heap dumps show that `responseListeners` contains the same number of elements as the value of `inFlight`.
- Commands are not being processed at all and `inFlight` counter is stuck, eventually increasing past the value of `limit` when the execution of white-listed commands is requested.

filipecampos:
## Description

A first implementation of approach 5 described in #26807, which logs some warning messages if the `CommandRateLimiter` hasn't accepted any command for more than 10 minutes.

## Related issues

related to #26807

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiter.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiter.java
@@ -20,6 +20,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -40,10 +41,12 @@ public final class CommandRateLimiter extends AbstractLimiter<Intent>
   private final Map<ListenerId, Listener> responseListeners = new ConcurrentHashMap<>();
   private final int partitionId;
   private final BackpressureMetrics metrics = new BackpressureMetrics();
+  private Long lastAcceptedCommandTimestamp;
 
   protected CommandRateLimiter(final CommandRateLimiterBuilder builder, final int partitionId) {
     super(builder);
     this.partitionId = partitionId;
+    lastAcceptedCommandTimestamp = 0L;
     metrics.setInflight(partitionId, 0);
     metrics.setNewLimit(partitionId, getLimit());
   }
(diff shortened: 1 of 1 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiter.java
```diff
@@ -68,17 +71,32 @@ private void registerListener(final int streamId, final long requestId, final Li
     responseListeners.put(new ListenerId(streamId, requestId), listener);
   }
 
+  private boolean isProcessingStuck() {
+    final long timeSinceLastCommand = System.nanoTime() - lastAcceptedCommandTimestamp;
+    if (timeSinceLastCommand > TimeUnit.MINUTES.toNanos(10)) {
```
npepinpe, line 76: :wrench: Interval should be configurable

zeebe/broker/src/main/java/io/camunda/zeebe/broker/transport/backpressure/CommandRateLimiter.java
```diff
@@ -68,17 +71,32 @@ private void registerListener(final int streamId, final long requestId, final Li
...
+      LOG.warn(
+          "This partition hasn't accepted commands for more than 10 minutes! Validate if the partition is stuck.");
+      LOG.warn(
+          "Current status: inFlight={} limit={} responseListeners={}",
+          getInflightCount(),
+          getLimit(),
+          responseListeners.size());
```
npepinpe, line 83: :wrench: Log statements can be merged.

npepinpe (commented):
:+1: I like the approach, that's clever - so if I understood correctly, since the lambda in `tryAcquire` is called only if we did acquire, we can then check every time we reject if it's been over some amount of time, and if it's too long, print that.

:x: Let's combine that with a metric every time we detect, a simple binary switch, so we can alert on it. You can alert on logs, but it's a little bit messier.

PR #26852, opened 2025-01-14
test: POC for new way of integration tests

ChrisKujawa:
## Description

As discussed @tmetzke @npepinpe I did a small POC to try it out. I would say it works.

### Details

Create an Camunda IT invocation provider (based on earlier existing provider), that can be registered with @ExtendWith. It will resolve parameters and static fields, to fill necessary CamundaClient instances. Allows to migrate easily existing test cases.

The extension will make sure that the Platform is started, but not secondary storage. It detects (right now defaults to ES) the secondary storage (fex via env var) and then sets the correct connection details. Here we need to use good defaults.

It will not set up ES, OS or RDBMS. It will expect they are started already. Allows to parallize tests, and reuse the same instance.

Test data should be prefixed with test class names.

Execution was successful for migrated test:

external-db

Normal test execution before:

normalexecution

I think the test container start up is not counted in the statistics in Intellij.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


npepinpe:
Brainstorming notes:

# Limitations
- junit 5 does not allow parameterizing at the class level

# Potential solutions
## Per-DB job
- Somehow we can identify and run DB specific
- Job runs for each DB this group of test
- Likely have a junit extension to perform the wiring based on some simple setting (e.g. env var or property or some easy toggle)
- Still use ZeebeIntegration (or some variant) to handle starting/configuring the application
- Pros:
	- Possibly faster by removing the overhead of starting ES/OS/RDBMS (possibly since we may want to support different vendors and so on)
	- Centralized container configuration, making it easier to work with
	- Simplified support solution for AWS OpenSearch, possibly in the future vendor-specific DBs
	- Jobs run in parallel "for free" (except it costs money)
- Cons:
	- Pushes more logic into CI
	- Poor integration with IntelliJ workflow
		- There are workarounds/improvements so may not be a big deal
	- Possible issues if data isolation doesn't work, might not be immediately obvious what the issue is
### TestInvocationContextProvider
- Is only for DBs, so one context per DB
- To avoid overhead of running a single test, DB is instantiated on first access
- Still use ZeebeIntegration (or some variant) to handle starting/configuring the application
-

entangled90:
I think in the screenshots even the time to startup our application is not taken into account.
After running that tests in my machine I noticed that the time is spent in the following sections:
- 1 second: testcontainers initialization
- 10 seconds: start ES container
- 11 seconds: starting Camunda (up to the line `io.camunda.zeebe.qa.util.cluster.TestSpringApplication - Started TestSpringApplication`)
- 1 second: stopping Camunda

So if we don't start the ES container in each test we will probably gain around 10 seconds out of 23 seconds of "overhead" for each class (around ~ 43%).

Do you think we can cut down the app startup time as well? This would make the move to the external DB instance even more compelling.

ChrisKujawa:
To document what I did today.

 * I was able to create a new extension that solely focus on detecting the database type, and configure the related TestApplication
 * Right now I created a separate TestApplication - copied from TestStandaloneCamunda; For simplicity and to not break other tests
 * The new test application has been extended such that it allows to be configured for ES completely and OS (this didn't worked before)
 * The extension was a copy from previous BrokerItProvider class, it has been cleaned and lot of code removed.
   * I decided to add the authentication client feature later. Likely I will separate this into a extra extension, to separate concerns. 
   * This can then still work with annotations, as it right now does. 
   * The workflow would be to create outside the test application, let it configured based on the new CamundaDBExtension and then use the AuthClientExtension (or however it is called) to create a authenticated client.
 * The new extension uses the test class name as prefix to isolate test data.
 * Furthermore, is the extension able to delete all data that has been created
   * We might can provide this as a functionality for Backups tests as well 
 * We have tested the migrated test with ES but also with OS (we got the OS setup running). OS needed some adjustments, and more configuration but it is working now
 *

ChrisKujawa:
I got it working the first time ! 🥳 You can't imagine how relieved I am 😅 
https://github.com/camunda/camunda/actions/runs/12813891301/job/35729049971#step:9:149

ChrisKujawa:
POC was successful, and MVP has been implemented as well (https://github.com/camunda/camunda/pull/27106, https://github.com/camunda/camunda/pull/27109, https://github.com/camunda/camunda/pull/27113)  . I will close this PR.

PR #26855, opened 2025-01-14
feat: use ProcessModelReader for rdbms + fix reading of start formId …

### LINKED ISSUE #26848: Use new ProcessModelReader instead of XMLUtil for Process Record extraction in RDBMS
The old `XMLUtil` will be replaced with this PR with a new `io.camunda.zeebe.util.modelreader.ProcessModelReader` which should be used by the new exporters. 

Additionally, we currently don't extract the start formId correctly, for this, `ProcessModelReader.extractStartFormLink` needs to be used. 

AC:
* XMLUtil is not longer used
* formId is set correctly

p-wunderlich:
## Description

Use new ProcessModelReader instead of XMLUtil for rdbms + fix reading of start formId

## Related issues

closes #26848


stoerti (approved):
💭 When there is an error in parsing the XML, there is no error/warning in the log. But also don't know who would see the error when we just continue processing...

PR #26862, opened 2025-01-14
feat: support form-based login for identity

emulbreh:
## Description

This PR adds a `formLogin` security filter chain for auth-basic and a login form for identity client.


authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
```diff
@@ -47,22 +51,25 @@ public class WebSecurityConfig {
   public static final String LOGIN_URL = "/login";
```
Ben-Sheppard, line 51: 🔧 Its not in this PR but these can be descoped down to `private`

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
```diff
@@ -47,22 +51,25 @@ public class WebSecurityConfig {
...
-        // deprecated Tasklist v1 Public Endpoints
-        "/v1/external/process/**",
-        "/new/**"
-      };
+  public static final Set<String> UNAUTHENTICATED_PATHS =
+      Set.of(
+          LOGIN_URL,
+          "/",
+          // identity is a single page app
+          "/identity/**",
```
Ben-Sheppard, line 59: ❓ Would we need this for all applications? I think current the other apps are still functioning and are single page applications

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
```diff
@@ -129,50 +128,117 @@ private void genericSuccessHandler(
...
 
   @Bean
   public FilterRegistrationBean<TenantRequestAttributeFilter>
       tenantRequestAttributeFilterRegistration(final MultiTenancyConfiguration configuration) {
     return new FilterRegistrationBean<>(new TenantRequestAttributeFilter(configuration));
   }
+
+  private static boolean isBasicAuthRequest(final HttpServletRequest request) {
+    final String authorizationHeader = request.getHeader(HttpHeaders.AUTHORIZATION);
+    // If the client sends an `Authorization: Basic` header, treat this as a basic auth request.
```
Ben-Sheppard, line 188: 💭 NABD but I typically try to avoid comments unless its something difficult to understand or has a specific not visible purpose. I think some of these comments don't meet that criteria for me but I'm not going to push on this :)

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
```diff
@@ -129,50 +128,117 @@ private void genericSuccessHandler(
...
+    if (authorizationHeader != null && authorizationHeader.startsWith("Basic ")) {
+      return true;
+    }
+    // If it's an API request and there's no session cookie, also treat the request as a basic auth
+    // request. This is a workaround for Java's HTTP client, which only sends an Authorization
+    // header after the server sends a WWW-Authenticate header in a 401 response.
+    if (isApiRequest(request) && !hasSessionCookie(request)) {
+      return true;
+    }
+    return false;
```
Ben-Sheppard, line 198:
```suggestion
return isApiRequest(request) && !hasSessionCookie(request);
```

dist/src/main/java/io/camunda/webapps/controllers/IndexController.java
```diff
@@ -29,7 +29,7 @@ public String index() {
    * Redirects the old frontend routes (common tasklist and operate routes) to the default-app
    * sub-path. This can be removed after the creation of the auto-discovery service.
    */
-  @GetMapping({"/processes", "/login"})
+  @GetMapping({"/processes"})
```
Ben-Sheppard, line 32: ❓ Are we now stopping the redirect of `/login` for any specific reason? I was trying to think of if this impacts us with the Identity login page

Ben-Sheppard (commented):
Some comments on my side. I had a look at the code locally and it acts as I would expect for the basic login, I'm able to login and logout with the default user (and also with a created user).

I couldn't see where it is in the code but also there is a default placeholder text in the login page for Identity :D

One thing I noticed (as we have discussed), when I navigate to `localhost:8080` I'm redirected to `localhost:8080/operate` (which is expected) however I am shown the following error:

```
{"type":"about:blank","title":"Unauthorized","status":401,"detail":"An Authentication object was not found in the SecurityContext","instance":"/"}
```

This means that with this change I cannot use `identity` and the other webapp profiles at the same time.

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
emulbreh: I'd prefer to keep the structure the same for the two conditions, but I don't feel strongly about it. This whole method is pulled apart in the next PR anyway: https://github.com/camunda/camunda/pull/26925

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
emulbreh: I've added this comment so the whole purpose of the method is explained (including this boring base-case). In general, I find there are too few comments in this code base, and maybe that makes these feel out of place. Would it be better as a Javadoc comment?

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
Ben-Sheppard: I see, thank you for the context

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
Ben-Sheppard: I don't think a javadoc would be the right option here (typically javadocs are only used for the public code here i.e. the java client.

Its fine to leave them in to me but yea maybe its because of fewer comments elsewhere that make it a flag for me :D

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
Ben-Sheppard: Ah if we're working with the method in the next PR then lets just ignore this point for now :)

Ben-Sheppard (approved):
I'm going to approve because functionally (the login of Identity) its doing as expected. I have concerns about the webapps not working together with Identity, will this be addressed in a following PR?

emulbreh:
> I'm going to approve because functionally (the login of Identity) its doing as expected. I have concerns about the webapps not working together with Identity, will this be addressed in a following PR?

Yes, eventually. But this will only be fixed when the apps are adapted for the new filter chain. Where do we run the webapps together except for development?

Ben-Sheppard:
> > I'm going to approve because functionally (the login of Identity) its doing as expected. I have concerns about the webapps not working together with Identity, will this be addressed in a following PR?
> 
> Yes, eventually. But this will only be fixed when the apps are adapted for the new filter chain. Where do we run the webapps together except for development?

And helm although they use the `auth-oidc` profile

megglos:
> > > > I'm going to approve because functionally (the login of Identity) its doing as expected. I have concerns about the webapps not working together with Identity, will this be addressed in a following PR?
> > > 
> > > 
> > > Yes, eventually. But this will only be fixed when the apps are adapted for the new filter chain. Where do we run the webapps together except for development?
> > 
> > 
> > And helm although they use the `auth-oidc` profile
> 
> I've changed it so that operate and tasklist are accessible after you log in to identity.

so this PR is unblocked from getting merged then? Let's do it :)

emulbreh:
> > > > > I'm going to approve because functionally (the login of Identity) its doing as expected. I have concerns about the webapps not working together with Identity, will this be addressed in a following PR?
> > > > 
> > > > 
> > > > Yes, eventually. But this will only be fixed when the apps are adapted for the new filter chain. Where do we run the webapps together except for development?
> > > 
> > > 
> > > And helm although they use the `auth-oidc` profile
> > 
> > 
> > I've changed it so that operate and tasklist are accessible after you log in to identity.
> 
> so this PR is unblocked from getting merged then? Let's do it :)

Sure, once the flaky tests let me.

megglos:
> Sure, once the flaky tests let me.

Do we know whether that one may already be addressed on `main`?

emulbreh:
> > Sure, once the flaky tests let me.
> 
> Do we know whether that one may already be addressed on `main`?

I rebased it yesterday, maybe it's been fixed since then. If the current run doesn't pass I'll rebase again.

Issue #26858, opened 2025-01-14
Change id from key to username in `UserDeletedHandler`


Ben-Sheppard:

remcowesterhoud:
Solved in https://github.com/camunda/camunda/pull/26863

PR #26863, opened 2025-01-14
fix: use username instead of key in user export handlers

### LINKED ISSUE #26857: Change id from key to username in `UserCreatedUpdatedHandler`

Ben-Sheppard:
## Description

This PR updates the exporters (both OS/ES and RDBMS) to use the username instead of the key as the ID of the user object during exports.

I've tested this locally and observed the exporters successfully exporting the ID of the record as the username, I was also able to successfully perform updates and deletes.

## Related issues

closes https://github.com/camunda/camunda/issues/26857 https://github.com/camunda/camunda/issues/26858 https://github.com/camunda/camunda/issues/26859

### DIFF
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
@@ -32,12 +32,7 @@ public class UserWriter {
     final boolean wasMerged =
         mergeToQueue(
             user.username(),
-            b ->
-                b.userKey(user.userKey())
-                    .username(user.username())
-                    .name(user.name())
-                    .email(user.email())
-                    .password(user.password()));
+            b -> b.name(user.name()).email(user.email()).password(user.password()));
 
     if (!wasMerged) {
       executionQueue.executeInQueue(
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
@@ -40,10 +40,13 @@ public class UserWriter {
     }
   }
 
-  public void delete(final long userKey) {
+  public void delete(final UserDbModel user) {
     executionQueue.executeInQueue(
         new QueueItem(
-            ContextType.USER, userKey, "io.camunda.db.rdbms.sql.UserMapper.delete", userKey));
+            ContextType.USER,
+            user.userKey(),
+            "io.camunda.db.rdbms.sql.UserMapper.delete",
+            user.username()));
   }
 
   private boolean mergeToQueue(
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
@@ -30,28 +30,28 @@ public class UserWriter {
 
   public void update(final UserDbModel user) {
     final boolean wasMerged =
-        mergeToQueue(
-            user.userKey(), b -> b.username(user.username()).name(user.name()).email(user.email()));
+        mergeToQueue(user.username(), b -> b.name(user.name()).email(user.email()));
 
     if (!wasMerged) {
       executionQueue.executeInQueue(
           new QueueItem(
-              ContextType.USER, user.userKey(), "io.camunda.db.rdbms.sql.UserMapper.update", user));
+              ContextType.USER,
+              user.username(),
+              "io.camunda.db.rdbms.sql.UserMapper.update",
+              user));
     }
   }
 
-  public void delete(final UserDbModel user) {
+  public void delete(final String username) {
     executionQueue.executeInQueue(
         new QueueItem(
-            ContextType.USER,
-            user.userKey(),
-            "io.camunda.db.rdbms.sql.UserMapper.delete",
-            user.username()));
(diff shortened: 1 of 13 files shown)


db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
```diff
@@ -40,10 +40,13 @@ public void update(final UserDbModel user) {
     }
   }
 
-  public void delete(final long userKey) {
+  public void delete(final UserDbModel user) {
```
stoerti, line 43: ❌  Since the userKey is open for deletion later on, we the should only use the username here and not the whole userDbModel object.

stoerti (changes requested):
When the username will be the new primary key of the user, we should use it everywhere in the user context as PK and treat the userKey only as some additional value field. With this I also would:
- change the primary key in the database definitions to the username (db/rdbms-schema)
- change the queueItem.id to the username and use it also in UserWriter#mergeToQueue

But I see, that there is also still an open discussion about these changes. So let's see, what the outcome is there.

Ben-Sheppard:
@stoerti - thank you for looking! As discussed on Slack I've made the changes you suggested (hopefully catching all the places that are required. Would you be able to take another look? :)

koevskinikola (approved):
LGTM @Ben-Sheppard 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserDeleteProcessor.java
```diff
@@ -148,6 +148,8 @@ private void deleteUser(final PersistedUser user) {
     }
 
     stateWriter.appendFollowUpEvent(
-        userKey, UserIntent.DELETED, new UserRecord().setUserKey(userKey));
+        userKey,
+        UserIntent.DELETED,
+        new UserRecord().setUserKey(userKey).setUsername(user.getUsername()));
```
remcowesterhoud, line 153: 🔧 It might make more sense to just return the entire record. It's not much data so I'm not worried about running out of the maximum message size. So it'd be `user.getUser()`.

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
```diff
@@ -30,25 +30,28 @@ public void create(final UserDbModel user) {
 
   public void update(final UserDbModel user) {
     final boolean wasMerged =
-        mergeToQueue(
-            user.userKey(), b -> b.username(user.username()).name(user.name()).email(user.email()));
+        mergeToQueue(user.username(), b -> b.name(user.name()).email(user.email()));
```
remcowesterhoud, line 33: ❓ Is there a reason to not just put all properties in here? So also the user key? Sure, it can't get updated but does the Exporter care about that? I'd leave the exporter to be stupid and just update eveything it receives from Zeebe. That way there won't be a data incosistency between the two.

It seems we also did this before. Here we also added the username in the update query even though we didn't update the username.

remcowesterhoud (approved):
Just some small comments, the rest looks fine to me.

stoerti (approved):

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
Ben-Sheppard: My stance here is to be more defensive, I believe that if I can't update the field then there is no benefit to adding it here, I also believe that if you leave that in here and down the line someone reads this code without prior knowledge the implication is that "the key is in here so the field supports an update".

I'd also argue that the query should actually be changed to only update the fields that can be updated - wdyt @stoerti ?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserDeleteProcessor.java
Ben-Sheppard: Lets do that, I'll update it now

db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/UserWriter.java
remcowesterhoud: > As far as I understood, the userKey will be removed in a future iteration anyway?

We won't fully remove it for a few reasons. For me the main reason would be regarding how distribution works in Zeebe. A key is generated on a specific partition. This partition is encoded within the key. So with this key we know which partition is the owner (source of truth) for this entity.
Currently this will always be partition 1, as this is the source of truth of every entity. The other partitions will follow it and become eventually consistent. A side-effect from this is that partition gets an unfair amount of load compared to the other partitions. If we want to change this in the future we need the keys to know which partition we must forward the requests to.

Anyway long story, but what I'm trying to say is that the keys could be useful in the future. We will keep track of them in Zeebe as well, as keeping track of it prevents annoying migrations once they do become useful.

Ben-Sheppard:
@stoerti / @remcowesterhoud - I've added the fields into the update path and also removed the username from the update query.

remcowesterhoud:
> closes https://github.com/camunda/camunda/issues/26857 https://github.com/camunda/camunda/issues/26858 https://github.com/camunda/camunda/issues/26859

@Ben-Sheppard Fyi you gotta write this as:

closes #26857 
closes #26858
closes #26859

Otherwise GitHub sees it as just a reference to the tickets and doesn't auto-close them 😉 It only closed the first one automatically now. I'll do the rest 🙂

PR #26871, opened 2025-01-14
Create user REST API endpoint returns full user object

### LINKED ISSUE #26780: Respond with full user object on user creation
## Description

In the Java client we only return a user key as a response. We should change this with the full user object. This is in line with REST best practices.

koevskinikola:
## Description

The create user REST API endpoint returns a full user object on user creation.

ℹ️ Some things that may be discussed:
- I didn't include the User's `password` property for security reasons.
- I included the `UserType` enum property as it might be relevant if we decide to expand user types in the future. I didn't see any security risks with adding it.

## Checklist

- [ ] for CI changes:

## Related issues

closes #26780


remcowesterhoud (changes requested):
Thanks @koevskinikola. I should've clarified in detail what we should return, sorry about that. Would you mind remove the user type? I don't think it adds any value and we shold remove it from Zeebe entirely.

koevskinikola:
@remcowesterhoud I removed the `userType` from the REST API response. Have another look when you have the time.

remcowesterhoud (approved):
Thank you 🙂

PR #26877, opened 2025-01-14
fix: Improve user task assignment handling and `changedAttributes` population

### LINKED ISSUE #26614: Unassign Flow with userTask Listeners Does Not Properly Update Task Assignee
The unassign flow is not functioning as expected. Upon unassigning a task, the assigned label temporarily updates correctly but then reverts to an inconsistent state when a delayed message is displayed. The unassign job removes the assignee value but does not set it to null (assignee: "") so the task in neither unassigned nor assigned to specific user so user has to manually press unassign button again to refelect the correct state

**Steps to Reproduce:**
1. Assign a task and wait until it is successfully assigned.
2. Unassign the task and monitor the assigned label closely.
    - Expected Behavior: The label should change from "Assigned to me" to "Unassigned."
    - Observed Behavior: After the delayed message is displayed, the label reverts to "Assigned to me" and then to "Assigned to."
3. Wait until the unassign job is completed and Refresh the page.
    - Observed Behavior: The unassign job removes the assignee value but does not set it to null (assignee: "") so the task in neither unassigned nor assigned to specific user
5. Press the Unassign button again.
    - Expected Behavior: The delayed message should not be displayed again, instead the rejection message if the unassignment job was not completed or "assign to me" button is displayed if job was completed
    - Observed Behavior: The issue repeats, with the delayed message being shown instead of the rejection message.

**Expected Behavior:**
- When a task is unassigned, the assigned label should consistently display "Unassigned" without reverting.
- Upon refreshing the page and attempting to unassign again, the system should either reflect the correct unassigned state or show the appropriate rejection message, not the delayed message.

**Actual Behavior:**
- The assigned label reverts from "Unassigned" to "Assigned to me" after a delayed message is displayed.
- Refreshing the page and unassigning again triggers the delayed message instead of a rejection message.

### Environment

- OS: [e.g. MacOS]
- Browser: [e.g. chrome, safari]
- Tasklist Version: 8.7.0-alpha3-rc4

### Additional context

https://github.com/user-attachments/assets/f8806c8b-bc2f-4e36-98be-3a3d192f73e9
Screenshot 2025-01-08 at 2 31 54 PM

related to: #25419

### LINKED ISSUE #26791: When unassigning, assignee is shown multiple times as a corrected attribute in Operate
**Describe the bug**

When un-assigning a user task, the user task details in Operate show `assignee` multiple times in the `correctedAttributes` property.

I expect that this happens due to `UserTaskHandler#handleAssignment` adding `assignee` to this property:

https://github.com/camunda/camunda/blob/222ffa7aba9bcfec95d034ab9a6e381ce86fa88d/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java#L236-L237

However, the record should already contain all relevant information, so there should be no need to add this property again.

**To Reproduce**

**Expected behavior**

Don't add this data to the record in the handler. Assume that Zeebe provides the record completely and correctly.

**Environment:**
- OS: 
- Zeebe Version: 8.7.0-alpha3
- Configuration:

ce-dmelnych:
## Description
PR description:

This PR addresses the following issues related to user task assignment handling and ensures consistent updates to the `changedAttributes` property:

- **Fixed Issue 1**: Correcting the `assignee` to an empty string (`""`) now properly clears the `assignee` in Tasklist instead of assigning an empty string.
- **Fixed Issue 2**: Resolved the issue where the `assignee` was shown multiple times as a corrected attribute in Operate after unassigning a user task.

### Key Changes:
1. Updated Zeebe code to include the `assignee` as an entry in the `changedAttributes` property after performing **assign**, **unassign**, or **claim** operations. This behavior is consistent regardless of whether user task listeners are configured or not.
2. Modified `camunda-exporter/UserTaskHandler.java` to rely on `record.getValue().getChangedAttributes()` when processing the `ASSIGNED` event for user tasks, and properly set the value for `entity.assignee` property as `null` instead of empty string `""`. This ensures downstream applications like Operate and Tasklist receive consistent and accurate data.

## Checklist

- [ ] for CI changes:

## Related issues

closes #26614
closes #26791

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
@@ -330,6 +330,7 @@ public final class BpmnUserTaskBehavior {
   public void userTaskAssigning(final UserTaskRecord userTaskRecord, final String assignee) {
     final long userTaskKey = userTaskRecord.getUserTaskKey();
     userTaskRecord.setAssignee(assignee);
+    userTaskRecord.setAssigneeChanged();
     stateWriter.appendFollowUpEvent(userTaskKey, UserTaskIntent.ASSIGNING, userTaskRecord);
   }
 
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
@@ -329,8 +329,10 @@ public final class BpmnUserTaskBehavior {
 
   public void userTaskAssigning(final UserTaskRecord userTaskRecord, final String assignee) {
     final long userTaskKey = userTaskRecord.getUserTaskKey();
-    userTaskRecord.setAssignee(assignee);
-    userTaskRecord.setAssigneeChanged();
+    if (!userTaskRecord.getAssignee().equals(assignee)) {
+      userTaskRecord.setAssignee(assignee);
+      userTaskRecord.setAssigneeChanged();
+    }
     stateWriter.appendFollowUpEvent(userTaskKey, UserTaskIntent.ASSIGNING, userTaskRecord);
   }
 
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
@@ -55,6 +55,7 @@ public final class UserTaskAssignProcessor implements UserTaskCommandProcessor {
     final long userTaskKey = command.getKey();
 
     userTaskRecord.setAssignee(command.getValue().getAssignee());
+    userTaskRecord.setAssigneeChanged();
     userTaskRecord.setAction(command.getValue().getActionOrDefault(DEFAULT_ACTION));
 
     stateWriter.appendFollowUpEvent(userTaskKey, UserTaskIntent.ASSIGNING, userTaskRecord);
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
@@ -54,8 +54,11 @@ public final class UserTaskAssignProcessor implements UserTaskCommandProcessor {
       final TypedRecord<UserTaskRecord> command, final UserTaskRecord userTaskRecord) {
     final long userTaskKey = command.getKey();
 
-    userTaskRecord.setAssignee(command.getValue().getAssignee());
-    userTaskRecord.setAssigneeChanged();
+    final var newAssignee = command.getValue().getAssignee();
+    if (!userTaskRecord.getAssignee().equals(newAssignee)) {
+      userTaskRecord.setAssignee(newAssignee);
+      userTaskRecord.setAssigneeChanged();
+    }
     userTaskRecord.setAction(command.getValue().getActionOrDefault(DEFAULT_ACTION));
 
     stateWriter.appendFollowUpEvent(userTaskKey, UserTaskIntent.ASSIGNING, userTaskRecord);
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
@@ -121,7 +121,8 @@ public class TaskListenerTest {
         userTask ->
             Assertions.assertThat(userTask)
                 .hasAction("my_custom_action")
-                .hasVariables(Map.of("foo_var", "bar")));
+                .hasVariables(Map.of("foo_var", "bar"))
+                .hasNoChangedAttributes());
     assertThatProcessInstanceCompleted(processInstanceKey);
   }
 
@@ -167,7 +168,10 @@ public class TaskListenerTest {
         processInstanceKey,
         UserTaskIntent.ASSIGNED,
         userTask ->
-            Assertions.assertThat(userTask).hasAssignee("me").hasAction("my_assign_action"));
+            Assertions.assertThat(userTask)
+                .hasAssignee("me")
+                .hasAction("my_assign_action")
+                .hasOnlyChangedAttributes(UserTaskRecord.ASSIGNEE));
   }
 
   @Test
@@ -388,7 +392,10 @@ public class TaskListenerTest {
         processInstanceKey,
         UserTaskIntent.ASSIGNED,
         userTask ->
-            Assertions.assertThat(userTask).hasAssignee("test_user").hasAction("claim_action"));
+            Assertions.assertThat(userTask)
+                .hasAssignee("test_user")
+                .hasAction("claim_action")
+                .hasOnlyChangedAttributes(UserTaskRecord.ASSIGNEE));
   }
 
   @Test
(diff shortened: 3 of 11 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/usertask/AssignUserTaskTest.java
```diff
@@ -84,6 +85,7 @@ public void shouldEmitAssignEventsForUserTask() {
                     .hasUserTaskKey(userTaskKey)
                     .hasAction(DEFAULT_ACTION)
                     .hasAssignee("foo")
+                    .hasOnlyChangedAttributes(UserTaskRecord.ASSIGNEE)
```
korthout, line 88: 😍

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
```diff
@@ -233,6 +233,14 @@ private boolean refersToPreviousVersionRecord(final long key) {
     return key < exporterMetadata.getFirstUserTaskKey(TaskImplementation.ZEEBE_USER_TASK);
   }
 
+  private void setAssignee(final Record<UserTaskRecordValue> record, final TaskEntity entity) {
+    if (ExporterUtil.isEmpty(record.getValue().getAssignee())) {
+      entity.setAssignee(null);
+    } else {
+      entity.setAssignee(record.getValue().getAssignee());
+    }
+  }
```
korthout: 🔧 This method can be static
🔧 To increase clarity, please consider swapping the parameters. Having the entity that will be changed first is common practice in Java APIs (although not documented for our codebase)

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
```diff
@@ -373,17 +380,20 @@ public void shouldTriggerAssignmentListenersAfterUserTaskCreationWithDefinedAssi
...
-            tuple(UserTaskIntent.COMPLETE_TASK_LISTENER, assignee, action),
-            tuple(UserTaskIntent.COMPLETE_TASK_LISTENER, assignee, action),
-            tuple(UserTaskIntent.ASSIGNED, assignee, action));
+            tuple(UserTaskIntent.CREATING, StringUtils.EMPTY, action, List.of()),
+            tuple(UserTaskIntent.CREATED, StringUtils.EMPTY, action, List.of()),
+            tuple(UserTaskIntent.ASSIGNING, assignee, action, List.of(UserTaskRecord.ASSIGNEE)),
+            tuple(UserTaskIntent.COMPLETE_TASK_LISTENER, assignee, action, List.of()),
+            tuple(UserTaskIntent.COMPLETE_TASK_LISTENER, assignee, action, List.of()),
+            tuple(UserTaskIntent.COMPLETE_TASK_LISTENER, assignee, action, List.of()),
+            tuple(UserTaskIntent.ASSIGNED, assignee, action, List.of(UserTaskRecord.ASSIGNEE)));
```
korthout, line 396: 🔧 Would be interesting to see this same assertion for a case with two additional corrected properties

❌ I've quickly tested this and noticed that the corrected attributes don't appear in the changed attributes of the `ASSIGNED` event

korthout (changes requested):
Nice work @ce-dmelnych 👏 

❌ After some testing, I feel like we're not there yet completely.

- Corrected attributes don't appear in the changed attributes anymore, e.g. correct priority on assigning should have both in changed
- We should only add values to changed attributes, if they actually differ with the previous state. I.e. don't add assignee as changed when assigning to the same assignee a second time.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
```diff
@@ -233,15 +233,12 @@ private boolean refersToPreviousVersionRecord(final long key) {
     return key < exporterMetadata.getFirstUserTaskKey(TaskImplementation.ZEEBE_USER_TASK);
   }
 
-  private void handleAssignment(final Record<UserTaskRecordValue> record, final TaskEntity entity) {
-    entity.getChangedAttributes().add("assignee");
+  private void setAssignee(final Record<UserTaskRecordValue> record, final TaskEntity entity) {
     if (ExporterUtil.isEmpty(record.getValue().getAssignee())) {
       entity.setAssignee(null);
```
tsedekey, line 238: I was testing your code locally, and it seems that setting the assignee property to null is not being persisted for some reason. The get task service still returns the assignee property as an empty string. I wasn’t able to verify this on ES because I’m not very familiar with ES yet.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
ce-dmelnych: Thank you for the suggestion!
I’ve tested this scenario in the `TaskListenerTest#shouldTrackOnlyChangedAttributesDuringUserTaskAssignmentWithMultipleListeners` test case.

This test specifically verifies how multiple listeners handle corrections to user task properties and ensures that only the truly changed attributes are tracked in the `ASSIGNED` event.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/usertask/AssignUserTaskTest.java
```diff
@@ -70,10 +74,17 @@ public void shouldEmitAssigningEventForUserTask() {
...
     final var assigningRecordValue = assigningRecord.getValue();
-    Assertions.assertThat(assigningRecordValue)
-        .hasUserTaskKey(userTaskKey)
-        .hasAction("assign")
-        .hasTenantId(TenantOwned.DEFAULT_TENANT_IDENTIFIER);
+    final var assignedRecordValue =
+        RecordingExporter.userTaskRecords(UserTaskIntent.ASSIGNED).getFirst().getValue();
+
+    assertThat(List.of(assigningRecordValue, assignedRecordValue))
+        .allSatisfy(
```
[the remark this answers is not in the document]
ce-dmelnych: I've added a `describedAs` message to clarify that the data consistency between `ASSIGNING`/`CLAIMING` and `ASSIGNED` records is essential, as dependent applications like Tasklist and Operate will rely on it to update user task data internally.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
ce-dmelnych: Thank you, Nico, for the feedback!  

We addressed this together by transforming:  
```java
private setAssignee(final Record<UserTaskRecordValue> record, final TaskEntity entity)
```  
into:  
```java
private static String getAssigneeOrNull(final Record<UserTaskRecordValue> record)
```  

This refactor enabled us to reuse the `getAssigneeOrNull` method in both the `updateChangedAttributes` and `createTaskEntity` methods.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskHandler.java
ce-dmelnych: Thank you for sharing your feedback, Eddie!

Could you share how you tested the new code from the branch?
Did you use the `Makefile` from the `tasklist` mvn module, or did you follow a different approach?

I'd like to understand your setup so I can try to reproduce the issue as well.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/usertask/processors/UserTaskAssignProcessor.java
```diff
@@ -54,7 +54,11 @@ public void onCommand(
       final TypedRecord<UserTaskRecord> command, final UserTaskRecord userTaskRecord) {
     final long userTaskKey = command.getKey();
 
-    userTaskRecord.setAssignee(command.getValue().getAssignee());
+    final var newAssignee = command.getValue().getAssignee();
+    if (!userTaskRecord.getAssignee().equals(newAssignee)) {
+      userTaskRecord.setAssignee(newAssignee);
+      userTaskRecord.setAssigneeChanged();
+    }
```
ce-dmelnych, line 61: @korthout 

As we are introducing this check to compare the new `assignee` with the persisted user task `assignee` in both `UserTaskAssignProcessor` and `UserTaskClaimProcessor`. This ensures that the `changedAttributes` and `assignee` will be updated only when `assignee` was actually changed.

With this in place, it seems like there’s no need to add an additional check like `modifiedRecord.setDiffAsChangedAttributes(persistedRecord);` in the `UserTaskProcessor#handleCommandProcessing` when no task listeners are defined before finalizing command.

I left the `UserTaskProcessor` class unchanged for now, but we might revisit this approach during the implementation of `updating` listeners.

What do you think about this?

korthout: Should be fine for now I think

ce-dmelnych:
Hey @panagiotisgts,

It seems @tsedekey is on a long vacation 👶🏻🍼. Could you, on behalf of the HTO team, review this PR?
This is especially relevant as we modified the [UserTaskHandler](https://github.com/camunda/camunda/pull/26877/files#diff-9b1d49038165041139dbd9c3172f71f3702b14380388f23731fded99dac0a8e6) class in the `exporter` module.  

The key change is that with the `ASSIGNED` user task record, Zeebe now includes the "assignee" entry in the `changedAttributes` list when the `assignee` property is modified. This allows the `UserTaskHandler` to rely solely on `record.getValue().getChangedAttributes()` to apply updates to the `entity` based on the changed attributes.  

Additionally, we've fixed an issue where the `entity.assignee` property was incorrectly assigned an empty string `""` after handling user task assignment, instead of being set to `null`.  

Your review would be greatly appreciated!

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/listeners/task/TaskListenerTest.java
```diff
@@ -1363,6 +1363,116 @@ private void testProvideCorrectedUserTaskDataToSubsequentTaskListener(
...
+            tuple(UserTaskIntent.CORRECTED, List.of("assignee", "priority")),
+            tuple(UserTaskIntent.COMPLETE_TASK_LISTENER, List.of("candidateGroupsList")),
+            tuple(UserTaskIntent.CORRECTED, List.of("candidateGroupsList")),
+            // As a result, `changedAttributes` contains only `dueDate` and `candidateGroupsList`
+            // because these are the only properties that differ from the original values of the
+            // created user task. All other properties were either unchanged or reset to their
+            // initial values that were before task assignment using the corrections.
+            tuple(UserTaskIntent.ASSIGNED, List.of("candidateGroupsList", "dueDate")));
+
+    final var createUserTaskValue =
```
korthout, line 1451:
```suggestion
    final var createdUserTaskValue =
```

korthout (approved):
Thanks @ce-dmelnych 👏 

LGMT 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
```diff
@@ -330,6 +330,7 @@ public void userTaskCreated(final UserTaskRecord userTaskRecord) {
   public void userTaskAssigning(final UserTaskRecord userTaskRecord, final String assignee) {
     final long userTaskKey = userTaskRecord.getUserTaskKey();
     userTaskRecord.setAssignee(assignee);
+    userTaskRecord.setAssigneeChanged();
```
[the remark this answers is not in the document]
ce-dmelnych: Currently, we invoke the `BpmnUserTaskBehavior#userTaskAssigning` method only when the user task is configured with a defined (non-empty string) assignee. 

However, to make it safer and consistent with the checks we already have in the `UserTaskAssignProcessor` and `UserTaskClaimProcessor` classes, I've added a similar condition here. This improvement ensures better alignment and might also be helpful if we decide to reuse this method in `UserTaskAssignProcessor` later.

PR #26904, opened 2025-01-14
Add inclusive gateway migration validations

### LINKED ISSUE #25675: Add inclusive gateway migration validations
## Description
After implementing #25674, we need validations to make sure inclusive gateway migration executes correctly. Following validations are needed:

- Reject inclusive gateway migration if joining inclusive gateway with at least one sequence flow taken is not mapped
- Reject inclusive gateway migration if joining inclusive gateway with at least one sequence flow taken is mapped to a different gateway type. E.g. mapped to an parallel gateway
- Reject inclusive gateway migration if all incoming sequence flows of the gateway are taken (concurrent command). It is because `ACTIVATE_ELEMENT` command for the joining gateway is written to the log but not processed before the `MIGRATE` command. If we do not reject this, it will cause wrong (with old process definition data) gateway record written to the log.
- Reject inclusive gateway migration if the target gateway has less incoming sequence flows than the source gateway.
- Reject inclusive gateway migration if a taken sequence flow does not exist in target definition

The last rejection case might need more clarification. The last case will be explained in following sections.

## How Inclusive Gateway Works in Zeebe
For the joining inclusive gateway, activation conditions are different from parallel gateway due to its execution logic. A joining inclusive gateway should be activated/completed when all paths that satisfy the condition are taken. One of the following conditions (in order) should be satisfied for joining inclusive gateway to be activated/completed:
1) If the number of incoming sequence flows taken is equal to the total number of incoming sequence flows (same as parallel gateway)
2) No active element in the flow scope has path to the inclusive gateway
3) None of the target elements of active sequence flows has path to inclusive gateway

Let’s consider the following example:

Screenshot 2024-12-03 at 11 56 26

Assume that only “Pasta” and “Steak” satisfies the condition.
1) When the task “Cook Pasta” is completed, the outgoing sequence flow (1) of it will be marked as taken.
2) The active element “Stir-fry Steak” still has a path to the joining inclusive gateway, so we will not activate the joining inclusive gateway.
3) Later, when the task “Stir-fry Steak” is completed and the outgoing sequence flow (2) is taken.
4) The joining inclusive gateway (join) will be activated/completed since there is no path from an active element to the joining inclusive gateway.

There is one important missing piece of information above; in step 2 and 3 there is one more step applied:
The sequence flow id `1` will be stored in the state. (details of why we store it is explained in the [internal discussion doc](https://docs.google.com/document/d/1iGoCVgHY1ZGK9HR0I8GwDCZan-S0hfvReSbxYinBAFo/edit?tab=t.0#heading=h.12sho1jivrb8))

### The issue with keeping track of sequence flow ids
We store ids of the sequence flows taken but not activate an element. In the example above, we store sequence flow id `1` in the state until the joining inclusive gateway is activated/completed. If the sequence flow id `1` no longer exists in the target process definition after migration, a `NullPointerException` will be thrown while validating a path from a sequence flow to the joining inclusive gateway.

Because of the reason explained above, we need to make sure taken sequence flow ids to joining gateways stay the same between process instance and the target process definition.

berkaycanbc:
## Description

**Note: Only test cases are added in this PR since all validations implemented in #26109.**

Validations are added for the following cases:
- Reject inclusive gateway migration if joining inclusive gateway with at least one sequence flow taken is not mapped
- Reject inclusive gateway migration if joining inclusive gateway with at least one sequence flow taken is mapped to a different gateway type. E.g. mapped to an parallel gateway
- Reject inclusive gateway migration if all incoming sequence flows of the gateway are taken (concurrent command). It is because `ACTIVATE_ELEMENT` command for the joining gateway is written to the log but not processed before the `MIGRATE` command. If we do not reject this, it will cause wrong (with old process definition data) gateway record written to the log.
- Reject inclusive gateway migration if the target gateway has less incoming sequence flows than the source gateway.
- Reject inclusive gateway migration if a taken sequence flow does not exist in target definition

Additionally, test cases are added to verify the the existing validations works for the following:
- Reject inclusive gateway migration if mapped gateway does not exist in source
- Reject inclusive gateway migration if mapped gateway does not exist in target

## Related issues

closes #25675

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateInclusiveGatewayTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateInclusiveGatewayTest.java
@@ -740,4 +740,83 @@ public class MigrateInclusiveGatewayTest {
                 .formatted(processInstanceKey, nonExistingTargetGatewayId))
         .hasKey(processInstanceKey);
   }
+
+  @Test
+  public void
+      shouldRejectMigrationForInclusiveJoiningGatewayIfTakenSequenceFlowIdIsChangedInTarget() {
+    final String sourceProcessId = helper.getBpmnProcessId();
+    final String targetProcessId = helper.getBpmnProcessId() + "_v2";
+
+    final var deployment =
+        ENGINE
+            .deployment()
+            .withXmlResource(
+                Bpmn.createExecutableProcess(sourceProcessId)
+                    .startEvent()
+                    .inclusiveGateway("fork")
+                    .conditionExpression("= true")
+                    .serviceTask("task1", b -> b.zeebeJobType("type1"))
+                    .sequenceFlowId("flow1")
+                    .inclusiveGateway("join1")
+                    .endEvent()
+                    .moveToNode("fork")
+                    .conditionExpression("= true")
+                    .serviceTask("task2", b -> b.zeebeJobType("type2"))
+                    .connectTo("join1")
+                    .done())
+            .withXmlResource(
+                Bpmn.createExecutableProcess(targetProcessId)
+                    .startEvent()
+                    .inclusiveGateway("fork")
+                    .conditionExpression("= true")
+                    .serviceTask("task1", b -> b.zeebeJobType("type1"))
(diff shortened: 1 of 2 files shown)


korthout (approved):
LGTM 💯 

Thanks @berkaycanbc 🥇

PR #26914, opened 2025-01-15
fix: ignore MIGRATED variable records

### LINKED ISSUE #26490: Variable can't be found in UI after process migration

Operate can't find variable after process migration.

* Backport to all support versions that have the problem (8.7 to 8.4; 8.3 doesn't have migration)

### Steps to reproduce (required on creation)

1. Start Operate with demo data creation
2. Select one process instance (eg Order process) and note the variable name and value (eg orderNo)
3. Migrate this process instance (eg Call Activity)
4. Search for process instances by filtering by variable name and value 

### Current behavior (required on creation)

The process instance is not found

### Expected behavior (required on creation)

The process instance is found, even after migration

### Rootcause (required on prioritization)

Operate processes variable data from Zeebe two times. One for `variable` index and one for `list-view` (search) index. 
* For variables it [checks for intent](https://github.com/camunda/camunda/blob/8e78d28bdda7eccbeba7fa7efc157a9af525dd44/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/VariableZeebeRecordProcessor.java#L82-L82) `MIGRATED` and updates then only `processDefinitionKey` and `bpmnProcessId`.

* For `list-view` it [doesn’t check](https://github.com/camunda/camunda/blob/8e78d28bdda7eccbeba7fa7efc157a9af525dd44/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/processors/ListViewZeebeRecordProcessor.java#L221-L221) for intent, so it overwrites with the “new” value. 

The Zeebe data for variables with intent `MIGRATED` has a variable value of `null`. Therefore
Operate writes a `null` in variable value in `list-view` index.
This can be confirmed by searching for variable value of `null` after process migrating.

### Solution ideas

Handle intent in `ListViewZeebeRecordProcessor` similar as in `VariableZeebeRecordProcessor`.
For example skip writing variable value in case of intent `MIGRATED`
```java
  ...
        final var initialIntent = cachedVariable.getLeft();
        if (initialIntent == VariableIntent.MIGRATED) {
          LOGGER.info("Migrated variable record found, skipping processing");
          continue;
        }
        final var value = cachedVariable.getRight().getVarValue();
  ...
```

Any references that the variable document has to the process definition should still be updated to the new version (e.g. process definition key). Needs assessment if and to what degree that is required.

### Additional context

* [Support 25085](https://jira.camunda.com/browse/SUPPORT-25085)
* [Support 25238](https://jira.camunda.com/browse/SUPPORT-25238)

### Handover Dev to QA (required before manual testing)

- Resources: Any BPMN model
  
- Versions to validate: Run `make env-up` from the `operate` directory on the `stable/8.6` branch: https://github.com/camunda/camunda/tree/stable/8.6/operate (otherwise we don't have snapshot deployments yet)
- Release version (in which version this feature will be released): see version labels for the patches; 8.7.0-alpha5, 8.8.0-alpha1
  

### Links

[Support 25085](https://jira.camunda.com/browse/SUPPORT-25085)

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/26914
- [ ] https://github.com/camunda/camunda/pull/27482
- [ ] https://github.com/camunda/camunda/pull/27094
- [ ] https://github.com/camunda/camunda/pull/27093
- [ ] https://github.com/camunda/operate/pull/6844
```

sdorokhova:
## Description

1. Variable MIGRATED Zeebe record does not provide the value, therefore we need to fully skip it when persisting in `list-view` index. More context: values store in `list-view` index are used to perform filtering by variable name and value in Processes tab.
2. I additionally reworked how the new variables are persisted (basically I brought back the code that was there[ before introducing and removing again of concurrency mode](https://github.com/camunda/camunda/commit/992afbf741f25b2d04d2ac1dc0254516a32b8add#diff-c871546cf084ee89e36386017b5ee7350014b13f14df4f466cbf867ba35ee449)): when variable is created in this batch, we execute add query and only if it was updated - upsert query. This should improve performance.

## Related issues

closes #26490

### DIFF
--- a/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/v8_6/processors/ListViewZeebeRecordProcessor.java
+++ b/operate/importer-8_6/src/main/java/io/camunda/operate/zeebeimport/v8_6/processors/ListViewZeebeRecordProcessor.java
@@ -44,6 +44,7 @@ import io.camunda.zeebe.protocol.record.Record;
 import io.camunda.zeebe.protocol.record.intent.IncidentIntent;
 import io.camunda.zeebe.protocol.record.intent.Intent;
 import io.camunda.zeebe.protocol.record.intent.JobIntent;
+import io.camunda.zeebe.protocol.record.intent.VariableIntent;
 import io.camunda.zeebe.protocol.record.value.*;
 import java.time.Instant;
 import java.time.OffsetDateTime;
@@ -166,15 +167,15 @@ public class ListViewZeebeRecordProcessor {
       final var scopedVariables = variableRecords.getValue();
 
       for (final var scopedVariable : scopedVariables) {
+        if (!shouldProcessVariableRecord(scopedVariable)) {
+          continue;
+        }
         final var intent = scopedVariable.getIntent();
         final var variableValue = scopedVariable.getValue();
         final var variableName = variableValue.getName();
         final var cachedVariable =
             temporaryVariableCache.computeIfAbsent(
-                variableName,
-                (k) -> {
-                  return Tuple.of(intent, new VariableForListViewEntity());
-                });
+                variableName, (k) -> Tuple.of(intent, new VariableForListViewEntity()));
         final var variableEntity = cachedVariable.getRight();
         processVariableRecord(scopedVariable, variableEntity);
       }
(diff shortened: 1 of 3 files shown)


ralfpuchert (approved):
Thank you!

Issue #26804, opened 2025-01-13
Support for Basic authentication on Webapp Login, REST API & GRPC API Access


megglos:

emulbreh:
Initial user configuration was implemented in https://github.com/camunda/camunda/pull/25762 and still works for me in main.

PR #26925, opened 2025-01-15
feat: add configuration for disabling authentication for the API

### LINKED ISSUE #26819: Allow unauthenticated REST API and GRPC requests
While with https://github.com/camunda/camunda/issues/26804 we enable the protection of the Webapps, the REST API and the GRPC API by basic authentication, this issue is about allowing to expose the REST and GRPC API without any authentication, even if basic auth is enabled for the webapplications.

Acceptance Criteria:

- [x] As an admin I can configure the REST and GRPC API to not have any authentication protection - even if basic authentication is enabled for the web applications

emulbreh:
We introduce a new boolean settings, `camunda.security.authentication.basic.allowUnauthenticatedApiAccess`, that toggles whether API requests require authentication. The default is false. This setting will be enabled in the default configuration.

closes #26819
relates to #26804

### DIFF
--- a/authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
+++ b/authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
@@ -7,11 +7,17 @@
  */
 package io.camunda.authentication.config;
 
+import static org.springframework.security.web.util.matcher.AntPathRequestMatcher.antMatcher;
+
+import com.google.common.collect.Sets;
 import io.camunda.authentication.CamundaUserDetailsService;
 import io.camunda.authentication.filters.TenantRequestAttributeFilter;
 import io.camunda.authentication.handler.AuthFailureHandler;
 import io.camunda.authentication.handler.CustomMethodSecurityExpressionHandler;
+import io.camunda.security.configuration.AuthenticationConfiguration;
+import io.camunda.security.configuration.BasicAuthenticationConfiguration;
 import io.camunda.security.configuration.MultiTenancyConfiguration;
+import io.camunda.security.configuration.SecurityConfiguration;
 import io.camunda.service.AuthorizationServices;
 import io.camunda.service.RoleServices;
 import io.camunda.service.TenantServices;
@@ -19,6 +25,7 @@ import io.camunda.service.UserServices;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import java.util.Arrays;
+import java.util.Optional;
 import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -48,8 +55,13 @@ import org.springframework.security.web.util.matcher.RequestMatcher;
 @Profile("auth-basic|auth-oidc")
 public class WebSecurityConfig {
   public static final String SESSION_COOKIE = "camunda-session";
+
+  private static final Logger LOG = LoggerFactory.getLogger(WebSecurityConfig.class);
+
   private static final String LOGIN_URL = "/login";
   private static final String LOGOUT_URL = "/logout";
+
+  private static final Set<String> API_PATHS = Set.of("/v1/**", "/v2/**", "/api/**");
   private static final Set<String> UNAUTHENTICATED_PATHS =
       Set.of(
           LOGIN_URL,
@@ -71,8 +83,6 @@ public class WebSecurityConfig {
           // deprecated Tasklist v1 Public Endpoints
           "/v1/external/process/**",
           "/new/**");
-
-  private static final Logger LOG = LoggerFactory.getLogger(WebSecurityConfig.class);
 
   @Bean
   @ConditionalOnMissingBean(MethodSecurityExpressionHandler.class)
@@ -100,7 +110,7 @@ public class WebSecurityConfig {
       final AuthFailureHandler authFailureHandler,
       final ClientRegistrationRepository clientRegistrationRepository)
       throws Exception {
-    return baseHttpSecurity(httpSecurity, authFailureHandler)
+    return baseHttpSecurity(httpSecurity, authFailureHandler, UNAUTHENTICATED_PATHS)
         .oauth2ResourceServer(
             oauth2 ->
                 oauth2.jwt(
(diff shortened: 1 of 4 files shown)


megglos:
@emulbreh for better context please link the relevant github issues to the PRs. I appears like covering partially 
https://github.com/camunda/camunda/issues/26804 & fully https://github.com/camunda/camunda/issues/26819 ?

security/security-core/src/main/java/io/camunda/security/configuration/BasicAuthenticationConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.security.configuration;
+
+public class BasicAuthenticationConfiguration {
+  private boolean apiDangerouslyUnprotected = false;
```
megglos, line 11: 🔧  The name made my day 😄  still as this is operator/admin facing I would be in for a simpler less fancy name. Wdyt of `camunda.security.authentication.basic.api.allow-unauthenticated=false`

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
```diff
@@ -130,31 +136,66 @@ private void genericSuccessHandler(
...
+      final HttpSecurity httpSecurity,
+      final AuthFailureHandler authFailureHandler,
+      final SecurityConfiguration securityConfiguration)
       throws Exception {
-    return buildLoginAuthSecurityFilterChain(httpSecurity, authFailureHandler);
+    if (!securityConfiguration.getAuthentication().getBasic().isApiDangerouslyUnprotected()) {
+      return null;
+    }
+    LOG.warn(
+        "The API is accessible without authentication. Remember to disable camunda.security.authentication.basic.api-dangerously-unprotected");
```
megglos, line 182: 👍  Nice that you make that transparent in the log! Just one additional suggestion
```suggestion
    LOG.warn(
        "The API is accessible without authentication. Remember to disable camunda.security.authentication.basic.api-dangerously-unprotected for a production server deployment.");
```

security/security-core/src/main/java/io/camunda/security/configuration/BasicAuthenticationConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+  private boolean httpBasicAuthEnabled = true;
```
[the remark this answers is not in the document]
emulbreh: You can use method=basic without allowing authentication via an `Authorization: Basic` header. This can be relevant for a self-managed setup. We can remove it if you believe it's unnecessary, but I think it's actually something you'd want to do if you choose to run basic auth in production.

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
emulbreh: It also should be disabled for a staging env, it's really only appropriate for local development. Maybe "for a non-local setup"?

emulbreh: It's now 
> Please disable camunda.security.authentication.basic.api-dangerously-unprotected for any deployment.

security/security-core/src/main/java/io/camunda/security/configuration/BasicAuthenticationConfiguration.java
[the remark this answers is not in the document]
megglos: I'm not aware of us planning to support that right now, thus I would not introduce it.

I will follow-up with PM if this is needed, then we can handle it separtely.

security/security-core/src/main/java/io/camunda/security/configuration/BasicAuthenticationConfiguration.java
megglos: Just aligned with PM, we would prefer to not replicate the react approach/naming and go with `camunda.security.authentication.basic.api.allow-unauthenticated=false` or something similar without warnings in the config name. We can use those words in the docs.

I get the intention but as the default will be false + the app logs a warning and the config name cleary indicates that unauthenticated requests are allowed I see no reason strong for including `dangerous` in the config name.

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
```diff
@@ -134,31 +141,61 @@ private void genericSuccessHandler(
...
+      final HttpSecurity httpSecurity,
+      final AuthFailureHandler authFailureHandler,
+      final SecurityConfiguration securityConfiguration)
       throws Exception {
-    return buildLoginAuthSecurityFilterChain(httpSecurity, authFailureHandler);
+    if (!securityConfiguration.getAuthentication().getBasic().getAllowUnauthenticatedApiAccess()) {
+      return null;
+    }
+    LOG.warn(
+        "The API is accessible without authentication. Please disable camunda.security.authentication.basic.allow-unauthenticated-api-access for any deployment.");
```
megglos, line 182: ❌ 
```suggestion
        "The API is accessible without authentication. Please disable camunda.security.authentication.basic.allowUnauthenticatedApiAccess for any deployment.");
```

megglos: 💭  although I consider it unlikely to change, do you see any good way to obtain the config name based off the config class dynamically?

megglos (approved):
Taking over review from @Ben-Sheppard as announced by him [here](https://camunda.slack.com/archives/C06UYJMMETZ/p1737106818535819?thread_ts=1737100824.757539&cid=C06UYJMMETZ).

LGTM! Just one thing that needs to be adjusted before merge + failing ITs due to the NPE
```
io.camunda.security.configuration.BasicAuthenticationConfiguration.getAllowUnauthenticatedApiAccess()" because the return value of "io.camunda.security.configuration.AuthenticationConfiguration.getBasic()" is null
```

Pre-approving, unless you see a need for another review.

authentication/src/main/java/io/camunda/authentication/config/WebSecurityConfig.java
emulbreh: I don't know how to do that, this is all Spring magic. I've extracted it as a constant in the next PR though.

PR #26936, opened 2025-01-15
Update user by username

### LINKED ISSUE #26784: Update users by username instead of key
## Description

- Modify OpenAPI specification
- Change the controller and service to work with username
- Modify the `UserUpdateProcessor` to lookup the user by username instead of key

koevskinikola:
## Description

Refator the "user update" feature to locate users by username instead of user key, for updating. 

## Checklist

- [ ] for CI changes:

## Related issues

closes #26784

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -3613,6 +3613,45 @@ paths:
                 $ref: "#/components/schemas/ProblemDetail"
         "500":
           $ref: "#/components/responses/InternalServerError"
+    patch:
+      tags:
+        - User
+      operationId: updateUser
+      summary: Update user
+      description: |
+        Updates a user.
+      parameters:
+        - name: username
+          in: path
+          required: true
+          description: The username of the user to update.
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UserUpdateRequest"
+        required: true
+      responses:
+        "204":
+          description: The user was updated successfully.
+        "400":
+          description: The provided data is not valid.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "404":
+          description: The user was not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "500":
+          $ref: "#/components/responses/InternalServerError"
 
   /incidents/search:
     post:
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
@@ -189,13 +189,13 @@ public class RequestMapper {
                 getStringOrEmpty(updateRequest, UserTaskUpdateRequest::getAction)));
   }
 
-  public static Either<ProblemDetail, UpdateUserRequest> toUserUpdateRequest(
+  public static Either<ProblemDetail, UserDTO> toUserUpdateRequest(
       final UserUpdateRequest updateRequest, final String username) {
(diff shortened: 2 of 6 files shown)


koevskinikola:
@remcowesterhoud this one is ready for review as well.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -190,13 +190,13 @@ public static Either<ProblemDetail, UpdateUserTaskRequest> toUserTaskUpdateReque
...
   public static Either<ProblemDetail, UpdateUserRequest> toUserUpdateRequest(
-      final UserUpdateRequest updateRequest, final long userKey) {
+      final UserUpdateRequest updateRequest, final String username) {
     final UserChangeset changeset = updateRequest.getChangeset();
     return getResult(
         validateUserUpdateRequest(updateRequest),
         () ->
             new UpdateUserRequest(
-                userKey, changeset.getName(), changeset.getEmail(), changeset.getPassword()));
+                username, changeset.getName(), changeset.getEmail(), changeset.getPassword()));
```
remcowesterhoud, line 199: 🤔 Is this legacy code? I don't quite understand why we don't return a `UserDTO` here directly, similar to what we do with user creation. Do you know?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -3613,6 +3613,45 @@ paths:
...
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UserUpdateRequest"
+        required: true
+      responses:
+        "204":
```
remcowesterhoud, line 3637: ❓ WDYT about returning the user object as a response here as well?

remcowesterhoud (approved):

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
koevskinikola: 👍 We should, it's defined in the [internal definition doc](https://docs.google.com/document/d/1rS989EESojWRrCMLH7Nscy1UppbgOanvzVeOljCc_iA/edit?pli=1&tab=t.0#heading=h.hxbk2uscj4u0). Other update endpoints follow this pattern as well.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
koevskinikola: You mean that we should return a `UserDTO` instance instead of `UpdateUserRequest`?

koevskinikola: Yeah, I refactored it. We don't need the `UpdateUserRequest` step.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
koevskinikola: @remcowesterhoud I had another look, and most "update" endpoints that we currently have don't return a response. Should we align with the REST API team before we make this change?

remcowesterhoud: Yes please check in with them. I'm fine with not doing it if that's the result. It's something we can always add later on.

koevskinikola: Results from the REST API team discussion [here](https://camunda.slack.com/archives/C06UKS51QV9/p1736953822611169) (internal link). The general feeling is that our idea hasn't been explored thoroughly, so if we make this change:
- we would break UX consistency of the REST API (update) endpoints.
- we would make the eventual consistency between the (Zeebe) runtime and secondary storage more visible, which is bad UX.

I will instead create a `feature` issue where this approach can be explored in more detail.

remcowesterhoud: Thanks for raising it @koevskinikola 👍 In that case this is good to merge 🚀

PR #26921, opened 2025-01-15
Authorization creation endpoint

nicpuppa:
## Description

This PR adds:

- authorization creation endpoint
- new fields in the authorization record
- service layer to send request to the broker
- new CREATE/CREATED authorization intents

## Related issues

closes #26876 
closes #26889

PR #26941, opened 2025-01-15
Authorization delete endpoint

### LINKED ISSUE #26922: As a User I can delete an authorization by its key (DELETE)
Create a new endpoint DELETE /authorizations/{key}.

The service layer to send the request to the broker need to be created, as well as the open api specification

nicpuppa:
## Description

This PR adds an endpoint to delete an authorization given its key

⚠️ This PR depends on https://github.com/camunda/camunda/pull/26921

## Related issues

closes #26922

### DIFF
--- a/service/src/main/java/io/camunda/service/AuthorizationServices.java
+++ b/service/src/main/java/io/camunda/service/AuthorizationServices.java
@@ -19,6 +19,7 @@ import io.camunda.service.search.core.SearchQueryService;
 import io.camunda.service.security.SecurityContextProvider;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
 import io.camunda.zeebe.gateway.impl.broker.request.BrokerAuthorizationCreateRequest;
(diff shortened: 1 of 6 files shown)


nicpuppa:
I'm not sure why the REST linting is complaining here, I don't think this two endpoints are ambigous
- DELETE `/authorizations/{authorizationKey}`
- PATCH `/authorizations/{ownerKey}`

We are going to remove the PATCH endpoint anyway so it's not a big deal, but IMO the linter is wrongly configured. The verbs of them is different. I think the linter should take into account not only the path but also the verbs.

Happy to hear your opinion here @tmetzke

pepopowitz (approved):
👍 The spec changes look good to me!

koevskinikola (approved):
Looks good @nicpuppa 👍 

I have some small improvement feedback for the `BrokerAuthorizationDeleteRequest` but I don't need to re-review the PR for it.

PR #26951, opened 2025-01-15
Modify DbUserState to use usernames as identifier instead of a key

### LINKED ISSUE #26801: Modify `USERS` ColumnFamily key to work with a username instead of a user key
## Description

- Change the key of the `USERS` CF in the `DbUserState` from user key to username.
- Go through the usages of the CF and make sure we only retrieve users by username
- The method to get a user using the user key can be removed

### LINKED ISSUE #26803: Remove `USER_KEY_BY_USERNAME` ColumnFamily
## Description

- Remove the `USER_KEY_BY_USERNAME` CF from the `DbUserState` and deprecate the enum in `ZbColumnFamilies`. 
- Remove any remaining usages.

remcowesterhoud:
## Description

With this PR the CFs get refactored to work with a username instead of the `userKey`. Some noteworthy things:
1. We had a CF tracking keys by username. I had to swap this around so we can retrieve the username by the key. The initial plan was to remove it but since we haven't refactored the other entities this would become a huge PR in which I'd have to tackle it all.
2. Since we didn't refactor the roles/groups/tenants yet I had to add a user lookup in a few places which we didn't have before. These can/should be removed once we refacor these other entities.

## Checklist

- [ ] for CI changes:

## Related issues

closes #26801 
closes #26803

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/GroupEntityAddedApplier.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/GroupEntityAddedApplier.java
@@ -37,7 +37,10 @@ public class GroupEntityAddedApplier implements TypedEventApplier<GroupIntent, G
     final var entityKey = value.getEntityKey();
     final var entityType = value.getEntityType();
     switch (entityType) {
-      case USER -> userState.addGroup(entityKey, groupKey);
+      case USER ->
+          userState
+              .getUser(entityKey)
+              .ifPresent(user -> userState.addGroup(user.getUsername(), groupKey));
       case MAPPING -> mappingState.addGroup(entityKey, groupKey);
       default ->
           throw new IllegalStateException(
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleDeletedApplier.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleDeletedApplier.java
@@ -37,7 +37,11 @@ public class RoleDeletedApplier implements TypedEventApplier<RoleIntent, RoleRec
     // Remove roles from users if EntityType.USER exists
     final var userEntities = entities.get(EntityType.USER);
     if (userEntities != null) {
-      userEntities.forEach(userKey -> userState.removeRole(userKey, roleKey));
+      userEntities.forEach(
+          userKey ->
+              userState
+                  .getUser(userKey)
+                  .ifPresent(user -> userState.removeRole(user.getUsername(), roleKey)));
     }
     // todo remove entity from mapping state
     // delete role from authorization state
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleDeletedApplier.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleDeletedApplier.java
@@ -10,40 +10,24 @@ package io.camunda.zeebe.engine.state.appliers;
 import io.camunda.zeebe.engine.state.TypedEventApplier;
 import io.camunda.zeebe.engine.state.mutable.MutableAuthorizationState;
 import io.camunda.zeebe.engine.state.mutable.MutableRoleState;
-import io.camunda.zeebe.engine.state.mutable.MutableUserState;
 import io.camunda.zeebe.protocol.impl.record.value.authorization.RoleRecord;
 import io.camunda.zeebe.protocol.record.intent.RoleIntent;
-import io.camunda.zeebe.protocol.record.value.EntityType;
 
 public class RoleDeletedApplier implements TypedEventApplier<RoleIntent, RoleRecord> {
 
   private final MutableRoleState roleState;
-  private final MutableUserState userState;
   private final MutableAuthorizationState authorizationState;
 
   public RoleDeletedApplier(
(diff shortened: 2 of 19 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/GroupEntityAddedApplier.java
```diff
@@ -37,7 +37,10 @@ public void applyState(final long key, final GroupRecord value) {
     final var entityKey = value.getEntityKey();
     final var entityType = value.getEntityType();
     switch (entityType) {
-      case USER -> userState.addGroup(entityKey, groupKey);
+      case USER ->
+          userState
+              .getUser(entityKey)
+              .ifPresent(user -> userState.addGroup(user.getUsername(), groupKey));
```
koevskinikola, line 43: 💭 The user check is redundant here, as we perform it in the processor. But I guess this is what you meant in the PR description, that first, we need to refactor the tenant/group/role states to work with IDs instead of keys.

And you need to retrieve the `username` somehow 🙂

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleDeletedApplier.java
```diff
@@ -37,7 +37,11 @@ public void applyState(final long key, final RoleRecord value) {
...
     final var userEntities = entities.get(EntityType.USER);
     if (userEntities != null) {
-      userEntities.forEach(userKey -> userState.removeRole(userKey, roleKey));
+      userEntities.forEach(
+          userKey ->
+              userState
+                  .getUser(userKey)
+                  .ifPresent(user -> userState.removeRole(user.getUsername(), roleKey)));
     }
     // todo remove entity from mapping state
```
koevskinikola, line 46: 🔧 We can remove this code completely. In the `RoleDeleteProcessor` [here](https://github.com/camunda/camunda/blob/a406f791bb017b34f4d20f0c5304f267efb814ac/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/RoleDeleteProcessor.java#L107-L124), we emmit `RoleIntent.ENTITY_REMOVED` events for each member entity, so the `RoleEntityRemovedApplier` is already invoked to remove these entries from the `userState`.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/TenantDeletedApplier.java
```diff
@@ -47,7 +47,11 @@ private void handleUserEntities(
...
     final List<Long> userEntities = entities.get(EntityType.USER);
     if (userEntities != null) {
-      userEntities.forEach(userKey -> userState.removeTenant(userKey, tenantId));
+      userEntities.forEach(
+          userKey ->
+              userState
+                  .getUser(userKey)
+                  .ifPresent(user -> userState.removeTenant(user.getUsername(), tenantId)));
     }
   }
```
koevskinikola, line 56: 🔧 We can remove this code completely. In the `TenantDeleteProcessor` [here](https://github.com/camunda/camunda/blob/a406f791bb017b34f4d20f0c5304f267efb814ac/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/tenant/TenantDeleteProcessor.java#L127-L144), we emit `TenantIntent.ENTITY_REMOVED` events for each member entity, so the `TenantEntityRemovedApplier` is already invoked to remove these entries from the `userState`.

koevskinikola (approved):
LGTM @remcowesterhoud 👍 

I found some code that we can clean up since we're actually removing member entities twice for Roles and Tenants.

On Role and Tenant deletion, we're emitting `ENTITY_REMOVED` events, so that they can be processed by the Camunda Exporter, and the appropriate data is also removed in ES/OS.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/GroupEntityAddedApplier.java
remcowesterhoud: Exactly. It's not here to check if the user exist, it's to find the username 😄

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/RoleDeletedApplier.java
remcowesterhoud: Intersting, I didn't know that. I wonder if we should remove it from the applier, or from the processor. I worry that if a role has a lot of users we could run into maximum batch size problems. 

On the other hand, it'd remove the user state from the applier which is quite nice. Especially since we can also remove the authorization state from it soon, since we can just write an authorization delete command.

remcowesterhoud: I think not having it in the applier is nicer. I'll remove it from there. If we run into batch size issues we'll deal with it then. There's an easy workaround, which is to remove a bunch of users manually before deleting the role.

koevskinikola: We shouldn't have removed it from the processor. That way, the exporter won't be able to clean up the ES/OS entries since we don't read from the ES/OS state, we just append to it. We can discuss this in more detail if you like.

PR #26952, opened 2025-01-15
[CPT] feat: Assert a process instance using selectors

### LINKED ISSUE #23091: [CPT] Assert a process instance by key
## Description

I want to use the assertions without providing a `ProcessInstanceEvent` or `ProcessInstanceResult`, for example, by providing a process instance key. 

In ZPT, we have the `InspectionUtility` to find a process instance. For CPT, we could integrate a similar way. 

## Tasks
- I can use the process instance assertion by providing a process instance key
- I can use the process instance assertion if the process is created from a message or timer start event
- The assertion should block until the process instance exists, similar to the usage with the process instance event

```
// by event
CamundaAssert.assertThat(processInstance);

// (new) be selector
CamundaAssert.assertThat(ProcessInstanceSelector.byKey(key));
CamundaAssert.assertThat(ProcessInstanceSelector.byProcessId(processId));
CamundaAssert.assertThat(processInstance -> {..});
```

saig0:
## Description

- Extend the assertion API by selectors for a process instance
    - Provide two predefined selectors to find a process instance by key or by process ID
    - Allow custom selectors to find a process instance
    - Align the selector API with the element selectors
- Refactoring: replace the internal `ProcessInstanceDto` with the response object
    - A step towards using the Camunda query API internally
    - Allow using the response object in the selector interface
- Refactoring: split element assertions from the process instance assertions         
    - Reduce assertion class size and increase the cohesion
    - Align the structure of the element assertions with the variable assertions 
    - Split the assertion test cases into separate classes

**Reviewer note**
The refactorings might make a review of the individual commits easier than the final changes. You could skip over the refactoring and focus on the actual changes, which are much less than 1k. :sweat_smile:     

## Related issues

closes #23091


testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/assertions/ProcessInstanceSelectors.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public boolean test(final ProcessInstance processInstance) {
+      return processInstance.getProcessInstanceKey().equals(processInstanceKey);
+    }
+
+    @Override
+    public String describe() {
+      return String.format("key: %s", processInstanceKey);
```
berkaycanbc, line 58: 🔧 In existing `ElementSelectors` we return the asserted field itself without the prefix (e.g. `key: `). Would be precise to have alignment between them. But is it going to break backward compatibility if we change it in `ElementSelectors`?

berkaycanbc (approved):
Thanks for refactoring! ⭐ 

I only have a minor possible change request, not a blocker.

LGTM 👍

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/api/assertions/ProcessInstanceSelectors.java
saig0: Good point. :+1: 

As a result, the failure message would be:
```
// by ID
Process instance [key: %d] should have active elements [id: 'A', id: 'C', id: 'D'] but the following elements were not active:
  - id: 'C': completed
  - id: 'D': not activated

// by name
Process instance [key: %d] should have active elements [name: 'A', name: 'C', name: 'D'] but the following elements were not active:
  - name: 'C': completed
  - name: 'D': not activated
```

I'm not sure about the format. :thinking: I'll keep it in mind and take it as an optional follow-up task. :+1:

PR #26954, opened 2025-01-15
feat: ignore failure to sync with newly added cluster member

lenaschoenburg:
When a member is just added to a cluster, we try to update it with the latest cluster configuration. This is quite likely to fail because the new member is still starting up and not ready to sync yet. This would cause "Failed to sync" warnings that can look a bit scary.

Instead of trying to silence the warning, for example by tracking that a member is newly added and therefor ignoring a couple of expected `MessagingException` types, we now ignore the outcome of the first sync by unicasting a gossip request instead of requesting a sync.

Since gossip requests and syncs are handled identically on the receiver side, we only change the behavior on the sender side: with sync, the sender could in theory receive an updated state from the newly added member directly. With the switch to gossip, this sync would occur a bit later as part of the regularly scheduled syncs.
This change in behavior is acceptable because an immediate sync with newly added members is not required for correctness, it is only an improvement to the cluster reconciliation time in the unlikely case that a newly added member has a newer configuration state.

The benefit of switching to gossip is that we don't log out a warning in case the newly added member isn't quite ready yet.


filipecampos (approved):
LGTM

Thanks for the changes @lenaschoenburg

PR #26964, opened 2025-01-15
ci: switch to the default maven resolver

lenaschoenburg:
We were using the old wagon resolver because the new native one was buggy in Maven versions < 3.9.2. 

This change lets maven choose the resolver itself but still stes connection TTL, connection pooling and retry counts to maintain the previous behavior mostly.


cmur2 (approved):
Let's give it a try 👍🏼

PR #26971, opened 2025-01-15
Prefix migration bash script

EuroLew:
## Description

This is a bash script that will handle the migration of the documents and historic indices.

Dependencies:
- `jq`
- `curl`

The script assumes that elasticsearch is reachable at localhost:9200

Not sure about the exact location to place the script, any suggestions would be appreciated

## Related issues

relates to https://github.com/camunda/camunda-docs/pull/4844

### DIFF
--- a/zeebe/migrate.sh
+++ b/zeebe/migrate.sh
@@ -1,0 +1,185 @@
+#!/bin/bash
+
+# Default values
+operate_prefix=""
+tasklist_prefix=""
+new_prefix=""
+
+# Help function
+display_help() {
+    echo "Usage: $0 [options]"
+    echo
+    echo "Options:"
+    echo "  --operate <value>     Prefix defined for operate"
+    echo "  --tasklist <value>    Prefix defined for tasklist"
+    echo "  --prefix <value>      New prefix for harmonised indices"
+    echo "  -h, --help            Display this help message."
+    exit 0
+}
+
+# Parse command-line arguments
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        --operate)
+            operate_prefix="$2"
+            shift 2
+            ;;
+        --tasklist)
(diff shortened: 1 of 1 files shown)


zeebe/migrate.sh
```diff
@@ -0,0 +1,185 @@
...
+echo "Cloning indices to new prefix"
+
+replace_prefix() {
+  local index="$1"
+
+  if [[ $index == "$tasklist_prefix"* ]]; then
+    echo "${index/$tasklist_prefix/$new_prefix-tasklist}"
+  elif [[ $index == "$operate_prefix"* ]]; then
+    echo "${index/$operate_prefix/$new_prefix-operate}"
+  else
```
ChrisKujawa, line 150: 👍🏼

ChrisKujawa (commented):
Thanks @EuroLew 🎉 

I think this will already help a lot in migrating 👍🏼 Please take a look at my comments and I think you should check your script also with https://www.shellcheck.net/, there is also an cli tool for it. 

Regarding testing: I guess you did, right? Do you have a sequence of steps to test it? Can you share it? Will be useful for qa as well. 

Regarding place to put I think would be good if it ends in the dist at the end, so user already have it available if they download the platform.

----


I'm currently a bit concerned about automated tests, I guess it will be a bit harder to set them up. Wdyt?
I have to admit I think you first intention was actually right to put it into a java application. Do you think it is easily doable the same also with the ES client (I guess so?). With that we might be able to set up the test scenario in an integration test? With test container using the old version using old prefixes? 🤔 

Do you have an opinion?

ChrisKujawa:
For reference https://camunda.slack.com/archives/C05QBRGMA9F/p1737123140169309?thread_ts=1737123140.169309&cid=C05QBRGMA9F

PR #26978, opened 2025-01-15
docs: improve REST API spec documentation

### LINKED ISSUE #26741: [C8 API] Fix API docs switcher for `oneOf`
**Describe the bug**

Open API documentation displays advanced filter keys with a switcher but with weird title:

**To Reproduce**

- Open any API docs containing an advanced filter and find the filter switcher, for instance: https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/find-process-instances/
- And navigate into Request -> BODY -> filter -> processInstanceKey.
- See switcher with `MOD1` title.

**Expected behavior**

The switcher is displayed with a proper title.

**Environment:**
Latest API docs containing advanced filters.

**Solution:**

Add a `title` for the `oneOf` primitive cases:

```yaml
    LongFilterProperty:
      description: Long property with full advanced search capabilities.
      type: object
      oneOf:
        - type: integer
          format: int64
          title: Integer
        - $ref: "#/components/schemas/AdvancedLongFilter"
```

pepopowitz:
## Description

Improves the REST API spec documentation, including: 

1. Brings some subtle documentation changes upstream from https://github.com/camunda/camunda-docs/pull/4811.
2. Addresses #26741, by adding `title` properties to filtering union types.

### Filtering union titles

Previously, the union type switcher which appears on multiple filtering requests would emit a nonsense title for the "exact match" tab, and a schema name for the "advanced filter" tab, as shown in #26741.

With this PR, these switchers now all emit the title "Exact match" for the exactly matching type, and "Advanced filter" for the advanced filtering type, like this:

image

There is one specific change that introduces more than a `title` or `description` property, and I'll call it out with a review comment. 

#### Strange lack of precision in exact match type descriptions

Strangely, the documentation plugin doesn't emit as much precision in the types for the "exact match" option. `date-time` strings are described as `string`, and `int32` integers are described as `integer`. This can be seen in the above screenshot.

This strangeness isn't introduced by this PR, but it will be obvious when you review, so I wanted to call it out as a known issue. 

## Related issues

closes #26741

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -4557,6 +4557,7 @@ components:
           allOf:
             - $ref: "#/components/schemas/ProcessInstanceFilterRequest"
     AdvancedIntegerFilter:
+      title: Advanced filter
       description: Advanced integer (int32) filter.
       type: object
       properties:
@@ -4599,8 +4600,11 @@ components:
       oneOf:
         - type: integer
           format: int32
+          title: Exact match
+          description: Matches the value exactly.
         - $ref: "#/components/schemas/AdvancedIntegerFilter"
     BasicLongFilter:
+      title: Advanced filter
       description: Basic advanced long (int64) filter.
       type: object
       properties:
@@ -4622,6 +4626,7 @@ components:
             type: integer
             format: int64
     AdvancedLongFilter:
+      title: Advanced filter
       description: Advanced long (int64) filter.
       allOf:
         - $ref: "#/components/schemas/BasicLongFilter"
@@ -4644,6 +4649,7 @@ components:
               type: integer
               format: int64
     AdvancedStringFilter:
+      title: Advanced filter
       description: Advanced string filter.
       type: object
       properties:
@@ -4667,6 +4673,7 @@ components:
             Supported wildcard characters depend on the configured search client.
           type: string
     AdvancedProcessInstanceStateFilter:
+      title: Advanced filter
       description: Advanced ProcessInstanceStateEnum filter.
       type: object
       properties:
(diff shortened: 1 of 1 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4735,32 +4742,44 @@ components:
...
     ProcessInstanceStateFilterProperty:
       description: ProcessInstanceStateEnum property with full advanced search capabilities.
       type: object
       oneOf:
-        - $ref: "#/components/schemas/ProcessInstanceStateEnum"
+        - type: string
+          title: Exact match
+          description: Matches the value exactly.
+          allOf:
+            - $ref: "#/components/schemas/ProcessInstanceStateEnum"
```
pepopowitz, line 4773: 💭 I don't know if this affects the API itself, and if there are non-spec changes I need to make to go with this. 

This is introduced because the `ProcessInstanceStateEnum` is used many places, and if I were to attach the "exact match" title and description to that schema itself, it would not make sense for any consuming schemas other than this one. 

So, I swapped this to use `allOf` inheritance, so that I could override the title and description.

In doing so, I added `type: string`...which is not precisely accurate, but (a) it was the closest type I could think of, and (b) not including a type results in the description and type not appearing, like this:

image

I don't enjoy introducing changes like this, which are only necessary for the downstream documentation generator, but this seemed like a viable pragmatic approach. I'm definitely open to suggestions.

danielkelemen: If the CI runs fine then it should be okay code-wise.

It doesn't look like it shows any type on the screenshot below the `Exact match`.
Shouldn't it show `string` or the enum values? It would be nice it would show the enum values.

pepopowitz: Sorry for the confusion @danielkelemen , that screenshot is of the format I _didn't_ choose, for the reasons you described.

This is what it looks like with what's in this PR:

image

danielkelemen (approved):
👍 Looks good to me. Nice improvement!

tmetzke (approved):
👍 This is great, thanks a lot, @pepopowitz 🚀

PR #26986, opened 2025-01-16
pg-23718-consolidate-abstractions

panagiotisgts:
## Description

- Remove component specific abstractions TasklistEntity & OperateZeebeEntity
- Fixed a test case onCamundaExporterIT#shouldExportRecords, the case for UserTaskVariableBatch was not actually testing whether the persisted entities were correct, since they are not of the same Class of the ExporterHandler itself
## Checklist

- [ ] for CI changes:

## Related issues

closes #


webapps-schema/src/main/java/io/camunda/webapps/schema/entities/tasklist/TaskVariableEntity.java
```diff
@@ -8,8 +8,12 @@
 package io.camunda.webapps.schema.entities.tasklist;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
+import io.camunda.webapps.schema.entities.AbstractExporterEntity;
+import io.camunda.zeebe.protocol.record.value.TenantOwned;
 
-public class TaskVariableEntity extends TasklistEntity<TaskVariableEntity> {
+public class TaskVariableEntity extends AbstractExporterEntity<TaskVariableEntity> {
+
+  private String tenantId = TenantOwned.DEFAULT_TENANT_IDENTIFIER;
```
ChrisKujawa, line 16: ❓ Should this tenant Id be in a separate abstraction then maybe?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -390,14 +392,19 @@ void shouldExportRecord(
...
           try {
             responseEntity =
                 clientAdapter.get(
-                    expectedEntity.getId(),
-                    exportHandler.getIndexName(),
-                    exportHandler.getEntityType());
+                    expectedEntity.getId(), exportHandler.getIndexName(), getType(exportHandler));
+            if (exportHandler.getEntityType().equals(UserTaskVariableBatch.class)) {
+              assertThat(responseEntity).isInstanceOf(TaskVariableEntity.class);
+            }
```
ChrisKujawa, line 407: 🤔 ❓ Why was this change necessary?

operate/webapp/src/test/java/io/camunda/operate/webapp/rest/FlowNodeInstanceMetadataBuilderTest.java
```diff
@@ -244,6 +244,7 @@ void buildBusinessRuleTaskMetadata() {
 
   private void fillStandardValues(final FlowNodeInstanceEntity flowNodeInstance) {
     flowNodeInstance
+        .setKey(0L)
```
ChrisKujawa, line 247: ❓ Was there a reason why we had to do this?

ChrisKujawa (commented):
thanks @panagiotisgts looks good in general. I had some questions and comments, we should clarify first. 

I asked @EuroLew to review your other PRs as I'm currently quite short on time

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -390,14 +392,19 @@ void shouldExportRecord(
     assertThat(expectedHandlers).isNotEmpty();
     expectedHandlers.forEach(
         exportHandler -> {
-          final ExporterEntity expectedEntity = getExpectedEntity(record, exportHandler);
+          ExporterEntity expectedEntity = getExpectedEntity(record, exportHandler);
```
EuroLew, line 395: ❌  This seems like a generic test to test exporting if there is a special case for this I would write it in a different test

EuroLew (commented):
I would like the test to be own it's own otherwise great job

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
panagiotisgts: The actual flushed entity is a TaskVariableEntity and not a UserTaskVariableBatch, the test case essentially was not creating correct assertions on the persisted entity.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/OperateZeebeEntity.java
```diff
@@ -1,64 +0,0 @@
...
-package io.camunda.webapps.schema.entities.operate;
-
-import io.camunda.webapps.schema.entities.AbstractExporterEntity;
-
-public abstract class OperateZeebeEntity<T extends OperateZeebeEntity<T>>
-    extends AbstractExporterEntity<T> {
-
-  private long key;
-
-  private int partitionId;
```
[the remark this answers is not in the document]
panagiotisgts: The fields moved to `AbstractExporterEntity` for Key and PartitionId are boxed since they are not necessarily mandatory for all entities, thus need to be excluded from serialization when flushing to ES/OS to avoid inserting new fields in the indies.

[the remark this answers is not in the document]
ChrisKujawa: Does it make sense to have then maybe a new layer/sub class for it?

operate/webapp/src/test/java/io/camunda/operate/webapp/rest/FlowNodeInstanceMetadataBuilderTest.java
panagiotisgts: Mocks use the `flowNodeInstance.getKey()` to return predetermined results. Now that the Key is a boxed Long and not primitive, this caused NPEs on test cases.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
panagiotisgts: This test is specific for all handlers of `ValueType.VARIABLE`, which are 4 (i think). I'm 50-50 on whether this should be a different case or remain on this one.

EuroLew: imo it's just an extra test not like we're adding any complexity to prod.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/ImportPositionEntity.java
```diff
@@ -108,19 +97,13 @@ public ImportPositionEntity setCompleted(final boolean completed) {
...
   public int hashCode() {
     return Objects.hash(
-        super.hashCode(),
-        aliasName,
-        partitionId,
-        position,
-        sequence,
-        postImporterPosition,
-        indexName);
+        super.hashCode(), aliasName, position, sequence, postImporterPosition, indexName);
```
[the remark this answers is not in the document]
panagiotisgts: The partitionId is now part of the AbstractExporterEntity, not sure how this would cause any implications. The hashCode will change since it's part of another class but why would it cause any issue, since it will be calculated anew.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/OperateZeebeEntity.java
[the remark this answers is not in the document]
panagiotisgts: IMO I believe these make sense to be part of the AbstractExporterEntity since it's data that are exposed by Zeebe on most records (AFAIK), even though not every entity is making use of them.

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/tasklist/TaskVariableEntity.java
panagiotisgts: @ChrisKujawa I forgot to have these classes implement the TenantOwned interface that is already exposed. Having another abstract class over the ExporterEntity is another option, but it would cause confusion probably again, WDYT?

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/FlowNodeInstanceEntity.java
```diff
@@ -8,11 +8,14 @@
...
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import io.camunda.webapps.schema.entities.AbstractExporterEntity;
+import io.camunda.zeebe.protocol.record.value.TenantOwned;
 import java.time.OffsetDateTime;
 import java.util.Arrays;
 import java.util.Objects;
 
-public class FlowNodeInstanceEntity extends OperateZeebeEntity<FlowNodeInstanceEntity> {
+public class FlowNodeInstanceEntity extends AbstractExporterEntity<FlowNodeInstanceEntity>
+    implements TenantOwned {
```
ChrisKujawa, line 18: 🔧 Maybe another layer for tenants make sense

webapps-schema/src/main/java/io/camunda/webapps/schema/entities/operate/OperateZeebeEntity.java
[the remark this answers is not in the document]
ChrisKujawa: I feel this is a wrong use of inheritance then. It sounds like not all entities are similar, so they should not have the same parent or not have the property in the parent. 


IMHO: It violates `interface segregation principle` 

>A client should never be forced to implement an interface that it doesn’t use, or clients shouldn’t be forced to depend on methods they do not use.

This also applies to properties.

IMHO: It is also hurting sort of `Liskov Substitution Principle `

> This means that every subclass or derived class should be substitutable for their base or parent class.

Which is not true if properties are not really present/used in certain cases.


To be more abstract (example):

It would be like you have a cherry Tree, a plum tree, and a fir. All of them are a tree. But not all of them have eatable fruits.You wouldn't put fruit property to the tree, but more likely introduce like a separate abstraction of fruit tree. 


I hope this makes sense.

I see here two/three options:

 1. We introduce another layer for the partition-related entities, and likely also for tenants aware
 2. We completely remove this hierarchy stuff and use plain old objects. In our current case, it is likely that the entities do not change much anymore. It would reduce the instability, that we currently have if we change something on the base that it affects all of them. The question is really for me do we need the inheritance here? Is it just for reusing some simple getter and setter? 
     3. We could use records instead, which auto-generates all getter/setter and equals/hashcode, etc.

ChrisKujawa (changes requested):
@panagiotisgts sorry for the delay, the last weeks were quite busy and rough. 

I think it would be helpful if we could split this PR into multiple smaller ones, I promise you we can more easily review them if they have a clearer and smaller scope.

 1. Take care of your test fix, you have encountered https://github.com/camunda/camunda/pull/26986/files#diff-34c14eb11449ae1dcbebfe45d784f738025bb052006ae34dd9e0851094d1e3ff
 2. Remove of obsolete classes https://github.com/camunda/camunda/pull/26986/files#diff-fb0bde51d2adb60d890ddbc2a3a63c9dc909c1f8f625ea0e8608e4bb528c9e0b
 3. Remodel the inheritance or remove it completelty.

panagiotisgts:
@ChrisKujawa Looking back on this after a while I believe we'd go the simpler route:
- :x:  Using records since they are immutable will be a bit tedious to work with and will require changes throughout Exporter, Tasklist & Operate on runtime as well on the way test case are currently created
- :wrench: The test case I discovered is actually out of context of the parent test class and should be removed, it's already covere by the Handler UTs
- :wrench: Maintain only the reference to the `TenantOwned` interface since it provides context regarding the usage of the entity
- :wrench: I'll split this PR also in two parts. One for Tasklist and another for Operate entities

WDYT?

Issue #26761, opened 2025-01-10
[C8 API] Add OpenAPI request String keys


danielkelemen:
## Description

String keys are also accepted in request objects (or params) therefore we should also define them in the OpenAPI request specification.

## Acceptance criteria

- Define OpenAPI requests with String and Long (in64) keys for the required APIs.
  - Check how parameters are defined and if it's also necessary to duplicate.
- 8.7 and 8.6. 
- No code changes required.

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/26992
- [ ] https://github.com/camunda/camunda/pull/26995
- [ ] https://github.com/camunda/camunda/pull/27062
```

tmetzke:
@pepopowitz, any ideas why the `oneOf` does not work with docusaurus in that case? Does this work with version 3?

pepopowitz:
No, I'm not sure. It also does the same thing in docusaurus 3, unfortunately. It is likely a bug in how the plugin handles oneOf in a parameter. I couldn't find any issues in their repo, open or closed.

I also noticed that Swagger doesn't render it properly either; it chooses the first type in the array:

<img width="1219" alt="Image" src="https://github.com/user-attachments/assets/ec93c717-945d-49a4-b8b6-b185296351ec" />

As far as I can tell from the OpenAPI spec there's no reason you can't put a oneOf in a parameter...it seems like tooling just isn't expecting it for some reason....

Which is also what both ChatGPT and Claude told me about this. They both suggested an alternative approach of specifying it as a string with a `pattern` that fits the two different types. 


(And then they both linked me to multiple specs and conversations that had nothing to do with the problem at hand.....and that is why they can't be trusted.)

danielkelemen:
## Request Naming

#### Existing request objects:
- `*SearchQueryRequest`
- `*Request`
- `*FilterRequest`
- `*Instruction` (ModifyProcessInstanceActivateInstruction, ModifyProcessInstanceTerminateInstruction)

#### New String key object naming
- It's not easy to find something that fits all of them, so I decided to use `RequestBody` for all to be consistent.
  - `UserTaskSearchQueryRequest` -> `UserTaskSearchQueryRequestBody`
  - `UserTaskFilterRequest` -> `UserTaskFilterRequestBody`
  - `ModifyProcessInstanceActivateInstruction` -> `ModifyProcessInstanceActivateRequestBody`

tmetzke:
[...]

PR #26992, opened 2025-01-16
feat: prepare advanced string key filter

danielkelemen:
## Description

- Move keys to BasicLongFilter
- Prepare BasicStringFilter for string keys
- Adjust client & gateway-rest

## Related issues

related to https://github.com/camunda/camunda/issues/26761


danielkelemen:
### TODOs:
- [ ] Drop LongFilter?
   - It's not used at the moment.
   - Could it be used in the future? Usually non-key number values are just integers, so not likely.

tmetzke (approved):
👍 This makes a lot more sense for the key attributes, thanks 🦖 
💭 I think rebasing onto `main` and adjusting the OpenAPI spec to incorporate the changes from https://github.com/camunda/camunda/pull/26978 might make sense here.

danielkelemen:
Added the new `title` properties, too.

PR #26993, opened 2025-01-16
fix: in HeartbeatHandler.Client check the payload & status only if heartbeat

entangled90:
## Description

In HeartbeatHandler.Client payload and status were checked even if the reply is not a heartbeat, creating a lot of logs at WARN. The map oustandingHeartbeats was empty, so all messages were not considered heartbeats.

This commit fixes the two issues mentioned above.

## Related issues

relates #26602

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
@@ -10,6 +10,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.timeout.IdleState;
 import io.netty.handler.timeout.IdleStateEvent;
 import io.netty.handler.timeout.IdleStateHandler;
+import io.netty.util.ReferenceCountUtil;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
@@ -97,6 +98,8 @@ abstract sealed class HeartbeatHandler extends ChannelDuplexHandler {
       // Handlers downstream can receive heartbeats as well
       if (!heartbeatReceived || forwardHeartbeats) {
         ctx.fireChannelRead(msg);
+      } else {
+        ReferenceCountUtil.release(msg);
       }
     }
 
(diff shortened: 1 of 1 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -158,32 +161,35 @@ public void handlerAdded(final ChannelHandlerContext ctx) {
...
-          heartbeatReceived = true;
-        }
+        final var isHeartbeat = outstandingHeartbeats.contains(reply.id());
         // remove all heartbeats sent before (and equal) to the last ProtocolReply received.
         // Note that all ProtocolReply messages (not just heartbeat messages) can indicate that
         // the channel is still open and transmitting.
         // If we remove from the set only the last message received, the size of
         // oustandingHeartbeats will increase for every missed
         // heartbeat and the memory will be freed only when the channel is closed.
         outstandingHeartbeats.removeIfLong(id -> id <= reply.id());
```
npepinpe, line 171: :question: Considering we remove earlier heartbeats, I may later get the reply - it would just be ignored if I understood correctly, right? Cause it wouldn't be able to tell it's a heartbeat.

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -97,6 +98,8 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
       // Handlers downstream can receive heartbeats as well
       if (!heartbeatReceived || forwardHeartbeats) {
         ctx.fireChannelRead(msg);
+      } else {
+        ReferenceCountUtil.release(msg);
       }
```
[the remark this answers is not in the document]
entangled90: If it's not a heartbeat message the branch `ctx.fireChannelRead` should be taken, so it can be consumed by other handlers:
```
heartbeatReceived = false
forwardHeartbeats = false
(!false) || false = true || false = true
```

The implementation of `super.channelRead()` is to just `fireChannelRead`

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
entangled90: That should be and edge case as the server replies immediately, since the server `HeartbeatHandler` runs before the "MessageDispatcher".

Anyway if it happens, then it's not marked as an heartbeat response, so it may be forwarded to handlers, which will result in a log line saying some thing like:
"Received message with id[2] but no handler registered"

npepinpe: Hm, ideally we would not log this, as it's a really unnecessary warning, and we've had customers open support issues for those :smile: 

I get not keeping track of all forever though :thinking: So we would need a way to know the reply is for a heartbeat, right? 

We can merge this for now, can we create a follow up to address this? In case we do run into this edge case, it's not fully undocumented.

npepinpe (approved):

entangled90:
I just realized we can check the SBE message header to see if it's `HeartbeatSetupResponse`, so in that case we are sure if it's a heartbeat or not :+1:

PR #26995, opened 2025-01-16
feat: add request parameter string key support

### LINKED ISSUE #26761: [C8 API] Add OpenAPI request String keys

danielkelemen:
## Description

- Add request KeyParameter OpenAPI type
- Add request string & number header support
- Add request string object types

## Related issues

closes https://github.com/camunda/camunda/issues/26761

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -160,7 +160,7 @@ paths:
           required: true
           description: The key of the job to fail.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: false
         content:
@@ -208,7 +208,7 @@ paths:
           required: true
           description: The key of the job.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: true
         content:
@@ -256,7 +256,7 @@ paths:
           required: true
           description: The key of the job to complete.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: false
         content:
@@ -303,7 +303,7 @@ paths:
           required: true
           description: The key of the job to update.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: true
         content:
@@ -352,7 +352,7 @@ paths:
           required: true
           description: Key of the incident to resolve.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "204":
           description: The incident is marked as resolved.
@@ -588,7 +588,7 @@ paths:
           required: true
           description: The unique identifier of the tenant.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
         - name: userKey
           in: path
           required: true
@@ -629,7 +629,7 @@ paths:
           required: true
           description: The unique identifier of the tenant.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
         - name: mappingKey
           in: path
           required: true
@@ -669,7 +669,7 @@ paths:
           required: true
           description: The unique identifier of the tenant.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
         - name: mappingKey
           in: path
           required: true
@@ -710,7 +710,7 @@ paths:
           required: true
           description: The unique identifier of the tenant.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
         - name: groupKey
           in: path
           required: true
@@ -750,7 +750,7 @@ paths:
           required: true
           description: The unique identifier of the tenant.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
         - name: groupKey
           in: path
           required: true
@@ -836,7 +836,7 @@ paths:
           required: true
           description: The key of the user task to complete.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: false
         content:
@@ -885,7 +885,7 @@ paths:
           required: true
           description: The key of the user task to assign.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: true
         content:
@@ -934,7 +934,7 @@ paths:
           required: true
           description: The user task key.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "200":
           description: >
@@ -982,7 +982,7 @@ paths:
           required: true
           description: The key of the user task to update.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: false
         content:
@@ -1033,7 +1033,7 @@ paths:
           required: true
           description: The user task key.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "200":
           description: >
@@ -1083,7 +1083,7 @@ paths:
           required: true
           description: The key of the user task.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "204":
           description: The user task was unassigned successfully.
@@ -1174,7 +1174,7 @@ paths:
           required: true
           description: The key of the user task.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       requestBody:
         required: false
         content:
@@ -1269,7 +1269,7 @@ paths:
           required: true
           description: The variable key.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "200":
           description: >
@@ -1417,7 +1417,7 @@ paths:
           required: true
           description: The assigned key of the process definition, which acts as a unique identifier for this process definition.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "200":
           description: >
@@ -1469,7 +1469,7 @@ paths:
           required: true
           description: The assigned key of the process definition, which acts as a unique identifier for this process.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "200":
           description: >
@@ -1524,7 +1524,7 @@ paths:
           required: true
           description: The process key.
           schema:
-            $ref: "#/components/schemas/KeyParameter"
+            type: string
       responses:
         "200":
           description: >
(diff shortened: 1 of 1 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1593,7 +1586,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/CreateProcessInstanceRequest"
+              $ref: "#/components/schemas/CreateProcessInstanceRequestBody"
```
tmetzke, line 1589: 🔧 The example below also needs to move to using a string key instead of a number

tmetzke: ❓ @pepopowitz, do I recall correctly that we have an automation in `camunda-docs` regarding this endpoint? Does that need an adjustment now?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -2337,7 +2369,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/EvaluateDecisionRequest"
+              $ref: "#/components/schemas/EvaluateDecisionRequestBody"
```
tmetzke, line 2372: 🔧 Same as above, the example needs to move to a string key as well.
❓ Same as above, @pepopowitz, do we need to adjust any automation in `camunda-docs` for this?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4224,21 +4265,38 @@ components:
...
+      properties:
         filter:
           description: The user task search filters.
           allOf:
             - $ref: "#/components/schemas/UserTaskFilterRequest"
+    UserTaskSearchQueryRequestBody:
+      description: User task search query request. Key attributes as string values.
+      type: object
+      allOf:
+        - $ref: "#/components/schemas/SearchQueryRequest"
```
tmetzke, line 4293: ❓ Could this also move into the `UserTaskSearchQueryRequestBase`?

tmetzke (commented):
👍 Looks good in general, on a structural level, good job 🍪 
🔧 There are some smaller suggestions from my end.

💭 As mentioned in https://github.com/camunda/camunda/issues/26761#issuecomment-2601951616, we could also move to shorter names than even longer ones. But it's just a thought, your names work fine as well in the end. Personally, I think they are becoming a bit too "expressive" now. They basically now describe the location they are used in rather than their inherent nature, e.g. it's a `UserTaskSearchQuery` with a `UserTaskFilter` and they are used in request bodies. I don't see a need to call them `UserTaskSearchQueryRequestBody` or `UserTaskFilterRequestBody`.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: Yes, it actually could.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: Ah right, the example! I'll update it.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -8400,21 +8900,41 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/ModifyProcessInstanceTerminateInstruction"
-        operationReference:
-          description: >
-            A reference key chosen by the user that will be part of all records resulting from this operation.
-            Must be > 0 if provided.
-          type: integer
-          format: int64
-          minimum: 1
```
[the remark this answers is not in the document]
danielkelemen: Yes, that's why it was moved into the base object.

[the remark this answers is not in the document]
tmetzke: Haha, now I see, sorry for the mix-up 😄 I thought we're having this twice now, as a string and a number attribute. Too many changes for me to handle apparently 🤷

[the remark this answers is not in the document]
danielkelemen: Np, it's too many for everyone 😄

tmetzke (approved):

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5156,37 +5357,42 @@ components:
...
+        - $ref: "#/components/schemas/FlowNodeInstanceSearchQueryRequestBase"
+      properties:
         filter:
           description: The flow node instance search filters.
           allOf:
             - $ref: "#/components/schemas/FlowNodeInstanceFilterRequest"
-    FlowNodeInstanceFilterRequest:
-      description: Flow node instance filter.
+    FlowNodeInstanceSearchQuery:
+      description: Flow node instance search request Key attributes as string values.
```
christinaausley, line 5383:
```suggestion
      description: Flow node instance search request key attributes as string values.
```

Nitpick -- non-blocking.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5862,33 +6184,65 @@ components:
...
           type: integer
           format: int64
+    AuthorizationFilter:
+      description: Authorization search filter. Key attributes as string values.
+      type: object
+      allOf:
+        - $ref: "#/components/schemas/AuthorizationFilterRequestBase"
+      properties:
+        ownerKey:
+          description: The id of the owner of permissions.
```
christinaausley, line 6244:
```suggestion
          description: The ID of the owner of permissions.
```

Is "the owner of permissions" the admin?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -5862,33 +6184,65 @@ components:
...
             - $ref: "#/components/schemas/OwnerTypeEnum"
+    AuthorizationFilterRequest:
+      description: Authorization search filter.
+      deprecated: true
+      type: object
+      allOf:
+        - $ref: "#/components/schemas/AuthorizationFilterRequestBase"
+      properties:
         ownerKey:
           description: The id of the owner of permissions.
```
christinaausley, line 6234:
```suggestion
          description: The ID of the owner of permissions.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -8247,6 +8693,28 @@ components:
...
           format: int64
+    CreateProcessInstanceRequest:
+      deprecated: true
+      type: object
+      allOf:
+        - $ref: "#/components/schemas/CreateProcessInstanceRequestBase"
+      properties:
+        processDefinitionKey:
+          description: |
+            The unique key identifying the process definition, e.g. returned for a process in the
```
christinaausley, line 8704:
```suggestion
            The unique key identifying the process definition, for example, returned for a process in the
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -8247,6 +8693,28 @@ components:
...
+          type: integer
+          format: int64
+    ProcessInstanceCreationInstruction:
+      type: object
+      allOf:
+        - $ref: "#/components/schemas/CreateProcessInstanceRequestBase"
+      properties:
+        processDefinitionKey:
+          description: |
+            The unique key identifying the process definition, e.g. returned for a process in the
```
christinaausley, line 8715:
```suggestion
            The unique key identifying the process definition, for example, returned for a process in the
```

christinaausley (approved):
Some non-blocking suggestions.

pepopowitz:
❓ I'm assuming based on this PR that we are accepting the limitations discussed in [this conversation](https://github.com/camunda/camunda/issues/26761#issuecomment-2593827554), and that the `KeyParameter` requests will not be very clearly described?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: Nice catch, that should be actually a separate sentence! 😀
I'll fix that.

danielkelemen:
> ❓ I'm assuming based on this PR that we are accepting the limitations discussed in https://github.com/camunda/camunda/issues/26761#issuecomment-2593827554, and that the KeyParameter requests will not be very clearly described?

Good point. There are 2 possibilities:

- A. We keep `KeyParameter`.
   - Docosaurus and swaggerui displays it as `string` which would be fine.
   - I tried it in redocly, and there it's displayed with a switcher:
image
   - Maybe it will be helpful for some code generation? I'm not sure.
- B. We define it as `string`.
  - Simpler.
  - Types for http query parameters don't matter that much.
  - We will drop the number key support soon anyway.
  - Even if users generate clients they can stringify their number keys easily.

Now thinking about this again, I'd rather vote for `B` as it seems like there is no real difference and it's less work in the future.
Let me know if you don't agree @pepopowitz, @tmetzke.

tmetzke:
> Now thinking about this again, I'd rather vote for `B` as it seems like there is no real difference and it's less work in the future. Let me know if you don't agree @pepopowitz, @tmetzke.

What could happen in the worst case for users?
The only thing I am currently aware of is if they generated REST clients using the paths as well. Those clients will change now since they don't expect a number anymore but a string. Models stay the same, and sending requests is based on string values anyway. You create a URL to send the request to, and this URL is a string value.

Maybe I am overlooking another impact here but if this is the only thing, I am fine with moving to string directly for path parameters. This makes the spec a lot simpler to maintain.

danielkelemen:
> What could happen in the worst case for users?

@tmetzke, I can also only think of the same. They generate the whole client code and the generated methods would expect a String parameter instead of Long, for instance: `migrateProcessInstance(String processInstanceKey, ...)`

To be honest, the generated code would also look different even with `KeyParameter`. (IF the generator can work with that.)

So for generated clients both are kind of a breaking change but with an easy fix in both cases.
I'll update to `B`.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: It's just the unique key (or ID) of the user, if I'm not mistaken.

PR #27008, opened 2025-01-16
Remove user type from UserRecord

remcowesterhoud:
## Description

`UserType` was an older concept which not relevant. We would use it to determine if a user was authorized. A default user would have all permissions. We changed this to providing explicit permissions through an Admin role instead.

 As there was no way to create users in previous released version we can safely delete the property from the records without breaking backwards compatibility. This is also why we should do it now, because in the near future we can never remove it again.

## Checklist

- [ ] for CI changes:

## Related issues

N/A


lenaschoenburg (commented):
Seems like we should be able to delete `UserType` completely now?

remcowesterhoud:
Can't believe I missed that 🤦

lenaschoenburg (approved):

PR #27012, opened 2025-01-16
Cleanup task listener jobs on element instance termination

### LINKED ISSUE #26786: When canceling instance, not all listener jobs are cleaned up
**Describe the bug**

While showcasing the new task listener feature in `8.7.0-alpha3`, I noticed that sometimes listener jobs remain activatable after the related process instance is canceled.

**To Reproduce**

- deploy process with user task and task listener
- create process instance
- trigger task listener, but don't complete job
- cancel instance
- notice that job can still be activated

**Expected behavior**

Listener jobs should be canceled when the process instance is canceled.

**Environment:**
- OS: 
- Zeebe Version:8.7.0-alpha3
- Configuration:

korthout:
## Description

This resolves a bug where the job of a task listener is not cleaned up after the element instance is terminated (e.g. when canceling the process instance, or when interrupting the user task).

This had an additional effect: an incident associated to such a job was also not cleaned up.

This PR solves both problems by:
- canceling the associated job
- resolving the associated incident

## Related issues

closes #26786


ce-dmelnych (approved):
LGTM 💯👍🏻

Thank you for enhancing the test coverage for Execution Listeners as well!

PR #27013, opened 2025-01-16
Feature/26965 add oracle db support

### LINKED ISSUE #26965: Add Oracle Database support to RDBMS
Additionally to the originally targeted databases (H2, MariaDB, Postgres), now also Oracle should be added to the list of supported vendors.

Acceptance Criteria:
- OracleDB is added to the list of supported and tested database systems
- All RDBMS ITests are also executed successfully against OracleDB

stoerti:
## Description

Added OracleDB support for RDBMS

closes #26965

### DIFF
--- a/db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
+++ b/db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
@@ -80,7 +80,7 @@
       <column name="VAR_NAME" type="VARCHAR(255)" />
       <column name="DOUBLE_VALUE" type="DOUBLE"/>
       <column name="LONG_VALUE" type="BIGINT"/>
-      <column name="VAR_VALUE" type="CLOB"/> <!-- max length of 8191 -->
+      <column name="VAR_VALUE" type="VARCHAR(8192)"/>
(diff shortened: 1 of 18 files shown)


db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
```diff
@@ -92,6 +92,11 @@
     <createIndex tableName="VARIABLE" indexName="IDX_VARIABLE_TENANT_ID">
       <column name="TENANT_ID" />
     </createIndex>
+
+    <modifySql dbms="oracle">
+      <!-- Oracle only supports a size 4000 characters in varchar2 -->
+      <replace replace="VARCHAR2(8192)" with="VARCHAR2(4000)"/>
```
p-wunderlich, line 98: You used VARCHAR in the original sql, not VARCHAR2. Or does the generated sql already contain varchar2?

stoerti: Yes, the Oracle name for the VARCHAR type is VARCHAR2 and the replace is executed on the generated native SQL. I also have not found a better solution than this modifySql-replace pattern. Alternative would be to duplicate the whole changeset.

p-wunderlich (approved):

PR #27023, opened 2025-01-16
feat: exchange heartbeat information on channel creation to avoid false-positives

### LINKED ISSUE #26879: Send heartbeat timeout from the client to the server during handshake
**Description**
In `NettyMessagingService` heartbeats rely on timeouts to be correctly triggered, but they are currently configured separately between client and server.

If these timeouts are not the same, it's easy to false-positives to be reported.
Suppose the two instances have these timeouts configured:
- Client `A`: `{"heartbeatInterval:  "15s", "heartbeatTimeout": "30s}` 
- Server `B`: `{"heartbeatInterval:  "5s", "heartbeatTimeout": "10s}`

When the connection is Idle, `client A` sends an heartbeat every 15 seconds, but `Server B` expects to receive one heartbeat every 10 seconds. As a result, Server will close the connection every `10 seconds`. 

> See #26812 for the original discussion

**Implementation** 
A new protocol version (`V3`) can be added which, as part as the handshake, will include the `int heartbeatTimeout` field  that the server must use for this connection.
If the server is an older version, the client must fallback to the protocol `V2`,  without heartbeats as the server does not  support them.

entangled90:
## Description

`HeartbeatSetupHandler.Client` is responsible for sending heartbeat informations to the server.
`HeartbeatSetupHandler.Server` receives these informations and replies to the client if heartbeats are enabled or not.

The client, upon receiving when receiving `heartbeatsEnable=true` adds to the `ChannelPipeline` the `HeartbeatHandler.Client`, which registers autonomously the `IdleStateHandler`.

The server will add to the `ChannelPipeline` the `HeartbeatHandler.Server` when it receives the first `HeartbeatSetupRequest`.

Both `HeartbeatSetupHandler` remove themselves from the `ChannelPipeline` when the setup has been done.

The `HeartbeatSetup{Request,Response}` are encoded using a new schema defined in `cluster-messaging.xml`

## Related issues
closes #26879

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
@@ -9,8 +9,10 @@ import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.timeout.IdleState;
 import io.netty.handler.timeout.IdleStateEvent;
+import io.netty.handler.timeout.IdleStateHandler;
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import org.agrona.collections.LongHashSet;
 import org.slf4j.Logger;
@@ -46,6 +48,9 @@ abstract sealed class HeartbeatHandler extends ChannelDuplexHandler {
   // DO NOT CHANGE: changing the subject is a backward INCOMPATIBLE change.
   static final String HEARTBEAT_SUBJECT = "internal-heartbeat";
   static final byte[] HEARTBEAT_PAYLOAD = new byte[0];
+  static final String IDLE_STATE_HANDLER_NAME = "idle";
+  static final String HEARTBEAT_HANDLER_NAME = "heartbeat";
+
   protected final Logger log;
   protected final boolean forwardHeartbeats;
 
@@ -55,10 +60,22 @@ abstract sealed class HeartbeatHandler extends ChannelDuplexHandler {
   }
 
   static final class Server extends HeartbeatHandler {
-    boolean receivedHeartbeat = false;
+    private final Duration heartbeatTimeout;
 
-    Server(final Logger log, final boolean fireHeartbeats) {
+    Server(final Logger log, final Duration heartbeatTimeout, final boolean fireHeartbeats) {
       super(log, fireHeartbeats);
+      this.heartbeatTimeout = heartbeatTimeout;
+    }
+
+    @Override
+    public void handlerAdded(final ChannelHandlerContext ctx) {
+      if (ctx.channel().isActive()) {
+        ctx.pipeline()
+            .addFirst(
+                IDLE_STATE_HANDLER_NAME,
+                // server is only interested in readIdleTime
+                new IdleStateHandler(heartbeatTimeout.toMillis(), 0, 0, TimeUnit.MILLISECONDS));
+      }
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatHandler.java
```diff
@@ -55,18 +60,29 @@ protected HeartbeatHandler(final Logger log, final boolean forwardHeartbeats) {
...
+      this.heartbeatTimeout = heartbeatTimeout;
+    }
+
+    @Override
+    public void handlerAdded(final ChannelHandlerContext ctx) {
+      if (ctx.channel().isActive()) {
+        ctx.pipeline()
+            .addFirst(
+                IDLE_STATE_HANDLER_NAME,
+                // server is ony interested in readIdleTime
```
filipecampos, line 76: typo - is only

zeebe/atomix/cluster/src/test/java/io/atomix/cluster/messaging/impl/NettyMessagingServiceTest.java
```diff
@@ -761,11 +760,36 @@ void shouldNotCloseTheConnectionFromTheServerIfNoHeartbeatsIsReceived() {
       assertThat(receivedHeartbeats.get()).isEqualTo(0);
     }
 
+    @Test
+    void shouldExchangeHeartbeatTimeoutFromClientToServer() throws Exception {
```
filipecampos, line 764: Just my taste, that I think the test could be named something like `shouldGetHeartbeatTimeoutValueFromClientToServer` or `shouldConveyHeartbeatTimeoutValueFromClientToServer`.

zeebe/atomix/cluster/src/test/java/io/atomix/cluster/messaging/impl/NettyMessagingServiceTest.java
```diff
@@ -725,29 +725,28 @@ void shouldGetChannelClosedWhenNotSendingHeartbeatsAfterAWhile() {
 
       // when - removing the `IdleStateHandler` from the pipeline such that `HeartBeatHandler` is
       // not triggered
-      clientChannel.pipeline().remove("idle");
+      clientChannel.pipeline().remove(HeartbeatHandler.IDLE_STATE_HANDLER_NAME);
 
       // then - the other side notices a lack of heartbeats and closes the channel
-      assertThat(clientChannel.closeFuture()).succeedsWithin(Duration.ofSeconds(5));
+      assertThat(clientChannel.closeFuture()).succeedsWithin(Duration.ofSeconds(500));
```
filipecampos, line 731: Doesn't this increase the duration of the test case too much?

zeebe/atomix/cluster/src/main/java/io/atomix/cluster/messaging/impl/HeartbeatSetupHandler.java
```diff
@@ -0,0 +1,205 @@
...
+      } catch (final IllegalStateException e) {
+        log.warn("Unable to decode heartbeat response, heartbeats are disabled.", e);
+        return false;
+      }
+    }
+  }
+
+  /**
+   * The server waits for a message to enable heartbeats on this connection. When such request is
+   * receive, he replies to the client in order to enable the heartbeats. In that case, it sets up a
```
filipecampos, line 141: typo: is received

filipecampos (approved):
Thanks Carlo for the patch. Just noticed a few typos. Apart from that LGTM!

zeebe/atomix/cluster/src/test/java/io/atomix/cluster/messaging/impl/NettyMessagingServiceTest.java
entangled90: thanks for spotting this, I bumped it when debugging :fearful:

zeebe/atomix/cluster/src/test/java/io/atomix/cluster/messaging/impl/NettyMessagingServiceTest.java
entangled90: what do you think of`shouldSendHeartbeatTimeoutFromClientToServer` ?

Issue #25935, opened 2024-12-11
Add E2E test for backups


entangled90:
A suite of E2e tests for backups will be added to the root `qa` project.
The tests will test the entire user journey for backups, including all apps.

> [!NOTE]
> - Zeebe's backup tests from zeebe/qa can used as a blueprint/ moved into the common module.
> - Zeebe must be configured with multiple partitions to ensure backup works for multiple partitions.
> - CamundaProcessTest assertions can be used to verify that the process are in expected state
> - Tests should be black boxed as much as possible, using queries and not break abstractions to simulate the end user journey

1. Tests must be able to be run on different deployment configurations:
   - zeebe + operate
   - zeebe + operate, tasklist & optimize
   - standalone camunda

2. The test must be able to test different block storages (only for standalone camunda):
  - GCP
  - Azure
  - S3
  - FS

3. The test must be able to test different databases (when all apps will use the `SearchClientProperties`): 
  - Elasticsearch
  - Opensearch

Tasks:
- [ ] Test all other deployment configurations
   - [ ] zeebe + operate
   - [ ] zeebe + operate, tasklist & optimize
   - [X] TestSimplecamundaApplication (zeebe, operate, tasklist)
- [x] Test multiple databases
  - [x] Elasticsearch
  - [x] Opensearch

ChrisKujawa:
@entangled90 is this done?

entangled90:
I checked what's been done in the checklist.
Currently only the webapps backup has been done, I would also add the Zeebe backup to the procedure so we can mimic what should be done in production exactly.
For the backup repository I don't think that we really need to test all combinations, as we would be testing ES/OS implementation of the backups, considering zeebe backup are already tested independently.

ChrisKujawa:
I think with https://github.com/camunda/camunda/pull/28318 we can mark this as done

PR #27029, opened 2025-01-16
Add OpenSearch testcase to BackupRestoreIT

entangled90:
## Description
Adds support for a test case using OpenSearch

- missing IN_PROGRESS enum case in opensearch `SnapshotState`
- basic test with OS has been added, but it was failing because the OS java client is broken on the snapshot API. Fortunately, this bug was just fixed, so I bumped the version and cleaned up some code that's not in use anymore in optimize (part of the code in optimize was also a for this OS client bug)

## Related issues
relates  #25935

### DIFF
--- a/operate/common/src/main/java/io/camunda/operate/conditions/DatabaseInfo.java
+++ b/operate/common/src/main/java/io/camunda/operate/conditions/DatabaseInfo.java
@@ -12,12 +12,13 @@ import static io.camunda.operate.conditions.DatabaseCondition.DATABASE_PROPERTY;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.DisposableBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.stereotype.Component;
 
 @Component("databaseInfo")
-public class DatabaseInfo implements ApplicationContextAware {
+public class DatabaseInfo implements ApplicationContextAware, DisposableBean {
 
   static final DatabaseType DEFAULT_DATABASE = DatabaseType.Elasticsearch;
   private static final Logger LOGGER = LoggerFactory.getLogger(DatabaseInfo.class);
@@ -59,6 +60,12 @@ public class DatabaseInfo implements ApplicationContextAware {
 
   public boolean isElasticsearchDb() {
     return isElasticsearch();
+  }
+
+  @Override
+  public void destroy() {
+    // unset current so it can be reinitialized
+    current = null;
   }
 
   @Override
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -78,6 +78,7 @@
     <version.netty>4.1.117.Final</version.netty>
     <version.objenesis>3.4</version.objenesis>
     <version.opensearch>2.14.0</version.opensearch>
+    <version.opensearch-java>2.19.0</version.opensearch-java>
     <version.opensearch.testcontainers>2.1.2</version.opensearch.testcontainers>
     <version.prometheus>0.16.0</version.prometheus>
     <version.protobuf>4.29.3</version.protobuf>
@@ -1351,7 +1352,7 @@
       <dependency>
         <groupId>org.opensearch.client</groupId>
         <artifactId>opensearch-java</artifactId>
-        <version>${version.opensearch}</version>
+        <version>${version.opensearch-java}</version>
       </dependency>
 
       <dependency>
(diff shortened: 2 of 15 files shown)


parent/pom.xml
```diff
@@ -77,7 +77,8 @@
     <version.msgpack>0.9.9</version.msgpack>
     <version.netty>4.1.117.Final</version.netty>
     <version.objenesis>3.4</version.objenesis>
-    <version.opensearch>2.14.0</version.opensearch>
+    <version.opensearch-java>2.19.0</version.opensearch-java>
+    <version.opensearch>2.18.0</version.opensearch>
```
entangled90, line 81: `opensearch-java` has published 2.19.0 with the fix, but `opensearch-rest-client` has not released 2.19.0 yet.

operate/common/src/main/java/io/camunda/operate/conditions/DatabaseInfo.java
```diff
@@ -61,6 +62,12 @@ public boolean isElasticsearchDb() {
     return isElasticsearch();
   }
 
+  @Override
+  public void destroy() {
+    // unset current so it can be reinitialized
+    current = null;
```
npepinpe, line 68: ❓️When do we reinitialize it? Only in tests?

parent/pom.xml
```diff
@@ -78,6 +78,7 @@
     <version.netty>4.1.117.Final</version.netty>
     <version.objenesis>3.4</version.objenesis>
     <version.opensearch>2.14.0</version.opensearch>
```
npepinpe, line 80: ❓️What is this version used for now then?

qa/integration-tests/src/test/java/io/camunda/it/backup/BackupDBClient.java
```diff
@@ -0,0 +1,33 @@
...
+  static BackupDBClient create(final String url, final DatabaseType databaseType)
+      throws IOException {
+    return switch (databaseType) {
+      case ELASTICSEARCH -> new ESDBClientBackup(url);
+      case OPENSEARCH -> new OSDBClientBackup(url);
+      default -> throw new IllegalStateException("Unsupported database type: " + databaseType);
+    };
+  }
+
+  void deleteAllIndices(final String indexPrefix) throws IOException;
```
npepinpe, line 30: 💭 We can put a todo to replace cleaning the DB with the purge functionality we have (once it's finished).

qa/integration-tests/src/test/java/io/camunda/it/backup/BackupDBClient.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package io.camunda.it.backup;
+
+import io.camunda.search.connect.configuration.DatabaseType;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+
+/** interface to abstract Elasticsearch/Opensearch clients for testing purposes */
+public interface BackupDBClient extends AutoCloseable {
```
npepinpe, line 16: 🔧I know it's test only, but I would document the contract in case we need to implement it again in the future.

npepinpe (commented):
Will finish later, but already some comments

operate/common/src/main/java/io/camunda/operate/conditions/DatabaseInfo.java
entangled90: I think so, this is probably the first test the reuses the application.

parent/pom.xml
entangled90: this is used for "opensearch-rest-client". "opensearch-java" has a direct dependency on it.
For example opensearch-java:2.19.0 depends on opensearch-rest-client:2.12.0

qa/util/src/main/java/io/camunda/qa/util/cluster/TestRestManagementClient.java
npepinpe, line 1: :question: What's the advantage of this over the usual actuator pattern (e.g. `BackupActuator`?

entangled90: My mistake, it's not in use. I forgot to remove it when rebasing the origina ES PR

entangled90:
In hindsight it's not really useful to keep the changes related to `TestStandaloneCamunda`, considering that it will be deprecated in favour of the MultiDb approach.
I remove all changes related to that class except for adding the schema manager to the scanned classes

npepinpe (approved):

entangled90:
JVM crashed in java-unit-tests

entangled90:
JVM crashed in unit-tests

PR #27034, opened 2025-01-16
fix: avoid unnecessary caching of incidents

sdorokhova:
## Description

We can safely write process definition related fields, as we only process CREATED and MIGRATED records and they both will have the fields filled in.

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/IncidentHandler.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/IncidentHandler.java
@@ -37,7 +37,6 @@ import org.slf4j.LoggerFactory;
 public class IncidentHandler implements ExportHandler<IncidentEntity, IncidentRecordValue> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(IncidentHandler.class);
-  private final Map<String, Record<IncidentRecordValue>> recordsMap = new HashMap<>();
   private final String indexName;
   private final ExporterEntityCache<Long, CachedProcessEntity> processCache;
   private final IncidentNotifier incidentNotifier;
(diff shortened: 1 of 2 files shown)


mustafadagher (approved):
Looks good, thanks @sdorokhova ! 🚀 

As a result of this, we can't now separate migrated from created incidents for incident notifications, but that's to be addressed separately.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/IncidentHandler.java
```diff
@@ -107,19 +105,14 @@ public void updateEntity(final Record<IncidentRecordValue> record, final Inciden
         .setTenantId(ExporterUtil.tenantOrDefault(recordValue.getTenantId()));
 
     entity.setTreePath(buildTreePath(record));
-
-    recordsMap.put(entity.getId(), record);
   }
 
   @Override
   public void flush(final IncidentEntity entity, final BatchRequest batchRequest) {
```
mustafadagher, line 111: let's fix the related tests in `IncidentHandlerTest`

mustafadagher:
@sdorokhova shall we rebase and merge this?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/IncidentHandler.java
```diff
@@ -114,26 +113,25 @@ public void updateEntity(final Record<IncidentRecordValue> record, final Inciden
...
     entity.setTreePath(buildTreePath(record));
 
-    recordsMap.put(entity.getId(), record);
-  }
-
-  @Override
-  public void flush(final IncidentEntity entity, final BatchRequest batchRequest) {
-    final String id = entity.getId();
-    final Record<IncidentRecordValue> record = recordsMap.get(id);
     final Intent intent = (record == null) ? null : record.getIntent();
```
mustafadagher, line 116: this check for a null record and null intent is obsolete now

sdorokhova: @mustafadagher I now moved that from `flush` to `updateEntity` method including incident notification call. Do you think it's OK?

mustafadagher: I think so. It's a bit different than where we did it in operate. but I don't think it matters a lot

PR #27045, opened 2025-01-16
refactor: use tenant id instead of tenant key

### LINKED ISSUE #27020: Modify tenant handler to store/update tenants by the tenantId as ID instead of key

### LINKED ISSUE #27021: Modify tenant mapping to use the tenantId as ID instead of key

Ben-Sheppard:
## Description

This PR refactors the exporters to use the tenantId as the primary id instead of the tenant key. I have tested locally and I'm able to observe the correct key being set during creation and update, along with deletes successfully working too.

To note: I have not done anything with the tenant members yet, that is going to be tackled in a separate PR, I'm aware that means the memberships on the RDBMS side are still using the tenant key, but I want to avoid PRs becoming too large.

closes #27020
closes #27021

### DIFF
--- a/db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
+++ b/db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
@@ -343,10 +343,10 @@
 
   <changeSet id="create_tenant_table" author="cthiel">
     <createTable tableName="TENANT">
-      <column name="TENANT_KEY" type="BIGINT">
+      <column name="TENANT_ID" type="VARCHAR(255)">
         <constraints primaryKey="true"/>
       </column>
-      <column name="TENANT_ID" type="VARCHAR(255)" />
+      <column name="TENANT_KEY" type="BIGINT"/>
       <column name="NAME" type="VARCHAR(255)" />
     </createTable>
   </changeSet>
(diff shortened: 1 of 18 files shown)


db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
```diff
@@ -343,10 +343,10 @@
 
   <changeSet id="create_tenant_table" author="cthiel">
     <createTable tableName="TENANT">
-      <column name="TENANT_KEY" type="BIGINT">
+      <column name="TENANT_ID" type="VARCHAR(255)">
         <constraints primaryKey="true"/>
       </column>
-      <column name="TENANT_ID" type="VARCHAR(255)" />
+      <column name="TENANT_KEY" type="BIGINT"/>
```
lenaschoenburg, line 349: :question: Is this still being used anywhere?

lenaschoenburg (approved):

db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
Ben-Sheppard: Good question, there is still the concept of a `tenant key` for the memberships (I am planning to follow up with a separate PR about that), for this field on the table though its likely something that we can remove as part of that clean up.

qa/integration-tests/src/test/java/io/camunda/it/operate/OperateMultiTenancyIT.java
```diff
@@ -76,6 +77,7 @@ public void beforeEach() {
   }
 
   @Test
+  @Disabled
```
lenaschoenburg, line 80: :thought_balloon: Let's make sure we at least have a note in an issue somewhere to re-enable it. I don't think I would remember :sweat_smile:

Ben-Sheppard: Raised an issue https://github.com/camunda/camunda/issues/27160 and linked it to the epic :)

PR #27047, opened 2025-01-16
ci: fix variable substitution for e2e tests

### LINKED ISSUE #27042: E2E tests are broken on `main`
The workflow is failing since three months already: https://github.com/camunda/camunda/actions/workflows/zeebe-e2e-testbench.yaml?query=branch%3Amain

We should
1. Fix the bug
2. Ensure we get notified when this workflow fails
3. Fix our release process to make sure we don't release without a successful E2E run

lenaschoenburg:
The JSON variables need to be provided un-escaped because the testbench workflow no longer does this for us.

closes #27042

### DIFF
--- a/.github/workflows/zeebe-e2e-testbench.yaml
+++ b/.github/workflows/zeebe-e2e-testbench.yaml
@@ -78,23 +78,23 @@ jobs:
       processId: e2e_testbench_protocol
       variables: >
         {
-        \"zeebeImage\":\"$IMAGE\",
-        \"generationTemplate\":\"${{ inputs.generation || 'Zeebe SNAPSHOT' }}\",
-        \"channel\": \"Internal Dev\",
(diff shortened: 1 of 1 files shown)


nicpuppa:
@lenaschoenburg does it work if in the variables input there are no `\` ? This workflow is used in two places, Zeebe E2E Tests and Zeebe QA Testbench. When I made this changes I focused only on the QA Testbench, and in this file `zeebe-qa-testbench.yaml` the variables are passed without the backslash

lenaschoenburg:
@nicpuppa Yeah that's much better :+1: 
Test run here: https://github.com/camunda/camunda/actions/runs/12825691800/job/35764145732

lenaschoenburg:
@nicpuppa works :tada:

nicpuppa (approved):

PR #27053, opened 2025-01-16
fix: use global prefix in TaskTemplate

sdorokhova:
## Description

* use global prefix in TaskTemplate: this helps when running E2E tests with `make start-e2e`, as we define `e2eoperate` index prefix there
* take into account JOIN field when querying for user tasks: since we joined index with Tasklist task index, we now store hierarchies of documents using "join" field type (similar to `list-view` index). This means that one index stores not only tasks, but also process instances and variables. Therefore to make sure that we search for tasks and not variables e.g. we should put additional filter on `join` field.


kristinkomschow:
Are there any tests that check if tasks get correctly read from the new index?

PR #27062, opened 2025-01-17
feat: add request parameter string key support

danielkelemen:
## Description

Add request KeyParameter OpenAPI type
Add request string & number header support
Add request string object types

## Related issues

closes https://github.com/camunda/camunda/issues/26761

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -671,6 +671,12 @@ paths:
                   processDefinitionId: "1234-5678"
                   version: 1
                   variables: {}
+          application/vnd.camunda.api.keys.number+json:
+            schema:
+              $ref: "#/components/schemas/CreateProcessInstanceRequest"
+          application/vnd.camunda.api.keys.string+json:
+            schema:
+              $ref: "#/components/schemas/CreateProcessInstanceRequestBody"
       responses:
         "200":
           description: The process instance was created.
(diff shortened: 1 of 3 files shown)


danielkelemen:
## TODOs
- [ ] rename `StringKeys`.

tmetzke:
Let's revisit when https://github.com/camunda/camunda/pull/26995 is settled.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
```diff
@@ -133,7 +133,7 @@ public static Either<ProblemDetail, DecisionInstanceQuery> toDecisionInstanceQue
     return buildSearchQuery(filter, sort, page, SearchQueryBuilders::decisionInstanceSearchQuery);
   }
 
-  private static DecisionInstanceFilter toDecisionInstanceFilter(
+  private static io.camunda.service.search.filter.DecisionInstanceFilter toDecisionInstanceFilter(
```
tmetzke, line 136: ❓ Do we need these explicit imports? Do we use the same class from the REST package as well?

tmetzke (approved):
👍 Nice, looks good to me 🚀 
🔧 Very minor question left, not blocking though

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java
danielkelemen: No, we use the number key variants which are called differently.
It's only needed due to the `*` imports. I'll fix that.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -156,8 +155,7 @@ paths:
           required: true
           description: The key of the job.
           schema:
-            type: integer
-            format: int64
+            type: string
```
tmetzke, line 158: ❓ Just realizing this now: should we stick to numbers in 8.6 at least to not introduce a breaking change there? Fine for 8.7 to move to string, of course, as discussed (cannot find anymore where we discussed this 😆)

PR #27064, opened 2025-01-17
fix: in NettyMessagingServiceTest don't consider udp sockets already opened

entangled90:
## Description
This test can be flaky if in the machine there are other programs which have already opened a few UDP sockets.

Ignore the sockets that were already opened in the machine before the test started


lenaschoenburg (approved):
Thanks!

PR #27095, opened 2025-01-17
feat: Allow Camunda Exporter index descriptors to decide how their persistence errors are handled

### LINKED ISSUE #27090: Allow Camunda Exporter index descriptors to decide how their persistence errors are handled
Usually errors during persistence of exporter entities to ES/OS are propagated, and the export for the error records is retried. But with #26093, a user could pass a wrong/non-existent operation-reference, which will fail the operation update, and will result in infinite retries.

To overcome the scenario above, we want to be able to mark some indices that they should fail silently, typically in this case is the `operate-operation` index.

mustafadagher:
## Description
- Added ErrorHandlingBehavior interfaces
- IndexDescriptor interface implements ErrorHandlingBehavior
- Operate OperationTemplate implements ErrorSwallowing
- Mark all current indices and templates as ErrorThrowing
- refactor: handlers receive their respective IndexDescriptor as a constructor parameter
- BatchRequest can register how to handle errors for certain indices, and use the behavior for persistence errors
- pass error handling behavior to batchRequest when flushing entities to be exported

## Related issues

closes #27090

### DIFF
--- a/migration/process-migration/src/main/java/io/camunda/migration/process/adapter/MigrationRepositoryIndex.java
+++ b/migration/process-migration/src/main/java/io/camunda/migration/process/adapter/MigrationRepositoryIndex.java
@@ -7,10 +7,11 @@
  */
 package io.camunda.migration.process.adapter;
 
+import io.camunda.webapps.schema.descriptors.errorHandling.ErrorThrowing;
 import io.camunda.webapps.schema.descriptors.operate.OperateIndexDescriptor;
 
 /* Fork of Operate's MigrationRepositoryIndex descriptor for correct prefixing */
-public class MigrationRepositoryIndex extends OperateIndexDescriptor {
+public class MigrationRepositoryIndex extends OperateIndexDescriptor implements ErrorThrowing {
 
   public static final String INDEX_NAME = "migration-steps-repository";
   public static final String VERSION = "1.1.0";
--- a/operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/util/TestDynamicIndex.java
+++ b/operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/util/TestDynamicIndex.java
@@ -7,9 +7,10 @@
  */
 package io.camunda.operate.schema.util;
 
+import io.camunda.webapps.schema.descriptors.errorHandling.ErrorThrowing;
 import io.camunda.webapps.schema.descriptors.operate.OperateIndexDescriptor;
 
-public class TestDynamicIndex extends OperateIndexDescriptor {
+public class TestDynamicIndex extends OperateIndexDescriptor implements ErrorThrowing {
 
   public TestDynamicIndex(final String indexPrefix, final boolean isElasticsearch) {
     super(indexPrefix, isElasticsearch);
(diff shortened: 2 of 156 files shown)


migration/process-migration/src/main/java/io/camunda/migration/process/adapter/MigrationRepositoryIndex.java
```diff
@@ -7,10 +7,11 @@
  */
 package io.camunda.migration.process.adapter;
 
+import io.camunda.webapps.schema.descriptors.errorHandling.ErrorThrowing;
 import io.camunda.webapps.schema.descriptors.operate.OperateIndexDescriptor;
 
 /* Fork of Operate's MigrationRepositoryIndex descriptor for correct prefixing */
-public class MigrationRepositoryIndex extends OperateIndexDescriptor {
+public class MigrationRepositoryIndex extends OperateIndexDescriptor implements ErrorThrowing {
```
ChrisKujawa, line 14: 💭 I'm not sure whether I would expect this on the index classes tbh. I thought we can do this on the handlers.

ChrisKujawa (changes requested):
I think syntactically I feel this is not correct. The index descritor is just a descriptor of the index, here we shouldn't have some more runtine related like whether errors are handled or not. This is more an Exporter internal, the other components do not care about.

I would propose we do the following:

1. The BatchRequest#execute gets a new parameter. This is an error handler
2. The error handler accepts the index, id, errorType/description, exception? (and more if necessary)
3. The id and index allow already to uniquely identifier an handler
4. The handler get a new method onError, to implement error handling logic that can cover cases based on their record handling
5. Per default this is simply throwing
6. For the AbstractOperation handler we implement a swallowing, if the error type matches of not existing entity. 
7. This approach should be easily extensible also for other handlers, if necessary. 
8. Either we make sure that only one handler is called per error (tbh I think entity id and index, should be enough)  - otherwise we break, or we validate this somehow with tests - or we document that this might be the case; But right now this is actually not a problem in my opinion.

mustafadagher:
> I think syntactically I feel this is not correct. The index descritor is just a descriptor of the index, here we shouldn't have some more runtine related like whether errors are handled or not. This is more an Exporter internal, the other components do not care about.
> 
> I would propose we do the following:
> 
> 1. The BatchRequest#execute gets a new parameter. This is an error handler
> 2. The error handler accepts the index, id, errorType/description, exception? (and more if necessary)
> 3. The id and index allow already to uniquely identifier an handler
> 4. The handler get a new method onError, to implement error handling logic that can cover cases based on their record handling
> 5. Per default this is simply throwing
> 6. For the AbstractOperation handler we implement a swallowing, if the error type matches of not existing entity. 
> 7. This approach should be easily extensible also for other handlers, if necessary. 
> 8. Either we make sure that only one handler is called per error (tbh I think entity id and index, should be enough)  - otherwise we break, or we validate this somehow with tests - or we document that this might be the case; But right now this is actually not a problem in my opinion.

Thanks for the comments Chris, the approach you described is already implemented in another draft, please take a look at my comments there too.

For what its worth, I don't like both approaches and I was still thinking about something simpler

Issue #26896, opened 2025-01-14
[EPIC]: Create new foundation for platform wide integration test


ChrisKujawa:
## Current challenge

At the time of writing, creating a Camunda platform specific integration test [on the root level](https://github.com/camunda/camunda/tree/main/qa) is not an easy task. There are several challenges engineers are facing:

- We have too many different approaches on how to write such tests. 
  - Using SpringBoot-related application classes for tests: TestStandaloneCamunda, TestStandaloneBroker, etc.
  - Using test extensions
- Some of the approaches have more downsides then others, like bootstrapping multiple brokers, multiple secondary storage, etc.
- There is no consistency and clarity on how to write such integration tests correctly
- We are even not able to support all backends fully right now, we often can only test against ES (and maybe RDBMS). 
  - We have to support OS and AWS OS as well.

![Image](https://github.com/user-attachments/assets/ae77958d-dd53-464d-9689-6c6701047b8c)

Right now we bootstrap the secondary storage per test class, or even per test. This allows us to have data isolation, with the cost of performance (bootstrapping containers, etc.)

We (@tmetzke @npepinpe and myself) have discussed different approaches, test containers (like we have now), test extensions, spring tests, and externalizing the dependency. We have seen for example some limitations with the previous approaches, like junit 5 does not allow parameterizing at the class level.

## Proposed solution

Instead of bootstrapping our secondary storage per test class, via test containers, we are externalizing them. This means we move the complexity one layer up to the IT job test execution, into the CI. With this we will go in a similar direction we have in C7 and had with the web application ITs before.

![Image](https://github.com/user-attachments/assets/ce254745-d95c-42cc-89d1-1dc4890d9b71)

* We should have a per-database CI job to run the same set of integration tests.
   * Such CI job should create the database as a container, before executing the integration tests
* We should be able to somehow identify the corresponding database, fex. via env vars
* Test applications are configured correctly based on connection/database type with specific defaults (like port, etc.)
   * We likely have some junit extension to to perform the wiring based on some simple setting (e.g. env var or property or some easy toggle)
* We still want to make sure to support and accept the TestApplication to be passed into the extension, so users can configure separately their applications for specific tests 
* We should migrate existing root IT tests, to create some consistency
* We should provide some guidance, via Javadoc and guidelines (docs/readme) on how to write the integration tests correctly

### Pros:

- possibly faster by removing the overhead of starting ES/OS/RDBMS (possibly since we may want to support different vendors and so on)
[...]

PR #27106, opened 2025-01-17
Add new IT job for externalize ES database

ChrisKujawa:
## Description

* Add a new IT job to run with the ES service container
* Service containers are hosted/maintained by GH
* IT is run against external ES, reducing set-up time for individual tests and sharing ES instances
* This setup allows to add more similar jobs, to run tests parallel against external different database types

## See in action

## Checklist

- [ ] for CI changes:

## Related issues

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -292,7 +292,7 @@ jobs:
           -D skipUTs -D skipChecks
           -D failsafe.rerunFailingTestsCount=3 -D flaky.test.reportDir=failsafe-reports
           -P parallel-tests,extract-flaky-tests,skipFrontendBuild,multi-db-test
-          -D camunda.it.database.type=es
+          -D test.integration.camunda.database.type=es
           -pl 'qa/integration-tests'
           verify
           | tee "${BUILD_OUTPUT_FILE_PATH}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,6 +227,96 @@ jobs:
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
 
+  elasticsearch-integration-tests:
+    name: "[IT] Elasticsearch"
+    timeout-minutes: 25
+    permissions: {}  # GITHUB_TOKEN unused in this job
+    runs-on: gcp-perf-core-8-default
+    env:
+      ZEEBE_TEST_DOCKER_IMAGE: localhost:5000/camunda/zeebe:current-test
+    services:
+      # Set up an elastic search service container
+      # https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/about-service-containers
+      elasticsearch:
+        image: docker.elastic.co/elasticsearch/elasticsearch:8.17.0
+        # Exposing the ports allow to be accessible via http://localhost:9200 in the runner
+        ports:
+          - 9200:9200
+        env:
+          # Disabling the security to make sure we can access without tls in our tests
+          # https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#general-security-settings
+          xpack.security.enabled: "false"
+          # We run a single node for our integration tests
+          discovery.type: "single-node"
+          # Disabling to make sure we can delete indices via wildcards
+          # https://www.elastic.co/guide/en/elasticsearch/reference/current/index-management-settings.html
+          action.destructive_requires_name: "false"
+      # Set up a registry service container to host our custom docker image
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-build
+        with:
+          maven-cache-key-modifier: it-elasticsearch
+          vault-address: ${{ secrets.VAULT_ADDR }}
+          vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
+          vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
+      # Build the platform before running the integration tests; skipping frontend
+      - uses: ./.github/actions/build-zeebe
+        id: build-zeebe
+        with:
+          maven-extra-args: -T1C -PskipFrontendBuild
+      - name: Create build output log file
+        run: echo "BUILD_OUTPUT_FILE_PATH=$(mktemp)" >> $GITHUB_ENV
+      # To fail early if set up with service container doesn't work and ES is not accessible
+      - name: Validate connection with ES
+        run: curl http://localhost:9200
+      # Run specific integration tests with ES service container
+      # During the migration of tests we run only specific package, where classes moved into
+      #
+      # To communicate to the IT what database to expect and run against:
+      # -Dcamunda.it.database.type=es
+      #
+      - name: Run integration test with externalized ES
+        run: >
+          ./mvnw -B -T 2 --no-snapshot-updates
+          -D forkCount=7
+          -D maven.javadoc.skip=true
+          -D skipUTs -D skipChecks
+          -D failsafe.rerunFailingTestsCount=3 -D flaky.test.reportDir=failsafe-reports
+          -P parallel-tests,extract-flaky-tests,skipFrontendBuild
+          -Dit.test=io/camunda/v2/**
+          -Dcamunda.it.database.type=es
+          -pl 'qa/integration-tests'
+          verify
+          | tee "${BUILD_OUTPUT_FILE_PATH}"
+      - name: Analyze Test Runs
+        id: analyze-test-run
+        if: always()
+        uses: ./.github/actions/analyze-test-runs
+        with:
+          buildOutputFilePath: ${{ env.BUILD_OUTPUT_FILE_PATH }}
+      - name: Upload test artifacts
+        uses: ./.github/actions/collect-test-artifacts
+        if: ${{ failure() || cancelled() || steps.analyze-test-run.outputs.flakyTests != '' }}
+        with:
+          name: "[IT] Elasticsearch"
+      - name: Observe build status
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          job_name: "integration-tests/elasticsearch"
+          build_status: ${{ job.status }}
+          user_reason: ${{ (steps.analyze-test-run.outputs.flakyTests != '') && 'flaky-tests' || '' }}
+          user_description: ${{ steps.analyze-test-run.outputs.flakyTests }}
+          secret_vault_secretId: ${{ secrets.VAULT_SECRET_ID }}
+          secret_vault_address: ${{ secrets.VAULT_ADDR }}
+          secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
+
   integration-tests:
     if: needs.detect-changes.outputs.java-code-changes == 'true'
     name: "[IT] ${{ matrix.name }}"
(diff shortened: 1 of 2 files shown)


cmur2:
Looks promising - I'll have a closer look on Monday!

cmur2 (commented):
In Slack you discussed that debugging of GHA service containers and resource allocation is hard(er) to do - would you from your experience recommend using more GHA service containers also outside this POC, or rather use `docker-compose`?

Left a couple of comments on the GHA pipeline code

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+    timeout-minutes: 25
+    permissions: {}  # GITHUB_TOKEN unused in this job
+    runs-on: gcp-perf-core-8-default
+    env:
+      ZEEBE_TEST_DOCKER_IMAGE: localhost:5000/camunda/zeebe:current-test
+    services:
+      # Set up an elastic search service container
+      # https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/about-service-containers
+      elasticsearch:
+        image: docker.elastic.co/elasticsearch/elasticsearch:8.17.0
```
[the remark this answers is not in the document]
ChrisKujawa: Can we somehow check whether it will be detected by renovate?

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+      # Build the platform before running the integration tests; skipping frontend
+      - uses: ./.github/actions/build-zeebe
+        id: build-zeebe
+        with:
+          maven-extra-args: -T1C -PskipFrontendBuild
+      - name: Create build output log file
+        run: echo "BUILD_OUTPUT_FILE_PATH=$(mktemp)" >> $GITHUB_ENV
+      # To fail early if set up with service container doesn't work and ES is not accessible
+      - name: Validate connection with ES
+        run: curl http://localhost:9200
```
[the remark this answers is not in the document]
ChrisKujawa: It needs some time, which is why I did it after. It took around 10 seconds to be up

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+        uses: ./.github/actions/collect-test-artifacts
+        if: ${{ failure() || cancelled() || steps.analyze-test-run.outputs.flakyTests != '' }}
+        with:
+          name: "[IT] Elasticsearch"
+      - name: Observe build status
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/observe-build-status
+        with:
+          job_name: "integration-tests/elasticsearch"
```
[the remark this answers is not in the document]
ChrisKujawa: thanks will

ChrisKujawa:
> In Slack you discussed that debugging of GHA service containers and resource allocation is hard(er) to do - would you from your experience recommend using more GHA service containers also outside this POC, or rather use `docker-compose`?
> 

Interesting question. 

Regarding debugging, I guess we or Nicolas meant looking into container what has happened, no actual debugging (like with breakpoints). For this we can get the logs in failure cases for example see here https://github.com/orgs/community/discussions/26282#discussioncomment-3251184

I think docker-compose is in general cool as well if we want to set up the same also locally, so a engineer could simple run docker-compose to get the same environment. I guess this is a plus. 

I think a benefit for the service container is that it is not running in the runner (afaik?) it is a separate container (somewhere?), reducing the needed resources vs running the docker compose inside the same runner. 

I would start with the service container, see how far we come and whether we ran into issues. If so, we can re-iterate.

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
 
+  elasticsearch-integration-tests:
+    if: needs.detect-changes.outputs.java-code-changes == 'true'
+    needs: [ detect-changes ]
+    name: "[IT] Elasticsearch"
+    timeout-minutes: 25
```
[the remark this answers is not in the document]
cmur2: Since this is a new CI job I'd appreciate to [stick to the best practice timings](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria) and 25 minutes is definitely above of what we should tolerate for new CI jobs.

.github/workflows/ci.yml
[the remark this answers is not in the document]
cmur2: It might be mentioned in the Renovate documentation/issues or we have to reproduce it in a dedicated repository since Renovate doesn't check PRs - or after merge of this PR as a follow-up if we are sure it doesn't get lost.

cmur2:
> I think a benefit for the service container is that it is not running in the runner (afaik?) it is a separate container (somewhere?), reducing the needed resources vs running the docker compose inside the same runner.

I'm pretty sure it is running on the same runner and just a fancy way to configure the local Docker daemon, so it will share resources with the build job.

.github/workflows/ci.yml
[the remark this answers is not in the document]
cmur2: Apparently it is supported: https://github.com/renovatebot/renovate/issues/16351

.github/workflows/ci.yml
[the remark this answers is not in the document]
ChrisKujawa: fine by me

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+      - name: Create build output log file
+        run: echo "BUILD_OUTPUT_FILE_PATH=$(mktemp)" >> $GITHUB_ENV
+      # To fail early if set up with service container doesn't work and ES is not accessible
+      - name: Validate connection with ES
+        run: curl http://localhost:9200
+      # Run specific integration tests with ES service container
+      # During the migration of tests we run only specific package, where classes moved into
+      #
+      # To communicate to the IT what database to expect and run against:
+      # -Dcamunda.it.database.type=es
```
npepinpe, line 283: :wrench: This means we're reserving the `camunda.it` namespace for configuration. Probably OK, but also maybe not. I would propose something that has no risk of clashing with our configuration, so basically not starting with `camunda`, but more like `test.camunda.`

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+      #
+      - name: Run integration test with externalized ES
+        run: >
+          ./mvnw -B -T 2 --no-snapshot-updates
+          -D forkCount=7
+          -D maven.javadoc.skip=true
+          -D skipUTs -D skipChecks
+          -D failsafe.rerunFailingTestsCount=3 -D flaky.test.reportDir=failsafe-reports
+          -P parallel-tests,extract-flaky-tests,skipFrontendBuild
+          -Dit.test=io/camunda/v2/**
```
npepinpe, line 293: :question: What does this do? Selecting the right files? If that's the case, then :x: let's use junit tagging instead, that way of selecting files which are included is more flexible, and not hidden in the CI job but instead visible directly in the code.

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+          -Dcamunda.it.database.type=es
+          -Dfailsafe.failIfNoSpecifiedTests=false
+          -pl 'qa/integration-tests'
```
npepinpe, line 296: :upside_down_face: I can imagine in the future we may want to run tests against multiple DBs in different modules. Does it hurt to just use tagging and a meta annotation (e.g. `CamundaMultiDBTest` or something? Not a blocker right now, can be an improvement later.

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+      # Set up a registry service container to host our custom docker image
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-build
+        with:
+          maven-cache-key-modifier: it-elasticsearch
```
npepinpe, line 265: :question: @cmur2 - I know we did some cache changes, I didn't follow it all, but just want to flag and make sure this is correct - so if you can please confirm. Just in case you would have glossed over (not saying you did, just want to make sure).

npepinpe (approved):
No major blocker, but please address all comments before merging (and ofc you can ask for another review if you want).

.github/workflows/ci.yml
```diff
@@ -227,6 +227,99 @@ jobs:
...
+    name: "[IT] Elasticsearch"
+    timeout-minutes: 25
+    permissions: {}  # GITHUB_TOKEN unused in this job
+    runs-on: gcp-perf-core-8-default
+    env:
+      ZEEBE_TEST_DOCKER_IMAGE: localhost:5000/camunda/zeebe:current-test
+    services:
+      # Set up an elastic search service container
+      # https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/about-service-containers
+      elasticsearch:
```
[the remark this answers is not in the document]
ChrisKujawa: Thanks i was alao thinking whether we could use something like that, but when i looked at the examples where not really clear to me how customizable they are

.github/workflows/ci.yml
ChrisKujawa: That is an intermediate state. Yes i wanted to run here only test that are migrated. After we migrated all i wanted to remove this again.

.github/workflows/ci.yml
ChrisKujawa: That would only work if they use the same extensions or way to detect. Right now this is out of scope.

.github/workflows/ci.yml
ChrisKujawa: Fine by me

.github/workflows/ci.yml
ChrisKujawa: I can add tags thanks 👍

Can we make sure that the tests are not executed by the other IT job? I guess `-DexcludedGroups=externalDb`

.github/workflows/ci.yml
cmur2: This is in line with other usages of `setup-build` in the `ci.yml` 👍🏼 

> (not saying you did, just want to make sure).

All good, no worries 😊

.github/workflows/ci.yml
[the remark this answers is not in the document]
ChrisKujawa:
```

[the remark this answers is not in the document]
Comment (ChrisKujawa) [reply .github/workflows/ci.yml:242]:
```
        options: >-
          --health-cmd "curl --silent --fail http://localhost:9200/_cluster/health"
          --health-interval 5s
          --health-timeout 5s
          --health-retries 10

```

Works https://github.com/camunda/camunda/actions/runs/12886542433/job/35927356462#step:3:115

.github/workflows/ci.yml
ChrisKujawa: I tried it with adding a [tag](https://github.com/camunda/camunda/pull/26852/commits/9e5d36cd9f7f479ad719beeca1c1eb64583d4c16) and running [maven with groups flag](https://github.com/camunda/camunda/pull/26852/commits/5b6334e14a3bdd50bc7c81735ef01bc8a9ecbfce), but it doesn't seem to work. It executed other tests as well

https://github.com/camunda/camunda/actions/runs/12886542433/job/35927356462#step:10:1

If you don't have a direct solution/hint I will go ahead and merge it first and find out this separately.

.github/workflows/ci.yml
ChrisKujawa: Additionally it also reduces the build/execution. I will keep it for now.

.github/workflows/ci.yml
ChrisKujawa: Got it working with a profile. The issue seem to be that groups have been overwritten, before in build step

ChrisKujawa: https://github.com/camunda/camunda/pull/26852/commits/77de90be131f13209a6a0e88338c2f067e624a4d

Issue #25686, opened 2024-12-05
Server ports opened by Optimize


buccarel:
### Context

Currently, Optimize is opening 2 ports on startup; if default settings are used:

* port 8090 is open (HTTP)
* port 8091 is open (HTTPs, optionally configurable as HTTP/2)

Moving forward with the single-application project, we need to understand whether or not this behavior shall be kept or changed in favor of opening one port only. Part of this understanding could be viewed as an alignment to what the other applications of the stack are doing.

It is worth noticing that, in some cases, it could be recommended to leave the HTTPs capability to a wrapper around an application that only supports HTTP. This could be the case when a reverse proxy is viable or when cloud services are used to host the application. Within the context of this task, we should understand if this could be a viable option or if this approach is already being used by other applications.

### Acceptance Criteria

[ ] decide whether we want to open only 1 port at once
[ ] decide whether optimize should handle the HTTPs configuration explicitly, instead of delegating it to a HTTPs reverse proxy or cloud environment

### Links

* Part of https://github.com/camunda/product-hub/issues/2289

### Breakdown

```[tasklist]
### Pull Requests
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [ ] The change is implemented as described on all target environments/versions/modes
- [ ] The documentation changes are as expected

### Completion

- [ ] All Review stages are successfully completed
- [ ] All associated PRs are merged to the main branch(es) and stable branches
- [ ] The correct version labels are applied to the issue

buccarel:
Currently, Optimize's default port is 8090 (HTTP).
Optimize is also opening port 8091, which is HTTPs.

This behavior is different than what happens with the stack of the single application.

In particular, once Optimize starts as single application, its HTTP port is supposed to be 8080, with the HTTP protocol. This is what other applications do. The same thing should happen when Optimize is launched with the standalone-optimize profile, from within the single application context.

This makes me think that:
1) the default ports should change (documentation and config change)
[...]

PR #27107, opened 2025-01-17
build: remove Optimize's web server port customization

buccarel:
## Description

As Optimize is going to become part of the single application, the web server customization it is installing need to be removed. With this PR, we are still supporting the legacy default port because, currently, integration tests are being executing using `optimize-distro`. Once they are also migrated to the single JAR distribution, the support for the legacy default port can also be removed.

In addition to that, Optimize is currently opening an extra port, configured with SSL. Even that behavior is supposed to change, leaving the entire configuration in the hands of the common web server used in the single application instead.

## Checklist

- [ ] for CI changes:

## Related issues

* part of #25686


matthewBearCamunda (approved):
The code LGTM and I will be approving

I'm curious though, with the removal of the methods that set up http & https, does https still work for Optimize?

Issue #26043, opened 2024-12-12
Implement `Exporter#purge` for the Camunda exporter


npepinpe:
## Description

Since CPT is the first main use case for purging, and it will use only the Camunda exporter, we will first implement `CamundaExporter#purge`.

The first step is figuring out what data we need to delete. Data that we clearly need to delete:

- All archived indices, e.g. operate-list-view-8.3.0-_2023-01-01
- All documents in the so-called live indices, e.g. operate-list-view-8.3.0-_

> [!Warning]
> Any manual changes to the schema will persist through a purge operation at the moment; this is considered out of scope for now.

## Acceptance criteria

- [ ] There are no more archived indices after a purge
- [ ] The live indices still exist but are empty
- [ ] The schema is the same as if the cluster was just recreated
- [ ] Any data that was _not_ managed by C8 is still available.

npepinpe:
We've decided to change the scope here, and will only purge the data, not the schema. There's a high overhead to deleting and recreating the schema, and within the scope of the CPT integration, this is likely not necessary. So we would propose that we focus on only deleting the data produced by the exporter, and ignore other data (e.g. batch operations written directly by Operate) as such things will be moved to Zeebe before the next release, and not deleting the schema.

This would reduce complexity and allow us to wrap up faster.

```
## Description

Since CPT is the first main use case for purging, and it will use only the Camunda exporter, we will first implement `CamundaExporter#purge`.

The first step is figuring out what data we need to delete. Data that we clearly need to delete:

- All indices managed by C8
- All index templates managed by C8
- All component templates managed by C8
- All ILM policies managed by C8
- All index aliases managed by C8

Maybe some other things?

## Acceptance criteria

- [ ] All data that must be deleted is discovered and defined
- [ ] All data that must be deleted is deleted from ES after `CamundaExporter#purge` returns.
- [ ] Any data that was _not_ managed by C8 is still available.
```

PR #27109, opened 2025-01-17
New external DB integration test junit extension

ChrisKujawa:
## Description

### Preparation work for the new extension

> [!Important]
>
> We could split the preparation more into smaller PRs if necessary.

 * Add OS Schemamanager
   * Default SchemaManager is excluded for test profiles
   * To run IT against OpenSearch we need a proxy schema manager, enabling them for the IT
   * This is already done for ES but not for OS
 * Support prefix in IT
   * Descriptors configuration doesn't set prefixes before
   * It was not necessary before for IT, as ES/OS were separated
   * It will become mandatory with new IT
 * New test application
    * Create a copy of TestStandaloneCamunda; _to not disrupt other tests and workflows_
    * Adjust to our needs, that means:
      * Support for ES configuration
      * Support for OS configuration
      * Support for RDBMS H2 configuration (not yet fully working)
    * Removed some not yet needed code; that have been part of standalone
     camunda
    * for supported database, necessary configurations are set (ES/OS)
      * RDBMS was not working yet
    * Validated test with both set ups ES and OS
 
 
### Changes for the actual new extension

 * Copy BrokerITInvocationProvider, adjust to create a new extension
 * Extension can be registered with ExtendWith
   * It will resolve parameters and static fields, to fill necessary CamundaClient instances.
   * Allows to migrate easily existing test cases.
   * The extension will make sure that the Platform is started, but not secondary storage.
   * It detects (right now defaults to ES) the secondary storage, via property and then sets the
     correct connection details. Here we need to use good defaults.
   * It will not set up ES, OS or RDBMS. It will expect they are started already. Allows to
parallize tests, and reuse the same instance.
  * Test data will be prefixed with test class names.

## See in action

## Related issues

### DIFF
--- a/qa/integration-tests/pom.xml
+++ b/qa/integration-tests/pom.xml
@@ -378,6 +378,11 @@
       <artifactId>tasklist-webapp</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.platform</groupId>
+      <artifactId>junit-platform-commons</artifactId>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
   <repositories>
--- a/qa/integration-tests/pom.xml
+++ b/qa/integration-tests/pom.xml
@@ -383,6 +383,16 @@
       <artifactId>junit-platform-commons</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>operate-common</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>tasklist-common</artifactId>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
   <repositories>
--- a/qa/integration-tests/pom.xml
+++ b/qa/integration-tests/pom.xml
@@ -393,7 +393,11 @@
       <artifactId>tasklist-common</artifactId>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.agrona</groupId>
+      <artifactId>agrona</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <repositories>
     <repository>
--- a/qa/integration-tests/pom.xml
+++ b/qa/integration-tests/pom.xml
@@ -492,6 +492,30 @@
         </plugins>
       </build>
     </profile>
+
+    <!--  Profile to configure tests against an external ELASTICSEARCH container -->
+    <profile>
+      <id>e2e-elasticsearch-test</id>
+      <properties>
+        <test.integration.camunda.database.type>es</test.integration.camunda.database.type>
+      </properties>
+    </profile>
+
+    <!--  Profile to configure tests against an external OPENSEARCH container -->
+    <profile>
+      <id>e2e-opensearch-test</id>
+      <properties>
+        <test.integration.camunda.database.type>es</test.integration.camunda.database.type>
+      </properties>
+    </profile>
+
+    <!--  Profile to configure tests against an external RDBMS container -->
+    <profile>
+      <id>e2e-rdbms-test</id>
+      <properties>
+        <test.integration.camunda.database.type>rdbms</test.integration.camunda.database.type>
+      </properties>
+    </profile>
   </profiles>
 
 </project>
--- a/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.it.utils;
+
+import io.camunda.client.CamundaClient;
+import io.camunda.qa.util.cluster.NewCamundaTestApplication;
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.net.http.HttpResponse.BodyHandlers;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+import org.junit.platform.commons.support.ModifierSupport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** TBD */
+public class CamundaExternalDBExtension
+    implements AfterAllCallback, BeforeAllCallback, ParameterResolver {
+
+  public static final String PROP_CAMUNDA_IT_DATABASE_TYPE = "camunda.it.database.type";
+  public static final String DEFAULT_ES_URL = "http://localhost:9200";
+  public static final String DEFAULT_OS_URL = "http://localhost:9200";
+  public static final String DEFAULT_OS_ADMIN_USER = "admin";
+  public static final String DEFAULT_OS_ADMIN_PW = "yourStrongPassword123!";
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CamundaExternalDBExtension.class);
+  private final DatabaseType databaseType;
+  private final List<AutoCloseable> closeables = new ArrayList<>();
+  private final NewCamundaTestApplication testApplication;
+  private String testPrefix;
+
+  public CamundaExternalDBExtension() {
+    this(new NewCamundaTestApplication());
+
+    closeables.add(testApplication);
+    testApplication
+        .withBrokerConfig(cfg -> cfg.getGateway().setEnable(true))
+        .withRecordingExporter(true);
+  }
+
+  public CamundaExternalDBExtension(final NewCamundaTestApplication testApplication) {
+    this.testApplication = testApplication;
+    // resolve active database and exporter type
+    final String property = System.getProperty(PROP_CAMUNDA_IT_DATABASE_TYPE);
+    databaseType =
+        property == null ? DatabaseType.ES : DatabaseType.valueOf(property.toUpperCase());
+  }
+
+  @Override
+  public void beforeAll(final ExtensionContext context) {
+    LOGGER.info("Starting up Camunda instance, with {}", databaseType);
+    final Class<?> testClass = context.getRequiredTestClass();
+    testPrefix = testClass.getSimpleName().toLowerCase();
+
+    switch (databaseType) {
+      case ES -> testApplication.withElasticsearchSupport(DEFAULT_ES_URL, testPrefix);
+      case OS ->
+          testApplication.withOpensearchSupport(
+              DEFAULT_OS_URL, testPrefix, DEFAULT_OS_ADMIN_USER, DEFAULT_OS_ADMIN_PW);
+      case RDBMS -> testApplication.withRdbmsExporter();
+      default -> throw new RuntimeException("Unknown exporter type");
+    }
+    testApplication.start();
+    testApplication.awaitCompleteTopology();
+
+    injectFields(testClass, null, ModifierSupport::isStatic);
+  }
+
+  private void injectFields(
+      final Class<?> testClass, final Object testInstance, Predicate<Field> predicate) {
+    predicate = predicate.and(field -> field.getType() == CamundaClient.class);
+    for (final Field field : testClass.getDeclaredFields()) {
+      try {
+        if (predicate.test(field)) {
+          field.setAccessible(true);
+          field.set(testInstance, createCamundaClient());
+        }
+      } catch (final Exception ex) {
+        throw new RuntimeException(ex);
+      }
+    }
+  }
+
+  @Override
+  public void afterAll(final ExtensionContext context) {
+    if (databaseType == DatabaseType.ES || databaseType == DatabaseType.OS) {
+      final URI deleteEndpoint = URI.create(String.format("%s/%s*", DEFAULT_ES_URL, testPrefix));
+      final HttpRequest httpRequest = HttpRequest.newBuilder().DELETE().uri(deleteEndpoint).build();
+      try (final HttpClient httpClient = HttpClient.newHttpClient()) {
+        final HttpResponse<String> response = httpClient.send(httpRequest, BodyHandlers.ofString());
+        final int statusCode = response.statusCode();
+        if (statusCode / 100 == 2) {
+          LOGGER.info("Test data deleted.");
+        } else {
+          LOGGER.warn("Failure on deleting test data. Status code: {}", statusCode);
+        }
+      } catch (final IOException | InterruptedException e) {
+        LOGGER.warn("Failure on deleting test data.", e);
+      }
+    }
+
+    closeables.parallelStream()
+        .forEach(
+            c -> {
+              try {
+                c.close();
+              } catch (final Exception e) {
+                throw new RuntimeException(e);
+              }
+            });
+  }
+
+  @Override
+  public boolean supportsParameter(
+      final ParameterContext parameterContext, final ExtensionContext extensionContext)
+      throws ParameterResolutionException {
+    return parameterContext.getParameter().getType() == CamundaClient.class;
+  }
+
+  @Override
+  public Object resolveParameter(
+      final ParameterContext parameterContext, final ExtensionContext extensionContext)
+      throws ParameterResolutionException {
+    return createCamundaClient();
+  }
+
+  private CamundaClient createCamundaClient() {
+    return testApplication.newClientBuilder().build();
+  }
+
+  public enum DatabaseType {
+    ES,
+    OS,
+    RDBMS
+  }
+}
--- a/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
@@ -37,7 +37,8 @@ import org.testcontainers.elasticsearch.ElasticsearchContainer;
 public class CamundaMultiDBExtension
     implements AfterAllCallback, BeforeAllCallback, ParameterResolver {
 
-  public static final String PROP_CAMUNDA_IT_DATABASE_TYPE = "camunda.it.database.type";
+  public static final String PROP_CAMUNDA_IT_DATABASE_TYPE =
+      "test.integration.camunda.database.type";
   public static final String DEFAULT_ES_URL = "http://localhost:9200";
   public static final String DEFAULT_OS_URL = "http://localhost:9200";
   public static final String DEFAULT_OS_ADMIN_USER = "admin";
--- a/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
@@ -30,7 +30,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /** TBD */
-public class CamundaExternalDBExtension
+public class CamundaMultiDBExtension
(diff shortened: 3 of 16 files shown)


qa/util/src/main/java/io/camunda/qa/util/cluster/NewCamundaTestApplication.java
```diff
@@ -0,0 +1,386 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.util.unit.DataSize;
+
+/** Represents an instance of the {@link BrokerModuleConfiguration} Spring application. */
+@SuppressWarnings("UnusedReturnValue")
+public final class NewCamundaTestApplication
+    extends TestSpringApplication<NewCamundaTestApplication>
+    implements TestGateway<NewCamundaTestApplication> {
+
```
ChrisKujawa, line 54: > [!Important]
>
> This is intermediate state, as I don't wanted to disrupt existing tests and workflows. 
> This is actually the most important part to discuss.

Do we want to have this configuration logic inside such a TestApplication class like we have [here](https://github.com/camunda/camunda/pull/27109/files#diff-7598c5ed62ccc7145a0db843a441fd5697f9cbea13144d7347e0a22ad5a3e0bcR271).

Benefit: We could use it also outside of our extension, so when using only the TestApplication.

But we could also do it in the new extension, accepting just a TestApplication and calling the withProperties and withBeans methods, such that this class becomes obsolete. 

I would like to hear your opinion here @npepinpe @tmetzke

qa/util/src/main/java/io/camunda/qa/util/cluster/NewCamundaTestApplication.java
```diff
@@ -0,0 +1,386 @@
...
+   * @param modifier a configuration function
+   * @return itself for chaining
+   */
+  public NewCamundaTestApplication withExporter(
+      final String id, final Consumer<ExporterCfg> modifier) {
+    registeredExporters.merge(id, modifier, (key, cfg) -> cfg.andThen(modifier));
+    return this;
+  }
+
+  public NewCamundaTestApplication withElasticsearchSupport(
```
ChrisKujawa, line 271: To be discussed here https://github.com/camunda/camunda/pull/27109/files#r1920413439

qa/util/src/main/java/io/camunda/qa/util/cluster/NewCamundaTestApplication.java
ChrisKujawa: Just that you understand what I mean. I created a separate POC to move everything in the extension https://github.com/camunda/camunda/commit/5a45578ed2c36521335c5ccf20af27b64afb1f97 Not perfect, but just to get the idea.

npepinpe: Some applications may not touch the DB - though to be fair, they might get removed in the future, or replaced (e.g. the current restore app which only restores Zeebe). So tentatively I would compose the behavior around the history storage in a new interface, and have the others extend or implement it, e.g. `TestDatabaseApplication`. This interface would then declare an appropriate `withDatabaseConfig(Consumer<SearchClientProperties>)` to allow configuring the DB, and then whatever other helpers (e.g. `withRdbmsExporter`) which are just defaults on that interface, but using the declared `withDatabaseConfig`, meaning implementations only really need to implement a very small set of things.

But like I said, it may be that this is a temporary thing, and I could imagine in the future _all_ applications are touching the DB, so we might as well put it on `TestApplication`.

This is where I miss the structural typing of Go, where you could just "give me anything which has a method `withDbConfig`" or something :smile: 

:wrench: Considering you only added 3 methods right now, which by the way are mutually exclusive, I would instead collapse them into a single `withDatabaseConfig(Consumer<SearchClientProperties>)`, and have the extension just configure it. Unless we expect people to also be doing this manually often, in which case the helpers make sense - but we can just start small. Then we can add this to `TestApplication`, and just have `TestRestoreApp` throw `UnsupportOperationException` for now.

npepinpe: :question: Or are there any other differences between this and `TestApplication`?

npepinpe (approved):
:rocket: Please address all comments (except the thoughts, you can choose to ignore if you want :smile:)

:thought_balloon: Looking at it, it makes me wonder if I didn't make a mistake allowing programmatic access of the configuration classes via injected beans. This breaks the configuration resolution of certain properties, which means certain classes of tests (where you want to ensure a property is used) are not possible. It also breaks some user (as in, the devs) expectation, since setting a property sometimes has no effect, which is a bit of a gotcha. But at the same time, it's more convenient... :disappointed: 

:question: I don't really get why we need a proxy schema manager, but you mention this is already a thing for ES, so...I guess? :shrug: I am curious to understand it though :smile:

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
```diff
@@ -0,0 +1,152 @@
...
+    switch (databaseType) {
+      case ES -> testApplication.withElasticsearchSupport(DEFAULT_ES_URL, testPrefix);
+      case OS ->
+          testApplication.withOpensearchSupport(
+              DEFAULT_OS_URL, testPrefix, DEFAULT_OS_ADMIN_USER, DEFAULT_OS_ADMIN_PW);
+      case RDBMS -> testApplication.withRdbmsExporter();
+      default -> throw new RuntimeException("Unknown exporter type");
+    }
+    testApplication.start();
+    testApplication.awaitCompleteTopology();
```
npepinpe, line 80: :wrench: :wrench: :hammer_and_wrench:  I'm not a fan of this managing the state of the application to be honest. We're just re-doing the same code. Also, there may be tests where I don't want a completely shared C8 instance, but want to reconfigure it between tests, or shut it down, etc. I know all these can be waved as "we'll implement it later", but I think it can be done better by just using the existing test extension to manage application lifecycle, and focusing this one on just configuring the application based on the DB and nothing else.

I know that right now we have a specific problem, and we want to focus on solving it, but I'm afraid we might miss the forest for the tree and optimize for the short term. I'm not putting this as a blocker, because I think it can be changed later, but I would strongly urge us to improve it ASAP.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
```diff
@@ -0,0 +1,152 @@
...
+          LOGGER.info("Test data deleted.");
+        } else {
+          LOGGER.warn("Failure on deleting test data. Status code: {}", statusCode);
+        }
+      } catch (final IOException | InterruptedException e) {
+        LOGGER.warn("Failure on deleting test data.", e);
+      }
+    }
+
+    closeables.parallelStream()
```
npepinpe, line 118: :wrench: `CloseHelper.quietCloseAll(closeables)`

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
```diff
@@ -0,0 +1,152 @@
...
+        throw new RuntimeException(ex);
+      }
+    }
+  }
+
+  @Override
+  public void afterAll(final ExtensionContext context) {
+    if (databaseType == DatabaseType.ES || databaseType == DatabaseType.OS) {
+      final URI deleteEndpoint = URI.create(String.format("%s/%s*", DEFAULT_ES_URL, testPrefix));
+      final HttpRequest httpRequest = HttpRequest.newBuilder().DELETE().uri(deleteEndpoint).build();
```
npepinpe, line 104: :question: Is this _really_ deleting everything? Indices, templates, component templates, aliases, etc.? All things listed here? https://github.com/camunda/camunda/issues/26043

Not a blocker though - in fact, why not make use of the purge feature once it's done? Right now it's only implemented for RDBMS, but we will have it for all DBs soon, and it will purge your C8 runtime (so log stream, snapshots, etc.) and the history storage (e.g. ES, RDBMS). We can make it as a TODO.

npepinpe: See: https://github.com/camunda/camunda/issues/25861

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
```diff
@@ -0,0 +1,152 @@
...
+      final ParameterContext parameterContext, final ExtensionContext extensionContext)
+      throws ParameterResolutionException {
+    return parameterContext.getParameter().getType() == CamundaClient.class;
+  }
+
+  @Override
+  public Object resolveParameter(
+      final ParameterContext parameterContext, final ExtensionContext extensionContext)
+      throws ParameterResolutionException {
+    return createCamundaClient();
```
npepinpe, line 140: :question: If the client is injected, who is closing it? The extension here doesn't keep track of it, so the test method has to close it? Might be safer to keep track of it and just close it ourselves on an `afterEach` :shrug:

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
ChrisKujawa: > Not a blocker though - in fact, why not make use of the purge feature once it's done?

Because its not done

qa/util/src/main/java/io/camunda/qa/util/cluster/NewCamundaTestApplication.java
ChrisKujawa: Can you check the other branch?

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
npepinpe: It was a rhetorical question :smile: I hope I would know if it was done or not :smile:

ChrisKujawa: Yeah i guess i would create an issue for it. I feel todos are more an antipattern and not really useful.

But tbh not sure who would take care of it.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
ChrisKujawa: I think you have a specific thing in mind (some expectation) which is not satisfiied. Maybe you can elaborate more on what it would look like. 

Be aware that the set up code shouldn't be too complex and too much boiler plate also doesnt make sense to me. 

I can write down what i understood and please confirm or decline whether this matches:

1. Test application created outside
2. An extension for the db detections and configuration
3. An extension to manage test application
4. Both configured with RegisterExtension

The issue is that the test application cant come from outside when using extendWith. This means we would always need to define extra register extension.

Complex scenarios are supported, with the trade of that simple set ups always have this boilerplate.

I would like to have something in between. That i can simple use the one extension and it sets everything up and if i need more i can use maybe a additional one or simple the extension exposes the some utilities to start and stop.

Maybe we want too many things at once. Maybe we need two different extensions depending on the use case. To separate the/your concerns. 🤔


Additional idea we have three extensions: the third one ia combining the both above to support the simple scenario. If more complex scenario we use the two above directly.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
ChrisKujawa: That sounds so evil 💀

npepinpe: dark side

ChrisKujawa: Thanks. I created a follow up to use this feature in integration tests https://github.com/camunda/camunda/issues/27264

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
ChrisKujawa: I will take it as a follow up.

qa/util/src/main/java/io/camunda/qa/util/cluster/NewCamundaTestApplication.java
ChrisKujawa: I reiterated about it

ChrisKujawa:
Hey @npepinpe I reiterated over the solution, and I think followed most of your suggestions and hints. 


Furthermore, I synced with @p-wunderlich and @stoerti to get some input, feedback, and share what will be are working on.


What I did:

 * We now make use of junit tagging, multi-db-test. The tagged tests are not executed on normal IT execution. The separate job takes care of it (similar what we already have with RDBMS)
 * We have a new annotation, that combines the tagging, the extending with the extension, etc.
 * We have introduced a new interface that is for the standalone application specifically, that allows to configure, gateway, exporter and broker properties. Discussable, but it helped to generalize the support for the Extension. Now we can use the extension also with TestStandloneBroker.
 * There is a new helper class that takes care of configuring the TestStandaloneApplication. This is tested as well.
 *  We now have better support for local execution: When no property is set we are spinning up an ES test container, allowing the same workflow as before. There is also no longer an ignoring of tests, when the property is not set.
 * I changed as requested the used test property, to not clash with the common camunda namespace.
 * There have been maybe some smaller other changes.


I would like to go ahead and merge as several people waiting for it. I think we can iterate over it, we see some flaws and issues with it. Please stop me if something is totally wrong.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaExternalDBExtension.java
npepinpe: :+1: I don't want to block you, I think we sort of understand each other :smile: I don't have a super clear idea, but my goal is to be able to compose extensions, instead of having one uber extension, or having to replicate the same behavior everywhere. I think I can look more into it post Kick Off.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -0,0 +1,238 @@
...
+        multiDbConfigurator.configureElasticsearchSupport(DEFAULT_ES_URL, testPrefix);
+      }
+      case OS ->
+          multiDbConfigurator.configureOpenSearchSupport(
+              DEFAULT_OS_URL, testPrefix, DEFAULT_OS_ADMIN_USER, DEFAULT_OS_ADMIN_PW);
+      case RDBMS -> multiDbConfigurator.configureRDBMSSupport();
+      default -> throw new RuntimeException("Unknown exporter type");
+    }
+    testApplication.start();
+    testApplication.awaitCompleteTopology();
```
npepinpe, line 144: :upside_down_face: You can chain the calls

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -0,0 +1,238 @@
...
+    final String property = System.getProperty(PROP_CAMUNDA_IT_DATABASE_TYPE);
+    databaseType =
+        property == null ? DatabaseType.LOCAL : DatabaseType.valueOf(property.toUpperCase());
+  }
+
+  @Override
+  public void beforeAll(final ExtensionContext context) {
+    LOGGER.info("Starting up Camunda instance, with {}", databaseType);
+    final Class<?> testClass = context.getRequiredTestClass();
+    testPrefix = testClass.getSimpleName().toLowerCase();
```
npepinpe, line 124: :question: Is simple name like, given a class `io.camunda.it.qa.MyTest`, just `MyTest`? Because we could have collisions without package names. :wrench: So using the fully qualified name would be a better option here, just in case.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -0,0 +1,238 @@
...
+    testApplication.awaitCompleteTopology();
+
+    injectFields(testClass, null, ModifierSupport::isStatic);
+  }
+
+  private ElasticsearchContainer setupElasticsearch() {
+    final ElasticsearchContainer elasticsearchContainer =
+        TestSearchContainers.createDefeaultElasticsearchContainer();
+    elasticsearchContainer.start();
+    closeables.add(elasticsearchContainer);
```
npepinpe, line 153: :upside_down_face: We can add it before starting the container for safety.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -0,0 +1,238 @@
...
+    return elasticsearchContainer;
+  }
+
+  private static void validateESConnection(final String url) {
+    final HttpRequest httpRequest =
+        HttpRequest.newBuilder().GET().uri(URI.create(String.format("%s/", url))).build();
+    try (final HttpClient httpClient = HttpClient.newHttpClient()) {
+      final HttpResponse<String> response = httpClient.send(httpRequest, BodyHandlers.ofString());
+      final int statusCode = response.statusCode();
+      assert statusCode / 100 == 2
```
npepinpe, line 163: :question: Why / 100? I don't get it :smile:

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -0,0 +1,238 @@
...
+          : "Expected to have a running ES service available under: " + url;
+    } catch (final IOException | InterruptedException e) {
+      assert false
+          : "Expected no exception on validating connection under: "
+              + url
+              + ", failed with: "
+              + e
+              + ": "
+              + e.getMessage();
```
npepinpe, line 172: :wrench: You can also `throw new AssertionError`, which is what this does ultimately, instead of a weird `assert false`.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -0,0 +1,238 @@
...
+      final Class<?> testClass, final Object testInstance, Predicate<Field> predicate) {
+    predicate = predicate.and(field -> field.getType() == CamundaClient.class);
+    for (final Field field : testClass.getDeclaredFields()) {
+      try {
+        if (predicate.test(field)) {
+          field.setAccessible(true);
+          field.set(testInstance, createCamundaClient());
+        }
+      } catch (final Exception ex) {
+        throw new RuntimeException(ex);
```
npepinpe, line 186: :wrench: You can use `LangUtil.rethrowUnchecked` to avoid the unnecessary wrapping.

qa/integration-tests/src/test/java/io/camunda/it/utils/MultiDbConfigurator.java
```diff
@@ -0,0 +1,142 @@
...
+                  "index",
+                  Map.of("prefix", indexPrefix),
+                  "bulk",
+                  Map.of("size", 1)));
+        });
+    testApplication.withProperty(
+        "camunda.database.type",
+        io.camunda.search.connect.configuration.DatabaseType.ELASTICSEARCH);
+    testApplication.withProperty("camunda.database.indexPrefix", indexPrefix);
+    testApplication.withProperty("camunda.database.url", elasticsearchUrl);
```
npepinpe, line 72: :clown_face: I know it's not your fault, but beautiful :smile:

npepinpe (commented):
:rocket: Reviewing after merge, no worries, don't want to block you, but still want to point out some things or just comment, ask questions, etc.

- :question: For local support, do we plan on supporting more DBs? I would have expected instead that I can still configure the DB type (e.g. `RDMBS`, `OS`), but if `CI=false` in the environment, the extension would spin up the right thing. Is that the future plan?

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
entangled90: So that 201 is ok :-D

qa/integration-tests/pom.xml
```diff
@@ -477,6 +496,30 @@
...
+      <properties>
+        <test.integration.camunda.database.type>es</test.integration.camunda.database.type>
+      </properties>
+    </profile>
+
+    <!--  Profile to configure tests against an external OPENSEARCH container -->
+    <profile>
+      <id>e2e-opensearch-test</id>
+      <properties>
+        <test.integration.camunda.database.type>es</test.integration.camunda.database.type>
```
mustafadagher, line 512: @ChrisKujawa was this intentionally set to `es`? or is it a typo?

ChrisKujawa: This was 1 year ago 🤣 i miss a bit context now. I guess probably a typo idk - i guess this was adjusted on further PRs

mustafadagher: it's still the same on main haha 😄 I got here from git blame

PR #27113, opened 2025-01-17
Migrate the first integration test to external db foundation

ChrisKujawa:
## Description

 Took ProcessInstanceAndFlowNodeInstanceQueryTest as example test to migrate to new external database approach and new extension

Note:

 * Locally I was able to run it against ES AND OS container 🚀 

## See in action

## Related issues

### DIFF
--- a/qa/integration-tests/pom.xml
+++ b/qa/integration-tests/pom.xml
@@ -433,7 +433,7 @@
               <value>${version.identity}</value>
             </property>
           </systemProperties>
-          <groups>!rdbms</groups>
+          <groups>!rdbms,!multi-db-test</groups>
         </configuration>
         <dependencies>
           <dependency>
@@ -490,7 +490,7 @@
             <artifactId>maven-failsafe-plugin</artifactId>
             <configuration>
               <failIfNoTests>false</failIfNoTests>
-              <groups>multiDbTest</groups>
+              <groups>multi-db-test</groups>
             </configuration>
           </plugin>
         </plugins>
--- a/qa/integration-tests/pom.xml
+++ b/qa/integration-tests/pom.xml
@@ -433,7 +433,7 @@
               <value>${version.identity}</value>
             </property>
           </systemProperties>
-          <groups>!rdbms & !multi-db-test</groups>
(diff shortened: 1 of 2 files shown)


ChrisKujawa:
@npepinpe @tmetzke has someone time to review this?

qa/integration-tests/pom.xml
```diff
@@ -433,7 +433,7 @@
               <value>${version.identity}</value>
             </property>
           </systemProperties>
-          <groups>!rdbms</groups>
+          <groups>!rdbms,!multiDbTest</groups>
```
tmetzke, line 436: ⚠️ I am missing too much context from the POC to judge if this is necessary/correct (I didn't find enough capacity yet to look into that thoroughly, sorry).

tmetzke (approved):
👍 The changes to the test itself look good to me 🍪 
⚠️ I am lacking context from the POC I couldn't look into enough yet to judge if the change is now consistent and all required CI is successful. I can see the Root QA is failing but I cannot judge why 🙈

ChrisKujawa:
>  I can see the Root QA is failing but I cannot judge why 🙈

Thanks, yeah the problem was that the RDBMS test was executed during the normal integration tests. This was a bit unexpected. 

It seems that negations, with multiple groups, didn't work well like:

```
<groups>!rdbms,!multi-db-test</groups>
```

I tried multiple things, for example ampersand [as documented in JUnit 5 docs ](https://junit.org/junit5/docs/snapshot/user-guide/#running-tests-build-maven-filter-tags), but this also didn't work.

I tried with excluded groups because that is essentially what we wanted to have. This caused that in the separate rdbms and multi db tests no tests are executed anymore 😆 

This was because of our separate profiles, where we want to run the rdbsm and multi db tests, as they didn't overwrite the excluded groups. 

In the end, I had to:

1. Add rdbms and multi-db-tests as excluded in normal failsafe execution, for our normal IT
2. Add rdbms as excluded in multidb profile
3. Add multidb as excluded in rdbms profile

This works fine, and all tests are running correctly.

Issue #22347, opened 2024-09-16
Add base path in Optimize app URLs


buccarel:
### Context

Once Optimize is part of the single application, such application should serve requests to Optimize together with the requests of the rest of the stack.

Also, as API calls are currently done with a relative URL, they should be changed to be done against the servers's hostname and not against the new `/optimize` context path, as they will be part of the unified API (e.g., a call to `/api/something` should be done against `schema://hostname:port/api/something` and not against `schema://hostname:port/optimize/api/something`).

### Acceptance Criteria

[ ] To dispatch between Optimize's and the rest of the stack's UI, we need to add the `/optimize` base path, and this shall be used by Optimize's UI.
[ ] The same base paths will be used when Optimize is run in standalone mode. This achieves consistency:
* within Optimize itself, meaning that Optimize would work the same way whether it's launched standalone or together with the rest of the stack
* with respect to other applications of the stack, as they're following the same strategy
[ ] in standalone mode, if a request is made against the root, it shall be redirected to `/optimize`
[x] The API requests made by the frontend should be relative to the webserver's host and not relative to optimize's context path

### Links

* Part of https://github.com/camunda/product-hub/issues/2289

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/27138
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [ ] The change is implemented as described on all target environments/versions/modes
- [ ] The documentation changes are as expected

### Completion

- [ ] All Review stages are successfully completed
- [ ] All associated PRs are merged to the main branch(es) and stable branches
- [ ] The correct version labels are applied to the issue

PR #27138, opened 2025-01-20
build: optimize calls the API with the absolute URL

buccarel:
## Description

As the single application will introduce the `/optimize` context path, we need to make sure that the calls to the API are done against the server hostname without such context path. This is doable by calling the API URIs with an absolute URL, instead of appending a relative URL to the current path.

For more context:
https://github.com/camunda/camunda/issues/22347

## Checklist

- [ ] for CI changes:

## Related issues

Part of:
* https://github.com/camunda/camunda/issues/22347
* https://github.com/camunda/product-hub/issues/2289

### DIFF
--- a/optimize/client/src/modules/api.ts
+++ b/optimize/client/src/modules/api.ts
@@ -6,12 +6,10 @@
  * except in compliance with the Camunda License 1.0.
  */
 
-export function getAbsoluteURL(url: string) {
-  try {
-    const absoluteURL = new URL(url);
-    return absoluteURL.toString();
-  } catch {
-    const baseURL = `${window.location.protocol}//${window.location.host}`;
-    return `${baseURL}/${url}`;
+export function getFullURL(url: string) {
+  if (typeof window.location.origin !== 'string') {
+    throw new Error('window.location.origin is not a set');
   }
+
+  return new URL(url, window.location.origin).toString();
 }
--- a/optimize/client/src/modules/api.ts
+++ b/optimize/client/src/modules/api.ts
@@ -1,0 +1,17 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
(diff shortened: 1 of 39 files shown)


optimize/client/src/modules/api.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function getAbsoluteURL(url: string) {
+  try {
+    const absoluteURL = new URL(url);
+    return absoluteURL.toString();
+  } catch {
+    const baseURL = `${window.location.protocol}//${window.location.host}`;
+    return `${baseURL}/${url}`;
+  }
+}
```
vsgoulart, line 17: @buccarel I think it's better to copy the implementation from Tasklist and Operate since it's already working with no problems and when we kick-off the frontend streamlining of the component apps we will be able to extract into a common util across apps.

Your implementation also has some problems. Because you're always trying to create a URL without a domain the try will always throw an error and only the catch will ever execute.
Another problem is that you don't take into account that a user might call the function with an extra slash when adding a new endpoint which might break things.

You can check it how we do on the apps [here (main function to get full URL)](https://github.com/camunda/camunda/blob/main/tasklist/client/src/modules/api.ts#L18-L24) and [here (function to handle merging path)](https://github.com/camunda/camunda/blob/main/tasklist/client/src/modules/utils/mergePathname.tsx) (you can remove the `BASENAME` part on your implementation since Optimize still uses a hash router)

vsgoulart (commented):
Besides the comment I mentioned I think we should write 1 or 2 tests for this function

vsgoulart (approved):
LGTM

@buccarel I believe the E2E tests are just flaky, correct? If yes, then feel free to merge whenever

buccarel:
> LGTM
> 
> @buccarel I believe the E2E tests are just flaky, correct? If yes, then feel free to merge whenever

@vsgoulart yes, that pipeline has 1 unstable test. I am about to deactivate it, indeed ^_^ Thanks

PR #27161, opened 2025-01-20
fix: Store duration values in a normalized format

### LINKED ISSUE #27098: [FEEL] Store duration values in a normalized format
**Description**

In C8, the engine stores a duration value from a FEEL expression in a string format. In some cases, the string format is not normalized, for example, the value `@"P5D"` is stored as `PT120H`. Users expect that the string format is formalized. 

Internally, the duration value is transformed in
[FeelToMessagePackTransformer](https://github.com/camunda/camunda/blob/main/zeebe/feel/src/main/java/io/camunda/zeebe/feel/impl/FeelToMessagePackTransformer.java#L100). Here, we use the `toString` function of the [Duration](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/time/Duration.html#toString()) object that returns a normalized format but only in hours.

Instead, we should use the `toString()` function of the [ValDayTimeDuration](https://github.com/camunda/feel-scala/blob/main/src/main/scala/org/camunda/feel/syntaxtree/Val.scala#L245) object directly that returns the expected normalized format.

saig0:
## Description

Use `toString()` of FEEL duration to serialize the value into a JSON variable. Compared to the Java's duration value, the
FEEL value returns the duration in a normalized format (i.e. in the biggest available unit).

This fix aligns the engine with the [docs](https://docs.camunda.io/docs/components/modeler/feel/language-guide/feel-temporal-expressions/#subtraction) and avoids the workaround using the [string()](https://docs.camunda.io/docs/components/modeler/feel/builtin-functions/feel-built-in-functions-conversion/#stringfrom) function.

```
// before
"PT120H"

// after
"P5D"
```

## Related issues

closes #27098

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/variable/mapping/ActivityInputMappingTest.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/variable/mapping/ActivityInputMappingTest.java
@@ -40,7 +40,7 @@ public final class ActivityInputMappingTest {
   private static final String A_DATE_VALUE = "\"2021-02-24\"";
   private static final String A_LOCAL_DATE_AND_TIME_VALUE = "\"2021-02-24T04:20\"";
   private static final String A_DATE_AND_TIME_VALUE = "\"2021-02-24T04:20+01:00\"";
-  private static final String A_DAY_TIME_DURATION_VALUE = "\"PT42H56M33S\"";
+  private static final String A_DAY_TIME_DURATION_VALUE = "\"P1DT18H56M33S\"";
   private static final String A_YEAR_MONTH_DURATION_VALUE = "\"P2Y3M\"";
 
   @Rule
--- a/zeebe/expression-language/src/test/java/io/camunda/zeebe/el/EvaluationResultTest.java
+++ b/zeebe/expression-language/src/test/java/io/camunda/zeebe/el/EvaluationResultTest.java
@@ -11,6 +11,7 @@ import static io.camunda.zeebe.test.util.MsgPackUtil.asMsgPack;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import io.camunda.zeebe.el.util.TestFeelEngineClock;
+import io.camunda.zeebe.util.buffer.BufferUtil;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.Period;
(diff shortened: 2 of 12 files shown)


zeebe/expression-language/src/test/java/io/camunda/zeebe/el/EvaluationResultTest.java
```diff
@@ -227,6 +232,34 @@ public void contextWithDateExpression() {
...
+    assertThat(evaluationResult.getBoolean()).isNull();
+    assertThat(evaluationResult.getString()).isNull();
+    assertThat(evaluationResult.getNumber()).isNull();
+    assertThat(evaluationResult.getList()).isNull();
+    assertThat(evaluationResult.toBuffer()).isEqualTo(asMsgPack("\"P5D\""));
+  }
+
+  @Test
+  public void shouldReturnNormalizedPeriod() {
+    final var evaluationResult = evaluateExpression("=duration(\"P2Y\")");
```
korthout, line 251: ❓ Should we also add a test case for period and duration with two separate units?

e.g.
- `P2Y3M`
- `P5DT3H2M`

zeebe/expression-language/src/test/java/io/camunda/zeebe/el/EvaluationResultTest.java
```diff
@@ -227,6 +232,34 @@ public void contextWithDateExpression() {
     assertThat(evaluationResult.toBuffer()).isEqualTo(asMsgPack("{'x':\"2020-04-02\"}"));
   }
 
+  @Test
+  public void shouldReturnNormalizedDuration() {
+    final var evaluationResult = evaluateExpression("=duration(\"P5D\")");
```
korthout, line 237: 🔧 We could show the normalization by starting with a non-normalized format:

- `PT120H` -> `P5D`

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/variable/mapping/ActivityInputMappingTest.java
```diff
@@ -40,7 +40,7 @@ public final class ActivityInputMappingTest {
   private static final String A_DATE_VALUE = "\"2021-02-24\"";
   private static final String A_LOCAL_DATE_AND_TIME_VALUE = "\"2021-02-24T04:20\"";
   private static final String A_DATE_AND_TIME_VALUE = "\"2021-02-24T04:20+01:00\"";
-  private static final String A_DAY_TIME_DURATION_VALUE = "\"PT42H56M33S\"";
+  private static final String A_DAY_TIME_DURATION_VALUE = "\"P1DT18H56M33S\"";
```
korthout, line 43: 😍

korthout (approved):
Nice work @saig0 🚀 

LGTM 👍 

🔧 I have some suggestions to extend test coverage

PR #27195, opened 2025-01-21
Add `PERMISSIONS_BY_AUTHORIZATION_KEY ` cf

### LINKED ISSUE #27192: Add a new column family to track the permissions given an authorization key
- Create PERMISSIONS_BY_AUTHORIZATION_KEY column family
- The key should be a composite key like `authorizataionKey + owner id + ownerType + resource type -> nil`
- On authorization creation a record should be added into this cf
- On authorization deletion, we need to lookup into this cf to get the permission key to remove

nicpuppa:
## Description

This PR adds a new column family to keep track of the authorization records and the related permissions

## Related issues

closes #27192


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -47,6 +48,13 @@ public class DbAuthorizationState implements MutableAuthorizationState {
   private final ColumnFamily<DbLong, PersistedAuthorization>
       authorizationByAuthorizationKeyColumnFamily;
 
+  // authorization key -> owner type + owner id + resource type
+  private final DbForeignKey<DbCompositeKey<DbCompositeKey<DbString, DbString>, DbString>>
```
koevskinikola, line 52: 👍 The inception here is amazing! 😄

koevskinikola (approved):
LGTM @nicpuppa 🚀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
nicpuppa: Yeah, that's why I had the idea of build the key merging the ownerType and ownerId. 😄

PR #27196, opened 2025-01-21
pg-26966-unbind-process-migration-from-tasklist

### LINKED ISSUE #26966: Figure out where should `process-migration` be run
The `process-migration` profile is being triggered on the `tasklist-importer-archiver` deployment as well. 

```
"ApplicationErrorListener.java","line":24},"message":"Failed to start application with message: Error creating bean with name 'migrationsRunner' defined in URL [jar:file:/usr/local/tasklist/lib/camunda-zeebe-8.7.0-alpha3.jar!/io/camunda/application/commons/migration/MigrationsRunner.class]: 
Unsatisfied dependency expressed through constructor parameter 0: Error creating bean with name 'process-migrator'
```

Also, another consideration is that if a SaaS cluster/SM installation does not make use of Tasklist, the migration of existing `operate-process` records might be mandatory as well for the V2 APIs. This needs some further investigation.

panagiotisgts:
## Description

- Tasklist profile no longer runs process-migration
- Include migration modules in StandaloneCamunda
## Checklist

- [ ] for CI changes:

## Related issues

closes #26966

### DIFF
--- a/dist/src/main/java/io/camunda/application/StandaloneCamunda.java
+++ b/dist/src/main/java/io/camunda/application/StandaloneCamunda.java
@@ -10,7 +10,6 @@ package io.camunda.application;
 import static org.springframework.core.env.AbstractEnvironment.ACTIVE_PROFILES_PROPERTY_NAME;
 
 import io.camunda.application.commons.CommonsModuleConfiguration;
-import io.camunda.application.commons.migration.MigrationsRunner;
 import io.camunda.application.initializers.DefaultAuthenticationInitializer;
 import io.camunda.application.initializers.HealthConfigurationInitializer;
 import io.camunda.application.initializers.WebappsConfigurationInitializer;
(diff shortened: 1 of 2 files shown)


dist/src/main/java/io/camunda/application/StandaloneCamunda.java
```diff
@@ -46,7 +47,8 @@ public static void main(final String[] args) {
                 IdentityModuleConfiguration.class,
                 WebappsModuleConfiguration.class,
                 BrokerModuleConfiguration.class,
-                GatewayModuleConfiguration.class)
```
ChrisKujawa, line 49: 💭 ❓  I think we also need to enable the profile in Line 32 right?

panagiotisgts: This will cause the profile to run as default. Adding just it's configuration allows it to be enabled when the profile is actually specified when running the StandaloneCamunda application.

dist/src/main/java/io/camunda/application/StandaloneCamunda.java
```diff
@@ -46,7 +47,8 @@ public static void main(final String[] args) {
...
+                GatewayModuleConfiguration.class,
+                MigrationsRunner.class)
```
houssain-barouni, line 51: 🔧 I think this is not needed to add this . `MigrationsRunner` is part of `io.camunda.application.commons` package so it should be scanned by `CommonsModuleConfiguration` and enabled when the right profiles are set

houssain-barouni (changes requested):
Remove unnecessary MigrationsRunner in application sources

houssain-barouni (approved):

houssain-barouni (approved):

PR #26704, opened 2025-01-09
[CPT] feat: Assert the BPMN elements by ID

saig0:
## Description

* Change the API of BPMN element assertions from element name to element ID
  * `hasActiveElements()`
  * `hasCompletedElements()`
  * `hasTerminatedElements()` 
* Extend the API by element selector to identify BPMN elements by ID or name or custom logic 

**Migration**

:fire: This is a **breaking change**. User tests written in version 8.6 use the BPMN element name for the assertions. With 8.7 the assertions expect the BPMN element ID instead. As a result, the tests compile but may fail.

:warning: Users must **migrate their tests** using the new element selector `ElementSelectors.byName()` for the assertions:
  * `hasActiveElements()`
  * `hasCompletedElements()`
  * `hasTerminatedElements()` 

```
// (before) with 8.6:
assertThat(processInstance).hasCompletedElements("A", "B", "C");

// (after) with 8.7:
import static io.camunda.process.test.api.assertions.ElementSelectors.byName;

assertThat(processInstance).hasCompletedElements(byName("A"), byName("B"), byName("C"));
``` 

## Related issues

closes #23294

PR #27200, opened 2025-01-21
[CPT] feat: Configure the BPMN element selector

saig0:
## Description

Add an option to configure the element selector used for the BPMN element assertions. The default selector is the
by-element-id selector.

This is a follow-up task of #26704 to ease the migration from 8.6. Instead of changing all assertions from BPMN element name to element ID, you need to set only the default selector and keep the test cases as before. 

```
CamundaAssert.setElementSelector(ElementSelectors::byName);
```

## Related issues

related to #26704


berkaycanbc (approved):
Nice improvement for users, LGTM 🚀

PR #27209, opened 2025-01-21
feat: Document the default page size applied in the search

### LINKED ISSUE #26587: Document the default page size applied in the search
## Description

When the API call does not specify a page size in the search query, a [default page size of 100](https://github.com/camunda/camunda/blob/b6a26fb9e20a68df32f28b2c4d9832f4ba982536/search/search-domain/src/main/java/io/camunda/search/page/SearchQueryPage.java#L17) is automatically applied.

This behavior needs to be documented in the API documentation (rest-api spec).

ghost:
feat: Document the default page size applied in the search

-

Closes https://github.com/camunda/camunda/issues/26587


tmetzke (approved):
👍 Simple change, making the default explicit everywhere, I like that 🍪

PR #27213, opened 2025-01-21
Refactor `DbAuthorizationState` PERMISSIONS column

### LINKED ISSUE #27036: AuthorizationState PERMISSIONS column family uses `owner type + owner id + resource type`
- Refactor the PERMISSIONS column family, the new key should be `owner type + owner id + resource type`. 
- On Authorization creation we need to add a record here as well

koevskinikola:
## Description

Refactors the PERMISSIONS column family, so a new key is used with `owner type + owner id + resource type` as composite key.

More details can be found in [this comment](https://github.com/camunda/camunda/pull/27213#issuecomment-2607038620).

## Checklist

- [ ] for CI changes:

## Related issues

closes #27036

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
@@ -27,14 +27,19 @@ public class DbAuthorizationState implements MutableAuthorizationState {
 
   private final PersistedAuthorization persistedAuthorization = new PersistedAuthorization();
 
-  private final DbLong ownerKey;
-  private final DbString resourceType;
-  private final DbCompositeKey<DbLong, DbString> ownerKeyAndResourceType;
-  // owner key + resource type -> permissions
-  private final ColumnFamily<DbCompositeKey<DbLong, DbString>, Permissions> permissionsColumnFamily;
-
   private final DbString ownerType;
-  // owner key -> owner type
+  private final DbString ownerId;
+  private final DbString resourceType;
+  private final DbCompositeKey<DbString, DbString> ownerTypeAndOwnerId;
+  private final DbCompositeKey<DbCompositeKey<DbString, DbString>, DbString>
+      ownerTypeOwnerIdAndResourceType;
+  // owner type + owner id + resource type -> permissions
+  private final ColumnFamily<
+          DbCompositeKey<DbCompositeKey<DbString, DbString>, DbString>, Permissions>
+      permissionsColumnFamily;
+
+  // owner key + owner id -> owner type
+  private final DbLong ownerKey;
   private final ColumnFamily<DbLong, DbString> ownerTypeByOwnerKeyColumnFamily;
 
   // authorization key -> authorization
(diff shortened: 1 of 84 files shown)


koevskinikola:
@nicpuppa the PR is done. Sorry for the large size, but most of the changes are refactorings in the tests. I needed to make sure that an `ownerId` is always passed. 

At the moment, we're not consistent with all entities, i.e. Users and Tenants already provide a String-based `username` and `tenantId` as `ownerId`s, but Groups, Roles, and Mappings don't have such properties yet. So I applied a temporary workaround where `long` keys are converted to `String.valueOf(key)`. We can refactor them when the appropriate entities provide String-based identifiers.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AddPermissionAuthorizationMultiPartitionTest.java
```diff
@@ -40,20 +40,22 @@ public class AddPermissionAuthorizationMultiPartitionTest {
...
-            .getKey();
+            .create();
+    final var ownerKey = userRecord.getValue().getUserKey();
+    final var ownerId = userRecord.getValue().getUsername();
 
     engine
         .authorization()
         .permission()
         .withOwnerKey(ownerKey)
+        .withOwnerId(ownerId)
```
nicpuppa, line 58: ❓I think you missed to set the `ownerType` here, right ? I think in all test classes of this [commit](https://github.com/camunda/camunda/pull/27213/commits/a964fffdb3521d560f2b006248a6b1b636062669) you missed to set the `ownerType`, can you double check ?

nicpuppa: Not in all the classes, but I think where the authorization is enabled

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -27,14 +27,19 @@ public class DbAuthorizationState implements MutableAuthorizationState {
...
-
   private final DbString ownerType;
-  // owner key -> owner type
+  private final DbString ownerId;
+  private final DbString resourceType;
+  private final DbCompositeKey<DbString, DbString> ownerTypeAndOwnerId;
+  private final DbCompositeKey<DbCompositeKey<DbString, DbString>, DbString>
+      ownerTypeOwnerIdAndResourceType;
+  // owner type + owner id + resource type -> permissions
+  private final ColumnFamily<
```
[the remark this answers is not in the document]
koevskinikola: @nicpuppa I think merging the `ownerType` and `ownerId` would give us less flexibility in the future. We wouldn't be able to search by `ownerType` for example, and we would need to split the composite String if we ever need to extract it.

What would be the benefit of merging it vs how it is now? Do you think the current solution is less performant or too complex?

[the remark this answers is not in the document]
nicpuppa: Not really, it was just an idea. Your approach is more future-proof. So go for it 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AddPermissionAuthorizationMultiPartitionTest.java
koevskinikola: Agh! Sorry for that, too many changes. It's fixed now.

koevskinikola:
@nicpuppa I applied your feedback. I squashed the changes in the tests in the same commit so we can easily find them when we refactor again in the future.

The failing IT tests are disabled until we untangle the code. An issue to track the re-enablement is here https://github.com/camunda/camunda/issues/27289

nicpuppa (approved):

PR #27215, opened 2025-01-21
docs: create an entity relationship diagram for identity

lenaschoenburg:
This adds an initial diagram describing the _desired_ data model of the Identity entities.

Rendered version: https://github.com/camunda/camunda/blob/ls/identity-er-diagram/docs/identity.md

### DIFF
--- a/docs/identity.md
+++ b/docs/identity.md
@@ -1,0 +1,44 @@
+# Identity
+
+## Data model
+
+```mermaid
+erDiagram
+  User {
+    string username PK
+    string name
+    string email
+    string password
+  }
+  Role {
+    string id PK
+    string name
+  }
+  Group {
+    string id PK
+    string name
+  }
+  Tenant {
+    string id PK
+    string name
+  }
+  Authorization {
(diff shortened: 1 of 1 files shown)


lenaschoenburg:
Note that we have very little control over the exact layout, I had to try out a couple of different declaration orderings to find a layout that doesn't look super confusing.

docs/identity.md
```diff
@@ -0,0 +1,43 @@
...
+  Group {
+    string id PK
+    string name
+  }
+  Tenant {
+    string id PK
+    string name
+  }
+  Authorization {
+    enum ownerType
```
megglos, line 26: 🔧 We have the ownerId here as well, right?
```suggestion
    enum ownerType
    string ownerId
```

megglos (approved):
LGTM!

Shall we add Mappings right away too? I can also follow-up on that.

lenaschoenburg:
I'd like to leave out mappings for now, I spent enough time to get the layout nice and tidy already :sweat_smile:

Issue #27217, opened 2025-01-21
Get all users assigned to a tenant (as a search endpoint to be able to use pagination)


Ben-Sheppard:
### Description
To be able to support the [Figma designs](https://www.figma.com/design/0FhSUskdeATmYkEyXAj9S4/Identity-%26-Console-8.7?node-id=193-97189&t=tQB2ERCpxsxROx7W-0) for the Identity UI in which we show a list of Users within a given tenant, we need to expose a new endpoint.

This endpoint should:

- Accept the tenant ID
- Support searching the list of users
- Return a list of user entities

PR #27218, opened 2025-01-21
refactor: support generic key types in entityjoinrelation

### LINKED ISSUE #27266: Modify tenant handler to store/update tenant assignments by the tenantId as ID instead of key

### LINKED ISSUE #27326: Modify tenant assignment handlers to store assignment by username/userId instead of key

Ben-Sheppard:
## Description

The intention of this PR is to refactor three things:
1. Enable a smoother transition from keys to ids by switching the EntityJoinRelation to use generics, this allows the parent key to be a `String` in the case of a tenant, or a 'long` in the case of the other entities yet to be moved.
2. Remove the `assignedMemberKeys` from the `TenantEntity` object for clarity. This field is not populated when using ES/OS due to the document structures so we would have to populate them in different ways, although we can map them when using the RDBMS side, we would likely prefer a consistent approach (such as accessing the members of a tenant in a targeted call i.e. `/tenant/{id}/members?memberType=user`
3. Use entityId in the tenant member mappings instead of entityKey

For 2, I'm aware that we are removing a field however the functionality will be reintroduced with https://github.com/camunda/camunda/issues/27217 (the same will be true for other entities where memberships are resolved)

Closes https://github.com/camunda/camunda/issues/27326
Closes https://github.com/camunda/camunda/issues/27266

### DIFF
--- a/db/rdbms/src/main/resources/mapper/TenantMapper.xml
+++ b/db/rdbms/src/main/resources/mapper/TenantMapper.xml
@@ -20,15 +20,11 @@
     resultMap="io.camunda.db.rdbms.sql.TenantMapper.tenantResultMap">
     SELECT * FROM (
     SELECT
-    t.TENANT_KEY,
-    t.TENANT_ID,
-    t.NAME,
-    t.DESCRIPTION,
-    tm.TENANT_ID AS MEMBER_TENANT_ID,
-    tm.ENTITY_KEY AS MEMBER_ENTITY_KEY,
-    tm.ENTITY_TYPE AS MEMBER_ENTITY_TYPE
+    TENANT_KEY,
+    TENANT_ID,
+    NAME,
+    DESCRIPTION
     FROM TENANT t
-    LEFT JOIN TENANT_MEMBER tm ON t.TENANT_ID = tm.TENANT_ID
     <include refid="io.camunda.db.rdbms.sql.TenantMapper.searchFilter"/>
     ) t
     <include refid="io.camunda.db.rdbms.sql.Commons.keySetPageFilter"/>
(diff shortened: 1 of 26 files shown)


webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/EntityJoinRelation.java
```diff
@@ -20,12 +20,16 @@ public EntityJoinRelationFactory(
...
 
-    public EntityJoinRelation createParent() {
-      return new EntityJoinRelation(parent.getType(), null);
+    public EntityJoinRelation<?> createParent() {
+      return new EntityJoinRelation<>(parent.getType(), null);
     }
 
-    public EntityJoinRelation createChild(final long parentKey) {
-      return new EntityJoinRelation(child.getType(), parentKey);
+    public EntityJoinRelation<Long> createChild(final long parentKey) {
```
emulbreh, line 27: We could add a type parameter on the factory too and then only have one `createChild()` method. I know it's only needed temporarily, but it avoids creating an EntityJoinRelation of the wrong kind accidentally.

db/rdbms/src/main/resources/mapper/TenantMapper.xml
```diff
@@ -20,14 +20,10 @@
...
-    t.TENANT_KEY,
-    t.TENANT_ID,
-    t.NAME,
-    tm.TENANT_ID AS MEMBER_TENANT_ID,
-    tm.ENTITY_KEY AS MEMBER_ENTITY_KEY,
-    tm.ENTITY_TYPE AS MEMBER_ENTITY_TYPE
+    TENANT_KEY,
+    TENANT_ID,
+    NAME
     FROM TENANT t
```
emulbreh, line 26: If we don't use the alias, let's drop it here too.

emulbreh (approved):

stoerti (commented):
With this PR we are still storing the tenant members in the database but we
a) never test this (test is removed)
b) never access this data. Where and how do I get the members of a tenant now?

Ben-Sheppard:
Good questions @stoerti - I hadn't spotted that we're still storing members with the test removed, it is quite possible that I need to reinstate those tests as we do want to store the members, however we have removed them from the `TenantEntity` so at the moment the tests don't have a way to retrieve just the members to verify.

My intention was to address the retrieval of members with https://github.com/camunda/camunda/issues/27217 which is one of the next tasks we will look at, ultimately the end state we're heading for would be that I have a tenant id and I intentionally/directly retrieve the members just for that tenant ID (likely also using a specific type so I know the members I'm getting back).

Is that ok from your perspective? Or should we instead look to do it a different way within the RDBMS paths? (Its known that our structure means we cannot join in ES/OS) there is some more context in Slack [here](https://camunda.slack.com/archives/C07CTSU4Q9Z/p1733235204314249)

stoerti:
Ok. As long as it is fixed in the next PR then, I'm good with it.

emulbreh (approved):

PR #27229, opened 2025-01-21
feat: configure custom error handling per index

### LINKED ISSUE #26093: New OperationReference on Client commands are not validated and can break system

### LINKED ISSUE #27090: Allow Camunda Exporter index descriptors to decide how their persistence errors are handled

mustafadagher:
## Description

- ExporterResourceProvider to provide a map of indices and their custom error handling behavior
- Exporter passes the custom error handlers to the ExporterBatchWriter
- ExporterBatchWriter passes the error handlers through to be used when executing a BatchRequest
- Created ErrorHandler functional interface that receives Error details
- Created an enum with a couple of error handling implementations
- Only ignore operation index errors if the error was document missing

## Related issues

closes #27090
closes #26093

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
@@ -54,5 +54,12 @@ public interface ExporterResourceProvider {
    */
   Set<ExportHandler<?, ?>> getExportHandlers();
 
+  /**
+   * Possible custom error handlers to be used if certain indices threw persistence errors. The
+   * first parameter is the name of the index that threw the error and the second is the error
+   * details
+   *
+   * @return A {@link BiConsumer} of {@link String} and {@link Error} to handle custom errors
+   */
   BiConsumer<String, Error> getCustomErrorHandlers();
 }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
@@ -15,6 +15,7 @@ import io.camunda.webapps.schema.descriptors.IndexTemplateDescriptor;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.util.Collection;
 import java.util.Set;
+import java.util.function.BiConsumer;
 
 public interface ExporterResourceProvider {
 
@@ -51,4 +52,6 @@ public interface ExporterResourceProvider {
    * @return A {@link Set} of {@link ExportHandler} to be registered with the exporter
    */
   Set<ExportHandler<?, ?>> getExportHandlers();
+
+  BiConsumer<String, String> getCustomErrorHandlers();
 }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
@@ -9,6 +9,7 @@ package io.camunda.exporter;
 
 import io.camunda.exporter.cache.ExporterEntityCacheProvider;
 import io.camunda.exporter.config.ExporterConfiguration;
+import io.camunda.exporter.errorhandling.Error;
 import io.camunda.exporter.handlers.ExportHandler;
 import io.camunda.webapps.schema.descriptors.IndexDescriptor;
 import io.camunda.webapps.schema.descriptors.IndexTemplateDescriptor;
@@ -53,5 +54,5 @@ public interface ExporterResourceProvider {
    */
   Set<ExportHandler<?, ?>> getExportHandlers();
 
-  BiConsumer<String, String> getCustomErrorHandlers();
+  BiConsumer<String, Error> getCustomErrorHandlers();
 }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/Error.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/Error.java
@@ -7,4 +7,12 @@
  */
 package io.camunda.exporter.errorhandling;
 
+/**
+ * Represents error details for Elasticsearch/Opensearch persistence errors when executing batch
+ * requests in the exporter.
+ *
+ * @param message the error message
+ * @param type the error type communicated by the search engine
+ * @param status the HTTP status code of the error
+ */
 public record Error(String message, String type, int status) {}
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/Error.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/Error.java
@@ -1,0 +1,10 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.errorhandling;
+
+public record Error(String message, String type, int status) {}
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/ErrorHandlers.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/ErrorHandlers.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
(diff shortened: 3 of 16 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterResourceProvider.java
```diff
@@ -51,4 +53,6 @@ void init(
    * @return A {@link Set} of {@link ExportHandler} to be registered with the exporter
    */
   Set<ExportHandler<?, ?>> getExportHandlers();
+
+  BiConsumer<String, Error> getCustomErrorHandlers();
```
ChrisKujawa, line 57: 🔧 Can you please some java doc here :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/Error.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.errorhandling;
+
+public record Error(String message, String type, int status) {}
```
ChrisKujawa, line 10: 🔧 I think here some more java doc would be beneficial as well, why we have it where is it used or so

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/ErrorHandlers.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package io.camunda.exporter.errorhandling;
+
+import io.camunda.exporter.exceptions.PersistenceException;
+
+public enum ErrorHandlers implements ErrorHandler {
+  IGNORE_DOCUMENT_DOES_NOT_EXIST {
+    @Override
+    public void handle(final Error error) {
+      if (error.status() == 404 || error.type().equals("document_missing_exception")) {
```
ChrisKujawa, line 16: 👍🏼

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/errorhandling/ErrorHandlers.java
```diff
@@ -0,0 +1,28 @@
...
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.exporter.errorhandling;
+
+import io.camunda.exporter.exceptions.PersistenceException;
+
+public enum ErrorHandlers implements ErrorHandler {
+  IGNORE_DOCUMENT_DOES_NOT_EXIST {
```
ChrisKujawa, line 13: 🔧 Can we document these different types please as well

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ElasticsearchBatchRequest.java
```diff
@@ -221,16 +224,19 @@ public BatchRequest deleteWithRouting(final String index, final String id, final
...
   @Override
-  public void execute() throws PersistenceException {
-    execute(false);
+  public void execute(final BiConsumer<String, Error> customErrorHandlers)
+      throws PersistenceException {
+    execute(customErrorHandlers, false);
   }
 
   @Override
   public void executeWithRefresh() throws PersistenceException {
```
ChrisKujawa, line 233: ❓ Who is actually calling this?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ElasticsearchBatchRequest.java
```diff
@@ -241,15 +247,43 @@ private void execute(final boolean shouldRefresh) throws PersistenceException {
...
+            .map(
+                item ->
+                    String.format(
+                        "%s failed for type [%s] and id [%s]: %s",
+                        item.operationType(), item.index(), item.id(), item.error().reason()))
+            .collect(Collectors.joining(", \n"));
+    LOGGER.warn("Bulk request execution failed: \n[{}]", errorMessages);
+
+    errorItems.forEach(
+        item -> {
```
ChrisKujawa, line 276: 🔧 We iterate twice over the errorItems. Maybe we can combine these

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -411,6 +414,72 @@ void shouldExportRecord(
         });
   }
 
+  @TestTemplate
+  void shouldNotFailWhenUpdatingOperationWithNoDocument(
+      final ExporterConfiguration config, final SearchClientAdapter clientAdapter) {
+    // given
+    final ValueType valueType = ValueType.INCIDENT;
+    final Record record =
+        factory.generateRecord(
```
ChrisKujawa, line 423: ❓ Where are we setting the Operation reference, to cause the error?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -411,6 +414,72 @@ void shouldExportRecord(
...
+
+    camundaExporter.configure(exporterTestContext);
+    camundaExporter.open(new ExporterTestController());
+
+    // act
+    assertThatCode(() -> camundaExporter.export(record)).doesNotThrowAnyException();
+  }
+
+  @TestTemplate
+  void shouldThrowIfDateFormatIsInvalid(
```
ChrisKujawa, line 449: 👍🏼

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/store/ElasticsearchBatchRequestTest.java
```diff
@@ -373,7 +377,7 @@ void shouldExecuteWithMultipleOperationsInBatch() throws PersistenceException, I
     // When
     batchRequest.add(INDEX, entity);
     batchRequest.update(INDEX, ID, entity);
-    batchRequest.execute();
+    batchRequest.execute(null);
```
ChrisKujawa, line 380: 🔧 We could also overload the method, to have one without parameter.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/store/ElasticsearchBatchRequestTest.java
```diff
@@ -434,10 +438,33 @@ void shouldThrowPersistenceExceptionIfAResponseItemHasError() throws IOException
...
+
+    when(elasticsearchClient.bulk(any(BulkRequest.class))).thenReturn(bulkResponse);
+    final BiConsumer<String, Error> errorHandler = mock(BiConsumer.class);
+
+    // When
+    batchRequest.add(INDEX_WITH_HANDLER, entity);
+    batchRequest.execute(errorHandler);
+
+    // Then
+    verify(errorHandler).accept(eq(INDEX_WITH_HANDLER), any());
```
ChrisKujawa, line 468: 🔧 We could also validate the error that is passed in, might be good to validate that we pass in the correct data.
```suggestion
    verify(errorHandler).accept(eq(INDEX_WITH_HANDLER), eq(new Error(...)));
```

ChrisKujawa (approved):
Thanks @mustafadagher I think this is the best approach we could do for now 👍🏼 I like it more than the others. Good job. 

I have some smaller nitpicks, and requests regarding java doc, other than that looks good. Pre-approving, please take a look at my comments before merging :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ElasticsearchBatchRequest.java
mustafadagher: My goal here was to log all the errors, disregarding if we will throw for them or not.
The first iteration is to collect the complete log message. if I combine them, we might end up throwing before logging the rest of the errors

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/store/ElasticsearchBatchRequest.java
mustafadagher: only tests, as of now, I think

PR #27233, opened 2025-01-22
refactor: cleanup C7 demo distribution logic

abremard:
## Description

This PR removes C7 demo distribution logic from `camunda-optimize` module (slack [thread](https://camunda.slack.com/archives/CUZUYDMMG/p1737438854104589)).
There's a lot of lines being deleted but it's mostly configs and scripts being removed entirely.

## Related issues

related to https://github.com/camunda/camunda-optimize/issues/13981


optimize-distro/src/README.txt
```diff
@@ -15,7 +15,7 @@ Contents:
                     which can be used to overwrite default values of Optimize configuration. Also the logging
                     level can be configured here.
 
-        elasticsearch/elasticsearch-${elasticsearch.demo.version}/ (full distribution only)
+        elastic/config/
```
abremard, line 18: It seems like the current version of the README doesn't reflect the current folders structure so I'm updating it to point to the config folder (also removing (full distribution only) as we no longer need the distinction).

optimize-distro/src/README.txt
```diff
@@ -30,6 +30,6 @@ Contents:
 ==================
 
 Camunda Optimize version: ${project.version}
-Elasticsearch version: ${elasticsearch.demo.version}
+Elasticsearch version: ${version.elasticsearch}
```
abremard, line 33: I'm updating the version to be the one from `zeebe-parent` which is also used in `optimize-backend`.

optimize-distro/src/README.txt
grlimacan: Since we are removing the `demo` distribution, you can also delete everything that is related to elastic search, including this line and the line below in the readme. Please also remove the `elastic` folder completely

optimize-distro/assembly-distro-production.xml
```diff
@@ -20,7 +20,6 @@
       <directory>src/config/</directory>
```
grlimacan, line 20: in the `config` folder you also need to remove the `OptimizeLicense.txt`, since this type of license is only applicable to C7

optimize-distro/assembly-distro-production.xml
```diff
@@ -20,7 +20,6 @@
...
       <excludes>
         <exclude>${exclude.from.config}</exclude>
-        <exclude>environment-config-demo.yaml</exclude>
```
grlimacan, line 23: Please also remove that file which is located in the `config` folder

optimize-distro/assembly-distro-production.xml
```diff
@@ -20,7 +20,6 @@
...
       </excludes>
       <outputDirectory>config/</outputDirectory>
```
grlimacan, line 24: the `curl.exe` file is also only used by the demo files, so please remove it as well

grlimacan (changes requested):
Some more files to be removed. Please add your PR to the `Pull requests` section in the issue. And please do not link the PR directly to the issue as it is now, because this will lead GitHub to close the issue once you merge the PR, but we don't want that. Please reach out if you have any questions about this

grlimacan (approved):
LGTM!

PR #27241, opened 2025-01-22
test: fix user cache tests ordering assertion

abremard:
## Description

When running unit tests locally, we've noticed a [small bug](https://camunda.slack.com/archives/CUZUYDMMG/p1737534872163439) where 3 tests were failing because the assertion was checking for strict ordering of items in a list. Since the ordering shouldn't matter, the fix is to replace `containsExactly` with `containsExactlyInAnyOrder`.


grlimacan (approved):
LGTM! Thanks!

PR #27247, opened 2025-01-22
ci: allow ignoring vacuum lint rules

lenaschoenburg:
This allows us to ignore certain vacuum lint rules to make exceptions during development, break rules when it's necessary etc.

I've used the opportunity to redefine our ruleset as a customization of the recommended ruleset. This way it's easier to maintain and see where we deviate from the recommendations.


lenaschoenburg:
Successful run is here: https://github.com/camunda/camunda/actions/runs/12905627942/job/35985518938

tmetzke (approved):
👍 Looks good to me, thanks, @lenaschoenburg 🚀 
ℹ️ The old custom ruleset removed all OAS2 rules from the recommended ruleset but I think that's not necessary, so we're good with what we have now with this PR.

PR #27248, opened 2025-01-22
Implement CamundaExporter::purge and parameterize ClusterPurgeIT

### LINKED ISSUE #26043: Implement `Exporter#purge` for the Camunda exporter

srsp:
## Description

Implement `purge()` for the Camunda Exporter using ES.

- Delete indices
- Delete index templates
- Delete ILM policies

## Related issues

closes #26043

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
@@ -29,6 +29,7 @@ import static io.camunda.zeebe.protocol.record.ValueType.VARIABLE_DOCUMENT;
 import io.camunda.exporter.adapters.ClientAdapter;
 import io.camunda.exporter.config.ConfigValidator;
 import io.camunda.exporter.config.ExporterConfiguration;
+import io.camunda.exporter.config.ExporterConfiguration.RetentionConfiguration;
 import io.camunda.exporter.exceptions.PersistenceException;
 import io.camunda.exporter.metrics.CamundaExporterMetrics;
 import io.camunda.exporter.schema.MappingSource;
@@ -51,6 +52,8 @@ import io.camunda.zeebe.protocol.record.ValueType;
 import io.camunda.zeebe.util.SemanticVersion;
 import io.camunda.zeebe.util.VisibleForTesting;
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Set;
 import org.agrona.CloseHelper;
 import org.slf4j.Logger;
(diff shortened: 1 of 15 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -206,6 +208,31 @@ public void export(final Record<?> record) {
...
 
+  @Override
+  public void purge() throws Exception {
+    provider.getIndexDescriptors();
+    provider.getIndexTemplateDescriptors();
+    final var client = clientAdapter.getSearchEngineClient();
+
+    //    final var indexMapping = client.getMappings("operate-*,tasklist-*", MappingSource.INDEX);
+    final var indexMapping =
+        client.getMappings("camunda-*,operate-*,tasklist-*", MappingSource.INDEX);
```
srsp, line 219: Are these all the indices we need to delete or are there more?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -206,6 +208,31 @@ public void export(final Record<?> record) {
...
+    //    final var indexMapping = client.getMappings("*", MappingSource.INDEX);
+    LOG.info("Purging '{}' existing indices.", indexMapping.size());
+    indexMapping.keySet().forEach(client::deleteIndex);
+
+    //    List.of("operate-*", "tasklist-*", "metrics-*", "ilm-*")
+    Stream.of("operate-*", "tasklist-*")
```
srsp, line 225: Are these all index templates we need to delete?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
ChrisKujawa: You need to make sure that the indices are prefixed, you can find this in the config.

ChrisKujawa: Otherwise it is correct (for now, util we add Optimize)

This overview might help (showing runtime, indices) History indices are not shown, but they follow similar pattern.
harmonized-indices-miro-v2

PR #27252, opened 2025-01-22
feat: Improve `changedAttributes` handling for (`UPDATING`, `UPDATED`, `CORRECTED`)  user task events

### LINKED ISSUE #27212: Refactor how `changedAttributes` is filled for `UPDATING`, `UPDATED`, `CORRECTED` events
## Description  
Currently, the `record.changedAttributes` field in user task events (`UPDATING`, `UPDATED`, and `CORRECTED`) does not accurately reflect the attributes that were truly modified through the `UPDATE` command or the corrections feature. Attributes are marked as changed even if their values remain the same as those already persisted in the user task.

This task focuses on refactoring the mechanism for populating the user task `changedAttributes` list to ensure it accurately tracks only the attributes that have been effectively changed. 

### Objectives  
1. Ensure that `changedAttributes` includes only the attributes whose values differ from the persisted user task state.
2. Avoid tracking attributes that are updated or corrected to their current persisted values.
3. Improve the corrections functionality, ensuring `CORRECTED` event triggers only when meaningful changes were done.

### Acceptance criteria  
- [x] The `recor.changedAttributes` accurately reflects only the truly modified attributes.
- [x] The changes are applied consistently across all relevant user task events (`UPDATING`, `UPDATED`, `CORRECTED`).
- [x] No regressions in existing functionality, validated through comprehensive testing.

ce-dmelnych:
## Description
This PR refactors the mechanism for populating the `changedAttributes` field in user task events (`UPDATING`, `UPDATED`, and `CORRECTED`) to ensure it accurately tracks only the attributes that have been genuinely modified.  
Previously, the `changedAttributes` list could include attributes that were set to the same value as their persisted state or unknown attributes, leading to unnecessary updates and redundant tracking.

### Key improvements  
- The `changedAttributes` field now includes only attributes that were actually changed:  
  - For `UPDATING` events, it contains only attributes whose values differ from the [persisted state](https://github.com/camunda/camunda/blob/ad03aca0e7baef3a33a1177455233829304b179c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/DbUserTaskState.java#L181).  
  - For `UPDATED` events, it contains only attributes whose values differ either from the [persisted state](https://github.com/camunda/camunda/blob/ad03aca0e7baef3a33a1177455233829304b179c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/DbUserTaskState.java#L181) or from the [intermediate state](https://github.com/camunda/camunda/blob/ad03aca0e7baef3a33a1177455233829304b179c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/DbUserTaskState.java#L190), if any task listeners were executed during the update process.  
  - For `CORRECTED` events, it contains only attributes that differ from the previous event (either the event that triggered the listener or a prior listener correction), stored in the [intermediate user task record state](https://github.com/camunda/camunda/blob/ad03aca0e7baef3a33a1177455233829304b179c/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/instance/DbUserTaskState.java#L190).  
- Attributes updated/corrected to the same value as the current persisted/intermediate state are excluded.
- The `CORRECTED` intent is emitted only when meaningful corrections are applied.  
- Ensures unknown attributes are ignored during updates, preventing them from being included in the `changedAttributes` list.  
- Comprehensive tests validate the behavior of `changedAttributes` tracking across all relevant scenarios, including updates, corrections, and handling of unknown attributes.  

## Checklist

- [ ] for CI changes:

## Related issues

closes #27212

### DIFF
--- a/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
+++ b/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
@@ -177,13 +177,31 @@ public final class UserTaskRecord extends UnifiedRecordValue implements UserTask
     return copy;
   }
 
+  /**
+   * Updates the attributes of this {@link UserTaskRecord} based on the given record.
+   *
+   * @apiNote If {@code includeTrackingProperties} is {@code true}, all attributes in the given
+   *     record's `changedAttributes` list will be added to this record's `changedAttributesProp`,
+   *     even if their values match the existing values. To update attributes and track only truly
+   *     changed attributes, use the {@link #wrapChangedAttributesIfValueChanged} method instead.
+   * @param record the record containing the changed attributes list and new attribute values
+   * @param includeTrackingProperties whether to include all changed attributes in the tracking list
+   */
   public void wrapChangedAttributes(
       final UserTaskRecord record, final boolean includeTrackingProperties) {
-    record.getChangedAttributesProp().stream()
-        .forEach(attribute -> updateAttribute(bufferAsString(attribute.getValue()), record));
     if (includeTrackingProperties) {
-      setChangedAttributesProp(record.getChangedAttributesProp());
+      changedAttributesProp.reset();
     }
+
+    record
+        .getChangedAttributes()
+        .forEach(
+            attribute -> {
+              updateAttribute(attribute, record);
+              if (includeTrackingProperties) {
+                addChangedAttribute(attribute);
+              }
+            });
   }
 
   /**
--- a/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
+++ b/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
@@ -589,18 +589,14 @@ public final class UserTaskRecord extends UnifiedRecordValue implements UserTask
   }
 
   private boolean isAttributeValueChanged(final String attribute, final UserTaskRecord other) {
-    final var attributeGetter = resolveAttributeGetter(attribute);
+    final var attributeGetter = ATTRIBUTE_GETTER_MAP.get(attribute);
+    if (attributeGetter == null) {
+      return false;
+    }
+
     final var thisAttributeValue = attributeGetter.apply(this);
     final var otherAttributeValue = attributeGetter.apply(other);
     return !Objects.equals(thisAttributeValue, otherAttributeValue);
-  }
-
-  private Function<UserTaskRecord, ?> resolveAttributeGetter(final String attribute) {
-    final Function<UserTaskRecord, ?> getter = ATTRIBUTE_GETTER_MAP.get(attribute);
-    if (getter == null) {
-      throw new IllegalArgumentException("Unknown attribute: " + attribute);
-    }
-    return getter;
   }
 
   @JsonIgnore
--- a/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
+++ b/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
@@ -105,23 +105,23 @@ public final class UserTaskRecord extends UnifiedRecordValue implements UserTask
       new StringProperty("tenantId", TenantOwned.DEFAULT_TENANT_IDENTIFIER);
 
   /**
-   * Tracks the names of user task attributes that have been modified.
+   * Tracks the names of user task attributes that are intended to be modified (e.g. on `UPDATE`),
+   * or have been modified (e.g. on `UPDATED` or `CORRECTED`).
    *
-   * <p>This property was introduced to enable tracking of attributes that were truly modified.
-   * Since our `msgpack` implementation doesn't support `null` values, all record attributes have
-   * default non-null values. As a result, without `changedAttributesProp`, it would be impossible
-   * to determine whether an attribute was explicitly modified or simply retained its default value.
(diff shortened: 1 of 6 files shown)


korthout (approved):
Great pull request @ce-dmelnych 👏 

👍 This contains a lot of great improvements!

In the pull request body (description) you mention the following:

>The changedAttributes field now includes only attributes whose values differ from the persisted state.

This was a bit confusing to me, because this is not the case for the `CORRECTED` events. For the `CORRECTED` events, the `changedAttributes` field contains only those attributes that differ from the previous event, either the event that triggered the listener, or the completion of a prior listener. Iternally this is now implemented by diffing with the intermediate state, which is probably where this statement came from.

🔧 I think it would be useful to document what the `changedAttributes` property of a `UserTaskRecord` contains under certain conditions with some JavaDoc on the field. We can clarify that:
- it is used to specify which attributes are intended to be changed by an `UPDATE` command to allow updating to `null`
- it is used to specify which attributes are `UPDATING` or have been `UPDATED`, including any corrections made by listeners
- it is used to specify which attributes have been `CORRECTED` by a listener, but that this does not contain the changes that the listener initially responded to

Aside from this suggestions, I have no other remarks.

LGTM 👍

ce-dmelnych:
Thanks a lot, @korthout! I really appreciate your review and feedback. 🙌  

I've updated the PR description to better reflect what was changed for `UPDATING`, `UPDATED`, and `CORRECTED` events in context of `changeAttributes` handling.  
Additionally, I’ve added JavaDoc for `changedAttributesProp` in this commit: https://github.com/camunda/camunda/pull/27252/commits/1a8b3abeadc2f470dd93f2cf6e00173d05075634.  

When you have a moment, could you take a glance and let me know if I’ve covered all the cases for how this property is used depending on the context?

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
```diff
@@ -103,8 +103,42 @@ public final class UserTaskRecord extends UnifiedRecordValue implements UserTask
...
       new StringProperty("tenantId", TenantOwned.DEFAULT_TENANT_IDENTIFIER);
+
+  /**
+   * Tracks the names of user task attributes that have been modified.
+   *
+   * <p>This property was introduced to enable tracking of attributes that were truly modified.
+   * Since our `msgpack` implementation doesn't support `null` values, all record attributes have
+   * default non-null values. As a result, without `changedAttributesProp`, it would be impossible
+   * to determine whether an attribute was explicitly modified or simply retained its default value.
+   * This property solves that issue by explicitly listing attributes that were changed.
```
korthout, line 114: 🔧 We can add some more clarity to the first sentence by being a bit more explicit. This allows us to remove the second sentence. I also suggest to use `changedAttributes` over `changedAttributesProp` as this is the name of the property.

```suggestion
   * Tracks the names of user task attributes that are intended to be modified (e.g. on `UPDATE`), or have been modified (e.g. on `UPDATED`).
   *
   * Since our `msgpack` implementation doesn't support `null` values, all record attributes have
   * default non-null values. As a result, without `changedAttributes`, it would be impossible
   * to determine whether an attribute was explicitly modified or simply retained its default value.
   * This property solves that issue by explicitly listing attributes that were changed.
```

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
```diff
@@ -103,8 +103,42 @@ public final class UserTaskRecord extends UnifiedRecordValue implements UserTask
...
+   *
+   * <p>The content of `changedAttributesProp` field depend on the type of event being processed:
```
korthout, line 116: 🔧 It's not just events but also for commands, so any record.
```suggestion
   * <p>The content of `changedAttributesProp` field depend on the type of record:
```

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
```diff
@@ -103,8 +103,42 @@ public final class UserTaskRecord extends UnifiedRecordValue implements UserTask
...
+   *
+   * <ul>
+   *   <li><b>UPDATE command:</b> Specifies which attributes are intended to be updated, allowing
+   *       explicit changes (including clearing values to default). The presence of an attribute in
+   *       this list indicates that it was explicitly modified by the update request.
+   *   <li><b>UPDATING and UPDATED events:</b> Contains only attributes whose values differ from the
+   *       persisted state or, in the case of `UPDATED`, from the intermediate user task record
+   *       state if task listeners were executed, including any corrections made by listeners.
```
korthout, line 124: ❌ I don't think this is correct, or we have a bug. `UPDATED` should contain all changes compared to the current persisted state.
```suggestion
   *   <li><b>UPDATING and UPDATED events:</b> Contains only attributes whose values differ from the
   *       persisted state.
```

We could add a line about listener corrections, but it's the same thing right?
```suggestion
   *   <li><b>UPDATING and UPDATED events:</b> Contains only attributes whose values differ from the
   *       persisted state. In the case of `UPDATED` this can include any corrections made by 
   *       listeners, unless they corrected them back to the prior state.
```

korthout (approved):
👍

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/usertask/UserTaskRecord.java
ce-dmelnych: Thanks a lot, @korthout! 🙌  

I've applied your recommendations 👍

ana-vinogradova-camunda (approved):
Great PR and improvements, thanks @ce-dmelnych

PR #27262, opened 2025-01-22
POC for env variables in feel expressions

eppdot:
While onboarding to the overall codebase and as preparation for the hackathon, I played around with the custom context we can provide to the `ExpressionLanguage`. 

I also stumbled across this [product hub epic](https://github.com/camunda/product-hub/issues/637) and thought it would be a good exercise to provide access to **environment variables within feel expressions**, for example:

```
// My environment variables on the (broker) node
export MY_TARGET_ENVIRONMENT=dev

// My Feel Expression in Modeler
= "https://" + MY_TARGET_ENVIRONMENT + ".service.cluster.de"
```

This PR is the outcome of my little PoC and I wanted to get some feedback if it's worth to continue.

What I did:
- Moved the `ExpressionProcessor` in a separate package that contains more expression relevant files.
- Removed the `combine` method in the `EvaluationContext` interface and replaced it, with a `CombinedEvaluationContext` allowing a bit more flexibility (especially for scoping)
- Refactored `ExpressionProcessor` to use a bit more flexible context handling
- Created a new `EnvVariableEvluationContext` that allows accessing environment variables with a pre-defined prefix
- Cleaned up some things and added a few tests

What's open:
- [ ] Check the way to convert to `MsgPack` (also left a comment)
- [ ] Define set of allowed prefixes by default
- [ ] write a couple of tests (e.g. for the `scoped` implementations of different contexts)
- [ ] Add some JavaDoc
- [ ] Add to camunda-docs (??)
- [ ] clarify the naming/prefix handling, e.g. `env.VAR_NAME` (based on [this message](https://github.com/camunda/product-hub/issues/637#issuecomment-1564433972) from @korthout)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/expression/EnvVariableEvaluationContext.java
```diff
@@ -0,0 +1,54 @@
...
+      if (value != null) {
+        return asMsgPack(value);
+      }
+    }
+
+    return null;
+  }
+
+  private static UnsafeBuffer asMsgPack(final String value) {
+    // @TODO is this the best way to convert to MsgPack?
```
eppdot, line 46: 💭 Not sure, if this is the best way to convert the value of the environment variable to something the feel expression can properly unpack.

[Here](https://github.com/camunda/camunda/blob/ab273e28f9920688886fb53f9cb6c62224aa4d07/zeebe/expression-language/src/main/java/io/camunda/zeebe/el/impl/FeelExpressionLanguage.java#L49) is defined that the feel engine uses the `MessagePackValueMapper` to resolve variable values. Also the `VariableState` seems to have variables stored using `MsgPack` encoding. But I feel that this connection is quite hidden, so it feels a bit awkward that the `EnvVariableEvaluationContext` needs to know that and is thereby implicitly coupled.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/expression/ScopedEvaluationContext.java
```diff
@@ -0,0 +1,20 @@
...
+
+import io.camunda.zeebe.el.EvaluationContext;
+
+public interface ScopedEvaluationContext extends EvaluationContext {
+  default ScopedEvaluationContext scoped(final long scopeKey) {
+    return this;
+  }
+
+  static ScopedEvaluationContext none() {
+    return variableName -> null;
```
eppdot, line 18: this is probably better defined once in a static field.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/expression/VariableStateEvaluationContext.java
```diff
@@ -0,0 +1,41 @@
...
+    this.scopeKey = scopeKey;
+  }
+
+  @Override
+  public DirectBuffer getVariable(final String variableName) {
+    if (scopeKey < 0) {
+      return null;
+    }
+
+    return variableState.getVariable(scopeKey, BufferUtil.wrapString(variableName));
```
eppdot, line 34: 💭 Here we are using `BufferUtil` for the variable name, but get back a value that is encoded via `MsgPack`. I think I need some more insights where we use which encoding, etc. 😅

eppdot:
@saig0 @korthout do you think we can close this PR? or does it still serve a purpose?

korthout:
@eppdot IMO it can be closed. Lot's of this is already available on `main`, and for reference the changes remains visible even when the PR is closed.

Issue #27166, opened 2025-01-20
A description field is added to the Tenant entity


Ben-Sheppard:

lenaschoenburg:
We will not support filtering/sorting for the tenant description, at least for now: https://camunda.slack.com/archives/C06UYJMMETZ/p1737544402972389?thread_ts=1737544190.274919&cid=C06UYJMMETZ

lenaschoenburg:
Updating a tenant requires both name and description fields to be set, even if only one of them changes. The unchanged field must be set to the old value. https://camunda.slack.com/archives/C06UYJMMETZ/p1737554902650929?thread_ts=1737544190.274919&cid=C06UYJMMETZ

PR #27263, opened 2025-01-22
feat: add description field to tenants in Identity client

Ben-Sheppard:
**Requires the backend implementation adding a description to a Tenant**

## Description

This PR adds in the description field for a Tenant in the Identity UI, the Figma designs for the Tenant pages can be found [here](https://www.figma.com/design/0FhSUskdeATmYkEyXAj9S4/Identity-%26-Console-8.7?node-id=239-175225&t=5yKzr15ZbTVBvzHq-0)

## Related issues

This PR is the UI parts of https://github.com/camunda/camunda/issues/27166

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/tenant/TenantUpdateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/tenant/TenantUpdateProcessor.java
@@ -104,7 +104,7 @@ public class TenantUpdateProcessor implements DistributedTypedRecordProcessor<Te
       existingTenant.setName(updatedName);
     }
     final var updatedDescription = updateRecord.getDescription();
-    if (!updatedDescription.isEmpty()) {
+    if (!existingTenant.getDescription().equals(updatedDescription)) {
       existingTenant.setDescription(updatedDescription);
     }
   }
--- a/identity/client/src/pages/tenants/List.tsx
+++ b/identity/client/src/pages/tenants/List.tsx
@@ -88,14 +88,22 @@ const List: FC = () => {
             label: t("Rename"),
             icon: Edit,
             onClick: (tenant) =>
-              editTenant({ tenantId: tenant.tenantId, name: tenant.name }),
+              editTenant({
+                tenantId: tenant.tenantId,
+                name: tenant.name,
+                description: tenant.description,
+              }),
           },
           {
             label: t("Delete"),
             icon: TrashCan,
             isDangerous: true,
             onClick: (tenant) =>
-              deleteTenant({ tenantId: tenant.tenantId, name: tenant.name }),
+              deleteTenant({
+                tenantId: tenant.tenantId,
+                name: tenant.name,
+                description: tenant.description,
+              }),
           },
         ]}
       />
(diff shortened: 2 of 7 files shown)


lenaschoenburg (approved):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/tenant/TenantUpdateProcessor.java
```diff
@@ -104,7 +104,7 @@ private void updateExistingTenant(
       existingTenant.setName(updatedName);
     }
     final var updatedDescription = updateRecord.getDescription();
-    if (!updatedDescription.isEmpty()) {
+    if (!existingTenant.getDescription().equals(updatedDescription)) {
```
emulbreh, line 107: ❓ Isn't the purpose of these `isEmpty` checks to enable partial updates, e.g. only setting the description and leaving the name unchanged?

emulbreh (approved):

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/tenant/TenantUpdateProcessor.java
Ben-Sheppard: I noticed an issue with the isEmpty checks blocking the removal of a description (i.e. if I set it to `""` it keeps the old description which felt counter intuitive.

lenaschoenburg: We decided against supporting partial updates. Both name and description have to be provided when updating a tenant.

emulbreh: If we don't do partial updates, can we remove all the `isEmpty()` checks in the update processors?

emulbreh: One more question: can we set the description unconditionally or is there a side effect when we call the setter?

lenaschoenburg: I'm only seeing one `isEmpty` check for the name which has to stay because we don't want a blank name. Do you mean other `isEmpty` checks than that?

emulbreh: There are several in the [other update processors](https://github.com/camunda/camunda/blob/59e3d72b4253852f3075851bd166a923c05bc851/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/user/UserUpdateProcessor.java#L104).
 If we don't support partial updates, wouldn't validation catch an empty name? The way it works now is that if a full update with an empty name is processed we silently ignore the name, although this indicates a problem upstream.

lenaschoenburg: I see. The decision to not support partial updates was made only for tenants, not necessarily the other entities.

I agree that ignoring a blank name here is a bit weird :+1:

PR #27267, opened 2025-01-22
feat: add tenant descriptions

lenaschoenburg:
This adds a description field to the Tenant entity.
Tenants can be created without specifying a description, the description is set to `""` internally.
Tenants can be updated but both name and description have to be provided, we don't support partial updates.
We do not support filtering or sorting based on the description.

The description is exported as a `text` field on ES/OS and as a `varchar(255)` on RDBMS.

relates to https://github.com/camunda/camunda/issues/27166

### DIFF
--- a/service/src/test/java/io/camunda/service/TenantServiceTest.java
+++ b/service/src/test/java/io/camunda/service/TenantServiceTest.java
@@ -140,8 +140,6 @@ public class TenantServiceTest {
   public void shouldUpdateTenantName() {
 
     // given
-    final var result =
-        new SearchQueryResult<>(1, List.of(tenantEntity), Arrays.array(), Arrays.array());
     final var tenantDTO =
         new TenantDTO(tenantEntity.key(), tenantEntity.tenantId(), "UpdatedTenantId", null);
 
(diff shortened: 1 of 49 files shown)


webapps-schema/src/main/resources/schema/elasticsearch/create/index/camunda-tenant.json
```diff
@@ -14,6 +14,9 @@
       "name": {
         "type": "keyword"
       },
+      "description": {
+        "type": "text"
```
Ben-Sheppard, line 18: 💭 Just thinking out loud here, as this is a text field it means the field is effectively parsed/tokenised and stored in the index in a way that allows partial matching, I think thats fine here as it would match the future usage of the field of finding all the tenants where a description contains `something`. The alternative is to use `keyword` which doesn't perform the parsing so I'm not sure of the implications there.

Ben-Sheppard (approved):
One NABD comment and a thought on my side but I'll approve early :)

webapps-schema/src/main/resources/schema/elasticsearch/create/index/camunda-tenant.json
lenaschoenburg: Right but `keyword` has problems too, for example failing on very large values. Not sure which is better here tbh.

service/src/test/java/io/camunda/service/TenantServiceTest.java
```diff
@@ -121,7 +121,8 @@ public void shouldThrowExceptionIfNotFoundByKey() {
   @Test
   public void shouldCreateTenant() {
     // given
-    final var tenantDTO = new TenantDTO(100L, "NewTenantName", "NewTenantId");
+    final var tenantDTO =
+        new TenantDTO(100L, "NewTenantName", "NewTenantId", "NewTenantDescription");
```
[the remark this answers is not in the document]
lenaschoenburg: Added one in 5cb3c10

webapps-schema/src/main/resources/schema/elasticsearch/create/index/camunda-tenant.json
Ben-Sheppard: Yea, like I say I was just thinking out loud, I think for our purposes `text` is fine (happy for you to seek other opinions though!)

PR #27268, opened 2025-01-22
Migrated backup metrics from prometheus to micrometer

### LINKED ISSUE #27083: Migrate all Prometheus metrics in zeebe-backup
**Description**

Migrate all Prometheus metrics in the zeebe-backup module.

**TODO**

- [x] BackupManagerMetrics
- [x] CheckPointMetrics

filipecampos:
## Description

Converted the existing Prometheus metrics into Micrometer ones.

## Checklist

TODO BEFORE MERGING

- [ ] Duplicate timers metrics without time units to match the Prometheus histogram ones or update the Grafana dashboard

## Related issues

closes #27083

### DIFF
--- a/zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
+++ b/zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
@@ -7,12 +7,12 @@
  */
 package io.camunda.zeebe.backup.metrics;
 
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.Timer.Sample;
-import io.prometheus.client.Counter;
-import io.prometheus.client.Gauge;
-import io.prometheus.client.Histogram;
-import io.prometheus.client.Histogram.Timer;
 import java.time.Duration;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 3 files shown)


zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
```diff
@@ -72,36 +76,81 @@ public OperationMetrics startDeleting() {
...
-      return new OperationMetrics(partitionId, timer, operation);
+      final Timer.Sample timerSample = Timer.start(METER_REGISTRY);
+      getOperationsInProgressGauge(partitionId, operation).incrementAndGet();
+      return new OperationMetrics(partitionId, timerSample, operation);
     }
 
     public <T> void complete(final T ignored, final Throwable throwable) {
-      timer.close();
-      OPERATIONS_IN_PROGRESS.labels(partitionId, operation).dec();
+      if (timerSample != null) {
```
entangled90, line 120: :question: why is the null check needed? Is `null` passed to the constructor?

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
```diff
@@ -25,29 +31,27 @@ public class BackupManagerMetrics {
...
-          .name("backup_operations_latency")
-          .help("Latency of backup operations")
-          .labelNames(LABEL_NAME_PARTITION, LABEL_NAME_OPERATION)
-          .buckets(0.01, 0.1, 1, 10, 60, 5 * 60)
-          .register();
+  private static final String BACKUP_OPERATIONS_TIMER_NAME =
+      NAMESPACE + "_backup_operations_latency";
+  private static final String BACKUP_OPERATIONS_IN_PROGRESS_NAME =
+      NAMESPACE + "_backup_operations_in_progress";
+  private static final MeterRegistry METER_REGISTRY = Metrics.globalRegistry;
```
entangled90, line 38: :x: : I would use the `MetricRegistry` that is associated to a partition, so it cannot be static. This will allow metrics to be unregistered when the registry is closed (for example a broker is not part of a partition anymore).

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
```diff
@@ -25,29 +31,27 @@ public class BackupManagerMetrics {
...
+  private static final Counter.Builder TOTAL_OPERATIONS_BUILDER =
+      Counter.builder(NAMESPACE + "_backup_operations_total")
+          .description("Total number of backup operations");
+  private static final ConcurrentHashMap<String, AtomicLong> OPERATIONS_IN_PROGRESS_GAUGES =
```
entangled90, line 42: :wrench: with the suggestion above, this map should not be static and it can just take the `operation` as key instead of a composed string with partition + operation.

The map can also be "preallocated", since all possible operations are known beforehand (in contrast to partitions)

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/CheckpointMetrics.java
```diff
@@ -7,39 +7,30 @@
...
-          .register();
-
-  private static final Gauge CHECKPOINT_ID =
-      Gauge.build()
-          .namespace(NAMESPACE)
-          .name("checkpoint_id")
-          .help("Id of the last checkpoint")
-          .labelNames(LABEL_NAME_PARTITION)
-          .register();
+  private static final MeterRegistry METER_REGISTRY = Metrics.globalRegistry;
```
entangled90, line 23: :x: Same as the other file, can we use a `MeterRegistry` tied to a partition?

entangled90 (changes requested):
I would avoid using a static MeterRegistry in favour of a MeterRegistry that is associated to a partition.
I know it's a bit more work to properly pass it down all the components, but it has many advantages.

PR #27278, opened 2025-01-22
[CPT] fix: Process instance state assertion

saig0:
## Description

The integration test fails at the assertion:

```
assertThat(processInstance)
        .hasCompletedElements(byName("Request tracking code")) // --> ok
        .hasActiveElements(byName("Received tracking code"))   // --> ok
        .isActive();                                           // --> failure!
```

Fixing the issue by resetting the internal cache of the process instance assertion. Before, the `failFast` condition was fulfilled because the process instance had the expected state. As a result, it threw an exception and the assertion failed.

## Related issues

no issue


berkaycanbc (approved):
Tricky one 🔎 

LGTM 👍

Issue #27237, opened 2025-01-22
ListView index timestamps serialization


panagiotisgts:
https://jira.camunda.com/browse/SUPPORT-25490
https://jira.camunda.com/browse/SUPPORT-25356

### Describe the bug

When an imported UserTask record timestamp is on the 0 nano (i.e 2025-01-22T07:45:10.000+0000) the list-view entity which takes the OffsetDateTime of the original UserTask entity and simply toString's it to get the value of the timestamp in String format, omits the nanos, thus generating a string of 2025-01-22T07:45:10Z , which is not compatible with the ES index format. Causing the importer to get stuck reimporting the record with the error :
```
Operation failed: ElasticsearchException[Elasticsearch exception [type=document_parsing_exception, reason=[1:231] failed to parse field
 [creationTime] of type [date] in document with id '2251799813685343'. Preview of field's value: '2025-01-22T07:45:10Z']]; nested: 
ElasticsearchException[Elasticsearch exception [type=illegal_argument_exception, reason=failed to parse date field [2025-01-22T07:45:10Z] 
with format [date_time || epoch_millis]]]; nested: ElasticsearchException[Elasticsearch exception [type=date_time_parse_exception, 
reason=Failed to parse with all enclosed parsers]];
```
### To Reproduce
In order to reproduce this you'd have to inject a record on Zeebe with a predetermined timestamp that's on **0** nanos. Or you could simply modify the [UserTaskEntityMapper](https://github.com/camunda/camunda/blob/017676656213fab8ae30636a4c367e56b4130fbe/tasklist/importer-860/src/main/java/io/camunda/tasklist/zeebeimport/v860/processors/common/UserTaskRecordToTaskEntityMapper.java#L103) to modify the creation/completion time:
```
 .setCreationTime(
  - DateUtil.toOffsetDateTime(Instant.ofEpochMilli(record.getTimestamp())));
  + DateUtil.toOffsetDateTime(Instant.ofEpochMilli(record.getTimestamp()))
  +  .withNano(0));
```

### Actual behavior
The importer throws the above error and get's stuck trying to import the same record.

### Expected behavior
The record is imported by Tasklist importer and it's visible on ES/OS `tasklist-list-view` index.

### Environment

- First encountered on SaaS, reproduced locally
- Tasklist Version: 8.6

### Additional context
Reference incident channel: [#inc-support-25356-no-tasks-created-in-tasklist](https://camunda.slack.com/archives/C088WM94JNS)

[...]

PR #27293, opened 2025-01-22
refactor: remove tasklist-list-view index

panagiotisgts:
## Description
The `tasklist-list-view` index introduced with 8.6 is now obsolete, since on 8.8 the `tasklist-task` index is planned to accommodate all the required changes to support the V2 APIs. In order not to create any more data that will be trashed later on it's better to remove it already.

## Checklist

- [ ] for CI changes:

## Related issues

closes #[27237](https://github.com/camunda/camunda/issues/27237)


nathansandi (approved):
Looks good to me, I have only one question related to the removal. 

❓  Is it possible verify if the index exist on Schema Validator/Manager, and if exists, delete it?

💭  I think also we can discuss and may be work on that in a separate issue (if it is the case)

panagiotisgts:
> ❓ Is it possible verify if the index exist on Schema Validator/Manager, and if exists, delete it?

We could take a look in another issue regarding this, it makes sense to delete this index since it may already have occupied a big chunk of size for Tasklist heavy customers. I'm not familiar if there's any specific approach/strategy for index deletion, we could discuss this.

Issue #27204, opened 2025-01-21
[C8 API] Make string keys the default in stable/8.7


danielkelemen:
## Description

Make String keys the default in the new 8.7 branch once it's available.

## Acceptance criteria

- OpenAPI default `application/json` returns String requests and responses.
- Gateway `LongSerializer` returns default string keys.
- Java client:
  - Theoretically no change required as it can work with string keys by default.
    - Alternatively, we could add the extra key header to make sure.
  - Backport agent header based serializing. (https://github.com/camunda/camunda/issues/26653)

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda/pull/27303
```

PR #27303, opened 2025-01-22
feat: change API default keys to string

danielkelemen:
## Description

- Adjust LongSerializer with string default & read client agent header
- Change OpenAPI default requests & responses to string keys
- Adjust controller tests for default string

## Related issues

closes https://github.com/camunda/camunda/issues/27204


tmetzke (approved):
👍 Nice, great that the switch is that easy 😅 

❓ Just to make sure I don't get lost:
* The request part is still pending for 8.6 and thus also for 8.7, right?
* We'll need to create a follow-up for 8.7 when this is merged for 8.6 and backported to 8.7, so we flip the default there as well in 8.7?
* This PR technically does not close #27204 yet?

danielkelemen:
No, this closes the ticket.

The requests are also [merged in 8.6](https://github.com/camunda/camunda/pull/27062) and [forward ported](https://github.com/camunda/camunda/pull/27269) into stable/8.7.
Or maybe I am missing something? 😄

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -559,7 +559,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/UserTaskSearchQueryResponse"
+                $ref: "#/components/schemas/UserTaskSearchQueryResult"
```
tmetzke, line 562: 🔧 I believe we missed the user task search **request** here; it's only number-based and string is not the default.

tmetzke:
> No, this closes the ticket.
> 
> The requests are also [merged in 8.6](https://github.com/camunda/camunda/pull/27062) and [forward ported](https://github.com/camunda/camunda/pull/27269) into stable/8.7. Or maybe I am missing something? 😄

Ah, forget what I wrote, yes, of course, all is fine. I was wondering about the few changes but we don't have much API in 8.7 now, of course. I can also see the adjusted request types now. Some might be missing though, at least the user task search query is missing media types completely (probably in 8.6 as well then).

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
danielkelemen: Yes, the `/user-tasks/search` is indeed missing.

tmetzke: As far as I can see, it is the only one, though.

danielkelemen: I couldn't find anything else either. 🤞

tmetzke: Do we need a fix for 8.6 as well, adding the alternative media types with number as default?

danielkelemen: Yes, but good that you say that there number is default. First I just cherry picked the same commit: https://github.com/camunda/camunda/pull/27349
It's confusing... 😄

PR #27316, opened 2025-01-23
build: deactivate unstable optimize drag report test

abremard:
## Description

The "drag a report in a Dashboard" e2e test has been unstable, so we're temporarily removing it.
It will be reactivated again, once there is the fix for it.


grlimacan (approved):
I am actually not able to review this, as I have no clue how the E2E tests work, but since no FE Engineer can take it and this looks plausible to me, I'm approving it

PR #27321, opened 2025-01-23
Rbms bugfixes for flow nodes and process instances

### LINKED ISSUE #27296: Bug: Fix flow node and process instances in RDBMS module
After testing with ProcessInstanceAndFlowNodeInstanceQueryTest against RDBMS module, we found two problems there:

1) In the setup of the test, rdbms module produces 26 flow node entries while camunda exporter just creates 20 entries:

**RDBMS:** 

Event_1p0nsc7, StartEvent_1, StartEvent_nyzldfv, call_activity, child_process_v1, endEvent, end_event, exclusiveGateway, incident_process_v1, manual_process, noOpTask, parent_process_v1, service_tasks_v1, service_tasks_v2, start, start, startEvent, start_event, taskA, taskAIncident, taskC, taskM, taskN, taskP, taskX, taskY  

**Camunda Exporter:**

Call Activity, End, End Event, Event_1p0nsc7, No Op, Start, Start Event, StartEvent_1, StartEvent_nyzldfv, Task A Incident, Task M, Task N, Task P, TaskX, TaskY, exclusiveGateway, start, start, task A, task C

2) For camunda exporter, the flowNodeIds are mapped (When reading in FlowNodeInstanceController with help of processCache) and for rdbms search client not. 

3) Process instance search with hasIncident filter doesn't work

### AC:

- [ ] RDBMS module produces same amount of flow node instances (we export too much)
- [ ] RDBMS results are also mapped in a way that the flowNodeName ist not just the flowNodeId (maybe we do this directly when writing it)
- [ ] RDBMS process instance search should support has incident filter

p-wunderlich:
## Description

See issue

## Related issues

closes #27296

### DIFF
--- a/db/rdbms/src/main/resources/mapper/ProcessInstanceMapper.xml
+++ b/db/rdbms/src/main/resources/mapper/ProcessInstanceMapper.xml
@@ -109,6 +109,12 @@
         AND TENANT_ID <include refid="io.camunda.db.rdbms.sql.Commons.operationCondition"/>
       </foreach>
     </if>
+    <if test="filter.hasIncident != null and filter.hasIncident == true">
+      AND NUM_INCIDENTS > 0
+    </if>
+    <if test="filter.hasIncident != null and filter.hasIncident == false">
+      AND NUM_INCIDENTS = 0
+    </if>
 
     <!-- date filters -->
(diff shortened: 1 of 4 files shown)


stoerti (approved):
LGTM

PR #27339, opened 2025-01-23
fix: fix "Camunda Performance" dashboard performance

multani:
## Description

The dashboard can be very expensive to run:

1. It defaults to all the Kubernetes namespaces, including non-Camunda ones. This pulls in a lot of metrics from Prometheus by default, even metrics completely unrelated to Camunda services. Some visualization are extremely expensive to display on large Kubernetes clusters if no namespace has been specifically selected.
2. It refreshed by default every 30 seconds. Combined with 1., this makes the dashboard even more expensive to display or interact with.
3. It removes extremely low refresh intervals. They are lower than most default Prometheus scrape interval and are not useful.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/monitor/grafana/camunda-performance.json
+++ b/monitor/grafana/camunda-performance.json
@@ -3291,7 +3291,7 @@
       "type": "row"
     }
   ],
-  "refresh": "30s",
+  "refresh": "",
   "schemaVersion": 38,
   "style": "dark",
   "tags": [],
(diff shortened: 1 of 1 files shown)


ChrisKujawa (approved):
Thanks Jon 👍

While i not agreed regarding not having a default set, but im fine for now.

Issue #26068, opened 2024-12-13
flaky SIGSEV when running AuthorizationCheckBehaviorTest 


entangled90:
**Summary**

> Try to answer the following as best as possible

- How often does the test fail? Randomly
- Does it block your work? no
- Do we suspect that it is a real failure? Yes

**Failures**

> Outline known failure cases, e.g. a failed assertion and its stacktrace obtained from Jenkins

The VM crashes because of SIGSEV, attached hs_err.log

**Logs**
Failing pipeline: https://github.com/camunda/camunda/actions/runs/12316493858/job/34376922384

```

Current thread (0x00007a5ac59866f0):  JavaThread "zb-actors-4"        [_thread_in_native, id=102646, stack(0x00007a5a3a1ff000,0x00007a5a3a2ff000) (1024K)]

Stack: [0x00007a5a3a1ff000,0x00007a5a3a2ff000],  sp=0x00007a5a3a2fd370,  free space=1016k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [librocksdbjni1421240772440967531.so+0x9115cf]
C  [librocksdbjni1421240772440967531.so+0x912a40]  std::_Hashtable<std::string, std::pair<std::string const, rocksdb::TrackedKeyInfo>, std::allocator<std::pair<std::string const, rocksdb::TrackedKeyInfo> >, std::__detail::_Select1st, std::equal_to<std::string>, std::hash<std::string>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, false, true> >::clear()+0x20
C  [librocksdbjni1421240772440967531.so+0x912a95]  std::_Hashtable<std::string, std::pair<std::string const, rocksdb::TrackedKeyInfo>, std::allocator<std::pair<std::string const, rocksdb::TrackedKeyInfo> >, std::__detail::_Select1st, std::equal_to<std::string>, std::hash<std::string>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, false, true> >::~_Hashtable()+0x15
[...]

PR #27341, opened 2025-01-23
test: In AuthorizationCheckBehaviourTest run actions on the same thread as the engine

entangled90:
## Description
Before this change, authorizationCheckBehaviour interact with mutables classes that are used concurrently by the engine in another thread.

All methods called in authorizationCheckBehaviour are run in the same thread as the StreamProcessor.
Although this may not be the case of the SIGSEV, it can help us identify if it's generated from somewhere else if it's not fixed by this commit.

## Related issues
relates #26068

### DIFF
--- a/zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/StreamProcessorRule.java
+++ b/zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/StreamProcessorRule.java
@@ -21,6 +21,7 @@ import io.camunda.zeebe.logstreams.util.TestLogStream;
 import io.camunda.zeebe.protocol.impl.encoding.AuthInfo;
 import io.camunda.zeebe.protocol.impl.record.UnifiedRecordValue;
 import io.camunda.zeebe.protocol.record.intent.Intent;
+import io.camunda.zeebe.scheduler.ActorScheduler;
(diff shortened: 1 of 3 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/StreamProcessorRule.java
```diff
@@ -100,6 +101,7 @@ public StreamProcessorRule(
       final TemporaryFolder temporaryFolder) {
     this.startPartitionId = startPartitionId;
     this.partitionCount = partitionCount;
+    actorSchedulerRule = new ActorSchedulerRule(partitionCount, clock);
```
lenaschoenburg, line 104: :+1: Good idea to use the partition count here.

lenaschoenburg (changes requested):
:x: The idea is cool but doesn't work the way we want it to: `runTaskAsActor` submits a new actor task to the actor scheduler which has multiple threads. Nothing guarantees that this actor task doesn't run at the same time as the StreamProcessor/Engine.
If we want to follow this direction (instead of preventing the unsafe access of processing state), then we'd have to submit new jobs to the StreamProcessor actor task, that way they cannot run concurrently.

lenaschoenburg (approved):
Alright, let's give this a try :+1:

PR #27342, opened 2025-01-23
Migrate all tests in client package to new IT/E2E foundation

ChrisKujawa:
## Description

Migrates all tests that are in `package io.camunda.it.client;` to the new IT foundation. that means to the usage of the `@MultiDbTest` annotation.

### Details

Tests are not separate, and executed as part of `[IT] Elasticsearch`. Later, we will add support for Opensearch as well, which will be added as separate CI JOB.

#### Performance

Due to setting up database once, and application we were able to reducing the execution by a factor 10 for each individual test. **Most test were running before around 100s (due to set up code) and now running within 10s.**

Examples from Main branch:

main-client2
main-client

Examples from branch:

migrate-client

migrate-client2

## Checklist

- [ ] for CI changes:

## Related issues

related to https://github.com/camunda/camunda/issues/26896

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,8 +286,8 @@ jobs:
       #
       - name: Run integration test with externalized ES
         run: >
-          ./mvnw -B -T 2 --no-snapshot-updates
-          -D forkCount=7
+          ./mvnw -B -T 1 --no-snapshot-updates
+          -D forkCount=1
           -D maven.javadoc.skip=true
           -D skipUTs -D skipChecks
           -D failsafe.rerunFailingTestsCount=3 -D flaky.test.reportDir=failsafe-reports
--- a/qa/integration-tests/src/test/java/io/camunda/it/client/CreateDocumentBatchTest.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/client/CreateDocumentBatchTest.java
@@ -12,32 +12,21 @@ import static org.assertj.core.api.Assertions.within;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import io.camunda.client.CamundaClient;
-import io.camunda.qa.util.cluster.TestStandaloneCamunda;
-import io.camunda.zeebe.qa.util.junit.ZeebeIntegration;
-import io.camunda.zeebe.qa.util.junit.ZeebeIntegration.TestZeebe;
+import io.camunda.it.utils.MultiDbTest;
 import java.io.ByteArrayInputStream;
 import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.time.temporal.ChronoUnit;
 import org.junit.jupiter.api.Test;
 
-@ZeebeIntegration
+@MultiDbTest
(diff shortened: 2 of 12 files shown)


ChrisKujawa:
Tests are currently failing because of a race condition see https://camunda.slack.com/archives/C07PNCGJHAA/p1737637610301689

.github/workflows/ci.yml
```diff
@@ -286,8 +286,8 @@ jobs:
       #
       - name: Run integration test with externalized ES
         run: >
-          ./mvnw -B -T 2 --no-snapshot-updates
-          -D forkCount=7
+          ./mvnw -B -T 1 --no-snapshot-updates
+          -D forkCount=1
           -D maven.javadoc.skip=true
```
ChrisKujawa, line 291: :i: I realized that we were also stressing the ES node, I saw some log saying Yellow. I just wanted to try with lower fork count if this helps to reduce the load. It did.

Instead before having test execution around 30s (and before in old setup 60s) - it was further reduced to 10 seconds ! When I run the tests with fork count 7 it needs 6-7 minutes, now it takes [~1:40](https://github.com/camunda/camunda/actions/runs/12943304681/job/36102432356?pr=27342#step:8:375). I'm a bit shocked

ChrisKujawa:
I have no idea why the BasicAuthIt fails, see discussion [here](https://camunda.slack.com/archives/C06UYJMMETZ/p1737695740364649)

ChrisKujawa:
> **Most test were running before around 100s (due to set up code) and now running within 10s.**

While this is I think great, there is a caveat that we now run a separate job, which might increase cost, but we were able to reduce the execution time. Maybe @cmur2 has some data to judge whether this balances out. 

Would be in general interesting how many jobs we have per day, and what this means for the execution time reduction then.

.github/workflows/ci.yml
tmetzke: OK, I wouldn't have expected that 😄

tmetzke (approved):
👍 Changes look good to me, the tests look a lot better now 🚀 
💡 I didn't expect the result of the config change, really 🤣

.github/workflows/ci.yml
entangled90: It also probably means that it will be much "lighter" on resources when running locally. It happened to me to hit OOMs.
Great! :rocket:

entangled90 (approved):
:rocket:

.github/workflows/ci.yml
npepinpe: So things were slow because we were overloading ES? We discussed that before, right? :smile:

tmetzke:
How much flakiness can one PR encounter upon being merged? 🤣

ChrisKujawa:
> How much flakiness can one PR encounter upon being merged? 🤣

It is so unbelievably sad. 


I created some issues for the Backup tests:

- https://github.com/camunda/camunda/issues/27744
- https://github.com/camunda/camunda/issues/27743

PR #27352, opened 2025-01-23
build: remove non-optimize CIs

buccarel:
## Description

As this branch is dedicated to Optimize, we don't need to run the CIs of other teams. This is because the version of Optimize of this branch is not yet integrated in the single application and other CIs would not be useful.

## Checklist

- [ ] for CI changes:

## Related issues

closes #


RomanJRW:
The approach seems reasonable to me and the CI on this PR seems to verify what we want, but let's have @cmur2 have a look before merging.

Also, shall we do this for the `stable/optimize-8.6` branch too? It would save on costs to not run unnecessary things there

cmur2:
> The approach seems reasonable to me and the CI on this PR seems to verify what we want, but let's have @cmur2 have a look before merging.

Looking at the list of removed jobs I think it should be fine if the branch is totally dedicated to Optimize. Since this change is reversible (in case we later identify some CI jobs would've been good to keep) I'm 👍🏼 

> Also, shall we do this for the `stable/optimize-8.6` branch too? It would save on costs to not run unnecessary things there

The same reasoning applies to `stable/optimize-8.6` indeed from my perspective.

grlimacan (approved):
LGTM!

optimize/test-file
grlimacan, line 1: Don't forget to remove this before merging

PR #27368, opened 2025-01-23
refactor: introduce a TenantMemberEntity to ES/OS exporters

Ben-Sheppard:
## Description

When sharing the progress of tenant members in the exporters it was suggested that the member docs are not optimal and also are missing a field. This PR achieves two things:

1. Adds in the `memberType` field such that you are able to filter for a specific type of member of a tenant
2. Adds in a `TenantMemberEntity` object that is exported instead of the overall `TenantEntity` object therefore reducing the fields in the child relation docs.

As a before and after, here is the ES indexes:

Before:
```
{
                "_index": "camunda-tenant-8.7.0_",
                "_id": "test-id-demo",
                "_score": 1.0,
                "_routing": "test-id",
                "_source": {
                    "id": "test-id-demo",
                    "key": null,
                    "tenantId": null,
                    "name": null,
                    "description": null,
                    "memberKey": null,
                    "join": {
                        "name": "member",
                        "parent": "test-id"
                    }
                }
            }
```

After: 
```
            {
                "_index": "camunda-tenant-8.8.0_",
                "_id": "test-id-demo",
                "_score": 1.0,
                "_routing": "test-id",
                "_source": {
                    "id": "test-id-demo",
                    "memberId": "demo",
                    "memberType": "USER",
                    "join": {
                        "name": "member",
                        "parent": "test-id"
                    }
                }
            }
```

## Related issues

_no related issue_

### DIFF
--- a/webapps-schema/src/main/resources/schema/elasticsearch/create/index/camunda-tenant.json
+++ b/webapps-schema/src/main/resources/schema/elasticsearch/create/index/camunda-tenant.json
@@ -17,8 +17,11 @@
       "description": {
         "type": "text"
       },
-      "memberKey": {
-        "type": "long"
+      "memberId": {
+        "type": "keyword"
+      },
+      "memberType": {
+        "type": "keyword"
       },
       "join": {
         "type": "join",
--- a/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/TenantEntity.java
+++ b/webapps-schema/src/main/java/io/camunda/webapps/schema/entities/usermanagement/TenantEntity.java
@@ -15,8 +15,6 @@ public class TenantEntity extends AbstractExporterEntity<TenantEntity> {
   private String tenantId;
   private String name;
   private String description;
-  private Long memberKey;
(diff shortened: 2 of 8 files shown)


webapps-schema/src/main/resources/schema/elasticsearch/create/index/camunda-tenant.json
```diff
@@ -17,8 +17,11 @@
       "description": {
         "type": "text"
       },
-      "memberKey": {
-        "type": "long"
+      "memberId": {
+        "type": "keyword"
+      },
+      "memberType": {
+        "type": "keyword"
```
emulbreh, line 24: We probably need the same change for the opensearch mapping.

emulbreh (approved):

webapps-schema/src/main/resources/schema/elasticsearch/create/index/camunda-tenant.json
Ben-Sheppard: Good catch, added :)

PR #27376, opened 2025-01-23
feat: renamed camunda.database.index-prefix property to camunda.datab…

stoerti:
…ase.prefix

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/search/search-client-connect/src/main/java/io/camunda/search/connect/configuration/ConnectConfiguration.java
+++ b/search/search-client-connect/src/main/java/io/camunda/search/connect/configuration/ConnectConfiguration.java
@@ -28,7 +28,12 @@ public class ConnectConfiguration {
   private String username;
   private String password;
   private SecurityConfiguration security = new SecurityConfiguration();
+  private String prefix;
+
+  /** replaced by <code>prefix</code> */
+  @Deprecated(forRemoval = true)
   private String indexPrefix;
+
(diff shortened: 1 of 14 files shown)


search/search-client-connect/src/main/java/io/camunda/search/connect/configuration/ConnectConfiguration.java
```diff
@@ -117,10 +122,22 @@ public void setSecurity(final SecurityConfiguration security) {
...
+  }
+
+  public void setPrefix(final String indexPrefix) {
+    prefix = indexPrefix;
+  }
+
+  /** replaced by <code>prefix</code> */
+  @Deprecated(forRemoval = true)
   public String getIndexPrefix() {
     return indexPrefix;
```
p-wunderlich, line 136: This and setIndexPrefix could already work on the "new" prefix property.

p-wunderlich: Do we even need to keep the old setters / getters? They are not longer used and I think they are also not used by the customer?

romansmirnov (commented):
@stoerti, the changes look good. Is there a way to have one test that ensures the deprecated `index-prefix` still works? (If there is already a test, then just let me know)

PR #27391, opened 2025-01-23
fix: adapt c8run to auth config changes

### LINKED ISSUE #28187: I can run C8Run with Basic Auth enabled for WebApps and an unprotected API
Acceptance Criteria:
- [ ] C8Run is running with basic auth enabled for web application
- [x]  C8Run is running with an unauthenticated API by default

emulbreh:
## Description

This PR adapts c8run to work with the basic auth config introduced in https://github.com/camunda/camunda/pull/26960.

It can't be included in the PR directly because `c8run package` pulls the latest release from GitHub instead of using the snapshot from the branch. 

See [this slack thread](https://camunda.slack.com/archives/C074T8BTTM2/p1737644000219619) for context.

closes #28187

### DIFF
--- a/c8run/main.go
+++ b/c8run/main.go
@@ -134,7 +134,9 @@ func adjustJavaOpts(javaOpts string, settings C8RunSettings) string {
 	if settings.password != "" {
 		javaOpts = javaOpts + " -Dcamunda.security.initialization.users[0].password=" + settings.password
 	}
-	javaOpts = javaOpts + " -Dspring.profiles.active=operate,tasklist,broker,identity,auth-basic"
+	javaOpts += " -Dspring.profiles.active=operate,tasklist,broker,identity,consolidated-auth"
+	javaOpts += " -Dcamunda.security.authentication.method=basic"
+	javaOpts += " -Dcamunda.security.authentication.basic.allow-unauthenticated-api-access=true"
 	os.Setenv("CAMUNDA_OPERATE_ZEEBE_RESTADDRESS", protocol+"://localhost:"+strconv.Itoa(settings.port))
 	fmt.Println("Java opts: " + javaOpts)
 	return javaOpts
--- a/c8run/main_test.go
+++ b/c8run/main_test.go
@@ -22,7 +22,10 @@ func TestCamundaCmdWithKeystoreSettings(t *testing.T) {
 		keystore:         "/tmp/camundatest/certs/secret.jks",
 		keystorePassword: "changeme",
 	}
-	expectedJavaOpts := "JAVA_OPTS= -Dserver.ssl.keystore=file:" + settings.keystore + " -Dserver.ssl.enabled=true" + " -Dserver.ssl.key-password=" + settings.keystorePassword + " -Dspring.profiles.active=operate,tasklist,broker,identity,auth-basic"
+	expectedJavaOpts := "JAVA_OPTS= -Dserver.ssl.keystore=file:" + settings.keystore + " -Dserver.ssl.enabled=true" + " -Dserver.ssl.key-password=" + settings.keystorePassword
+	expectedJavaOpts += " -Dspring.profiles.active=operate,tasklist,broker,identity,consolidated-auth"
+	expectedJavaOpts += " -Dcamunda.security.authentication.method=basic"
+	expectedJavaOpts += " -Dcamunda.security.authentication.basic.allow-unauthenticated-api-access=true"
 
 	javaOpts := adjustJavaOpts("", settings)
 	c8runPlatform := getC8RunPlatform()


megglos:
Hey @jessesimpson36 I'd like to follow-up on
> It can't be included in the PR directly because c8run package pulls the latest release from GitHub instead of using the snapshot from the branch.

I think it would be worth allowing to use local archives of the camunda 8 application (e.g. on local machine, or also in a github ci workflow) when building c8run. This way we could actually have c8run ci that integrates with the current state of the c8 applications of the same revision.
It would allow easy local testing based on a local build c8 application + in https://github.com/camunda/camunda/blob/main/.github/workflows/c8run-build.yaml we could build the current state of the camunda 8 single app and integrate.

wdyt?

jessesimpson36:
@megglos 

> It can't be included in the PR directly because c8run package pulls the latest release from GitHub instead of using the snapshot from the branch.

we now have a release 8.8.0-alpha1, which we can now pull from the package function in c8run
https://github.com/camunda/camunda/releases/tag/8.8.0-alpha1

As for the option to build camunda from source within the CI, the github action that I call from the c8run integration tests has an option to support building from source and feeding that into c8run.
https://github.com/camunda/camunda/blob/5978dc0117d97fda31beb4e8c29063529ea8dd01/.github/actions/setup-c8run/action.yml#L19-L22

This does give you a point-in-time record of how c8run behaves with that version of the codebase. However, I do not recommend enabling it inside the camunda repository, and would instead suggest that it be used as part of the nightly QA e2e test runs. the branch name where I try to add that in is called `gha-c8run-refactor`.

https://github.com/camunda/c8-cross-component-e2e-tests/compare/main...gha-c8run-refactor

hisImminence:
> > snapshot from the branch.
> 
> I think it would be worth allowing to use local archives of the camunda 8 application (e.g. build on local machine, or also in a github ci workflow) when building c8run. This way we could actually have c8run ci that integrates with the current state of the c8 applications of the same revision. It would allow easy local testing based on a local build c8 application + in https://github.com/camunda/camunda/blob/main/.github/workflows/c8run-build.yaml we could build the current state of the camunda 8 single app and integrate.

@megglos, had a chat about it with @jessesimpson36 about it!

--> We have the following suggestion:

1. @jessesimpson36 actually already implemented the possibility to build C8run from source. Its the [source-build](https://github.com/camunda/camunda/blob/main/.github/actions/setup-c8run/action.yml#L19) option that you can set in the c8run-build action. Its set to false by default as when enabled it was failing back then most of the time due to camunda build errors and debugging the cause was hard.
--> We suggest to create a new workflow that enables the `source-build` and see if its now more stable. This workflow should be owned by a core engineer as debugging the build should be way more efficient if done by a them than a distro team member (if cause we support if its c8run related). 

2. @jessesimpson36 also has a flow to build c8run from source locally already, which is sufficient as a first iteration imo (we can improve the flow once we confirm its generally the approach we want to go):

```
How to build and run c8run from a branch (e.g. main/ stable/8.7, ...):

1. From root dir ./camunda run: /mvnw clean dependency:resolve -DskipTests -DskipOptimize -DskipChecks -Dskip.fe.build=false install
   --> This will build the local SNAPSHOT in: ./dist/target/camunda-zeebe-8.8.0-SNAPSHOT.tar.gz
4. Set "8.8.0-SNAPSHOT" as the value in the [.env file](https://github.com/camunda/camunda/blob/main/c8run/.env#L3) for CAMUNDA_VERSION.
5. run: go build 
6. run: bash package.go
7. start your c8run build
```

--> If you agree @megglos I would document the steps in 2. in the c8run README and if the core engineers are fine with owning the c8run workflow build from source - we can add this workflow as well. Let me know if thats they way you want to go!

megglos:
@jessesimpson36 
> However, I do not recommend enabling it inside the camunda repository

For me to better understand, why would you not recommend that?

> --> If you agree @megglos I would document the steps in 2. in the c8run README and if the core engineers are fine with owning the c8run workflow build from source - we can add this workflow as well. Let me know if thats they way you want to go!

I actually think we should integrate with the current source state of camunda 8 within the c8run CI by default.
As otherwise we are not really integrating and are blocked adopting to breaking changes, like we have been here, which are likely to happen with 8.8 within the next months.

We will try to incorporate this change into this PR to evaluate the impact on build time of the C8Run CI etc. as we should be able to keep the time needed for the c8 maven build rather low (single digit minutes, locally it takes 2 minutes running `mvn clean install -T1C -DskipTests -DskipChecks -Dquickly`). Let's align on that then?

jessesimpson36:
> For me to better understand, why would you not recommend that? I would argue we should as only this way we perform real CI with the current state of the application and c8run to be released next.

It takes extra time to build the application from source. About 20 minutes.  and majority of the time, the reasons the build fails has nothing to do with c8run issues. When zeebe is built and tested, its always tested against a version of zeebe built with skipFrontendBuild=true,  which means when you set that to false (which c8run requires), we'll get build failures until someone actually tries that and fixes it.  Some of the timing could be addressed if the build of camunda set skipFrontendBuild=false and if that build was passed into c8run rather than us building camunda core redundantly.

Most PRs in the camunda repo don't touch c8run logic, and most c8run changes don't change application logic. so constantly testing that they're compatible with eachother is wasteful... not just of infrastructure resources, but also time developers spend waiting for builds to pass, and also time I spend debugging application-level logic when it's irrelevant to my PRs.

> I think we should integrate with the current source state of camunda 8 within the c8run CI by default and it's also in the interest of the distro team to spot incompatibilities asap even before the next release, isn't it?

Ideally yes, but the camunda repo is not at this level of stability yet.  If we're going with the fix-everything-until-it's-all-green approach, then I think we should address all the current CI failures against main before enabling extra CI.

2025-02-19-095815_grim

> Otherwise what's the value to verify that the current c8run state works with a previous release of the application stack but not the current one of the same git revision?

the value is that no new changes in c8run breaks c8run. When camunda core is ready to be built and released on the new version, then the version is updated and I fix any incompatibilities, which there hasn't really hasn't been many changes like this.

hisImminence:
How about this?

Lets have a dedicated workflow running every night that builds C8Run from source. If the workflow is green for the next weeks/ month we will add this action to the C8Run CI (ran per C8run PR).

megglos:
I'm also fine with the nightly compromise, however that will still not allow us to merge potentially breaking config changes to c8run to adapt to recent camunda 8 changes, given on a PR we integrate still with the previous release.

What about this, let's spike the integration of a build with the lowest overhead we can achieve into the PR workflow of c8run and check how long it actually takes/how stable it is. If the outcome is not promising let's go for nightly a nightly one?

Let's also do this separate from this change, as we don't need to couple this.

hisImminence:
> I'm also fine with the nightly compromise, however that will still not allow us to merge potentially breaking config changes to c8run to adapt to recent camunda 8 changes, given on a PR we integrate still with the previous release.
> 
> What about this, let's spike the integration of an ad-hoc build with the lowest overhead we can achieve into the PR workflow of c8run and check how long it actually takes/how stable it is. If the outcome is not promising let's go for nightly a nightly one?
> 
> Let's also do this separate from this change, as we don't need to couple this.

Okay - as just discussed, lets enable it on any PR then - @jessesimpson36, ready to rumble!

hisImminence:
PR is ready for review: https://github.com/camunda/camunda/pull/28444

PR #27454, opened 2025-01-24
feat: don't expose tenant key anywhere

### LINKED ISSUE #27059: Remove (tenant) key from API and use/return the tenantId instead
Depends on https://github.com/camunda/camunda/issues/27219 and https://github.com/camunda/camunda/issues/27220

lenaschoenburg:
Remove tenant key from the API and Exporters, tenant id is used as primary key everywhere except the runtime state.
Endpoints to assign or unassign groups and mappings from tenants are removed for now, we will add them later on again when these entities become id-based too.

depends on #27382
closes #27059


remcowesterhoud:
@lenaschoenburg Do we not want to keep exposing the key? We left it in for users as well. It could be useful to have in the future. For example, maybe distribution will be more shared more fairly amongst partitions. We can use the key to determine which partition is the "owner" of this tenant.

PR #27457, opened 2025-01-24
feat: add id to Mapping

### LINKED ISSUE #27375: Create mapping by mapping id instead of key
## Description
In this issue we are going to add a unique column called id to Mapping and change the CreateMappingProcessor accordingly to handle this change. 

## Out of scope

- Other processors related to Mapping
- Exporter and retrieval

maryarm:
## Description

Handle external mapping id during mapping creation.

## Related issues

closes #27375


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbMappingState.java
```diff
@@ -34,38 +34,52 @@ public class DbMappingState implements MutableMappingState {
...
     mappingKey = new DbLong();
     fkClaim = new DbForeignKey<>(claim, ZbColumnFamilies.MAPPINGS);
     claimByKeyColumnFamily =
         zeebeDb.createColumnFamily(
             ZbColumnFamilies.CLAIM_BY_KEY, transactionContext, mappingKey, fkClaim);
+
+    mappingId = new DbString();
+    claimByIdColumnFamily =
+        zeebeDb.createColumnFamily(
+            ZbColumnFamilies.CLAIM_BY_KEY, transactionContext, mappingId, fkClaim);
```
remcowesterhoud, line 60: ❌ We cannot reuse the same ColumnFamily. The name also doesn't suit with the contents of the CF. I'd expect the name to be `CLAIM_BY_ID`.

I do not see a reason to keep the `CLAIM_BY_KEY` CF. Why did you decide to keep this one?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbMappingState.java
```diff
@@ -34,38 +34,52 @@ public class DbMappingState implements MutableMappingState {
...
   }
 
   @Override
   public void create(final MappingRecord mappingRecord) {
```
remcowesterhoud, line 64: ❓ I assume you will deal with the other methods (delete/get/update) in different issues?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/MappingCreateProcessor.java
```diff
@@ -26,8 +26,11 @@
 
 public class MappingCreateProcessor implements DistributedTypedRecordProcessor<MappingRecord> {
 
-  private static final String MAPPING_ALREADY_EXISTS_ERROR_MESSAGE =
+  private static final String MAPPING_SAME_CLAIM_ALREADY_EXISTS_ERROR_MESSAGE =
       "Expected to create mapping with claimName '%s' and claimValue '%s', but a mapping with this claim already exists.";
+  private static final String MAPPING_SAME_ID_ALREADY_EXISTS_ERROR_MESSAGE =
+      "Expected to create mapping with id '%s' but a mapping with this id already exists.";
```
remcowesterhoud, line 32:
```suggestion
      "Expected to create mapping with id '%s', but a mapping with this id already exists.";
```

🙃 Let's add a comma here to keep it aligned with the other error message

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/identity/MappingCreateProcessor.java
```diff
@@ -95,7 +108,7 @@ final var record = command.getValue();
         .ifPresentOrElse(
             existingMapping -> {
               final var errorMessage =
-                  MAPPING_ALREADY_EXISTS_ERROR_MESSAGE.formatted(
+                  MAPPING_SAME_CLAIM_ALREADY_EXISTS_ERROR_MESSAGE.formatted(
```
remcowesterhoud, line 111: 🔧 Let's change this so we get the mapping by id instead of by key, and reject the command with the ID already exists error message.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/PersistedMapping.java
```diff
@@ -40,7 +41,8 @@ public PersistedMapping() {
         .declareProperty(nameProp)
         .declareProperty(roleKeysProp)
         .declareProperty(tenantIdsProp)
-        .declareProperty(groupKeysProp);
+        .declareProperty(groupKeysProp)
+        .declareProperty(idProp);
```
remcowesterhoud, line 45: ❌ You've added a property so you must increment the number in the call to the super class to 8.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/AuthorizationCheckBehaviorTest.java
```diff
@@ -801,6 +801,7 @@ public void shouldBeAuthorizedThroughMultipleMappings() {
             .mapping()
             .newMapping(firstClaimName)
             .withClaimValue(firstClaimValue)
+            .withId(UUID.randomUUID().toString())
```
remcowesterhoud, line 804: ❓ Do you have an issue to actually use this mapping id when retrieving authorizations? Currently it seems we're still doing it using the mapping key and I fear we are running out of sync with the parallel streams of work that are happening.

remcowesterhoud: For example I'd expect that when creating authorization on line 826 we'd provide the id now, instead of the key as a string. This comment applies to other methods as well.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/IdentitySetupInitializeTest.java
```diff
@@ -309,10 +309,12 @@ public void shouldCreateConfiguredMappings() {
...
     final var mapping1 =
         new MappingRecord()
+            .setId(UUID.randomUUID().toString())
             .setClaimName(UUID.randomUUID().toString())
             .setClaimValue(UUID.randomUUID().toString());
     final var mapping2 =
         new MappingRecord()
+            .setId(UUID.randomUUID().toString())
             .setClaimName(UUID.randomUUID().toString())
             .setClaimValue(UUID.randomUUID().toString());
```
remcowesterhoud, line 319: ❓ Do you have an issue to allow users to configure an id when configuring the initial mappings? Currently we only allow them to configure a claim name and value. Since id is user provided as well, we must ensure users can configure these as well.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/MappingClient.java
```diff
@@ -61,6 +61,11 @@ public MappingCreateClient withClaimValue(final String claimValue) {
       return this;
     }
 
+    public MappingCreateClient withId(final String id) {
+      mappingRecord.setId(id);
+      return this;
+    }
+
```
remcowesterhoud, line 68: 🔧 As id is the identifier I'd prefer having a method `withClaimName(String)` and putting id in the `newMapping(String)` method. So it'd become:

```java
engine.mapping()
  .newMapping(id)
  .withClaimName(claimName)
  .withClaimValue(claimValue)
  .withName(name)
  .create()
```

WDYT?

zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/MappingRecord.java
```diff
@@ -19,13 +19,15 @@ public class MappingRecord extends UnifiedRecordValue implements MappingRecordVa
...
+  private final StringProperty idProp = new StringProperty("id", "");
 
   public MappingRecord() {
     super(4);
     declareProperty(mappingKeyProp)
         .declareProperty(claimNameProp)
         .declareProperty(claimValueProp)
-        .declareProperty(nameProp);
+        .declareProperty(nameProp)
+        .declareProperty(idProp);
```
remcowesterhoud, line 30: ❌ You've added a property so you must increment the 4 in the super() call to 5.

remcowesterhoud (changes requested):
Thank @maryarm, please take a look at my comments and question 🙂 


❌ I'm missing some test cases:

- State test to verify creating a mapping rule stores the id as expected
- State test to verify getting a mapping rule by id.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbMappingState.java
maryarm: It has two reason:
I was looking on the TenentState as a reference and on that time it has kept the key. (after later discussions with @lenaschoenburg and @megglos we were still not sure to keep or remove it, but at the end we decided to remove it for now although we may lost the followup event feature here.  
The second reason was that I can remove the key after I complete the changes in all processors like delete tenant.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/MappingClient.java
maryarm: The reason that I didn't go for this option was that already there are a lot usage (34 times) for newMapping(name) method and I didn't want to touch all these files in this PR.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbMappingState.java
remcowesterhoud: So what you're saying is that this is temporary and will be removed later?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/MappingClient.java
remcowesterhoud: I think it's a valid refactoring to do. If you keep it isolated in a single commit it doesn't make the PR harder to review to me 🤷

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/authorization/IdentitySetupInitializeTest.java
remcowesterhoud: https://github.com/camunda/camunda/issues/27820

remcowesterhoud (approved):
Sorry for taking long to review this one. I somehow missed it in my notificiations.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/MappingClient.java
maryarm: I'll do it in a separate PR then.

PR #27460, opened 2025-01-24
feat: use one ObjectMapper for all classes in the camunda-exporter package

### LINKED ISSUE #24599: Streamline ObjectMapper usage and configuration
## Description

We use and instantiate ObjectMapper in several places.

With the next Camunda Exporter we do this even more often, and often with default configurations, while in the "old" Importers for Tasklist and Operate we had specific configurations. 

In general, it causes confusion about what to use and can also cause data conflicts when writing and reading with different formats, etc. 

We should align our ObjectMapper user and also pair up with the REST API team (@danielkelemen as he is currently also looking into this).

### Context

1. https://camunda.slack.com/archives/C06F0GLJNFM/p1730397911982409
2. https://camunda.slack.com/archives/C06F0GLJNFM/p1731318495433169

entangled90:
## Description
All classes instantiating their own ObjectMapper now use a common instance created using `io.camunda.search.connect.jackson.JacksonConfiguration` (from the common `search` module)
This instance is injected at runtime in all exporter classes and it's shared with ES/OS clients.

The `ObjectMapper` is available as a method in `ClientAdapter`. 

When possible, type specific readers/writers have been used in favour of an ObjectMapper as they are thread-safe and immutable.

As a result, there is `new ObjectMapper()` call in production code, and a `TestObjectMapper` is created which contains a static shared instance for when there is no ES/OS connection.

## Related issues

closes #24599

### DIFF
--- a/migration/process-migration/src/test/java/io/camunda/migration/process/it/AdapterTest.java
+++ b/migration/process-migration/src/test/java/io/camunda/migration/process/it/AdapterTest.java
@@ -20,6 +20,7 @@ import co.elastic.clients.elasticsearch.core.IndexRequest;
 import co.elastic.clients.elasticsearch.core.SearchRequest;
 import co.elastic.clients.elasticsearch.core.SearchResponse;
 import co.elastic.clients.elasticsearch.core.search.Hit;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.exporter.config.ExporterConfiguration.IndexSettings;
 import io.camunda.exporter.schema.elasticsearch.ElasticsearchEngineClient;
 import io.camunda.exporter.schema.opensearch.OpensearchEngineClient;
@@ -75,6 +76,9 @@ public abstract class AdapterTest {
   private static final OpensearchContainer<?> OS_CONTAINER =
       TestSearchContainers.createDefaultOpensearchContainer();
 
+  private static ObjectMapper esObjectMapper;
+  private static ObjectMapper osObjectMapper;
+
   protected boolean isElasticsearch = true;
 
   @BeforeAll
@@ -87,15 +91,19 @@ public abstract class AdapterTest {
     ES_CONFIGURATION.setUrl("http://localhost:" + ES_CONTAINER.getMappedPort(9200));
     OS_CONFIGURATION.setType("opensearch");
     OS_CONFIGURATION.setUrl("http://localhost:" + OS_CONTAINER.getMappedPort(9200));
-    esClient = new ElasticsearchConnector(ES_CONFIGURATION).createClient();
-    osClient = new OpensearchConnector(OS_CONFIGURATION).createClient();
+    final var esConnector = new ElasticsearchConnector(ES_CONFIGURATION);
+    esObjectMapper = esConnector.objectMapper();
+    esClient = esConnector.createClient();
+    final var osConnector = new OpensearchConnector(OS_CONFIGURATION);
+    osObjectMapper = osConnector.objectMapper();
+    osClient = osConnector.createClient();
     esMigrator = new MigrationRunner(properties, ES_CONFIGURATION, meterRegistry);
     osMigrator = new MigrationRunner(properties, OS_CONFIGURATION, meterRegistry);
     createIndices();
   }
 
   private static void createIndices() {
-    final OpensearchEngineClient osEngine = new OpensearchEngineClient(osClient);
+    final OpensearchEngineClient osEngine = new OpensearchEngineClient(osClient, osObjectMapper);
     processIndex = new ProcessIndex(ES_CONFIGURATION.getIndexPrefix(), false);
     migrationRepositoryIndex =
         new MigrationRepositoryIndex(ES_CONFIGURATION.getIndexPrefix(), false);
@@ -104,7 +112,8 @@ public abstract class AdapterTest {
     osEngine.createIndex(migrationRepositoryIndex, new IndexSettings());
     osEngine.createIndex(importPositionIndex, new IndexSettings());
 
-    final ElasticsearchEngineClient esEngine = new ElasticsearchEngineClient(esClient);
+    final ElasticsearchEngineClient esEngine =
+        new ElasticsearchEngineClient(esClient, esObjectMapper);
     processIndex = new ProcessIndex(ES_CONFIGURATION.getIndexPrefix(), true);
     migrationRepositoryIndex =
         new MigrationRepositoryIndex(ES_CONFIGURATION.getIndexPrefix(), true);
--- a/search/search-client-connect/src/main/java/io/camunda/search/connect/es/ElasticsearchConnector.java
+++ b/search/search-client-connect/src/main/java/io/camunda/search/connect/es/ElasticsearchConnector.java
@@ -86,6 +86,10 @@ public final class ElasticsearchConnector {
     return new ElasticsearchAsyncClient(transport);
   }
 
+  public ObjectMapper objectMapper() {
+    return objectMapper;
+  }
+
   private RestClient createRestClient(final ConnectConfiguration configuration) {
     final var httpHost = getHttpHost(configuration);
     final var restClientBuilder = RestClient.builder(httpHost);
--- a/zeebe/exporters/camunda-exporter/pom.xml
+++ b/zeebe/exporters/camunda-exporter/pom.xml
@@ -138,6 +138,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jsr310</artifactId>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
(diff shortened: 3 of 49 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -66,18 +66,19 @@ public class CamundaExporter implements Exporter {
...
   public CamundaExporter() {
-    this(new DefaultExporterResourceProvider());
+    // the metadata will be initialized on open
+    this(new DefaultExporterResourceProvider(), null);
   }
 
   @VisibleForTesting
   public CamundaExporter(final ExporterResourceProvider provider) {
-    this(provider, new ExporterMetadata());
+    this(provider, null);
```
ChrisKujawa, line 81: 🔧 Or could we also do it here: `new ExporterMetadata(clientAdapter.objectMapper());`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/ExporterMetadata.java
```diff
@@ -42,6 +44,12 @@ public final class ExporterMetadata {
         }
       };
 
+  public ExporterMetadata(final ObjectMapper objectMapper) {
+    // Specialized reader/writer for this class for efficiency
+    objectWriter = objectMapper.writerFor(ExporterMetadata.class);
+    objectReader = objectMapper.readerForUpdating(this);
```
ChrisKujawa, line 50: 👍🏼 TIL

migration/process-migration/src/test/java/io/camunda/migration/process/it/AdapterTest.java
```diff
@@ -75,6 +76,9 @@ public abstract class AdapterTest {
   private static final OpensearchContainer<?> OS_CONTAINER =
       TestSearchContainers.createDefaultOpensearchContainer();
 
+  private static ObjectMapper esObjectMapper;
+  private static ObjectMapper osObjectMapper;
```
ChrisKujawa, line 80: ❓ Do we need two different ones?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/SchemaResourceSerializer.java
```diff
@@ -18,11 +18,14 @@
...
 
 public final class SchemaResourceSerializer {
-  private static final ObjectMapper MAPPER = new ObjectMapper();
 
-  private SchemaResourceSerializer() {}
+  private final ObjectMapper objectMapper;
 
-  public static Map<String, Object> serialize(
+  public SchemaResourceSerializer(final ObjectMapper objectMapper) {
+    this.objectMapper = objectMapper;
```
ChrisKujawa, line 25: ❓ Should we also use here the reader?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ClientAdapter.java
```diff
@@ -7,22 +7,27 @@
...
 
   static ClientAdapter of(final ExporterConfiguration configuration) {
-    return switch (ConnectionTypes.from(configuration.getConnect().getType())) {
-      case ELASTICSEARCH -> new ElasticsearchAdapter(configuration);
-      case OPENSEARCH -> new OpensearchAdapter(configuration);
+    final var databaseType = configuration.getConnect().getTypeEnum();
+    return switch (databaseType) {
+      case DatabaseType.ELASTICSEARCH -> new ElasticsearchAdapter(configuration.getConnect());
+      case DatabaseType.OPENSEARCH -> new OpensearchAdapter(configuration.getConnect());
+      default -> throw new IllegalArgumentException("Unsupported databaseType: " + databaseType);
```
ChrisKujawa, line 25: ❓ Why was this change necessary? Shouldn't this fail compilation before?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/SearchEngineClientUtils.java
```diff
@@ -41,19 +43,20 @@ public static InputStream appendToFileSchemaSettings(
...
         .collect(Collectors.joining(","));
   }
 
-  public static <T> T mapToSettings(
+  public <T> T mapToSettings(
       final Map<String, String> settingsMap, final Function<InputStream, T> settingsDeserializer) {
     try (final var settingsStream =
-        IOUtils.toInputStream(MAPPER.writeValueAsString(settingsMap), StandardCharsets.UTF_8)) {
+        IOUtils.toInputStream(
+            objectMapper.writeValueAsString(settingsMap), StandardCharsets.UTF_8)) {
```
ChrisKujawa, line 59: ❓ Why we are not using here the writer/readers in this class?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/SearchClientAdapter.java
```diff
@@ -114,7 +114,7 @@ public JsonNode getPolicyAsNode(final String policyName) throws IOException {
       final var request =
           Requests.builder().method("GET").endpoint("_plugins/_ism/policies/" + policyName).build();
 
-      return MAPPER.readTree(osClient.generic().execute(request).getBody().get().body());
+      return objectMapper.readTree(osClient.generic().execute(request).getBody().get().body());
```
ChrisKujawa, line 117: ❓ Can we use the reader in this class as well?

search/search-client-connect/src/main/java/io/camunda/search/connect/es/ElasticsearchConnector.java
```diff
@@ -86,6 +86,10 @@ public ElasticsearchAsyncClient createAsyncClient() {
     return new ElasticsearchAsyncClient(transport);
   }
 
+  public ObjectMapper objectMapper() {
+    return objectMapper;
```
ChrisKujawa, line 90: 👍🏼 So we reuse now the ObjectMapper created in the search module, everywhere right?

ChrisKujawa (approved):
Thanks @entangled90 🚀 

I have only smaller comments/questions, but other than that looks good! Pre-approving, please have a look at the comments still.

Just to clarify: we use the object mapper created in the Search connectors now, right?

zeebe/exporters/camunda-exporter/pom.xml
```diff
@@ -138,6 +138,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jsr310</artifactId>
+      <scope>test</scope>
```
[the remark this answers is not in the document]
entangled90: :shrug: probably it's used now only from the camunda-search module

migration/process-migration/src/test/java/io/camunda/migration/process/it/AdapterTest.java
entangled90: Not great, but in these tests everything is "doubled" so I just kept two of them just in case, because they could have different configurations potentially.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/adapters/ClientAdapter.java
entangled90: So the difference is that now we are using the `DatabaseType` from the `search-connect` module, which contains `RDBMS` as a case, while previously  `ConnectionTypes` only contained `ES/OS`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/SearchEngineClientUtils.java
entangled90: I don't remember, but there a case when a method was not present in the reader or writer, but was present in the object mapper.  I need to double check

search/search-client-connect/src/main/java/io/camunda/search/connect/es/ElasticsearchConnector.java
entangled90: Yes, it's then pased everywhere from here

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/SchemaResourceSerializer.java
entangled90: For some reason there is no method in the reader `readValue(String, TypedReference)` :shrug:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/utils/SearchEngineClientUtils.java
entangled90: I checked, `ObjectReader` is missing the overload `readValue(String, TypedReference)`

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/utils/SearchClientAdapter.java
entangled90: I left the objectMapper because we need both the reader and the writer in this case

ChrisKujawa:
@entangled90 somehow a lot of things broke, looks like a compilation error :/

PR #27468, opened 2025-01-24
[stable/8.3] Raise incident at specific called process depth

korthout:
## Description

Manual port of https://github.com/camunda/camunda/pull/22735 to `stable/8.3`.

There were many conflicts to resolve, going back all the way to 8.3 first will make the other ports easier for me. It will also allow me to correctly document the versioning the `@since` javadocs.

## Related issues

relates to #16410


korthout:
@npepinpe Thanks for checking. It is still relevant. I was working through the remaining problems in the PR that it ports #22735. I'll wrap this up soon.

korthout:
This manual backport is pulling in changes to the activating applier from main that shouldn't be on there.

We decided not to merge this PR, as the effort to get it merged is expected to be too high compared to our expectations of the benefits (given that 8.3 will be EOL soon anyway). We don't expect that this is a problem typically occurring on production, rather users typically encounter this during development. However, if you happen to run into the related issue on 8.3, please let me know here so we can reconsider.

PR #27470, opened 2025-01-24
Ensure `withAuthentication()` is called before doing a service call.

remcowesterhoud:
## Description

This ArchUnit test verifies that every call to a Service that is performed by a Controller adds the authentication by
calling withAuthentication first. The test will fail if this is not the case.

This is essential, as not making this call could result in cases where forget to add it, or accidentally remove it (for
example during a merge conflict). As a result the endpoint would not verify a user's authorization anymore. Potentially
causing data-leaks.

It is important to note that this method assumes that the service field is only called once, and chained afterward. If there is multiple separate calls to the service field it will not be accurate and should be modified! Currently, we do not have such use-cases. I tried to make it all-encompassing but the code became too complex and it was not worth it imo.

## Checklist

- [ ] for CI changes:

## Related issues

N/A

### DIFF
--- a/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/ControllerAuthorizationArchTest.java
+++ b/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/ControllerAuthorizationArchTest.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.gateway;
+
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.fields;
+
+import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.core.domain.JavaField;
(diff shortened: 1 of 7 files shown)


zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/ControllerAuthorizationArchTest.java
```diff
@@ -0,0 +1,135 @@
...
+                  if (call1.getName().equals(WITH_AUTHENTICATION_METHOD_NAME)) {
+                    return -1;
+                  } else if (call2.getName().equals(WITH_AUTHENTICATION_METHOD_NAME)) {
+                    return 1;
+                  } else {
+                    return 0;
+                  }
+                })
+            .toList()
+            .getFirst();
```
houssain-barouni, line 131: ❓ maybe I misunderstood why we do sorting here, but can't we just look for any call with Name equals to `WITH_AUTHENTICATION_METHOD_NAME` ?

houssain-barouni (approved):
LGTM! 🚀 
I had one question ❓

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/ControllerAuthorizationArchTest.java
remcowesterhoud: @houssain-barouni You are right, this is senseless. I should order by line number first. I've added a new commit if you'd like to check it out 🙂 The detailed explanation is in the commit message.

houssain-barouni (approved):
LGTM! 👍

PR #27483, opened 2025-01-27
pg-27022-schema-validator-dynamic-field

### LINKED ISSUE #27022: CustomHeaders exposes values as indexable fields
### Describe the bug

The property `customHeaders` of the `tasklist-task` index, being set as `dynamic:true` is allowed to expose set values as indexable nested fields. This creates issues when doing the validation of the existing schema on ES/OS and the one present on the Tasklist's resources, i.e `tasklist-task.json`. 

Update: The same is true for the `tasklist-list-view` index, however the SchemaValidator does not seem to check it for this index. (?)

Either: 
- The Tasklist Schema Validator must be able to skip fields with `dynamic:true` so as to allow the usage of such fields and also allow them to be searchable
- or this field must be set to `dynamic:false` so as not to index these user provided fields

### To Reproduce

1. Deploy & start a process containing a UserTask with specified _customHeader_ key-value pair
2. Wait for the UserTask to be exported in Elasticsearch
3. Check the Elasticsearch index mappings and you'll find that under `tasklist-task.customHeaders` a new property has been exposed with the name of the provided key.

### Actual behavior
A difference is reported in the SchemaValidator, since the index mapping for `customHeaders` is now changed to :
```json
        "customHeaders" : {
          "dynamic" : "true",
          "properties" : {
            "custom-header-user-defined-key" : {
              "type" : "text",
              "fields" : {
                "keyword" : {
                  "type" : "keyword",
                  "ignore_above" : 256
                }
              }
            }
          }
        },
```

### Expected behavior
The schema migration is able to be completed as normal.

### Environment

- SaaS cluster
- Tasklist Version: 8.6.6 to 8.6.7 upgrade

### Additional context
Incident thread: https://camunda.slack.com/archives/C08903YEZ5J

panagiotisgts:
## Description

- Do not throw validation exception if difference is present on dynamic fields when comparing schemas
- Test case for ambiguous mappings between runtime and dated indices

## Checklist

- [ ] for CI changes:

## Related issues

closes #27022

### DIFF
--- a/tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
+++ b/tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
@@ -97,20 +97,15 @@ public class IndexSchemaValidatorUtil {
     }
 
     // Validate if the difference is dynamic
-    if (difference.getEntriesDiffering() != null) {
-      for (final PropertyDifference propertyDifference : difference.getEntriesDiffering()) {
-        final Object typeDefinition = propertyDifference.getLeftValue().getTypeDefinition();
-        if (propertyDifference.getLeftValue().getTypeDefinition() instanceof Map) {
-          final Map<String, Object> typeDefMap = (Map<String, Object>) typeDefinition;
-          final Object dynamicValue = typeDefMap.getOrDefault("dynamic", false);
-          if (dynamicValue.equals(true)) {
-            LOGGER.debug(
-                String.format(
-                    "Difference is on dynamic field - continue initialization: %s.",
-                    indexDescriptor.getIndexName()));
-            return;
-          }
-        }
+    if (difference.getEntriesDiffering() != null && difference.getEntriesDiffering().size() > 0) {
+      if (difference.getEntriesDiffering().stream().noneMatch(this::nonDynamicPropertyDifference)) {
+        LOGGER.debug(
+            String.format(
+                "Difference is on dynamic field - continue initialization - left %s, right %s. Actual diff values: %s",
+                difference.getLeftIndexMapping().getIndexName(),
+                difference.getRightIndexMapping().getIndexName(),
+                difference.getEntriesDiffering()));
+        return;
       }
     }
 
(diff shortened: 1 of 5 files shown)


tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
```diff
@@ -97,20 +97,15 @@ public void validateDifferenceAndCollectNewFields(
...
-          if (dynamicValue.equals(true)) {
-            LOGGER.debug(
-                String.format(
-                    "Difference is on dynamic field - continue initialization: %s.",
-                    indexDescriptor.getIndexName()));
-            return;
-          }
-        }
+    if (difference.getEntriesDiffering() != null && difference.getEntriesDiffering().size() > 0) {
+      if (difference.getEntriesDiffering().stream().noneMatch(this::nonDynamicPropertyDifference)) {
```
houssain-barouni, line 101: 🔧 Use AND instead of nesting if statements
```suggestion
    if (difference.getEntriesDiffering() != null && difference.getEntriesDiffering().size() > 0
      && difference.getEntriesDiffering().stream().noneMatch(this::nonDynamicPropertyDifference)) {
```

tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
```diff
@@ -97,20 +97,15 @@ public void validateDifferenceAndCollectNewFields(
...
+        LOGGER.debug(
+            String.format(
+                "Difference is on dynamic field - continue initialization - left %s, right %s. Actual diff values: %s",
+                difference.getLeftIndexMapping().getIndexName(),
+                difference.getRightIndexMapping().getIndexName(),
+                difference.getEntriesDiffering()));
```
houssain-barouni, line 107: ❓ this log will show-up twice in the logs, isn't?

tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
```diff
@@ -179,14 +174,37 @@ private IndexMappingDifference getIndexMappingDifference(
...
+          if (currentDifference.getEntriesDiffering().stream()
+              .anyMatch(this::nonDynamicPropertyDifference)) {
+            throw new TasklistRuntimeException(
+                "Ambiguous schema update. First bring runtime and data indices to one schema."
+                    + " Difference 1:"
+                    + difference
+                    + ". Difference 2: "
+                    + currentDifference);
+          } else {
+            LOGGER.debug(
```
houssain-barouni, line 186: ❓ is it ok if we keep `difference = currentDifference` in this case ? probably yes, but I am not 100% sure how this works?

tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
```diff
@@ -179,14 +174,37 @@ private IndexMappingDifference getIndexMappingDifference(
...
     }
     return difference;
   }
+
+  private boolean nonDynamicPropertyDifference(final PropertyDifference propertyDifference) {
+    final Object typeDefinition = propertyDifference.getLeftValue().getTypeDefinition();
+    if (propertyDifference.getLeftValue().getTypeDefinition() instanceof Map) {
+      final Map<String, Object> typeDefMap = (Map<String, Object>) typeDefinition;
+      final Object dynamicValue = typeDefMap.getOrDefault("dynamic", false);
+      if (dynamicValue.equals("true")) {
```
houssain-barouni, line 204: previously this was
```java
final Object dynamicValue = typeDefMap.getOrDefault("dynamic", false);
 if (dynamicValue.equals(true)) {
```
it is unclear to me whether the map contains strings or booleans or both ??

I propose we do the following

```suggestion
final boolean isDynamic = Boolean.parseBoolean(typeDefMap.getOrDefault("dynamic", "false").toString());
if (isDynamic) {
```

"true" → true
"false", null, and anything else → false
true (Boolean) → "true".toString() → true

houssain-barouni (commented):
Thanks for fixing this 👍 
I had mostly questions and requested some clean-up

tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
panagiotisgts: Not sure I get your point here. This check is to ensure that all indices runtime and dated ones contain the same diffs as the ones present when comparing the current JSON index mapping with the current Index mapping present on Elasticsearch.

houssain-barouni: my point is that the currentDifference may contain some ignored dynamic mappings, but I guess it should be fine if we use it as a reference

tasklist/els-schema/src/main/java/io/camunda/tasklist/util/IndexSchemaValidatorUtil.java
panagiotisgts: I removed the second instance of this log message, as I think it's not gonna get triggered anytime.

houssain-barouni (approved):

PR #27459, opened 2025-01-24
refactor: migrate to Identity v2 authentication in Operate frontend (with shim)

emulbreh:
## Description

This is a rebased version of https://github.com/camunda/camunda/pull/25912 with a shim for the /v2/authentication/me endpoint.

PR #27485, opened 2025-01-27
Adapt operate to work with new security filter chains

emulbreh:
## Description
Builds on https://github.com/camunda/camunda/pull/27459 and https://github.com/camunda/camunda/pull/26960

Part of https://github.com/camunda/camunda/issues/26804


megglos:
I will extract https://github.com/camunda/camunda/pull/27485/commits/bd6734b8d9f48cc32f140584d8a0d329ed3aa9d5 into a dedicated PR and align on it with the exporter team.

operate/Makefile
```diff
@@ -3,14 +3,18 @@
 .PHONY: env-up
 env-up:
 	docker compose -f docker-compose.yml up -d elasticsearch \
-    && mvn install -DskipTests=true -Dskip.fe.build=false -DskipChecks -DskipQaBuild \
+    && mvn install -DskipTests=true -Dskip.fe.build=true -DskipChecks -DskipQaBuild \
```
emulbreh, line 6: I'm not sure if this change is right.

operate/client/e2e-playwright/api/camunda8.ts
```diff
@@ -11,7 +11,8 @@ import {Camunda8} from '@camunda8/sdk';
 const camunda8 = new Camunda8({
   CAMUNDA_OAUTH_DISABLED: true,
   CAMUNDA_SECURE_CONNECTION: false,
-  ZEEBE_REST_ADDRESS: 'http://localhost:8086',
+  ZEEBE_REST_ADDRESS:
+    process.env.ZEEBE_REST_GATEWAY_ADDRESS || 'http://localhost:8086',
```
emulbreh, line 15:
```suggestion
    process.env.ZEEBE_REST_GATEWAY_ADDRESS || 'http://localhost:8080',
```

Maybe that would allow us to remove the overrides for `ZEEBE_REST_GATEWAY_ADDRESS`.

operate/client/playwright.config.ts
```diff
@@ -69,6 +69,10 @@ const config: PlaywrightTestConfig = {
     actionTimeout: 0,
     baseURL: `http://localhost:${getPort()}${getBasePath()}`,
     trace: 'retain-on-failure',
+    httpCredentials: {
+      username: 'demo',
+      password: 'demo',
+    },
```
emulbreh, line 75: these might not be necessary now

megglos:
Created the exporter PR https://github.com/camunda/camunda/pull/28154

operate/client/e2e-playwright/constants.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+
+const URL_PATTERN = /^.*\/(api|v2).*$/i;
```
megglos, line 9: 🔧 To make clear this is about api urls?
```suggestion
const URL_API_PATTERN = /^.*\/(api|v2).*$/i;
```

operate/client/src/App/Decisions/Decision/tests/decisionOperations.test.tsx
```diff
@@ -61,37 +61,13 @@ describe('<Decision /> - operations', () => {
     ).not.toBeInTheDocument();
   });
 
-  it('should not show delete button when user has no permissions', () => {
```
megglos, line 64: ❌ should we bring that back?

operate/client/e2e-playwright/api/camunda8.ts
Ben-Sheppard: My concern with changing this default is that I don't know 100% that its not in use anymore so for now I'll leave it at that

operate/client/src/App/Decisions/Decision/tests/decisionOperations.test.tsx
Ben-Sheppard: Perhaps one for @pedesen - I'm not sure this test restores much value as its based on the old permission formats of `read` etc. We should replace it with a new test for sure but I'm not sure we are in a place to do that yet (we need authorizations in place and exposed)

pedesen: It's based also on the resource based permissions which are still in place on the (old) internal Operate API as far as I know. Or did this change?

Ben-Sheppard: Nothing has changed there AFAIK but the permissions object is not returned by the `/me` endpoint so the test in its old state needs to change. As I understood @pedesen - we now just look at the `authorizedApplications` field to know if a user can access the component however the buttons in the UI need to have the new authorizations passed via something like `/me/authorizations`. 

That was my understanding based on [this](https://camunda.slack.com/archives/C06UYJMMETZ/p1737026427216549) thread, and if thats correct I think this test is fine to leave out (for 8.8)

zeebe/exporters/rdbms-exporter/pom.xml
```diff
@@ -80,5 +80,10 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>camunda-authentication</artifactId>
```
megglos, line 86: ❌ why is it here? 🤔

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/BasicAuthenticationUserService.java
```diff
@@ -0,0 +1,34 @@
...
+public class BasicAuthenticationUserService extends AuthUserService {
+
+  @Override
+  public UserDto createUserDtoFrom(final UsernamePasswordAuthenticationToken authentication) {
+    final CamundaUser user = (CamundaUser) authentication.getPrincipal();
+    return new UserDto()
+        .setUserId(user.getUserId())
+        .setDisplayName(user.getDisplayName())
+        .setCanLogout(true)
+        .setPermissions(List.of(Permission.READ, Permission.WRITE));
```
megglos, line 32: ❓ I guess here we actually need to resolve the actual permissions? Can we actually cover this with https://github.com/camunda/camunda/issues/27885 ? Will we actually keep that service going forward or have something central from the authentication module?

operate/qa/util/src/main/java/io/camunda/operate/qa/util/TestContainerUtil.java
```diff
@@ -524,7 +525,8 @@ public ZeebeContainer startZeebe(final String version, final TestContext testCon
           .withEnv("ZEEBE_CLOCK_CONTROLLED", "true")
           .withEnv("ZEEBE_BROKER_DATA_DISKUSAGEREPLICATIONWATERMARK", "0.99")
           .withEnv("ZEEBE_BROKER_DATA_DISKUSAGECOMMANDWATERMARK", "0.98")
-          .withEnv("ZEEBE_BROKER_DATA_SNAPSHOTPERIOD", "1m");
+          .withEnv("ZEEBE_BROKER_DATA_SNAPSHOTPERIOD", "1m")
+          .withEnv(AuthenticationProperties.getAllowUnauthenticatedApiAccessEnvVar(), "true");
```
megglos, line 529: 🔧 given it's the default for now, I think we can remove that here + the dependency on the auth module in the pom.
```suggestion
          .withEnv("ZEEBE_BROKER_DATA_SNAPSHOTPERIOD", "1m");
```

operate/client/e2e-playwright/api/camunda8.ts
megglos: We should be able to track that down by usage of `yarn run test:e2e` ? I also see the `8086` port only being used in the operate e2e workflow, so I would argue it's safe to change to the new default port `8080` indeed.

.github/workflows/operate-e2e-tests.yml
```diff
@@ -115,14 +104,18 @@ jobs:
         # we're excluding optimize from the build, not to impact this operate's workflow.
         run: ./mvnw clean install -T1C -DskipChecks -P -docker,skipFrontendBuild -DskipTests=true -B -DskipRemoteStaging=true -Dmaven.deploy.skip=true
       - name: Start Operate
-        run: ./mvnw -q -B spring-boot:start -pl dist -Dspring-boot.run.fork=true -Dspring-boot.run.main-class=io.camunda.application.StandaloneOperate
+        run: >
+          ./mvnw -q -B spring-boot:start -pl dist -Dspring-boot.run.fork=true
+          -Dspring-boot.run.main-class=io.camunda.application.StandaloneCamunda
+          -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=operate,consolidated-auth,broker -Dzeebe.broker.exporters.camundaexporter.classname=io.camunda.exporter.CamundaExporter -Dzeebe.broker.exporters.camundaexporter.args.connect.url=http://localhost:9200 -Dzeebe.broker.exporters.camundaexporter.args.bulk.size=100 -Dzeebe.broker.backpressure.enabled=false -Dcamunda.security.authentication.basic.allow-unauthenticated-api-access=true -Dzeebe.broker.exporters.camundaexporter.args.index.shouldWaitForImporters=false"
```
megglos, line 110: 🔧 let's remove the default config value for unauthenticated api access here as well
```suggestion
          -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=operate,consolidated-auth,broker -Dzeebe.broker.exporters.camundaexporter.classname=io.camunda.exporter.CamundaExporter -Dzeebe.broker.exporters.camundaexporter.args.connect.url=http://localhost:9200 -Dzeebe.broker.exporters.camundaexporter.args.bulk.size=100 -Dzeebe.broker.backpressure.enabled=false -Dzeebe.broker.exporters.camundaexporter.args.index.shouldWaitForImporters=false"
```

operate/Makefile
```diff
@@ -5,12 +5,16 @@ env-up:
 	docker compose -f docker-compose.yml up -d elasticsearch \
     && mvn install -DskipTests=true -Dskip.fe.build=false -DskipChecks -DskipQaBuild \
 	&& CAMUNDA_OPERATE_IMPORTERENABLED=false \
+	CAMUNDA_SECURITY_AUTHORIZATIONS_ENABLED=false \
+	CAMUNDA_OPERATE_ELASTICSEARCH_CREATESCHEMA=false \
```
megglos, line 9: 🔧  I guess this also helped to avoid the exporter getting stuck? I wonder if we stick to this approach or also use `ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_INDEX_SHOULDWAITFORIMPORTERS=false` here, wdyt?

operate/Makefile
```diff
@@ -96,4 +100,6 @@ start-e2e:
 	CAMUNDA_OPERATE_CSRF_PREVENTION_ENABLED=false \
 	SERVER_PORT=8081 \
 	MANAGEMENT_SERVER_PORT=9601 \
-	mvn clean install -DskipTests -DskipChecks -f ../dist/pom.xml exec:java -Dexec.mainClass="io.camunda.application.StandaloneCamunda" -Dspring.profiles.active=operate,broker,auth
+	mvn clean install -DskipTests -DskipChecks -f ../dist/pom.xml exec:java \
```
megglos, line 103: ❓ should we align the env vars used here with `env-up:`? especially either `CAMUNDA_OPERATE_ELASTICSEARCH_CREATESCHEMA` or  `ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_INDEX_SHOULDWAITFORIMPORTERS` as this should be relevant here too?

megglos (approved):
A couple of suggestions I'd like to align on, only one thing blocking being a, from my perspective unneeded dependency on the exporter. Otherwise LGTM, pre-approving!

For context I only reviewed CI/backend changes, I'll leave the frontend to @pedesen .

megglos:
Quoting @Ben-Sheppard for reference @pedesen 🤞 😜 
> Ben: Thank you, I will take a look at the comments shortly so its compete from a BE perspective today **(hopefully Paddy will see no issues and we could merge on Monday**

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/BasicAuthenticationUserService.java
Ben-Sheppard: I think the context of [this](https://camunda.slack.com/archives/C06UYJMMETZ/p1737026427216549) thread may help here but the way I see it is the change to support this is larger, for example we need to use the new authorizations to correctly derive if `READ` and/or `WRITE` is present (using some form of mapping logic between new and old). 

I would agree with your suggestion though that we could consider it a part of  https://github.com/camunda/camunda/issues/27885 and make the issue a little broader to not only consider app access but also authorizations

zeebe/exporters/rdbms-exporter/pom.xml
Ben-Sheppard: I can't see a reason for it tbh, sometimes I know these are added because of the dependency checks detecting that its used (somehow) but not included in the pom

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/BasicAuthenticationUserService.java
megglos: Actually READ/Write will vanish and it will only be ACCESS according to https://docs.google.com/document/d/1ikxw9ARCv-MBwASnj7WhW5uHCndCRk2B35xi-cB2N0M/edit?tab=t.0#heading=h.c0kx9z3moo6t

operate/Makefile
Ben-Sheppard: Possible, I will switch it to the should wait for importers config though (along with the other places)

operate/webapp/src/main/java/io/camunda/operate/webapp/security/auth/BasicAuthenticationUserService.java
Ben-Sheppard: Eventually yes, but there are some stage between now and then, specifically I think the Operate UI changes (maybe the v2 api switch?)

operate/client/src/App/Decisions/Decision/tests/decisionOperations.test.tsx
pedesen: @Ben-Sheppard Okay let's try to get on the same page here. First I'd like to understand if the resource based permission objects are still returned along with the data by the (old) internal Operate API, because that's what we're testing here. 

In this case the `/api/decisions/grouped` endpoint returned a permission object for the different decision definitions. Here you can see the [test data](https://github.com/camunda/camunda/blob/5ab0ed3ce7eafe51d43d43323e094807da598336/operate/client/src/modules/mocks/groupedDecisions.ts#L64): Only `invoiceClassification` has a DELETE permission. 

Are these permissions objects still in place? Then I would be in favor of keeping the test for now and update it later.

pedesen: Next we will do is to migrate Operate to the V2 API, so we need to rethink how we get the resource based permission data. As you said, one approach would be [this](https://camunda.slack.com/archives/C06UYJMMETZ/p1737026427216549), but I'd like to further discuss this (probably best in sync meeting).

operate/client/e2e-playwright/api/camunda8.ts
pedesen: I agree, port `8086` was only used by the the zeebe service container [here](https://github.com/camunda/camunda/pull/27485/files#diff-d046a4c1810127d86e1a7d99c5953d3bcbf91529d672c7fc4a6fa98236b4bdb1L54-L64). Since this is removed now, we should change the default to `8080`.

operate/client/src/setupProxy.js
```diff
@@ -15,7 +15,7 @@ module.exports = function (app) {
     BASENAME,
     createProxyMiddleware({
       target: `http://localhost:${process.env.IS_E2E ? '8081' : '8080'}`,
-      pathFilter: ['/client-config.js', '/api', '/v2'],
+      pathFilter: ['/client-config.js', '/api', '/v2', '/login', '/logout'],
```
pedesen, line 18: This is needed to make our local frontend dev environment work. We're using a separate frontend setup with hot-reloasing on port 3000 (resp. 3001 for local e2e testing). Since the URL path for login and logout changed, this needs to be proxied here too.

pedesen (approved):
It looks very good to me :partying_face: 

I checked the local dev setup and the local e2e tests. After a small adjustments everything works fine :smile: 

There is still an open [discussion about removing a test case](https://github.com/camunda/camunda/pull/27485#pullrequestreview-2617109126), but it's not a blocker in my opinion.

operate/client/src/App/Decisions/Decision/tests/decisionOperations.test.tsx
megglos: > @Ben-Sheppard Okay let's try to get on the same page here. First I'd like to understand if the resource based permission objects are still returned along with the data by the (old) internal Operate API, because that's what we're testing here.

If I'm not missing anyhting, In case of the new auth setup, the userService will statically return READ & WRITE in the permissions, see [here](https://github.com/camunda/camunda/pull/27485/files#diff-65f7063fc2dfe9e1a9408445eb0fb6cc7766d78b282ca8286589ccd9e3f14182).

So from that regard resource based auths are not enforced anymore when the new setup is used, we need to migrate to the new Application authorizations there, which will be done with https://github.com/camunda/camunda/issues/27885

.github/workflows/operate-e2e-tests.yml
```diff
@@ -115,14 +104,18 @@ jobs:
         # we're excluding optimize from the build, not to impact this operate's workflow.
         run: ./mvnw clean install -T1C -DskipChecks -P -docker,skipFrontendBuild -DskipTests=true -B -DskipRemoteStaging=true -Dmaven.deploy.skip=true
       - name: Start Operate
-        run: ./mvnw -q -B spring-boot:start -pl dist -Dspring-boot.run.fork=true -Dspring-boot.run.main-class=io.camunda.application.StandaloneOperate
+        run: >
+          ./mvnw -q -B spring-boot:start -pl dist -Dspring-boot.run.fork=true
+          -Dspring-boot.run.main-class=io.camunda.application.StandaloneCamunda
+          -Dspring-boot.run.jvmArguments="-Dspring.profiles.active=operate,consolidated-auth,broker -Dzeebe.broker.exporters.camundaexporter.classname=io.camunda.exporter.CamundaExporter -Dzeebe.broker.exporters.camundaexporter.args.connect.url=http://localhost:9200 -Dzeebe.broker.exporters.camundaexporter.args.bulk.size=100 -Dzeebe.broker.backpressure.enabled=false -Dzeebe.broker.exporters.camundaexporter.args.index.shouldWaitForImporters=false"
```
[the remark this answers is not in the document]
megglos: that we have the exporter or that this is so "readable"? 😬 

Once CamundaExporter runs by default, we may only have `-Dspring.profiles.active=operate,broker -Dzeebe.broker.exporters.camundaexporter.args.index.shouldWaitForImporters=false` 🤞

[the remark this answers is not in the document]
pedesen: No, I was genuinely happy about the switch to CamundaStandalone, which had not worked before. 😊

PR #27496, opened 2025-01-28
Fix dependency management: use specified versions

npepinpe:
## Description

This PR fixes an issue with our dependency management. It seems since April, we introduced `spring-boot-dependencies` into our parent POM. This means our project generally tries to use the same version for shared dependencies as Spring does (e.g. Netty, Awaitility, Mockito, etc.). I assume the goal was to avoid dependency convergence issues.

This was a nuclear bomb kind of solution however, because what it ended up doing is overwriting any version we later specify via other `POM/IMPORT` dependencies.

So take, for example, Micrometer (this is how I stumbled on it). We specify in the parent POM version `1.14.3`. If you look at the `mvn dependency:tree`, you'll see it's pulling in 1.13.9. This is extremely confusing, and hard to pinpoint: where is this coming from? We explicitly include the `micrometer-bom` in our `parent/pom` and set the version, after all.

The trick is, since the `micrometer-bom` is declared after the `spring-boot-dependencies`, any shared dependencies are ignored. The fix is to ensure all explicit, library-specific `POM/IMPORT` declarations are _before_ `spring-boot-dependencies`.

I personally would prefer _not_ use `spring-boot-dependencies`, and having explicit declarations: it's more verbose, but much easier to know and ensure we use the right versions. It also decouples us from the choices of another company on which library versions to use.

For this PR, I didn't do this yet, I'd like to know what others think (in particular those who included it originally), and opted for just declaring our library versions higher. I also had to move the `google-libraries` declaration second-last, as it overwrote our Protobuf, Netty, gRPC, etc., all so-called "first party libraries" from Google.

Note that this implies we may have been shipping with different versions than we thought, but generally this should not affect CVEs as our scanner runs on the resolved tree, not the POMs themselves.

It **does mean** however that we may have been merging dependency updates which break things unknowingly, because the version was not even used.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -1076,22 +1076,6 @@
       </dependency>
 
       <dependency>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter-webflux</artifactId>
-        <version>${version.spring-boot}</version>
-        <exclusions>
-          <exclusion>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
-          </exclusion>
-          <exclusion>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-to-slf4j</artifactId>
-          </exclusion>
-        </exclusions>
-      </dependency>
-
-      <dependency>
         <groupId>io.camunda</groupId>
         <artifactId>zeebe-broker</artifactId>
         <version>${project.version}</version>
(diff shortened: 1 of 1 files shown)


npepinpe:
@entangled90 - somehow I thought I requested a review, but I guess not :smile:

entangled90 (approved):
:rocket: 
However we potentially have the same problems, but not from spring now.
We probably need to think how to enforce this in a more consistent way

PR #27526, opened 2025-01-29
FT: Removing unnecessary files after installation

hawkab:
#### `apt-get purge --auto-remove -y`

This command removes unnecessary dependencies installed automatically when installing packages. It:

* Removes packages that were installed as dependencies but are not used (auto-remove).
* Removes package configuration files entirely (purge).

#### `rm -rf /var/lib/apt/lists/* /tmp/* ~/.cache/*`
Removing /var/lib/apt/lists/* files after installing a package is a common practice for Docker's look and feel

Pros of removing /var/lib/apt/lists/*

1 Reducing image size

* Cached package lists (/var/lib/apt/lists/*) can take up tens of megabytes.
* Removing these files reduces the final image size, especially if the image has a lot of dependencies.

2 Avoiding stale metadata

* If package lists are left in the image, when the container is reused (e.g. with docker exec), it will not be able to update correctly, since the lists are out of date.
* This is also important if the container will be running for a long time without being rebuilt.

3 Best practices for production environments

* In production, apt-get update is rarely needed in an already running container, since images must be immutable.
* This makes cached lists of installed packages unnecessary.


tmetzke:
Hi @hawkab,

Thanks for your proposal. In general, this contribution is reasonable if one wants to minimize Docker image size, and we see many examples in practice that do exactly this. However, as noted [in the comment](https://github.com/camunda/camunda/pull/27526/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R24-R25) above your proposed changes:

> \# FYI, installing packages via APT also updates the dpkg files, which are few MBs, but removing or
> \# caching them could break stuff (like not knowing the package is present) or container scanners

I'm closing this PR as `Not planned` due to that. Let us know if you have further thoughts.

Best,
Tobias

PR #27530, opened 2025-01-30
feat: create mapping accept mapping id

### LINKED ISSUE #27373: Create mapping accept mapping id

maryarm:
## Description

Add mapping id to api and UI.

## Related issues

closes #27373


clients/java/src/test/java/io/camunda/client/mapping/CreateMappingTest.java
```diff
@@ -27,6 +27,7 @@ public class CreateMappingTest extends ClientRestTest {
   public static final String CLAIM_NAME = "claimName";
   public static final String CLAIM_VALUE = "claimValue";
   public static final String NAME = "mappingName";
+  public static final String ID = "mappingID";
```
Ben-Sheppard, line 30: 🔧 I think this should be `mappingId`

migration/identity-migration/src/main/java/io/camunda/migration/identity/TenantMappingRuleMigrationHandler.java
```diff
@@ -57,6 +57,7 @@ protected MigrationStatusUpdateRequest processTask(final TenantMappingRule tenan
           new MappingDTO(
               tenantMappingRule.getClaimName(),
               tenantMappingRule.getClaimValue(),
+              tenantMappingRule.getName(),
```
Ben-Sheppard, line 60: ❓ Do we have any special considerations around what is an `id` for example can it contain spaces? do we need to, as an example, convert this to all lower case and replace spaces with `-`?

Ben-Sheppard (changes requested):
I can see that we have a `name` and now an `id` for mappings but these are not reflected in the `MappingCreatedHandler` class (we only set the claimName and ClaimValue). 

Also the `mappingId` needs to be added to `camunda-mapping.json` schemas

maryarm:
> I can see that we have a `name` and now an `id` for mappings but these are not reflected in the `MappingCreatedHandler` class (we only set the claimName and ClaimValue).
> 
> Also the `mappingId` needs to be added to `camunda-mapping.json` schemas

I think this comments is related to export, yes? Exporter changes will come later in separate PRs.

Ben-Sheppard (approved):

Issue #26078, opened 2024-12-13
Non-Prometheus Micrometer implementations only expose few C8 metrics


npepinpe:
**Describe the bug**

I would like to use a different backend than Prometheus for metrics with my C8 deployment. Unfortunately, when I configure one (e.g. Dynatrace), I only see a handful of C8 metrics. I would like to have access to the same metrics as the Prometheus backend.

**To Reproduce**

1. Enable the metrics endpoint via `management.endpoint.metrics.enabled=true`
2. Enable the simple metrics backend via `management.simple.metrics.export.enabled=true`
3. Query `http://localhost:9600/actuator/metrics`

You will only see the few metrics from Zeebe which are registered via Micrometer, instead of via Prometheus.

**Expected behavior**

All metrics are emitted by C8 are visible in _every_ Micrometer backend (except for those who do not support certain types).

**Environment:**
- Zeebe Version: >= 8.6.+

```[tasklist]
### Initial Assessment
- [ ] https://github.com/camunda/camunda/issues/26898
- [ ] https://github.com/camunda/camunda/issues/26909
```

```[tasklist]
### Migrate metrics
- [ ] https://github.com/camunda/camunda/issues/27009
- [ ] https://github.com/camunda/camunda/issues/26910
- [ ] https://github.com/camunda/camunda/issues/26906
```

```[tasklist]
### Testing and Validation
- [ ] https://github.com/camunda/camunda/issues/26901
- [ ] https://github.com/camunda/camunda/issues/26945
- [ ] https://github.com/camunda/camunda/issues/26946
```

```[tasklist]
### Configuration Changes
- [ ] https://github.com/camunda/camunda/issues/26897
- [ ] https://github.com/camunda/camunda/issues/26905
- [ ] https://github.com/camunda/camunda/issues/26950
```

```[tasklist]
### Follow-ups
- [ ] https://github.com/camunda/camunda/issues/26947
- [ ] https://github.com/camunda/camunda/issues/26943
```

```[tasklist]
### Documentation
- [ ] https://github.com/camunda/camunda/issues/26903
- [ ] https://github.com/camunda/camunda/issues/26938
```

npepinpe:
The goal here should be to _fully_ migrate away from the Prometheus simple client. The core of C8 should not use Prometheus directly, but instead Micrometer _only_.

The only modules which can use Prometheus should be: the benchmark applications (though it would be nice to migrate them to Micrometer, this can wait), and the dist module.

[...]

PR #27532, opened 2025-01-30
Migrate exporter director metrics to Micrometer

npepinpe:
## Description

This PR migrates the exporter director metrics from Prometheus simple client to Micrometer. The main changes is that the histograms now have `_seconds_bucket` as suffix, instead of just `_bucket`. There are also no `_created` metrics for all, but we were not using them either in the past. Everything else is kept the same, including the histogram buckets.

> [!Note]
> Micrometer suggests using more buckets to properly quantify percentiles, but it does mean an increase from ~350 bytes per histogram to ~1.5KB per histogram, so we should do it only if we think it's a good idea. It likely is, but I have no time to evaluate this right now.

With Micrometer, we now have to track metrics as instance level state. While their recommended usage is to recreate the builder or the `Meter.Id` every time you access a metric, this creates unnecessary garbage and possible allocations. To avoid this, you can simply keep track of the meters - or in the case of gauges, of the object tracking the state.

Since every combination of meter name + tag creates a new metric, the state has to be kept based on that. This can include multiple tags, though typically we don't use much more than two (if we ignore the common tags).

To help, I introduced a very simple collection, `Table`. This isn't a new concept - for example, Guava has a much more extensive implementation of this. I opted against using Guava's for now, mostly because we need so little from it. As usual, if we do need more functionality in the future, we should consider switching to a well established implementation. **This makes the PR bigger, but should help keep things simple in future PRs.**

> [!Note]
> A table is basically a map, where values are indexed by two keys, row and column. Think of it like a matrix.

I also decided to start using the Micrometer documentation format, in the hopes that we can in the future use their documentation generator (see [here](https://docs.micrometer.io/micrometer-docs-generator/reference/index.html))

Finally, the dashboard using the histograms with new names has been updated, now showing the union of both series (since the usage of one or the other is mutually exclusive).

## Related issues

related to #26078

### DIFF
--- a/monitor/grafana/camunda-performance.json
+++ b/monitor/grafana/camunda-performance.json
@@ -1579,7 +1579,7 @@
                 "uid": "$DS_PROMETHEUS"
               },
               "editorMode": "code",
-              "expr": "histogram_quantile(0.5, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "expr": "histogram_quantile(0.5, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval]) or rate(zeebe_exporting_latency_seconds_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
               "format": "time_series",
               "intervalFactor": 1,
               "legendFormat": "p50 (avg)",
@@ -1592,7 +1592,7 @@
                 "uid": "$DS_PROMETHEUS"
               },
               "editorMode": "code",
-              "expr": "histogram_quantile(0.99, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "expr": "histogram_quantile(0.99, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval]) or rate(zeebe_exporting_latency_seconds_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
               "hide": false,
               "instant": false,
               "legendFormat": "p99",
@@ -1605,7 +1605,7 @@
                 "uid": "$DS_PROMETHEUS"
               },
               "editorMode": "code",
-              "expr": "histogram_quantile(0.90, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "expr": "histogram_quantile(0.90, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval]) or rate(zeebe_exporting_latency_seconds_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
               "hide": false,
               "instant": false,
               "legendFormat": "p90",
@@ -1970,7 +1970,7 @@
                 "uid": "$DS_PROMETHEUS"
               },
               "editorMode": "code",
-              "expr": "histogram_quantile(0.50, sum(rate(zeebe_exporter_exporting_duration_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "expr": "histogram_quantile(0.50, sum(rate(zeebe_exporter_exporting_duration_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval]) or rate(zeebe_exporter_exporting_duration_seconds_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval])) by (le))",
               "format": "time_series",
               "intervalFactor": 1,
               "legendFormat": "p50",
@@ -1983,7 +1983,7 @@
                 "uid": "$DS_PROMETHEUS"
               },
               "editorMode": "code",
-              "expr": "histogram_quantile(0.99, sum(rate(zeebe_exporter_exporting_duration_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "expr": "histogram_quantile(0.99, sum(rate(zeebe_exporter_exporting_duration_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval]) or rate(zeebe_exporter_exporting_duration_seconds_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval])) by (le))",
               "hide": false,
               "instant": false,
               "legendFormat": "p99",
@@ -1996,7 +1996,7 @@
                 "uid": "$DS_PROMETHEUS"
               },
               "editorMode": "code",
-              "expr": "histogram_quantile(0.90, sum(rate(zeebe_exporter_exporting_duration_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "expr": "histogram_quantile(0.90, sum(rate(zeebe_exporter_exporting_duration_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval]) or rate(zeebe_exporter_exporting_duration_seconds_bucket{namespace=~\"$namespace\",partition=~\"$partition\"}[$__rate_interval])) by (le))",
               "hide": false,
               "instant": false,
               "legendFormat": "p90",
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
@@ -33,8 +33,8 @@ public final class ExporterMetrics {
   private final Map<String, AtomicLong> lastUpdatedExportedPositions = new HashMap<>();
   private final AtomicInteger exporterState = new AtomicInteger();
   private final Map<ValueType, Timer> exportingLatency = new HashMap<>();
-  private final Table<Timer, String, ValueType> exporterExportingDuration = Table.simple();
-  private final Table<Counter, ExporterActionKeyNames, ValueType> exporterEvents =
+  private final Table<String, ValueType, Timer> exporterExportingDuration = Table.simple();
+  private final Table<ExporterActionKeyNames, ValueType, Counter> exporterEvents =
       Table.ofEnum(ExporterActionKeyNames.class, ValueType.class, Counter[]::new);
 
   private final MeterRegistry meterRegistry;
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
@@ -7,120 +7,99 @@
  */
 package io.camunda.zeebe.broker.exporter.stream;
 
+import io.camunda.zeebe.broker.exporter.stream.ExporterMetricsDoc.ExporterActionKeyNames;
 import io.camunda.zeebe.protocol.record.ValueType;
-import io.prometheus.client.Counter;
-import io.prometheus.client.Gauge;
-import io.prometheus.client.Histogram;
+import io.camunda.zeebe.util.CloseableSilently;
+import io.camunda.zeebe.util.collection.Table;
+import io.camunda.zeebe.util.micrometer.ExtendedMeterDocumentation;
+import io.camunda.zeebe.util.micrometer.MicrometerUtil;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 public final class ExporterMetrics {
-
-  private static final String LABEL_NAME_PARTITION = "partition";
   private static final String LABEL_NAME_EXPORTER = "exporter";
   private static final String LABEL_NAME_ACTION = "action";
   private static final String LABEL_NAME_VALUE_TYPE = "valueType";
-  private static final String NAMESPACE_ZEEBE = "zeebe";
 
-  private static final Histogram EXPORTING_LATENCY =
-      Histogram.build()
-          .namespace(NAMESPACE_ZEEBE)
-          .name("exporting_latency")
-          .help("Time between a record is written until it is picked up for exporting (in seconds)")
-          .labelNames(LABEL_NAME_PARTITION, LABEL_NAME_VALUE_TYPE)
-          .register();
+  private final Map<String, AtomicLong> lastExportedPositions = new HashMap<>();
+  private final Map<String, AtomicLong> lastUpdatedExportedPositions = new HashMap<>();
+  private final AtomicInteger exporterState = new AtomicInteger();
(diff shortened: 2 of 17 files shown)


npepinpe:
To review, I would recommend looking at the `ExporterMetricsDoc` and `ExporterMetrics` first. Then looking at the `Table` and other utilities created, but you would have an idea of their purpose.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetrics.java
npepinpe, line 1: :exclamation: I don't assign the partition label here, because it's alright a common tag on the partition meter registry used.

monitor/grafana/camunda-performance.json
npepinpe, line 1: :exclamation: Using the union of both series - since a broker only outputs one or the other - allows us to have a single continuous series in the end. It's not backwards compatible per se, because you have to update the dashboard when you update the application, but it is forward compatible, so you can just update the dashboard first.

I know, not perfect, but I think it's an acceptable solution.

zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterMetricsDoc.java
npepinpe, line 1: :exclamation: This will allow us to generate documentation for our metrics via https://docs.micrometer.io/micrometer-docs-generator/reference/index.html. I've already tried it - it's not perfect by any means, but it's 100% better than nothing (which is what we have now :upside_down_face:). There is a [small bug](https://github.com/micrometer-metrics/micrometer-docs-generator/issues/338) but it has a workaround. I would propose generating docs at the end in the dist module, so I'm not doing it now.

The sad news is it seems Docusaurus does not support AsciiDoc, so we wouldn't be able to ingest the metrics. I think having the Javadoc in itself is useful, and later we can look into making this public in some shape or form.

zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExporterDisableTest.java
npepinpe, line 1: :exclamation: I had some flakiness here, and looking into it...having a 100ms max interval is bound to have some flakiness in general, no? esp. on a machine doing multiple things...

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/Table.java
npepinpe, line 1: :exclamation: This allows us to keep track of individual metrics with up to 2 labels/tags, without having to recreate the builders/meter IDs every time. It doesn't scale neatly to 3, but generally thanks to the `CompositeMeterRegistry` instances with common tags, we don't have more than 2 tags. If we do, you could technically do a `Table<Map<String, T>, String, String>`, but I guess it's pretty clunky. In that case we might want to use a recursive structure. 

I'm now thinking a graph (or forest, I guess) is much better representation, where each level represents a tag :thinking:

npepinpe: For fun, I created such an intrusive "forest" structure:

<details><summary>FixedDepthForest.java</summary>

```java
/*
 * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
 * one or more contributor license agreements. See the NOTICE file distributed
 * with this work for additional information regarding copyright ownership.
 */
package io.camunda.zeebe.util.collection;

import java.util.HashMap;
import java.util.Map;

/**
 * A forest (i.e. a map of trees) with a fixed depth, such that every tree is expected to always
 * have the exact same depth.
 *
 * <p>This is primarily used to track the {@link io.micrometer.core.instrument.Meter} or {@link
 * io.micrometer.core.instrument.Meter.Id} to avoid creating garbage every time we increment a
 * counter or set a gauge, and have to recreate the original {@link
 * io.micrometer.core.instrument.Meter.Id} based on the name and a series of tag.
 *
 * <p>You should only use this class if your metric has 3 or more tags. If you have 2 tags, use a
 * {@link Table}. If you have a single tag, use a {@link Map}. If you don't have a tag, just track
 * the meter or ID directly (or the state object in the case of a gauge).
 *
 * <p>The expected usage is to use {@link #computeIfAbsent(ValueFunction, String...)}, and pass
 * exactly the number of keys as you expect the tags. This should <em>always</em> be in the same
 * order, though we cannot enforce it directly. <strong>Make sure to always call them with the exact
 * same order they are registered in</strong>.
 *
 * <p>Example usage:
 *
 * <pre>{@code
 * // assume a counter metric with 4 tags: a, b, c, and d
 * final FixedDepthForest<Counter> counters = new FixedDepthForest(4);
 *
 * public void incrementCounter(
 *   final String aValue,
 *   final String bValue,
 *   final String cValue,
 *   final String dValue) {
 *   counters.computeIfAbsent(this::registerCounter, aValue, bValue, cValue, dValue).increment();
 * }
 *
 * private Counter registerCount(final String... tags) {
 *   return Counter.builder("my.counter")
 *     .description("My super counter")
 *     .tag("a", tags[0])
 *     .tag("b", tags[1])
 *     .tag("c", tags[2])
 *     .tag("d", tags[3])
 *     .register(meterRegistry);
 * }
 * }</pre>
 *
 * @param <T> the type of the values to store (e.g. {@link io.micrometer.core.instrument.Counter},
 *     {@link java.util.concurrent.atomic.AtomicLong})
 */
public final class FixedDepthForest<T> {
  private final TreeNode<T> root = TreeNode.root();
  private final int depth;

  /**
   * @param depth the expected depth of each tree
   * @throws IllegalArgumentException if the depth is less than or equal to 2; in such cases, use a
   *     {@link Table} or a {@link Map}
   */
  public FixedDepthForest(final int depth) {
    this.depth = depth;

    if (depth < 2) {
      throw new IllegalArgumentException(
          "Expected at least two levels; if you only have one, use a map");
    }
  }

  /** Clears the forest entirely. */
  public void clear() {
    root.clear();
  }

  /** Returns how many leaves are contained in the forest. */
  public int size() {
    return root.size();
  }

  /**
   * Returns the current leaf value down the branch identified by the given keys. If there is none,
   * will call the {@code valueSupplier} with the keys passed here, set it as the leaf, and return
   * it.
   *
   * @param valueSupplier a function to compute a new value if there is none stored yet
   * @param keys the branch to follow
   * @return the existing value, or the result of calling valueSupplier if none
   * @throws IllegalArgumentException if the number of keys is not equal to the depth of the forest
   */
  public T computeIfAbsent(final ValueFunction<T> valueSupplier, final String... keys) {
    if (keys.length != depth) {
      throw new IllegalArgumentException(
          "Expected %d keys (i.e. the depth of each tree), but got %d"
              .formatted(depth, keys.length));
    }

    final var leaf = root.computeIfAbsent(valueSupplier, 0, keys);
    return leaf == null ? null : leaf.value();
  }

  /**
   * Sets the current leaf value down the branch identified by the given keys, overwriting anything
   * there previously (if any).
   *
   * @param value the new value to set
   * @param keys the branch to follow
   * @throws IllegalArgumentException if the number of keys is not equal to the depth of the forest
   */
  public void put(final T value, final String... keys) {
    if (keys.length != depth) {
      throw new IllegalArgumentException(
          "Expected %d keys (i.e. the depth of each tree), but got %d"
              .formatted(depth, keys.length));
    }

    root.put(value, 0, keys);
  }

  /**
   * Returns the current leaf value down the branch identified by the given keys, or null if there
   * is none.
   *
   * @param keys the branch to follow
   * @return the existing value, or null if there is none
   * @throws IllegalArgumentException if the number of keys is not equal to the depth of the forest
   */
  public T get(final String... keys) {
    if (keys.length != depth) {
      throw new IllegalArgumentException(
          "Expected %d keys (i.e. the depth of each tree), but got %d"
              .formatted(depth, keys.length));
    }

    final var leaf = root.get(0, keys);
    return leaf == null ? null : leaf.value();
  }

  private record TreeNode<T>(T value, Map<String, TreeNode<T>> children) {
    private static <T> TreeNode<T> leaf(final T value) {
      return new TreeNode<>(value, null);
    }

    private static <T> TreeNode<T> root() {
      return new TreeNode<>(null, new HashMap<>());
    }

    @Override
    public T value() {
      if (value == null) {
        throw new IllegalStateException(
            "Expected to access a value node, but instead hit a sub-tree");
      }

      return value;
    }

    private TreeNode<T> computeIfAbsent(
        final ValueFunction<T> valueSupplier, final int cursor, final String... keys) {
      if (cursor == keys.length - 1) {
        final var value = children.get(keys[cursor]);
        if (value != null) {
          return value;
        }

        final var newValue = leaf(valueSupplier.apply(keys));
        children.put(keys[cursor], newValue);
        return newValue;
      }

      return children
          .computeIfAbsent(keys[cursor], k -> root())
          .computeIfAbsent(valueSupplier, cursor + 1, keys);
    }

    private void clear() {
      children.clear();
    }

    private int size() {
      if (value != null) {
        return 1;
      }

      return children.values().stream().mapToInt(TreeNode::size).sum();
    }

    private void put(final T value, final int cursor, final String... keys) {
      if (cursor == keys.length - 1) {
        children.put(keys[cursor], leaf(value));
        return;
      }

      children.computeIfAbsent(keys[cursor], k -> root()).put(value, cursor + 1, keys);
    }

    private TreeNode<T> get(final int cursor, final String... keys) {
      if (cursor == keys.length) {
        return this;
      }

      final var key = keys[cursor];
      final var node = children.get(key);
      return node == null ? null : node.get(cursor + 1, keys);
    }
  }

  /**
   * Computes the new value to set as the leaf of the branch identified by the given keys.
   *
   * @param <T> the type of the leaf value
   */
  @FunctionalInterface
  public interface ValueFunction<T> {
    T apply(final String... keys);
  }
}
```

</details>

And some basic usage/test:

```java
public static void main(final String[] args) {
  final var forest = new FixedDepthForest<>(3);
  forest.put(5, "1", "2", "3");
  forest.put(4, "1", "2", "4");
  forest.put(6, "4", "5", "6");

  System.out.println("[1,2,3] = " + forest.get("1", "2", "3"));
  System.out.println("[1,2,4] = " + forest.get("1", "2", "4"));
  System.out.println("[1,2,2] = " + forest.get("1", "2", "2"));
  System.out.println("[1,3,2] = " + forest.get("1", "3", "2"));
  System.out.println("[10,3,2] = " + forest.get("10", "3", "2"));
  System.out.println("[4,5,6] = " + forest.get("4", "5", "6"));
  System.out.println("[5,6,7] = " + forest.computeIfAbsent(keys -> 7, "5", "6", "7"));
  System.out.println("[5,6,7] = " + forest.computeIfAbsent(keys -> 8, "5", "6", "7"));

  System.out.println("Size = " + forest.size());
  forest.clear();
  System.out.println("Size = " + forest.size());
}
```

npepinpe: I'd still wait to see if we really need it, I don't really like having untyped keys and variadic functions :shrug:

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/EnumTable.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+package io.camunda.zeebe.util.collection;
+
+import java.util.function.BiFunction;
+import java.util.function.IntFunction;
+
+/**
+ * A table implementation optimized for {@link Enum} keys. See {@link MapTable} for more on tables.
+ *
+ * <p>Enums allow us to define a finite keyspace, so we can efficiently use a single array to
```
entangled90, line 16: :wrench: We can also mention that we are using `row-major` ordering as well in the intro, so that it's quicker to double check if we are using row-major or column-major ordering. 
Great explanation anyway! :rocket:

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/EnumTable.java
```diff
@@ -0,0 +1,98 @@
...
+ * BAR | 4 | 5 | 6
+ * </pre>
+ *
+ * Let FOO_A denote the value at row FOO and column A (i.e. 1). If we map this as an array, we have:
+ *
+ * <pre>
+ * [ FOO_A, FOO_B, FOO_C, BAR_A, BAR_B, BAR_C ]
+ * </pre>
+ */
+final class EnumTable<T, RowT extends Enum<RowT>, ColT extends Enum<ColT>>
```
entangled90, line 55: :x: I find it confusing having `T` as the first type paramete because it's the opposite of `Map<K,V>`. Guava's `Table` puts the keys first in the type signature. 
```suggestion
final class EnumTable<RowT extends Enum<RowT>, ColT extends Enum<ColT>, T>
```

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/MapTable.java
```diff
@@ -0,0 +1,40 @@
...
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.util.collection;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiFunction;
+
+/** A very simple implementation using a map-of-map approach. */
+final class MapTable<T, RowT, ColT> implements Table<T, RowT, ColT> {
```
entangled90, line 15: :x: same as above
```suggestion
final class MapTable<RowT, ColT> implements Table<T, RowT, ColT, T> {
```

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/Table.java
entangled90: A simpler solution for such cases it's to just use a map with composite key (tuples).
Not sure if we already have implemented tuple as it does not exist in the standard library, 
The downside is that the composite key must be allocated every time a `get` is performed on the map, but the scope is so small that the tuple most of the times is allocated on the stack. I can't guarantee this (who can?) but if it's important we can check with a simple benchmark and or look at the C2 Jitted asm.

monitor/grafana/camunda-performance.json
```diff
@@ -1592,7 +1592,7 @@
                 "uid": "$DS_PROMETHEUS"
               },
               "editorMode": "code",
-              "expr": "histogram_quantile(0.99, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
+              "expr": "histogram_quantile(0.99, sum(rate(zeebe_exporting_latency_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval]) or rate(zeebe_exporting_latency_seconds_bucket{namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (le))",
```
entangled90, line 1595: :question: is it correct that we don't need the `sum` in the second metric?

entangled90 (changes requested):
Looks great :rocket:, I would just change the type parameters ordering for `Table`.

I guess it would be impossible to know if customers will be impacted by the change in names of the metrics related to histograms. Probably not many of them, but we need to remember to mention this when we release it.

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/Table.java
npepinpe: Yeah I also thought even for Meter.Id it might sometimes get scalar-replaced, but I don't know enough to guarantee that :smile: So it'd be the same with tuples. We do have a tuple somewhere (or used to anyway), but it would also get verbose really fast in Java, e.g. `Map<Tuple<Tuple<String, String>, String>, Counter>` :smile:

monitor/grafana/camunda-performance.json
npepinpe: `sum` is the outer call, the `or` is inside of it. So basically: `sum(rate(a) or rate(b))`, where `a` is the first series, and `b` is the new one.

npepinpe:
> I guess it would be impossible to know if customers will be impacted by the change in names of the metrics related to histograms. Probably not many of them, but we need to remember to mention this when we release it.

I think we can't discard that, and we should have a release announcement. But generally we don't have metrics/dashboards in the public API for this reason (though that will likely change). And the workaround is to update the dashboard, or if you have a custom one, to use the union of both series.

zeebe/util/src/main/java/io/camunda/zeebe/util/collection/Table.java
entangled90: Generally tuples in most languages have N-arity, so you would have` Tuple2, Tuple3` and so on..

entangled90 (approved):
:rocket:

PR #27534, opened 2025-01-30
Migrate job stream metrics to Micrometer

npepinpe:
## Description

This PR migrates the job stream metrics in the broker from Prometheus to Micrometer. Nothing special changed. Depends on #27532 for some Micrometer utilities.

No required dashboard changes because none of the labels/metric names have changed.

## Related issues

related to #26078

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/jobstream/JobStreamMetrics.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/jobstream/JobStreamMetrics.java
@@ -9,63 +9,68 @@ package io.camunda.zeebe.broker.jobstream;
 
 import io.camunda.zeebe.transport.stream.api.RemoteStreamMetrics;
 import io.camunda.zeebe.transport.stream.impl.messages.ErrorCode;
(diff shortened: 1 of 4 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/jobstream/JobStreamMetrics.java
```diff
@@ -9,63 +9,68 @@
...
-    PUSH_FAILED_COUNT.inc();
+    pushFailedCount.increment();
   }
 
   @Override
   public void pushTryFailed(final ErrorCode code) {
-    PUSH_TRY_FAILED_COUNT.labels(code.name()).inc();
+    final var meterDoc = JobStreamMetricsDoc.PUSH_TRY_FAILED_COUNT;
+    pushTryFailedCount
+        .computeIfAbsent(code, ignored -> registerCounter(meterDoc, Tag.of("code", code.name())))
```
entangled90, line 73: :wrench: ErrorCode is an enum, so we could use an array directly, similar to Table ?

entangled90 (approved):

Issue #27379, opened 2025-01-23
Optimize is not showing current user name in Profile in a Self-Managed setup with a custom IDP


megglos:
Engineering DRI: @abremard

### Describe the bug

In a Self-Managed OAuth setup, Optimize tries to load the user name from the getName getter from the UserDetails [here](https://github.com/camunda/camunda/blob/main/optimize/backend/src/main/java/io/camunda/optimize/service/security/CCSMTokenService.java#L211) and if null uses the bare userId  which is the `sub` from the access token.
However the UserDetails also serves a dedicated username [property](https://github.com/camunda-cloud/identity/blob/main/sdk-java/src/main/java/io/camunda/identity/sdk/authentication/UserDetails.java#L22), which Optimize does not make use of.
Operate on the other side, only makes use of username see [here](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/IdentityAuthentication.java#L225).

In UserDetails the username property is populated from the `preferred_username `claim and name from the `name` claim of the access token.

### Expected behaviour

Optimize should make use of `username` before falling back to the bare userId

### To Reproduce

1. Setup a [Self-Managed setup using a IDP](https://docs.camunda.io/docs/next/self-managed/setup/guides/connect-to-an-oidc-provider/) that is not adding the `name` claim to the access token, but the `preferred_username` claim
2. Login to Optimize and click on user profile and see something like this
![Image](https://github.com/user-attachments/assets/63999679-96e2-4aa9-80fa-3d01a0f9e37d)

### Environment observed:

- Optimize mode(s): CCSM
- Optimize version(s): all that support CCSM?
- Database(s) (Elasticsearch/OpenSearch): shouldn't matter

### Links

https://jira.camunda.com/browse/SUPPORT-24363

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/27571
- [ ] https://github.com/camunda/camunda/pull/27632
- [ ] https://github.com/camunda/camunda/pull/27633
```

## Bug Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

### QA Verification

Is this bug reproducible?

- [x] Yes. If so:
  - [x] The bug description is clear
  - [x] The steps to reproduce are clear
  - [x] The environments observed are correct and complete
- [ ] No

#### Engineering Review

[...]

PR #27571, opened 2025-02-03
fix: get name from token before falling back to username and id

abremard:
## Description

When a user click on "Open User", they will see the `name` claim populated in the side tab, which falls back to the `id` claim. In some cases, the `name` claim is empty but not the `username` claim. In such case, we expect the side tab to first fall back to the `username` value before falling back to the bare user id.

This PR adds that additional fallback logic in the middle and adds unit tests for the `io.camunda.optimize.service.security.CCSMTokenService::getUserInfoFromToken` method.

I ran the service locally to validate with a user missing the `name` claim by tweaking the `KEYCLOAK_USERS_0_***` values.

## Related issues

related to #27379

### DIFF
--- a/optimize/backend/src/main/java/io/camunda/optimize/service/security/CCSMTokenService.java
+++ b/optimize/backend/src/main/java/io/camunda/optimize/service/security/CCSMTokenService.java
@@ -208,7 +208,7 @@ public class CCSMTokenService {
         verifyToken(extractTokenFromAuthorizationValue(accessToken)).getUserDetails();
     return new UserDto(
         userId,
-        userDetails.getName().orElse(userId),
+        userDetails.getName().orElseGet(() -> userDetails.getUsername().orElse(userId)),
         userDetails.getEmail().orElse(userId),
         Collections.emptyList());
   }
--- a/optimize/backend/src/test/java/io/camunda/optimize/service/security/CCSMTokenServiceTest.java
+++ b/optimize/backend/src/test/java/io/camunda/optimize/service/security/CCSMTokenServiceTest.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.optimize.service.security;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+import com.google.common.collect.ImmutableList;
+import io.camunda.identity.sdk.Identity;
+import io.camunda.identity.sdk.authentication.AccessToken;
+import io.camunda.identity.sdk.authentication.Authentication;
+import io.camunda.identity.sdk.authentication.UserDetails;
+import io.camunda.optimize.dto.optimize.UserDto;
+import io.camunda.optimize.rest.exceptions.NotAuthorizedException;
+import io.camunda.optimize.service.util.configuration.ConfigurationService;
+import java.util.Optional;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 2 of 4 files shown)


optimize/backend/src/test/java/io/camunda/optimize/service/security/CCSMTokenServiceTest.java
```diff
@@ -0,0 +1,119 @@
...
+import io.camunda.optimize.service.util.configuration.ConfigurationService;
+import java.util.Optional;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class CCSMTokenServiceTest {
```
abremard, line 29: There were no tests for the current class so I wrote some tests for the method that I'm modifying. Let me know if I should write tests for the other methods as well.

abremard:
The PR is a little messy because I added a space between [type] and semicolon [:] so the `Commitlint` action doesn't validate it, let me know if I should close this one and open a new one

optimize/backend/src/test/java/io/camunda/optimize/service/security/CCSMTokenServiceTest.java
grlimacan: Thanks for writing the tests :-)

optimize/backend/src/test/java/io/camunda/optimize/service/security/CCSMTokenServiceTest.java
```diff
@@ -0,0 +1,119 @@
...
+  void getUserInfoFromTokenValidTokenReturnsUserDto() {
+    when(accessToken.getUserDetails()).thenReturn(userDetails);
+    when(userDetails.getName()).thenReturn(Optional.of(NAME));
+    when(userDetails.getEmail()).thenReturn(Optional.of(EMAIL));
+
+    final UserDto result = ccsmTokenService.getUserInfoFromToken(ID, ACCESS_TOKEN_VALUE);
+
+    assertEquals(ID, result.getId());
+    assertEquals(NAME, result.getFirstName());
+    assertEquals(EMAIL, result.getEmail());
```
grlimacan, line 67: For completeness in all tests, please also assert the values of the `lastName` and `roles`. I believe in all test cases below `lastName` should be null and the `roles` should be an empty list.

grlimacan (changes requested):
Thanks for the implementation, a change in `CCSMUserCache` is also necessary

grlimacan:
@abremard , since this is categorized as a bug, I believe we need to backport it to `stable/optimize-8.7` and also to `stable/optimize-8.6` re [this message](https://camunda.slack.com/archives/CUZUYDMMG/p1737629891608019?thread_ts=1737547239.966909&cid=CUZUYDMMG) from @RomanJRW . Pls ping if me you have never done a backport before

optimize/backend/src/main/java/io/camunda/optimize/service/security/CCSMTokenService.java
```diff
@@ -208,7 +208,7 @@ public UserDto getUserInfoFromToken(final String userId, final String accessToke
         verifyToken(extractTokenFromAuthorizationValue(accessToken)).getUserDetails();
     return new UserDto(
         userId,
-        userDetails.getName().orElse(userId),
+        userDetails.getName().orElseGet(() -> userDetails.getUsername().orElse(userId)),
```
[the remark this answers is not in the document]
abremard: After reading through the code, the CCSMUserCache method you've referenced is mapping a 
`io.camunda.identity.sdk.users.dto.User` to `io.camunda.optimize.dto.optimize.UserDto` 

But the CCSMTokenService method is mapping a 
`io.camunda.identity.sdk.authentication.UserDetails` to `io.camunda.optimize.dto.optimize.UserDto`

The difference being the `User` object don't have optional fields. I assume identity already sets the correct user values as opposed to `CCSMTokenService` which extracts bare values from the token and expects optional.empty values.

grlimacan (approved):
Pre-approving, just make sure the commit lint pipeline is passing before merging. Tip: run `mvn spotless:apply`

Issue #27286, opened 2025-01-22
Add update authorization processor/applier


nicpuppa:
## Processor:

1. Create new class `AuthorizationUpdateProcessor` and implement the `DistributedTypedRecordProcessor<AuthorizationRecord>` interface
2. Register the processor in the `EngineProcessor`
3. In the `processNewCommand method:
    - Check if the user is authorized to update Authorization
        - Resource type: `AUTHORIZATION`
        - Permission type: `UPDATE`
     - Check if the Authorization already exists. If no, reject the command with `NOT_FOUND`. If yes continue.
     - Write a `Authorization.UPDATED` event and respond to the client
     - Distribute the command to the other partitions on the `DistributionQueue.IDENTITY` queueId.
4. In the `processDistributedCommand` method:
    - Write a `Authorization.UPDATED` event
    - Acknowledge the distribution
5. Unit test the `Authorization.UPDATE` command
6. Unit test the `Authorization.UPDATE` command distribution.

## Applier:

1. Create the `AuthorizationUpdatedApplier`
2. Implement the `TypedEventApplier<AuthorizationIntent, AuthorizationRecord>` interface
4. In the `AuthorizationUpdatedApplier` call the update method from the state.

PR #27596, opened 2025-02-03
Partial authorization update is not supported

nicpuppa:
## Description

Based on this [discussion](https://camunda.slack.com/archives/C06UYJMMETZ/p1738580108815709) on slack we should support only full record updates, and NOT partial records update (PATCH)

## Related issues

related to https://github.com/camunda/camunda/issues/27286

### DIFF
--- a/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/AuthorizationRecord.java
+++ b/zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/AuthorizationRecord.java
@@ -23,7 +23,6 @@ import io.camunda.zeebe.util.buffer.BufferUtil;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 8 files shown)


zeebe/protocol-impl/src/main/java/io/camunda/zeebe/protocol/impl/record/value/authorization/AuthorizationRecord.java
```diff
@@ -63,8 +60,7 @@ public AuthorizationRecord() {
         .declareProperty(resourceIdProp)
         .declareProperty(resourceTypeProp)
         .declareProperty(permissionsProp)
-        .declareProperty(authorizationPermissionsProp)
-        .declareProperty(changedAttributesProp);
+        .declareProperty(authorizationPermissionsProp);
```
remcowesterhoud, line 63: ❌ You forgot to decrement the call to the super class.

remcowesterhoud (approved):
Thanks @nicpuppa! One small thing you need to change, but I'll pre-approve 👍

PR #27605, opened 2025-02-03
[Backport stable/8.5] fix: use new micrometer metric names

npepinpe:
## Description

Manually backports a fix based on https://github.com/camunda/camunda/pull/23197, but modified to use the union approach.

backports #22458


npepinpe:
I had to manually backport because merging Grafana dashboards is impossible :smile: 

I made sure to update the same metrics, but used the union approach instead of duplication.

entangled90 (approved):

PR #27608, opened 2025-02-03
fix: allow unauthenticated api access by default

emulbreh:
This should fix the failing tests in CI, caused by changing the default to have basic auth enabled done with https://github.com/camunda/camunda/pull/26960 previously.

Context see this thread https://camunda.slack.com/archives/C06UYJMMETZ/p1738600178706609?thread_ts=1738594463.864289&cid=C06UYJMMETZ


megglos (approved):
LGTM!

PR #27611, opened 2025-02-03
refactor: use batch document upload endpoint instead of single document upload

vsgoulart:
## Description

Before we were using the file upload endpoint to upload multiple documents but this was not a great solution because if one request failed it would fail the entire task completion. Now that we have the batch upload we can handle this better.

You can check the endpoint specs [here](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/create-documents/)

## Checklist

- [ ] for CI changes:

### DIFF
--- a/tasklist/client/src/modules/utils/buildDocumentMultipart.ts
+++ b/tasklist/client/src/modules/utils/buildDocumentMultipart.ts
@@ -1,0 +1,62 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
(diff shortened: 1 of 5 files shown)


vsgoulart:
I'll have to make some changes in this PR so I'll have to convert it to draft again

tasklist/client/src/modules/utils/buildDocumentMultipart.ts
vsgoulart, line 1: This is necessary because when using `FormData` we can't set a part header and we need the part header to associate a file to a Zeebe variable because the endpoint can't guarantee the order of the files returned on success

vsgoulart:
Should be ready for review again

pedesen (approved):
LGTM

PR #27615, opened 2025-02-04
Update authorizations

### LINKED ISSUE #27203: The `AuthorizationState` can update an authorization
- Delete the old authorization record and create the new one
- Remove old permission and re-add the new one
- Remove the old authorization key - permission key record from the cf and add the new one

nicpuppa:
## Description

This PR adds a method in the state to update an authorization record. In order to update also the permission related to an authorization record, the easiest way to go is by deleting the old authorization record and permission and re-create them with the new provided values

## Related issues

closes #27203

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
@@ -148,11 +148,6 @@ public class DbAuthorizationState implements MutableAuthorizationState {
     final var persistedAuthorization =
         authorizationByAuthorizationKeyColumnFamily.get(this.authorizationKey);
 
-    if (persistedAuthorization == null) {
-      // if there are no persisted authorization, we should not update the authorization
-      return;
-    }
-
     // remove the old permissions
     persistedAuthorization
         .getPermissionTypes()
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
@@ -144,25 +144,8 @@ public class DbAuthorizationState implements MutableAuthorizationState {
 
   @Override
   public void update(final long authorizationKey, final AuthorizationRecord authorization) {
-    this.authorizationKey.wrapLong(authorizationKey);
-    final var persistedAuthorization =
-        authorizationByAuthorizationKeyColumnFamily.get(this.authorizationKey);
-
-    // remove the old permissions
-    persistedAuthorization
-        .getPermissionTypes()
-        .forEach(
-            permissionType -> {
-              removePermission(
-                  persistedAuthorization.getOwnerType(),
-                  persistedAuthorization.getOwnerId(),
-                  persistedAuthorization.getResourceType(),
-                  permissionType,
-                  Set.of(persistedAuthorization.getResourceId()));
-            });
-
-    // delete the old authorization record
-    authorizationByAuthorizationKeyColumnFamily.deleteExisting(this.authorizationKey);
+    // delete old authorization record
+    delete(authorizationKey);
 
     // create the new authorization record
     create(authorizationKey, authorization);
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
@@ -143,6 +143,37 @@ public class DbAuthorizationState implements MutableAuthorizationState {
   }
 
   @Override
+  public void update(final long authorizationKey, final AuthorizationRecord authorization) {
+    this.authorizationKey.wrapLong(authorizationKey);
+    final var persistedAuthorization =
+        authorizationByAuthorizationKeyColumnFamily.get(this.authorizationKey);
+
+    if (persistedAuthorization == null) {
+      // if there are no persisted authorization, we should not update the authorization
+      return;
+    }
+
+    // remove the old permissions
+    persistedAuthorization
+        .getPermissionTypes()
+        .forEach(
+            permissionType -> {
+              removePermission(
+                  persistedAuthorization.getOwnerType(),
+                  persistedAuthorization.getOwnerId(),
+                  persistedAuthorization.getResourceType(),
+                  permissionType,
+                  Set.of(persistedAuthorization.getResourceId()));
+            });
+
+    // delete the old authorization record
+    authorizationByAuthorizationKeyColumnFamily.deleteExisting(this.authorizationKey);
+
+    // create the new authorization record
+    create(authorizationKey, authorization);
+  }
+
+  @Override
   public void delete(final long authorizationKey) {
     this.authorizationKey.wrapLong(authorizationKey);
     final var persistedAuthorization =
@@ -150,7 +181,7 @@ public class DbAuthorizationState implements MutableAuthorizationState {
 
     // remove the old permissions
     persistedAuthorization
-        .getPermissions()
+        .getPermissionTypes()
         .forEach(
             permissionType -> {
               removePermission(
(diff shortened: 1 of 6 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -162,6 +162,38 @@ public void deleteAuthorizationsByOwnerTypeAndIdPrefix(
...
+                  persistedAuthorization.getOwnerType(),
+                  persistedAuthorization.getOwnerId(),
+                  persistedAuthorization.getResourceType(),
+                  permissionType,
+                  Set.of(persistedAuthorization.getResourceId()));
+            });
+
+    // delete the old authorization record
+    authorizationByAuthorizationKeyColumnFamily.deleteExisting(this.authorizationKey);
+    permissionKeyByAuthorizationKeyColumnFamily.deleteExisting(this.authorizationKey);
```
nicpuppa, line 191: 💭 This CF might not be needed anymore, but I'll wait until also the delete method is implemented to properly evaluate this

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -162,6 +162,38 @@ public void deleteAuthorizationsByOwnerTypeAndIdPrefix(
...
+        .getPermissions()
+        .forEach(
+            permissionType -> {
+              removePermission(
+                  persistedAuthorization.getOwnerType(),
+                  persistedAuthorization.getOwnerId(),
+                  persistedAuthorization.getResourceType(),
+                  permissionType,
+                  Set.of(persistedAuthorization.getResourceId()));
+            });
```
remcowesterhoud, line 187: ❌ No need to do this since you're going to delete the entry from the CF anyways.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
```diff
@@ -162,6 +162,38 @@ public void deleteAuthorizationsByOwnerTypeAndIdPrefix(
...
+
+    // delete the old authorization record
+    authorizationByAuthorizationKeyColumnFamily.deleteExisting(this.authorizationKey);
+    permissionKeyByAuthorizationKeyColumnFamily.deleteExisting(this.authorizationKey);
+
+    // create the new authorization record
+    create(authorizationKey, authorization);
```
remcowesterhoud, line 194: ❓ Why delete and recreate instead of doing an update?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
remcowesterhoud: ❓ Do we still need the `permissionsColumnFamily`? Don't we have the permissions inside of the `PersistedAuthorization` as well?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/authorization/AuthorizationStateTest.java
```diff
@@ -274,4 +274,271 @@ void shouldRemoveAllPermissionsByOwnerTypeAndID() {
...
+    assertThat(anotherResourceIdentifiers2).isEmpty();
+
+    final var anotherResourceIdentifiers3 =
+        authorizationState.getResourceIdentifiers(
+            ownerType, ownerId, resourceType, PermissionType.DELETE);
+    assertThat(anotherResourceIdentifiers3).isEmpty();
+  }
+
+  @Test
+  void shouldUpdateAuthorizationChangingPermissionsKey() {
```
remcowesterhoud, line 343:
```suggestion
  void shouldUpdateAuthorizationChangingOwnerId() {
```

Unless I'm missing something you're updating the owner id. I'd consider not updating the permissions here to keep the test case limited to one scenario.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/authorization/AuthorizationStateTest.java
```diff
@@ -274,4 +274,271 @@ void shouldRemoveAllPermissionsByOwnerTypeAndID() {
...
+            ownerType, ownerId, resourceType, PermissionType.READ);
+    assertThat(anotherResourceIdentifiers3).isEmpty();
+    final var anotherResourceIdentifiers4 =
+        authorizationState.getResourceIdentifiers(
+            ownerType, ownerId, resourceType, PermissionType.DELETE);
+    assertThat(anotherResourceIdentifiers4).isEmpty();
+  }
+
+  @Test
+  void shouldUpdateAuthorizationChangingPermission() {
```
remcowesterhoud, line 409:
```suggestion
  void shouldUpdateAuthorizationChangingResourceType() {
```

I'd consider not updating the permissions here to keep the test case limited to one scenario.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/state/authorization/AuthorizationStateTest.java
```diff
@@ -274,4 +274,271 @@ void shouldRemoveAllPermissionsByOwnerTypeAndID() {
                 ownerType, ownerId, resourceType, permissionType))
         .isEmpty();
   }
+
+  @Test
+  void shouldUpdateAuthorization() {
```
remcowesterhoud, line 279:
```suggestion
  void shouldUpdateAuthorizationChangingPermissions() {
```

remcowesterhoud (changes requested):
Please check out my comments 👀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
nicpuppa: No, I need to delete the permissions related to this authorization record. I'm can't delete the entry from the PERMISSION CF as it might contains more permissions

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
nicpuppa: We have the permission related only to one resourceId, the PERMISSION cf store all the permission for a specific owner and resource type

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
nicpuppa: Because I think this is the easiest approach, the persisted authorization can be easily updated. However to handle correctly the permissions we need to first remove them from the related record and re-add the new ones.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
remcowesterhoud: Do we still need a CF which has all permissions? Isn't the individual ones enough?

remcowesterhoud: Imo it becomes very confusing the way it is now. Where we have authorization containing partial permissions and another CF containing all permissions.

nicpuppa: Which is the individual one ? We store the authorization record in the AUTHORIZATIONS CF and then we store the PERMISSIONS collected by ownerId + ownerType + resourceType.

nicpuppa: I can rename the property

remcowesterhoud: Thanks for explaining and renaming. I understand now 👍

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/authorization/DbAuthorizationState.java
remcowesterhoud: I see, so mainly to ensure the permissions are added. I can live with that.

remcowesterhoud (approved):
👍

PR #27641, opened 2025-02-04
fix: cannot assign zeebe user tasks in Tasklist running in the single app

### LINKED ISSUE #24808: Unable to Assign Zeebe User Task in C8Run 8.7.0-alpha1 & 8.6.5
**Describe the issue:**

https://jira.camunda.com/browse/SUPPORT-25953

When using C8Run with `8.7.0-alpha1` and `8.6.5` , I am unable to assign a Zeebe User Task. 

**Actual behavior:**

When attempting to assign a Zeebe User Task, the task does not get assigned as expected.

**Expected behavior:**

I would expect to be able to assign a Zeebe User Task without any issues.

**How to reproduce:**

1. Set up C8Run with version `8.7.0-alpha1` or `8.6.5` .
2. Deploy a diagram with a Zeebe User Task
3. Attempt to assign a Zeebe User Task in the UI.
4. Observe that the task is not assigned and an error appears in the UI (see screenshot)
Screenshot 2024-11-15 at 17 33 31

*Note:* The issue occurs across all operating systems.

**Logs:**

I have attached logs and a screenshot of the UI behavior for reference.
[camunda.log](https://github.com/user-attachments/files/17799747/camunda.log)

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- **Platform:** Mac, Windows, Linux
- **Architecture:** x86, ARM
- **Version:** 8.7.0-alpha1 and 8.6.5

houssain-barouni:
## Description

Fixes the assignment and completion of Zeebe user tasks in Tasklist when running in single app mode.

This PR backports some of the changes from https://github.com/camunda/camunda/pull/25799, replacing direct Zeebe client usage with broker requests. However, unlike the referenced PR, which disables the mode using the `camunda.tasklist.zeebe.compatibility.enabled` configuration property, this implementation conditions the change on the presence of the `standalone` Spring profile.

Since this fix will be included as a patch for 8.6, it is safer to ensure that the Zeebe client remains in use for production configurations while restricting broker requests to single app mode. This minimizes the risk of unforeseen issues that may arise from switching to broker requests.

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/24808


qa/integration-tests/src/test/java/io/camunda/it/tasklist/TasklistCreateProcessInstanceIT.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private void ensureProcessInstanceCreated(final String processDefinitionId) {
+    Awaitility.await(
+            "should have started process instance with id %s".formatted(processDefinitionId))
+        .atMost(Duration.ofSeconds(15))
+        .ignoreExceptions() // Ignore exceptions and continue retrying
+        .untilAsserted(
+            () -> {
+              final var result =
+                  operateRestClient
```
panagiotisgts, line 109: Could we maybe use another condition to remove the dependency of Operate client? Such as a User Task is present for said Process Instance or make use of the ZeebeClient?

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/service/TaskService.java
```diff
@@ -283,13 +247,11 @@ public TaskDTO unassignTask(final String taskId) {
...
-        zeebeClient.newUserTaskUnassignCommand(taskBefore.getKey()).send().join();
-      } catch (final ClientException exception) {
-        taskStore.persistTaskClaim(taskBefore, taskBefore.getAssignee());
-        throw new TasklistRuntimeException(exception.getMessage());
-      }
+    try {
+      tasklistServicesAdapter.unassignUserTask(taskEntity);
+    } catch (final Exception e) {
+      taskStore.persistTaskClaim(taskBefore, taskBefore.getAssignee());
+      throw e;
```
panagiotisgts, line 254: Should we wrap this exception into TasklistRuntimeException?

panagiotisgts (approved):
Nice work, also pretty nice clean up all over Tasklist.  Just some minor questions/suggestions.

Would it make sense/is it possible to follow the approach of the profile instead of the `camunda.tasklist.zeebe.compatibility.enabled` variable in 8.8 as well? It's one less env var to have to think about.

houssain-barouni:
> Nice work, also pretty nice clean up all over Tasklist. Just some minor questions/suggestions.
> 
> Would it make sense/is it possible to follow the approach of the profile instead of the `camunda.tasklist.zeebe.compatibility.enabled` variable in 8.8 as well? It's one less env var to have to think about.

In version 8.8, Zeebe engine will introduce authorization and permission checks, meaning that using the Zeebe client will require M2M credentials with the necessary permissions. However, this is not always the case. For example, in SaaS, we do not use M2M credentials when interacting with the Zeebe client. To address this, Tasklist needs to send broker requests to the engine while forwarding the caller’s context (with the appropriate permissions).

Additionally, this should better align with the architectural direction, where Tasklist acts as any gateway, forwarding requests to the engine via broker requests.

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/service/TaskService.java
houssain-barouni: if we do that all the exception will be returned as 500 to the caller, the exception can be a `ForbiddenActionException` which is mapped to 403 in https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/rest/InternalAPIErrorController.java

qa/integration-tests/src/test/java/io/camunda/it/tasklist/TasklistCreateProcessInstanceIT.java
houssain-barouni: The test is about process instances, so better to keep it this way. This is actually not a REST call to Operate API but a ES search query call. Using ZeebeClient search process instance would require creating a session (as all endpoints are secured by Operate filters in single app in 8.7), which is a bit overkill

PR #27654, opened 2025-02-04
Migrate ProcessingMetrics from stream platform to Micrometer 

entangled90:
## Description

This PR migrates the `ProcessingMetrics` class from the stream platform to Micrometer. To ease migration, it adds a new helper, `EnumMeter`, which is inspired by Prometheus' `Enumeration` meter.

> [!Note]
> An `Enumeration` is essentially a series of gauges with the same name, one for every value of a specific `enum`. The enum value is specified as a label/tag on the metric. Every gauge is always 0, except the one which is currently the active state, which will be set to 1.

The two histograms in there had their names changed (i.e. with the `_seconds` suffix), so I've updated the dashboard consequently to use the union of both series.

> Original PR was 27556, but it was merged to the wrong branch

## Related issues

related #26078

### DIFF
--- a/monitor/grafana/zeebe.json
+++ b/monitor/grafana/zeebe.json
@@ -9195,7 +9195,7 @@
                 "uid": "${DS_PROMETHEUS}"
               },
               "editorMode": "code",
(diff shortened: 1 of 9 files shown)


npepinpe (approved):

npepinpe:
@korthout - FYI, haven't seen these "Server Error" comments in the past :smile:

PR #27659, opened 2025-02-04
Update authorization endpoint

### LINKED ISSUE #26985: As a User I can update an authorization, changing one or multiple fields
Create a new endpoint PUT `/authorizations/{authorizationKey}`. 

The request body is:

```
{
  ownerId: string,
  ownerType: enum,
  resourceId: string,
  resourceType: enum,
  permissions: [enum]
}
``` 

The service layer to send the request to the broker need to be created, as well as the open api specification

nicpuppa:
## Description

This PR adds the authorization update endpoint.

## Related issues

closes #26985

### DIFF
--- a/service/src/main/java/io/camunda/service/AuthorizationServices.java
+++ b/service/src/main/java/io/camunda/service/AuthorizationServices.java
@@ -17,9 +17,10 @@ import io.camunda.security.auth.Authorization;
 import io.camunda.service.search.core.SearchQueryService;
 import io.camunda.service.security.SecurityContextProvider;
 import io.camunda.zeebe.broker.client.api.BrokerClient;
-import io.camunda.zeebe.gateway.impl.broker.request.BrokerAuthorizationCreateRequest;
+import io.camunda.zeebe.gateway.impl.broker.request.BrokerAuthorizationRequest;
 import io.camunda.zeebe.gateway.impl.broker.request.BrokerAuthorizationDeleteRequest;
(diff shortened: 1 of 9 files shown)


christinaausley (approved):

service/src/main/java/io/camunda/service/AuthorizationServices.java
```diff
@@ -132,10 +132,31 @@ public CompletableFuture<AuthorizationRecord> deleteAuthorization(final long aut
...
       AuthorizationResourceType resourceType,
       Set<PermissionType> permissions) {}
+
+  public record UpdateAuthorizationRequest(
+      long authorizationKey,
+      String ownerId,
+      AuthorizationOwnerType ownerType,
+      String resourceId,
+      AuthorizationResourceType resourceType,
+      Set<PermissionType> permissions) {}
```
remcowesterhoud, line 161: 🤔 I wonder if it's easier to just put the `CreateAuthorizationRequest` in here instead of copying all the properties. But the naming won't make sense that way so maybe it'll cause more confusion than good.

remcowesterhoud (approved):
Nice work 🚀

service/src/main/java/io/camunda/service/AuthorizationServices.java
nicpuppa: I'll keep it like this, more clear

PR #27674, opened 2025-02-04
Merge archiver and retention config

### LINKED ISSUE #26492: Merge archiver configuration with retention configuration
**Description**

Right now, you can configure retention via `ZEEBE_BROKER_EXPORTERS_CAMUNDA_ARGS_RETENTION_*`, and archiver properties (e.g. rollover interval) via `ZEEBE_BROKER_EXPORTERS_CAMUNDA_ARGS_ARCHIVER_*`. 

This is because we consider retention to be tied to ILM/data deletion, and archiving to just creating the rollover indices. I think that's fine in terms of implementation details, but from a user perspective, it's likely the same general concern.

Archiving is implementation details of how we handle data retention, so the archiver properties should be merged with the retention configuration. That would mean:

```java
public static final class RetentionConfiguration {
    private boolean enabled = false;
    private String minimumAge = "30d";
    private String policyName;
}
```

Becomes:

```java
public static final class RetentionConfiguration {
    private boolean enabled = false;
    private String minimumAge = "30d";
    private String policyName;

    private boolean rolloverEnabled = true;
    private String elsRolloverDateFormat = "date";
    private String rolloverInterval = "1d";
    private int rolloverBatchSize = 100;
    private String waitPeriodBeforeArchiving = "1h";
    private int delayBetweenRuns = 2000;
}
```

You could also make rollover a sub type of `RetentionConfiguration`, I have no strong opinion on this.

I suspect most users will only configure the retention configuration, and advanced users _may_ configure rollover, but likely only the interval and format. Everything else is advanced or ways for us to mitigate issues with archiving.

EuroLew:
## Description

Merge archiver and retention config

## Related issues

relates to #26492

### DIFF
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -991,13 +991,6 @@
         #     numberOfReplicas: 0
         #     variableSizeThreshold: 8191
         #
-        #   retention:
-        #     enabled: false
-        #     minimumAge: 30d
-        #     policyName: camunda-retention-policy
-        #
-        #   createSchema: true
-        #
         #   archiver:
         #     enabled: true
         #     elsRolloverDateFormat: "date"
@@ -1005,6 +998,13 @@
         #     rolloverBatchSize: 100
         #     waitPeriodBeforeArchiving: "1h"
         #     delayBetweenRuns: 2000
+        #     maxDelayBetweenRuns: 60000
+        #     retention:
+        #       enabled: false
+        #       minimumAge: 30d
+        #       policyName: camunda-retention-policy
+        #
+        #   createSchema: true
         #
         #   processCache:
         #     maxCacheSize: 10000
--- a/dist/src/main/config/broker.yaml.template
+++ b/dist/src/main/config/broker.yaml.template
@@ -124,7 +124,7 @@
           # password:
 
       # commandApi:
-        # Overrides the host used for gateway-to-broker communication
+      # Overrides the host used for gateway-to-broker communication
         # This setting can also be overridden using the environment variable ZEEBE_BROKER_NETWORK_COMMANDAPI_HOST.
         # host: 0.0.0.0
 
@@ -905,13 +905,6 @@
         #     numberOfReplicas: 0
         #     variableSizeThreshold: 8191
         #
-        #   retention:
-        #     enabled: false
-        #     minimumAge: 30d
-        #     policyName: camunda-retention-policy
-        #
-        #   createSchema: true
-        #
         #   archiver:
         #     enabled: true
         #     elsRolloverDateFormat: "date"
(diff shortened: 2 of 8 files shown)


npepinpe (commented):
:question: As a user, this means I now have to write `ZEEBE_BROKER_EXPORTERS_CAMUNDA_ARGS_RETENTION_ARCHIVER_BLAH`? :smile:

I was more thinking of merging them into a single class to be honest, but I suppose if we document the configuration properly then it shouldn't matter :+1: 

I don't think we even have an issue to document this - can you check if we do? If not, can you please create one? We should document these settings, what they do, how they work together for common use cases, etc.

dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -995,17 +995,16 @@
         #     enabled: false
         #     minimumAge: 30d
         #     policyName: camunda-retention-policy
+        #     archiver:
+        #       enabled: true
```
ChrisKujawa, line 999: ❓ How is the behavior now with the retention disabled? What is the expectation here? @npepinpe ?

As user I would expect based on the nesting that if retention is disabled archiver is also disabled? Does it even make sense to disable the archiver?

dist/src/main/config/broker.yaml.template
```diff
@@ -124,7 +124,7 @@
           # password:
 
       # commandApi:
-        # Overrides the host used for gateway-to-broker communication
+       # Overrides the host used for gateway-to-broker communication
```
ChrisKujawa, line 127:
```suggestion
        # Overrides the host used for gateway-to-broker communication
```

dist/src/main/config/broker.standalone.yaml.template
ChrisKujawa: @npepinpe can you answer this?

npepinpe: Good point. No, in that sense archiving should be enabled anyway to keep "live" indices fast.

So in that sense they are separate :thinking: But still, to configure retention properly, the archiver also needs to be updated. So I guess the dependency is the other way around? Archiver > Retention? :thinking:

ChrisKujawa: Ok to clarify and confirm:

 * We always want to run archiver - there shouldn't be an option to disable it?
 * Retention is related to ILM - which are based on historic indices - which are created by archiver
   * Thus, retention should be below archiver config
 * Retention can be disabled, that means no ILM for historic indices, which is valid

EuroLew: Moved retention to be the child of archiver in configuration

ChrisKujawa:
@EuroLew can you try to rebase the branch, there some fixes that might help in fixing the OpenSearch IT

ChrisKujawa (approved):
Thanks 🙇🏼

PR #27684, opened 2025-02-04
feat: allocate a partition MeterRegistry in RaftPartition

entangled90:
## Description

A MeterRegistry with tag "partition" is allocated in RaftPartition constructor.
The registry is then passed everywhere and closed when RaftPartition.stop() method is called.

## Related issues
relates #26078

### DIFF
--- a/zeebe/atomix/cluster/src/main/java/io/atomix/raft/partition/RaftPartition.java
+++ b/zeebe/atomix/cluster/src/main/java/io/atomix/raft/partition/RaftPartition.java
@@ -32,6 +32,10 @@ import io.camunda.zeebe.util.VisibleForTesting;
 import io.camunda.zeebe.util.health.FailureListener;
 import io.camunda.zeebe.util.health.HealthMonitorable;
 import io.camunda.zeebe.util.health.HealthReport;
+import io.camunda.zeebe.util.micrometer.MicrometerUtil.PartitionKeyNames;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 import java.io.File;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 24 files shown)


zeebe/atomix/cluster/src/main/java/io/atomix/raft/partition/RaftPartition.java
```diff
@@ -273,10 +281,22 @@ public boolean shouldStepDown() {
   }
 
   public CompletableFuture<Void> stop() {
-    return server.stop();
+    // close the registry regardless of errors in server.stop()
+    return server.stop().whenComplete((unused, error) -> meterRegistry.close());
```
npepinpe, line 285: :x: Don't we need to also clear it? I don't think it's sufficient to close it.

npepinpe (changes requested):

zeebe/atomix/cluster/src/main/java/io/atomix/raft/partition/RaftPartition.java
entangled90: Yes you are correct, clearing removes all metrics, while closing it closes all outstanding timers (and some background tasks)

npepinpe (approved):

zeebe/atomix/cluster/src/main/java/io/atomix/raft/partition/RaftPartition.java
```diff
@@ -273,10 +281,28 @@ public boolean shouldStepDown() {
   }
 
   public CompletableFuture<Void> stop() {
-    return server.stop();
+    // close the registry regardless of errors in server.stop()
+    return server
+        .stop()
+        .whenComplete(
+            (unused, error) -> {
+              meterRegistry.clear();
```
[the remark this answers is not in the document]
entangled90: - `clear` just remove all elements from a map, it may fail only if it allocates and raises an OOM.
- yeah sure, why not!

PR #27686, opened 2025-02-04
Pass and store meter registry down to ProcessEngineMetrics

npepinpe:
## Description

This PR sets up migrating the engine metrics to Micrometer by passing the meter registry all the way down to the process engine metrics, and ensuring we only create one instance of the `ProcessEngineMetrics` (instead of two).

It's not used yet, but will be in upcoming PRs.

## Related issues

related to #26078

### DIFF
--- a/zeebe/backup/pom.xml
+++ b/zeebe/backup/pom.xml
@@ -71,6 +71,12 @@
     </dependency>
 
     <dependency>
+      <groupId>io.micrometer</groupId>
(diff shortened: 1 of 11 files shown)


entangled90 (approved):
:+1: so we are using the MeterRegistry from the StreamProcessorContext

npepinpe:
Yeah, I didn't see a need to create an engine specific meter registry, since its lifecycle is tied to the stream processor.

PR #27696, opened 2025-02-05
fix: remove business key from customer onboarding dataset

abremard:
## Description

This PR adds a filter to exclude "business key" field from process instance raw data. This means exporting to csv will result in that field being excluded, however when exporting to json, the fields will be present will null values (as explained [here](https://github.com/camunda/camunda-optimize/issues/14172#issuecomment-2633100401)). The unit tests have been updated to reflect the changes described in the issue.

## Related issues

related to https://github.com/camunda/camunda-optimize/issues/14172


grlimacan (approved):
Congratulations! Elegant, simple and clean solution, I have nothing to complain about :-)

PR #27718, opened 2025-02-05
Bugfix/27717 gateway fix loading of xml for decision requirements

### LINKED ISSUE #27717: Fix loading of xml for decision requirements
Currently there is different behaviour for loading the xml of decision requirements in ES and RDBMS. 

Basically the `DecisionRequirementsQueryResultConfig.includeXml` uses "false" as default if the xml should be loaded. But since the DecisionRequirementsQueryResultConfig is sometimes null, the default is not always applied. 

* The RDBMS module initializes an empty config if it's null, so the default of DecisionRequirementsQueryResultConfig.includeXml is used correctly. 
* In ES the empty config get's never initialized, and the transformers than simply ignores the config. 
https://github.com/camunda/camunda/blob/ff26d9d5fb15a0de004dabe7cdfd9f038a7b0e32/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/query/TypedSearchQueryTransformer.java#L69-L72
Since the xml needs to be actively be excluded, per default it's than always queried. 
https://github.com/camunda/camunda/blob/ff26d9d5fb15a0de004dabe7cdfd9f038a7b0e32/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/result/DecisionRequirementsResultConfigTransformer.java#L23-L25

**AC:**

* Per default, xml should not be loaded for decision requirements
* If xml is needed, it should be actively activated

p-wunderlich:
## Description

* Use default for resultConfig in DecisionRequirementsQuery
* Actively enable loading of xml for decision requirements where needed
* Fixed typo in DecisionRequirementsController

Details can be found in #27717 

## Related issues
closes #27717

### DIFF
--- a/search/search-domain/src/main/java/io/camunda/search/query/DecisionRequirementsQuery.java
+++ b/search/search-domain/src/main/java/io/camunda/search/query/DecisionRequirementsQuery.java
@@ -41,6 +41,8 @@ public record DecisionRequirementsQuery(
         FilterBuilders.decisionRequirements().build();
     private static final DecisionRequirementsSort EMPTY_SORT =
         SortOptionBuilders.decisionRequirements().build();
+    private static final DecisionRequirementsQueryResultConfig EMPTY_RESULT_CONFIG =
+        DecisionRequirementsQueryResultConfig.of(b -> b);
 
     private DecisionRequirementsFilter filter;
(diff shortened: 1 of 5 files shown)


houssain-barouni (approved):
LGTM

PR #27720, opened 2025-02-05
Refactor RDBMS and Camunda Exporters to new Authorization structure

### LINKED ISSUE #27579: Create `AuthorizationCreatedUpdatedHandler`
## Description

- Rename `AuthorizationHandler` to `AuthorizationCreatedUpdatedHandler`
- Add a unit test
- Query authorization by owner id, resource type and permission type in the `AuthorizationChecker`.

-----

Depends on #26983. We need to be able to get authorizations by owner id so we can check for authorized permissions in the rest gateway.

### LINKED ISSUE #27664: Convert from owner keys to owner ids
## Description

- Update AuthorizationExportHandler with the new field(s)
- Update AuthorizationsMapper with the new field(s)
- Update AuthorizationWriter to use the key instead of the current composite key
- Update AuthorizationReader to map the new field(s)

remcowesterhoud:
## Description

There's no way easy way split this PR, so as a it's a bit larger than I'd like. I tried to make clear commits.

1. Some refactorings had to be made to Authorization entity. This is used in the search API and both exporters
2. The RDBMS Exporter was changed so it stores the data in the new structure. In the end the structure didn't really change except for an extra column (authorization key) and a rename (owner key to owner id). Some of the mapping had to be changed because of the changes in the Authorization entity
3. The handler of the RDBMS exporter needed to change. The old intents will be removed soon. Instead it needs to act on CREATE/UPDATE/DELETE.
4. The Camunda Exporter had to be slightly changd to listen to Authorization CREATE/UPDATE events. I also had to change the exported properties and modify the index template accordingly.

## Checklist

- [ ] for CI changes:

## Related issues

closes #27664 
closes #27579


qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/fixtures/AuthorizationFixtures.java
```diff
@@ -24,15 +24,21 @@ private AuthorizationFixtures() {}
   public static AuthorizationDbModel createRandomized(
       final Function<Builder, Builder> builderFunction) {
     final var ownerId = nextStringId();
+
+    final var randomPermissionType1 = randomEnum(PermissionType.class);
+    var randomPermissionType2 = randomEnum(PermissionType.class);
+    while (randomPermissionType2 == randomPermissionType1) {
+      randomPermissionType2 = randomEnum(PermissionType.class);
+    }
```
nicpuppa, line 32: 🔧 Just to make this more fancy and a bit optimized

```
final var randomPermissionType1 = randomEnum(PermissionType.class);
PermissionType randomPermissionType2;

do {
    randomPermissionType2 = randomEnum(PermissionType.class);
} while (randomPermissionType2 == randomPermissionType1);
```

nicpuppa (commented):
@remcowesterhoud looks good so far, I left some small questions.

There are some failures in the CI

qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/fixtures/AuthorizationFixtures.java
remcowesterhoud: Why is this more optimized?

db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
```diff
@@ -411,15 +411,16 @@
   <changeSet id="create_authorization_permission_table" author="cthiel">
     <!-- no PK possible, would have too many columns (ALL) -->
     <createTable tableName="${prefix}AUTHORIZATIONS">
-      <column name="OWNER_KEY" type="BIGINT" />
+      <column name="AUTHORIZATION_KEY" type="BIGINT" />
```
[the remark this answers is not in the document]
remcowesterhoud: No, because it'll violate the uniqueness constraint. For each permission type a new row gets added to the table. But these are all part of the same authorization key. So there'd be dupes.

db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
```diff
@@ -411,15 +411,16 @@
...
       <column name="OWNER_TYPE" type="VARCHAR(255)" />
       <column name="RESOURCE_TYPE" type="VARCHAR(255)" />
-      <column name="PERMISSION_TYPE" type="VARCHAR(255)" />
       <column name="RESOURCE_ID" type="VARCHAR(255)" />
+      <column name="PERMISSION_TYPE" type="VARCHAR(255)" />
     </createTable>
 
     <createIndex tableName="${prefix}AUTHORIZATIONS" indexName="${prefix}IDX_AUTHORIZATIONS">
-      <column name="OWNER_KEY"/>
+      <column name="OWNER_ID"/>
```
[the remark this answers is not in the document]
remcowesterhoud: Yes, you can create indexes on multiple columns.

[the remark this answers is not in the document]
nicpuppa: Thanks 👍

qa/integration-tests/src/test/java/io/camunda/it/rdbms/db/fixtures/AuthorizationFixtures.java
remcowesterhoud: If the first assignment is equal we will do 2 assignments. It doesn't matter which structure we use as far as I can see? I'm happy to change it but I don't see a different 😄 Maybe it's just too early for me 😛

remcowesterhoud:
@nicpuppa I processed you review and hopefully resolved the tests 🤞 Please have a look at the new commits.

db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
```diff
@@ -411,15 +411,16 @@
...
+      <column name="AUTHORIZATION_KEY" type="BIGINT" />
+      <column name="OWNER_ID" type="VARCHAR(255)" />
       <column name="OWNER_TYPE" type="VARCHAR(255)" />
       <column name="RESOURCE_TYPE" type="VARCHAR(255)" />
-      <column name="PERMISSION_TYPE" type="VARCHAR(255)" />
       <column name="RESOURCE_ID" type="VARCHAR(255)" />
+      <column name="PERMISSION_TYPE" type="VARCHAR(255)" />
     </createTable>
 
     <createIndex tableName="${prefix}AUTHORIZATIONS" indexName="${prefix}IDX_AUTHORIZATIONS">
```
stoerti, line 422: ❌ The primary JOIN used in AuthorizationsMapper now uses the Authorization_key. We should add an authorizationKey index for this JOIN. (Adding it to the existing index may be too big).

💭 not sure if we even need the existing index anymore when we have an authorizationKey now.

search/search-domain/src/main/java/io/camunda/search/entities/AuthorizationEntity.java
```diff
@@ -8,9 +8,14 @@
 package io.camunda.search.entities;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import io.camunda.security.entity.Permission;
-import java.util.List;
+import io.camunda.zeebe.protocol.record.value.PermissionType;
```
stoerti, line 11: 💭 I know this is not introduced in this PR, but it feels odd for me to use an internal protocol enum in the API. It should be rather an enum of the API. Would it be a big effort to change this and introduce a mapped enum in the API?

search/search-domain/src/main/java/io/camunda/search/entities/AuthorizationEntity.java
```diff
@@ -8,9 +8,14 @@
 package io.camunda.search.entities;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import io.camunda.security.entity.Permission;
```
stoerti, line 11: 🔧 Is the Permission class still used in the code or can it be deleted?

stoerti (changes requested):

search/search-domain/src/main/java/io/camunda/search/entities/AuthorizationEntity.java
remcowesterhoud: This can go, good catch 👍

search/search-domain/src/main/java/io/camunda/search/entities/AuthorizationEntity.java
remcowesterhoud: I get what you're saying but I also feel this PR is already growing too large. I'd rather have a separate PR for this.

db/rdbms/src/main/resources/mapper/AuthorizationsMapper.xml
```diff
@@ -40,7 +42,7 @@
     <include refid="io.camunda.db.rdbms.sql.Commons.orderBy"/>
     <include refid="io.camunda.db.rdbms.sql.Commons.paging"/>
     ) t
-    JOIN ${prefix}AUTHORIZATIONS a ON t.OWNER_KEY = a.OWNER_KEY AND t.OWNER_TYPE = a.OWNER_TYPE AND
+    JOIN ${prefix}AUTHORIZATIONS a ON t.AUTHORIZATION_KEY = a.AUTHORIZATION_KEY AND t.OWNER_ID = a.OWNER_ID AND t.OWNER_TYPE = a.OWNER_TYPE AND
```
stoerti, line 45: ❓ is `resourceId` missing here?

db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
stoerti: I would create a separate index for authorizationKey since this probably the most accurate index for the JOIN in the SQL.

remcowesterhoud:
Thanks @stoerti, please check out my last 2 commits 👍

db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
stoerti: The indexes both have the same name, this will not work.

stoerti (approved):
The Indexes have the same name which will result in an error in the tests.

Aside from that LGTM (I have mostly looked at the RDBMS part)

nicpuppa (approved):

PR #27728, opened 2025-02-05
migrate LeaderMetrics to micrometer

entangled90:
## Description
Migrate LeaderMetrics to micrometer.

## Checklist
- [x] Grafana dashboard checks
  
## Related issues

relates #26708

### DIFF
--- a/monitor/grafana/zeebe.json
+++ b/monitor/grafana/zeebe.json
@@ -11348,7 +11348,7 @@
               "datasource": {
                 "uid": "$DS_PROMETHEUS"
               },
-              "expr": "sum(increase(atomix_append_entries_latency_bucket{cluster=~\"$cluster\", namespace=~\"$namespace\", pod=~\"$pod\", partition=~\"$partition\"}[$__rate_interval])) by (le)",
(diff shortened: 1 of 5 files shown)


entangled90:
Forgot to add  `KeyNames` to the metrics, will do in a new commit

monitor/grafana/zeebe.json
```diff
@@ -10979,7 +10979,7 @@
                 "uid": "${DS_PROMETHEUS}"
               },
               "editorMode": "code",
-              "expr": "sum(rate(atomix_append_entries_data_rate_total{cluster=~\"$cluster\", namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (partition, follower)",
+              "expr": "sum(rate(atomix_append_entries_data_rate_total{cluster=~\"$cluster\", namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval]) or rate(atomix_append_entries_data_rate_seconds_total{cluster=~\"$cluster\", namespace=~\"$namespace\", partition=~\"$partition\"}[$__rate_interval])) by (partition, follower)",
```
npepinpe, line 10982: :question: This is a counter, right? Why are we adding `_seconds` here? Is it because it's the base unit?

npepinpe: Same for the one below.

npepinpe: Seems like we don't specify units. So this seems wrong to me :x:

npepinpe (approved):
:rocket: 

Only one thing about the `_seconds` suffix to a counter? Doesn't look right to me.

Pre-approving because it's not something we need to re-review, but do double check it.

monitor/grafana/zeebe.json
entangled90: Removed it, did one too many :-D

PR #27733, opened 2025-02-05
feat: implement get all users for tenant

### LINKED ISSUE #27217: Get all users assigned to a tenant (as a search endpoint to be able to use pagination)

Ben-Sheppard:
## Description

This PR introduces the endpoint that allows retrieving the users for a given tenant and searching within those users. There is support for both the ES/OS side and the RDBMS side (this is also reflected in the choices on where/how to handle the different access patterns of the data (two queries vs JOIN`).

## Related issues

closes #27217

### DIFF
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -7,7 +7,7 @@
  */
 package io.camunda.search.clients;
 
-import static io.camunda.search.entities.TenantMemberEntity.TenantMemberType.USER;
+import static io.camunda.zeebe.protocol.record.value.EntityType.USER;
 
 import io.camunda.search.clients.auth.DocumentAuthorizationQueryStrategy;
 import io.camunda.search.clients.transformers.ServiceTransformers;
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -7,7 +7,7 @@
  */
 package io.camunda.search.clients;
 
-import static io.camunda.search.entities.TenantMemberEntity.MemberType.USER;
+import static io.camunda.search.entities.TenantMemberEntity.TenantMemberType.USER;
 
 import io.camunda.search.clients.auth.DocumentAuthorizationQueryStrategy;
 import io.camunda.search.clients.transformers.ServiceTransformers;
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -24,6 +24,7 @@ import io.camunda.search.entities.ProcessDefinitionEntity;
 import io.camunda.search.entities.ProcessInstanceEntity;
 import io.camunda.search.entities.RoleEntity;
 import io.camunda.search.entities.TenantEntity;
+import io.camunda.search.entities.TenantMemberEntity;
 import io.camunda.search.entities.UsageMetricsEntity;
 import io.camunda.search.entities.UserEntity;
 import io.camunda.search.entities.UserTaskEntity;
@@ -51,6 +52,8 @@ import io.camunda.security.auth.SecurityContext;
 import io.camunda.webapps.schema.descriptors.IndexDescriptors;
 import io.camunda.zeebe.util.CloseableSilently;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public class SearchClients
     implements AuthorizationSearchClient,
@@ -249,7 +252,7 @@ public class SearchClients
     final var overlayQuery = injectUsernamesIntoQuery(usernames, userQuery);
 
     return getSearchExecutor()
-        .search(filter, io.camunda.webapps.schema.entities.usermanagement.UserEntity.class);
+        .search(overlayQuery, io.camunda.webapps.schema.entities.usermanagement.UserEntity.class);
   }
 
   @Override
@@ -310,5 +313,21 @@ public class SearchClients
                 securityContext)
             .findAll(filter, io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
     return metrics.stream().map(UsageMetricsEntity::value).distinct().count();
+  }
+
+  private UserQuery injectUsernamesIntoQuery(
+      final Set<String> usernames, final UserQuery userQuery) {
+    return UserQuery.of(
+        builder ->
+            builder
+                .filter(
+                    f ->
+                        f.name(userQuery.filter().name())
+                            .email(userQuery.filter().email())
+                            .key(userQuery.filter().key())
+                            .username(userQuery.filter().username())
+                            .usernames(usernames))
+                .sort(userQuery.sort())
+                .page(userQuery.page()));
   }
 }
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
@@ -7,6 +7,8 @@
  */
 package io.camunda.search.clients;
 
+import static io.camunda.search.entities.TenantMemberEntity.MemberType.USER;
+
 import io.camunda.search.clients.auth.DocumentAuthorizationQueryStrategy;
 import io.camunda.search.clients.transformers.ServiceTransformers;
 import io.camunda.search.entities.AuthorizationEntity;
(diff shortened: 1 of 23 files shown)


search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
```diff
@@ -288,4 +314,20 @@ private Long distinctCountUsageMetricsFor(final String event, final UsageMetrics
             .findAll(filter, io.camunda.webapps.schema.entities.operate.UsageMetricsEntity.class);
     return metrics.stream().map(UsageMetricsEntity::value).distinct().count();
   }
+
+  private UserQuery injectUsernamesIntoQuery(
+      final Set<String> usernames, final UserQuery userQuery) {
+    return UserQuery.of(
```
emulbreh, line 320: This can be a bit more robust with a `toBuilder()...build()` pattern. I've tried that for roles in https://github.com/camunda/camunda/pull/25555/files#diff-c7ee74abafdf2e69d0afefb6b2c26ec904c4b8b7534592b6070f550c8524ffeeR312.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
```diff
@@ -224,9 +229,30 @@ public List<GroupEntity> findAllGroups(final GroupQuery query) {
...
+                    .filter(f -> f.joinParentId(userQuery.filter().tenantId()).memberType(USER))
+                    .build(),
+                io.camunda.webapps.schema.entities.usermanagement.TenantMemberEntity.class);
+    final var usernames =
+        tenantMembers.stream().map(TenantMemberEntity::id).collect(Collectors.toSet());
+
+    final var overlayQuery = injectUsernamesIntoQuery(usernames, userQuery);
+
+    return getSearchExecutor()
+        .search(overlayQuery, io.camunda.webapps.schema.entities.usermanagement.UserEntity.class);
```
emulbreh, line 255:
```suggestion
    return searchUsers(overlayQuery);
```
This makes it a bit easier to add other special cases, e.g. filtering users by groups or roles.

search/search-domain/src/main/java/io/camunda/search/filter/TenantFilter.java
```diff
@@ -7,11 +7,18 @@
...
 import java.util.function.Function;
 
-public record TenantFilter(Long key, String tenantId, String name, Set<Long> memberKeys)
+public record TenantFilter(
+    Long key,
+    String tenantId,
+    String name,
+    String joinParentId,
+    TenantMemberType tenantMemberType,
+    Set<Long> memberKeys)
```
emulbreh, line 21:
```suggestion
    String joinParentId,
    TenantMemberType tenantMemberType,
    Set<Long> memberKeys)
```
We could extract those into a `TenantMemberFilter`, which is nicer because it separates the two use-cases of the tenant filter. On the other hand, it's another class and the type union pattern is used in many places already.

emulbreh: Oh, nevermind, they can actually be used as a combined filter.

search/search-domain/src/main/java/io/camunda/search/entities/TenantMemberEntity.java
```diff
@@ -0,0 +1,20 @@
...
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.search.entities;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record TenantMemberEntity(String id, TenantMemberType tenantMemberType) {
+  public enum TenantMemberType {
```
emulbreh, line 14: Is this the same as `io.camunda.zeebe.protocol.record.value.EntityType`?

Ben-Sheppard: Yea, I couldn't find a nice place to lift `EntityType` and wanted to keep the PR smaller (without introducing a shared module other than the shared security module etc). 

I can look at this but at the moment it seems like there isn't a nice (non security scoped) method of sharing these common enums between both the engine and the search classes

Ben-Sheppard: Ah, I've just seen that the search query transformers module has a dependency to the zeebe protocol module so I'm just going to add that in here for the search domain module

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/SearchClients.java
Ben-Sheppard: Good suggestion, changed now

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/TenantFilterTransformer.java
```diff
@@ -31,16 +29,16 @@ public TenantFilterTransformer(final IndexDescriptor indexDescriptor) {
...
-                ? matchNone()
-                : hasChildQuery(
-                    IdentityJoinRelationshipType.MEMBER.getType(),
-                    longTerms(TenantIndex.MEMBER_KEY, filter.memberKeys())));
+            : term(TenantIndex.MEMBER_TYPE, filter.tenantMemberType().name()),
+        filter.joinParentId() == null
+            ? term(TenantIndex.JOIN, IdentityJoinRelationshipType.TENANT.getType())
+            : hasParentQuery(
+                IdentityJoinRelationshipType.TENANT.getType(),
+                term(TENANT_ID, filter.joinParentId())));
```
emulbreh, line 42: It's not from this PR, but this always feels off to me: there are two separate entities (Tenant and TenantMember) that are squashed into one object. As a result, some of the filter configurations don't make sense (e.g. you can't combine a name and a parent id filter). Ideally, I'd split the entities completely and just let them share the index. Filters could be very clear then:
```java
record TenantFilter(String name, …) {…}
record TenantMemberFilter(TenantFilter parentFilter, EntityType memberType) {…}
```
This is really out of scope here though and should be a separate refactoring.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/TenantFilterTransformer.java
```diff
@@ -31,16 +29,16 @@ public TenantFilterTransformer(final IndexDescriptor indexDescriptor) {
   @Override
   public SearchQuery toSearchQuery(final TenantFilter filter) {
     return and(
-        term(TenantIndex.JOIN, IdentityJoinRelationshipType.TENANT.getType()),
         filter.key() == null ? null : term(KEY, filter.key()),
         filter.tenantId() == null ? null : term(TENANT_ID, filter.tenantId()),
         filter.name() == null ? null : term(NAME, filter.name()),
-        filter.memberKeys() == null
```
emulbreh, line 38: `memberKeys` is still on TenantFilter. Do we drop it here intentionally?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/tenant/TenantController.java
```diff
@@ -167,6 +183,35 @@ private ResponseEntity<TenantSearchQueryResponse> search(final TenantQuery query
...
+            builder
+                .filter(
+                    f ->
+                        f.name(userQuery.filter().name())
+                            .email(userQuery.filter().email())
+                            .key(userQuery.filter().key())
+                            .username(userQuery.filter().username())
+                            .tenantId(tenantId))
+                .sort(userQuery.sort())
+                .page(userQuery.page()));
```
emulbreh, line 212: We can use the toBuilder() pattern here now. Maybe we should also assert that the userQuery doesn't contain a tenant id filter already.

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/TenantFilterTransformer.java
Ben-Sheppard: I dropped it intentionally as we were shifting through the resources anyway but good point, I'll reinstate it for now

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/tenant/TenantController.java
Ben-Sheppard: I wondered about the tenantId on the filter object also but in this case I wanted it to be very intentional that the tenant ID you pass in via the path is the one that we will use moving down the stack (we can also document this in the rest spec). We could add in validation as you say but I wasn't completely convinced in the benefit of blocking the request based on that

search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/filter/TenantFilterTransformer.java
Ben-Sheppard: I agree here in both point, firstly that it would be a cleaner approach and clearer but secondly that its not for this PR :)

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/tenant/TenantController.java
emulbreh: It points to a problem somewhere downstream and it'd be best for the client if they'd get a 400. We have the same problem with the member-only filters though and maybe this isn't the right layer for this validation.

emulbreh (approved):

Issue #26711, opened 2025-01-09
Reenable adding report to dashboard Optimize e2e


RomanJRW:
### Context

With [this PR](https://github.com/camunda/camunda/pull/26710), we deleted the flaky dashboard e2e. We should investigate the flakiness and readd this test to our running e2e suite.

The underlying behaviour still works (verified [here](https://camunda.slack.com/archives/C04FM0DRJT0/p1736423821109049?thread_ts=1736347237.691859&cid=C04FM0DRJT0))

### Links

- https://github.com/camunda/camunda/pull/26710
- [Context](https://camunda.slack.com/archives/C04FM0DRJT0/p1736423821109049?thread_ts=1736347237.691859&cid=C04FM0DRJT0)

### Breakdown

```[tasklist]
### Pull Requests
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [ ] The change is implemented as described on all target environments/versions/modes
- [ ] The documentation changes are as expected

### Completion

- [ ] All Review stages are successfully completed
- [ ] All associated PRs are merged to the main branch(es) and stable branches
- [ ] The correct version labels are applied to the issue

RomanJRW:
_something_ broke this e2e in the last month (perhaps an automated dependency update). If you feel brave enough to take a look at reenabling it @abremard , this could also be a nice starter issue for exposure to e2es

abremard:
> If you feel brave enough to take a look at reenabling it @abremard , this could also be a nice starter issue for exposure to e2es

I haven't looked into e2es yet so this is great timing. I've mainly worked with `Selenium` before so it's a good opportunity to try out `testcafe` 🧪

PR #27739, opened 2025-02-06
test: re-enable adding report to dashboard Optimize e2e

abremard:
## Description

After investigation of the `add a report from the dashboard` e2e test, it turns out the flakiness is mainly happening at 
```
await t.click(Common.modalConfirmButton).hover(Common.addButton).click('.DashboardRenderer');
```

In order to counter this, I've added `await t.expect(Common.templateModalProcessField.visible).ok();` before we interact with `Common.templateModalProcessField` so that we ensure that the element is visible before interacting with it. This is quite common in async e2e testing unfortunately because of rendering time or network delays.

- Before : The test would fail 50% of the time (on about 10 runs)
- Now : 10 consecutive successful runs

## Related issues

related to #26711

### DIFF
--- a/optimize/client/e2e/sm-tests/Dashboard.js
+++ b/optimize/client/e2e/sm-tests/Dashboard.js
@@ -463,6 +463,43 @@ test('version selection', async (t) => {
   await t.expect(Common.option('Test alert').exists).notOk();
 });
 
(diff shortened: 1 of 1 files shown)


grlimacan (approved):
Amazing stuff! Thanks for the investigation!!

PR #27742, opened 2025-02-06
fix: add rest api fixed wildcard support

### LINKED ISSUE #26924: [C8 API] Like search fixed wildcard support and search client mapping
## Description

API like search should have fixed defined wildcards and not depending on the underlying database types. This way users don't need to know about the data layer.

## Acceptance criteria

- Define wildcard(s) for like search.
  - TBD:
  - Separate wildcard for single and multiple characters?
  - What are the actual characters?
- Search clients map API wildcards to actual supported wildcards by their db.
- Tests.

### Links 
- SQL wildcards: https://www.w3schools.com/sql/sql_like.asp
   - The percent sign `%` represents zero, one, or multiple characters
   - The underscore sign `_` represents one, single character
- ES wildcards: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html
   - `?`, which matches any single character
   - `*`, which can match zero or more characters, including an empty one
   - OS: https://opensearch.org/docs/latest/query-dsl/term/wildcard/ (same)

```[tasklist]
### Tasks
```

danielkelemen:
## Description

- Add support for fixed ES style wildcards in REST API
- Escape sql wildcards

## Related issues

closes https://github.com/camunda/camunda/issues/26924

### DIFF
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/sql/typehandler/WildcardTransformingStringTypeHandler.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/sql/typehandler/WildcardTransformingStringTypeHandler.java
@@ -15,36 +15,55 @@ import org.apache.ibatis.type.StringTypeHandler;
 
 public class WildcardTransformingStringTypeHandler extends StringTypeHandler {
 
+  // Regular expression to match unescaped wildcards
+  private static final String REGEX_ASTERISK = "(?<!\\\\)\\*";
+  private static final String REGEX_QUESTION_MARK = "(?<!\\\\)\\?";
+  private static final String REGEX_ESCAPED_ASTERISK = "\\\\\\*";
+  private static final String REGEX_ESCAPED_QUESTION_MARK = "\\\\\\?";
+  private static final String REGEX_PERCENT = "(?<!\\\\)%";
+  private static final String REGEX_UNDERSCORE = "(?<!\\\\)_";
+  private static final Pattern PATTERN_ASTERISK = Pattern.compile(REGEX_ASTERISK);
+  private static final Pattern PATTERN_QUESTION_MARK = Pattern.compile(REGEX_QUESTION_MARK);
+  private static final Pattern PATTERN_ESCAPED_ASTERISK = Pattern.compile(REGEX_ESCAPED_ASTERISK);
+  private static final Pattern PATTERN_ESCAPED_QUESTION_MARK =
+      Pattern.compile(REGEX_ESCAPED_QUESTION_MARK);
+  private static final Pattern PATTERN_PERCENT = Pattern.compile(REGEX_PERCENT);
+  private static final Pattern PATTERN_UNDERSCORE = Pattern.compile(REGEX_UNDERSCORE);
+
   @Override
   public void setNonNullParameter(
       final PreparedStatement ps, final int i, final String parameter, final JdbcType jdbcType)
       throws SQLException {
-    ps.setString(i, transformElasticsearchToSql(parameter));
+    ps.setString(i, transformParameter(parameter));
   }
 
-  public static String transformElasticsearchToSql(final String elasticsearchQuery) {
-    // Regular expression to match unescaped wildcards
-    final var regexAsterisk = "(?<!\\\\)\\*";
-    final var regexQuestionMark = "(?<!\\\\)\\?";
+  private static String replace(
+      final Pattern pattern, final String input, final String replacement) {
+    final var matcher = pattern.matcher(input);
+    final var builder = new StringBuilder();
+    while (matcher.find()) {
+      matcher.appendReplacement(builder, replacement);
+    }
+    matcher.appendTail(builder);
+    return builder.toString();
+  }
+
+  public static String transformParameter(String parameter) {
+    // Escape unescaped %
+    parameter = replace(PATTERN_PERCENT, parameter, "\\\\%");
+    // Escape unescaped _
+    parameter = replace(PATTERN_UNDERSCORE, parameter, "\\\\_");
 
     // Replace unescaped * with %
-    final var patternAsterisk = Pattern.compile(regexAsterisk);
-    final var matcherAsterisk = patternAsterisk.matcher(elasticsearchQuery);
-    final var sqlQuery = new StringBuilder();
-    while (matcherAsterisk.find()) {
-      matcherAsterisk.appendReplacement(sqlQuery, "%");
-    }
-    matcherAsterisk.appendTail(sqlQuery);
-
+    parameter = replace(PATTERN_ASTERISK, parameter, "%");
     // Replace unescaped ? with _
-    final var patternQuestionMark = Pattern.compile(regexQuestionMark);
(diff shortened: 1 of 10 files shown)


db/rdbms/src/main/java/io/camunda/db/rdbms/sql/typehandler/WildcardTransformingStringTypeHandler.java
```diff
@@ -15,36 +15,44 @@
...
+    final var builder = new StringBuilder();
+    while (matcher.find()) {
+      matcher.appendReplacement(builder, replacement);
     }
-    matcherAsterisk.appendTail(sqlQuery);
+    matcher.appendTail(builder);
+    return builder.toString();
+  }
 
+  public static String transformElasticsearchToSql(String exp) {
```
tmetzke, line 46: 🔧 The name of this method is a bit misleading, I believe. Are we replacing API wildcards with SQL wildcards? If so, let's name it according to that. The fact that ES is using the same wildcard notion as our API is simply a "coincidence".

db/rdbms/src/main/resources/db/vendor-properties/h2.properties
```diff
@@ -9,3 +9,4 @@
 paging.after=LIMIT #{page.size} OFFSET #{page.from}
 keysetPaging.limit=LIMIT #{page.size}
 variableValue.previewSize=8191
+escapeChar='\\'
```
tmetzke, line 12: ❓ Is this configurable, so customers can set a different escape char? Do we need to consider this on the application level?

db/rdbms/src/test/java/io/camunda/db/rdbms/sql/typehandler/WildcardTransformingStringTypeHandlerTest.java
```diff
@@ -32,4 +32,21 @@ private static Stream<Arguments> provideElasticsearchQueries() {
...
   }
+
+  private static Stream<Arguments> provideSqlQueries() {
+    return Stream.of(
+        Arguments.of("some%query", "some\\%query"),
+        Arguments.of("some\\%query", "some\\%query"),
+        Arguments.of("some_query", "some\\_query"),
+        Arguments.of("some\\_query", "some\\_query"),
+        Arguments.of("some%query_with%wildcards", "some\\%query\\_with\\%wildcards"),
+        Arguments.of("some\\*query\\_with%wild*cards?", "some\\*query\\_with\\%wild%cards_"));
```
tmetzke, line 43: ❓ Should we test this for escaped `?` as well? Plus, what does the RDBMS do with those escaped `*`s and `?`s? Does it take those literally in the end or how is this interpreted?
```suggestion
        Arguments.of("some\\*query\\_with%wild*car\\?ds?", "some\\*query\\_with\\%wild%car\\?ds_"));
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4331,7 +4331,11 @@ components:
             $like:
               description: |
                 Checks if the property matches the provided like value.
-                Supported wildcard characters depend on the configured search client.
+
+                Supported wildcard characters are:
+
+                * `*`: matches zero, one, or multiple characters.
+                * `?`: matches one, single character.
```
tmetzke, line 4338: 👍 So much better to be able to claim this

qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
```diff
@@ -368,10 +367,12 @@ public void shouldRetrieveTaskByCandidateGroup() {
...
     // when
-    final var filter = new StringFilterProperty();
-    filter.$like("grou?");
     final var result =
-        camundaClient.newUserTaskQuery().filter(f -> f.candidateGroup("group")).send().join();
+        camundaClient
+            .newUserTaskQuery()
+            .filter(f -> f.candidateGroup(b -> b.like("grou?")))
+            .send()
+            .join();
```
tmetzke, line 375: ❓ Should we add a couple more ITs that ensure more edge cases work as expected e2e, like escaping things and using characters that would normally be interpreted as wildcards in SQL? I know we have unit tests that test they are transformed as expected but I think it makes sense to also ensure e2e they are executable and lead to the results we expect in every data layer.

tmetzke (commented):
👍 Looks really good already, great improvement!
🔧 There are smaller suggestions plus one question regarding escape chars for RDBMS.

qa/integration-tests/src/test/java/io/camunda/it/client/UserTaskQueryTest.java
danielkelemen: I was also thinking about this. I'll add a couple tests to see that everything is interpreted in all dbs properly.

db/rdbms/src/main/resources/db/vendor-properties/h2.properties
danielkelemen: Right now it's not configurable. We could probably make it, although it is also DB specific, so we would need to add multiple properties.

I'll need test it with values containing backslashes, like `C:\Users` or similar. But I think in worst case we also need to escape these backslashes, too. I don't think it needs to be configurable though.

tmetzke: So, this hardwires the properties for our connection to the specific database? If that's the case, then this can stay as-is. As soon as this is configurable by users, we need to consider this in the wildcard transformation code as well.

tmetzke: Perfect 👍 Yeah, sorry, I didn't make that very specific in the beginning 🙈

tmetzke (approved):
Pre-approving as the suggestions are not blocking

db/rdbms/src/main/java/io/camunda/db/rdbms/sql/typehandler/WildcardTransformingStringTypeHandler.java
danielkelemen: Renamed to `transformParameter`.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Should we explain how to search for values that contain those values already, i.e. how to escape this in search queries in the API so I can find `my*value` using like search?

danielkelemen:
I added one more round of replace in [WildcardTransformingStringTypeHandler](https://github.com/camunda/camunda/blob/deb3b32d1c9d0b28d6ab8b690459ffd5dcb00175/db/rdbms/src/main/java/io/camunda/db/rdbms/sql/typehandler/WildcardTransformingStringTypeHandler.java#L51):
1. Escape unescaped `%` and `_`
   - Searching for variable `business%` will be executed in SQL as `business\%`.
3. Replace unescaped `*` with `%` and `?` with `_`
   - Regular wildcard search translated into SQL.
4. Unescape escaped `*` and `?` (new)
   - Searching for variable `business\*` will be executed in SQL as `business*`, without escaping.

PR #27745, opened 2025-02-06
Bugfix/fix rdbms flickering tests

p-wunderlich:
* Fixes some flickering tests in rdbms
* Log records in rdbms which are not getting exported


zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/RdbmsExporter.java
```diff
@@ -125,6 +127,10 @@ public void export(final Record<?> record) {
     } else {
       LOG.trace("[RDBMS Exporter] No registered handler found for {}", record.getValueType());
     }
+
+    if (!exported) {
+      LOG.debug("[RDBMS Exporter] Record could not be exported {}", record);
+    }
```
stoerti, line 133: Is this purely for debugging? `Record could not be exported` sounds like a problem.

p-wunderlich: I had it before als warning, but the "problem" ist that we ignore lot's of Records. Our canHandle checks are very specific. I think there are many ignored records which gets ignored on purpose ...

stoerti (approved):

PR #27525, opened 2025-01-29
Idempotent message correlation retries

korthout:
## Description

This pull request resolves a bug related to message correlation retries by ensuring that message correlation commands are processed idempotently.

### Problem context
A message correlation is split across two partitions: the message partition (where the message is expected) and the process instance partition (where the process instance exists that awaits a message). When a message is published to the message partition, and that partition is aware of a process instance awaiting the message (represented by a message subscription on the message partition), it attempts to correlate the message by sending a `CORRELATE` command to the process instance partition. This command must be processed and then acknowledged by sending a `CORRELATE` command back to the message partition. The communication between these partition is unreliable, and so either of these commands may fail to arrive at the other partition. Therefore, a retry mechanism exists in the message subscription partition to re-send this `CORRELATE` command to the process instance partition. The checker runs every 30s and only retries commands that haven't been communicated about at least 10s.

If the acknowledgement would be lost, or the partition is down for some time, then multiple `CORRELATE` commands for the same message may exist on the process instance partition due to the retry. Each of these commands would trigger a non-interrupting message catch event an additional time.

```mermaid
sequenceDiagram
    participant MP as Message Partition
    participant PIP as Process Instance Partition

    MP->>PIP: CORRELATE (Message 1)
    Note right of MP: Initial correlation attempt

    PIP->>PIP: Trigger non-interrupting message catch event
    PIP-->>MP: ACK (CORRELATE)
    Note left of PIP: ACK lost due to network issues

    MP->>PIP: CORRELATE (Message 1) [Retry]
    Note right of MP: Retry due to missing ACK
    
    PIP->>PIP: Trigger non-interrupting message catch event again
    Note right of PIP: Multiple correlations trigger the event multiple times
```

### Solving the problem

This pull request solves the problem by deduplicating the `CORRELATE` commands on the process instance partition. It can do this because only one message will be correlating to the message subscription at a time. This means that ordering is guaranteed already, and we can simply reject a secondary `CORRELATE` command for the same message (by its message key). Note that the last correlated message key is stored in the process message subscription.

```mermaid
sequenceDiagram
    participant MP as Message Partition
    participant PIP as Process Instance Partition

    MP->>PIP: CORRELATE (Message 1)
    Note right of MP: Initial correlation attempt

    PIP->>PIP: Trigger non-interrupting message catch event
    PIP-->>MP: ACK (CORRELATE)
    Note left of PIP: ACK lost due to network issues

    MP->>PIP: CORRELATE (Message 1) [Retry]
[...]

PR #27779, opened 2025-02-06
Fail test on unexpected user task command rejection

korthout:
## Description

The `SUCCESS_SUPPLIER` (success expectation) of the user task test client only failed if no records were written as a result. However, typically we fail it also if a rejection was written.

## Related issues

NA

I encountered this while working on #27525


remcowesterhoud (approved):
I like how you used 3 commits to change a whopping 8 lines 😄 Very clear explanation in those commits though 👍

Issue #25279, opened 2024-11-26
RPA files can be linked to ServiceTask jobs


kzraitel:
**Description**:

We can use our `Resource` generalization to link RPA files to `ServiceTask` jobs.

On the Zeebe engine side, resource linking occurs in 3 steps:
1. A BPMN process containing a `ServiceTask` with a `linkedResources` tag is validated on deployment. Some design-time validators with similar functionality can be seen
   * [here](https://github.com/camunda/camunda/blob/599a9c7e65695732887220f70b23dfbad1a6cb80/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/validation/zeebe/ZeebeDesignTimeValidators.java#L49-L52)
   * [here](https://github.com/camunda/camunda/blob/599a9c7e65695732887220f70b23dfbad1a6cb80/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/validation/zeebe/ZeebeDesignTimeValidators.java#L82-L85)
   * [here](https://github.com/camunda/camunda/blob/599a9c7e65695732887220f70b23dfbad1a6cb80/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/validation/zeebe/ZeebeDesignTimeValidators.java#L92-L99)
   * and [here](https://github.com/camunda/camunda/blob/fd489e785e0ac4e2728ce2993f4be76925a2664d/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/validation/zeebe/ExecutionListenersValidator.java#L32).
2. On Job activation, the job and its extension elements are transformed [here](https://github.com/camunda/camunda/blob/ff8dbe135d1523283ba1324ed42c98824150432d/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/JobWorkerElementTransformer.java#L38-L49). The new extension element `ZeebeLinkedResources.class` will need to be fetched and transformed in the same transformer as well.
   - A similar transformation is done [here](https://github.com/camunda/camunda/blob/7270681a81e1264625d66897587522eef3639e40/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/CallActivityTransformer.java#L39-L65).
[...]

Issue #24774, opened 2024-11-15
ServiceTasks support a new `linkedResource` extension element


koevskinikola:
### Description
We should add a new `linkedResource` extension element class to the BPMN Model API which represents the `camunda:linkedResource` XML element (that can be used to link, for example, an RPA script to a ServiceTask Job).
The new extension element class will be used by Zeebe to retrieve the appropriate linking information from a diagram's XML when it gets deployed or processed.

It should support the following attributes:
- `resourceId`
- `bindingType` (`latest` (default), `deployment`, `versionTag`)
- `versionTag` (if `bindingType` is set to `versionTag`)

marstamm:
Sync with @koevskinikola , @kzraitel and @marstamm :

## Topics: 
 - What attributes should the LinkedResource support
 - How is the linked Resource consumed in the Client

## Outcomes:

### LinkedResource Structure
- LinkedResources will have the following sctucture
  ```
  <bpmn:extensionElements>
    <LinkedResource
      resourceId="=myScript"
      bindingType="latest|deployment|versionTag"
      versionTag?="v1"
      resourceType="RPA"
      linkName?="customMetadata"
    />
  </>
  ```
- We support multiple LinkedResources for a single element. This can be useful for future usecases, e.g. Connectors where configs might be used in conjunction with scripts
  ```
  <bpmn:extensionElements>
    <LinkedResource
      resourceId="restAPIConfig"
      bindingType="latest"
      resourceType="config"
    />
    <LinkedResource
      resourceId="=myScript"
      bindingType="latest"
      resourceType="RPA"
    />
  </>
  ```

### Consumption in the client

The linked resource will be evaluated at Job activation. The resolved key will be used in a custom Task header (`camunda::linked_resources`, **TODO**: check if header name fits current naming convention and adjust if necessary). 
The Header will be an encoded JSON object with the following info: 

```JSON
[
  {
    "resourceId": "resolvedId",
    "key": "uniqueKey",
    "resourceType": "RPA",
    "linkName": "RPAScript"
  },
   ...
]
```

Optionally, you can add a metadata field `linkName` to the Extension element. This will be used to annotate the resolved resource in the Task header. This allows easy handling in the client, since ID and key might change.

### Example
Linked Resource:
```
    <LinkedResource
      resourceId="=myScript"
      bindingType="latest"
      resourceType="RPA"
      linkName="RPAScript"
    />
```

Resolves to the Task Header:
```JSON
[
  {
[...]

PR #27785, opened 2025-02-06
feat: add LinkedResource transformation (part 3)

kzraitel:
## Description
Issue ref: https://github.com/camunda/camunda/issues/25279
Covers items 3 & 4 
ResourceKey lookup will be in a separate PR

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
@@ -9,6 +9,7 @@ package io.camunda.zeebe.engine.processing.bpmn.behavior;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Function;
 import com.google.common.base.Strings;
 import io.camunda.zeebe.el.Expression;
 import io.camunda.zeebe.engine.metrics.JobMetrics;
@@ -18,6 +19,7 @@ import io.camunda.zeebe.engine.processing.common.Failure;
 import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableJobWorkerElement;
 import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutionListener;
 import io.camunda.zeebe.engine.processing.deployment.model.element.JobWorkerProperties;
+import io.camunda.zeebe.engine.processing.deployment.model.element.LinkedResource;
 import io.camunda.zeebe.engine.processing.deployment.model.element.TaskListener;
 import io.camunda.zeebe.engine.processing.deployment.model.transformer.ExpressionTransformer;
 import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
@@ -45,6 +47,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import org.agrona.DirectBuffer;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -97,6 +100,19 @@ public final class BpmnJobBehavior {
     return Either.<Failure, JobProperties>right(new JobProperties())
         .flatMap(p -> evalTypeExp(jobWorkerProps.getType(), scopeKey).map(p::type))
         .flatMap(p -> evalRetriesExp(jobWorkerProps.getRetries(), scopeKey).map(p::retries))
+        .flatMap(p -> evalLinkedResourceKey(jobWorkerProps).map(p::resourceKey))
+        .flatMap(
+            p ->
+                evalLinkedResourceProperty(jobWorkerProps, LinkedResource::getResourceId)
+                    .map(p::resourceId))
+        .flatMap(
+            p ->
+                evalLinkedResourceProperty(jobWorkerProps, LinkedResource::getResourceType)
+                    .map(p::resourceType))
+        .flatMap(
+            p ->
+                evalLinkedResourceProperty(jobWorkerProps, LinkedResource::getLinkName)
+                    .map(p::linkName))
         .flatMap(
             p ->
                 userTaskBehavior
@@ -136,6 +152,34 @@ public final class BpmnJobBehavior {
                         scopeKey)
                     .map(key -> Objects.toString(key, null))
                     .map(p::formKey));
+  }
+
+  private Either<Failure, String> evalLinkedResourceProperty(
+      final JobWorkerProperties props, final Function<LinkedResource, String> propertyFunction) {
+    final List<LinkedResource> linkedResources = props.getLinkedResources();
+    if (linkedResources == null || linkedResources.isEmpty()) {
+      return Either.right(null);
+    }
+    return Either.right(
+        linkedResources.stream()
+            .map(propertyFunction)
+            .collect(Collectors.joining(PROPERTY_DELIMITER)));
+  }
+
+  private Either<Failure, String> evalLinkedResourceKey(final JobWorkerProperties props) {
+    final List<LinkedResource> linkedResources = props.getLinkedResources();
+    if (linkedResources == null || linkedResources.isEmpty()) {
+      return Either.right(null);
+    }
+    return Either.right(
+        linkedResources.stream()
+            .map(l -> resolveLinkedResourceKey(l))
+            .collect(Collectors.joining(PROPERTY_DELIMITER)));
+  }
+
+  private String resolveLinkedResourceKey(final LinkedResource linkedResource) {
+    // TODO Add implementation in the next PR
+    return "";
   }
 
   private static String asListLiteralOrNull(final List<String> list) {
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
@@ -9,7 +9,6 @@ package io.camunda.zeebe.engine.processing.bpmn.behavior;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.base.Function;
 import com.google.common.base.Strings;
 import io.camunda.zeebe.el.Expression;
 import io.camunda.zeebe.engine.metrics.JobMetrics;
@@ -39,6 +38,7 @@ import io.camunda.zeebe.protocol.record.value.JobKind;
 import io.camunda.zeebe.protocol.record.value.JobListenerEventType;
 import io.camunda.zeebe.stream.api.state.KeyGenerator;
 import io.camunda.zeebe.util.Either;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -47,7 +47,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 import org.agrona.DirectBuffer;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -100,19 +99,7 @@ public final class BpmnJobBehavior {
     return Either.<Failure, JobProperties>right(new JobProperties())
         .flatMap(p -> evalTypeExp(jobWorkerProps.getType(), scopeKey).map(p::type))
         .flatMap(p -> evalRetriesExp(jobWorkerProps.getRetries(), scopeKey).map(p::retries))
-        .flatMap(p -> evalLinkedResourceKey(jobWorkerProps).map(p::resourceKey))
-        .flatMap(
-            p ->
-                evalLinkedResourceProperty(jobWorkerProps, LinkedResource::getResourceId)
-                    .map(p::resourceId))
-        .flatMap(
-            p ->
-                evalLinkedResourceProperty(jobWorkerProps, LinkedResource::getResourceType)
-                    .map(p::resourceType))
-        .flatMap(
-            p ->
-                evalLinkedResourceProperty(jobWorkerProps, LinkedResource::getLinkName)
-                    .map(p::linkName))
+        .flatMap(p -> evalLinkedResourceProps(jobWorkerProps, scopeKey).map(p::linkedResources))
         .flatMap(
             p ->
                 userTaskBehavior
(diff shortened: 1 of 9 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/JobWorkerElementTest.java
```diff
@@ -326,4 +327,42 @@ public void shouldCreateJobWithJobRetriesExpression() {
...
+
+  @Test
+  public void shouldCompleteJobWithLinkedResources() {
+    final BpmnModelInstance modelInstance =
+        Bpmn.createExecutableProcess("process")
+            .startEvent()
+            .serviceTask(
+                "my_linked_resource",
+                t ->
+                    t.zeebeLinkedResources(
```
korthout, line 339: ❌ This test case is specific to service task only, but this test class is for any [job worker](https://docs.camunda.io/docs/next/components/concepts/job-workers/) element (i.e. [service task](https://docs.camunda.io/docs/components/modeler/bpmn/service-tasks/), [business rule task](https://docs.camunda.io/docs/components/modeler/bpmn/business-rule-tasks/), [script task](https://docs.camunda.io/docs/components/modeler/bpmn/script-tasks/), [send task](https://docs.camunda.io/docs/components/modeler/bpmn/send-tasks/), [message end event, and message intermediate throw event](https://docs.camunda.io/docs/components/modeler/bpmn/message-events/#message-throw-events)) and executes the test case parameterized for each of those elements (redundantly).

❓ This raises the question: Do we want to support `zeebeLinkedResources` on service tasks only, or on all elements that can be implemented by a Job Worker?

1. If we want to support all job worker elements, then we should keep the test here, but change it to use `.withXmlResource(process(t -> t.zeebeJobType("test").zeebeLinkedResources(`, which currently produces a compile error because `zeebeLinkedResources` is implemented in [`AbstractServiceTaskBuilder`](https://github.com/camunda/camunda/blob/main/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/AbstractServiceTaskBuilder.java) rather than in [`ZeebeJobWorkerPropertiesBuilderImpl`](https://github.com/camunda/camunda/blob/main/zeebe/bpmn-model/src/main/java/io/camunda/zeebe/model/bpmn/builder/ZeebeJobWorkerPropertiesBuilderImpl.java).
2. If we only want to support service tasks, then we should create a new `ServiceTaskTest` class in `io.camunda.zeebe.engine.processing.bpmn.activity` to describe the new unique behavior of service tasks, and move this test there.

❓ I'd also like to expand my previous question: What about [Execution Listeners](https://docs.camunda.io/docs/components/concepts/execution-listeners/) and [User task Listeners](https://docs.camunda.io/docs/next/components/concepts/user-task-listeners/)? These are also simply implemented as Job Workers, and may want to run an RPA script. Personally, I think we can keep it out of scope as this will require more effort and tests. However, I could not find any documentation about this.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/JobWorkerElementTest.java
```diff
@@ -326,4 +327,42 @@ public void shouldCreateJobWithJobRetriesExpression() {
...
+    // then
+    assertThat(
+            RecordingExporter.processInstanceRecords()
+                .withProcessInstanceKey(processInstanceKey)
+                .limitToProcessInstanceCompleted())
+        .extracting(r -> r.getValue().getBpmnElementType(), Record::getIntent)
+        .containsSubsequence(
+            tuple(BpmnElementType.SERVICE_TASK, ProcessInstanceIntent.ELEMENT_COMPLETING),
+            tuple(BpmnElementType.SERVICE_TASK, ProcessInstanceIntent.ELEMENT_COMPLETED),
+            tuple(BpmnElementType.PROCESS, ProcessInstanceIntent.ELEMENT_COMPLETED));
```
korthout, line 366: 💭 While this test does what is says (`shouldCompleteJobWithLinkedResources`), I don't see how this is particularly useful. A job worker will receive a job with linked resources like any other, with the only difference being the addition of linked resources in the custom headers of the activated job.

❌ I'd much rather see a test case that shows that a `JobRecord.CREATED` was appended with `customHeaders` containing one or more specific `linkedResources`. Such a test would cover the code added in `BpmnJobBehavior`, and aligns with the test case above this one (`shouldCreateJobWithJobRetriesExpression`). Such a test can be written even without #27903, but `resourceKey` will not yet be available. I'm marking this as a change request, as I see a need for us to test the actual behavior implemented in this PR.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/zeebe/LinkedResourcesTransformer.java
```diff
@@ -0,0 +1,46 @@
...
+import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableJobWorkerElement;
+import io.camunda.zeebe.engine.processing.deployment.model.element.JobWorkerProperties;
+import io.camunda.zeebe.engine.processing.deployment.model.element.LinkedResource;
+import io.camunda.zeebe.model.bpmn.instance.FlowElement;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeLinkedResource;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeLinkedResources;
+import java.util.List;
+import java.util.Optional;
+
+public class LinkedResourcesTransformer {
```
korthout, line 19: 🔧 This class is among the more interesting parts of code that was added/changed in this PR. A small unit test for the transformer (similar to [`ExecutionListenerTransformerTest`](https://github.com/camunda/camunda/blob/main/zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/zeebe/ExecutionListenerTransformerTest.java)) could help build confidence in the implementation of this transformer class.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/zeebe/LinkedResourcesTransformer.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  private LinkedResource toLinkedResourceModel(final ZeebeLinkedResource zeebeLinkedResource) {
+    final LinkedResource linkedResource = new LinkedResource();
+    linkedResource.setResourceId(zeebeLinkedResource.getResourceId());
+    linkedResource.setLinkName(zeebeLinkedResource.getLinkName());
+    linkedResource.setResourceType(zeebeLinkedResource.getResourceType());
+    linkedResource.setBindingType(zeebeLinkedResource.getBindingType());
+    return linkedResource;
+  }
```
korthout, line 45: 👍 Simple and straightforward code is always easiest to review and maintain. Thanks!

korthout (changes requested):
Thanks for this contribution @kzraitel 👏 I'm taking over from my teammate. Please note that I use [review emoji code](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md#review-emoji-code) to better express my intentions.

👍 I really like how you kept the changes condensed in this PR. It made reviewing easier.

💭 In future PRs, I recommend splitting your commits into more meaningful and smaller commits (e.g. by [separating structural from behavioral changes](https://github.com/camunda/camunda/wiki/Pull-Requests-and-Code-Reviews#ive-separated-structural-from-behavioral-changes)) and [describing the reasoning for the changes](https://github.com/camunda/camunda/wiki/Pull-Requests-and-Code-Reviews#ive-written-a-clear-changelist-description) in the commit messages. This makes life for the reviewer easier.

❌ I'm missing a test case covering the main behavioral change: a created job should now have linked resources in its custom headers. This must be added for us to merge this.

Please have a look at my other comments and questions.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/JobWorkerElementTest.java
korthout: Oh wait. Right after I post my review, I find [this comment](https://github.com/camunda/camunda/issues/24774#issuecomment-2587439011) stating that we will focus on service tasks only.

That means:
- please create a new `ServiceTaskTest` class in `io.camunda.zeebe.engine.processing.bpmn.activity` to describe the new unique behavior of service tasks, and move this test there
- you can ignore my expanded question regarding listeners

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
korthout, line 1: 💭 By the way, if you wanted to keep this PR focused on the transformation only, then this class should not have seen any changes. You could consider moving these changes to your other PR to fully split the PRs semantically:
- this PR is about transforming the service task
- the other PR is about linking resources for an activating service task at runtime

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
```diff
@@ -138,6 +141,29 @@ public Either<Failure, JobProperties> evaluateJobExpressions(
...
+    if (linkedResources == null || linkedResources.isEmpty()) {
+      return Either.right(null);
+    }
+    final List<LinkedResourceProps> linkedResourceProps = new ArrayList<>();
+    for (final LinkedResource linkedResource : linkedResources) {
+      final LinkedResourceProps resourceProps = new LinkedResourceProps();
+      resourceProps.setResourceKey(resolveLinkedResourceKey(linkedResource, scopeKey));
+      resourceProps.setResourceType(linkedResource.getResourceType());
+      resourceProps.setLinkName(linkedResource.getLinkName());
+      linkedResourceProps.add(resourceProps);
```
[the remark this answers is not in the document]
kzraitel: As discussed we will assume that FEEL expression is not yet needed. We will add later in another PR if needed.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/zeebe/LinkedResourcesTransformer.java
kzraitel: Added test case

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -0,0 +1,80 @@
...
+    final LinkedResourceProps resourceProps = new LinkedResourceProps();
+    resourceProps.setResourceType("RPA");
+    resourceProps.setLinkName("my_link");
+    // TODO: key lookup will be implemented in the next PR
+    resourceProps.setResourceKey("");
+
+    Assertions.assertThat(jobCreated.getValue())
+        .hasCustomHeaders(
+            Map.of(
+                "linkedResources", MAPPER.writeValueAsString(Lists.newArrayList(resourceProps))));
```
korthout, line 78: 🔧 We can just use the standard library's `List.of` instead of importing the third party `com.tngtech.archunit.thirdparty.com.google.common.collect.Lists`

```suggestion
            Map.of("linkedResources", MAPPER.writeValueAsString(List.of(resourceProps))));
```

korthout (commented):
Thanks for the changes @kzraitel 👍 

👍 The changes look good to me

🔧 I have a few suggestions. Please have a look at them and apply those that you agree with.

Let me know when it's ready for a last look.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: That is an accidental import :(

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/JobWorkerElementTransformer.java
```diff
@@ -46,5 +50,8 @@ public void transform(final T element, final TransformContext context) {
 
     final var taskHeaders = element.getSingleExtensionElement(ZeebeTaskHeaders.class);
     taskHeadersTransformer.transform(jobWorkerElement, taskHeaders, element);
+
+    final var linkedResources = element.getSingleExtensionElement(ZeebeLinkedResources.class);
+    linkedResourcesTransformer.transform(jobWorkerElement, linkedResources);
```
korthout, line 55: ❌ As we decided to limit the Linked Resources to Service Tasks only, we should actually only transform it if this concerns a service task

```suggestion

    if (type.equals(ServiceTask.class)) {
      final var linkedResources = element.getSingleExtensionElement(ZeebeLinkedResources.class);
      linkedResourcesTransformer.transform(jobWorkerElement, linkedResources);
    }
```

❌ We also need to validate against usage of `ZeebeLinkedResources` in other elements by adding a new validation to `ZeebeDesignTimeValidators`. It's not clear to me what it should exactly look like, but I recommend writing a test case adding a linked resources to some other element, and register a new implementation of `ModelElementValidator<ZeebeLinkedResources>` to see if/how you can reject it.

korthout (changes requested):
Thanks @kzraitel 🚀 

👍 Almost there!

❌ I overlooked a problem on my previous review pass that must be addressed. Please have a look.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/JobWorkerElementTransformer.java
kzraitel: I've added the check for ServiceTask. I also added a test so that when we add `LinkedResources` to a component that is handled by `JobWorkerElementTransformer` (`SendTask` in our case), linkedResources don't appear as a header.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -74,4 +77,24 @@ public void shouldCreateJobWithLinkedResources() throws JsonProcessingException
...
+    // when
+    final long processInstanceKey = ENGINE.processInstance().ofBpmnProcessId(PROCESS_ID).create();
+
+    // then
+    final Record<JobRecordValue> jobCreated =
+        RecordingExporter.jobRecords(JobIntent.CREATED)
+            .withProcessInstanceKey(processInstanceKey)
+            .getFirst();
+
+    assertThat(jobCreated.getValue().getCustomHeaders()).doesNotContainKey("linkedResources");
```
korthout, line 98: This is sadly not enough, because the model is deployed and will be transformed on restarts including version updates. So if we ever expand the transformer without being careful, we'd all of the sudden change what this deployed process does.

❌ Instead, we need to reject this deployment by adding the validation I mentioned [here](https://github.com/camunda/camunda/pull/27785#discussion_r1961864257).

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/JobWorkerElementTransformer.java
```diff
@@ -50,8 +51,9 @@ public void transform(final T element, final TransformContext context) {
 
     final var taskHeaders = element.getSingleExtensionElement(ZeebeTaskHeaders.class);
     taskHeadersTransformer.transform(jobWorkerElement, taskHeaders, element);
-
-    final var linkedResources = element.getSingleExtensionElement(ZeebeLinkedResources.class);
-    linkedResourcesTransformer.transform(jobWorkerElement, linkedResources);
+    if (element instanceof ServiceTask) {
```
korthout, line 54: 🔧 No need for instanceof check. You can use `if (type.equals(ServiceTask.class)) {`

korthout (changes requested):
👍 Thanks for the changes and the added test @kzraitel 

❌ We must reject the deployment if it's used wrongly to avoid changing behavior of the process in future versions.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/JobWorkerElementTransformer.java
kzraitel: We don't have any type there. Am I missing something?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: I didn't find any place where we use explicit prohibition for extension elements in other components. For example, there is no explicit prohibition for listeners in ServiceTask though it doesn't support them. Adding validation for `linkedResources` absence in. other components doesn't feel right taking into account current architecture.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/JobWorkerElementTransformer.java
korthout: 👍 You do, it's here:
https://github.com/camunda/camunda/blob/64fa223e18ab3e339ae037a888dd42dd2bb0b4e1/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/JobWorkerElementTransformer.java#L23

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
korthout: Interesting point. Thanks for sharing your argument. I agree 👍

korthout (approved):
Thanks! LGTM 💯

PR #27798, opened 2025-02-06
refactor: validate grpc and rest URIs in spring sdk properties

houssain-barouni:
## Description

Same as in https://github.com/camunda/camunda/pull/22241
Adding validation on gRPC and REST URIs provided in spring SDK properties

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/clients/java/src/main/java/io/camunda/client/impl/util/ClientPropertiesValidationUtils.java
+++ b/clients/java/src/main/java/io/camunda/client/impl/util/ClientPropertiesValidationUtils.java
@@ -21,14 +21,6 @@ public final class ClientPropertiesValidationUtils {
 
   private ClientPropertiesValidationUtils() {}
 
-  public static void validateGrpcAddress(final URI grpcAddress) {
-    validateAddress(grpcAddress, "grpcAddress");
-  }
-
-  public static void validateRestAddress(final URI restAddress) {
-    validateAddress(restAddress, "restAddress");
-  }
-
   /**
    * Validates that the provided address is an absolute URI.
    *
@@ -43,7 +35,7 @@ public final class ClientPropertiesValidationUtils {
    *
    * throws IllegalArgumentException if the provided address is not an absolute URI
    */
-  private static void validateAddress(final URI address, final String propertyName) {
+  public static void checkIfUriIsAbsolute(final URI address, final String propertyName) {
     if (address != null && address.getHost() == null) {
       throw new IllegalArgumentException(
           String.format("'%s' must be an absolute URI", propertyName));
--- a/clients/java/src/main/java/io/camunda/client/impl/util/ClientPropertiesValidationUtils.java
+++ b/clients/java/src/main/java/io/camunda/client/impl/util/ClientPropertiesValidationUtils.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright © 2017 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.client.impl.util;
+
+import java.net.URI;
+
+public final class ClientPropertiesValidationUtils {
+
+  private ClientPropertiesValidationUtils() {}
+
+  /**
+   * Validates that the provided gRPC address is an absolute URI.
+   *
+   * <p>We use {@code URI.getHost() == null} to check for absolute URIs because:
+   *
+   * <ul>
+   *   <li>For absolute URIs (with a scheme) (e.g., "https://example.com"), {@code URI.getHost()}
+   *       returns the hostname (e.g., "example.com").
+   *   <li>For relative URIs (without a scheme) (e.g., "example.com"), {@code URI.getHost()} returns
+   *       {@code null}.
+   * </ul>
+   *
+   * throws IllegalArgumentException if the provided gRPC address is not an absolute URI
+   */
+  public static void validateGrpcAddress(final URI grpcAddress) {
+    if (grpcAddress != null && grpcAddress.getHost() == null) {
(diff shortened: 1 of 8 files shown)


jonathanlukas:
@houssain-barouni is this a validation that the CamundaConfiguration of the plain java client does as well?
If yes, I would tend to unify the behaviour and use the configuration of the plain java client as bean instead. wdyt?

houssain-barouni:
> @houssain-barouni is this a validation that the CamundaConfiguration of the plain java client does as well? If yes, I would tend to unify the behaviour and use the configuration of the plain java client as bean instead. wdyt?

Yes, in the java client, it is done in [CamundaClientBuilderImpl](https://github.com/camunda/camunda/blob/9d5d775aea8c74a16f9e3efa48f1568602f93c6b/clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java#L427-L443)
However since in Spring SDK we directly implement `CamundaClientConfiguration` class (CamundaClientBuilder is not used), we have to replicate it in the properties classes that are used in the configuration implementation CamundaClientConfigurationImpl

jonathanlukas:
For me, the reason seems to be that the plain java camunda client does not offer a native impl for the configuration, just a builder...
So, maybe we could create a Util class for the client properties validation so that we do not have to duplicate code and have consistent behaviour. wdyt?

tmetzke:
I generally agree with @jonathanlukas here regarding code duplication. Let's try to reuse what we provide in the client itself already. It may seem over-engineered, but this validation is not super obvious, so I'd also vote not to duplicate it. We may overlook changing this here when we adjust this in the client at some point.

clients/spring-boot-starter-camunda-sdk/src/test/java/io/camunda/spring/client/properties/CamundaClientPropertiesPostProcessorTest.java
```diff
@@ -58,6 +59,30 @@ void shouldMapLists() {
     }
   }
 
+  @Nested
```
jonathanlukas, line 62: These test should go to `JavaClientPropertiesTest` as these tests focus on the actual properties handling.

`CamundaClientPropertiesPostProcessorTest` performs assertions on the post processor behaviour:

* activating modes
* mapping legacy properties

clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/properties/CamundaClientProperties.java
```diff
@@ -175,6 +175,18 @@ public URI getGrpcAddress() {
   }
 
   public void setGrpcAddress(final URI grpcAddress) {
+    /*
```
jonathanlukas, line 178: Move to Util class (CamundaClientConfigurationUtil?) in the java client and call the method (CamundaClientConfigurationUtil.validateGrpcAddress) from there.

clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/properties/CamundaClientProperties.java
```diff
@@ -183,6 +195,18 @@ public URI getRestAddress() {
   }
 
   public void setRestAddress(final URI restAddress) {
+    /*
```
jonathanlukas, line 198: Move to Util class (CamundaClientConfigurationUtil?) in the java client and call the method (CamundaClientConfigurationUtil.validateRestAddress) from there.

jonathanlukas (changes requested):
As discussed previously.

Also, I think the tests should go to another test class

clients/java/src/main/java/io/camunda/client/impl/util/ClientPropertiesValidationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.client.impl.util;
+
+import java.net.URI;
+
+public final class ClientPropertiesValidationUtils {
```
tmetzke, line 20: 🙃 It's just a nitpick, but we're basically duplicating the validation code. Only parts of the error message are different.

houssain-barouni: I always question myself about when to stop extracting common code for the sake of readability. Introducing a generic method will require additional parameter just to differentiate error messages, which I don't find elegant neither :) .

houssain-barouni: I hope this looks better now https://github.com/camunda/camunda/pull/27798/commits/bcb4f7e5122a4e2659fee9fd539c82bb15c92e88

houssain-barouni: maybe not a good idea, uri.toString() can return strange things

clients/java/src/main/java/io/camunda/client/impl/util/ClientPropertiesValidationUtils.java
```diff
@@ -0,0 +1,52 @@
...
+   * <ul>
+   *   <li>For absolute URIs (with a scheme) (e.g., "https://example.com"), {@code URI.getHost()}
+   *       returns the hostname (e.g., "example.com").
+   *   <li>For relative URIs (without a scheme) (e.g., "example.com"), {@code URI.getHost()} returns
+   *       {@code null}.
+   * </ul>
+   *
+   * throws IllegalArgumentException if the provided address is not an absolute URI
+   */
+  private static void validateAddress(final URI address, final String propertyName) {
```
tmetzke, line 46: 🔧 My suggestion here would be:
* Make this `public` and let the consumer decide about the message (either completely or by stating what the `propertyName` is concatenated with). We don't really need separate methods for this, I think.
* Rename this to `checkIfUriIsAbsolute` or something because that is all it does

Alternative:
* This only is an `isAbsolute` method that returns a boolean and we explain why we need this instead of the `isAbsolute` provided by the URI class.

houssain-barouni: done!

tmetzke (approved):
👍 Looks good to me

houssain-barouni:
Hi @jonathanlukas, does it look good to you now?

jonathanlukas (approved):
Great work! Thank you @houssain-barouni

PR #27818, opened 2025-02-07
feat: change patch to put in user update endpoint

### LINKED ISSUE #27626: Change Update user HTTP method from PATCH to PUT

maryarm:
## Description

Switch PATCH to PUT to support updating a full user record 

## Related issues

closes #27626


identity/client/src/utility/api/users/index.ts
```diff
@@ -32,8 +32,10 @@ export const createUser: ApiDefinition<undefined, CreateUserParams> = (user) =>
 
 export const updateUser: ApiDefinition<undefined, User> = (user) => {
   const { name, email, username, password } = user;
-  return apiPatch(`${USERS_ENDPOINT}/${username}`, {
-    changeset: { name, email, password: password ?? "" },
+  return apiPut(`${USERS_ENDPOINT}/${username}`, {
+    name,
+    email,
+    password: password ?? "",
```
Ben-Sheppard, line 38: 💭 this is a source of a bug (not introduced by you), if you update a user and don't set a password the request contains an empty string and then we will set the password to an empty string in the data stores (because we don't check validation etc [here](https://github.com/camunda/camunda/pull/27818/files#diff-4d443ca94da97e614acb13ef011b4ea527f23933462a28b9f1a539c32e6babe0R202).

I think as we're updating this here we should also think about that bug and resolve it.

Ben-Sheppard (commented):
Changes look as expected in general but I have one thought about a bug we're experiencing that would be good to resolve also

identity/client/src/utility/api/users/index.ts
maryarm: Actually it was also my question, it seems previously it was possible to update without setting a password but then it changed. Should I remove this condition here and check for password to not be null in controller?

Ben-Sheppard: I think its best handled in the controller/mapper layer because then we can ensure a consistent approach for the UI and API, what I am not sure on though is how the processors further down the stack will react to null, will they set it to empty or ignore the field on the record?

If it turns out to be a bigger thing I'm fine with keeping this as it is for now to keep things moving and create a targeted PR later that includes the complete set of changes required for that

maryarm: I checked the code. I changed the code to avoid encoding the empty password, so it will be ok and the current password won't change.

Ben-Sheppard (approved):
Thank you for handling the password issue, there is a bug ticket we can also close off with this change (but we can handle that separately).

It looks like there are dependency issues in the build but I'm going to approve early so this can be merged once thats resolved.

PR #27828, opened 2025-02-07
feat: add AWS OS support for multi-DB test

igpetrov:
## Description

feat: add AWS OS support for multi-DB test

## Checklist

- [ ] for CI changes:

## Related issues

closes #


ghost (approved):
lgtm

qa/integration-tests/src/test/java/io/camunda/it/utils/MultiDbConfigurator.java
```diff
@@ -164,6 +164,45 @@ public void configureRDBMSSupport() {
         });
   }
 
+  public void configureAWSOSupport(final String opensearchUrl, final String indexPrefix) {
+    final OperateOpensearchProperties operateOpensearch = operateProperties.getOpensearch();
```
ChrisKujawa, line 168: ❓ 🔧  I think we should simply reuse the other Opensearch configuration, can we call/delegate to the other and then afterwards set the was enabled?

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -167,6 +169,10 @@ public void beforeAll(final ExtensionContext context) {
             () -> validateSchemaCreation(DEFAULT_ES_URL, testPrefix, expectedDescriptors);
       }
       case RDBMS -> multiDbConfigurator.configureRDBMSSupport();
+      case AWS_OS -> {
+        final var awsOSUrl = System.getProperty(PROP_CAMUNDA_IT_OPENSEARCH_AWS_URL);
+        multiDbConfigurator.configureAWSOSupport(awsOSUrl, UUID.randomUUID() + "multidb");
```
ChrisKujawa, line 174: ❌ Let's use the same prefix, as we have in the others. 

❌ Let's set the `databaseSetupReadinessWaitStrategy =
￼            () -> validateSchemaCreation(DEFAULT_ES_URL, testPrefix, expectedDescriptors);		￼            () -> validateSchemaCreation(DEFAULT_ES_URL, testPrefix, expectedDescriptors);` so we can validate that the schema is created.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -99,6 +100,7 @@ public class CamundaMultiDBExtension
     implements AfterAllCallback, BeforeAllCallback, ParameterResolver {
   public static final String PROP_CAMUNDA_IT_DATABASE_TYPE =
       "test.integration.camunda.database.type";
+  public static final String PROP_CAMUNDA_IT_OPENSEARCH_AWS_URL = "camunda.it.opensearch.aws.url";
```
ChrisKujawa, line 103: 🔧 ❌ Based on recent discussions with @npepinpe we determined it would be better to leave the `camunda` namespace for production related configs.

Can we change this to `test.integration.opensearch.aws.url`

.github/workflows/zeebe-aws-os-dispatch-tests.yml
```diff
@@ -108,7 +115,8 @@ jobs:
           -D maven.javadoc.skip=true
           -D skipUTs -D skipChecks -D skipOptimize
           -D camunda.it.opensearch.aws.url="${{ env.OS_URL }}"
-          -P parallel-tests,extract-flaky-tests
+          -D test.integration.camunda.database.type=AWS_OS
+          -P parallel-tests,extract-flaky-tests,multi-db-test
```
ChrisKujawa, line 119: 🚀

ChrisKujawa (commented):
Thanks @igpetrov cool stuff 🚀 just some smaller things to discuss

qa/integration-tests/src/test/java/io/camunda/it/utils/MultiDbConfigurator.java
igpetrov: Reused

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
igpetrov: Regarding prefix: used as in other code blocks.
Regarding schema validation: not relevant anymore since this method moved to `MultiDbSetupHelper` interface. And I am using that via `ElasticOpenSearchSetupHelper`.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
igpetrov: Done here and in all other tests

igpetrov:
Hi @ChrisKujawa, please take a look once again. Related test execution is here: https://github.com/camunda/camunda/actions/runs/13409014277.

Regarding previously discussed https://camunda.slack.com/archives/C05QBRGMA9F/p1739276759677299, the issue turned out to be AWS OS not being fast enough in creating indexes, so I increased some await times in tests.

qa/integration-tests/src/test/java/io/camunda/it/client/DecisionInstanceQueryTest.java
```diff
@@ -341,7 +341,7 @@ private static EvaluateDecisionResponse evaluateDecision(
   private static void waitForDecisionsToBeEvaluated(
       final CamundaClient camundaClient, final int expectedCount) {
     Awaitility.await("should deploy decision definitions and import in Operate")
-        .atMost(Duration.ofSeconds(15))
+        .atMost(Duration.ofSeconds(120))
```
ChrisKujawa, line 344: TBH I'm not sure whether I like that we increase everywhere the timeouts 🤔 I'm wondering whether we could speed up the index creation somehow. 

I tried to parallelize it here https://github.com/camunda/camunda/pull/28433 let's see what the results are

igpetrov: @ChrisKujawa do you not like the fact that we introduce wait time in general or that OS doesn't create indexes on time?

If concern is in general, we can make delay a CI-based variable. AWS OS CI tests have intention to run weekly once, hence I don't see a reason to optimize that: after all this `await` will hold 120 seconds in worst case but default polling interval is 100ms.

If concern is about AWS OS, your proposed solution might work.

igpetrov:
So reversing to original 15 seconds values didn't help. Even after increasing to 1m, I see few of these:
`Error:    UserTaskQueryTest.beforeAll:67->waitForTasksBeingExported:1302 ? ConditionTimeout Condition with alias 'should receive data from ES' didn't complete within 1 minutes because assertion condition defined as a Lambda expression in io.camunda.it.client.UserTaskQueryTest `

igpetrov:
Adjusting time outs to 2m

igpetrov:
@ChrisKujawa these timeout are the smallest I can do. Also, I got a hint from the Infra team, that AWS OS is running on `t3.small` instance (2 vCPU, 2GB RAM), so that might be a reason for often timeouts, considering amount of tests running. But at this moment I have exhausted all software options for optimization.

Ref, passing tests: https://github.com/camunda/camunda/actions/runs/13503836488

qa/integration-tests/src/test/java/io/camunda/it/client/QueryTest.java
```diff
@@ -111,7 +111,7 @@ public static void waitForFlowNodeInstances(
   public static void waitForProcessesToBeDeployed(
       final CamundaClient camundaClient, final int expectedProcessDefinitions) {
     Awaitility.await("should deploy processes and import in Operate")
-        .atMost(Duration.ofSeconds(15))
+        .atMost(Duration.ofSeconds(120))
```
ChrisKujawa, line 114: 🙃 :x: I feel like this is a suboptimal approach and quite fragile.

* In general I think we should try to have a minimal timeout.
* I do understand that we can't have this for AWS as we use small instances, etc. But there is a problem with this approach, new tests were likely not to be written with this in mind. This means engineers will put again, random values and likely too small ones on new tests.
* How can we make sure that new tests are not failing, because of this? How can we make sure that people are not reducing this again, because they are wondering why the defaults are so high?


@npepinpe do you have anything to add here or maybe some idea?

ChrisKujawa: We might want to have some constants, and better documentation regarding the reasons.

Maybe as you [proposed](https://github.com/camunda/camunda/pull/27828/files#r1963123281) we can make it somehow configurable from the CI, that only the AWS pipeline has such big timeouts.

ChrisKujawa (commented):
Thanks @igpetrov I still see some issue with the timeout see below, other than that it looks good 👍🏼

qa/integration-tests/src/test/java/io/camunda/it/client/QueryTest.java
igpetrov: One major issue is that different tests use different time outs and I don't know whether those timeout are taken out of thin air or that actually something required. I see few 15 seconds timeouts and few 60 seconds. This may mean that we might need two separate constants which is not great because whenever we have 2, there will be 3-4-5 at some point.

With my next commit I will try to make all unified with 15 seconds for local and parametrized for CI and see if it helps.

igpetrov: I will also ask tomorrow infra team to increase AWS instance types with more CPU/memory

igpetrov: Instance type tickets: https://github.com/camunda/team-infrastructure/issues/787

igpetrov: Hi @ChrisKujawa do you have other suggestions or can we get it moving? Maintaining this PR becomes a challenging task.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -137,12 +138,20 @@ public class CamundaMultiDBExtension
     implements AfterAllCallback, BeforeAllCallback, ParameterResolver {
   public static final String PROP_CAMUNDA_IT_DATABASE_TYPE =
       "test.integration.camunda.database.type";
+  public static final String TEST_INTEGRATION_OPENSEARCH_AWS_URL =
+      "test.integration.opensearch.aws.url";
+  public static final String PROP_TEST_INTEGRATION_OPENSEARCH_AWS_TIMEOUT =
+      "test.integration.opensearch.aws.timeout.seconds";
+  public static final Duration TIMEOUT_DATABASE_EXPORTER_READINESS_TIMEOUT =
```
ChrisKujawa, line 145: 🔧 I would probably rename them if they are used in the tests to something like:

```suggestion
  public static final Duration TIMEOUT_DATA_AVAILABILITY =
```

This is also somehow duplicated TIMEOUT__TIMEOUT

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
```diff
@@ -137,12 +138,20 @@ public class CamundaMultiDBExtension
...
+      Optional.ofNullable(System.getProperty(PROP_TEST_INTEGRATION_OPENSEARCH_AWS_TIMEOUT))
+          .map(val -> Duration.ofSeconds(Long.parseLong(val)))
+          .orElse(Duration.ofSeconds(30));
```
ChrisKujawa, line 148: ❓ Should we use here the other static value? TIMEOUT_DATABASE_EXPORTER_READINESS

ChrisKujawa (approved):
Looks good thanks for the changes, smaller remarks, please check before merging :) 🙇🏼

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
igpetrov: That one is used for other time outs. Would you like to unify those as well? The `TIMEOUT_DATABASE_EXPORTER_READINESS` is 3 minutes, while tests using this value work fine with 30 seconds timeout. My suggestion is not to reuse that constant. If the person set 3 minutes, it is likely for a reason.

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
igpetrov: That's weird I didn't noticed it. Probably changed somewhere down refactoring 🙈 . Will rename it to as you suggested

qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
ChrisKujawa: Don't you set it to 3 minutes in this pr? 🤔

igpetrov: Sorry, maybe we're talking on different matters: we have a new timeout that I introduced: `TIMEOUT_DATABASE_EXPORTER_READINESS_TIMEOUT` aka `TIMEOUT_DATA_AVAILABILITY` whish is 30 seconds by default. And we have old ones `TIMEOUT_DATABASE_EXPORTER_READINESS` and `TIMEOUT_DATABASE_READINESS` which are 3 minutes now. Those are used in different places.

The reason why I increased those - I didn't want to add additional parametrized value.

igpetrov: BTW, infra team increased instance class for AWS OS. So I will actually decrease timeouts again and see if it is any better now.

igpetrov: There's now `m5.large.search` if you're curious

ChrisKujawa:
@igpetrov interesting would be whether we have an process about following up with test failures or an dri to look into these?

igpetrov:
@ChrisKujawa yes, my team is the DRI. We monitor weekly tests at related [Slack group](https://camunda.slack.com/archives/C0740PUGURH). So we are taking full responsibility for these and other modules mentioned in [this workflow](https://github.com/camunda/camunda/blob/main/.github/workflows/zeebe-aws-os-dispatch-tests.yml).

ChrisKujawa:
That is what I wanted to hear. Thanks Igor!

PR #27859, opened 2025-02-07
fix: ensure we always close background async client

npepinpe:
## Description

This PR closes the ES/OS async client for the batch operations repository. It also moves this behavior to the shared repository to ensure we always close it.

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/ElasticsearchRepository.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/ElasticsearchRepository.java
@@ -24,9 +24,10 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import javax.annotation.WillCloseWhenClosed;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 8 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/ElasticsearchRepository.java
```diff
@@ -35,7 +36,9 @@ public class ElasticsearchRepository {
   protected final Logger logger;
 
   public ElasticsearchRepository(
-      final ElasticsearchAsyncClient client, final Executor executor, final Logger logger) {
+      @WillCloseWhenClosed final ElasticsearchAsyncClient client,
```
ChrisKujawa, line 39: 🤡 What does this annotation? It looks wild.

ChrisKujawa (approved):
Thanks 🚀 Good catch!

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/ElasticsearchRepository.java
npepinpe: It doesn't "do" anything (unless you use some compiler add ons or google's error prone), but it's still helpful imo :)

PR #27878, opened 2025-02-09
Migrate gateway/gRPC metrics to Micrometer

### LINKED ISSUE #26078: Non-Prometheus Micrometer implementations only expose few C8 metrics

npepinpe:
## Description

This PR migrates the gateway and gateway gRPC metrics to Micrometer. For the gateway metrics, the main file was simply not used anymore, so I deleted it.

For the long polling metrics, we had a bit of a bind: we create two different handlers, one for REST, one for gRPC. It's unclear to me why we don't reuse the same for each (or just reuse the job services for both protocol), but as it is now, we end up with two distinct instances.

The problem is they both want to keep track of a set count, which may conflict. You could argue that in the future job polling will be REST only, but there will be a migration period where people may use both, thus showing incorrect values. It's likely not a huge problem, but could derail debugging or diagnosing. 

To circumvent this, I added a new tag to the metric via a composite registry/common tags, `gateway`. On gRPC, the metric will show `gateway=gRPC`, and on REST, it will be tagged `gateway=rest`. This should allow us to differentiate if we need to.

This label can be removed later when we only do polling on REST.

The rest of the changes were getting rid of the Prometheus gRPC interceptor, and replacing it with Micrometer's built-in one. Since that one emits different metrics, the dashboard was amended.

One thing I found out is that you cannot easily show the union of two queries as a histogram, the values are all weird. The union of two series works great for time series though - just not for pre-bucketed/heatmap values :disappointed: 

## Related issues

relates #26078


npepinpe:
@entangled90 - FYI about the heatmaps - the `or` trick doesn't quite work when joining two queries, it seems to sum them incorrectly (merging with the wrong timestamps and so on, ignoring a bunch of values, etc.). It looks fine once there's only one returning values though, so it's not so horrible, but not great. For heatmaps we may want to duplicate the series. For time series data it shows up fine though.

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/metrics/GatewayMetrics.java
```diff
@@ -1,52 +0,0 @@
...
- * except in compliance with the Camunda License 1.0.
- */
-package io.camunda.zeebe.gateway.metrics;
-
-import io.prometheus.client.Counter;
-import io.prometheus.client.Histogram;
-
-public final class GatewayMetrics {
-
-  private static final Histogram REQUEST_LATENCY =
```
entangled90, line 15: It's a shame they were not in use anymore, could be useful when debugging latency

parent/pom.xml
```diff
@@ -67,7 +67,6 @@
     <version.identity>8.6.7</version.identity>
     <version.surefire>3.5.2</version.surefire>
     <version.jackson>2.18.2</version.jackson>
-    <version.java-grpc-prometheus>0.6.0</version.java-grpc-prometheus>
```
entangled90, line 70: :tada:

entangled90 (approved):

PR #27881, opened 2025-02-10
refactor: renamed camunda.client.auth.issuer to camunda.client.auth.token-url

jonathanlukas:
## Description

This PR renames a property as the original name could give a wrong impression about the expected value.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
+++ b/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
@@ -47,7 +47,7 @@ public class CredentialsProviderConfiguration {
             .audience(camundaClientProperties.getAuth().getAudience())
             .scope(camundaClientProperties.getAuth().getScope())
             .authorizationServerUrl(
-                ofNullable(camundaClientProperties.getAuth().getTokenEndpoint())
+                ofNullable(camundaClientProperties.getAuth().getTokenUrl())
                     .map(URI::toString)
                     .orElse(null))
             .credentialsCachePath(camundaClientProperties.getAuth().getCredentialsCachePath())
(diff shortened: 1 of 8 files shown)


houssain-barouni (commented):
Hi @jonathanlukas,
The term `issuer` is defined in the [OAuth 2.0 specification](https://datatracker.ietf.org/doc/html/rfc9207) and is commonly recognized in that context. Could you provide more context on the need behind this change?

jonathanlukas:
Hi @houssain-barouni ,

the change is motivated by the fact that we are currently using the term `issuer` for the wrong thing.

What we are using it for: The endpoint that credentials are sent to to issue a token (this is actually the `token_endpoint`)

What it actually is: The address from where all other oidc endpoints can be reached (via `.well-known/openid-configuration`)

Also, this is currently not in line with our helm charts where we refer to the `global.identity.auth.issuer` as the issuer as what it actually is. On top, there is another property `global.identity.auth.tokenUrl` which is what we refer to as issuer right now.

Source: https://docs.camunda.io/docs/self-managed/setup/guides/connect-to-an-oidc-provider/?optionsType=helm&authPlatform=microsoftEntraId#configuration

houssain-barouni:
> Hi @houssain-barouni ,
> 
> the change is motivated by the fact that we are currently using the term `issuer` for the wrong thing.
> 
> What we are using it for: The endpoint that credentials are sent to to issue a token (this is actually the `token_endpoint`)
> 
> What it actually is: The address from where all other oidc endpoints can be reached (discovery via `.well-known/openid-configuration`)
> 
> Also, this is currently not in line with our helm charts where we refer to the `global.identity.auth.issuer` as the issuer as what it actually is. On top, there is another property `global.identity.auth.tokenUrl` which is what we refer to as issuer right now.
> 
> Source: https://docs.camunda.io/docs/self-managed/setup/guides/connect-to-an-oidc-provider/?optionsType=helm&authPlatform=microsoftEntraId#configuration

Thanks for the clarification, it makes total sense
What about using token_url to be consistent with helm ? in the end it is an url that is required for this property

houssain-barouni (approved):

PR #27903, opened 2025-02-10
feat: add LinkedResource id lookup (part 4)

kzraitel:
## Description
Issue ref: https://github.com/camunda/camunda/issues/25279
Final part of the issue - resourceKey lookup

## Checklist

- [ ] for CI changes:

## Related issues

closes #


zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
```diff
@@ -159,9 +166,98 @@ private Either<Failure, List<LinkedResourceProps>> evalLinkedResourceProps(
...
+      final String versionTag,
+      final String tenantId,
+      final long scopeKey) {
+    return resourceState
+        .findResourceByIdAndVersionTag(resourceId, versionTag, tenantId)
+        .<Either<Failure, PersistedResource>>map(Either::right)
+        .orElseGet(
+            () ->
+                Either.left(
+                    new Failure(
```
korthout, line 252: 🔧 We have a nice helper method to construct an Either from an Optional which allows you to avoid the juggling with generics: `Either.ofOptional`

```suggestion
    return Either.ofOptional(
            resourceState.findResourceByIdAndVersionTag(resourceId, versionTag, tenantId))
        .orElse(
            new Failure(
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
```diff
@@ -159,9 +166,98 @@ private Either<Failure, List<LinkedResourceProps>> evalLinkedResourceProps(
...
-    return "";
+      final LinkedResource linkedResource, final BpmnElementContext context, final long scopeKey) {
+    return findLinkedResource(
+            linkedResource.getResourceId(),
+            linkedResource.getBindingType(),
+            linkedResource.getVersionTag(),
+            context,
+            scopeKey)
+        .map(PersistedResource::getResourceKey)
+        .get()
```
korthout, line 177: ❌ By calling `get` on the Either, the potential Left (i.e. `Failure`) is ignored. However, the calling method also wants to return Either some result or a failure, so we must ensure that we provide the failure if we have one. This is used by the callers of `BpmnJobBehavior.evaluateJobExpressions` to determine how to handle the failure cases (e.g. raise an incident at the service task).

One way to do this is to change this method to return `Either<Failure, String>` and then change how it's called. For example, you can apply an early return if a Left is encountered (alternatively you can use something like [`Either.collectorFoldingLeft()`](https://github.com/camunda/camunda/blob/baf6f7287bedab490db45ea1eeb5ffb27e69e603/zeebe/util/src/main/java/io/camunda/zeebe/util/Either.java#L194)):

```java
final var linkedResourceKey = resolveLinkedResourceKey(linkedResource, context, scopeKey);
if (linkedResourceKey.isLeft()) {
  return Either.left(linkedResourceKey.getLeft());
}
resourceProps.setResourceKey(linkedResourceKey.get());
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
```diff
@@ -159,9 +166,98 @@ private Either<Failure, List<LinkedResourceProps>> evalLinkedResourceProps(
...
+                    .orElseGet(
+                        () ->
+                            Either.left(
+                                new Failure(
+                                    String.format(
+                                        """
+                                        Expected to use a resource with id '%s' with binding type 'deployment', \
+                                        but no such resource found in the deployment with key %s which contained the current process. \
+                                        To resolve this incident, migrate the process instance to a process definition \
+                                        that is deployed together with the intended resource to use.\
```
korthout, line 214: 🔧 We can be more explicit about what was attempted

```suggestion
                                        Expected to link a resource with id '%s' and binding type 'deployment', \
                                        but no such resource found in the deployment with key %s which contained the current process. \
                                        To resolve this incident, migrate the process instance to a process definition \
                                        that is deployed together with the intended resource to use.\
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
```diff
@@ -159,9 +166,98 @@ private Either<Failure, List<LinkedResourceProps>> evalLinkedResourceProps(
...
+                    .findResourceByIdAndDeploymentKey(
+                        resourceId, deploymentKey, context.getTenantId())
+                    .<Either<Failure, PersistedResource>>map(Either::right)
+                    .orElseGet(
+                        () ->
+                            Either.left(
+                                new Failure(
+                                    String.format(
+                                        """
+                                        Expected to use a resource with id '%s' with binding type 'deployment', \
```
korthout, line 211: 🔧 Currently there are 40 chars of indentation on this text block. I recommend extracting the template strings to class level constants to declutter the code.

korthout (changes requested):
Thanks for the contribution @kzraitel 👏 

👍 You've resolved the happy path

❌ This code ignores the failure case. We need to handle this accordingly and raise an incident.

❌ There are no tests added to show the new behavior. Please add them.

korthout (changes requested):
Thanks for the changes @kzraitel 🏗️ 

👍 We're well on our way to support this now.

❌ I'm still missing tests that show the behavior of the different binding types. We only test version tag binding right now. We should also show the success and failure cases (i.e. raising an incident) on the other binding types: deployment binding and latest binding.

❌ I'm also missing test cases that show that after deploying the missing resource, the raised incident can be resolved, i.e. a job with linked resource is created at the end.

💡 Please note that for deployment binding type, you can only fix the issue after deploying both the process and the resource together, and then migrate the process instance to the new process definition, before resolving the incident.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -33,13 +36,21 @@ public class ServiceTaskTest {
...
   public final RecordingExporterTestWatcher recordingExporterTestWatcher =
       new RecordingExporterTestWatcher();
 
   @Test
   public void shouldCreateJobWithLinkedResources() throws JsonProcessingException {
+    final var record = createResourceRecord();
+    final MutableResourceState resourceState =
+        ((ProcessingDbState) ENGINE.getProcessingState()).getResourceState();
+    resourceState.storeResourceInResourceColumnFamily(record);
+    resourceState.storeResourceInResourceKeyByResourceIdAndVersionTagColumnFamily(record);
```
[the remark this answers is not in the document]
kzraitel: Fixed direct state modification

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnJobBehavior.java
```diff
@@ -142,26 +166,100 @@ public Either<Failure, JobProperties> evaluateJobExpressions(
...
+            linkedResource.getResourceId(),
+            linkedResource.getBindingType(),
+            linkedResource.getVersionTag(),
+            context,
+            scopeKey);
+    if (resource.isRight()) {
+      return Either.right(resource.map(PersistedResource::getResourceKey).get().toString());
+    } else {
+      return Either.left(resource.getLeft());
+    }
```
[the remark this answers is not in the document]
kzraitel: Fixed

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -70,14 +81,56 @@ public void shouldCreateJobWithLinkedResources() throws JsonProcessingException
...
+
+    // when
+    final long processInstanceKey = ENGINE.processInstance().ofBpmnProcessId(PROCESS_ID).create();
+
+    // then
+    assertThat(
+            RecordingExporter.jobRecords(JobIntent.CREATED)
+                .withProcessInstanceKey(processInstanceKey)
+                .exists())
+        .isFalse();
```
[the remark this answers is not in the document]
kzraitel: Added validation

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -36,20 +33,24 @@ public class ServiceTaskTest {
...
-    resourceState.storeResourceInResourceColumnFamily(record);
-    resourceState.storeResourceInResourceKeyByResourceIdAndVersionTagColumnFamily(record);
+  public void shouldCreateJobWithLinkedResourcesVersionTagBinding() throws JsonProcessingException {
+    final var rpaScript =
+        """
+          {
+          "type": "default",
+          "id": "Rpa_0w7r08e",
+          "executionPlatform": "Camunda Cloud",
+          "executionPlatformVersion": "8.1.0",
```
korthout, line 54: 🔧 As this feature releases with 8.7.0, let's adjust this to avoid confusion about supported versions

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -36,20 +33,24 @@ public class ServiceTaskTest {
...
       new RecordingExporterTestWatcher();
 
   @Test
-  public void shouldCreateJobWithLinkedResources() throws JsonProcessingException {
-    final var record = createResourceRecord();
-    final MutableResourceState resourceState =
-        ((ProcessingDbState) ENGINE.getProcessingState()).getResourceState();
-    resourceState.storeResourceInResourceColumnFamily(record);
-    resourceState.storeResourceInResourceKeyByResourceIdAndVersionTagColumnFamily(record);
+  public void shouldCreateJobWithLinkedResourcesVersionTagBinding() throws JsonProcessingException {
```
korthout, line 47: 🔧 Bonus: if we want to show that version tag binding really works, then you should deploy a similar resource with the same id but a different versionTag after deploying the one that you want to bind to. You should also deploy the resource and the process separately. This shows it's different from latest binding and different from deployment binding.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -59,16 +60,78 @@ final var record = createResourceRecord();
...
+    assertThat(resourcePropsList).hasSize(1);
+    final LinkedResourceProps resourceProps = resourcePropsList.get(0);
+    assertThat(resourceProps)
+        .hasFieldOrPropertyWithValue("resourceType", "RPA")
+        .hasFieldOrPropertyWithValue("linkName", "my_link");
+    assertThat(resourceProps.getResourceKey()).matches("\\d+");
+  }
+
+  @Test
+  public void shouldCreateJobWithLinkedResourcesLatestBinding() throws JsonProcessingException {
```
korthout, line 107: 🔧 Bonus: To show that this uses latest binding, you should deploy a second version of the same resource (use the same ID) in a separate deployment, and check that the latest is linked. (Tip: You can find the latest by using `.skip()` on the RecordingExporter stream)

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -59,16 +60,78 @@ final var record = createResourceRecord();
...
+        MAPPER.readValue(
+            jobCreated.getValue().getCustomHeaders().get("linkedResources"),
+            new TypeReference<>() {});
+
+    assertThat(resourcePropsList).hasSize(1);
+    final LinkedResourceProps resourceProps = resourcePropsList.get(0);
+    assertThat(resourceProps)
+        .hasFieldOrPropertyWithValue("resourceType", "RPA")
+        .hasFieldOrPropertyWithValue("linkName", "my_link");
+    assertThat(resourceProps.getResourceKey()).matches("\\d+");
```
korthout, line 103: ❌ You must verify that the correct resource was linked (please do this in all cases)

Try something like 
```suggestion
    assertThat(resourceProps.getResourceKey())
        .isEqualTo(
            String.valueOf(
                RecordingExporter.resourceRecords()
                    .withResourceId("Rpa_0w7r08e")
                    .withIntent(ResourceIntent.CREATED)
                    .getFirst()
                    .getKey()));
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -78,14 +141,77 @@ final var record = createResourceRecord();
...
+  }
+
+  @Test
+  public void shouldCreateJobWithLinkedResourcesDeploymentBinding() throws JsonProcessingException {
```
korthout, line 165: 🔧 Bonus: To show that this uses deployment binding, you should deploy another resource with the same id afterward. The first one should be linked.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -241,10 +245,148 @@ public void shouldHandleNotFound() {
...
+            .withProcessInstanceKey(processInstanceKey)
+            .getFirst()
+            .getValue();
+    assertThat(incidentRecordValue.getErrorType()).isEqualTo(ErrorType.RESOURCE_NOT_FOUND);
+    assertThat(incidentRecordValue.getErrorMessage())
+        .matches(
+            String.format(
+                BpmnJobBehavior.FIND_RESOURCE_BY_ID_IN_SAME_DEPLOYMENT_FAILED_MESSAGE,
+                "2",
+                "\\d+"));
```
korthout, line 296: ❌ Please ensure that the right key is used here, don't just match on any digit. Use my previous suggestion to find the resource key.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -241,10 +245,148 @@ public void shouldHandleNotFound() {
...
+        .deploy();
+    // resolve incident
+    ENGINE.incident().ofInstance(processInstanceKey).withKey(incident.getKey()).resolve();
+    // then
+    assertThat(RecordingExporter.incidentRecords().onlyEvents().limit(2))
+        .extracting(io.camunda.zeebe.protocol.record.Record::getKey, Record::getIntent)
+        .describedAs("resource not found incident is resolved and no new incident is created")
+        .containsExactly(
+            tuple(incident.getKey(), IncidentIntent.CREATED),
+            tuple(incident.getKey(), IncidentIntent.RESOLVED));
```
korthout, line 389: ❌ Please show the effect of resolving the incident as well:

Something like:

```java
    final Record<JobRecordValue> jobCreated =
        RecordingExporter.jobRecords(JobIntent.CREATED)
            .getFirst();

    assertThat(jobCreated.getValue().getCustomHeaders()).containsKey("linkedResources");
```

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -39,7 +44,19 @@ public class ServiceTaskTest {
...
+        """
+          {
+          "type": "default",
+          "id": "Rpa_0w7r08e",
+          "executionPlatform": "Camunda Cloud",
+          "executionPlatformVersion": "8.1.0",
+          "schemaVersion": 7,
+          "versionTag": "1v",
+          "resource": "Script content"
+        }""";
```
korthout, line 58: 👍 Seems like you took a real one! I like it

korthout (changes requested):
Thanks @kzraitel for the new changes

👍 I really like your expressive test case names

❌ There are a few remaining change requests. Please have a look

🔧 I've also added some bonus suggestions. They concern additional test cases that fit the described behavior (based on binding type). We should have these test cases, but @natanielstrack [had mentioned the idea of backporting some test improvements](https://camunda.slack.com/archives/C037W9NMATG/p1740052874647439?thread_ts=1739800049.028959&cid=C037W9NMATG) after the code freeze, and these could be candidates for that in my opinion. However, we need them eventually regardless.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: Fixed

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: Fixed

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: This was a nice catch! As a result I've found out that in this test I had incorrect id's

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: Done!

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: Done

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: Done!

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
```diff
@@ -100,21 +139,22 @@ public void shouldCreateJobWithLinkedResourcesVersionTagBinding() throws JsonPro
     assertThat(resourceProps)
         .hasFieldOrPropertyWithValue("resourceType", "RPA")
         .hasFieldOrPropertyWithValue("linkName", "my_link");
-    assertThat(resourceProps.getResourceKey()).matches("\\d+");
+
+    final List<Record<Resource>> resourceRecords =
+        RecordingExporter.resourceRecords()
+            .withResourceId("Rpa_0w7r08e")
+            .withIntent(ResourceIntent.CREATED)
+            .toList();
```
korthout, line 147: ❌ The stream must be limited otherwise this waits for 5 seconds. Please also apply to the other test cases where you look them up like this.

```suggestion
    final List<Record<Resource>> resourceRecords =
        RecordingExporter.resourceRecords()
            .withResourceId("Rpa_0w7r08e")
            .withIntent(ResourceIntent.CREATED)
            .limit(2)
            .toList();
```

🔧 Alternatively, you can use the returned value of the call to the `.deploy()` method above to lookup the resource metadata. You can apply this to any call to deploy. Consider this the response of the client request to deploy. The resource metadata contains the same information as the recorded `CREATED` event.

```java
final var resourceMetadata =
    ENGINE
        .deployment()
        .withJsonResource(RPA_RESOURCE_2.getBytes(StandardCharsets.UTF_8), "my_resource.rpa")
        .deploy()
        .getValue()
        .getResourceMetadata()
        .getFirst();
```

korthout (changes requested):
Nice work @kzraitel 👏 Thanks for taking the bonuses as well

❌ One last step before merging 🚧

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/bpmn/activity/ServiceTaskTest.java
kzraitel: Fixed!

korthout (approved):
Awesome 👏  Great work on this PR @kzraitel 

LGTM 💯

PR #27830, opened 2025-02-07
migrate RaftServiceMetrics & RaftStartupMetrics to micrometer

entangled90:
## Description
migrate RaftServiceMetrics & RaftStartupMetrics to micrometer. Only RaftServiceMetrics contains a timer which needs a dashboard fix.

- [x] Dashboard change

## Related issues

relates #26708

PR #27907, opened 2025-02-10
Migrate zeebe-backup to micrometer

### LINKED ISSUE #27083: Migrate all Prometheus metrics in zeebe-backup

entangled90:
## Description
Remove prometheus in favour of micrometer metrics

- [x] Dashboard changes

See PR #27830 for similar changes in other metrics 

## Related issues

closes #27083

### DIFF
--- a/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/BackupService.java
+++ b/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/BackupService.java
@@ -18,6 +18,7 @@ import io.camunda.zeebe.scheduler.Actor;
 import io.camunda.zeebe.scheduler.future.ActorFuture;
 import io.camunda.zeebe.scheduler.future.CompletableActorFuture;
 import io.camunda.zeebe.snapshots.PersistedSnapshotStore;
+import io.micrometer.core.instrument.MeterRegistry;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Optional;
@@ -44,13 +45,14 @@ public final class BackupService extends Actor implements BackupManager {
       final BackupStore backupStore,
       final PersistedSnapshotStore snapshotStore,
       final Path segmentsDirectory,
-      final JournalInfoProvider raftMetadataProvider) {
+      final JournalInfoProvider raftMetadataProvider,
+      final MeterRegistry brokerMeterRegistry) {
     this.nodeId = nodeId;
     this.partitionId = partitionId;
     this.numberOfPartitions = numberOfPartitions;
     this.snapshotStore = snapshotStore;
     this.segmentsDirectory = segmentsDirectory;
-    metrics = new BackupManagerMetrics(partitionId);
+    metrics = new BackupManagerMetrics(partitionId, brokerMeterRegistry);
     internalBackupManager = new BackupServiceImpl(backupStore);
     actorName = buildActorName("BackupService", partitionId);
     journalInfoProvider = raftMetadataProvider;
--- a/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/BackupService.java
+++ b/zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/BackupService.java
@@ -46,13 +46,13 @@ public final class BackupService extends Actor implements BackupManager {
       final PersistedSnapshotStore snapshotStore,
       final Path segmentsDirectory,
       final JournalInfoProvider raftMetadataProvider,
-      final MeterRegistry brokerMeterRegistry) {
+      final MeterRegistry partitionRegistry) {
     this.nodeId = nodeId;
     this.partitionId = partitionId;
     this.numberOfPartitions = numberOfPartitions;
     this.snapshotStore = snapshotStore;
     this.segmentsDirectory = segmentsDirectory;
-    metrics = new BackupManagerMetrics(partitionId, brokerMeterRegistry);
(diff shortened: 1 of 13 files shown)


zeebe/backup/src/main/java/io/camunda/zeebe/backup/management/BackupService.java
```diff
@@ -44,13 +45,14 @@ public BackupService(
...
-      final JournalInfoProvider raftMetadataProvider) {
+      final JournalInfoProvider raftMetadataProvider,
+      final MeterRegistry brokerMeterRegistry) {
     this.nodeId = nodeId;
     this.partitionId = partitionId;
     this.numberOfPartitions = numberOfPartitions;
     this.snapshotStore = snapshotStore;
     this.segmentsDirectory = segmentsDirectory;
-    metrics = new BackupManagerMetrics(partitionId);
+    metrics = new BackupManagerMetrics(partitionId, brokerMeterRegistry);
```
lenaschoenburg, line 55: :thought_balloon: Given that we have a partition id here, I'm 99% sure that we should use the partition scoped registry here, not the broker one.

entangled90: I was a bit hesitant to avoid changing the code too much, but it's just probably going to be simpler :+1:

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
```diff
@@ -7,102 +7,112 @@
...
-          .buckets(0.01, 0.1, 1, 10, 60, 5 * 60)
-          .register();
-
-  private final String partitionId;
-
-  public BackupManagerMetrics(final int partitionId) {
-    this.partitionId = String.valueOf(partitionId);
+
+  private final MeterRegistry registry;
+  private final Table<OperationType, OperationResult, Counter> totalOperations = Table.simple();
```
npepinpe, line 26: :wrench: If `OperationType` and `OperationResult` are both enums, then let's use `Table.ofEnum`.

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/CheckpointMetrics.java
```diff
@@ -7,57 +7,56 @@
...
-          .labelNames(LABEL_NAME_PARTITION)
-          .register();
-
-  final String partitionId;
-
-  public CheckpointMetrics(final int partitionId) {
-    this.partitionId = String.valueOf(partitionId);
+public class CheckpointMetrics {
+  private final MeterRegistry registry;
+  private final EnumMap<CheckpointMetricsResultValue, Counter> checkpointRecords =
```
npepinpe, line 21: :wrench: Same as before, let's use the least specific type for field declaration.

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
```diff
@@ -7,102 +7,112 @@
...
-    OPERATIONS_IN_PROGRESS.labels(partitionId, TAKE_OPERATION).set(0);
-    OPERATIONS_IN_PROGRESS.labels(partitionId, DELETE_OPERATION).set(0);
-    OPERATIONS_IN_PROGRESS.labels(partitionId, STATUS_OPERATION).set(0);
-    OPERATIONS_IN_PROGRESS.labels(partitionId, LIST_OPERATION).set(0);
+    for (final var operation : OperationType.values()) {
+      final var value = operationInProgress.get(operation);
+      if (value != null) {
+        value.set(0L);
+      }
+    }
```
npepinpe, line 58: :wrench: Can't we use `EnumMeter` for this?

npepinpe (approved):
:wrench: Please double check the heatmaps based on the issue I previously found, where the union of two series wouldn't result in the right value if both series returned something (but would be correct if only one did).

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
entangled90: In EnumMeter the value is 0 | 1, while here we have counters. Ideally they should be either 0 or 1 as well, but I'm not sure if we should change it

zeebe/backup/src/main/java/io/camunda/zeebe/backup/metrics/BackupManagerMetrics.java
```diff
@@ -7,102 +7,112 @@
...
-          .register();
-
-  private final String partitionId;
-
-  public BackupManagerMetrics(final int partitionId) {
-    this.partitionId = String.valueOf(partitionId);
+
+  private final MeterRegistry registry;
+  private final Table<OperationType, OperationResult, Counter> totalOperations = Table.simple();
+  private final EnumMap<OperationType, AtomicLong> operationInProgress =
```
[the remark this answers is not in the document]
entangled90: I'm generally agree with your comment, but in this case it's just an implementation detail that's not exposed to the outside. Leaving `EnumMap` may help the JIT as it's a `final class` compared to `Map` which is a mega morphic interface

PR #27911, opened 2025-02-10
ci: remove broken benchmark measurement support

lenaschoenburg:
The measurements were broken anyway so let's remove support for now. This reduces our benchmarking load a bit, especially on PR benchmarks where measurements are still tried by default.


ChrisKujawa (approved):
Thanks 👍🏼

Issue #27169, opened 2025-01-20
[EPIC] Remove number key objects in C8 API


danielkelemen:
## Introduction

With the [Return key attributes as string values in the C8 REST API by default](https://github.com/camunda/camunda/issues/25424) EPIC we added a new set of request and response objects in the OpenAPI in order to support both string and number keys.

On the long term this will cause confusion to users and also complicates the future development for us.

Number key objects should be removed in a future release.

## Concept

Remove number key objects from OpenAPI and code and switch to string entirely.

## Task Breakdown

```[tasklist]
### Tasks
- [x] Remove number key objects from OpenAPI specification
- [x] Update REST Gateway to use string key objects
- [x] Convert Longs to Strings in REST controllers
- [x] Remove Rest Gateway `LongSerializer`
- [x] Update Java Client to use string key objects on the wire
- [ ] https://github.com/camunda/camunda/pull/27814
- [ ] https://github.com/camunda/camunda/pull/27927
- [x] Document changes
- [ ] https://github.com/camunda/camunda-docs/pull/4955
- [x] Drop number key headers from controller annotations & RequestMapper
- [ ] https://github.com/camunda/camunda/pull/28588
- [ ] https://github.com/camunda/camunda/pull/28902
```

## Discussions and Decisions

String key objects = Request or response objects containing key attributes as String.
Number key objects = Request or response objects containing key attributes as Long.

As an intermediate phase, we could already add string key attribute methods to the Java Client and deprecate the existing number key methods. When gRPC methods are removed, the number methods can be removed as well potentially.

## Links

- https://github.com/camunda/camunda/issues/25424

aleksander-dytko:
I think we could remove it with 8.8. Can we mark `number-based key attributes` as deprecated with 8.7 somehow?

danielkelemen:
After dropping all the previous number key objects, we still have around 20-30 objects ending either with `*SearchQuery`, `*Response` or `*FilterRequest`.
All the new string key objects use a different naming. They end either with just `*Query`, `*Result` of `*Filter`, so it's a little inconsistency.

Should we rename these as well?

tmetzke:
> After dropping all the previous number key objects, we still have around 20-30 objects ending either with `*SearchQuery`, `*Response` or `*FilterRequest`. All the new string key objects use a different naming. They end either with just `*Query`, `*Result` of `*Filter`, so it's a little inconsistency.
> 
> Should we rename these as well?
> 
[...]

PR #27927, opened 2025-02-10
refactor: revert client to long key search

danielkelemen:
## Description

- Revert client key filter methods to long
- Convert keys to string in BasicLongProperty
- Adjust tests

## Related issues


tmetzke (approved):
👍 Code looks good to me, and the `BasicStringProperty` class is gone, so a green CI indicates all required methods have been adjusted.

PR #27944, opened 2025-02-10
Migrate log stream metrics to Micrometer for stable/8.5

npepinpe:
## Description

This PR migrates all log stream metrics to Micrometer. Nothing has changed - while there are two timers which technically have new names, we were not using them in any dashboard, as they're mostly used on demand for debugging.

## Related issues

related to #26078

### DIFF
--- a/zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/flowcontrol/AppendMetricsDoc.java
+++ b/zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/flowcontrol/AppendMetricsDoc.java
@@ -1,0 +1,211 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Zeebe Community License 1.1. You may not use this file
+ * except in compliance with the Zeebe Community License 1.1.
+ */
+package io.camunda.zeebe.logstreams.impl.flowcontrol;
+
+import io.camunda.zeebe.util.micrometer.ExtendedMeterDocumentation;
+import io.micrometer.common.docs.KeyName;
+import io.micrometer.core.instrument.Meter.Type;
+
+/** Flow control metrics for the log storage appender. */
+@SuppressWarnings("NullableProblems")
+public enum AppendMetricsDoc implements ExtendedMeterDocumentation {
+  /** Number of deferred appends due to backpressure */
+  TOTAL_DEFERRED_APPEND_COUNT {
+    @Override
+    public String getDescription() {
+      return "Number of deferred appends due to backpressure";
+    }
+
+    @Override
+    public String getName() {
+      return "zeebe.deferred.append.count.total";
+    }
+
+    @Override
+    public Type getType() {
+      return Type.COUNTER;
+    }
+  },
+  /** Number of tries to append */
+  TOTAL_APPEND_TRY_COUNT {
+    @Override
+    public String getDescription() {
+      return "Number of tries to append";
+    }
+
+    @Override
+    public String getName() {
+      return "zeebe.try.to.append.total";
+    }
+
+    @Override
+    public Type getType() {
+      return Type.COUNTER;
+    }
+  },
+  /** Current number of append inflight */
+  CURRENT_INFLIGHT {
+    @Override
+    public String getDescription() {
+      return "Current number of append inflight";
+    }
+
+    @Override
+    public String getName() {
+      return "zeebe.backpressure.inflight.append.count";
+    }
+
+    @Override
+    public Type getType() {
+      return Type.GAUGE;
+    }
+  },
+  /** Current limit for number of inflight appends */
+  CURRENT_LIMIT {
+    @Override
+    public String getDescription() {
+      return "Current limit for number of inflight appends";
+    }
+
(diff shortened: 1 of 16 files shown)


zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/flowcontrol/AppendMetricsDoc.java
```diff
@@ -0,0 +1,211 @@
...
+
+import io.camunda.zeebe.util.micrometer.ExtendedMeterDocumentation;
+import io.micrometer.common.docs.KeyName;
+import io.micrometer.core.instrument.Meter.Type;
+
+/** Flow control metrics for the log storage appender. */
+@SuppressWarnings("NullableProblems")
+public enum AppendMetricsDoc implements ExtendedMeterDocumentation {
+  /** Number of deferred appends due to backpressure */
+  TOTAL_DEFERRED_APPEND_COUNT {
```
entangled90, line 18: :wrench:  should we add the partition to the KeyName docs ?
Even though it comes from the registry, it's a field in the meter that should be documented.
I think we forgot to add it in the past few meters, but it's probably better to add it?

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/log/SequencerMetrics.java
```diff
@@ -7,58 +7,128 @@
...
+      }
+
+      @Override
+      public String getName() {
+        return "zeebe.sequencer.batch.length.bytes";
+      }
+
+      @Override
+      public String getBaseUnit() {
+        return "bytes";
```
entangled90, line 120: :x: is this "KiB"? The buckets are in KiB and the metric is recorded in KiB as well

npepinpe:
The failing Go tests is flakiness, not related.

zeebe/logstreams/src/test/java/io/camunda/zeebe/logstreams/util/SyncLogStreamBuilder.java
```diff
@@ -66,6 +67,12 @@ public SyncLogStreamBuilder withLogName(final String logName) {
     return this;
   }
 
+  @Override
+  public LogStreamBuilder withMeterRegistry(final MeterRegistry meterRegistry) {
+    delegate.withMeterRegistry(meterRegistry);
+    return null;
```
entangled90, line 73: :question: why null? shouldn't be `this` ?

zeebe/logstreams/src/test/java/io/camunda/zeebe/logstreams/impl/log/SequencerTest.java
```diff
@@ -30,7 +31,7 @@ final class SequencerTest {
   @Test
   void notifiesConsumerOnWrite() {
     // given
-    final var sequencer = new Sequencer(0, 16, new SequencerMetrics(1));
+    final var sequencer = new Sequencer(0, 16, new SequencerMetrics(new SimpleMeterRegistry()));
```
entangled90, line 34: :thought_balloon:  I usually create a registry in the tests class with `@AutoClose` or `AutoCloseResource`. Probably an overkill, but using these patterns even in test code is helpful for consistency and to learn/remember about scopes and resources

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/flowcontrol/AppendMetricsDoc.java
npepinpe: Yeah, good point.

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/log/SequencerMetrics.java
npepinpe: Good point, yes.

zeebe/logstreams/src/test/java/io/camunda/zeebe/logstreams/util/SyncLogStreamBuilder.java
npepinpe: yes :see_no_evil:

zeebe/logstreams/src/test/java/io/camunda/zeebe/logstreams/impl/log/SequencerTest.java
npepinpe: You're right, I was going a bit fast :see_no_evil:

npepinpe:
@entangled90 - please take over and apply your suggestions :+1:

entangled90:
Dashboard changes are missing for two timers, I'll add them now

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/flowcontrol/AppendMetricsDoc.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    @Override
+    public String getName() {
+      return "zeebe.log.appender.append.latency";
+    }
+
+    @Override
+    public Type getType() {
+      return Type.TIMER;
```
entangled90, line 133: Missing changes in dashboard here

entangled90 (approved):

Issue #25639, opened 2024-12-05
[Optimize] Support ad-hoc subprocesses


saig0:
Engineering DRI: @abremard 

### Context

With Camunda 8.7, we will support a new BPMN element: the ad-hoc subprocess. For Optimize, we may write tests to ensure that data from such elements can be imported and used for report evaluation.

### Acceptance Criteria

- [ ] I can see the data from an ad-hoc subprocess

### Testing Notes

Example process:

![Screenshot from 2024-12-05 10-10-03](https://github.com/user-attachments/assets/963c4cd3-24a6-4b66-8713-5cea7393fe02)

### Links

Part of https://github.com/camunda/product-hub/issues/2546.

### Breakdown

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/27968
- [ ] https://github.com/camunda/camunda/pull/28233
```

## Task Lifecycle

For managing the issue lifecycle, please use the workflow commands. You can see the available
commands by writing `/help` as a comment on this issue.

#### Engineering Review

- [ ] All code targeting the main branch has been reviewed by at least one Engineer
- [ ] The PR targeting the main branch has been approved by the reviewing Engineer
- [ ] If the API has changed, the API documentation is updated
- [ ] All other PRs (docs, controller etc.) in the breakdown have been approved

#### QA Review

- [ ] The change is implemented as described on all target environments/versions/modes
- [ ] The documentation changes are as expected

### Completion

- [x] All Review stages are successfully completed
- [x] All associated PRs are merged to the main branch(es) and stable branches
- [x] The correct version labels are applied to the issue

abremard:
I've checked regarding single app release with Mattia (see [comment](https://github.com/camunda/camunda/pull/28233#discussion_r1958987246))

Issue #25272, opened 2024-11-26
A process instance completes an ad-hoc subprocess when the last child instance is completed


saig0:
**Description**

The process instance should complete an ad-hoc subprocess when the last child instance is completed or terminated. So, the subprocess waits for the completion of all child instances, similar to regular embedded subprocesses. On completion, the ad-hoc subprocess should apply variable output mappings and invoke end execution listeners.

### Tasks
- Complete the ad-hoc subprocess when all child instances are completed
- Apply variable output mappings
- Invoke end execution listeners
- Terminate all child instances when the ad-hoc subprocess is terminated
- Verify ad-hoc subprocess with multi-instance marker
- Verify ad-hoc subprocess with compensation marker

Requires https://github.com/camunda/camunda/issues/25271.

PR #27968, opened 2025-02-11
test: support ad-hoc subprocesses

abremard:
## Description

In Camunda 8.7, a new BPMN element was introduced : the [ad-hoc subprocess](https://www.omg.org/spec/BPMN/2.0/PDF#13.2.5%20Ad-Hoc%20Sub-Process). It is supported in Zeebe but hasn't been tested in Optimize yet. This PR adds integration tests for the new element, to validate that an ad-hoc subprocess definition as well as process instances can be exported to Optimize for evaluation.

See [thread](https://camunda.slack.com/archives/CUZUYDMMG/p1739181377577979) for details on the test approach.

## Related issues

related to #25639

### DIFF
--- a/optimize/backend/pom.xml
+++ b/optimize/backend/pom.xml
@@ -421,8 +421,7 @@
     <dependency>
       <groupId>io.camunda</groupId>
       <artifactId>zeebe-bpmn-model</artifactId>
-      <version>8.6.7</version>
-      <scope>test</scope>
+      <version>8.8.0-SNAPSHOT</version>
     </dependency>
 
     <dependency>
--- a/optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
+++ b/optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
@@ -9,6 +9,8 @@ package io.camunda.optimize.service.importing.processdefinition;
 
 import static io.camunda.optimize.service.db.DatabaseConstants.ZEEBE_PROCESS_DEFINITION_INDEX_NAME;
 import static io.camunda.optimize.service.util.importing.ZeebeConstants.ZEEBE_DEFAULT_TENANT_ID;
+import static io.camunda.optimize.util.ZeebeBpmnModels.ACTIVATE_ELEMENTS;
+import static io.camunda.optimize.util.ZeebeBpmnModels.ADHOC_SUB_PROCESS;
 import static io.camunda.optimize.util.ZeebeBpmnModels.END_EVENT;
 import static io.camunda.optimize.util.ZeebeBpmnModels.SIGNAL_CATCH;
 import static io.camunda.optimize.util.ZeebeBpmnModels.SIGNAL_GATEWAY_CATCH;
@@ -20,7 +22,9 @@ import static io.camunda.optimize.util.ZeebeBpmnModels.SIGNAL_START_INT_SUB_PROC
 import static io.camunda.optimize.util.ZeebeBpmnModels.SIGNAL_START_NON_INT_SUB_PROCESS;
 import static io.camunda.optimize.util.ZeebeBpmnModels.SIGNAL_THROW;
 import static io.camunda.optimize.util.ZeebeBpmnModels.START_EVENT;
+import static io.camunda.optimize.util.ZeebeBpmnModels.TASK;
 import static io.camunda.optimize.util.ZeebeBpmnModels.USER_TASK;
+import static io.camunda.optimize.util.ZeebeBpmnModels.createAdHocSubProcess;
 import static io.camunda.optimize.util.ZeebeBpmnModels.createProcessWith83SignalEvents;
 import static io.camunda.optimize.util.ZeebeBpmnModels.createSimpleServiceTaskProcess;
 import static io.camunda.optimize.util.ZeebeBpmnModels.createSimpleUserTaskProcess;
@@ -296,6 +300,31 @@ public class ZeebeProcessDefinitionImportIT extends AbstractCCSMIT {
                         SIGNAL_INTERRUPTING_BOUNDARY,
                         SIGNAL_NON_INTERRUPTING_BOUNDARY,
                         SIGNAL_PROCESS_END));
+  }
+
+  @Test
+  public void importZeebeProcess_processContainsAdHocSubProcess() {
+    // given
+    final BpmnModelInstance adHocSubProcessModel =
+        createAdHocSubProcess(
+            "someProcess",
+            process ->
+                process.zeebeActiveElementsCollectionExpression(ACTIVATE_ELEMENTS).task(TASK));
+    zeebeExtension.deployProcess(adHocSubProcessModel).getBpmnProcessId();
+
+    // when
+    waitUntilNumberOfDefinitionsExported(1);
+    importAllZeebeEntitiesFromScratch();
+
+    // then
+    assertThat(databaseIntegrationTestExtension.getAllProcessDefinitions())
+        .singleElement()
+        .extracting(ProcessDefinitionOptimizeDto::getFlowNodeData)
+        .satisfies(
+            flowNodeDataDtos ->
+                assertThat(flowNodeDataDtos)
+                    .extracting(FlowNodeDataDto::getId)
+                    .contains(START_EVENT, ADHOC_SUB_PROCESS, TASK, END_EVENT));
   }
 
   // Test backwards compatibility for default tenantID applied when importing records pre multi
--- a/optimize/backend/src/main/java/io/camunda/optimize/service/util/BpmnModelUtil.java
+++ b/optimize/backend/src/main/java/io/camunda/optimize/service/util/BpmnModelUtil.java
@@ -9,6 +9,13 @@ package io.camunda.optimize.service.util;
 
 import io.camunda.optimize.dto.optimize.FlowNodeDataDto;
 import io.camunda.optimize.service.exceptions.OptimizeRuntimeException;
+import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
+import io.camunda.zeebe.model.bpmn.instance.BaseElement;
+import io.camunda.zeebe.model.bpmn.instance.FlowNode;
+import io.camunda.zeebe.model.bpmn.instance.Process;
+import io.camunda.zeebe.model.bpmn.instance.SubProcess;
+import io.camunda.zeebe.model.bpmn.instance.UserTask;
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -24,13 +31,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.camunda.bpm.model.bpmn.Bpmn;
-import org.camunda.bpm.model.bpmn.BpmnModelInstance;
-import org.camunda.bpm.model.bpmn.instance.BaseElement;
-import org.camunda.bpm.model.bpmn.instance.FlowNode;
-import org.camunda.bpm.model.bpmn.instance.Process;
-import org.camunda.bpm.model.bpmn.instance.SubProcess;
-import org.camunda.bpm.model.bpmn.instance.UserTask;
 import org.slf4j.Logger;
 
 public final class BpmnModelUtil {
--- a/optimize/backend/src/test/java/io/camunda/optimize/service/util/BpmnModelUtilTest.java
+++ b/optimize/backend/src/test/java/io/camunda/optimize/service/util/BpmnModelUtilTest.java
@@ -18,12 +18,12 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import io.camunda.optimize.dto.optimize.FlowNodeDataDto;
 import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.assertj.core.groups.Tuple;
-import org.camunda.bpm.model.bpmn.BpmnModelInstance;
 import org.junit.jupiter.api.Test;
 
 public class BpmnModelUtilTest {
--- a/optimize/pom.xml
+++ b/optimize/pom.xml
@@ -63,7 +63,7 @@
     <zeebe.version>8.6.0</zeebe.version>
     <identity.version>8.6.0</identity.version>
     <!-- We use this for the Zeebe test container version only -->
-    <zeebe.docker.version>${zeebe.version}</zeebe.docker.version>
+    <zeebe.docker.version>8.7.0-alpha3</zeebe.docker.version>
 
     <!-- maven plugins -->
     <surefire-plugin.version>3.5.2</surefire-plugin.version>
--- a/optimize/pom.xml
+++ b/optimize/pom.xml
@@ -63,7 +63,7 @@
     <zeebe.version>8.6.0</zeebe.version>
     <identity.version>8.6.0</identity.version>
     <!-- We use this for the Zeebe test container version only -->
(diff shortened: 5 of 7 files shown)


optimize/backend/pom.xml
```diff
@@ -421,8 +421,7 @@
     <dependency>
       <groupId>io.camunda</groupId>
       <artifactId>zeebe-bpmn-model</artifactId>
-      <version>8.6.7</version>
-      <scope>test</scope>
+      <version>8.8.0-SNAPSHOT</version>
```
abremard, line 424: The ad-hoc subprocess is only supported in Camunda 8.7 and above, let me know if there's a better way to set this version.

optimize/backend/src/main/java/io/camunda/optimize/service/util/BpmnModelUtil.java
```diff
@@ -9,6 +9,13 @@
 
 import io.camunda.optimize.dto.optimize.FlowNodeDataDto;
 import io.camunda.optimize.service.exceptions.OptimizeRuntimeException;
+import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
+import io.camunda.zeebe.model.bpmn.instance.BaseElement;
+import io.camunda.zeebe.model.bpmn.instance.FlowNode;
+import io.camunda.zeebe.model.bpmn.instance.Process;
+import io.camunda.zeebe.model.bpmn.instance.SubProcess;
+import io.camunda.zeebe.model.bpmn.instance.UserTask;
```
abremard, line 18: I had to migrate from `org.camunda.bpmn` to `io.camunda.zeebe` for bpmn imports because `AdHocSubprocess` element does not exist in `org.camunda.bpm.model.bpmn.instance` (Camunda 7) but exists in io.camunda.zeebe.model.bpmn.instance (Camunda 8). 

This is needed specifically inside [extractFlowNodeData method](https://github.com/camunda/camunda/blob/ca6adf6427a4fd990b7c1cfe588ff27f364b914c/optimize/backend/src/main/java/io/camunda/optimize/service/util/BpmnModelUtil.java#L52-L54) where an `AdHocSubprocess` would be parsed as a FlowNode in `io.camunda.zeebe` (because `AdHocSubprocess` extends `Subprocess` which extends `Activity` which extends `FlowNode`). In the incorrect case, it would not be recognized as an known element and would be "skipped".

optimize/backend/src/test/java/io/camunda/optimize/service/util/BpmnModelUtilTest.java
```diff
@@ -18,12 +18,12 @@
 
 import io.camunda.optimize.dto.optimize.FlowNodeDataDto;
 import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
```
abremard, line 21: This is needed for similar reasons as stated in the comment above

optimize/pom.xml
```diff
@@ -63,7 +63,7 @@
     <zeebe.version>8.6.0</zeebe.version>
     <identity.version>8.6.0</identity.version>
     <!-- We use this for the Zeebe test container version only -->
-    <zeebe.docker.version>${zeebe.version}</zeebe.docker.version>
+    <zeebe.docker.version>8.7.0-alpha3</zeebe.docker.version>
```
abremard, line 66: `zeebe.version` (see above) is set to 8.6.0 which doesn't support the ad-hoc subprocess, let me know if there's a better way to set this

optimize/backend/pom.xml
RomanJRW: Could we use the `8.7.0-alpha4` version here? Using the SNAPSHOT introduces the risk of accidentally breaking things

My assumption is also that this will go away with the single application progress

optimize/backend/src/main/java/io/camunda/optimize/service/util/BpmnModelUtil.java
RomanJRW: That sounds reasonable to me, thanks for sharing the reasoning here 👍

optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
```diff
@@ -298,6 +302,31 @@ public void importZeebeProcess_processContainsNewBpmnElementsIntroducedWith830()
...
+
+    // then
+    assertThat(databaseIntegrationTestExtension.getAllProcessDefinitions())
+        .singleElement()
+        .extracting(ProcessDefinitionOptimizeDto::getFlowNodeData)
+        .satisfies(
+            flowNodeDataDtos ->
+                assertThat(flowNodeDataDtos)
+                    .extracting(FlowNodeDataDto::getId)
+                    .contains(START_EVENT, ADHOC_SUB_PROCESS, TASK, END_EVENT));
```
RomanJRW, line 327: have you checked how this renders in the Optimize UI? For example: does the heatmap look as you might expect?

optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
```diff
@@ -298,6 +302,31 @@ public void importZeebeProcess_processContainsNewBpmnElementsIntroducedWith830()
...
   }
 
+  @Test
+  public void importZeebeProcess_processContainsAdHocSubProcess() {
+    // given
+    final BpmnModelInstance adHocSubProcessModel =
+        createAdHocSubProcess(
+            "someProcess",
+            process ->
+                process.zeebeActiveElementsCollectionExpression(ACTIVATE_ELEMENTS).task(TASK));
```
RomanJRW, line 312: can you explain to me the purpose of this line? I am not sure I follow what this task-related behaviour is

optimize/pom.xml
RomanJRW: hmm ok. Maybe we could use the `alpha4` though?

RomanJRW: we could also consider having a follow up issue to change the `zeebe.version` property to 8.7 after the 8.7 release in April

RomanJRW (changes requested):
Great work, and congrats on taking on a task relating to a feature request. You did a nice job of not only making the changes, but explaining why each change was necessary 💪 

I have left some comments/questions with a change request. Generally, I think the approach makes a lot of sense, I just want to ensure we have clarity around the UI behaviour and also the versioning we use in our poms before we look to getting this merged 🙂

optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
abremard: The line adds a task inside of the ad-hoc subprocess :
```
    <adHocSubProcess id="adhoc" name="adhoc">
      <extensionElements>
        <ns0:adHoc activeElementsCollection="=activateElements"/>
      </extensionElements>
      <incoming>sequenceFlow_d594f747-3ac8-43cc-b9b4-99bde1267e21</incoming>
      <outgoing>sequenceFlow_f5c3e393-235f-4aca-ad5c-392754b382c3</outgoing>
      <task id="A" name="A"/>
    </adHocSubProcess>
```

Based on the [bpmn specifications](https://www.omg.org/spec/BPMN/2.0/PDF#13.2.5%20Ad-Hoc%20Sub-Process), [8.8 docs](https://docs.camunda.io/docs/next/components/modeler/bpmn/ad-hoc/) and this [issue](https://github.com/camunda/camunda/issues/25272), I was trying to simulate :
- Task A completion (`zeebeExtension.startProcessInstanceWithVariables(
        processId, Map.of("activateElements", List.of("A")));`)
- Check that the ad-hoc sub-process completes as a result

The `activeElementsCollection` attribute essentially allows the execution of child elements based on variables that are being passed (in this case I specifically request execution of `Task A`)

optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
abremard: I haven't checked that, I will do this now

abremard: I can't deploy or run the process through the modeler right now because 
```
- ERROR: Elements of type 'AdHocSubProcess' are currently not supported. Please refer to the documentation for a list of supported elements: https://docs.camunda.io/docs/components/modeler/bpmn/bpmn-coverage/
```
@RomanJRW do you know how can I download the latest non-public version of the modeler? Currently the highest version I can get is 8.6

In the meantime, I'll try to import this through "customer_onboarding" jsons. It's hacky but should give us some visibility on the UI

optimize/backend/pom.xml
abremard: I tried setting it to `8.7.0-alpha4` but it raises a dependency convergence error
```
[ERROR] Dependency convergence error for io.camunda:zeebe-bpmn-model:jar:8.8.0-SNAPSHOT paths to dependency are:
[ERROR] +-io.camunda.optimize:optimize-backend:jar:8.8.0-SNAPSHOT
[ERROR]   +-io.camunda:camunda-search-client-connect:jar:8.8.0-SNAPSHOT:compile
[ERROR]     +-io.camunda:zeebe-util:jar:8.8.0-SNAPSHOT:compile
[ERROR]       +-io.camunda:zeebe-bpmn-model:jar:8.8.0-SNAPSHOT:compile
[ERROR] and
[ERROR] +-io.camunda.optimize:optimize-backend:jar:8.8.0-SNAPSHOT
[ERROR]   +-io.camunda:zeebe-bpmn-model:jar:8.7.0-alpha4:compile
[ERROR] 
```

When setting it to `${project.version}` or `SNAPSHOT`, the github actions are also working, should we keep it this way 🤔

optimize/pom.xml
abremard: Setting this to `8.7.0-alpha4` works, I will create a follow-up issue to set it to 8.7

optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
abremard: I'm able to render the ad-hoc subprocess bpmn with incident heatmap as an example. To be clear, I've modified the `customer_onboarding` jsons locally and added some random flow node instances [from here](https://camunda.slack.com/archives/CUZUYDMMG/p1739258040770239?thread_ts=1739181377.577979&cid=CUZUYDMMG). 

Screenshot 2568-02-13 at 12 34 21

RomanJRW: I don't know the answer, but nice work around! 👍 

For the flow node instance, is it possible for the adhoc subprocess element itself to have data? Or is it rather a wrapper around other nodes (like in your picture above). This is more a functional question than a technical one, I have not taken the time to look properly into what adhoc subprocesses are

optimize/backend/pom.xml
RomanJRW: Hmm ok, let's keep the SNAPSHOT then I guess. What would be ideal is that we updating it to 8.7.0 once that Zeebe release is complete. However, we would need a reminder in order to do so basically 🤔 
Can you think of a good way to track this so that we change it before conducting the Optimize 8.7.0 release?

optimize/pom.xml
RomanJRW: Ah I just saw this comment. The trick/challenge is doing it after the Zeebe 8.7.0 release, but before the Optimize 8.7.0

optimize/backend/pom.xml
abremard: > Can you think of a good way to track this so that we change it before conducting the Optimize 8.7.0 release?

2 ideas come to mind :
- Setup a webhook using OptimizeInfo slack bot and try to track Zeebe release. I don't know how feasible and how reliable would this be. Also, it could take more time than necessary (unless there's some setup that exists in Camunda for this)
- Setup a naive weekly reminder in `#optimize-status` channel (that could tag `@optimize-release-manager` since they're DRI on Optimize 8.7.0 release), this will be noisy but easier to implement. We can maybe anticipate Zeebe's release cadence and set the reminder frequency accordingly.

In both cases, the slack message would be linked to an issue describing the update needed.

optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
abremard: > For the flow node instance, is it possible for the adhoc subprocess element itself to have data?

Because adhoc subprocesses are built on top of subprocesses which are built on top of flow nodes, it's technically possible to have them contain data. However, based on the [bpmn specifications](https://www.omg.org/spec/BPMN/2.0/PDF#13.2.5%20Ad-Hoc%20Sub-Process), [8.8 docs](https://docs.camunda.io/docs/next/components/modeler/bpmn/ad-hoc/) and the few examples I've seen reading around the linked issues, adhoc subprocesses are mainly used to wrap task/activities and have them executed in a more flexible way than normal subprocesses.

Here's a diagram that might be clearer :
ad-hoc-subprocess-activation-8f493e153c948b8f06d2bec5a40e1146

optimize/backend/pom.xml
RomanJRW: We actually already get notified about Zeebe releases ([example](https://camunda.slack.com/archives/C03NFMH4KC6/p1739364622190789)). The challenge is really remembering after that notification that we also need to update the version used in this pom

optimize/backend/src/it/java/io/camunda/optimize/service/importing/processdefinition/ZeebeProcessDefinitionImportIT.java
RomanJRW: Nice, and thanks for the thorough explanation. In that case, I think the behaviour you have implemented here looks correct 🙂

optimize/backend/pom.xml
abremard: Looking at some github workflows, I think rather than consuming the slack notification we could have a workflow that :
1. Parses Zeebe version from Artifactory/JFrog (kind of like [this step](https://github.com/camunda/camunda/blob/c9b3c83cb5fa596a47ec240547d7ee5d3660fdac/.github/workflows/optimize-zeebe-compatibility.yml#L48-L56) with the difference being we check `zeebe-bpmn-model` release instead of `zeebe-protocol`).
2. Sends a slack notification (we can use [this example](https://github.com/camunda/camunda/blob/470044bccc4af502dd4786507b98cba982ece118/.github/workflows/dispatch-release-8-7.yaml#L57) as a reference), we can send the issue link inside the message. We can also note in the issue that we'll delete the github workflow once the update is done.

Let me know if this works, and I'll open a separate PR for the workflow.

RomanJRW: Nice that you thought of a solution to this problem 💪 

My general thought is that your proposal would actually work, and it definitely would give us better visibility than we currently have 🤔 

On the other hand, I am thinking that this problem _probably_ goes away by itself over the coming months anyway. Here I mean that we are planning on delivering Optimize in the Single Application. As part of that, we can probably use this dependency directly rather than pinning a version here. But this is absolutely based on an assumption rather than concretely knowing that this will be the case.

My suggestion would be to proceed with how you have it, and then sync with @buccarel next week when he returns to see if my above assumption is correct. If it is not correct, then we can revisit whether we need to be proactive with a plan to solve this problem (for example, as you propose above).

Does that make sense to you?

abremard: That makes sense, I wrote a reminder to myself to see with Mattia when he's back 👍

RomanJRW (approved):
Based on the above discussions, happy to approve and trust you to take the appropriate action to understand what changes need making with respect to the single application 🙂

PR #27981, opened 2025-02-11
feat: add new command processor for fetching resources

### LINKED ISSUE #24773: Process `ResourceIntent.FETCH` command
### Description
Zeebe should support a new "fetch resource" command that can be used to retrieve deployed resources from the Zeebe state. This will be used by the new REST API endpoint for fetching RPA resources (https://github.com/camunda/camunda/issues/25281).

### Acceptance Criteria
- [x] A new command processor is added that allows to fetch a resource by resource key.
- [x] If a resource with the given key exists in the resource state, the corresponding resource record is written to the response.
  - _Note_: A look-up from the state requires both `resourceKey` and `tenantId`. The user's `authorized_tenants` should be used to get the relevant tenant IDs (similar to how it's done for resource deletion). This ensures that a resource is only returned if the user is authorized to access it.
- [x] If no resource can be found, the command is rejected with type `NOT_FOUND` and an appropriate error message.

### Dependencies
- required by #25281

jfriedenstab:
## Description
Adds a new command processor that allows to fetch a deployed (RPA) resource from the Zeebe state by resource key.

**Notes**:
- ~The issue title suggested to add a new enum constant `FETCH` to the existing `ResourceIntent`. However, I figured it would be more appropriate to have a dedicated `ResourceFetchIntent.FETCH`, to ensure a unique mapping between intent, value type (`ValueType.RESOURCE_FETCH`) and record (`ResourceFetchRecord`) – similar to how it's done for resource deletion.~
  - see https://github.com/camunda/camunda/pull/27981#pullrequestreview-2615807004
- I used `stable/8.7` as the base branch because the overall RPA feature needs to be ready for the 8.7 release. So it has higher priority to get the changes included there (they will be forward-ported to `main`).

## Related issues
closes #24773

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceFetchProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceFetchProcessor.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.engine.processing.resource;
+
+import io.camunda.zeebe.auth.impl.Authorization;
+import io.camunda.zeebe.engine.processing.streamprocessor.TypedRecordProcessor;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.StateWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedRejectionWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.TypedResponseWriter;
+import io.camunda.zeebe.engine.processing.streamprocessor.writers.Writers;
+import io.camunda.zeebe.engine.state.deployment.PersistedResource;
+import io.camunda.zeebe.engine.state.immutable.ProcessingState;
+import io.camunda.zeebe.engine.state.immutable.ResourceState;
+import io.camunda.zeebe.protocol.impl.record.value.deployment.ResourceRecord;
+import io.camunda.zeebe.protocol.record.RejectionType;
+import io.camunda.zeebe.protocol.record.intent.ResourceIntent;
+import io.camunda.zeebe.stream.api.records.TypedRecord;
+import io.camunda.zeebe.util.buffer.BufferUtil;
+import java.util.List;
+
+public class ResourceFetchProcessor implements TypedRecordProcessor<ResourceRecord> {
+
+  private final TypedResponseWriter responseWriter;
+  private final TypedRejectionWriter rejectionWriter;
+  private final StateWriter stateWriter;
+  private final ResourceState resourceState;
+
+  public ResourceFetchProcessor(final Writers writers, final ProcessingState processingState) {
+    responseWriter = writers.response();
+    rejectionWriter = writers.rejection();
+    stateWriter = writers.state();
+    resourceState = processingState.getResourceState();
+  }
+
+  @Override
+  public void processRecord(final TypedRecord<ResourceRecord> command) {
+    final var resourceKey = command.getValue().getResourceKey();
+    for (final var tenantId : getAuthorizedTenants(command)) {
+      final var optionalResource = resourceState.findResourceByKey(resourceKey, tenantId);
+      if (optionalResource.isPresent()) {
+        final var record = asResourceRecord(optionalResource.get());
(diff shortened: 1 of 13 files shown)


berkaycanbc (changes requested):
@jfriedenstab Thanks for the PR and clear PR description. ⭐ 

❌ I see a new record is created for fetching resources however the implementation can be simplified by using existing `ResourceRecord` and `ResourceIntent`.

A new record and intent is created for resource deletion. But that serves a purpose of having a single endpoint to delete different kind of resources (process definitions, decision definitions or forms). In our case, we would like to implement fetching resources from the state as a temporary solution (related Slack [thread](https://camunda.slack.com/archives/C07DA2WBMAL/p1737557015819819?thread_ts=1737388797.061439&cid=C07DA2WBMAL)). Therefore, it will not be needed for other resource types, so we do not need a generic implementation for fetching resources but replace it with the Exporter Query API in 8.8.

👍 Processor implementation and tests LGTM but I will still do a final review once the implementation is simplified.

jfriedenstab:
Thanks for the review, @berkaycanbc! I refactored the code now so that the existing `ResourceRecord` and `ResourceIntent` will be used (I needed to add some default values to the `ResourceRecord` to make it work).

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/ResourceFetchClient.java
```diff
@@ -0,0 +1,55 @@
...
+    this.requestStreamId = requestStreamId;
+    return this;
+  }
+
+  public ResourceFetchClient withRequestId(final long requestId) {
+    this.requestId = requestId;
+    return this;
+  }
+
+  public void fetch() {
```
berkaycanbc, line 47: 🔧  Please apply success and rejection expectation after writing the command (you can take a look at `ResourceDeletionClient` for an example). That way, we make sure the expected follow up event is written as well. It also helps us to prevent writing additional assertion each time.

berkaycanbc (approved):
Thanks @jfriedenstab! 🚀 

🔧 I added one suggestion for the test client, please take a look.

I will pre-approve the PR since another review is not needed once you apply my suggestion. LGTM 👍

zeebe/engine/src/test/java/io/camunda/zeebe/engine/util/client/ResourceFetchClient.java
jfriedenstab: Hi @berkaycanbc, in the success case there is no follow-up event because the record is just written to the response. Do you think it makes sense to add such an event?

berkaycanbc: @jfriedenstab yes, we need to write a follow up event as well to export the result of the command as an event.

jfriedenstab: Ok, let me quickly add that.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceFetchProcessor.java
```diff
@@ -42,13 +44,8 @@ public void processRecord(final TypedRecord<ResourceRecord> command) {
...
       if (optionalResource.isPresent()) {
         final var record = asResourceRecord(optionalResource.get());
-        responseWriter.writeResponse(
-            resourceKey,
-            ResourceIntent.FETCHED,
-            record,
-            ValueType.RESOURCE,
-            command.getRequestId(),
-            command.getRequestStreamId());
+        stateWriter.appendFollowUpEvent(resourceKey, ResourceIntent.FETCHED, record);
```
jfriedenstab, line 47: One thing I was unsure about: Is it okay to use the `resourceKey` as the event key? I've checked how it's done at other places in the code and sometimes a new key is generated for the follow-up event (but not always).

jfriedenstab:
@berkaycanbc Sorry for the delay, but I got sick on Monday and was out for two days. I added two more commits – could you have another look? Please also see my [question](https://github.com/camunda/camunda/pull/27981#discussion_r1963273467) above.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/resource/ResourceFetchProcessor.java
berkaycanbc: @jfriedenstab yes, it is okay to use resource key. We generate new keys for the new records that is going to be created as a result of processing. Additionally, when we would like to distribute a record, we generate a new key to propagate that key as the unique key of the event.

berkaycanbc (approved):
@jfriedenstab thanks for latest changes, LGTM! 🚀

PR #27986, opened 2025-02-11
migrate TopologyMetrics to micrometer

### LINKED ISSUE #27085: Migrate all Prometheus metrics in zeebe-cluster-config
**Description**

Migrate all Prometheus metrics in the zeebe-cluster-config module.

entangled90:
## Description
migrate TopologyMetrics to micrometer.
- [x] Dashboard changes 

## Related issues

closes #27085


npepinpe (approved):
Generally looks good, but as we said, 

:x: let's split the metrics class between gossiper and manager, since they are different actors.

No need for another review.

PR #27998, opened 2025-02-11
fix: CVE 2024 57699 [Backport/stable 8.6]

### LINKED ISSUE #27989: fix: CVE-2024-57699
(fix: CVE-2024-57699) 
Present in all stable versions of Tasklist
8.3 to 8.6

nathansandi:
## Description

More info https://nvd.nist.gov/vuln/detail/CVE-2024-57699

**Is this a vulnerability in Camunda? Why?**
Yes, this is a potential vulnerability in Camunda due to the usage of the json-smart library within its dependencies, which are susceptible to stack exhaustion when processing specially crafted JSON inputs. If Camunda relies on these vulnerable versions for JSON processing, attackers could potentially exploit this to cause a Denial of Service (DoS).
**Impact: What does it allow an attacker to do? (Required before prioritization; Completed by Camunda):**
This vulnerability allows an attacker to craft malicious JSON payloads containing a large number of { characters, which can trigger stack exhaustion in the JSON parser. This can lead to a Denial of Service (DoS), causing the affected component or service
**Risk: How likely is it to be exploited? (Required before prioritization; Completed by Camunda):**
the risk is lower. However, given the simplicity of the attack (sending a crafted JSON payload), the likelihood is moderate to high in exposed environments.

to validate the correct version is used now
`mvn dependency:tree `

closes #27989

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -1866,7 +1866,7 @@
       <dependency>
         <groupId>net.minidev</groupId>
         <artifactId>json-smart</artifactId>
-        <version>2.5.1</version>
+        <version>2.4.11</version>
       </dependency>
 
       <!-- needed for tests in elasticsearch-exporter -->


zeebe/gateway-rest/pom.xml
```diff
@@ -254,6 +254,19 @@
       <groupId>com.jayway.jsonpath</groupId>
       <artifactId>json-path</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>net.minidev</groupId>
```
nathansandi, line 259: @RomanJRW I have excluded the libraries that use the dependency 2.5.1 and when is necessary declared the 2.4.11
Please let me know if I should only ignore this case, so I can revert and use only the `parent/pom.xml`

To validate this, it can be used `mvn dependency:tree`

RomanJRW (approved):

RomanJRW (changes requested):
I don't think the renovate file change is needed here for two reasons:
- I believe the `main` renovate config is the only one that is respected and that the branch rules are specified there
- We already have a package rule that limits updates to stable branches to just patch updates (which we would still want in this case)

RomanJRW (approved):

PR #28011, opened 2025-02-11
fix: non created user task intents should be routed in the old fashion

panagiotisgts:
## Description

Non created intents on User Task records should be persisted and routed in the old way to match previous version User Task ids, when an 8.8 `CREATED` intent has not been yet ingested by the Exporter.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskHandlerTest.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskHandlerTest.java
@@ -636,11 +636,12 @@ public class UserTaskHandlerTest {
             ValueType.USER_TASK,
             r ->
                 r.withIntent(UserTaskIntent.MIGRATED)
+                    .withKey(111)
                     .withValue(taskRecordValue)
                     .withTimestamp(System.currentTimeMillis()));
 
     // when
-    final TaskEntity taskEntity = new TaskEntity().setId(String.valueOf(123L));
+    final TaskEntity taskEntity = new TaskEntity().setId(String.valueOf(111L));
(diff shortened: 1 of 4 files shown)


zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskHandlerTest.java
```diff
@@ -636,11 +636,12 @@ void flushedEntityShouldContainMigratedData() {
             ValueType.USER_TASK,
             r ->
                 r.withIntent(UserTaskIntent.MIGRATED)
+                    .withKey(111)
                     .withValue(taskRecordValue)
                     .withTimestamp(System.currentTimeMillis()));
 
     // when
-    final TaskEntity taskEntity = new TaskEntity().setId(String.valueOf(123L));
+    final TaskEntity taskEntity = new TaskEntity().setId(String.valueOf(111L));
```
ChrisKujawa, line 644: ❓ Ah we didn't detected it because our test data was not good enough? We used the same keys right?

ChrisKujawa (approved):
Thanks 👍🏼 Good catch. 

I reverted the fix and tests are failing now, that is good 👍🏼 Thanks.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/UserTaskHandlerTest.java
panagiotisgts: Technically this was missed from multiple sides because we are doing upserts, so the update will go through and create a new record, which is then retrievable.

PR #28017, opened 2025-02-11
ci: properly push tags and version updates on optimize release

matthewBearCamunda:
## Description

Optimize is not properly updating its version in pom.xml file when a release is run. This has lead to having our 8.6.5 release have 8.6.4 versions in its pom.xml

I noticed that `DpushChanges` and `DremoteTagging` were always `false` when the release job was run. I updated the script to properly set these options when the job is run

Some context for this problem and how it was found: https://camunda.slack.com/archives/C071KP5BTHB/p1739179242814939

This is also aligning `main` and `stable` branches. `stable` was using the `mvn` command to do the tagging, whereas `main` was not. The change to `stable` was made to work around some `git` issues, and this change apparenrtly did not properly make it to `main`

https://camunda.slack.com/archives/C071KP5BTHB/p1736345765024329

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/.github/workflows/optimize-release-optimize-c8-only.yml
+++ b/.github/workflows/optimize-release-optimize-c8-only.yml
@@ -184,12 +184,6 @@ jobs:
             --fail-at-end \
             -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
 
-      - name: Push changes
-        if: env.IS_DRY_RUN == 'false'
-        run: |
-          git push --tags
-          git push --all
-
       - name: Perform release
         run: |
           ./mvnw -f optimize \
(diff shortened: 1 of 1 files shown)


.github/workflows/optimize-release-optimize-c8-only.yml
```diff
@@ -122,6 +131,7 @@ jobs:
             echo "IS_ALPHA=${{ steps.release-type.outputs.is_alpha }}"
             echo "IS_RC=${{ steps.release-type.outputs.is_rc }}"
             echo "IS_DRY_RUN=${{ github.event.inputs.IS_DRY_RUN || 'true' }}"
+            echo "SHOULD_PUSH=${{ steps.release-type.outputs.should_push }}"
```
RomanJRW, line 134: why do we need a new property rather than just reusing (and inverting) the `IS_DRY_RUN` property?

RomanJRW (changes requested):

RomanJRW (approved):
LGTM. Is there any way to test this?

matthewBearCamunda:
@RomanJRW - If there is a way to test this change on a branch, I'm unsure how to do it.

The only was I can think to test would be to merge this PR and perform a dry run release to see if the proper variables are being set on the maven command

PR #28026, opened 2025-02-11
Provide per-partition startup and transition/role meter registries

npepinpe:
## Description

This PR introduces a root, per-partition meter registry using a common partition tag, which is created on partition startup (bootstrap and join). This is distinguished from the `ZeebePartition`'s per transition meter registry, which is used when we have metrics that report the state of objects recreated during a transition.

> [!Note]
> As @entangled90 raised, there are probably nicer ways to do this, and we may not want to have a per-transition registry, but it was outside of scope here.

This is in preparation to migrating the snapshot metrics.

## Related issues

related to #26078

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/PartitionStartupAndTransitionContextImpl.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/PartitionStartupAndTransitionContextImpl.java
@@ -106,8 +106,8 @@ public class PartitionStartupAndTransitionContextImpl
   private BackupStore backupStore;
   private AdminApiRequestHandler adminApiService;
   private PartitionAdminAccess adminAccess;
-  private final MeterRegistry brokerMeterRegistry;
-  private MeterRegistry partitionMeterRegistry;
+  private final MeterRegistry startupMeterRegistry;
+  private MeterRegistry transitionMeterRegistry;
 
   public PartitionStartupAndTransitionContextImpl(
       final int nodeId,
@@ -129,7 +129,7 @@ public class PartitionStartupAndTransitionContextImpl
       final DiskSpaceUsageMonitor diskSpaceUsageMonitor,
       final AtomixServerTransport gatewayBrokerTransport,
       final TopologyManager topologyManager,
-      final MeterRegistry brokerMeterRegistry) {
+      final MeterRegistry startupMeterRegistry) {
     this.nodeId = nodeId;
     this.clusterCommunicationService = clusterCommunicationService;
     this.raftPartition = raftPartition;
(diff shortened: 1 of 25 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/PartitionStartupAndTransitionContextImpl.java
```diff
@@ -151,8 +151,10 @@ public PartitionStartupAndTransitionContextImpl(
     this.diskSpaceUsageMonitor = diskSpaceUsageMonitor;
     this.gatewayBrokerTransport = gatewayBrokerTransport;
     this.topologyManager = topologyManager;
-    this.brokerMeterRegistry = new CompositeMeterRegistry().add(brokerMeterRegistry);
-    this.brokerMeterRegistry.config().commonTags(Tags.of("partition", String.valueOf(partitionId)));
+    this.startupMeterRegistry = new CompositeMeterRegistry().add(startupMeterRegistry);
+    this.startupMeterRegistry
+        .config()
+        .commonTags(Tags.of("partition", String.valueOf(partitionId)));
```
entangled90, line 157: :wrench: should we use a static field for the partition name as it's shared by both partitionMeterRegistries? This should be the same tag as `PartitionKeyNames`

zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/partitions/impl/steps/MetricsStep.java
```diff
@@ -31,25 +31,22 @@ public ActorFuture<Void> prepareTransition(
...
+      context.setPartitionTransitionMeterRegistry(null);
     }
     return context.getConcurrencyControl().createCompletedFuture();
   }
 
   @Override
   public ActorFuture<Void> transitionTo(
       final PartitionTransitionContext context, final long term, final Role targetRole) {
-    final var brokerRegistry = context.getBrokerMeterRegistry();
+    final var brokerRegistry = context.getPartitionStartupMeterRegistry();
```
entangled90, line 42: :question: I find a bit counter-intuitive that the "brokerRegistry" already has a partition tag registered in it. Has it always been like this?

npepinpe: Oops, no, it should be renamed. I renamed the getter but not the variable :see_no_evil:

zeebe/restore/src/main/java/io/camunda/zeebe/restore/RestoreManager.java
```diff
@@ -92,20 +95,39 @@ private void logSuccessfulRestore(
...
+
+    final var registry = partition.registry();
+    return new PartitionRestoreService(
+            backupStore, raftPartition, new ChecksumProviderRocksDBImpl())
         .restore(backupId, validator)
-        .thenAccept(backup -> logSuccessfulRestore(backup, partition.id().id(), backupId));
+        .thenAccept(backup -> logSuccessfulRestore(backup, raftPartition.id().id(), backupId))
+        .whenComplete(
+            (ok, error) -> {
+              registry.clear();
```
entangled90, line 118: :thought_balloon: this three calls are popping up all over the place :-D

entangled90 (approved):

zeebe/restore/src/main/java/io/camunda/zeebe/restore/RestoreManager.java
npepinpe: Go for it and add a helper in MicrometerUtil ;)

PR #28028, opened 2025-02-11
migrate ActorMetrics to micrometer

### LINKED ISSUE #27150: Migrate all Prometheus metrics in zeebe-scheduler
**Description**

Migrate all the Prometheus metrics in the zeebe-scheduler module.

entangled90:
## Description

 Each actor gets an instance of ActorMetricsScoped which is unique per actor in order to avoid concurrency issues with metrics.

- Actor histograms were using exponential buckets, which is a missing feature in micrometer.
- [x] Dashboard changes

## Related issues

closes #27150

### DIFF
--- a/dist/src/main/java/io/camunda/zeebe/broker/ActorSchedulerConfiguration.java
+++ b/dist/src/main/java/io/camunda/zeebe/broker/ActorSchedulerConfiguration.java
@@ -13,6 +13,7 @@ import io.camunda.zeebe.broker.system.configuration.ThreadsCfg;
 import io.camunda.zeebe.scheduler.ActorScheduler;
 import io.camunda.zeebe.shared.ActorClockConfiguration;
 import io.camunda.zeebe.shared.IdleStrategyConfig.IdleStrategySupplier;
+import io.micrometer.core.instrument.MeterRegistry;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -21,12 +22,16 @@ import org.springframework.context.annotation.Configuration;
 public final class ActorSchedulerConfiguration {
   private final BrokerCfg brokerCfg;
   private final ActorClockConfiguration actorClockConfiguration;
+  private final MeterRegistry meterRegistry;
 
   @Autowired
   public ActorSchedulerConfiguration(
-      final BrokerConfiguration config, final ActorClockConfiguration actorClockConfiguration) {
+      final BrokerConfiguration config,
+      final ActorClockConfiguration actorClockConfiguration,
+      final MeterRegistry meterRegistry) {
     brokerCfg = config.config();
     this.actorClockConfiguration = actorClockConfiguration;
+    this.meterRegistry = meterRegistry;
   }
 
   @Bean(destroyMethod = "close")
@@ -42,7 +47,7 @@ public final class ActorSchedulerConfiguration {
             .setActorClock(actorClockConfiguration.getClock().orElse(null))
             .setCpuBoundActorThreadCount(cpuThreads)
             .setIoBoundActorThreadCount(ioThreads)
-            .setMetricsEnabled(metricsEnabled)
+            .setMeterRegistry(metricsEnabled ? meterRegistry : null)
             .setSchedulerName(String.format("Broker-%d", brokerCfg.getCluster().getNodeId()))
             .setIdleStrategySupplier(idleStrategySupplier)
             .build();
--- a/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/Actor.java
+++ b/zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/Actor.java
@@ -7,6 +7,7 @@
  */
 package io.camunda.zeebe.scheduler;
 
+import io.camunda.zeebe.scheduler.ActorMetrics.ActorMetricsScoped;
 import io.camunda.zeebe.scheduler.future.ActorFuture;
 import io.camunda.zeebe.util.Loggers;
 import java.time.Duration;
@@ -25,6 +26,7 @@ public abstract class Actor implements AutoCloseable, AsyncClosable, Concurrency
 
   private static final int MAX_CLOSE_TIMEOUT = 300;
   protected final ActorControl actor = new ActorControl(this);
+  ActorMetricsScoped metrics;
   private Map<String, String> context;
 
   /**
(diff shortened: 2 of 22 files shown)


zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/ActorMetricsDoc.java
```diff
@@ -0,0 +1,137 @@
...
+      return "The length of the job queue for an actor task";
+    }
+
+    @Override
+    public KeyName[] getKeyNames() {
+      return new KeyName[] {ActorMetricsKeyName.ACTOR_NAME};
+    }
+  };
+
+  public enum ActorMetricsKeyName implements KeyName {
```
npepinpe, line 123: :wrench: Please add documentation here as well.

zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/Actor.java
```diff
@@ -99,6 +101,9 @@ public void close() {
 
   @Override
   public ActorFuture<Void> closeAsync() {
+    if (metrics != null) {
+      metrics.close();
+    }
```
npepinpe, line 106: :x: We should close the metrics as the last thing the actor does. This is too early - this is when requesting the close, and can be called from a different thread. Instead, the actor task should take care of closing the metrics. Basically, we should close it when we're transition the actor lifecycle state to closed.

entangled90: I moved it to `ActorTask.onClose()`, is that the right place?

npepinpe: I think so yes :+1:

entangled90:
I still need to double check the dashboard with grizzly

entangled90:
Any idea why the base unit for "Execution Latency" is in minutes? looks like a typo (instead of millis).
I'll fix it as well

zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/ActorThread.java
```diff
@@ -103,13 +99,14 @@ private void doWork() {
     currentTask = taskScheduler.getNextTask();
 
     if (currentTask != null) {
-      final var actorName = currentTask.actor.getName();
-      try (final var timer = actorMetrics.startExecutionTimer(actorName)) {
+      final var metrics = currentTask.actor.metrics;
+      try (final var timer = metrics != null ? metrics.startExecutionTimer() : null) {
```
npepinpe, line 103: :wrench: I would prefer a noop implementation here when disabled instead of null checks. It doesn't even have to allocate, can return the same dumb noop instance every time.

zeebe/util/src/main/java/io/camunda/zeebe/util/micrometer/MicrometerUtil.java
```diff
@@ -96,6 +98,25 @@ public static void closeRegistry(final MeterRegistry registry) {
     registry.close();
   }
 
+  public static Duration[] exponentialBucketDuration(
```
npepinpe, line 101: :x: Please document what it returns, and esp. why/when you should use it.

npepinpe (changes requested):

dist/src/main/java/io/camunda/zeebe/broker/ActorSchedulerConfiguration.java
```diff
@@ -42,7 +47,7 @@ public ActorScheduler scheduler(final IdleStrategySupplier idleStrategySupplier)
             .setActorClock(actorClockConfiguration.getClock().orElse(null))
             .setCpuBoundActorThreadCount(cpuThreads)
             .setIoBoundActorThreadCount(ioThreads)
-            .setMetricsEnabled(metricsEnabled)
+            .setMeterRegistry(metricsEnabled ? meterRegistry : null)
```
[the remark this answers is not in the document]
entangled90: Some checks are done in the scheduler to see if the metrics are enabled or not, if they are not, entire steps are skipped (such as getting timestamps and so on) which is probably more eficient than using a MeterFilter, so I'm not sure tbh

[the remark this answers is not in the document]
npepinpe: MeterFilter from what I get results in noop metrics on the Micrometer side. Of course, we're still doing some caching, setting gauges, etc., so it's not free. But I think it's worth exploring in the future. As I said, out of scope here, possible follow up. We can omit writing an issue as of now.

zeebe/util/src/main/java/io/camunda/zeebe/util/micrometer/MicrometerUtil.java
entangled90: I'll add it

entangled90: :heavy_check_mark:  added

zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/ActorMetricsDoc.java
npepinpe: Do you think we shouldn't?

zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/ActorThread.java
npepinpe: I see the thumbs up but no changes? :smile:

npepinpe (approved):
Some comments were ignored? At any rate, no need for another review.

zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/ActorThread.java
entangled90: ops I did the noop implementation, but didn't hook it up here !

zeebe/scheduler/src/main/java/io/camunda/zeebe/scheduler/ActorMetricsDoc.java
entangled90: Strange, I thought I added them, let me do it.

entangled90: I added the docs to the enum cases

PR #28030, opened 2025-02-11
Refactor the multi DB setup to clean up test data correctly

ChrisKujawa:
## Description

**Challenge**: Data is not cleaned up properly, and we have a growing complexity in the extension, handling setup-related code for ES/OS.

### Without this fix (main branch)

ES/OS Exporter indices and templates are not cleaned up correctly. 

### With this refactoring and the fixes:

Context: 

 * CI Incident: [index templates are not cleaned up, causing to fail CI jobs](https://app.slack.com/client/T0PM0P1SA/C08CAP74823) 
 * Investigation showed that ES indices are not cleaned up at all
 * @igpetrov reported that [AWS Opensearch has several indices left ](https://camunda.slack.com/archives/C05QBRGMA9F/p1739276759677299)

## PR contains

This PR tried to clean this up, to overcome the above-mentioned challenge. This is done via:

 * Introduction of a new interface to encapsulate the logic of handling with
 multi db setup like: 
   * validation of connection 
   * schema creation
   * clean up data.
 * Create new implementation for ES/OS setup handling:
   * validate connection
   * validate schema creation
   * deletion of test data
 * Add additional test to verify logic
 * Several fixes are contained here as well:
   * Configure ES exporter bulk size to 1
     * To ensure that flushing happens immediately
     * This reduces potential race conditions, such that indices are earlier created
   and data is exported
   * Validate template creation, to reduce further race conditions and validate clean-up
   * Delete all data not only from Camunda Exporter but also from ES exporter
     * Make sure that we delete all indices and templates.
     * ES Exporter uses a different pattern for prefix _ instead of -. This caused us to not clean up test data correctly.

## Related issues

related #26896

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/utils/CamundaMultiDBExtension.java
@@ -32,11 +32,13 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.Callable;
 import java.util.function.Predicate;
 import java.util.stream.StreamSupport;
 import org.agrona.CloseHelper;
 import org.awaitility.Awaitility;
 import org.awaitility.core.ThrowingRunnable;
+import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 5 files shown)


entangled90 (approved):
:rocket:

PR #28038, opened 2025-02-12
test: Check usage of methods annotated with @VisibleForTesting

### LINKED ISSUE #26687: Add ArchUnit test to ensure methods annotated by `@VisibleForTesting` are called only in appropriate contexts
**Description**

We have several private or package private, internal methods in classes which are made visible primarily for testing. These are then annotated with `@VisibleForTesting` to indicate to developers that they should not be called outside of the class _except_ for testing.

We should add an ArchUnit test which verifies this on the whole code base. This means, any method annotated with `@VisibleForTesting` (from our own `zeebe-util` module) should only be called:

- In test code
- By the containing class itself

githubEshan:
## Description

Added Arch Unit test with the scope of io.camunda.zeebe.engine that verifies methods annotated with @VisibleForTesting should only be called in test code or by the containing class itself.

## Related issues

closes #26687


githubEshan:
Hi @nicpuppa, I’ve created a new pull request. Please let me know if any changes are required. Thanks!

nicpuppa (changes requested):
Hey @githubEshan, thanks for your contribution. I have some comments to share regarding your PR.

❌ As mentioned also in the [issue](https://github.com/camunda/camunda/issues/26687#issuecomment-2580925321) the validation should be done for the whole codebase, not only for Zeebe related packages. That means that we need to scan the whole codebase

❌ I'd prefer to add the new arch unit test into a separate class, so my suggestion is to create a new test class in the `zeebe-util` module. In this way the annotation and the test are in the same module

Please have a look, thanks 👍

githubEshan:
Hi @nicpuppa, I added the classpath element to the Surefire plugin to enable the test to run across the codebase. So far, I’ve tested it successfully for document/api. Should I proceed with applying this to all classpaths?

rdegamwar (approved):

nicpuppa:
Hey @githubEshan 👋 
I'm going to close this PR due to inactivity. Feel free to open a new PR and assign it to me if you find time to continue with this issue.
Thanks

PR #28046, opened 2025-02-12
feat: update treepath for migrated element instances

### LINKED ISSUE #26094: TreePath needs to be updated after PI modification/migration
## Description

We recently introduced the feature of building the treePath (an internal data structure, that allows to show Operate data correctly) as part of the ACTIVIATING records.

This allows to always build the treePath if an element is created, like PROCESS, START EVENT, TASK, etc, and publish this as part of the event. 

The Camunda Exporter can consume such, and publish the structure to ES/OS as well.

### Challenge

As we support process instance modification (inside a PI) and process instance migration (to migrate different process model version), we have to cover these case as well.

1. For PI modification we write an ACTIVATING event as well, we have to verify whether this is correctly filled in this case.
2. For PI migration we need to publish a new treePath (as for example call activities can be added in the model, changing lexicographical ordering)

Right now we don't handle these cases which would mean we would break the PI view in Operate after migration (or modification).

mustafadagher:
## Description

- Recalculates the tree-path properties for migrated process instances 
- Adjusts the migrated event appliers to update the treepath in state. No need for a new applier version because treepath fields already existed in the record, and will stay the same unless altered during migration
- ~~Migrate the elements of a called subprocess (@berkaycanbc , @korthout please review)~~
- Introduced a new intent `ANCESTOR_MIGRATED` to mark elements of a called sub-process when it has an ancestor that has been migrated. 
   - When migrating the parent process instance, new call activities might be included. Therefore, we need to adjust the tree path of the called subprocess elements accordingly.
- Update the tree-path when exporting a migrated process instances and incidents
- Updated CompactRecordLogger to include a summary for tree-path information when found. To help with debugging

## Related issues

closes #26094

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -12,6 +12,7 @@ import static io.camunda.zeebe.engine.state.immutable.IncidentState.MISSING_INCI
 
 import io.camunda.zeebe.engine.Loggers;
 import io.camunda.zeebe.engine.processing.bpmn.behavior.BpmnBehaviors;
+import io.camunda.zeebe.engine.processing.common.ElementTreePathBuilder;
 import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableFlowNode;
 import io.camunda.zeebe.engine.processing.deployment.model.element.ExecutableSequenceFlow;
 import io.camunda.zeebe.engine.processing.distribution.CommandDistributionBehavior;
@@ -305,14 +306,10 @@ public class ProcessInstanceMigrationMigrateProcessor
     requireNoConcurrentCommand(
         eventScopeInstanceState, elementInstanceState, elementInstance, processInstanceKey);
 
+    updateElementInstanceRecord(elementInstance, targetProcessDefinition, targetElementId);
+
     stateWriter.appendFollowUpEvent(
-        elementInstance.getKey(),
-        ProcessInstanceIntent.ELEMENT_MIGRATED,
-        elementInstanceRecord
-            .setProcessDefinitionKey(targetProcessDefinition.getKey())
-            .setBpmnProcessId(targetProcessDefinition.getBpmnProcessId())
-            .setVersion(targetProcessDefinition.getVersion())
-            .setElementId(targetElementId));
+        elementInstance.getKey(), ProcessInstanceIntent.ELEMENT_MIGRATED, elementInstanceRecord);
 
     final Set<ExecutableSequenceFlow> sequenceFlows =
         getSequenceFlowsToMigrate(
@@ -423,6 +420,42 @@ public class ProcessInstanceMigrationMigrateProcessor
           processInstanceKey,
           elementId);
     }
+  }
+
+  /**
+   * Updates the element instance record with the new process definition key, bpmn process id,
+   * version and recalculates the tree path.
+   *
+   * @param elementInstance the element instance to be updated
+   * @param targetProcessDefinition the new process definition
+   * @param targetElementId the new element id
+   */
+  private void updateElementInstanceRecord(
+      final ElementInstance elementInstance,
+      final DeployedProcess targetProcessDefinition,
+      final String targetElementId) {
+    final var elementInstanceRecord = elementInstance.getValue();
+
+    elementInstanceRecord
+        .setProcessDefinitionKey(targetProcessDefinition.getKey())
+        .setBpmnProcessId(targetProcessDefinition.getBpmnProcessId())
+        .setVersion(targetProcessDefinition.getVersion())
+        .setElementId(targetElementId);
+
+    // recalculating the tree path is necessary because the element id changed
+    final var elementTreePath =
+        new ElementTreePathBuilder()
+            .withElementInstanceProvider(elementInstanceState::getInstance)
+            .withCallActivityIndexProvider(processState::getFlowElement)
+            .withElementInstanceKey(elementInstance.getKey())
+            .withFlowScopeKey(elementInstance.getParentKey())
+            .withRecordValue(elementInstanceRecord)
+            .build();
+
+    elementInstanceRecord
+        .setElementInstancePath(elementTreePath.elementInstancePath())
+        .setProcessDefinitionPath(elementTreePath.processDefinitionPath())
+        .setCallingElementPath(elementTreePath.callingElementPath());
   }
 
   private void appendIncidentMigratedEvent(
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -508,7 +508,7 @@ public class ProcessInstanceMigrationMigrateProcessor
         .setCallingElementPath(elementTreePath.callingElementPath());
 
     stateWriter.appendFollowUpEvent(
-        instance.getKey(), ProcessInstanceIntent.ELEMENT_MIGRATED, elementInstanceRecord);
+        instance.getKey(), ProcessInstanceIntent.ANCESTOR_MIGRATED, elementInstanceRecord);
 
     if (elementInstanceRecord.getBpmnElementType() == BpmnElementType.CALL_ACTIVITY) {
       // found more call activities? add the called child instance to the queue to continue going
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -359,13 +359,13 @@ public class ProcessInstanceMigrationMigrateProcessor
         incidentState.getProcessInstanceIncidentKey(elementInstance.getKey());
     if (processIncidentKey != MISSING_INCIDENT) {
       appendIncidentMigratedEvent(
-          processIncidentKey, targetProcessDefinition, targetElementId, processInstanceKey);
+          processIncidentKey, targetProcessDefinition, targetElementId, elementInstanceRecord);
     }
 
     final var jobIncidentKey = incidentState.getJobIncidentKey(elementInstance.getJobKey());
     if (jobIncidentKey != MISSING_INCIDENT) {
       appendIncidentMigratedEvent(
-          jobIncidentKey, targetProcessDefinition, targetElementId, processInstanceKey);
+          jobIncidentKey, targetProcessDefinition, targetElementId, elementInstanceRecord);
     }
 
     if (elementInstance.getUserTaskKey() > 0) {
@@ -523,7 +523,7 @@ public class ProcessInstanceMigrationMigrateProcessor
       final long incidentKey,
       final DeployedProcess targetProcessDefinition,
       final String targetElementId,
-      final long processInstanceKey) {
+      final ProcessInstanceRecord elementInstanceRecord) {
     final var incidentRecord = incidentState.getIncidentRecord(incidentKey);
     if (incidentRecord == null) {
       throw new SafetyCheckFailedException(
@@ -532,7 +532,7 @@ public class ProcessInstanceMigrationMigrateProcessor
               Expected to migrate a user task for process instance with key '%d', \
               but could not find incident with key '%d'. \
               Please report this as a bug""",
-              processInstanceKey, incidentKey));
+              elementInstanceRecord.getProcessInstanceKey(), incidentKey));
     }
     stateWriter.appendFollowUpEvent(
         incidentKey,
@@ -540,7 +540,10 @@ public class ProcessInstanceMigrationMigrateProcessor
         incidentRecord
             .setProcessDefinitionKey(targetProcessDefinition.getKey())
             .setBpmnProcessId(targetProcessDefinition.getBpmnProcessId())
-            .setElementId(BufferUtil.wrapString(targetElementId)));
+            .setElementId(BufferUtil.wrapString(targetElementId))
+            .setElementInstancePath(elementInstanceRecord.getElementInstancePath())
+            .setProcessDefinitionPath(elementInstanceRecord.getProcessDefinitionPath())
+            .setCallingElementPath(elementInstanceRecord.getCallingElementPath()));
   }
 
   private Set<ExecutableSequenceFlow> getSequenceFlowsToMigrate(
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
@@ -420,6 +420,10 @@ public class ProcessInstanceMigrationMigrateProcessor
           processInstanceKey,
           elementId);
     }
+
+    if (elementInstanceRecord.getBpmnElementType() == BpmnElementType.CALL_ACTIVITY) {
+      migrateCalledSubProcessElements(elementInstance.getCalledChildInstanceKey());
+    }
   }
 
   /**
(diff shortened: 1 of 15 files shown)


zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateProcessInstanceTest.java
```diff
@@ -133,7 +134,12 @@ public void shouldWriteElementMigratedEventForProcessInstance() {
         .hasBpmnProcessId(otherProcessId)
         .hasElementId(otherProcessId)
         .describedAs("Expect that version number did not change")
-        .hasVersion(1);
+        .hasVersion(1)
+        .hasElementInstancePath(List.of(processInstanceKey))
+        .describedAs(
+            "Expect that process definition path changed to contain new process definition key")
+        .hasProcessDefinitionPath(List.of(otherProcessDefinitionKey))
+        .hasCallingElementPath(List.of());
```
ChrisKujawa, line 142: ❓ Do we have a test with call activities? So we can validate this works also when we change the path or depth here?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -466,6 +470,63 @@ private void updateElementInstanceRecord(
...
 
+  /**
+   * Migrates the elements of a called subprocess.
+   *
+   * <p>When migrating the parent process instance, new call activities might be included.
+   * Therefore, we need to adjust the tree path of the called subprocess elements accordingly.
+   *
+   * @param calledChildInstanceKey the key of the called subprocess instance
+   */
+  private void migrateCalledSubProcessElements(final long calledChildInstanceKey) {
```
ChrisKujawa, line 481: ❓ I'm not much into this migration topics, so please bare with me.

But how is the migration done? Recursively on each element instances that are below a certain element instance?

ChrisKujawa: This is then the reason why we react on the call activity element instance, to update all childs? As on the childs we don't know whether we are part of a call activity that have been migrated?

ChrisKujawa: Potentially we could do it always when parentPIkey is set, but I guess this is less efficient than your approach? 🤔 But this now causes additional events? 🤔 So maybe we didn't migrated the childs before?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -466,6 +470,63 @@ private void updateElementInstanceRecord(
...
+    final var calledInstance = elementInstanceState.getInstance(calledChildInstanceKey);
+    final var elementInstances = new ArrayDeque<>(List.of(calledInstance));
```
ChrisKujawa, line 483: 👍🏼 Good no recursion - stack trace exception 😆

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceElementMigratedApplier.java
```diff
@@ -55,7 +55,8 @@ public void applyState(final long elementInstanceKey, final ProcessInstanceRecor
               .setCallingElementPath(value.getCallingElementPath());
         });
 
-    if (value.getBpmnElementType() == BpmnElementType.PROCESS) {
+    if (value.getBpmnElementType() == BpmnElementType.PROCESS
+        && previousProcessDefinitionKey.get() != value.getProcessDefinitionKey()) {
```
ChrisKujawa, line 59: ❓ Why we do this?

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCallActivityTest.java
```diff
@@ -449,4 +471,101 @@ public void shouldAllowToTerminateFlowAfterMigratingChildProcessInstance() {
...
+            .getFirst();
+    assertThat(endEventInstance.getValue())
+        .describedAs(
+            "Expect that the end event instance is activated with the correct paths after migration")
+        .hasElementInstancePath(
+            List.of(
+                List.of(processInstanceKey, callActivityElementInstanceKey),
+                List.of(childProcessInstanceKey, endEventInstance.getKey())))
+        .hasProcessDefinitionPath(List.of(targetProcessDefinitionKey))
+        .hasCallingElementPath(List.of(1));
```
ChrisKujawa, line 569: ❓ 🔧 Does it maybe make sense to have a test with deeper nesting?

zeebe/test-util/src/main/java/io/camunda/zeebe/test/util/record/CompactRecordLogger.java
```diff
@@ -569,9 +569,35 @@ private String summarizeProcessInstance(final Record<?> record) {
         .append(formatId(value.getElementId()))
         .append(
             summarizeProcessInformation(value.getBpmnProcessId(), value.getProcessInstanceKey()))
+        .append(summarizeTreePath(value))
```
ChrisKujawa, line 572: 👍🏼

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandler.java
```diff
@@ -177,6 +154,37 @@ public String getIndexName() {
     return indexName;
   }
 
+  /**
+   * Sets the tree path and level of the flow node instance.
```
ChrisKujawa, line 158: 🔧 I would like to keep the terminology `inter tree path`

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandlerTest.java
```diff
@@ -409,5 +409,6 @@ public void shouldUpdateMigratedEntityFromRecord() {
     assertThat(flowNodeInstanceEntity.getStartDate()).isNull();
     assertThat(flowNodeInstanceEntity.getEndDate()).isNull();
     assertThat(flowNodeInstanceEntity.getPosition()).isNull();
+    assertThat(flowNodeInstanceEntity.getTreePath()).isNotNull();
```
ChrisKujawa, line 412: 🔧 ❌  I think we should validate the exact value

ChrisKujawa (approved):
Thanks @mustafadagher in general, it looks good and straightforward to me.

Unfortunately, my knowledge is quite limited regarding the migration. So I hope you get some input from other ZPA members. 

It would be great if you could re-iterate the tests, to enhance them a bit. But I will pre-approve already.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateProcessInstanceTest.java
mustafadagher: we do in `MigrateCallActivityTest`

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
mustafadagher: Not recursively but we go breadth first using a queue. and we do the migration top down in general.

Previously when we migrated a process we migrated all its children, but when one of the children was a call activity that called another process, we didn't migrate the sub-process

I've now introduced migration to the subprocesses too to reflect the possible changes in the tree-paths.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceElementMigratedApplier.java
mustafadagher: now that we are also migrating subprocesses that were called from a migrated parent process, the sub-process model didn't really change, so no need for the deletion and insertion below in that case. the extra check was to save on performance reducing unnecessary operations.

zeebe/engine/src/test/java/io/camunda/zeebe/engine/processing/processinstance/migration/MigrateCallActivityTest.java
mustafadagher: we can, but will it add any extra value to the tests?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandler.java
mustafadagher: it's the `intra` not `inter`, right? it's already part of the block comment (line 165)
>This intra tree path shows the position of a flow node instance within a single process instance.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
```diff
@@ -466,6 +470,63 @@ private void updateElementInstanceRecord(
...
+            .withRecordValue(elementInstanceRecord)
+            .build();
+
+    elementInstanceRecord
+        .setElementInstancePath(elementTreePath.elementInstancePath())
+        .setProcessDefinitionPath(elementTreePath.processDefinitionPath())
+        .setCallingElementPath(elementTreePath.callingElementPath());
+
+    stateWriter.appendFollowUpEvent(
+        instance.getKey(), ProcessInstanceIntent.ELEMENT_MIGRATED, elementInstanceRecord);
```
berkaycanbc, line 519: ❌ Writing `ELEMENT_MIGRATED` event for each child element is semantically incorrect as we do not migrate child instances themselves. Please consider introducing a new intent `ProcessInstanceIntent.PARENT_MIGRATED` for the child elements and apply changes to the state through the applier of that event.

`PARENT_MIGRATED` might not be the best name since we go deeper if the called activity is also a call activity and so on. We might name it as `TREE_PATH_UPDATED` but then it is specific to the tree path concept but not the migration. Any ideas @korthout?

berkaycanbc (changes requested):
🙇‍♂️ Sorry, I finally found time to review.

⭐ Good job @mustafadagher with changes plus clear commit separation and test cases. 

❌ I've requested a change to semantically fix the written element migrated event for child processes. Please take a look. 👀

zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/processinstance/ProcessInstanceMigrationMigrateProcessor.java
korthout: That's an interesting point. However, please note that this means that Camunda Exporter needs to ingest a new event as well. That's fine with me of course. The alternative of `ELEMENT_MIGRATED` is not semantically correct, but seems a pragmatic solution to me.


Alternatively, `ANCESTOR_MIGRATED` would be the generic term you're looking for I think

mustafadagher: @berkaycanbc and @korthout good point, and thanks for the hint. I just updated the PR accordingly, please check again! 🙏

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceElementMigratedApplier.java
```diff
@@ -49,10 +49,14 @@ public void applyState(final long elementInstanceKey, final ProcessInstanceRecor
...
-              .setFlowScopeKey(value.getFlowScopeKey());
+              .setFlowScopeKey(value.getFlowScopeKey())
+              .setElementInstancePath(value.getElementInstancePath())
+              .setProcessDefinitionPath(value.getProcessDefinitionPath())
+              .setCallingElementPath(value.getCallingElementPath());
         });
 
-    if (value.getBpmnElementType() == BpmnElementType.PROCESS) {
+    if (value.getBpmnElementType() == BpmnElementType.PROCESS
+        && previousProcessDefinitionKey.get() != value.getProcessDefinitionKey()) {
```
berkaycanbc, line 59: 🔧 We do not need this check as new introduced a new intent for subprocesses.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/ProcessInstanceAncestorMigratedApplier.java
```diff
@@ -0,0 +1,42 @@
...
+  private final MutableElementInstanceState elementInstanceState;
+
+  public ProcessInstanceAncestorMigratedApplier(
+      final MutableElementInstanceState elementInstanceState) {
+    this.elementInstanceState = elementInstanceState;
+  }
+
+  @Override
+  public void applyState(final long elementInstanceKey, final ProcessInstanceRecord value) {
+    if (value.getBpmnElementType() == BpmnElementType.SEQUENCE_FLOW) {
```
berkaycanbc, line 28: ❌ This check is not needed since we will not write `ANCESTOR_MIGRATED` event for sequence flows.

berkaycanbc (approved):
🚀 Thanks for updates @mustafadagher!

🔧 I've requested two more changes but they do not require an additional review.

✅ Pre-approving since changes I requested are straightforward.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/FlowNodeInstanceFromProcessInstanceHandlerTest.java
```diff
@@ -52,24 +52,23 @@ public void testGetEntityType() {
...
+  @EnumSource(
+      value = ProcessInstanceIntent.class,
+      names = {
+        "ELEMENT_ACTIVATING",
+        "ELEMENT_COMPLETED",
+        "ELEMENT_TERMINATED",
+        "ELEMENT_MIGRATED",
+        "ANCESTOR_MIGRATED"
+      },
+      mode = Mode.INCLUDE)
```
ChrisKujawa, line 65: 🚀

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandlerTest.java
```diff
@@ -455,13 +438,33 @@ void shouldUpdateStateForMigrateRecord() {
...
     final ProcessInstanceForListViewEntity processInstanceForListViewEntity =
         new ProcessInstanceForListViewEntity();
     underTest.updateEntity(processInstanceRecord, processInstanceForListViewEntity);
 
+    // then
     assertThat(processInstanceForListViewEntity.getEndDate()).isNull();
     assertThat(processInstanceForListViewEntity.getStartDate()).isNull();
     assertThat(processInstanceForListViewEntity.getState()).isEqualTo(ProcessInstanceState.ACTIVE);
+    assertThat(processInstanceForListViewEntity.getTreePath()).isEqualTo(treePath);
+  }
```
ChrisKujawa, line 456: ❓ Wondering whether we should further extend the test cases to also cover the ancestor intent now to make sure we consume the tree path

ChrisKujawa (approved):

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/handlers/ListViewProcessInstanceFromProcessInstanceHandlerTest.java
mustafadagher: I missed this comment, was added after I added the request to the merge queue already.
Yes, the extra test would have been nice indeed.

PR #28057, opened 2025-02-12
fix: RaftRequestMetrics is threadsafe

entangled90:
## Description
ConcurrentHashMap is used instead of HashMap as these metrics are called from Netty's threads and not from RaftContext.

### DIFF
--- a/zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
+++ b/zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
@@ -50,9 +50,7 @@ public class RaftRequestMetricsTest {
           });
     }
     // when
-    for (final var task : tasks) {
-      executorService.submit(task);
-    }
+    tasks.forEach(executorService::submit);
     // shutdown the executor and wait for all tasks
     executorService.shutdown();
     // then
--- a/zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
+++ b/zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.atomix.raft.metrics;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import java.util.ArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.junit.jupiter.api.AutoClose;
+import org.junit.jupiter.api.Test;
+
+public class RaftRequestMetricsTest {
+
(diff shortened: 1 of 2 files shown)


zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
```diff
@@ -0,0 +1,61 @@
...
+  @AutoClose private MeterRegistry meterRegistry = new SimpleMeterRegistry();
+  @AutoClose private ExecutorService executorService = Executors.newWorkStealingPool();
+  private final RaftRequestMetrics raftMetrics = new RaftRequestMetrics("1", meterRegistry);
+
+  private final AtomicBoolean failed = new AtomicBoolean(false);
+
+  @Test
+  public void shouldBeThreadSafe() {
+    // given
+    final var tasks = new ArrayList<Runnable>(20_000);
```
npepinpe, line 32: :smile: Not my preferred way of testing thread safety, but we don't have support yet for jcstress or things like this.

zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
```diff
@@ -0,0 +1,61 @@
...
+              raftMetrics.sendMessage(String.valueOf(iFinal % 5), String.valueOf(iFinal % 100));
+            } catch (final Exception e) {
+              failed.set(true);
+            }
+          });
+    }
+    // when
+    for (final var task : tasks) {
+      executorService.submit(task);
+    }
```
npepinpe, line 55: :upside_down_face: 
```suggestion
    tasks.forEach(executorService::submit);
```

npepinpe (approved):
No need for a second review :)

zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
entangled90: yeah I know, but at least it works :shrug:

zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.junit.jupiter.api.AutoClose;
+import org.junit.jupiter.api.Test;
+
+public class RaftRequestMetricsTest {
+
+  @AutoClose private MeterRegistry meterRegistry = new SimpleMeterRegistry();
+  @AutoClose private ExecutorService executorService = Executors.newWorkStealingPool();
```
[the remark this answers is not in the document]
entangled90: The default constructors uses the number of available CPUs for parallelism. I don't think using more threads will lead to higher contention, as in the end, only platform threads actually execute in parallel.

zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
npepinpe: Yeah sure, I don't have a better suggestion. jcstress might be something we want to invest in in the future though :shrug:

zeebe/atomix/cluster/src/test/java/io/atomix/raft/metrics/RaftRequestMetricsTest.java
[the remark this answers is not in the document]
npepinpe: Yeah, my main concern is that (very unlikely) we run this on a single node machine or something, and then it's pointless :smile:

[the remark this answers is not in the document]
npepinpe: I guess it's more of a nit :upside_down_face:

PR #28058, opened 2025-02-12
feat: add Rdbms history cleanup #28101

### LINKED ISSUE #28101: RDBMS - History Cleanup MVP
History cleanup is a feature that removes data - produced by the process-engine and exported to the secondary storage - based on a configurable time-to-live setting. 

It deletes:

- Historic Process Instances plus all related historic data
- Historic Decision Instances plus all related historic data
- Historic Batch Operation plus all related historic data (currently not present in rdbms)

With historic, it refers to completed (or terminated) process instances, decision instances, and batch operations.

For the first iteration, we defined the following scope:

- (Default) History TTL configuration in application.yml that applies to all historical data
- History TTL applies to individual process instances and decision instances.

What is NOT in scope:

- History TTL per definition
- Root Process Instance handling
- Advanced configuration
- Historic Batch Operations

We defined in a Kickoff meeting, after learnings that were made in C7, that we want to use the "[Removal-time-based](https://docs.camunda.org/manual/latest/user-guide/process-engine/history/history-cleanup/#removal-time-based-strategy)" from C7.

stoerti:
## Description

Added a history cleanup for RDBMS

Functionality:
1. When processInstance is finished, a historyCleanupDate is calculated and added to all entities related to a processInstance (variables, userTasks...)
2. A scheduled task cleans all entities with `historyCleanupDate < now()`

Implementation features:
- added history_cleanup_date column to processInstance, variable, userTask, decisionInstance, incident, variable
  - fixed foreign keys for sub-tables for userTask and decisionInstance
  - fixed rdbmsPurger and FK handling: Deactivated FKs while purging
- added logic to entities and mybatis mapper files to store a calculated historyCleanupDate
  - new ordering logic to assure that the historyCleanupDate is updated for all entities (needed for short-running PIs)
- added scheduler to RdbmsExporter to actually cleanup the history
  - calculation algorithm for dynamic interval

Possible configurations:
- defaultHistoryTTL: default time-to-live for processInstances after they finish
- minHistoryCleanupInterval: shortest possible interval between cleanup runs
- maxHistoryCleanupInterval: longest possible interval between cleanup runs
- historyCleanupBatchSize: max number of rows to delete in a cleanup run (per entity)

## Related issues

closes #28101


dist/src/main/resources/application-rdbmsPostgres.yaml
```diff
@@ -86,8 +86,13 @@ zeebe:
         # Classname doesn't matter, since this exporter is present via spring bean
         className: RdbmsExporter
         args:
-          flushInterval: 500
+          flushInterval: PT0.5S
           maxQueueSize: 1000
+          cleanupBatchSize: 1000
+          defaultHistoryTTL: PT1M
```
p-wunderlich, line 92: We should clarify these values maybe with product team?

p-wunderlich (approved):

PR #28059, opened 2025-02-12
fix: clean up index version patterns

### LINKED ISSUE #28066: Index version validation can work with different archiver rollover patterns

We faced the problem in Zeebe benchmark setup, that when archiver is configured with the following rollover pattern:
```
zeebe.broker.exporters.CamundaExporter.args.retention.archiver.elsRolloverDateFormat: yyyy-MM-dd_HH
```
, index schema validation is failing on restart with the following errors:

```
Caused by: io.camunda.operate.exceptions.OperateRuntimeException: Error(s) in index schema: Newer version(s) for variable (8.3.0) already exists: [8.3.0_2025-02-11];Newer version(s) for flownode-instance (8.3.1) already exists: [8.3.1_2025-02-11];Newer version(s) for list-view (8.3.0) already exists: [8.3.0_2025-02-11];Newer version(s) for event (8.3.0) already exists: [8.3.0_2025-02-11];Newer version(s) for sequence-flow (8.3.0) already exists: [8.3.0_2025-02-11];Newer version(s) for job (8.6.0) already exists: [8.6.0_2025-02-11]
```

We need to make sure that even when rollover pattern it configured including underscore, index validation can still extract proper versions from index names.

### Acceptance Criteria (required on creation)

- Validate during tests  that all of the index templates have normal semantic version strings (X.Y.Z)
- Adjust the [VERSION_PATTERN](https://github.com/camunda/camunda/blob/bc05f48f181e9a5b898f128b34cdc50c339543d9/operate/schema/src/main/java/io/camunda/operate/schema/IndexSchemaValidator.java#L34) to be accurate, i.e. it has to extract the version regardless of the rollover pattern
- Add test cases to verify that schema validation passes, even when weird rollover patterns are used

### Links

```[tasklist]
### Pull Requests
- [ ] https://github.com/camunda/camunda/pull/28059
```

sdorokhova:
## Description

We faced the problem in Zeebe benchmark setup, that when archiver is configured with the following rollover pattern:
```
zeebe.broker.exporters.CamundaExporter.args.retention.archiver.elsRolloverDateFormat: yyyy-MM-dd_HH
```
, index schema validation is failing on restart.

Within this PR:
* index and template versions can be of only "X.Y.Z" pattern. Added test to validate that.
* extend `IndexOldSchemaValidatorIT` test to also test archived indices that contain underscore in rollover pattern


webapps-schema/src/test/java/io/camunda/webapps/schema/descriptors/IndexDescriptorTest.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.junit.Assert.assertNotNull;
+
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.platform.commons.support.ReflectionSupport;
+
+public class IndexDescriptorTest {
```
lenaschoenburg, line 20: :+1: Nice test case

operate/schema/src/main/java/io/camunda/operate/schema/IndexSchemaValidator.java
```diff
@@ -31,7 +31,7 @@ public class IndexSchemaValidator {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(IndexSchemaValidator.class);
 
-  private static final Pattern VERSION_PATTERN = Pattern.compile(".*-(\\d+\\.\\d+\\.\\d+.*)_.*");
```
lenaschoenburg, line 34: :x: This may still falsely match some (admittedly weird) rollover patterns such as `yyyy.MM.dd_HH`, right?
What if instead of using one regex on the entire index name, we try to strip the prefix and index name first, knowing that the very next thing has to be the version? That would allow for a much more precise pattern I think.

operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/IndexOldSchemaValidatorIT.java
```diff
@@ -256,7 +256,16 @@ private void whenDatabaseClientReturnsIndexNames(
...
 
   private Set<String> versionsOf(final IndexDescriptor index, final Set<String> versions) {
-    return new HashSet<>(map(versions, version -> getFullQualifiedIndexName(index, version)));
+    return map(
+            versions,
+            version ->
+                Set.of(
+                    getFullQualifiedIndexName(index, version),
+                    // also test archived indices with underscore in suffix
+                    getFullQualifiedIndexName(index, version) + "2025-02-11_20"))
```
lenaschoenburg, line 265: :wrench: Extending the test coverage for all tests in this class is nice but I'd suggest adding one dedicated test method that specifically tests various rollover patterns.

lenaschoenburg (changes requested):
Thanks @sdorokhova, please take a look at my comments to make this even more reliable :safety_vest:

operate/schema/src/main/java/io/camunda/operate/schema/IndexSchemaValidator.java
sdorokhova: I don't think that it will fail, as it takes [1st group found](https://github.com/camunda/camunda/blob/2d9addcc020337a4cf824ce148c98ad922ee790d/operate/schema/src/main/java/io/camunda/operate/schema/IndexSchemaValidator.java#L82). But let me try to improve it the way you propose.

operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/IndexOldSchemaValidatorIT.java
```diff
@@ -167,6 +168,51 @@ public void testOlderVersionsForIndex() {
...
+    whenDatabaseClientReturnsIndexNames(givenIndexNames);
+    assertThat(indexSchemaValidator.newerVersionsForIndex(processIndex)).isEmpty();
+
+    // Only current version
+    whenDatabaseClientReturnsIndexNames(
+        versionsOf(
+            processIndex,
+            Set.of(processIndex.getVersion()),
+            "2022.01.01_10",
+            "2022_01_01_10",
```
lenaschoenburg, line 185:
```suggestion
            "2022_01_01-10",
```

Looks like a copy-paste mistake?

operate/qa/integration-tests/src/test/java/io/camunda/operate/schema/IndexOldSchemaValidatorIT.java
```diff
@@ -167,6 +168,51 @@ public void testOlderVersionsForIndex() {
...
+            "2022-01-01"));
+    assertThat(indexSchemaValidator.olderVersionsForIndex(processIndex)).isEmpty();
+    // Only older versions
+    whenDatabaseClientReturnsIndexNames(
+        versionsOf(processIndex, olderVersions, "2022.01.01_10", "2022_01_01_10", "2022-01-01"));
```
lenaschoenburg, line 212:
```suggestion
        versionsOf(processIndex, olderVersions, "2022.01.01_10", "2022_01_01-10", "2022-01-01"));
```

lenaschoenburg (approved):
Thanks so much! :rocket:

PR #28071, opened 2025-02-12
[Operate CI] Make use of current-test docker image in IT

### LINKED ISSUE #28070: Use current-test docker image for Operate IT in CI
## Description

Make sure that we use the current-test docker image, build for specific branch, when running IT in the CI

ChrisKujawa:
## Description

For Operate CI:

- Build a branch corresponding docker image for Zeebe (introduce separate step)
- Make use of current-test docker image for Zeebe test container
   * Introduce separate property to override the docker repo/registry
   * Make use of property, this allows to use local registry
   * Make sure that current-test tag is accepted by IT infrastructure
   * Read system properties - and default to previous logic if necessary
     Otherwise properties were not accepted somehow
- Current solution also allows to run test locally still with SNAPSHOT

## Related issues

closes https://github.com/camunda/camunda/issues/28070

See for context: https://camunda.slack.com/archives/C06F0GLJNFM/p1739174602278719?thread_ts=1739172753.912819&cid=C06F0GLJNFM

This is to at least make sure that we better test changes, but limit our efforts/work in refactoring the Tasklist CI/IT, as this covered in a follow up PDP

### DIFF
--- a/.github/workflows/operate-ci-test-reusable.yml
+++ b/.github/workflows/operate-ci-test-reusable.yml
@@ -56,7 +56,15 @@ jobs:
     timeout-minutes: 40
     runs-on: ${{ inputs.runner-type }}
     if: ${{ !startsWith(inputs.branch, 'fe-') && !(startsWith(inputs.branch, 'renovate/') && contains(inputs.branch, '-fe-')) }}
-
+    env:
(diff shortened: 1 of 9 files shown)


sdorokhova (approved):
LGTM!

PR #28082, opened 2025-02-12
Migrate workflow engine metrics to Micrometer

### LINKED ISSUE #27144: Migrate all Prometheus metrics in zeebe-workflow-engine
**Description**

Migrate all Prometheus metrics in the zeebe-workflow-engine module.

- [x] BannedInstanceMetrics
- [x] BufferedMessagesMetrics
- [x] IncidentMetrics
- [x] JobMetrics
- [ ] JobMetricsTest
- [x] ProcessEngineMetrics
- [ ] ProcessEngineMetricsTest
- [ ] MetricsTestHelper

npepinpe:
## Description

This PR migrates the workflow engine metrics to Micrometer. No timers were present, so no metric or label names changed.

## Related issues

closes #27144


zeebe/engine/pom.xml
```diff
@@ -125,6 +120,12 @@
       <artifactId>micrometer-commons</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>net.jcip</groupId>
+      <artifactId>jcip-annotations</artifactId>
+      <scope>provided</scope>
```
entangled90, line 126: :question: why the scope is provided?

zeebe/engine/src/main/java/io/camunda/zeebe/engine/metrics/JobProcessingMetrics.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public void countJobEvent(final JobAction action, final String type, final int amount) {
+    jobEvents.computeIfAbsent(action, type, this::registerJobEventCounter).increment(amount);
+  }
+
+  private Counter registerJobEventCounter(final JobAction jobAction, final String jobType) {
+    return Counter.builder(EngineMetricsDoc.JOB_EVENTS.getName())
+        .description(EngineMetricsDoc.JOB_EVENTS.getDescription())
+        .tag(EngineKeyNames.JOB_ACTION.asString(), jobAction.getLabel())
+        .tags(EngineKeyNames.JOB_TYPE.asString(), jobType)
```
entangled90, line 45:
```suggestion
        .tag(EngineKeyNames.JOB_TYPE.asString(), jobType)
```

zeebe/engine/src/main/java/io/camunda/zeebe/engine/metrics/JobProcessingMetrics.java
```diff
@@ -0,0 +1,48 @@
...
+import net.jcip.annotations.NotThreadSafe;
+
+/**
+ * This class is not thread safe, and is meant to be used only in the context of the engine actor.
+ *
+ * <p>If you need to use it from the scheduling actor, or from somewhere else, then you need to
+ * first make this thread safe.
+ */
+@NotThreadSafe
+public final class JobProcessingMetrics {
```
entangled90, line 24: :thought_balloon: :crossed_fingers: renaming this class will not create more conflicts across all backports

entangled90 (approved):

zeebe/engine/pom.xml
npepinpe: The annotations are not retained at runtime, so there's no need to have the JAR packaged with the project.

zeebe/engine/src/main/java/io/camunda/zeebe/engine/metrics/JobProcessingMetrics.java
npepinpe: :see_no_evil:

zeebe/engine/src/main/java/io/camunda/zeebe/engine/metrics/JobProcessingMetrics.java
npepinpe: I renamed it to differentiate it from the others, since its accessed during processing and not from the state. We've discussed with Chris in the past to extract the state classes into their own modules, and the fact they're opened outside of the engine module is a strong hint to that. Otherwise you couple things to the engine (and its processing classes) only for the state, leading to a "wide" module.

Issue #26065, opened 2024-12-13
Include processInstanceKey in 8.7 in order to archive in 8.8 be able to process the data


marcosgvieira:
### Description

The problem arises primarily due to the switch from importers to exporters, combined with the inclusion of the processInstanceKey field in the task-variables template. Since archiving in exporters only considers data with the processInstanceKey, any documents without the processInstanceKey would not be archived.

With the return of importers, we can add that the processInstanceKey field to the task-variables template, and update the importers. This would mitigate the problem in the exporter, as the current Tasklist Archiving process would already be responsible for archiving the data and the new data indexed via the importer would already contain the processInstanceKey, meaning we wouldn’t need to worry about this during migration or develop additional jobs for the exporter (coming on version 8.8).

Currently, in version 8.7, the processInstanceKey is not included in the task-variables template (checked on stable/8.7). Based on this solution, the task breakdown would be as follows:
• Add the processInstanceKey to the task-variables template.
• Add this field to the importer for both Elasticsearch and Opensearch.

Please let me know if this solution does not make sense or if I am missing something. Once validated, I will update the description of this issue, with the acceptance criteria.

**Acceptance Criteria**

- [x] Have the processInstanceKey in stable/8.7 in the task-variables template
- [ ] The importer should be able to handle the new field for Elasticsearch and Opensearch

marcosgvieira:
I made progress on this one Today - decided to use Archiver structure as all the logic to archive/reindex it's easier to be used there, I'm creating an extra feature on the archiver tasks that will allow a task to stop being rescheduled based in a condition.

I also thought on using Migration structure to achieve the same, but it would be more complex to replicate/reuse the archiving part inside migration module without code duplication.

@romansmirnov fyi and let me know if the above doesn't make sense.

nathansandi:
Hello @romansmirnov (cc: @marcosgvieira),

I would like to share a quick update based on my analysis and ask for feedback.

With the return of importers, we can ensure that the `processInstanceKey` field is added to the `task-variables` template. This would mitigate the problem in the exporter, as the current Tasklist Archiving process would already be responsible for archiving the data. Additionally, the new data indexed via the importer would already contain the `processInstanceKey`, meaning we wouldn’t need to worry about this during migration or develop additional jobs for the exporter.

Currently, in version 8.7, the `processInstanceKey` is not included in the `task-variables` template (checked on stable/8.7). Based on this solution, the task breakdown would be as follows:
[...]

PR #28083, opened 2025-02-12
feat: include processInstanceKey to the tasklist-task-variables

nathansandi:
## Description

With the return of importers, we can add that the processInstanceKey field to the task-variables template, and update the importers. This would mitigate the problem in the exporter, as the current Tasklist Archiving process would already be responsible for archiving the data and the new data indexed via the importer would already contain the processInstanceKey, meaning we wouldn’t need to worry about this during migration or develop additional jobs for the exporter (coming on version 8.8).

On this PR we are covering:
- Add the processInstanceKey to the task-variables template.
- Add this field to the importer for both Elasticsearch and Opensearch.
- Adjust variables to include the processInstanceKey once task is concluded
- Integration tests

## Related issues

closes #https://github.com/camunda/camunda/issues/26065

### DIFF
--- a/tasklist/importer-870/src/main/java/io/camunda/tasklist/zeebeimport/v870/processors/es/UserTaskZeebeRecordProcessorElasticSearch.java
+++ b/tasklist/importer-870/src/main/java/io/camunda/tasklist/zeebeimport/v870/processors/es/UserTaskZeebeRecordProcessorElasticSearch.java
@@ -66,7 +66,7 @@ public class UserTaskZeebeRecordProcessorElasticSearch {
         final List<TaskVariableEntity> variables =
             userTaskRecordToVariableEntityMapper.mapVariables(record);
         for (final TaskVariableEntity variable : variables) {
-          bulkRequest.add(getVariableQuery(variable));
+          bulkRequest.add(getVariableQuery(variable, record.getValue().getProcessInstanceKey()));
         }
       }
     }
@@ -97,7 +97,8 @@ public class UserTaskZeebeRecordProcessorElasticSearch {
     }
   }
 
-  private UpdateRequest getVariableQuery(final TaskVariableEntity variable)
+  private UpdateRequest getVariableQuery(
+      final TaskVariableEntity variable, final Long processInstanceKey)
       throws PersistenceException {
     try {
       LOGGER.debug("Variable instance for list view: id {}", variable.getId());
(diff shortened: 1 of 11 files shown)


tasklist/importer-870/src/main/java/io/camunda/tasklist/zeebeimport/v870/processors/es/UserTaskZeebeRecordProcessorElasticSearch.java
```diff
@@ -97,14 +97,16 @@ private UpdateRequest getTaskQuery(final TaskEntity entity, final Record record)
...
+  private UpdateRequest getVariableQuery(
+      final TaskVariableEntity variable, final Long processInstanceKey)
       throws PersistenceException {
     try {
       LOGGER.debug("Variable instance for list view: id {}", variable.getId());
       final Map<String, Object> updateFields = new HashMap<>();
       updateFields.put(TaskVariableTemplate.VALUE, variable.getValue());
       updateFields.put(TaskVariableTemplate.FULL_VALUE, variable.getFullValue());
       updateFields.put(TaskVariableTemplate.IS_PREVIEW, variable.getIsPreview());
+      updateFields.put(TaskVariableTemplate.PROCESS_INSTANCE_KEY, processInstanceKey);
```
panagiotisgts, line 109: :wrench:  Should we also add routing to this via the Process Instance key for ES and OS aswell?

tasklist/qa/integration-tests/src/test/java/io/camunda/tasklist/webapp/api/rest/v1/controllers/TaskControllerIT.java
```diff
@@ -1381,6 +1406,24 @@ public void completeZeebeUserTaskWithVariables() throws Exception {
                 assertThat(task.getImplementation()).isEqualTo(TaskImplementation.ZEEBE_USER_TASK);
               });
 
+      // Validate variables has processInstanceKey in ES/OS
```
panagiotisgts, line 1409: Same here, these comments can be removed.

panagiotisgts (changes requested):
Some minor cleanups, and a suggestion regarding routing of the ES/OS request.

tasklist/importer-870/src/main/java/io/camunda/tasklist/zeebeimport/v870/processors/es/UserTaskZeebeRecordProcessorElasticSearch.java
nathansandi: We validated together on Slack, and base on https://github.com/camunda/camunda/blob/5e768f0143a09ccc42a721884fc1400fd1982a5b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/handlers/UserTaskCompletionVariableHandler.java 
It is not necessary to add routing for this case.

panagiotisgts (approved):

Issue #26046, opened 2024-12-13
Importer(s) are not communicating import done correctly


ChrisKujawa:
**Describe the bug**

When running a benchmark, we observed that the Camunda Exporter failed after restart, [because the Importer doesn't seem to be completed (where it should be).](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22zeebe-io%22%0Aresource.labels.location%3D%22europe-west1-b%22%0Aresource.labels.cluster_name%3D%22zeebe-cluster%22%0Aresource.labels.namespace_name%3D%22ck-main-exporter%22%0Alabels.k8s-pod%2Fapp%3D%22camunda-platform%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fcomponent%3D%22zeebe-broker%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Finstance%3D%22zeebe-benchmark-test%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fmanaged-by%3D%22Helm%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fname%3D%22camunda-platform%22%0Alabels.k8s-pod%2Fapp_kubernetes_io%2Fpart-of%3D%22camunda-platform%22;pinnedLogId=2024-12-12T21:20:50.478899335Z%2F58efh5mbgslw9b7p;cursorTimestamp=2024-12-12T21:20:50.489646788Z;duration=PT1H?inv=1&invt=Abj9IQ&project=zeebe-io)

```
INFO 2024-12-12T21:20:50.478899335Z Waiting for importers to finish, cached record with key 2251799814997670 but did not flush
```

The importer seems to not complete all documents.

![import-completed](https://github.com/user-attachments/assets/244bd959-82a4-4557-a6a1-f8398e560a6b)

ChrisKujawa:
https://zeebe-io.github.io/zeebe-chaos/2024/12/12/News-from-Camunda-Exporter-project

michaldytko:
@ChrisKujawa I can still reproduce it with 8.7.0-alpha3. I'm using docker-compose with CamundaExporter for Zeebe.

EuroLew:
> [@ChrisKujawa](https://github.com/ChrisKujawa) I can still reproduce it with 8.7.0-alpha3. I'm using docker-compose with CamundaExporter for Zeebe.

[...]

PR #28105, opened 2025-02-13
Camunda exporter startup bug

EuroLew:
## Description

There is a bug where the camunda exporter can enter an error loop in a new cluster where it is waiting for importers to complete which never will (as importers will not do any work in 8.8). This fix will as per the slack thread referred to here https://github.com/camunda/camunda/issues/26046#issuecomment-2647813197, assume that a lack of existence of zeebe indices (which the importer reads from), means that it is a fresh cluster, thus the exporter does not need to wait and we avoid the error loop.

Questions
- There is a requirement that the new `zeebeIndexPrefix` match the zeebe index prefix defined for the ES exporter, I have documented this in the yaml but I feel like this should be explicitly stated elsewhere otherwise the 8.8 camunda exporter can enter boot loops. (Although it will work work with default configuration).
- Would this need to get backported for people running benchmarks? (to which version?)

## Related issues

relates to #26046

### DIFF
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -987,8 +987,8 @@
         #
         #   index:
         #     prefix:
-        #     zeebeIndexPrefix: zeebe-record
-        #     The 'zeebeIndexPrefix' must be the same value as the Elasticsearch exporter 'elasticsearch.args.index.prefix'
+        #     zeebeExporterIndexPrefix: zeebe-record
+        #     The 'zeebeIndexPrefix' must be the same value as the Elasticsearch/OpenSearch exporter 'elasticsearch.args.index.prefix'
         #     otherwise the exporter may never start.
         #     numberOfShards: 3
         #     numberOfReplicas: 0
--- a/dist/src/main/config/broker.standalone.yaml.template
+++ b/dist/src/main/config/broker.standalone.yaml.template
@@ -987,6 +987,9 @@
         #
         #   index:
         #     prefix:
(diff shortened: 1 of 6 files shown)


dist/src/main/config/broker.standalone.yaml.template
```diff
@@ -987,6 +987,9 @@
         #
         #   index:
         #     prefix:
+        #     zeebeIndexPrefix: zeebe-record
```
ChrisKujawa, line 990:
```suggestion
        #     zeebeExporterIndexPrefix: zeebe-record
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/CamundaExporter.java
```diff
@@ -284,13 +285,20 @@ private void checkImportersCompletedAndReschedule() {
...
-      importersCompleted =
-          searchEngineClient.importersCompleted(partitionId, importPositionIndices);
+      final var zeebeIndicesExist =
+          !searchEngineClient
+              .getMappings(
+                  configuration.getIndex().getZeebeIndexPrefix() + "*", MappingSource.INDEX)
+              .isEmpty();
+
+      exporterCanFlush =
+          !zeebeIndicesExist
```
ChrisKujawa, line 295: 🔧 As this is a repeating check, we could also think about whether we want to safe this result, when we the first time encounter zeebe indices, we don't need to recheck for them.

ChrisKujawa (approved):
Thanks @EuroLew 🚀 Looks good, I had smaller comments, please take a look before merging, but I pre-approve already.

PR #28115, opened 2025-02-13
migrate zeebeDb to micrometer

### LINKED ISSUE #27153: Migrate all Prometheus metrics in zeebe-db
**Description**

Migrate all the Prometheus metrics in the zeebe-db module.

entangled90:
## Description
The metrics in ZeebeDb are a bit different compared to the rest as well directly rocksDB for them.

Now that the meterRegistry is created inside ZeebeDb is better to have the ZeebeDbMetricsExporter couple to the lifecycle of ZeebeDb directly, so it has been moved inside it and a new default method defined in the interface, so that the timer can be scheduled at startup as it was done before.

The MeterDocumentation for the rocksDB metrics are available using the static method:
```
ZeebeRocksDBMetricExporter.allMeterDocumentations()
```
## Related issues

closes #27153

### DIFF
--- a/zeebe/zb-db/src/main/java/io/camunda/zeebe/db/ZeebeDb.java
+++ b/zeebe/zb-db/src/main/java/io/camunda/zeebe/db/ZeebeDb.java
@@ -65,7 +65,14 @@ public interface ZeebeDb<ColumnFamilyType extends Enum<? extends EnumValue> & En
    */
   boolean isEmpty(ColumnFamilyType column, TransactionContext context);
 
+  /**
+   * @return the StatefulMeterRegister instance that is tied to this ZeebeDb instance. When ZeebeDb
+   *     is closed, all metrics registered through this instances will be removed from the
+   *     underlying registry.
+   *     <p>Do not register metrics that must outlive the scope of this ZeebeDb instance
+   */
   StatefulMeterRegistry getMeterRegistry();
 
+  /** Export the metrics from RocksDB into micrometer */
   default void exportMetrics() {}
 }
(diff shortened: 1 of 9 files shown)


zeebe/zb-db/src/main/java/io/camunda/zeebe/db/ZeebeDb.java
```diff
@@ -66,4 +66,6 @@ ColumnFamily<KeyType, ValueType> createColumnFamily(
   boolean isEmpty(ColumnFamilyType column, TransactionContext context);
 
   StatefulMeterRegistry getMeterRegistry();
```
npepinpe, line 68: :wrench: if you could please fix my mistake, and add docs to describe when to use this registry, and its lifecycle.

zeebe/zb-db/src/main/java/io/camunda/zeebe/db/ZeebeDb.java
```diff
@@ -66,4 +66,6 @@ ColumnFamily<KeyType, ValueType> createColumnFamily(
...
+
+  default void exportMetrics() {}
```
npepinpe, line 70: :wrench: Please add docs as to the expected usage

npepinpe (approved):
:rocket: Pre-approving based on conversation

npepinpe:
@entangled90 - as a follow up, please move the exporting task to the state controller implementation and get rid of the rocksdb metrics exporter startup step.

npepinpe:
Classic atomix flaky test :|

PR #28119, opened 2025-02-13
Migrates job client stream metrics to Micrometer

npepinpe:
## Description

This PR migrates the job client stream metrics to Micrometer. There was a histogram, but it's really a distribution summary, so no metric name changes (or labels, for that matter).

Incidentally, this also removes Prometheus from the dist module :tada: 

## Related issues

related to #26078


zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/stream/JobClientStreamMetricsDoc.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    @Override
+    public Type getType() {
+      return Type.GAUGE;
+    }
+  },
+
+  /** Distribution of client count per aggregated stream */
+  AGGREGATED_CLIENTS {
```
entangled90, line 71: :wrench: I think we need to add the distributionSLOs as the default is empty? Not sure if it's needed though

entangled90 (approved):
Pre-approving, just double check if we need to add the distributionSLOs for compatibility :rocket:

zeebe/gateway/src/main/java/io/camunda/zeebe/gateway/impl/stream/JobClientStreamMetricsDoc.java
npepinpe: Can you take this over? Check what are the default right now with Prometheus, and just use them?

entangled90:
flaky atomix-cluster unit tests

PR #28140, opened 2025-02-13
fix: handle failure on first migration element

panagiotisgts:
## Description

Identified a case where the process-migration would skip records if the first entity (Process Definition key sorted) to migrate had errors when flushing to ES/OS
## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/migration/process-migration/src/main/java/io/camunda/migration/process/adapter/os/OpensearchAdapter.java
+++ b/migration/process-migration/src/main/java/io/camunda/migration/process/adapter/os/OpensearchAdapter.java
@@ -221,9 +221,7 @@ public class OpensearchAdapter implements Adapter {
     final var sorted = items.stream().sorted(Comparator.comparing(BulkResponseItem::id)).toList();
     for (int i = 0; i < sorted.size(); i++) {
       if (sorted.get(i).error() != null) {
-        return i > 0
-            ? Objects.requireNonNull(sorted.get(i - 1).id())
-            : Objects.requireNonNull(sorted.get(i).id());
+        return i == 0 ? null : Objects.requireNonNull(sorted.get(i - 1).id());
       }
     }
 
--- a/migration/process-migration/src/test/java/io/camunda/migration/process/it/MigrationRunnerIT.java
+++ b/migration/process-migration/src/test/java/io/camunda/migration/process/it/MigrationRunnerIT.java
@@ -387,4 +387,33 @@ public class MigrationRunnerIT extends AdapterTest {
     final var ex = assertThrows(MigrationException.class, migrator::call);
     assertThat(ex.getMessage()).isEqualTo("Failed to fetch last migrated process");
   }
+
+  @ParameterizedTest
+  @ValueSource(booleans = {true, false})
+  public void shouldNotFlushStepOnError(final boolean isElasticsearch) throws IOException {
+    this.isElasticsearch = isElasticsearch;
+
+    if (isElasticsearch) {
+      ES_CONFIGURATION.setIndexPrefix(MISCONFIGURED_PREFIX);
+    } else {
+      OS_CONFIGURATION.setIndexPrefix(MISCONFIGURED_PREFIX);
+    }
+
+    final Adapter adapter =
+        isElasticsearch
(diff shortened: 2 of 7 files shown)


ChrisKujawa (commented):
Great that we found this. Can we somehow write a unit test for this?

migration/process-migration/src/main/java/io/camunda/migration/process/adapter/os/OpensearchAdapter.java
```diff
@@ -221,9 +221,7 @@ private String lastUpdatedProcessDefinition(final List<BulkResponseItem> items)
     final var sorted = items.stream().sorted(Comparator.comparing(BulkResponseItem::id)).toList();
     for (int i = 0; i < sorted.size(); i++) {
       if (sorted.get(i).error() != null) {
-        return i > 0
-            ? Objects.requireNonNull(sorted.get(i - 1).id())
-            : Objects.requireNonNull(sorted.get(i).id());
```
ChrisKujawa, line 226: 🤔 TBH I'm not sure whether I understand the problem and fix

panagiotisgts: Basically, if the first item of a batch, sorted by key, on ES/OS response of the bulk update had an error, I was returning that key. However, that record had never been updated thus it was being skipped.

migration/process-migration/src/test/java/io/camunda/migration/process/it/MigrationRunnerIT.java
```diff
@@ -387,4 +387,33 @@ public void shouldThrowException(final boolean isElasticsearch) {
...
+        adapter.migrate(List.of(MigrationUtil.migrate(entityToBeMigrated)));
+
+    assertThat(migratedEntityId).isNull();
+
+    if (isElasticsearch) {
+      ES_CONFIGURATION.setIndexPrefix(null);
+    } else {
+      OS_CONFIGURATION.setIndexPrefix(null);
+    }
+  }
```
ChrisKujawa, line 418: 🔧 i feel like we should refactor the tests, to split up the concerns ES/OS or maybe we can do it somehow on the infrastructure of the test. As these conditions, etc, make them much harder to read/maintain, I think. Nothing you need to do now.

ChrisKujawa, line 418:
```suggestion
  @ParameterizedTest
  @ValueSource(booleans = {true, false})
  public void shouldNotFlushStepOnError(final boolean isElasticsearch) throws IOException {
    // given
    this.isElasticsearch = isElasticsearch;

    if (isElasticsearch) {
      ES_CONFIGURATION.setIndexPrefix(MISCONFIGURED_PREFIX);
    } else {
      OS_CONFIGURATION.setIndexPrefix(MISCONFIGURED_PREFIX);
    }

    final Adapter adapter =
        isElasticsearch
            ? new ElasticsearchAdapter(properties, ES_CONFIGURATION)
            : new OpensearchAdapter(properties, OS_CONFIGURATION);
    final ProcessEntity entityToBeMigrated = TestData.processEntityWithPublicFormId(1L);
    writeToMisconfiguredProcessToIndex(entityToBeMigrated);
    
    // when
    final String migratedEntityId =
        adapter.migrate(List.of(MigrationUtil.migrate(entityToBeMigrated)));

    // then
    assertThat(migratedEntityId).isNull();

    if (isElasticsearch) {
      ES_CONFIGURATION.setIndexPrefix(null);
    } else {
      OS_CONFIGURATION.setIndexPrefix(null);
    }
  }
```

ChrisKujawa (approved):
Thanks for fixing this @panagiotisgts 👍🏼

migration/process-migration/src/test/java/io/camunda/migration/process/it/MigrationRunnerIT.java
panagiotisgts: In general I believe the whole movement for the arch streamlining has led to the creation of multiple test frameworks and setups that should be consolidated since many of them are overlapping.

PR #28150, opened 2025-02-13
feat: rdbms - add partitionId for history cleanup relevant tables #28101

p-wunderlich:
## Description

Add partitionId for history cleanup relevant tables. This is needed to do cleanup per partition, so that multiple exporters (one per partition) can cleanup there "own" data.

### DIFF
--- a/db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
+++ b/db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
@@ -60,6 +60,7 @@
       <column name="NUM_INCIDENTS" type="SMALLINT" />
       <column name="ELEMENT_ID" type="VARCHAR(255)"/>
       <column name="VERSION" type="SMALLINT" />
+      <column name="PARTITION_ID" type="NUMBER"/>
     </createTable>
 
     <modifySql dbms="mariadb">
@@ -84,6 +85,7 @@
       <column name="VAR_FULL_VALUE" type="CLOB"/>
       <column name="TENANT_ID" type="VARCHAR(255)"/>
       <column name="IS_PREVIEW" type="BOOLEAN"/>
+      <column name="PARTITION_ID" type="NUMBER"/>
     </createTable>
 
     <createIndex tableName="${prefix}VARIABLE" indexName="${prefix}IDX_VARIABLE_PROCESS_INSTANCE_KEY">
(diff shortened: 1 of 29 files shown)


db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
```diff
@@ -232,6 +237,7 @@
       <column name="VERSION" type="SMALLINT" />
       <column name="RESOURCE_NAME" type="VARCHAR(4000)" />
       <column name="XML" type="CLOB" />
+      <column name="PARTITION_ID" type="NUMBER"/>
```
stoerti, line 240: ❓ Decision requirements are not deleted by the history cleanup, why is it part of this PR?

db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/DecisionInstanceDbModel.java
```diff
@@ -185,6 +192,7 @@ public DecisionInstanceDbModel build() {
           rootDecisionDefinitionKey,
           decisionType,
           tenantId,
+          Long.valueOf(partitionId),
```
stoerti, line 195: ❓  Is it a long or an int? Why the difference?

zeebe/exporters/rdbms-exporter/src/main/java/io/camunda/exporter/rdbms/handlers/ProcessInstanceExportHandler.java
```diff
@@ -55,18 +56,20 @@ public void export(final Record<ProcessInstanceRecordValue> record) {
...
-        0,
-        null,
-        value.getVersion());
+    return new ProcessInstanceDbModelBuilder()
+        .processInstanceKey(value.getProcessInstanceKey())
+        .processDefinitionId(value.getBpmnProcessId())
+        .processDefinitionKey(value.getProcessDefinitionKey())
+        .state(ProcessInstanceState.ACTIVE)
+        .startDate(DateUtil.toOffsetDateTime(record.getTimestamp()))
+        .endDate(null)
```
stoerti, line 65: 🔧  Maybe move the default `null` values to the builder?

db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml
p-wunderlich: A that's true, reverted it.

db/rdbms/src/main/java/io/camunda/db/rdbms/write/domain/DecisionInstanceDbModel.java
p-wunderlich: Forget this place. Its a int. Was wrong because we have places where we used it as long. Fixed it also at the other places.

stoerti (approved):

PR #28152, opened 2025-02-14
fix: remove unnecessary info from ui config response

abremard:
## Description
There is some webapps data that is currently being returned by the `ui-configuration` endpoint which is not needed in C8. This PR removes that part of the response.

## Related issues

related to https://github.com/camunda/camunda-optimize/issues/10593


optimize/backend/src/main/java/io/camunda/optimize/service/UIConfigurationService.java
```diff
@@ -40,7 +35,6 @@ public class UIConfigurationService {
 
   public static final DateTimeFormatter DATE_TIME_FORMATTER =
       DateTimeFormatter.ofPattern("uuuu-MM-dd'T'HH:mm:ss'Z'").withZone(ZoneOffset.UTC);
-  private static final Logger LOG = org.slf4j.LoggerFactory.getLogger(UIConfigurationService.class);
```
abremard, line 43: Not related to the change but the LOG here is not used

RomanJRW:
@abremard - please don't consider this a complete review, but you might want to consider going the whole way with this. What I mean is also removing this from the API, by changing:
- The UI configuration response DTO
- The FE consumption of the API
- Updating the [internal API docs](https://confluence.camunda.com/display/CO/UI+Configuration)

abremard:
> @abremard - please don't consider this a complete review, but you might want to consider going the whole way with this. What I mean is also removing this from the API, by changing:
> 
> * The UI configuration response DTO
> * The FE consumption of the API
> * Updating the [internal API docs](https://confluence.camunda.com/display/CO/UI+Configuration)

That's a good point, I'll update those

optimize/client/src/modules/components/ReportRenderer/visualizations/Table/processRawData.test.js
```diff
@@ -155,7 +137,7 @@ describe('Process table', () => {
     expect(cells[0]).toBe('1d 10h 10min 23s 423ms');
   });
 
-  it('should not make the processInstanceId a link if no endpoint is specified', () => {
+  it('should display the processInstanceId as text', () => {
```
abremard, line 140: see comment above

grlimacan (approved):
LGTM! On the FE side, just limited review as it's not my area of expertise

PR #28157, opened 2025-02-14
IdentitySetupInitializeProcessor: don't write commands on distributed partitions

### LINKED ISSUE #28278: IdentitySetupInitializeProcessor: append create authorization command only while processing new commands
## Description

The `IdentitySetupInitializeProcessor` creates Authorizations. For this it has to generate an Authorization key. This is fine, were it not that we generate the key on processing a distributed command as well. This will result in a state inconsistency, where each of the partitions will have the authorization stored with a different key.

We need to ensure we don't generate keys on distributed commands. For Authorizations this is failry simple. We need to ensure that:
- On processing a new command we write all the authorizations that need to be created as a follow-up command. These follow-up commands already have distribution covered.
- On processing a distributed command we should not worry about creating any authorizations. This will be handles with different distributions in the `AuthorizationCreateProcessor`.

nicpuppa:
## Description

Since now from the IdentitySetupInitializeProcessor we send a follow up command (not an event anymore), the permissions need to be added only while processing new commands. Then the AuthorizationCreateProcessor will take care of distribute properly the authorization command

## Related issues

closes #28278

### DIFF
--- a/security/security-core/src/main/java/io/camunda/zeebe/protocol/record/value/AuthorizationResourceType.java
+++ b/security/security-core/src/main/java/io/camunda/zeebe/protocol/record/value/AuthorizationResourceType.java
@@ -10,6 +10,7 @@ package io.camunda.zeebe.protocol.record.value;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public enum AuthorizationResourceType {
   AUTHORIZATION(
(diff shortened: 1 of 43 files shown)


nicpuppa:
I'm not sure what is going on with the IT tests

Ben-Sheppard (commented):
On the parent issue it states

> On processing a distributed command we should not worry about creating any authorizations. This will be handles with different distributions in the AuthorizationCreateProcessor.

I can't see any changes here related to this in this PR, is it because this was always the case? Or am I missing something :)

remcowesterhoud:
> I can't see any changes here related to this in this PR, is it because this was always the case? Or am I missing something :)

You are missing something. We have removed the adding of permissions in the `createRole` method which. This method was called during distribution. We now only add the permissions when processing the new command.

https://github.com/camunda/camunda/pull/28157/files#diff-b992ce3dab140671c219345e3b86945f0349b0f567b2ead4c67ba93305999575L207

Ben-Sheppard:
Ah I see, thank you for the confirmation @remcowesterhoud :)

Ben-Sheppard (approved):
Thank you for addressing my comment and also giving context (I initially missed the change because I hadn't unfolded code to see where the `createRole` method was used.

PR #28165, opened 2025-02-14
refactor: remove explicit dynatrace support; we will document it instead

npepinpe:
## Description

This PR removes explicit Dynatrace support. We would recommend usage of the OTLP exporter instead, but if not possible, Dynatrace can still be used by mounting the registry JAR and configuring the application, so the it's also not too terrible.

If there's enough demand, we can always add it back.

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -210,12 +210,6 @@
     </dependency>
 
     <dependency>
-      <groupId>io.micrometer</groupId>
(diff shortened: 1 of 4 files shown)


dist/pom.xml
```diff
@@ -209,12 +209,6 @@
       <scope>runtime</scope>
     </dependency>
 
-    <dependency>
```
entangled90, line 212: :tada:

entangled90 (approved):

PR #28189, opened 2025-02-14
Fix flaky BackupRestoreIT

### LINKED ISSUE #27850: io.camunda.it.backup.BackupRestoreIT is flaky
**Summary**

- How often does the test fail? 15 times in the last 48 hours
- Does it block your work? Sort of
- Do we suspect that it is a real failure? No

**Failures**

See: https://dashboard.int.camunda.com/d/ae2j69npxh3b4f/flaky-tests-camunda-camunda-monorepo?orgId=1&from=now-2d&to=now&timezone=browser&var-branch=main

You can find links to failed runs associated to the tests in the dashboard.

rodrigo-lourenco-lopes:
## Description

```
  <testcase name="shouldBackupAndRestoreToPreviousState(BackupRestoreTestConfig)[2] config=BackupRestoreTestConfig[databaseType=opensearch, bucket=bucket] (Flaky Test)" classname="io.camunda.it.backup.BackupRestoreIT" time="39.106">
    <failure type="feign.FeignException$NotFound (Flaky Test)"><![CDATA[feign.FeignException$NotFound: [404] during [GET] to [http://0.0.0.0:1431/actuator/backupHistory/1] [HistoryBackupClient#getBackup(long)]: [{"message":"No backup with id [1] found."}]
	at feign.FeignException.clientErrorStatus(FeignException.java:249)
	at feign.FeignException.errorStatus(FeignException.java:223)
	at feign.FeignException.errorStatus(FeignException.java:213)
	at feign.codec.ErrorDecoder$Default.decode(ErrorDecoder.java:103)
	at feign.InvocationContext.decodeError(InvocationContext.java:133)
	at feign.InvocationContext.proceed(InvocationContext.java:80)
	at feign.ResponseHandler.handleResponse(ResponseHandler.java:69)
	at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:109)
	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:53)
	at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:104)
	at jdk.proxy2/jdk.proxy2.$Proxy214.getBackup(Unknown Source)
	at io.camunda.it.backup.BackupRestoreIT.lambda$shouldBackupAndRestoreToPreviousState$1(BackupRestoreIT.java:137)
	at org.awaitility.core.AssertionCondition.lambda$new$0(AssertionCondition.java:53)
	at org.awaitility.core.ConditionAwaiter$ConditionPoller.call(ConditionAwaiter.java:248)
	at org.awaitility.core.ConditionAwaiter$ConditionPoller.call(ConditionAwaiter.java:235)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
]]></failure>
```

Error happens when checking for the first time for the backup state. This might be a race condition as this is still not ready to poll even though the backup is already in progress. 

The solution might be to add one poll delay for the first check to avoid this race condition. 

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/27850

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/backup/BackupRestoreIT.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/backup/BackupRestoreIT.java
@@ -47,7 +47,6 @@ public class BackupRestoreIT {
   private String dbUrl;
   private GenericContainer<?> searchContainer;
   private DataGenerator generator;
-  private BackupRestoreTestConfig config;
   private HistoryBackupClient historyBackupClient;
 
   @AfterEach
@@ -93,7 +92,6 @@ public class BackupRestoreIT {
     configurator.getOperateProperties().getBackup().setRepositoryName(REPOSITORY_NAME);
     configurator.getTasklistProperties().getBackup().setRepositoryName(REPOSITORY_NAME);
 
-    this.config = config;
     testStandaloneApplication.start().awaitCompleteTopology();
 
     camundaClient = testStandaloneApplication.newClientBuilder().build();
--- a/qa/integration-tests/src/test/java/io/camunda/it/backup/BackupRestoreIT.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/backup/BackupRestoreIT.java
@@ -9,6 +9,7 @@ package io.camunda.it.backup;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import feign.FeignException.NotFound;
 import io.camunda.client.CamundaClient;
 import io.camunda.it.utils.MultiDbConfigurator;
 import io.camunda.management.backups.TakeBackupHistoryResponse;
@@ -127,13 +128,17 @@ public class BackupRestoreIT {
     final var snapshots = takeResponse.getScheduledSnapshots();
 
     Awaitility.await("Backup completed")
-        .pollDelay(Duration.ofSeconds(1))
         .atMost(Duration.ofSeconds(600))
         .untilAsserted(
             () -> {
-              final var backupResponse = historyBackupClient.getBackup(1L);
-              assertThat(backupResponse.getState()).isEqualTo(BackupStateDto.COMPLETED);
-              assertThat(backupResponse.getDetails()).allMatch(d -> d.getState().equals("SUCCESS"));
+              try {
+                final var backupResponse = historyBackupClient.getBackup(1L);
+                assertThat(backupResponse.getState()).isEqualTo(BackupStateDto.COMPLETED);
+                assertThat(backupResponse.getDetails())
+                    .allMatch(d -> d.getState().equals("SUCCESS"));
+              } catch (final NotFound e) {
+                throw new AssertionError("Backup not found:", e);
+              }
             });
 
     // when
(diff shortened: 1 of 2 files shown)


qa/integration-tests/src/test/java/io/camunda/it/backup/BackupRestoreIT.java
```diff
@@ -131,6 +127,7 @@ public void shouldBackupAndRestoreToPreviousState(final BackupRestoreTestConfig
     final var snapshots = takeResponse.getScheduledSnapshots();
 
     Awaitility.await("Backup completed")
+        .pollDelay(Duration.ofSeconds(1))
```
entangled90, line 130: :thought_balloon: I'm not sure if this is enough, as it just waits 1 second before checking the condition the first time, but it should retry afterwards anyway..

rodrigo-lourenco-lopes: We do retry after, at an unspecified poll interval, up until 600 seconds correct? Poll delay: "Specify the delay that will be used before Awaitility starts polling for the result the first time. If you don't specify a poll delay explicitly it'll be the same as the poll interval."
In the tests I did locally from the take backup command until we check this backup status was just a few miliseconds, thats why I though one second should be eneugh, or should it be larger?

entangled90: I don't think that by waiting a bit before checking the first time the flakyness would resolve. Awaitility will retry to check the condition every `pollDelay`.
Were you able to reproduce the flakyness on your local machine before doing any changes?

rodrigo-lourenco-lopes: Yes, but only when changing the code so that the get request is imediatly followed by the backup:

```
final var takeResponse = historyBackupClient.takeBackup(1L);
final var backupResponse = historyBackupClient.getBackup(1L);
 ```
 That's why I believe its just a race condition on the first get request being to fast after the backup is taken. 
 With the original test where we check the `takeResponse` contents first I was never able to reproduce, likely too slow on my machine.

entangled90: I think i get why it works now:
if the first call is done immediately it may throw an exception, which is not catched.
By adding a second of sleep this is exception is not thrown.
I think we can get rid of the initial `pollDelay`, but catch exception thrown by `backupClient.getBackup(1L)`. 
It should work as long as the exception is considered an `AssertionFailure`

entangled90 (approved):
:rocket: Thank you very much!

qa/integration-tests/src/test/java/io/camunda/it/backup/BackupRestoreIT.java
rodrigo-lourenco-lopes: Thanks! Implemented your suggestion :)

PR #28223, opened 2025-02-17
feat: add new REST API endpoints for fetching resource content + metadata

### LINKED ISSUE #25281: RPA files can be fetched through the `/TBD` REST API endpoint

jfriedenstab:
## Description
Add new REST API endpoints for fetching the content and metadata of a deployed RPA resource.

## Related issues

closes #25281

### DIFF
--- a/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
+++ b/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
@@ -1869,6 +1869,78 @@ paths:
             application/problem+json:
               schema:
                 $ref: "#/components/schemas/ProblemDetail"
+  /resources/{resourceKey}:
+    get:
+      tags:
+        - Resource
+      summary: Get resource
+      description: |
+        Returns a deployed resource.
+        :::info
+        Currently, this endpoint only supports RPA resources.
+        :::
+      parameters:
+        - name: resourceKey
+          in: path
+          required: true
+          description: The unique key identifying the resource.
+          schema:
+            type: string
+      responses:
+        "200":
+          description: The resource is successfully returned.
+          content:
+            application/json:
+              schema:
+                type: "#/components/schemas/ResourceResult"
+        "404":
+          description: A resource with the given key was not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "500":
+          description: An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+  /resources/{resourceKey}/content:
+    get:
+      tags:
+        - Resource
+      summary: Get resource content
+      description: |
+        Returns the content of a deployed resource.
+        :::info
+        Currently, this endpoint only supports RPA resources.
+        :::
+      parameters:
+        - name: resourceKey
+          in: path
+          required: true
+          description: The unique key identifying the resource.
+          schema:
+            type: string
+      responses:
+        "200":
+          description: The resource content is successfully returned.
+          content:
+            application/json:
+              schema:
+                type: string
+        "404":
+          description: A resource with the given key was not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "500":
+          description: An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
 
   /element-instances/{elementInstanceKey}/variables:
     put:
@@ -4914,6 +4986,28 @@ components:
       properties:
         signalKey:
           description: The unique ID of the signal that was broadcast.
+          type: string
+    ResourceResult:
+      type: object
+      properties:
+        resourceName:
+          description: The resource name from which this resource was parsed.
+          type: string
+        version:
+          description: The assigned resource version.
+          type: integer
+          format: int32
+        versionTag:
+          description: The version tag of this resource.
+          type: string
+        resourceId:
+          description: The resource ID of this resource.
+          type: string
+        tenantId:
+          description: The tenant ID of this resource.
+          type: string
+        resourceKey:
+          description: The unique key of this resource.
           type: string
   responses:
     InternalServerError:
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/ResourceController.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/ResourceController.java
@@ -10,12 +10,15 @@ package io.camunda.zeebe.gateway.rest.controller;
 import io.camunda.service.ResourceServices;
 import io.camunda.service.ResourceServices.DeployResourcesRequest;
 import io.camunda.service.ResourceServices.ResourceDeletionRequest;
+import io.camunda.service.ResourceServices.ResourceFetchRequest;
 import io.camunda.zeebe.gateway.impl.configuration.MultiTenancyCfg;
 import io.camunda.zeebe.gateway.protocol.rest.DeleteResourceRequest;
 import io.camunda.zeebe.gateway.rest.RequestMapper;
 import io.camunda.zeebe.gateway.rest.ResponseMapper;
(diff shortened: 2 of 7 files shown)


zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1869,6 +1869,78 @@ paths:
...
+  /resources/{resourceKey}:
+    get:
+      tags:
+        - Resource
+      summary: Get resource
+      description: |
+        Returns a deployed resource.
+        :::note
+        Currently, this endpoint only supports RPA resources.
+        :::
```
korthout, line 1881: 👍 It took me some time to verify whether these note admonitions are correctly rendered by our docs, because they aren't used in the current published docs, but they work!

🔧 I think we can increase the visibility to [INFO or WARNING](https://docusaurus.io/docs/markdown-features/admonitions).

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1869,6 +1869,78 @@ paths:
...
+      parameters:
+        - name: resourceKey
+          in: path
+          required: true
+          description: The assigned key of the resource, which acts as a unique identifier for this resource.
```
korthout, line 1886: 🔧 All keys in Camunda 8 are assigned. Let's align this to the other usage of the term key in resources: 

```suggestion
          description: The unique key identifying the resource.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4915,6 +4987,28 @@ components:
         signalKey:
           description: The unique ID of the signal that was broadcast.
           type: string
+    ResourceResult:
+      type: object
+      properties:
+        resourceName:
+          description: Resource name of this resource.
```
korthout, line 4994: 🔧 Let's align to other endpoints

```suggestion
          description: The resource name from which this resource was parsed.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4915,6 +4987,28 @@ components:
...
+          type: string
+        version:
+          description: Version of this resource.
```
korthout, line 4997: 🔧 Let's align to other endpoints:

```suggestion
          description: The assigned resource version.
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4915,6 +4987,28 @@ components:
...
+          type: integer
+          format: int32
+        versionTag:
+          description: Version tag of this resource.
```
korthout, line 5001: 🔧 Let's align to other endpoints:

```suggestion
          description: The version tag of the resource, as parsed during deployment.
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/ResourceController.java
```diff
@@ -58,6 +60,18 @@ public CompletableFuture<ResponseEntity<Object>> deleteResource(
...
+  public CompletableFuture<ResponseEntity<Object>> getResource(
+      @PathVariable final long resourceKey) {
+    return getResource(new ResourceFetchRequest(resourceKey));
+  }
+
+  @CamundaGetMapping(path = "/resources/{resourceKey}/content")
+  public CompletableFuture<ResponseEntity<Object>> getResourceContent(
+      @PathVariable final long resourceKey) {
+    return getResourceContent(new ResourceFetchRequest(resourceKey));
+  }
```
korthout, line 73: 🔧 The methods called by these methods look a lot alike. There's some re-use possible:

```suggestion
  @CamundaGetMapping(path = "/resources/{resourceKey}")
  public CompletableFuture<ResponseEntity<Object>> getResource(
      @PathVariable final long resourceKey) {
    return RequestMapper.executeServiceMethod(
        () -> fetchResource(resourceKey), ResponseMapper::toGetResourceResponse);
  }

  @CamundaGetMapping(path = "/resources/{resourceKey}/content")
  public CompletableFuture<ResponseEntity<Object>> getResourceContent(
      @PathVariable final long resourceKey) {
    return RequestMapper.executeServiceMethod(
        () -> fetchResource(resourceKey), ResponseMapper::toGetResourceContentResponse);
  }

  private CompletableFuture<ResourceRecord> fetchResource(final long resourceKey) {
    return resourceServices
        .withAuthentication(RequestMapper.getAuthentication())
        .fetchResource(new ResourceFetchRequest(resourceKey));
  }
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1869,6 +1869,78 @@ paths:
...
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
+        "500":
+          description: An internal error occurred while processing the request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: "#/components/schemas/ProblemDetail"
```
korthout, line 1907: 🔧 There's no test case for the 500 responses in both endpoints. However, I think such tests are missing in general (I only found one for decision definition query controller). Might be nice to add, but not mandatory apparently.

korthout (approved):
Nicely done @jfriedenstab 🚀 

👍 I found no flaws in the implementation
👍 Both happy and failure cases are covered by tests

🔧 I have a few suggestions here and there. Please have a look, perhaps they are helpful.

LGTM 👍

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
jfriedenstab: Actually, they are already used in this `rest-api.yaml` file for a couple of endpoints that are marked as `alpha` 🙂 (e.g. https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/query-decision-definitions-alpha/).

I've increased the visibility to `INFO` (I wouldn't say that it's worth a `WARNING`).

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/ResourceController.java
jfriedenstab: Yep, good idea 👍🏻

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
jfriedenstab: I added tests for this case.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
jfriedenstab: I took the description from the `processDefinitionKey` in [this endpoint](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/get-process-definition/), but I'm happy to use a shorter one.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
jfriedenstab: I noticed that we use "as parsed during deployment" in some other places, but I'm not sure if it's really useful here, given that the version tag is a custom string provided by the user at development time.

jfriedenstab:
Thanks a lot for your review, @korthout! I've addressed your comments.

By the way, I needed to add https://github.com/camunda/camunda/pull/28223/commits/4e93af842bf4c3b2b595e36a9c24df5d682546af which I missed in https://github.com/camunda/camunda/pull/27981.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
korthout: I don't have a strong opinion on this. If you look around the rest-api doc, the usage of the term ID is anyways not documented consistently. However, I'd rather see it aligned on any of the currently used conventions rather than adding another.

korthout (approved):
Thanks for the changes @jfriedenstab 👍 

LGTM 💯

PR #28226, opened 2025-02-17
Camunda Client out of the box Basic Auth support

### LINKED ISSUE #26982: Support for Basic authentication for REST and GRPC in camunda Client & SDK
The Zeebe/Camunda client & sdk should allow to configure basic auth credentials in case the API is secured by such.

Acceptance Critieria:
- [x] basic auth credentials can be configured on the bare camunda client using java api
- [x] basic auth credentials can be configured on the bare camunda client using environment variables
- [x] basic auth credentials can be configured on via properties of the spring sdk
- [x] basic auth credentials, when configured, are sent eagerly with every GRPC request
- [x] basic auth credentials, when configured, are sent eagerly with every REST API request

remcowesterhoud:
## Description

If a user wants to use basic auth they'd always have to create their own credentials provider. Since with the Identity rearchitecture basic auth is valid and fully supported option, we can make this easier by providing a `BasicAuthCredentialsProvider`. This provider takes a username and password and will set these on the `Authorization` header of the request.

I've also added back basic auth support in the Spring SDK. The username and password properties were already configurable, though never used. Now they have a use again.
We used to have basic auth support in the Spring SDK before, but it was called Simple auth. I chose to rename it to basic auth, as that's what it's called 😄 

## Checklist

- [ ] for CI changes:

## Related issues

closes #26982

### DIFF
--- a/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
+++ b/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
@@ -39,6 +39,11 @@ public class CredentialsProviderConfiguration {
   @ConditionalOnMissingBean
   public CredentialsProvider camundaClientCredentialsProvider(
       final CamundaClientProperties camundaClientProperties) {
+    return buildOAuthCredentialsProvider(camundaClientProperties);
+  }
+
+  private CredentialsProvider buildOAuthCredentialsProvider(
+      final CamundaClientProperties camundaClientProperties) {
     final OAuthCredentialsProviderBuilder credBuilder =
         CredentialsProvider.newCredentialsProviderBuilder()
             .applyEnvironmentOverrides(false)
--- a/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
+++ b/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
@@ -22,7 +22,6 @@ import io.camunda.client.impl.NoopCredentialsProvider;
 import io.camunda.client.impl.basicauth.BasicAuthCredentialsProviderBuilder;
 import io.camunda.client.impl.oauth.OAuthCredentialsProviderBuilder;
 import io.camunda.spring.client.properties.CamundaClientProperties;
-import io.camunda.spring.client.properties.CamundaClientProperties.ClientMode;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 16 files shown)


remcowesterhoud:
@jonathanlukas It'd be great if you could have a look at the Spring SDK files 🙂

clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
```diff
@@ -39,6 +41,34 @@ public class CredentialsProviderConfiguration {
   @ConditionalOnMissingBean
   public CredentialsProvider camundaClientCredentialsProvider(
       final CamundaClientProperties camundaClientProperties) {
+    if (camundaClientProperties.getMode() == ClientMode.basic) {
+      return buildBasicAuthCredentialsProvider(camundaClientProperties);
+    } else {
+      return buildOAuthCredentialsProvider(camundaClientProperties);
```
jonathanlukas, line 47: We could use a switch - case logic here for the client mode:

saas, self-managed: OAuth
basic: BasicAuch
default: Noop

clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
```diff
@@ -39,6 +40,33 @@ public class CredentialsProviderConfiguration {
   @ConditionalOnMissingBean
   public CredentialsProvider camundaClientCredentialsProvider(
       final CamundaClientProperties camundaClientProperties) {
+    return switch (camundaClientProperties.getMode()) {
+      case basic -> buildBasicAuthCredentialsProvider(camundaClientProperties);
+      case saas, selfManaged -> buildOAuthCredentialsProvider(camundaClientProperties);
```
jonathanlukas, line 45: The case `null` is not covered here.

remcowesterhoud: There's no reason. If the mode return `null` the switch would throw a `NullPointerException`. But even then it is impossible to be `null` as the `CamundaClientPropertiesPostProcessor` will throw an `IllegalArgumentException` if it finds an invalid client mode.

remcowesterhoud: Actually the tests do fail in case of null. Not sure why it sometimes goes through the properties post processor and sometimes nowt 🤔 

Anyway I'll handle it.

jonathanlukas (approved):
cool

clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
```diff
@@ -39,6 +40,37 @@ public class CredentialsProviderConfiguration {
...
+        new BasicAuthCredentialsProviderBuilder()
+            .applyEnvironmentOverrides(false)
+            .username(username)
+            .password(password);
+
+    try {
+      return builder.build();
+    } catch (final Exception e) {
+      LOG.warn("Failed to configure credential provider", e);
+      return new NoopCredentialsProvider();
```
jonathanlukas, line 68: As we can now use the noop provider by setting no mode, this could throw to prevent configuration issues.

jonathanlukas (commented):
We could discuss my comments

clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CredentialsProviderConfiguration.java
jonathanlukas: I see.. let's leave it that way.

nicpuppa (approved):

PR #28233, opened 2025-02-17
[Backport stable/optimize-8.7] test: support ad-hoc subprocesses

abremard:
## Description

In Camunda 8.7, a new BPMN element was introduced : the [ad-hoc subprocess](https://www.omg.org/spec/BPMN/2.0/PDF#13.2.5%20Ad-Hoc%20Sub-Process). It is supported in Zeebe but hasn't been tested in Optimize yet. This PR adds integration tests for the new element, to validate that an ad-hoc subprocess definition as well as process instances can be exported to Optimize for evaluation.

See [thread](https://camunda.slack.com/archives/CUZUYDMMG/p1739181377577979) for details on the test approach.

## Related issues

related to https://github.com/camunda/camunda/issues/25639

### DIFF
--- a/optimize/backend/pom.xml
+++ b/optimize/backend/pom.xml
@@ -81,6 +81,11 @@
     <!-- Zeebe dependencies -->
     <dependency>
       <groupId>io.camunda</groupId>
+      <artifactId>zeebe-bpmn-model</artifactId>
+      <version>8.8.0-SNAPSHOT</version>
+    </dependency>
+    <dependency>
+      <groupId>io.camunda</groupId>
       <artifactId>zeebe-protocol</artifactId>
       <version>${zeebe.version}</version>
     </dependency>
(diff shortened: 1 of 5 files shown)


optimize/backend/pom.xml
```diff
@@ -79,6 +79,11 @@
     </dependency>
 
     <!-- Zeebe dependencies -->
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>zeebe-bpmn-model</artifactId>
+      <version>8.8.0-SNAPSHOT</version>
```
RomanJRW, line 85: we would also need this to be updated to 8.7.0 on this branch too right?

optimize/pom.xml
```diff
@@ -17,7 +17,7 @@
     <zeebe.version>8.7.0-alpha4</zeebe.version>
     <identity.version>8.7.0-alpha4</identity.version>
     <!-- We use this for the Zeebe test container version only -->
-    <zeebe.docker.version>${zeebe.version}</zeebe.docker.version>
+    <zeebe.docker.version>8.7.0-alpha4</zeebe.docker.version>
```
RomanJRW, line 20: I don't think we need to change this do we? The `zeebe.version` property is already set to 8.7.0-alpha4

RomanJRW (approved):
Generally LGTM, pre-approving with some small comments! 👍

optimize/backend/pom.xml
abremard: Yes, but after checking with @buccarel regarding [this](https://github.com/camunda/camunda/pull/27968#discussion_r1955815740), the expected behavior is that once the single app goes live, we wouldn't need to pin a version here (= your assumption is correct), and we will get an error on build telling us to resolve the dependency convergence

optimize/pom.xml
abremard: Oops, that's an oversight from cherry-picking the changes from `main`, good catch

PR #28243, opened 2025-02-17
fix: always close exporter container and distribution service

lenaschoenburg:
Fixes two bugs:
1. `ExporterContainer`s were not closed by followers
3. `ExporterStateDistributionService` was not closed by leaders.

### DIFF
--- a/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
+++ b/zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
@@ -407,11 +407,8 @@ public final class ExporterDirector extends Actor implements HealthMonitorable, 
   @Override
   protected void onActorCloseRequested() {
     isOpened.set(false);
-    if (exporterMode == ExporterMode.ACTIVE) {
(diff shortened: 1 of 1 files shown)


zeebe/broker/src/main/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirector.java
```diff
@@ -407,11 +407,8 @@ protected void onActorClosed() {
   @Override
   protected void onActorCloseRequested() {
     isOpened.set(false);
-    if (exporterMode == ExporterMode.ACTIVE) {
-      containers.forEach(ExporterContainer::close);
-    } else {
-      exporterDistributionService.close();
-    }
+    containers.forEach(ExporterContainer::close);
```
entangled90, line 410: :wrench: `close` may throw, we should make sure that we call close on everything neverthless

lenaschoenburg: `ExporterContainer#close` doesn't throw, it logs all errors and doesn't rethrow them.

entangled90 (approved):
:rocket:

PR #28255, opened 2025-02-17
feat: make IncidentUpdateTask async

### LINKED ISSUE #27894: Make Incident update task async
`IncidentUpdateTask` is calling `.join()` on every `CompletableFuture`  returned by ES/OS, we should make it completely async to reduce resouce usage and parallelize what we can

entangled90:
## Description
The purpose of this PR is refactor the IncidentUpdateTask in a way that it does not block any threads with `.join()` calls and try to parallelize as much as possible.
In order to perform causal ordering when processing incidents, only one incident is processed at a given time.
A couple of tasks have been parallelized and are mostly about fetching necessary data before processing the incidents:
- `searchForInstances > checkDataAndCollectParentTreePaths` : data is collected in parallel 
- `processIncident > createFlowNodeInstanceUpdates`: flow node instances are fetched in parallel if necessary

All async operation like `thenComposeAsync` are done in the provided executor, which is the executor created in `BackgroundManager`. The size of the executor has been kept the same even if the task is not blocking anymore, to avoid starving other threads.
Considering that some tasks may be spawned in parallel, this task could use more than 1 thread, but in general it should be non-blocking so it should not use 1 thread in average.

A simple mechanism for trying to gracefully shutdown this tasks have been added:
before rescheduling any task check if it's been cancelled and it if it, it will not reschedule itself.
The executor is then terminated "gracefully" with `shutdown` first, giving some time for the tasks to stop gracefully before running `shutdownNow`, if they are not yet terminated.
This mechanism can be removed if not deemed necessary

### Comments
The code is much more hard to follow and understand after this change and the parallelism that we gain is really just limited to a couple of places. 
So I think it would be ok if we decide to keep using a synchronous style except  for those two places, where we can actually gain some performance.

## Related issues
closes #27894

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
@@ -12,9 +12,11 @@ import io.camunda.exporter.tasks.batchoperations.BatchOperationUpdateRepository;
 import io.camunda.exporter.tasks.incident.IncidentUpdateRepository;
 import io.camunda.zeebe.util.CloseableSilently;
 import io.camunda.zeebe.util.VisibleForTesting;
+import java.time.Duration;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.WillCloseWhenClosed;
 import org.agrona.CloseHelper;
 import org.slf4j.Logger;
@@ -26,7 +28,8 @@ public final class BackgroundTaskManager implements CloseableSilently {
   private final BatchOperationUpdateRepository batchOperationUpdateRepository;
   private final Logger logger;
   private final ScheduledThreadPoolExecutor executor;
-  private final List<Runnable> tasks;
+  private final List<RunnableTask> tasks;
+  private final Duration closeTimeout;
 
   private int submittedTasks = 0;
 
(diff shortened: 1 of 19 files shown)


npepinpe:
I will take over the PR this week.

entangled90:
This PR now includes #31624 as well.

rodrigo-lourenco-lopes (approved):
Thanks for the great improvement 🚀 
I only reviewed the last 5 commits added by @panagiotisgts related with spliting parallel and batched paths in IncidentUpdate and associated reactors as discussed [here](https://camunda.slack.com/archives/C06F0GLJNFM/p1746524312396799)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
```diff
@@ -38,7 +41,8 @@ public final class BackgroundTaskManager implements CloseableSilently {
       final @WillCloseWhenClosed BatchOperationUpdateRepository batchOperationUpdateRepository,
       final Logger logger,
       final @WillCloseWhenClosed ScheduledThreadPoolExecutor executor,
-      final List<Runnable> tasks) {
+      final List<RunnableTask> tasks,
+      final Duration closeTimeout) {
```
rodrigo-lourenco-lopes, line 45: ❓ do we expect to make this configurable?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/incident/IncidentUpdateTask.java
```diff
@@ -81,30 +94,55 @@ public String getCaption() {
     return "Incident update task";
   }
 
-  private int processNextBatch() {
+  /**
```
rodrigo-lourenco-lopes, line 97: Nice to add docs here 🚀

rodrigo-lourenco-lopes (approved):
Thanks for the comments in the PR description and the code, help me a lot in trying to follow the changes 🙌

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTaskManager.java
entangled90: I don't think so, it's just a timeout when closing the jobs.

Issue #28219, opened 2025-02-17
Ad-Hoc Subprocess: Implement activate activity


maff:
## Description

As a user (human or AI agent), I want to be able to activate an activity within an instance of an ad-hoc subprocess.

## Acceptance Criteria

- **Given** a running instance key of an ad-hoc subprocess **and** a flow node that exists within the ad-hoc subprocess, **when** activation is triggered, **then** the flow node should be transitioned to the active state and it should start executing.
- **Given** a running instance key of an ad-hoc subprocess **and** multiple flow nodes that exists within the ad-hoc subprocess, **when** activation is triggered, **then** the selected flow nodes should be transitioned to the active state and they should start executing.
- **Given** a running instance key of an ad-hoc subprocess, **and** a the user attempts to activate a flow node that does not exist in the given ad-hoc-subprocess, **when** attempting to activate, **then** no changes should occur to the ad-hoc suprocess **and** an error message is returned.
- ~~**Given** a running instance key of an ad-hoc subprocess, **and** the given flow node is already in an active state, **when** attempting to activate the flow node, **then** the activation is rejected **and** an error message stating that an activated flow node is already activated is returned.~~ (Removed because it is desired that an activity can be activated multiple times)
- **Given** an instance key of an ad-hoc subprocess that has already completed execution, **when** attempting to activate flow nodes, **then** the activation is rejected **and** an error message stating that activation of flow nodes in an ad-hoc subprocess that has been completed isn't allowed.
- **Given** an instance key of an ad-hoc subprocess that has been terminated, **when** attempting to activate flow nodes, **then** the activation is rejected **and** an error message stating that activation of flow nodes in an ad-hoc subprocess that has been terminated isn't allowed.
- **Given** an instance key of an ad-hoc subprocess that has not started, **when** attempting to activate flow nodes, **then** the activation is rejected **and** an error message stating that activation of flow nodes in an ad-hoc subprocess that hasn't started isn't allowed.
- **Given** an instance key of an ad-hoc suprocess that does not exist, **when** attempting to activate the flow nodes, **then** the activation is rejected **and** an error message stating that the ad-hoc subprocess does not exist is returned.
- **Given** a running instance of an ad-hoc subprocess **and** duplicate flow nodes, **when** attempting to activate duplicate flow nodes, **then** the activation is rejected **and** an error message stating that duplicate flow node ids are not allowed is returned.

### Technical Tasks

- Broker: Implement new activate activity record and record processor
[...]

Issue #25192, opened 2024-11-25
I can define custom headers for an execution listener 


saig0:
**Is your feature request related to a problem? Please describe.**
Currently, I can't define custom headers for execution listeners. I can define only the event type (start | end), the job type, and the retries ([ref](https://docs.camunda.io/docs/components/concepts/execution-listeners/#define-an-execution-listener)). For other jobs, I can define custom headers to pass additional information to the job worker ([ref](https://docs.camunda.io/docs/components/concepts/job-workers/)). 

Because of this limitation, I can't use **connectors** as execution listeners. In many cases, the connectors pass configuration information as custom headers, for example, for the result expression or the error expression.    

**Describe the solution you'd like**
I can define custom headers for execution listeners in the same way as for other jobs. So in the end, I can use my connectors as execution listeners. 

**Describe alternatives you've considered**
None. Currently, the connectors depend on the custom headers.

**Additional context**
Example: the configuration of a GitHub outbound connector to create an issue.

<details>
  <summary>BPMN XML</summary>

![Screenshot from 2024-11-25 15-04-13](https://github.com/user-attachments/assets/9812bcd5-cdeb-41bf-a371-302ef1c7fb90)

</details>

sbuettner:
Thanks for creating this issue @saig0. In addition to the headers we would also need `zeebe:ioMapping` support in case we need additional/dynamic config properties as required by our existing Connectors.  What do you think about extending this issue?

cc @mattli2024 This feature is the enabler for working on Connectors as listeners.

xevien96:
This ticket is also a nice enabler for some migration use cases where we need to put delegate expressions from conditional sequence flows into the execution listener of gateways to have the decision ready as a local process variable. Having custom headers will enable the c7-adapter way of dealing with that.

saig0:
> In addition to the headers we would also need zeebe:ioMapping support in case we need additional/dynamic config properties as required by our existing Connectors. What do you think about extending this issue?

@sbuettner I see that the variable mappings would be necessary too. However, it doesn't fit the lifecycle to evaluate the mappings when creating the job for the execution listener. Input mappings create local variables of the BPMN element.

It might be an option to extend the input mappings when adding a connector template as an execution listener.

sbuettner:
> However, it doesn't fit the lifecycle to evaluate the mappings when creating the job for the execution listener.
[...]

PR #28272, opened 2025-02-18
Ad-Hoc Subprocess: Search activatable activities REST API

### LINKED ISSUE #27930: Ad-Hoc Subprocess: Implement search activities REST API based on process definition
## Description

Search for BPMN elements inside an ad-hoc subprocess that can be activated. 

Only elements which can be activated should be returned by the API call (root nodes within the ad-hoc subprocess). This first iteration should return a list of activities for an ad-hoc subprocess from reading the process definition -- this will be enhanced to be based on process instance data and to include paging/sorting in #28432.

- Update OpenAPI specification
- Implement controller + service layers
- Update client
- TBD: documentation

## API Specification

**URL**: http://localhost:8080/v2/ad-hoc-activities/search
**Method**: POST
**Request body**:
```json
{
   "filter":{
      "processDefinitionKey":0,
      "adhocSubprocessId": "string"
   }
}
```

**Response body**:
```json
{
   "items":[
      {
         "processDefinitionKey":0,
         "processDefinitionId":"string",         
         "adhocSubprocessId":"string",
         "flowNodeId":"string",
         "flowNodeName":"string",
         "type":"UNSPECIFIED",
         "description": "string",
         "tenantId":"string"
      }
   ]
}
```

maff:
## Description

Implements a REST endpoint to search for activities within an ad-hoc subprocess.

While the final implementation of this endpoint should include process instance data and list all activities within running ad-hoc subprocesses, this first iteration returns a static list of activities as resolved from a process definition.

Therefore this endpoint requires both `processDefinitionKey` and `adHocSubprocessId` on the filter object and does not allow any sorting/paging yet.

Example request:

```json5
// POST http://localhost:8080/v2/element-instances/ad-hoc-activities/search
{
  "filter": {
    "processDefinitionKey": 2251799813685269,
    "adHocSubprocessId": "MyAdHocSubProcess"
  }
}
```

Example response:

```json5
{
  "items": [
    {
      "processDefinitionKey": "2251799813685269",
      "processDefinitionId": "TestProcessId",
      "adHocSubprocessId": "MyAdHocSubProcess",
      "elementId": "fooCalculation",
      "elementName": "foo calculation",
      "documentation": "This is a foo calculation",
      "type": "SCRIPT_TASK",
      "tenantId": "<default>"
    },
    {
      "processDefinitionKey": "2251799813685269",
      "processDefinitionId": "TestProcessId",
      "adHocSubprocessId": "AdHocSubProcess",
      "elementId": "barCalculation",
      "elementName": "bar calculation",
      "documentation": "",
      "type": "SCRIPT_TASK",
      "tenantId": "<default>"
    }
  ]
}
```

## Related issues

closes #27930

### DIFF
--- a/clients/java/src/main/java/io/camunda/client/CamundaClient.java
+++ b/clients/java/src/main/java/io/camunda/client/CamundaClient.java
@@ -79,6 +79,7 @@ import io.camunda.client.api.fetch.UserTaskGetRequest;
 import io.camunda.client.api.fetch.VariableGetRequest;
 import io.camunda.client.api.response.ActivatedJob;
 import io.camunda.client.api.response.DocumentReferenceResponse;
+import io.camunda.client.api.search.query.AdHocSubprocessActivityQuery;
 import io.camunda.client.api.search.query.DecisionDefinitionQuery;
 import io.camunda.client.api.search.query.DecisionInstanceQuery;
 import io.camunda.client.api.search.query.DecisionRequirementsQuery;
@@ -850,6 +851,54 @@ public interface CamundaClient extends AutoCloseable, JobClient {
    * @return a builder for the request to get a flow node instance
    */
   FlowNodeInstanceGetRequest newFlowNodeInstanceGetRequest(long flowNodeInstanceKey);
+
+  /**
+   * Executes a search request to query activities within ad-hoc subprocesses.
+   *
+   * <p>Note that this API currently requires filters for both process definition key and ad-hoc
+   * subprocess ID and does not support paging or sorting.
+   *
+   * <pre>
+   * long processDefinitionId = ...;
+   * String adHocSubprocessId = ...;
+   *
+   * camundaClient
+   *  .newAdHocSubprocessActivityQuery()
+   *  .filter((f) -> f
+   *     .processDefinitionId(processDefinitionId)
+   *     .adHocSubprocessId(adHocSubprocessId)
+   *  )
+   *  .send();
+   * </pre>
+   *
+   * @return a builder for the ad-hoc subprocess activity query
+   */
+  @ExperimentalApi("https://github.com/camunda/camunda/issues/27930")
+  AdHocSubprocessActivityQuery newAdHocSubprocessActivityQuery();
+
+  /**
+   * Executes a search request to query activities within ad-hoc subprocesses.
+   *
+   * <p>Note that this API currently requires filters for both process definition key and ad-hoc
+   * subprocess ID and does not support paging or sorting.
+   *
+   * <pre>
+   * long processDefinitionId = ...;
+   * String adHocSubprocessId = ...;
+   *
+   * camundaClient
+   *  .newAdHocSubprocessActivityQuery(
+   *    processDefinitionId,
+   *    adHocSubprocessId
+   *  )
+   *  .send();
+   * </pre>
+   *
+   * @return a builder for the ad-hoc subprocess activity query
+   */
+  @ExperimentalApi("https://github.com/camunda/camunda/issues/27930")
+  AdHocSubprocessActivityQuery newAdHocSubprocessActivityQuery(
+      long processDefinitionKey, String adHocSubprocessId);
 
   /**
    * Executes a search request to query user tasks.
--- a/clients/java/src/main/java/io/camunda/client/CamundaClient.java
+++ b/clients/java/src/main/java/io/camunda/client/CamundaClient.java
@@ -859,13 +859,13 @@ public interface CamundaClient extends AutoCloseable, JobClient {
    * subprocess ID and does not support paging or sorting.
    *
    * <pre>
-   * long processDefinitionId = ...;
+   * long processDefinitionKey = ...;
    * String adHocSubprocessId = ...;
    *
    * camundaClient
    *  .newAdHocSubprocessActivityQuery()
    *  .filter((f) -> f
-   *     .processDefinitionId(processDefinitionId)
+   *     .processDefinitionKey(processDefinitionKey)
    *     .adHocSubprocessId(adHocSubprocessId)
    *  )
    *  .send();
@@ -883,12 +883,12 @@ public interface CamundaClient extends AutoCloseable, JobClient {
    * subprocess ID and does not support paging or sorting.
    *
    * <pre>
-   * long processDefinitionId = ...;
+   * long processDefinitionKey = ...;
    * String adHocSubprocessId = ...;
    *
    * camundaClient
    *  .newAdHocSubprocessActivityQuery(
-   *    processDefinitionId,
+   *    processDefinitionKey,
    *    adHocSubprocessId
    *  )
    *  .send();
--- a/clients/java/src/test/java/io/camunda/client/adhocsubprocess/AdHocSubprocessActivitySearchTest.java
+++ b/clients/java/src/test/java/io/camunda/client/adhocsubprocess/AdHocSubprocessActivitySearchTest.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright © 2017 camunda services GmbH (info@camunda.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.client.adhocsubprocess;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
+
+import io.camunda.client.api.search.response.AdHocSubprocessActivityResponse;
+import io.camunda.client.api.search.response.AdHocSubprocessActivityResponse.AdHocSubprocessActivity;
+import io.camunda.client.api.search.response.AdHocSubprocessActivityResponse.AdHocSubprocessActivity.AdHocSubprocessActivityType;
+import io.camunda.client.protocol.rest.AdHocSubprocessActivityResult;
+import io.camunda.client.protocol.rest.AdHocSubprocessActivityResult.TypeEnum;
+import io.camunda.client.protocol.rest.AdHocSubprocessActivitySearchQuery;
+import io.camunda.client.protocol.rest.AdHocSubprocessActivitySearchQueryResult;
+import io.camunda.client.util.ClientRestTest;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.Test;
+
+public class AdHocSubprocessActivitySearchTest extends ClientRestTest {
+
+  private static final Long PROCESS_DEFINITION_KEY = 2251799813685281L;
+  private static final String PROCESS_DEFINITION_ID = "TestParentAdHocSubprocess";
+  private static final String AD_HOC_SUBPROCESS_ID = "TestAdHocSubprocess";
+
+  @Test
+  void shouldSearchAdHocSubprocessActivities() {
+    // given
+    final AdHocSubprocessActivitySearchQueryResult searchQueryResult =
+        new AdHocSubprocessActivitySearchQueryResult();
+    searchQueryResult.addItemsItem(
+        adHocSubprocessActivityResult(
+            r -> {
+              r.processDefinitionKey(String.valueOf(PROCESS_DEFINITION_KEY));
+              r.processDefinitionId(PROCESS_DEFINITION_ID);
+              r.setAdHocSubprocessId(AD_HOC_SUBPROCESS_ID);
+              r.setFlowNodeId("task1");
+              r.setFlowNodeName("Task #1");
+              r.setType(TypeEnum.SERVICE_TASK);
+              r.setDocumentation("The first task in the ad-hoc subprocess");
+              r.setTenantId("<default>");
+            }));
+    searchQueryResult.addItemsItem(
+        adHocSubprocessActivityResult(
+            r -> {
+              r.processDefinitionKey(String.valueOf(PROCESS_DEFINITION_KEY));
+              r.processDefinitionId(PROCESS_DEFINITION_ID);
+              r.setAdHocSubprocessId(AD_HOC_SUBPROCESS_ID);
+              r.setFlowNodeId("task2");
+              r.setFlowNodeName("Task #2");
+              r.setType(TypeEnum.USER_TASK);
+              r.setDocumentation("The second task in the ad-hoc subprocess");
+              r.setTenantId("<default>");
+            }));
+
+    gatewayService.onAdHocSubprocessActivitySearch(searchQueryResult);
+
+    // when
+    final AdHocSubprocessActivityResponse response =
+        client
+            .newAdHocSubprocessActivityQuery(PROCESS_DEFINITION_KEY, AD_HOC_SUBPROCESS_ID)
+            .send()
+            .join();
+
+    // then
+    final AdHocSubprocessActivitySearchQuery request =
+        gatewayService.getLastRequest(AdHocSubprocessActivitySearchQuery.class);
+    assertThat(request.getFilter().getProcessDefinitionKey())
+        .isEqualTo(PROCESS_DEFINITION_KEY.toString());
+    assertThat(request.getFilter().getAdHocSubprocessId()).isEqualTo(AD_HOC_SUBPROCESS_ID);
+
+    assertThat(response.getItems())
+        .hasSize(2)
+        .extracting(
+            AdHocSubprocessActivity::getProcessDefinitionKey,
+            AdHocSubprocessActivity::getProcessDefinitionId,
+            AdHocSubprocessActivity::getAdHocSubprocessId,
+            AdHocSubprocessActivity::getFlowNodeId,
+            AdHocSubprocessActivity::getFlowNodeName,
+            AdHocSubprocessActivity::getType,
+            AdHocSubprocessActivity::getDocumentation,
+            AdHocSubprocessActivity::getTenantId)
+        .containsExactly(
+            tuple(
+                PROCESS_DEFINITION_KEY,
+                PROCESS_DEFINITION_ID,
+                AD_HOC_SUBPROCESS_ID,
+                "task1",
+                "Task #1",
+                AdHocSubprocessActivityType.SERVICE_TASK,
+                "The first task in the ad-hoc subprocess",
+                "<default>"),
+            tuple(
+                PROCESS_DEFINITION_KEY,
+                PROCESS_DEFINITION_ID,
+                AD_HOC_SUBPROCESS_ID,
+                "task2",
+                "Task #2",
+                AdHocSubprocessActivityType.USER_TASK,
+                "The second task in the ad-hoc subprocess",
+                "<default>"));
+  }
+
+  @Test
+  void shouldMapUnknownEnumValueToFallbackValueWithoutBreaking() {
+    // given
+    final String responseJson =
+        "{\n"
+            + "  \"items\": [\n"
+            + "    {\n"
+            + "      \"processDefinitionKey\": \"2251799813685281\",\n"
+            + "      \"processDefinitionId\": \"TestParentAdHocSubprocess\",\n"
+            + "      \"adHocSubprocessId\": \"TestAdHocSubprocess\",\n"
+            + "      \"flowNodeId\": \"unknownTask\",\n"
+            + "      \"flowNodeName\": \"Unknown Task\",\n"
+            + "      \"type\": \"START_EVENT\",\n"
+            + "      \"tenantId\": \"<default>\"\n"
+            + "    }\n"
+            + "  ]\n"
+            + "}";
+
+    gatewayService.onAdHocSubprocessActivitySearch(responseJson);
+
+    // when
+    final AdHocSubprocessActivityResponse response =
+        client
+            .newAdHocSubprocessActivityQuery(PROCESS_DEFINITION_KEY, AD_HOC_SUBPROCESS_ID)
+            .send()
+            .join();
+
+    // then
+    assertThat(response.getItems())
+        .hasSize(1)
+        .extracting(
+            AdHocSubprocessActivity::getProcessDefinitionKey,
+            AdHocSubprocessActivity::getProcessDefinitionId,
+            AdHocSubprocessActivity::getAdHocSubprocessId,
+            AdHocSubprocessActivity::getFlowNodeId,
+            AdHocSubprocessActivity::getFlowNodeName,
+            AdHocSubprocessActivity::getType,
+            AdHocSubprocessActivity::getDocumentation,
+            AdHocSubprocessActivity::getTenantId)
+        .containsExactly(
+            tuple(
+                PROCESS_DEFINITION_KEY,
+                PROCESS_DEFINITION_ID,
+                AD_HOC_SUBPROCESS_ID,
+                "unknownTask",
+                "Unknown Task",
+                AdHocSubprocessActivityType.UNKNOWN_ENUM_VALUE,
+                null,
+                "<default>"));
+  }
+
+  private static AdHocSubprocessActivityResult adHocSubprocessActivityResult(
+      final Consumer<AdHocSubprocessActivityResult> consumer) {
+    final AdHocSubprocessActivityResult result = new AdHocSubprocessActivityResult();
+    consumer.accept(result);
+    return result;
+  }
+}
--- a/clients/java/src/test/java/io/camunda/client/adhocsubprocess/AdHocSubprocessActivitySearchTest.java
+++ b/clients/java/src/test/java/io/camunda/client/adhocsubprocess/AdHocSubprocessActivitySearchTest.java
@@ -46,8 +46,8 @@ public class AdHocSubprocessActivitySearchTest extends ClientRestTest {
               r.processDefinitionKey(String.valueOf(PROCESS_DEFINITION_KEY));
               r.processDefinitionId(PROCESS_DEFINITION_ID);
               r.setAdHocSubprocessId(AD_HOC_SUBPROCESS_ID);
-              r.setFlowNodeId("task1");
-              r.setFlowNodeName("Task #1");
+              r.setElementId("task1");
+              r.setElementName("Task #1");
               r.setType(TypeEnum.SERVICE_TASK);
               r.setDocumentation("The first task in the ad-hoc subprocess");
               r.setTenantId("<default>");
@@ -58,8 +58,8 @@ public class AdHocSubprocessActivitySearchTest extends ClientRestTest {
               r.processDefinitionKey(String.valueOf(PROCESS_DEFINITION_KEY));
               r.processDefinitionId(PROCESS_DEFINITION_ID);
               r.setAdHocSubprocessId(AD_HOC_SUBPROCESS_ID);
-              r.setFlowNodeId("task2");
-              r.setFlowNodeName("Task #2");
+              r.setElementId("task2");
+              r.setElementName("Task #2");
               r.setType(TypeEnum.USER_TASK);
               r.setDocumentation("The second task in the ad-hoc subprocess");
               r.setTenantId("<default>");
@@ -87,8 +87,8 @@ public class AdHocSubprocessActivitySearchTest extends ClientRestTest {
             AdHocSubprocessActivity::getProcessDefinitionKey,
             AdHocSubprocessActivity::getProcessDefinitionId,
             AdHocSubprocessActivity::getAdHocSubprocessId,
-            AdHocSubprocessActivity::getFlowNodeId,
-            AdHocSubprocessActivity::getFlowNodeName,
+            AdHocSubprocessActivity::getElementId,
+            AdHocSubprocessActivity::getElementName,
             AdHocSubprocessActivity::getType,
             AdHocSubprocessActivity::getDocumentation,
             AdHocSubprocessActivity::getTenantId)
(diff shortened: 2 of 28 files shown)


service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  public List<AdHocSubprocessActivity> findActivatableActivities(
+      final Long processDefinitionKey, final String adHocSubprocessId) {
+    final ProcessDefinitionEntity processDefinitionEntity =
+        processDefinitionServices.getByKey(processDefinitionKey);
+
+    // TODO should we cache this lookup somewhere? There are multiple ProcessCache implementations
+    // in operate, tasklist, gateway-rest, but none in this service layer. Implement caching later
+    // or introduce a process definition cache at this level?
```
maff, line 60: Maybe something which we can do in a dedicated PR later on if there is no proper cache which would fit our use case right now.

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,115 @@
...
+    final BpmnModelInstance modelInstance =
+        Bpmn.readModelFromStream(
+            new ByteArrayInputStream(
+                processDefinitionEntity.bpmnXml().getBytes(StandardCharsets.UTF_8)));
+
+    final var processElement = modelInstance.getModelElementById(adHocSubprocessId);
+    if (processElement instanceof final AdHocSubProcess adHocSubProcess) {
+      final var tasks = adHocSubProcess.getChildElementsByType(Task.class);
```
maff, line 68: Most likely this is not enough (e.g. we might also have call activities or other subprocesses in there) - do we need to filter at all here?

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,115 @@
...
+      return tasks.stream()
+          .filter(task -> task.getIncoming().isEmpty())
```
maff, line 70: no incoming flows = root node/can be activated

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,115 @@
...
+          .map(
+              task ->
+                  new AdHocSubprocessActivity(
+                      processDefinitionKey,
+                      adHocSubprocessId,
+                      task.getId(),
+                      task.getName(),
+                      BpmnElementType.bpmnElementTypeFor(task.getElementType().getTypeName()),
```
maff, line 78: This might need some more error handling.

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,115 @@
...
+            adHocSubprocessId,
+            processDefinitionKey);
+      } else {
+        LOGGER.warn(
+            "Found element with id '{}' in process definition with key '{}' but it was of type '{}', expected 'AdHocSubProcess'",
+            adHocSubprocessId,
+            processDefinitionKey,
+            processElement.getElementType());
+      }
+    }
```
maff, line 96: Mostly for debugging purposes for the moment. I didn't check how much we usually log, but we can also remove this before merging.

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,115 @@
...
+        "Failed to find Ad-Hoc Subprocess with ID '%s'".formatted(adHocSubprocessId));
+  }
+
+  private String extractDocumentation(final Task task) {
+    return task.getDocumentations().stream()
+        .map(ModelElementInstance::getTextContent)
+        .collect(Collectors.joining("\n"));
+  }
+
+  public record AdHocSubprocessActivity(
```
maff, line 108: I did not (yet) find a good package for these data objects for use cases not related to search so for the moment I just put this here also if it does not 100% fit how other services look like. Open to discuss :)

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1822,6 +1823,43 @@ paths:
         "500":
           $ref: "#/components/responses/InternalServerError"
 
+  /ad-hoc-activities/activatable:
```
maff, line 1826: `activatable` was one of the variants we discussed, but as this might not be a real term open to find a better one :)

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4796,6 +4834,60 @@ components:
...
+        - processDefinitionKey
+        - adHocSubprocessId
+      properties:
+        processDefinitionKey:
+          description: The process definition key.
+          type: string
+        adHocSubprocessId:
+          description: The ad-hoc subprocess process element ID.
+          type: string
+    AdHocSubprocessActivitySearchQueryResult:
```
maff, line 4850: As we're not doing a real search I opted for not including paging/sorting on this API. Open to discuss if we want this in terms of consistency, but I guess this would make the implementation more complex as we're not relying on the datastore here and would need to implement this directly.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4796,6 +4834,60 @@ components:
...
+        flowNodeId:
+          description: The flow node ID for this activity within the ad-hoc subprocess.
+          type: string
+        flowNodeName:
+          description: The flow node name for this activity within the ad-hoc subprocess.
+          type: string
+        documentation:
+          description: The documentation for this activity within the ad-hoc subprocess.
+          type: string
+        # TODO which types do we need to support here? e.g. (sub-)process?
```
maff, line 4876: Maybe we don't need to restrict it here and can use the whole set of types - do we have a reusable list somewhere?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/AdHocSubprocessActivityController.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private final AdHocSubprocessActivityServices adHocSubprocessActivityServices;
+
+  public AdHocSubprocessActivityController(
+      final AdHocSubprocessActivityServices adHocSubprocessActivityServices) {
+    this.adHocSubprocessActivityServices = adHocSubprocessActivityServices;
+  }
+
+  @CamundaPostMapping(path = "/activatable")
+  public ResponseEntity<AdHocSubprocessActivitySearchQueryResult> findActivatableActivities(
```
maff, line 35: TODO add some validation to the request body. Not sure if we're able to use bean validation on the generated models, but otherwise we can implement a simple validation function (validating adHocSubprocessId not blank for example) on the controller level as other controllers already do.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/controller/AdHocSubprocessActivityController.java
```diff
@@ -0,0 +1,63 @@
...
+      final var result = new AdHocSubprocessActivitySearchQueryResult();
+      result.setItems(activities.stream().map(this::toAdHocActivity).toList());
+
+      return ResponseEntity.ok(result);
+    } catch (final Exception e) {
+      return mapErrorToResponse(e);
+    }
+  }
+
+  private AdHocSubprocessActivityResult toAdHocActivity(final AdHocSubprocessActivity activity) {
```
maff, line 53: Similar to the record I opted for not adding this to the global response mappers but rather to keep this close to the controller code here. Open to discuss what fits better in terms of consistency, but for my initial consideration it was easier to read/follow when keeping this close (also because we're very likely to not reuse it outside of this controller).

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
maff: Filtering on `FlowNode` now as it also might include intermediate events.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
maff: Updated to include most types besides gateways and start/end events

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,110 @@
...
+
+  public SearchQueryResult<AdHocSubprocessActivityEntity> search(
+      final AdHocSubprocessActivityQuery query) {
+    final ProcessDefinitionEntity processDefinitionEntity =
+        processDefinitionServices.getByKey(query.filter().processDefinitionKey());
+
+    final BpmnModelInstance modelInstance =
+        Bpmn.readModelFromStream(
+            new ByteArrayInputStream(
+                processDefinitionEntity.bpmnXml().getBytes(StandardCharsets.UTF_8)));
```
maff, line 58: For this first iteration we agreed to not add any caching at this point. As this will be updated to also query process instance data in #28432 we can address that later.

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,110 @@
...
+
+    final var processElement =
+        modelInstance.getModelElementById(query.filter().adHocSubprocessId());
+    if (processElement instanceof final AdHocSubProcess adHocSubProcess) {
+      final var flowNodes = adHocSubProcess.getChildElementsByType(FlowNode.class);
+      final var rootActivities =
+          flowNodes.stream()
+              .filter(flowNode -> flowNode.getIncoming().isEmpty())
```
maff, line 66: This now filters any flow node which is in the definition and relies on the deployment validation to only return valid elements. Filtering on `Activity` would be too restrictive as having intermediate events as root node is a valid use case.

service/src/test/java/io/camunda/service/AdHocSubprocessActivityServicesTest.java
```diff
@@ -0,0 +1,145 @@
...
+      assertThatThrownBy(
+              () ->
+                  adHocSubprocessActivityServices.search(
+                      defaultSearchQuery(filter -> filter.adHocSubprocessId("nonExistingId"))))
+          .isInstanceOf(NotFoundException.class)
+          .hasMessage("Failed to find Ad-Hoc Subprocess with ID 'nonExistingId'");
+    }
+
+    @Test
+    void shouldThrowNotFoundExceptionWhenElementWithAdHocSubprocessIsUnexpectedType() {
```
maff, line 120:
```suggestion
    void shouldThrowNotFoundExceptionWhenElementWithAdHocSubprocessIdIsUnexpectedType() {
```

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1822,6 +1823,46 @@ paths:
...
+    post:
+      tags:
+        - Ad-hoc subprocess
+      operationId: findAdHocSubprocessActivities
+      summary: Query activatable activities within ad-hoc subprocesses
+      description: |
+        Search for activatable activities within ad-hoc subprocesses based on given criteria.
+
+        Note that this API currently requires filters for both process definition key and ad-hoc
+        subprocess ID and does not support paging or sorting.
```
maff, line 1836: The API tries to align to the other search APIs as much as possible (naming, structure, ...), but request/response models are a subset of the usual search ones without paging + sorting. Will be addressed in #28432.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4796,6 +4837,83 @@ components:
         incidentKey:
           description: Incident key associated with this flow node instance.
           type: string
+    AdHocSubprocessActivitySearchQuery:
+      description: Ad-hoc subprocess activities search request.
+      type: object
+      required:
+        - filter
```
maff, line 4844: For now, this is required, but we can loosen this later when implementing the enhanced query functionality in #28432.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4796,6 +4837,83 @@ components:
...
+            - RECEIVE_TASK
+            - USER_TASK
+            - MANUAL_TASK
+            - TASK
+            - MULTI_INSTANCE_BODY
+            - CALL_ACTIVITY
+            - BUSINESS_RULE_TASK
+            - SCRIPT_TASK
+            - SEND_TASK
+            - UNKNOWN
```
maff, line 4910: Currently everything which is also defined in the flow node instance queries minus start/end events, gateways, and sequence flows. Is there anything in here we should remove?

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RequestMapper.java
```diff
@@ -787,6 +795,41 @@ public static Either<ProblemDetail, TenantDTO> toTenantUpdateDto(
                 tenantUpdateRequest.getDescription()));
   }
 
+  public static Either<ProblemDetail, AdHocSubprocessActivityQuery> toAdHocSubprocessActivityQuery(
```
maff, line 798: Currently validating that all filter properties are required - restrictions will be removed in #28432.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4796,6 +4837,83 @@ components:
...
+    AdHocSubprocessActivityFilter:
+      description: Flow node instance filter.
+      type: object
+      required:
+        - processDefinitionKey
+        - adHocSubprocessId
+      properties:
+        processDefinitionKey:
+          description: The process definition key.
+          type: string
```
maff, line 4859: I wasn't completely sure how to handle this as in the code we need to do some casting between string and long on this property. Kept it as string here to be consistent with other APIs and if I remember correctly the plan is to treat all keys as strings in the future?

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1822,6 +1823,46 @@ paths:
         "500":
           $ref: "#/components/responses/InternalServerError"
 
+  /ad-hoc-activities/search:
+    post:
+      tags:
+        - Ad-hoc subprocess
+      operationId: findAdHocSubprocessActivities
+      summary: Query activatable activities within ad-hoc subprocesses
```
pepopowitz, line 1831:
```suggestion
      summary: Query activatable activities
```

🔧 On the docs side, we prefer for summaries to be relatively short, since they appear in the left nav, and wordier summaries can make it harder to find what they're looking for. The user can tell from the navigation hierarchy that they are within the Ad-hoc subprocesses, so my suggestion removes that bit of context. If there are other ways you can think of to shorten, that's great too. 

(Here are screenshots of the docs nav with the original, compared to my suggestion:)

image

image

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -4796,6 +4837,83 @@ components:
...
+          description: The ad-hoc subprocess process element ID.
+          type: string
+        flowNodeId:
+          description: The flow node ID for this activity.
+          type: string
+        flowNodeName:
+          description: The flow node name for this activity.
+          type: string
+        type:
+          description: Type of activity as defined set of values.
```
pepopowitz, line 4890:
```suggestion
          description: Type of activity with a defined set of values.
```

🔧 This wording aligns with other endpoints.

pepopowitz (approved):
👍 A few minor comments from the docs perspective!

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
```diff
@@ -1822,6 +1823,46 @@ paths:
         "500":
           $ref: "#/components/responses/InternalServerError"
 
+  /ad-hoc-activities/search:
+    post:
+      tags:
+        - Ad-hoc subprocess
```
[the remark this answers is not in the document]
maff: Thanks for the input! I'd agree that this is more descriptive and that it also would be more consistent with how it is actually implemented (e.g. `AdHocSubprocessActivityController`).

@saig0 as you initially designed the API spec, any objections from your side to changing the endpoint as suggested?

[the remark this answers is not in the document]
saig0: `/ad-hoc-subprocesses/activities/search` would be possible too. However, I don't see any other use for the endpoint `ad-hoc-subprocesses/`. 

The endpoint is about activities with an ad-hoc subprocess that can be activated using the activate API.

aleksander-dytko:
Hi @maff, 

I have a few suggestions from the API guidance standpoint:
* please use `element` instead of `flowNode` - we're currently aligning that for 8.8 and we should keep unified naming 
* I propose to have this endpoint in `/element-instances/ad-hoc-activities/search` path - we already have other endpoints for elements

For technical considerations, please involve @tmetzke if needed.

clients/java/src/test/java/io/camunda/client/adhocsubprocess/AdHocSubprocessActivitySearchTest.java
```diff
@@ -0,0 +1,173 @@
...
+            + "      \"processDefinitionKey\": \"2251799813685281\",\n"
+            + "      \"processDefinitionId\": \"TestParentAdHocSubprocess\",\n"
+            + "      \"adHocSubprocessId\": \"TestAdHocSubprocess\",\n"
+            + "      \"flowNodeId\": \"unknownTask\",\n"
+            + "      \"flowNodeName\": \"Unknown Task\",\n"
+            + "      \"type\": \"START_EVENT\",\n"
+            + "      \"tenantId\": \"<default>\"\n"
+            + "    }\n"
+            + "  ]\n"
+            + "}";
```
tmetzke, line 132: 🙃 You can consider using text blocks to make this a bit nicer, see [this test](https://github.com/camunda/camunda/blob/main/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/TopologyControllerTest.java#L43) for an example.

zeebe/gateway-protocol/src/main/proto/rest-api.yaml
tmetzke: Yep, all keys are supposed to be string values 👍 The other controllers show to handle it, I believe.

tmetzke:
> Hi @maff,
> 
> I have a few suggestions from the API guidance standpoint:
> 
> * please use `element` instead of `flowNode` - we're currently aligning that for 8.8 and we should keep unified naming
> * I propose to have this endpoint in `/element-instances/ad-hoc-activities/search` path - we already have other endpoints for elements
> 
> For technical considerations, please involve @tmetzke if needed.

This is quite an interesting use case, really.

I believe, but please correct me if I'm wrong, this is not really suited for being under `element-instances` as it is rather a sub-resource of the process definition, isn't it? This is static information from the BPMN XML, not dynamic instances of activities.

To help define naming, I need more context on how we want to use this API. For example:
* Do we always have a fixed path of process definition key and sub-process ID for which we need to find those ad-hoc activities? 
* Or do we need to support searching for something like "All ad-hoc activities of process definitions whose IDs are named like 'foo*'"?
* If the path is fixed by the two attributes anyway, a completely different idea would be making this a process definition sub-resource, accessed directly, for example: `GET /v2/process-definitions/38572950284287/sub-processes/my-subprocess-id/ad-hoc-activities`. This returns the list of ad-hoc activities and could theoretically be nicely cached even before going to the REST API gateway if needed (I think we don't support this yet, just for showing possible directions).

clients/java/src/test/java/io/camunda/client/adhocsubprocess/AdHocSubprocessActivitySearchTest.java
maff: I did so on the other modules (e.g. the REST controller tests), but unfortunately this module is still Java 8.

maff:
@aleksander-dytko @tmetzke thank you for your comments! I'll align the `flowNode -> element` naming accordingly.

Regarding the use case, you're right, at the moment this is only returning static information based on the process definition data, however this is only the first step towards an API which is able to search for "activatable" activities within existing ad-hoc subprocesses. The actual final solution as envisioned in the [design document](https://docs.google.com/document/d/1EK90iIbqCpP-USxw2thzQPxXdw4ngvw4JXym6hR-YO0/edit?tab=t.0#heading=h.be9ipaujnu2z) would be implemented in #28432 to achieve the following:

- Search all active ad-hoc subprocesses matching the given search criteria (making this an actual search without requiring the 2 fixed properties as it does now) - this needs to come from process instance data and could most likely build on the existing flow node instance search.
- For each of these found subprocesses, resolve them to a list of activities which can be activated (kind of a tasklist for ad-hoc activities) - an easy approach could be to merge this with the data we currently resolve from the definition (which I agree could be cached for each process definition key), so for each found ad-hoc subprocess matching the criteria, the end result would include its activatable activities.

So basically this one is only a step towards the desired goal as defined in the design document. If you prefer not to merge this at this point I'd be ok with pulling #28432 into this issue and not merging this in the current state.

Maybe @saig0 you could shed some additional light on the actual use case to find the best option for the REST API path?

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
maff: Thanks, I found that one while looking at the flow node instance API, but could not use it in the current state mainly because it currently only caches flow node names. I would have planned to first finish #28432 and then look into how this could be properly cached.

tmetzke: That makes sense, I just wanted to make sure you're aware 👍

clients/java/src/test/java/io/camunda/client/adhocsubprocess/AdHocSubprocessActivitySearchTest.java
tmetzke: Ugh, yeah, the client, I forgot 😄 Nevermind, then 😅

tmetzke:
> So basically this one is only a step towards the desired goal as defined in the design document. If you prefer not to merge this at this point I'd be ok with pulling https://github.com/camunda/camunda/issues/28432 into this issue and not merging this in the current state.

This is absolutely fine. The constraints are clearly documented here and in the follow-up, we should go iteratively here 👍 

> * Search all active ad-hoc subprocesses matching the given search criteria (making this an actual search without requiring the 2 fixed properties as it does now) - this needs to come from process instance data and could most likely build on the existing flow node instance search.
> * For each of these found subprocesses, resolve them to a list of activities which can be activated (kind of a tasklist for ad-hoc activities) - an easy approach could be to merge this with the data we currently resolve from the definition (which I agree could be cached for each process definition key), so for each found ad-hoc subprocess matching the criteria, the end result would include its activatable activities.

OK, that is super helpful, thank you, @maff!
So, we want to support finding ad-hoc activities in active element instances (of a specific type, which is ad-hoc subprocesses here). Those ad-hoc activities are a new kind of resource/search result, we don't have this yet, I would say.
Additionally, we want to support activating ad-hoc activities at one specific element instance (they are denoted in the ad-hoc activities search result by an `elementInstanceKey`, the internal key of the ad-hoc subprocess instance they belong to).

With this in mind, I could think of the following API:
* `POST /v2/element-instances/ad-hoc-activities/search` to find ad-hoc activities in ad-hoc subprocess instances.
* `POST /v2/element-instances/:elementInstanceKey/ad-hoc-activation` to activate ad-hoc activities in one specific ad-hoc subprocess instance.

The endpoints above are generic regarding the element instance. They don't restrict ad-hoc activities to being attached to ad-hoc subprocess instances. They could be related to ANY element, at least from the API's perspective.

Not sure if this makes sense overall but this looks most consistent to me regarding naming and reuse of existing resources while staying semantically correct. The placement under element instances might not be intuitive/discoverable enough, I am not entirely sure about that, to be honest. The question here ultimately is: should ad-hoc subprocesses become a first-class citizen in the API? It would more or less replace the element instance-named suggestions above.

Another naming path is going for ad-hoc activities directly, e.g.:
* `POST /v2/ad-hoc-activities/search` to find ad-hoc activities with whatever search criteria.
* `POST /v2/ad-hoc-activities/activation` to activate ad-hoc activities based on defined criteria, e.g. by defining one but maybe also multiple element instance keys, i.e., activating ad-hoc activities at multiple ad-hoc subprocess instances at once.

It depends a bit on what we consider the first-class citizen here, what we want to achieve in the end, and how we think about discoverability in the API. I believe all those paths are totally valid, we should just settle for a consistent path and execute accordingly.

I hope this helps more than it creates too many new options 😅 
As with most cases regarding API: there is no one correct way 🤷

berkaycanbc (changes requested):
🚀 Thanks for the PR @maff!

⭐ Very good job with separate commits and comments on implementation details. Made it a lot easier to review.

🔧 I requested bunch of minor changes that needs to be addressed before merging the PR. Please note that I use [emoji codes](https://github.com/camunda/camunda/blob/bc8fbdc0b0c6338e45ab18f45c75c5a11ab60b1a/CONTRIBUTING.md#review-emoji-code) while reviewing.

clients/java/src/main/java/io/camunda/client/CamundaClient.java
```diff
@@ -851,6 +852,54 @@ static CamundaClientCloudBuilderStep1 newCloudClientBuilder() {
...
   FlowNodeInstanceGetRequest newFlowNodeInstanceGetRequest(long flowNodeInstanceKey);
 
+  /**
+   * Executes a search request to query activities within ad-hoc subprocesses.
+   *
+   * <p>Note that this API currently requires filters for both process definition key and ad-hoc
+   * subprocess ID and does not support paging or sorting.
+   *
+   * <pre>
+   * long processDefinitionId = ...;
```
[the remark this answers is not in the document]
maff: Good catch! 🤦‍♂️

service/src/main/java/io/camunda/service/AdHocSubprocessActivityServices.java
```diff
@@ -0,0 +1,110 @@
...
+                      toAdHocSubprocessActivityEntity(
+                          processDefinitionEntity, adHocSubProcess, flowNode))
+              .toList();
+
+      return new SearchQueryResult.Builder<AdHocSubprocessActivityEntity>()
+          .items(rootActivities)
+          .build();
+    }
+
+    throw new NotFoundException(
```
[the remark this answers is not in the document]
maff: Fully agree - I initially had implemented this way to make use of the pattern matching on `adHocSubProcess`, but I wasn't aware this is also available outside the if block. I'll update it accordingly.

berkaycanbc (approved):
LGTM 👍

maff:
@aleksander-dytko @tmetzke thanks for the valuable discussion!

Regarding the `/element-instances` segment in the path: we had some internal discussions around the scope of this API and agreed to deprioritize #28432 for now as it might not be needed.

Even if we would rely on process instance data as initially planned, the results returned by this API would never refer to actual element instances (there is no flow node key in the response) as at the time of the activation of the subprocess the child activity instances might not exist yet. For example in the following example the ad-hoc subprocess was activated, but its activities weren't:

image

Given that, would you still suggest including `/element-instances` in the path? Otherwise I'd suggest keeping it a dedicated root path as it is currently implemented to indicate that this is kind of a special case which will be paired with the activation API implemented in https://github.com/camunda/camunda/issues/28219.

tmetzke:
> @aleksander-dytko @tmetzke thanks for the valuable discussion!
> 
> Regarding the `/element-instances` segment in the path: we had some internal discussions around the scope of this API and agreed to deprioritize #28432 for now as it might not be needed.
> 
> Even if we would rely on process instance data as initially planned, the results returned by this API would never refer to actual element instances (there is no flow node key in the response) as at the time of the activation of the subprocess the child activity instances might not exist yet. For example in the following example the ad-hoc subprocess was activated, but its activities weren't:
> 
> image
> 
> Given that, would you still suggest including `/element-instances` in the path? Otherwise I'd suggest keeping it a dedicated root path as it is currently implemented to indicate that this is kind of a special case which will be paired with the activation API implemented in #28219.

Yes, I still would. The access path explanation for `POST /v2/element-instances/ad-hoc-activities/search` is the following:

* Create a `search` for `ad-hoc-activities` as a sub-resource of `element-instances`. All result items are ad-hoc activities that are somehow connected to existing element instances (here: ad-hoc subprocess instances). The return type is still a list of ad-hoc activities, not element instances.
* This is similar to `GET /v2/user-tasks/:userTaskKey/variables/search`, where you search for variables connected to a user task. The return type here is a list of variables, not user tasks.
  * The difference with our current path is that we don't restrict the search to one specific ad-hoc subprocess instance. Whether we want to do this depends on the use case/access pattern that makes the most sense here.
    * Do I only query for ad-hoc activities for one specific ad-hoc subprocess instance anyway?
    * Or do I regularly query for ad-hoc activities across ad-hoc subprocess instances, e.g., all ad-hoc activities of one process definition across all the current instances?

I cannot answer those questions, I believe. It depends on what we want to use the endpoint for.
It all comes down to where and how we want to use those endpoints.

maff:
> I cannot answer those questions, I believe. It depends on what we want to use the endpoint for.
It all comes down to where and how we want to use those endpoints.

Thanks again for the input. As this is a technical enabler for still unclear use cases that's hard to judge 😅 I updated the endpoint to reflect the suggested changes - happy to revisit this if we have more knowledge.

maff:
> Where do we plan to use the endpoint for now? Any web app or anything?

The first consumer will be an execution listener on the subprocess and/or its activities which is able to decide which activities to activate based on this search, however in the future this might be superseded by passing this information to the EL directly via https://github.com/camunda/camunda/issues/25192. 

I marked the API as alpha on both the OpenAPI spec and the client now and we'll make sure to revisit this as we proceed (also happy to remove the API if it turns out it won't be needed in this current state).

tmetzke:
> > Where do we plan to use the endpoint for now? Any web app or anything?
> 
> The first consumer will be an execution listener on the subprocess and/or its activities which is able to decide which activities to activate based on this search, however in the future this might be superseded by passing this information to the EL directly via #25192.
> 
> I marked the API as alpha on both the OpenAPI spec and the client now and we'll make sure to revisit this as we proceed (also happy to remove the API if it turns out it won't be needed in this current state).

OK, interesting case 👍 Sounds reasonable to me, we need to make progress and explore how we want to use it. If you have any questions or would like feedback on this topic/API later again, just ping me or reach out in the REST API Slack channel. Happy to help 🙂

PR #28279, opened 2025-02-18
refactor: remove C7 engine config from Optimize C8

abremard:
## Description

This is part of the "C7 config cleanup from C8". Because there's a lot of configs that need cleaning, I'm breaking this work down into multiple PRs. This one is focused on "engine config" cleanup. Most of the lines removed are setters/getters and boilerplates. For these updates I don't see any references to the `engine config` in [C8 docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/)

- it's not being used/referenced in `camunda/camunda`
- it's not expected in `camunda-cloud` repos
- it's not expected in `camunda/camunda-platform-helm`

## Related issues


optimize-distro/src/config/environment-config.yaml
```diff
@@ -36,43 +36,8 @@ engines:
   # An alias of the engine, which will be used for internal purposes like
   # logging and displaying which data belong to which engine.
   'camunda-bpm':
```
RomanJRW, line 38: can't this also be deleted?

optimize-distro/src/config/environment-config.yaml
```diff
@@ -36,43 +36,8 @@ engines:
...
-    # For details consult the Multi-Tenancy documentation.
-    defaultTenant:
-      # the id used for this default tenant on persisted entities
-      id: null
-      # the name used for this tenant when displayed in the UI
-      name: null
-    # A base URL that will be used for connections to the Camunda Engine REST API.
-    rest: 'http://localhost:8080/engine-rest'
     # Determines whether this instance of Optimize should import definition & historical data from this engine.
     importEnabled: true
```
RomanJRW, line 40: why does this stay?

optimize/util/optimize-commons/src/test/resources/environment-variable-default-value-test-config.yaml
```diff
@@ -11,27 +11,9 @@ api:
 
 engines:
   'camunda-bpm-1':
```
RomanJRW, line 13: I feel like this whole block _should_ be removable, but I would like to learn why if this is not the case

RomanJRW (changes requested):
Generally no issue with the changes, just a comment as I suspect we can even delete more than is here

optimize-distro/src/config/environment-config.yaml
abremard: The `importEnabled` config is still exposed in [ConfigurationService::isImportEnabled](https://github.com/camunda/camunda/blob/fb7b200a6f9c4f755bdeef04d586929cefb0b9be/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java#L874) which is used by [ImportSchedulerManagerService::startSchedulers](https://github.com/camunda/camunda/blob/dcc5a37552e13acc6474ed24da8274bd8e5d1a62/optimize/backend/src/main/java/io/camunda/optimize/service/importing/ImportSchedulerManagerService.java#L88). I don't have the full context around that scheduler but if it's something we can remove, I'm happy to do it as part of this PR

abremard: That works, thanks for the input here!

RomanJRW: Happy to help. Although you did all the hard work - thanks! 😆

optimize-distro/src/config/environment-config.yaml
```diff
@@ -30,50 +30,6 @@ container:
   # If no value specified the container host and port are used instead.
   accessUrl: null
 
-# Configuration for engines used to import data. Please note that you have to have at
-# least one engine configured at all times.
-engines:
```
RomanJRW, line 35: much nicer that we can now drop the whole block 💪

RomanJRW (approved):

PR #28281, opened 2025-02-18
feat: new property mapping mechanism, just like in 8.8

jonathanlukas:
## Description

This PR introduces the property mapping pattern from 8.8 to 8.7.

#### OAuthCredentialsProvider

* got a new flag to make environment variable overrides optional
* this is useful as the plain env variables should not pollute the spring properties configuration

#### New CredentialsProviderConfiguration

* instantiates the credentials provider as a separate bean
* useful to extend later
* more spring-style configuration of the zeebe client

#### Removal of PropertyUtil

* actual goal of this PR
* was an intermediate solution for mapping "old" and "new" properties
* con: very vague and generic, error-prone

=> there will be a lot of replacements of the usage, the default replacement if `ofNullable(property).orElse(defaultProperty)`

#### Make ZeebeClientConfigurationProperties usused

* all "old" properties are now mapped to new ones
* classes are kept in place to generate the properties mapping json

#### Change in properties structure - introduction of `camunda.client.cloud`

* will be part of 8.8 as well
* separation of cloud-related properties like cluster-id and region from the rest of the client properties
* old properties structure will still be mapped

#### PropertiesPostProcessor

* now uses the CamundaClientLegacyPropertiesMappingsLoader to load mappings from classpath
* they are used to "find" old properties and map them to "new" ones as new ones are absent -> will never override new ones
* additional mappings of all worker-specific property overrides
* detection of implicit saas configuration (by searching for a cluster id)

#### Removal of PropertiesUtil

* unused static method removed

#### AuthProperties

* introduce new property tokenUrl and make issuer deprecated
* added configuration for connectTimeout and readTimeout as this can be configured to the credentialsprovider as well

#### ZeebeClientProperties

* introduced `DataSize` as format for message size and metadata size -> is non-breaking as numbers will still be interpreted as bytes

#### Moving of "mode" presets

* they could be used as spring profiles which is NOT intended

#### Legacy property mappings

* all mapping rules are kept in place here
* all previous mappings are in place again
* in addition, some new mappings have been found

#### Test classes

* all tests need to be @SpringBootTests now as the properties post processor is only applied then
* the credentials provider configuration needs to be included to bootstrap the zeebe client
* tests for unused functions have been removed
* tests that have duplicate assertions in other places have been removed

#### ZeebeClientPropertiesPostProcessorTest

* verifies compatibility with older (documented) properties -> used the docs for this

## Checklist

- [ ] for CI changes:

## Related issues

closes #


ana-vinogradova-camunda:
❓ Is there a summary of the changes introduced in this PR somewhere? That would help, thank you 👍

ana-vinogradova-camunda:
❓ My assumption is that the main goal of this PR is to replace backward compatibility in code with a new mapping of properties. This looks good from my best knowledge 👍 , please let me know if I am missing something.

jonathanlukas:
@ana-vinogradova-camunda I added a detailed documentation of the introduced changes for your review.

ana-vinogradova-camunda:
@jonathanlukas thank you for the detailed explanation here, that helped a lot 🥇 👍 
💭 is there a need to update documentation as well?

ana-vinogradova-camunda (approved):

PR #28283, opened 2025-02-18
refactor: remove C7 engine-commons config from Optimize C8

abremard:
## Description

This is part of the "C7 config cleanup from C8". Because there's a lot of configs that need cleaning, I'm breaking this work down into multiple PRs. This one is focused on "engine-commons config" cleanup. I'm not seeing any getter/setter calls so this config should be relatively safe to remove. Most of the lines removed are setters/getters and boilerplates. For these updates I don't see any references to the engine-commons config in [C8 docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/)

## Related issues

### DIFF
--- a/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java
+++ b/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java
@@ -367,31 +367,6 @@ public class ConfigurationService {
     this.maximumBackoff = maximumBackoff;
   }
 
-  public int getEngineConnectTimeout() {
-    if (engineConnectTimeout == null) {
-      engineConnectTimeout =
-          configJsonContext.read(
-              ConfigurationServiceConstants.ENGINE_CONNECT_TIMEOUT, Integer.class);
-    }
-    return engineConnectTimeout;
-  }
-
-  public void setEngineConnectTimeout(final Integer engineConnectTimeout) {
-    this.engineConnectTimeout = engineConnectTimeout;
-  }
-
-  public int getEngineReadTimeout() {
-    if (engineReadTimeout == null) {
-      engineReadTimeout =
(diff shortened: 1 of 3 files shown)


optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java
```diff
@@ -367,27 +367,10 @@ public void setMaximumBackoff(final Long maximumBackoff) {
...
-      engineConnectTimeout =
-          configJsonContext.read(
-              ConfigurationServiceConstants.ENGINE_CONNECT_TIMEOUT, Integer.class);
-    }
-    return engineConnectTimeout;
-  }
-
   public void setEngineConnectTimeout(final Integer engineConnectTimeout) {
     this.engineConnectTimeout = engineConnectTimeout;
   }
```
RomanJRW, line 372: can the setter also be removed then? And the actual variables too

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java
```diff
@@ -367,27 +367,10 @@ public void setMaximumBackoff(final Long maximumBackoff) {
...
 
-  public int getEngineReadTimeout() {
-    if (engineReadTimeout == null) {
-      engineReadTimeout =
-          configJsonContext.read(ConfigurationServiceConstants.ENGINE_READ_TIMEOUT, Integer.class);
-    }
-    return engineReadTimeout;
-  }
-
   public void setEngineReadTimeout(final Integer engineReadTimeout) {
```
RomanJRW, line 374: same here

RomanJRW (approved):
Thanks for breaking all of this work into such small chunks, it makes reviewing super easy 👍 

Pre-approving with some minor suggestions

PR #28296, opened 2025-02-18
ci: update benchmark workflows

lenaschoenburg:
For benchmarks we now always build the full camunda image, including frontends. That allows us to remove the special handling of Operate.

Additionally, there are some minor adjustments to align with the latest helm chart values.

### DIFF
--- a/.github/workflows/zeebe-benchmark.yml
+++ b/.github/workflows/zeebe-benchmark.yml
@@ -119,10 +119,22 @@ jobs:
           vault-address: ${{ secrets.VAULT_ADDR }}
           vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
           vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: ./.github/actions/build-frontend
+        id: build-operate-fe
+        with:
+          directory: ./operate/client
+      - uses: ./.github/actions/build-frontend
+        id: build-tasklist-fe
+        with:
+          directory: ./tasklist/client
+      - uses: ./.github/actions/build-frontend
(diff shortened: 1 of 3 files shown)


lenaschoenburg:
Here's a workflow run: https://github.com/camunda/camunda/actions/runs/13390728637

.github/workflows/zeebe-benchmark.yml
```diff
@@ -119,10 +109,22 @@ jobs:
           vault-address: ${{ secrets.VAULT_ADDR }}
           vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
           vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
+      - uses: ./.github/actions/build-frontend
```
npepinpe, line 112: :thought_balloon: Surprised there's not a global `build-frontend` action :shrug:

zeebe/benchmarks/setup/default/values-stable.yaml
```diff
@@ -1,5 +1,17 @@
 # Additional values file to run Zeebe on stable VMs
 camunda-platform:
+  core:
```
npepinpe, line 3: :exclamation: Ups, totally missed we have to update this :see_no_evil:

npepinpe (approved):

.github/workflows/zeebe-benchmark.yml
ChrisKujawa: That would be too easy.

ChrisKujawa (approved):
Thanks 👍🏼

PR #28297, opened 2025-02-18
feat: fix and enhance execution order in rdbms statement queue #28101

stoerti:
## Description

While working on #28101 a problem with the execution order of rdbms writer statements was discovered. It could lead to a situation where a record is updated before the insert was executed. The new implementation now executes INSERT statements first.
For a few entities (authorization, userTask) it is configurable because there an UPDATE is executed as DELETE + INSERT. There the order is left untouched.

## Related issues

relates to #28101

### DIFF
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/DefaultExecutionQueue.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/DefaultExecutionQueue.java
@@ -14,6 +14,8 @@ import java.util.Comparator;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 import org.apache.ibatis.executor.BatchResult;
 import org.apache.ibatis.session.ExecutorType;
 import org.apache.ibatis.session.SqlSessionFactory;
@@ -136,11 +138,10 @@ public class DefaultExecutionQueue implements ExecutionQueue {
         sessionFactory.openSession(ExecutorType.BATCH, TransactionIsolationLevel.READ_UNCOMMITTED);
 
     var flushedElements = 0;
-    final var items = new ArrayList<>(queue);
-    items.sort(Comparator.comparing(QueueItem::contextType).thenComparing(QueueItem::statementId));
+    final var optimizedItems = optimizeQueueOrder(queue);
 
     try {
-      for (final var entry : items) {
+      for (final var entry : optimizedItems) {
         LOG.trace("[RDBMS ExecutionQueue, Partition {}] Executing entry: {}", partitionId, entry);
         session.update(entry.statementId(), entry.parameter());
         queue.remove();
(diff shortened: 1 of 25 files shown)


db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/DefaultExecutionQueue.java
```diff
@@ -187,6 +188,29 @@ private int doFLush() {
...
+      if (!itemsByContextType.containsKey(contextType)) {
+        continue;
+      }
+
+      if (contextType.preserveOrder()) {
+        resultList.addAll(itemsByContextType.get(contextType));
+      } else {
+        final var contextItems = new ArrayList<>(itemsByContextType.get(contextType));
+        contextItems.sort(
+            Comparator.comparing(QueueItem::statementType).thenComparing(QueueItem::statementId));
```
p-wunderlich, line 206: How is the order defined? By the order of the rows in the StatementType enum? Isn't it normally the name?

db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/StatementType.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.db.rdbms.write.queue;
+
+public enum StatementType {
```
p-wunderlich, line 10: Hm should we either add also Select as type, or rename it to WriteStatementType to be clear?

db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/DefaultExecutionQueue.java
stoerti: By default Enum classes are sorted by natural order (order of values in the Enum class).

db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/StatementType.java
stoerti: Renamed it.

p-wunderlich (approved):

db/rdbms/src/main/java/io/camunda/db/rdbms/write/queue/DefaultExecutionQueue.java
```diff
@@ -187,6 +188,46 @@ private int doFLush() {
...
+   *
+   * @param items queue of items
+   * @return optimized queue of items
+   */
+  private List<QueueItem> optimizeQueueOrder(final List<QueueItem> items) {
+    final Map<ContextType, List<QueueItem>> itemsByContextType =
+        items.stream().collect(Collectors.groupingBy(QueueItem::contextType));
+
+    final List<QueueItem> resultList = new ArrayList<>();
+    for (final var contextType : ContextType.values()) {
```
romansmirnov, line 213: ❓ Would be `itemsByContextType.entries()` (instead `ContextType.values()`) sufficient? If yes, then the next if-statement could be deleted.

stoerti: Oh yes... seems to be a refactoring leftover. I replaced it 👍

db/rdbms/src/test/java/io/camunda/db/rdbms/write/queue/DefaultExecutionQueueTest.java
```diff
@@ -200,4 +269,57 @@ public QueueItem merge(final QueueItem originalItem) {
...
+
+    // then
+    verify(session).update(eq("statement5"), any());
+    verify(session).update(eq("statement2"), any());
+    verify(session).update(eq("statement6"), any());
+    verify(session).update(eq("statement7"), any());
+    verify(session).update(eq("statement8"), any());
+    verify(session).update(eq("statement4"), any());
+    verify(session).update(eq("statement1"), any());
+    verify(session).update(eq("statement3"), any());
```
danielkelemen, line 323: 🔧 This doesn't actually check the order. I'd change it to an arg captor and verify the actual order.

PR #28299, opened 2025-02-18
fix: renew an OIDC access token when the refresh token is not a JWT

### LINKED ISSUE #27710: Operate cannot renew an OIDC access token when the refresh token is not a JWT

When I configure Operate with an OIDC provider that doesn't use JWT for its refresh tokens (e.g. with MS Entry ID), then Operate is not successful with refreshing access tokens.

In consequence, the logged in user is logged out and must authenticate again.

In addition, this is logged on error level. It's not error, as this is expected behavior in such a setup:

```
io.camunda.operate.webapp.security.identity.IdentityAuthentication - Unable to decode refresh token XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX-1 with exception: error decoding jwt 
```

Backport to 8.7 and 8.6

### Steps to reproduce (required on creation)

1. Configure Operate and Identity to use MS Entra ID
1. Sign into Operate
1. wait for the access token to expire.

### Current behavior (required on creation)

* Instead of getting a new access token, the user is logged out.
* The error message above is logged

### Expected behavior (required on creation)

* The user remains logged in
* The error is not logged (at most on debug level)

### Environment (required on creation)

- MS Entra ID and other OIDC implementations that use non-JWT refresh tokens
- Operate configured with Identity

### Root cause (required on prioritization)

* With this PR we introduced the ability to use OIDC probiders with non-JWT refresh tokens: https://github.com/camunda/operate/pull/5708. The goal of that was to support MS Azure AD. See also this code review discussion on this exact matter: https://github.com/camunda/operate/pull/5708#pullrequestreview-1712788084
* With this PR we have refactored the code and broke the flow that determines if the token should be used: https://github.com/camunda/camunda/pull/17107. The code to parse the token was moved, so that now we ignore a decoding exception and set the refresh date to `null` ([ref](https://github.com/camunda/camunda/blob/8.6.7/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java#L249-L257)). If null, we treat the refresh token as expired ([ref](https://github.com/camunda/camunda/blob/8.6.7/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java#L80)). And accordingly don't attempt to use it to get a new access token ([ref](https://github.com/camunda/camunda/blob/8.6.7/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java#L98)).
* In the addition, in the current code you can still see a catch clause that has no chance to ever trigger: [https://github.com/camunda/camunda/blob/8.6.7/operate/webapp/src/main/java/io/camu[…]da/operate/webapp/security/identity/IdentityAuthentication.java](https://github.com/camunda/camunda/blob/8.6.7/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java#L77-L87)

### Solution ideas

* Option 1: Fix bug and keep existing JWT parsing logic (if it applies)
  * If the token cannot be parsed, use it for refreshing in any case
  * Remove the non-functional catch clause: https://github.com/camunda/camunda/blob/8.6.7/operate/webapp/src/main/java/io/camu[…]da/operate/webapp/security/identity/IdentityAuthentication.java](https://github.com/camunda/camunda/blob/8.6.7/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java#L77-L87
  * Change the log level for JWT parsing errors from error to debug
* Option 2: Remove the parsing of the refresh token entirely
  * Always use the refresh token, treating it as an opaque value that it is according to the [OAuth spec](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5); if refreshing is not possible, only then let the user log in again
  * This simplifies the implementation and aligns it with [how Identity handles this case](https://camunda.slack.com/archives/C08A4323G83/p1737711793171749)
  * Was introduced with https://github.com/camunda/operate/pull/2415
* In any case, write a test case for this situation, e.g. using Wiremock

### Handover Dev to QA (required before manual testing)

- Resources:
  
- Versions to validate:
  
- Release version (in which version this feature will be released)
  

### Link to the test case
[User remains logged when configure operate with an OIDC](https://camunda.testrail.com/index.php?/cases/view/589535)

### Links

- https://jira.camunda.com/browse/SUPPORT-25491

```[tasklist]
### Pull Requests
- [ ] main: https://github.com/camunda/camunda/pull/28299
- [ ] 8.7: https://github.com/camunda/camunda/pull/28508
- [ ] 8.6: https://github.com/camunda/camunda/pull/28579
```

kristinkomschow:
## Description

Removed parsing of refresh tokens because this leads to errors when they are not JWTs and goes against the OIDC specification of treating them as opaque values. The error was that the parsing would throw an exception, because of which `refreshTokenExpiresAt` was set to null. The refresh token was therefore never used. Instead, the user has to log in again as soon as the access token expired.

Added test case to verify authentication works with non-JWT refresh tokens.

## Checklist

- [ ] for CI changes:

## Related issues

related to https://github.com/camunda/camunda/issues/27710

### DIFF
--- a/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java
+++ b/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java
@@ -28,8 +28,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.InsufficientAuthenticationException;
-import org.springframework.web.context.request.RequestContextHolder;
-import org.springframework.web.context.request.ServletRequestAttributes;
 
 public class IdentityAuthentication extends AbstractAuthenticationToken
     implements Serializable, TenantAwareAuthentication {
@@ -215,17 +213,6 @@ public class IdentityAuthentication extends AbstractAuthenticationToken
     expires = accessToken.getToken().getExpiresAt();
     LOGGER.info("Access token will expire at {}", expires);
     setAuthenticated(!hasExpired());
-  }
-
-  private boolean isPolling() {
-    final ServletRequestAttributes requestAttributes =
-        (ServletRequestAttributes) RequestContextHolder.getRequestAttributes();
-    if (requestAttributes != null) {
-      return Boolean.TRUE.equals(
-          Boolean.parseBoolean(requestAttributes.getRequest().getHeader(POLLING_HEADER)));
-    } else {
-      return false;
-    }
   }
 
   private void retrieveName(final UserDetails userDetails) {
--- a/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java
+++ b/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java
@@ -35,7 +35,6 @@ public class IdentityAuthentication extends AbstractAuthenticationToken
   @Serial private static final long serialVersionUID = 1L;
 
   private static final Logger LOGGER = LoggerFactory.getLogger(IdentityAuthentication.class);
-  private static final String POLLING_HEADER = "x-is-polling";
   private Tokens tokens;
   private String id;
   private String name;
@@ -43,8 +42,6 @@ public class IdentityAuthentication extends AbstractAuthenticationToken
   @JsonIgnore private List<IdentityAuthorization> authorizations;
   private String subject;
   private Date expires;
-
-  private Date refreshTokenExpiresAt;
 
   private List<OperateTenant> tenants;
 
--- a/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java
+++ b/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java
@@ -12,7 +12,6 @@ import io.camunda.identity.sdk.Identity;
 import io.camunda.identity.sdk.authentication.AccessToken;
 import io.camunda.identity.sdk.authentication.Tokens;
 import io.camunda.identity.sdk.authentication.UserDetails;
-import io.camunda.identity.sdk.authentication.exception.TokenDecodeException;
 import io.camunda.identity.sdk.impl.rest.exception.RestException;
 import io.camunda.operate.util.SpringContextHolder;
 import io.camunda.operate.webapp.security.Permission;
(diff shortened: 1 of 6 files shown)


kristinkomschow:
There is one issue that this will introduce (see https://github.com/camunda/camunda/issues/27710#issuecomment-2665923614):
> The downside of not checking refresh tokens is that the renew token request is still retried multiple times even when the error source is clear because of this code:
> 
> [camunda/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java](https://github.com/camunda/camunda/blob/63f7a76ae9ab1226466160e43ab24a1bca7b1c5c/operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java#L283-L288)
> 
> Lines 283 to 288 in [63f7a76](/camunda/camunda/commit/63f7a76ae9ab1226466160e43ab24a1bca7b1c5c)
> 
>  private Tokens renewTokens(final String refreshToken) throws Exception { 
>    return getIdentityRetryService() 
>        .requestWithRetry( 
>            () -> getIdentity().authentication().renewToken(refreshToken), 
>            "IdentityAuthentication#renewTokens"); 
>  } 
> I could not find any justification for why this request is done with retries in the git history, so I'm unsure if we can just remove the retries on that request or if we should add better error handling.

I'd appreciate opinions on what I should do with this: remove retries, create new issue or adjust error handling (this may not be feasible because of time pressure)

operate/webapp/src/main/java/io/camunda/operate/webapp/security/identity/IdentityAuthentication.java
```diff
@@ -244,17 +214,6 @@ public void authenticate(final Tokens tokens) {
     subject = accessToken.getToken().getSubject();
     expires = accessToken.getToken().getExpiresAt();
     LOGGER.info("Access token will expire at {}", expires);
-    if (!isPolling()) {
```
sdorokhova, line 247: It look like `isPolling` method can also be removed, right?

sdorokhova (approved):
LGTM! 
Just a tiny note in in `IdenityAuthentication` class.

kristinkomschow:
I removed the unused method.

I also added another change, after seeing this: https://learn.microsoft.com/en-us/entra/identity-platform/refresh-tokens#token-lifetime
According to the Microsoft Entra documentation, refresh tokens should be revoked once a new one has been fetched because this does not happen automatically. I added revoking the token in the refresh method.

sdorokhova (approved):

kristinkomschow:
I added another change: retries for requests to Identity will only happen for `RestException` instead of all `IdentityException` classes now. This prevents retries to be executed when the issue is an expired/faulty token rather than a connection issue.
See discussion here: https://camunda.slack.com/archives/CRQAL5A1M/p1740053857648029?thread_ts=1739778227.737889&cid=CRQAL5A1M

Retries are only used for the token refresh and for getting authentication. I think it makes sense in both cases to only attempt a retry if the error isn't "just" a connection error.

sdorokhova (approved):
That's a nice catch with `IdentityException` vs `RestException`!

kristinkomschow:
Removed `refreshTokenExpiresAt ` field after Svetlana noted it and deemed it safe to delete despite session deserializatio that Spring does in `SessionService `:  "When persistent sessions are enabled, Spring will store sessions in Elasticsearch and read them whenever needed, this allows us to have several webapp nodes in SaaS. The user may be logged in on one node, but not on the other, but as sessions are persisted they can be reused. I'm actually sure now that it's safe to delete the field. I checked the code and if session can't be serialized Spring will simply consider that there is no session yet and will create the new one."

PR #28310, opened 2025-02-18
test: verify that passive director closes all exporters

lenaschoenburg:
Adds a regression test for https://github.com/camunda/camunda/pull/28243

### DIFF
--- a/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirectorTest.java
+++ b/zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirectorTest.java
@@ -310,6 +310,20 @@ public final class ExporterDirectorTest {
   }
 
   @Test
+  public void shouldIgnoreErrorsOnClose() throws Exception {
+    // given
+    startExporterDirector(exporterDescriptors);
+
+    // when -- closing the first exporter will throw an exception
+    doThrow(new RuntimeException()).when(exporters.get(0)).close();
+    rule.closeExporterDirector();
+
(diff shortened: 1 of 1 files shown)


zeebe/broker/src/test/java/io/camunda/zeebe/broker/exporter/stream/ExporterDirectorTest.java
```diff
@@ -320,6 +322,19 @@ public void shouldCloseAllExportersOnClose() throws Exception {
     verify(exporters.get(1), TIMEOUT).close();
   }
 
+  @Test
+  public void shouldCloseAllExportersOnCloseInPassiveMode() throws Exception {
+    // given
+    passiveExporterRule.startExporterDirector(exporterDescriptors);
```
entangled90, line 328: :wrench: it may be worth also adding a "mixed" test case where we have:
- 1 exporter active
- 1 exporter passive
- the first exporter registered throws an exception when `close()` is called
and then we check that every `close()` method has been called

lenaschoenburg: The "mixed" scenario shouldn't exist, it's the entire `ExporterDirector` which is either in Active or in Passive mode.

Good call on the error handling though, it looks like we don't have any tests for that :+1:

lenaschoenburg: I've added  0da020a for this. It's not specific to the Active/Passive mode so I've left that part out.

entangled90 (approved):

PR #28320, opened 2025-02-18
fix: Add missing flow-node type for ad-hoc subprocesses

### LINKED ISSUE #28304: Add new BPMN element type for ad-hoc subprocess
## Description

With Camunda 8.7, we support a new BPMN element: the ad-hoc subprocess. Process instance records of this type has the `BpmnElementType.AD_HOC_SUB_PROCESS`. 

We need to support the new element type in all places:
- Camunda exporter
- REST API
- Java SDK search API

saig0:
## Description

Add the missing flow-node type for ad-hoc subprocesses.

Add test cases to ensure we don't forget to add new types again. 

## Related issues

closes #28304


berkaycanbc (approved):
Super good job with test cases added. ⭐ 

LGTM 👍

PR #28338, opened 2025-02-18
deps: use 8.6.0 for identity and zeebe

RomanJRW:
## Description

External contributors cannot build the repo because Optimize is using alpha versions on the `stable/8.6` branch. This is already bumped on the `stable/optimize-8.6` branch ([example](https://github.com/camunda/camunda/commit/a65a1b7eb9b008d3986e8ab593c4fb76b499fece)), which is where Optimize is released from.

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)
  - [ ] enable backports [when recommended](https://github.com/camunda/camunda/wiki/CI-&-Automation#when-to-backport-ci-changes)

## Related issues

closes #

PR #28341, opened 2025-02-18
deps: use 8.6.8 for identity and zeebe in Optimize

RomanJRW:
## Description

8.7 equivalent of https://github.com/camunda/camunda/pull/28338

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/optimize/pom.xml
+++ b/optimize/pom.xml
@@ -14,8 +14,8 @@
     <camunda.maven.artifacts.version>7.21.0</camunda.maven.artifacts.version>
 
     <!-- We use this version to compile but run IT against containers with zeebe.docker.version -->
-    <zeebe.version>8.6.0-alpha5</zeebe.version>
-    <identity.version>8.6.0-alpha5</identity.version>
+    <zeebe.version>8.6.8</zeebe.version>
+    <identity.version>8.6.8</identity.version>
     <!-- We use this for the Zeebe test container version only -->
     <zeebe.docker.version>${zeebe.version}</zeebe.docker.version>
 


optimize/pom.xml
```diff
@@ -14,8 +14,8 @@
     <camunda.maven.artifacts.version>7.21.0</camunda.maven.artifacts.version>
 
     <!-- We use this version to compile but run IT against containers with zeebe.docker.version -->
-    <zeebe.version>8.6.0-alpha5</zeebe.version>
-    <identity.version>8.6.0-alpha5</identity.version>
+    <zeebe.version>8.6.8</zeebe.version>
+    <identity.version>8.6.8</identity.version>
```
korthout, line 18: 🤔 Shouldn't this just use something like `${project.version}`, i.e. `8.6.1-SNAPSHOT`?

korthout (approved):
I'll pre-approve to not block merging this, as it unblocks others. But please consider my questions above

optimize/pom.xml
RomanJRW: On 8.6 and 8.7, Optimize is not part of the single application, and doesn't inherit from the project root pom. I think the project version property here would therefore refer to the Optimize version, which are versioned independently from Identity and Zeebe

RomanJRW: > Also, why is it 8.6.1-SNAPSHOT on stable/8.7

I guess it could also be `8.7.0-SNAPSHOT` in this case. Either way, I would say it doesn't really matter, given that the Optimize 8.7 version is maintained from the `stable/optimize-8.7` branch rather than here

PR #28343, opened 2025-02-18
Migrate broker client metrics to Micrometer

npepinpe:
## Description

This PR migrates the broker client metrics to Micrometer for main. I will backport to 8.6/8.7 from here.

There were some required changes due to monorepo, and a new additional error code to account for. The metrics themselves are the same though.

## Related issues

related to #26078


zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/api/BrokerClientRequestMetrics.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /** Increments the count of successful requests for the given parameters, using these as tags. */
+  void registerSuccessfulRequest(
+      final int partitionId, final String requestType, final long latencyMs);
+
+  /** Increments the count of failed requests for the given parameters, using these as tags. */
+  void registerFailedRequest(final int partitionId, final String requestType, final Enum<?> error);
+
+  /** An implementation which simply does nothing, mostly useful for testing. */
+  final class Noop implements BrokerClientRequestMetrics {
```
entangled90, line 33: :wrench: if this is useful only for testing, then maybe we should move it to the `test  folder`?

zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/api/BrokerClientTopologyMetrics.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Sets the role of the broker with the given {@code brokerId} for the partition with the given
+   * {@code partitionId}
+   */
+  void setRoleForPartition(
+      final int partitionId, final int brokerId, final PartitionRoleValues roleValue);
+
+  /** An implementation which simply does nothing, mostly useful for testing. */
+  final class Noop implements BrokerClientTopologyMetrics {
```
entangled90, line 34: :wrench:  same as above

entangled90 (approved):
Pre-approving, just some minor comments

zeebe/broker-client/src/main/java/io/camunda/zeebe/broker/client/api/BrokerClientRequestMetrics.java
npepinpe: Honestly, I'm starting to think the interface is just not that useful in both cases, and we could just fold it. :shrug:

entangled90: if there's just one implementation :shrug:

PR #28353, opened 2025-02-18
Migrate log stream metrics to Micrometer

npepinpe:
## Description

This PR migrates the log stream metrics to Micrometer. There were new metrics added in comparison from 8.5, notably around the flow control, so this needs to be reviewed again, unfortunately.

## Related issues

related to #26078


zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/log/LogStreamImpl.java
```diff
@@ -68,7 +70,6 @@ public void close() {
     LOG.debug("Closing {} with {} readers", logName, readers.size());
     readers.forEach(LogStreamReader::close);
     logStorage.removeCommitListener(this);
-    logStreamMetrics.remove();
```
npepinpe, line 71: :exclamation: This is taken care of by closing the meter registry during a transition.

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/log/LogStreamBuilderImpl.java
```diff
@@ -24,6 +26,7 @@ public final class LogStreamBuilderImpl implements LogStreamBuilder {
   private InstantSource clock;
   private Limit requestLimit;
   private RateLimit writeRateLimit;
+  private MeterRegistry meterRegistry = new SimpleMeterRegistry();
```
entangled90, line 29: :question: is this really needed? This may cause hard to diagnose bugs down the line if the method `withMeterRegistry` is not called (which is a real possibility). If it's a required dependency we should add it to the builder constructor.
I see that it's the same as in `stable/8.5`, but it's probably worth changing it if does not implies too many changes

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/log/SequencerMetrics.java
```diff
@@ -7,42 +7,107 @@
...
   }
 
   void observeBatchLengthBytes(final int lengthBytes) {
     final int batchLengthKiloBytes = Math.floorDiv(lengthBytes, 1024);
-    batchLengthBytes.observe(batchLengthKiloBytes);
+    batchLengthBytes.record(batchLengthKiloBytes);
+  }
+
+  @SuppressWarnings("NullableProblems")
+  public enum SequencerMetricsDoc implements ExtendedMeterDocumentation {
```
entangled90, line 47: :thought_balloon: I think the Documnetation for most of the other metrics are in a separate file, so it may be more consistent to have them in a separate file. However, in branch `stable/8.5` is already like this so it would probably create conflicts in the feature, so let's not change it

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/log/LogStreamBuilderImpl.java
npepinpe: I think in 8.5 it's because there were tons of calls to the constructor and so on. It's much more limited on main, so let's go for it.

zeebe/logstreams/src/main/java/io/camunda/zeebe/logstreams/impl/log/SequencerMetrics.java
npepinpe: I guess it's because it was so small :smile: I don't think it's a huge deal, imo. If it's 1 metric, it's like OK, as long as it's public.

entangled90 (approved):

PR #28354, opened 2025-02-18
fix: Error message links on dashboard show wrong process instances

ghost:
## Description

- we are going to add the field `incidentErrorHashCode` to the filter for process instances (internal API). This field will be used in conjunction with the `errorMessage` field, to narrow down the search results, for example when the error message is too long
- the field  `incidentErrorHashCode` is also returned by the `/api/incidents/byError` endpoint to front-end, so that front-end can pass it back as filter to back-end
- the field `incidentErrorHashCode` is already stored in the incident index, so we don't need to add it to the list-view index, i.e. we don't need to migrate / import
- the field will work in the following way as filter:
  - if the field `incidentErrorHashCode` is provided in the API request for process instances, back-end will make an additional request, to first get the (full) error message(s) that match the given hash code, from the incident index. Then, using the full error message(s), back-end will make a normal filter request to the list-view index, to get the matching process instances, by error message.
  - if both  `errorMessage` and `incidentErrorHashCode` are provided in the filter, then there will be an AND filter request, between the `errorMessage` provided, and the error message matching `incidentErrorHashCode`. This means that, in order to get a hit, the `errorMessage` provided must be a substring of the error message matching `incidentErrorHashCode`

Closes https://github.com/camunda/operate/issues/6630

### DIFF
--- a/operate/client/e2e-playwright/mocks/dashboard.mocks.ts
+++ b/operate/client/e2e-playwright/mocks/dashboard.mocks.ts
@@ -20,6 +20,7 @@ const mockStatistics = {
 const mockIncidentsByError = [
   {
     errorMessage: 'No more retries left.',
+    incidentErrorHashCode: 1111,
     instancesWithErrorCount: 170,
     processes: [
       {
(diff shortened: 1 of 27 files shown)


pedesen:
@vsgoulart Could you please review the FE part? This is a backport for 8.8, the code is the same as https://github.com/camunda/camunda/pull/28138

vsgoulart (approved):

vsgoulart:
@pedesen are the E2E failures related?

pedesen:
> @pedesen are the E2E failures related?

Seems to be unrelated flaky tests.

sdorokhova (approved):

PR #28365, opened 2025-02-19
Schema manager updates settings

### LINKED ISSUE #28111: Camunda exporter schema manager not updating index template settings
**Describe the bug**

When the Camunda Exporter starts up it runs the `SchemaManager.startup` which will update the mappings of index templates to that defined in the json files of those index templates. 

However the json files can also contain a settings block, and currently it will not update the settings block of the existing index template with that of the json file. 

This is resulting in such bugs such as the deserialisation issue with type: custom with elasticsearch referenced here https://github.com/elastic/elasticsearch-java/issues/696, which is also causing issues with the camunda migration to 8.8.

**To Reproduce**

- Create an 8.7 cluster.
- check that operate-list-view template, lowercase normalizer does not have a type custom field
- Create an 8.8 cluster, the operate-list-view template json file for 8.8 has a type custom field so after startup the index template setting should be updated

**Expected behavior**

The index template settings gets updated with the value from the source json.

EuroLew:
## Description

This is a first draft  to confirm the approach solves the problems it needs to. Based on the requirements defined https://github.com/camunda/camunda/issues/32872

Add ability to update settings of index and index templates:

Indices:
- Only able to update the replicas of live indices

Index templates:
- Able to update all settings for the index template.

Note the configuration for replica/shards in `ExporterConfiguration` which can be templated from envs/files etc.. takes ultimate precedence.

1. If you update shards/replica through configuration it will apply fully to index templates 
2. If you update shards/replica through configuration ONLY the replica change will apply to live indices.
3. if you update the json schema file it will fully apply to index templates (with the exception of replicas and shards) with NO changes to the corresponding index.

## Related issues

closes #28111

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
@@ -59,7 +59,7 @@ public class SchemaManager {
     updateSchemaMappings(newIndexProperties);
     LOG.info(
         "Update index template schema. '{}' index templates need to be updated",
-        newIndexProperties.size());
+        newIndexTemplateProperties.size());
     updateSchemaMappings(newIndexTemplateProperties);
     updateSchemaSettings();
 
@@ -89,7 +89,7 @@ public class SchemaManager {
         .forEach(
             desc -> {
               searchEngineClient.updateIndexTemplateSettings(
-                  desc, getIndexSettings(desc.getIndexName()));
+                  desc, getIndexSettingsFromConfig(desc.getIndexName()));
             });
 
     // update matching index for the index template descriptor
@@ -103,10 +103,10 @@ public class SchemaManager {
   }
 
   private void updateIndexReplicaCount(final IndexDescriptor indexDescriptor) {
-    final var indexSettings = getIndexSettings(indexDescriptor.getIndexName());
+    final var indexReplicaCount =
+        String.valueOf(getNumberOfReplicasFromConfig(indexDescriptor.getIndexName()));
     searchEngineClient.putSettings(
-        List.of(indexDescriptor),
-        Map.of("index.number_of_replicas", indexSettings.getNumberOfReplicas().toString()));
+        List.of(indexDescriptor), Map.of("index.number_of_replicas", indexReplicaCount));
   }
 
   public void initialiseResources() {
@@ -133,7 +133,7 @@ public class SchemaManager {
             descriptor -> {
               LOG.info("Create missing index '{}'", descriptor.getFullQualifiedName());
               searchEngineClient.createIndex(
-                  descriptor, getIndexSettings(descriptor.getIndexName()));
+                  descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()));
             });
   }
 
@@ -158,13 +158,13 @@ public class SchemaManager {
             descriptor -> {
               LOG.info("Create missing index template '{}'", descriptor.getTemplateName());
               searchEngineClient.createIndexTemplate(
-                  descriptor, getIndexSettings(descriptor.getIndexName()), true);
+                  descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()), true);
               LOG.info(
                   "Create missing index '{}', for template '{}'",
                   descriptor.getFullQualifiedName(),
                   descriptor.getTemplateName());
               searchEngineClient.createIndex(
-                  descriptor, getIndexSettings(descriptor.getIndexName()));
+                  descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()));
             });
   }
 
@@ -179,7 +179,7 @@ public class SchemaManager {
             "Updating template: '{}'", ((IndexTemplateDescriptor) descriptor).getTemplateName());
         searchEngineClient.createIndexTemplate(
             (IndexTemplateDescriptor) descriptor,
-            getIndexSettings(descriptor.getIndexName()),
+            getIndexSettingsFromConfig(descriptor.getIndexName()),
             false);
       } else {
         LOG.info(
@@ -191,12 +191,8 @@ public class SchemaManager {
     }
   }
 
-  private IndexSettings getIndexSettings(final String indexName) {
-    final var templateReplicas =
-        config
-            .getIndex()
-            .getReplicasByIndexName()
-            .getOrDefault(indexName, config.getIndex().getNumberOfReplicas());
+  private IndexSettings getIndexSettingsFromConfig(final String indexName) {
+    final var templateReplicas = getNumberOfReplicasFromConfig(indexName);
     final var templateShards =
         config
             .getIndex()
@@ -208,6 +204,13 @@ public class SchemaManager {
     settings.setNumberOfReplicas(templateReplicas);
 
     return settings;
+  }
+
+  private int getNumberOfReplicasFromConfig(final String indexName) {
+    return config
+        .getIndex()
+        .getReplicasByIndexName()
+        .getOrDefault(indexName, config.getIndex().getNumberOfReplicas());
   }
 
   private Map<IndexDescriptor, Collection<IndexMappingProperty>> validateIndices(
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
@@ -14,6 +14,7 @@ import io.camunda.exporter.config.ExporterConfiguration.RetentionConfiguration;
 import io.camunda.webapps.schema.descriptors.IndexDescriptor;
 import io.camunda.webapps.schema.descriptors.IndexTemplateDescriptor;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
@@ -60,6 +61,7 @@ public class SchemaManager {
         "Update index template schema. '{}' index templates need to be updated",
         newIndexProperties.size());
     updateSchemaMappings(newIndexTemplateProperties);
+    updateSchemaSettings();
 
     final RetentionConfiguration retention = config.getArchiver().getRetention();
     if (retention.isEnabled()) {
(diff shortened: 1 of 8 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -73,6 +75,40 @@ public void startup() {
...
+
+    final var existingIndexNames =
+        searchEngineClient.getMappings(allIndexNames(), MappingSource.INDEX).keySet();
+
+    indexTemplateDescriptors.stream()
+        .filter(desc -> existingTemplateNames.contains(desc.getTemplateName()))
+        .forEach(
+            desc -> {
+              searchEngineClient.updateIndexTemplateSettings(
+                  desc, getIndexSettings(desc.getIndexName()));
```
ChrisKujawa, line 92: 🔧 Just a nitpick: Maybe we call it `getIndexSettingsFromConfig` or something, because for me it was not clear where these settings come from and whether it is not from ES or config, etc.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -73,6 +75,40 @@ public void startup() {
...
+        .filter(desc -> existingIndexNames.contains(desc.getFullQualifiedName()))
+        .forEach(this::updateIndexReplicaCount);
+
+    indexDescriptors.stream()
+        .filter(desc -> existingIndexNames.contains(desc.getFullQualifiedName()))
+        .forEach(this::updateIndexReplicaCount);
+  }
+
+  private void updateIndexReplicaCount(final IndexDescriptor indexDescriptor) {
+    final var indexSettings = getIndexSettings(indexDescriptor.getIndexName());
```
ChrisKujawa, line 106: 🔧 I feel we could also simply do:

```
        config
            .getIndex()
            .getReplicasByIndexName()
            .getOrDefault(indexName, config.getIndex().getNumberOfReplicas());
```

Instead of constructing the settings object with unnecessary values, etc. You can also put this into an own method to get the replicaSetting from config, and use it in both places.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
```diff
@@ -227,6 +227,29 @@ public boolean importersCompleted(
...
+        updateTemplateSettings(indexTemplateDescriptor, currentSettings);
+
+    try {
+      client.indices().putIndexTemplate(updateIndexTemplateSettingsRequest);
+    } catch (final IOException | ElasticsearchException e) {
+      throw new ElasticsearchExporterException(
+          String.format(
+              "Failed to update index template settings for [%s]",
+              indexTemplateDescriptor.getTemplateName()),
+          e);
```
ChrisKujawa, line 249: 🔧 Let's try to follow our error guidelines https://github.com/camunda/camunda/wiki/Error-Guidelines and give some more context

Example:

```suggestion
      throw new ElasticsearchExporterException(
          String.format(
              "Expected to update index template settings '%s' with '%s', but failed ",
              indexTemplateDescriptor.getTemplateName(),
              updateIndexTemplateSettingsRequest),
          e);
```

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
```diff
@@ -386,6 +411,47 @@ private PutIndexTemplateRequest putIndexTemplateRequest(
...
+                      .mappings(currentIndexTemplateState.mappings())
+                      .aliases(currentIndexTemplateState.aliases()))
+          .build();
+
+    } catch (final IOException e) {
+      throw new ElasticsearchExporterException(
+          "Failed to load file "
+              + indexTemplateDescriptor.getMappingsClasspathFilename()
+              + " from classpath.",
+          e);
```
ChrisKujawa, line 451: ❓  I feel like this is only partially true, can this happen also on other calls like the client.indices?

zeebe/exporters/camunda-exporter/src/test/resources/mappings-and-updated-settings.json
```diff
@@ -0,0 +1,18 @@
+{
+  "settings": {
+    "index": {
+      "refresh_interval": "5s"
```
ChrisKujawa, line 4: 👍🏼

ChrisKujawa: Can we have also one with updated replicas here and additionally one with using this file and configure it, so we see the order in test (as you have defined in your PR description). Simply, to show that it is not some random behavior it is intended.

ChrisKujawa (approved):
Great @EuroLew thanks 👍🏼 looks in general good to me, I had smaller comments please see below.

I think it makes sense that @ThorbenLindhauer also looks at it, whether it is what he expected.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
ThorbenLindhauer, line 92: For templates, is this effectively already happening through `updateSchemaMappings` already, particularly https://github.com/camunda/camunda/blob/36e2bdf57d673a874a820cbdc0ad4c3f5245da6f/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java#L180-L183 (note the boolean false parameter means that the template gets overwritten)? I'm not super confident in my observation here, but want to ask.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -55,11 +56,12 @@ public void startup() {
 
     //  used to update existing indices/templates
     LOG.info("Update index schema. '{}' indices need to be updated", newIndexProperties.size());
-    updateSchema(newIndexProperties);
+    updateSchemaMappings(newIndexProperties);
     LOG.info(
         "Update index template schema. '{}' index templates need to be updated",
         newIndexProperties.size());
```
ThorbenLindhauer, line 62: Not related to your change, but I think we want to log `newIndexTemplateProperties.size()` here, or?

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
```diff
@@ -435,6 +435,120 @@ void shouldAlsoUpdateCorrespondingIndexWhenIndexTemplateUpdated(
...
+        searchClientAdapter.getIndexTemplateAsNode(indexTemplate.getTemplateName());
+    final var updatedMatchingIndex =
+        searchClientAdapter.getIndexAsNode(indexTemplate.getFullQualifiedName());
+
+    assertThat(updatedTemplate.at(indexTemplateSettingsToBeAppended).asText()).isEqualTo("5s");
+    assertThat(updatedMatchingIndex.at(indexSettingsToBeAppended).asText()).isEqualTo("");
+  }
+
+  @TestTemplate
+  void shouldUpdateIndexTemplateWithNewReplicaAndShardCount(
```
ThorbenLindhauer, line 483: Maybe we can also have a test case that the main index of the template is updated correctly. We have dedicated code for this case, so I think that justifies a dedicated test case.

ThorbenLindhauer (approved):
Looks good to me. This seems to achieve exactly what we need. I left some detail comments.

Note that the [tasklist implementation](https://github.com/camunda/tasklist/pull/4697) does this more defensively by first checking what the current settings are and only changing them if they differ. But I don't have a particular argument why that approach should be better.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
ThorbenLindhauer: Slight correction: I mean justifies *test coverage*, not necessarily a test case.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
EuroLew: yes it is happening from `updateSchemaMappings` but note that function ONLY calls `createIndexTemplate` with create = false (so it's an update and overwritten), if there are new fields in the mappings. Therefore:
- Change mappings and settings block in json, both changes affect the index template
- only change settings block in json, no changes happen

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/elasticsearch/ElasticsearchEngineClient.java
EuroLew: you're right two different causes, I added two different catchecs

zeebe/exporters/camunda-exporter/src/test/resources/mappings-and-updated-settings.json
EuroLew: hey sorry my behaviour documented in the PR description was wrong, if you put replicas and shards in the json file, it will never have an effect, I changed my description to be correct.

`if you update the json schema file it will fully apply to index templates (with the exception of replicas and shards) with NO changes to the corresponding index.`

This means I will write a test where the json file specifies replicas and shards but show that the configuration always takes precendence

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
EuroLew: not sure what you mean by this. any updates the the json file of an index template should have no effect on the corresponding index I wrote a test for this `shouldUpdateSettingsForIndexTemplatesButNotUpdateIndexSettingsWhenSchemaChanges`

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
EuroLew: although your comment means there is a way better solution, I don't need to implement `updateIndexTemplateSettings` I can just remove all code associated and use the createIndexTemplate as it also doubles as a put

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/schema/SchemaManagerIT.java
ThorbenLindhauer: It's about the number of replicas config, not the schema mapping json.

I'll try to rephrase: Will a test fail, if we remove the following code?

https://github.com/camunda/camunda/blob/3165ca2b131d9c75523577380db8730f98d5b91b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java#L95-L98

I want to make sure it's covered :) Maybe it is already, as I am not super deep into the matter here.

EuroLew: `shouldUpdateIndexWithNewReplicaCountButNotNewShardCount` will fail if that is removed, as corresponding index for index templates are not updated

PR #28368, opened 2025-02-19
refactor: remove C7 plugin config from Optimize C8

abremard:
## Description

This is part of the "C7 config cleanup from C8". Because there's a lot of configs that need cleaning, I'm breaking this work down into multiple PRs. This one is focused on "plugins config" cleanup. I'm not seeing any usage of these plugin configs so this change should be relatively safe. For this update there is a reference in [C8 docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/#other) which needs cleaning up. I will do this once this PR is approved and merged.

## Related issues


RomanJRW (approved):
This is indeed a safe removal - thanks!

PR #28373, opened 2025-02-19
refactor: remove C7 decision cleanup config from Optimize C8

abremard:
## Description

This is part of the "C7 config cleanup from C8". Because there's a lot of configs that need cleaning, I'm breaking this work down into multiple PRs. This one is focused on "decision cleanup config" cleanup. I'm removing the decision data cleanup service here because we've disabled decision data in C8. Most of the lines removed are setters/getters and boilerplates. For this update there is a reference in [C8 docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/#history-cleanup-settings) which needs cleaning up. I will do this once this PR is approved and merged.

## Related issues


RomanJRW (approved):
LGTM!

PR #28376, opened 2025-02-19
refactor: remove C7 tenants cache config from Optimize C8

abremard:
## Description

This is part of the "C7 config cleanup from C8". Because there's a lot of configs that need cleaning, I'm breaking this work down into multiple PRs. This one is focused on "tenants cache config" cleanup. Most of the lines removed are setters/getters and boilerplates. For this update I don't see any references to the `tenants cache config` in [C8 docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/)

## Related issues

### DIFF
--- a/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/GlobalCacheConfiguration.java
+++ b/optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/GlobalCacheConfiguration.java
@@ -9,7 +9,6 @@ package io.camunda.optimize.service.util.configuration;
 
 public class GlobalCacheConfiguration {
 
-  private CacheConfiguration tenants;
   private CacheConfiguration definitions;
   private CacheConfiguration definitionEngines;
   private CloudUserCacheConfiguration cloudUsers;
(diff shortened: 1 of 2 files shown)


optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/GlobalCacheConfiguration.java
```diff
@@ -18,14 +17,6 @@ public class GlobalCacheConfiguration {
...
   public GlobalCacheConfiguration() {}
 
-  public CacheConfiguration getTenants() {
-    return tenants;
-  }
-
-  public void setTenants(final CacheConfiguration tenants) {
-    this.tenants = tenants;
-  }
-
```
abremard, line 28: These are not used anywhere. Worth noting that there is a "tenants cache" being created in [this test](https://github.com/camunda/camunda/blob/4aa292960bf17f49a9d978912f4b5fd7881cef5a/optimize/backend/src/test/java/io/camunda/optimize/service/tenant/CCSMTenantAuthorizationServiceTest.java#L52) but my understanding is that the "tenant cache" there refers to "cloud tenant authorizations" ([this yaml config](https://github.com/camunda/camunda/blob/eae72cb5272ffb680358ab76e55d2fd5aaa6cf86/optimize/util/optimize-commons/src/main/resources/service-config.yaml#L797))

RomanJRW (approved):

PR #28379, opened 2025-02-19
refactor: remove C7 variable ingestion config from Optimize C8

abremard:
## Description

This is part of the "C7 config cleanup from C8". Because there's a lot of configs that need cleaning, I'm breaking this work down into multiple PRs. This one is focused on "variable ingestion config" cleanup. Most of the lines removed are setters/getters and boilerplates. For this update I'm seeing a reference in the [C8 docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/#external-variable-ingestion-rest-api-configuration) which needs cleaning up. I will do this once this PR is approved and merged.

## Related issues


optimize/util/optimize-commons/src/test/java/io/camunda/optimize/service/util/configuration/ConfigurationValidatorTest.java
```diff
@@ -155,8 +155,7 @@ public void deletedAccessTokenConfigs() {
     // then
     assertThat(deletions)
         .hasSize(2)
-        .containsEntry("eventBasedProcess.eventIngestion.accessToken", DOC_URL)
-        .containsEntry("externalVariable.variableIngestion.accessToken", DOC_URL);
```
abremard, line 159: The access token config has been removed in a previous PR I believe and it seems like there's leftovers in test files.

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java
```diff
@@ -1272,11 +1272,6 @@ public void setExternalVariableConfiguration(
     this.externalVariableConfiguration = externalVariableConfiguration;
   }
 
-  @JsonIgnore
-  public VariableIngestionConfiguration getVariableIngestionConfiguration() {
```
RomanJRW, line 1276: I'm confused as I thought that the external variable ingestion was enabled in C8? 🤔

RomanJRW (changes requested):

RomanJRW:
@abremard marked this as change requested but not actually requesting any changes, per se. It would just be good to get clarity on this together

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/service/util/configuration/ConfigurationService.java
abremard: I'm not finding any calls to this method, the only reference to "variableIngestion" I find is [this comment](https://github.com/camunda/camunda/blob/a299ca5ba759d90f83fb0ded48bae7ef0aa98211/optimize/backend/src/main/java/io/camunda/optimize/FilterBeansConfig.java#L19) and I'm not able to find issues mentioning it in [github search](https://github.com/search?q=repo%3Acamunda%2Fcamunda+variableingestion&type=code). Looks like it was first introduced by this [old commit](https://github.com/camunda/camunda/commit/01baac0092f437151f8cc24222dea0426214f650#diff-55c413615854c652f4f8a2a26ceab4a0fe80fb684d62e47be5ff81b0665825bd)

RomanJRW: Ok I think we are probably safe to remove this 🙂

RomanJRW (approved):

RomanJRW:
@abremard - I actually wouldn't consider this a bug so don't see a need for this to be backported 🙂

PR #28388, opened 2025-02-19
ci: fetch depth of 0 for deploy workflow

megglos:
## Description

Workaround for https://github.com/yarnpkg/berry/issues/4014 .

Caused builds of the workflow to fail with `[INFO] Usage Error: No ancestor could be found between any of HEAD and master, origin/master, upstream/master, main, origin/main, upstream/main`, see e.g. https://github.com/camunda/camunda/actions/runs/13411100239


megglos:
Test run https://github.com/camunda/camunda/actions/runs/13411533959

Ben-Sheppard (approved):
LGTM - this issue came up before but we hadn't committed to the fix as it was only happening in certain circumstances, that being said, this is the fix from what I've read and you've shown it works

Issue #27193, opened 2025-01-21
Publish 8.7.0-SNAPSHOT from stable/8.7 and bump main to 8.8.0-SNAPSHOT


cmur2:
## Description

Sub-task of https://github.com/camunda/product-hub/issues/2618

Goal of this task is to adjust CI pipelines to achieve that the following SNAPSHOT releases are published on push to respective branches:

* `8.7.0-SNAPSHOT` from `stable/8.7` + `stable/optimize-8.7` (to Artifactory, DockerHub)
* `8.8.0-SNAPSHOT` from `main` (to Artifactory)
* `SNAPSHOT` from `main` for [Operate](https://hub.docker.com/r/camunda/operate/tags?name=SNAPSHOT), [Tasklist](https://hub.docker.com/r/camunda/tasklist/tags?name=SNAPSHOT), [Zeebe](https://hub.docker.com/r/camunda/zeebe/tags?name=SNAPSHOT) (to DockerHub, we don't include version info for SNAPSHOT Docker images)
* `8-SNAPSHOT` from `main` for [Optimize](https://hub.docker.com/r/camunda/optimize/tags?name=SNAPSHOT) (to DockerHub, due to [coexistence with C7 flavor](https://camunda.slack.com/archives/C55U06YRG/p1737547747874549?thread_ts=1737545048.821109&cid=C55U06YRG))

## Tasks

Once the `stable/8.7` branch is created:

* ✔ backport https://github.com/camunda/camunda/pull/24309 to `stable/8.7`
  * https://github.com/camunda/camunda/pull/27246
* ✔ on `stable/8.7` branch adjust GHA to also publish snapshots to Artifactory: [1](https://github.com/camunda/camunda/blob/8eca29a7962e6991146cdd85187f8c2744ee9922/.github/workflows/ci.yml#L772)
  * https://github.com/camunda/camunda/pull/27243 waits on https://github.com/camunda/camunda/pull/27210
* ✔ on `stable/optimize-8.7` branch adjust GHA to also publish snapshots to Artifactory: [1](https://github.com/camunda/camunda/blob/f84319e60a9ba485e390f60ee6dc724bbf030132/.github/workflows/optimize-deploy-artifacts.yml#L89)
[...]

PR #28400, opened 2025-02-19
ci: snapshot uploads for 8.7 and 8.8 are independent

cmur2:
## Description

Currently, the snapshot uploads on `main` and `stable/8.7` (target of this PR!) use the same GHA concurrency group names. This can lead to unnecessary waiting when e.g. an upload is running for `stable/8.7` and one wants to run for `main`.

Technically, those uploads are independent because they target different artifacts (different Maven package files or Docker image tags). So there is no race condition between those 2 kinds of uploads thus different GHA concurrency groups should be used to avoid unnecessary waiting.

## Checklist

- [x] for CI changes:
  - [ ] enable backports [when recommended](https://github.com/camunda/camunda/wiki/CI-&-Automation#when-to-backport-ci-changes) - not needed since `stable/8.7` is the only other branch with this behavior

## Related issues

Related #27193

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -727,7 +727,7 @@ jobs:
     permissions: {}  # GITHUB_TOKEN unused in this job
     if: github.repository == 'camunda/camunda' && github.ref == 'refs/heads/stable/8.7'
     concurrency:
-      group: deploy-maven-snapshot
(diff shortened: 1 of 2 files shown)


Kerruba (approved):
If I'm understanding correctly, this is moving the 8.7 snapshot uploads to a dedicated concurrency group right?

PR #28412, opened 2025-02-19
Write followup events/command of `User.UPDATE` with correct user key

### LINKED ISSUE #28404: User key is set to -1 on user update
## Description

When creating a user a key is generated for this user. We will store this in RocksDB and Elasticsearch. When you update this same user, the key gets overridden to `-1L`. We should make sure we keep the initially generated key.

remcowesterhoud:
## Description

We set the user key to the key that was on the command. This will always be -1 since it's a command received from the 
gateway. As users are identified by their username there's no need to have a userkey set.

We should make sure on followup commands that we set the userkey properly. We can do this by taking it from the 
persisted user and setting it on the command. As well as ensuring that we write any followup commands/events with the 
correct user key.

## Checklist

- [ ] for CI changes:

## Related issues

closes #28404


Ben-Sheppard (approved):
LGTM :)

Issue #28418, opened 2025-02-19
Uploading a document in local storage without an explicit ID does not return its ID


georgios-goulos:
**Describe the bug**

Uploading a document in local storage without an explicit ID does not return its ID in the API response

**To Reproduce**

Upload a document to local storage using the API

**Expected behavior**

The API response contains the document ID

**Actual behavior**
The document ID is missing from the API response. Example of API response
```
{
    "camunda.document.type": "camunda",
    "storeId": "local",
    "contentHash": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "metadata": {
        "contentType": "application/octet-stream",
        "fileName": "doc.bpmn",
        "size": 3173,
        "customProperties": {}
    }
}
```

**Environment:**
- OS: any
- Zeebe Version: 8.8, 8.7

PR #28419, opened 2025-02-19
fix: Return document ID when storing doc locally

georgios-goulos:
## Description

The document ID was not being returned in the API response for document uploads to local storage. This PR fixes that

## Related issues

1st PR towards closing https://github.com/camunda/camunda/issues/28418


panagiotisgts (approved):
LGTM

PR #28430, opened 2025-02-19
fix: Align states of process instance and flow-node instance

### LINKED ISSUE #27156: [C8 API] Align states of process instance and flow-node instance
**Description**

The C8 REST API (v2) contains Get/Search APIs for process instances and flow-node instances (i.e. the BPMN element instances of a process instance). The [process instance](https://github.com/camunda/camunda/blob/dc945dbabf6ad522e3b55458d72f03ae0503ab36/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L5149-L5154) and the [flow-node instance](https://github.com/camunda/camunda/blob/dc945dbabf6ad522e3b55458d72f03ae0503ab36/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L5345-L5351) entities have a property **state** with similar values but one difference:
* `ACTIVE` (both)
* `COMPLETED` (both)
* `CANCELED` (process instance) **vs.** `TERMINATED` (flow-node instance) 

The Operate v1 API had the same difference: [Search flownode-instances](https://docs.camunda.io/docs/next/apis-tools/operate-api/specifications/search-4/) vs. [Search process instances](https://docs.camunda.io/docs/next/apis-tools/operate-api/specifications/search-1/)

However, there is no good reason to have two different terms for the same state. We should align the terms to streamline the experience. 

---

Usage of the term CANCELED:
* The [Cancel process instance API](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/cancel-process-instance/)

Usage of the term TERMINATED:
* On the record stream and for exporters, both entities have the record intent: `ELEMENT_TERMINATED`
* The [modification API](https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/modify-process-instance/)  contains `terminateInstructions`

ghost:
fix: Align states of process instance and flow-node instance

- replaced state CANCELED with state TERMINATED in V2 API

Closes https://github.com/camunda/camunda/issues/27156

### DIFF
--- a/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ProcessInstanceAssertj.java
+++ b/testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ProcessInstanceAssertj.java
@@ -84,7 +84,7 @@ public class ProcessInstanceAssertj
   @Override
   public ProcessInstanceAssert isTerminated() {
     hasProcessInstanceInState(
-        "terminated", ProcessInstanceState.CANCELED::equals, ProcessInstanceAssertj::isEnded);
+        "terminated", ProcessInstanceState.TERMINATED::equals, ProcessInstanceAssertj::isEnded);
     return this;
   }
 
@@ -95,7 +95,7 @@ public class ProcessInstanceAssertj
         state ->
             state == ProcessInstanceState.ACTIVE
                 || state == ProcessInstanceState.COMPLETED
-                || state == ProcessInstanceState.CANCELED,
+                || state == ProcessInstanceState.TERMINATED,
         Objects::nonNull);
     return this;
   }
@@ -266,8 +266,6 @@ public class ProcessInstanceAssertj
   private static String formatState(final ProcessInstanceState state) {
     if (state == null || state == ProcessInstanceState.UNKNOWN_ENUM_VALUE) {
       return "not created";
-    } else if (state == ProcessInstanceState.CANCELED) {
-      return "terminated";
     } else {
       return state.name().toLowerCase();
     }
--- a/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
+++ b/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
@@ -28,7 +28,20 @@ public class AdvancedSearchFilterUtil {
     return (final Object filter) -> mapToOperations(filter, tClass);
   }
 
+  public static <T> Function<Object, List<Operation<T>>> mapToOperations(
+      final Class<T> tClass, final CustomConverter<T> customConverter) {
+    return (final Object filter) -> mapToOperations(filter, tClass, customConverter);
+  }
+
   protected static <T> T convertValue(final Class<T> tClass, final Object value) {
+    return convertValue(tClass, value, null);
+  }
+
+  protected static <T> T convertValue(
+      final Class<T> tClass, final Object value, final CustomConverter<T> customConverter) {
+    if (customConverter != null && customConverter.canConvert(value)) {
+      return customConverter.convertValue(value);
+    }
     if (value == null) {
       return null;
     } else if (tClass.isInstance(value)) {
@@ -51,6 +64,11 @@ public class AdvancedSearchFilterUtil {
 
   protected static <T> List<Operation<T>> mapToOperations(
       final Object filter, final Class<T> tClass) {
+    return mapToOperations(filter, tClass, null);
+  }
+
+  protected static <T> List<Operation<T>> mapToOperations(
+      final Object filter, final Class<T> tClass, final CustomConverter<T> customConverter) {
     final var fClass = filter.getClass();
     final var operations = new ArrayList<Operation<T>>();
     for (final Operator operator : Operator.values()) {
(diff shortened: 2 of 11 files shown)


tmetzke (changes requested):
👍 API changes and response mapping look good to me
❌ The client enum should simply be adjusted as well
❓ Do we need to adjust the request mapping as well, i.e. mapping the state filter in the search request so we use the right value in ES/OS/RDBMS queries?

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ProcessInstanceAssertj.java
```diff
@@ -266,7 +266,7 @@ private static boolean isEnded(final ProcessInstanceDto processInstance) {
   private static String formatState(final ProcessInstanceState state) {
     if (state == null || state == ProcessInstanceState.UNKNOWN_ENUM_VALUE) {
       return "not created";
-    } else if (state == ProcessInstanceState.CANCELED) {
+    } else if (state == ProcessInstanceState.TERMINATED) {
       return "terminated";
     } else {
       return state.name().toLowerCase();
```
tmetzke, line 272: 🙃 This can be simplified now, I think
```suggestion
    } else {
      return state.name().toLowerCase();
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
```diff
@@ -34,6 +36,11 @@ protected static <T> T convertValue(final Class<T> tClass, final Object value) {
     } else if (tClass.isInstance(value)) {
       return tClass.cast(value);
     } else if (tClass == String.class) {
+      if (value instanceof ProcessInstanceStateEnum processInstanceStateEnum) {
+        if (processInstanceStateEnum == ProcessInstanceStateEnum.TERMINATED) {
+          return (T) ProcessInstanceEntity.ProcessInstanceState.CANCELED.toString();
+        }
+      }
       return (T) value.toString();
```
mihail-ca, line 44: @tmetzke I adjusted the request to map TERMINATED to CANCELED here, let me know if this is the correct place to do it, thank you.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
```diff
@@ -34,6 +36,11 @@ protected static <T> T convertValue(final Class<T> tClass, final Object value) {
     } else if (tClass.isInstance(value)) {
       return tClass.cast(value);
     } else if (tClass == String.class) {
+      if (value instanceof ProcessInstanceStateEnum processInstanceStateEnum) {
+        if (processInstanceStateEnum == ProcessInstanceStateEnum.TERMINATED) {
+          return (T) ProcessInstanceEntity.ProcessInstanceState.CANCELED.toString();
+        }
+      }
```
tmetzke, line 43: 💭 I am wondering if we should do the conversion somewhere earlier, e.g. in the request mapping. That way, we don't introduce this in a generic place. When we create the search API filter from the REST DTO filter, we should transform the `ProcessInstanceStateEnum` to a `ProcessInstanceEntity.ProcessInstanceState`. This is where we can map from one enum to another.

mihail-ca: Based on the stack trace, this is kind of the earliest place where we can do the conversion. This `AdvancedSearchFilterUtil` is actually called by `SearchQueryRequestMapper`, among others.
image

mihail-ca: In the `SearchQueryRequestMapper`, the filter parameter is the Open API model, so we cannot change the enum value earlier:
https://github.com/camunda/camunda/blob/main/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/SearchQueryRequestMapper.java#L490

zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/ProcessInstanceQueryControllerTest.java
```diff
@@ -534,4 +536,42 @@ void shouldSearchProcessInstancesWithAdvancedFilter(
     verify(processInstanceServices)
         .search(new ProcessInstanceQuery.Builder().filter(filter).build());
   }
+
+  @Test
+  void shouldSearchProcessInstancesByState() {
```
tmetzke, line 541: 🔧 I believe we can make this a parameterized test, using the enum class. You can remove the loop from the test method then.

```suggestion
  @ParameterizedTest
  @EnumSource(ProcessInstanceStateEnum.class) 
  void shouldSearchProcessInstancesByState(ProcessInstanceStateEnum state) {
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/util/AdvancedSearchFilterUtil.java
```diff
@@ -28,7 +28,20 @@ public static <T> Function<Object, List<Operation<T>>> mapToOperations(final Cla
...
+
   protected static <T> T convertValue(final Class<T> tClass, final Object value) {
+    return convertValue(tClass, value, null);
+  }
+
+  protected static <T> T convertValue(
+      final Class<T> tClass, final Object value, final CustomConverter<T> customConverter) {
+    if (customConverter != null && customConverter.canConvert(value)) {
+      return customConverter.convertValue(value);
+    }
```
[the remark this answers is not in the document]
mihail-ca: I had to add the `canConvert` method, because the converter is called to convert also things that are not enum, for example the string "act" in [this test](https://github.com/camunda/camunda/blob/be-27156-c8-api/zeebe/gateway-rest/src/test/java/io/camunda/zeebe/gateway/rest/controller/ProcessInstanceQueryControllerTest.java#L490). I am not really sure why it is supposed to work like this, i.e. why it is called with different data types. Regarding you suggested changes, personally I find functional code difficult to read/debug. But I will try to adjust as you said.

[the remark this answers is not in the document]
danielkelemen: It's because of the like search, that still requires a simple `string`. Alternatively, we could consider dropping like search for enum values.

tmetzke (approved):

PR #28433, opened 2025-02-19
SchemaManager: Parallelize index creation

ChrisKujawa:
## Description

To make index and template creation quicker, as suggested by @entangled90.

For that we use virtual threads for the creation of indices and templates. This is to make bootstrapping quicker, especially for tests quite useful. We can see a test reduction (will post results tomorrow).

> [!Note]
>
> @npepinpe mentioned it might make more sense, to use ES async client, but for index creations I did find any async method. Open for any hints.

## Related issues

closes #

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
@@ -21,11 +21,15 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import org.agrona.LangUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class SchemaManager {
+
+  public static final int INDEX_CREATION_TIMEOUT_SECONDS = 60;
   private static final Logger LOG = LoggerFactory.getLogger(SchemaManager.class);
   private final SearchEngineClient searchEngineClient;
   private final Collection<IndexDescriptor> indexDescriptors;
@@ -154,7 +158,7 @@ public class SchemaManager {
 
     // We need to wait for the completion, to make sure all indices has been created successfully
     // Doing this in parallel is still speeding up the bootstrap time
-    CompletableFuture.allOf(futures).join();
+    joinOnFutures(futures);
   }
 
   private void initialiseIndexTemplates() {
@@ -200,7 +204,23 @@ public class SchemaManager {
     // We need to wait for the completion, to make sure all indices and templates have been created
     // successfully
     // Doing this in parallel is still speeding up the bootstrap time
-    CompletableFuture.allOf(futures).join();
+    joinOnFutures(futures);
+  }
+
+  /**
+   * Join on given futures with {@link SchemaManager#INDEX_CREATION_TIMEOUT_SECONDS} as timeout.
+   *
+   * <p>All exceptions, including timeout exception, are rethrown as unchecked exception. To reduce
+   * boilerplate (exception handling), but make sure startup fails.
+   *
+   * @param futures futures that be joined on
+   */
+  private void joinOnFutures(final CompletableFuture<?>[] futures) {
+    try {
+      CompletableFuture.allOf(futures).get(INDEX_CREATION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    } catch (final Exception e) {
+      LangUtil.rethrowUnchecked(e);
+    }
   }
 
   public void updateSchemaMappings(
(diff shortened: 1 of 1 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -85,13 +86,14 @@ private void initialiseIndices() {
...
-        searchEngineClient.getMappings(allIndexNames(), MappingSource.INDEX).keySet();
+        Collections.synchronizedSet(
+            searchEngineClient.getMappings(allIndexNames(), MappingSource.INDEX).keySet());
 
     LOG.info(
         "Found '{}' existing indices. Create missing index templates based on '{}' descriptors.",
         existingIndexNames.size(),
         indexTemplateDescriptors.size());
-    indexDescriptors.stream()
+    indexDescriptors.parallelStream()
```
entangled90, line 96: :wrench: `.parallelStream()` will spawn tasks in the static `ForkJoinPool.commonPool`. In general ForkJoinPool should not be used for blocking tasks, but:
- the max parallelism by default is 256 (which is enough for these tasks and to not block others using it)
- it's done only at startup (if needed)

Do you know if we use already this pool? If we are not then it will be allocated here and then probably stay alive. 

If  we are not using one already, we can create a local executor that we close the schema is created.

ChrisKujawa: Thanks @entangled90 for the input! I knew there was something not optimal with the fork-join pool, but didn't remember what it was. But anyhow it was mostly more a POC whether it helps :) 


We could also think about virtual threads maybe? 🤔 We have already an executor created here for the incident notification https://github.com/camunda/camunda/blob/main/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/DefaultExporterResourceProvider.java#L152

entangled90: I was under the impression that until this [JEP](https://openjdk.org/jeps/491) lands it's a bit risky to use virtual threads, but i'm not up to date on those decisions.

Anyway for this task it should be ok, even if they lock the native thread it's still a better parallelism than 1 :-D

ChrisKujawa: Ok let me try this

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -124,20 +130,31 @@ private void initialiseIndices() {
...
+                          LOG.info("Create missing index '{}'", descriptor.getFullQualifiedName());
+                          searchEngineClient.createIndex(
+                              descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()));
+                        },
+                        virtualThreadExecutor))
+            .toArray(CompletableFuture[]::new);
+
+    // We need to wait for the completion, to make sure all indices has been created successfully
+    // Doing this in parallel is still speeding up the bootstrap time
+    CompletableFuture.allOf(futures).join();
```
entangled90, line 157: :wrench: I think it's better to always put a timeout in these occasion, probably in the order of 1 minute or more. Just to not have something waiting indefinitely, but just throw and exception and maybe restart the pod?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -147,28 +164,43 @@ private void initialiseIndexTemplates() {
...
+                          searchEngineClient.createIndex(
+                              descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()));
+                        },
+                        virtualThreadExecutor))
+            .toArray(CompletableFuture[]::new);
+
+    // We need to wait for the completion, to make sure all indices and templates have been created
+    // successfully
+    // Doing this in parallel is still speeding up the bootstrap time
+    CompletableFuture.allOf(futures).join();
```
entangled90, line 203: :wrench: same as above

entangled90 (approved):
:rocket:

entangled90:
If [Structured concurrency](https://openjdk.org/jeps/453) would not have been in Preview, it could have a been a good place to use it :+1:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
ChrisKujawa: Oh yes, I somehow oversaw this. Thanks for spotting!

ChrisKujawa:
> If [Structured concurrency](https://openjdk.org/jeps/453) would not have been in Preview, it could have a been a good place to use it 👍

Yeah that looks interesting. Actually I would like to have more something like C# has https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/

PR #28437, opened 2025-02-19
ci: c8run build camunda from source

### LINKED ISSUE #28187: I can run C8Run with Basic Auth enabled for WebApps and an unprotected API

megglos:
## Description

Builds c8run based of the current revision of camunda 8, building the distribution once to be used further down in the actual c8run jobs of the c8run build workflow - the price is 5 minutes overhead which I consider worth having actual accurate feedback on the integration state of c8run and the rest of the mono-repo.

This way we ensure integration of c8run development state with the current development state of camunda 8, enabling us to adapt to potential breaking changes asap.

This already contains two fixes that were needed to get the c8run CI working with the current state of camunda 8:
* fa5a9cfb4c8582a63774adbb2a227f5e34b01d33 which removes the setting`spring.config.location`  - It actually breaks the single app, as it leads to internal defaults not being loaded.
See https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.files. - I suspect this issue should already be present in the last alpha release, I wonder why it didn't surface during release already
* 4247d376f22c3f61e73cf98de185e35c7bf425ec adopts to the new auth setup including test adaptation superseding https://github.com/camunda/camunda/pull/27391

## Checklist

- [ ] for CI changes:

## Related issues

closes #28187

### DIFF
--- a/.github/workflows/c8run-build.yaml
+++ b/.github/workflows/c8run-build.yaml
@@ -47,6 +47,36 @@ jobs:
         with:
           working-directory: ./c8run
 
+  camunda-dist-build:
+    name: Build camunda-dist
+    runs-on: gcp-perf-core-16-default
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: ./.github/actions/setup-build
+        with:
+          dockerhub-readonly: true
+          vault-address: ${{ secrets.VAULT_ADDR }}
+          vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
+          vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
+
+      - name: Package camunda-dist
+        shell: bash
+        id: build-dist
+        run: |
+          ./mvnw -B -T1C -DskipTests -DskipChecks -Dflatten.skip=true package
+          export ARTIFACT=$(./mvnw -pl dist/ help:evaluate -Dexpression=project.build.finalName -q -DforceStdout)
+          echo "distball=dist/target/${ARTIFACT}.tar.gz" >> $GITHUB_OUTPUT
+
+      - name: Upload camunda-dist
+        uses: actions/upload-artifact@v4
+        with:
+          name: camunda-platform-dist
+          path: ${{ steps.build-dist.outputs.distball }}
+
   test_c8run:
     strategy:
       matrix:
@@ -55,12 +85,19 @@ jobs:
     name: C8Run Test ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
+    needs: camunda-dist-build
     steps:
       - uses: actions/checkout@v4
 
       - uses: actions/setup-node@v4
         with:
           node-version: 18
+
+      - name: Download Camunda Core Dist
+        uses: actions/download-artifact@v4
+        with:
+          name: camunda-platform-dist
+          path: c8run
 
       - name: Setup C8Run
         uses: ./.github/actions/setup-c8run
@@ -70,6 +107,8 @@ jobs:
           vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
           os: ${{ matrix.os }}
           github-token: ${{ github.token }}
+          local-archive-build: 'true'
+          checkout: 'false'
 
       - name: Install dependencies
         run: npm ci
@@ -109,6 +148,7 @@ jobs:
     name: C8Run Test Windows
     runs-on: windows-latest
     timeout-minutes: 15
+    needs: camunda-dist-build
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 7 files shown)


.github/workflows/c8run-build.yaml
```diff
@@ -109,6 +148,7 @@ jobs:
     name: C8Run Test Windows
```
megglos, line 148: @jessesimpson36 @hisImminence I wonder why the windows step is not reusing the setup-c8run action, the action actually already has platform specific logic baked in, right?

megglos:
and there is more, now it fails with
```
      io.camunda.application - Failed to start application with message: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: liquibase.exception.CommandExecutionException: liquibase.exception.ChangeLogParseException: classpath:/db/changelog/db.changelog-master.yaml does not exist
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: liquibase.exception.CommandExecutionException: liquibase.exception.ChangeLogParseException: classpath:/db/changelog/db.changelog-master.yaml does not exist
```

megglos:
both encountered issues are fixed, see main description

c8run/main.go
```diff
@@ -502,13 +502,7 @@ func main() {
 		if err != nil {
 			fmt.Print("Failed to write to file: " + connectorsPidPath + " continuing...")
 		}
-		var extraArgs string
```
megglos, line 505: @jessesimpson36 we removed this as this =broke the camunda 8 setup, as it bypassed all the internal default configs see https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.files.

hisImminence:
Thanks @megglos ! After @jessesimpson36s review I am fine with merging this. 

As discussed:

Afterwards we should still have a gha that runs this CI nightly with a slack notification informing in [#team-distribution-alerts](https://camunda.slack.com/archives/C0796EZJHQD). (longterm we can route to a core-engneering alerting channel if that makes more sense).

megglos:
Comparing runtime of the old setup:

https://github.com/camunda/camunda/actions/runs/13430384351 => 
Total duration [10m 55s]

vs. new with source build:
https://github.com/camunda/camunda/actions/runs/13433997083/job/37532369555?pr=28437
Total duration: [15m 24s]

Generally the windows step seems to take most of the time, 10 minutes alone vs. 5 for e.g. linux.
So reducing build could be achieved by removing the windows step and e.g. putting it into a nightly.
when it's gone we would actually be back to ~10 minutes for the whole CI.

megglos:
Failing Operate CI originates from `main` and is not relevant for this PR. The whole CI just run because this PR touches the `.githunb` folder, only relevant CI is the c8run CI https://github.com/camunda/camunda/actions/runs/13433997083?pr=28437

.github/workflows/c8run-build.yaml
jessesimpson36: it was either that compiling the camunda core did not work on windows, or one of the actions (setup-build or build-zeebe) was not built for running on windows nodes. I don't recall the exact error.

c8run/main.go
megglos: I found older docs from spring where they kept the old config, it might have been caused by a recent update then.

megglos: the new equivalent may be `spring.config.additional-location`
>If you prefer to add additional locations, rather than replacing them, you can use spring.config.additional-location. Properties loaded from additional locations can override those in the default locations.

megglos: Would you prefer to migrate to the new property, as removing it would be a regression? Happy to test that then.

jessesimpson36: I would like that, but it doesn't need to be part of this PR if you'd rather merge this first. the `--config` option is pretty hard to test (at least on windows)

megglos: Yeah, happy to assist on this in a follow-up if you would add it back, would keep it out of scope here under the assumption this is an acceptable regression from your perspective.

jessesimpson36: This is an acceptable regression. :)

Ben-Sheppard (approved):
I'm approving based on this being what I would hope for, I would also wait for @jessesimpson36's input from a broader C8Run approach :)

jessesimpson36:
I think we're good to merge this.  followup tasks I see are:
1. fix the --config option
2. adjust when the c8run-build action runs (on schedule, and if theres any updates to the c8run-build workflow)
3. have some notification mechanism for when the nightly runs fails

I'm personally fine if these followup tasks are done in a separate PR. Can even be done by immi or I .

jessesimpson36 (approved):

hisImminence:
> I think we're good to merge this. followup tasks I see are:
> 
> 1. fix the --config option
> 2. adjust when the c8run-build action runs (on schedule, and if theres any updates to the c8run-build workflow)
> 3. have some notification mechanism for when the nightly runs fails
> 
> I'm personally fine if these followup tasks are done in a separate PR. Can even be done by immi or I .

Thanks for the summary - created 2 follow up tickets:
1. config option: https://github.com/camunda/camunda/issues/28573
2. nigthly run: https://github.com/camunda/camunda/issues/28574

PR #20723, opened 2024-08-02
refactor: remove webhooks from optimize

buccarel:
## Description

Webhooks are a feature that is present in Optimize 7 and shall be removed from Optimize 8

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

* https://github.com/camunda/camunda/issues/33560

PR #28453, opened 2025-02-20
refactor: remove webhooks from optimize DB

abremard:
## Description

I'm resuming the work being done in [this PR](https://github.com/camunda/camunda/pull/20723) which attempted to remove webhooks (C7 feature) from C8. I broke down the PR into a few smaller pieces to make it easier to review and roll out the changes. This one specifically removes the alert webhooks logic from database.

Also worth noting that this will require a migration (see [comment](https://github.com/camunda/camunda/pull/20723#pullrequestreview-2218254284))

## Related issues

related to https://github.com/camunda/camunda-optimize/issues/13439


abremard:
@RomanJRW Before this PR can be ready to be reviewed, I need to ensure the migration is done for the `AlertIndex`, see [comment](https://github.com/camunda/camunda/pull/20723#pullrequestreview-2218254284). I'm seeing this [8.8.0 DB change log file](https://github.com/camunda/camunda/blob/main/db/rdbms-schema/src/main/resources/db/changelog/rdbms-exporter/changesets/8.8.0.xml) but it looks like it's only RDBMS, is there an equivalent migration file I should be writing updates to?

matthewBearCamunda (approved):

optimize/util/optimize-commons/src/main/java/io/camunda/optimize/dto/optimize/query/alert/AlertCreationRequestDto.java
```diff
@@ -20,7 +20,6 @@ public class AlertCreationRequestDto {
   private boolean fixNotification;
   private AlertInterval reminder;
   private List<String> emails = new ArrayList<>();
-  private String webhook;
```
RomanJRW, line 23: is this an API change? If so, please ensure the docs get updated too

abremard: It's changing the `ui-configuration` API endpoint, which IIRC is only used by our UI. I don't see any reference to that endpoint in `camunda-docs`

PR #21131, opened 2024-08-14
refactor: remove combined report from Optimize FE

OmranAbazid:
## Description

This PR removes the combined reports functionality from the front-end application of Optimize. 
In scope of this, I also did a small refactoring to decouple `hyperReportRenderer` from `CombinedReportRenderer`

## Checklist

- [ ] for CI changes:
  - [ ] structural/foundational changes signed off by [CI DRI](https://github.com/cmur2)
  - [ ] [ci.yml](https://github.com/camunda/camunda/blob/main/.github/workflows/ci.yml) modifications comply with ["Unified CI" requirements](https://github.com/camunda/camunda/wiki/CI-&-Automation#workflow-inclusion-criteria)

## Related issues

related to https://github.com/camunda/camunda-optimize/issues/13439

PR #28456, opened 2025-02-20
refactor: remove combined and reportType from report payload

abremard:
## Description

This PR is removing the "combined" and "reportType" fields as we don't support combined reports in C8 (see [this PR](https://github.com/camunda/camunda/pull/21131) for context).

## Related issues

related to https://github.com/camunda/camunda-optimize/issues/13439

### DIFF
--- a/optimize/client/src/modules/services/reportService.ts
+++ b/optimize/client/src/modules/services/reportService.ts
@@ -7,7 +7,7 @@
  */
 
 import {post} from 'request';
-import {Report, ReportType} from 'types';
+import {Report} from 'types';
 import { getFullURL } from '../api';
 
 interface ConfigParams {
@@ -43,13 +43,13 @@ type DeepPartial<T> = T extends object
     }
   : T;
 
-export type ReportEvaluationPayload<T extends ReportType> = DeepPartial<Report<T>>;
+export type ReportEvaluationPayload = DeepPartial<Report>;
 
-export async function evaluateReport<T extends ReportType>(
-  payload: ReportEvaluationPayload<T>,
+export async function evaluateReport(
+  payload: ReportEvaluationPayload,
   filter = [],
   query = {}
-): Promise<Report<T>> {
+): Promise<Report> {
   let response;
 
   if (typeof payload !== 'object') {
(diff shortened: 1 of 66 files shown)


optimize/client/src/modules/services/reportService.ts
```diff
@@ -59,7 +59,7 @@ export async function evaluateReport<T extends ReportType>(
     // evaluate unsaved report
     // we dont want to send report result in payload to prevent exceedeing request size limit
     const {result: _result, ...evaluationPayload} = payload;
-    response = await post(getFullURL(`api/report/evaluate`), evaluationPayload, {query});
+    response = await post(getFullURL(`api/report/evaluate`), {...evaluationPayload, combined: false, reportType: "process"}, {query});
```
abremard, line 62: To prevent the BE from breaking, I'm sending these two dummy key/values for now. "combined" should always be `false` since we don't support combined reports, and `reportType` should always be `process`. Once the backend part of the cleanup is done, we can remove these dummy values.

OmranAbazid: > To prevent the BE from breaking, I'm sending these two dummy key/values for now. "combined" should always be `false` since we don't support combined reports, and `reportType` should always be `process`. Once the backend part of the cleanup is done, we can remove these dummy values.

Do we keep track of this somehow? Do we have an issue for it or can we have mentioned in the backend issue?

OmranAbazid (approved):
pre-approving.
I think some of the files are not being formatting property like the reportService file.
I think you need to use prettier formatter for it to work correctly.

optimize/client/src/modules/services/reportService.ts
abremard: Thanks for raising this, I've added a checkbox in https://github.com/camunda/camunda-optimize/issues/13211 so we don't forget

abremard:
> I think some of the files are not being formatting property like the reportService file. I think you need to use prettier formatter for it to work correctly.

Interesting, shouldn't this [lint-staged](https://github.com/camunda/camunda/blob/2320bcc73a5590e8a4dd80b413f4aa881175fe21/optimize/client/package.json#L59-L67) have fixed it automatically?

I just manually ran a `prettier --write .`

Issue #27883, opened 2025-02-10
Backport StandaloneSchemaManager to 8.6/.7


berndruecker:
BAckport the Schema Manager from https://github.com/camunda/camunda/issues/25922 to 8.6
Because of architecture streamlining, the code is different in 8.6. 
I pushed code that works on 8.6.8 on this branch: https://github.com/camunda/camunda/compare/8.6.8...25922-standalone-schema-manager-8.6.8

Todo
- [x] Merge for 8.6.9
- [x] Push to 8.7 branch to avoid a gap there (in 8.8.alpha1 it is included via #25922 already)

Usage:

Run the `schema.bat`/`schema.sh` and provide an `application.yaml`, I tested with:

```yaml
# Zeebe Standalone Broker configuration file (with embedded gateway)

# This file is based on broker.standalone.yaml.template but stripped down to contain only a limited
# set of configuration options. These are a good starting point to get to know Zeebe.
# For advanced configuration options, have a look at the templates in this folder.

# !!! Note that this configuration is not suitable for running a standalone gateway. !!!
# If you want to run a standalone gateway node, please have a look at gateway.yaml.template

# ----------------------------------------------------
# Byte sizes
# For buffers and others must be specified as strings and follow the following
# format: "10U" where U (unit) must be replaced with KB = Kilobytes, MB = Megabytes or GB = Gigabytes.
# If unit is omitted then the default unit is simply bytes.
# Example:
# sendBufferSize = "16MB" (creates a buffer of 16 Megabytes)
#
# Time units
# Timeouts, intervals, and the likes, must be specified either in the standard ISO-8601 format used
# by java.time.Duration, or as strings with the following format: "VU", where:
#   - V is a numerical value (e.g. 1, 5, 10, etc.)
#   - U is the unit, one of: ms = Millis, s = Seconds, m = Minutes, or h = Hours
#
# Paths:
# Relative paths are resolved relative to the installation directory of the broker.
# ----------------------------------------------------

zeebe:
  broker:
    gateway:
      # Enable the embedded gateway to start on broker startup.
      # This setting can also be overridden using the environment variable ZEEBE_BROKER_GATEWAY_ENABLE.
      enable: true

      network:
        # Sets the port the embedded gateway binds to.
        # This setting can also be overridden using the environment variable ZEEBE_BROKER_GATEWAY_NETWORK_PORT.
        port: 26500

      security:
        # Enables TLS authentication between clients and the gateway
[...]

PR #28460, opened 2025-02-20
Add standalone application to create schema

rodrigo-lourenco-lopes:
This implements a `StandaloneSchemaManager` that can create the Elasticsearch schema for Operate, Tasklist and Zeebe as well as the ILM policies.
The regular applications are extended with new configuration options to _avoid_ creating schemas and policies.

## Related issues

related to #27883

### DIFF
--- a/dist/src/main/config/application.yaml
+++ b/dist/src/main/config/application.yaml
@@ -81,6 +81,18 @@ zeebe:
       # Controls the number of io threads to be used.
       # This setting can also be overridden using the environment variable ZEEBE_BROKER_THREADS_IOTHREADCOUNT
       ioThreadCount: 2
+	  
+	exporters:
+      # Important to export data into Operate and Tasklist, needs the elasticsearch (or opensearch) endpoint configured
+	  elasticsearch:
+	    className: io.camunda.zeebe.exporter.ElasticsearchExporter
+	    args:
+	      url: http://localhost:9200
+	      #authentication:
+	      #   username: user
+	      #   password: pass
+	      index:
+	        prefix: zeebe-record
 
 camunda:
   database:
--- a/dist/src/main/config/application.yaml
+++ b/dist/src/main/config/application.yaml
@@ -82,18 +82,6 @@ zeebe:
       # This setting can also be overridden using the environment variable ZEEBE_BROKER_THREADS_IOTHREADCOUNT
       ioThreadCount: 2
 
-    exporters:
-      # Important to export data into Operate and Tasklist, needs the elasticsearch (or opensearch) endpoint configured
-      elasticsearch:
-        className: io.camunda.zeebe.exporter.ElasticsearchExporter
-        args:
-          url: http://localhost:9200
-          #authentication:
-          #   username: user
-          #   password: pass
-          index:
-            prefix: zeebe-record
-
 camunda:
   database:
     # Controls which database must be used, possible values: elasticsearch, opensearch
--- a/dist/src/main/config/application.yaml
+++ b/dist/src/main/config/application.yaml
@@ -81,18 +81,18 @@ zeebe:
       # Controls the number of io threads to be used.
       # This setting can also be overridden using the environment variable ZEEBE_BROKER_THREADS_IOTHREADCOUNT
       ioThreadCount: 2
-	  
-	exporters:
+
+    exporters:
       # Important to export data into Operate and Tasklist, needs the elasticsearch (or opensearch) endpoint configured
-	  elasticsearch:
-	    className: io.camunda.zeebe.exporter.ElasticsearchExporter
-	    args:
-	      url: http://localhost:9200
-	      #authentication:
-	      #   username: user
-	      #   password: pass
-	      index:
-	        prefix: zeebe-record
+      elasticsearch:
+        className: io.camunda.zeebe.exporter.ElasticsearchExporter
+        args:
+          url: http://localhost:9200
+          #authentication:
+          #   username: user
+          #   password: pass
+          index:
+            prefix: zeebe-record
 
 camunda:
   database:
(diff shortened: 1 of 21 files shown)


dist/src/main/config/application.yaml
```diff
@@ -82,6 +82,18 @@ zeebe:
       # This setting can also be overridden using the environment variable ZEEBE_BROKER_THREADS_IOTHREADCOUNT
       ioThreadCount: 2
 
+    exporters:
```
grlimacan, line 85: to be deleted

operate/archiver/src/main/java/io/camunda/operate/archiver/ElasticsearchArchiverRepository.java
```diff
@@ -189,28 +193,22 @@ public CompletableFuture<Void> deleteDocuments(
...
       final List<Object> processInstanceKeys) {
-    final var deleteFuture = new CompletableFuture<Void>();
+    final var deleteFuture = new CompletableFuture<Long>();
+    final var deleteRequest =
+        createDeleteByQueryRequestWithDefaults(sourceIndexName)
+            .setQuery(termsQuery(idFieldName, processInstanceKeys))
+            .setMaxRetries(UPDATE_RETRY_COUNT);
 
     final var startTimer = Timer.start();
-    deleteAsyncWithConnectionRelease(
```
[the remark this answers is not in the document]
npepinpe: The method implementation was using the Elasticsearch Task API, which requires cluster level privileges.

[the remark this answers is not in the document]
lenaschoenburg: These are coming from c66ff883fcd9832759030aba54289438797cd45b. They were necessary because we previously used async tasks which require elevated ES permissions. @npepinpe copied the code from Tasklist's archiver, which works without async tasks, into the Operate archiver here. It has nothing to do with releasing connections.

zeebe/exporters/elasticsearch-exporter/src/main/java/io/camunda/zeebe/exporter/ElasticsearchExporter.java
```diff
@@ -70,7 +68,6 @@ public void configure(final Context context) {
     pluginRepository.load(configuration.getInterceptorPlugins());
 
     context.setFilter(new ElasticsearchRecordFilter(configuration));
-    indexTemplatesCreated = false;
```
grlimacan, line 73: This is the root-cause of the failures from the 3 tests in `ElasticsearchExporterIT$IndexSettingsTest`

lenaschoenburg (approved):

operate/common/src/main/java/io/camunda/operate/JacksonConfig.java
```diff
@@ -28,7 +28,7 @@
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
 
-@Configuration
+@Configuration("operateJacksonConfig")
```
grlimacan, line 31: Why is this necessary?

qa/integration-tests/src/test/java/io/camunda/it/StandaloneSchemaManagerTest.java
```diff
@@ -0,0 +1,323 @@
...
+  @BeforeEach
+  void setup() throws IOException, InterruptedException {
+    // setup ES users
+    es.execInContainer("elasticsearch-users", "useradd", APP_USER, "-p", APP_PASSWORD);
+    es.execInContainer("elasticsearch-users", "useradd", ADMIN_USER, "-p", ADMIN_PASSWORD);
+    es.execInContainer("elasticsearch-users", "roles", ADMIN_USER, "-a", ADMIN_ROLE);
+    es.execInContainer("elasticsearch-users", "roles", APP_USER, "-a", APP_ROLE);
+    // Connect to ES
+    schemaManager.withProperty(
+        "zeebe.broker.exporters.elasticsearch.args.url", "http://" + es.getHttpHostAddress());
```
grlimacan, line 156: Small nit: you could save the host address in a variable.

qa/integration-tests/src/test/java/io/camunda/it/StandaloneSchemaManagerTest.java
```diff
@@ -0,0 +1,323 @@
...
+import org.testcontainers.containers.BindMode;
+import org.testcontainers.elasticsearch.ElasticsearchContainer;
+import org.testcontainers.images.builder.Transferable;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.utility.DockerImageName;
+
+@ZeebeIntegration
+@Testcontainers
+final class StandaloneSchemaManagerTest {
```
grlimacan, line 42: Awesome that you manage to cover so much in such short notice, thanks!

qa/integration-tests/src/test/java/io/camunda/it/StandaloneSchemaManagerTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+  public static final String ADMIN_USER = "camunda-admin";
```
grlimacan, line 44: Given our time constraints, only do this if it's really quick, otherwise let's do this as a follow-up: Could you add an additional test-case that verifies that without executing the schema manager beforehand, that running the application with those parameters will fail due to non-existing schema? This would be important to verify that the positive cases you are testing really are running on an application with limited elasticsearch privileges

tasklist/common/src/main/java/io/camunda/tasklist/JacksonConfig.java
```diff
@@ -25,7 +25,7 @@
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
 
-@Configuration
+@Configuration("tasklistJacksonConfig")
```
grlimacan, line 28: Ah ok, now I see why :-) Ignore my comment above

grlimacan (approved):
Great job @lenaschoenburg ! Pre-approving! Just some small nits and a suggestion for an extra test-case

operate/common/src/main/java/io/camunda/operate/JacksonConfig.java
lenaschoenburg: See the commit message in 844faf70584dbd6132131d2bd9f9cef4616bbf37, it's necessary because we want to load both and not giving them a name will result in a conflict that prevents Spring startup.

qa/integration-tests/src/test/java/io/camunda/it/StandaloneSchemaManagerTest.java
lenaschoenburg: I verified this manually now but would skip adding an actual test for two reasons:
1. It's not super straightforward because the `camunda.start()` doesn't fail with a nice exception
2. We'd add ~20 seconds test runtime again

PR #28463, opened 2025-02-20
fix: delete zeebe-record indices to fix flakiness

panagiotisgts:
## Description

CamundaExporterIT was flaky because the `zeebe-record` created indices for the Importers finished checks were not being cleaned up after test cases, causing the `shouldFlushIfImporterNotCompletedButNoZeebeIndices` case to fail

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
+++ b/zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
@@ -129,8 +129,10 @@ final class CamundaExporterIT {
         Optional.ofNullable(System.getProperty(IT_OPENSEARCH_AWS_INSTANCE_URL_PROPERTY)).orElse("");
     if (openSearchAwsInstanceUrl.isEmpty()) {
       searchDB.esClient().indices().delete(req -> req.index(CUSTOM_PREFIX + "*"));
+      searchDB.esClient().indices().delete(req -> req.index("zeebe-record*"));
     }
     searchDB.osClient().indices().delete(req -> req.index(CUSTOM_PREFIX + "*"));
+    searchDB.osClient().indices().delete(req -> req.index("zeebe-record*"));
   }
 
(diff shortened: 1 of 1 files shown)


EuroLew (approved):
Nice catch my bad here

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/CamundaExporterIT.java
```diff
@@ -129,8 +129,10 @@ public void afterEach() throws IOException {
         Optional.ofNullable(System.getProperty(IT_OPENSEARCH_AWS_INSTANCE_URL_PROPERTY)).orElse("");
     if (openSearchAwsInstanceUrl.isEmpty()) {
       searchDB.esClient().indices().delete(req -> req.index(CUSTOM_PREFIX + "*"));
+      searchDB.esClient().indices().delete(req -> req.index("zeebe-record*"));
```
ChrisKujawa, line 132: Can someone point me where we creating these indices?

mustafadagher: they're added in the `shouldNotFlushIfImportersAreNotCompletedAndThereAreZeebeIndices` and `shouldFailIfWaitingForImportersAndCachedRecordsCountReachesBulkSize`

🔧 And in my opinion, they should be removed at the end of these methods only not in the `afterEach()`

panagiotisgts: We could also prefix these indices with the `CUSTOM_PREFIX` but that would require a bit more changes, that way they would be cleaned up by the same query. I can possibly look into this after we unblock the CI

panagiotisgts:
@mustafadagher I've applied the change and it seems to be OK, mind checking this PR again?

ChrisKujawa (approved):

PR #28465, opened 2025-02-20
Opensearch support for prefix migration

### LINKED ISSUE #28155: Opensearch support for prefix migration script
The prefix migration java application should support opensearch this involves implementing the empty implementations of `OpensearchPrefixMigrationClient.java`.

- `OpensearchPrefixMigrationClient.reindex` should take a source and target as strings and perform a [reindex](https://opensearch.org/docs/latest/im-plugin/reindex-data/)
- `OpensearchPrefixMigrationClient.clone` should take a source and target as strings and perform a [clone](https://opensearch.org/docs/latest/api-reference/index-apis/clone/)
- `OpensearchPrefixMigrationClient. getAllHistoricIndices` - should take a string prefix and return all historical index names with that prefix.
- Either extend `PrefixMigrationIT` or write a new test that ensures prefix migration with opensearch works.

EuroLew:
## Description

Prefix migration support for opensearch.

## Related issues

closes #28155


ChrisKujawa (commented):
@EuroLew might make to sync with @panagiotisgts and his PR https://github.com/camunda/camunda/pull/28471 as this might cause conflicts

ChrisKujawa:
@EuroLew can you rebase?

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationIT.java
```diff
@@ -139,12 +106,12 @@ void shouldReindexDocumentsDuringPrefixMigration() throws IOException, Interrupt
             .join();
 
     // Wait for documents to be written to indices
-    Thread.sleep(5000);
+    Thread.sleep(10000);
```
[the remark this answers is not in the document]
ChrisKujawa: Still think it would be better to use awaitility.

qa/integration-tests/src/test/java/io/camunda/it/migration/util/PrefixMigrationHelper.java
```diff
@@ -0,0 +1,55 @@
...
+
+import io.camunda.client.CamundaClient;
+import io.camunda.client.CredentialsProvider;
+import java.io.IOException;
+import java.net.URI;
+import java.util.Base64;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+
+public final class PrefixMigrationHelper {
```
ChrisKujawa, line 18: ❌ This is super confusing. We have now two classes with the same name, in different packages. Can we rename one of them, please.

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationOSIT.java
```diff
@@ -0,0 +1,141 @@
...
+        .withNetwork(NETWORK)
+        .withNetworkAliases("camunda")
+        .withExposedPorts(SERVER_PORT, MANAGEMENT_PORT, GATEWAY_GRPC_PORT)
+        .withEnv("CAMUNDA_DATABASE_TYPE", "opensearch")
+        .withEnv("CAMUNDA_OPERATE_DATABASE", "opensearch")
+        .withEnv("CAMUNDA_TASKLIST_DATABASE", "opensearch")
+        .withEnv("CAMUNDA_OPERATE_CSRFPREVENTIONENABLED", "false")
+        .withEnv("CAMUNDA_TASKLIST_CSRFPREVENTIONENABLED", "false")
+        .withEnv("SPRING_PROFILES_ACTIVE", "broker,operate,tasklist,identity,consolidated-auth")
+        .withEnv("CAMUNDA_SECURITY_AUTHENTICATION_METHOD", "BASIC")
```
ChrisKujawa, line 66: ❓ Couldn't we simply disable the auth?

ChrisKujawa (changes requested):
Thanks @EuroLew there are some requests for change we need to go over, before merging.

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationOSIT.java
EuroLew: I don't know how to do that, I tried asking for help and this is what @Ben-Sheppard and I could get working

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationIT.java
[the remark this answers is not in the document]
EuroLew: will use the new rebased test fix as this is flaky

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/opensearch/OpensearchPrefixMigrationClient.java
```diff
@@ -22,17 +33,70 @@ public OpensearchPrefixMigrationClient(final OpenSearchClient client) {
...
+  }
+
+  private void markIndexReadOnly(final String index) throws IOException {
+    client
+        .indices()
+        .putSettings(r -> r.index(index).settings(s -> s.index(i -> i.blocks(b -> b.write(true)))));
+  }
+
+  private void cloneIndex(final String src, final String target) throws IOException {
+    final var targetAlias = target.substring(0, target.length() - 10) + "alias";
```
panagiotisgts, line 87: :wrench:  This `10` magic number refers to the dated prefix right? Maybe we can make this a bit prettier and clearer like last index of `_` or regex pattern replace for the suffix , since the date regex is already present in this context. 

I also noticed this is present on the Elasticsearch side of things

panagiotisgts (commented):
Overall looks good to me, one comment left for improvement. We could move these tests in the new Migration test framework when it's merged as well  :muscle:

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/opensearch/OpensearchPrefixMigrationClient.java
EuroLew: yes the 10 refers to the yyyy-mm-dd suffix for historic indices.

I'll just capture it with a regex should be better

qa/integration-tests/src/test/java/io/camunda/it/migration/util/PrefixMigrationITUtils.java
```diff
@@ -0,0 +1,119 @@
...
+          String.format("%sapi/login?username=%s&password=%s", endpoint, "demo", "demo"),
+          null);
+
+      return sendPOSTRequest(
+          httpClient,
+          String.format("%sv1/process-instances/search", endpoint),
+          String.format(
+              "{\"filter\":{\"key\":%d},\"sort\":[{\"field\":\"endDate\",\"order\":\"ASC\"}],\"size\":20}",
+              processInstanceKey));
+    }
```
ChrisKujawa, line 99: 🔧 / ❌  @EuroLew I think we can replace this completely with the usage of the TestRestOperateClient which is now part of the testSimpleCamundaApplication. Wdyt?

I don't see it necessarily as a blocker for the PR, but I think we should clean this up then.

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationOSIT.java
```diff
@@ -0,0 +1,149 @@
...
+
+@Testcontainers
+public class PrefixMigrationOSIT {
+
+  @Container
+  private final OpensearchContainer osContainer =
+      TestSearchContainers.createDefaultOpensearchContainer()
+          .withNetwork(NETWORK)
+          .withNetworkAliases(OPENSEARCH_ALIAS);
+
```
ChrisKujawa, line 50: 🔧 💭  I was wondering whether we may simply make use of the @MultiDbTestApplication (where we disable the starting of the application, to better control it).

Part of https://github.com/camunda/camunda/pull/29636

With that we can have the container first doing things. We connect in our test always to `localhost:9200` (we can get the URL from the extension as well), so this should also work for the PrefixMigration App. 

After running the migration APP, we can start the TestSimpleApplication.

The only problem is distinguishing between OS and ES for the container configs. We could get this if the MultiDbExtension would also support containers, I guess 🤔 But yeah maybe this should be a follow-up.

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationOSIT.java
```diff
@@ -0,0 +1,149 @@
...
+        .withEnv("CAMUNDA_OPERATE_OPENSEARCH_INDEX_PREFIX", OLD_OPERATE_PREFIX)
+        .withEnv("CAMUNDA_TASKLIST_OPENSEARCH_INDEX_PREFIX", OLD_TASKLIST_PREFIX)
+        .withEnv("CAMUNDA_TASKLIST_OPENSEARCH_INDEXPREFIX", OLD_TASKLIST_PREFIX)
+        .withEnv(
+            "ZEEBE_BROKER_EXPORTERS_OPENSEARCH_CLASSNAME",
+            "io.camunda.zeebe.exporter.opensearch.OpensearchExporter")
+        .withEnv("CAMUNDA_OPERATE_OPENSEARCH_URL", osUrl)
+        .withEnv("CAMUNDA_OPERATE_ZEEBEOPENSEARCH_URL", osUrl)
+        .withEnv("CAMUNDA_TASKLIST_OPENSEARCH_URL", osUrl)
+        .withEnv("CAMUNDA_TASKLIST_ZEEBEOPENSEARCH_URL", osUrl)
```
ChrisKujawa, line 80: ❓ Don't we need to configure the components for the credentials OS expects? AFAIK Opensearch requires user and password?

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationIT.java
```diff
@@ -220,28 +145,15 @@ void shouldReindexDocumentsDuringPrefixMigration() throws IOException, Interrupt
     camunda87.stop();
 
     // when
-    prefixMigration(NEW_PREFIX);
+    prefixMigration();
 
     // then
-    final var currentCamundaClient = startLatestCamunda();
+    final var currentCamundaClient =
+        startLatestCamunda(esContainer.getHttpHostAddress(), NEW_PREFIX, true);
```
ChrisKujawa, line 152: 🔧 Should we also use here `try (...) {` with resources>?

ChrisKujawa (approved):
Thanks @EuroLew 👍🏼 I had some smaller comments, but looks good!

There was one bigger topic, regarding using the new multi db extension, but yeah I guess let's get this merged and iterate from there.

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationOSIT.java
EuroLew: Seems like quite a bit of work but would be worth it for sure

qa/integration-tests/src/test/java/io/camunda/it/migration/util/PrefixMigrationITUtils.java
EuroLew: This will be a follow up, can't quite get it working right now

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationOSIT.java
EuroLew: Looking at the logs, it seems that, basic authentication is disabled for opensearch so we can just connect with no username/password.

image

qa/integration-tests/src/test/java/io/camunda/it/migration/PrefixMigrationOSIT.java
EuroLew: https://github.com/camunda/camunda/issues/29814

qa/integration-tests/src/test/java/io/camunda/it/migration/util/PrefixMigrationITUtils.java
EuroLew: https://github.com/camunda/camunda/issues/29814

PR #28472, opened 2025-02-20
fix: rdbms - use rule index as additional PK for decision outputs

p-wunderlich:
## Description

Fixes the "Unique index or primary key violation: `PUBLIC.PRIMARY_KEY_F ON PUBLIC.DECISION_INSTANCE_OUTPUT(DECISION_INSTANCE_ID, OUTPUT_ID) VALUES ( /* key:3 */ '2251799813685275-2', 'OutputClause_1cthd0w')` problem in rdbms module. 

Since DECISION_INSTANCE_ID and OUTPUT_ID ist not unique as PK, I extended the PK to also use the RULE_INDEX.


ChrisKujawa (approved):

ChrisKujawa:
💭 I guess we could remove the output id from the PK 🤔

p-wunderlich:
@ChrisKujawa nope don't think so, since you can also have multiple columns.

stoerti (approved):

Issue #28301, opened 2025-02-18
Reintroduce User Task Restrictions by candidateGroup 


nathansandi:
### Description

Reintroduce the User Task Restriction into Tasklist V1

This was disabled in (https://github.com/camunda/camunda/pull/28134/files) 

### Acceptance Criteria:

- [ ] all disabled tests in [TaskControllerTest.java](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/test/java/io/camunda/tasklist/webapp/api/rest/v1/controllers/TaskControllerTest.java) are enabled again
- [ ] User Task restriction properly working with Camunda User Service retrieving the groups

Issue #28122, opened 2025-02-13
Retrieval of Groups IdentityAuthorizationService


nathansandi:
## Description

The current Identity Authorization Service is Stub,
The goal of this issue is use CamundaUserService in order to collect the current user and return the groups, once the old implementation is still using the identity.

[IdentityAuthorizationService](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/IdentityAuthorizationService.java#L13)

## Acceptance Criteria

- [ ] Tasklist should point to IdentityAuthorizationService instead of Stub Configuration
- [ ] Use CamundaService to retrieve the user groups instead of identity

PR #28505, opened 2025-02-20
feat: implement get groups in identity auth service and user task restrictions

nathansandi:
## Description

This pull request updates Tasklist V1 to support consolidated-auth for group retrieval using Camunda User Service (and replace group retrieval using Identity SDK), ensuring compatibility regardless of the authentication method (OIDC or Basic). 

Additionally, it reintroduces User Task Restrictions for group-based access control, independent of the authentication mechanism. In case it is not consolidated-auth, the group retrieval will give full access. 

## Related issues

closes #https://github.com/camunda/camunda/issues/28301
clsoes #https://github.com/camunda/camunda/issues/28122


dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
```diff
@@ -55,13 +61,47 @@ public class TasklistModuleConfiguration {
...
+   * the profile "consolidated-auth" is active.
+   *
+   * <p>This bean provides the **main authorization logic** when consolidated authentication is
+   * enabled.
+   */
+  @Bean
+  @Primary
+  @Profile("consolidated-auth")
+  public IdentityAuthorizationService consolidatedIdentityAuthorizationService() {
+    return new IdentityAuthorizationServiceImpl();
```
megglos, line 87: ❌  this class and also the other impl are already annotated as `@Component` so they will be loaded via class path scanning already, right? Would change it then to avoid duplicates, my spring knowledge might also just be rusty though 😅

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
```diff
@@ -55,13 +61,47 @@ public class TasklistModuleConfiguration {
...
+   *
+   * <p>- This bean ensures that authorization is always available, even when consolidated
+   * authentication is not enabled. - It provides a **default, full-access authorization service**.
+   *
+   * <p>This bean declaration can be removed after the consolidated authentication is fully
+   * implemented
+   */
+  @Bean
+  public IdentityAuthorizationService defaultIdentityAuthorizationService() {
+    return new DefaultIdentityAuthorizationServiceImpl();
```
megglos, line 102: 🔧  same here, this is already declared as `@Component`

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
megglos: it should get picked up by line 36
```
      @ComponentScan.Filter(
          type = FilterType.REGEX,
          pattern = "io\\.camunda\\.tasklist\\.webapp\\..*"),
```

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/DefaultIdentityAuthorizationServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package io.camunda.tasklist.webapp.security.identity;
+
+import static io.camunda.tasklist.property.IdentityProperties.FULL_GROUP_ACCESS;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.springframework.stereotype.Component;
+
+@Component
```
megglos, line 16: ❌  let's make sure it's not loaded with the new identity setup
```suggestion
@Component
@Profile("!consolidated-auth")
```

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/api/rest/v1/controllers/TaskController.java
```diff
@@ -230,7 +230,25 @@ private void checkTaskImplementation(final LazySupplier<TaskDTO> taskSupplier) {
...
+    final boolean allUsersTask =
+        task.getCandidateUsers() == null && task.getCandidateGroups() == null;
+    final boolean candidateGroupTasks =
+        task.getCandidateGroups() != null
+            && !Collections.disjoint(Arrays.asList(task.getCandidateGroups()), listOfUserGroups);
+    final boolean candidateUserTasks =
+        task.getCandidateUsers() != null
+            && Arrays.asList(task.getCandidateUsers()).contains(userName);
+    final boolean assigneeTasks = task.getAssignee() != null && task.getAssignee().equals(userName);
+    final boolean noRestrictions = listOfUserGroups.contains(IdentityProperties.FULL_GROUP_ACCESS);
```
megglos, line 245: 🔧  this is actually not supported with the new group membership setup 🤔  could we remove it already alongside relevant tests?

megglos (changes requested):
Good job on adapting to the new service!

I have some concerns on having the services declared as beans and components at the same, thus requesting changes for us to align. Happy to do that in sync if you think that might be more effective.

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/IdentityAuthorizationServiceImpl.java
```diff
@@ -7,52 +7,19 @@
...
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.security.core.Authentication;
-import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Component;
 
 @Component
+@Profile("consolidated-auth")
 public class IdentityAuthorizationServiceImpl implements IdentityAuthorizationService {
```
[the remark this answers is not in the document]
nathansandi: Great opportunity for do that :), going to rename it!

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/api/rest/v1/controllers/TaskController.java
nathansandi: Lets do a quick sync on this one. 
From my understanding V1 should still support Task Restrictions, but lets have a quick chat to align this.

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
nathansandi: Interesting :) Thanks for point this.

megglos: Sync call: we just found that these are actualyl exclusions, so it's config beans only for tasklist setup and the component annotation is just present for autowire convenience

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
megglos: same as here https://github.com/camunda/camunda/pull/28505/files/2bb71ded0b0d6df1f966b985aab08294acc8ec05#r1969433454

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/api/rest/v1/controllers/TaskController.java
megglos: to calrify, this was about the special value of `FULL_GROUP_ACCESS` granting access to all groups

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/DefaultIdentityAuthorizationServiceImpl.java
megglos: irrelevant because of https://github.com/camunda/camunda/pull/28505/files/2bb71ded0b0d6df1f966b985aab08294acc8ec05#r1969433454

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
```diff
@@ -55,13 +61,47 @@ public class TasklistModuleConfiguration {
...
+   * Fallback bean for `IdentityAuthorizationService`, used when the "consolidated-auth" profile is
+   * **not active**.
+   *
+   * <p>- This bean ensures that authorization is always available, even when consolidated
+   * authentication is not enabled. - It provides a **default, full-access authorization service**.
+   *
+   * <p>This bean declaration can be removed after the consolidated authentication is fully
+   * implemented
+   */
+  @Bean
```
megglos, line 100:
```suggestion
  @Bean
  @Profile("!consolidated-auth")
```

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/api/rest/v1/controllers/TaskController.java
nathansandi: The `FULL_GROUP_ACCESS`  will return an empty string, and it is an special value, used in case of the fallback.
This wont cause issues for consolidated-auth, once it is not a valid value over there. 

When the `DefaultUserGroupService` is no longer necessary, this can be removed, once we can rely only in the `CamundaUserService` Groups

tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/DefaultUserGroupService.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static io.camunda.tasklist.property.IdentityProperties.FULL_GROUP_ACCESS;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+@Component
+@Profile("!consolidated-auth")
```
megglos, line 18: 🔧 Given the setup is based on configuration classes only and component scans are deactives, I think you can omit the profile here, it should actually not have any effect.
```suggestion
```

megglos (approved):
LGTM! Pre-approving as nothing is blocking, please check the comments I left still.

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
nathansandi: Not so far, once I am using the @Primary bean in the other one.

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
houssain-barouni: `io.camunda.tasklist.webapp` is scanned by https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/WebappModuleConfiguration.java

Issue #28302, opened 2025-02-18
Daily notification of flakiest tests in Slack


npepinpe:
## Description

I would like to have an easy way to keep track of the flakiest tests currently affecting the mono repo CI, in order to better focus our effort in providing a stable pipeline, reduce CI costs due to unnecessary builds, increase confidence in a green pipeline, and short feedback loop for developers.

At the moment, I currently check the flaky test dashboard every Monday, and I manually triage which tests are owned by my team, then assign them based on how many flakes where on main (where the bulk of contributions are).

My goal with this issue is to reduce effort and bring awareness to the flaky tests in order to ensure we are dealing with them in a timely manner.

I would like a daily notification to a shared Slack channel (e.g. `#core-ci`), which tells me how many builds total where there in the last 24 hours, how many of these builds had flaky tests, a list of very flaky tests which need attention ASAP, and the top flakiest tests in the last 24 hours, with a link to the latest build where the flake was detected.

I'm not sure yet if we want to just always show the 10 top tests, or the ones with at least X flakes in the last window of time. Please share feedback there. I could also imagine some combination, e.g.

```
📈 62% of builds were flaky in the last 24 hours (62 flaky tests in 100 builds).

👀 Flakiest test which need attention 👀  ( > 50 flakes in 24 hours)

Test             | Count |
-------------------------|
MySuperFlakyTest |    60 |

🥇 Top flakiest test in the last 24 hours

Test             | Count |
-------------------------|
MySuperFlakyTest |    60 |
MyLessFlakyTest  |    05 |
MyLeastFlakyTest |    02 |

```

cmur2:
First iteration is live in #zeebe-ci starting today to gather feedback and catch (usability) problems, before wider rollout

cmur2:
Positive feedback on the first iteration (improves visibility). Will rollout to a wider audience to increase awareness next.

Open feature requests:

* link to existing GH issue if available (or automatically create one)
* link to GHA job with failure (partially covered by timestamped Grafana dashboard link)

PR #28509, opened 2025-02-20
ci: add job posting daily flaky test statistics to Slack

cmur2:
## Description

Experiment with data retrieval from BigQuery to post a first draft of daily flaky test statistics to Slack. The job takes the data from the previous day and runs Tuesday to Friday (since I figured the data from Sunday wouldn't be super interesting, can be made smarter later e.g. to lookback to fridays).

To gather initial feedback it is set up to post to the Zeebe CI Slack channel, to be changed later for wider audience.

image

## Checklist

- [x] for CI changes:
  - [ ] enable backports [when recommended](https://github.com/camunda/camunda/wiki/CI-&-Automation#when-to-backport-ci-changes) - not needed, new workflow mainly run on schedule

## Related issues

Related https://github.com/camunda/camunda/issues/28302

### DIFF
--- a/.github/workflows/statistics-daily.yml
+++ b/.github/workflows/statistics-daily.yml
@@ -1,0 +1,79 @@
+---
+name: Statistics Daily
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: '0 8 * * 2-5'  # Tuesdays to Fridays
+
+jobs:
+  flaky-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions: {}  # GITHUB_TOKEN unused in this job
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+
+      - name: Import Secrets
+        id: secrets
+        uses: hashicorp/vault-action@a1b77a09293a4366e48a5067a86692ac6e94fdc0 # v3.1.0
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          method: approle
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          exportEnv: false # we rely on step outputs, no need for environment variables
+          secrets: |
+            secret/data/products/camunda/ci/github-actions SLACK_ZEEBECI_WEBHOOK_URL;
+            secret/data/products/zeebe/ci/ci-analytics gcloud_sa_key;
+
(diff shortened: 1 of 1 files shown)


.github/workflows/statistics-daily.yml
```diff
@@ -0,0 +1,79 @@
+---
+name: Statistics Daily
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: '0 8 * * 2-5'  # Tuesdays to Fridays
```
npepinpe, line 7: :thinking: Is the idea that we can't do Monday because Sunday there are no builds? I mean, we still have some builds, but :shrug: I think it might OK, but we can also wait.

.github/workflows/statistics-daily.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Print Google Cloud SDK version used
+        shell: bash
+        run: |
+          gcloud info
+
+      - name: Create message with BigQuery data
+        id: message
+        shell: bash
+        run: |
+          NUMBER_OF_ALL_JOBS=$(bq query --format=json --use_legacy_sql=false --quiet=true 'SELECT COUNT(*) as number_of_all_jobs FROM `ci-30-162810.prod_ci_analytics.build_status_v2` WHERE TIMESTAMP_SUB(TIMESTAMP_TRUNC(CURRENT_TIMESTAMP, DAY), INTERVAL 1 DAY)<=report_time AND report_time<=TIMESTAMP_TRUNC(CURRENT_TIMESTAMP, DAY) AND ci_url="https://github.com/camunda/camunda"' | jq -r '.[0].number_of_all_jobs')
```
npepinpe, line 48: :question: This is all jobs, not just those on main, right? That includes all branches and so on, including jobs which may not run code checks and so on?

npepinpe (approved):
:+1: 

Looks good! I think you're right, we may want to iterate on how we could these things to get closer to the actual experience from engineers (e.g. it looks like 0% of jobs are flaky, but we all know this is not true in practice :smile:)

.github/workflows/statistics-daily.yml
cmur2: Yea ideally I would want Monday to look back to Friday (maybe incl Sat+Sun).

.github/workflows/statistics-daily.yml
cmur2: Yes, all jobs even if most do not extract flaky tests. I tried limiting to only "Unified CI" but still it is 1.1k (instead of 1.5k) jobs where there was 69 flaky jobs

PR #28524, opened 2025-02-20
fix: re-add the --config flag but with it being used under spring.config.additional-location

### LINKED ISSUE #28573: Add --config flag again
**Describe the issue:**

To enable the build from source the `--config` option [was removed](https://github.com/camunda/camunda/commit/fa5a9cfb4c8582a63774adbb2a227f5e34b01d33) as with the latest spring version an empty configuration folder overrwrites the default configuration (see [here](https://github.com/camunda/camunda/pull/28437#issuecomment-2670721869) for more details).

**Actual behavior:**

With this, the current versions does not allow config overrides anymore.

**Expected behavior:**

As a user I can add config overrides again by adding a config file to a configuration folder.

**How to reproduce:**

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Architecture: 
- Version:

jessesimpson36:
## Description

This PR adds back the --config flag.  We found that this flag was broken after an upgrade in spring, and was also taken out because of issues it caused overriding all of the internal default configuration.

This PR adds it back but under `spring.config.additional-configuration` that way all the options are appended or overridden but not entirely replaced.

## Checklist

- [ ] for CI changes:

## Related issues

closes https://github.com/camunda/camunda/issues/28573

### DIFF
--- a/c8run/main.go
+++ b/c8run/main.go
@@ -281,6 +281,7 @@ func getBaseCommandSettings(baseCommand string) (C8RunSettings, error) {
 
 func createStartFlagSet(settings *C8RunSettings) *flag.FlagSet {
 	startFlagSet := flag.NewFlagSet("start", flag.ExitOnError)
+	startFlagSet.StringVar(&settings.config, "config", "", "Applies the specified configuration file.")
 	startFlagSet.BoolVar(&settings.detached, "detached", false, "Starts Camunda Run as a detached process")
 	startFlagSet.IntVar(&settings.port, "port", 8080, "Port to run Camunda on")
 	startFlagSet.StringVar(&settings.keystore, "keystore", "", "Provide a JKS filepath to enable TLS")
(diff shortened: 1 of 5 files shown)


jessesimpson36:
https://github.com/camunda/camunda/pull/28437

jessesimpson36:
@megglos , I included a test in api_tests.sh that will grab out the overridden setting to prove that the --config flag works. I would be fine with keeping this test or removing it (since setting the prefix could one day break the CI).

c8run/main.go
```diff
@@ -499,7 +500,13 @@ func main() {
 		if err != nil {
 			fmt.Print("Failed to write to file: " + connectorsPidPath + " continuing...")
 		}
-		camundaCmd := c8.CamundaCmd(camundaVersion, parentDir, "", javaOpts)
+
+                var extraArgs string
+		if settings.config != "" {
+                        extraArgs = "--spring.config.additional-location=file:" + filepath.Join(parentDir, settings.config)
```
megglos, line 506: ❌  there is one thing we need to take care of as well, in case the path points to a directoty a trailing slash must be present, from what I understand `filepath.Join` will always output paths without a trailing slash, right? I can add a test case for this 

See the [spring docs](https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.files)
>If spring.config.location contains directories (as opposed to files), they should end in /.

jessesimpson36: is there a reason that anyone would put a directory?

megglos: it's generally supported by spring, you could organize config across multiple files.
If we don't want to support it, c8run should fail eagerly on dir paths, otherwise the user may only find out why app startup fails by reading the camunda.log.

jessesimpson36: I added code to support it.

c8run/main.go
```diff
@@ -499,7 +500,30 @@ func main() {
 		if err != nil {
 			fmt.Print("Failed to write to file: " + connectorsPidPath + " continuing...")
 		}
-		camundaCmd := c8.CamundaCmd(camundaVersion, parentDir, "", javaOpts)
+
+		var extraArgs string
```
hisImminence, line 504: I would make this a separate function outside of main - but can be done later.

hisImminence (approved):

PR #28555, opened 2025-02-21
fix: NettyDnsMetrics use a MeterProvider instead of a HashMap to avoid concurrency issues

### LINKED ISSUE #28556: NettyDnsMetrics is not safe under concurrent access
`NettyDNSMetrics` uses a HashMapo which is not safe to access on concurrent access as noted in #28192.
Given that is not in the hotpath a `MeterProvider` can be used

entangled90:
## Description
Followup of the comments in #28192:
NettyDnsMetrics is used by multiple threads and the HashMap in use is not concurrent.
Considering the small number of usages of these metrics it's ok to use the `MeterProvider` from micrometer

## Related issues
closes #28556


npepinpe (approved):
We assume it's a small usage because we only send DNS requests seldom and cache the results?

entangled90:
I got the impression from you that we do. 
Checking the codebase I see that we do not override the ttl/negativeTtl of `DnsNameResolverBuilder`. This means that we respect the caching ttls from the DNS server. 
Even though this may be okay in a platform we control or k8s ( yet to be verified) this may pose some problems for SM users.

entangled90:
Anyway the MeterProvider is used only for DNS resolution failures, so it should be not that frequent.

PR #28557, opened 2025-02-21
Create matching index if index template already exists

### LINKED ISSUE #28173: Camunda Exporter is not creating indices for templates, when templates already exist
## Description

We ran into this during implementing the new IT foundation #26896 as we detected that indices are not cleaned up correctly, and later some indices are not re-created.

It turned out that if the templates were not cleaned up as well, the corresponding indices haven't been recreated (even if they were missing). Causing endless loops of waiting for all indices.

In an production environment this might be less likely, but in general I think it makes sense to de-couple this logic to always create the index (and log the failure if it already exists).

Happy to discuss this further @EuroLew

EuroLew:
## Description

It turned out that if the templates were not cleaned up as well, the corresponding indices haven't been recreated (even if they were missing). Causing endless loops of waiting for all indices.

Instead now the schema manager will check for any templates that exist without their corresponding index and create them.

## Related issues

closes #28173

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
@@ -11,6 +11,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.exporter.config.ExporterConfiguration;
 import io.camunda.exporter.config.ExporterConfiguration.IndexSettings;
 import io.camunda.exporter.config.ExporterConfiguration.RetentionConfiguration;
+import io.camunda.exporter.exceptions.ElasticsearchExporterException;
+import io.camunda.exporter.exceptions.OpensearchExporterException;
 import io.camunda.webapps.schema.descriptors.IndexDescriptor;
 import io.camunda.webapps.schema.descriptors.IndexTemplateDescriptor;
 import java.util.Collection;
@@ -92,10 +94,7 @@ public class SchemaManager {
             .getMappings(config.getIndex().getPrefix() + "*", MappingSource.INDEX_TEMPLATE)
             .keySet();
 
-    final var existingIndexNames =
-        allIndexNames().isBlank()
-            ? Set.of()
-            : searchEngineClient.getMappings(allIndexNames(), MappingSource.INDEX).keySet();
+    final var existingIndexNames = existingIndexNames();
 
     indexTemplateDescriptors.stream()
         .filter(desc -> existingTemplateNames.contains(desc.getTemplateName()))
@@ -133,9 +132,7 @@ public class SchemaManager {
       return;
     }
 
-    final var existingIndexNames =
-        Collections.synchronizedSet(
-            searchEngineClient.getMappings(allIndexNames(), MappingSource.INDEX).keySet());
+    final var existingIndexNames = Collections.synchronizedSet(existingIndexNames());
 
     LOG.info(
         "Found '{}' existing indices. Create missing index templates based on '{}' descriptors.",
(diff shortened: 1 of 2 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -155,6 +153,22 @@ private void initialiseIndexTemplates() {
...
+                    && !existingIndexNames.contains(desc.getFullQualifiedName()))
+        .forEach(
+            desc -> {
+              LOG.warn(
+                  "Index template '{}' already exists but corresponding index '{}' does not, this is an anomalous state which should not happen, fixing by creating the index",
+                  desc.getTemplateName(),
+                  desc.getFullQualifiedName());
+
+              searchEngineClient.createIndex(desc, getIndexSettingsFromConfig(desc.getIndexName()));
+            });
```
ChrisKujawa, line 170: ❌ Tbh I think we should simply go over the templates, create them and then go over them again and create the indices again.

This means we don't need to check for existing index 🤔 

```java
    indexTemplateDescriptors.stream()
        .filter(descriptor -> !existingTemplateNames.contains(descriptor.getTemplateName()))
        .forEach(
            descriptor -> {
              LOG.info("Create missing index template '{}'", descriptor.getTemplateName());
              searchEngineClient.createIndexTemplate(
                  descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()), true);
            });

// we simply create all indices - we simply ignore already exist errors
    indexTemplateDescriptors.stream()
        .forEach(
            descriptor -> {
              LOG.info(
                  "Create missing index '{}', for template '{}'",
                  descriptor.getFullQualifiedName(),
                  descriptor.getTemplateName());
              searchEngineClient.createIndex(
                  descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()));
            });

```

ChrisKujawa (commented):
I think we can simply create all without requesting the indices, please correct me if I misunderstood something.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
EuroLew: there's a very annoying bug that occurs if we do this. We only really want to call `createIndexTemplate` when we can be sure it is a creation and not an update. This is a current restriction on the way schema validation works. When we validate index templates, if we remove a mapping field it counts as field deletion but we currently just ignore this as it's not something ES supports (or should support for indices). 
image

Therefore if we call a naked createIndexTemplate then we can actually remove fields from our index template mappings, e.g.

template currently has mappings 
{ foo: string, bar: keyword }

the json changes to be
{ foo:string }

The current schema manager will just log out `Index '{}': Field deletion is requested, will be ignored. Fields: {}",` and nothing happens. The index template still has mappings { foo: string, bar: keyword }

but if we did your solution of recreating them all the time what would happen is the index template would then have mappings of {foo: string} which is field deletion which is something we explicitly ignore.

Now the obvious solution is not to call createIndexTemplate for cases of field deletion but that is more work than this solution as it would require a refactoring of the SchemaValidator.

EuroLew: A solution I would be partial to due to it's simplicity would just be to throw an exception for field deletion being requested by the schema, failing early, and then there's nothing stopping your solution. I'll do that unless you have any objections.

ChrisKujawa (commented):
Hm i think the current approach doesnt work well.

What about:

1. We iterate once over all templates
2. Try to create the template
3. It might fail due to already existing -> we still create index
4. Might succeed then we create index as well
5. Might fail for other reason -> we dont create index

Wdyt @EuroLew

EuroLew:
> Hm i think the current approach doesnt work well.
> 
> What about:
> 
> 1. We iterate once over all templates
> 2. Try to create the template
> 3. It might fail due to already existing -> we still create index
> 4. Might succeed then we create index as well
> 5. Might fail for other reason -> we dont create index
> 
> Wdyt @EuroLew

sounds good to me

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -207,6 +189,34 @@ private void initialiseIndexTemplates() {
...
+          descriptor.getTemplateName());
+
+    } catch (final ElasticsearchExporterException | OpensearchExporterException e) {
+      final var errMsg =
+          String.format("Index template '%s' could not be created", descriptor.getTemplateName());
+      LOG.error(errMsg, e);
+      throw new IllegalStateException(errMsg, e);
+    }
+  }
+
```
ChrisKujawa, line 219: ❓ The creation of the template will fail if it exists right?

❌ I think with that we still don't cover that if the template exists, we create an index.

ChrisKujawa: or do I miss something ? 😅

EuroLew: If the template already exists, we swallow the error 
image

This means 
- already exists template
- make creation call
- error swallowed
- makes call to create index.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -179,26 +176,11 @@ private void initialiseIndexTemplates() {
         indexTemplateDescriptors.size());
     final var futures =
         indexTemplateDescriptors.stream()
-            .filter(descriptor -> !existingTemplateNames.contains(descriptor.getTemplateName()))
```
ChrisKujawa, line 182: :i: This is the actual fix.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -179,26 +176,11 @@ private void initialiseIndexTemplates() {
...
-                              true);
-                          LOG.info(
-                              "Create missing index '{}', for template '{}'",
-                              descriptor.getFullQualifiedName(),
-                              descriptor.getTemplateName());
-                          searchEngineClient.createIndex(
-                              descriptor, getIndexSettingsFromConfig(descriptor.getIndexName()));
-                        },
-                        virtualThreadExecutor))
+                        () -> createIndexTemplate(descriptor), virtualThreadExecutor))
```
ChrisKujawa, line 183: :i: This is a refactoring

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
```diff
@@ -133,9 +132,7 @@ private void initialiseIndices() {
       return;
     }
 
-    final var existingIndexNames =
-        Collections.synchronizedSet(
-            searchEngineClient.getMappings(allIndexNames(), MappingSource.INDEX).keySet());
+    final var existingIndexNames = Collections.synchronizedSet(existingIndexNames());
```
ChrisKujawa, line 135: 🔧 I think we actually can removed this now. It is only used for logging. Feels a bit like a waste.

ChrisKujawa (approved):
Looked at it together with @EuroLew and we discussed that the solution solved it now with removing the filter we had before. Means we no longer check whether the template exists, we always try to create it and ignore if it already exists. The same applies for index creation.

Issue #24551, opened 2024-11-08
ProcessDefinitionQueryTest is flaky


ana-vinogradova-camunda:
**Summary**

- How often does the test fail? Once so far
- Does it block your work? No
- Do we suspect that it is a real failure? Not sure

**Failures**

<details><summary>Example assertion failure</summary>
<pre>
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 198.6 s <<< FAILURE! -- in io.camunda.it.client.ProcessDefinitionQueryTest
Error:  io.camunda.it.client.ProcessDefinitionQueryTest -- Time elapsed: 198.6 s <<< ERROR!
org.awaitility.core.ConditionTimeoutException: 
Condition with alias 'should deploy processes and import in Operate' didn't complete within 3 minutes because assertion condition defined as a Lambda expression in io.camunda.it.client.ProcessDefinitionQueryTest 
expected: 28
 but was: 7.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1006)
	at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:790)
	at io.camunda.it.client.ProcessDefinitionQueryTest.waitForProcessesToBeDeployed(ProcessDefinitionQueryTest.java:441)
	at io.camunda.it.client.ProcessDefinitionQueryTest.beforeAll(ProcessDefinitionQueryTest.java:71)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:194)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)
[...]

PR #28560, opened 2025-02-21
fix: improve test case for process definition deploys

nathansandi:
## Description

The process query test is flaky because it is not deploying the processes in 3 minutes: (As it i deploying 8 processes + 2 forms)

Error:
Condition with alias 'should deploy processes and import in Operate' didn't complete within 3 minutes because assertion condition defined as a Lambda expression in io.camunda.it.client.ProcessDefinitionQueryTest 

This PR covers: 
- Going to reduce the number of processes deployed (there are process deployed that are not really used in test case)
- Increase the time to check if the process are deployed

## Related issues

closes #https://github.com/camunda/camunda/issues/24551


tmetzke (approved):

tmetzke:
Does this exist in 8.6 as well and should be backported there, too?

PR #28567, opened 2025-02-21
feat: updating settings should happen first to avoid errors with deserialisation

EuroLew:
## Description

Settings should get updated before mappings, this is because there are some settings which will cause errors with deserialisation when retrieving index templates https://github.com/elastic/elasticsearch-java/issues/696.


ChrisKujawa (commented):
TIL;

Can we add a test for this somehow?

PR #28585, opened 2025-02-21
Fix bug blocking migration

EuroLew:
## Description

the operate-list-view index template is blocking migration as the one created by previous versions of the schema manager are not compatible with the client used by the current camunda exporter schema manager.

The issue is manifold
1. we must get the index template mappings to validate them.
2. we cannot retrieve just the mappings (indices have a getMapping which does not exist of index templates), this means we must also retrieve index template settings and deserialise them, this causes issues with our client hence blocking migration.
3. we cannot just update the settings of the index template, as the java client is limited in that a put index template request will OVERWRITE everything there is no append functionality. therefore if we wanted to just update the breaking settings we would also need to retrieve the current mappings of the index template, hence we are back at step 2.

Unfortunately this is the simplest fix I can think of, this shouldn't break anything as after deleting the list view index template that breaks migration, we will immediately after recreate it.

If the solution is not ideal, there is still the test which creates a copy of the breaking index template which can be used to check if another solution would work.

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/schema/SchemaManager.java
@@ -49,6 +49,10 @@ public class SchemaManager {
       return;
     }
     LOG.info("Schema creation is enabled. Start Schema management.");
+
+    // remove legacy index template which is not compatible with current schema manager
+    searchEngineClient.removeBreakingIndexTemplates(indexTemplateDescriptors);
+
     final var schemaValidator = new IndexSchemaValidator(objectMapper);
     final var newIndexProperties = validateIndices(schemaValidator);
     final var newIndexTemplateProperties = validateIndexTemplates(schemaValidator);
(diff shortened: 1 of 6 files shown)


ThorbenLindhauer:
@ChrisKujawa I don't fully understand the problem description here (the first paragraph is very vague). But based on what I read, no I haven't heard this before.

panagiotisgts:
> I feel this logic is too coupled/hard coded to the migration of one version. Might make sense to have this in the migration application instead ?

@ChrisKujawa Not really sure it fits there. It's true that this has to be done for a single version upgrade only but I don't believe this is in the scope of the Migration. To my understanding this falls most in the Schema Manager territory. We could _hide_ it in the migration so as not to have to remove it later on but should we really do that?

ChrisKujawa:
I looked a bit more deeply into this topic and created https://github.com/camunda/camunda/issues/29132

I think it is critical as it completely blocks our migration/update on ES. I think the proposed solution doesn't work either, as it contains some race conditions with the Archiver, as it is able to still create indices, that are not following potentially no template at all.

I'm closing this, but keep the reference.

PR #28588, opened 2025-02-21
refactor: removed number & string key headers

danielkelemen:
## Description

- Removed number & string key headers

## Related issues


tmetzke:
Just a thought: we could also leave them in (or at least the string header) so people that used this in 8.6 to fix applications and are migrating would not run into failures directly 🤔

danielkelemen:
We could definitely leave the string header.
The number probably doesn't make much sense, does it? We always return the same response.

houssain-barouni (approved):

tmetzke (approved):

PR #28600, opened 2025-02-21
test: add ArchiveTest in IT to check that only completed instances are archived

entangled90:
## Description
Basic implementation of an IT test for the archiver:
- clock is pinned to now - 2h
- 4 process instances are create
- 1 instance is completed
- clock is pinned to now - 1.5 h
- 2 instances are completed

We check that only the completed instance from 2 hours ago is deleted, after forcing ES to move the indices.

Inspired by @npepinpe test shared on slack.
It's a bit flaky on my machine:
- [ ] ensure it's not flaky

## Related issues

relates #24095


entangled90:
Superseded by #28997

PR #28646, opened 2025-02-23
Fix flaky PrefixMigrationIT

ChrisKujawa:
## Description

The PrefixMigrationIT used thread sleep to delay test execution, so we somewhat make sure importers see data before migration.

Problem with this was that first thread sleep is not reliable and we have no guarantee that we have seen data. Secondly, with using the MultiDbConfigurator we are configure the ES Exporter with a prefix, this can cause that data is written to a different index as read by Operate importers (as here we don't set the prefix).

These combinations caused flakiness. \cc @EuroLew 

### What this PR does

Replacing the `Thread.sleep` with a more reliable method, querying the v1 API to make sure Importers have seen data before doing the migration.

## Related

Slack thread https://camunda.slack.com/archives/C06F0GLJNFM/p1740163747867229


panagiotisgts (approved):

npepinpe:
Should we ban `Thread.sleep`? Only half joking :smile:

cmur2:
> Should we ban `Thread.sleep`? Only half joking 😄

Is there a good way in Java ecosystem to detect it/flag it as code smell in test code?

npepinpe:
[ArchUnit](https://www.archunit.org/) is a great tool for this (banning anyway). Detection can be done with other tools too, like Spotbugs, Sonaqube, etc.

PR #28659, opened 2025-02-24
Bound long polling metrics

npepinpe:
## Description

The `LongPollingMetrics` were previously technically unbounded in their cardinality, as we just accepted all job types as possible tags.

This introduces a small wrapper around a Caffeine cache, such that once a certain amount of tags is reached, the least frequently used is evicted, ensuring we stay within some bounds. At the same time, when the tag is evicted, the meter is removed from the underlying registry as well (ensuring it _also_ does not grow forever).

Micrometer does have a similar feature, but it's usage is a bit different: you set up a filter on the underlying registry specifying the maximum number of tags. This filter is global for the entire registry, but specific per metric.

Since filters need to be registered _before_ any meters are, we need to know at boot time which metrics will need this limit/filter, which is in itself pretty ugly. Additionally, a filter can just accept or deny - meaning, if we reach the limit, the registry returns a no-op meter. This means, when combined with a map/cache, you're still using up memory.

> [!Note]
> This does seem to imply using caching with Micrometer and filters is a bad idea in general. If everything is bounded, it's likely OK, but if not, then we could have a memory issue and a DoS vector.

Very open to other suggestions (other than migrate to OpenTelemetry - which the more I work with Micrometer, the more inclined I am to do so...)


npepinpe:
One last option is ignoring this - after all, we had the same issue with Prometheus in the past, where we would keep metrics in memory forever based on user input :)

npepinpe:
I will backport from 8.6 to 8.5 to reduce friction.

zeebe/util/src/main/java/io/camunda/zeebe/util/micrometer/BoundedMeterCache.java
```diff
@@ -0,0 +1,146 @@
...
+  private static <T extends Meter> void removeEvictedMeter(
+      final MeterRegistry registry, final T value) {
+    if (value != null) {
+      registry.remove(value);
+    }
+  }
+
+  /**
+   * Builds a bounded meter cache. By default, it sets the maximum size to 100, and expires/removes
+   * meters that haven't been accessed in the last.
```
lenaschoenburg: :upside_down_face: 
> haven't been accessed in the last

in the last what? :thinking:

zeebe/util/src/main/java/io/camunda/zeebe/util/micrometer/BoundedMeterCache.java
```diff
@@ -0,0 +1,146 @@
...
+
+  private static <T extends Meter> void removeEvictedMeter(
+      final MeterRegistry registry, final T value) {
+    if (value != null) {
+      registry.remove(value);
+    }
+  }
+
+  /**
+   * Builds a bounded meter cache. By default, it sets the maximum size to 100, and expires/removes
```
lenaschoenburg: :x: Looks like it now defaults to 500

lenaschoenburg (approved):

npepinpe:
Yes, I guess I pushed while you were review :s

I replaced the builder with some static factories. The reasoning behind it:
- Still some logic when building, so good to encapsulate this
- Eviction using a TTL _greatly_ affects performance, even in the plain "get" case. Since it was just a "nice to have", I figured let's drop it for now.
- Since the builder now only had one method, I figured let's drop it for now, we'll bring it back if we add more parameters.

That also coincided with updating the docs :sweat_smile:

PR #28673, opened 2025-02-24
ci: Create reopen_issues_without_version_label.yml

vheinila:
This GHA will reopen issues that are labeled "support" but do not have the "version:" label in them. This will also add a comment to these issues that the "version:" label needs to be added before closing the issue.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/.github/workflows/reopen_issues_without_version_label.yml
+++ b/.github/workflows/reopen_issues_without_version_label.yml
@@ -1,0 +1,42 @@
+on:
+  issues:
+    types: [closed]
+
+jobs:
+  reopenIfNoVersionLabel:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Reopen issue and add comment if no version label and has support label
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const issue = context.payload.issue;
+            const hasVersionLabel = issue.labels.some(label => label.name.startsWith('version:'));
+            const hasSupportLabel = issue.labels.some(label => label.name === 'support');
+
+            if (!hasVersionLabel && hasSupportLabel) {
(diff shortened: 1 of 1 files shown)


.github/workflows/reopen_issues_without_version_label.yml
```diff
@@ -0,0 +1,42 @@
+on:
+  issues:
+    types: [closed]
```
houssain-barouni, line 3: I think this will be also triggered when an issue is closed but not completed (not planned or duplicate)
I am not sure how we can distinguish between these states, it seems they are not part of the GH webhook payload
image

megglos: Even on completed we have a chicken-egg problem at least from zeebe perspective.
This is because version labels are [only added during the release process](https://github.com/camunda/camunda/blob/main/.github/workflows/camunda-platform-release.yml#L235) but not before/on/right close.

megglos: Given labeling is automated, we should never have support cases that got fixed and included in a release that are not labeled. I would thus propose to check the cases where it didn't work as expected and invest into fixing the causes of that.

vheinila: Update:

1. I will change the script not to re-open issues, but to only add a comment to these issues. 
2. We still need this script even though Monorepo has automated version labeling built into the release process, since we will be transferring issues into Monorepo as part of the Central public issue tracker project, that are not part of this process. 
3. Instead of running this script every time an issue is closed, we should run in once every month e.g. 2 days before a release date to check for issues lacking this version label from the previous 30 days.

megglos (changes requested):
please reassign for review one https://github.com/camunda/camunda/pull/28673/files#r1969578158 is addressed

PR #28675, opened 2025-02-24
fix: migrations should not block spring's context from being initialized

panagiotisgts:
## Description

Make MigrationsRunner execute the migrations without blocking Spring's context from being initialized, as the migrations are also dependent on other components of the single-app.
## Checklist

- [ ] for CI changes:

## Related issues

ref: https://camunda.slack.com/archives/C06F0GLJNFM/p1740134490554779?thread_ts=1740121225.157419&cid=C06F0GLJNFM

### DIFF
--- a/dist/src/main/java/io/camunda/application/commons/migration/MigrationsRunner.java
+++ b/dist/src/main/java/io/camunda/application/commons/migration/MigrationsRunner.java
@@ -9,6 +9,7 @@ package io.camunda.application.commons.migration;
 
 import io.camunda.migration.api.Migrator;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import org.slf4j.Logger;
@@ -34,7 +35,26 @@ public class MigrationsRunner implements ApplicationRunner {
   @Override
   public void run(final ApplicationArguments args) throws Exception {
     LOG.info("Starting {} migration tasks", migrators.size());
-    final Exception migrationsExceptions = new Exception("migration is failed");
+    /* Detach from main Spring thread */
+    CompletableFuture.runAsync(
+            () -> {
+              try {
+                startMigrators(args);
+              } catch (final Exception e) {
+                throw new RuntimeException(e);
+              }
+            },
+            Executors.newSingleThreadExecutor())
+        .exceptionally(
+            e -> {
+              LOG.error("Migration failed", e);
+              return null;
+            })
+        .whenComplete((ignored, throwable) -> LOG.info("All migration tasks finished"));
+  }
+
+  private void startMigrators(final ApplicationArguments args) throws Exception {
+    final Exception migrationsExceptions = new Exception("migration failed");
     try (final var executor = Executors.newFixedThreadPool(migrators.size())) {
       final var results = executor.invokeAll(migrators);
       for (final var result : results) {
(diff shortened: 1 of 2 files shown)


dist/src/main/java/io/camunda/application/commons/migration/MigrationsRunner.java
```diff
@@ -34,7 +35,26 @@ public MigrationsRunner(final List<Migrator> migrators) {
...
+    /* Detach from main Spring thread */
+    CompletableFuture.runAsync(
+            () -> {
+              try {
+                startMigrators(args);
+              } catch (final Exception e) {
+                throw new RuntimeException(e);
+              }
+            },
+            Executors.newSingleThreadExecutor())
```
lenaschoenburg, line 47: :x: Please set a meaningful thread name.
:wrench: I think a simple `new Thread(...).run()` might be preferable here.

lenaschoenburg (changes requested):

dist/src/main/java/io/camunda/application/commons/migration/MigrationsRunner.java
```diff
@@ -34,7 +34,20 @@ public MigrationsRunner(final List<Migrator> migrators) {
   @Override
   public void run(final ApplicationArguments args) throws Exception {
     LOG.info("Starting {} migration tasks", migrators.size());
-    final Exception migrationsExceptions = new Exception("migration is failed");
+    /* Detach from main Spring thread */
+    new Thread(
```
lenaschoenburg, line 38: :x: I still think we should set a thread name here.

:wrench: Ideally add one too for the executor used to run the actual migrations.

lenaschoenburg (commented):
:question: This is also changing the error handling. Before, we'd `MigrationsRunner#run` would throw an exception if any migration task fails. Now it is thrown in the new thread, without bubbling up. Is that acceptable and intended?

lenaschoenburg:
> Since we don't want to have the Migration run in a blocking way, would that matter?

I think that's the question we both want answers for :sweat_smile: 
I'd assume that migration failure is something serious that needs to be looked at. With this PR, I don't see how we or users would notice it. I guess @ChrisKujawa has an opinion on this already.

panagiotisgts:
The migration profile for SaaS will be run in the migration job (whatever that job is, I'm not familiar with it :sweat_smile: ) and not as part of the Zeebe deployment.

lenaschoenburg:
It doesn't matter if it runs as part of the single app deployment or as a separate job, both should be able to signal when a migration task fails.

ChrisKujawa:
> @lenaschoenburg Had a chat with @ChrisKujawa about this one an what to do with the exceptions. For now we'll simply log and swallow them on the Runner. We'll be revisiting this since for Identity as well there are open points and things that need to be adjusted. Shall we merge it for now and I'll create a ticket to track this down afterwards?

Just to add more context:

* I haven't reviewed the changes (and right now have no time for it)
* The migration apps **need** to run in parallel to the brokers and importers
  * To be specific in HC they will be part of the importer deployment
* Identity migration need to send commands to the broker, to publish related auth data from previous external systems
* Process migration need to migrate old imported processes, to enhance the data
  * The error space is here rather limited, and the migration is idempotent
  * We are able to retry this as well
  * If errors happen this is not the end of the world, and processing can be still done (there might be "just" some feature not fully working end to end). Something we can't prevent here.
* I synced previously with @megglos and likely the identity migration app/business logic need to anyway revisited, but this is done in a later milestone
  * Current target greenfield
  * @megglos please correct me if I'm wrong, but I see this then in the scope of the identity project to come up with a solution for the error handling special to the identity migration

With that my opinion we should no longer block this development, and move forward with the current approach/idea to make it run in parallel.

lenaschoenburg (approved):
I'll approve this then. Please consider my suggestion regarding thread names still :pray:

Issue #28158, opened 2025-02-14
Improve error handling - make sure stacktraces show which task fails


npepinpe:
# Description

Right now when an error occurs in a background task, especially on the Elastic side, we will log the exception as it bubbles up. The downside is that the stack trace only contains ES code, meaning we lose any information about which task was trying to access ES, which task failed, etc.

Additionally, some of the ES errors are not really errors - e.g. logging an error every time ES is temporarily down is not appropriate, because the expectation is we will recover automatically from this. So it should be a warning.

## Acceptance criteria

- [ ] I can identify which task produced an exception when the error is logged (through stack trace or just logging)
- [x] Recoverable errors (such as ConnectException, TimeoutException, IOException) are logged as warning, not errors.

sdorokhova:
Second point was implemented within #27746
The first is fixed in #28682

ChrisKujawa:
Closing. @sdorokhova please correct me if I'm wrong

PR #28682, opened 2025-02-24
feat: avoid log pollution with similar error messages

### LINKED ISSUE #25539: Cannot archive if a shard is unhealthy
**Describe the bug**

Archiving does not work with the new exporter is any of the shards are currently unhealthy. We had a case where this happened because ES went out of disk space, and when it restarted, it failed to start and would shut down because it cannot load it's persistent cache (due to out of disk space).

```
[2024-12-03T12:16:35,514][INFO ][o.e.n.Node               ] [elastic-0] starting ...
[2024-12-03T12:16:35,531][ERROR][o.e.b.Elasticsearch      ] [elastic-0] fatal exception while booting Elasticsearch java.io.UncheckedIOException: Failed to load persistent cache
	at org.elasticsearch.searchablesnapshots@8.15.3/org.elasticsearch.xpack.searchablesnapshots.cache.full.PersistentCache.repopulateCache(PersistentCache.java:272)
	at org.elasticsearch.searchablesnapshots@8.15.3/org.elasticsearch.xpack.searchablesnapshots.cache.full.CacheService.doStart(CacheService.java:196)
	at org.elasticsearch.server@8.15.3/org.elasticsearch.common.component.AbstractLifecycleComponent.start(AbstractLifecycleComponent.java:50)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.elasticsearch.server@8.15.3/org.elasticsearch.node.Node.start(Node.java:285)
	at org.elasticsearch.server@8.15.3/org.elasticsearch.bootstrap.Elasticsearch.start(Elasticsearch.java:457)
	at org.elasticsearch.server@8.15.3/org.elasticsearch.bootstrap.Elasticsearch.initPhase3(Elasticsearch.java:257)
	at org.elasticsearch.server@8.15.3/org.elasticsearch.bootstrap.Elasticsearch.main(Elasticsearch.java:76)
Caused by: java.io.IOException: No space left on device
	at java.base/sun.nio.ch.FileDispatcherImpl.write0(Native Method)
	at java.base/sun.nio.ch.FileDispatcherImpl.write(FileDispatcherImpl.java:62)
	at java.base/sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:132)
	at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:97)
	at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:67)
	at java.base/sun.nio.ch.FileChannelImpl.write(FileChannelImpl.java:288)
	at java.base/java.nio.channels.Channels.writeFullyImpl(Channels.java:74)
	at java.base/java.nio.channels.Channels.writeFully(Channels.java:96)
	at java.base/java.nio.channels.Channels$1.write(Channels.java:171)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.store.FSDirectory$FSIndexOutput$1.write(FSDirectory.java:400)
	at java.base/java.util.zip.CheckedOutputStream.write(CheckedOutputStream.java:73)
	at java.base/java.io.BufferedOutputStream.flushBuffer(BufferedOutputStream.java:81)
	at java.base/java.io.BufferedOutputStream.flush(BufferedOutputStream.java:142)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.store.OutputStreamIndexOutput.getChecksum(OutputStreamIndexOutput.java:104)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.codecs.CodecUtil.writeCRC(CodecUtil.java:644)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.codecs.CodecUtil.writeFooter(CodecUtil.java:412)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.index.SegmentInfos.write(SegmentInfos.java:691)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.index.SegmentInfos.write(SegmentInfos.java:569)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.index.SegmentInfos.prepareCommit(SegmentInfos.java:912)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.index.IndexWriter.startCommit(IndexWriter.java:5647)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.index.IndexWriter.prepareCommitInternal(IndexWriter.java:3790)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.index.IndexWriter.commitInternal(IndexWriter.java:4131)
	at org.apache.lucene.core@9.11.1/org.apache.lucene.index.IndexWriter.commit(IndexWriter.java:4093)
	at org.elasticsearch.searchablesnapshots@8.15.3/org.elasticsearch.xpack.searchablesnapshots.cache.full.PersistentCache$CacheIndexWriter.commit(PersistentCache.java:525)
	at org.elasticsearch.searchablesnapshots@8.15.3/org.elasticsearch.xpack.searchablesnapshots.cache.full.PersistentCache.repopulateCache(PersistentCache.java:261)
	... 7 more

ERROR: Elasticsearch did not exit normally - check the logs at /opt/bitnami/elasticsearch/logs/elastic.log
[2024-12-03T12:16:35,538][INFO ][o.e.n.Node               ] [elastic-0] stopping ...
[2024-12-03T12:16:35,547][INFO ][o.e.n.Node               ] [elastic-0] stopped
[2024-12-03T12:16:35,547][INFO ][o.e.n.Node               ] [elastic-0] closing ...
[2024-12-03T12:16:35,557][INFO ][o.e.n.Node               ] [elastic-0] closed
```

This means the shard is never assigned until there is more disk space, and archiving does not work, leading to more disk usage for other nodes as well.

**To Reproduce**

1. Create a cluster with the new Camunda Exporter.
1. Generate load quickly (<1h) until one node is out of disk space.
1. Restart the node.

Check that at least one shard for process instance related indices is unhealthy. This will prevent reindexing, which archiving relies on.

**Expected behavior**

1. We have a better error message describing what is happening. Right now, the error simply says we're missing some shards, and nothing else. Looking at ES is what led to diagnose this.
1. Log errors less often. We logged this error hundreds of thousands of time now in a short time, we can probably do better.

And if anyone has any other workaround, I'm all ears.

sdorokhova:
## Description

Improve logging in background tasks:
* add information about which background task has failed
* when one and the same exception is reoccurring, log only every 10th occurrence. Other are logged on DEBUG level.

## Related issues

closes #25539, #28158

### DIFF
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTask.java
@@ -11,4 +11,12 @@ import java.util.concurrent.CompletionStage;
 
 public interface BackgroundTask {
   CompletionStage<Integer> execute();
+
+  default String getCaption() {
+    final String simpleName = getClass().getSimpleName();
+    if (simpleName == null || simpleName.isEmpty()) {
+      return getClass().getName();
+    }
+    return simpleName;
+  }
 }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTask.java
@@ -13,10 +13,6 @@ public interface BackgroundTask {
   CompletionStage<Integer> execute();
 
   default String getCaption() {
-    final String simpleName = getClass().getSimpleName();
-    if (simpleName == null || simpleName.isEmpty()) {
-      return getClass().getName();
-    }
-    return simpleName;
+    return getClass().getSimpleName();
   }
 }
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
@@ -7,31 +7,23 @@
  */
 package io.camunda.exporter.tasks;
 
+import io.camunda.exporter.tasks.util.ReschedulingTaskLogger;
 import io.camunda.zeebe.util.ExponentialBackoff;
-import java.net.ConnectException;
-import java.net.SocketException;
-import java.net.SocketTimeoutException;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 
 public final class ReschedulingTask implements Runnable {
-  private static final Set<Class<? extends Exception>> BACKGROUND_SUPPRESSED_EXCEPTIONS =
-      Set.of(SocketTimeoutException.class, ConnectException.class, SocketException.class);
-
-  private static final Integer FAILURE_LOGGING_SKIP_COUNT = 10;
   private final BackgroundTask task;
   private final int minimumWorkCount;
   private final ScheduledExecutorService executor;
   private final Logger logger;
+  private final ReschedulingTaskLogger periodicLogger;
   private final ExponentialBackoff idleStrategy;
   private final ExponentialBackoff errorStrategy;
   private long delayMs;
   private long errorDelayMs;
-  private Integer failureCount = 0;
-  private String lastErrorMessage;
 
   public ReschedulingTask(
       final BackgroundTask task,
@@ -45,6 +37,7 @@ public final class ReschedulingTask implements Runnable {
     this.executor = executor;
     this.logger = logger;
 
+    periodicLogger = new ReschedulingTaskLogger(logger);
     idleStrategy = new ExponentialBackoff(maxDelayBetweenRunsMs, delayBetweenRunsMs, 1.2, 0);
     errorStrategy = new ExponentialBackoff(10_000, delayBetweenRunsMs, 1.2, 0);
   }
@@ -86,37 +79,12 @@ public final class ReschedulingTask implements Runnable {
     return errorDelayMs;
   }
 
-  /**
-   * To avoid log pollution with similar error messages we will log on ERROR level every 10th
-   * occurrence and all others on DEBUG level. Additionally, for some exception classes like
-   * `ConnectException` we will log the on WARN level.
-   *
-   * @param error
-   */
   private void logError(final Throwable error) {
     final String errorMessage =
         String.format(
             "Error occurred while performing a background task %s; error message `%s`; operation will be retried",
             task.getCaption(), error.getCause().getMessage());
-    if (lastErrorMessage != null && lastErrorMessage.equals(errorMessage)) {
-      failureCount++;
-    } else {
-      failureCount = 1;
-      lastErrorMessage = errorMessage;
-    }
-    // only log the error message if it's different from the last one, or if it's the same, only log
-    // every n-th time
-    if (lastErrorMessage == null
-        || !lastErrorMessage.equals(errorMessage)
-        || failureCount % FAILURE_LOGGING_SKIP_COUNT == 1) {
-      if (BACKGROUND_SUPPRESSED_EXCEPTIONS.contains(error.getCause().getClass())) {
-        logger.warn(errorMessage);
-      } else {
-        logger.error(errorMessage, error);
-      }
-    } else {
-      logger.debug(errorMessage);
-    }
+    periodicLogger.logError(errorMessage, error);
   }
 
   private void reschedule(final long delay) {
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
@@ -80,11 +80,11 @@ public final class ReschedulingTask implements Runnable {
   }
 
   private void logError(final Throwable error) {
-    final String errorMessage =
-        String.format(
-            "Error occurred while performing a background task %s; error message `%s`; operation will be retried",
-            task.getCaption(), error.getCause().getMessage());
-    periodicLogger.logError(errorMessage, error);
+    periodicLogger.logError(
+        "Error occurred while performing a background task {}; error message {}; operation will be retried",
+        error,
+        task.getCaption(),
+        error.getCause().getMessage());
   }
 
   private void reschedule(final long delay) {
--- a/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
+++ b/zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
@@ -20,15 +20,18 @@ import org.slf4j.Logger;
 public final class ReschedulingTask implements Runnable {
   private static final Set<Class<? extends Exception>> BACKGROUND_SUPPRESSED_EXCEPTIONS =
       Set.of(SocketTimeoutException.class, ConnectException.class, SocketException.class);
+
+  private static final Integer FAILURE_LOGGING_SKIP_COUNT = 10;
   private final BackgroundTask task;
   private final int minimumWorkCount;
   private final ScheduledExecutorService executor;
   private final Logger logger;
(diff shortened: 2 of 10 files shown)


zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTask.java
```diff
@@ -11,4 +11,12 @@
...
 public interface BackgroundTask {
   CompletionStage<Integer> execute();
+
+  default String getCaption() {
+    final String simpleName = getClass().getSimpleName();
+    if (simpleName == null || simpleName.isEmpty()) {
+      return getClass().getName();
+    }
+    return simpleName;
+  }
```
npepinpe, line 21: :wrench: imo we don't overload it with much benefit, I would just directly use `getClass().getSimpleName()`. I don't think simple name can ever be null though - but it could be empty if it's an unnamed class. I think this is only for anonymous classes though:

> Returns the simple name of the underlying class as given in the source code. An empty string is returned if the underlying class is `#isAnonymousClass()` anonymous or `#isUnnamedClass()` unnamed. A `#isSynthetic()` synthetic class, one not present in source code, can have a non-empty name including special characters, such as `$`.

So I think it's fine to just use `getClass().getSimpleName()` directly.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
```diff
@@ -76,19 +79,46 @@ private long onWorkPerformed(final int count) {
...
   }
 
+  /**
+   * To avoid log pollution with similar error messages we will log on ERROR level every 10th
+   * occurrence and all others on DEBUG level. Additionally, for some exception classes like
+   * `ConnectException` we will log the on WARN level.
+   *
+   * @param error
+   */
+  private void logError(final Throwable error) {
```
npepinpe, line 96: :wrench: We have a utility called `ThrottledLogger` which logs every X duration interval. e.g. `new ThrottledLogger(MY_LOGGER, Duration.ofSeconds(1))` will log at most once per second. We could use this instead, unless there's a specific benefit to logging every X log instead of every interval.

Though it's not quite what you implemented here, I think - the `ThrottledLogger` doesn't care about new messages, it just throttles all equally, whereas here you want to show different errors, right? I don't have a strong opinion here, as I'm not sure in practice which one is truly more useful (i.e. if we basically often have the same error in a short period, ThrottledLogger is fine - if we often alternate, then probably not).

:x: At any rate, I would encapsulate this in a utility somewhere to avoid having this class perform too many different things.

zeebe/exporters/camunda-exporter/src/test/java/io/camunda/exporter/tasks/ReschedulingTaskTest.java
```diff
@@ -40,6 +45,71 @@ void shouldRescheduleTaskOnError() {
...
+            10,
+            10,
+            executor,
+            loggerMock);
+
+    // when
+    task.run();
+
+    // then - when errors are different they are all logged on error level
+    final String errorMessageSubstring =
```
npepinpe, line 99: :wrench: Check out the `GrpcErrorMapperTest` for an example of how to unit test logging without mocking. You can use a recording appender, such that the logger ends up logging to it and not to the STDOUT or whatever, letting you run assertions on the log messages.

Of course, this implies having the class accept different loggers, but the `ReschedulingTask` already does :)

npepinpe (changes requested):

npepinpe:
[I just added this improvement on how to provide more context to logging in general.](https://github.com/camunda/camunda/issues/28729) I could imagine this would be a better alternative in the future to provide context that is tied to a specific logger instance, and not to a thread. That would be much more useful in multi-threading/asynchronous cases, where an operation crosses thread boundaries.

Note I'm not asking you to work on it, but once done, we may want to iterate on this using whatever the outcome of the issue is.

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/ReschedulingTask.java
sdorokhova: I'm not sure if I like the idea of simply skipping errors. I could understand skipping warnings or any other lower debug levels, but errors are kind of important.

I will separate the logging logic to some utility class.

npepinpe: Sounds good :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/ReschedulingTaskLogger.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public void logError(final String errorMessage, final Throwable error) {
+    if (lastErrorMessage != null && lastErrorMessage.equals(errorMessage)) {
+      failureCount++;
+    } else {
+      failureCount = 1;
+      lastErrorMessage = errorMessage;
+    }
+    // only log the error message if it's different from the last one, or if it's the same, only log
+    // every n-th time
```
npepinpe, line 47: :question: Does this do what the comment says? From what I can see, it just checks if there is a last error message, or if it's been enough skipped entries. I guess the check if it's the same is done higher up?

Also, it doesn't look to me like `lastErrorMessage` can ever be null, unless `errorMessage` itself is null?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTask.java
npepinpe: Did you decide not to apply this?

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/util/ReschedulingTaskLogger.java
```diff
@@ -0,0 +1,64 @@
...
+  public ReschedulingTaskLogger(final Logger logger) {
+    this.logger = logger;
+  }
+
+  public ReschedulingTaskLogger(final Integer skipEntriesCount, final Logger logger) {
+    this.skipEntriesCount = skipEntriesCount;
+    this.logger = logger;
+  }
+
+  public void logError(final String errorMessage, final Throwable error) {
```
npepinpe, line 39: :wrench: I would prefer keeping the same interface as `Logger`. So instead of pre-formatting the error message, we support something like

```java
  public void logError(final String errorMessage, final Throwable error, final Object... args) {
    if (lastErrorMessage != null && lastErrorMessage.equals(errorMessage)) {
      failureCount++;
    } else {
      failureCount = 1;
      lastErrorMessage = errorMessage;
    }
    // only log the error message if it's different from the last one, or if it's the same, only log
    // every n-th time
    final Level logLevel;
    if (lastErrorMessage == null || failureCount % skipEntriesCount == 1) {
      if (shouldBeSupressed(error)) {
        logLevel = Level.WARN;
      } else {
        logLevel = Level.ERROR;
      }
    } else {
      logLevel = Level.DEBUG;
    }

    logger.atLevel(logLevel).setCause(error).log(errorMessage, args);
  }
```

That way, you can call it as:

```java
logger.logError("Expected something to be {} but instead got {}", error, expected, actual);
```

This keeps the interpolation of arguments into the logger consistent (as it's not always a plain %s :)), keeps the API familiar, and it's a very tiny optimization by avoiding the interpolation if the log level is not high enough.

npepinpe (approved):
Pre-approving, but you put a thumbs up on a comment, and then ignored it - was that on purpose? For things with a :wrench:, which are strong suggestions, please justify when you decide to not apply them :)

zeebe/exporters/camunda-exporter/src/main/java/io/camunda/exporter/tasks/BackgroundTask.java
sdorokhova: Sorry, wanted to apply, but forgot 🤦

sdorokhova: Basically I observed nulls in test for inner classes, but I agree, it does not worth keeping this check only for those rare test occasions.

npepinpe (approved):

PR #28695, opened 2025-02-24
fix: wait for 8.7 records to be present

panagiotisgts:
## Description

Add `performOneRoundOfImport` on `OperateTester` to ensure the importer thread queue does not have records to import before progressing when doing single imports.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
+++ b/operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
@@ -18,7 +18,6 @@ import io.camunda.operate.util.OperateZeebeAbstractIT;
 import io.camunda.operate.util.TestSupport;
 import io.camunda.operate.zeebe.ImportValueType;
 import io.camunda.operate.zeebeimport.RecordsReaderHolder;
-import io.camunda.operate.zeebeimport.ZeebeImporter;
 import io.camunda.operate.zeebeimport.elasticsearch.ElasticsearchRecordsReader;
 import io.camunda.webapps.schema.descriptors.operate.index.ImportPositionIndex;
 import io.camunda.zeebe.exporter.ElasticsearchExporter;
@@ -36,7 +35,6 @@ import java.io.IOException;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Map;
-import org.awaitility.Awaitility;
 import org.elasticsearch.action.admin.indices.refresh.RefreshRequest;
 import org.elasticsearch.action.get.GetRequest;
 import org.elasticsearch.action.search.SearchRequest;
@@ -53,8 +51,8 @@ public class FinishedImportingIT extends OperateZeebeAbstractIT {
   private static final ElasticsearchExporter EXPORTER = new ElasticsearchExporter();
   private static final ElasticsearchExporterConfiguration CONFIG =
       new ElasticsearchExporterConfiguration();
+
   @Autowired public SchemaManager schemaManager;
-  @Autowired protected ZeebeImporter zeebeImporter;
   @Autowired private OperateProperties operateProperties;
   @Autowired private RecordsReaderHolder recordsReaderHolder;
   @Autowired private ImportPositionIndex importPositionIndex;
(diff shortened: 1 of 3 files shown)


operate/qa/integration-tests/src/test/java/io/camunda/operate/elasticsearch/FinishedImportingIT.java
```diff
@@ -162,10 +163,16 @@ final var record = generateRecord(ValueType.PROCESS_INSTANCE, "8.7.0", 1);
...
+    Awaitility.await()
+        .atMost(Duration.ofSeconds(30))
+        .pollInterval(Duration.ofSeconds(1))
+        .until(
+            () -> {
+              esClient.indices().refresh(new RefreshRequest("*"), RequestOptions.DEFAULT);
+              zeebeImporter.performOneRoundOfImport();
+              return isRecordIngested("1-process-instance")
+                  && isRecordIngested("2-process-instance");
+            });
```
ChrisKujawa, line 175: ❓ I'm wondering about this. Why does this help? We are also using await on the `then` part. We or the importer should be able to import the records and mark as complete no? How and why does this change? Can you elaborate on this? 

I'm afraid that we might simply hide a bug here.

panagiotisgts: I believe this is the same issue I've been facing with the Importers on the migration test setup. Sometimes, the flush of the `import-positions` is not happening when the import of the actual records is happening but on a later time, this is what creates this issue.

The actual record might be present but the `import-position` pointing to it is not guaranteed to be flushed on the same batch. By forcing the `performOneRoundOfImport` the flush of the `import-position` is being triggered. That's what I understand, not 100% sure about this one too. Maybe @sdorokhova can help clarify/verify this.

The await on the `then` part, is checking that the `import-position` is marked as completed, which is something done after the 8.8 record is ingested.

sdorokhova: I checked the code and I also don't understand why we would need to wait for `import-position` records beforehand. We decreased import-position flush interval till 5 seconds in `@Before` method and ideally after `zeebeImporter.performOneRoundOfImport();` was called 5 times in `when`, after 5 seconds `import-position` must be updated/inserted. And this should happen no matter whether import-position records were created before or not 🤔 

I now see that parent class `OperateZeebeAbstractIT` also declares `@Before` method. Can it be that two before methods are executed in unpredictable order and therefore update interval is not properly set or taken into account?

panagiotisgts: @ChrisKujawa @sdorokhova I think we are indeed hiding a bug. I believe the checks for the importers expect all partitions to have flushed records prior to 8.8 in order to update their _counts_. I'm still looking into this one.

operate/qa/integration-tests/src/test/java/io/camunda/operate/util/OperateTester.java
```diff
@@ -966,4 +978,11 @@ public OperateTester withAuthenticationToken(final String token) {
     SecurityContextHolder.getContext().setAuthentication(jwtAuthenticationConverter.convert(jwt));
     return this;
   }
+
+  public void performOneRoundOfImport() {
+    zeebeImporter.performOneRoundOfImport();
```
ChrisKujawa, line 983: 💭 ❓ We might need to move this into the `.untilAsserted` so the trigger is repeated? 🤔 What do you think? Can this be missed?

ChrisKujawa: ❓ Is it sufficient to simply check for the active job count in the executor 🤔 

Does it maybe make sense to check the import position? First get position - X. Then schedule and wait until it is larger then X?

ChrisKujawa: Looking at the code it might be sufficient

ChrisKujawa (approved):

PR #28715, opened 2025-02-25

### LINKED ISSUE #27234: Embedded gateway properties and standalone gateway properties are not handled correctly
**Describe the bug**

We are using `ZEEBE_STANDALONE_GATEWAY=true` so that zeebe is started in gateway mode.
Despite that, it seems that the default setting from application.yaml

```
zeebe:
  broker:
    gateway:
      authentication:
        security:
          mode: none
```

overrides the environment variable `ZEEBE_GATEWAY_AUTHENTICATION_SECURITY_MODE=identity` . Is this expected from your side?

**To Reproduce**

1. Configure zeebe in gateway mode using `ZEEBE_STANDALONE_GATEWAY=true`
2. Activate Identity using `ZEEBE_GATEWAY_AUTHENTICATION_SECURITY_MODE=identity`
3. Configure the identity sdk using `CAMUNDA_IDENTITY_...` environment
4. Start the configured gateway
5. Try to fetch the topology without a token (or call any other endpoint on the rest api)
6. It works!

**Expected behavior**

The REST API should be protected as identity is enabled.

**Log/Stacktrace**

n/a

**Environment:**

- OS: all
- Zeebe Version: 8.6.6
- Configuration: see above

**Links**

* 🔒 Slack thread: https://camunda.slack.com/archives/CSQ2E3BT4/p1737113739972799
* https://jira.camunda.com/browse/SUPPORT-25058

abremard:
## Description
An issue was raised where some identity authentication configs were applied to the wrong context. For example when running in standalone gateway mode, the `ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_MODE` is applied. This is not expected because it only serves to configure the embedded gateway (different mode). This PR is fixing the inconsistency issue by introducing more restrictive Spring conditions.

## Related issues

related to #27234


dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -117,16 +121,36 @@ public GatewaySecurityAuthenticationEnabledCondition() {
       super(ConfigurationPhase.REGISTER_BEAN);
     }
 
+    @Conditional(ProfileGatewayCondition.class)
     @ConditionalOnProperty(
         prefix = "zeebe.gateway",
         value = "security.authentication.mode",
-        havingValue = "none")
+        havingValue = "none",
+        matchIfMissing = true)
```
korthout, line 129: ❌ Introducing the `matchIfMissing` means that the default changes from security is on by default to security if disabled by default:
- The `GatewaySecurityAuthenticationEnabledCondition` is a `NoneNestedConditions`
- That means that it is True only if none of it's nested conditions are True
- In other words, `NONE` must be explicitly set to disable the security
- By using `matchIfMissing` the nested conditions could become True by default meaning that security is disabled

I suggest simply removing it.

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -117,16 +121,36 @@ public GatewaySecurityAuthenticationEnabledCondition() {
...
     static class EmbeddedGatewayCondition {}
   }
+
+  static class ProfileGatewayCondition implements Condition {
+    @Override
+    public boolean matches(final ConditionContext context, final AnnotatedTypeMetadata metadata) {
+      final Environment env = context.getEnvironment();
+      return env.acceptsProfiles(Profiles.of(io.camunda.application.Profile.GATEWAY.getId()));
+    }
+  }
```
korthout, line 147: ❓ I wonder whether the `GATEWAY` profile is present as well on embedded gateway, but I'm not sure. Please verify.

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -117,16 +121,36 @@ public GatewaySecurityAuthenticationEnabledCondition() {
...
+
+  static class ProfileBrokerCondition implements Condition {
+    @Override
+    public boolean matches(final ConditionContext context, final AnnotatedTypeMetadata metadata) {
+      final Environment env = context.getEnvironment();
+      return env.acceptsProfiles(Profiles.of(io.camunda.application.Profile.BROKER.getId()));
+    }
+  }
```
korthout, line 155: 🔧 This idea makes sense to me, but please verify that BROKER profile is present on embedded gateway and missing on standalone gateway.

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -117,16 +121,36 @@ public GatewaySecurityAuthenticationEnabledCondition() {
...
+    @Conditional(ProfileGatewayCondition.class)
     @ConditionalOnProperty(
         prefix = "zeebe.gateway",
         value = "security.authentication.mode",
-        havingValue = "none")
+        havingValue = "none",
+        matchIfMissing = true)
     static class StandaloneGatewayCondition {}
 
+    @Conditional(ProfileBrokerCondition.class)
```
korthout, line 132: ❓ Would this be the same `@Profile("broker")`, or is the implementation of `ProfileBrokerCondition` different?

korthout (commented):
Thanks for the fix @abremard 👏 

❌ Please add some tests to show that the behavior is now working as expected

👍 I like the idea to use profiles as a way to focus the conditions

🔧 If that doesn't work, you could check for `zeebe.broker.gateway.enabled=true` to know whether this concerns an embedded gateway, and `zeebe.gateway.enabled=true` to know whether it concerns a standalone gateway

abremard:
> Thanks for the fix @abremard 👏

Thanks for the quick review!

> ❌ Please add some tests to show that the behavior is now working as expected

I'm working on it, will add them on my next commit 🔨 

> 🔧 If that doesn't work, you could check for `zeebe.broker.gateway.enabled=true` to know whether this concerns an embedded gateway, and `zeebe.gateway.enabled=true` to know whether it concerns a standalone gateway

I was also thinking about these but didn't know which env variables to use, this is super helpful thanks!

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
abremard: Ah, thanks for raising that, I wasn't sure what the expected default should be, makes sense

korthout: @abremard For context on my previous fix (which didn't consider https://github.com/camunda/camunda/issues/27234), see this commit message a120b4b771e395833f8bbd2a06bc1a20f5287455

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
abremard: > :question: I wonder whether the GATEWAY profile is present as well on embedded gateway, but I'm not sure. Please verify.

I tried to run the [StandaloneBroker](https://github.com/camunda/camunda/blob/cab5fb9664aa502d0b1e528a8f8dbd5b064f1603/dist/src/main/java/io/camunda/application/StandaloneBroker.java#L16) with both `BROKER` and `GATEWAY` profiles + configs and it raises `[io.camunda.zeebe.broker.BrokerJobStreamService] conflicts with existing, non-compatible bean definition of same name and class [io.camunda.zeebe.gateway.GatewayJobStreamService]`

This is because the [broker service](https://github.com/camunda/camunda/blob/ff8dbe135d1523283ba1324ed42c98824150432d/dist/src/main/java/io/camunda/zeebe/broker/BrokerJobStreamService.java#L23) and [gateway service](https://github.com/camunda/camunda/blob/ff8dbe135d1523283ba1324ed42c98824150432d/dist/src/main/java/io/camunda/zeebe/gateway/GatewayJobStreamService.java#L22) have the same `@Component(value = "jobStreamService")` which means (if I understand correctly) both profiles are exclusive

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
abremard: > :wrench: This idea makes sense to me, but please verify that BROKER profile is present on embedded gateway

I can't find any code that says
- `zeebe.gateway.enabled=true` <=> GATEWAY profile
- `zeebe.broker.gateway.enabled=true` <=> BROKER profile

so probably using `zeebe.broker.gateway.enabled=true` and `zeebe.gateway.enabled=true` is safer, I just updated the condition

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -111,22 +112,22 @@ private HttpSecurity configureSecurity(final HttpSecurity http) throws Exception
    * set to {@code NONE}. It helps deal with the fact that the gateway can be embedded in the broker
    * or run standalone.
    */
-  static class GatewaySecurityAuthenticationEnabledCondition extends NoneNestedConditions {
+  static class GatewaySecurityAuthenticationEnabledCondition implements Condition {
```
abremard, line 115: The condition became a bit more complex with the additional checks I had to make, so I'm going for a `Condition` with a combination of booleans instead of a `NoneNestedConditions`. It might seem less concise/elegant but I think it's easier to test and debug

dist/src/main/java/io/camunda/zeebe/shared/security/SecurityConfiguration.java
```diff
@@ -41,7 +42,7 @@
 @EnableWebSecurity
 @EnableMethodSecurity
 @Configuration(proxyBeanMethods = false)
-public final class SecurityConfiguration {
+public class SecurityConfiguration {
```
abremard, line 45: This was a warning from IntelliJ, not related to this issue

dist/src/test/java/io/camunda/zeebe/shared/security/SecurityConfigurationTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Test
+  void matchesWhenStandaloneGatewayEnabledAndSecurityModeIsNotNone() {
+    when(environment.getProperty("zeebe.gateway.enabled")).thenReturn("true");
+    when(environment.getProperty("zeebe.gateway.security.authentication.mode"))
+        .thenReturn("identity");
+
+    final boolean result = condition.matches(conditionContext, methodMetadata);
+    assertThat(result).isTrue();
+  }
```
korthout, line 56: 🔧 Now this test contains multiple test cases that operate the same way but verify different cases. That's a great opportunity to use a ParameterizedTest to highlight exactly which different cases are tested.

🔧 I recommend using a record `TestCase` with a static inner class `Builder` to easily construct the different cases.
🔧 I recommend providing a `description` String to the `TestCase` record to be used to easily distinguish the case from others and use it in something like: `assertThat(result).describedAs(..).isEqualTo(case.isGatewaySecurityEnabled())`

This exercise may highlight that there are some cases missing:
- what happens when both `zeebe.gateway.enabled` and `zeebe.broker.gateway.enabled` are true? Can the class under test handle it?
- what happens when `zeebe.gateway.security.authentication.mode` is explicitly `none` but `zeebe.broker.gateway.security.authentication.mode` is explicitly `identity` for both the standalone gateway and the embedded gateway?
- etc

korthout (commented):
Thanks for the changes @abremard 🙌 

🔧 Please have a look at my suggestion

dist/src/test/java/io/camunda/zeebe/shared/security/SecurityConfigurationTest.java
```diff
@@ -0,0 +1,194 @@
...
+import org.springframework.core.type.MethodMetadata;
+
+public class SecurityConfigurationTest {
+
+  @Mock private Environment environment;
+
+  private ConditionContext conditionContext;
+  private MethodMetadata methodMetadata;
+  private final SecurityConfiguration.GatewaySecurityAuthenticationEnabledCondition condition =
+      new SecurityConfiguration.GatewaySecurityAuthenticationEnabledCondition();
```
korthout, line 33: 🔧 The class may be under test, but only the `Condition` interface is actually used in the test

```suggestion
  private final Condition condition =
      new SecurityConfiguration.GatewaySecurityAuthenticationEnabledCondition();
```

korthout (approved):
Thanks @abremard 🙌 

👍 I love that we can now easily see how it behaves under different combinations

🔧 I have a few final suggestions. Please have a look before merging.

LGTM 👍

dist/src/test/java/io/camunda/zeebe/shared/security/SecurityConfigurationTest.java
```diff
@@ -36,62 +40,155 @@ void setUp() throws Exception {
...
+                .build()),
+        Arguments.of(
+            new TestCase.Builder()
+                .description(
+                    "Standalone gateway enabled, standalone gateway security mode none, embedded gateway security mode identity")
+                .enableStandaloneGateway()
+                .standaloneGatewaySecurityMode("none")
+                .embeddedGatewaySecurityMode("identity")
+                .expectedResult(false)
+                .build()),
```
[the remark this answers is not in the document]
abremard: Good point, I've added warning logs for ambiguity. I will log off now so I'll merge this tomorrow morning

dist/src/test/java/io/camunda/zeebe/shared/security/SecurityConfigurationTest.java
```diff
@@ -59,57 +59,48 @@ void testGatewaySecurityAuthenticationEnabledCondition(final TestCase testCase)
...
                 .build()),
-        Arguments.of(
-            new TestCase.Builder()
-                .description(
-                    "Standalone gateway enabled, standalone gateway security mode none, embedded gateway security mode identity")
-                .enableStandaloneGateway()
-                .standaloneGatewaySecurityMode("none")
-                .embeddedGatewaySecurityMode("identity")
-                .expectedResult(false)
-                .build()),
```
korthout: 🔧 I did not intend for you to remove this test case, I see it as useful. My previous point about needing a test for it was about the warning log itself.

abremard: Ah I misunderstood your previous comment, just added the tests back

abremard (commented):
@korthout I know you've pre-approved this but since I had to fix a couple of small issues, just wanted to double-check this can get merged 🙇

korthout (approved):
Thanks @abremard 💯 

👍 Well done spotting those last few issues

💭 I saw that you updated your branch using a merge commit. Generally, we try to rebase our branches instead to ensure the history stays simpler, making it easier to read and navigate. It also helps keep pull requests easier to [backport using our automation](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md#backporting-changes) because merge commits cannot be automatically cherry-picked. I see that our [Contributing Guide](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md#creating-a-pull-request) doesn’t specify this clearly enough. I’ll see if I can highlight this somewhere.

LGTM 👍

PR #28735, opened 2025-02-25
Update routing for post importer action

ThorbenLindhauer:
related to #28468


sdorokhova (approved):
LGTM! I like that you take `processInstanceKey` from `treePath`. My first idea was to request them from `flow-node-instance` index when requesting for index names but this looks much better.

ThorbenLindhauer:
@sdorokhova the fix is the same that we applied on main as well. See the link in the ticket description to the PR.

Issue #28701, opened 2025-02-24
Decide on CI to ensure that the upgrade paths still working up to version 8.6


nathansandi:
### Description

Reintroduce on Task V1 the Migration tests for the version 8.6.0 and 8.7.0-SNAPSHOT

### Acceptance Criteria

- Migration tests working for the version `8.6` and `8.7-SNAPSHOT `

PR #28736, opened 2025-02-25
test: add migration tests for 8.7 snapshot path 

nathansandi:
## Description

Reintroduce on Task V1 the Migration tests for the version `8.6.0` and `8.7.0-SNAPSHOT`

## Related issues

closes https://github.com/camunda/camunda/issues/28701


tasklist/qa/migration-tests/migration-test/src/test/resources/test.properties
```diff
@@ -1 +1 @@
-upgrade.path=8.1.27,8.2.26,8.3.10,8.4.6,8.5.0
+upgrade.path=8.1.27,8.2.26,8.3.10,8.4.6,8.5.0,8.6.0,8.7.0-SNAPSHOT
```
nathansandi, line 1: This should be updated yo 8.7.0 after minor release.

houssain-barouni:
Hi @nathansandi , did you try to run the migration CI on your branch, to see if it works

nathansandi:
> Hi @nathansandi , did you try to run the migration CI on your branch, to see if it works

Hello @houssain-barouni, yes, I ran it locally as well, and it worked 👍.

tasklist/qa/migration-tests/migration-test/src/test/resources/test.properties
marcosgvieira, line 1: Should we keep 8.1 and 8.2 as they are not officially supported anymore, or would it be ok to remove it?
Also, would make any difference about the patch versions to not be the latest? I assume not because we don't change schemas or do schema migrations as part of patches, but double-checking.

tasklist/qa/migration-tests/pom.xml
```diff
@@ -19,6 +19,7 @@
     <module>test-fixture-840</module>
     <module>test-fixture-850</module>
     <module>test-fixture-860</module>
+    <module>test-fixture-870</module>
```
marcosgvieira, line 22: Same here, can we remove 810, 820?

marcosgvieira (changes requested):
Just minor question about versions that are not supported anymore, if we can remove them we can save a bit of time/processing.

tasklist/qa/migration-tests/migration-test/src/test/resources/test.properties
nathansandi: > Should we keep 8.1 and 8.2 as they are not officially supported anymore, or would it be ok to remove it?

Nice suggestion, I am going to remove this 

> Also, would make any difference about the patch versions to not be the latest? I assume not because we don't change schemas or do schema migrations as part of patches, but double-checking.

No difference so far, I have double check this when I was working on this PR, and as mentioned, schemas are not changed, as well as we dont deliver resources that could affect the migration over the patch releases.

marcosgvieira (approved):
LGTM!

Issue #28161, opened 2025-02-14
[EPIC]: Migrate existing root IT to new MultiDbExtension


ChrisKujawa:
## Description

In https://github.com/camunda/camunda/issues/26896 we created a new foundation. We now need to migrate all existing tests to such if possible.

```[tasklist]
- [ ] https://github.com/camunda/camunda/pull/27342
- [ ] https://github.com/camunda/camunda/issues/28162
- [ ] https://github.com/camunda/camunda/issues/28163
- [x] Migrate Backups E2E
- [ ] https://github.com/camunda/camunda/pull/28997
- [ ] https://github.com/camunda/camunda/issues/29479
- [ ] https://github.com/camunda/camunda/pull/29466
- [ ] https://github.com/camunda/camunda/pull/29478
- [ ] https://github.com/camunda/camunda/issues/29907
```

ChrisKujawa:
All tests (more than 40 tests) have been migrated.

PR #28746, opened 2025-02-25
Support authenticated client in MultiDb extension

### LINKED ISSUE #28160: Support Auth in root IT foundation
## Description

In https://github.com/camunda/camunda/issues/26896 we created a new foundation. We now need to migrate all existing tests to such if possible.

As some make use of authorization we need to support this in the MultiDb extension as well.

This would allow us to migrate more tests.

### LINKED ISSUE #28162: Migrate test in auth package to MultiDbExtension
## Description

Similar to https://github.com/camunda/camunda/pull/27342 we want to migrate all tests in `io.camunda.it.auth` to make use of the MultiDbExtension if possible. 

Prerequisite is https://github.com/camunda/camunda/issues/28160

ChrisKujawa:
## Description

We have several auth integration tests in our root QA module. Several of them are using the "old" BrokerITInvocationProvider, which only supports ES.

We need to make sure that the authentication mechanism works with different secondary storages, as this might depend on certain search implementations. \cc @megglos 

 In order to support such the CamundaMultiDbExtension needs to support creation of authenticated clients

### What this PR does

 * Refactoring: Extract certain classes out of the ClientFactory, moving them into a own utils.auth package. Allows to be used by other classes, like the extension.
 * Add necessary authentication configuration methods to the TestStandaloneCamunda and TestStandaloneBroker classes, so they can be used in the tests moving forward. Allowing configuring the standalone test classes in the specific tests.
 * Make use of the ClientFactory in the MultiDb Extension, allowing to either create authenticated, or default clients with no auth
 * Extracting static User properties (permissions, etc.) from test classes, to make sure to create necessary permissions, and create authenticated clients.
 * Introduce a marker annotation, for the user definition, to clear communicate that these are used/consumed by the extension
 * Updated documentations, and extend examples
 * Migrated ALL authentication test to make use of the new multi db extension

## Related issues

closes #28160 
closes https://github.com/camunda/camunda/issues/28162
relates to https://github.com/camunda/camunda/issues/28161


qa/integration-tests/src/test/java/io/camunda/it/auth/AuthorizationSearchIT.java
```diff
@@ -47,28 +51,18 @@ class AuthorizationSearchIT {
           ADMIN,
           DEFAULT_PASSWORD,
           List.of(new Permissions(AUTHORIZATION, PermissionTypeEnum.READ, List.of("*"))));
-
   private static final User RESTRICTED_USER = new User(RESTRICTED, DEFAULT_PASSWORD, List.of());
```
ChrisKujawa, line 54: 🔧 I'm thinking about introducing a marker annotation, so to make more clear that the extension is actually consuming these properties, to create permissions,etc. and for the authenticated clients.

ChrisKujawa:
@megglos I just took the momentum and migrated ALL auth package tests to the MultiDbExtension. I hope that was fine doing it as part of the PR, locally all succeeding.

qa/util/src/main/java/io/camunda/qa/util/auth/Authenticated.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to be passed along with {@link BrokerITInvocationProvider}'s {@link
+ * org.junit.jupiter.api.TestTemplate}. When applied, this indicates that the {@link CamundaClient}
+ * should be created with the provided user's credentials.
```
megglos, line 21: 🔧 needs updating as this is now also used by the MultiDbExtension?

qa/integration-tests/src/test/java/io/camunda/it/auth/AuthorizationSearchIT.java
megglos: We should do that, otherwise it's just unclear that these are picked up by the extension 😅

megglos (approved):
Nice! Generally looks good to me, except the user pickup magic, we should make this explicit with an annotation like you proposed, pre-approving.

qa/integration-tests/src/test/java/io/camunda/it/auth/AuthorizationSearchIT.java
ChrisKujawa: @megglos I introduced with https://github.com/camunda/camunda/pull/28746/commits/ea650fc97098fb54e962e7f7fd792e39eca2a7e0 a new marker interface.

Locally all tests succeed.

I also extended the docs to make this clear. https://github.com/camunda/camunda/pull/28746/commits/1da4fb7f2bea64f86544411f8fb37b3fe08c0916

PR #28753, opened 2025-02-25
feat: add rbdms architectural documentation

p-wunderlich:
## Description

Initial architectural docs for rdbms module.

### DIFF
--- a/docs/rdbms.md
+++ b/docs/rdbms.md
@@ -1,0 +1,285 @@
+# Rdbms Module
+
+**This document is WIP!**
+
+## Introduction and Goals
+
+The rdbms module adds support to Camunda for relational database management systems (RDBMS) as an
+alternative to
+Elasticsearch (ES) and OpenSearch (OS) for second-level storage.
+
+Key Features & Benefits:
+
+* Seamless RDBMS Integration: Supports SQL-based databases, enabling customers to leverage existing
+  DBA expertise.
+* Simplified Onboarding & Testing: Removes the dependency on ES/OS, making it easier to use
+  lightweight in-memory databases for local development and testing scenarios.
+* Operational Efficiency: Facilitates easier maintenance, backups, and upgrades with established
+  database procedures.
+* Enterprise Readiness: Addresses licensing constraints, compliance requirements, and cost concerns.
+
+### Requirements Overview
+
+### Quality Goals
+
+tbd
+
+### Stakeholders
+
+* Daniel Meyer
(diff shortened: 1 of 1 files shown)


docs/rdbms.md
```diff
@@ -0,0 +1,411 @@
...
+flowchart LR
+  USER(["User"])
+  Camunda_8["Camunda"]
+  RDBMS_DB[("RDBMS (e.g. Postgres)")]
+  Customer --> Camunda_8 --> RDBMS_DB
+```
+
+| Entity  | Description                                                      |
+|---------|------------------------------------------------------------------|
+| User    | The user which uses Camunda.                                     |
```
stoerti, line 57: In the graph we have a user (unconnected) and a customer, in the table only the user.

docs/rdbms.md
```diff
@@ -0,0 +1,411 @@
...
+    end
+    subgraph db[DB RDBMS]
+      RDBMS_SERVICE["RDBMS Service"]
+    end
+  end
+  USER --> CAMUNDA
+  RDBMS_SERVICE --> RDBMS_DB
+  broker --> RDBMS_SERVICE
+  restGateway --> RDBMS_SERVICE
+```
```
stoerti, line 85: 💭  Maybe it's just the alignment of the entities, but it looks like that the User uses the broker but not the REST gateway.

docs/rdbms.md
```diff
@@ -0,0 +1,411 @@
...
+      REST_GATEWAY["REST Gateway"]
+    end
+    subgraph camundaService["Camunda Service"]
+      CAMUNDA_SERVICE_SERVICE["Camunda Service"] --> SEARCH_CLIENT
+      BROKER_CLIENT["Broker Client"]
+    end
+    REST_GATEWAY --> CAMUNDA_SERVICE_SERVICE --> BROKER_CLIENT --> broker
+    REST_GATEWAY --> CAMUNDA_SERVICE_SERVICE
+    subgraph broker["`**Broker (Partition)** - _Command Side_`"]
+      RECORD_EXPORTER["Record Exporter"]
```
stoerti, line 126: I would add the engine here as a black box which produces records.

stoerti (approved):

docs/rdbms.md
p-wunderlich: For this simple diagramm, the user should just use Camunda itself. It's just the auto alignment of mermaid ..

PR #28779, opened 2025-02-26
fix: default addresses are plaintext

jonathanlukas:
## Description

This PR changes the protocol of the default addresses from https to http.

As a side effect, the client will by default assume that the `grpcAddress` is used instead of the `gatewayAddress`.

Otherwise, an unconfigured client will still attempt to connect to the grpc gateway using https as the property `usePlaintextConnection` is applied.

One thing to check would be how much sense it actually makes to use TLS by default and "opt-out" for the gateway address.

## Checklist

- [ ] for CI changes:

## Related issues

closes #

### DIFF
--- a/clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
+++ b/clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
@@ -81,8 +81,8 @@ public final class CamundaClientBuilderImpl
 
   public static final String DEFAULT_GATEWAY_ADDRESS = "0.0.0.0:26500";
   public static final URI DEFAULT_GRPC_ADDRESS =
-      getURIFromString("https://" + DEFAULT_GATEWAY_ADDRESS);
-  public static final URI DEFAULT_REST_ADDRESS = getURIFromString("https://0.0.0.0:8080");
+      getURIFromString("http://" + DEFAULT_GATEWAY_ADDRESS);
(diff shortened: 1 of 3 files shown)


koevskinikola (dismissed):
Hey @jonathanlukas, sorry I've been busy with release manager tasks and haven't been able to look into this PR.

❌ Can you link the GitHub issue this PR relates to?

clients/java/src/main/java/io/camunda/client/impl/CamundaClientBuilderImpl.java
```diff
@@ -81,8 +81,8 @@ public final class CamundaClientBuilderImpl
 
   public static final String DEFAULT_GATEWAY_ADDRESS = "0.0.0.0:26500";
   public static final URI DEFAULT_GRPC_ADDRESS =
-      getURIFromString("https://" + DEFAULT_GATEWAY_ADDRESS);
-  public static final URI DEFAULT_REST_ADDRESS = getURIFromString("https://0.0.0.0:8080");
+      getURIFromString("http://" + DEFAULT_GATEWAY_ADDRESS);
+  public static final URI DEFAULT_REST_ADDRESS = getURIFromString("http://0.0.0.0:8080");
```
ana-vinogradova-camunda, line 85: Thank you for fixing that @jonathanlukas 👍

ana-vinogradova-camunda (approved):
Changes look good to me 👍 
Thank you @jonathanlukas 

Please let me know if there are any outstanding questions.

PR #28782, opened 2025-02-26
fix: adjust bean condition for not consolidated auth to avoid conflict

### LINKED ISSUE #28778: Latest SNAPSHOT Tasklist does not start
### Describe the bug

The latest SNAPSHOT Tasklist docker container does not start with the following exception:
```
io.camunda.application - Failed to start application with message: Error creating bean with name 'io.camunda.tasklist.webapp.api.rest.v1.controllers.TaskController': 
Unsatisfied dependency expressed through field 'userGroupService': 
No qualifying bean of type 'io.camunda.tasklist.webapp.security.identity.UserGroupService' available: expected single matching bean but found 2: io.camunda.tasklist.webapp.security.identity.DefaultUserGroupService,defaultUserGroupService
```

Once fixed, we need to reactivate the IT tests that has been disabled in https://github.com/camunda/camunda/pull/28774

nathansandi:
## Description

The annotation @Profile("!consolidated-auth") does not fully cover the opposite case, leading to a bean conflict when no active profile is explicitly set. This issue occurs because, by default, Spring attempts to register multiple beans of the same type, resulting in ambiguity.

To take advantage of Component Scam, the beans declaration was removed, and the component implementation moved to a group, in order to be detected by spring component scan.

## Related issues

closes https://github.com/camunda/camunda/issues/28778


houssain-barouni (changes requested):
[DefaultUserGroupService](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/DefaultUserGroupService.java) and [UserGroupServiceImpl](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/UserGroupServiceImpl.java) are already defined as Spring components in Tasklist, so I propose the following:
- remove the beans definition from `TasklistModuleConfiguration`, these beans are needed by the webapp, so no need to create them for the importer (in SaaS for example)
- move `DefaultUserGroupService` to `UserGroupServiceImpl` out of `io.camunda.tasklist.webapp.security` package (to `io.camunda.tasklist.webapp.group` package) so that it gets created in the Single app (where the security package is disabled in favor of Operate's one)

This way I think it will work.

nathansandi:
> [DefaultUserGroupService](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/DefaultUserGroupService.java) and [UserGroupServiceImpl](https://github.com/camunda/camunda/blob/main/tasklist/webapp/src/main/java/io/camunda/tasklist/webapp/security/identity/UserGroupServiceImpl.java) are already defined as Spring components in Tasklist, so I propose the following:
> 
> * remove the beans definition from `TasklistModuleConfiguration`, these beans are needed by the webapp, so no need to create them for the importer (in SaaS for example)
> * move `DefaultUserGroupService` to `UserGroupServiceImpl` out of `io.camunda.tasklist.webapp.security` package (to `io.camunda.tasklist.webapp.group` package) so that it gets created in the Single app (where the security package is disabled in favor of Operate's one)
> 
> This way I think it will work.

Nice suggestion @houssain-barouni ;) This way also works. I did the refactoring.

dist/src/main/java/io/camunda/tasklist/TasklistModuleConfiguration.java
```diff
@@ -67,39 +62,6 @@ public TasklistModuleConfiguration(
...
-  /**
-   * Bean definition for `UserGroupService` when the application is running under the
-   * "consolidated-auth" profile.
-   *
-   * <p>- The `@Profile("consolidated-auth")` annotation ensures that this bean is only loaded when
-   * the profile "consolidated-auth" is active.
-   *
-   * <p>This bean provides the **main authorization logic** when consolidated authentication is
-   * enabled.
-   */
```
houssain-barouni, line 79: MINOR: should we keep these javadocs and move them to the classes?

houssain-barouni (approved):
LGTM!

PR #28822, opened 2025-02-26
[CPT] feat: Replace the internal HTTP client with the Camunda client

### LINKED ISSUE #19263: [CPT] Replace the internal client by the official client to consume the new REST API (v2)
## Description

Camunda Process Test should use the Camunda Java SDK to consume the new REST API (v2). To speed up the development, the testing library consumes the old API until now.  

## Tasks
- Replace the proprietary API access with the new Java SDK (ZeebeClient)
- Write a test to verify that the REST API can be consumed
- Configure pagination of the search API

saig0:
## Description

Major changes:
- Enable the search API of the Camunda client by using the `consolidated-auth` profile
- Replace the internal HTTP client for the v1 REST API with the v2 search API from the Camunda client.

The PR contains a bunch of refactorings to replace the internal client and DTOs with the Camunda client interfaces. 

The usage of the Camunda client opens some improvements of queries in the assertions and the result collector. I'll follow up on these improvements in separate PRs.    

## Related issues

closes #19263


testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/CamundaDataSource.java
```diff
@@ -15,44 +15,64 @@
...
-import io.camunda.process.test.impl.client.VariableDto;
-import java.io.IOException;
+import io.camunda.client.api.search.response.Variable;
 import java.util.List;
+import java.util.function.Consumer;
 
 public class CamundaDataSource {
 
-  private final CamundaApiClient camundaApiClient;
+  private static final Consumer<SearchRequestPage> DEFAULT_PAGE_REQUEST = page -> page.limit(100);
```
berkaycanbc, line 29: 💭 100 is probably enough for most cases but would be good to add this default behaviour to the documentation in any case.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/testresult/CamundaProcessTestResultCollector.java
```diff
@@ -49,7 +47,7 @@ public ProcessTestResult collect() {
               .map(this::collectProcessInstanceResult)
               .collect(Collectors.toList());
       result.setProcessInstanceTestResults(processInstanceResults);
-    } catch (final IOException e) {
+    } catch (final Exception e) {
```
berkaycanbc, line 51: ❓No checked exception is thrown inside the try block anymore. Why do we still have a catch here? Or why don't we catch a `RuntimeException` which is more accurate to use if we still want to catch exceptions for proper logging.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/assertions/ElementAssertj.java
```diff
@@ -246,7 +245,7 @@ private void hasElementInState(
 
     try {
       Awaitility.await()
-          .ignoreException(CamundaClientNotFoundException.class)
+          .ignoreExceptions()
```
berkaycanbc, line 248: ❓Do we want to ignore all exceptions? The previous version was ignoring `CamundaClientNotFoundException` because it might be a temporary issue. But after the refactoring we now ignore all types of exceptions.

testing/camunda-process-test-java/src/main/java/io/camunda/process/test/impl/testresult/CamundaProcessTestResultCollector.java
```diff
@@ -74,26 +75,12 @@ private Map<String, String> collectVariables(final long processInstanceKey) {
...
 
-  private List<OpenIncident> collectOpenIncidents(final long processInstanceKey) {
-    return dataSource.getFlowNodeInstancesByProcessInstanceKey(processInstanceKey).stream()
-        .filter(
-            flowNodeInstance ->
-                flowNodeInstance.getIncident() != null && flowNodeInstance.getIncident())
-        .map(this::getIncident)
+  private List<Incident> collectOpenIncidents(final long processInstanceKey) {
+    return dataSource.getIncidentsByProcessInstanceKey(processInstanceKey).stream()
+        .filter(incident -> incident.getState().equals(IncidentState.ACTIVE))
```
berkaycanbc, line 80: 🚀 Nice!

berkaycanbc (approved):
Big refactoring, very good job @saig0! 🚀 

❓I've asked couple of questions that might lead to changes that are not a blocker.

👍 LGTM

PR #28857, opened 2025-02-27

abremard:
## Description
An issue was raised where some identity authentication configs were applied to the wrong context. For example when running in standalone gateway mode, the `ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_MODE` is applied. This is not expected because it only serves to configure the embedded gateway (different mode). This PR is applying the fix from [8.6](https://github.com/camunda/camunda/pull/28715) to 8.7

## Related issues

related to #27234


korthout (approved):
Thanks @abremard 👍 

💭 There's often no need to manually port pull requests to other branches. We have [automation for this](https://github.com/camunda/camunda/blob/main/CONTRIBUTING.md#backporting-changes). If it fails (e.g. due to conflicts), you'll still have to do some manual work, but at least the PR will be automatically created for you. If it succeeds, the automatically created PR is even auto-approved, so it also reduces effort on reviewers. It also ensures that commits are cherry-picked with back-references. I strongly suggest you use it next time.

PR #28858, opened 2025-02-27

abremard:
## Description
An issue was raised where some identity authentication configs were applied to the wrong context. For example when running in standalone gateway mode, the `ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_MODE` is applied. This is not expected because it only serves to configure the embedded gateway (different mode). This PR is applying the fix from [8.6](https://github.com/camunda/camunda/pull/28715) to 8.8

## Related issues

related to #27234


korthout (approved):
LGTM 👍

PR #28867, opened 2025-02-27
Explain why our contributing guide advises rebasing instead of merge commits

korthout:
## Description

We advise rebasing rather than using a merge commit to update a branch, but it was hidden inside the section where a merge was already queued. That's not always the case; sometimes, folks need to update their branch before it is approved.

Therefore, we should highlight this more explicitly. I've also explained why this matters and improved the wording from 'rebase `main` into the branch' to 'rebase onto the target branch'. Note that officially, the target branch is called the 'base', but 'target' is probably a more generally understood term.

This PR also improves the backport instructions, as I noticed that they were outdated when I linked them in the section where we advise rebasing over merge commits.

## Related issues

https://github.com/camunda/camunda/pull/28715#pullrequestreview-2646870880
https://github.com/camunda/camunda/pull/28857#pullrequestreview-2646885409


lenaschoenburg (approved):
Nice improvement in clarity, thanks :+1:

tmetzke (approved):
🚀 Nice improvements

PR #28889, opened 2025-02-27
fix: fix possible sigsev in tests

npepinpe:
## Description

The performance test closes the DB without closing the state controller, meaning the state controller may still try to access the DB afterwards. This can cause a sigsev when that happens.

Since the state controller is now doing the metrics export, there can be a race condition where the export runs concurrently while the DB is closed. This is only a test issue, since in production, the controller is the only one closing the DB.

To avoid this, we don't close the DB explicitly, but instead close the state.


lenaschoenburg (approved):

npepinpe:
No need to backport, I found it while working on the backports of the original PR :sweat_smile:

PR #28890, opened 2025-02-27
Forward port migrating the job stream client metrics to main

npepinpe:
## Description

This PR forward ports #28119 to main (which will then be backported to 8.7 and 8.6). No real important conflicts, but it had a merge commit so it couldn't be backported automatically :(

The one small change I did here was the last commit, which migrates the standard gauges to `StatefulGauge`.

## Related issues

forward ports #28119


lenaschoenburg (approved):
Looks good , but I haven't given it much thought since it's just a forward-port.

PR #28902, opened 2025-02-27
refactor: OpenAPI cleanup base objects

danielkelemen:
## Description

- Cleanup OpenAPI base objects 

## Related issues


pepopowitz (approved):
❤️

tmetzke (approved):
👍 Thank you 
❓ Do we need to have a look at 8.7, to check if we flipped the default for those DTOs?

danielkelemen:
> ❓ Do we need to have a look at 8.7, to check if we flipped the default for those DTOs?

Yeah, I suspect that's the case since code change was also required.

Issue #28904, opened 2025-02-27
Clean up not used test classes


ChrisKujawa:
With migrating our tests to the new foundation https://github.com/camunda/camunda/issues/28161 we should be able to remove old obsolete test utils and classes.

```[tasklist]
### Clean up
- [x] Move MultiDbClasses to util
- [x] Deprecate TestStandaloneCamunda
- [x] Deprecate BrokerITInvocationProvider
- [x] After migrating all tests https://github.com/camunda/camunda/issues/28161
- [ ] Delete test classes, from above
- [ ] https://github.com/camunda/camunda/pull/29255
- [ ] https://github.com/camunda/camunda/pull/30869
```

PR #28915, opened 2025-02-27
Migrate MultiDb extension

ChrisKujawa:
## Description

 * To better separate concerns and reference classes. We migrate the MultiDb extensions, annotations, and helper classes to the util module we have for our integration tests

## Related issues

related to https://github.com/camunda/camunda/issues/28904


megglos (approved):
👍

Issue #24392, opened 2024-11-05
Operate nightly CI jobs are not running on stable branches


ThorbenLindhauer:
### Description

* The following builds are only running on main, not on the stable branches:
  * https://github.com/camunda/camunda/actions/workflows/operate-test-importer.yml
  * https://github.com/camunda/camunda/actions/workflows/operate-test-migrate-elasticsearch-data.yml
* Reason: Time-based github workflow triggers run only once (expected) and target only the repository's default branch, e.g. see https://stackoverflow.com/questions/58798886/github-actions-schedule-operation-on-branch

### TODO

* The nightly builds also run on the stable branches: `stable/8.6` and `stable/operate-8.5`
* They run on future stable branches (if needed)
* They have a timeout to not take up unnecessary resources for too long after a test failure

### Additional Information

PRs:

* main: https://github.com/camunda/camunda/pull/32751
* 8.7: https://github.com/camunda/camunda/pull/32570
* 8.6: https://github.com/camunda/camunda/pull/32753
* 8.5: https://github.com/camunda/camunda/pull/32752

ThorbenLindhauer:
Migration job is addressed via https://github.com/camunda/camunda/issues/24840

ThorbenLindhauer:
`ImporterMetricsZeebeImportIT` and similar test cases are failing on the 8.6 PR, because they assert the value of Prometheus metrics and expect

`operate_events_processed_total{partition=\"1\",status=\"succeeded\",type=\"JOB\",}`

while the actual value is

`operate_events_processed_total{partition=\"1\",status=\"succeeded\",type=\"JOB\"}`

Note the only difference is the last comma in the list of labels. Also note that a comma is optionally allowed at this place, so it's not an invalid server response: https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details

What's curious is that the tests succeed on both 8.7 and 8.5 (i.e. commas are appended), and it's not clear to me right now what the difference is.

ThorbenLindhauer:
State of progress and what's still left to do:

* I had a [slack discussion with Christian](https://camunda.slack.com/archives/C071KP5BTHB/p1741111865809949) to change this from a nightly build to a build that runs on every PR. Reasoning is that the nightly builds just don't work, nobody acts on failures. We should change the scope here to 
  * make them run on main + stable branches
  * make them run with every PR (not nightly trigger)
  * make them succeed
[...]

PR #28931, opened 2025-02-27
24392: enable Operate importer test cases on 8.7

ThorbenLindhauer:
related to #24392


cmur2:
Maybe it makes sense to raise a generic PR against `main` that handles `push: ['stable/**']` and backport it to avoid code divergence?

PR #28962, opened 2025-02-28
Deprecate old test classes

ChrisKujawa:
## Description

With the new IT foundation we build, where we support multi secondary storages, we no longer should use the classes:

 * TestStandaloneCamunda, as this is coupled to ES and starting an ES container
 * BrokerITInvocationProvider, as this approach is outdated, using test containers and
 starting all at once, and running the test multiple times. Causing long
 test executions that can't be broken up into multiple CI jobs, and
 also only supports a limited set of secondary storage

## Related issues

related https://github.com/camunda/camunda/issues/28904


megglos (approved):
👋

PR #28963, opened 2025-02-28
test: ensure gRPC gateway security config is consistent with REST

### LINKED ISSUE #28859: Ensure gRPC gateway security config is consistent with REST
**Description**

We've fixed the REST API gateway security configs to be consistent with standalone/embedded `authentication.security.mode` in #27234 (see [PR](https://github.com/camunda/camunda/pull/28715)). This issue is a follow-up task to make sure the behavior in gRPC is consistent with the one in REST.

abremard:
## Description

This PR is adding more integration tests to verify that the REST and gRPC behaviors are consistent for both standalone and embedded gateway authentication modes.

Already existing tests coverage :
- REST/gRPC embedded gateway with security mode NONE
- REST/gRPC embedded gateway with security mode IDENTITY

This PR is adding :
- REST/gRPC standalone gateway with security mode NONE
- REST/gRPC standalone gateway with security mode IDENTITY

## Related issues

related to #28859
follow-up issue from #27234

### DIFF
--- a/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/EmbeddedGatewayAuthenticationNoneIT.java
+++ b/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/EmbeddedGatewayAuthenticationNoneIT.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.zeebe.it.gateway;
+
+import io.camunda.application.Profile;
+import io.camunda.zeebe.qa.util.cluster.TestGateway;
+import io.camunda.zeebe.qa.util.cluster.TestStandaloneBroker;
+
+public class EmbeddedGatewayAuthenticationNoneIT
+    extends GatewayAuthenticationNoneAbstractIT<TestStandaloneBroker> {
+
+  private final TestStandaloneBroker zeebe =
+      new TestStandaloneBroker()
+          .withAdditionalProfile(Profile.IDENTITY_AUTH)
+          .withProperty("zeebe.broker.gateway.security.authentication.mode", "none")
+          .withProperty("camunda.identity.issuerBackendUrl", getKeycloakRealmAddress())
+          .withProperty("camunda.identity.audience", ZEEBE_CLIENT_AUDIENCE);
+
+  @Override
+  protected TestGateway getZeebe() {
+    return zeebe;
+  }
+}
--- a/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/EmbeddedGatewayAuthenticationNoneIT.java
+++ b/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/EmbeddedGatewayAuthenticationNoneIT.java
@@ -22,7 +22,7 @@ public class EmbeddedGatewayAuthenticationNoneIT
           .withProperty("camunda.identity.audience", ZEEBE_CLIENT_AUDIENCE);
 
   @Override
-  protected TestGateway getZeebe() {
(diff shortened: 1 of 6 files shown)


zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/EmbeddedGatewayAuthenticationNoneIT.java
```diff
@@ -0,0 +1,28 @@
...
+import io.camunda.zeebe.qa.util.cluster.TestGateway;
+import io.camunda.zeebe.qa.util.cluster.TestStandaloneBroker;
+
+public class EmbeddedGatewayAuthenticationNoneIT
+    extends GatewayAuthenticationNoneAbstractIT<TestStandaloneBroker> {
+
+  private final TestStandaloneBroker zeebe =
+      new TestStandaloneBroker()
+          .withAdditionalProfile(Profile.IDENTITY_AUTH)
+          .withProperty("zeebe.broker.gateway.security.authentication.mode", "none")
```
abremard, line 20: For reasons unclear to me, I'm unable to use the `withBrokerConfig` modifier with `AuthMode.None`. The existing tests is also setting the config directly [here](https://github.com/camunda/camunda/blob/stable/8.6/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/GatewayAuthenticationNoneIT.java#L124) as opposed to using the modifier [here](https://github.com/camunda/camunda/blob/stable/8.6/zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/GatewayAuthenticationIdentityIT.java#L125-L129)

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/GatewayAuthenticationIdentityAbstractIT.java
```diff
@@ -53,16 +49,16 @@
 
 @Testcontainers
 @ZeebeIntegration
-public class GatewayAuthenticationIdentityIT {
```
abremard, line 56: It's worth noting that `GatewayAuthenticationIdentityIT` have been removed in #24757 with a plan to be re-added with adjustments in #25718 . This means I'm planning to port this to 8.7 but not to 8.8.

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/StandaloneGatewayAuthenticationNoneIT.java
```diff
@@ -0,0 +1,33 @@
...
+          .withGatewayConfig(
+              cfg -> {
+                final var auth = cfg.getSecurity().getAuthentication();
+                auth.setMode(AuthMode.NONE);
+              })
+          .withProperty("camunda.identity.issuerBackendUrl", getKeycloakRealmAddress())
+          .withProperty("camunda.identity.audience", ZEEBE_CLIENT_AUDIENCE);
+
+  @Override
+  protected TestGateway getZeebe() {
```
korthout, line 30: 🔧 You can provide the generics here to be more explicit
```suggestion
  protected TestGateway<TestStandaloneGateway> getZeebe() {
```

zeebe/qa/integration-tests/src/test/java/io/camunda/zeebe/it/gateway/EmbeddedGatewayAuthenticationNoneIT.java
korthout: 💭 The problem is that we're using Spring magic in the `GatewaySecurityAuthenticationEnabledCondition` (only inspecting Spring properties). Setting the `brokerCfg` directly circumvents Spring properties (actually the Spring properties are used to build up the `brokerCfg`).

💭 This means that `GatewaySecurityAuthenticationEnabledCondition` isn't able to function correctly when the properties are not loaded through a spring context.

👍 I believe this is not an issue because our applications always run within a Spring context.

💭 It might be an issue in other tests when people need to control this, and they run into the same issue spending time on it. One idea would be to re-implement `.withBrokerConfig` to set up spring properties rather than directly provide the config object, but IMO this exceeds the current scope.

korthout (approved):
Nice work! @abremard 🚀 

🔧 Just a few minor suggestions. Please have a look before merging.

LGTM 👍

PR #28974, opened 2025-02-28
Clean up Tenants handler in the RDBMS Exporter

### LINKED ISSUE #28966: Clean up the Tenant handler in the RDBMS Exporter
## Description

- `deleteAllMembers` script can go. There will be individual events to remove members from a tenant when a tenant gets deleted.
- `tenantId` should be immutable and not updated in the update script

remcowesterhoud:
## Description

Some smaller improvements:
- We don't need the clean up all members script. Members will be deleted with individual events
- We shouldn't update the tenant id in the update query. The tenant id is the primary key of a tenant.

I was looking at tests, but these scenarios seem to be covered already. The deleteAllMembers script was untested, so no need to remove anything there. There is a test for updating the tenant id. We can't really tests if the tenantId remains unchanged as the update script matches the tenant on the id. So this scenario isn't really possible.

## Checklist

- [ ] for CI changes:

## Related issues

closes #28966

### DIFF
--- a/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/TenantWriter.java
+++ b/db/rdbms/src/main/java/io/camunda/db/rdbms/write/service/TenantWriter.java
@@ -82,13 +82,6 @@ public class TenantWriter {
             tenant.tenantId(),
             "io.camunda.db.rdbms.sql.TenantMapper.delete",
             tenant.tenantId()));
-    executionQueue.executeInQueue(
-        new QueueItem(
-            ContextType.TENANT,
-            WriteStatementType.DELETE,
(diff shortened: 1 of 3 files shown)


maryarm (approved):
Looks good to me, I'm just wondering do we need to add any assertion in existing test TenantIT#shouldSaveAndDelete to check if the members are n't in TENAT_MEMBER table?

remcowesterhoud:
> Looks good to me, I'm just wondering do we need to add any assertion in existing test TenantIT#shouldSaveAndDelete to check if the members are n't in TENAT_MEMBER table?

I don't think it fits in the specific case you mention, but I do agree these test cases are missing. I've added some.

PR #28984, opened 2025-02-28
PoC listen to the `creating` event

### LINKED ISSUE #28252: PoC listen to user task `creating` and `canceling` events
As part of https://github.com/camunda/product-hub/issues/2625 and https://github.com/camunda/product-hub/issues/2657, we want to explore how we can implement `creating` and `canceling` user task listeners.

The requirements are specified [here](https://docs.google.com/document/d/1g3g4gLSFMSbcXkiNoQBKLisDxgiAnAv2im0uRvjUI8I/edit?tab=t.0#heading=h.rupp1hvtxsyw).

korthout:
## Description

Includes:
- responding to the `creating` event
- providing access to the `assignee` in the `creating` event if user task defines an assignee in the process
- allowing `creating` listener to correct the `assignee`
  - this is useful when the user task does not define an assignee but one should be set when creating
  - note that the correction may be ignored by subsequent `assigning` listeners

## Related issues

closes #28252

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserTaskCreatingApplier.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserTaskCreatingApplier.java
@@ -8,6 +8,7 @@
 package io.camunda.zeebe.engine.state.appliers;
 
 import io.camunda.zeebe.engine.state.TypedEventApplier;
+import io.camunda.zeebe.engine.state.immutable.UserTaskState.LifecycleState;
 import io.camunda.zeebe.engine.state.instance.ElementInstance;
 import io.camunda.zeebe.engine.state.mutable.MutableElementInstanceState;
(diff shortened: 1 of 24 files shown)


zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/appliers/UserTaskCreatingApplier.java
```diff
@@ -28,7 +29,14 @@ public UserTaskCreatingApplier(final MutableProcessingState processingState) {
 
   @Override
   public void applyState(final long key, final UserTaskRecord value) {
-    userTaskState.create(value);
+    // todo: should we actually store the user task fully already, or wait until we finalize?
+    final var copy = value.copy();
+    copy.unsetAssignee();
+    userTaskState.create(copy);
```
korthout, line 35: 💭  This is necessary to ensure the UserTaskRecord still has the assignee set, which is used after creation in the task listener (to set `assignee` property in custom headers)

korthout:
Closing this, as the PoC is completed

PR #28997, opened 2025-02-28
Run HistoryCleanupIT with CamundaExporter

ChrisKujawa:
## Description

Migrating the HistoryCleanupIT, to remove the BrokerITInvocationProvider eventually.

As a pre-requisite we had to make sure that the CamundaExporter is correctly configured, such that it is able to delete data quickly. Furthermore, per default ISM/ILM is run only 5-10 minutes per default, we need to reduce this to test history cleanup reliably and quickly.

## What the PR does

 * Introducing a new annotation HistoryMultiDbTest, that extends the MultiDbTest, to mark a test to be run against multi databases and enabling history clean up. This is to not run into issues with other tests.
 * Configuring the CamundaExporter with archiver configs, including retention.
 * When HistoryMultiDbTest is detected retention is enabled
 * Configuring the ES/OS service container in CI and local test container ILM/ISM setting, such as they are executed more frequently, to clean up data earlier
 * Migrate the HistoryCleanupIT

Successful run of the test:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,6 +258,11 @@ jobs:
           # Disabling to make sure we can delete indices via wildcards
           # https://www.elastic.co/guide/en/elasticsearch/reference/current/index-management-settings.html
           action.destructive_requires_name: "false"
+          # We need to configure ILM to run more often, to make sure data is cleaned up earlier
+          # Useful for tests where we verify history clean up
+          # Default is 10m
+          # https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-settings.html
+          indices.lifecycle.poll_interval: "1s"
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-build
@@ -317,7 +322,7 @@ jobs:
     if: needs.detect-changes.outputs.java-code-changes == 'true'
     needs: [ detect-changes ]
     name: "[IT] OpenSearch"
-    timeout-minutes: 10
+    timeout-minutes: 15
     permissions: {}  # GITHUB_TOKEN unused in this job
     runs-on: gcp-perf-core-8-default
     env:
@@ -347,6 +352,14 @@ jobs:
           # We have to specify an initial password to bootstrap OpenSearch
           # https://opensearch.org/blog/replacing-default-admin-credentials/
           OPENSEARCH_INITIAL_ADMIN_PASSWORD: "yourStrongPassword123!"
+          # We need to configure ISM to run more often, to make sure data is cleaned up earlier
+          # Useful for tests where we verify history clean up
+          # Default is 5 - unit is minutes
+          # https://opensearch.org/docs/latest/im-plugin/ism/settings/
+          plugins.index_state_management.job_interval: "1"
+          # A randomized delay that is added to a job’s base run time to prevent a surge of activity from all indexes at the same time.
+          plugins.index_state_management.jitter: "0"
+          knn.algo_param.index_thread_qty: 4
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-build
@@ -370,7 +383,7 @@ jobs:
       - name: Run integration test with externalized OS
         run: >
           ./mvnw -B -T 1 --no-snapshot-updates
-          -D forkCount=1
+          -D forkCount=2
           -D maven.javadoc.skip=true
           -D skipUTs -D skipChecks
           -D failsafe.rerunFailingTestsCount=3 -D flaky.test.reportDir=failsafe-reports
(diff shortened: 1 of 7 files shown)


npepinpe:
FYI, history cleanup is a thing in RDBMS - it's not the same as the archiver. So I think it makes sense it doesn't run for the Camunda Exporter :smile:

ChrisKujawa:
@npepinpe thanks for looking into it 👍🏼 I also had a closer look the last days, and was wondering why this is actually allowed in RDBMS 😅 and thought this would need more tuning if this should work with Es exporter, regarding ILM, etc. in ES.

I will migrate in in the next step only for RDBMs, and look into the Archiver E2E separately.

ChrisKujawa:
Oh I got it working 💪🏼 It is just a matter of getting ILM configured, etc. I feel this is a good first archiving E2E test 👍🏼

ChrisKujawa:
Oh I got it working for the PI 💪🏼 It is just a matter of getting ILM configured, etc. I feel this is a good first archiving E2E test 👍🏼 

It looks like tasks are not deleted. Is this expected? Not sure here, I guess no? @npepinpe 

```
org.awaitility.core.ConditionTimeoutException: Condition with alias 'should wait until process is ended' didn't complete within 1 minutes because assertion condition defined as a Lambda expression in io.camunda.it.historycleanup.HistoryCleanupIT 
expected: 0L
 but was: 1L.
Expected :0L
Actual   :1L
```

ChrisKujawa:
Ok looks like Archiving itself works.
We have dated indices, with data:



User tasks have also right key



ILM Policy has been created



And can also trigger



All dated indices are then deleted again


**Challenge here**: It takes ~12 minutes.

ChrisKujawa:
> indices.lifecycle.poll_interval
> ([Dynamic](https://www.elastic.co/guide/en/elasticsearch/reference/current/settings.html#dynamic-cluster-setting), [time unit value](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#time-units)) How often index lifecycle management checks for indices that meet policy criteria. Defaults to 10m.

https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-settings.html

ChrisKujawa:
@cmur2 I have an question. I was working on some history clean up tests (migrating them), on ES they are quite performant, but on OS they are not.

https://github.com/camunda/camunda/actions/runs/13647000564/job/38147614578?pr=28997#step:8:905

It takes ~4 minutes to run them, this is because the ISM on Opensearch is quite, naja. We can only set a check interval of minutes, not seconds, for example. The test and pipeline is green now, and in their bounds, but I had to increase the timeout of the CI job.

Wondering whether I should completely separate this test into own CI job, with OS service container again, to parallize this, and not impact other tests. What do you think? I did that here for example: https://github.com/camunda/camunda/pull/28997/commits/0ed0b8103bb7d33d6c087ffea9373ecc8968337d

(I reverted this for testing purposes). Alternative would be to increase the CI job timeout, as I did for now.

ChrisKujawa:
@npepinpe any opinion on the above?

npepinpe:
Not keen on having more and more jobs which only slightly differ from each other, as that sounds like it'll lead to confusion in the long run.

I get where you're coming from though :thinking: Setting e.g. `plugins.index_state_management.job_interval: 30s` really fails? Or does it just round up to 1m? That's kind of weird choice on their part regardless :shrug:

npepinpe (approved):
I tried a couple things, but I can't get ISM to reliably run :(

ES ILM for example also lets you force moving to the next phase in the lifecycle, but ISM doesn't seem to. So I don't know what else we can do. Do you think modifying the system clock would cause it to work? I've done this before with very simple extensions (e.g. you inject a library via LD_PRELOAD which gives you a very simple API to change the clock).

npepinpe:
Modifying the system clock: https://github.com/camunda-community-hub/zeebe-test-container/pull/217

That was how I initially attempted it for zeebe-test-container :shrug:

cmur2:
> @cmur2 I have an question. I was working on some history clean up tests (migrating them), on ES they are quite performant, but on OS they are not.
> 
> https://github.com/camunda/camunda/actions/runs/13647000564/job/38147614578?pr=28997#step:8:905

The ES job in there takes 9min21s and the OS job takes 10m25s so not a huge difference?

> It takes ~4 minutes to run them, this is because the ISM on Opensearch is quite, naja. We can only set a check interval of minutes, not seconds, for example. The test and pipeline is green now, and in their bounds, but I had to increase the timeout of the CI job.

In the code I see an increase from 10min to 25min but as noted above the job barely needs above 10mins, so why not go for 15mins timeout? 15mins is without the bounds of the Unified CI goals so that should be fine, 25mins not. If the 25mins are really needed I'd go for splitting the job (e.g. with a matrix build to avoid code repetition).

ChrisKujawa:
Thanks @npepinpe @cmur2 

I will go with the higher timeout for now, let us see whether it is an issue. I have set it to 15 minutes, as you mentioned @cmur2

PR #29001, opened 2025-02-28
PoC: Migrate from Job-based user task to Camunda User Task

berkaycanbc:
#### Summary

During the first iteration, I was able to migrate from Job Based to Camunda User Task successfully. While doing so, I was able to map related fields between different types. In addition, user task assignment task listener is also triggered if previous user task has an assignee.

Job based user tasks can be migrated to Camunda user tasks in following steps:
- Map job based user task custom headers to user task properties
- Create new user task using mapped task properties and task headers (non user task related headers)
- Assign user task if any assignee exist
- Create task listener job listening user task assignments
- Cancel job based user task job

#### Remarks

- For job based user tasks, `formKey` is used to reference either internal form or an external form. To determine if the `formKey` is external or internal, a check against `formId` is applied. If `formId` is present in the job worker properties, we update `formKey` in Camunda User Task, if not we update `externalFormReference`.

- If the target Camunda User Task listens for assignment events and Job Worker User Task has an assignee, a task listener job will be created during the migration.

- `priority` field is not part of a Job-based User Task. It will be set to default value `50` during the migration.

#### Next
I will verify if task listener job can successfully be completed after the migration, implement embedded form handling and add more test cases to verify correct mapping for different user task properties.

### DIFF
--- a/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
+++ b/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/bpmn/behavior/BpmnUserTaskBehavior.java
@@ -35,7 +35,9 @@ import java.time.ZonedDateTime;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import org.agrona.DirectBuffer;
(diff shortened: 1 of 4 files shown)


berkaycanbc:
@aleksander-dytko, @eppdot I completed first iteration of PoC. Please take a look at my findings in the description and let me know if you have any questions.

PR #29003, opened 2025-02-28
Remove `entityKey` from `TenantRecord`

### LINKED ISSUE #28831: Remove `entityKey` from `TenantRecord`
## Description

The `TenantRecord` currently contains a property called `entityKey`. Since we are switching from keys to ids we no longer need this property. We should remove it from the record and refactor any usages of this property to use ids instead.

------

Blocked by #27137

remcowesterhoud:
## Description

Tenants are now id-based, and so will all the entities that can get assigned to a tenant. We no longer have a need for the `entityKey` and should remove it. The property is used in 3 scenarios:

- Assigning/Unassigning a user. This is simple and mostly worked already since the user migration from key to username is already completed.
- Assigning/unassigning a group. Groups don't have an id yet. Because of this we use the key for now. By parsing the key to a String we can pretend it's an id. This parsing needs to be replaced when we refactor Groups to work with ids.
- Assigning/unassigning a mapping. Mappings do have an id. Because of this we can just use the id. It did require some
fixes in code that weren't done before.

## Checklist

- [ ] for CI changes:

## Related issues

closes #28831

### DIFF
--- a/qa/integration-tests/src/test/java/io/camunda/it/rdbms/exporter/RecordFixtures.java
+++ b/qa/integration-tests/src/test/java/io/camunda/it/rdbms/exporter/RecordFixtures.java
@@ -255,14 +255,6 @@ public class RecordFixtures {
 
   protected static ImmutableRecord<RecordValue> getTenantRecord(
       final Long tenantKey, final String tenantId, final TenantIntent intent) {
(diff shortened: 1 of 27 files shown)


koevskinikola (approved):
LGTM @remcowesterhoud 🎉

Sorry for the review delay, too many release tasks recently 😄

PR #29005, opened 2025-02-28
ci: include Tasklist docker tests in the common CI

houssain-barouni:
## Description

This PR addresses a recent issue where the Tasklist application failed to start due to changes that were merged without being blocked by the Tasklist Docker tests workflow, as it is not a required check.

To avoid that, we are integrating Tasklist Docker tests into the common CI. This includes:
* Building the full project (excluding the frontend and Optimize)
* Building the Tasklist Docker image and pushing it to a local registry
* Running a Docker test that starts a Tasklist container and performs basic sanity checks

The job takes around 8-9 minutes in average, I set the timeout to 15 minutes

This will replace the existing `build-tasklist-backend` job in the common CI, as it contains similar steps.

## Checklist

- [x] for CI changes:
  - [x] enable backports [when recommended](https://github.com/camunda/camunda/wiki/CI-&-Automation#when-to-backport-ci-changes)

## Related issues

closes #

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -747,45 +747,11 @@ jobs:
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
 
-  build-tasklist-backend:
+  tasklist-backend-tests:
     if: needs.detect-changes.outputs.tasklist-backend-changes == 'true'
-    needs: [detect-changes]
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
-    permissions: {}  # GITHUB_TOKEN unused in this job
-    env:
-      BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
-      - uses: ./.github/actions/setup-build
-        name: Build setup
-        with:
-          java-distribution: adopt
-          maven-cache-key-modifier: tasklist
-          maven-version: 3.8.6
-          vault-address: ${{ secrets.VAULT_ADDR }}
-          vault-role-id: ${{ secrets.VAULT_ROLE_ID }}
-          vault-secret-id: ${{ secrets.VAULT_SECRET_ID }}
-      - name: Build Maven
-        run: |
-          ./mvnw clean deploy -B -T1C -PskipFrontendBuild -DskipTests -DskipChecks -DaltStagingDirectory=${{ github.workspace }}/staging-${{ env.BRANCH_NAME }} -DskipRemoteStaging=true -Dmaven.deploy.skip=true
-      # Build the docker image to verify the built artifacts and the Dockerfile
-      # It does NOT push the image to the registry
-      - name: Build Docker image
-        uses: ./.github/actions/build-platform-docker
-        with:
-          dockerfile: tasklist.Dockerfile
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -747,45 +747,11 @@ jobs:
           secret_vault_address: ${{ secrets.VAULT_ADDR }}
           secret_vault_roleId: ${{ secrets.VAULT_ROLE_ID }}
 
-  build-tasklist-backend:
+  tasklist-backend-tests:
```
ChrisKujawa, line 750: ❓ Should it be called 
```suggestion
  tasklist-docker-tests:
```

ChrisKujawa (approved):
👍🏼

.github/workflows/ci.yml
houssain-barouni: in the beginning, I named like that, then I decided to rename it as it can be extended to other tasklist backend tests that we can include in the common CI in the future

cmur2:
I will have a look but might not make it today since I'm feeling sick-ish.

> enable backports [when recommended](https://github.com/camunda/camunda/wiki/CI-&-Automation#when-to-backport-ci-changes)

Do you think backporting to stable 8.6 and 8.7 makes sense for this? Those branches also have the Unified CI and no required Tasklist status check.

houssain-barouni:
> I will have a look but might not make it today since I'm feeling sick-ish.

There is no urgency, Get Well Soon!


> Do you think backporting to stable 8.6 and 8.7 makes sense for this? Those branches also have the Unified CI and no required Tasklist status check.

yes, it makes sense to backport it.

.github/workflows/ci.yml
ChrisKujawa: Yeah i know but right now it isnt so i would reflect the current state

cmur2: I'm fine with either name but I'd like it to be consistent with the name of the called workflow, which is currently `tasklist-docker-tests.yml`.

> as it can be extended to other tasklist backend tests that we can include in the common CI in the future

🔧 If you want to have that as a place for future extension, I think it makes sense to have a more generic name for that workflow file.

> would reflect the current state

💭 Could leave a comment in the `ci.yml` indicating the current state + future aspiration aka `# Currently only runs Docker tests but should be extended in future to also include backend tests`

.github/workflows/tasklist-docker-tests.yml
```diff
@@ -1,111 +1,50 @@
...
-      - 'stable/**'
-      - 'release**'
-  pull_request:
-
-# Limit workflow to 1 concurrent run per ref (branch): new commit -> old runs are canceled to save costs
-# Exception for main branch: complete builds for every commit needed for confidenence
-concurrency:
-  cancel-in-progress: true
-  group: ${{ format('{0}-{1}', github.workflow, github.ref == 'refs/heads/main' && github.sha || github.ref) }}
+  workflow_dispatch: { }
```
cmur2, line 4: 💭 Is the capability to run this workflow manually needed? Otherwise `workflow_call` is enough for using it in Unified CI.

cmur2 (approved):
Thanks for this contribution!

.github/workflows/tasklist-docker-tests.yml
houssain-barouni: I made it consistent with `zeebe-ci.yml`, I think there is not harm of making the flow runnable with a manual action

.github/workflows/ci.yml
houssain-barouni: I renamed the job to `tasklist-docker-tests` and added a comment to tell that we need to extends it with more IT tests

PR #29030, opened 2025-03-02
fix: REST API Exceptions improvements

### LINKED ISSUE #26868: `CamundaSearchException` causes are logged as unexpected errors
As Zeebe medic I was checking an [error report](https://console.cloud.google.com/logs/query;cursorTimestamp=2025-01-13T21:18:26.471219775Z;endTime=2025-01-13T21:48:56.471Z;query=error_groups.id%3D%22CNvhsrmCq9a1-QE%22%0AlogName:%22stdout%22%0Aresource.type%3D%22k8s_container%22%0Aresource.labels.container_name%3D%22zeebe%22%0Aresource.labels.namespace_name%3D%22498c3fff-85be-4af1-87ab-4f81814430e0-zeebe%22%0Aresource.labels.pod_name%3D%22zeebe-2%22%0Aresource.labels.cluster_name%3D%22prod-worker-3%22%0Aresource.labels.project_id%3D%22camunda-cloud-240911%22%0Aresource.labels.location%3D%22us-central1%22%0Atimestamp%3D%222025-01-13T21:18:26.471219775Z%22%0AinsertId%3D%22dj9cijgdldgzycsv%22;startTime=2025-01-13T20:48:56.471Z?project=camunda-cloud-240911):
```
org.apache.http.ConnectionClosedException: Connection is closed
	at org.elasticsearch.client.RestClient.extractAndWrapCause(RestClient.java:924) ~[elasticsearch-rest-client-8.13.4.jar:8.13.4]
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:304) ~[elasticsearch-rest-client-8.13.4.jar:8.13.4]
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:292) ~[elasticsearch-rest-client-8.13.4.jar:8.13.4]
	at co.elastic.clients.transport.rest_client.RestClientHttpClient.performRequest(RestClientHttpClient.java:91) ~[elasticsearch-java-8.13.4.jar:?]
	at co.elastic.clients.transport.ElasticsearchTransportBase.performRequest(ElasticsearchTransportBase.java:144) ~[elasticsearch-java-8.13.4.jar:?]
	at co.elastic.clients.elasticsearch.ElasticsearchClient.search(ElasticsearchClient.java:1923) ~[elasticsearch-java-8.13.4.jar:?]
	at io.camunda.search.es.clients.ElasticsearchSearchClient.search(ElasticsearchSearchClient.java:55) ~[camunda-search-client-elasticsearch-8.7.0-alpha2.jar:8.7.0-alpha2]
[...]
```
At first I was super confused because `ElasticsearchSearchClient#search` actually catches `IOException`  and so I would have expected a  `CamundaSearchException` with `org.apache.http.ConnectionClosedException` as cause. Then I found out that the `RestErrorMapper` handles `CamundaSearchException` by first logging a debug message (that makes sense!) but then further tries to unwrap the cause. Since we don't have special handling of  `ConnectionClosedException` or `IOException`, it get's logged as an unexpected error.

In my opinion, this unwrapping does not make much sense. If we accept `CamundaSearchException` as an _expected_ exception, we should not unwrap it further and eventually log the cause as an unexpected exception.

This was reported before actually: https://github.com/camunda/camunda/issues/26050

Error groups:
- https://console.cloud.google.com/errors/detail/CP_nhvvHr-H3ag;service=zeebe;time=P7D;locations=global?inv=1&invt=Abm0IA&project=camunda-cloud-240911
- https://console.cloud.google.com/errors/detail/CNvhsrmCq9a1-QE;locations=global?project=camunda-cloud-240911&inv=1&invt=Abm0cQ
- https://console.cloud.google.com/errors/detail/CNvhsrmCq9a1-QE;service=zeebe;time=P7D;locations=global?inv=1&invt=Abm0cQ&project=camunda-saas-prod
- https://console.cloud.google.com/errors/detail/CKGqmP6S67bzbw;service=zeebe;time=P7D;locations=global?inv=1&invt=Abm_wg&project=camunda-saas-prod-sin

### LINKED ISSUE #28181: [REST API] `ConnectionException` error mapping is not specific enough
**Description**

There is at least 2 ways the REST API can run into a `ConnectionException`:
1. The gateway cannot reach a Broker
2. The gateway cannot reach Elasticsearch

At the moment both of these scenarios result in the message:

> 14:07:49.597 [] [http-nio-0.0.0.0-8080-exec-4] [] WARN  io.camunda.zeebe.gateway.rest - Expected to handle REST API request, but there was a connection error with one of the brokers

This is misleading when the problem is Elasticsearch. If that's the response mentioned the Broker is unreachable, which is false.

The problem lies in the `RestErrorMapper`. [Here we map](https://github.com/camunda/camunda/blob/3360da078bfd20a30658f56f5490fa8ffa5b0792/zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java) the `ConnectionException` to error message saying the Broker is unreachable. We should aim to identify if it's actually the Broker, or if it's really Elasticsearch, and create a message accordingly.

ghost:
## Description

fix: REST API Exceptions improvements

- replaced derived classes of CamundaSearchException whit enum states
- actual exceptions are logged in situ, where they happen
- added custom ProblemDetail, based on the type of CamundaSearchException encountered

Closes https://github.com/camunda/camunda/issues/26868
Closes https://github.com/camunda/camunda/issues/28181


search/search-domain/src/main/java/io/camunda/util/ExceptionUtil.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH under
+ * one or more contributor license agreements. See the NOTICE file distributed
+ * with this work for additional information regarding copyright ownership.
+ * Licensed under the Camunda License 1.0. You may not use this file
+ * except in compliance with the Camunda License 1.0.
+ */
+package io.camunda.util;
```
tmetzke, line 8: 🔧 Let's move this to `io.camunda.search.exception`

search/search-domain/src/main/java/io/camunda/util/ExceptionUtil.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class ExceptionUtil {
```
tmetzke, line 10: 🔧 We only collect error messages here, so let's maybe name this accordingly, like `ErrorMessages`. We have a similar thing in the REST gateway.

tmetzke (changes requested):
👍 This is really cool, nice work, @mihail-ca 🚀 
⭐ This makes the exception handling a lot easier to understand, I believe. And it's nicely extendable, too 👍 

🔧 I added some suggestions to simplify the new structure even further and to move the responsibilities closer to the origin of the exceptions.

Overall, I really like this approach 👍

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,53 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
       final DocumentException e) {
     return mapProblemToResponse(mapDocumentHandlingExceptionToProblem(e));
   }
+
+  public static ProblemDetail mapCamundaSearchExceptionToProblem(
+      CamundaSearchException cse, final Function<BrokerRejection, ProblemDetail> rejectionMapper) {
```
tmetzke, line 324: 🔧 The `rejectionMapper` is unused now, let's remove it from this method.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,53 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+    final CamundaSearchException.Reason reason =
+        (cse.getReason() == null) ? CamundaSearchException.Reason.UNKNOWN : cse.getReason();
+    final Throwable cause = (cse.getCause() == null) ? cse : cse.getCause();
+    final String name = cause.getClass().getName();
```
tmetzke, line 328: 🔧 We should only need the name in case of an UNKNOWN reason. Otherwise, the reason should be the `title` in the problem detail.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,53 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+    final String message = cause.getMessage();
```
tmetzke, line 329: 🔧 We should only need the cause's message in case of an UNKNOWN reason. Otherwise, the message should be the one from the `CamundaSearchException`. In the logger messages, we can go for the "Expected to handle ..., but ..." texts. Those should NOT go into the user-facing problem details though. They already know they sent a REST request. But in the logs it's helpful.

tmetzke (commented):
👍 Really nice improvements.
🔧 I added some last suggestions for the finishing touch, should be good to merge afterward.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+
+  public static ProblemDetail mapCamundaSearchExceptionToProblem(CamundaSearchException cse) {
+    final CamundaSearchException.Reason reason =
+        (cse.getReason() == null) ? CamundaSearchException.Reason.UNKNOWN : cse.getReason();
+    final Throwable cause = (cse.getCause() == null) ? cse : cse.getCause();
+    final String title =
+        (reason == CamundaSearchException.Reason.UNKNOWN)
+            ? cause.getClass().getName()
+            : reason.name();
+    final String errorMessage = cause.getMessage();
```
tmetzke, line 331: 🔧 I think we can simplify the prolog of this method a bit.
* The `CamundaSearchException` should always have a reason. Let's enforce this in the exception itself by setting to UNKNOWN if `null` was provided. We can remove the check here.
* The reason should always be the `title`.
* The exception message should be what the search exception brings along. We can add information if needed in the switch cases.

```suggestion
    final CamundaSearchException.Reason reason = cse.getReason();
    final String title = reason.name();
    final String errorMessage = cse.getMessage();
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+    final String logPrefix = "Expected to handle REST request, but error occured: ";
```
tmetzke, line 332: 🔧 The prefix can be a bit more generic, since it's not always an error. Plus, the placeholder can be embedded already and can be removed in all switch cases.
```suggestion
    final String logPrefix = "Expected to handle REST request, but: {}";
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+    switch (reason) {
+      case NOT_FOUND:
+        {
+          REST_GATEWAY_LOGGER.debug(logPrefix + "{}", errorMessage);
```
tmetzke, line 336: 🔧 The `{}` can be removed in all switch cases.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+          return createProblemDetail(
+              HttpStatus.NOT_FOUND, errorMessage, RejectionType.NOT_FOUND.name());
```
tmetzke, line 338:
```suggestion
              HttpStatus.NOT_FOUND, errorMessage, title);
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+        }
+      case NOT_UNIQUE:
+        {
+          REST_GATEWAY_LOGGER.debug(logPrefix + "{}", errorMessage);
+          return createProblemDetail(
+              HttpStatus.CONFLICT, errorMessage, RejectionType.INVALID_STATE.name());
+        }
+      case CONNECTION_FAILED:
+        {
+          final String detail = "The search client could to connect to the search server.";
```
tmetzke, line 348: 🔧 Let's keep the original message and add information.
```suggestion
          final String detail = errorMessage + ". The search client could not connect to the search server.";
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+          REST_GATEWAY_LOGGER.debug(logPrefix + "{}", detail);
+          return createProblemDetail(HttpStatus.SERVICE_UNAVAILABLE, detail, title);
+        }
+      case SEARCH_SERVER_FAILED:
+        {
+          final String detail = "The search server was unable to process the request.";
```
tmetzke, line 354: 🔧 Let's keep the original message and add information.
```suggestion
          final String detail = errorMessage + ". The search server was unable to process the request.";
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+          REST_GATEWAY_LOGGER.debug(logPrefix + "{}", detail);
+          return createProblemDetail(HttpStatus.INTERNAL_SERVER_ERROR, detail, title);
+        }
+      case SEARCH_CLIENT_FAILED:
+        {
+          final String detail = "The search client was unable to process the request.";
```
tmetzke, line 360: 🔧 Let's keep the original message and add information.
```suggestion
          final String detail = errorMessage + ". The search client was unable to process the request.";
```

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +319,55 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
...
+          REST_GATEWAY_LOGGER.debug(logPrefix + "{}", detail);
+          return createProblemDetail(HttpStatus.INTERNAL_SERVER_ERROR, detail, title);
+        }
+      default:
+        {
+          final String detail = "Unexpected error: %s".formatted(errorMessage);
+          REST_GATEWAY_LOGGER.debug(logPrefix + "{}", detail);
+          return createProblemDetail(
+              HttpStatus.INTERNAL_SERVER_ERROR, detail.formatted(errorMessage), title);
```
tmetzke, line 369: 🔧 The error is not unexpected in that sense; we caught it explicitly. An unexpected error is usually uncaught.
```suggestion
          REST_GATEWAY_LOGGER.debug(logPrefix + "{}", errorMessage);
          return createProblemDetail(
              HttpStatus.INTERNAL_SERVER_ERROR, errorMessage, title);
```

tmetzke (commented):
👍 Changes look good; I left some simplification suggestions.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
```diff
@@ -327,4 +318,51 @@ public static ResponseEntity<Object> mapDocumentHandlingExceptionToResponse(
       final DocumentException e) {
     return mapProblemToResponse(mapDocumentHandlingExceptionToProblem(e));
   }
+
+  public static ProblemDetail mapCamundaSearchExceptionToProblem(CamundaSearchException cse) {
+    final CamundaSearchException.Reason reason = cse.getReason();
+    final Throwable cause = (cse.getCause() == null) ? cse : cse.getCause();
```
tmetzke, line 324: 🔧 I think this `cause` is not used anymore, right?

tmetzke (approved):
👍 Nice, changes look good to me, nice job ⭐ 
🔧 ~We can remove an unused variable, I believe, good to go afterward.~ Fixed it already

ℹ️ For transparency, why is this also fixing https://github.com/camunda/camunda/issues/28181? Because we don't unpack search exceptions anymore, so search-related `ConnectExceptions` will be handled separately and correctly flagged as search connection issues.

zeebe/gateway-rest/src/main/java/io/camunda/zeebe/gateway/rest/RestErrorMapper.java
tmetzke:
```suggestion
```

mihail-ca:
> 👍 Nice, changes look good to me, nice job ⭐ 🔧 ~We can remove an unused variable, I believe, good to go afterward.~ Fixed it already
> 
> ℹ️ For transparency, why is this also fixing #28181? Because we don't unpack search exceptions anymore, so search-related `ConnectExceptions` will be handled separately and correctly flagged as search connection issues.

It's fixing because now we return the error CONNECTION_FAILED when ES is down, i.e. ES client cannot connect to server. Initially it was something about brokers.

PR #29038, opened 2025-03-03
refactor: remove C7 serialization config from Optimize C8

abremard:
## Description

This is part of the "C7 config cleanup from C8". Because there's a lot of configs that need cleaning, I'm breaking this work down into multiple PRs. This one is focused on "serialization config" cleanup. For these updates the corresponding section in [C8 docs](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/system-configuration/) have been updated in [this PR](https://github.com/camunda/camunda-docs/pull/5089)

## Related issues

related to #13375

### DIFF
--- a/optimize/backend/src/test/java/io/camunda/optimize/service/util/ObjectMapperFactoryTest.java
+++ b/optimize/backend/src/test/java/io/camunda/optimize/service/util/ObjectMapperFactoryTest.java
@@ -9,7 +9,6 @@ package io.camunda.optimize.service.util;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.optimize.dto.optimize.ReportType;
 import io.camunda.optimize.dto.optimize.query.report.ReportDefinitionDto;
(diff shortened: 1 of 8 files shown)


optimize/backend/src/test/java/io/camunda/optimize/service/util/ObjectMapperFactoryTest.java
```diff
@@ -176,7 +165,7 @@ public void testOptimizeMapperDateSerialization() throws Exception {
 
     // when
     final DateHolder instance = new DateHolder();
-    instance.setDate(OffsetDateTime.parse(dateString, createEngineDateFormatter()));
```
RomanJRW, line 179: was this using the wrong mapper before in the test?

optimize/backend/src/test/java/io/camunda/optimize/service/util/ObjectMapperFactoryTest.java
```diff
@@ -188,7 +177,7 @@ public void testOptimizeMapperDateDeserialization() throws Exception {
     // given
     final String dateString = "2017-12-11T17:28:38.222+0100";
     final OffsetDateTime expectedOffsetDateTime =
-        OffsetDateTime.parse(dateString, createEngineDateFormatter());
+        OffsetDateTime.parse(dateString, createOptimizeDateFormatter());
```
RomanJRW, line 180: same as above

RomanJRW (approved):

optimize/backend/src/test/java/io/camunda/optimize/service/util/ObjectMapperFactoryTest.java
abremard: Yes, I should have clarified this. I believe it was copy-pasted issue that slipped from the engine mapper tests below (the naming is the same with the exception of OptimizeMapper vs EngineMapper).

PR #29058, opened 2025-03-03
Identity http authorization tests

sergiizakharov:
## Description

Add tests for the authorization api.

## Checklist

- [ ] http tests

## Related issues

closes #


megglos (approved):
@sergiizakharov please do a test run from this branch, this would help to be confident that the tests are working as intended

you should be able to do that via https://github.com/camunda/camunda/actions/workflows/identity-regression-test.yml selecting your branch as reference.

Pre-approving, assuming you do that and eventually address potential issues.

